PR #3456, opened 2025-11-17
Add Web Chat UI for any agent that can be launched using `clai web` or `Agent.to_web()`

dsfaccini:
web-based chat interface for Pydantic AI agents

1. new module `pydantic_ai.ui.web`
2. new method `Agent.to_web()`

### fastapi
- `app = create_chat_app(agent)`

- the following endpoints come preconfigured:
  - `GET /` and `/:id` - serve the chat UI
  - `POST /api/chat` - Main chat endpoint using VercelAIAdapter
  - `GET /api/configure` - Returns available models and builtin tools
  - `GET /api/health` - Health check
  - NOTE: I'm counting on FastAPI to complain if the user tried adding conflicting routes, otherwise we could add a warning on the respective docs.

### options and example

NOTE: the module for options is currently `pydantic_ai.ui.web.`
- pre-configured model options:
  - `anthropic:claude-sonnet-4-5`
  - `openai-responses:gpt-5`
  - `google-gla:gemini-2.5-pro`

- supported builtin tools:
  - `web_search`
  - `code_execution`
  - `image_generation`

```python
# app.py
import logfire
from pydantic_ai import Agent

logfire.configure(send_to_logfire='if-token-present')
logfire.instrument_pydantic_ai()

agent = Agent('openai:gpt-5')

@agent.tool
def get_weather(city: str) -> str:
    return f"The weather in {city} is sunny"

app = agent.to_web()

logfire.instrument_fastapi(app, capture_headers=True)

# Run with: uvicorn app:app
```

## testing

- 7 tests in `tests/test_ui_web.py`

## notes

- UI is served from CDN: `@pydantic/ai-chat-ui@0.0.2`
- Uses Vercel AI protocol for chat streaming
- [x] TODO: add `clai web` command to launch from the CLI (as in `uvx pydantic-work` without the whole URL magic)
- [x] TODO: should I add a new doc at `docs/ui/to_web.md`? I'd also reference this in `docs/ui/overview.md` and `docs/agents.md`

EDIT: if you try it out it's worth noting that the current hosted UI doesn't handle `ErrorChunks`, so you will get no spinner and no response when there's a model-level error and fastapi will return a 200 any way.
This will happen for instance when you use a model for which you don't have a valid API key in your environment
I opened a PR for the error chunks here https://github.com/pydantic/ai-chat-ui/pull/4.

Closes https://github.com/pydantic/pydantic-ai/issues/3295

PR #3412, opened 2025-11-13
Add support for aspect ratio in gemini image generation

mwildehahn:
Fix for https://github.com/pydantic/pydantic-ai/issues/3119

PR #3599, opened 2025-11-30
Add OpenRouter image generation support

### LINKED ISSUE #3258: can't generate image using openrouter
### Question

Hi, I have openrouter account and trying to use the nano banana model via openrouter with pydantic_ai, the problem is that the model response is str and the image that is being generated by model is not in the response object. Can someone Help me? 
sample code:
```
from pydantic_ai import Agent, BinaryImage
from pydantic_ai.models.openai import OpenAIModel

def generate_image(prompt: str, api_key: str):
    """Generate image using pydantic_ai with OpenRouter."""
    
    # Configure model
    provider = OpenRouterProvider(api_key=api_key)
    model = OpenAIModel(
        model_name="google/gemini-2.5-flash-image-preview",  # Doesn't exist on OpenRouter
        provider=provider,
    )
    
    # Create agent expecting BinaryImage
    agent = Agent(model=model, output_type=BinaryImage)
    
    # This fails: model returns text instead of image data
    result = agent.run_sync(prompt)  # Error: "Exceeded maximum retries for output validation"
    
    return result.output.data

if __name__ == "__main__":
    try:
        image_data = generate_image("A cat", "your_api_key")
        print("Success!")
    except Exception as e:
        print(f"Error: {e}")
```

### Additional Context

_No response_

ajac-zero:
Closes #3258

This PR enables using OpenRouter for image generation.

It also adds a couple of new hooks to `OpenAIChatModel._process_response ` to make it easier for subclasses to implement their own logic for `FilePart` responses, which  `OpenAIChatModel` doesn't support.


docs/models/openrouter.md
```diff
@@ -73,3 +73,58 @@ model = OpenRouterModel('openai/gpt-5')
 agent = Agent(model, model_settings=settings)
 ...
 ```
+
+## Image Generation
```
DouweM: We should also make this work with `builtin_tools=[ImageGenerationTool()]` and document it here: https://ai.pydantic.dev/builtin-tools/#image-generation-tool

As with Google, which doesn't expose that as a tool, using that tool or `BinaryImage` in `output_type` should automatically enable the modality.

docs/models/openrouter.md
```diff
@@ -73,3 +73,58 @@ model = OpenRouterModel('openai/gpt-5')
...
+```
+
+You can further customize image generation using `openrouter_image_config`:
+
+```python
+from pydantic_ai.models.openrouter import OpenRouterModelSettings
+
+settings = OpenRouterModelSettings(
+    openrouter_modalities=['image', 'text'],
+    openrouter_image_config={'aspect_ratio': '3:2'}
```
DouweM: I want this to be an option on `ImageGenerationTool` as in https://github.com/pydantic/pydantic-ai/pull/3412.

DouweM: I'm OK with it _also_ being a model setting if it supports more keys than `ImageGenerationTool` does currently (or really: after that PR), but if this is the only field supported so far I'd prefer to leave that and the modality settings off.

If you want, you can finish that PR as we're at it to make your life here easier.

tests/models/test_openrouter.py
```diff
@@ -406,3 +407,40 @@ class FindEducationContentFilters(BaseModel):
...
+    )
+    settings = OpenRouterModelSettings(
+        openrouter_modalities=['image', 'text'], openrouter_image_config={'aspect_ratio': '3:2'}
+    )
+
+    agent = Agent(model=model, output_type=str | BinaryImage, model_settings=settings)
+
+    async with agent.run_stream('A dog') as result:
+        async for output in result.stream_output():
+            if isinstance(output, str):
```
DouweM: This means we may never actually get to the image assertions!

docs/models/openrouter.md
ajac-zero: Yep! #3672

DouweM:
@ajac-zero https://github.com/pydantic/pydantic-ai/pull/3672 was merged! Can you please rebase and update? I'll do another round of review after that.

docs/models/openrouter.md
```diff
@@ -73,3 +73,49 @@ model = OpenRouterModel('openai/gpt-5')
 agent = Agent(model, model_settings=settings)
 ...
 ```
+
+## Image Generation
```
DouweM: I don't think we need this section here; I'd rather have this feature documented only in ai.pydantic.dev/builtin-tools#image-generation-tool. Can you update the table etc there with OpenRouter support? Also in the `ImageGenerationTool` docstring

pydantic_ai_slim/pydantic_ai/models/openrouter.py
```diff
@@ -510,6 +530,18 @@ def _openrouter_settings_to_openai_settings(model_settings: OpenRouterModelSetti
     if usage := model_settings.pop('openrouter_usage', None):
         extra_body['usage'] = usage
 
+    for builtin_tool in model_request_parameters.builtin_tools:
+        if isinstance(builtin_tool, ImageGenerationTool):  # pragma: lax no cover
```
DouweM: Why the `lax no cover`?

pydantic_ai_slim/pydantic_ai/models/openrouter.py
```diff
@@ -510,6 +530,18 @@ def _openrouter_settings_to_openai_settings(model_settings: OpenRouterModelSetti
...
+            extra_body['modalities'] = ['text', 'image']
+
+            image_config: dict[str, str] = {}
+            if aspect_ratio := builtin_tool.aspect_ratio:
+                image_config['aspect_ratio'] = aspect_ratio
+            extra_body['image_config'] = image_config
+
+    if isinstance(model_request_parameters.output_object, BinaryImage):  # pragma: lax no cover
```
DouweM: We should check `model_request_parameters.allow_image_output` instead

tests/models/test_openrouter.py
```diff
@@ -409,6 +411,70 @@ class FindEducationContentFilters(BaseModel):
     )
 
 
+async def test_openrouter_image_generation(allow_model_requests: None, openrouter_api_key: str) -> None:
+    provider = OpenRouterProvider(api_key=openrouter_api_key)
+    model = OpenRouterModel(
+        model_name='google/gemini-2.5-flash-image-preview',
+        provider=provider,
+    )
+    agent = Agent(model=model, output_type=str | BinaryImage)
```
DouweM, line 653: We should also test a scenario with `ImageGenerationTool` but no `output_type=BinaryImage`, where we read the files off of `response.files`, like here:

https://github.com/pydantic/pydantic-ai/blob/d2b08ad12e103e81bd0a9dc0eb7f2a7c3472931d/tests/models/test_openai_responses.py#L6026-L6040

DouweM:
@ajac-zero FYI https://github.com/pydantic/pydantic-ai/pull/3456 just got merged which will require you to update the `supported_builtin_tools` method on `OpenRouterModel`

ajac-zero:
@DouweM pls take a look when you get a chance!

pydantic_ai_slim/pydantic_ai/models/openrouter.py
ajac-zero: If an invalid builtin tool (which right now is anything other than `ImageGenerationTool`) is passed. The model will raise an error before getting to this point because it is not in the `supported_builtin_tools`. So here the builtin_tool can really only be of type `ImageGenerationTool`.

docs/builtin-tools.md
```diff
@@ -245,6 +245,7 @@ The [`ImageGenerationTool`][pydantic_ai.builtin_tools.ImageGenerationTool] enabl
 |----------|-----------|-------|
 | OpenAI Responses | ✅ | Full feature support. Only supported by models newer than `gpt-5`. Metadata about the generated image, like the [`revised_prompt`](https://platform.openai.com/docs/guides/tools-image-generation#revised-prompt) sent to the underlying image model, is available on the [`BuiltinToolReturnPart`][pydantic_ai.messages.BuiltinToolReturnPart] that's available via [`ModelResponse.builtin_tool_calls`][pydantic_ai.messages.ModelResponse.builtin_tool_calls]. |
 | Google | ✅ | Limited parameter support. Only supported by [image generation models](https://ai.google.dev/gemini-api/docs/image-generation) like `gemini-2.5-flash-image` and `gemini-3-pro-image-preview`. These models do not support [function tools](tools.md) and will always have the option of generating images, even if this built-in tool is not explicitly specified. |
+| OpenRouter | ✅ | Full feature support. Only supported by models models that have "image" in their output_modalities. Visit the [Models](https://openrouter.ai/models) page and filter by output modalities to find models capable of image generation. |
```
DouweM:
```suggestion
| OpenRouter | ✅ | Limited parameter support. Only supported by models that support the "image" output modality as indicated on <https://openrouter.ai/models>. |
```

DouweM: We should also update the table below of which options are supported by which model

pydantic_ai_slim/pydantic_ai/models/openrouter.py
```diff
@@ -591,6 +653,17 @@ def _map_response_thinking_part(self, item: ThinkingPart) -> None:
...
                 super()._map_response_thinking_part(item)
 
+        @override
+        def _map_response_file_part(self, item: FilePart) -> None:
+            if item.content.media_type in (
+                'image/png',
+                'image/jpeg',
+                'image/webp',
+                'image/gif',
+            ):  # pragma: lax no cover
```
DouweM: It seems unexpected to silently ignore other image files

pydantic_ai_slim/pydantic_ai/models/openrouter.py
DouweM: "lax no cover" doesn't seem appropriate here; it's for lines that are run on CI, but not locally. Any line covered by running the test suite locally shouldn't need a `pragma` at all.

I think this should have a `pragma: no branch` instead, because as you said elsewhere, if there's a builtin tool here, it's always `ImageGenerationTool`

pydantic_ai_slim/pydantic_ai/models/openrouter.py
```diff
@@ -575,11 +628,20 @@ def _process_provider_details(self, response: chat.ChatCompletion) -> dict[str,
...
 
         def _into_message_param(self) -> chat.ChatCompletionAssistantMessageParam:
             message_param = super()._into_message_param()
             if self.reasoning_details:
                 message_param['reasoning_details'] = self.reasoning_details  # type: ignore[reportGeneralTypeIssues]
+            if self.file_inputs:
+                content = message_param.get('content')
+                if isinstance(content, str):  # pragma: lax no cover
+                    message_param['content'] = [{'type': 'text', 'text': content}] + self.file_inputs  # type: ignore[reportGeneralTypeIssues]
+                elif isinstance(content, list):  # pragma: lax no cover
```
DouweM: All of these "lax no cover"s seem wrong; can you see what error we get if we remove them? In this case, I think the str and list cases should all have tests.

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/models/openrouter.py
ajac-zero: Added a raise error instead of ignoring

pydantic_ai_slim/pydantic_ai/models/openrouter.py
ajac-zero: I added a test for the str case. For the list case, it turns out it is unreachable due to some upstream code in the parent `_into_message_param` method, so I removed it

pydantic_ai_slim/pydantic_ai/models/openrouter.py
```diff
@@ -511,6 +531,18 @@ def _openrouter_settings_to_openai_settings(model_settings: OpenRouterModelSetti
...
+        if isinstance(builtin_tool, ImageGenerationTool):  # pragma: no branch
+            extra_body['modalities'] = ['text', 'image']
+
+            image_config: dict[str, str] = {}
+            if aspect_ratio := builtin_tool.aspect_ratio:
+                image_config['aspect_ratio'] = aspect_ratio
+            extra_body['image_config'] = image_config
+
+    if model_request_parameters.allow_image_output:
+        extra_body['modalities'] = ['text', 'image']
```
DouweM, line 566: Should we remove `text` if `not model_request_parameters.allow_text_output`?

pydantic_ai_slim/pydantic_ai/models/openrouter.py
```diff
@@ -592,6 +652,19 @@ def _map_response_thinking_part(self, item: ThinkingPart) -> None:
...
 
+        @override
+        def _map_response_file_part(self, item: FilePart) -> None:
+            if item.content.media_type in (
+                'image/png',
+                'image/jpeg',
+                'image/webp',
+                'image/gif',
+            ):
+                encoding = base64.b64encode(item.content.data).decode('utf-8')
```
DouweM, line 685: We can use `item.content.base64`

pydantic_ai_slim/pydantic_ai/models/openrouter.py
```diff
@@ -592,6 +652,19 @@ def _map_response_thinking_part(self, item: ThinkingPart) -> None:
...
+                self.file_inputs.append({'type': 'image_url', 'imageUrl': {'url': encoding}})
```
DouweM, line 686: Should we use `item.content.data_uri` instead?

ajac-zero:
Will get back on this soon! My bad. Pinging so it doesn't get closed

PR #3601, opened 2025-11-30
Support json object output for Deepseek provider

zhaob1n:
All models provided by deepseek official support json object https://api-docs.deepseek.com/quick_start/pricing


DouweM:
@xcpky Thanks!

PR #3602, opened 2025-11-30
New PromptedOutput logic

### LINKED ISSUE #3490: Some models ran through OpenAI-compatible APIs need JSON schema in instructions even when native structured output is used
### Description

Currently, `reponse_format` is set to json_object for PromptedOutput if json object is supported.

It's better to set it to json schema if json schema is supported just like native output, before checking the json_object.

The problem with NativeOutput is that, for many endpoints, enabling structured output alone without guiding in the prompt causes bad and unstable response. After all the model doesn't know the schema, it's just logits mask. Openai is a notable exception though.

I can contribute to this if the proposal is OK for the maintainers.

### References

_No response_

zhaob1n:
This PR closes #3490 

This PR handles `supports_json_schema_output` and `supports_json_object_output` together:
1. Check both `supports_json_schema_output` and `supports_json_object_output` for prompted output
2. Add `use_structured_output_for_prompted_output` field to ModelProfile to toggle the checking.

Todos:
- [ ] support other models than openai
- [ ] doc
- [ ] test

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/openai.py
+++ b/pydantic_ai_slim/pydantic_ai/models/openai.py
@@ -56,7 +56,7 @@ from . import Model, ModelRequestParameters, StreamedResponse, check_allow_model
 
 try:
     from openai import NOT_GIVEN, APIConnectionError, APIStatusError, AsyncOpenAI, AsyncStream
-    from openai.types import AllModels, chat, responses
+    from openai.types import AllModels, chat, responses, ResponseFormatJSONObject
     from openai.types.chat import (
         ChatCompletionChunk,
         ChatCompletionContentPartImageParam,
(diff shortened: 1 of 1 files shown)


DouweM:
@xcpky Please see https://github.com/pydantic/pydantic-ai/issues/3490#issuecomment-3598088204. 

I think it is more consistent with current behavior and expectations of what the different modes mean to have a model profile flag that makes `NativeOutput` include the schema in the prompt, than a flag that makes `PromptedOutput` use the provider API native output feature. 

Essentially the question is whether this behavior (schema in prompt + native output feature) is a flavor of "native output" or a flavor of "prompted output", and I feel strongly it's the latter: you're still using the provider's native JSON schema enforcement feature, you're just including the schema in the prompt manually instead of relying on the provider to do so automatically (as OpenAI etc do). The "prompted output" mode is for the case where you don't want any provider-side schema enforcement.

Issue #2513, opened 2025-08-12
Add support for Free-Form Function Calling and Context Free Grammar constraints over string output


matthewfranglen:
### Description

The release of GPT-5 has added a different form of structured output. This uses a context free grammar implemented using either the LARK or regex formats. It is described [here](https://cookbook.openai.com/examples/gpt-5/gpt-5_new_params_and_tools#3-contextfree-grammar-cfg).

With this you can write agents that can reliably produce structured output strings. The example in the link above is for SQL, I have also found this useful for the lucene query language. Composing valid queries of arbitrary complexity using pydantic models is both tricky programmatically and very likely difficult for a given LLM to work with.

I would like to propose the addition of an `output_grammar` to the agent constructor and or run method which takes a suitable grammar, validates it, and then provides it to the completion endpoint. It would be an error to provide the grammar if the `output_type` is not a string.

### References

https://cookbook.openai.com/examples/gpt-5/gpt-5_new_params_and_tools#3-contextfree-grammar-cfg

my lucene experiment:

```python
from openai import OpenAI
from lark import Lark

client = OpenAI()

prompt = (
    "Call the lucene_query to generate a query for the Solr database that will "
    "match the documents that reference coca-cola by name or alias"
)
grammar = LUCENE_GRAMMAR_FILE.read_text()

# check validity of grammar
Lark(grammar, parser="lalr")

# make request, this works with gpt-5 and gpt-5-mini
response = client.responses.create(
    model="gpt-5",
    input=prompt,
    text={"format": {"type": "text"}},
    tools=[
        {
            "type": "custom",
            "name": "lucene_query",
            "description": (
                "Executes read-only lucene queries that match the text of "
                "the document. YOU MUST REASON HEAVILY ABOUT THE QUERY AND "
                "MAKE SURE IT OBEYS THE GRAMMAR."
            ),
            "format": {
                "type": "grammar",
                "syntax": "lark",
                "definition": grammar,
            },
        },
    ],
    parallel_tool_calls=False,
)

print("--- Lucene Query ---")
print(response.output[1].input)
```

The grammar is available [here](https://gist.github.com/matthewfranglen/059a0c226fbcc929d0230520a62140d0).

matthewfranglen:
I've been working on this today and I came up with the following custom OpenAI model:

```python
from typing import Literal, cast

from lark import Lark
from openai import NOT_GIVEN, APIStatusError, AsyncOpenAI
from openai.types import responses
[...]

PR #2572, opened 2025-08-15
Add support for GPT-5 Free-Form Function Calling and Context Free Grammar constraints over tools

matthewfranglen:
- Closes https://github.com/pydantic/pydantic-ai/issues/2513

This is the PR to implement free-form function calling and context free constraints over tools. This implements:

- [x] Free Form Function calling for Tool, @agent.tool, @agent.tool_plain
- [x] Free Form Function calling for output tools
- [x] Context Free constraints over the free form functions
- [x] Documentation
- [x] Tests

I'm hoping to get a working implementation done today, after which we can have the rounds of feedback and improvement required before this can be merged. I will try to keep in mind the feedback that I got on the last PR.

Since there is a specific company voice that you wish to maintain, when suggesting changes to documentation et al please provide the exact wording that you desire.

PR #3612, opened 2025-12-02
Support GPT-5 Freeform Function Calling and Context Free Grammar for tool args and output

cjohnhanson:
- Builds on #2572 by @matthewfranglen, implementing the annotation-based API refactor [suggested by @DouweM](https://github.com/pydantic/pydantic-ai/pull/2572#issuecomment-2893826569).
- Closes https://github.com/pydantic/pydantic-ai/issues/2513

Picks this back up in hopes of bringing it across the finish line.


docs/models/openai.md
```diff
@@ -210,6 +210,137 @@ print(result2.output)
...
+agent = Agent(model)
+
+timestamp_pattern = r'^\d{4}-(0[1-9]|1[0-2])-(0[1-9]|[12]\d|3[01]) (?:[01]\d|2[0-3]):[0-5]\d$'
+
+@agent.tool_plain
+def timestamp_accepting_tool(timestamp: Annotated[str, RegexGrammar(timestamp_pattern)]): ...  # (2)!
+```
+
+1. The GPT-5 family (`gpt-5`, `gpt-5-mini`, `gpt-5-nano`) all support freeform function calling with context free grammar constraints. Unfortunately `gpt-5-nano` often struggles with these calls.
+2. If the tool or model cannot be used with freeform function calling then it will be invoked in the normal way, which may lead to invalid input.
```
DouweM: It would be super cool if we could perform our own agent-side validation of the input by defining `__get_pydantic_core_schema__ ` on the grammar classes. That way this could work even with models other than gpt-5, provided we share the grammar in the tool description and models understand what it means, by relying on the same retry behavior for JSON args: https://ai.pydantic.dev/agents/#reflection-and-self-correction. That'd be similar to Prompted Output mode and non-strict JSON tool args, where we're just relying on the model's understanding instead of strict token constraints.

If we do that, then most of this can be documented outside of OpenAI context.

docs/models/openai.md
```diff
@@ -210,6 +210,137 @@ print(result2.output)
 #> This is an excellent joke invented by Samuel Colvin, it needs no explanation.
 ```
 
+### Freeform Function Calling
```
DouweM: The fact that this can be used for output is a bit buried now, I'd like that to be clearer. If we do what I wrote in the other comment about validating on the agent side, this would warrant sections on the Output and Tool docs.

docs/models/openai.md
```diff
@@ -210,6 +210,137 @@ print(result2.output)
...
+column: "id" | "user_id" | "name" | "age"
+condition: column ("=" | ">" | "<") (NUMBER | STRING)
+%import common.NUMBER
+%import common.ESCAPED_STRING -> STRING
+%import common.WS
+%ignore WS
+''' # (1)!
+
+model = OpenAIResponsesModel('gpt-5')
+agent = Agent(model, output_type=Annotated[str, LarkGrammar(sql_grammar)])
```
DouweM: Let's try to implement support for this feature in `OutlinesModel` as well, as I believe `outlines` supports similar grammar based constraints. That'd be a good way of verifying that the implementation is generic enough to work with providers other than OpenAI.

pydantic_ai_slim/pydantic_ai/_function_schema.py
```diff
@@ -147,6 +150,13 @@ def function_schema(  # noqa: C901
                 errors.append('RunContext annotations can only be used as the first argument')
                 continue
 
+            # Extract text format annotation if present
+            if extracted_format := _extract_text_format(annotation):
+                if text_format is not None:
+                    errors.append('Only one parameter may have a TextFormat annotation')
```
DouweM, line 157: We may be able to weaken this requirement and support multiple grammar-constrained str args, if we can do the validation on our side. Then we'd use OpenAI's custom tools functionality only if there is a single arg with a format annotation.

pydantic_ai_slim/pydantic_ai/_function_schema.py
```diff
@@ -301,3 +312,39 @@ def _build_schema(
...
+def _extract_text_format(annotation: Any) -> TextFormat | None:
+    """Extract a TextFormat annotation from an Annotated type hint.
+
+    Args:
+        annotation: The type annotation to check.
+
+    Returns:
+        The TextFormat instance if found, None otherwise.
+    """
+    from typing import Annotated, get_args, get_origin
```
DouweM: Please move imports to the top of the file

pydantic_ai_slim/pydantic_ai/_function_schema.py
```diff
@@ -301,3 +312,39 @@ def _build_schema(
...
+
+    from .tools import FreeformText, LarkGrammar, RegexGrammar
```
DouweM: Maybe we should define the types in this file, and import them there

pydantic_ai_slim/pydantic_ai/_function_schema.py
```diff
@@ -301,3 +312,39 @@ def _build_schema(
...
+    base_type = args[0]
+    metadata = args[1:]
+
+    # Check if base type is str
+    if base_type is not str:
+        return None
+
+    # Look for TextFormat in metadata
+    for item in metadata:
+        if isinstance(item, (FreeformText, RegexGrammar, LarkGrammar)):
```
DouweM: Or if we make these all subclasses of one type that's defined here, the more interesting subtypes can be defined in `tools`

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -1194,9 +1195,25 @@ def _process_response(  # noqa: C901
             elif isinstance(item, responses.ResponseComputerToolCall):  # pragma: no cover
                 # Pydantic AI doesn't yet support the ComputerUse built-in tool
                 pass
-            elif isinstance(item, responses.ResponseCustomToolCall):  # pragma: no cover
-                # Support is being implemented in https://github.com/pydantic/pydantic-ai/pull/2572
-                pass
+            elif isinstance(item, responses.ResponseCustomToolCall):
+                # Handle custom tool calls (freeform function calling)
+                if item.name not in model_request_parameters.tool_defs:
```
DouweM, line 1290: Let's make it clear this is for the scenario where the model calls a non-existent tool

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -1194,9 +1195,25 @@ def _process_response(  # noqa: C901
...
+                    argument_name = 'input'
+                else:
+                    tool = model_request_parameters.tool_defs[item.name]
+                    tool_argument_name = tool.single_string_argument_name
+                    if tool_argument_name is None:
+                        raise UnexpectedModelBehavior(
```
DouweM: We shouldn't be able to get here, right, as we wouldn't send the tool definition as being a custom tool unless there was a `single_string_argument_name`?

I'd prefer to just use `input` as the key in that case, and let the existing tool args validation deal with errors

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -1194,9 +1195,25 @@ def _process_response(  # noqa: C901
...
+                            f'Custom tool call made to function {item.name} which has unexpected arguments'
+                        )
+                    argument_name = tool_argument_name
+                items.append(
+                    ToolCallPart(
+                        item.name,
+                        {argument_name: item.input},
+                        tool_call_id=_combine_tool_call_ids(item.call_id, item.id),
```
DouweM: We shouldn't need to `_combine_tool_call_ids` anymore, we can use both `tool_call_id` and `id` fields

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -1467,15 +1493,44 @@ def _get_builtin_tools(self, model_request_parameters: ModelRequestParameters) -
             tools.append({'type': 'image_generation'})
         return tools
 
-    def _map_tool_definition(self, f: ToolDefinition) -> responses.FunctionToolParam:
+    def _map_tool_definition(self, f: ToolDefinition) -> responses.FunctionToolParam | responses.CustomToolParam:
+        model_profile = OpenAIModelProfile.from_profile(self.profile)
+        if f.text_format:
+            if not model_profile.openai_supports_freeform_function_calling:
+                raise UserError(
+                    f'Tool {f.name!r} uses freeform function calling but {self._model_name!r} does not support freeform function calling.'
```
DouweM: I'd rather not raise an error, as we won't in other model classes either. We can just use the normal behavior, if possible with agent-side grammar validation

docs/models/openai.md
```diff
@@ -210,6 +210,137 @@ print(result2.output)
...
+from pydantic_ai import Agent, RegexGrammar
+from pydantic_ai.models.openai import OpenAIResponsesModel
+
+model = OpenAIResponsesModel('gpt-5')  # (1)!
+agent = Agent(model)
+
+timestamp_pattern = r'^\d{4}-(0[1-9]|1[0-2])-(0[1-9]|[12]\d|3[01]) (?:[01]\d|2[0-3]):[0-5]\d$'
+
+@agent.tool_plain
+def timestamp_accepting_tool(timestamp: Annotated[str, RegexGrammar(timestamp_pattern)]): ...  # (2)!
```
DouweM: What do you think about supporting the Pydantic pattern `Annotated[str, Field(pattern=...)]` as well?

dsfaccini:
Tests are failing for python 3.10 because `unhashable types` in the new classes

pydantic_ai_slim/pydantic_ai/_function_schema.py
dsfaccini: made it a union

tests/json_body_serializer.py
```diff
@@ -86,7 +89,20 @@ def serialize(cassette_dict: Any):  # pragma: lax no cover
                     # Responses will have the body under a field called 'string'
                     body = body.get('string')
                 if body:
-                    # NOTE(Marcelo): This doesn't handle gzip compression.
+                    if isinstance(body, bytes):
+                        content_encoding = headers.get('content-encoding', [])
```
dsfaccini, line 132: this goes away when https://github.com/pydantic/pydantic-ai/pull/3900 is merged

typings/brotli.pyi
```diff
@@ -0,0 +1 @@
+def decompress(data: bytes) -> bytes: ...
```
dsfaccini: please remove this folder!

tests/test_tools_grammar.py
```diff
@@ -0,0 +1,240 @@
...
+    def test_annotated_non_str_returns_none(self):
+        """Test that Annotated[non-str, TextFormat] returns None."""
+        from pydantic_ai._function_schema import extract_text_format
+
+        # TextFormat on non-str type should return None
+        assert extract_text_format(Annotated[int, RegexGrammar(r'\d+')]) is None
+
+    def test_annotated_without_text_format_returns_none(self):
+        """Test that Annotated without TextFormat returns None."""
+        from pydantic_ai._function_schema import extract_text_format
```
dsfaccini: this import repeats in every function, please extract once to the top

tests/test_tools_grammar.py
```diff
@@ -0,0 +1,240 @@
...
+        json_schema = schema.json_schema
+        phone_desc = json_schema['properties']['phone'].get('description', '')
+
+        # Description should combine docstring description with grammar description
+        assert 'The phone number to look up' in phone_desc
+        assert r'Input must match regex pattern: \d{3}-\d{4}' in phone_desc
+
+    def test_description_without_docstring(self):
+        """Test that grammar description is used when no docstring description exists."""
+        from pydantic.json_schema import GenerateJsonSchema
```
dsfaccini: this one too, import from the top unless strictly necessary to do it inline (bc of circular imports etc)

dsfaccini:
hey @cjohnhanson sorry for hijacking this over the holidays!

DouweM:
@cjohnhanson Great to see some updates here, I'm still excited about this feature! Let me know when it's ready for a full review again or if there were still things you wanted to address first.

PR #3613, opened 2025-12-02
Add latest Grok (xAI) models

darjeeling:
added grok models from [xai-sdk-python](https://github.com/xai-org/xai-sdk-python/blob/main/src/xai_sdk/types/model.py#L5-L26)
* grok-4-1-fast
* grok-4-1-fast-reasoning
* grok-4-1-fast-non-reasoning
* grok-4-fast
* grok-4-fast-reasoning
* grok-4-fast-non-reasoning
* grok-code-fast-1


DouweM:
@darjeeling Thanks!

PR #3618, opened 2025-12-02
Fix structured output with nested definitions with Gemini via OpenRouter

### LINKED ISSUE #3617: OpenRouter + Google models produce degraded output due to schema format changes
### Initial Checks

## Summary

Since #3357, when using Google/Gemini models via OpenRouter (`openrouter:google/gemini-2.5-flash`), the JSON schema sent to the API uses modern JSON Schema features (`$defs`, `$ref`, `anyOf` for nullable types) that Gemini models don't handle well unless the new `parameters_json_schema`/`response_json_schema` field is used, resulting in degraded tool call output.

## Environment

- pydantic-ai version: post-1.19.0
- Model: `openrouter:google/gemini-2.5-flash`

## Expected behavior

The model should correctly populate the `level` field as an object with `level_name` and `level_type`, and `spaces` as a list of objects with `space_name`.

## Actual behavior

The model returns degraded output:
- `level` becomes a simple string (`"ground"`) instead of an object
- `spaces` becomes a list of strings instead of objects

## Root cause

PR #3357 simplified `GoogleJsonSchemaTransformer` by removing workarounds that are now handled natively by the Gemini API as of https://blog.google/technology/developers/gemini-api-structured-outputs/. However, when models are accessed via OpenRouter, these native features may not be properly supported through the compatibility layer.

**Before (v1.19.0)** - Schema sent to API:
```json
{
  "level": {
    "type": "object",
    "nullable": true,
    "properties": { "level_name": {...}, "level_type": {...} }
  }
}
```

**After (main)** - Schema sent to API:
```json
{
  "level": {
    "anyOf": [
      {"$ref": "#/$defs/InsertLevelArg"},
      {"type": "null"}
    ]
  },
  "$defs": {
    "InsertLevelArg": {...}
  }
}
```

The new format with `$defs`, `$ref`, and `anyOf` causes Gemini via OpenRouter to misunderstand the schema structure, even though it works when using `GoogleModel` directly.

## Suggested fix

For OpenRouter specifically, use a transformer that restores the v1.19.0 behavior:
- `prefer_inlined_defs=True` - inline `$defs` instead of using `$ref`
- `simplify_nullable_unions=True` - convert `anyOf: [{...}, {type: null}]` to `nullable: true`

This would be applied only to Google models accessed via OpenRouter, since direct Google API access may work correctly with the new schema format.

## Related

- PR #3357: Support Gemini enhanced JSON Schema features

### Example Code

```Python
import asyncio
import logfire
from pydantic import BaseModel, Field
from enum import Enum

logfire.configure()
logfire.instrument_httpx(capture_all=True)

class LevelType(str, Enum):
    ground = "ground"
    basement = "basement"

class InsertLevelArg(BaseModel):
    level_name: str = Field(description="Name of the level")
    level_type: LevelType = Field(description="Type of level")

class SpaceArg(BaseModel):
    space_name: str = Field(description="Name of the space")

class InsertLevelWithSpacesArgs(BaseModel):
    level: InsertLevelArg | None = Field(default=None, description="Level definition (None for annexes)")
    spaces: list[SpaceArg] = Field(description="List of spaces")

agent = Agent(
    "openrouter:google/gemini-2.5-flash",
    system_prompt="You help configure buildings.",
)

@agent.tool
def insert_level_with_spaces(ctx: RunContext, args: InsertLevelWithSpacesArgs) -> str:
    """Insert a level with its spaces."""
    return f"Level created: {args.level}, spaces: {args.spaces}"

async def main():
    result = await agent.run("Create a ground floor with an entrance and a garage.")

asyncio.run(main())
```

```Text
Python 3.12
pydantic-ai 1.25
gemini-flash-2.5 via openrouter (see MRE)
```

dsfaccini:
Fixes #3617 temporarily while support is added.

### DIFF
--- a/tests/models/test_openrouter.py
+++ b/tests/models/test_openrouter.py
@@ -440,38 +440,33 @@ async def test_openrouter_google_nested_schema(allow_model_requests: None, openr
         space_name: str
         space_type: SpaceType
 
-    class InsertLevelWithSpaces(BaseModel):
-        """Insert a level with its spaces."""
+    class InsertedLevel(BaseModel):
+        """Result of inserting a level."""
 
-        level: InsertLevelArg | None = None
-        spaces: list[SpaceArg]
+        level_name: str
+        level_type: LevelType
+        space_count: int
 
     model = OpenRouterModel('google/gemini-2.5-flash', provider=provider)
-    response = await model_request(
-        model,
-        [
-            ModelRequest.user_text_prompt(
-                "It's a house with a ground floor that has an entryway, a living room and a garage."
-            )
-        ],
-        model_request_parameters=ModelRequestParameters(
-            function_tools=[
-                ToolDefinition(
-                    name='insert_level_with_spaces',
-                    description=InsertLevelWithSpaces.__doc__ or '',
-                    parameters_json_schema=InsertLevelWithSpaces.model_json_schema(),
-                )
-            ],
-            allow_text_output=False,
-        ),
-    )
+    agent: Agent[None, InsertedLevel] = Agent(model, output_type=InsertedLevel)
 
-    assert response.parts == snapshot(
+    @agent.tool_plain
+    def insert_level_with_spaces(level: InsertLevelArg | None, spaces: list[SpaceArg]) -> str:
+        """Insert a level with its spaces."""
+        return f'Inserted level {level} with {len(spaces)} spaces'
+
+    result = await agent.run("It's a house with a ground floor that has an entryway, a living room and a garage.")
+
+    tool_call_message = result.all_messages()[1]
+    assert tool_call_message.parts == snapshot(
         [
             ToolCallPart(
                 tool_name='insert_level_with_spaces',
-                args='{"level":{"level_name":"ground floor","level_type":"ground"},"spaces":[{"space_name":"entryway","space_type":"entryway"},{"space_type":"living-room","space_name":"living room"},{"space_type":"garage","space_name":"garage"}]}',
-                tool_call_id='tool_insert_level_with_spaces_Dp2OGkGjAwsmqUUwHwmj',
+                args='{"spaces":[{"space_type":"entryway","space_name":"entryway"},{"space_name":"living_room","space_type":"living-room"},{"space_name":"garage","space_type":"garage"}],"level":{"level_type":"ground","level_name":"ground_floor"}}',
+                tool_call_id='tool_insert_level_with_spaces_3ZiChYzj8xER8HixJe7W',
             )
         ]
     )
+
+    assert result.output.level_type == LevelType.ground
+    assert result.output.space_count == 3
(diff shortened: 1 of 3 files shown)


pydantic_ai_slim/pydantic_ai/providers/openrouter.py
```diff
@@ -31,6 +32,70 @@
...
+
+def openrouter_google_model_profile(model_name: str) -> ModelProfile | None:
+    """Get the model profile for a Google model accessed via OpenRouter.
+
+    Uses the legacy transformer to maintain compatibility with OpenRouter's
+    translation layer, which doesn't fully support modern JSON Schema features.
+    """
+    is_image_model = 'image' in model_name
+    is_3_or_newer = 'gemini-3' in model_name
+    return GoogleModelProfile(
```
DouweM: We shouldn't copy paste all of this! We should call the original `google_model_profile`, and then do the `replace` thing to swap out just the transformer

pydantic_ai_slim/pydantic_ai/providers/openrouter.py
```diff
@@ -31,6 +32,70 @@
     ) from _import_error
 
 
+class OpenRouterGoogleJsonSchemaTransformer(JsonSchemaTransformer):
```
DouweM: This is public now, I don't think it should be :)

DouweM (changes requested):
@dsfaccini We should also have a recorded test that verifies that a nested schema now works properly with OpenRouter + Gemini again

tests/providers/test_openrouter.py
```diff
@@ -170,3 +173,40 @@ def test_openrouter_provider_model_profile(mocker: MockerFixture):
...
+
+    # oneOf -> anyOf conversion (line 68)
+    assert 'anyOf' in result['properties']['category']
+    assert 'oneOf' not in result['properties']['category']
+
+    # format -> description with existing description (lines 73-75)
+    assert result['properties']['email']['description'] == 'User email (format: email)'
+    assert 'format' not in result['properties']['email']
+
+    # format -> description without existing description (lines 76-77)
```
DouweM: These line numbers will get outdated; please don't include them

tests/models/test_openrouter.py
```diff
@@ -406,3 +406,72 @@ class FindEducationContentFilters(BaseModel):
...
+            ModelRequest.user_text_prompt(
+                "It's a house with a ground floor that has an entryway, a living room and a garage."
+            )
+        ],
+        model_request_parameters=ModelRequestParameters(
+            function_tools=[
+                ToolDefinition(
+                    name='insert_level_with_spaces',
+                    description=InsertLevelWithSpaces.__doc__ or '',
+                    parameters_json_schema=InsertLevelWithSpaces.model_json_schema(),
```
DouweM: Let's use an `Agent` with a `@agent.tool` so we are sure that it goes through the exact same processing that users would see

PR #3626, opened 2025-12-03
Use custom reasoning field for OpenRouter

zhaob1n:
Set custom reasoning field for openrouter.

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/openrouter.py
+++ b/pydantic_ai_slim/pydantic_ai/models/openrouter.py
@@ -567,7 +567,7 @@ class OpenRouterModel(OpenAIChatModel):
             if item.provider_name == self._model.system:
                 if reasoning_detail := _into_reasoning_detail(item):  # pragma: lax no cover
                     self.reasoning_details.append(reasoning_detail.model_dump())
-            else:
+            else:  # pragma: lax no cover
                 super()._map_response_thinking_part(item)
 
     @property


pydantic_ai_slim/pydantic_ai/models/openrouter.py
```diff
@@ -567,11 +567,8 @@ def _map_response_thinking_part(self, item: ThinkingPart) -> None:
             if item.provider_name == self._model.system:
                 if reasoning_detail := _into_reasoning_detail(item):  # pragma: lax no cover
                     self.reasoning_details.append(reasoning_detail.model_dump())
-            elif content := item.content:  # pragma: lax no cover
-                start_tag, end_tag = self._model.profile.thinking_tags
-                self.texts.append('\n'.join([start_tag, content, end_tag]))
             else:
-                pass
+                super()._map_response_thinking_part(item)
```
DouweM, line 571: Can you please add a test for this behavior?

zhaob1n: I'll do it when i find time to learn how to write pydantic-ai test code.

By the way, I don't think this branch will be hit anyway. openrouter use both reasoning and reasoning_detail. So dealing with reasoning_detail guarantees correctness already.

DouweM: I think it'll be hit when using message history from another provider and passing it to `OpenRouterModel`. But the code looks reasonable to me so I'll just go ahead and merge it

pydantic_ai_slim/pydantic_ai/models/openrouter.py
```diff
@@ -567,11 +567,8 @@ def _map_response_thinking_part(self, item: ThinkingPart) -> None:
             if item.provider_name == self._model.system:
                 if reasoning_detail := _into_reasoning_detail(item):  # pragma: lax no cover
                     self.reasoning_details.append(reasoning_detail.model_dump())
-            elif content := item.content:  # pragma: lax no cover
-                start_tag, end_tag = self._model.profile.thinking_tags
-                self.texts.append('\n'.join([start_tag, content, end_tag]))
             else:
```
DouweM:
```suggestion
            else:  # pragma: lax no cover
```

PR #3620, opened 2025-12-02
feat: Add ToolSearchTool and defer_loading for dynamic tool discovery

ryx2:
## Summary

Adds support for dynamic tool discovery without loading all tool definitions upfront, reducing token usage for agents with many tools.

- Add `ToolSearchTool` builtin tool for on-demand tool discovery
  - Supports `search_type` of 'regex' or 'bm25' (defaults to `None` = provider's default)
- Add `defer_loading` field to `ToolDefinition` and `Tool` class
- Automatically add `ToolSearchTool` when any tool has `defer_loading=True`
- Update Anthropic model to map `defer_loading` to API
- Enable `advanced-tool-use-2025-11-20` beta automatically when features are used

Supported by Anthropic models.
See https://docs.anthropic.com/en/docs/agents-and-tools/tool-use/tool-search-tool

## Test plan

- [x] Added unit tests in `tests/models/anthropic/test_tool_search.py`
- [x] All tests pass
- [x] No new pyright errors introduced

## Related

Split from #3550 per @DouweM's recommendation to separate the features into individual PRs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3619, opened 2025-12-02
Add `examples` field to `ToolDefinition` and send it to Anthropic

ryx2:
## Summary

Adds support for providing example inputs to help the model understand correct tool usage patterns.

- Add `input_examples` field to `ToolDefinition` dataclass
- Add `input_examples` parameter to `Tool` class
- Update `FunctionToolset.tool`, `add_function` decorators
- Update `Agent.tool_plain` decorator
- Update Anthropic model to pass `input_examples` to API
- Automatically enable `advanced-tool-use-2025-11-20` beta header when `input_examples` is used

Supported by Anthropic models.
See https://docs.anthropic.com/en/docs/agents-and-tools/tool-use/tool-use-examples

## Test plan

- [x] Added unit tests in `tests/models/anthropic/test_input_examples.py`
- [x] All tests pass
- [x] No new pyright errors introduced

## Related

Split from #3550 per @DouweM's recommendation to separate the features into individual PRs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3550, opened 2025-11-25
feat: Add Anthropic advanced tool use features

ryx2:
## Summary

This PR adds support for Anthropic's new advanced tool use features from the `advanced-tool-use-2025-11-20` beta, announced at https://www.anthropic.com/engineering/advanced-tool-use.

### Features Added

- **Tool Search Tool** (`ToolSearchTool`): Enables dynamic tool discovery without loading all definitions upfront. Supports both regex and BM25 search types. New `defer_loading` field on `ToolDefinition` marks tools for on-demand loading.

- **Programmatic Tool Calling** (`ProgrammaticCodeExecutionTool`): Enhanced code execution that allows Claude to write Python code that calls your custom tools programmatically. New `allowed_callers` field on `ToolDefinition` specifies which tool types can call a tool.

- **Tool Use Examples**: New `input_examples` field on `ToolDefinition` provides example inputs demonstrating correct tool usage patterns.

### Changes

- Added `defer_loading`, `allowed_callers`, and `input_examples` fields to `ToolDefinition`
- Added same parameters to `Tool` class and `@agent.tool`/`@agent.tool_plain` decorators
- Added `ToolSearchTool` and `ProgrammaticCodeExecutionTool` builtin tools
- Updated Anthropic model to map new fields and add beta header automatically
- Updated `_map_server_tool_use_block` to handle tool search responses
- Added comprehensive tests

## Test plan

- [x] Added unit tests for all new `ToolDefinition` fields
- [x] Added tests for `Tool` class with new parameters  
- [x] Added tests for new builtin tools
- [x] Added tests for Anthropic model `_map_tool_definition` with new fields
- [x] Added tests for beta header auto-detection
- [x] Added tests for Agent integration with new parameters
- [x] All 31 new tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3627, opened 2025-12-03
feat: Add Anthropic programmatically_callable (PTC) support

ryx2:
## Summary

This adds support for Anthropic's programmatic tool calling feature, allowing tools to be called from within code execution blocks.

**Part 3 of 3** - Following the review recommendations on #3550, splitting into:
1. PR #3619: `input_examples` feature
2. PR #3620: Tool Search + `defer_loading` feature  
3. **This PR**: Programmatic Tool Calling (`programmatically_callable`)

### Changes

- Add `programmatically_callable` field to `ToolDefinition` dataclass
  - `False` (default): Tool can only be called directly by the model
  - `True`: Tool callable both directly and from code execution
  - `'only'`: Tool only callable from code execution

- Add `programmatically_callable` parameter to:
  - `Tool` class constructor
  - `FunctionToolset.add_function()` and `@tool` decorator
  - `Agent.tool()` and `Agent.tool_plain()` decorators

- Update Anthropic model:
  - Map `programmatically_callable` to Anthropic's `allowed_callers` API
  - Auto-add `CodeExecutionTool` when `programmatically_callable` is used
  - Use newer code execution tool (20250825) when PTC is enabled
  - Store `caller` and `container_id` in `ToolCallPart.provider_details` for tools called from code execution

### Example Usage

```python
from pydantic_ai import Agent
from pydantic_ai.builtin_tools import CodeExecutionTool

agent = Agent('anthropic:claude-sonnet-4-5', builtin_tools=[CodeExecutionTool()])

# Tool callable both directly by model and from code execution
@agent.tool_plain(programmatically_callable=True)
def add_numbers(x: int, y: int) -> int:
    """Add two numbers."""
    return x + y

# Tool only callable from code execution (not directly by model)
@agent.tool_plain(programmatically_callable='only')
def calculate_pi(precision: int) -> float:
    """Calculate pi to given precision."""
    # Only code execution can call this
    return 3.14159265358979
```

## Test plan

- [x] Added unit tests for `programmatically_callable` values (True, 'only', False)
- [x] Added test for auto-adding CodeExecutionTool when PTC is used
- [x] Added tests verifying correct code execution tool version is used
- [x] All existing tests pass

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/agent/__init__.py
+++ b/pydantic_ai_slim/pydantic_ai/agent/__init__.py
@@ -8,7 +8,7 @@ from asyncio import Lock
 from collections.abc import AsyncIterator, Awaitable, Callable, Iterator, Sequence
 from contextlib import AbstractAsyncContextManager, AsyncExitStack, asynccontextmanager, contextmanager
 from contextvars import ContextVar
-from typing import TYPE_CHECKING, Any, ClassVar, overload
+from typing import TYPE_CHECKING, Any, ClassVar, Literal, overload
 
 from opentelemetry.trace import NoOpTracer, use_span
 from pydantic.json_schema import GenerateJsonSchema
@@ -1031,6 +1031,7 @@ class Agent(AbstractAgent[AgentDepsT, OutputDataT]):
         sequential: bool = False,
         requires_approval: bool = False,
         metadata: dict[str, Any] | None = None,
+        programmatically_callable: bool | Literal['only'] = False,
     ) -> Callable[[ToolFuncContext[AgentDepsT, ToolParams]], ToolFuncContext[AgentDepsT, ToolParams]]: ...
 
     def tool(
@@ -1049,6 +1050,7 @@ class Agent(AbstractAgent[AgentDepsT, OutputDataT]):
         sequential: bool = False,
         requires_approval: bool = False,
         metadata: dict[str, Any] | None = None,
+        programmatically_callable: bool | Literal['only'] = False,
     ) -> Any:
         """Decorator to register a tool function which takes [`RunContext`][pydantic_ai.tools.RunContext] as its first argument.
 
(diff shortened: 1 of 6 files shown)


DouweM:
@ryx2 Thanks for splitting this out! 

I wrote some notes in https://github.com/pydantic/pydantic-ai/issues/3666#issuecomment-3629161845 about how I'd like to support this feature for all models by implementing it in Pydantic AI, while leveraging provider-native support if possible, so while we don't need to do all of that right in this PR, we do need to make sure the API we commit to here is compatible with that next step. 

Specifically, let's think through how the `programatically_callable` field interacts with the proposed `ProgrammaticallyCallableToolset` (name TBD!), which can both explicitly wrap another toolset (which would typically have tools with `programmatically_callable` set to the default value), and can be created automatically when any of the agent's tools have `programmatically_callable` set explicitly.

That suggests that the default value for `programmatically_callable` should be `None`, which `ProgrammaticallyCallableToolset` would then interpret as `'only'` when wrapping, as the expectation of creating the toolset would be that these are _only_ available from code, and toolset will expose just a single `run_code` tool. 

If the field is set explicitly to `False`, the toolset will return it directly from `get_tools` (and not make it available to code). If it's set to `True`, it'll be returned from `get_tools` _and_ be available from code. (If Anthropic is used, the tool def will be returned from `get_tools` and the value will be mapped by `AnthropicModel`)

If we detect any `ToolDefinition` with `programmatically_callable` set to a value other than the default of `None`, we automatically create a `ProgrammaticallyCallableToolset` for it, with the behavior as above.

I don't love the `programmatically_callable` name, and it's not clear that `True` means "both directly and programmatically". 

I'm thinking that Anthropic's `allowed_callers` does make more sense, which in our case would become `None | set[Literal['direct', 'code']] = None`, with `None` effectively meaning `{'direct'}` for top-level tools (that wouldn't get any special handling), and `None` meaning `{'code'}` for tools passed into `ProgrammaticallyCallableToolset`. An empty set is a valid value and I suppose should mean that the tool is not callable at all.

Can you make those changes to field name and default please?

I don't like the `ProgrammaticallyCallableToolset` name one bit, but we can figure out something better later.

AlanPonnachan:
Hi @ryx2, this PR seems inactive. Would it be okay if I worked on this and opened a new PR?

DouweM:
@AlanPonnachan Go for it, but note that it's not trivial because of https://github.com/pydantic/pydantic-ai/pull/3627#issuecomment-3629493972 and https://github.com/pydantic/pydantic-ai/issues/3666#issuecomment-3629161845

PR #3634, opened 2025-12-03
Consistently raise `ContentFilterError` when model response is empty because of content filter

### LINKED ISSUE #1035: Content Filtering Exception Handling
I noticed two issues with the Content Filtering behavior of Pydantic AI when used with Azure OpenAI:

1. **Inconsistent error handling for content filtering**: Azure OpenAI uses two different types of content filters:
   - **Prompt filter**: Azure raises a `BadRequestError` [PromptFilterBadRequestBody.txt](https://github.com/user-attachments/files/19054705/PromptFilterBadRequestBody.txt)
   - **Completion filter**: The `finish_reason` on the response will be `content_filter` ([CompletionFilter.txt](https://github.com/user-attachments/files/19054704/CompletionFilter.txt)) 
   
   (Reference: [Azure OpenAI Content Filter documentation](https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/content-filter?tabs=warning%2Cuser-prompt%2Cpython-new#scenario-details))

   This translates to two different errors in Pydantic AI:
   - For prompt filtering: `ModelHTTPError`
   - For completion filtering: `UnexpectedModelBehavior` error. Here we don't even see the filter type.

2. **Lack of model-agnostic content filter handling**: Different providers handle content filtering differently. For example, [Vertex AI handles content filters differently](https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/configure-safety-filters), which makes the current content filter handling in Pydantic AI not model-agnostic.

## Proposed Solution

1. Implement a unique content filter exception that can be used to handle only these specific cases. This exception should:
   - Contain filter type information (hate, sexual, violence, and self-harm) to help users identify problematic and sensitive information in their prompts
   - Include child exceptions (prompt filter and completion filter) to help users handle these cases separately if needed
   - Provide a consistent interface for handling content filtering errors
   - If relevant, make sure to not leake any private user information contained in prompt or generation

2. Ensure that Pydantic AI's behavior on content filters is consistent across different model providers, making the library truly model-agnostic.

These improvements would allow users to better handle content filtering scenarios, potentially by altering prompts or model inputs when content filtering is triggered.

### References

_No response_

AlanPonnachan:
## Unified Content Filtering Exception Handling
This PR closes #1035

Standardizing how content filtering events are handled across different model providers.

Previously, triggering a content filter resulted in inconsistent behaviors: generic `ModelHTTPError` (Azure), `UnexpectedModelBehavior` (Google), or silent failures depending on the provider. This PR introduces a dedicated exception hierarchy to allow users to catch and handle prompt refusals and response interruptions programmatically and consistently.

**Key Changes:**

*   **New Exceptions**: Added `ContentFilterError` (base), `PromptContentFilterError` (for input rejections, e.g., Azure 400), and `ResponseContentFilterError` (for output refusals).
*   **OpenAI & Azure**: Updated logic to raise `PromptContentFilterError` for Azure's specific 400 error body and `ResponseContentFilterError` when `finish_reason='content_filter'`.
*   **Google Gemini**: Updated `_process_response` to raise `ResponseContentFilterError` instead of `UnexpectedModelBehavior` when safety thresholds are triggered.
*   **Anthropic**: Added mapping for refusal stop reasons to raise `ResponseContentFilterError`.
*   **Tests**: Added comprehensive tests covering synchronous and streaming scenarios for OpenAI, Google, and Anthropic in `tests/models/`.

**Example Usage:**

```python
from pydantic_ai import Agent
from pydantic_ai.exceptions import ContentFilterError

try:
    # If the prompt or generation triggers a safety filter
    await agent.run("Generate unsafe content...")
except ContentFilterError as e:
    # Catches both PromptContentFilterError and ResponseContentFilterError
    print(f"Request halted by safety filter: {e}")
```


rahim-figs:
Is it possible to handle AWS Bedrock as well? Thanks.

pydantic_ai_slim/pydantic_ai/exceptions.py
```diff
@@ -179,6 +182,28 @@ def __init__(self, status_code: int, model_name: str, body: object | None = None
...
+
+
+class ResponseContentFilterError(ContentFilterError):
+    """Raised when the generated response triggers a content filter."""
+
+    def __init__(self, message: str, model_name: str, body: object | None = None):
+        self.body = body
+        super().__init__(model_name, message)
+
+
```
dsfaccini: the inheritance chain here is a bit weird: both the `PromptContentFilterError` and `ResponseContentFilterError` will be triggered by an HTTP response, right? 
`ModelAPIError` is used when no connection is established, so I imagine something like the following would be more appropriate:

```
class ContentFilterError(ModelHTTPError):
    """Raised when content filtering is triggered by the model provider."""
    def __init__(self, ...):
        self.reason = reason
        super().__init__(...)

class PromptContentFilterError(ContentFilterError):
    """Raised when the prompt triggers a content filter."""

    def __init__(self, status_code: int, model_name: str, body: object | None = None):
        # if you want to set a specific message for this error type, you should probably do it for both for consistency
        message = f"Prompt triggered content filter for model '{model_name}: {body}'"
        super().__init__(status_code=status_code, model_name=model_name, body=body)

class ResponseContentFilterError(ContentFilterError):
    """Raised when the generated response triggers a content filter."""

    def __init__(self, status_code: int, model_name: str, body: object | None = None):
        super().__init__(status_code=status_code, model_name=model_name, body=body)
```

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -526,6 +526,12 @@ def _process_response(self, response: BetaMessage) -> ModelResponse:
         if raw_finish_reason := response.stop_reason:  # pragma: no branch
             provider_details = {'finish_reason': raw_finish_reason}
             finish_reason = _FINISH_REASON_MAP.get(raw_finish_reason)
+            if finish_reason == 'content_filter':
+                raise ResponseContentFilterError(
+                    f'Content filter triggered for model {response.model}',
```
dsfaccini: if the `message` is the same/similar for all, more of a reason to set it in the model itself to not repeat it everywhere

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -552,7 +552,27 @@ async def _completions_create(
...
+                    error = err_dict.get('error')
+
+                    if isinstance(error, dict):
+                        error_dict = cast(dict[str, Any], error)
+                        if error_dict.get('code') == 'content_filter':
+                            raise PromptContentFilterError(
+                                status_code=status_code,
+                                model_name=self.model_name,
+                                body=cast(dict[str, Any], err_body),
+                            ) from e
```
dsfaccini: if we need to do this, something like 

```
if (
    status_code == 400 and 
    isinstance(e.body, dict) and
    'error' in e.body and
    code := e.body['error'].get('code') == 'content_filter'
):
    raise PromptContentFilterError(
        status_code=status_code,
        model_name=self.model_name,
        body=cast(dict[str, Any], err_body),
    ) from e
```

and if the logic further down is the same you can write a utility like `_handle_content_filter`

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -1390,7 +1424,27 @@ async def _responses_create(  # noqa: C901
...
```
dsfaccini: looks the same as above so the utility is a good idea to deduplicate

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -552,7 +552,27 @@ async def _completions_create(
...
+
+                raise ModelHTTPError(
+                    status_code=status_code,
+                    model_name=self.model_name,
+                    body=cast(dict[str, Any], err_body),
```
dsfaccini: we weren't casting the body before so you don't need to do it here

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -1390,7 +1424,27 @@ async def _responses_create(  # noqa: C901
...
```
dsfaccini: undo cast, leave as before

tests/models/test_anthropic.py
```diff
@@ -7858,3 +7858,58 @@ async def test_anthropic_cache_messages_real_api(allow_model_requests: None, ant
...
+    m = AnthropicModel('claude-haiku-4-5', provider=AnthropicProvider(anthropic_client=mock_client))
+    agent = Agent(m)
+
+    with pytest.raises(ResponseContentFilterError) as exc_info:
+        await agent.run('hello')
+
+    # The mock completion_message uses this model name hardcoded
+    assert exc_info.value.model_name == 'claude-3-5-haiku-123'
+    assert 'Content filter triggered' in str(exc_info.value)
+
```
dsfaccini: since we're not testing the name you can remove that assertion

and instead of the `in exc_info.value` assertion a ptest match would be better

```
        with pytest.raises(ResponseContentFilterError, match='Content filter triggered for model...'): # be thorough in the match, if the error message has the model name use it here too
            await a2a_client.send_message(message=message)
```

tests/models/test_anthropic.py
```diff
@@ -7858,3 +7858,58 @@ async def test_anthropic_cache_messages_real_api(allow_model_requests: None, ant
...
+    m = AnthropicModel('claude-haiku-4-5', provider=AnthropicProvider(anthropic_client=mock_client))
+    agent = Agent(m)
+
+    with pytest.raises(ResponseContentFilterError) as exc_info:
+        async with agent.run_stream('hello') as result:
+            async for _ in result.stream_text():
+                pass
+
+    assert exc_info.value.model_name == 'claude-3-5-haiku-123'
+    assert 'Content filter triggered' in str(exc_info.value)
```
dsfaccini: same here, use pytest match

tests/models/test_google.py
```diff
@@ -4425,3 +4433,51 @@ def test_google_missing_tool_call_thought_signature():
             ],
         }
     )
+
+
+async def test_google_response_filter_error_sync(
+    allow_model_requests: None, google_provider: GoogleProvider, mocker: MockerFixture
+):
+    model = GoogleModel('gemini-1.5-flash', provider=google_provider)
```
dsfaccini: this is a new test so use gemini-2.5-flash instead

tests/models/test_google.py
```diff
@@ -4425,3 +4433,51 @@ def test_google_missing_tool_call_thought_signature():
...
+    response = GenerateContentResponse(candidates=[candidate], model_version='gemini-1.5-flash')
+    mocker.patch.object(model.client.aio.models, 'generate_content', return_value=response)
+
+    agent = Agent(model=model)
+
+    with pytest.raises(ResponseContentFilterError) as exc_info:
+        await agent.run('bad content')
+
+    assert exc_info.value.model_name == 'gemini-1.5-flash'
+    assert 'Content filter' in str(exc_info.value)
```
dsfaccini: - no model name assertion needed
- yes pytest match assertion of the full error message

tests/models/test_openai.py
```diff
@@ -3296,3 +3297,55 @@ async def test_openai_reasoning_in_thinking_tags(allow_model_requests: None):
...
+
+def test_azure_prompt_filter_error(allow_model_requests: None) -> None:
+    mock_client = MockOpenAI.create_mock(
+        APIStatusError(
+            'content filter',
+            response=httpx.Response(status_code=400, request=httpx.Request('POST', 'https://example.com/v1')),
+            body={'error': {'code': 'content_filter', 'message': 'The content was filtered.'}},
+        )
+    )
+    m = OpenAIChatModel('gpt-4o', provider=OpenAIProvider(openai_client=mock_client))
```
dsfaccini: use newer models `gpt-5-mini` whenever possible, unless you require a specific model for content filtering

dsfaccini (changes requested):
Hey @AlanPonnachan thank you for the PR! I've requested a couple small changes, let me know if you have any questions

dsfaccini:
one more thing I missed, please include the name of the new exception in the [fallback model docs](https://ai.pydantic.dev/models/overview/#per-model-settings)

`ModelAPIError, which includes ModelHTTPError and ...`

AlanPonnachan:
@dsfaccini Thank you for the review. I’ve made the requested changes.

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -1398,6 +1430,9 @@ async def _responses_create(  # noqa: C901
             )
         except APIStatusError as e:
             if (status_code := e.status_code) >= 400:
+                _check_azure_content_filter(e, self.model_name)
+
+                # Reverted cast
```
dsfaccini: needless comment, please remove

tests/models/test_openai.py
```diff
@@ -3296,3 +3297,71 @@ async def test_openai_reasoning_in_thinking_tags(allow_model_requests: None):
...
+            body={'error': {'code': 'invalid_parameter', 'message': 'Invalid param.'}},
+        )
+    )
+    m = OpenAIChatModel('gpt-5-mini', provider=OpenAIProvider(openai_client=mock_client))
+    agent = Agent(m)
+    with pytest.raises(ModelHTTPError) as exc_info:
+        agent.run_sync('hello')
+
+    assert not isinstance(exc_info.value, PromptContentFilterError)
+    assert exc_info.value.status_code == 400
```
dsfaccini: the generic 400 is very likely already being tested, so please remove this test and only re-add if coverage asks for it

pydantic_ai_slim/pydantic_ai/exceptions.py
```diff
@@ -179,6 +182,30 @@ def __init__(self, status_code: int, model_name: str, body: object | None = None
...
+        message = f"Prompt content filtered by model '{model_name}'"
+        super().__init__(message, status_code, model_name, body)
+
+
+class ResponseContentFilterError(ContentFilterError):
+    """Raised when the generated response triggers a content filter."""
+
+    def __init__(self, model_name: str, body: object | None = None, status_code: int = 200):
+        message = f"Response content filtered by model '{model_name}'"
+        super().__init__(message, status_code, model_name, body)
```
dsfaccini: this might seem a bit picky but `Prompt content filtered by the model` doesn't seem right

**Prompt**
please change to `Model '{model_name}' content filter was triggered by the user's prompt`

**Response**
please change to `Model '{model_name}' triggered its content filter while generating a response`

you'll need to update the tests that match the exception string as well

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -160,6 +160,24 @@
 }
 
 
+def _check_azure_content_filter(e: APIStatusError, model_name: str) -> None:
+    """Check if the error is an Azure content filter error and raise PromptContentFilterError if so."""
+    if e.status_code == 400:
+        body_any: Any = e.body
+
```
dsfaccini: do we need the type hint here?

dsfaccini (changes requested):

dsfaccini:
hey @AlanPonnachan thanks a lot for your work! It looks very good now, I requested a couple more changes but once that's done and coverage passes I think the PR will be ready for merge.

pydantic_ai_slim/pydantic_ai/models/openai.py
AlanPonnachan: oh sorry, it was for my reference. I will remove

pydantic_ai_slim/pydantic_ai/exceptions.py
AlanPonnachan: makes sense

pydantic_ai_slim/pydantic_ai/models/openai.py
AlanPonnachan: The explicit Any annotation is required to bypass Pyright's strict mode. Without it, e.body is inferred as dict[Unknown, Unknown], which causes reportUnknownMemberType errors on subsequent access.

tests/models/test_openai.py
AlanPonnachan: I have to  add these cases since coverage is asking for it.

AlanPonnachan:
@dsfaccini Thanks again for the review! I’ve applied the requested changes. Test coverage is now at 100%.

pydantic_ai_slim/pydantic_ai/exceptions.py
```diff
@@ -179,6 +182,30 @@ def __init__(self, status_code: int, model_name: str, body: object | None = None
...
 
+class ContentFilterError(ModelHTTPError):
+    """Raised when content filtering is triggered by the model provider."""
+
+    def __init__(self, message: str, status_code: int, model_name: str, body: object | None = None):
+        super().__init__(status_code, model_name, body)
+        self.message = message
+
+
+class PromptContentFilterError(ContentFilterError):
```
DouweM: Can we name this `RequestContentFilterError` for consistency with our `ModelRequest`/`ModelResponse`

pydantic_ai_slim/pydantic_ai/exceptions.py
```diff
@@ -179,6 +182,30 @@ def __init__(self, status_code: int, model_name: str, body: object | None = None
...
+    """Raised when the prompt triggers a content filter."""
+
+    def __init__(self, status_code: int, model_name: str, body: object | None = None):
+        message = f"Model '{model_name}' content filter was triggered by the user's prompt"
```
DouweM: Or by the instructions, right? I'd prefer it to be a bit more generic

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -526,6 +526,11 @@ def _process_response(self, response: BetaMessage) -> ModelResponse:
         if raw_finish_reason := response.stop_reason:  # pragma: no branch
             provider_details = {'finish_reason': raw_finish_reason}
             finish_reason = _FINISH_REASON_MAP.get(raw_finish_reason)
+            if finish_reason == 'content_filter':
+                raise ResponseContentFilterError(
```
DouweM: I think we should do this in `ModelRequestNode`, so that it'll work for all models, not just Anthropic.

That would mean that it wouldn't trigger `FallbackModel` to try another model, but that will be possible once we do https://github.com/pydantic/pydantic-ai/issues/3640.

pydantic_ai_slim/pydantic_ai/models/google.py
```diff
@@ -495,8 +495,8 @@ def _process_response(self, response: GenerateContentResponse) -> ModelResponse:
 
         if candidate.content is None or candidate.content.parts is None:
             if finish_reason == 'content_filter' and raw_finish_reason:
-                raise UnexpectedModelBehavior(
-                    f'Content filter {raw_finish_reason.value!r} triggered', response.model_dump_json()
+                raise ResponseContentFilterError(
```
DouweM: This is a breaking change for users that currently have `except UnexpectedModelBehavior`. So if we add a new exception, I think it should be a subclass of `UnexpectedModelBehavior`, not `ModelAPIError` (as it isn't really an API/HTTP/connection error).

DouweM: Combined with the above, the solution may be to remove this check here, and to implement the same `if response.finish_reason == 'content_filter'` check in `ModelRequestNode` that will apply to all models.

If we want to get some details from the response up to that level, we can store them in `response.provider_details`

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -160,6 +160,24 @@
...
+        if isinstance(body_any, dict):
+            body_dict = cast(dict[str, Any], body_any)
+
+            if (error := body_dict.get('error')) and isinstance(error, dict):
+                error_dict = cast(dict[str, Any], error)
+                if error_dict.get('code') == 'content_filter':
+                    raise PromptContentFilterError(
```
DouweM: What do you think about instead handling this by returning a `ModelResponse` with empty `parts` and `finish_reason == 'content_filter'`, so that it can go through the same logic as the Anthropic and Google models?

I also don't think they distinguish between input/request and output/response content filter errors, so I don't think we need to here either.

DouweM (changes requested):
@AlanPonnachan Thanks for working on this! My main concern is that this doesn't actually make it consistent for all models that respond with `finish_reason=='content_filter'`, just for Anthropic/Google/OpenAI.

docs/models/overview.md
```diff
@@ -229,7 +229,7 @@ contains all the exceptions encountered during the `run` execution.
 
 By default, the `FallbackModel` only moves on to the next model if the current model raises a
 [`ModelAPIError`][pydantic_ai.exceptions.ModelAPIError], which includes
-[`ModelHTTPError`][pydantic_ai.exceptions.ModelHTTPError]. You can customize this behavior by
+[`ModelHTTPError`][pydantic_ai.exceptions.ModelHTTPError] and [`ContentFilterError`][pydantic_ai.exceptions.ContentFilterError]. You can customize this behavior by
```
DouweM: This is no longer true

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -534,6 +534,9 @@ def _finish_handling(
         ctx: GraphRunContext[GraphAgentState, GraphAgentDeps[DepsT, NodeRunEndT]],
         response: _messages.ModelResponse,
     ) -> CallToolsNode[DepsT, NodeRunEndT]:
+        if response.finish_reason == 'content_filter':
```
DouweM: Before this change, some content filter errors would likely have gone unnoticed because the model could've returned some response parts before the filter was hit and the agent run would've continued as normal, or the model could've returned an empty parts list and we would've auto-retried, which may have succeeded.

So I'm very hesitant to turn this behavior, which is likely working fine for some in production, into a hard error, especially one that can't be recovered by `FallbackModel`...

So at the very least I think we should only do this if the model returned no parts and it's very unlikely that retrying will help, by moving the check here:

https://github.com/pydantic/pydantic-ai/blob/eaedf8ae8738a402c57c7f2c9c6aee8809e0de8c/pydantic_ai_slim/pydantic_ai/_agent_graph.py#L602-L609

Note that that's also what the `GoogleModel` logic did.

---

Of course there are also scenarios where you would want to catch every content filter hit, and always try a different model or fail hard. 

The previous `ModelAPIError` based approach would've worked with `FallbackModel`, but still feels too aggressive to me (unless we limit it to responses with 0 parts, which would also make it less useful). 

So perhaps the upcoming https://github.com/pydantic/pydantic-ai/issues/3640 that will make it very easy to fallback based on `response.finish_reason` is enough? That in combination with only raising this error if there are 0 parts.

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -584,6 +598,20 @@ async def _completions_create(
...
             )
         except APIStatusError as e:
+            if _check_azure_content_filter(e):
+                return chat.ChatCompletion(
+                    id='content_filter',
+                    choices=[
+                        chat.chat_completion.Choice(
+                            finish_reason='content_filter',
+                            index=0,
+                            message=chat.ChatCompletionMessage(content='', role='assistant'),
```
DouweM: Do we need this? I'd want it to result in a `ModelResponse` with 0 parts, and this _may_ result in a `TextPart(content='')`

DouweM: It may be cleaner to just build that `ModelResponse` directly...

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/_agent_graph.py
DouweM: Either way the behavior we end up with will need documentation!

pydantic_ai_slim/pydantic_ai/models/openai.py
AlanPonnachan: a ModelResponse directly would be cleaner but _completions_create is strictly typed to return ChatCompletion (or AsyncStream). 
I updated  to constructing a ChatCompletion with content=None (not '') which will successfully results in a ModelResponse with 0 parts via _process_response.

AlanPonnachan:
@DouweM , I’ve made a few changes. Let me know your thoughts.

pydantic_ai_slim/pydantic_ai/exceptions.py
```diff
@@ -152,6 +153,10 @@ def __str__(self) -> str:
             return self.message
 
 
+class ContentFilterError(UnexpectedModelBehavior):
+    """Raised when content filtering is triggered by the model provider."""
```
DouweM:
```suggestion
    """Raised when content filtering is triggered by the model provider resulting in an empty response."""
```

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -176,6 +176,20 @@ def _resolve_openai_image_generation_size(
     return mapped_size
 
 
+def _check_azure_content_filter(e: APIStatusError) -> bool:
+    """Check if the error is an Azure content filter error."""
```
DouweM, line 237: Can we do this only if `self.system == 'azure'`?

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -603,6 +603,12 @@ async def _run_stream() -> AsyncIterator[_messages.HandleResponseEvent]:  # noqa
                             f'Model token limit ({max_tokens or "provider default"}) exceeded before any response was generated. Increase the `max_tokens` model setting, or simplify the prompt to result in a shorter response that will fit within the limit.'
                         )
 
+                    # Check for content filter on empty response
+                    if self.model_response.finish_reason == 'content_filter':
+                        raise exceptions.ContentFilterError(
+                            f'Content filter triggered for model {self.model_response.model_name}'
```
DouweM: Instead of naming the model (which we don't do anywhere else), I'd prefer to include `model_response.provider_details['finish_reason']` if it exists, like we did in the Google exception

tests/models/test_google.py
```diff
@@ -4610,3 +4619,57 @@ def get_country() -> str:
             ),
         ]
     )
+
+
+async def test_google_stream_empty_chunk(
```
DouweM: Do we need this?

pydantic_ai_slim/pydantic_ai/models/google.py
```diff
@@ -503,11 +503,7 @@ def _process_response(self, response: GenerateContentResponse) -> ModelResponse:
             finish_reason = _FINISH_REASON_MAP.get(raw_finish_reason)
```
DouweM, line 576: Are you able to test against Vertex AI?

It may be worth exposing these extra data points on `ModelResponse.provider_details` https://docs.cloud.google.com/vertex-ai/generative-ai/docs/multimodal/configure-safety-filters#gemini-api-in-vertex-ai

pydantic_ai_slim/pydantic_ai/_agent_graph.py
DouweM: Related to what I wrote in google.py about storing more of the actual error response context on `provider_details`, what do you think about including (JSONified) `model_response` on this exception, so that it can be accessed by the user (or in an observability platform)?

pydantic_ai_slim/pydantic_ai/models/google.py
DouweM: Similar for the `content_filter_result` returned by Azure:

```python
{
	'message': "The response was filtered due to the prompt triggering Azure OpenAI's content management policy. Please modify your prompt and retry. To learn more about our content filtering policies please read our documentation: https://go.microsoft.com/fwlink/?linkid=2198766", 
	'type': None, 
	'param': 'prompt', 
	'code': 'content_filter', 
	'status': 400, 
	'innererror': {
		'code': 'ResponsibleAIPolicyViolation', 
		'content_filter_result': {
			'hate': {
				'filtered': True,
				'severity': 'high'
			}, 
			'jailbreak': {
				'filtered': False, 
				'detected': False
			}, 
			'self_harm': {
				'filtered': False, 
				'severity': 'safe'
			}, 
			'sexual': {
				'filtered': False, 
				'severity': 'safe'
			}, 
			'violence': {
				'filtered': False, 
				'severity': 'medium'
			}
		}
	}
}
	
	
```

DouweM (changes requested):

tests/models/test_google.py
AlanPonnachan: required for test coverage for this line:
```
if candidate.content is None or candidate.content.parts is None:
                continue
 ```

File: pydantic_ai_slim/pydantic_ai/models/openai.py [hunk 29]
```diff
@@ -642,6 +655,24 @@ async def _completions_create(
...
+                            message=chat.ChatCompletionMessage(content=None, role='assistant'),
+                        )
+                    ],
+                    created=0,
+                    model=self.model_name,
+                    object='chat.completion',
+                )
+                # Attach inner error details for _process_provider_details to pick up
+                if inner_error := azure_error.get('innererror'):
+                    setattr(resp, 'azure_content_filter_result', inner_error)
```
DouweM: Hmm, now that we're having to resort to this, I do think it'd be easier to have `_check_azure_content_filter` return a `ModelResponse` directly, and then change the signature (and call sites) of this method to allow that as an alternative to `chat.ChatCompletion`.

That'd be similar to what we do here:

https://github.com/pydantic/pydantic-ai/blob/c1213e4a7a978099b64c642b38e1e041d18afad6/pydantic_ai_slim/pydantic_ai/models/groq.py#L202-L221

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -1499,6 +1535,19 @@ async def _responses_create(  # noqa: C901
                 extra_body=model_settings.get('extra_body'),
             )
         except APIStatusError as e:
+            if _check_azure_content_filter(e, self.system):
```
DouweM: Shouldn't we handle the innererror here as well?

tests/models/test_google.py
```diff
@@ -5020,3 +5029,58 @@ async def test_google_system_prompts_and_instructions_ordering(google_provider:
         }
     )
     assert contents == snapshot([{'role': 'user', 'parts': [{'text': 'Hello'}]}])
+
+
+async def test_google_stream_empty_chunk(
```
DouweM: What do we need this test for?

I'd like to see a test that shows that we capture the `safety_ratings` on `provider_details` and expose it in the error

tests/models/test_model_function.py
```diff
@@ -538,3 +539,75 @@ async def test_return_empty():
     with pytest.raises(ValueError, match='Stream function must return at least one item'):
         async with agent.run_stream(''):
             pass
+
+
+async def test_central_content_filter_handling():
```
DouweM: This file is for _testing the function model_, not using the function model to test other things :) Can you move this to `test_agent.py` please?

tests/models/test_openai.py
```diff
@@ -3560,6 +3562,166 @@ async def test_openai_reasoning_in_thinking_tags(allow_model_requests: None):
...
+            body=body,
+        )
+    )
+
+    m = OpenAIChatModel('gpt-5-mini', provider=AzureProvider(openai_client=cast(AsyncAzureOpenAI, mock_client)))
+    agent = Agent(m)
+
+    try:
+        agent.run_sync('bad prompt')
+    except ContentFilterError as e:
```
DouweM: This test will succeed if no exception is raised :) `with pytest.raises(...) with exc_info` let's you access the `exc_info` after the `with` context.

tests/models/test_openai.py
```diff
@@ -3560,6 +3562,166 @@ async def test_openai_reasoning_in_thinking_tags(allow_model_requests: None):
...
+        # Check that the exception was raised
+        assert "Content filter triggered. Finish reason: 'content_filter'" in str(e)
+        # Verify body contains the inner error details (serialized)
+        assert 'ResponsibleAIPolicyViolation' in str(e.body)
```
DouweM: Can we test `assert e.body == snapshot()` so we can see the entire structure?

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -194,6 +194,19 @@ def _resolve_openai_image_generation_size(
     return mapped_size
 
 
+def _check_azure_content_filter(e: APIStatusError, system: str) -> dict[str, Any] | None:
```
DouweM: Would it be possible to do this using Pydantic models as we do here?

https://github.com/pydantic/pydantic-ai/blob/c1213e4a7a978099b64c642b38e1e041d18afad6/pydantic_ai_slim/pydantic_ai/models/groq.py#L663-L688

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -660,7 +724,11 @@ def _process_provider_details(self, response: chat.ChatCompletion) -> dict[str,
 
         This method may be overridden by subclasses of `OpenAIChatModel` to apply custom mappings.
         """
-        return _map_provider_details(response.choices[0])
+        details = _map_provider_details(response.choices[0])
+
+        if hasattr(response, 'azure_content_filter_result'):
+            details['content_filter_result'] = getattr(response, 'azure_content_filter_result')
```
DouweM: This shouldn't be needed anymore

tests/models/test_google.py
```diff
@@ -998,7 +1005,10 @@ async def test_google_model_safety_settings(allow_model_requests: None, google_p
     )
     agent = Agent(m, instructions='You hate the world!', model_settings=settings)
 
-    with pytest.raises(UnexpectedModelBehavior, match="Content filter 'SAFETY' triggered"):
+    with pytest.raises(
+        ContentFilterError,
+        match="Content filter triggered. Finish reason: 'SAFETY'",
```
DouweM, line 1030: Can we verify in this test, that uses a real response we got from the API, that the `safetySettings` stuff is accessible?

tests/models/test_google.py
```diff
@@ -5020,3 +5030,60 @@ async def test_google_system_prompts_and_instructions_ordering(google_provider:
...
+    # Verify exception message
+    assert 'Content filter triggered' in str(exc_info.value)
+
+    # Verify safety ratings are present in the body (serialized ModelResponse)
+    assert exc_info.value.body is not None
+    body_json = json.loads(exc_info.value.body)
+
+    response_msg = body_json[0]
+    assert response_msg['provider_details']['finish_reason'] == 'SAFETY'
+    assert response_msg['provider_details']['safety_ratings'][0]['category'] == 'HARM_CATEGORY_HATE_SPEECH'
```
DouweM, line 5446: See above; I think we can do this in the existing test that uses a real API response, so we don't need this one

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -164,6 +164,34 @@
...
+    filtered: bool
+    severity: str
+
+
+class _AzureContentFilterResult(BaseModel):
+    hate: _AzureContentFilterResultDetail | None = None
+    self_harm: _AzureContentFilterResultDetail | None = None
+    sexual: _AzureContentFilterResultDetail | None = None
+    violence: _AzureContentFilterResultDetail | None = None
+    jailbreak: dict[str, Any] | None = None
```
DouweM: Isn't this also `_AzureContentFilterResultDetail`?

See https://learn.microsoft.com/en-us/azure/ai-foundry/openai/reference?view=foundry-classic#contentfilterpromptresults

Also looks like there's `profanity` we could support

tests/models/test_openai.py
```diff
@@ -3560,6 +3562,205 @@ async def test_openai_reasoning_in_thinking_tags(allow_model_requests: None):
...
+    m = OpenAIChatModel('gpt-5-mini', provider=AzureProvider(openai_client=cast(AsyncAzureOpenAI, mock_client)))
+    agent = Agent(m)
+
+    with pytest.raises(ModelHTTPError) as exc_info:
+        agent.run_sync('hello')
+
+    assert exc_info.value.status_code == 400
+
+
+def test_azure_prompt_filter_no_inner_error(allow_model_requests: None) -> None:
```
DouweM: Is this testing the same as the `test_responses_azure_prompt_filter_error` above?

tests/models/test_openai.py
```diff
@@ -3560,6 +3562,205 @@ async def test_openai_reasoning_in_thinking_tags(allow_model_requests: None):
...
+    )
+
+    m = OpenAIResponsesModel('gpt-5-mini', provider=AzureProvider(openai_client=cast(AsyncAzureOpenAI, mock_client)))
+    agent = Agent(m)
+
+    with pytest.raises(ContentFilterError, match=r"Content filter triggered. Finish reason: 'content_filter'"):
+        agent.run_sync('bad prompt')
+
+
+async def test_openai_response_filter_error_sync(allow_model_requests: None):
```
DouweM: Not sure what "sync" is meant to indicate here

tests/test_agent.py
```diff
@@ -7165,3 +7166,75 @@ async def test_dynamic_tool_in_run_call():
...
+
+    # Should NOT raise ContentFilterError
+    result = await agent.run('Trigger filter')
+    assert result.output == 'Partially generated content...'
+
+
+async def test_content_filter_serialization_fallback():
+    """
+    Test fallback to str() when ModelResponse cannot be serialized to JSON.
+    Covers the except block in _agent_graph.py.
```
DouweM: I don't think we need the except block or this test; `ModelResponse` is guaranteed to be serializable, if it isn't, I'd rather see the error so we fix it ASAP, instead of hiding it.

DouweM (changes requested):

tests/models/test_google.py
```diff
@@ -1008,9 +1008,23 @@ async def test_google_model_safety_settings(allow_model_requests: None, google_p
...
         ContentFilterError,
         match="Content filter triggered. Finish reason: 'SAFETY'",
-    ):
+    ) as exc_info:
         await agent.run('Tell me a joke about a Brazilians.')
 
+    # Verify that we captured the safety settings in the exception body
+    assert exc_info.value.body is not None
+    body_json = json.loads(exc_info.value.body)
+    assert len(body_json) == 1
```
DouweM: Can we replace all the below with `assert body_json == snapshot()`? When you run the test, the snapshot will be filled in. That way we can see the entire structure, same as we did for Azure

tests/models/test_openai.py
```diff
@@ -3574,6 +3574,8 @@ def test_azure_prompt_filter_error(allow_model_requests: None) -> None:
                     'self_harm': {'filtered': False, 'severity': 'safe'},
                     'sexual': {'filtered': False, 'severity': 'safe'},
                     'violence': {'filtered': False, 'severity': 'medium'},
+                    'jailbreak': {'filtered': False, 'detected': False},
+                    'profanity': {'filtered': False, 'detected': True},  # Added profanity
```
DouweM: Unnecessary comment :)

DouweM (changes requested):
@AlanPonnachan Thanks Alan; code looks good now, some minor points on tests, and please have a look at the failing coverage.

AlanPonnachan:
@DouweM The failing coverage issue has been resolved. The other failing tests appear to be unrelated.

tests/models/test_google.py
```diff
@@ -5307,3 +5388,151 @@ async def test_google_system_prompts_and_instructions_ordering(google_provider:
         }
     )
     assert contents == snapshot([{'role': 'user', 'parts': [{'text': 'Hello'}]}])
+
+
+async def test_google_stream_tool_call(
+    allow_model_requests: None, google_provider: GoogleProvider, mocker: MockerFixture
+):
+    """Test streaming a tool call from Google to ensure function_call branch is covered."""
```
DouweM: Why do we need this now and not before? I'd rather remove this

tests/models/test_openai.py
```diff
@@ -3731,6 +3733,183 @@ async def test_openai_reasoning_in_thinking_tags(allow_model_requests: None):
...
+        )
+    )
+    m = OpenAIChatModel('gpt-5-mini', provider=AzureProvider(openai_client=cast(AsyncAzureOpenAI, mock_client)))
+    agent = Agent(m)
+
+    with pytest.raises(ModelHTTPError) as exc_info:
+        agent.run_sync('hello')
+
+    assert exc_info.value.status_code == 400
+    assert not isinstance(exc_info.value, ContentFilterError)
```
DouweM: we can drop this because the `pytest.raises` already ensures it's a ModelHTTPError, which means not a ContentFilterError

DouweM (changes requested):

DouweM:
@AlanPonnachan Thanks Alan!

Issue #3365, opened 2025-11-07
Support Anthropic and OpenAI Skills built-in tool


mattbrandman:
### Description

Add support for native and custom anthropic skills via API as documented here. I think it’s mainly an addition to the code interpreter built in tool https://docs.claude.com/en/api/skills-guide

DouweM:
@mattbrandman OK so this feature requires the following things:

1. Somehow let users specify skills to be passed to the API as `container.skills` with the `skills-2025-10-02` beta header
1. Store `response.container.id` (likely on `ModelResponse.provider_metadata`) and pass it in again when continuing a run (https://docs.claude.com/en/api/skills-guide#multi-turn-conversations)
1. Handle `pause_turn` (https://docs.claude.com/en/api/skills-guide#long-running-operations): https://github.com/pydantic/pydantic-ai/pull/3051
1. Handle the `bash_code_execution_tool_result` event type (https://docs.claude.com/en/api/skills-guide#downloading-generated-files), store the result in `ModelResponse`, and make sure generated `file_id`s can be read easily.
    1. It looks like this is a newer version of the code execution tool than the one we currently support: https://docs.claude.com/en/docs/agents-and-tools/tool-use/code-execution-tool. 


It makes sense to me to consider this a feature of the [`CodeExecutionTool`](https://ai.pydantic.dev/builtin-tools/#code-execution-tool) as you suggest; since it's so Anthropic specific, I think it should be a new `anthropic_skills` field that takes their own skill objects.

I don't think we need to concern ourselves with creating skills or uploading files into a container.

Regarding handling generated files: OpenAI Code Interpreter can generate images that we turn into a `FilePart` with `BinaryContent` in `ModelResponse.parts` (see second example at https://ai.pydantic.dev/builtin-tools/#usage_1), but in this case we don't get the actual file binary content, just a file ID. So it wouldn't make sense to make that a `FilePart`, unless we automatically download the file which could be unexpected. So until we have a native way to represent uploaded files (https://github.com/pydantic/pydantic-ai/issues/2574) the file IDs would only be available on the `BuiltinToolReturnPart.content`, which is annoying but better than nothing. It'd require similar logic to get the ID as the `extract_file_ids` function in Anthropic's examples.

@mattbrandman Would you be up for working on this? :)

mattbrandman:
@DouweM i definitely would be! Don't have a ton of time as I have a few other pressing projects but if they wrap up when I hope could potentially tackle late this month early next.

[...]

PR #3637, opened 2025-12-04
Pass `container_id` back to Anthropic API

mattbrandman:
- Store and resends the container id on the modelresponse.provider_details.
- Always goes by the most recent container id 
- Introduces a setting on modelprofile to force a new container (this is relevant for resetting and handling expiring containers)

solves containerid part of #3365.

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/anthropic.py
+++ b/pydantic_ai_slim/pydantic_ai/models/anthropic.py
@@ -207,7 +207,7 @@ class AnthropicModelSettings(ModelSettings, total=False):
     By default, if previous messages contain a container_id (from a prior response),
     it will be reused automatically.
 
-    Set to `False` to force a fresh container (ignore any container_id from history).
+    Set to `False` to force a fresh container (ignore any `container_id` from history).
     Set to a dict (e.g. `{'id': 'container_xxx'}`) to explicitly specify a container.
     """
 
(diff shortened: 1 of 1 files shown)


mattbrandman:
@DouweM  for review whenever you have time

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -439,6 +452,18 @@ def _get_betas_and_extra_headers(
...
 
+    def _get_container(
+        self, messages: list[ModelMessage], model_settings: AnthropicModelSettings
+    ) -> BetaContainerParams | None:
+        """Get container config for the API request."""
+        if (container := model_settings.get('anthropic_container')) is not None:
+            return None if container is False else container
+        for m in reversed(messages):
+            if isinstance(m, ModelResponse) and m.provider_details:
+                if cid := m.provider_details.get('container_id'):
```
DouweM, line 463: We should verify that the `provider_name` matches `self.system` as well

mattbrandman:
@DouweM updated

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -1239,7 +1270,7 @@ async def _get_event_iterator(self) -> AsyncIterator[ModelResponseStreamEvent]:
             elif isinstance(event, BetaRawMessageDeltaEvent):
                 self._usage = _map_usage(event, self._provider_name, self._provider_url, self._model_name, self._usage)
                 if raw_finish_reason := event.delta.stop_reason:  # pragma: no branch
-                    self.provider_details = {'finish_reason': raw_finish_reason}
+                    self.provider_details = {**(self.provider_details or {}), 'finish_reason': raw_finish_reason}
```
DouweM: I like the 2-line way we do it on line 1158 slightly better

pydantic_ai_slim/pydantic_ai/models/anthropic.py
DouweM: We can merge these 2 `if`s

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -200,6 +201,16 @@ class AnthropicModelSettings(ModelSettings, total=False):
     See https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching for more information.
     """
 
+    anthropic_container: BetaContainerParams | Literal[False]
+    """Container configuration for multi-turn conversations.
+
+    By default, if previous messages contain a container_id (from a prior response),
+    it will be reused automatically.
+
+    Set to `False` to force a fresh container (ignore any container_id from history).
```
DouweM:
```suggestion
    Set to `False` to force a fresh container (ignore any `container_id` from history).
```

DouweM (changes requested):

mattbrandman:
@DouweM  updated

PR #3641, opened 2025-12-04
Replace OTel events with logs

alexmojaki:
The events SDK has been deprecated.


tests/models/test_instrumented.py
```diff
@@ -222,7 +215,7 @@ async def test_instrumented_model(capfire: CaptureLogfire):
         [
             {
                 'body': {'role': 'system', 'content': 'system_prompt'},
-                'severity_number': 9,
+                'severity_number': None,
```
DouweM, line 218: Is this a problem (for Logfire)?

alexmojaki: no

PR #3644, opened 2025-12-04
Disable use of `tool_choice=required` for deepseek-reasoner

ZeroAurora:
I've been experimenting with DeepSeek v3.2 recently, and find the reasoner model from the official endpoint yields 400 Bad Request when `openai_supports_tool_choice_required` is not set to False. That is, deepseek-reasoner does not support `tool_choice=required`, at least for now.

I'm not sure if it's a provider-specific issue or a model characteristic, so I'm changing the provider side.


ZeroAurora:
The test is a false positive I think

PR #3645, opened 2025-12-04
Support `us-gov.` and other multi-character Bedrock geo prefixes

Leundai:
## Summary

The `BedrockProvider.model_profile` method previously only handled 2-character regional prefixes (e.g., `us.`, `eu.`), causing issues with models using longer prefixes like `us-gov.` (AWS GovCloud) and `global.`.

## Problem

When using extended thinking with `us-gov.anthropic.*` models via Bedrock, the `model_profile` method doesn't recognize the `us-gov.` prefix because it only strips 2-character prefixes:

```python
# Previous implementation:
parts = model_name.split('.', 2)
if len(parts) > 2 and len(parts[0]) == 2:
    parts = parts[1:]
```

This causes `bedrock_send_back_thinking_parts` to stay at its default `False` value. As a result, `ThinkingPart` blocks from previous conversation turns are converted to text blocks with thinking tags instead of `reasoningContent`, causing Bedrock to reject the request with:

```
Expected `thinking` or `redacted_thinking`, but found `text`
```

## Solution

1. Added `_strip_geo_prefix()` helper function that properly handles all known geo prefixes including `us-gov.` and `global.`
2. Updated `_AWS_BEDROCK_INFERENCE_GEO_PREFIXES` constant to include `us-gov.` (the comment in the code already suggested this)
3. Added comprehensive parametrized tests for all 8 geo prefixes

## Affected Models

- `us-gov.anthropic.claude-sonnet-4-5-20250929-v1:0`
- `us-gov.anthropic.claude-3-7-sonnet-20250219-v1:0`
- `global.anthropic.claude-opus-4-5-20251101-v1:0`
- And any other cross-region inference profile using multi-character prefixes

## Testing

All existing tests pass, plus new parametrized tests for all geo prefixes

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/providers/bedrock.py
+++ b/pydantic_ai_slim/pydantic_ai/providers/bedrock.py
@@ -59,20 +59,7 @@ def bedrock_deepseek_model_profile(model_name: str) -> ModelProfile | None:
 
 
 # Known geo prefixes for cross-region inference profile IDs
-BEDROCK_GEO_PREFIXES: tuple[str, ...] = ('us.', 'eu.', 'apac.', 'jp.', 'au.', 'ca.', 'global.', 'us-gov.')
-
-
-def _strip_geo_prefix(model_name: str) -> str:
-    """Strip geographic/regional prefix from model name if present.
-
-    AWS Bedrock cross-region inference uses prefixes like 'us.', 'eu.', 'us-gov.', 'global.'
-    to route requests to specific regions. This function strips those prefixes so we can
-    identify the underlying provider and model.
-    """
-    for prefix in BEDROCK_GEO_PREFIXES:
-        if model_name.startswith(prefix):
-            return model_name.removeprefix(prefix)
-    return model_name
+BEDROCK_GEO_PREFIXES: tuple[str, ...] = ('us', 'eu', 'apac', 'jp', 'au', 'ca', 'global', 'us-gov')
 
 
 class BedrockProvider(Provider[BaseClient]):
@@ -104,10 +91,14 @@ class BedrockProvider(Provider[BaseClient]):
             'deepseek': bedrock_deepseek_model_profile,
         }
 
-        model_name = _strip_geo_prefix(model_name)
+        # Split the model name into parts
+        parts = model_name.split('.', 2)
 
-        # Split the model name into provider and model parts
-        parts = model_name.split('.', 1)
+        # Handle regional prefixes
+        if len(parts) > 2 and parts[0] in BEDROCK_GEO_PREFIXES:
+            parts = parts[1:]
+
+        # required format is provider.model-name-with-version
         if len(parts) < 2:
             return None
 
--- a/tests/providers/test_bedrock.py
+++ b/tests/providers/test_bedrock.py
@@ -109,10 +109,19 @@ def test_bedrock_provider_model_profile_all_geo_prefixes(env: TestEnv, prefix: s
     env.set('AWS_DEFAULT_REGION', 'us-east-1')
     provider = BedrockProvider()
 
-    model_name = f'{prefix}anthropic.claude-sonnet-4-5-20250929-v1:0'
+    model_name = f'{prefix}.anthropic.claude-sonnet-4-5-20250929-v1:0'
     profile = provider.model_profile(model_name)
 
     assert profile is not None, f'model_profile returned None for {model_name}'
+
+
+def test_bedrock_provider_model_profile_with_unknown_geo_prefix(env: TestEnv):
+    env.set('AWS_DEFAULT_REGION', 'us-east-1')
+    provider = BedrockProvider()
+
+    model_name = 'narnia.anthropic.claude-sonnet-4-5-20250929-v1:0'
+    profile = provider.model_profile(model_name)
+    assert profile is None, f'model_profile returned {profile} for {model_name}'
 
 
 def test_latest_bedrock_model_names_geo_prefixes_are_supported():
@@ -131,7 +140,7 @@ def test_latest_bedrock_model_names_geo_prefixes_are_supported():
         # - With prefix: "us.anthropic.claude-xxx" (3 parts)
         parts = model_name.split('.')
         if len(parts) >= 3:
-            geo_prefix = parts[0] + '.'
+            geo_prefix = parts[0]
             if geo_prefix not in BEDROCK_GEO_PREFIXES:  # pragma: no cover
                 missing_prefixes.add(geo_prefix)
 
--- a/pydantic_ai_slim/pydantic_ai/models/bedrock.py
+++ b/pydantic_ai_slim/pydantic_ai/models/bedrock.py
@@ -687,8 +687,8 @@ class BedrockConverseModel(Model):
     def _remove_inference_geo_prefix(model_name: BedrockModelName) -> BedrockModelName:
         """Remove inference geographic prefix from model ID if present."""
         for prefix in BEDROCK_GEO_PREFIXES:
-            if model_name.startswith(prefix):
-                return model_name.removeprefix(prefix)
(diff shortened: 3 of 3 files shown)


pydantic_ai_slim/pydantic_ai/providers/bedrock.py
```diff
@@ -58,6 +58,23 @@ def bedrock_deepseek_model_profile(model_name: str) -> ModelProfile | None:
     return profile  # pragma: no cover
 
 
+# Known geo prefixes for cross-region inference profile IDs
+_BEDROCK_GEO_PREFIXES: tuple[str, ...] = ('us.', 'eu.', 'apac.', 'jp.', 'au.', 'ca.', 'global.', 'us-gov.')
```
Leundai: I think at some point it might be worth unifying this and the other private constant in `models/bedrock.py`

But this just extends it to support govcloud and global. Its been a real issue since Claude 4.5 came out and AWS doing cross-region inferencing

dsfaccini: I think you should definitely not duplicate it, since we're already importing the `BedrockModelProfile` in `models/bedrock.py` we should make this public and import it there. https://github.com/pydantic/pydantic-ai/blob/8d111fec483038582cbcf5f0afe45b847480c6ce/pydantic_ai_slim/pydantic_ai/models/bedrock.py#L46
 
@DouweM do you agree?

pydantic_ai_slim/pydantic_ai/providers/bedrock.py
```diff
@@ -87,13 +104,11 @@ def model_profile(self, model_name: str) -> ModelProfile | None:
             'deepseek': bedrock_deepseek_model_profile,
         }
 
-        # Split the model name into parts
-        parts = model_name.split('.', 2)
-
-        # Handle regional prefixes (e.g. "us.")
-        if len(parts) > 2 and len(parts[0]) == 2:
-            parts = parts[1:]
+        # Strip regional/geo prefix if present (e.g. "us.", "eu.", "us-gov.", "global.")
```
dsfaccini: you can remove this comment

tests/providers/test_bedrock.py
```diff
@@ -100,3 +100,50 @@ def test_bedrock_provider_model_profile(env: TestEnv, mocker: MockerFixture):
 
     unknown_model = provider.model_profile('unknown.unknown-model')
     assert unknown_model is None
+
+
+@pytest.mark.parametrize('prefix', ['us.', 'eu.', 'apac.', 'jp.', 'au.', 'ca.', 'global.', 'us-gov.'])
```
dsfaccini: can we use the `BEDROCK_GEO_PREFIXES` you defined above here?

tests/providers/test_bedrock.py
```diff
@@ -100,3 +100,50 @@ def test_bedrock_provider_model_profile(env: TestEnv, mocker: MockerFixture):
...
+    provider = BedrockProvider()
+
+    # Test Anthropic model with geo prefix
+    model_name = f'{prefix}anthropic.claude-sonnet-4-5-20250929-v1:0'
+    profile = provider.model_profile(model_name)
+
+    assert profile is not None, f'model_profile returned None for {model_name}'
+    assert isinstance(profile, BedrockModelProfile)
+    assert profile.bedrock_supports_tool_choice is True
+    assert profile.bedrock_send_back_thinking_parts is True
```
dsfaccini: if these don't vary across models we don't need to test them, right? the test just tests that all prefixes are hadnled properly

tests/providers/test_bedrock.py
```diff
@@ -100,3 +100,50 @@ def test_bedrock_provider_model_profile(env: TestEnv, mocker: MockerFixture):
...
+
+    # Test global. prefix
+    profile = provider.model_profile('global.anthropic.claude-opus-4-5-20251101-v1:0')
+    anthropic_model_profile_mock.assert_called_with('claude-opus-4-5-20251101')
+    assert isinstance(profile, BedrockModelProfile)
+    assert profile.bedrock_supports_tool_choice is True
+    assert profile.bedrock_send_back_thinking_parts is True
```
dsfaccini: we're already testing this above, aren't we?

let's add a test to check that we don't have any bedrock shorthand in `KnownModelName` with a geo prefix that we haven't added in the `BEDROCK_GEO_PREFIXES`

pydantic_ai_slim/pydantic_ai/providers/bedrock.py
Leundai: Sounds good!

Leundai:
@dsfaccini Thanks so much for taking a look! I've addressed your comments here.

pydantic_ai_slim/pydantic_ai/providers/bedrock.py
```diff
@@ -87,13 +104,10 @@ def model_profile(self, model_name: str) -> ModelProfile | None:
...
-        # Split the model name into parts
-        parts = model_name.split('.', 2)
-
-        # Handle regional prefixes (e.g. "us.")
-        if len(parts) > 2 and len(parts[0]) == 2:
-            parts = parts[1:]
+        model_name = _strip_geo_prefix(model_name)
 
+        # Split the model name into provider and model parts
+        parts = model_name.split('.', 1)
```
dsfaccini: we can get rid of the new `_strip_geo_prefix` function, (we went from 5 lines to 25 for the same thing)
let's simply fix the check

```python
        # Split the model name into parts
        parts = model_name.split('.', 2)

        # Handle regional prefixes
        if len(parts) > 2 and parts[0] in BEDROCK_GEO_PREFIXES:
            parts = parts[1:]
```

tests/providers/test_bedrock.py
```diff
@@ -100,3 +101,42 @@ def test_bedrock_provider_model_profile(env: TestEnv, mocker: MockerFixture):
 
     unknown_model = provider.model_profile('unknown.unknown-model')
     assert unknown_model is None
+
+
+@pytest.mark.parametrize('prefix', BEDROCK_GEO_PREFIXES)
```
dsfaccini, line 106: tests look good now, thanks!

can we add another test like this (non-parametrized though) to show what happens when a user passes a model with a prefix that's not in the list? you can make one up like `narnia.`

dsfaccini:
@Leundai thanks again for the changes! Code looks great now, I think after the two requested changes we'll be good to go.

Leundai:
Great! Thanks for taking another look here, simplified this even further
@dsfaccini

tests/providers/test_bedrock.py
```diff
@@ -100,3 +101,51 @@ def test_bedrock_provider_model_profile(env: TestEnv, mocker: MockerFixture):
...
+    assert profile is not None, f'model_profile returned None for {model_name}'
+
+
+def test_bedrock_provider_model_profile_with_unknown_geo_prefix(env: TestEnv):
+    env.set('AWS_DEFAULT_REGION', 'us-east-1')
+    provider = BedrockProvider()
+
+    model_name = 'narnia.anthropic.claude-sonnet-4-5-20250929-v1:0'
+    profile = provider.model_profile(model_name)
+    assert profile is None, f'model_profile returned {profile} for {model_name}'
```
dsfaccini, line 124: @DouweM this is in line with previous behavior but probably bad UX, right? should we parse the anthropic provider properly even if we haven't registered a potential new region prefix? and print a warning?

Leundai: i don't know if we should properly parse if it's an incorrect geo prefix, as a user id be confused that a non existent geo prefix actually worked. 

might be worth printing a warning and returning None still? or just raising an exception (unsure if that's a huge nono)

DouweM: I think this behavior is fine, you'd end up getting an error from the API anyway

PR #3646, opened 2025-12-04
fix: Skip llama_cpp tests on CPUs without AVX2 support

dsfaccini:
## Summary

- Adds AVX2 CPU support detection before running llama_cpp tests
- Gracefully skips tests on machines without AVX2 instead of crashing with `Fatal Python error: Illegal instruction`
- Fixes flaky CI failures on GitHub Actions runners that lack AVX2 support

## Background

The llama_cpp tests have been regularly failing in CI with:
```
Fatal Python error: Illegal instruction
```

This happens because `llama_cpp` uses AVX2 CPU instructions that aren't available on all GitHub Actions runners. Since runners are heterogeneous, the tests are flaky - they pass on some runners and crash on others.

Example failure: https://github.com/pydantic/pydantic-ai/actions/runs/19935894476

## Solution

Add a check for AVX2 support on Linux by reading `/proc/cpuinfo`. If AVX2 is not available, the tests are skipped gracefully rather than crashing.

## Test plan

- [x] Pre-commit checks pass
- [ ] CI should now skip llama_cpp tests on runners without AVX2 instead of crashing

### DIFF
--- a/tests/models/test_outlines.py
+++ b/tests/models/test_outlines.py
@@ -7,6 +7,7 @@ from __future__ import annotations as _annotations
 
 import json
 import os
+import platform
 from collections.abc import Callable
 from pathlib import Path
 from typing import Any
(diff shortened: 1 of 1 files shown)


dsfaccini:
obviously isn't fixing the error, this outlines test is too flaky on CI, some runners obviously miss an instruction that `test_outlines` triggers but I can't pinpoint which

dsfaccini:
@DouweM smart code fixes didn't work. To disable it for now you'd need to change the `ci.yml`

just changing the var:
```
      # RUN_LLAMA_CPP_TESTS: ${{ matrix.python-version == '3.13' && matrix.install.name == 'all-extras' }}
      RUN_LLAMA_CPP_TESTS: false
```

PR #3647, opened 2025-12-05
Fix GoogleModel thinking signature not stored on tool calls when streaming

jerry-heygen:
## Summary

- Fix regression where `ThinkingPart.signature` was never populated for Google model responses (introduced in #3516)
- **This caused thinking context to be lost in follow-up requests**, as Google requires the `thought_signature` to be sent back to maintain reasoning continuity
- Google's API returns `thought_signature` on the part **following** the thinking content (e.g., TextPart or ToolCallPart), not on the ThinkingPart itself
- The fix applies the signature from the following part back to the preceding `ThinkingPart.signature`

## Problem

After #3516, `ThinkingPart.signature` was always `null`. The signature was being stored in `provider_details` on the TextPart/ToolCallPart, but **never preserved on the ThinkingPart**.

This is critical because when sending message history back to Google, `_content_model_response()` reads from `ThinkingPart.signature` to include the signature on the next part. Without it, **thinking context is lost in multi-turn conversations**.

Per [Google's documentation](https://ai.google.dev/gemini-api/docs/thinking#signatures):
> "Signatures are returned from the model within other parts in the response, for example function calling or text parts."
> "Always send the thought_signature back to the model inside its original Part."

## Solution

**Non-streaming path**: Track `last_thinking_part` and when a non-thinking part arrives with `thought_signature`, apply it back to `ThinkingPart.signature`.

**Streaming path**: When a non-thinking part with `thought_signature` arrives, emit a `handle_thinking_delta` event with the signature to update the previous thinking part.

## Test plan

- [x] All existing Google model tests pass (88 passed, 14 skipped)
- [x] `test_google_thought_signature_on_thinking_part` verifies round-trip behavior
- [x] Updated snapshots to include `signature=IsStr()` on ThinkingParts

### DIFF
--- a/tests/test_parts_manager.py
+++ b/tests/test_parts_manager.py
@@ -171,30 +171,64 @@ def test_handle_tool_call_deltas():
     assert manager.get_parts() == snapshot([])
 
     # Now that we have a tool name, we can produce a part:
-    event = manager.handle_tool_call_delta(vendor_part_id='first', tool_name='tool', args=None, tool_call_id='call')
+    event = manager.handle_tool_call_delta(
+        vendor_part_id='first',
+        tool_name='tool',
+        args=None,
+        tool_call_id='call',
+        provider_details={'foo': 'bar'},
+    )
     assert event == snapshot(
         PartStartEvent(
             index=0,
-            part=ToolCallPart(tool_name='tool', args='{"arg1":', tool_call_id='call', part_kind='tool-call'),
+            part=ToolCallPart(
+                tool_name='tool',
+                args='{"arg1":',
+                tool_call_id='call',
+                part_kind='tool-call',
+                provider_details={'foo': 'bar'},
+            ),
             event_kind='part_start',
         )
     )
     assert manager.get_parts() == snapshot(
-        [ToolCallPart(tool_name='tool', args='{"arg1":', tool_call_id='call', part_kind='tool-call')]
+        [
+            ToolCallPart(
+                tool_name='tool',
+                args='{"arg1":',
+                tool_call_id='call',
+                part_kind='tool-call',
+                provider_details={'foo': 'bar'},
+            ),
+        ]
     )
 
-    event = manager.handle_tool_call_delta(vendor_part_id='first', tool_name='1', args=None, tool_call_id=None)
+    event = manager.handle_tool_call_delta(
+        vendor_part_id='first', tool_name='1', args=None, tool_call_id=None, provider_details={'baz': 'qux'}
+    )
     assert event == snapshot(
         PartDeltaEvent(
             index=0,
             delta=ToolCallPartDelta(
-                tool_name_delta='1', args_delta=None, tool_call_id='call', part_delta_kind='tool_call'
+                tool_name_delta='1',
+                args_delta=None,
+                tool_call_id='call',
+                part_delta_kind='tool_call',
+                provider_details={'baz': 'qux'},
             ),
             event_kind='part_delta',
         )
     )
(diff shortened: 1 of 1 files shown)


DouweM:
@jerry-heygen Do you have example code that is failing on `main` but works on this branch?

Because this part is incorrect:
 
> This is critical because when sending message history back to Google, `_content_model_response()` reads from `ThinkingPart.signature` to include the signature on the next part. Without it, **thinking context is lost in multi-turn conversations**.

The logic here prefers reading `thought_signature` from the `provider_details` of the part itself:

https://github.com/pydantic/pydantic-ai/blob/f2bb374b9661705fe9b992542dc0f3808fc9663b/pydantic_ai_slim/pydantic_ai/models/google.py#L793-L803

So it shouldn't be necessary for the signature to (also) be on `ThinkingPart.signature`.

As you can see in the PR diff, all of the `ThinkingPart`s that now have a `signature` were followed by parts that already had `provider_details['thought_signature']`, so that _should_ have been getting sent to Google correctly already.

I'd like to understand the specific scenario in which that wasn't happening, so we can fix it -- adding `ThinkingPart.signature` shouldn't be necessary.

jerry-heygen:
Hi @DouweM , so what we have observed on the last release version (v1.26.0) was that the thought signature had been both missing from the `ThinkingPart.signature` and the `provider_details['thought_signature']` (from relevant parts, like `ToolCallPart`). I have searched for the serialized `list[ModelMessage]` (where we obtained from `agent_run.result.all_messages()`) and there was no thought signature strings in it. This has caused the tool call quality in the multi-turn conversation to degrade, but indeed the strange part of the story is Google didn't reject these requests (we use Vertex AI).

We had to downgrade to v1.21.0 where the thought signature is preserved in the ThinkingPart and we have extensively tested that, the tool call quality is higher and more following to the previous thinking when either we use v1.21.0 or this modified branch as in this PR, compared to v1.26.0.

DouweM:
@jerry-heygen Thanks for the additional context.

I think the bug I introduced in https://github.com/pydantic/pydantic-ai/pull/3516 is that this logic passes `provider_details` (with the `thought_signature`) to `handle_tool_call_delta` correctly:

https://github.com/pydantic/pydantic-ai/blob/f2bb374b9661705fe9b992542dc0f3808fc9663b/pydantic_ai_slim/pydantic_ai/models/google.py#L734-L743

Which then passes it into `ToolCallPartDelta` correctly as well:

https://github.com/pydantic/pydantic-ai/blob/f2bb374b9661705fe9b992542dc0f3808fc9663b/pydantic_ai_slim/pydantic_ai/_parts_manager.py#L295-L297

https://github.com/pydantic/pydantic-ai/blob/f2bb374b9661705fe9b992542dc0f3808fc9663b/pydantic_ai_slim/pydantic_ai/_parts_manager.py#L306-L308

But then the `provider_details` are not passed along to the `ToolCallPart` here:

https://github.com/pydantic/pydantic-ai/blob/f2bb374b9661705fe9b992542dc0f3808fc9663b/pydantic_ai_slim/pydantic_ai/messages.py#L1646-L1655

And here:

https://github.com/pydantic/pydantic-ai/blob/f2bb374b9661705fe9b992542dc0f3808fc9663b/pydantic_ai_slim/pydantic_ai/messages.py#L1690-L1741

Can you please look into fixing it at that level, instead of by restoring the `ThinkingPart.signature`?

jerry-heygen:
@DouweM sure, I've reverted previous changes and applied the correct fix. PTAL

PR #3648, opened 2025-12-05
Add `provider_url` to `ModelResponse` and use it in `cost()`

### LINKED ISSUE #3608: Add `provider_url` to `ModelResponse` and use it in `cost()`
### Description

by passing it to `calc_price`

### References

_No response_

mahiro72:
## Summary
- Add `provider_url` field to `ModelResponse` to store the provider's API endpoint URL
- Pass `provider_url` to `calc_price()` as `provider_api_url` for more accurate pricing calculations
- Add `provider_url` abstract property to `StreamedResponse` and implement it in all subclasses

Closes #3608


pydantic_ai_slim/pydantic_ai/models/gemini.py
```diff
@@ -148,8 +148,8 @@ def __init__(
 
     @property
     def base_url(self) -> str:
-        assert self._url is not None, 'URL not initialized'  # pragma: no cover
-        return self._url  # pragma: no cover
+        assert self._url is not None, 'URL not initialized'
```
mahiro0x00, line 151: [note]
This change now calls this property to retrieve provider_url, so it's now covered by tests.
Removing # pragma: no cover.

tests/models/test_gemini_vertex.py
```diff
@@ -187,6 +188,7 @@ async def test_url_input_force_download(allow_model_requests: None) -> None:  #
                 usage=IsInstance(RequestUsage),
                 model_name='gemini-2.0-flash',
                 timestamp=IsDatetime(),
+                provider_url=IsStr(),
```
alexmojaki: Why IsStr?

mahiro0x00: You're right, `IsStr()` was too permissive since the URL is deterministic. I've replaced it with the concrete URL in the following commit.
[fix: replace IsStr with actual provider_url](https://github.com/pydantic/pydantic-ai/pull/3648/commits/1042538507a247ac8e1d9c5c9a549a5103670c12)

alexmojaki:
Thanks!

PR #3652, opened 2025-12-05
Don't require `anthropic` dependency when using Anthropic model with other provider

### LINKED ISSUE #3625: `pydantic-ai-slim[bedrock]` silently depends on `anthropic`
### Initial Checks

### Description

Since https://github.com/pydantic/pydantic-ai/commit/9106868f11e7f7093317eef61cfdc257abca5d15 bedrock requests to anthropic models can hit https://github.com/pydantic/pydantic-ai/blob/88f6e92ce863995eeceeed5becf69434c5eb5fa6/pydantic_ai_slim/pydantic_ai/profiles/anthropic.py#L53 where the import fails if `anthropic` is not installed. This occurs even if using a model that does not support structured output (eg: `anthropic.claude-haiku-4-5-20251001-v1:0`)

I was able to solve the problem by changing [AnthropicJsonSchemaTransformer](https://github.com/pydantic/pydantic-ai/blob/88f6e92ce863995eeceeed5becf69434c5eb5fa6/pydantic_ai_slim/pydantic_ai/profiles/anthropic.py#L25)`.walk()` like so:
```python
    def walk(self) -> JsonSchema:
        schema = super().walk()

        # The caller (pydantic_ai.models._customize_tool_def or _customize_output_object) coalesces
        # - output_object.strict = self.is_strict_compatible
        # - tool_def.strict = self.is_strict_compatible
        # the reason we don't default to `strict=True` is that the transformation could be lossy
        # so in order to change the behavior (default to True), we need to come up with logic that will check for lossiness
        # https://github.com/pydantic/pydantic-ai/issues/3541
        self.is_strict_compatible = (
            self.strict is True
        )  # not compatible when strict is False/None

        if self.strict is True:
            from anthropic import transform_schema

            return transform_schema(schema)
        else:
            return schema
```
But that seems pretty fragile. And would still fail if you ended up on that code path without `anthropic` installed.

Maybe it's better to modify [anthropic_model_profile()](https://github.com/pydantic/pydantic-ai/blob/88f6e92ce863995eeceeed5becf69434c5eb5fa6/pydantic_ai_slim/pydantic_ai/profiles/anthropic.py#L9) to not include `AnthropicJsonSchemaTransformer` if `supports_json_schema_output` == `False`?

I guess the most robust solution is to attempt importing anthropic and raise a descriptive error if it fails 🤷 

### Example Code

```Python
import boto3
from pydantic_ai import Agent

bedrock_client = boto3.client(
    "bedrock-runtime",
)
model = BedrockConverseModel(
    "anthropic.claude-haiku-4-5-20251001-v1:0",
    provider=BedrockProvider(bedrock_client=bedrock_client),
)


class OutputModel(BaseModel):
    answer: str
    confidence: float


agent = Agent(
    model,
    instructions="Respond to user's question, include your confidence in the answer as a float between 0 and 1",
    output_type=OutputModel,
)

result = agent.run_sync("What city is the capital of France?") # will error here

print(result.output)
```

```Text
python v3.12.10
pydantic-ai-slim[bedrock] v1.25.1
pydantic v2.12.5
boto3 v1.42.1
```

dsfaccini:
Fixes #3625

many thanks to @nataziel for the reporting and suggesting a fix


pydantic_ai_slim/pydantic_ai/providers/bedrock.py
```diff
@@ -75,9 +75,14 @@ def client(self) -> BaseClient:
...
-            'anthropic': lambda model_name: BedrockModelProfile(
-                bedrock_supports_tool_choice=True, bedrock_send_back_thinking_parts=True
-            ).update(anthropic_model_profile(model_name)),
+            'anthropic': lambda model_name: replace(
+                BedrockModelProfile(bedrock_supports_tool_choice=True, bedrock_send_back_thinking_parts=True).update(
+                    anthropic_model_profile(model_name)
+                ),
+                # Bedrock does not support native structured output
+                supports_json_schema_output=False,
+                json_schema_transformer=None,
```
DouweM: `BedrockProvider` is not the only one that uses `anthropic_model_profile`, so with this changes the other providers still have the same issue.

I think we should implement this as I suggested in https://github.com/pydantic/pydantic-ai/issues/3625#issuecomment-3618020309:

> That means moving the `AnthropicJsonSchemaTransformer` bit out of `pydantic_ai.profiles.anthropic.anthropic_model_profile`, and into `AnthropicProvider.model_profile`.

So `anthropic_model_profile` should not have any `json_schema_transformer`, and we should add it in `AnthropicProvider.model_profile` That way, `AnthropicJsonSchemaTransformer` will only be used with `AnthropicProvider`, not any other provider.

pydantic_ai_slim/pydantic_ai/providers/anthropic.py
```diff
@@ -39,7 +42,10 @@ def client(self) -> AsyncAnthropicClient:
         return self._client
 
     def model_profile(self, model_name: str) -> ModelProfile | None:
-        return anthropic_model_profile(model_name)
+        profile = anthropic_model_profile(model_name)
+        if profile:
```
DouweM: I liked the `ModelProfile.from_profile(profile).update(...)` (or `ModelProfile(...).update(profile)`) approach you had before better exactly because it also does the right thing when `anthropic_model_profile` returns `None`

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/providers/anthropic.py
dsfaccini: fixed it now, sorry about that

Issue #3590, opened 2025-11-28
Support deferred loading of tools and discovery via tool search tool


DouweM:
### Description

Inspired by Anthropic's https://www.anthropic.com/engineering/advanced-tool-use and https://platform.claude.com/docs/en/agents-and-tools/tool-use/tool-search-tool, but implemented in Pydantic AI so it works for all providers. From those docs:

> Prompt caching note: Tool Search Tool doesn't break prompt caching because deferred tools are excluded from the initial prompt entirely. They're only added to context after Claude searches for them, so your system prompt and core tool definitions remain cacheable.

Unfortunately this wouldn't apply to our native implementation as the only way we have of passing in tool definitions is through the official list at the start of the context, that we'd vary at each step of the agent run based on what's already been discovered. It looks like Anthropic has a special method of injecting new tool definitions at a later point of the conversation/context, 

Related notes from https://github.com/pydantic/pydantic-ai/pull/3550#discussion_r2566752517:

> But on the other hand, defer_loading seems like a reasonable not-Anthropic-specific name, and I could see how we could implement support for this in Pydantic AI using a new flag that will insert our own tool search tool.

> So I'm OK with this field, but I'd want it to automatically add the ToolSearchTool when used. Similar to how specifying output_type=BinaryImage on OpenAIResponsesModel automatically adds the ImageGenerationTool, and how using strict=True on a ToolDefinition automatically adds the structured-outputs-2025-11-13 Anthropic beta header.

DouweM:
Context on how we could support this for all models _and_ use Anthropic's native support when available: https://github.com/pydantic/pydantic-ai/issues/3666#issuecomment-3629161845

DouweM:
Note to self, contributors, and AI agents who may work on this: https://github.com/pydantic/pydantic-ai/pull/3680 and its spinoff https://github.com/pydantic/pydantic-ai/pull/3831 have been closed as the author has indicated he won't keep working on them, but you should take the code and especially the comments into consideration.

dsfaccini:
I've opened #4090 to hopefully tackle this once and for all

Issue #3179, opened 2025-10-15
Add support for algorithmic optimizers (e.g., GEPA, TextGrad, MIPRO)


linjer:
### Description

It would be helpful to have direct support for integrating various optimizers to algorithmically improve `pydantic-ai` agents.

This is the appeal of DSPy, but the DSPy framework typically also requires adopting their declarative authoring framework, and has very limited support for sophisticated agents.

Examples of optimizers:
- GEPA: https://github.com/gepa-ai/gepa
- TextGrad: https://github.com/zou-group/textgrad

DouweM:
@mwildehahn recently started working on https://github.com/mwildehahn/pydantic-ai-gepa and has made some contributions on our side to make that integration easier: https://github.com/pydantic/pydantic-ai/pull/2926.

@mwildehahn I'd love to hear your thoughts on how we could make the integration stronger/native!

mwildehahn:
@DouweM I've been actively working on pydantic-ai-gepa and using it for our app. It's been really nice 😄 . I feel pretty solid about the api -- it adds support for an "InputType" which fits well with the "OutputType" pydantic-ai already uses. This gets combined into a "Signature" that you can optimize, which is similar to what dspy allows for. I've only been focused on enabling gepa though so probably could use a pass at making sure it's generic enough to support other optimizers.

I'd love to see this be a native part of pydantic-ai. LMK how y'all want to work towards that.

DouweM:
Related: https://github.com/davidberenstein1957/dspydantic

DouweM:
@mwildehahn Great! @dmontagu is going to be looking into this soon

jerry-reevo:
@mwildehahn I tried your `pydantic-ai-gepa` repo a few weeks ago and it ran significantly slower than the DSPy implementation. Any idea why and if there's anything `pydantic-ai` could do to help it run faster?

@dmontagu While you're looking into it, first class support for general optimization techniques not related to DSPy like [ACE](https://arxiv.org/abs/2510.04618) or [Agent Lightning](https://github.com/microsoft/agent-lightning) would also be appreciated but since these are more framework agnostic, it's easier to leverage with custom glue.

mwildehahn:
What was slow about it?

mwildehahn:
There are so many variables here that "runs slowly" is hard to do anything with 😂

jerry-reevo:
> There are so many variables here that "runs slowly" is hard to do anything with 😂

[...]

PR #3656, opened 2025-12-06
Allow customization of `ModelRetry` and `ToolReturn` prompts, `tool_args` and `tool_descriptions` sent to the LLM

### LINKED ISSUE #3566: Allow customization of tool validation error messages
## Problem

  When tool calls fail validation, Pydantic AI sends `"2 validation errors: ... Fix the errors and try again"` to the LLM. Some models interpret this as user input rather than an internal error, causing confused responses.

## Proposed 

Add a way to customize/override the validation error message format. For example with a callable. 
Even better if we'd be able to inject RunContext into that formatter so we could do conditionals based on what the state of the conversation is. 

  ```python

  class Agent:
      def __init__(
          self,
          ...,`
          tool_retry_formatter: Callable[[str, RunContext], str] | None = None`
      ):
          ...

  # Usage
  def format_tool_error(error_message: str, ctx: RunContext) -> str:
      return f"Customizable text that tells the agent exactly what to do, that couldn't be explained pre-validation failure\n{error_message}"
```

###  Use Case

  Prepending custom error 

  Benefits
  - Clearer error context for LLMs (being able to inform that _THIS IS NOT THE USER_)
  - Custom agent behavior on validation failures

### References

[_Slack thread_](https://pydanticlogfire.slack.com/archives/C083V7PMHHA/p1764166495119879)

### PR DESCRIPTION
**Pull request by adtyavrdhn**
Closes #3566

---

## Overview

This PR introduces `PromptConfig`, a new configuration system for customizing all system-generated messages and tool metadata that PydanticAI sends to models.

---

## PromptConfig Overview

`PromptConfig` serves as the central configuration class with two components:

- **`PromptTemplates`**: Customizes system-generated messages (retry prompts, tool return confirmations, validation errors, prompted output instructions, etc.)
- **`ToolConfig`**: Customizes tool metadata (tool descriptions and tool argument descriptions)

```python
from pydantic_ai import Agent, PromptConfig, PromptTemplates, ToolConfig

agent = Agent(
    'openai:gpt-4',
    prompt_config=PromptConfig(
        templates=PromptTemplates(
            validation_errors_retry='Please correct the validation errors and try again.',
            final_result_processed='Result received successfully.',
            prompted_output_template='''
                Respond with JSON matching this schema:
                {schema}
            ''',
        ),
        tool_config=ToolConfig(
            tool_descriptions={
                'search_database': 'Search the customer database for user records.',
            },
            tool_args_descriptions={
                'search_database': {
                    'query': 'Search term - name, ID, or keyword.',
                    'limit': 'Maximum results to return (1-100).',
                },
            },
        ),
    ),
)
```

---

## PromptTemplates

Allows customization of system-injected messages. Each template can be either:
- A **static string** for simple customization
- A **callable** `(part, RunContext) -> str` for dynamic, context-aware messages

### Available Template Fields

| Field | Description |
|-------|-------------|
| `final_result_processed` | Confirmation when a final result is successfully processed |
| `output_tool_not_executed` | Message when an output tool call is skipped because a result was already found |
| `function_tool_not_executed` | Message when a function tool call is skipped due to early termination |
| `tool_call_denied` | Message when a tool call is rejected by an approval handler |
| `validation_errors_retry` | Message appended to validation errors when asking the model to retry |
| `model_retry_string_tool` | Message when `ModelRetry` is raised from a tool |
| `model_retry_string_no_tool` | Message when `ModelRetry` is raised outside of a tool context |
| `prompted_output_template` | Template for prompted output schema instructions (uses `{schema}` placeholder) |

### Prompted Output Template

The `prompted_output_template` field allows customizing the instructions sent to models when using `PromptedOutput`. This template uses a `{schema}` placeholder that gets replaced with the JSON schema definition:

```python
from pydantic_ai import Agent, PromptConfig, PromptTemplates

class UserInfo(BaseModel):
    name: str
    age: int

agent = Agent(
    'openai:gpt-4o',
    output_type=PromptedOutput(UserInfo),
    prompt_config=PromptConfig(
        templates=PromptTemplates(
            prompted_output_template='''
                You must respond with valid JSON matching this schema:
                {schema}
                
                Do not include any markdown formatting or extra text.
            '''
        )
    ),
)
```

**Template Precedence:**
1. `PromptConfig.templates.prompted_output_template` (highest priority)
2. `PromptedOutput.template` (per-output-type override)
3. Model profile default template (lowest priority)

---

## ToolConfig

Allows overriding tool descriptions and argument descriptions at runtime without modifying the original tool definitions. This is useful for providing different descriptions in different contexts or agent runs.

```python
from pydantic_ai import Agent, PromptConfig, ToolConfig

@agent.tool_plain
def search(query: str, limit: int) -> list[str]:
    '''Search for items.'''
    return []

result = agent.run_sync(
    'Find products',
    prompt_config=PromptConfig(
        tool_config=ToolConfig(
            tool_descriptions={'search': 'Search product catalog by name or SKU.'},
            tool_args_descriptions={
                'search': {
                    'query': 'Product name or SKU code.',
                    'limit': 'Maximum results to return (1-100).',
                }
            },
        )
    ),
)
```

### Available Configuration Options

| Field | Description |
|-------|-------------|
| `tool_descriptions` | Dict mapping tool names to custom descriptions |
| `tool_args_descriptions` | Nested dict mapping tool names to argument descriptions: `{'tool_name': {'arg_name': 'description'}}` |

---

## Integration Points

The `prompt_config` parameter is integrated throughout the agent system:

- **Agent constructor**: `Agent(..., prompt_config=...)`
- **Run methods**: `agent.run(..., prompt_config=...)` — allows per-run overrides
- **All execution modes**: `run()`, `run_sync()`, `run_stream()`, `run_stream_events()`, `iter()`
- **Override context manager**: `agent.override(prompt_config=...)`
- **All agent types**: Standard Agent, WrappedAgent, DBOSAgent, PrefectAgent, TemporalAgent

Templates and tool configurations are applied right before messages are sent to the model.

---

## Additional Change: `return_kind` field on `ToolReturnPart`

Added a new `return_kind` field to `ToolReturnPart` that provides contextual visibility into how a tool call was resolved:

| Value | Description |
|-------|-------------|
| `'tool-executed'` | Tool ran successfully and produced a return value |
| `'final-result-processed'` | An output tool produced the run's final result |
| `'output-tool-not-executed'` | An output tool was skipped because a final result already existed |
| `'function-tool-not-executed'` | A function tool was skipped due to early termination after a final result |
| `'tool-denied'` | The tool call was rejected by an approval handler |

This field enables `PromptTemplates` to apply the appropriate template based on the tool return context, and provides useful debugging information when inspecting message history.

---

## Default Constants Exported

The following default constants are exported from `pydantic_ai` for reference and customization:

| Constant | Default Value |
|----------|---------------|
| `DEFAULT_FINAL_RESULT_PROCESSED` | `'Final result processed.'` |
| `DEFAULT_OUTPUT_TOOL_NOT_EXECUTED` | `'Output tool not used - a final result was already processed.'` |
| `DEFAULT_FUNCTION_TOOL_NOT_EXECUTED` | `'Tool not executed - a final result was already processed.'` |
| `DEFAULT_TOOL_CALL_DENIED` | `'The tool call was denied.'` |
| `DEFAULT_MODEL_RETRY` | `'Fix the errors and try again.'` |
| `DEFAULT_OUTPUT_VALIDATION_FAILED` | `'Output tool not used - output failed validation.'` |
| `DEFAULT_PROMPTED_OUTPUT_TEMPLATE` | Multi-line template with `{schema}` placeholder |

---

### Example 1: Using Static Strings

```python
from pydantic_ai import Agent, PromptConfig, PromptTemplates

agent = Agent(
    'openai:gpt-4o',
    prompt_config=PromptConfig(
        templates=PromptTemplates(
            validation_errors_retry='Please correct the validation errors and try again.',
            final_result_processed='Result received successfully.',
        ),
    ),
)
```

### Example 2: Using Callable Functions for Dynamic Messages

```python
from pydantic_ai import Agent, PromptConfig, PromptTemplates
from pydantic_ai.messages import RetryPromptPart
from pydantic_ai.tools import RunContext


def custom_retry_message(part: RetryPromptPart, ctx: RunContext) -> str:
    return f'Attempt #{ctx.retry + 1}: Please fix the errors and try again.'

agent = Agent(
    'openai:gpt-4o',
    prompt_config=PromptConfig(
        templates=PromptTemplates(
            validation_errors_retry=custom_retry_message,
        ),
    ),
)
```

### Example 3: Per-Run Override

```python
from pydantic_ai import Agent, PromptConfig, PromptTemplates

# Override at runtime
result = agent.run_sync(
    'Hello',
    prompt_config=PromptConfig(
        templates=PromptTemplates(validation_errors_retry='Custom retry message for this run.')
    ),
)
```

### Example 4: Using agent.override() Context Manager

```python
from pydantic_ai import Agent, PromptConfig, PromptTemplates

with agent.override(
    prompt_config=PromptConfig(
        templates=PromptTemplates(validation_errors_retry='Another custom message.')
    )
):
    result = agent.run_sync('Hello')
```

### Example 5: Customizing Tool Descriptions at Runtime

```python
from pydantic_ai import Agent, PromptConfig, ToolConfig

agent = Agent(
    'openai:gpt-4o',
    prompt_config=PromptConfig(
        tool_config=ToolConfig(
            tool_descriptions={
                'search_database': 'Search the customer database for user records by name or email.',
                'send_notification': 'Send an urgent notification to the user via their preferred channel.',
            },
            tool_args_descriptions={
                'search_database': {
                    'query': 'Search term to match against user names or email addresses.',
                },
                'send_notification': {
                    'user_id': 'The unique identifier of the user to notify.',
                    'message': 'The notification message content (max 500 characters).',
                },
            },
        ),
    ),
)


@agent.tool_plain
def search_database(query: str) -> list[str]:
    """Original description that will be overridden."""
    return ['result1', 'result2']


@agent.tool_plain
def send_notification(user_id: str, message: str) -> bool:
    """Original description that will be overridden."""
    return True
```


adtyavrdhn:
@DouweM Please take a look whenever you get some time, I have not added a retry-kind on RetryPromptPart because regardless of the way we reach we do end up with the same default string at the end that is "Fix and try again". So I think covering this does the job for us. 

Would you say knowing why this particular RetryPromptPart is being fired off(with a separate field) is worth it on its own as well, I can take that up as a separate PR or in this one as well if we think there is enough use case to warrant it.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -509,6 +510,10 @@ async def _prepare_request(
         # Update the new message index to ensure `result.new_messages()` returns the correct messages
         ctx.deps.new_message_index -= len(original_history) - len(message_history)
 
+        prompt_templates = ctx.deps.prompt_templates
+        if prompt_templates:
+            _apply_prompt_templates(message_history, prompt_templates, run_context)
```
DouweM: We should not modify the `ModelRequest`s or parts in the history in place, so please have this method build new object and a new list.

We should make sure the modified final `ModelRequest` shows up in `result.all_messages()` etc, so we need `ctx.state.message_history[:] = message_history` as above

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -871,13 +881,15 @@ async def process_tool_calls(  # noqa: C901
                     tool_name=call.tool_name,
                     content='Final result processed.',
```
DouweM: Could we remove the hard-coded values here, and make them defaults on `PromptTemplates` or something? Then we may want to _always_ have `PromptTemplates` instead of it being none-able.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -1129,6 +1144,7 @@ async def _call_tool(
                 tool_name=tool_call.tool_name,
                 content=tool_call_result.message,
```
DouweM, line 1215: Similar to above; it'd be nice if we could move the default `'The tool call was denied.'` message off of `ToolDenied`, and only have it on `PromptTemplates`

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -1129,6 +1144,7 @@ async def _call_tool(
...
                 tool_call_id=tool_call.tool_call_id,
+                return_kind='tool-denied',
```
DouweM, line 1217: I'm not too opposed to having this new field, but I wonder if it's strictly necessary. Since we build the `RetryPromptPart`s in this file, would it be an option to explicitly pass something like `content=self.prompt_templates.generate(self.prompt_templates.tool_denied, ctx)`?

pydantic_ai_slim/pydantic_ai/_agent_graph.py
DouweM: This only applies to message history passed in by the user, that's not our to edit. We can of course modify `self.request` and other `ModelRequest`s we build ourselves

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -1361,3 +1378,11 @@ def _clean_message_history(messages: list[_messages.ModelMessage]) -> list[_mess
...
                 clean_messages.append(message)
     return clean_messages
+
+
+def _apply_prompt_templates(
+    messages: list[_messages.ModelMessage], prompt_templates: _messages.PromptTemplates, ctx: RunContext[Any]
+):
+    for msg in messages:
+        for msg_part in msg.parts:
+            prompt_templates.apply_template(msg_part, ctx)
```
DouweM: See above; I'm a bit worried about doing this in-place

pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -410,7 +421,7 @@ def event_stream_handler(self) -> EventStreamHandler[AgentDepsT] | None:
         return self._event_stream_handler
 
     def __repr__(self) -> str:
-        return f'{type(self).__name__}(model={self.model!r}, name={self.name!r}, end_strategy={self.end_strategy!r}, model_settings={self.model_settings!r}, output_type={self.output_type!r}, instrument={self.instrument!r})'
+        return f'{type(self).__name__}(model={self.model!r}, name={self.name!r}, end_strategy={self.end_strategy!r}, model_settings={self.model_settings!r}, prompt_templates={self.prompt_templates!r},output_type={self.output_type!r}, instrument={self.instrument!r})'
```
DouweM: We can exclude it here, we don't include most fields here anyway

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -1929,3 +1957,61 @@ class BuiltinToolResultEvent:
 
 AgentStreamEvent = Annotated[ModelResponseStreamEvent | HandleResponseEvent, pydantic.Discriminator('event_kind')]
 """An event in the agent stream: model response stream events and response-handling events."""
+
+
+@dataclass
+class PromptTemplates:
```
DouweM: This should live in a separate file since it's not a message/part

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -1929,3 +1957,61 @@ class BuiltinToolResultEvent:
...
+    """Templates for customizing messages that Pydantic AI sends to models.
+
+    Each template can be a static string or a callable that receives context and returns a string.
+    """
+
+    retry_prompt: str | Callable[[RetryPromptPart, _RunContext[Any]], str] | None = None
+    """Message sent to the model after validation failures or invalid responses.
+
+    Default: "Validation feedback: {errors}\\n\\nFix the errors and try again."
```
DouweM: Can we actually store the defaults here as default values?

DouweM: This is also not actually the default, because `RetryPromptPart.model_response()` depends on whether `self.content` is a string or validation errors, and whether `self.tool_name` is set.

I think we should have separate templates for each of those scenarios

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -1929,3 +1957,61 @@ class BuiltinToolResultEvent:
...
+            if message.return_kind == 'final-result-processed' and self.final_result_processed:
+                self._apply_tool_template(message, ctx, self.final_result_processed)
+            elif message.return_kind == 'output-tool-not-executed' and self.output_tool_not_executed:
+                self._apply_tool_template(message, ctx, self.output_tool_not_executed)
+            elif message.return_kind == 'function-tool-not-executed' and self.function_tool_not_executed:
+                self._apply_tool_template(message, ctx, self.function_tool_not_executed)
+
+        elif isinstance(message, RetryPromptPart) and self.retry_prompt:
+            if isinstance(self.retry_prompt, str):
+                message.retry_message = self.retry_prompt
```
DouweM: I think we should respect the `{errors}` placeholder, similar to `PromptedOutputSchema.build_instructions`.

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -1929,3 +1957,61 @@ class BuiltinToolResultEvent:
 
 AgentStreamEvent = Annotated[ModelResponseStreamEvent | HandleResponseEvent, pydantic.Discriminator('event_kind')]
 """An event in the agent stream: model response stream events and response-handling events."""
+
+
+@dataclass
+class PromptTemplates:
+    """Templates for customizing messages that Pydantic AI sends to models.
```
DouweM: Ultimately I want every single string that is sent to the model (except for user input) to be overridable using this feature.

That includes the https://ai.pydantic.dev/output/#prompted-output template that's currently derived from `ModelProfile.prompted_output_template`. 

Can you make it so that that's also overridable through this object?

DouweM: For https://github.com/pydantic/pydantic-ai/issues/3179, we're also going to need (static) instructions, tool descriptions, and ideally tool arg descriptions to be overridable.

That doesn't _need_ to be implemented in this PR, but we should be confident we're building a foundation that can be extended in that direction.

Instructions and tool arg descriptions will be trickier, but tool descriptions should be relatively straightforward. So can you look into adding a `tool_descriptions: dict[str, str]` mapping tool names to descriptions? That can then be handled either in `Model.prepare_request` (modifying `function_tools` and `output_tools` based on a new ModelRequestParameters field `tool_descriptions`), or in `Agent._get_toolset` that can wrap the final toolset in a `PreparedToolset` that can override `ToolDefinition` descriptions.

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/_agent_graph.py
adtyavrdhn: I think we could but I like the idea of this kind in the messages, I think the visibility of the ToolReturnPart's context increases.

pydantic_ai_slim/pydantic_ai/messages.py
adtyavrdhn: I added it, although at the cost of a bit of duplication(to find the reason of the retry) I populate the string on the RetryPromptPart and use it along with the errors we were already populating.

pydantic_ai_slim/pydantic_ai/messages.py
adtyavrdhn: I have added it as a suffix to the end of the error descriptions we were created. What do you think?

pydantic_ai_slim/pydantic_ai/messages.py
adtyavrdhn: I'll read about this I've no idea how these algorithm optimizers work and what they need.

adtyavrdhn:
Moving to draft, more of a WIP because I have to figure out how to keep it extendible for prompt optimizers use and how to make instructions, tool desc etc also overridable(if not in this PR then at least provide the interface and arch for it)

docs/agents.md
DouweM, line 1: (I'll review the docs when the API has stablized)

docs/api/prompt_config.md
```diff
@@ -0,0 +1,9 @@
+# `pydantic_ai.prompt_config`
+
+::: pydantic_ai.prompt_config
+    options:
```
DouweM, line 4: I think we can do with the `options` if we use `__all__` in the module

docs/api/prompt_config.md
```diff
@@ -0,0 +1,9 @@
...
+      inherited_members: true
+      members:
```
DouweM, line 6: Don't we also want to include the constants?

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -379,9 +389,16 @@ async def _prepare_request_parameters(
...
+    # Get the prompted output template with precedence:
+    # PromptConfig template > PromptedOutput.template > model profile default (handled downstream)
+    prompted_output_template: str | None = None
+    if isinstance(output_schema, _output.PromptedOutputSchema):
+        prompt_config = ctx.deps.prompt_config
+        templates = prompt_config.templates if prompt_config else None
+        config_template = templates.prompted_output_template if templates else None
+        # Only use config_template if explicitly set, otherwise preserve original behavior
+        # and let the model apply its profile-specific default
+        prompted_output_template = config_template or output_schema.template
```
DouweM: I think this is easier to follow than the multiple `x if y.foo else None` constructions:

```suggestion
    if isinstance(output_schema, _output.PromptedOutputSchema):
        if (
            (prompt_config := ctx.deps.prompt_config)
            and (templates := prompt_config.templates)
            and (template := templates.prompted_output_template)
        ):
            prompted_output_template = template
        else:
            prompted_output_template = output_schema.template
```

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -780,6 +808,12 @@ def _handle_final_result(
 
         # For backwards compatibility, append a new ModelRequest using the tool returns and retries
         if tool_responses:
+            # Only apply templates if explicitly configured
+            if ctx.deps.prompt_config and ctx.deps.prompt_config.templates is not None:
```
DouweM: As above, I prefer to write these things as `if (x := y.z) and (foo := x.bar)` etc, so that we can just use `templates` below instead of repeating the method chain

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -780,6 +808,12 @@ def _handle_final_result(
...
+                run_ctx = build_run_context(ctx)
+                tool_responses = [
+                    ctx.deps.prompt_config.templates.apply_template(part, run_ctx) for part in tool_responses
```
DouweM: Could this use `_apply_prompt_templates_to_message_history` if we run it only on the newly built `_messages.ModelRequest`?

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -865,17 +899,19 @@ async def process_tool_calls(  # noqa: C901
         if final_result and final_result.tool_call_id == call.tool_call_id:
             part = _messages.ToolReturnPart(
                 tool_name=call.tool_name,
-                content='Final result processed.',
+                content=_prompt_config.DEFAULT_FINAL_RESULT_PROCESSED,
```
DouweM, line 923: Instead of having these constants, could we have a `DEFAULT_PROMPT_CONFIG` exported by `_prompt_config` and get the values from there? Or if we have a `prompt_config` that could also be the user-provided one, maybe we don't need to do the `apply_template` stuff in some cases?

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -892,7 +928,7 @@ async def process_tool_calls(  # noqa: C901
                     yield _messages.FunctionToolCallEvent(call)
                     part = _messages.ToolReturnPart(
                         tool_name=call.tool_name,
-                        content='Output tool not used - output failed validation.',
+                        content=_prompt_config.DEFAULT_OUTPUT_VALIDATION_FAILED,
                         tool_call_id=call.tool_call_id,
```
DouweM, line 953: `return_kind` please

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -1380,3 +1421,18 @@ def _clean_message_history(messages: list[_messages.ModelMessage]) -> list[_mess
...
+    messages_with_templates_applied: list[_messages.ModelMessage] = []
+
+    for msg in messages:
+        if isinstance(msg, _messages.ModelRequest):
+            parts_template_applied = [prompt_templates.apply_template(part, ctx) for part in msg.parts]
+            messages_with_templates_applied.append(replace(msg, parts=parts_template_applied))
+        else:
+            messages_with_templates_applied.append(msg)
+
+    return messages_with_templates_applied
```
DouweM: I like list comprehensions: 😄 

```suggestion
def _apply_prompt_templates_to_message_history(
    messages: list[_messages.ModelMessage], prompt_templates: _prompt_config.PromptTemplates, ctx: RunContext[Any]
) -> list[_messages.ModelMessage]:
    return [
        replace(msg, parts=[prompt_templates.apply_template(part, ctx) for part in msg.parts])
        if isinstance(msg, _messages.ModelRequest)
        else msg
        for msg in messages
    ]
```

If we do that `replace(msg, parts=[prompt_templates.apply_template(part, ctx) for part in msg.parts])` thing in a few places, maybe `prompt_templates` should have a method that takes a `ModelRequest`

pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -1409,13 +1455,23 @@ def copy_dynamic_toolsets(toolset: AbstractToolset[AgentDepsT]) -> AbstractTools
...
+        # 2. Per-call parameter (agent.run(..., prompt_config=...))
+        # 3. Agent-level default (Agent(..., prompt_config=...))
+        tool_config = (
+            effective_prompt_config.tool_config
+            if (effective_prompt_config := self._get_prompt_config(prompt_config))
+            else None
+        )
+
+        if self._prepare_tools or tool_config:
+            toolset = PreparedToolset(toolset, self._prepare_tools, tool_config=tool_config)
```
DouweM: Instead of making `PreparedToolset` `tool_config`-aware, I think we can wrap the existing toolsets in a _new_ `PreparedToolset` with a function that applies the transformations from `prompt_config`.

Or, and this is fine by me as well, we make `ToolManager` directly aware of `tool_config`.

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -874,6 +874,26 @@ class ToolReturnPart(BaseToolReturnPart):
     part_kind: Literal['tool-return'] = 'tool-return'
     """Part type identifier, this is available on all parts as a discriminator."""
 
+    return_kind: (
+        Literal[
```
DouweM: Let's have a new type for this

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -949,7 +972,13 @@ def model_response(self) -> str:
             description = (
                 f'{len(self.content)} validation error{"s" if plural else ""}:\n```json\n{json_errors.decode()}\n```'
             )
-        return f'{description}\n\nFix the errors and try again.'
+
+        if self.retry_message is None:
```
DouweM: The "Validation feedback" bit above needs to be customizable as well!

DouweM: And "validation error". I imagine people will use this to translate prompts to other languages as well, so that the LLM doesn't get confused about the mix of English and suddenly start speaking it to the user

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -949,7 +972,13 @@ def model_response(self) -> str:
...
+            from .prompt_config import DEFAULT_MODEL_RETRY
+
+            return f'{description}\n\n{DEFAULT_MODEL_RETRY}'
```
DouweM: Instead of duplicating this, let's set a new local `retry_message` var

pydantic_ai_slim/pydantic_ai/prompt_config.py
```diff
@@ -0,0 +1,306 @@
...
+
+from collections.abc import Callable
+from dataclasses import dataclass, field, replace
+from textwrap import dedent
+from typing import TYPE_CHECKING, Any
+
+from .messages import ModelRequestPart, RetryPromptPart, ToolReturnPart
+
+if TYPE_CHECKING:
+    from ._run_context import RunContext as _RunContext
```
DouweM: No need to alias it

pydantic_ai_slim/pydantic_ai/prompt_config.py
```diff
@@ -0,0 +1,306 @@
...
+                )
+                message_part = self._apply_tool_template(message_part, ctx, template)
+            elif message_part.return_kind == 'function-tool-not-executed':
+                template = (
+                    self.function_tool_not_executed
+                    if self.function_tool_not_executed is not None
+                    else DEFAULT_FUNCTION_TOOL_NOT_EXECUTED
+                )
+                message_part = self._apply_tool_template(message_part, ctx, template)
+            elif message_part.return_kind == 'tool-denied':
```
DouweM: Please add an `else: assert_never(message_part)` branch so we can't forget to add a new branch here when we add a literal value

pydantic_ai_slim/pydantic_ai/prompt_config.py
```diff
@@ -0,0 +1,306 @@
...
+                    else DEFAULT_OUTPUT_TOOL_NOT_EXECUTED
+                )
+                message_part = self._apply_tool_template(message_part, ctx, template)
+            elif message_part.return_kind == 'function-tool-not-executed':
+                template = (
+                    self.function_tool_not_executed
+                    if self.function_tool_not_executed is not None
+                    else DEFAULT_FUNCTION_TOOL_NOT_EXECUTED
+                )
+                message_part = self._apply_tool_template(message_part, ctx, template)
```
DouweM: No need to repeat this line

pydantic_ai_slim/pydantic_ai/prompt_config.py
```diff
@@ -0,0 +1,306 @@
...
+                    if self.output_tool_not_executed is not None
+                    else DEFAULT_OUTPUT_TOOL_NOT_EXECUTED
+                )
+                message_part = self._apply_tool_template(message_part, ctx, template)
+            elif message_part.return_kind == 'function-tool-not-executed':
+                template = (
+                    self.function_tool_not_executed
+                    if self.function_tool_not_executed is not None
+                    else DEFAULT_FUNCTION_TOOL_NOT_EXECUTED
+                )
```
DouweM: I think we can just use `or` here

pydantic_ai_slim/pydantic_ai/prompt_config.py
```diff
@@ -0,0 +1,306 @@
...
+                    self.output_tool_not_executed
+                    if self.output_tool_not_executed is not None
+                    else DEFAULT_OUTPUT_TOOL_NOT_EXECUTED
+                )
+                message_part = self._apply_tool_template(message_part, ctx, template)
+            elif message_part.return_kind == 'function-tool-not-executed':
+                template = (
+                    self.function_tool_not_executed
+                    if self.function_tool_not_executed is not None
+                    else DEFAULT_FUNCTION_TOOL_NOT_EXECUTED
```
DouweM: It'd be nice to read this from a dict that has the same `'function-tool-not-executed'` etc values as keys

pydantic_ai_slim/pydantic_ai/prompt_config.py
```diff
@@ -0,0 +1,306 @@
...
+        message_part: RetryPromptPart,
+        ctx: _RunContext[Any],
+        template: str | Callable[[RetryPromptPart, _RunContext[Any]], str],
+    ) -> RetryPromptPart:
+        if isinstance(template, str):
+            message_part = replace(message_part, retry_message=template)
+        else:
+            message_part = replace(message_part, retry_message=template(message_part, ctx))
+
+        return message_part
```
DouweM: This can be much simpler:

```suggestion
        if not isinstance(template, str):
            template = template(message_part, ctx)
        return replace(message_part, retry_message=template)
```

pydantic_ai_slim/pydantic_ai/prompt_config.py
```diff
@@ -0,0 +1,306 @@
...
+        ```
+    """
+
+    tool_descriptions: dict[str, str] = field(default_factory=lambda: {})
+    """Custom descriptions for tools, keyed by tool name."""
+
+    tool_args_descriptions: dict[str, dict[str, str]] = field(default_factory=lambda: {})
+    """Custom descriptions for tool arguments: `{'tool_name': {'arg_name': 'description'}}`."""
+
+    def get_tool_args_for_tool(self, tool_name: str) -> dict[str, str] | None:
```
DouweM: Not sure if this is worth a method

pydantic_ai_slim/pydantic_ai/prompt_config.py
```diff
@@ -0,0 +1,306 @@
...
+                            'limit': 'Maximum results to return (1-100).',
+                        }
+                    },
+                )
+            ),
+        )
+        ```
+    """
+
+    tool_descriptions: dict[str, str] = field(default_factory=lambda: {})
```
DouweM: Looking at `ToolDefinition`, optimizers should also have the ability to modify `name`, and it'd be useful to test if using `strict` has an impact on accuracy/performance as well. Also, `tool_args_descriptions` is doesn't extend to descriptions inside arg types (e.g. basemodels), which are optimization targets as well.

So I propose we make `tool_config` a `dict[str, ToolConfig]` where `ToolConfig` describes a single tool's `name`, `description`, and `strict`. For nested descriptions in the schema, we could either allow the full `parameters_json_schema` to be replaced, but optimizing LLMs will be too likely to mess it up in an incompatible way (causing validation errors) so perhaps we could support nested fields on `tool_args_descriptions` using `foo.bar` or `$defs.foo.bar`.

pydantic_ai_slim/pydantic_ai/prompt_config.py
```diff
@@ -0,0 +1,306 @@
...
+    tool_args_descriptions: dict[str, dict[str, str]] = field(default_factory=lambda: {})
+    """Custom descriptions for tool arguments: `{'tool_name': {'arg_name': 'description'}}`."""
+
+    def get_tool_args_for_tool(self, tool_name: str) -> dict[str, str] | None:
+        """Get the tool argument descriptions for the given tool name."""
+        return self.tool_args_descriptions.get(tool_name)
+
+
+@dataclass
+class PromptConfig:
```
DouweM, line 409: It should also be able to _generate_ a `PromptConfig` from an `Agent` (+ values you may pass to `run`) with all of the values filled in with their defaults or generated values like tool descriptions, as a starting point for the optimizer.

pydantic_ai_slim/pydantic_ai/prompt_config.py
```diff
@@ -0,0 +1,306 @@
...
+
+    tool_config: ToolConfig | None = None
+    """Configuration for customizing tool descriptions and metadata.
+    See [`ToolConfig`][pydantic_ai.ToolConfig] for available configuration options.
+    """
+
+    def __post_init__(self):  # pragma: no cover
+        if self.templates is None and self.tool_config is None:
+            raise ValueError(
+                "PromptConfig requires at least 'templates' or 'tool_config' to be provided. "
```
DouweM: Do we really need to require this?

pydantic_ai_slim/pydantic_ai/toolsets/prepared.py
```diff
@@ -16,10 +19,57 @@ class PreparedToolset(WrapperToolset[AgentDepsT]):
     See [toolset docs](../toolsets.md#preparing-tool-definitions) for more information.
     """
 
-    prepare_func: ToolsPrepareFunc[AgentDepsT]
+    prepare_func: ToolsPrepareFunc[AgentDepsT] | None
+    tool_config: ToolConfig | None = None
```
DouweM: As mentioned above, I'd prefer to implement this with a new `PreparedToolset` with a custom function, _or_ a new `PreparedToolset` subclass, _or_ directly on `ToolManager`. I like the new `PreparedToolset` subclass option best, so we can return it right at the end of `Agent._get_toolset`.

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/_agent_graph.py
adtyavrdhn: Haha I feel so dumb, I had a default PromtpConfig before but I removed it for this because I wasn't liking the look of DEFAULT_PROMPT_CONFIG.templates.__ something. I realize this looks messier now with so many constants being imported from another class.

pydantic_ai_slim/pydantic_ai/messages.py
adtyavrdhn: Oh yep, that makes sense. Missed that

pydantic_ai_slim/pydantic_ai/prompt_config.py
adtyavrdhn: I was thinking users initializing an empty PromptConfig() should not have any bias that it will do something and needs at least one of the inner fields to operate something with.

pydantic_ai_slim/pydantic_ai/toolsets/prepared.py
adtyavrdhn: I'll take a look.

pydantic_ai_slim/pydantic_ai/messages.py
adtyavrdhn: So I added a description_template, which I am restricting to be a Callable so that the users are aware content is available to be used while generating it. It isn't pretty atm but I think quite useful.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
adtyavrdhn: I also did for the typing and avoiding casting to a string in one place. I'll give the older one another shot and see which one feels appropriate.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
adtyavrdhn:
```
            message = _messages.ModelRequest(parts=tool_responses, run_id=ctx.state.run_id)

            if (prompt_config := ctx.deps.prompt_config) and (prompt_templates := prompt_config.templates):
                run_ctx = build_run_context(ctx)
                message = _apply_prompt_templates_to_message_history([message], prompt_templates=prompt_templates, ctx=run_ctx)[0]

            messages.append(message)
```

Would you prefer this over directly using prompt_templates? Is this for consistency in not touching prompt_templates directly in this file and only via the method we add?

pydantic_ai_slim/pydantic_ai/_agent_graph.py
adtyavrdhn: I do too, admittedly this PR got a bit out of hand there for me. Reviewing it to cleanup has been a hit or miss(mostly miss :')) )

adtyavrdhn: I've cleaned this up, removed this helper and added it in prompt_templates itself.

pydantic_ai_slim/pydantic_ai/messages.py
adtyavrdhn: I am not sure if we need this anymore? I have refactored it with description as well. Let me know what you think of the new structure.

pydantic_ai_slim/pydantic_ai/prompt_config.py
adtyavrdhn: I have added a dict, I can also see using a mapping where we have attributes as these function names and fetching the attribute using return kind and using it.

```
mapping = {'function-tool-not-executed': 'function_tool_not_executed'}
template = getattr(mapping, return_kind)
```

We could do it but I wouldn't personally like looking at it.

pydantic_ai_slim/pydantic_ai/prompt_config.py
adtyavrdhn: Nope, isn't. will remove

pydantic_ai_slim/pydantic_ai/agent/__init__.py
adtyavrdhn: Yes I like the PreparedToolSet wrapper much more. Thanks. It looks a lot cleaner and keeps the concerns separate.

pydantic_ai_slim/pydantic_ai/prompt_config.py
adtyavrdhn: Added this.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
adtyavrdhn: Actually I refactored this whole thing. Removed the helper altogether so that we only trigger via prompt_templates.apply_template_message_history()

pydantic_ai_slim/pydantic_ai/prompt_config.py
adtyavrdhn:
```
    @staticmethod
    async def generate_prompt_config_from_agent(agent: Agent, model: Model) -> PromptConfig:
        """Generate a PromptConfig instance based on an Agent instance.

        The information we can find we will fill in, everything else can be None or defaults if any.

        """
        tool_config: dict[str, ToolConfig] = {}

        prompt_templates: PromptTemplates = PromptTemplates(
            final_result_processed=DEFAULT_FINAL_RESULT_PROCESSED,
            output_tool_not_executed=DEFAULT_OUTPUT_TOOL_NOT_EXECUTED,
            output_validation_failed=DEFAULT_OUTPUT_VALIDATION_FAILED,
            function_tool_not_executed=DEFAULT_FUNCTION_TOOL_NOT_EXECUTED,
            tool_call_denied=DEFAULT_TOOL_CALL_DENIED,
            validation_errors_retry=DEFAULT_MODEL_RETRY,
            model_retry_string_tool=DEFAULT_MODEL_RETRY,
            model_retry_string_no_tool=DEFAULT_MODEL_RETRY,
            prompted_output_template=DEFAULT_PROMPTED_OUTPUT_TEMPLATE,
            description_template=DEFAULT_VALIDATION_FEEDBACK,
        )

        run_ctx = RunContext(deps=None, model=model, usage=RunUsage())

        toolsets = agent.toolsets
        for toolset in toolsets:
            tools = await toolset.get_tools(run_ctx)
            for tool_name, toolset_tool in tools.items():
                tool_def = toolset_tool.tool_def
                config = ToolConfig(
                    name=tool_name,
                    tool_description=tool_def.description,
                    strict=tool_def.strict,
                    tool_args_descriptions=_convert_parameters_json_schema_to_dot_notation(
                        tool_def.parameters_json_schema
                    ),
                )
                tool_config[tool_name] = config

        return PromptConfig(
            tool_config=tool_config,
            templates=prompt_templates,
        )
```

Thinking of something like this?

docs/agents.md
adtyavrdhn: Yes, I'll rewrite it once API is done.

pydantic_ai_slim/pydantic_ai/prompt_config.py
adtyavrdhn: Added the whole thing.

adtyavrdhn:
Think the PR note is due for a overhaul again :'))

pydantic_ai_slim/pydantic_ai/_agent_graph.py
adtyavrdhn: I am leaning towards using this for the typing issue, I don't want to cast or ignore all those lines otherwise although it would be fine as well. Which would you prefer?

pydantic_ai_slim/pydantic_ai/prompt_config.py
adtyavrdhn: Removed after all, wasn't necessary.

docs/output.md
```diff
@@ -351,7 +351,7 @@ While we would generally suggest starting with tool or native output, in some ca
...
 
-To use this mode, you can wrap the output type(s) in the [`PromptedOutput`][pydantic_ai.output.PromptedOutput] marker class that also lets you specify a `name` and `description` if the name and docstring of the type or function are not sufficient. Additionally, it supports an `template` argument lets you specify a custom instructions template to be used instead of the [default][pydantic_ai.profiles.ModelProfile.prompted_output_template].
+To use this mode, you can wrap the output type(s) in the [`PromptedOutput`][pydantic_ai.output.PromptedOutput] marker class that also lets you specify a `name` and `description` if the name and docstring of the type or function are not sufficient. Additionally, it supports a `template` argument that lets you specify a custom instructions template to be used instead of the [default][pydantic_ai.profiles.ModelProfile.prompted_output_template].
```
adtyavrdhn, line 409: I just fixed this small grammatical error because I saw it :'))

pydantic_ai_slim/pydantic_ai/__init__.py
```diff
@@ -95,6 +95,18 @@
     ModelProfile,
     ModelProfileSpec,
 )
+from .prompt_config import (
+    DEFAULT_FINAL_RESULT_PROCESSED,
```
DouweM, line 105: I don't love exporting so many constants here, but see your point below about having to cast fields to `str` when using `DEFAULT_PROMPT_CONFIG`. Still leaving this comment here as a sign of "let's get back to this" :)

pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -557,6 +572,8 @@ async def main():
             instructions: Optional additional instructions to use for this run.
             deps: Optional dependencies to use for this run.
             model_settings: Optional settings to use for this model's request.
+            prompt_config: Optional prompt configuration to override how system-generated parts are
+                phrased for this specific run, falling back to the agent's defaults if omitted.
```
DouweM, line 602: Reading this quickly, I could think that these get merged with the agent's `prompt_config`, which would also be reasonable behavior, if you just want to override a few things at run time, like a tool description, while you define e.g. translations for standard prompts at agent time. Could we support that?

If not, we should make it a bit clearer here that this is an all-or-nothing fallback, not a per-field fallback.

pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -1413,13 +1459,24 @@ def copy_dynamic_toolsets(toolset: AbstractToolset[AgentDepsT]) -> AbstractTools
...
 
+        if tool_config:
+            toolset = ToolConfigPreparedToolset(toolset, tool_config)
+
         output_toolset = output_toolset if _utils.is_set(output_toolset) else self._output_toolset
         if output_toolset is not None:
             if self._prepare_output_tools:
                 output_toolset = PreparedToolset(output_toolset, self._prepare_output_tools)
+            if tool_config:
+                output_toolset = ToolConfigPreparedToolset(output_toolset, tool_config)
```
DouweM: Can't we add this once at the very end, applying to the entire `toolset` just before we return it?

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -873,6 +873,16 @@ def has_content(self) -> bool:
     __repr__ = _utils.dataclasses_no_defaults_repr
 
 
+ReturnKind: TypeAlias = Literal[
```
DouweM: `ToolReturnKind` please

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -882,6 +892,18 @@ class ToolReturnPart(BaseToolReturnPart):
     part_kind: Literal['tool-return'] = 'tool-return'
     """Part type identifier, this is available on all parts as a discriminator."""
 
+    return_kind: ReturnKind | None = None
+    """How the tool call was resolved, used for disambiguating return parts.
+
+    * `tool-executed`: the tool ran successfully and produced a return value
```
DouweM: Let's explain this on the type docstring instead

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -944,20 +966,27 @@ class RetryPromptPart:
...
+    """Pre-rendered retry message. When set by PromptTemplates, model_response() returns this directly."""
+
     def model_response(self) -> str:
         """Return a string message describing why the retry is requested."""
+        # If templates were applied, return the pre-rendered message
+        if self.retry_message is not None:
+            return self.retry_message
+
+        # Default fallback when no template was applied
+        from .prompt_config import DEFAULT_MODEL_RETRY, default_validation_error, default_validation_feedback
```
DouweM: Not sure if these should all be public methods

DouweM: And if they should be, should we export them from `pydantic_ai.__init__` as well?

pydantic_ai_slim/pydantic_ai/prompt_config.py
```diff
@@ -0,0 +1,414 @@
...
+
+    {schema}
+
+    Don't include any text or Markdown fencing before or after.
+    """
+)
+"""Default template for prompted output schema instructions."""
+
+
+def default_validation_feedback(content: str | list[pydantic_core.ErrorDetails]) -> str:
```
DouweM: We can't require `str` here?

pydantic_ai_slim/pydantic_ai/prompt_config.py
```diff
@@ -0,0 +1,414 @@
...
+    """Generate a default validation feedback message."""
+    assert isinstance(content, str)
+    return f'Validation feedback:\n{content}'
+
+
+def default_validation_error(content: str | list[pydantic_core.ErrorDetails]) -> str:
+    """Generate a default validation error message from a list of Pydantic `ErrorDetails`."""
+    from .messages import error_details_ta
+
```
DouweM, line 69: We can't require `lists[...]` here?

pydantic_ai_slim/pydantic_ai/prompt_config.py
```diff
@@ -0,0 +1,414 @@
...
+    assert isinstance(content, list)
+
+    json_errors = error_details_ta.dump_json(content, exclude={'__all__': {'ctx'}}, indent=2)
+    plural = len(content) != 1
+    return f'{len(content)} validation error{"s" if plural else ""}:\n```json\n{json_errors.decode()}\n```'
+
+
+return_kind_to_default_prompt_template: dict[str, str] = {
```
DouweM: Use `ToolReturnKind` here

pydantic_ai_slim/pydantic_ai/prompt_config.py
```diff
@@ -0,0 +1,414 @@
...
+    ([`ModelProfile.prompted_output_template`][pydantic_ai.profiles.ModelProfile.prompted_output_template])
+    is used.
+    """
+
+    description_template: Callable[[str | list[pydantic_core.ErrorDetails]], str] | None = None
+    """Format a description message while asking the model to retry."""
+
+    def apply_template(self, message_part: ModelRequestPart, ctx: RunContext[Any]) -> ModelRequestPart:
+        if isinstance(message_part, ToolReturnPart):
+            if message_part.return_kind in (None, 'tool-executed'):
```
DouweM: We can check if the key is in `return_kind_to_default_prompt_template`, right?

pydantic_ai_slim/pydantic_ai/prompt_config.py
```diff
@@ -0,0 +1,414 @@
...
+                return message_part
+
+            field_name = message_part.return_kind.replace('-', '_')
+            template = getattr(self, field_name, None)
```
DouweM: I'd rather have a type safe method that directly maps from ToolReturnKind to the self.methods

pydantic_ai_slim/pydantic_ai/prompt_config.py
```diff
@@ -0,0 +1,414 @@
...
+            # Map return_kind directly to template attribute name (e.g. 'final-result-processed' -> 'final_result_processed')
+
+            # Special case for tool-denied: only apply if template is explicitly set
```
DouweM: Can you explain why we need to do this in a code comment?

pydantic_ai_slim/pydantic_ai/prompt_config.py
```diff
@@ -0,0 +1,414 @@
...
+            replace(
+                message,
+                parts=[self.apply_template(part, ctx) for part in message.parts],
+            )
+            if isinstance(message, ModelRequest)
+            else message
+            for message in _messages
+        ]
+
+    def _apply_retry_template(
```
DouweM: This method is now mirroring `RetryPromptPart.model_response()` in terms of the logic; can we unify them somehow and only have the logic in one place?

Maybe `model_response()` could take `prompt_config: PromptConfig = DEFAULT_PROMPT_CONFIG` as an arg and keep the logic there, or we could have the logic as a method on `PromptConfig` and have `model_response` call `return self.retry_message or DEFAULT_PROMPT_CONFIG.format_retry_prompt(RetryPromptPart)`

pydantic_ai_slim/pydantic_ai/prompt_config.py
```diff
@@ -0,0 +1,414 @@
...
+            # List of ErrorDetails (validation errors)
+            description_template = self.description_template or default_validation_error
+            description = description_template(content)
+            retry_template = self.validation_errors_retry_prompt or DEFAULT_MODEL_RETRY
+
+        # Resolve callable if needed
+        if callable(retry_template):
+            retry_template = retry_template(message_part, ctx)
+
+        return replace(message_part, retry_message=f'{description}\n\n{retry_template}')
```
DouweM: Should we bail out if `retry_message` has already been set? I can imagine a user wanting to do that directly, instead of setting `content`, if they don't want the extra stuff around it.

DouweM: That would warrant a better field name and docstring

pydantic_ai_slim/pydantic_ai/prompt_config.py
```diff
@@ -0,0 +1,414 @@
...
+
+    This allows you to override tool metadata without modifying the original tool definitions.
+
+    This is applied via [`PromptConfig`][pydantic_ai.PromptConfig] and can be used to rename tools,
+    override descriptions, enable strict schema behavior for supported models, and override argument
+    descriptions.
+    """
+
+    name: str | None = None
+    tool_description: str | None = None
```
DouweM: No need to repeat `tool_` in a class named `ToolConfig` :)

pydantic_ai_slim/pydantic_ai/prompt_config.py
```diff
@@ -0,0 +1,414 @@
...
+    strict: bool | None = None
+    tool_args_descriptions: dict[str, str] | None = None
```
DouweM: Trying to mirror `ToolDefinition.parameters_json_schema`, let's call this `parameters_descriptions`

DouweM: Also we need a docstring to explain the nested stuff at the very least

pydantic_ai_slim/pydantic_ai/prompt_config.py
```diff
@@ -0,0 +1,414 @@
...
+    async def generate_prompt_config_from_agent(agent: Agent[Any, Any], model: Model) -> PromptConfig:
+        """Generate a PromptConfig instance based on an Agent instance.
+
+        This fills in per-tool metadata and default templates, producing a `PromptConfig` that can be
+        used as a starting point for prompt optimizers.
+
+        """
+        tool_config: dict[str, ToolConfig] = {}
+
+        prompt_templates: PromptTemplates = PromptTemplates(
```
DouweM: I think it's worth havin g this as `DEFAULT_PROMPT_TEMPLATES`

pydantic_ai_slim/pydantic_ai/prompt_config.py
```diff
@@ -0,0 +1,414 @@
...
+    """Template for prompted output schema instructions.
+
+    If set, this overrides the template used for [`PromptedOutput`][pydantic_ai.output.PromptedOutput].
+
+    If `None`, the template from `PromptedOutput` is used if set, otherwise the model profile default
+    ([`ModelProfile.prompted_output_template`][pydantic_ai.profiles.ModelProfile.prompted_output_template])
+    is used.
+    """
+
+    description_template: Callable[[str | list[pydantic_core.ErrorDetails]], str] | None = None
```
DouweM: This name is not nearly clear enough :) I think all the retry prompt stuff should be prefixed the same

pydantic_ai_slim/pydantic_ai/prompt_config.py
```diff
@@ -0,0 +1,414 @@
...
+    """
+
+    tool_config: dict[str, ToolConfig] | None = None
+    """Configuration for customizing tool descriptions and metadata, keyed by tool name.
+    See [`ToolConfig`][pydantic_ai.ToolConfig] for available configuration options.
+    """
+
+    @staticmethod
+    async def generate_prompt_config_from_agent(agent: Agent[Any, Any], model: Model) -> PromptConfig:
+        """Generate a PromptConfig instance based on an Agent instance.
```
DouweM: Shouldnm't this use `agent.prompt_config` as well, to override the defaults?

pydantic_ai_slim/pydantic_ai/prompt_config.py
```diff
@@ -0,0 +1,414 @@
...
+            final_result_processed=DEFAULT_FINAL_RESULT_PROCESSED,
+            output_tool_not_executed=DEFAULT_OUTPUT_TOOL_NOT_EXECUTED,
+            output_validation_failed=DEFAULT_OUTPUT_VALIDATION_FAILED,
+            function_tool_not_executed=DEFAULT_FUNCTION_TOOL_NOT_EXECUTED,
+            tool_denied=DEFAULT_TOOL_CALL_DENIED,
+            validation_errors_retry_prompt=DEFAULT_MODEL_RETRY,
+            tool_retry_prompt=DEFAULT_MODEL_RETRY,
+            no_tool_retry_prompt=DEFAULT_MODEL_RETRY,
+            prompted_output_template=DEFAULT_PROMPTED_OUTPUT_TEMPLATE,
+            description_template=None,
```
DouweM: Why none?

pydantic_ai_slim/pydantic_ai/prompt_config.py
```diff
@@ -0,0 +1,414 @@
...
+        )
+
+        run_ctx = RunContext(deps=None, model=model, usage=RunUsage())
```
DouweM: This is gonna cause errors with `PreparedToolset`s, `FilteredToolset`s etc that really do depend on `deps`. So we should let an actual `deps` object to be passed in, and make the type safety work by having it as a method on agent so we can use `AgentDepsT` like w do on `run` etc.

pydantic_ai_slim/pydantic_ai/prompt_config.py
```diff
@@ -0,0 +1,414 @@
...
+
+        # Include both regular and output tools
+        from .toolsets import CombinedToolset
+
+        all_toolsets = [*agent.toolsets]
+        if output_toolset := getattr(agent, '_output_toolset', None):
```
DouweM: If we're going to be using private methods anyway, could we use `Agent._get_toolset()`?

pydantic_ai_slim/pydantic_ai/prompt_config.py
DouweM: We may actually need a new approach that outputs _all_ the tools, not just the ones that would be sent to the model, otherwise the tool defs from searchable toolset (https://github.com/pydantic/pydantic-ai/issues/3590) won't show up (as they are deferred until the model searches for them), while you'd definitely want to be able to optimize on those descriptions exactly because they can affect how well the model will be able to find them.

So we need some way on each toolset to return all the tools without any filtering.

pydantic_ai_slim/pydantic_ai/toolsets/prepared.py
```diff
@@ -34,3 +38,115 @@ async def get_tools(self, ctx: RunContext[AgentDepsT]) -> dict[str, ToolsetTool[
...
+                        'strict': config.strict,
+                    }.items()
+                    if v is not None
+                },
+            )
+
+            updated_tool = replace(tool, tool_def=updated_tool_def)
+
+            # Handle renaming: remove old key if renamed, then add with final name
+            final_tool_name = config.name if config.name is not None else tool_name
```
DouweM: To keep the renamed tools working when they are called with the new names, we need to do mapping in `call_tool` as well. Check out `RenamedToolset`. 

I think we should be able to implement this toolset as a combination of `RenamedToolset` + `PreparedToolset`, both with custom functions.

pydantic_ai_slim/pydantic_ai/prompt_config.py
```diff
@@ -0,0 +1,414 @@
...
+
+# JSON Schema keys
+_PROPERTIES = 'properties'
+_DEFS = '$defs'
+_REF = '$ref'
+_REF_PREFIX = '#/$defs/'
+_DESCRIPTION = 'description'
+
+
+def _extract_descriptions_from_json_schema(parameters_json_schema: dict[str, Any]) -> dict[str, str]:
```
DouweM: I'd like this to live closer to the code for injecting descriptions into the schema. Maybe a new class with 2 methods? Or 2 utils? Either way, side by side.

pydantic_ai_slim/pydantic_ai/ui/_event_stream.py
```diff
@@ -170,6 +170,7 @@ async def transform_stream(  # noqa: C901
                                 tool_call_id=tool_call_id,
                                 tool_name=tool_name,
                                 content='Final result processed.',
```
DouweM: Let's use the constant here

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/prompt_config.py
adtyavrdhn: Yeah that makes sense to do, I was thinking for it to only behave as a starting point but we could start off with some other PromptConfig of a previous session as well. I'll also add it in a comment and docstring.

pydantic_ai_slim/pydantic_ai/prompt_config.py
adtyavrdhn: Doesn't that expose the user to losing out on using the content and the error surrounding it if they add it only expecting the retry message at the end to change? I was hoping to kind of keep the error present forcibly and overwriting it would mean ignoring the content arg being passed through the Callable(description_message atm) which would be visible to the user?

pydantic_ai_slim/pydantic_ai/prompt_config.py
adtyavrdhn: Yeah I don't like this either, I'll change to keep it in one place. We wouldn't want to maintain same code in two files.

pydantic_ai_slim/pydantic_ai/prompt_config.py
adtyavrdhn: Yeah I figured so as well, I changed it just for you to take a look but it just looks bad. I'll revert.

pydantic_ai_slim/pydantic_ai/prompt_config.py
adtyavrdhn: Type checker complains for the assert never statement, presumably because based on the dict look up it cannot know if tool-executed is covered or not. Safe to ignore it from our end knowing that is what it means for it to not be present I think

pydantic_ai_slim/pydantic_ai/agent/__init__.py
adtyavrdhn: No I agree, given that the entire premise is overriding. We should be sticking to that behaviour whenever possible.

pydantic_ai_slim/pydantic_ai/prompt_config.py
adtyavrdhn: Isn't it used only for the case where the content is a string, I added this assert to satisfy the type checker otherwise we would need two templates for when content is a string or list of errors? Or do you mean something else altogether?

pydantic_ai_slim/pydantic_ai/prompt_config.py
adtyavrdhn: Haha yes.

pydantic_ai_slim/pydantic_ai/prompt_config.py
adtyavrdhn: Yes I was thinking about it as well. Can also maybe add a roundtrip test to verify.

pydantic_ai_slim/pydantic_ai/prompt_config.py
adtyavrdhn: Yes I think it should be obvious that you get to format and display the retry error content with this, let me think about it.

pydantic_ai_slim/pydantic_ai/agent/__init__.py
adtyavrdhn: I am thinking about this merge more and I think the best behaviour would be to go down and nest as deep as we can recursively for both the prompt_configs to merge the conflicts and copy the others on the top layer where it is common. That would be pretty sweet and very useful for people using these optimizers.

adtyavrdhn: Issue is I am not sure if I can write it in a generic way and that would still be compatible when we expand PromptConfig with more features. I'll think about this more.

pydantic_ai_slim/pydantic_ai/prompt_config.py
adtyavrdhn: > This is gonna cause errors with PreparedToolsets, FilteredToolsets etc that really do depend on deps. So we should let an actual deps object to be passed in, and make the type safety work by having it as a method on agent so we can use AgentDepsT like w do on run etc.

Oh yep, makes sense to move it to deps aware Agent. 

About all toolsets, I'll read the toolsets and how they operate more then decide how I want to implement no filtering get_tools()

adtyavrdhn: Instead of returning all tools, leaning towards returning all tool defs instead because that is what we optimize on and the runtime details are not sent to the model anyway. It should be more efficient as well.

One thing I am a bit confused about is prepared toolset behaviour, we cannot add tools but we can remove them, I don't want to bypass it completely because it could have important tool_def changes, raising an error also doesn't seem okay because we do allow the same prepare_func for get_tools().

pydantic_ai_slim/pydantic_ai/toolsets/prepared.py
adtyavrdhn: Yes, removed the new ToolSet and moved to using this combination instead.

pydantic_ai_slim/pydantic_ai/prompt_config.py
adtyavrdhn: It can hold one of the two default callables which is resolved with context, cannot keep one of those as default here without any context.

adtyavrdhn: Added a default function to resolve via one of these two based on the content.

adtyavrdhn:
Ping!

Issue #2600, opened 2025-08-18
Anthropic `stop_reason` `pause_turn` is not handled correctly, resulting in errors with long-running built-in tools


eballesteros:
### Question

I ran into a tricky issue using Anthropic's web_search with Pydantic AI that took some debugging to figure out. I wanted to share my findings and the workaround I came up with, in case this may help others who hit the same issue. I'm also curious if anyone has found better approaches, or if this is something the library could handle internally.

The issue: When using the web search builtin_tools with Anthropic models, and setting an output_type, I was getting `ModelHTTPError: status_code: 400, model_name: claude-sonnet-4-20250514, body: {'type': 'error', 'error': {'type': 'invalid_request_error', 'message': 'messages.1: 'web_search' tool use with id '<some-id>' was found without a corresponding 'web_search_tool_result' block'}}` fairly often. Capturing the messages when this happened, they would look something like this:

```python
[
    ModelRequest(
        parts=[
            UserPromptPart(content='...', timestamp=...)
        ], 
        instructions='...'),
    ModelResponse(
        parts=[
            # succession of BuiltinToolCallPart, BuiltinToolReturnPart and TextPart blocks
            # ... ,
            BuiltinToolCallPart(tool_name='web_search', args={'query': '...'}, tool_call_id='some-id', provider_name='anthropic')
        ], 
        usage=Usage(...)
    ),
    ModelRequest(
        parts=[
            RetryPromptPart(content='Plain text responses are not permitted, please include your response in a tool call', tool_call_id=..., timestamp=...)
        ]
    )
]
```

My understanding of what's happening is the following: The last part in that `ModelResponse` in the middle is a `BuiltinToolCallPart`, without an associated `BuiltinToolReturnPart` (a query without it's answer). The server seems to be returning a response before it's finished. This results in a `ModelRetry`, since I've set `output_type` and that last part isn't a final output tool part. But when the framework tries to send this followup request, it's malformed (there's a "'web_search' tool use ... without a corresponding 'web_search_tool_result'"), so the server returns 400 and we get the `ModelHTTPError`.

My workaround: Use `history_processors` to detect that this is happening and "rewrite" the message history before the retry.

```python
def remove_hanging_search(messages: list[ModelMessage]) -> list[ModelMessage]:
    """Remove incomplete web searches and provide better retry guidance."""
    if (
        len(messages) >= 2
        and isinstance(last_message := messages[-1], ModelRequest)
        and len(last_message.parts) == 1
[...]

PR #3051, opened 2025-09-30
Fix Anthropic pause_turn handling to allow agent continuation

kazmer97:
## Summary
Map Anthropic's `pause_turn` finish reason to `None` instead of `'stop'` so the agent automatically continues with a new request when this finish reason is encountered.

## Background
When using Anthropic's builtin tools (like `web_search`) with `output_type` set, long-running operations may trigger the `pause_turn` finish reason. Previously, this was mapped to `'stop'`, causing the agent to halt execution and attempt a retry. This led to malformed requests with errors like:

```
ModelHTTPError: status_code: 400, model_name: claude-sonnet-4-20250514, 
body: {'type': 'error', 'error': {'type': 'invalid_request_error', 
'message': "messages.1: 'web_search' tool use with id '<some-id>' was found 
without a corresponding 'web_search_tool_result' block"}}
```

## Changes
- Updated `_FINISH_REASON_MAP` in `pydantic_ai_slim/pydantic_ai/models/anthropic.py` to map `'pause_turn'` to `None`
- Updated the type annotation to `dict[BetaStopReason, FinishReason | None]` to allow `None` values
- Added test `test_pause_turn_finish_reason_mapping` to verify the mapping

## How it works
When `finish_reason` is `None`, the agent treats the response as incomplete and automatically makes another request with the same message history, allowing the conversation to continue seamlessly without manual intervention.

## Testing
- All existing Anthropic model tests pass (51/51)
- New test verifies the `pause_turn` mapping is correct

Fixes #2600

PR #3661, opened 2025-12-07
Handle Anthropic `pause_turn` finish reason by resubmitting request with response so far

chetanreddyv:
- Fixes https://github.com/pydantic/pydantic-ai/issues/2600
- Replaces https://github.com/pydantic/pydantic-ai/pull/3051

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/anthropic.py
+++ b/pydantic_ai_slim/pydantic_ai/models/anthropic.py
@@ -69,6 +69,7 @@ try:
         BetaCacheControlEphemeralParam,
         BetaCitationsConfigParam,
         BetaCitationsDelta,
+        BetaCodeExecutionTool20250522Param,
         BetaCodeExecutionToolResultBlock,
         BetaCodeExecutionToolResultBlockContent,
         BetaCodeExecutionToolResultBlockParam,
@@ -81,6 +82,7 @@ try:
         BetaMCPToolResultBlock,
         BetaMCPToolUseBlock,
         BetaMCPToolUseBlockParam,
+        BetaMemoryTool20250818Param,
         BetaMessage,
         BetaMessageParam,
         BetaMessageTokensCount,
@@ -157,8 +159,6 @@ class AnthropicModelSettings(ModelSettings, total=False):
 
     Contains `user_id`, an external identifier for the user who is associated with the request.
     """
-
-
 
     anthropic_thinking: BetaThinkingConfigParam
     """Determine whether the model should generate a thinking block.
@@ -409,21 +409,19 @@ class AnthropicModel(Model):
                     extra_headers=extra_headers,
                     extra_body=model_settings.get('extra_body'),
                 )
-                
-                # If streaming, return immediately
-                if stream:
-                    return response
-                
+
                 # Handle pause_turn for non-streaming
                 assert isinstance(response, BetaMessage)
                 if response.stop_reason == 'pause_turn':
                     # Append assistant message to history and continue
-                    anthropic_messages.append({
-                        'role': 'assistant',
-                        'content': response.content,
-                    })
+                    anthropic_messages.append(
+                        {
+                            'role': 'assistant',
+                            'content': response.content,
+                        }
+                    )
                     continue
-                
+
                 return response
 
             except APIStatusError as e:
@@ -501,8 +499,6 @@ class AnthropicModel(Model):
         except APIConnectionError as e:
             raise ModelAPIError(model_name=self.model_name, message=e.message) from e
 
-
-
     def _process_response(self, response: BetaMessage) -> ModelResponse:
         """Process a non-streamed response, and prepare a message to return."""
         items: list[ModelResponsePart] = []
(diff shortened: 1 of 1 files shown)


pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -49,7 +49,7 @@
     'max_tokens': 'length',
     'stop_sequence': 'stop',
     'tool_use': 'tool_call',
-    'pause_turn': 'stop',
+    'pause_turn': 'stop',  # TODO: should this be a different finish reason?
```
DouweM, line 52: I think we can remove it here because we never end up returning a `ModelResponse` from `pause_turn`, right? As we always resubmit the request

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -385,33 +385,51 @@ async def _messages_create(
...
+                )
+
+                # Handle pause_turn for non-streaming
+                assert isinstance(response, BetaMessage)
+                if response.stop_reason == 'pause_turn':
+                    # Append assistant message to history and continue
+                    anthropic_messages.append(
+                        {
+                            'role': 'assistant',
+                            'content': response.content,
```
DouweM, line 420: Similar to https://github.com/pydantic/pydantic-ai/pull/3637, should we also send back `response.container.id` so that it can be reused?

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -385,33 +385,51 @@ async def _messages_create(
...
+                        }
+                    )
+                    continue
+
+                return response
```
DouweM, line 425: We're currently losing the initial `response.usage`. We have to make sure that this method (or really the `request` method) has the combined usage of both responses on it

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -385,33 +385,51 @@ async def _messages_create(
...
+                    temperature=model_settings.get('temperature', OMIT),
+                    top_p=model_settings.get('top_p', OMIT),
+                    timeout=model_settings.get('timeout', NOT_GIVEN),
+                    metadata=model_settings.get('anthropic_metadata', OMIT),
+                    extra_headers=extra_headers,
+                    extra_body=model_settings.get('extra_body'),
+                )
+
+                # Handle pause_turn for non-streaming
+                assert isinstance(response, BetaMessage)
```
DouweM, line 414: This currently breaks streaming entirely. We have to make this work for that scenario as well.

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -512,6 +530,16 @@ def _process_response(self, response: BetaMessage) -> ModelResponse:
...
+                # Fallback for new block types like `bash_code_execution_tool_result` if they aren't explicitly typed yet
+                # or if we want to handle them generically.
+                # For now, we'll try to handle `bash_code_execution_tool_result` if it appears as a dict or unknown type,
+                # but since `response.content` is typed as a union of specific blocks, we might need to rely on `model_dump` or similar if the SDK doesn't support it yet.
+                # However, the user request says "Handle the bash_code_execution_tool_result event type".
+                # If `anthropic` SDK doesn't have it, we might not see it here unless we upgrade or it's in `BetaContentBlock`.
+                # Assuming `BetaCodeExecutionToolResultBlock` covers it or we need to add a check.
+                # Let's assume for now `BetaCodeExecutionToolResultBlock` is sufficient or we'll see.
+                # But wait, `bash_code_execution_tool_result` implies a specific type.
+                # Let's check if we can import it.
```
DouweM, line 542: This comment doesn't seem necessary

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -1287,6 +1316,14 @@ def _map_server_tool_use_block(item: BetaServerToolUseBlock, provider_name: str)
             args=cast(dict[str, Any], item.input) or None,
             tool_call_id=item.id,
         )
+    elif item.name == 'bash_code_execution':
```
DouweM, line 1319: Why do we need this now if we didn't need it before?

tests/models/test_anthropic_pause_turn.py
```diff
@@ -0,0 +1,62 @@
...
+        BetaUsage,
+        BetaMessage,
+    )
+
+pytestmark = [
+    pytest.mark.skipif(not imports_successful(), reason='anthropic not installed'),
+    pytest.mark.anyio,
+]
+
+async def test_pause_turn_retry_loop(allow_model_requests: None):
```
DouweM, line 25: Please add this to the existing `test_anthropic.py`

tests/models/test_anthropic_pause_turn.py
```diff
@@ -0,0 +1,62 @@
...
+    # Mock a sequence of responses:
+    # 1. pause_turn response
+    # 2. final response
+    
+    c1 = completion_message(
+        [BetaTextBlock(text='paused', type='text')],
+        usage=BetaUsage(input_tokens=10, output_tokens=5),
+    )
+    c1.stop_reason = 'pause_turn' # type: ignore
```
DouweM, line 34: Instead of faking it, can we record a real interaction that has a `pause_turn` response? Note that in https://github.com/pydantic/pydantic-ai/issues/2600#issuecomment-3294057060 I mentioned that I reproduced the issue. We should reproduce it in a test, to make sure that bug is now fixed.

DouweM (changes requested):

PR #3662, opened 2025-12-07
Mistral: implement native output mode for structured outputs

nachollorca:
As noted in some [TODOs](https://github.com/pydantic/pydantic-ai/blob/8d111fec483038582cbcf5f0afe45b847480c6ce/pydantic_ai_slim/pydantic_ai/models/mistral.py#L299), Mistral has been supporting [constrained json grammar](https://docs.mistral.ai/capabilities/structured_output/custom) for some time now. Following Mistral 3 release last week, I wanted to check how well it does. This PR integrates it into pydantic-ai:

- Created `MistralJsonSchemaTransformer` for compatibility with Mistral's API
- Set `supports_json_object_output` to `True` 🥳 
- Added pseudoprivate methods to handle json responses. Here I tried to mimic how `pydantic_ai_slim/pydantic_ai/models/google.py` goes
- Added some unit tests for the new feature

I've been playing around with it and it looks fine so far. Please let me know of any changes u'd like to see <3


nachollorca:
I will fix the tests ASAP

pydantic_ai_slim/pydantic_ai/models/mistral.py
```diff
@@ -215,6 +218,32 @@ async def request_stream(
...
+
+        Returns None if no special format is needed.
+        """
+        if model_request_parameters.output_mode == 'native':
+            # Use native JSON schema mode
+            output_object = model_request_parameters.output_object
+            assert output_object is not None
+            json_schema = self._map_json_schema(output_object)
+            return MistralResponseFormat(type='json_schema', json_schema=json_schema)
+        elif model_request_parameters.output_mode == 'prompted' and not model_request_parameters.function_tools:
```
DouweM, line 232: I don't think we need `not model_request_parameters.function_tool`

pydantic_ai_slim/pydantic_ai/models/mistral.py
```diff
@@ -215,6 +218,32 @@ async def request_stream(
...
+        else:
+            return None
+
+    def _map_json_schema(self, o: OutputObjectDefinition) -> MistralJSONSchema:
+        """Map OutputObjectDefinition to Mistral JSONSchema format."""
+        return MistralJSONSchema(
+            name=o.name or 'output',
+            schema_definition=o.json_schema,
+            description=o.description or UNSET,
+            strict=o.strict if o.strict is not None else None,
```
DouweM, line 244: This is just `strict=o.strict` right?

pydantic_ai_slim/pydantic_ai/models/mistral.py
```diff
@@ -227,13 +256,25 @@ async def _completions_create(
         if model_request_parameters.builtin_tools:
             raise UserError('Mistral does not support built-in tools')
 
+        # Determine the response format based on output mode
+        response_format = self._get_response_format(model_request_parameters)
+
+        # When using native JSON schema mode, don't use tool-based output
```
DouweM, line 262: I don't think we need to branch on this: if `model_request_parameters.output_mode == 'native'`, `model_request_parameters.output_tools` will be empty anyway. Note that we don't have logic like this in the other models

pydantic_ai_slim/pydantic_ai/models/mistral.py
```diff
@@ -258,57 +299,41 @@ async def _stream_completions_create(
...
-                response_format={
-                    'type': 'json_object'
-                },  # TODO: Should be able to use json_schema now: https://docs.mistral.ai/capabilities/structured-output/custom_structured_output/, https://github.com/mistralai/client-python/blob/bc4adf335968c8a272e1ab7da8461c9943d8e701/src/mistralai/extra/utils/response_format.py#L9
-                stream=True,
-                http_headers={'User-Agent': get_user_agent()},
-            )
+        # Determine the response format based on output mode
+        response_format = self._get_response_format(model_request_parameters)
 
+        # When using native JSON schema mode, don't use tool-based output
```
DouweM, line 312: See above; I don't think we need this. And if we do, we should dedupe it between the request/request_stream methods :)

pydantic_ai_slim/pydantic_ai/models/mistral.py
DouweM, line 1: There are a few `TODO`s in the file that we should address at the same time:
- `# TODO: Should be able to use json_schema` 
- `# TODO: Port to native "manual JSON" mode`

These are related to an old approach for handling `output_tools` by passing the schemas to the API as user text parts, which I don't think we need anymore, so we should implement the `tool`, `native`, and `prompted` modes the way the other models do.

DouweM (changes requested):

PR #3663, opened 2025-12-07
Pass `s3://` file URLs directly to API in `BedrockConverseModel`

### LINKED ISSUE #3621: Pass `s3://` file URLs directly to API in `BedrockConverseModel`
See https://docs.aws.amazon.com/bedrock/latest/userguide/conversation-inference-call.html, under Request > `messages` > `video`:

> You can also specify an Amazon S3 URI instead of passing the bytes directly in the request body. The following shows a sample Message object with a content array containing the source passed through an Amazon S3 URI.
> 
> ```json
> {
>     "role": "user",
>     "content": [
>         {
>             "video": {
>                 "format": "mp4",
>                 "source": {
>                     "s3Location": {
>                         "uri": "s3://amzn-s3-demo-bucket/myVideo",
>                         "bucketOwner": "111122223333"
>                     }
>                 }
>             }
>         }
>     ]
> }
> ```
> 
> The assumed role must have the `s3:GetObject` permission to the Amazon S3 URI. The `bucketOwner` field is optional but must be specified if the account making the request does not own the bucket the Amazon S3 URI is found in. For more information, see Configure access to Amazon S3 buckets.

We’d just need to edit this: 

https://github.com/pydantic/pydantic-ai/blob/4ec0b32d3e82cb740c38ec6d7e3eb55ff5058ad5/pydantic_ai_slim/pydantic_ai/models/bedrock.py#L645-L652

We already support `gs://` URLs with `GoogleModel` as well, as documented on https://ai.pydantic.dev/input.

Also related to:
- #2574
- https://github.com/pydantic/pydantic-ai/pull/2611

mochow13:
Closes #3621


pydantic_ai_slim/pydantic_ai/models/bedrock.py
```diff
@@ -643,20 +643,25 @@ async def _map_user_prompt(part: UserPromptPart, document_count: Iterator[int])
...
+                        source = {'s3Location': {'uri': item.url}}
+                    else:
+                        downloaded_item = await download_item(item, data_format='bytes', type_format='extension')
+                        source = {'bytes': downloaded_item['data']}
+
                     if item.kind == 'image-url':
                         format = item.media_type.split('/')[1]
                         assert format in ('jpeg', 'png', 'gif', 'webp'), f'Unsupported image format: {format}'
-                        image: ImageBlockTypeDef = {'format': format, 'source': {'bytes': downloaded_item['data']}}
+                        image: ImageBlockTypeDef = {'format': format, 'source': cast(Any, source)}
```
DouweM: Instead of casting this to `Any`, can we fix the `source` type hint to be `DocumentSourceTypeDef`?

pydantic_ai_slim/pydantic_ai/models/bedrock.py
```diff
@@ -643,20 +643,25 @@ async def _map_user_prompt(part: UserPromptPart, document_count: Iterator[int])
...
                         content.append({'image': image})
 
                     elif item.kind == 'document-url':
                         name = f'Document {next(document_count)}'
                         document: DocumentBlockTypeDef = {
                             'name': name,
                             'format': item.format,
-                            'source': {'bytes': downloaded_item['data']},
+                            'source': cast(Any, source),
```
DouweM: Same as up

tests/models/test_bedrock.py
```diff
@@ -729,6 +729,45 @@ async def test_text_document_url_input(allow_model_requests: None, bedrock_provi
 """)
 
 
+@pytest.mark.vcr()
+async def test_s3_image_url_input(allow_model_requests: None, bedrock_provider: BedrockProvider):
+    """Test that s3:// image URLs are passed directly to Bedrock API without downloading."""
+    m = BedrockConverseModel('us.amazon.nova-pro-v1:0', provider=bedrock_provider)
+    agent = Agent(m, system_prompt='You are a helpful chatbot.')
+    image_url = ImageUrl(url='s3://my-bucket/images/test-image.jpg', media_type='image/jpeg')
```
DouweM, line 745: I'm guessing these are not real files in S3 and the cassettes were manually modified? That'll get us into trouble if we re-record the cassettes :) So I'd rather either use a _real_ S3 URL (may be hard), _or_ directly test the result of `_map_messages` as we do in other tests already

pydantic_ai_slim/pydantic_ai/models/bedrock.py
```diff
@@ -643,20 +643,25 @@ async def _map_user_prompt(part: UserPromptPart, document_count: Iterator[int])
                     else:
                         raise NotImplementedError('Binary content is not supported yet.')
                 elif isinstance(item, ImageUrl | DocumentUrl | VideoUrl):
-                    downloaded_item = await download_item(item, data_format='bytes', type_format='extension')
-                    format = downloaded_item['data_type']
+                    source: dict[str, Any]
+                    if item.url.startswith('s3://'):
+                        source = {'s3Location': {'uri': item.url}}
```
DouweM: There's also a `bucketOwner` field that users may want to set. Maybe we can tell them to encode it as a query param on the URL, and parse it out here?

pydantic_ai_slim/pydantic_ai/models/bedrock.py
```diff
@@ -643,20 +643,25 @@ async def _map_user_prompt(part: UserPromptPart, document_count: Iterator[int])
...
+                    else:
+                        downloaded_item = await download_item(item, data_format='bytes', type_format='extension')
```
DouweM, line 747: `download_item` currently has logic gating for `gs://` URLs; let's check `s3://` URLs there as well

pydantic_ai_slim/pydantic_ai/models/bedrock.py
DouweM, line 1: We should document this feature in `input.md`. At the bottom there's already a section on uploaded files to Google, can you mention S3 files + `BedrockConverseModel` there as well please?

pydantic_ai_slim/pydantic_ai/models/bedrock.py
mochow13: It seems the existing code in `download_item` checks for `gs://` and `youtube` URLs:

```
    if item.url.startswith('gs://'):
        raise UserError('Downloading from protocol "gs://" is not supported.')
    elif isinstance(item, VideoUrl) and item.is_youtube:
        raise UserError('Downloading YouTube videos is not supported.')
```

What check do you mean for `s3://` here?

pydantic_ai_slim/pydantic_ai/models/bedrock.py
mochow13: Do you mean something like `s3://my-bucket/key?bucketOwner=owner`?

tests/models/test_bedrock.py
mochow13: Yeah cassettes were manual (with help of AI of course). I have updated the tests to directly test `_map_messages` result. Not sure whether I should update the cassettes.

pydantic_ai_slim/pydantic_ai/models/bedrock.py
mochow13: Updated

pydantic_ai_slim/pydantic_ai/models/bedrock.py
mochow13: Updated

tests/models/test_bedrock.py
DouweM: I think the cassettes can be deleted then

pydantic_ai_slim/pydantic_ai/models/bedrock.py
DouweM: Yep that's what I was thinking

pydantic_ai_slim/pydantic_ai/models/bedrock.py
DouweM: Same check raising an error saying that `download_item` does not support `s3://` URLs

mochow13: Ah ok need to stop supporting download altogether. Updated. 
Kept the check pretty simple. Not sure whether we should go for a proper url parsing here since the expectation is just `bucketOwner` param.

DouweM: If it's not drastically more code, I'd prefer proper URL parsing

pydantic_ai_slim/pydantic_ai/models/bedrock.py
```diff
@@ -733,20 +734,28 @@ async def _map_user_prompt(  # noqa: C901
...
+                        source = {'s3Location': s3_location}
+                    else:
+                        downloaded_item = await download_item(item, data_format='bytes', type_format='extension')
+                        source = {'bytes': downloaded_item['data']}
+
                     if item.kind == 'image-url':
                         format = item.media_type.split('/')[1]
                         assert format in ('jpeg', 'png', 'gif', 'webp'), f'Unsupported image format: {format}'
-                        image: ImageBlockTypeDef = {'format': format, 'source': {'bytes': downloaded_item['data']}}
+                        image: ImageBlockTypeDef = {'format': format, 'source': cast(DocumentSourceTypeDef, source)}
```
DouweM: We shouldn't need to `cast` if hint the type of `source` to be `source: DocumentSourceTypeDef`

pydantic_ai_slim/pydantic_ai/models/bedrock.py
mochow13: @DouweM I have updated the doc. Added a paragraph for S3 + `BedrockConverseModel`. Also updated the section above to mention that S3 files will not be downloaded.

Looking at the doc, I am a bit confused. Given that we have updated `download_item` to skip downloading `s3://` URLs altogether, won't it apply to all models? For example, if we pass a `s3://` URL to a model that doesn't support downloading itself, because of our check, Pydantic AI will also stop downloading for that particular model, right?

Not sure if I question is clear enough. Basically, in this MR, we are passing the `s3://` URL directly to `BedrockConverseModel` but we have updated `downloaed_item` function to raise an error if `s3://` is passed. This would mean we will stop downloading from `s3://`URLs for other models, no?

DouweM: @mochow13 `download_item` is already not capable of downloading `s3://` URLs, but now that we officially support them for one model, people are more likely to pass them to non-supporting models, so out of consistency with what we did for `gs://`, I thought it'd be a good idea to add an explicit error.

docs/input.md
```diff
@@ -138,3 +138,5 @@ result = agent.run_sync(
 )
 print(result.output)
 ```
+
+For `BedrockConverseModel`, you can upload a file to S3 and pass the URL to the API directly. Pyadantic AI will pass it to `bedrock` as is. It expects the URL with the format: `s3://<bucket-name>/<object-key>`. An optional `bucketOwner` query parameter can be added to the URL to specify the owner of the bucket. For example: `s3://my-bucket/my-file.png?bucketOwner=123456789012`.
```
DouweM:
```suggestion
`BedrockConverseModel` supports `s3://<bucket-name>/<object-key>` URIs, provided that the assumed role has the `s3:GetObject` permission. An optional `bucketOwner` query parameter must be specified if the bucket is not owned by the account making the request. For example: `s3://my-bucket/my-file.png?bucketOwner=123456789012`.
```

pydantic_ai_slim/pydantic_ai/models/bedrock.py
```diff
@@ -11,6 +11,7 @@
 
 import anyio.to_thread
 from botocore.exceptions import ClientError
+from mypy_boto3_bedrock_runtime.type_defs import DocumentSourceTypeDef
```
DouweM: Let's move this line into the `if TYPE_CHECKING:` block

pydantic_ai_slim/pydantic_ai/models/bedrock.py
mochow13: Not drastically more code. Used Python's `urlib`.

pydantic_ai_slim/pydantic_ai/models/bedrock.py
mochow13: Good catch, I only replaced `any` with the type previously 🤦  
Updated to use proper hints.

PR #3671, opened 2025-12-08
Fix error when OpenRouter response includes file annotations

### LINKED ISSUE #3657: OpenRouter gpt-5.1-codex models returns annotations: [{type: "file"}] in Chat Completions, causing UnexpectedModelBehavior in pydantic-ai 1.27.0 (expects only url_citation)
### Initial Checks

### Description

# Pydantic AI + OpenRouter annotation validation failure

## Title
OpenRouter `gpt-5.1-codex-mini` returns `annotations: [{type: "file"}]` in Chat Completions, causing `UnexpectedModelBehavior` in pydantic-ai 1.27.0 (expects only `url_citation`)

## Summary
When using Pydantic AI with OpenRouter (chat completions, not Responses API) and model `gpt-5.1-codex-mini`, the provider returns `message.annotations` entries with `type: "file"` and `url_citation: null`. The OpenAI ChatCompletion schema in `pydantic_ai`/`openai` only allows `url_citation`, so validation fails with `UnexpectedModelBehavior`.

## Environment
- pydantic-ai 1.27.0 / pydantic-ai-slim 1.27.0  
- openai 2.9.0  
- Provider: OpenRouter (OpenAI-compatible), model `gpt-5.1-codex-mini`  
- Using Chat Completions via `OpenAIChatModel` + `OpenRouterProvider`

## Steps to Reproduce
1) Configure OpenRouter provider and call `Agent` with model `gpt-5.1-codex-mini`, passing files so the model emits annotations.  
2) Provider returns ChatCompletion containing:
   ```json
   "annotations": [
     {
       "type": "file",
       "start_index": ...,
       "end_index": ...,
       "url_citation": null
     }
   ]
   ```
3) Pydantic raises:
   ```
   UnexpectedModelBehavior: Invalid response from OpenAI chat completions endpoint:
   choices.0.message.annotations.0.type Input should be 'url_citation'
   choices.0.message.annotations.0.url_citation Input should be a valid dictionary...
   ```

## Expected
Pydantic AI should not crash on provider responses that include non-`url_citation` annotations. Options: extend schema to accept `file` annotations (as emitted by OpenRouter) or ignore/strip unknown annotation types for OpenRouter Chat Completions.

## Actual
Validation fails in `ChatCompletion.model_validate`, aborting the run before tool calls.

## Ask
- Add support for OpenRouter’s `file` annotations (or relax/strip unknown annotations for OpenRouter provider in Chat Completions).  
- Alternatively, provide a setting to bypass/ignore annotation validation for OpenRouter.

### Example Code

```Python
## Minimal Repro Script

# Requires: OPENROUTER_API_KEY set
import asyncio
from pathlib import Path
from tempfile import NamedTemporaryFile

from pydantic_ai import Agent
from pydantic_ai.messages import BinaryContent, UserPromptPart

# Create a tiny dummy PDF to ensure a file is sent
tmp = NamedTemporaryFile(suffix=".pdf", delete=False)
tmp.write(b"%PDF-1.1\n1 0 obj\n<<>>\nendobj\ntrailer\n<<>>\n%%EOF")
tmp.flush()
pdf_path = Path(tmp.name)

async def main():
    agent = Agent("openrouter:openai/gpt-5.1-codex-mini")
    # Passing a file triggers OpenRouter to emit an annotation with type="file"
    await agent.run(
        UserPromptPart(
            content=[
                "Summarize this file.",
                BinaryContent.from_path(pdf_path),
            ]
        )
    )

asyncio.run(main())

Expected: succeeds.  
Actual: `UnexpectedModelBehavior` due to `annotations[0].type == "file"` failing validation (only `url_citation` allowed).
```

```Text
- pydantic-ai 1.27.0 / pydantic-ai-slim 1.27.0  
- openai 2.9.0  
- Provider: OpenRouter (OpenAI-compatible), model `gpt-5.1-codex-mini` and `gpt-5.1-codex` and `gpt-5.1` 
- Using Chat Completions via `OpenAIChatModel` + `OpenRouterProvider`
```

dsfaccini:
Fixes #3657 

New `_OpenRouterAnnotation: TypeAlias = _OpenAIAnnotation | _OpenRouterFileAnnotation` is now able to validate file annotations

For reference this is what the annotation looks like https://logfire-us.pydantic.dev/public-trace/08504127-711c-433a-80d1-6e74034ae99f?spanId=bdf6140a9d7f98b2
image

the array of `['<file name="filename.pdf">', '</file>']` seems to me like someone is doing something they shouldn't somewhere, but I can't pinpoint this specific schema to any sdk right now, which is why I left the field typing as `dict[str, Any]`


tests/models/test_openrouter.py
```diff
@@ -470,3 +471,85 @@ def insert_level_with_spaces(level: InsertLevelArg | None, spaces: list[SpaceArg
...
+    result = await agent.run(
+        user_prompt=[
+            'What does this PDF contain? Answer in one short sentence.',
+            document_content,
+        ]
+    )
+
+    # The response should contain text (model may or may not include file annotations)
+    assert isinstance(result.output, str)
+    assert len(result.output) > 0
```
DouweM, line 522: Looking at the cassette, it doesn't actually include file annotations 😄 If we can't force the API to do that, we can just drop this test and only to the one below

dsfaccini: changed it to `5.1-codex`, that's the one that produces the file annotations

PR #3672, opened 2025-12-08
Add support for aspect ratio in gemini image generation

ajac-zero:
Finishes #3412 that has gone stale 😿

These changes will be used to complete #3599

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/openrouter.py
+++ b/pydantic_ai_slim/pydantic_ai/models/openrouter.py
@@ -261,21 +261,21 @@ class _ReasoningSummary(_BaseReasoningDetail, frozen=True):
     """Represents a high-level summary of the reasoning process."""
 
     type: Literal['reasoning.summary']
-    summary: str
+    summary: str = ''
 
 
 class _ReasoningEncrypted(_BaseReasoningDetail, frozen=True):
     """Represents encrypted reasoning data."""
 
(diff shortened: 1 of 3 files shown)


tests/models/test_openai_responses.py
```diff
@@ -128,6 +130,37 @@ async def test_openai_responses_image_detail_vendor_metadata(allow_model_request
...
+        ('3:2', 'auto', '1536x1024'),
+    ],
+)
+def test_openai_responses_image_generation_tool_aspect_ratio_mapping(
+    aspect_ratio: ImageAspectRatio,
+    explicit_size: Literal['1024x1024', '1024x1536', '1536x1024', 'auto'],
+    expected_size: Literal['1024x1024', '1024x1536', '1536x1024'],
+) -> None:
+    tool = ImageGenerationTool(aspect_ratio=aspect_ratio, size=explicit_size)
+    assert _resolve_openai_image_generation_size(tool) == expected_size
```
Kludex, line 147: @DouweM I suggest we don't include tests that import private methods. It gets hard to refactor.

DouweM: @Kludex I prefer not to but it's sometimes a lot easier to unit test something than to rely only on the public API surface, and I don't think this method should be public. I think of this as similar to using mocks,, not ideal but acceptable in tests

PR #3676, opened 2025-12-09
Let `TemporalAgent`s be registered to a Temporal workflow using `__pydantic_ai_agents__` field

### LINKED ISSUE #3650: Simplify registering `TemporalAgent`s with a Temporal worker via the workflow
### Description

Right now, `TemporalAgent`s need to be registered with a Temporal worker one by one using the `AgentPlugin`:

```py
import uuid

from temporalio import workflow
from temporalio.client import Client
from temporalio.worker import Worker

from pydantic_ai import Agent
from pydantic_ai.durable_exec.temporal import (
    AgentPlugin,
    PydanticAIPlugin,
    TemporalAgent,
)

agent = Agent(
    'gpt-5',
    instructions="You're an expert in geography.",
    name='geography',  
)

temporal_agent = TemporalAgent(agent)  


@workflow.defn
class GeographyWorkflow:  
    @workflow.run
    async def run(self, prompt: str) -> str:
        result = await temporal_agent.run(prompt)  
        return result.output


async def main():
    client = await Client.connect(  
        'localhost:7233',  
        plugins=[PydanticAIPlugin()],  
    )

    async with Worker(  
        client,
        task_queue='geography',
        workflows=[GeographyWorkflow],
        plugins=[AgentPlugin(temporal_agent)],  
    ):
        output = await client.execute_workflow(  
            GeographyWorkflow.run,
            args=['What is the capital of Mexico?'],
            id=f'geography-{uuid.uuid4()}',
            task_queue='geography',
        )
        print(output)
        #> Mexico City (Ciudad de México, CDMX)
```

A user has pointed out that it would be more convenient to register `TemporalAgent`s with the workflow that uses them via a special field like `pydantic_ai_agents` (possibly with double underscores?), and then for them (or more precisely: their activities) to be registered with the worker automatically when the workflow is registered with that worker:

```py
@workflow.defn
class GeographyWorkflow:  
    __pydantic_ai_agents__ = [temporal_agent]

    @workflow.run
    async def run(self, prompt: str) -> str:
        result = await temporal_agent.run(prompt)  
        return result.output
```

On the `async with Worker(...)` side, this could either look like a new `plugins=[AgentsPlugin()]`, or we could possibly implement this in the existing `PydanticAIPlugin` so that nothing needs to be passed to the worker at all.

To make this type-safe, workflow classes would need to inherit from a class we provide that has the `__pydantic_ai_agents__` field (or class method?) typed as `Sequence[TemporalAgent]`. 

But for convenience, I don't want to _require_ the workflow class to subclass that, so we should also do a `getattr(workflow_cls, '__pydantic_ai_agents__', None)` and check the types at runtime.

adtyavrdhn:
Closes #3650 

I went in favor of adding it to PydanticAIPlugin() because I think it is cleaner that way.

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/durable_exec/temporal/__init__.py
+++ b/pydantic_ai_slim/pydantic_ai/durable_exec/temporal/__init__.py
@@ -1,8 +1,9 @@
 from __future__ import annotations
 
 import warnings
+from collections.abc import Sequence
 from dataclasses import replace
-from typing import Any, Sequence
+from typing import Any
 
 from pydantic.errors import PydanticUserError
 from temporalio.contrib.pydantic import PydanticPayloadConverter, pydantic_data_converter
@@ -101,6 +102,8 @@ class PydanticAIPlugin(SimplePlugin):
 
         for workflow_class in workflows:  # type: ignore[reportUnknownMemberType]
             agents = getattr(workflow_class, '__pydantic_ai_agents__', None)  # type: ignore[reportUnknownMemberType]
+            if agents is None:
+                continue
             if not isinstance(agents, Sequence):
                 raise TypeError(  # pragma: no cover
                     f'__pydantic_ai_agents__ must be a Sequence of TemporalAgent instances, got {type(agents)}'
@@ -108,7 +111,7 @@ class PydanticAIPlugin(SimplePlugin):
             for agent in agents:  # type: ignore[reportUnknownVariableType]
                 if not isinstance(agent, TemporalAgent):
                     raise TypeError(  # pragma: no cover
-                        f'__pydantic_ai_agents__ must be a Sequence of TemporalAgent, got {type(agent)}' # type: ignore[reportUnknownVariableType]
+                        f'__pydantic_ai_agents__ must be a Sequence of TemporalAgent, got {type(agent)}'  # type: ignore[reportUnknownVariableType]
                     )
                 activities.extend(agent.temporal_activities)  # type: ignore[reportUnknownMemberType]
 
--- a/pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_workflow.py
+++ b/pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_workflow.py
@@ -5,6 +5,6 @@ from pydantic_ai.durable_exec.temporal import TemporalAgent
 
 
 class PydanticAIWorkflow:
-    """Temporal Workflow base class that provides `__pydantic_ai_agents__` for direct agent registration instead of using plugins."""
+    """Temporal Workflow base class that provides `__pydantic_ai_agents__` for direct agent registration."""
 
     __pydantic_ai_agents__: Sequence[TemporalAgent[Any, Any]]
--- a/tests/test_temporal.py
+++ b/tests/test_temporal.py
@@ -2374,6 +2374,16 @@ class WorkflowWithAgents(PydanticAIWorkflow):
         return result.output
 
 
+@workflow.defn
+class WorkflowWithAgentsWithoutPydanticAIWorkflow:
+    __pydantic_ai_agents__ = [simple_temporal_agent]
+
+    @workflow.run
+    async def run(self, prompt: str) -> str:
+        result = await simple_temporal_agent.run(prompt)
+        return result.output
+
+
 async def test_passing_agents_through_workflow(allow_model_requests: None, client: Client):
     async with Worker(
         client,
@@ -2384,6 +2394,21 @@ async def test_passing_agents_through_workflow(allow_model_requests: None, clien
             WorkflowWithAgents.run,
             args=['What is the capital of Mexico?'],
             id=WorkflowWithAgents.__name__,
+            task_queue=TASK_QUEUE,
+        )
+        assert output == snapshot('The capital of Mexico is Mexico City.')
+
+
+async def test_passing_agents_through_workflow_without_pydantic_ai_workflow(allow_model_requests: None, client: Client):
+    async with Worker(
+        client,
+        task_queue=TASK_QUEUE,
+        workflows=[WorkflowWithAgentsWithoutPydanticAIWorkflow],
+    ):
+        output = await client.execute_workflow(
+            WorkflowWithAgentsWithoutPydanticAIWorkflow.run,
+            args=['What is the capital of Mexico?'],
+            id=WorkflowWithAgentsWithoutPydanticAIWorkflow.__name__,
             task_queue=TASK_QUEUE,
         )
         assert output == snapshot('The capital of Mexico is Mexico City.')
--- a/tests/cassettes/test_temporal/test_passing_agents_through_workflow_without_pydantic_ai_workflow.yaml
+++ b/tests/cassettes/test_temporal/test_passing_agents_through_workflow_without_pydantic_ai_workflow.yaml
@@ -1,0 +1,79 @@
+interactions:
+- request:
+    headers:
+      accept:
+      - application/json
+      accept-encoding:
+      - gzip, deflate
+      connection:
+      - keep-alive
+      content-length:
+      - '105'
+      content-type:
+      - application/json
+      host:
+      - api.openai.com
+    method: POST
+    parsed_body:
+      messages:
+      - content: What is the capital of Mexico?
+        role: user
+      model: gpt-4o
+      stream: false
+    uri: https://api.openai.com/v1/chat/completions
+  response:
+    headers:
+      access-control-expose-headers:
+      - X-Request-ID
+      alt-svc:
+      - h3=":443"; ma=86400
+      connection:
+      - keep-alive
+      content-length:
+      - '838'
+      content-type:
+      - application/json
+      openai-organization:
+      - pydantic-28gund
+      openai-processing-ms:
+      - '403'
+      openai-project:
+      - proj_dKobscVY9YJxeEaDJen54e3d
+      openai-version:
+      - '2020-10-01'
+      strict-transport-security:
+      - max-age=31536000; includeSubDomains; preload
+      transfer-encoding:
+      - chunked
+    parsed_body:
+      choices:
+      - finish_reason: stop
+        index: 0
+        logprobs: null
+        message:
+          annotations: []
+          content: The capital of Mexico is Mexico City.
+          refusal: null
+          role: assistant
+      created: 1754675179
+      id: chatcmpl-C2LSVwAtcuMjKCHykKXgKphwTaQVB
+      model: gpt-4o-2024-08-06
+      object: chat.completion
+      service_tier: default
+      system_fingerprint: fp_ff25b2783a
+      usage:
+        completion_tokens: 8
+        completion_tokens_details:
+          accepted_prediction_tokens: 0
+          audio_tokens: 0
+          reasoning_tokens: 0
+          rejected_prediction_tokens: 0
+        prompt_tokens: 14
+        prompt_tokens_details:
+          audio_tokens: 0
+          cached_tokens: 0
+        total_tokens: 22
+    status:
+      code: 200
+      message: OK
+version: 1


pydantic_ai_slim/pydantic_ai/durable_exec/temporal/__init__.py
```diff
@@ -91,6 +93,22 @@ def __init__(self):
...
 
+    def configure_worker(self, config: WorkerConfig) -> WorkerConfig:
+        config = super().configure_worker(config)
+
+        workflows = list(config.get('workflows', []))  # type: ignore[reportUnknownMemberType]
+        activities = list(config.get('activities', []))  # type: ignore[reportUnknownMemberType]
+
+        for workflow_class in workflows:  # type: ignore[reportUnknownMemberType]
+            if issubclass(workflow_class, PydanticAIWorkflow):
+                agents = workflow_class.__pydantic_ai_agents__
```
DouweM: Note this line that I added to the issue description, possibly after you started working on this:

> But for convenience, I don't want to _require_ the workflow class to subclass that, so we should also do a `getattr(workflow_cls, '__pydantic_ai_agents__', None)` and check the types at runtime.

tests/test_temporal.py
```diff
@@ -207,6 +213,16 @@ async def run(self, prompt: str) -> str:
         return result.output
 
 
+@workflow.defn
+class WorkflowWithAgents(PydanticAIWorkflow):
```
DouweM: Please move this down to just before the test that uses it

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/__init__.py
```diff
@@ -91,6 +93,22 @@ def __init__(self):
...
+                for agent in agents:
+                    activities.extend(agent.temporal_activities)  # type: ignore[reportUnknownMemberType]
+
+        config['activities'] = activities
+
+        return config
+
 
 class AgentPlugin(SimplePlugin):
```
DouweM, line 123: We should document this new feature on `docs/durable_execution/temporal.md`. I like it so much better than `AgentPlugin` that we should just update the main example.

It should have a tooltip on the `PydanticAIWorkflow` line saying this is optional but helps with `__pydantic_ai_agents__` typing, and then a tooltip on the `__pydantic_ai_agents__` line that mentions `AgentPlugin` as an alternative, if it's easier to modify the worker than initialization than the workflow class.

adtyavrdhn: Screenshot 2025-12-10 at 9 52 59 AM

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/__init__.py
```diff
@@ -91,6 +93,22 @@ def __init__(self):
...
 
+    def configure_worker(self, config: WorkerConfig) -> WorkerConfig:
+        config = super().configure_worker(config)
+
+        workflows = list(config.get('workflows', []))  # type: ignore[reportUnknownMemberType]
+        activities = list(config.get('activities', []))  # type: ignore[reportUnknownMemberType]
+
+        for workflow_class in workflows:  # type: ignore[reportUnknownMemberType]
+            agents = getattr(workflow_class, '__pydantic_ai_agents__', None)  # type: ignore[reportUnknownMemberType]
+            if agents is not None:
```
DouweM: We should raise an error if it's not a `Sequence` or if any of its items are not `TemporalAgent`s

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_workflow.py
```diff
@@ -0,0 +1,10 @@
+from collections.abc import Sequence
+from typing import Any
+
+from pydantic_ai.durable_exec.temporal import TemporalAgent
+
+
+class PydanticAIWorkflow:
+    """Temporal Workflow base class that provides `__pydantic_ai_agents__` for direct agent registration instead of using plugins."""
```
DouweM: Please remove the "instead of using plugins" bit as it's only relevant to those who know what we used to recommend before, not to new readers of the docs

tests/test_temporal.py
```diff
@@ -2356,3 +2362,28 @@ async def test_beta_graph_parallel_execution_in_workflow(client: Client):
         # Results can be in any order due to parallel execution
         # 10 * 2 = 20, 10 * 3 = 30, 10 * 4 = 40
         assert sorted(output) == [20, 30, 40]
+
+
+@workflow.defn
+class WorkflowWithAgents(PydanticAIWorkflow):
```
DouweM, line 2368: Let's have another test to verify that it also works without subclassing

DouweM (changes requested):

DouweM:
@adtyavrdhn Thanks Aditya!

adtyavrdhn:
## Code Review

This PR adds a `PydanticAIWorkflow` base class and `__pydantic_ai_agents__` duck-typing convention so that `configure_worker()` can auto-register activities from workflow class definitions, without requiring an explicit `AgentPlugin`. The idea is clean and well-scoped. A few issues are worth addressing:

---

### 1. Circular import in `_workflow.py` ⚠️

**File:** `pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_workflow.py`

```python
from pydantic_ai.durable_exec.temporal import TemporalAgent  # imports __init__.py
```

`__init__.py` imports `_workflow.py` via `from ._workflow import PydanticAIWorkflow`, and `_workflow.py` imports back from `__init__.py`. This is a circular import. It happens to work at runtime today because `TemporalAgent` is added to `__init__.py`'s namespace via `from ._agent import TemporalAgent` on line 15, *before* `PydanticAIWorkflow` is imported on line 19 — meaning the partially-initialised `__init__` module already carries the name when `_workflow.py` needs it. But this is fragile: reordering imports in `__init__.py` silently breaks `_workflow.py`.

**Fix** — import directly from the submodule:
```python
from ._agent import TemporalAgent
```

---

### 2. `__pydantic_ai_agents__` is missing `ClassVar` ⚠️

**File:** `pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_workflow.py`

```python
class PydanticAIWorkflow:
    __pydantic_ai_agents__: Sequence[TemporalAgent[Any, Any]]
```

Without `ClassVar`, type checkers — and Pydantic, if this class is ever used as a mixin with a Pydantic model — treat this as an *instance* attribute declaration, not a class-level one. Assignments like `__pydantic_ai_agents__ = [agent]` in a subclass body will be flagged by pyright/mypy.

**Fix:**
```python
from typing import ClassVar

class PydanticAIWorkflow:
    __pydantic_ai_agents__: ClassVar[Sequence[TemporalAgent[Any, Any]]]
```

---

### 3. Duplicate `# (10)!` callout in the documentation 🐛

**File:** `docs/durable_execution/temporal.md`

Both `name='geography'` (line 97) and `client.execute_workflow(` (line 124) are annotated with `# (10)!`. Footnote (9) — *"The agent's `name` is used to uniquely identify its activities"* — therefore has no corresponding marker in the code sample and is an orphan, while footnote (10) is pointed to twice.

**Fix:** Change `name='geography',  # (10)!` → `name='geography',  # (9)!`

---

### 4. No deduplication guard when mixing `__pydantic_ai_agents__` and `AgentPlugin` ⚠️

**File:** `pydantic_ai_slim/pydantic_ai/durable_exec/temporal/__init__.py`, `configure_worker`

```python
activities.extend(agent.temporal_activities)
```

If a user lists an agent in `__pydantic_ai_agents__` *and* also registers it via `AgentPlugin(same_agent)` on the same `Worker`, `configure_worker` extends the activities list without deduplication. Temporal raises an error when the same activity function appears twice. At minimum, the docstring/docs should warn that the two registration mechanisms are mutually exclusive; better still, deduplicate by callable identity:

```python
existing = set(activities)
for activity in agent.temporal_activities:
    if activity not in existing:
        activities.append(activity)
        existing.add(activity)
```

---

### 5. `# pragma: no cover` on user-facing validation paths — missing tests for bad input 🧪

**File:** `pydantic_ai_slim/pydantic_ai/durable_exec/temporal/__init__.py`, lines 116–122

```python
if not isinstance(agents, Sequence):
    raise TypeError(...)  # pragma: no cover
...
if not isinstance(agent, TemporalAgent):
    raise TypeError(...)  # pragma: no cover
```

These are explicitly user-facing runtime checks with descriptive error messages, yet both are excluded from coverage. `# pragma: no cover` implies they are unreachable, which is not true — a user could write `__pydantic_ai_agents__ = "oops"`. Two small tests covering each `TypeError` path would validate the checks and allow the `pragma` comments to be dropped.

---

### 6. `config['activities']` is unconditionally overwritten

**File:** `pydantic_ai_slim/pydantic_ai/durable_exec/temporal/__init__.py`, `configure_worker`

```python
activities = list(config.get('activities', []))
# ... loop (may find no agents) ...
config['activities'] = activities  # always set, even if nothing changed
```

If the original config had `activities=None` (a valid Temporal default meaning "no activities registered"), `config.get('activities', [])` returns `[]`, then `config['activities'] = []` writes an empty list back — a subtle semantic change. The assignment should be guarded so it only runs when activities were actually appended, or the `None` case should be explicitly handled.

adtyavrdhn:
## Code Review

This PR introduces a nice ergonomic improvement — declaring `TemporalAgent`s directly on the workflow class via `__pydantic_ai_agents__` instead of registering `AgentPlugin` separately on the worker. A few issues worth addressing:

---

### 🔴 Circular Import in `_workflow.py`

**`pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_workflow.py` line 4**

```python
from pydantic_ai.durable_exec.temporal import TemporalAgent
```

This is a circular import. `__init__.py` imports from `_workflow.py` (`from ._workflow import PydanticAIWorkflow`), and `_workflow.py` imports back from `__init__.py`. It only works today because `TemporalAgent` happens to be added to `__init__.py`'s namespace *before* `PydanticAIWorkflow` is imported (line 16 vs 20). Any reordering by a linter/isort could silently break this at import time.

**Fix:** Use the direct relative import in `_workflow.py`:
```python
from ._agent import TemporalAgent
```

---

### 🔴 Duplicate Activity Registration

**`pydantic_ai_slim/pydantic_ai/durable_exec/temporal/__init__.py` lines 108–118**

```python
for agent in agents:
    activities.extend(agent.temporal_activities)
```

If the same `TemporalAgent` is referenced by more than one workflow class, or is *also* registered via `AgentPlugin` on the same worker, all of its activities are appended multiple times. Temporal may raise a `ValueError` at worker startup if the same activity function is registered twice.

**Fix:** Deduplicate by object identity before extending:
```python
registered_agent_ids: set[int] = set()
for workflow_class in workflows:
    ...
    for agent in agents:
        if id(agent) not in registered_agent_ids:
            activities.extend(agent.temporal_activities)
            registered_agent_ids.add(id(agent))
```

---

### 🟡 Missing `ClassVar` Annotation on `PydanticAIWorkflow`

**`pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_workflow.py` line 9**

```python
__pydantic_ai_agents__: Sequence[TemporalAgent[Any, Any]]
```

Without `ClassVar`, type checkers treat this as an *instance* variable annotation rather than a class variable. Subclasses assigning it at class scope (`__pydantic_ai_agents__ = [agent]`) can trigger errors with strict type checkers.

**Fix:**
```python
from typing import ClassVar
...
__pydantic_ai_agents__: ClassVar[Sequence[TemporalAgent[Any, Any]]]
```

---

### 🟡 `# pragma: no cover` on Reachable Validation Branches

**`pydantic_ai_slim/pydantic_ai/durable_exec/temporal/__init__.py` lines 109–115**

```python
if not isinstance(agents, Sequence):
    raise TypeError(  # pragma: no cover
        f'__pydantic_ai_agents__ must be a Sequence of TemporalAgent instances, got {type(agents)}'
    )
...
if not isinstance(agent, TemporalAgent):
    raise TypeError(  # pragma: no cover
        f'__pydantic_ai_agents__ must be a Sequence of TemporalAgent, got {type(agent)}'
    )
```

Both `TypeError` guards exist to protect against user error, yet both are marked `# pragma: no cover`, meaning they are completely untested. A reachable runtime guard that bypasses coverage provides false confidence. Either:
- Add simple unit tests for each branch (e.g. pass `__pydantic_ai_agents__ = "oops"` and assert the `TypeError`), or
- Remove the guards if you trust the type system to catch these at authoring time.

---

### 🟡 Excessive `# type: ignore` Comments

**`pydantic_ai_slim/pydantic_ai/durable_exec/temporal/__init__.py` lines 97–118**

The `configure_worker` method accumulates six separate `# type: ignore` suppressions (`reportUnknownMemberType`, `reportUnknownVariableType`). While some may be unavoidable given how the Temporal SDK types `WorkerConfig`, this density signals that the types are underspecified. At minimum, consider grouping the suppressions with a brief explanatory comment (e.g., `# WorkerConfig values are typed as object in the Temporal SDK`) to make the intent clear to future readers.

---

### 🟢 Minor: Missing Test for Plugin-Absent Failure Mode

There is no test verifying that a `Worker` *without* `plugins=[PydanticAIPlugin()]` but with a workflow carrying `__pydantic_ai_agents__` fails to run because activities are never registered. Adding such a test would close the loop on the expected error path for this feature.

PR #3678, opened 2025-12-09
Add prompt caching options to OpenAIChatModelSettings

kichanyurd:
Adding support for OpenAI's prompt caching configuration parameters (`prompt_cache_key` and `prompt_cache_retention`) to the `OpenAIResponsesModelSettings`, enabling users to optimize cache hit rates and control cache retention behavior.

###  Why This Is Needed

While OpenAI's automatic prompt caching works out-of-the-box, the API provides additional parameters that allow developers to control and significantly improve caching effectiveness:

Cache control is incredibly important for controlling cost and latency at scale, in applications with significant input token reuse.

#### Cache Routing Optimization (prompt_cache_key)
OpenAI routes requests to servers based on a hash of the prompt prefix. When multiple requests share long common prefixes, providing a consistent prompt_cache_key helps route them to the same machine, improving cache hit rates. This is particularly beneficial for applications with shared system prompts or instructions across many user requests.

####  Cache Retention Control (prompt_cache_retention)
Default in-memory caching retains prefixes for 5-10 minutes (up to 1 hour max). Extended retention (24h) keeps cached prefixes active for up to 24 hours on supported models (gpt-5.1, gpt-5.1-codex, gpt-5, gpt-4.1), enabling applications to benefit from caching across longer time periods

### Changes

- Added `openai_prompt_cache_key: str` field to `OpenAIResponsesModelSettings`
- Added `openai_prompt_cache_retention field: Literal["in-memory", "24h"]` to `OpenAIResponsesModelSettings`
- Both parameters are passed through to the OpenAI Responses API as optional parameters
- Includes docstrings with links to OpenAI's official prompt caching guide

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/openai.py
+++ b/pydantic_ai_slim/pydantic_ai/models/openai.py
@@ -201,6 +201,18 @@ class OpenAIChatModelSettings(ModelSettings, total=False):
     This feature is currently only supported for some OpenAI models.
     """
 
+    openai_prompt_cache_key: str
+    """Used by OpenAI to cache responses for similar requests to optimize your cache hit rates.
+
+    See the [OpenAI Prompt Caching documentation](https://platform.openai.com/docs/guides/prompt-caching#how-it-works) for more information.
+    """
+
+    openai_prompt_cache_retention: Literal['in-memory', '24h']
+    """The retention policy for the prompt cache. Set to 24h to enable extended prompt caching, which keeps cached prefixes active for longer, up to a maximum of 24 hours.
+
+    See the [OpenAI Prompt Caching documentation](https://platform.openai.com/docs/guides/prompt-caching#how-it-works) for more information.
+    """
+
 
 @deprecated('Use `OpenAIChatModelSettings` instead.')
 class OpenAIModelSettings(OpenAIChatModelSettings, total=False):
@@ -550,6 +562,8 @@ class OpenAIChatModel(Model):
                 logit_bias=model_settings.get('logit_bias', OMIT),
                 logprobs=model_settings.get('openai_logprobs', OMIT),
                 top_logprobs=model_settings.get('openai_top_logprobs', OMIT),
+                prompt_cache_key=model_settings.get('openai_prompt_cache_key', OMIT),
+                prompt_cache_retention=model_settings.get('openai_prompt_cache_retention', OMIT),
                 extra_headers=extra_headers,
                 extra_body=model_settings.get('extra_body'),
             )
@@ -1393,6 +1407,8 @@ class OpenAIResponsesModel(Model):
                 user=model_settings.get('openai_user', OMIT),
                 text=text or OMIT,
                 include=include or OMIT,
+                prompt_cache_key=model_settings.get('openai_prompt_cache_key', OMIT),
+                prompt_cache_retention=model_settings.get('openai_prompt_cache_retention', OMIT),
                 extra_headers=extra_headers,
                 extra_body=model_settings.get('extra_body'),
             )


pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -283,6 +283,18 @@ class OpenAIResponsesModelSettings(OpenAIChatModelSettings, total=False):
     Corresponds to the `web_search_call.action.sources` value of the `include` parameter in the Responses API.
     """
 
+    openai_prompt_cache_key: str
```
DouweM: These fields are supported by the Chat Completions API as well, so please move these to `OpenAIChatModelSettings` and implement support for them in `OpenAIChatModel` as well

kichanyurd: @DouweM I was considering that, as I saw in the API itself that I can pass them. But the official OpenAI docs don't mention them there. It might be risky to add them without an official statement of support. WDYT?

DouweM: @kichanyurd If they're allowed by the SDK types and we can try it out and it doesn't fail, I'm fine with it :)

kichanyurd: Done! @DouweM

Clement-Lelievre:
Screenshot 2025-12-11 at 08 16 19

hey @kichanyurd 

it seems your PR released in 1.30.0 broke something: when I tried my own addition from 1.30.0 by running [this code](https://github.com/pydantic/pydantic-ai/issues/3690#issue-3714210959) it didn't work anymore (see image) while it worked on my PR branch.

Could you pls check and let me know?

kichanyurd:
@Clement-Lelievre could it be that your openai package is not up to date?

Clement-Lelievre:
> @Clement-Lelievre could it be that your openai package is not up to date?

@kichanyurd yes indeed, I was running on pydantic-ai==1.30.0 and openai==2.2.0, now with openai==2.9.0 it works.

Because your addition to pydantic-ai requires openai to be at least a specific version, shouldn't you update this dep in pydantic-ai's deps, so that by upgrading pydantic-ai people get a compatible version of openai? cc @DouweM

kichanyurd:
> > @Clement-Lelievre could it be that your openai package is not up to date?
> 
> @kichanyurd yes indeed, I was running on pydantic-ai==1.30.0 and openai==2.2.0, now with openai==2.9.0 it works.
> 
> Because your addition to pydantic-ai requires openai to be at least a specific version, shouldn't you update this dep in pydantic-ai's deps, so that by upgrading pydantic-ai people get a compatible version of openai? cc @DouweM 

@Clement-Lelievre Yeah I think you're right. Looking at https://github.com/pydantic/pydantic-ai/blob/main/pydantic_ai_slim%2Fpyproject.toml it seems the OpenAI version is really old.

@DouweM is there a reason that it's not even 2.x? Waiting for your input on this and will remediate as soon as we've got clarity on how to proceed.

DouweM:
@Clement-Lelievre Sorry about that, I've created anew issue: https://github.com/pydantic/pydantic-ai/issues/3707

@kichanyurd I've responded to the openai version question there

Clement-Lelievre:
> @Clement-Lelievre Sorry about that, I've created anew issue: #3707
> 
> @kichanyurd I've responded to the openai version question there

@DouweM no worries

maybe we could catch these with e2e tests?

DouweM:
@Clement-Lelievre It actually should've been caught by CI because we run the test suite against the lowest versions. Let me see why that wasn't happening.

Edit: New issue https://github.com/pydantic/pydantic-ai/issues/3710

Issue #3260, opened 2025-10-25
Temporalio 1.18.1 incompatible with the latest Pydantic AI 1.6.0


dorkknightdev:
### Initial Checks

- [x] I confirm that I'm using the latest version of Pydantic AI
- [x] I confirm that I searched for my issue in https://github.com/pydantic/pydantic-ai/issues before opening this issue

### Description

When I try the documented example on https://ai.pydantic.dev/durable_execution/temporal/#durable-agent, I get the following error:

```
(assistant_poc) % python main.py
Traceback (most recent call last):
  File "/Users/peter.kim/proj/assistant_poc/main.py", line 54, in <module>
    asyncio.run(main())
    ~~~~~~~~~~~^^^^^^^^
  File "/Users/peter.kim/Library/Application Support/uv/python/cpython-3.13.1-macos-aarch64-none/lib/python3.13/asyncio/runners.py", line 194, in run
    return runner.run(main)
           ~~~~~~~~~~^^^^^^
  File "/Users/peter.kim/Library/Application Support/uv/python/cpython-3.13.1-macos-aarch64-none/lib/python3.13/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/Users/peter.kim/Library/Application Support/uv/python/cpython-3.13.1-macos-aarch64-none/lib/python3.13/asyncio/base_events.py", line 720, in run_until_complete
    return future.result()
           ~~~~~~~~~~~~~^^
  File "/Users/peter.kim/proj/assistant_poc/main.py", line 34, in main
    plugins=[PydanticAIPlugin()],
             ~~~~~~~~~~~~~~~~^^
TypeError: Can't instantiate abstract class PydanticAIPlugin without an implementation for abstract methods 'configure_replayer', 'connect_service_client', 'init_client_plugin', 'init_worker_plugin', 'run_replayer', 'run_worker'
```

The package installation instruction was missing, so I installed `pydantic-ai` and `temporalio` separately, and that's probably part of the reason:
```
uv add temporalio.   # installs 1.18.1
uv add pydantic-ai.  # installs 1.6.0
```

If I install pydantic-ai using one of these methods, tempralio 1.18.0 is installed and there's no error:
```
uv add 'pydantic-ai[temporal]'
uv add 'pydantic-ai-slim[temporal,openai]'
```

[...]

PR #3214, opened 2025-10-21
Bump temporalio to v1.19.0 and use SimplePlugin

DouweM:
Depends on https://github.com/temporalio/sdk-python/pull/1139

PR #3679, opened 2025-12-09
Upgrade temporalio to 1.20.0

akrivka:
Bump `temporalio` version to 1.20.0. Don't see any breaking changes that would need changing in the code.


Kludex:
Why do we pin temporal? @DouweM

M0dEx:
> Why do we pin temporal? @DouweM

From @DouweM's [response](https://github.com/pydantic/pydantic-ai/issues/3260#issuecomment-3452457655):
> When temporalio 1.19.0 lands, we'll be bumping it in https://github.com/pydantic/pydantic-ai/pull/3214. And once their plugins interface is stablized, we can make the range >= again.

Issue #2818, opened 2025-09-05
Parallel MCP servers: `RuntimeError: Attempted to exit cancel scope in a different task than it was entered in`


g-eoj:
### Initial Checks

- [x] I confirm that I'm using the latest version of Pydantic AI
- [x] I confirm that I searched for my issue in https://github.com/pydantic/pydantic-ai/issues before opening this issue

### Description

When setting `max_concurrency` to anything other than one, eventually the `RuntimeError: Attempted to exit cancel scope in a different task than it was entered in` error will pop up for evals with ~10 or more cases. It appears to be connected to using toolsets created from MCP servers.

I followed https://ai.pydantic.dev/api/mcp/#pydantic_ai.mcp.MCPServerStreamableHTTP - tried using this pattern everywhere an agent is used:
```
import some_agent

...

async with some_agent:
    run = await some_agent.run(...)

...

````

### Example Code

Here is a simple reproduction - this should work right?

```Python
import asyncio

from pydantic_ai.mcp import MCPServerStdio
from pydantic_evals import Case, Dataset
from pydantic_evals.evaluators import Evaluator, EvaluatorContext


class MyEvaluator(Evaluator[str, bool]):
    def evaluate(self, ctx: EvaluatorContext[str, bool]) -> bool:
        return True


mcp_server = MCPServerStdio('python', ['-m', 'tests.mcp_server'])


async def run(case: str) -> bool:
    async with mcp_server:
        await mcp_server.list_tools()
    return True


async def main():
    cases: list[Case[str, bool, dict[str, str]]] = []
    for _ in range(10):
        case = Case(
            name=str(_),
            inputs='',
            expected_output=True,
            metadata={'difficulty': 'easy'},
        )
        cases.append(case)

    dataset = Dataset(
        cases=cases,
        evaluators=[MyEvaluator()],
    )

    report = await dataset.evaluate(run, max_concurrency=4)
    report.print()


if __name__ == '__main__':
    asyncio.run(main())

```

### Python, Pydantic AI & LLM client version

```Text
Pydantic AI 1.0.1
Python 3.12.7
```

DouweM:
[...]

PR #3682, opened 2025-12-09
Add `DynamicToolset` support in Temporal

### LINKED ISSUE #3390: `@agent.toolset` decorator (`DynamicToolset`) does not work with Temporal
An attempt at fixing this was made in https://github.com/pydantic/pydantic-ai/pull/3165. That PR can likely be revived.

dsfaccini:
Closes #3390
- adds support for `@agent.toolset` decorator

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/toolsets/_dynamic.py
+++ b/pydantic_ai_slim/pydantic_ai/toolsets/_dynamic.py
@@ -2,8 +2,7 @@ from __future__ import annotations
 
 import inspect
 from collections.abc import Awaitable, Callable
-from dataclasses import dataclass
-from typing import Any, Generic, TypeAlias
+from typing import Any, TypeAlias
 
 from typing_extensions import Self
 
@@ -15,14 +14,6 @@ ToolsetFunc: TypeAlias = Callable[
     AbstractToolset[AgentDepsT] | None | Awaitable[AbstractToolset[AgentDepsT] | None],
 ]
 """A sync/async function which takes a run context and returns a toolset."""
-
-
-@dataclass
-class _RunState(Generic[AgentDepsT]):
-    """Per-run state for a DynamicToolset."""
-
-    toolset: AbstractToolset[AgentDepsT] | None = None
-    run_step: int | None = None
 
 
 class DynamicToolset(AbstractToolset[AgentDepsT]):
@@ -45,7 +36,8 @@ class DynamicToolset(AbstractToolset[AgentDepsT]):
         self.toolset_func = toolset_func
         self.per_run_step = per_run_step
         self._id = id
-        self._run_state: _RunState[AgentDepsT] | None = None
+        self._toolset: AbstractToolset[AgentDepsT] | None = None
+        self._run_step: int | None = None
 
     @property
     def id(self) -> str | None:
@@ -73,21 +65,17 @@ class DynamicToolset(AbstractToolset[AgentDepsT]):
     async def __aexit__(self, *args: Any) -> bool | None:
         try:
             result = None
-            if self._run_state is not None and self._run_state.toolset is not None:
-                result = await self._run_state.toolset.__aexit__(*args)
+            if self._toolset is not None:
+                result = await self._toolset.__aexit__(*args)
         finally:
-            self._run_state = None
+            self._toolset = None
+            self._run_step = None
         return result
 
     async def get_tools(self, ctx: RunContext[AgentDepsT]) -> dict[str, ToolsetTool[AgentDepsT]]:
-        if self._run_state is None:
-            self._run_state = _RunState()
-
-        run_state = self._run_state
-
-        if run_state.toolset is None or (self.per_run_step and ctx.run_step != run_state.run_step):
-            if run_state.toolset is not None:
-                await run_state.toolset.__aexit__()
+        if self._toolset is None or (self.per_run_step and ctx.run_step != self._run_step):
+            if self._toolset is not None:
+                await self._toolset.__aexit__()
 
             toolset = self.toolset_func(ctx)
             if inspect.isawaitable(toolset):
@@ -96,35 +84,33 @@ class DynamicToolset(AbstractToolset[AgentDepsT]):
             if toolset is not None:
                 await toolset.__aenter__()
 
-            run_state.toolset = toolset
-            run_state.run_step = ctx.run_step
+            self._toolset = toolset
+            self._run_step = ctx.run_step
 
-        if run_state.toolset is None:
+        if self._toolset is None:
             return {}
 
-        return await run_state.toolset.get_tools(ctx)
+        return await self._toolset.get_tools(ctx)
 
     async def call_tool(
         self, name: str, tool_args: dict[str, Any], ctx: RunContext[AgentDepsT], tool: ToolsetTool[AgentDepsT]
     ) -> Any:
-        assert self._run_state is not None and self._run_state.toolset is not None
-        return await self._run_state.toolset.call_tool(name, tool_args, ctx, tool)
+        assert self._toolset is not None
+        return await self._toolset.call_tool(name, tool_args, ctx, tool)
 
     def apply(self, visitor: Callable[[AbstractToolset[AgentDepsT]], None]) -> None:
-        if self._run_state is None or self._run_state.toolset is None:
+        if self._toolset is None:
             super().apply(visitor)
         else:
-            self._run_state.toolset.apply(visitor)
+            self._toolset.apply(visitor)
 
     def visit_and_replace(
         self, visitor: Callable[[AbstractToolset[AgentDepsT]], AbstractToolset[AgentDepsT]]
     ) -> AbstractToolset[AgentDepsT]:
-        if self._run_state is None or self._run_state.toolset is None:
+        if self._toolset is None:
             return super().visit_and_replace(visitor)
         else:
             new_toolset = self.copy()
-            new_toolset._run_state = _RunState(
-                toolset=self._run_state.toolset.visit_and_replace(visitor),
-                run_step=self._run_state.run_step,
-            )
+            new_toolset._toolset = self._toolset.visit_and_replace(visitor)
+            new_toolset._run_step = self._run_step
             return new_toolset
--- a/tests/test_temporal.py
+++ b/tests/test_temporal.py
@@ -1188,7 +1188,7 @@ def my_mcp_dynamic_toolset(ctx: RunContext[None]) -> MCPServerStreamableHTTP:
 
     This tests MCP lifecycle management (context manager enter/exit) within DynamicToolset + Temporal.
     """
-    return MCPServerStreamableHTTP('https://mcp.deepwiki.com/mcp', id='dynamic_deepwiki')
+    return MCPServerStreamableHTTP('https://mcp.deepwiki.com/mcp')
 
 
 mcp_dynamic_toolset_temporal_agent = TemporalAgent(
--- a/tests/test_toolsets.py
+++ b/tests/test_toolsets.py
@@ -26,7 +26,7 @@ from pydantic_ai._tool_manager import ToolManager
 from pydantic_ai.exceptions import ModelRetry, ToolRetryError, UnexpectedModelBehavior, UserError
 from pydantic_ai.models.test import TestModel
 from pydantic_ai.tools import ToolDefinition
-from pydantic_ai.toolsets._dynamic import DynamicToolset, _RunState  # pyright: ignore[reportPrivateUsage]
+from pydantic_ai.toolsets._dynamic import DynamicToolset
 from pydantic_ai.usage import RunUsage
 
 pytestmark = pytest.mark.anyio
@@ -34,7 +34,7 @@ pytestmark = pytest.mark.anyio
 T = TypeVar('T')
 
 
-def build_run_context(deps: T, run_step: int = 0, run_id: str = 'test-run-id') -> RunContext[T]:
+def build_run_context(deps: T, run_step: int = 0) -> RunContext[T]:
     return RunContext(
         deps=deps,
         model=TestModel(),
@@ -42,7 +42,6 @@ def build_run_context(deps: T, run_step: int = 0, run_id: str = 'test-run-id') -
         prompt=None,
         messages=[],
         run_step=run_step,
-        run_id=run_id,
     )
 
 
@@ -708,12 +707,12 @@ async def test_visit_and_replace():
     toolset1 = FunctionToolset(id='toolset1')
     toolset2 = FunctionToolset(id='toolset2')
 
-    run_ctx = build_run_context(None, run_id='visit-replace-run')
+    run_ctx = build_run_context(None)
 
     active_dynamic_toolset = DynamicToolset(toolset_func=lambda ctx: toolset2)
     await active_dynamic_toolset.get_tools(run_ctx)
-    assert active_dynamic_toolset._run_state is not None  # pyright: ignore[reportPrivateUsage]
-    assert active_dynamic_toolset._run_state.toolset is toolset2  # pyright: ignore[reportPrivateUsage]
+    assert active_dynamic_toolset._toolset is not None  # pyright: ignore[reportPrivateUsage]
+    assert active_dynamic_toolset._toolset is toolset2  # pyright: ignore[reportPrivateUsage]
 
     inactive_dynamic_toolset = DynamicToolset(toolset_func=lambda ctx: FunctionToolset())
 
@@ -731,11 +730,8 @@ async def test_visit_and_replace():
         per_run_step=active_dynamic_toolset.per_run_step,
         id=active_dynamic_toolset._id,  # pyright: ignore[reportPrivateUsage]
     )
-    assert active_dynamic_toolset._run_state is not None  # pyright: ignore[reportPrivateUsage]
-    expected_dynamic._run_state = _RunState(  # pyright: ignore[reportPrivateUsage]
-        toolset=WrapperToolset(toolset2),
-        run_step=active_dynamic_toolset._run_state.run_step,  # pyright: ignore[reportPrivateUsage]
-    )
+    expected_dynamic._toolset = WrapperToolset(toolset2)  # pyright: ignore[reportPrivateUsage]
+    expected_dynamic._run_step = active_dynamic_toolset._run_step  # pyright: ignore[reportPrivateUsage]
 
     assert visited_toolset == CombinedToolset(
         [
@@ -780,17 +776,16 @@ async def test_dynamic_toolset():
 
     toolset = DynamicToolset[None](toolset_func=toolset_factory)
 
-    run_context = build_run_context(None, run_id='test-dynamic-run')
+    run_context = build_run_context(None)
 
     def get_inner_toolset(toolset: DynamicToolset[None] | None) -> EnterableToolset | None:
         assert toolset is not None
-        run_state = toolset._run_state  # pyright: ignore[reportPrivateUsage]
-        inner_toolset = run_state.toolset if run_state else None
+        inner_toolset = toolset._toolset  # pyright: ignore[reportPrivateUsage]
         assert isinstance(inner_toolset, EnterableToolset) or inner_toolset is None
         return inner_toolset
 
     async with toolset:
-        assert toolset._run_state is None  # pyright: ignore[reportPrivateUsage]
+        assert toolset._toolset is None  # pyright: ignore[reportPrivateUsage]
 
         # Test that calling get_tools initializes the toolset
         tools = await toolset.get_tools(run_context)
@@ -820,22 +815,20 @@ async def test_dynamic_toolset_empty():
 
     toolset = DynamicToolset[None](toolset_func=no_toolset_func)
 
-    run_context = build_run_context(None, run_id='test-empty-run')
+    run_context = build_run_context(None)
 
     tools = await toolset.get_tools(run_context)
 
     assert tools == {}
 
     async with toolset:
-        run_state = toolset._run_state  # pyright: ignore[reportPrivateUsage]
-        assert run_state is None or run_state.toolset is None
+        assert toolset._toolset is None  # pyright: ignore[reportPrivateUsage]
 
         tools = await toolset.get_tools(run_context)
 
         assert tools == {}
 
-        run_state = toolset._run_state  # pyright: ignore[reportPrivateUsage]
-        assert run_state is not None and run_state.toolset is None
+        assert toolset._toolset is None  # pyright: ignore[reportPrivateUsage]
 
 
 def test_dynamic_toolset_id():


docs/durable_execution/temporal.md
```diff
@@ -156,7 +156,15 @@ There are a few considerations specific to agents and toolsets when using Tempor
...
+When `TemporalAgent` dynamically creates activities for the wrapped agent's model requests and toolsets (specifically those that implement their own tool listing and calling, i.e. [`FunctionToolset`][pydantic_ai.toolsets.FunctionToolset], [`MCPServer`][pydantic_ai.mcp.MCPServer], and dynamic toolsets created with [`@agent.toolset`][pydantic_ai.Agent.toolset]), their names are derived from the agent's [`name`][pydantic_ai.agent.AbstractAgent.name] and the toolsets' [`id`s][pydantic_ai.toolsets.AbstractToolset.id]. These fields are normally optional, but are required to be set when using Temporal. They should not be changed once the durable agent has been deployed to production as this would break active workflows.
```
DouweM: Let's not mention `@agent.toolset` here because the next paragrapg is more clear

docs/toolsets.md
```diff
@@ -598,6 +598,14 @@ To register a dynamic toolset, you can pass a function that takes [`RunContext`]
 
 By default, the function will be called again ahead of each agent run step. If you are using the decorator, you can optionally provide a `per_run_step=False` argument to indicate that the toolset only needs to be built once for the entire run.
 
+When using [Temporal durable execution](./durable_execution/temporal.md), the decorator also accepts an `id` parameter to uniquely identify the toolset. If not provided, the function name is used as the ID.
+
+```python {test="skip"}
```
DouweM: We shouldn't have a niche Temporal-specific example ahead of the regular example :) I think we can just have the sentence, without the example.

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_toolset.py
```diff
@@ -127,6 +127,20 @@ def temporalize_toolset(
             run_context_type=run_context_type,
         )
 
+    from pydantic_ai.toolsets._dynamic import DynamicToolset
```
DouweM: Please move this to the top of the file

pydantic_ai_slim/pydantic_ai/toolsets/_dynamic.py
```diff
@@ -27,13 +27,15 @@ class DynamicToolset(AbstractToolset[AgentDepsT]):
 
     toolset_func: ToolsetFunc[AgentDepsT]
     per_run_step: bool = True
+    _id: str | None = None
```
DouweM: - Let's move this down with the other private fields
- Let's add a custom `__init__` that takes this in as `id`, like we have on other toolsets. Then the docstring can be there as well

DouweM: Hmm, if `__init__` takes it as `id`, but the actual field is `_id`, `dataclasses.replace` may get into trouble, which may be OK but above it says explicitly "To use it multiple times, copy it using `dataclasses.replace`", and we do that here:

https://github.com/pydantic/pydantic-ai/blob/f42e52346dfb7189dfb5216412bd52e5c98f981d/pydantic_ai_slim/pydantic_ai/agent/__init__.py#L1378-L1385

So we may need a new `copy()` method and use that instead.

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_dynamic_toolset.py
```diff
@@ -0,0 +1,155 @@
...
+
+        async def call_tool_activity(params: CallToolParams, deps: AgentDepsT) -> CallToolResult:
+            """Activity that instantiates the dynamic toolset and calls the tool."""
+            ctx = self.run_context_type.deserialize_run_context(params.serialized_run_context, deps=deps)
+
+            async with self.wrapped:
+                tools = await self.wrapped.get_tools(ctx)
+                tool = tools.get(params.name)
+                if tool is None:
+                    from pydantic_ai.exceptions import UserError
```
DouweM: Move to the top please

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_dynamic_toolset.py
```diff
@@ -0,0 +1,155 @@
...
+
+                    raise UserError(
+                        f'Tool {params.name!r} not found in dynamic toolset {self.id!r}. '
+                        'The dynamic toolset function may have returned a different toolset than expected.'
+                    )
+
+                args_dict = tool.args_validator.validate_python(params.tool_args)
+                return await self._wrap_call_tool_result(self.wrapped.call_tool(params.name, args_dict, ctx, tool))
```
DouweM: This is duplicated from here, where we have a useful comment to explain why we do this:

https://github.com/pydantic/pydantic-ai/blob/f42e52346dfb7189dfb5216412bd52e5c98f981d/pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_function_toolset.py#L49-L52

What do you think about moving these 2 lines into a new `_call_tool_in_activity` method that we can use here and where we currently have these 2 lines?

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_dynamic_toolset.py
```diff
@@ -0,0 +1,155 @@
...
+                **merged_config,
+            )
+        )
+
+    def _tool_for_tool_def(self, tool_def: ToolDefinition) -> ToolsetTool[AgentDepsT]:
+        """Create a ToolsetTool from a ToolDefinition for use outside activities."""
+        return ToolsetTool(
+            toolset=self,
+            tool_def=tool_def,
+            max_retries=1,
```
DouweM: We should use the actual `max_retries` we got from `self.wrapper.get_tools` inside the `get_tools_activity`

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_dynamic_toolset.py
```diff
@@ -0,0 +1,155 @@
...
+            args_validator=TOOL_SCHEMA_VALIDATOR,
```
DouweM, line 160: Let's add a note to say that this is fine, because we run the actual args validator inside `call_tool_activity`

docs/durable_execution/temporal.md
```diff
@@ -156,7 +156,15 @@ There are a few considerations specific to agents and toolsets when using Tempor
...
+
+For dynamic toolsets created with the [`@agent.toolset`][pydantic_ai.Agent.toolset] decorator, the `id` parameter can be set explicitly or it will default to the function name:
```
DouweM: We should also specify that with Temporal, `per_run_step=False` is not respected, as the toolset always needs to be created on-the-fly in the activity.

tests/test_temporal.py
```diff
@@ -1104,6 +1105,244 @@ async def test_toolset_without_id():
...
+def dynamic_toolset_model_logic(messages: list[ModelMessage], info: AgentInfo) -> ModelResponse:
+    """Model logic for testing dynamic toolsets."""
+    # If no tool calls yet, call the tool
+    if len(messages) == 1:
+        return ModelResponse(parts=[ToolCallPart(tool_name='get_dynamic_data', args={'key': 'test'})])
+    # After tool call, return final result
+    return ModelResponse(parts=[TextPart(content='Dynamic result received')])
+
+
+dynamic_toolset_function_model = FunctionModel(dynamic_toolset_model_logic)
```
DouweM: Could we use `TestModel`, which already automatically calls the first tool?

tests/test_temporal.py
```diff
@@ -1104,6 +1105,244 @@ async def test_toolset_without_id():
...
+
+dynamic_toolset_agent = Agent(dynamic_toolset_function_model, name='dynamic_toolset_agent')
+
+
+@dynamic_toolset_agent.toolset
+def my_dynamic_toolset(ctx: RunContext[None]) -> FunctionToolset[None]:
+    toolset = FunctionToolset[None](id='inner_dynamic')
```
DouweM: From a lifecycle management perspective, a dynamic toolset that creates an `MCPServer` is more interesting. Can you try that? See examples further up with `MCPServerStdio`

tests/test_temporal.py
```diff
@@ -1104,6 +1105,244 @@ async def test_toolset_without_id():
...
+
+@workflow.defn
+class DynamicToolsetAgentWorkflow:
+    @workflow.run
+    async def run(self, prompt: str) -> str:
+        result = await dynamic_toolset_temporal_agent.run(prompt)
+        return result.output
+
+
+async def test_dynamic_toolset_temporal_agent_structure():
```
DouweM: We don't need to test any of this, just that the dynamic toolset actually works in Temporal

tests/test_temporal.py
```diff
@@ -1104,6 +1105,244 @@ async def test_toolset_without_id():
...
+        output = await client.execute_workflow(
+            DynamicToolsetAgentWorkflow.run,
+            args=['Call the dynamic tool with test key'],
+            id='test_dynamic_toolset_workflow',
+            task_queue=TASK_QUEUE,
+        )
+        assert output == 'Dynamic result received'
+
+
+# Test with explicit id parameter
```
DouweM: I'd rather test the new `id` arg where we test the `@agent.toolset` decorator, since this part is not Temporal-specific

tests/test_temporal.py
```diff
@@ -1104,6 +1105,244 @@ async def test_toolset_without_id():
...
+    return toolset
+
+
+dynamic_toolset_with_id_temporal_agent = TemporalAgent(
+    dynamic_toolset_agent_with_id,
+    activity_config=BASE_ACTIVITY_CONFIG,
+)
+
+
+async def test_dynamic_toolset_with_custom_id():
```
DouweM: Please remove this test

tests/test_temporal.py
```diff
@@ -1104,6 +1105,244 @@ async def test_toolset_without_id():
...
+
+    for toolset in toolsets:
+        if isinstance(toolset, TemporalDynamicToolset):
+            assert toolset.id == 'custom_toolset_id'
+            break
+    else:
+        pytest.fail('TemporalDynamicToolset not found in toolsets')
+
+
+async def test_dynamic_toolset_get_tools_outside_workflow():
```
DouweM: We should test this by running the same agent we created earlier outside of a workflow, and checking that we still get the correct result

tests/test_temporal.py
```diff
@@ -1104,6 +1105,244 @@ async def test_toolset_without_id():
...
+        if isinstance(toolset, TemporalDynamicToolset):
+            async with toolset:
+                tools = await toolset.get_tools(ctx)
+            assert 'get_dynamic_data' in tools
+            break
+    else:
+        pytest.fail('TemporalDynamicToolset not found')
+
+
+async def test_dynamic_toolset_call_tool_outside_workflow():
```
DouweM: Not needed

tests/test_temporal.py
```diff
@@ -1104,6 +1105,244 @@ async def test_toolset_without_id():
...
+                tools = await toolset.get_tools(ctx)
+                tool = tools['get_dynamic_data']
+                result = await toolset.call_tool('get_dynamic_data', {'key': 'mykey'}, ctx, tool)
+            assert result == 'dynamic_value_for_mykey'
+            break
+    else:
+        pytest.fail('TemporalDynamicToolset not found')
+
+
+async def test_dynamic_toolset_tool_not_found_in_activity():
```
DouweM: I'm fine with `pragma: no cover` for this `UserError` like we have here:

https://github.com/pydantic/pydantic-ai/blob/f42e52346dfb7189dfb5216412bd52e5c98f981d/pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_function_toolset.py#L41-L47

tests/test_temporal.py
dsfaccini: so `MCPServerStdio` will not work, see https://github.com/pydantic/pydantic-ai/issues/2818#issuecomment-3639083046

but I did the test with `FastMCPToolset`

pydantic_ai_slim/pydantic_ai/toolsets/_dynamic.py
dsfaccini: removed dataclass, added `__init__` with `id` param and replaced `replace` logic with `copy()`

docs/durable_execution/temporal.md
```diff
@@ -158,6 +158,27 @@ To ensure that Temporal knows what code to run when an activity fails or is inte
 
 When `TemporalAgent` dynamically creates activities for the wrapped agent's model requests and toolsets (specifically those that implement their own tool listing and calling, i.e. [`FunctionToolset`][pydantic_ai.toolsets.FunctionToolset] and [`MCPServer`][pydantic_ai.mcp.MCPServer]), their names are derived from the agent's [`name`][pydantic_ai.agent.AbstractAgent.name] and the toolsets' [`id`s][pydantic_ai.toolsets.AbstractToolset.id]. These fields are normally optional, but are required to be set when using Temporal. They should not be changed once the durable agent has been deployed to production as this would break active workflows.
 
+For dynamic toolsets created with the [`@agent.toolset`][pydantic_ai.Agent.toolset] decorator, the `id` parameter can be set explicitly or it will default to the function name:
```
DouweM: Please link to the dynamic toolset doc

docs/durable_execution/temporal.md
```diff
@@ -158,6 +158,27 @@ To ensure that Temporal knows what code to run when an activity fails or is inte
...
+from pydantic_ai import Agent, RunContext
+from pydantic_ai.toolsets import FunctionToolset
+
+agent = Agent('openai:gpt-5')
+
+@dataclass
+class MyDeps:
+    ...
+
+@agent.toolset(id='my_dynamic_tools')
```
DouweM: I don't think we need the example

docs/durable_execution/temporal.md
```diff
@@ -158,6 +158,27 @@ To ensure that Temporal knows what code to run when an activity fails or is inte
...
+def my_toolset(ctx: RunContext[MyDeps]) -> FunctionToolset:
+    ...
+```
+
+Note that with Temporal, `per_run_step=False` is not respected, as the toolset always needs to be created on-the-fly in the activity.
```
DouweM: This can be on the same single line about dynamic toolsets as the stuff above

docs/durable_execution/temporal.md
DouweM: I think it's actually better to NOT default to the function name and require `id` to be set explicitly when Temporal is used, as the ID is used in activity names that need to be stable (i.e. can't change from one release to the next). So it's better for it to be separate field with a clear docstring that the user won't accidentally rename when they're doing a code refactor

docs/toolsets.md
```diff
@@ -598,6 +598,8 @@ To register a dynamic toolset, you can pass a function that takes [`RunContext`]
 
 By default, the function will be called again ahead of each agent run step. If you are using the decorator, you can optionally provide a `per_run_step=False` argument to indicate that the toolset only needs to be built once for the entire run.
 
+When using [Temporal durable execution](./durable_execution/temporal.md), the decorator also accepts an `id` parameter to uniquely identify the toolset. If not provided, the function name is used as the ID.
```
DouweM: This makes it seem like it only accepts the parameter when Temporal is used, while of course it always accepts the parameter :) 

If we don't use the function name by default as I suggested above, we may not need this line here at all, as this won't be relevant unless they use Temporal, and if they do they'll get an error or read it in the docs

pydantic_ai_slim/pydantic_ai/toolsets/_dynamic.py
```diff
@@ -17,23 +16,52 @@
...
 
     toolset_func: ToolsetFunc[AgentDepsT]
-    per_run_step: bool = True
-
-    _toolset: AbstractToolset[AgentDepsT] | None = None
-    _run_step: int | None = None
+    per_run_step: bool
+    _id: str | None
+    _toolset: AbstractToolset[AgentDepsT] | None
+    _run_step: int | None
```
DouweM: I wonder if we could solve this "reuse a dynamic toolset across multiple agent runs" issue not by copying it for each run (which is tricky as we see here), but by storing the `_toolset` and `_run_step` on a dict keyed by `ctx.run_id`.

That way, we can keep using the same toolset instance, but state from different runs wouldn't interfere. I think that's worth trying in this PR, as an alternative to the new `copy` method

tests/test_temporal.py
```diff
@@ -1104,6 +1104,114 @@ async def test_toolset_without_id():
         TemporalAgent(Agent(model=model, name='test_agent', toolsets=[FunctionToolset()]))
 
 
+# --- DynamicToolset / @agent.toolset tests ---
+
+
+def get_dynamic_weather(location: str) -> str:
```
DouweM: We don't need this at the top level; we can register it to the `FunctionToolset` inline inside `my_dynamic_toolset` using `@toolset.tool`

tests/test_temporal.py
```diff
@@ -1104,6 +1104,114 @@ async def test_toolset_without_id():
...
+async def test_dynamic_toolset_outside_workflow():
+    """Test that the dynamic toolset agent works correctly outside of a workflow."""
+    result = await dynamic_toolset_temporal_agent.run('Get the weather for Paris')
+    assert result.output == snapshot('{"get_dynamic_weather":"The weather in a is sunny."}')
+
+
+# --- MCP-based DynamicToolset test ---
+# Tests that @agent.toolset with an MCP toolset works with Temporal workflows.
+# Uses FastMCPToolset (HTTP-based) rather than MCPServerStdio (subprocess-based) because
+# MCPServerStdio has issues when created dynamically inside Temporal activities.
```
DouweM: Let's link to the issue

tests/test_temporal.py
```diff
@@ -1104,6 +1104,114 @@ async def test_toolset_without_id():
...
+
+fastmcp_dynamic_toolset_agent = Agent(model, name='fastmcp_dynamic_toolset_agent')
+
+
+@fastmcp_dynamic_toolset_agent.toolset(per_run_step=False)
+def my_fastmcp_dynamic_toolset(ctx: RunContext[None]) -> FastMCPToolset:
+    """Dynamic toolset that returns an MCP toolset.
+
+    This tests MCP lifecycle management (context manager enter/exit) within DynamicToolset + Temporal.
+    Uses per_run_step=False so the toolset persists across run steps within an activity.
```
DouweM: Does this actually work?

tests/test_temporal.py
```diff
@@ -1104,6 +1104,114 @@ async def test_toolset_without_id():
...
+    """Get the weather for a location."""
+    return f'The weather in {location} is sunny.'
+
+
+dynamic_toolset_agent = Agent(TestModel(), name='dynamic_toolset_agent')
+
+
+@dynamic_toolset_agent.toolset
+def my_dynamic_toolset(ctx: RunContext[None]) -> FunctionToolset[None]:
+    return FunctionToolset(tools=[get_dynamic_weather], id='dynamic_weather')
```
DouweM: Let's use a value from `deps` inside the function (when we inline it), since the point of `DynamicToolset` is to build it dynamically. We can do that in the MCP server example as well, if that seems more appropriate

tests/test_toolsets.py
```diff
@@ -721,18 +721,28 @@ async def test_visit_and_replace():
         ]
     )
     visited_toolset = toolset.visit_and_replace(lambda toolset: WrapperToolset(toolset))
-    assert visited_toolset == CombinedToolset(
```
DouweM, line 724: Is this change necessary? The new code is a lot harder to read

DouweM (changes requested):

tests/test_temporal.py
dsfaccini: unsure what this was related to

docs/toolsets.md
dsfaccini: removed the wording about the id

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_dynamic_toolset.py
```diff
@@ -0,0 +1,172 @@
...
+
+        self.call_tool_activity = activity.defn(name=f'{activity_name_prefix}__dynamic_toolset__{self.id}__call_tool')(
+            call_tool_activity
+        )
+
+    @property
+    def temporal_activities(self) -> list[Callable[..., Any]]:
+        return [self.get_tools_activity, self.call_tool_activity]
+
+    async def __aenter__(self) -> Self:
```
DouweM: This behavior in aenter/aexit may make sense on `TemporalWrapperToolset`. Then we can also remove the overrides from `TemporalMCPServer`.

pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -1378,7 +1382,7 @@ def _get_toolset(
         # Copy the dynamic toolsets to ensure each run has its own instances
         def copy_dynamic_toolsets(toolset: AbstractToolset[AgentDepsT]) -> AbstractToolset[AgentDepsT]:
```
DouweM, line 1411: Now that we use `run_id` in `DynamicToolset`, I don't think we need to do this anymore, nor do we need the `copy` method

pydantic_ai_slim/pydantic_ai/toolsets/_dynamic.py
```diff
@@ -17,39 +16,92 @@
 """A sync/async function which takes a run context and returns a toolset."""
 
 
-@dataclass
+class ToolsetRunStep(TypedDict, Generic[AgentDepsT]):
```
DouweM: - Could this be a `dataclass`?
- Can we come up with a more clear name? Something with "state" in it

pydantic_ai_slim/pydantic_ai/toolsets/_dynamic.py
```diff
@@ -17,39 +16,92 @@
...
     """
 
     toolset_func: ToolsetFunc[AgentDepsT]
-    per_run_step: bool = True
-
-    _toolset: AbstractToolset[AgentDepsT] | None = None
-    _run_step: int | None = None
+    per_run_step: bool
+    _id: str | None
+    _toolset_runstep: dict[str, ToolsetRunStep[AgentDepsT]]
```
DouweM: Maybe `_run_state`?

pydantic_ai_slim/pydantic_ai/toolsets/_dynamic.py
```diff
@@ -17,39 +16,92 @@
...
+                    result = await run_state['toolset'].__aexit__(*args)
         finally:
-            self._toolset = None
-            self._run_step = None
+            self._toolset_runstep.clear()
+        return result
+
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, DynamicToolset):
+            return False
```
DouweM: It reads a little easier to ad `isinstance(other, DynamicToolset) and` to the next lines

pydantic_ai_slim/pydantic_ai/toolsets/_dynamic.py
```diff
@@ -17,39 +16,92 @@
...
+            for run_state in self._toolset_runstep.values():
+                if run_state['toolset'] is not None:
+                    result = await run_state['toolset'].__aexit__(*args)
         finally:
-            self._toolset = None
-            self._run_step = None
+            self._toolset_runstep.clear()
+        return result
+
+    def __eq__(self, other: object) -> bool:
```
DouweM: Why do we need this now? Because of snapshot equality stuff? Could we add `eq=False` to the new state field? I don't think that internal state should matter to snapshot equality

pydantic_ai_slim/pydantic_ai/toolsets/_dynamic.py
```diff
@@ -17,39 +16,92 @@
...
+            and self.per_run_step == other.per_run_step
+            and self._id == other._id
+            and self._toolset_runstep == other._toolset_runstep  # pyright: ignore[reportUnknownMemberType]
+        )
 
     async def get_tools(self, ctx: RunContext[AgentDepsT]) -> dict[str, ToolsetTool[AgentDepsT]]:
-        if self._toolset is None or (self.per_run_step and ctx.run_step != self._run_step):
-            if self._toolset is not None:
-                await self._toolset.__aexit__()
+        run_id = ctx.run_id or '__default__'
```
DouweM: I'd rather assert if this is not set, because I think it always should be

pydantic_ai_slim/pydantic_ai/toolsets/_dynamic.py
```diff
@@ -17,39 +16,92 @@
...
+
+        if run_id not in self._toolset_runstep:
```
DouweM: We could make the instance variable a `defaultdict` that automatically builds an empty state object

pydantic_ai_slim/pydantic_ai/toolsets/_dynamic.py
```diff
@@ -58,30 +110,43 @@ async def get_tools(self, ctx: RunContext[AgentDepsT]) -> dict[str, ToolsetTool[
...
         else:
-            self._toolset.apply(visitor)
+            for toolset in wrapped_toolsets:
+                toolset.apply(visitor)
 
     def visit_and_replace(
         self, visitor: Callable[[AbstractToolset[AgentDepsT]], AbstractToolset[AgentDepsT]]
     ) -> AbstractToolset[AgentDepsT]:
-        if self._toolset is None:
+        wrapped_items = [(run_id, rs) for run_id, rs in self._toolset_runstep.items() if rs['toolset'] is not None]
```
DouweM: Could this be a dict instead of a list of tuples?

tests/test_temporal.py
```diff
@@ -1104,6 +1104,125 @@ async def test_toolset_without_id():
...
+    """Test that the dynamic toolset agent works correctly outside of a workflow."""
+    result = await dynamic_toolset_temporal_agent.run(
+        'Get the weather for Paris', deps=DynamicToolsetDeps(user_name='Bob')
+    )
+    assert result.output == snapshot('{"get_dynamic_weather":"Weather in a for Bob: sunny."}')
+
+
+# --- MCP-based DynamicToolset test ---
+# Tests that @agent.toolset with an MCP toolset works with Temporal workflows.
+# See https://github.com/pydantic/pydantic-ai/issues/3390
```
DouweM: This is the issue we're fixing, I meant let's fix to the issue that's called out in the next lines that we haven't fixed yet!

tests/test_temporal.py
DouweM: Well we say "Uses per_run_step=False so the toolset persists across run steps within an activity.", but I'm not sure that actually works, as I think we regenerate the dynamic toolset each time, and we explicitly document that we don't respect `per_run_step=False`. And "run steps within an activity" doesn't really make sense

tests/test_temporal.py
```diff
@@ -1104,6 +1104,125 @@ async def test_toolset_without_id():
...
+
+
+@fastmcp_dynamic_toolset_agent.toolset(id='fastmcp_toolset', per_run_step=False)
+def my_fastmcp_dynamic_toolset(ctx: RunContext[None]) -> FastMCPToolset:
+    """Dynamic toolset that returns an MCP toolset.
+
+    This tests MCP lifecycle management (context manager enter/exit) within DynamicToolset + Temporal.
+    Uses per_run_step=False so the toolset persists across run steps within an activity.
+    """
+    return FastMCPToolset('https://mcp.deepwiki.com/mcp', id='dynamic_deepwiki')
```
DouweM: Can we use our `MCPServerStreamableHTTP` instead? Or does that also have the issue of `MCPServerStdio`? In that case, the comment above is incorrect, because it's not just `MCPServerStdio` that has issues, but all `MCPServer`s

pydantic_ai_slim/pydantic_ai/toolsets/_dynamic.py
DouweM: Hmm, I see we do this because of the tests among other things. I'd still rather assert, and then pass a run ID in explicitly.

tests/test_toolsets.py
```diff
@@ -815,10 +824,79 @@ def no_toolset_func(ctx: RunContext[None]) -> None:
...
+    toolset1 = DynamicToolset(toolset_func=lambda ctx: FunctionToolset())
+    toolset2 = DynamicToolset(toolset_func=lambda ctx: FunctionToolset())
+
+    # Same instance
+    assert toolset1 == toolset1
+
+    # Different instances with different functions
+    assert toolset1 != toolset2
+
+    # Compare with non-DynamicToolset
```
DouweM: This doesn't seem really worth testing

DouweM (changes requested):

tests/test_toolsets.py
dsfaccini: yeah that one was just for coverage

pydantic_ai_slim/pydantic_ai/toolsets/_dynamic.py
dsfaccini: `_run_state: _RunState` it is

tests/test_temporal.py
dsfaccini: yeah swapped out for `MCPServerStreamableHTTP`, the above comment is fixed by the change as well

tests/test_temporal.py
dsfaccini: my bad

pydantic_ai_slim/pydantic_ai/toolsets/_dynamic.py
```diff
@@ -18,38 +19,71 @@
...
 
     @property
     def id(self) -> str | None:
-        return None  # pragma: no cover
+        return self._id
+
+    def __eq__(self, other: object) -> bool:
+        return (
+            isinstance(other, DynamicToolset)
+            and self.toolset_func is other.toolset_func  # pyright: ignore[reportUnknownMemberType]
```
DouweM, line 49: Why do we need this ignore?

pydantic_ai_slim/pydantic_ai/toolsets/_dynamic.py
```diff
@@ -58,30 +92,48 @@ async def get_tools(self, ctx: RunContext[AgentDepsT]) -> dict[str, ToolsetTool[
...
+                    toolset=run_state.toolset.visit_and_replace(visitor),
+                    run_step=run_state.run_step,
+                )
+            new_toolset = DynamicToolset(
+                toolset_func=self.toolset_func,
+                per_run_step=self.per_run_step,
+                id=self._id,
+            )
+            new_toolset._run_state = new_run_state
+            return new_toolset
```
DouweM: Can this be simplified using a dict comprehension?

We can probably use `replace` on the `_RunState`

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/toolsets/_dynamic.py
dsfaccini: we can turn it into the dict comprehension but we lose the `defaultdict` and `assert`

```python
new_run_state: dict[str, _RunState[AgentDepsT]] = {
    run_id: _RunState(
        toolset=run_state.toolset.visit_and_replace(visitor),
        run_step=run_state.run_step,
    )
    for run_id, run_state in wrapped_items.items()
    if run_state.toolset is not None
}
```

dsfaccini: about the `replace`, I simply added the `_run_state` param to the class signature
```python
        else:
            new_run_state: dict[str, _RunState[AgentDepsT]] = defaultdict(_RunState)
            for run_id, run_state in wrapped_items.items():
                assert run_state.toolset is not None
                new_run_state[run_id] = _RunState(
                    toolset=run_state.toolset.visit_and_replace(visitor),
                    run_step=run_state.run_step,
                )
            new_toolset = DynamicToolset(
                toolset_func=self.toolset_func, per_run_step=self.per_run_step, id=self._id, _run_state=new_run_state
            )
            return new_toolset
```

pydantic_ai_slim/pydantic_ai/toolsets/_dynamic.py
dsfaccini:
```
PYRIGHT_PYTHON_IGNORE_WARNINGS=1 uv run pyright
/Users/david/projects/forks/add-support-for-toolset-decorator-in-temporal/pydantic_ai_slim/pydantic_ai/toolsets/_dynamic.py
  /Users/david/projects/forks/add-support-for-toolset-decorator-in-temporal/pydantic_ai_slim/pydantic_ai/toolsets/_dynamic.py:63:38 - error: Type of "toolset_func" is partially unknown
    Type of "toolset_func" is "(RunContext[Unknown]) -> (AbstractToolset[Unknown] | Awaitable[AbstractToolset[Unknown] | None] | None)" (reportUnknownMemberType)
1 error, 0 warnings, 0 informations
make: *** [typecheck-pyright] Error 1
```

DouweM:
@dsfaccini As discussed, let's go back to the copy way of doing things

pydantic_ai_slim/pydantic_ai/toolsets/_dynamic.py
```diff
@@ -58,30 +96,35 @@ async def get_tools(self, ctx: RunContext[AgentDepsT]) -> dict[str, ToolsetTool[
...
             super().apply(visitor)
         else:
-            self._toolset.apply(visitor)
+            self._run_state.toolset.apply(visitor)
 
     def visit_and_replace(
         self, visitor: Callable[[AbstractToolset[AgentDepsT]], AbstractToolset[AgentDepsT]]
     ) -> AbstractToolset[AgentDepsT]:
-        if self._toolset is None:
+        if self._run_state is None or self._run_state.toolset is None:
```
DouweM: The repeated `if self._run_state is None or self._run_state.toolset is None:` is a bit uglier than the `if self._toolset is None:` we had before, so I suggest either making that field non-optional with a default empty state object, or just going back to the original 2 instance variables (basically reverting this file except for the id and copy stuff). I don't mind the idea of a state object, but we typically just use private instance variables.

tests/test_temporal.py
```diff
@@ -1110,6 +1110,123 @@ async def test_toolset_without_id():
...
+mcp_dynamic_toolset_agent = Agent(model, name='mcp_dynamic_toolset_agent')
+
+
+@mcp_dynamic_toolset_agent.toolset(id='mcp_toolset')
+def my_mcp_dynamic_toolset(ctx: RunContext[None]) -> MCPServerStreamableHTTP:
+    """Dynamic toolset that returns an MCP toolset.
+
+    This tests MCP lifecycle management (context manager enter/exit) within DynamicToolset + Temporal.
+    """
+    return MCPServerStreamableHTTP('https://mcp.deepwiki.com/mcp', id='dynamic_deepwiki')
```
DouweM: Can you verify that in this case it's not required for the dynamically generated MCP server to have an ID, just the dynamic outer one? We can just drop it here to test that.

tests/test_toolsets.py
```diff
@@ -26,22 +26,23 @@
...
-from pydantic_ai.toolsets._dynamic import DynamicToolset
+from pydantic_ai.toolsets._dynamic import DynamicToolset, _RunState  # pyright: ignore[reportPrivateUsage]
 from pydantic_ai.usage import RunUsage
 
 pytestmark = pytest.mark.anyio
 
 T = TypeVar('T')
 
 
-def build_run_context(deps: T, run_step: int = 0) -> RunContext[T]:
```
DouweM, line 37: Same as up; let's revert this as it's not strictly necessary/better anymore

PR #3501, opened 2025-11-21
Fix media type inference for URLs with query parameters

fedexman:
# Fix media type inference for URLs with query parameters

When using presigned URLs (eg AWS S3) with `ImageUrl`, `AudioUrl`, or `VideoUrl`, the media type inference fails
```bash
https://pics.s3.ap-northeast-1.amazonaws.com/test/Capture-2025-11-21-112402.png?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=fdafdas%2F20251121%2Fap-northeast-1%2Fs3%2Faws4_request&X-Amz-Date=20251121T023200Z&X-Amz-Expires=3600&X-Amz-SignedHeaders=host&X-Amz-Security-Token=IQoJb3JpZ2luX2VjEDoaD
```

to the agent we get this error,
```bash
Internal server error: Could not infer media type from image URL: 
...
Explicitly provide a `media_type` instead
```

the reason is that the `_infer_media_type` function only check the end of the url with `url.endswith('.mkv')` but do not parse the url. 

I propose to parse the url with 
```python
from urllib.parse import urlparse
path = urlparse(self.url).path
if path.endswith('.mkv'):
            return 'video/x-matroska'
```

PR #3686, opened 2025-12-09
feat: support inferring media types from URLs with query params/fragments

h0rv:
Currently, pydantic-ai cannot infer media type for presigned S3 URLs due to the query parameters added to the URL. This improves the infer method to check just the base URL (query parameters and fragments stripped).


h0rv:
We can also use urlib from the std lib if we want to introduce a new dep:

```
from urllib.parse import urlparse

def base_url(url: str) -> str:
    return urlparse(url)._replace(query="", fragment="").geturl()
```

h0rv:
whoops - this is a dupe of https://github.com/pydantic/pydantic-ai/pull/3501

PR #3687, opened 2025-12-09
Encourage the usage of `instructions` in place of `system_prompt` by replacing existing examples and tests

dsfaccini:
`system_prompt` and `instructions` do provide different behaviors, so I didn't replace all instances, but I tried replacing all that weren't specifically targeting that difference

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/bedrock.py
+++ b/pydantic_ai_slim/pydantic_ai/models/bedrock.py
@@ -616,7 +616,7 @@ class BedrockConverseModel(Model):
             if isinstance(message, ModelRequest):
                 for part in message.parts:
                     if isinstance(part, SystemPromptPart):
-                        if part.content:
+                        if part.content:  # pragma: no branch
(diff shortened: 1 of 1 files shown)


tests/models/test_anthropic.py
```diff
@@ -2123,22 +2125,19 @@ async def test_anthropic_model_thinking_part_from_other_model(
     provider = OpenAIProvider(api_key=openai_api_key)
     m = OpenAIResponsesModel('gpt-5', provider=provider)
     settings = OpenAIResponsesModelSettings(openai_reasoning_effort='high', openai_reasoning_summary='detailed')
-    agent = Agent(m, system_prompt='You are a helpful assistant.', model_settings=settings)
```
DouweM, line 2320: If we're going to edit tests that have cassettes, we should re-record the cassettes.

Note that I'd be OK only updating this in the docs (for now).

DouweM:
@dsfaccini Please look at conflicts and CI; I can merge this quickly when those are green

PR #3689, opened 2025-12-10
Fix `BedrockConverseModel` error when `ModelResponse.parts` is empty

ffineis:
### Problem
When a Bedrock model returns a content-filtered response, the ModelResponse has `parts=[]`. Previously, `_map_messages` would still append an empty assistant message to the Bedrock request:

```
bedrock_messages.append({'role': 'assistant', 'content': []})
```

This causes ModelHTTPErrors when sending follow-up requests to Bedrock with an empty assistant message in the conversation history: e.g. `The content field in the Message object at messages.2 is empty...`

### Solution
Added a conditional check to only append assistant messages when there is actual content:
if content: 

```
bedrock_messages.append({'role': 'assistant', 'content': content})
```

This ensures that `ModelResponse` messages with empty `parts` (e.g., from content-filtered responses) are gracefully skipped during message mapping.

Adds tests. Thanks @DouweM for suggesting the fix.

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/bedrock.py
+++ b/pydantic_ai_slim/pydantic_ai/models/bedrock.py
@@ -584,7 +584,8 @@ class BedrockConverseModel(Model):
                     else:
                         assert isinstance(item, ToolCallPart)
                         content.append(self._map_tool_call(item))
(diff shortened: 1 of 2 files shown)


DouweM:
@ffineis Thanks Frank!

PR #3611, opened 2025-12-02
Add `tool_choice` setting

dsfaccini:
Closes #2799
Closes #3092

Adds `tool_choice` to `ModelSettings`, letting users control how the model interacts with function tools.

Currently Pydantic AI internally decides whether to use `tool_choice='auto'` or `'required'` based on output configuration, but users have no way to override this. The workaround was using `extra_body={'tool_choice': 'none'}` which is provider-specific and doesn't work everywhere.

This PR allows the user to set `tool_choice` to:
- 'auto' - model decides whether to call tools
- 'required' - model must call a tool
- 'none' - model can't use function tools
- ['tool_a', 'tool_b'] - model must use one of these specific tools

**One important distinction**: this only affects `function tools` (the ones you register on the agent), not output tools (used internally for structured output). So if you have an agent with `output_type=SomeModel`
and you set `tool_choice='none'`, the output tool stays available - you'll just get a warning about it.

Implementation is spread across all model providers since each has its own API format for tool_choice. 
Added a `resolve_tool_choice` utility that handles validation (checking tool names exist, warning about
conflicts with output tools) and returns a normalized representation that each provider then maps to their specific format.

**Bedrock** is a bit of a special case - it doesn't support 'none' at all, so we fall back to 'auto' with a warning. **Anthropic** has a constraint where 'required' and specific tool selection don't work with thinking/extended thinking enabled.

**TODO**
- [ ] document this somewhere in the docs

Issue #2586, opened 2025-08-16
More flexible tool exception handling


strawgate:
### Description

I think function Toolset is great for taking python code and making it runnable by an Agent.

An exception raised during a function call ends the agent run unless it's a model retry.

This means that often times users have to modify their functions in offer to use them with a function Toolset 

We should think about making the default behavior that it catches exceptions from the function call and raises a model retry and perhaps have an option on the function tool/toolset to do something else (maybe a new exception like ModelFunctionToolError)

DouweM:
@strawgate What kinds of exceptions are you thinking of? Sharing the error message with the model and telling it to retry may make sense in some cases, but really all it can do is change its arguments or just try again verbatim, and the latter would be better handled by retrying the tool call on the Python rather than LLM side (using tenacity like in https://ai.pydantic.dev/retries/). Other errors may also be unrecoverable.

Having some way to configure how to handle specific types of exceptions at the tool or `FunctionToolset` level makes sense to me, as an alternative to having to repeat `except` statements in every tool function, but I don't think treating every exception as "model retryable" by default would make sense. I'd imagine a dict of exception classes (or tuples thereof) pointing to `ModelRetry` or a `tenacity.AsyncRetrying`, or some other classes/string literals codifying exception handling behavior we may like. If you want to retry all errors, you'd end up with something like `on_exception={Exception: ModelRetry}` (all names TBD).

How does that sound to you?

jcaplan:
> I don't think treating every exception as "model retryable" by default would make sense

I definitely think it can make sense depending on the tools you're using.

I decided to switch over from openai agents sdk because on the surface they're quite similar so not a lot of code to rework, but overall there's a few key things in pydantic-ai I like better. but this issue is really hurting me right now.

DouweM:
> I definitely think it can make sense depending on the tools you're using.

@jcaplan I agree, I'm saying it probably shouldn't be the default for Pydantic AI, but it should definitely be something users should be able to configure.

Note that you can already implement this using a custom `WrapperToolset`: https://ai.pydantic.dev/toolsets/#changing-tool-execution. Something like:

```py
class RetryingToolset(WrapperToolset):
    async def call_tool(self, name: str, tool_args: dict[str, Any], ctx: RunContext, tool: ToolsetTool) -> Any:
        try:
            result = await super().call_tool(name, tool_args, ctx, tool)
        except ModelRetry:
            raise
        except Exception as e:
[...]

Issue #1820, opened 2025-05-24
Require agent to use tool(s) before final output


maxvdv94:
### Description

When assigning tools there's no method or parameter to add that forces the model to use the tool. You either have to instruct it through a system_prompt or assume it makes the decision itself to use the tool.

The OpenAI SDK has a tool_choice parameter that can be found [here](https://platform.openai.com/docs/api-reference/responses/create#responses/object-tool_choice).

Can you add this or a similar functionality that forces the model to use a specified tool on the first API request?

Example of the tool_choice parameter:

```
response = client.responses.create(
      model=model,
      tools=[{"type": "web_search_preview"}],
      input=search_query,
      tool_choice="required",
  )
```

### References

https://platform.openai.com/docs/api-reference/responses/create#responses/object-tool_choice

Whadup:
```
agent = Agent(
    ....
        model_settings=OpenAIModelSettings(
            temperature=chat_context.request.temperature,
            top_p=chat_context.request.top_p,
            extra_body={"tool_choice": "required},
        )
)

DouweM:
Do we just want to specify this for the _first_ request to the LLM, or do we want a way to dynamically modify the `tool_choice` for every request to the LLM, through some kind of hook function that takes `RunContext` (similar to `prepare_tools`)?

One factor that complicates this slightly is that PydanticAI uses special output tools for structured output (until we get https://github.com/pydantic/pydantic-ai/pull/1628 to use `format=json_schema` on models that support it), and uses `tool_choice=required` to indicate whether the model has to respond using one of the output tools, or `tool_choice=auto` if it's allowed to respond with plain text as well (e.g. `output_type=str | SomeModel`). 

If we let `tool_choice` be set to explicitly at a per-LLM-call level (rather than through the agent `model_settings` property which is convoluted enough that you'll realize you're now overriding default PydanticAI behavior), we'd have to think about how to handle output tools. It may make sense to _not_ register them with the model at all, so the model doesn't accidentally go straight for an output instead of using one of the user-provided tools.

As different APIs use different values (e.g. "required" or "any" to mean the same thing), it could be called `require_tool_call`, taking either a boolean or a list of possible tool names / references to already-registered tool functions.

simone-gaiarin:
First, could you clarify what you mean by "first request"? Are you referring to the very first call to run, or to the first internal request the agent makes when run is invoked?

[...]

PR #3691, opened 2025-12-10
Adding per tool usage limit

### LINKED ISSUE #3352: Add per-tool usage limits
### Description

Pydantic AI supports per-agent tool call limits, and per-tool retry limits, but it's currently not possible to specify per-tool usage limits, which seems like a gap. This would be useful in cases where some tools are much more expensive than others, and so their usage needs to be more tightly controlled. 

For example, it would be great to be able to do 

```python
@agent.tool(max_uses=2)
async def expensive_tool(ctx: RunContext[Deps], input: str) -> str:
    ...

@agent.tool(max_uses=10)
async def cheap_tool(ctx: RunContext[Deps], input: str) -> str:
    ...
```

The implementation should closely follow the tool [`max_retries`](https://ai.pydantic.dev/api/tools/#pydantic_ai.tools.RunContext.max_retries) and agent [`usage_limits`](https://ai.pydantic.dev/api/usage/#pydantic_ai.usage.UsageLimits) setting:

- The initial system prompt should include information about the max uses for each tool.
- Calling a tool too many times should raise a `UsageLimitExceeded` or similar exception.
- The agent should be allowed to call other tools according to `usage_limits` and any other per-tool usage limits

This should prevent some boilerplate in adding usage counts and limits to the deps and checking this in the tool definition, as in:

```python
@agent.tool
async def my_tool(ctx: RunContext[Deps], input: str) -> str:
    if ctx.deps.tool_uses >= ctx.deps.tool_use_limit:
        return "You have already used the my_tool tool the maximum number of allowed times."

    # tool logic
    ...

    ctx.deps.tool_uses += 1
```

Given potential issues with parallel tools (https://github.com/pydantic/pydantic-ai/issues/3120), I think it's worth a robust pydantic-ai side implementation here.

adtyavrdhn:
Closes #3352 

## Soft Tool Usage Limits

Introduces soft tool usage limits via `ToolPolicy` and `ToolsPolicy`. Unlike the hard `UsageLimits.tool_calls_limit` (raises `UsageLimitExceeded`), soft limits return a message to the model so it can adapt.

### APIs

**`ToolPolicy`** - Per-Tool Limits

Set on individual tools via `@agent.tool(usage_policy=...)`:

```python
from pydantic_ai import Agent, ToolPolicy

@agent.tool_plain(usage_policy=ToolPolicy(max_uses=3, max_uses_per_step=1))
def fetch_record(record_id: int) -> str:
    return f'Record {record_id}'
```

**`ToolsPolicy`** - Agent-Wide Limits

Set on Agent via `tools_policy` or per-run:

```python
from pydantic_ai import Agent, ToolPolicy, ToolsPolicy

agent = Agent('openai:gpt-4o', tools_policy=ToolsPolicy(max_uses=5, max_uses_per_step=2))

# Per-run override with per-tool runtime overrides
result = agent.run_sync(
    '...',
    tools_policy=ToolsPolicy(
        max_uses=10,
        per_tool={'expensive_api': ToolPolicy(max_uses=3)}
    )
)
```

### Parameters

| Parameter | ToolPolicy | ToolsPolicy | Description |
|-----------|:----------:|:-----------:|-------------|
| `max_uses` | ✓ | ✓ | Max successful calls (total run for ToolPolicy, aggregate across all tools for ToolsPolicy) |
| `max_uses_per_step` | ✓ | ✓ | Max successful calls per step |
| `partial_acceptance` | ✓ | ✓ | Accept partial batch or reject all (default: True) |
| `per_tool` | - | ✓ | Runtime per-tool overrides |

### Behavior

- When a per-tool `max_uses` is reached, the tool is removed from available tools
- When aggregate limits are hit, model receives a message to produce output without tools
- `partial_acceptance=True` (default): If model requests 5 calls but only 3 are allowed, 3 accepted, 2 rejected
- `partial_acceptance=False`: All-or-nothing — entire batch rejected if it would exceed limits

### Comparison to UsageLimits

| Parameter | Behavior | Use Case |
|-----------|----------|----------|
| `UsageLimits.tool_calls_limit` | Raises `UsageLimitExceeded` | Hard stop to prevent runaway costs |
| `ToolPolicy.max_uses` | Removes tool / returns message | Limit specific expensive/rate-limited tools |
| `ToolsPolicy.max_uses` | Returns message to model | Soft aggregate limit; model adapts |

### Not Included

Toolset-level handling (e.g., `MCPServer(usage_policy=...)` applying limits across all MCP tools as a group) is not part of this PR. Can take it up later? I am unsure of the value addition it will bring apart from what I've already added.


docs/agents.md
```diff
@@ -669,6 +669,51 @@ except UsageLimitExceeded as e:
     - Usage limits are especially relevant if you've registered many tools. Use `request_limit` to bound the number of model turns, and `tool_calls_limit` to cap the number of successful tool executions within a run.
     - The `tool_calls_limit` is checked before executing tool calls. If the model returns parallel tool calls that would exceed the limit, no tools will be executed.
 
+##### Soft Tool Call Limits with `max_tool_calls`
+
+If you want to limit tool calls but let the model decide how to proceed instead of raising an error, use the `max_tool_calls` parameter. This is a "soft" limit that returns a message to the model when exceeded, rather than raising a [`UsageLimitExceeded`][pydantic_ai.exceptions.UsageLimitExceeded] exception.
+
+```py {test="skip"}
```
DouweM: We shouldn't skip tests unless absolutely necessary

docs/agents.md
```diff
@@ -669,6 +669,51 @@ except UsageLimitExceeded as e:
...
+
+# The model can make up to 2 tool calls
+result = agent.run_sync('Please call the tool three times')
+print(result.output)
+#> I was able to call the tool twice, but the third call reached the limit.
+```
+
+1. Set the maximum number of tool calls allowed during runs. This can also be set per-run.
+
+When `max_tool_calls` is exceeded, instead of executing the tool, the agent returns a message to the model: `'Tool call limit reached for tool "{tool_name}".'`. The model then decides how to respond based on this information.
```
DouweM: This seems like it can be combined with the sentence before the example

docs/agents.md
```diff
@@ -669,6 +669,51 @@ except UsageLimitExceeded as e:
     - Usage limits are especially relevant if you've registered many tools. Use `request_limit` to bound the number of model turns, and `tool_calls_limit` to cap the number of successful tool executions within a run.
     - The `tool_calls_limit` is checked before executing tool calls. If the model returns parallel tool calls that would exceed the limit, no tools will be executed.
 
+##### Soft Tool Call Limits with `max_tool_calls`
```
DouweM: I think this belongs on `tools-advanced.md`, with a link from the usage limits section above.

And no need to mention the name of the field in the title, that'll make it too long for the ToC sidebar

docs/agents.md
```diff
@@ -669,6 +669,51 @@ except UsageLimitExceeded as e:
...
+
+@agent.tool_plain
+def calculate(x: int) -> int:
+    return x * 2
+
+# Override the limit for this specific run
+result = agent.run_sync('Calculate something', max_tool_calls=1)
+```
+
+**When to use `max_tool_calls` vs `tool_calls_limit`:**
```
DouweM: This sounds like something for the explanation in the previous section where we link to this feature. And for the primary docs of this feature where we mention how it's related to the UsageLimits hard limit.

docs/tools-advanced.md
```diff
@@ -411,7 +411,39 @@ If a tool requires sequential/serial execution, you can pass the [`sequential`][
...
 
 !!! note "Limiting tool executions"
-    You can cap tool executions within a run using [`UsageLimits(tool_calls_limit=...)`](agents.md#usage-limits). The counter increments only after a successful tool invocation. Output tools (used for [structured output](output.md)) are not counted in the `tool_calls` metric.
+    You can cap the total number of tool executions within a run using [`UsageLimits(tool_calls_limit=...)`](agents.md#usage-limits). For finer control, you can limit how many times a *specific* tool can be called by setting the `max_uses` parameter when registering the tool (e.g., `@agent.tool(max_uses=3)` or `Tool(func, max_uses=3)`). Once a tool reaches its `max_uses` limit, it is automatically removed from the available tools for subsequent steps in the run. The `tool_calls` counter increments only after a successful tool invocation. Output tools (used for [structured output](output.md)) are not counted in the `tool_calls` metric.
```
DouweM: The bit about `max_uses` should be in the _next_ paragraph that's about soft limits, o that we can leave this one about hard limits.

docs/tools-advanced.md
```diff
@@ -411,7 +411,39 @@ If a tool requires sequential/serial execution, you can pass the [`sequential`][
...
+
+    For a "soft" limit that lets the model decide how to proceed instead of raising an error, use the [`max_tool_calls`](agents.md#soft-tool-call-limits-with-max_tool_calls) parameter on the agent or run method.
+
+#### Raising Hard Errors on Tool Usage Limits
+
+By default, when a tool reaches its `max_uses` limit, it is silently removed from the available tools. If you want to raise an error instead, you can use a [`prepare`](#tool-prepare) function to check the tool usage and raise a [`UsageLimitExceeded`][pydantic_ai.exceptions.UsageLimitExceeded] exception:
```
DouweM: `Agent(max_tool_calls)` vs `tool(max_uses)` feels inconsistent; should both uses calls vs uses?

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -816,6 +818,8 @@ def build_run_context(ctx: GraphRunContext[GraphAgentState, GraphAgentDeps[DepsT
         else DEFAULT_INSTRUMENTATION_VERSION,
         run_step=ctx.state.run_step,
         run_id=ctx.state.run_id,
+        tool_usage=ctx.state.tool_usage,
```
DouweM: This is very similar to `RunContext.retries`, but that one is managed completely in `ToolManager`; do we _have_ to store this on `state` and pass it in here?

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -1039,14 +1054,38 @@ async def _call_tools(
         projected_usage.tool_calls += len(tool_calls)
         usage_limits.check_before_tool_call(projected_usage)
 
+    # Checks for soft limits(if any set on total tools)
+    can_make_tool_calls = tool_manager.can_make_tool_calls(len(tool_calls), deepcopy(usage))
```
DouweM: This seems to duplicate some of what the `if` above does, in terms of copying usage and adding tool call count on it

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -1018,6 +1022,17 @@ async def process_tool_calls(  # noqa: C901
         output_final_result.append(final_result)
 
 
+def _projection_count_of_tool_usage(
```
DouweM: I don't think this is worth a method, it'll be easier to follow if we inline it

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -1039,14 +1054,38 @@ async def _call_tools(
...
+
+    calls_to_run: list[_messages.ToolCallPart] = []
+
+    # For each tool, check how many calls are going to be made
+    tool_call_counts: defaultdict[str, int] = defaultdict(int)
+    _projection_count_of_tool_usage(tool_call_counts, tool_calls)
+
     for call in tool_calls:
-        yield _messages.FunctionToolCallEvent(call)
+        current_tool_use = tool_manager.get_current_use_of_tool(call.tool_name)
```
DouweM: Could this be managed entirely in `ToolManager`? That can already short-circuit what is returned for a given tool call

pydantic_ai_slim/pydantic_ai/_run_context.py
```diff
@@ -48,6 +48,8 @@ class RunContext(Generic[RunContextAgentDepsT]):
     """Instrumentation settings version, if instrumentation is enabled."""
     retries: dict[str, int] = field(default_factory=dict)
     """Number of retries for each tool so far."""
+    tool_usage: dict[str, int] = field(default_factory=dict)
+    """Number of calls for each tool so far."""
```
DouweM: We have to be consistent/specific about calls vs uses and how this relates to retries; does this imply _successful_ calls, or any call at all?

pydantic_ai_slim/pydantic_ai/_tool_manager.py
```diff
@@ -66,11 +66,22 @@ async def for_run_step(self, ctx: RunContext[AgentDepsT]) -> ToolManager[AgentDe
...
+        """
+        if self.tools is None or self.ctx is None:
             raise ValueError('ToolManager has not been prepared for a run step yet')  # pragma: no cover
 
-        return [tool.tool_def for tool in self.tools.values()]
+        result: list[ToolDefinition] = []
+        for tool in self.tools.values():
+            # Filter out tools that have reached their max_uses limit
+            if tool.max_uses is not None:
+                current_uses = self.ctx.tool_usage.get(tool.tool_def.name, 0)
```
DouweM: `get_current_use_of_tool`

pydantic_ai_slim/pydantic_ai/_tool_manager.py
```diff
@@ -66,11 +66,22 @@ async def for_run_step(self, ctx: RunContext[AgentDepsT]) -> ToolManager[AgentDe
...
+                if current_uses >= tool.max_uses:
+                    continue
+            result.append(tool.tool_def)
+        return result
```
DouweM: I'd rather have this be a list comprehension

pydantic_ai_slim/pydantic_ai/_tool_manager.py
```diff
@@ -159,8 +170,11 @@ async def _call_tool(
                 max_retries=tool.max_retries,
                 tool_call_approved=approved,
                 partial_output=allow_partial,
+                max_uses=tool.max_uses,
             )
 
+            self.ctx.tool_usage[name] = self.ctx.tool_usage.get(name, 0) + 1
```
DouweM: To me, "uses" implied _successful_ uses of a potentially costly thing, so excluding retries. Otherwise I think it should definitely be `max_calls`. But I think that only tracking successful uses probably makes sense, so we don't punish models unnecessarily for an arg validation error

pydantic_ai_slim/pydantic_ai/_tool_manager.py
```diff
@@ -272,3 +286,37 @@ async def _call_function_tool(
...
+        """Get the current number of uses of a given tool."""
+        if self.ctx is None:
+            raise ValueError('ToolManager has not been prepared for a run step yet')  # pragma: no cover
+
+        return self.ctx.tool_usage.get(tool_name, 0)
+
+    def _get_max_tool_calls(self) -> int | None:
+        """Get the maximum number of tool calls allowed during this run, or `None` if unlimited."""
+        if self.ctx is None:
+            raise ValueError('ToolManager has not been prepared for a run step yet')  # pragma: no cover
```
DouweM, line 459: Let's move this to a private helper method like `_assert_ctx() -> RunContext`

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/_tool_manager.py
DouweM: If we take it to mean "successful calls", then this part from https://github.com/pydantic/pydantic-ai/issues/3352#issuecomment-3507572317 will require special care, especially if we do it in `ToolManager`, because we don't want to run a `max_uses=1` tool twice in parallel; the >1 calls should fail:

> Another problem is when we are within one run step and the model does parallel function calling to the one-shot tool. How to stop the model technically (not by prompting or returning "do not call me again" - this works, but it is not real [and nice] control) from reusing this tool? Sequential=true appears too harsh since it makes other tool calls withing the same step sequential as well (I think).

pydantic_ai_slim/pydantic_ai/_agent_graph.py
DouweM: It would be interesting to implement this suggestion from https://github.com/pydantic/pydantic-ai/issues/3352#issuecomment-3507572317 at the same time:

> *this is the interesting part - if there was something like ModelPleaseDont exception or other reliable way to stop model from calling a tool within the same step and not break the run (use previous call output) at the same time, solutions above would have become a little cleaner. ModelRetry("Do not retry.") appears wrong and I believe it is not to be used this way.

docs/tools-advanced.md
```diff
@@ -411,7 +411,39 @@ If a tool requires sequential/serial execution, you can pass the [`sequential`][
...
+from pydantic_ai import Agent, RunContext, ToolDefinition
+from pydantic_ai.exceptions import UsageLimitExceeded
+
+agent = Agent('test')
+
+
+async def raise_on_limit(
+    ctx: RunContext[Any], tool_def: ToolDefinition
+) -> ToolDefinition | None:
+    if ctx.max_uses and ctx.tool_usage.get(tool_def.name, 0) >= ctx.max_uses:
```
DouweM: Another interesting example would be to check `ctx.tool_use` and do this from https://github.com/pydantic/pydantic-ai/issues/3352#issuecomment-3513478442:

> A better approach may be to have the tool when output "This tool can only be called M more times." after its real result, so that it works like a counter that increases during the run of the conversation. But that'd definitely be something for the user to do, I think, not Pydantic AI itself.

pydantic_ai_slim/pydantic_ai/_tool_manager.py
```diff
@@ -272,3 +286,37 @@ async def _call_function_tool(
...
+
+        return self.ctx.max_tool_calls
+
+    def can_make_tool_calls(self, num_tool_calls: int, usage: RunUsage) -> bool:
```
Jeremaiha: Shouldn't this require a `semaphore`?
@DouweM

DouweM: @Jeremaiha I don't think so as it's only called from the async event loop, not from threads.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
adtyavrdhn: I think I added it at the time to fix some ruff check, I'll try to inline it :)

pydantic_ai_slim/pydantic_ai/_run_context.py
adtyavrdhn: It implies successful calls only, I will update the docstrings.

docs/tools-advanced.md
adtyavrdhn: Yeah I think it is better to call it` max_tools_uses` instead, successful uses of all tools?
Calls implies to me even the ones that failed or were even not approved.

pydantic_ai_slim/pydantic_ai/_tool_manager.py
adtyavrdhn: Moved it to after the call happens, I agree I want it to mean successful uses and do not want the model to miss out on using the tool because of arg validation errors either.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
adtyavrdhn: Nope, moved the handling into `for_run_step`

pydantic_ai_slim/pydantic_ai/_agent_graph.py
adtyavrdhn: Yeah, reusing projected_usage for this moved it out of the if.

pydantic_ai_slim/pydantic_ai/_tool_manager.py
adtyavrdhn: > If we take it to mean "successful calls", then this part from https://github.com/pydantic/pydantic-ai/issues/3352#issuecomment-3507572317 will require special care, especially if we do it in ToolManager, because we don't want to run a max_uses=1 tool twice in parallel; the >1 calls should fail:

I was thinking both calls should fail because the model is attempting to use a tool twice but only one use is allowed, the response might not be expected by just using the tool once so what do we gain by letting the one call through?
With prompt_templates we will allow the users to override what gets sent to the model which could control the intentional outcome better if the model can again try with just one call(if it makes sense for it to use it once).

The test(`test_tool_max_uses`) sort of mimics my thinking about this scenario.

docs/tools-advanced.md
adtyavrdhn: I'll add this example as well :)

pydantic_ai_slim/pydantic_ai/_agent_graph.py
adtyavrdhn: Refactored into tool_manager.

> It would be interesting to implement this suggestion from https://github.com/pydantic/pydantic-ai/issues/3352#issuecomment-3507572317 at the same time:

I don't think we should just use the previous call's output though, it gives incorrect implication to the model that the response was from that set of args(if any) and I am also not sure about allowing some of the parallel calls and denying the rest. It does move away from UsageLimits behaviour that people might be familiar with. 
We could make this configurable(all-nothing / calls within limit allowed others returned) and default to all-nothing which is same as UsageLimits?

docs/agents.md
adtyavrdhn: Clubbed it together with max_uses and create a small table sort of to explain.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
adtyavrdhn: Okay so I misread this, having a seperate cap on per run step makes sense. We can have total max_uses of 10 and still only want to allow 1 call per step. We should support this.

adtyavrdhn:
Pingg!

vikigenius:
@DouweM anything holding this up? This is a crucial feature for us. I saw the CI logs and the test failures don't really make sense

DouweM:
@vikigenius I'll review it in next few days!

docs/agents.md
DouweM, line 1: Note to self: thoroughly review docs before merge

docs/agents.md
```diff
@@ -677,6 +677,53 @@ except UsageLimitExceeded as e:
     - Usage limits are especially relevant if you've registered many tools. Use `request_limit` to bound the number of model turns, and `tool_calls_limit` to cap the number of successful tool executions within a run.
     - The `tool_calls_limit` is checked before executing tool calls. If the model returns parallel tool calls that would exceed the limit, no tools will be executed.
 
+##### Soft Tool Use Limits
+
+If you want to limit successful tool uses but let the model decide how to proceed instead of raising an error, use the [`ToolsPolicy`][pydantic_ai.ToolsPolicy]. When exceeded, instead of executing the tool, the agent returns a message to the model (`'Tool use limit reached for tool "{tool_name}".'`), allowing it to adapt gracefully rather than raising a [`UsageLimitExceeded`][pydantic_ai.exceptions.UsageLimitExceeded] exception.
```
DouweM: Should specify that tools will be removed after the limit is hit, as this can affect the cache so people need to know. I wonder if once we have `tool_choice` support (https://github.com/pydantic/pydantic-ai/pull/3611), we could disallow specific tools to be used that way.

docs/agents.md
```diff
@@ -677,6 +677,53 @@ except UsageLimitExceeded as e:
...
+
+```py
+from pydantic_ai import Agent, ToolsPolicy
+
+agent = Agent('anthropic:claude-sonnet-4-5', tools_policy=ToolsPolicy(max_uses=2))
```
DouweM: - I like adding an object for this instead of new plain Agent/.run arguments!
- I think `tool_use_policy` is a bit more clear.
- https://github.com/pydantic/pydantic-ai/issues/1820 should then be implemented on this object as well. cc @dsfaccini 
- https://github.com/pydantic/pydantic-ai/pull/3981's new `tool_retries` argument feels like it should be on the policy object as well (as `max_retries`), although that's a hard limit that results in an error, unlike the ones we're adding here...
	- What do you think about supporting hard limits on this policy object as well? 
	- Probably as a mode per limit (hard vs soft vs soft-with-partial-acceptance or something), but maybe there's a scenario where it makes sens to have separate fields for soft and hard with different values.
	- That does raise questions about what we should do with `UsageLimits.tool_calls_limit`, which would be the same as the `ToolsPolicy.max_uses` in hard mode. I'm OK leaving it there (and not deprecating it) because it's conceptually correct to have it on `UsageLimits` as well; we just have to specify which takes precedence.

I definitely think it's worth spending some time on how to unify these various tool arguments.

docs/tools-advanced.md
```diff
@@ -415,7 +415,212 @@ If a tool requires sequential/serial execution, you can pass the [`sequential`][
...
+result = agent.run_sync('Quick search only', tools_policy=ToolsPolicy(max_uses=2))
+```
+
+`ToolsPolicy` provides the following options:
+
+| Option | Description |
+| ------ | ----------- |
+| `max_uses` | Maximum total successful uses allowed across all tools for the entire run. |
+| `max_uses_per_step` | Maximum total successful uses allowed across all tools within a single step. |
+| `per_tool` | A dict mapping tool names to `ToolPolicy` for per-tool overrides. |
```
DouweM: Configuring tool behavior _by name at the top level_ like this is really brittle as there's no type safety (we have to accept any str), and tool names are unpredictable when toolsets are nested, prefixed, prepared, etc, meaning that it's very easy to accidentally use a name that doesn't exist, and we have no way of letting you know (because it's also valid for tools to be filtered out, so an unknown tool name being used is not necessarily a mistake by the user). (We only do it like this on on `TemporalAgent` because we don't really have a choice there)

So:
- I don't want this field here
- If you want to set a tool policy, you have to do it as close to the tool as possible, meaning on the function tool decorator, or on the toolset that itself defines the tool (before it's wrapped by other toolsets that could change the name etc). 
- We should have a `tool_use_policy` argument on toolsets so it can be set for all the included tools in one go
- We should have a way to wrap a toolset and apply a policy to specific tools based on custom logic (e.g. name, metadata), basically by defining a custom function that takes `ToolDefinition` and returns the policy (similar to `PreparedToolset`)

Then I think we can use the same policy object at all levels, rather than having a singular and plural one.

docs/agents.md
DouweM: Also instead of the names "hard" and "soft", let's think about this as "what to do when the limit is hit", implying values/modes like "error" and "model_retry". That's also suggested here for "what to do if a tool raises an error":
- https://github.com/pydantic/pydantic-ai/issues/2586
- `pydantic_ai.toolsets.fastmcp.ToolErrorBehavior = Literal['model_retry', 'error']` that already implements that

That's also a candidate for merging with the policy...

docs/tools-advanced.md
```diff
@@ -415,7 +415,212 @@ If a tool requires sequential/serial execution, you can pass the [`sequential`][
...
+| `partial_acceptance` | Master switch for partial acceptance behavior (default: `None`, inherits `True` behavior). |
+
+#### Partial Acceptance vs All-or-Nothing
+
+By default, when a model requests more tool calls than allowed, Pydantic AI uses **partial acceptance**: it accepts as many calls as the limits allow and rejects the rest individually. This lets the model make progress with the calls that succeeded.
```
DouweM: - Worth checking what the behavior of `UsageLimit.tool_calls_limit` is the analogous case
- "partial acceptance" isn't very clear to me, I don't think of "acceptance" as meaning tool execution. Maybe `execution` is the word to use?
- In general like bool args to have a name that implies the default, e.g. if a field is named `skip_foo` or `disable_foo`, it's extremely likely that the default behavior is to run "foo" value is `False`. If it's just `foo`, it requires looking at the code/docs to know what will happen.
- this is similar to `Agent(end_strategy='early' | 'exhaustive')`, so that naming could also serve as inspiration
- Above I mentioned each limit could have a mode like `error` or `retry`; maybe this could be a third (and fourth) option?

docs/tools-advanced.md
```diff
@@ -415,7 +415,212 @@ If a tool requires sequential/serial execution, you can pass the [`sequential`][
...
+    tools_policy=ToolsPolicy(max_uses=4, partial_acceptance=False)
+)
+```
+
+For per-tool all-or-nothing behavior, set `partial_acceptance=False` on the tool's `ToolPolicy`. This is useful for tools that have transactional semantics or require all their calls to succeed together:
+
+```python
+from pydantic_ai import Agent, ToolPolicy
+
+agent = Agent('openai:gpt-4o')
```
DouweM: Latest gen models please

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -1052,6 +1055,43 @@ async def process_tool_calls(  # noqa: C901
...
+    tool_manager: ToolManager[DepsT],
+    calls_to_run: list[_messages.ToolCallPart],
+    projected_tool_uses: Counter[str],
+    projected_usage: _usage.RunUsage,
+) -> Iterator[_messages.HandleResponseEvent]:
+    accepted_per_tool: Counter[str] = Counter()
+    total_accepted = 0
+
+    for call in tool_calls:
+        rejection_reason = tool_manager.check_tool_call_allowed(
```
DouweM: Functions with many args (like this one, and the one we're calling here) should require kwargs for everything but the first 1 or 2 whose type/name is implied by the function name (i.e. here it's obvious that the first arg will be a tool call, but all the others are not so should be kwargs)

docs/tools-advanced.md
```diff
@@ -415,7 +415,212 @@ If a tool requires sequential/serial execution, you can pass the [`sequential`][
...
+
+Use [`ToolPolicy`][pydantic_ai.ToolPolicy] to control how many times a specific tool can be called. Set it via the `usage_policy` parameter when registering a tool:
+
+```python
+from pydantic_ai import Agent, ToolPolicy
+
+agent = Agent('test')
+
+
+@agent.tool_plain(usage_policy=ToolPolicy(max_uses=3))
```
DouweM, line 432: I'm less sure here that it makes sense to require a whole policy object, instead of just supporting the fields as arguments directly, as everything is already tool related. 

This decorator already has a `retries` arg, and like I mentioned above, I want to see if it makes sense to move those hard limits to the policy as well. So I think we either end up with all args, or with moving `retries` onto the object. And then all-args seems more convenient for the user.

pydantic_ai_slim/pydantic_ai/_run_context.py
```diff
@@ -50,6 +51,8 @@ class RunContext(Generic[RunContextAgentDepsT]):
     """Instrumentation settings version, if instrumentation is enabled."""
     retries: dict[str, int] = field(default_factory=dict)
     """Number of retries for each tool so far."""
+    tools_use_counts: dict[str, int] = field(default_factory=dict)
```
DouweM, line 53: Let's put some effort into making the `RunContext` arguments named consistently (with deprecation, validation aliases for backward compatibility etc), i.e. `retries` should probably also have `tool` in the name, and it should be clearer which fields are for _this_ tool vs _all_ tools.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -1052,6 +1055,43 @@ async def process_tool_calls(  # noqa: C901
         output_final_result.append(final_result)
 
 
+def _handle_tool_calls_parts(
+    tool_calls: list[_messages.ToolCallPart],
+    current_tool_calls: int,
+    output_parts: list[_messages.ModelRequestPart],
+    tool_manager: ToolManager[DepsT],
+    calls_to_run: list[_messages.ToolCallPart],
```
DouweM: It should be clearer which of these are "output args"

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -1069,19 +1109,29 @@ async def _call_tools(
...
 
+    projected_tool_uses = Counter[str]()
     for call in tool_calls:
-        yield _messages.FunctionToolCallEvent(call)
+        projected_tool_uses[call.tool_name] += 1
+
+    calls_to_run: list[_messages.ToolCallPart] = []
+
+    # Process tool calls with partial acceptance - accept as many as allowed by limits
+    for event in _handle_tool_calls_parts(
```
DouweM, line 1137: We have a few levels of nested function calls now that are all called something like "handle/call/process tools" 😄 I'd prefer inlining over yet another layer

DouweM: It's also worth investigating if we can move more of the logic to `ToolManager`, out of the graph.

DouweM (changes requested):
@adtyavrdhn Haven't thoroughly reviewed the code yet, but my comments on the API should give you plenty to work with :)

docs/tools-advanced.md
adtyavrdhn: I was also thinking about a toolset which does the same would be nice. I think it could be cleaner to implement if I can further chip away at the logic currently residing in the graph. I will take a look.

So we are okay with not providing any runtime overriding capabilities here like we do for retries?

What I also had in mind was the behaviour of this toolset, are the limits shared or not within this toolset?

If I want to allow 5 tools to use 7 tool calls, now this distribution is irrelevant. If we assign each tool some value we lose this shared state capability, I am unsure if there are cases when people would need this? What do you think?

docs/agents.md
adtyavrdhn: That makes sense but do you forsee further behavioural requirements(apart from error, retry) propping up later for us to go down the Literal path or is that what makes more sense semantically?

docs/tools-advanced.md
adtyavrdhn: Hmm, I like it being consistent with Agent and clubbing everything together in one object rather than having so many args live on the tool directly but I do see your point as well. I'll think about it

adtyavrdhn: I am currently thinking of allowing the tool to opt in to have a different mode than the agent run as well so that would mean the agent could be raising errors for every tool but this particular tool which opts in to allow model retry. I think this sort of configuration where I really just want to give the user all the control would be better within a n object with contextual closeness? 

Although now I do have this thought as well if with allowing all sorts of configuration I am making it very confusing for the user? Would appreciate your thoughts about this.

docs/tools-advanced.md
adtyavrdhn: I would rather keep it partial execution because it is more of a restriction than what happens when limits exceed which is what I would relate the mode to

adtyavrdhn: I am most concerned about what happens when both get configured? It makes sense to keep it on UsageLimits but I don't like the fact that there are two contradicting ways to configure it.

docs/agents.md
adtyavrdhn: > new tool_retries argument feels like it should be on the policy object as well (as `max_retries`), although that's a hard limit that results in an error, unlike the ones we're adding here...

I don't think it will make a lot of sense because `ToolPolicy` gives you control over the configuration on model-retry or error and having two fields which only raise errors even if documented feels incorrect? 

I think I have added somewhere that I think partial_execution being a boolean gives more sense to me than it being a part of different modes such as `model-retry-partial`?

adtyavrdhn: > That does raise questions about what we should do with UsageLimits.tool_calls_limit, which would be the same as the ToolsPolicy.max_uses in hard mode. I'm OK leaving it there (and not deprecating it) because it's conceptually correct to have it on UsageLimits as well; we just have to specify which takes precedence.


I don't like it being there, allows for conflicting configuration and confusion. I am not sure how straightforward deprecating it might be either. If we keep it, I would prefer to give UsageLimits precedence because it is a known feature and deals with ALL limits and not just tool usage.

pydantic_ai_slim/pydantic_ai/_tool_manager.py
```diff
@@ -181,7 +196,9 @@ async def _call_tool(
                     call.args or {}, allow_partial=pyd_allow_partial, context=ctx.validation_context
                 )
 
-            return await self.toolset.call_tool(name, args_dict, ctx, tool)
+            result = await self.toolset.call_tool(name, args_dict, ctx, tool)
+            self.ctx.tools_use_counts[name] = self.ctx.tools_use_counts.get(name, 0) + 1
+            return result
```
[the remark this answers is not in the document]
adtyavrdhn: Will read more on this but great find if correct.

Issue #3685, opened 2025-12-09
3.14: Tool definition not working with sync_to_async


nairb774:
### Initial Checks

- [x] I confirm that I'm using the latest version of Pydantic AI
- [x] I confirm that I searched for my issue in https://github.com/pydantic/pydantic-ai/issues before opening this issue

### Description

We have a few tool definitions that are wrapped with `sync_to_async` to make the boundary between pydantic-ai and the underlying Django layer work reliably. With Python 3.13 (and prior) this seems to work fine. Attempting to update to 3.14 causes the tool definitions to fail with some unique errors. From the sample below:

```
Traceback (most recent call last):
  File "script.py", line 38, in <module>
    Tool(foo)
    ~~~~^^^^^
  File ".cache/uv/environments-v2/script-b6576b051c3e2f31/lib/python3.14/site-packages/pydantic_ai/tools.py", line 358, in __init__
    self.function_schema = function_schema or _function_schema.function_schema(
                                              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        function,
        ^^^^^^^^^
    ...<3 lines>...
        require_parameter_descriptions=require_parameter_descriptions,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File ".cache/uv/environments-v2/script-b6576b051c3e2f31/lib/python3.14/site-packages/pydantic_ai/_function_schema.py", line 137, in function_schema
    annotation = type_hints[name]
                 ~~~~~~~~~~^^^^^^
KeyError: 'a'
```

From tracing down the code, it looks like the behavior of `functools.update_wrapper` [^1] changed between [3.13](https://github.com/python/cpython/blob/v3.13.11/Lib/functools.py#L34) and [3.14](https://github.com/python/cpython/blob/v3.14.2/Lib/functools.py#L33) - specifically the `__annotations__` wrapper assignment was swapped for `__annotate__`. This causes [`pydantic._internal._typing_extra.get_function_type_hints`](https://github.com/pydantic/pydantic/blob/v2.12.5/pydantic/_internal/_typing_extra.py#L570-L577) to return `{}` resulting in `function_schema` to explode.

I don't have an immediate need for 3.14 support, but this is the current issue we are seeing which is preventing any further evaluation.

[...]

PR #3693, opened 2025-12-10
Use `get_type_hints()` in `function_schema()`

Viicos:
Fixes https://github.com/pydantic/pydantic-ai/issues/3685.

Todo still (will do in a separate PR): update tests in https://github.com/pydantic/pydantic-ai/blob/main/tests/test_function_schema.py to generate the whole function schema to test the actual logic. The issue is that when `function_schema()` relied on `_takes_ctx()` previous to this PR, the logic to unwrap the original function differed: `_takes_ctx()` unwrapped a whole bunch of possible wrappers (properties, classmethods, partial, even classes having a `__call__()`), which is not the case with `_typing_extra.get_function_type_hints()` (which only unwrapped partials).

This is a first step in removing usage of this private Pydantic utility, I will then do follow up PRs.

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/_function_schema.py
+++ b/pydantic_ai_slim/pydantic_ai/_function_schema.py
@@ -7,6 +7,7 @@ from __future__ import annotations as _annotations
 
 from collections.abc import Awaitable, Callable
 from dataclasses import dataclass, field
+from functools import partial
 from inspect import Parameter, signature
 from typing import TYPE_CHECKING, Any, Concatenate, cast, get_origin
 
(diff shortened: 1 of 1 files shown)


pydantic_ai_slim/pydantic_ai/_function_schema.py
```diff
@@ -127,6 +126,9 @@ def function_schema(  # noqa: C901
             errors.append(f'Missing parameter descriptions for {", ".join(missing_params)}')
 
     for index, (name, p) in enumerate(sig.parameters.items()):
+        if index == 0 and takes_ctx is None:
+            takes_ctx = p.annotation is not sig.empty and _is_call_ctx(type_hints[name])
```
Viicos, line 120: Instead of using `_takes_ctx()`, do the logic here. This avoids computing the type hints twice.

pydantic_ai_slim/pydantic_ai/_function_schema.py
```diff
@@ -234,7 +236,7 @@ def function_schema(  # noqa: C901
 TargetCallable = WithCtx[P, R] | WithoutCtx[P, R]
 
 
-def _takes_ctx(callable_obj: TargetCallable[P, R]) -> TypeIs[WithCtx[P, R]]:
+def _takes_ctx(callable_obj: TargetCallable[P, R]) -> TypeIs[WithCtx[P, R]]:  # pyright: ignore[reportUnusedFunction]
```
Viicos, line 236: Still used in one other place in Pydantic AI, will see how this can be cleaned up in a future PR.

DouweM:
@Viicos Thank you!

PR #2611, opened 2025-08-19
Add basic support for UploadedFile UserContent

tarruda:
@DouweM here's some initial support for #2574 

The `UploadedFile` user content is simply wrapping an opaque reference to some return value of provider-specific file upload API, which is then validated in the corresponding model `_map_user_prompt`.

I've only added support for Google and OpenAI, but I believe the API should be flexible enough to add support for other providers. I started working on Anthropic, but decided to leave it out for now as the official SDK doesn't support this feature yet, and I was having trouble referencing it using the SDK data objects.

I've opted to not implement a `Provider.upload_file` abstraction, as the options can be different across providers and I would need to get more familiar with pydantic-ai before feeling confident enough to design a proper abstraction (Can follow up with another PR later!)

One caveat with the tests: The VCR framework apparently doesn't support requests containing binary content, so I had to turn off for uploading files. This is how I proceeded to add the tests:

- Wrote code to upload the file (a new smiley pdf which I've added to tests/assets/smiley.pdf) and turned vcr off.
- Ran the test with a print statement to show the return value
- Commented the code to upload the file (left it as reference for later)
- Re-ran the tests with recording on, and with a literal provider-specific file object with the same id I had previously uploaded.

Since this is just a recording and we are only verifying that we can reference an uploaded file, it probably doesn't matter much that we are not actually running the upload request for now. This can be changed later when VCR is fixed to support this type of request.

Close #2574

PR #3694, opened 2025-12-10
Support `FileUrl.force_download` in `AnthropicModel` and `OpenAIResponsesModel`

### LINKED ISSUE #3569: Review multi-modal input handling to send URLs when possible and don't error on types that are actually supported
### Description

- OpenAI Responses always downloads `FileUrl`s and sends them as `file_data`, but they support `file_url` now.
- Anthropic errors on `text/plain` `BinaryContent`, but this is supported now
- We should always respect `force_download`
- There likely is more

### References

_No response_

dsfaccini:
Closes #3569 - Review multi-modal input handling.

### Changes

**BinaryContent improvements:**
- Add `base64` property that returns raw base64-encoded string
- Update `data` field docstring to clarify it should contain raw bytes, not base64-encoded strings

**Anthropic provider:**
- Add `force_download` support for `ImageUrl` - downloads and sends as base64 when `force_download=True`
- Add `text/plain` support for `BinaryContent` - decodes UTF-8 and sends as plain text document
- Extract `_map_binary_content()` helper to reduce complexity of `_map_user_prompt()`

**Test fixes:**
- Fix `test_mcp_sampling.py`: was storing `base64.b64encode(b'data')` instead of raw `b'data'`
- Fix `test_mistral.py`: was storing base64 ASCII characters as bytes
- Update `test_anthropic.py` error message match for audio content

**OpenAI Responses API:**
- Add `force_download` support for `AudioUrl` and `DocumentUrl` - uses `file_url` to send URLs directly when `force_download=False` (default), downloads and sends as `file_data` when `force_download=True`
- Consolidate `AudioUrl` and `DocumentUrl` handling into single branch

**New tests:**
- `test_binary_content_base64` - verifies new property
- `test_image_url_input_force_download` - verifies Anthropic force_download
- `test_text_document_as_binary_content_input` - verifies Anthropic text/plain support
- `test_document_url_input_response_api` - verifies URL is sent directly (default behavior)
- `test_document_url_input_force_download_response_api` - verifies download when force_download=True

### DIFF
--- a/docs/input.md
+++ b/docs/input.md
@@ -114,8 +114,8 @@ Support for file URLs varies depending on type and provider:
 | [`OpenAIResponsesModel`][pydantic_ai.models.openai.OpenAIResponsesModel] | `ImageUrl`, `AudioUrl`, `DocumentUrl` | — | `VideoUrl` |
 | [`AnthropicModel`][pydantic_ai.models.anthropic.AnthropicModel] | `ImageUrl`, `DocumentUrl` (PDF) | `DocumentUrl` (`text/plain`) | `AudioUrl`, `VideoUrl` |
 | [`GoogleModel`][pydantic_ai.models.google.GoogleModel] (Vertex) | All URL types | — | — |
-| [`GoogleModel`][pydantic_ai.models.google.GoogleModel] (GLA) | [YouTube](models/google.md#document-image-audio-and-video-input), [Files API](https://ai.google.dev/gemini-api/docs/files) | All other URLs | — |
-| [`MistralModel`][pydantic_ai.models.mistral.MistralModel] | `ImageUrl`, `DocumentUrl` (PDF) | — | `AudioUrl`, `VideoUrl` |
+| [`GoogleModel`][pydantic_ai.models.google.GoogleModel] (GLA) | [YouTube](models/google.md#document-image-audio-and-video-input), [Files API](models/google.md#document-image-audio-and-video-input) | All other URLs | — |
+| [`MistralModel`][pydantic_ai.models.mistral.MistralModel] | `ImageUrl`, `DocumentUrl` (PDF) | — | `AudioUrl`, `VideoUrl`, `DocumentUrl` (non-PDF) |
 | [`BedrockConverseModel`][pydantic_ai.models.bedrock.BedrockConverseModel] | S3 URLs (`s3://`) | `ImageUrl`, `DocumentUrl`, `VideoUrl` | `AudioUrl` |
 
 A model API may be unable to download a file (e.g., because of crawling or access restrictions) even if it supports file URLs. For example, [`GoogleModel`][pydantic_ai.models.google.GoogleModel] on Vertex AI limits YouTube video URLs to one URL per request. In such cases, you can instruct Pydantic AI to download the file content locally and send that instead of the URL by setting `force_download` on the URL object:
@@ -131,25 +131,7 @@ DocumentUrl(url='https://example.com/doc.pdf', force_download=True)
 
 ## Uploaded Files
 
-Some model providers like Google's Gemini API support [uploading files](https://ai.google.dev/gemini-api/docs/files). You can upload a file using the provider's client and passing the resulting URL as input:
+Some model providers support passing URLs to files hosted on their platform:
 
-```py {title="file_upload.py" test="skip"}
-from pydantic_ai import Agent, DocumentUrl
-from pydantic_ai.models.google import GoogleModel
-from pydantic_ai.providers.google import GoogleProvider
-
-provider = GoogleProvider()
-file = provider.client.files.upload(file='pydantic-ai-logo.png')
-assert file.uri is not None
-
-agent = Agent(GoogleModel('gemini-2.5-flash', provider=provider))
-result = agent.run_sync(
-    [
-        'What company is this logo from?',
-        DocumentUrl(url=file.uri, media_type=file.mime_type),
-    ]
-)
-print(result.output)
-```
-
-`BedrockConverseModel` supports `s3://<bucket-name>/<object-key>` URIs, provided that the assumed role has the `s3:GetObject` permission. An optional `bucketOwner` query parameter must be specified if the bucket is not owned by the account making the request. For example: `s3://my-bucket/my-file.png?bucketOwner=123456789012`.
+- [`GoogleModel`][pydantic_ai.models.google.GoogleModel] supports the [Files API](models/google.md#document-image-audio-and-video-input) for uploading and referencing files.
+- [`BedrockConverseModel`][pydantic_ai.models.bedrock.BedrockConverseModel] supports `s3://<bucket-name>/<object-key>` URIs, provided that the assumed role has the `s3:GetObject` permission. An optional `bucketOwner` query parameter must be specified if the bucket is not owned by the account making the request. For example: `s3://my-bucket/my-file.png?bucketOwner=123456789012`.
--- a/docs/models/google.md
+++ b/docs/models/google.md
@@ -217,6 +217,27 @@ result = agent.run_sync(
 print(result.output)
 ```
 
+Files can be uploaded via the [Files API](https://ai.google.dev/gemini-api/docs/files) and passed as URLs:
+
+```py {title="file_upload.py" test="skip"}
+from pydantic_ai import Agent, DocumentUrl
+from pydantic_ai.models.google import GoogleModel
+from pydantic_ai.providers.google import GoogleProvider
+
+provider = GoogleProvider()
+file = provider.client.files.upload(file='pydantic-ai-logo.png')
+assert file.uri is not None
+
+agent = Agent(GoogleModel('gemini-2.5-flash', provider=provider))
+result = agent.run_sync(
+    [
+        'What company is this logo from?',
+        DocumentUrl(url=file.uri, media_type=file.mime_type),
+    ]
+)
+print(result.output)
+```
+
 See the [input documentation](../input.md) for more details and examples.
 
 ## Model settings


dsfaccini:
work's still ongoing to address the other points mentioned in the issue

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -1021,24 +1048,20 @@ async def _map_user_prompt(
...
-                                media_type='application/pdf',
-                                type='base64',
-                            ),
-                            type='document',
-                        )
                     else:
-                        raise RuntimeError('Only images and PDFs are supported for binary content')
-                elif isinstance(item, ImageUrl):
-                    yield BetaImageBlockParam(source={'type': 'url', 'url': item.url}, type='image')
+                        yield BetaImageBlockParam(source={'type': 'url', 'url': item.url}, type='image')
```
dsfaccini, line 1096: let me know if you find this diff confusing

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -1049,24 +1076,20 @@ async def _map_user_prompt(
                 elif isinstance(item, CachePoint):
                     yield item
                 elif isinstance(item, BinaryContent):
-                    if item.is_image:
+                    yield AnthropicModel._map_binary_content(item)
+                elif isinstance(item, ImageUrl):
+                    if item.force_download:
+                        downloaded = await download_item(item, data_format='bytes')
```
DouweM, line 1093: We should also respect `force_download` for `DocumentUrl` + `item.media_type == 'application/pdf'` further down, right?

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -1034,6 +1034,33 @@ def _add_cache_control_to_last_param(
...
+    @staticmethod
+    def _map_binary_content(item: BinaryContent) -> BetaContentBlockParam:
+        # Anthropic SDK accepts file-like objects (IO[bytes]) and handles base64 encoding internally
+        if item.is_image:
+            return BetaImageBlockParam(
+                source={'data': io.BytesIO(item.data), 'media_type': item.media_type, 'type': 'base64'},  # type: ignore
+                type='image',
+            )
+        elif item.media_type == 'application/pdf':
+            return BetaBase64PDFBlockParam(
```
DouweM: This is now an alias for `BetaRequestDocumentBlockParam`, so let's use the newer name

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -1034,6 +1034,33 @@ def _add_cache_control_to_last_param(
...
+                source=BetaBase64PDFSourceParam(
```
DouweM, line 1060: I'm looking at what other sources this supports, and there's also `BetaFileDocumentSourceParam`, which takes a `file_id` for the file upload API. 

We're adding support for uploaded files in https://github.com/pydantic/pydantic-ai/pull/2611, but that PR has been stale for a bit so may be interesting for you to pick up.

pydantic_ai_slim/pydantic_ai/models/anthropic.py
DouweM: Would it make sense to use `_map_binary_content` there, if we make it more generic so it can take the result of `download_item`? (Or take a `FileUrl | BinaryContent` and if it gets a `FileUrl`, do the download first)

tests/models/test_anthropic.py
```diff
@@ -1486,6 +1486,32 @@ async def test_image_url_input(allow_model_requests: None, anthropic_api_key: st
...
+    result = await agent.run(
+        [
+            'What is this vegetable?',
+            ImageUrl(
+                url='https://t3.ftcdn.net/jpg/00/85/79/92/360_F_85799278_0BBGV9OAdQDTLnKwAPBCcg1J7QtiieJY.jpg',
+                force_download=True,
+            ),
+        ]
+    )
+    assert result.output == snapshot(
```
DouweM, line 1509: Can we test that the forced download is actually happening?

tests/models/test_openai.py
```diff
@@ -873,6 +873,33 @@ async def test_document_url_input(allow_model_requests: None, openai_api_key: st
...
+
+async def test_document_url_input_force_download_response_api(allow_model_requests: None, openai_api_key: str):
+    """Test DocumentUrl with force_download=True downloads and sends as file_data."""
+    provider = OpenAIProvider(api_key=openai_api_key)
+    m = OpenAIResponsesModel('gpt-4.1-nano', provider=provider)
+    agent = Agent(m)
+
+    document_url = DocumentUrl(
+        url='https://www.w3.org/WAI/ER/tests/xhtml/testfiles/resources/pdf/dummy.pdf',
+        force_download=True,
```
DouweM, line 903: Same as up -- we're not testing that we're respecting this, although I suppose that fact that those lines have test coverage sort of does? But it'd be good if we could assert that explicitly

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -574,7 +578,12 @@ def identifier(self) -> str:
     @property
     def data_uri(self) -> str:
         """Convert the `BinaryContent` to a data URI."""
-        return f'data:{self.media_type};base64,{base64.b64encode(self.data).decode()}'
+        return f'data:{self.media_type};base64,{self.base64}'
+
+    @property
+    def base64(self) -> str:
```
DouweM, line 583: We're not using this anywhere but in the test! I searched for `base64.b64encode(` and found at least one place we could use it: in `UserPromptPart.otel_message_parts`

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/messages.py
dsfaccini: there are a couple places actually, I replaced all the ones I found

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -474,7 +474,11 @@ class BinaryContent:
     """Binary content, e.g. an audio or image file."""
 
     data: bytes
-    """The binary data."""
+    """Arbitrary binary data.
```
DouweM: "Arbitrary" may be a little too strong, as it is specifically a file matching the type in `media_type` 😄 

I think if we make this `The binary file data`, we can remove the specific mention of base64 that's added in the next lines

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -1034,6 +1034,35 @@ def _add_cache_control_to_last_param(
...
+        elif media_type == 'text/plain':
+            return BetaRequestDocumentBlockParam(
+                source=BetaPlainTextSourceParam(data=data.decode('utf-8'), media_type=media_type, type='text'),
+                type='document',
+            )
+        else:
+            raise RuntimeError(f'Unsupported binary content media type for Anthropic: {media_type}')
+
+    @staticmethod
+    def _map_binary_content(item: BinaryContent) -> BetaContentBlockParam:
```
DouweM: We don't need this method anymore as it's just one line that can be inlined

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -1034,6 +1034,35 @@ def _add_cache_control_to_last_param(
         # Add cache_control to the last param
         last_param['cache_control'] = self._build_cache_control(ttl)
 
+    @staticmethod
+    def _map_binary_to_block(data: bytes, media_type: str) -> BetaContentBlockParam:
```
DouweM: "to block" is implied, so I'd just make this `_map_binary_data`

pydantic_ai_slim/pydantic_ai/models/anthropic.py
DouweM: I wonder if `download_item` could be refactored to return a `BinaryContent`

dsfaccini: would've been cool if it had been like `BinaryContent.download` or `.from_url` but I don't think the replacement is worth the hassle now

pydantic_ai_slim/pydantic_ai/messages.py
dsfaccini: I'd like to keep one as it reduces the likelihood of users using it incorrectly

dsfaccini:
changed the docs as follows

image

`force_download` snippet -> better than describing it (personally, but also for llms)

dsfaccini:
## Addendum

`force_download=False` passthrough mode (addresses #3724):

Changed `force_download` from `bool` to `bool | None` to support explicit URL passthrough:

| Value          | Behavior                                                       |
|----------------|----------------------------------------------------------------|
| None (default) | Provider decides - downloads if needed, sends URL if supported |
| True           | Always download content locally and send as bytes              |
| False          | Never download - pass URL directly to API (useful for proxies) |

This enables use cases where users have middleware/proxies that intercept requests and handle custom URL schemes (e.g., file://).

Provider changes:
- All providers: Updated conditionals from if `item.force_download`: to `if item.force_download is True:`
- Google/Gemini: `force_download=False` bypasses `google-gla` download requirement
- Anthropic: force_download=False on text/plain `DocumentUrl` passes URL through (API will reject, it though unless the user knows what they're doing)
- Bedrock: Raises `RuntimeError` when `force_download=False` since Bedrock doesn't support URLs at all

New tests:
- `test_document_url_passthrough_non_http_url` - verifies non-http URLs pass through without validation
- `test_url_force_download_false_raises_error` - verifies Bedrock raises clear error
- `test_text_document_url_input_no_force_download` - verifies Anthropic API error for unsupported passthrough

Docs:
- Updated docs/input.md with force_download behavior table and examples

This is how the doc looks like after this change:
image

docs/input.md
```diff
@@ -104,21 +104,55 @@ print(result.output)
 
 ## User-side download vs. direct file URL
 
-When you provide a URL using any of `ImageUrl`, `AudioUrl`, `VideoUrl` or `DocumentUrl`, Pydantic AI will typically send the URL directly to the model API so that the download happens on their side.
+When using one of `ImageUrl`, `AudioUrl`, `VideoUrl` or `DocumentUrl`, Pydantic AI will default to sending the URL to the model, so the file is downloaded on their side.
```
DouweM:
```suggestion
When using one of `ImageUrl`, `AudioUrl`, `VideoUrl` or `DocumentUrl`, Pydantic AI will default to sending the URL to the model provider, so the file is downloaded on their side.
```

docs/input.md
```diff
@@ -104,21 +104,55 @@ print(result.output)
...
 
-Some model APIs do not support file URLs at all or for specific file types. In the following cases, Pydantic AI will download the file content and send it as part of the API request instead:
+Support for file URLs varies depending on type and provider. Pydantic AI handles this as follows:
```
DouweM:
```suggestion
Support for file URLs varies depending on type and provider:
```

docs/input.md
```diff
@@ -104,21 +104,55 @@ print(result.output)
 
 ## User-side download vs. direct file URL
 
-When you provide a URL using any of `ImageUrl`, `AudioUrl`, `VideoUrl` or `DocumentUrl`, Pydantic AI will typically send the URL directly to the model API so that the download happens on their side.
+When using one of `ImageUrl`, `AudioUrl`, `VideoUrl` or `DocumentUrl`, Pydantic AI will default to sending the URL to the model, so the file is downloaded on their side.
 
-Some model APIs do not support file URLs at all or for specific file types. In the following cases, Pydantic AI will download the file content and send it as part of the API request instead:
```
DouweM, line 109: We still need a sentence like this to explain what will happen if the table below has "Sends URL directly" as false, especially because the "false" is not explicitly stated but implied because of something like "`ImageUrl`, `AudioUrl`, `DocumentUrl` | `ImageUrl` only". It's not currently clear to the user what happens for `AudioUrl` and `DocumentUrl`

DouweM: Maybe it should be columns: "Send URL directly | Download URL and send bytes | Unsupported", listing each type explicitly for each provider

docs/input.md
```diff
@@ -104,21 +104,55 @@ print(result.output)
...
+
+```py {title="force_download_true.py" test="skip" lint="skip"}
+from pydantic_ai import ImageUrl, DocumentUrl
+
+# Force local download when the API can't fetch the URL
+ImageUrl(url='https://example.com/image.png', force_download=True)
+DocumentUrl(url='https://example.com/doc.pdf', force_download=True)
+```
+
+Use `force_download=False` to pass URLs directly without any download attempt. This is useful when you have a proxy or middleware that intercepts requests:
```
DouweM: So this is only relevant for models that support URLs of a given type, but for some URLs of that type prefer a download anyway, and now we're giving the user the option to in those cases send the URL anyway? 

I think that's _way too niche_ a feature to support, as it just affects people putting a file-URL-interpreting proxy in front of Google GLA, and the specific Anthropic+`text/plain` combination, which seems incredibly uncommon.

This doesn't solve https://github.com/pydantic/pydantic-ai/issues/3724's problem anyway, as they were using `OpenAIChatModel` which doesn't support document URLs anyway, only image URLs.

So please revert these changes, reopen that issue, and maybe we can find out what he would expect to happen because OpenAI Chat doesn't have a way of passing document URLs though anyway.

Edit: Also note that this is the risk of adding scope to an existing PR; I may have been able to merge this now and then we could have done this new feature in a followup PR, but now this entire PR is blocked waiting for it to be removed

experiments/test_anthropic_changes.py
DouweM, line 1: No experiments please

docs/input.md
dsfaccini: sorry for that, I've those changes now

docs/input.md
dsfaccini: leaving this open for checking it tomorrow

dsfaccini: image

docs/input.md
```diff
@@ -104,21 +104,34 @@ print(result.output)
...
-- [`BedrockConverseModel`][pydantic_ai.models.bedrock.BedrockConverseModel]: All URLs except S3 URLs, specifically starting with `s3://`.
+| Model | Send URL directly | Download and send bytes | Unsupported |
+|-------|-------------------|-------------------------|-------------|
+| [`OpenAIChatModel`][pydantic_ai.models.openai.OpenAIChatModel] | `ImageUrl` | `AudioUrl`, `DocumentUrl` | `VideoUrl` |
+| [`OpenAIResponsesModel`][pydantic_ai.models.openai.OpenAIResponsesModel] | `ImageUrl`, `AudioUrl`, `DocumentUrl` | — | `VideoUrl` |
+| [`AnthropicModel`][pydantic_ai.models.anthropic.AnthropicModel] | `ImageUrl`, `DocumentUrl` (PDF) | `DocumentUrl` (`text/plain`) | `AudioUrl`, `VideoUrl` |
+| [`GoogleModel`][pydantic_ai.models.google.GoogleModel] (Vertex) | All URL types | — | — |
+| [`GoogleModel`][pydantic_ai.models.google.GoogleModel] (GLA) | [YouTube](models/google.md#document-image-audio-and-video-input), [Files API](https://ai.google.dev/gemini-api/docs/files) | All other URLs | — |
+| [`MistralModel`][pydantic_ai.models.mistral.MistralModel] | `ImageUrl`, `DocumentUrl` (PDF) | — | `AudioUrl`, `VideoUrl` |
```
DouweM: Does this imply that "`DocumentUrl` (non-PDF)" should be in Unsupported?

docs/models/google.md
```diff
@@ -199,7 +199,25 @@ agent = Agent(model)
 
 ## Document, Image, Audio, and Video Input
 
-`GoogleModel` supports multi-modal input, including documents, images, audio, and video. See the [input documentation](../input.md) for details and examples.
+`GoogleModel` supports multi-modal input, including documents, images, audio, and video.
+
+YouTube video URLs can be passed directly to Google models:
```
DouweM, line 204: If we make this edit, I think we should add the Files API example from input.md "Uploaded Files" as well

tests/models/test_anthropic.py
DouweM: The new test below does this now, but we still have still have this old test, with a near-identical name 😄

Issue #3690, opened 2025-12-10
eval crashes when evaluating an image


Clement-Lelievre:
### Initial Checks

- [x] I confirm that I'm using the latest version of Pydantic AI
- [x] I confirm that I searched for my issue in https://github.com/pydantic/pydantic-ai/issues before opening this issue

### Description

When setting up an eval with a `LLMJudge` supposed to look at an image, the eval fails because the image output is stringified and thus breaches the max nb of tokens allowed. see [here](https://github.com/pydantic/pydantic-ai/blob/f42e52346dfb7189dfb5216412bd52e5c98f981d/pydantic_evals/pydantic_evals/evaluators/llm_as_a_judge.py#L233-L250)

### Example Code

```Python
from pydantic_ai import Agent
from pydantic_ai.messages import BinaryImage
from pydantic_evals import Case
from pydantic_evals import Dataset
from pydantic_evals.evaluators import LLMJudge

dataset = Dataset(
    cases=[
        Case(
            inputs="make an image of a cat",
        ),
    ],
    evaluators=[LLMJudge(rubric="The image depicts a cat")],
)


async def ask_agent(
    prompt: str,
) -> BinaryImage:
    agent = Agent("google-gla:gemini-3-pro-image-preview", output_type=BinaryImage)
    result = await agent.run(prompt)
    return result.output


report = await dataset.evaluate(ask_agent)
report.print()
"""
                                           Evaluation Summary: ask_agent                                           
┏━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━┓
┃ Case ID  ┃ Evaluator Failures                                                                        ┃ Duration ┃
┡━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━┩
│ Case 1   │ LLMJudge: ModelHTTPError: status_code: 400, model_name: gpt-4o, body: {'message': "This   │    14.2s │
│          │ model's maximum context length is 128000 tokens. However, your messages resulted in       │          │
[...]

PR #3696, opened 2025-12-10
Support multi-modal output in `LLMJudge`

Clement-Lelievre:
fixes https://github.com/pydantic/pydantic-ai/issues/3690

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -171,7 +171,7 @@ jobs:
 
       - uses: denoland/setup-deno@v2
         with:
-          deno-version: v2.x
+          deno-version: v2.5.x
 
       - run: mkdir .coverage
 
@@ -222,7 +222,7 @@ jobs:
 
       - uses: denoland/setup-deno@v2
         with:
-          deno-version: v2.x
+          deno-version: v2.5.x
 
       - run: mkdir .coverage
 
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -41,7 +41,7 @@ jobs:
 
       - uses: denoland/setup-deno@v2
         with:
-          deno-version: v2.x
+          deno-version: v2.5.x
 
       - run: uv tool install pre-commit
 
--- a/docs/api/models/cerebras.md
+++ b/docs/api/models/cerebras.md
@@ -1,0 +1,7 @@
+# `pydantic_ai.models.cerebras`
+
+## Setup
+
+For details on how to set up authentication with this model, see [model configuration for Cerebras](../../models/cerebras.md).
+
+::: pydantic_ai.models.cerebras
--- a/docs/models/cerebras.md
+++ b/docs/models/cerebras.md
@@ -1,0 +1,77 @@
+# Cerebras
+
+## Install
+
+To use `CerebrasModel`, you need to either install `pydantic-ai`, or install `pydantic-ai-slim` with the `cerebras` optional group:
+
+```bash
+pip/uv-add "pydantic-ai-slim[cerebras]"
+```
+
+## Configuration
+
+To use [Cerebras](https://cerebras.ai/) through their API, go to [cloud.cerebras.ai](https://cloud.cerebras.ai/?utm_source=3pi_pydantic-ai&utm_campaign=partner_doc) and generate an API key.
+
+For a list of available models, see the [Cerebras models documentation](https://inference-docs.cerebras.ai/models).
+
+## Environment variable
+
+Once you have the API key, you can set it as an environment variable:
+
+```bash
+export CEREBRAS_API_KEY='your-api-key'
+```
+
+You can then use `CerebrasModel` by name:
+
+```python
+from pydantic_ai import Agent
+
+agent = Agent('cerebras:llama-3.3-70b')
+...
+```
+
+Or initialise the model directly with just the model name:
+
+```python
+from pydantic_ai import Agent
+from pydantic_ai.models.cerebras import CerebrasModel
+
+model = CerebrasModel('llama-3.3-70b')
+agent = Agent(model)
+...
+```
+
+## `provider` argument
+
+You can provide a custom `Provider` via the `provider` argument:
+
+```python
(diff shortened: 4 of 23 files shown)


tests/evals/test_llm_as_a_judge.py
```diff
@@ -326,26 +350,26 @@ async def test_judge_input_output_expected_with_model_settings_mock(
...
-                '</Input>',
-                """\
+            """\
+<Input>
+
+
+123
+
+</Input>
+
```
dsfaccini: there are a couple too many line breaks here, where does the change come from?

Clement-Lelievre: @dsfaccini dropped them, lmk if lgty now

pydantic_evals/pydantic_evals/evaluators/llm_as_a_judge.py
```diff
@@ -221,39 +221,44 @@ def _stringify(value: Any) -> str:
         return repr(value)
 
 
+def _add_section(sections: list[str | UserContent], content: Any, tag: str) -> None:
+    """Add a tagged section *inplace* to the sections list, handling different content types."""
```
DouweM: There doesn't seem to be a particular reason why this needs to add to an existing list, so I'd prefer to just have it return a list that we can then combine at the call site

pydantic_evals/pydantic_evals/evaluators/llm_as_a_judge.py
```diff
@@ -221,39 +221,44 @@ def _stringify(value: Any) -> str:
...
-            else:
-                sections.append(_stringify(inputs))
-            sections.append('</Input>')
-
-    sections.append(f'<Output>\n{_stringify(output)}\n</Output>')
+        _add_section(sections, inputs, 'Input')
+
+    _add_section(sections, output, 'Output')
+
     sections.append(f'<Rubric>\n{rubric}\n</Rubric>')
```
DouweM: Let's use the method here as well just for consistency.

DouweM (changes requested):

tests/evals/test_llm_as_a_judge.py
```diff
@@ -166,6 +166,30 @@ async def test_judge_input_output_binary_content_list_mock(mocker: MockerFixture
...
+    assert result.reason == 'Test passed'
+    assert result.pass_ is True
+    assert result.score == 1.0
+
+    # Verify the agent was called with correct prompt
+    mock_run.assert_called_once()
+    raw_prompt = mock_run.call_args[0][0]
+
+    # 1) It must be a list
+    assert isinstance(raw_prompt, list), 'Expected prompt to be a list when passing binary'
```
DouweM: Can we instead test `assert call_args == snapshot()` as we have below? When you run the test, the snapshot will be filled in automatically

pydantic_evals/pydantic_evals/evaluators/llm_as_a_judge.py
Clement-Lelievre: done

tests/evals/test_llm_as_a_judge.py
Clement-Lelievre: I did that, except the problem with `BinaryContent` is that the `data` attr repr is extremely large and that would clutter the file.
I instead replaced it with the `image_content` fixture used in the test and `snapshot()` is happy with it

pydantic_evals/pydantic_evals/evaluators/llm_as_a_judge.py
Clement-Lelievre: done

pydantic_evals/pydantic_evals/evaluators/llm_as_a_judge.py
```diff
@@ -221,39 +221,51 @@ def _stringify(value: Any) -> str:
...
+            for item in content:  # type: ignore[reportUnknownVariableType]
+                if isinstance(item, str | MultiModalContent):
+                    sections.append(item)
+                else:
+                    sections.append(_stringify(item))
+        elif isinstance(content, MultiModalContent):
+            sections.append(content)
+        else:
+            sections.append(_stringify(content))
+        sections.append(f'</{tag}>')
```
DouweM: Now that we do the joining-if-all-strings at the end, I think this can be simplified as follows:

```suggestion
    content = content if isinstance(content, Sequence) else [content]
    
    sections.append(f'<{tag}>')
    for item in content:  # type: ignore[reportUnknownVariableType]
        if isinstance(item, str | MultiModalContent):
            sections.append(item)
        else:
            sections.append(_stringify(item))
    sections.append(f'</{tag}>')
```

DouweM (changes requested):

pydantic_evals/pydantic_evals/evaluators/llm_as_a_judge.py
Clement-Lelievre: @DouweM  I can see your refactoring point, however as is, no your snippet will not produce the intended result bcs strings are instances of `Sequence` and therefore we would newline after each character.
Then, with this out of the way (easily fixable), this means we change a little the list we produce. Specifically when at least one item in `sections` is not a str, then we return a list of all the items instead of concatenating the string parts.

Example above, look at the **dummy rubric** part: it goes from one concatenated string to its parts.

Can you confirm which format is best for the LLMJudge to receive, if any? Then I can proceed with the changes.


Screenshot 2025-12-10 at 19 20 29

DouweM: @Clement-Lelievre Ah good point on str being a sequence :) So yeah then we have to do `isinstance(str)` first.

The rubric being broken up into items is OK, the model will receive/interpret it identically.

Clement-Lelievre: ok done

(so in fact there's no strong reason to concatenate into a single str when all items are strs, but I left it as is as it doesn't harm)

Issue #2805, opened 2025-09-04
Sync instructions, output, validator functions don't work with Temporal


DouweM:
### Initial Checks

- [x] I confirm that I'm using the latest version of Pydantic AI
- [x] I confirm that I searched for my issue in https://github.com/pydantic/pydantic-ai/issues before opening this issue

### Description

This doesn't raise an error, but causes the Temporal workflow to stall for some reason:

```py
complex_agent = Agent(
    model,
    deps_type=Deps,
    name='agent',
)


@agent.instructions
def instructions(ctx: RunContext[Deps]) -> str:
    return 'Always be kind.'

temporal_agent = TemporalAgent(agent)

# now use the temporal_agent in a workflow
```

### Example Code

```Python

```

### Python, Pydantic AI & LLM client version

```Text
Pydantic AI v1.0.0b1
```

PR #3537, opened 2025-11-24
Allow `TemporalAgent` to switch model at `agent.run`-time

mattbrandman:
Fixes https://github.com/pydantic/pydantic-ai/issues/3407

This pull request introduces enhanced support for model selection and management within Temporal agents, as well as improved handling of run context propagation. The main changes allow registering multiple models with a Temporal agent, selecting models by name or provider string at runtime (inside workflows), and ensuring the current run context is properly tracked across async boundaries. These improvements make it easier to use and configure multiple models in Temporal workflows, while maintaining safety and clarity in model selection.

**Model selection and registration for Temporal agents:**

- Added support for registering multiple models with a Temporal agent via the new `additional_models` argument, and for selecting a model by name or provider string at runtime within workflows. This includes validation to prevent duplicate or invalid model names and ensures that only registered models or provider strings can be selected during workflow execution. (`pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py`, [[1]](diffhunk://#diff-190f8dc438fdab2b98b4f4d9705e01f01749994ad76f7606439d83f81fa7469eR60-R61) [[2]](diffhunk://#diff-190f8dc438fdab2b98b4f4d9705e01f01749994ad76f7606439d83f81fa7469eR87-R93) [[3]](diffhunk://#diff-190f8dc438fdab2b98b4f4d9705e01f01749994ad76f7606439d83f81fa7469eR115-R127) [[4]](diffhunk://#diff-190f8dc438fdab2b98b4f4d9705e01f01749994ad76f7606439d83f81fa7469eL123-L134) [[5]](diffhunk://#diff-190f8dc438fdab2b98b4f4d9705e01f01749994ad76f7606439d83f81fa7469eR169-R184) [[6]](diffhunk://#diff-190f8dc438fdab2b98b4f4d9705e01f01749994ad76f7606439d83f81fa7469eR350) [[7]](diffhunk://#diff-190f8dc438fdab2b98b4f4d9705e01f01749994ad76f7606439d83f81fa7469eL358-R417) [[8]](diffhunk://#diff-190f8dc438fdab2b98b4f4d9705e01f01749994ad76f7606439d83f81fa7469eL890-R950)

[...]

PR #3704, opened 2025-12-11
Fix using sync history processors, instructions functions, and output functions with `TemporalAgent`

### LINKED ISSUE #3607: AnyIO worker exception due to sync history processor function in TemporalAgent
### Initial Checks

### Description

When a sync history processor function is used in conjunction with a `TemporalAgent` instance, the function is executed within an AnyIO loop used by PydanticAI, which causes issues due to the event loop being overridden by Temporal (the history processor is executed within the workflow).

This results in an exception being thrown from the worker thread, due to the Temporal SDK not implementing `is_closed`:
```
<redacted>  | Exception in thread AnyIO worker thread:
<redacted>  | Traceback (most recent call last):
<redacted>  |   File "/<redacted>/cpython-3.13.9-linux-aarch64-gnu/lib/python3.13/threading.py", line 1043, in _bootstrap_inner
<redacted>  |     self.run()
<redacted>  |     ~~~~~~~~^^
<redacted>  |   File "/<redacted>/site-packages/anyio/_backends/_asyncio.py", line 973, in run
<redacted>  |     if not self.loop.is_closed():
<redacted>  |            ~~~~~~~~~~~~~~~~~~~^^
<redacted>  |   File "/<redacted>/cpython-3.13.9-linux-aarch64-gnu/lib/python3.13/asyncio/events.py", line 251, in is_closed
<redacted>  |     raise NotImplementedError
<redacted>  | NotImplementedError
```

The workaround is to make the history processor `async`, which makes it be executed within the workflow, but that can potentially cause issue if the history processor did any non-deterministic actions, as Temporal might raise an error then.

### Example Code

```Python
def dummy_history_processor(messages: list[ModelMessage]) -> list[ModelMessage]:
    return messages[1:] # some dummy processing


MY_AGENT = TemporalAgent(
    Agent(
        ...
        history_processors=[dummy_history_processor],
    )
)
```

```Text
Python: CPython 3.13.9
Pydantic: 1.23.0
LLM client: `anthropic 0.72.0`
```

dsfaccini:
- Fixes #3607
- Fixes https://github.com/pydantic/pydantic-ai/issues/2805

## Problem

Sync history processors fail inside Temporal workflows because `anyio.to_thread.run_sync` tries to create threads, which Temporal's sandboxed event loop doesn't support. This causes `NotImplementedError`.

## Solution

Use a ContextVar to control when `run_in_executor()` should execute sync functions directly (blocking) vs using threading. This follows the approach discussed in the issue comments and approved by @DouweM.

## Changes

- **`_utils.py`**: Added `_prefer_blocking_execution` ContextVar, removed Temporal detection code
- **`temporal/_agent.py`**: Set ContextVar in `_temporal_overrides()` context manager
- **`test_utils.py`**: Updated test to use ContextVar instead of mocking Temporal internals
- **`test_temporal.py`**: Added integration test for sync history processors in workflows

## Testing

- Unit test verifies ContextVar controls blocking behavior
- Integration test validates sync history processors work in Temporal workflows

### DIFF
--- a/pydantic_ai_slim/pyproject.toml
+++ b/pydantic_ai_slim/pyproject.toml
@@ -53,7 +53,7 @@ Changelog = "https://github.com/pydantic/pydantic-ai/releases"
 
 [tool.hatch.metadata.hooks.uv-dynamic-versioning]
 dependencies = [
-    "griffe>=1.3.2",
+    "griffe>=1.14.0",
     "httpx>=0.27",
     "pydantic>=2.10",
     "pydantic-graph=={{ version }}",
--- a/uv.lock
+++ b/uv.lock
@@ -5685,7 +5685,7 @@ requires-dist = [
     { name = "genai-prices", specifier = ">=0.0.40" },
     { name = "google-auth", marker = "extra == 'vertexai'", specifier = ">=2.36.0" },
     { name = "google-genai", marker = "extra == 'google'", specifier = ">=1.55.0" },
-    { name = "griffe", specifier = ">=1.3.2" },
+    { name = "griffe", specifier = ">=1.14.0" },
     { name = "groq", marker = "extra == 'groq'", specifier = ">=0.25.0" },
     { name = "httpx", specifier = ">=0.27" },
     { name = "httpx", marker = "extra == 'web'", specifier = ">=0.27.0" },


pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
```diff
@@ -243,17 +243,21 @@ def temporal_activities(self) -> list[Callable[..., Any]]:
 
     @contextmanager
     def _temporal_overrides(self) -> Iterator[None]:
+        from pydantic_ai._utils import _prefer_blocking_execution  # pyright: ignore[reportPrivateUsage]
+
         # We reset tools here as the temporalized function toolset is already in self._toolsets.
         with super().override(model=self._model, toolsets=self._toolsets, tools=[]):
-            token = self._temporal_overrides_active.set(True)
+            temporal_active_token = self._temporal_overrides_active.set(True)
+            blocking_token = _prefer_blocking_execution.set(True)
```
DouweM: Let's make this a contextmanager exposed by the utils module, so that we don't deal with the contextvar directly here. See the current_run_context feature in https://github.com/pydantic/pydantic-ai/pull/3537

DouweM: Then we can add that contextmanager to the `with` statement above

pydantic_ai_slim/pydantic_ai/_utils.py
DouweM, line 1: We should update/remove https://ai.pydantic.dev/durable_execution/temporal/#instructions-functions-output-functions-and-history-processors

tests/test_temporal.py
```diff
@@ -199,6 +199,19 @@ async def client_with_logfire(temporal_env: WorkflowEnvironment) -> Client:
 simple_temporal_agent = TemporalAgent(simple_agent, activity_config=BASE_ACTIVITY_CONFIG)
 
 
+def drop_first_message_sync(messages: list[ModelMessage]) -> list[ModelMessage]:
+    """Sync history processor for testing blocking execution mode."""
+    return messages[1:] if len(messages) > 1 else messages
+
+
+agent_with_sync_history_processor = Agent(
```
DouweM: Move this to just ahead of the workflow that uses it, to keep the test file consistently ordered

tests/test_temporal.py
```diff
@@ -199,6 +199,19 @@ async def client_with_logfire(temporal_env: WorkflowEnvironment) -> Client:
 simple_temporal_agent = TemporalAgent(simple_agent, activity_config=BASE_ACTIVITY_CONFIG)
 
 
+def drop_first_message_sync(messages: list[ModelMessage]) -> list[ModelMessage]:
```
DouweM: Let's make this an inline lambda as it's very simple

tests/test_utils.py
```diff
@@ -166,6 +166,31 @@ async def test_run_in_executor_with_contextvars() -> None:
     assert old_result != ctx_var.get()
 
 
+async def test_run_in_executor_with_blocking_execution_enabled() -> None:
+    from pydantic_ai._utils import _prefer_blocking_execution  # pyright: ignore[reportPrivateUsage]
```
DouweM: Same as above; let's not use the private var, but a new public contextmanager

tests/test_temporal.py
dsfaccini: pyright always complains about lambdas, and the length causes a line break, so I'd prefer to keep the extra 1-2 lines with proper types and a nice name

docs/durable_execution/temporal.md
```diff
@@ -162,7 +162,7 @@ Other than that, any agent and toolset will just work!
 
 ### Instructions Functions, Output Functions, and History Processors
```
DouweM: I'd just remove the entire section as they will "just work" now, so they don't really need to be mentioned under `Temporal Integration Considerations`.

The "use async when possible" recommendation is already in https://ai.pydantic.dev/tools-advanced/#parallel-tool-calls-concurrency

pydantic_ai_slim/pydantic_ai/_utils.py
```diff
@@ -41,8 +52,33 @@
 _P = ParamSpec('_P')
 _R = TypeVar('_R')
 
+_prefer_blocking_execution: ContextVar[bool] = ContextVar('_prefer_blocking_execution', default=False)
```
DouweM: "prefer" makes it sound like this expresses a preference, that may or may not be respected. So "enable" or no prefix at all would be better

pydantic_ai_slim/pydantic_ai/_utils.py
```diff
@@ -41,8 +52,33 @@
...
+
+
+@contextmanager
+def blocking_execution() -> Iterator[None]:
```
DouweM: `with _utils.blocking_execution()` makes it seems like all execution will be blocking, but that's not true because of async etc. So maybe we can rename this to something like `with disable_threads()`?

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
```diff
@@ -243,17 +243,22 @@ def temporal_activities(self) -> list[Callable[..., Any]]:
 
     @contextmanager
     def _temporal_overrides(self) -> Iterator[None]:
+        from pydantic_ai._utils import blocking_execution
```
DouweM: To the top please! Imports inside methods are _exclusively_ for circular import issues, and files that depend on optional packages

tests/test_temporal.py
```diff
@@ -1249,6 +1249,48 @@ async def test_temporal_agent_run_sync_in_workflow(allow_model_requests: None, c
...
+
+@workflow.defn
+class AgentWithSyncHistoryProcessorWorkflow:
+    @workflow.run
+    async def run(self, prompt: str) -> str:
+        result = await temporal_agent_with_sync_history_processor.run(prompt)
+        return result.output
+
+
+async def test_temporal_agent_with_sync_history_processor(allow_model_requests: None, client: Client):
```
DouweM, line 1278: Can we test sync instructions as well?

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
dsfaccini: good to know!

PR #3705, opened 2025-12-11
docs: Add Azure, Bedrock and Vertex examples to use Anthropic models

dsfaccini:
https://github.com/user-attachments/assets/826f9bee-7323-40b1-9bcc-81fc1f90459b

### DIFF
--- a/docs/evals.md
+++ b/docs/evals.md
@@ -1,3 +1,7 @@
+---
+title: Pydantic Evals
+---
+
 # Pydantic Evals
 
 **Pydantic Evals** is a powerful evaluation framework for systematically testing and evaluating AI systems, from simple LLM calls to complex multi-agent applications.
--- a/docs/models/anthropic.md
+++ b/docs/models/anthropic.md
@@ -105,7 +105,7 @@ agent = Agent(model)
 
 ### Google Vertex AI
 
-To use Claude models via [Google Cloud Vertex AI](https://cloud.google.com/vertex-ai/generative-ai/docs/partner-models/use-claude), use the [`AsyncAnthropicVertex`](https://docs.anthropic.com/en/api/claude-on-vertex-ai) client from the `anthropic` package:
+To use Claude models via [Google Cloud Vertex AI](https://cloud.google.com/vertex-ai/generative-ai/docs/partner-models/use-claude), follow the [Anthropic documentation](https://docs.anthropic.com/en/api/claude-on-vertex-ai) on how to set up an `AsyncAnthropicVertex` client and then pass it to `AnthropicProvider`:
 
 ```python {test="skip"}
 from anthropic import AsyncAnthropicVertex
@@ -126,7 +126,7 @@ agent = Agent(model)
 
 ### Microsoft Foundry
 
-To use Claude models via [Microsoft Foundry](https://ai.azure.com/), use the [`AsyncAnthropicFoundry`](https://platform.claude.com/docs/en/build-with-claude/claude-in-microsoft-foundry) client from the `anthropic` package:
+To use Claude models via [Microsoft Foundry](https://ai.azure.com/), follow the [Anthropic documentation](https://platform.claude.com/docs/en/build-with-claude/claude-in-microsoft-foundry) on how to set up an `AsyncAnthropicFoundry` client and then pass it to `AnthropicProvider`:
 
 ```python {test="skip"}
 from anthropic import AsyncAnthropicFoundry


docs/models/anthropic.md
```diff
@@ -78,6 +78,106 @@ agent = Agent(model)
...
+    vertex_client = AsyncAnthropicVertex(region='us-east5', project_id='your-project-id')
+    provider = AnthropicProvider(anthropic_client=vertex_client)
+    model = AnthropicModel('claude-sonnet-4-5', provider=provider)
+    agent = Agent(model)
+    ...
+    ```
+
+### Microsoft Azure
+
+Azure offers Claude models through their "Models as a Service" using an OpenAI-compatible API. Use [`OpenAIModel`][pydantic_ai.models.openai.OpenAIModel] with an Azure-configured client:
```
DouweM: Actually we should be able to use `AnthropicModel` with `AsyncAnthropicFoundry` now: https://platform.claude.com/docs/en/build-with-claude/claude-in-microsoft-foundry#api-key-authentication

So let's update these docs to be like the other 3, and then also update this:

https://github.com/pydantic/pydantic-ai/blob/ea2d9d367f1626c8fdbfe4adb83780d6b64b8818/pydantic_ai_slim/pydantic_ai/providers/anthropic.py#L26

docs/models/anthropic.md
```diff
@@ -78,6 +78,106 @@ agent = Agent(model)
...
+### AWS Bedrock
+
+To use Claude models via [AWS Bedrock](https://aws.amazon.com/bedrock/claude/):
+
+=== "With Pydantic AI Gateway"
+
+    ```python {title="<a href='/gateway/' style='float: right;'>Learn about Gateway</a>" test="skip"}
+    from pydantic_ai import Agent
+
+    agent = Agent('gateway/bedrock:us.anthropic.claude-haiku-4-5-20251001-v1:0')
```
DouweM: Unlike the rest of this page, this will not actually use the `AnthropicModel` or the Anthropic-API(-running-on-Bedrock), but rather `BedrockConverseModel` and the AWS Bedrock Converse API (proxied via Gateway of course). Using the Anthropic model will pretty much always give better performance, so I'd prefer to not present this as an equivalent alternative to the `AsyncAnthropicBedrock` approach. (The little note at the end of the section linking to the bedrock docs is sufficient)

docs/models/anthropic.md
```diff
@@ -78,6 +78,106 @@ agent = Agent(model)
...
+### Google Vertex AI
+
+To use Claude models via [Google Cloud Vertex AI](https://cloud.google.com/vertex-ai/generative-ai/docs/partner-models/use-claude):
+
+=== "With Pydantic AI Gateway"
+
+    ```python {title="<a href='/gateway/' style='float: right;'>Learn about Gateway</a>" test="skip"}
+    from pydantic_ai import Agent
+
+    agent = Agent('gateway/google-vertex:claude-sonnet-4-5@20250514')
```
DouweM: Same as up; this uses the Google Vertex API, rather than the Anthropic API hosted on Vertex.

docs/models/anthropic.md
```diff
@@ -78,6 +78,106 @@ agent = Agent(model)
...
+    ```python {test="skip"}
+    from anthropic import AsyncAnthropicBedrock
+
+    from pydantic_ai import Agent
+    from pydantic_ai.models.anthropic import AnthropicModel
+    from pydantic_ai.providers.anthropic import AnthropicProvider
+
+    bedrock_client = AsyncAnthropicBedrock()  # Uses AWS credentials from environment
+    provider = AnthropicProvider(anthropic_client=bedrock_client)
+    model = AnthropicModel('claude-sonnet-4-5', provider=provider)
```
DouweM: This model name is wrong! https://platform.claude.com/docs/en/build-with-claude/claude-on-amazon-bedrock#making-requests

docs/models/anthropic.md
```diff
@@ -78,6 +78,106 @@ agent = Agent(model)
...
+    agent = Agent(model)
+    ...
+    ```
+
+    !!! note "Bedrock vs BedrockConverseModel"
+        This approach uses Anthropic's SDK with AWS Bedrock credentials. For an alternative using AWS SDK (boto3) directly, see [`BedrockConverseModel`](bedrock.md).
```
DouweM: OK here I agree it is is worth mentioning the `BedrockConverseModel`, and then that page will also show the Gateway option.

docs/models/anthropic.md
DouweM: And let's also have a section pointing out `GoogleModel`

docs/models/anthropic.md
dsfaccini: ahh cool, changed it now

docs/models/anthropic.md
```diff
@@ -78,6 +78,75 @@ agent = Agent(model)
 ...
 ```
 
+## Cloud Platform Integrations
+
+You can use Anthropic models through cloud platforms by passing a custom client to [`AnthropicProvider`][pydantic_ai.providers.anthropic.AnthropicProvider].
+
+### AWS Bedrock
+
+To use Claude models via [AWS Bedrock](https://aws.amazon.com/bedrock/claude/), use the [`AsyncAnthropicBedrock`](https://docs.anthropic.com/en/api/claude-on-amazon-bedrock) client from the `anthropic` package:
```
DouweM: It's a little hidden now that the `AsyncAnthropicBedrock` code snippet links to very important docs on how to authenticate etc, so I'd rearrange it like this:

```suggestion
To use Claude models via [AWS Bedrock](https://aws.amazon.com/bedrock/claude/), follow the [Anthropic documentation](https://docs.anthropic.com/en/api/claude-on-amazon-bedrock) on how to set up an `AsyncAnthropicBedrock` client and then pass it to `AnthropicProvider`:
```

(Same goes for the other clouds)

Edit: I see now the secvtion for Microsoft already has a clearer docs link, that format works for me as well!

DouweM (changes requested):

docs/models/anthropic.md
dsfaccini: I'll change azure as well for consistency

docs/evals.md
```diff
@@ -1,3 +1,7 @@
+---
+title: Pydantic Evals
+---
+
```
dsfaccini, line 4: hijacking this PR to add a better social preview of the evals docs

Image

Issue #3654, opened 2025-12-05
Tool arguments of type `dict` are invisible to OpenAI-powered agents


hinnefe2:
### Initial Checks

- [x] I confirm that I'm using the latest version of Pydantic AI
- [x] I confirm that I searched for my issue in https://github.com/pydantic/pydantic-ai/issues before opening this issue

## Observed behavior
When a `pydantic-ai` agent powered by an OpenAI model has a tool with an argument of type `dict` that argument is invisible to the OpenAI model.

## Expected behavior
All tool arguments should be visible to the model powering the agent for all model providers.

## Example
With a basic agent like
```
from pydantic_ai import Agent

openai_agent = Agent("openai:gpt-4o")

@openai_agent.tool_plain
def tool_w_dict(name: str, data: dict[str, str]) -> str:
    """A tool with a dict argument.

    Args:
        name: A name string
        data: A dictionary with arbitrary data
    """
    return f"Got {name} with {data}"

@openai_agent.tool_plain
def tool_w_list(name: str, data: list[str]) -> str:
    """A tool with a list argument.

    Args:
        name: A name string
        data: A list with arbitrary data
    """
    return f"Got {name} with {data}"
```
running a request like
```
openai_result = openai_agent.run_sync("List all tools you have and their parameters with types")
print(openai_result.output)
```
produces results like
```
I have the following tools available:

1. **functions.tool_w_dict**
   - **Parameters**:
     - `name` (string): A name string.

2. **functions.tool_w_list**
   - **Parameters**:
     - `name` (string): A name string.
     - `data` (string[]): A list with arbitrary data.

These tools can handle dictionary-based or list-based inputs, respectively.
```
Note that the `tool_w_dict` parameters do not include the `dict`-typed `data` argument from the function. This behavior is consistent across many attempts and multiple OpenAI models (I tried `gpt-4o`, `gpt-4o-mini`, and `gpt-5`).

The same agent powered by Anthropic's `anthropic:claude-sonnet-4-20250514` model produces output which does include the `data` argument like
```
I have access to 2 tools:

## 1. tool_w_dict
**Description:** A tool with a dict argument.
**Parameters:**
- `name` (string, required): A name string
- `data` (object, required): A dictionary with arbitrary data
  - Additional properties: string values only

## 2. tool_w_list
**Description:** A tool with a list argument.
**Parameters:**
- `name` (string, required): A name string
- `data` (array, required): A list with arbitrary data
  - Items: string type only
```

[...]

PR #3712, opened 2025-12-11
Fix OpenAI never populating `dict[str, ...]` tool args

hinnefe2:
Closes https://github.com/pydantic/pydantic-ai/issues/3654 by adding a warning when an OpenAI-powered agent has a tool with a `dict`-typed argument.

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/profiles/openai.py
+++ b/pydantic_ai_slim/pydantic_ai/profiles/openai.py
@@ -250,8 +250,7 @@ class OpenAIJsonSchemaTransformer(JsonSchemaTransformer):
         if schema_type == 'object':
             # Always ensure 'properties' key exists - OpenAI drops objects without it
             if 'properties' not in schema:
-                _properties: dict[str, Any] = {}
-                schema['properties'] = _properties
+                schema['properties'] = dict[str, Any]()
 
             if self.strict is True:
                 # additional properties are disallowed


pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -1527,6 +1537,16 @@ def _get_builtin_tools(self, model_request_parameters: ModelRequestParameters) -
         return tools
 
     def _map_tool_definition(self, f: ToolDefinition) -> responses.FunctionToolParam:
+        if _has_dict_typed_params(f.parameters_json_schema):
+            warnings.warn(
```
DouweM: Let's move the warning into the helper method so that we don't repeat the text

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -1527,6 +1537,16 @@ def _get_builtin_tools(self, model_request_parameters: ModelRequestParameters) -
         return tools
 
     def _map_tool_definition(self, f: ToolDefinition) -> responses.FunctionToolParam:
```
DouweM, line 1603: I think we'll also need to do the check in `_map_json_schema`

tests/models/test_openai.py
```diff
@@ -3325,3 +3325,53 @@ async def test_openai_reasoning_in_thinking_tags(allow_model_requests: None):
...
+
+def test_has_dict_typed_params_array_of_dicts():
+    """Test detection of list[dict[str, int]] type."""
+    from pydantic_ai.models.openai import _has_dict_typed_params  # pyright: ignore[reportPrivateUsage]
+
+    schema = {
+        'properties': {
+            'dict_list': {'type': 'array', 'items': {'type': 'object', 'additionalProperties': {'type': 'integer'}}}
+        }
+    }
```
DouweM: I'd prefer to test this with a tool function that actually has a `dict` argument, or a `BaseModel` argument that itself has a `dict` field, instead of testing the schemas directly.

So maybe can we build an agent with a tool like that, then run it, and test that a warning was emitted?

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -913,6 +913,16 @@ def _map_json_schema(self, o: OutputObjectDefinition) -> chat.completion_create_
         return response_format_param
 
     def _map_tool_definition(self, f: ToolDefinition) -> chat.ChatCompletionToolParam:
+        if _has_dict_typed_params(f.parameters_json_schema):
+            warnings.warn(
+                f"Tool '{f.name}' has dict-typed parameters that OpenAI's API will silently ignore. "
+                f'Use a Pydantic BaseModel with explicit fields instead of dict types, '
+                f'or switch to a different provider which supports dict types. '
+                f'See: https://github.com/pydantic/pydantic-ai/issues/3654',
```
DouweM:
```suggestion
                f"Tool {f.name!r} has `dict`-typed parameters that OpenAI's API will silently ignore. "
                f'Use a Pydantic `BaseModel`, `dataclass`, or `TypedDict` with explicit fields instead, '
                f'or switch to a different provider which supports `dict` types. '
                f'See: https://github.com/pydantic/pydantic-ai/issues/3654',
```

hinnefe2:
Ok @DouweM I think this is ready for another look.

DouweM:
@hinnefe2 Please have a look at the coverage check!

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -954,6 +954,8 @@ def _map_tool_call(t: ToolCallPart) -> ChatCompletionMessageFunctionToolCallPara
         )
 
     def _map_json_schema(self, o: OutputObjectDefinition) -> chat.completion_create_params.ResponseFormat:
+        _warn_on_dict_typed_params(o.name or DEFAULT_OUTPUT_TOOL_NAME, o.json_schema)
```
DouweM: This class is also used by `OpenRouterModel` and tons of other OpenAI-compatible APIs  (https://ai.pydantic.dev/models/openai/#openai-compatible-models) so I'm thinking we should do this only when `self.system == 'openai'`

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -2830,3 +2836,57 @@ def _map_mcp_call(
...
+    if provider_name != 'openai':
+        return False
+
+    has_dict_params = False
+
+    properties: dict[str, dict[str, Any]] = json_schema.get('properties', {})
+    for prop_schema in properties.values():
+        # Check for object type with non False/absent additionalProperties
+        if (prop_schema.get('type') == 'object') and (prop_schema.get('additionalProperties') not in (False, None)):
+            has_dict_params = True
```
DouweM: Won't this accidentally catch things like `class Foo(BaseModel, extra='allow')`, or a `TypedDict` with `total=False` (or the other properties, I may be mixing them up)? My point is that maybe we should verify that `properties` is empty, assuming that OpenAI only ignores "empty" object schemas with only `additionalProperties`, but not those with at least one known field.

DouweM (changes requested):

hinnefe2:
Apologies for the slow progress here - I got caught up in end-of-year work and then forgot about it over a holiday break. I'll take another pass today.

pydantic_ai_slim/pydantic_ai/models/openai.py
hinnefe2: > assuming that OpenAI only ignores "empty" object schemas with only additionalProperties, but not those with at least one known field.

Ok, after investigating a bit more I think the OpenAI behavior depends on the presence of a `properties` key, not the `additionalProperties`. I repeated the [test](https://community.openai.com/t/tool-arguments-of-type-object-are-silently-dropped-from-tool-definitions-passed-to-the-model/1368533) where I pass a tool schema to an OpenAI request as a) part of a tool definition, and b) part of the regular prompt and ask the model to compare. For the new test I used the following schema which has all four combinations of `additionalProperties` (True or False) and `properties` (present or absent):
```{
  "properties": {
    "name": {
      "description": "A name string",
      "type": "string"
    },
    "data1": {
      "description": "Some arbitrary data",
      "type": "object",
      "additionalProperties": false
    },
    "data2": {
      "description": "Some more arbitrary data",
      "type": "object",
      "additionalProperties": true
    },
    "data3": {
      "description": "Even more arbitrary data",
      "type": "object",
      "properties": {},
      "additionalProperties": false
    },
    "data4": {
      "description": "Even even more arbitrary data",
      "type": "object",
      "properties": {},
      "additionalProperties": true
    },
    "config": {
      "description": "A configuration object with defined properties",
      "type": "object",
      "properties": {
        "timeout": {
          "description": "Timeout in seconds",
          "type": "integer"
        },
        "config_data": {
          "description": "Config data",
          "type": "object"
        },
        "config_data_w_properties": {
          "description": "Config data with properties",
          "type": "object",
          "properties": {}
        }
      },
      "required": [
        "timeout"
      ],
      "additionalProperties": false
    }
  },
  "required": [
    "name",
    "config"
  ],
  "type": "object"
}
```
which produced this response from OpenAI:
```
1. **Parameters in the tool "test_structured_dict":**
   - `name`: A name string
   - `data3?`: Even more arbitrary data (optional)
   - `data4?`: Even even more arbitrary data (optional)
   - `config`: A configuration object with defined properties
     - `timeout`: Timeout in seconds
     - `config_data_w_properties?`: Config data with properties (optional)

2. **Parameters in the provided JSON schema:**
   - `name`: A name string
   - `data1`: Some arbitrary data
   - `data2`: Some more arbitrary data
   - `data3`: Even more arbitrary data
   - `data4`: Even even more arbitrary data
   - `config`: A configuration object with defined properties
     - `timeout`: Timeout in seconds
     - `config_data`: Config data
     - `config_data_w_properties`: Config data with properties
```
from which I conclude that the presence of a `properties` key (even one with an empty dict for a value) is enough to make OpenAI see the parameter.

So based on all that I've updated the logic to look for schema entries which have:
- `"type": "object"`
- no `properties` key.

DouweM: What do you think about instead _fixing_ the behavior by ensuring a `type: object` will always have a `properties` key? That could be done in `OpenAIJsonSchemaTransformer`.

hinnefe2:
Ok looks like fixing this is much simpler than warning about it! I tested the changes with this script
```
from pydantic_ai import Agent

agent = Agent("openai:gpt-4o")


@agent.tool_plain
def process_data(name: str, metadata: dict[str, str]) -> str:
    """Process some data with metadata."""
    return f"Processed {name}"


result = agent.run_sync(
    "What parameters does the process_data tool accept? List them all."
)
print("Agent response:")
print(result.output)
```
Running this with the current changes gave me
```
Agent response:
The `process_data` tool accepts the following parameters:

1. `name` (string): The name associated with the data to be processed.
2. `metadata` (object): The metadata related to the data being processed.
```

while running the same script on `main` gave me:
```
Agent response:
The `process_data` tool accepts a single parameter:

- `name`: A string representing the name of the data to be processed.
```

DouweM:
@hinnefe2 Thanks Henry!

hinnefe2:
For sure, thanks for your patience as this progressed slowly!

Issue #3558, opened 2025-11-25
Support updated AG-UI UserMessage InputContent


davidjon-act:
### Description

Could you please look in to the updated UserMessage structure, defined by AG-UI as defined on their [website](https://docs.ag-ui.com/sdk/python/core/types#usermessage). This is available since ag-ui-protocol 0.1.9.

The AGUIAdapater load_message function currently only accepts a single string for `msg.content`. However since the update it is `Union[str, List["InputContent"]]`. 

It would be great if you could support this similar to the [Vercel-AI setup](https://github.com/pydantic/pydantic-ai/blob/main/pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py) you recently added, allowing us to add attachments to messages.

## Current load_messages

```python
 if isinstance(msg, UserMessage):
      // This currently assumes that the user message content is a string.
      // We should handle the case where the content is of Union[str, List[InputContent]]
      builder.add(UserPromptPart(content=msg.content))
```

## Vercel AI integration

```python
 user_prompt_content: str | list[UserContent] = []
                for part in msg.parts:
                    if isinstance(part, TextUIPart):
                        user_prompt_content.append(part.text)
                    elif isinstance(part, FileUIPart):
                        try:
                            file = BinaryContent.from_data_uri(part.url)
                        except ValueError:
                            media_type_prefix = part.media_type.split('/', 1)[0]
                            match media_type_prefix:
                                case 'image':
                                    file = ImageUrl(url=part.url, media_type=part.media_type)
                                case 'video':
                                    file = VideoUrl(url=part.url, media_type=part.media_type)
                                case 'audio':
                                    file = AudioUrl(url=part.url, media_type=part.media_type)
                                case _:
                                    file = DocumentUrl(url=part.url, media_type=part.media_type)
                        user_prompt_content.append(file)
                    else:  # pragma: no cover
                        raise ValueError(f'Unsupported user message part type: {type(part)}')

                if user_prompt_content:  # pragma: no branch
                    if len(user_prompt_content) == 1 and isinstance(user_prompt_content[0], str):
                        user_prompt_content = user_prompt_content[0]
                    builder.add(UserPromptPart(content=user_prompt_content))
```

### References

https://docs.ag-ui.com/concepts/messages#user-messages
https://docs.ag-ui.com/sdk/python/core/types#usermessage

DouweM:
@MikeRyanDev Any chance you could look into this? 🫶

[...]

PR #3715, opened 2025-12-11
Added support for AG-UI Multi-modal Messages

bojan2501:
Addresses [3558](https://github.com/pydantic/pydantic-ai/issues/3558#issuecomment-3637729528)
Refactored code for better readability.
Bumped version of ag-ui libs.

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
+++ b/pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
@@ -159,7 +159,7 @@ class AGUIAdapter(UIAdapter[RunAgentInput, Message, BaseEvent, AgentDepsT, Outpu
                                         binary_part = BinaryContent(
                                             data=b64decode(part.data), media_type=part.mime_type
                                         )
-                                    else:
+                                    else:  # pragma: no cover
                                         raise ValueError('BinaryInputContent must have either a `url` or `data` field.')
                                     user_prompt_content.append(binary_part)
                                 case _:  # pragma: no cover
--- a/tests/test_ag_ui.py
+++ b/tests/test_ag_ui.py
@@ -72,6 +72,7 @@ with try_import() as imports_successful:
         RunAgentInput,
         StateSnapshotEvent,
         SystemMessage,
+        TextInputContent,
         Tool,
         ToolCall,
         ToolMessage,
@@ -1584,23 +1585,26 @@ async def test_messages(image_content: BinaryContent, document_content: BinaryCo
         ),
         UserMessage(
             id='msg_1',
-            content=[BinaryInputContent(url=image_content.data_uri, mime_type=image_content.media_type)],
+            content=[
+                TextInputContent(text='this is an image:'),
+                BinaryInputContent(url=image_content.data_uri, mime_type=image_content.media_type),
+            ],
         ),
         UserMessage(
             id='msg2',
-            content=[BinaryInputContent(id='img1', url='http://example.com/image.png', mime_type='image/png')],
+            content=[BinaryInputContent(url='http://example.com/image.png', mime_type='image/png')],
         ),
         UserMessage(
             id='msg3',
-            content=[BinaryInputContent(url='http://example.com/video.mp4', mime_type='video/mp4', id='vid1')],
+            content=[BinaryInputContent(url='http://example.com/video.mp4', mime_type='video/mp4')],
         ),
         UserMessage(
             id='msg4',
-            content=[BinaryInputContent(url='http://example.com/audio.mp3', mime_type='audio/mpeg', id='aud1')],
+            content=[BinaryInputContent(url='http://example.com/audio.mp3', mime_type='audio/mpeg')],
         ),
         UserMessage(
             id='msg5',
-            content=[BinaryInputContent(url='http://example.com/doc.pdf', mime_type='application/pdf', id='doc1')],
+            content=[BinaryInputContent(url='http://example.com/doc.pdf', mime_type='application/pdf')],
         ),
         UserMessage(
             id='msg6', content=[BinaryInputContent(data=document_content.base64, mime_type=document_content.media_type)]
@@ -1691,7 +1695,7 @@ async def test_messages(image_content: BinaryContent, document_content: BinaryCo
                         timestamp=IsDatetime(),
                     ),
                     UserPromptPart(
-                        content=[image_content],
+                        content=['this is an image:', image_content],
                         timestamp=IsDatetime(),
                     ),
                     UserPromptPart(


DouweM:
@bojan2501 Can you have a look at the missing test coverage please?

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
```diff
@@ -117,79 +127,141 @@ def state(self) -> dict[str, Any] | None:
 
         return cast('dict[str, Any]', state)
 
+    @classmethod
+    def _load_binary_part(
```
DouweM: Please move private helper methods below the public ones that use them like `load_messages`

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
```diff
@@ -117,79 +127,141 @@ def state(self) -> dict[str, Any] | None:
...
+                    'video': VideoUrl,
+                    'audio': AudioUrl,
+                }
+                media_type_prefix = part.mime_type.split('/', 1)[0]
+                constructor = media_type_constructors.get(media_type_prefix, DocumentUrl)
+                return constructor(
+                    url=part.url,
+                    media_type=part.mime_type,
+                    identifier=part.id,
+                    vendor_metadata={'filename': part.filename},
```
DouweM: We can't do this this, this will be interpreted as `video_metadata` by GoogleModel

bojan2501: You mean we should drop this line: 
```
vendor_metadata={'filename': part.filename}
```

DouweM: @bojan2501 Yep

bojan2501:
I think this is it.
Let me know if I break something.
Also if something needs to be fixed. Will try to find some time to work on it.

bojan2501:
@DouweM To be honest I do not understand why the checks are failing now.

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
bojan2501: Moved them, and as I had issues with testing they are not anymore private.

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
bojan2501: Done.

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
```diff
@@ -124,72 +134,130 @@ def load_messages(cls, messages: Sequence[Message]) -> list[ModelMessage]:
...
+                case ToolMessage() as tool_msg:
+                    cls.add_tool_return_part(builder, tool_msg, tool_calls)
+
+                case ActivityMessage():
+                    raise ValueError(f'Unsupported message type: {type(msg)}')
 
         return builder.messages
+
+    @classmethod
+    def load_binary_part(cls, part: BinaryInputContent) -> BinaryContent | ImageUrl | VideoUrl | AudioUrl | DocumentUrl:
```
DouweM: These methods really shouldn't be part of the public interface, so please make them private, either has class methods, or if that doesn't work, as  private functions in the module, outside of the class

tests/test_ag_ui.py
```diff
@@ -348,6 +377,80 @@ async def test_messages_with_history() -> None:
...
+            id='msg_1',
+            content=[
+                TextInputContent(text='Write a summary of this file:', type='text'),
+                BinaryInputContent(data='VGVzdCBEb2M=', mime_type='text/plain'),
+            ],
+        )
+    )
+    events = await run_and_collect_events(agent, run_input)
+
+    assert events == simple_result()
```
DouweM: Let's drop all of the tests above because they're not verifying that the binary content actually made it through. I'd rather expand the `test_messages` test

tests/test_ag_ui.py
```diff
@@ -1576,6 +1679,221 @@ async def test_messages() -> None:
...
+        UserMessage(
+            id='msg_1',
+            content=[],
+        )
+    ]
+    loaded_messages = AGUIAdapter.load_messages(messages)
+    assert not loaded_messages
+
+
+def test_load_binary_part() -> None:
```
DouweM: I'd rather test this at the `load_messages` level so we can verify these end up in the correct Pydantic AI message history, not just individual parts

tests/test_ag_ui.py
```diff
@@ -1576,6 +1679,221 @@ async def test_messages() -> None:
...
+    assert isinstance(result, BinaryContent)
+    assert result.data == b'Hello world'
+    assert result.media_type == 'text/plain'
+
+    # Test ValueError
+    with pytest.raises(ValueError, match='BinaryInputContent must have either a `url` or `data` field.'):
+        AGUIAdapter.load_binary_part(BinaryInputContent(id='some_id', mime_type='text/plain'))
+
+
+def test_add_assistant_tool_parts() -> None:
```
DouweM: We shouldn't need to test these helper methods directly, as all of these features have already been tested through the existing tests, except for the new binary content. So I'd expect for the _only_ new tests to be related to those binary parts.

If these tests were necessary because of the refactoring of the original massive method into separate ones, I'd prefer to revert that and have a massive method again, which was 100% test-covered.

bojan2501:
@DouweM 
You were right, i made this too complex.

1. So I reverted all changes. 
2. Bumped ag-ui lib to latest version.
3. Updated existing method load_messages to handle new type of messages.
4. Updated test_messages test with new content.
5. Added # noqa: C901 as the method load_messages is complex.

https://github.com/pydantic/pydantic-ai/pull/3715/changes/bab3876a9315ae058e6d3b9e197425afdec38e1c#diff-84321598744d84dbee2318e634c74c9aae39a1c253f1c4bd17ebf9ef2f807b11

I see now that something went wrong, and there is a conflict.
Funny thing is that I used two different editors and two agents to test development.
In the end had to revert and do all by hand.

Should I cancel this pull request and create fresh one? Could be much more easy for review.

DouweM:
@bojan2501 I've fixed this branch PR as follows:

```bash
git co feature/3558-Add-Support-For-Multi-Modal-Messages-AG-UI # your branch
git fetch origin # my origin remote points at the canonical repo
git reset --hard origin/main # reset to match main branch, removing all history on the branch
git cherry-pick bab3876a9315ae058e6d3b9e197425afdec38e1c # apply just the commit with the intended changes
git push -f # force push to overwrite branch history
```

Let's see what CI says!

tseaver:
@bojan2501 What was the rationale for adding an exception to `ActivityMessages` passed by the client as part of the message history?

There is nothing in the AG-UI spec which requires such: you've now required clients to do this filtering, which might or might not be desirable.  For instance, agents implemented as graphs with human-in-the-loop or  client-side tool calling might use the `ActivityMessage` content to reconstruct their place in the graph's flow in a new run.

bojan2501:
Hi @tseaver,
Apologizing for the issue.

My intention was to add support for Multi Modal messages.
And while I was checking the implementation I did not find how to handle ActivityMessages.
Also checks/hooks were complaining that not all cases were covered...
So I added exception as this type of Message is not yet supported.

I could review this again.

tseaver:
@bojan2501 Thanks for the reply.

If the translation to `pyadantic_ai.ModelMessage` isn't clear, then ISTM cleaner just to ignore it:  the message is (likely) the result of events emitted from the *agent* involved (including message IDs passed as part of those events), if not the LLM.

DouweM:
@tseaver Feel free to submit a new PR to remove the error!

PR #3716, opened 2025-12-11
Fix incomplete OTel attributes with `Agent.run_stream_sync`

### LINKED ISSUE #3714: OTel attributes are incomplete when using `Agent.run_stream_sync`
### Initial Checks

### Description

As reported on Slack: https://pydanticlogfire.slack.com/archives/C083V7PMHHA/p1765485304294729

`run_stream_sync`:

Image

`run_stream`:

Image

### Example Code

```Python
import asyncio

import logfire
from pydantic_ai import Agent

logfire.configure()
logfire.instrument_pydantic_ai()

agent = Agent("openai:gpt-5.2")

response = agent.run_stream_sync("What is the capital of the UK?")
print(response.get_output())


async def main():
    async with agent.run_stream("What is the capital of the UK?") as response:
        print(await response.get_output())


asyncio.run(main())
```

```Text
main
```

DouweM:
Fixes https://github.com/pydantic/pydantic-ai/issues/3714

To do:
- [ ] Add test


DouweM:
As can be seen in the test snapshot, the run is failing with a `RuntimeError: Attempted to exit cancel scope in a different task than it was entered in`. 

I believe that's related to the fact that the agent graph uses a different task for each node in the graph (handle initial user prompt, send model request, handle response / call tools, send model request, etc), combined with the pretty hacky ways we try to turn async iterables into sync ones, resulting in before/after steps that are supposed to take place in the same task, running in different tasks instead.

With the current API, where `run_stream_sync` directly returns a `StreamedRunResultSync` object and its methods directly return iterators, it may not be possible to get OTel attributes to work properly. If we make `run_stream_sync` a contextmanager instead (like `run_sync`), we may be able to fix this more cleanly. For backward compatibility, the non-contextmanager approach could continue to _work_, but we could let make the typing (and docs) only document/allow the contextmanager approach.

DouweM:
Closing in favor of #6199

PR #3718, opened 2025-12-12
Set `ToolRetryError` message

jamesaud:
## Summary

`ToolRetryError` was not passing a message to `Exception.__init__()`, causing `str(error)` to return an empty string. 

This came up when the agent loop crashes due to max tool retries, which breaks error monitoring tools like Sentry or Langsmith that rely on `str(exception)` to display and group errors.

## Problem

```python
from pydantic_ai.exceptions import ToolRetryError
from pydantic_ai.messages import RetryPromptPart

part = RetryPromptPart(content='Test error', tool_name='test')
error = ToolRetryError(part)

str(error)  # Returns '' (empty string)
error.args  # Returns ()
```

## Solution

Pass `tool_retry.model_response()` to the parent `Exception.__init__()`:

```python
str(error)  # Now returns 'Test error\n\nFix the errors and try again.'
```

## Changes

- `pydantic_ai_slim/pydantic_ai/exceptions.py`: Pass message to parent class
- `tests/test_exceptions.py`: Add `ToolRetryError` to hashable test + add dedicated string representation test


pydantic_ai_slim/pydantic_ai/exceptions.py
```diff
@@ -188,7 +188,7 @@ class ToolRetryError(Exception):
 
     def __init__(self, tool_retry: RetryPromptPart):
         self.tool_retry = tool_retry
-        super().__init__()
+        super().__init__(tool_retry.model_response())
```
DouweM: Can we use `tool_retry.content` that contains just the error message, without the surrounding text instructing the model to retry?

Besides a `str`, `content` can also be a list of Pydantic error details, so we would need to stringify/dump those as to JSON.

jamesaud: Great, thanks for the info! Using `tool_retry.content` now. Added a bit of formatting code to make `list[ErrorDetail]` more human readable.

jamesaud: Plz wait for re-review - need to fix CI.

tests/test_exceptions.py
```diff
@@ -59,3 +64,26 @@ def test_exceptions_hashable(exc_factory: Callable[[], Any]):
...
+        {'type': 'missing', 'loc': ('field1',), 'msg': 'Field required', 'input': {}},
+        {'type': 'string_type', 'loc': ('field2',), 'msg': 'Input should be a valid string', 'input': 123},
+    ]
+    part = RetryPromptPart(content=error_details, tool_name='my_tool')
+    error = ToolRetryError(part)
+
+    expected = """\
+Tool 'my_tool' failed: 2 validation errors
+  field1: Field required
+  field2: Input should be a valid string"""
```
DouweM: This is a bit lossy because we don't get to see the original input or error type. Can you see if we can use the same rendering that is used when Pydantic `ValidationError` is raised and printed in terminal? We get the `ErrorDetails` from a `ValidationError` here, so maybe we can pass in the string message here where we still have that error:

https://github.com/pydantic/pydantic-ai/blob/d2b08ad12e103e81bd0a9dc0eb7f2a7c3472931d/pydantic_ai_slim/pydantic_ai/_tool_manager.py#L184-L190

jamesaud: Thank you for the suggestion! I simplified the scope of the PR to explicitly pass `str(e)` from the call sites. 
If callers do not provide `message`, then it will be blank.

Another option would be to fallback to lossy behavior as a backup if `message` is not passed.

pydantic_ai_slim/pydantic_ai/_tool_manager.py
```diff
@@ -187,14 +187,14 @@ async def _call_tool(
...
-                        e = ToolRetryError(m)
+                        e = ToolRetryError(m, message=str(e))
                     elif isinstance(e, ModelRetry):
                         m = _messages.RetryPromptPart(
                             tool_name=name,
                             content=e.message,
                             tool_call_id=call.tool_call_id,
                         )
-                        e = ToolRetryError(m)
+                        e = ToolRetryError(m, message=e.message)
```
DouweM: I wouldn't mind setting `message` automatically if `isinstance(tool_retry.content, str)`

tests/test_exceptions.py
```diff
@@ -59,3 +63,10 @@ def test_exceptions_hashable(exc_factory: Callable[[], Any]):
 
     assert exc in s
     assert d[exc] == 'value'
+
+
+def test_tool_retry_error_str():
```
DouweM: We should test that it is set correctly in the `ValidationError` case

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -637,7 +637,7 @@ async def process(
                 m = _messages.RetryPromptPart(
                     content=e.errors(include_url=False),
                 )
-                raise ToolRetryError(m) from e
+                raise ToolRetryError(m, message=str(e)) from e
```
DouweM: If `ToolRetryError.__init__` is able to access the `__cause__` that's set by `from e` (I'm not sure if it can), we could do this automatically

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/_output.py
jamesaud: Checked locally. Unfortunately, `__cause__` is `None` in `__init__`. It gets set after the exception is constructed, by the `raise ... from e` syntax

pydantic_ai_slim/pydantic_ai/_tool_manager.py
jamesaud: Do we want to support an automatic behavior for `list[ErrorDetails]`? There's a few options:
1. call `str()` on `list[ErrorDetails]` 
2. call `json.dumps` on `list[ErrorDetails]`
3. format `list[ErrorDetails]` to be more human readable (more code)
4. do nothing

Which do you prefer?

tests/test_exceptions.py
jamesaud: Sounds good!

pydantic_ai_slim/pydantic_ai/exceptions.py
jamesaud: closing comment as relevant context has been shifted to others.

pydantic_ai_slim/pydantic_ai/_tool_manager.py
DouweM: Hmm, since my idea below of getting it from `__cause__` won't work, I prefer option 2, because I don't want us to have to remember to pass in something special when we create a `ToolRetryError` for a `ValidationError`; so I'd rather have everything live inside that `ToolRetryError.__init__`. I do wonder if there's a way to get access to the logic `ValidationError` uses to produce its human-readable error output based on `ErrorDetails`.

tests/test_exceptions.py
jamesaud: Updated test

pydantic_ai_slim/pydantic_ai/_tool_manager.py
jamesaud: I took an alternative approach, which I think is pragmatic. 

1. If `tool_retry.content` is `list[ErrorDetails]`, we _know_ it's a `ValidationError` and can actually reconstruct the nice human-readable output with `ValidationError.from_exception_data`.

2. If `tool_retry.content` is a `str`, it means it was some other exception and we just set this as the message

And removed the optional `message` parameter from `ToolRetryError` entirely, so that it is implicitly handled now.

tests/test_exceptions.py
jamesaud: Closing as conversation is addressed in the comment below

pydantic_ai_slim/pydantic_ai/_tool_manager.py
jamesaud: Unfortunately I spoke too soon, some of the other tests fail because certain Pydantic error types (`value_error`, `assertion_error`, `json_invalid`, etc.) require an `error` key in their `ctx` dictionary.  When `ErrorDetails` are serialized/stored in `RetryPromptPart`, exception objects get stripped from `ctx` (since they're not JSON-serializable).

Resorting to parsing manually.

pydantic_ai_slim/pydantic_ai/exceptions.py
```diff
@@ -188,7 +190,31 @@ class ToolRetryError(Exception):
...
+
+        We format manually rather than using ValidationError.from_exception_data because
+        some error types (value_error, assertion_error, etc.) require an 'error' key in ctx,
+        but when ErrorDetails are serialized, exception objects are stripped from ctx.
+        The 'msg' field already contains the human-readable message, so we use that directly.
+        """
+        title = tool_name or 'ToolRetryError'
+        error_count = len(errors)
+        error_word = 'error' if error_count == 1 else 'errors'
+        lines = [f'{error_count} validation {error_word} for {title}']
```
DouweM: - For consistency with other errors, I want the tool name to be wrapped in quotes, which we can do with `{tool_name!r}`
- If there's no tool name, I don't think it's worth including `ToolRetryError` as that's already the exception name

DouweM (commented):
@jamesaud Thanks for working on this; one last change about the copy and I'll merge this!

pydantic_ai_slim/pydantic_ai/exceptions.py
jamesaud: Done, and consolidated the ternary logic into the string via f strings, to be consistent with the code style of ValidationError.

PR #3720, opened 2025-12-12
Add support for `ImageGenerationTool.size` to Gemini image models

h0rv:
This adds `resolution` configuration to the `ImageGenerationTool` to control the image size for Gemini 3 Pro Image: https://ai.google.dev/gemini-api/docs/image-generation.

### DIFF
--- a/tests/models/test_openai.py
+++ b/tests/models/test_openai.py
@@ -104,7 +104,38 @@ def test_init():
     assert m.model_name == 'gpt-4o'
 
 
-def test_openai_image_generation_tool_aspect_ratio_invalid_covered() -> None:
+@pytest.mark.parametrize(
+    'aspect_ratio,size,expected',
+    [
+        # aspect_ratio is None, various sizes
+        (None, None, 'auto'),
+        (None, 'auto', 'auto'),
+        (None, '1024x1024', '1024x1024'),
+        (None, '1024x1536', '1024x1536'),
+        (None, '1536x1024', '1536x1024'),
+        # Valid aspect_ratios with no size
+        ('1:1', None, '1024x1024'),
+        ('2:3', None, '1024x1536'),
+        ('3:2', None, '1536x1024'),
+        # Valid aspect_ratios with compatible sizes
+        ('1:1', 'auto', '1024x1024'),
+        ('1:1', '1024x1024', '1024x1024'),
+        ('2:3', '1024x1536', '1024x1536'),
+        ('3:2', '1536x1024', '1536x1024'),
+    ],
+)
+def test_openai_image_generation_size_valid_combinations(
+    aspect_ratio: Literal['1:1', '2:3', '3:2'] | None,
+    size: Literal['auto', '1024x1024', '1024x1536', '1536x1024'] | None,
+    expected: Literal['auto', '1024x1024', '1024x1536', '1536x1024'],
+) -> None:
+    """Test valid combinations of aspect_ratio and size for OpenAI image generation."""
+    tool = ImageGenerationTool(aspect_ratio=aspect_ratio, size=size)
+    assert _resolve_openai_image_generation_size(tool) == expected
+
+
+def test_openai_image_generation_tool_aspect_ratio_invalid() -> None:
+    """Test that invalid aspect_ratio raises UserError."""
     tool = ImageGenerationTool(aspect_ratio='16:9')
     with pytest.raises(UserError, match='OpenAI image generation only supports `aspect_ratio` values'):
         _resolve_openai_image_generation_size(tool)


h0rv:
We can add this under existing `quality` config, but they are semantically different for the Gemini 3 Pro image model, so I opted to making it separate.

pydantic_ai_slim/pydantic_ai/models/google.py
```diff
@@ -362,8 +362,14 @@ def _get_tools(
...
-                    if tool.aspect_ratio:
-                        image_config = ImageConfigDict(aspect_ratio=tool.aspect_ratio)
+                    # Build image_config from tool parameters, only including non-None values
+                    image_config_kwargs: dict[str, Any] = {}
+                    if tool.aspect_ratio is not None:
+                        image_config_kwargs['aspect_ratio'] = tool.aspect_ratio
+                    if tool.resolution is not None:
+                        image_config_kwargs['image_size'] = tool.resolution
+                    if image_config_kwargs:
+                        image_config = ImageConfigDict(**image_config_kwargs)
```
DouweM: Since `image_config` is already typed as `ImageConfigDict`, which is a `TypedDict`, we can drop `image_config_kwargs`, set `image_config` to `{}`, and then set dict fields on it as usual.

pydantic_ai_slim/pydantic_ai/builtin_tools.py
```diff
@@ -328,6 +328,14 @@ class ImageGenerationTool(AbstractBuiltinTool):
     * OpenAI Responses (maps '1:1', '2:3', and '3:2' to supported sizes)
     """
 
+    resolution: str | None = None
```
DouweM: The API field is called `image_size`, so it feels like we should use our `size` field for this. What do you think about adding the 3 options there, and changing the docstring to make it clear which values are supported by OpenAI vs Google?

docs/builtin-tools.md
```diff
@@ -353,6 +353,23 @@ assert isinstance(result.output, BinaryImage)
 
 _(This example is complete, it can be run "as is")_
 
+To control the image resolution with Gemini 3 Pro Image models, use the `resolution` parameter:
```
DouweM: This make it seem like it's specific to _only_ Gemini 3 Pro, while presumably future models will support it as well. So let's rephrase this to make it clear it's supported by models _starting with_ Gemini 3 Pro Image

docs/builtin-tools.md
```diff
@@ -366,8 +383,8 @@ For more details, check the [API documentation][pydantic_ai.builtin_tools.ImageG
 | `output_format` | ✅ | ❌ |
 | `partial_images` | ✅ | ❌ |
 | `quality` | ✅ | ❌ |
-| `size` | ✅ | ❌ |
-| `aspect_ratio` | ✅ (1:1, 2:3, 3:2) | ✅ |
+| `size` | ✅ (1024x1024, 1024x1536, 1536x1024, auto) | ✅ (1K, 2K, 4K) |
```
DouweM: Let's not mention `auto` here as it's not really a value the user would want to pick, but more the lack of a value, i.e. "let the model decide"

pydantic_ai_slim/pydantic_ai/builtin_tools.py
```diff
@@ -311,12 +311,13 @@ class ImageGenerationTool(AbstractBuiltinTool):
...
 
-    size: Literal['1024x1024', '1024x1536', '1536x1024', 'auto'] = 'auto'
+    size: Literal['1024x1024', '1024x1536', '1536x1024', '1K', '2K', '4K', 'auto'] = 'auto'
     """The size of the generated image.
 
     Supported by:
 
-    * OpenAI Responses
+    * OpenAI Responses: '1024x1024', '1024x1536', '1536x1024', 'auto'
+    * Google (Gemini 3 Pro Image and later): '1K', '2K', '4K'
```
DouweM: We effectively support `auto` on Google as well by not sending any value to the API, right? So I'd drop `auto` from the OpenAI Responses list above as it's not really OpenAI specific

pydantic_ai_slim/pydantic_ai/models/google.py
```diff
@@ -362,8 +362,12 @@ def _get_tools(
...
                             "`ImageGenerationTool` is not supported by this model. Use a model with 'image' in the name instead."
                         )
-                    if tool.aspect_ratio:
-                        image_config = ImageConfigDict(aspect_ratio=tool.aspect_ratio)
+
+                    image_config = ImageConfigDict()
+                    if tool.aspect_ratio is not None:
+                        image_config['aspect_ratio'] = tool.aspect_ratio
+                    if tool.size in ('1K', '2K', '4K'):
+                        image_config['image_size'] = tool.size
```
DouweM, line 383: Can we have an error for unsupported values like we have for OPenAI?

```
raise UserError(
            f'OpenAI image generation only supports `aspect_ratio` values: {supported}. '
            'Specify one of those values or omit `aspect_ratio`.'
        )
```

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -158,7 +158,11 @@ def _resolve_openai_image_generation_size(
     """Map `ImageGenerationTool.aspect_ratio` to an OpenAI size string when provided."""
     aspect_ratio = tool.aspect_ratio
     if aspect_ratio is None:
-        return tool.size
+        # Only return size if it's an OpenAI-supported value
+        if tool.size in ('auto', '1024x1024', '1024x1536', '1536x1024'):
+            return tool.size
```
DouweM: We should raise an error for unsupported values like we do below for `aspect_ratio`

pydantic_ai_slim/pydantic_ai/builtin_tools.py
h0rv: For Google - it defaults, rather than the model deciding:

```
  image_size: Optional[str]
  """Optional. Specifies the size of generated images. Supported
      values are `1K`, `2K`, `4K`. If not specified, the model will use default
      value `1K`."""
```

h0rv: Should we do something like the following for the doc str and docs?

```
| `size` | ✅ (auto (default), 1024x1024, 1024x1536, 1536x1024, auto) | ✅ (1K (default), 2K, 4K) |
```

h0rv: we can add a note for auto that the model decides

h0rv: or `auto` is just omitted for google, which would fallback to 1k, but that may be confusing.

DouweM: @h0rv Hmm, then the fact that we say `= 'auto'` as if it's the default for all models, is confusing. In retrospect it's a shame that OpenAI effectively used this for `aspect_ratio`, not really different `size`s, but then again if they add sizes in the future I guess they'll be in this argument as well.

Maybe don't have `auto` as the explicit default anymore, document for each of the providers what the default is, and make it clear that `auto` is indeed not supported by Google.

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -151,13 +151,28 @@
     '3:2': '1536x1024',
 }
 
+_OPENAI_IMAGE_SIZE_VALUES: tuple[Literal['auto', '1024x1024', '1024x1536', '1536x1024'], ...] = (
```
DouweM: We can make `Literal['auto', '1024x1024', '1024x1536', '1536x1024']` a type, and then use it in the return signature for `_resolve_openai_image_generation_size` _and_ at runtime using `_utils.get_union_args` (or if that doesn't work `typing.get_args`)

pydantic_ai_slim/pydantic_ai/models/google.py
```diff
@@ -137,6 +137,8 @@
     GoogleFinishReason.NO_IMAGE: 'error',
 }
 
+_GOOGLE_IMAGE_SIZE_VALUES: tuple[Literal['1K', '2K', '4K'], ...] = ('1K', '2K', '4K')
```
DouweM: If Google exposes these as a type, we can use the type directly to get the tuple of the strings; see the comment on OpenAI below

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/models/google.py
h0rv: Unfortunately its loosely typed as a str :/

```py
  image_size: Optional[str]
  """Optional. Specifies the size of generated images. Supported
      values are `1K`, `2K`, `4K`. If not specified, the model will use default
      value `1K`."""
```

PR #3725, opened 2025-12-14
Update docs about choosing priority of output and deferred tool calls with run_stream

### LINKED ISSUE #3636: Missing Docs: Inconsistency in `result.output` for `agent.run_sync` and `agent.run_stream` when both `OutputTool` and `DeferredTool` are called
### Initial Checks

### Description

I was working on mirrored tests in `test_agent.py` and `test_streaming.py` for #3523 and discovered there's a difference in sync and streaming version for `test_early_strategy_with_external_tool_call` test.

Streaming version: [link](https://github.com/pydantic/pydantic-ai/blob/48e3707348af73ab3863615848cbdf9df304d055/tests/test_streaming.py#L1087);
Sync version: [link](https://github.com/Danipulok/pydantic-ai/blob/c61a0de2034bee1fe7a17bc84c570a2afb83a47c/tests/test_agent.py#L3850);

Both tests have the same input, but the behavior is different (both tests pass).
In streaming, `defer` tool is handled as final output in [this](https://github.com/pydantic/pydantic-ai/blob/48e3707348af73ab3863615848cbdf9df304d055/pydantic_ai_slim/pydantic_ai/models/__init__.py#L1227-L1228) part.
But in sync, the output tool becomes the final output in `_agent_graph`.

### Example Code
https://gist.github.com/Danipulok/04ccb71269064e0bae8963f6e99884ff

```Text
Python 3.13.4
pydantic==2.12.4
pydantic-ai==1.25.1
```

Danipulok:
Fixes #3636

### DIFF
--- a/docs/agents.md
+++ b/docs/agents.md
@@ -124,7 +124,7 @@ It also takes an optional `event_stream_handler` argument that you can use to ga
 
 The example below shows how to stream events and text output. You can also [stream structured output](output.md#streaming-structured-output).
 
-!!! note "Streaming Methods Choose Final Result Differently"
+!!! note
     The `run_stream()` and `run_stream_sync()` methods will consider the first output that matches the [output type](output.md#structured-output) (which could be text, an [output tool](output.md#tool-output) call, or a [deferred](deferred-tools.md) tool call) to be the final output of the agent run, even when the model generates (additional) tool calls after this "final" output.
 
 	These "dangling" tool calls will not be executed unless the agent's [`end_strategy`][pydantic_ai.agent.Agent.end_strategy] is set to `'exhaustive'`, and even then their results will not be sent back to the model as the agent run will already be considered completed.
--- a/docs/output.md
+++ b/docs/output.md
@@ -319,7 +319,7 @@ When the model calls other tools in parallel with an output tool, you can contro
 
 The `'exhaustive'` strategy is useful when tools have important side effects (like logging, sending notifications, or updating metrics) that should always execute.
 
-!!! warning "Difference in choosing Final Result in Streaming Methods"
+!!! warning "Priority of output and deferred tools in streaming methods"
     The [`run_stream()`][pydantic_ai.agent.AbstractAgent.run_stream] and [`run_stream_sync()`][pydantic_ai.agent.AbstractAgent.run_stream_sync] methods will consider the first output that matches the [output type](output.md#structured-output) (which could be text, an [output tool](output.md#tool-output) call, or a [deferred](deferred-tools.md) tool call) to be the final output of the agent run, even when the model generates (additional) tool calls after this "final" output.
 
     This means that if the model calls deferred tools before output tools when using these methods, the deferred tool calls determine the agent run's final output, while the other [run methods](agents.md#running-agents) would have prioritized the tool output.
--- a/tests/test_agent.py
+++ b/tests/test_agent.py
@@ -3356,9 +3356,9 @@ class TestMultipleToolCalls:
     def test_early_strategy_with_external_tool_call(self):
         """Test that early strategy handles external tool calls correctly.
 
-        Streaming and sync modes differ in how they choose the final result:
+        Streaming and non-streaming modes differ in how they choose the final result:
         - Streaming: First tool call (in response order) that can produce a final result (output or deferred)
-        - Sync: First output tool (if none called, all deferred tools become final result)
+        - Non-streaming: First output tool (if none called, all deferred tools become final result)
 
         See https://github.com/pydantic/pydantic-ai/issues/3636#issuecomment-3618800480 for details.
         """
--- a/tests/test_streaming.py
+++ b/tests/test_streaming.py
@@ -1137,9 +1137,9 @@ class TestMultipleToolCalls:
     async def test_early_strategy_with_external_tool_call(self):
         """Test that early strategy handles external tool calls correctly.
 
-        Streaming and sync modes differ in how they choose the final result:
+        Streaming and non-streaming modes differ in how they choose the final result:
         - Streaming: First tool call (in response order) that can produce a final result (output or deferred)
-        - Sync: First output tool (if none called, all deferred tools become final result)
+        - Non-streaming: First output tool (if none called, all deferred tools become final result)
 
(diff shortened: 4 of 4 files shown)


docs/output.md
```diff
@@ -315,6 +315,12 @@ When the model calls other tools in parallel with an output tool, you can contro
 
 The `'exhaustive'` strategy is useful when tools have important side effects (like logging, sending notifications, or updating metrics) that should always execute.
 
+!!! warning "Streaming vs Sync Behavior Difference"
+    `run_stream()` behaves differently from `run()` and `run_sync()` when choosing the final result:
+
+    - **`run_stream()`**: The first called tool that **can** produce a final result (output or deferred) becomes the final result
+    - **`run()` / `run_sync()`**: The first **output** tool becomes the final result. If none are called, all **deferred** tools become the final result as `DeferredToolRequests`
```
Danipulok: > If none are called

The wording is like this, since we get an `UnexpectedModelBehavior` if both output and deferred tools are called, but none output tools are validated

Danipulok: If I understand correctly, the deferred tools args are not validated

I guess it means we could check that if any deferred tools calls are present and `tool_call_results` is `None`, we can skip ignore invalid output tool calls, as we do in here:

https://github.com/pydantic/pydantic-ai/blob/ccefddceb68eedab04619e7ab1fd12117e0c1153/pydantic_ai_slim/pydantic_ai/_agent_graph.py#L913-L915

But I'm not sure about it...
And IMHO, if do this at all, then it should be done in a separate PR

docs/output.md
```diff
@@ -315,6 +315,12 @@ When the model calls other tools in parallel with an output tool, you can contro
 
 The `'exhaustive'` strategy is useful when tools have important side effects (like logging, sending notifications, or updating metrics) that should always execute.
 
+!!! warning "Streaming vs Sync Behavior Difference"
+    `run_stream()` behaves differently from `run()` and `run_sync()` when choosing the final result:
```
DouweM: - Please link to the API docs for `run_stream`
- It's not just `run` and `run_sync`, also `iter`, `run_stream_events`, etc. So maybe link to agents.md#running-agents and just say "the other run methods"

docs/output.md
```diff
@@ -315,6 +315,12 @@ When the model calls other tools in parallel with an output tool, you can contro
...
+
+    - **`run_stream()`**: The first called tool that **can** produce a final result (output or deferred) becomes the final result
```
DouweM: Let's link to the deferred tool docs

docs/output.md
DouweM: > If I understand correctly, the deferred tools args are not validated

That's true for external tools, but not for function tools that raise `CallDeferred`, which do have args validated before the function is called

> I guess it means we could check that if any deferred tools calls are present and tool_call_results is None, we can skip ignore invalid output tool calls, as we do in here:

I don't quite understand what you mean

docs/output.md
```diff
@@ -315,6 +315,12 @@ When the model calls other tools in parallel with an output tool, you can contro
 
 The `'exhaustive'` strategy is useful when tools have important side effects (like logging, sending notifications, or updating metrics) that should always execute.
 
+!!! warning "Streaming vs Sync Behavior Difference"
```
DouweM: It's not exactly streaming vs sync; it's just `run_stream` that's odd. So maybe it makes sense to mention this under https://ai.pydantic.dev/agents/#streaming-events-and-final-output, either as well, or instead. I say "instead" because we already give context there on `run_stream` always preferring the first final result it sees, which is in line with what we're explaining here. So maybe that place should have the primary explanation, and then here we can link to there

DouweM (changes requested):

docs/output.md
Danipulok: > I don't quite understand what you mean

I mean we could try to guess if we have any _valid_ differed tools calls in the same model response. And if we have any, do not raise `ModelUnexpectedBehavior`, but ignore invalid output tool calls

DouweM: @Danipulok I think the behavior of all-invalid output tool calls + deferred tool calls (valid or otherwise) should be identical to the behavior of all-invalid output tool calls + function tool calls. That means that the `end_strategy` determines whether the tools are executed, but either way if the model called _any_ output tools (valid or otherwise) we interpret that to mean that model is ready to end the agent run, which means that if the output tools were valid, we return it directly, and if not, we tell the model to try again, but we don't send it the results of the other tool calls. 

With deferred tool calls, the consistent behavior would be to never return `DeferredToolRequests` if there were any output tool calls, because that would imply that the run will be resumed later with the deferred tool results. So if none of the output tools are valid, and the model is told to retry and keeps failing, I think the output validation `UnexpectedModelBehavior` exception is correct.

Danipulok: Got it, thank you!
So if both output tools and deferred tool calls were passed in the same response, and all output tools have failed, the `UnexpectedModelBehavior exception` is correct.
Then current behavior is okay

docs/agents.md
```diff
@@ -124,13 +124,13 @@ It also takes an optional `event_stream_handler` argument that you can use to ga
...
+!!! note "Streaming Methods Choose Final Result Differently"
+    [`run_stream()`][pydantic_ai.agent.AbstractAgent.run_stream] and [`run_stream_sync()`][pydantic_ai.agent.AbstractAgent.run_stream_sync] will consider the first tool call that can produce a final result (both [output](output.md#tool-output) and [deferred](deferred-tools.md) tools) to be the final output. This is different from all [other](#running-agents) run methods, which prioritize **output** tools over deferred tools.
 
-	These "dangling" tool calls will not be executed unless the agent's [`end_strategy`][pydantic_ai.agent.Agent.end_strategy] is set to `'exhaustive'`, and even then their results will not be sent back to the model as the agent run will already be considered completed.
+    "Dangling" tool calls generated after the "final output" will not be executed unless the agent's [`end_strategy`][pydantic_ai.agent.Agent.end_strategy] is set to `'exhaustive'`, and even then their results will not be sent back to the model as the agent run will already be considered completed.
```
Danipulok: Not really sure what to do here. It just seems that `end_strategy` should be mentioned in this file, so I kept it. But at the same time it's not a difference between sync and async, so maybe it shouldn't be in this `note` section...

@DouweM what would you recommend?

Issue #3393, opened 2025-11-11
Raising `ModelRetry` in output validator function crashes when streaming


petersli:
### Initial Checks

- [x] I confirm that I'm using the latest version of Pydantic AI
- [x] I confirm that I searched for my issue in https://github.com/pydantic/pydantic-ai/issues before opening this issue

### Description

When using `stream_output`, if I raise `ModelRetry` in my output validator function, instead of seeing another model request (like I do when I raise `ModelRetry` in a function tool), the exception is unhandled.

Based on the [docs](https://ai.pydantic.dev/output/#output-validator-functions) and [docstrings](https://github.com/pydantic/pydantic-ai/blob/aa1fe331e7fc9ddc54a6ccbe4e1b2cad5831724e/pydantic_ai_slim/pydantic_ai/messages.py#L855), I would not expect this to crash.

It works properly with `run_sync`, which is the case that's tested in `tests/test_agent.py` and the docs.

### Example Code

```Python
# add this test to tests/test_agent.py
async def test_output_validator_stream_output():
    """Test that ModelRetry in output validators works correctly with streaming."""

    async def stream_model(messages: list[ModelMessage], info: AgentInfo) -> AsyncIterator[DeltaToolCalls]:
        assert info.output_tools is not None
        if len(messages) == 1:
            # First attempt: return wrong value (a=41)
            yield {0: DeltaToolCall(name=info.output_tools[0].name, json_args='{"a": 41')}
            yield {0: DeltaToolCall(json_args=', "b": "f')}
            yield {0: DeltaToolCall(json_args='oo"}')}
        else:
            # Retry: return correct value (a=42)
            yield {0: DeltaToolCall(name=info.output_tools[0].name, json_args='{"a": 42')}
            yield {0: DeltaToolCall(json_args=', "b": "f')}
            yield {0: DeltaToolCall(json_args='oo"}')}

    agent = Agent(FunctionModel(stream_function=stream_model), output_type=Foo)

    @agent.output_validator
    def validate_output(ctx: RunContext[None], output: Foo) -> Foo:
        assert ctx.tool_name == 'final_result'
        # Only validate on final output, not partial streaming outputs
        if not ctx.partial_output:
            if output.a == 42:
                return output
            else:
                raise ModelRetry('"a" should be 42')
        return output

    async with agent.run_stream('Hello') as result:
        outputs = [output async for output in result.stream_output(debounce_by=None)]

    # Verify final output is correct after retry
[...]

PR #3726, opened 2025-12-14
Add docs and tests for `RunContext.partial_output` in output tools

### LINKED ISSUE #3624: Output tools are called twice or more times in streaming mode
### Initial Checks

### Summary
Output tools are called zero, twice or more times in streaming mode (but only once in sync mode)
(I'm not sure about the zero case for sync mode)

### What happens
The first call happens at [_agent_graph.py:861](https://github.com/pydantic/pydantic-ai/blob/88f6e92ce863995eeceeed5becf69434c5eb5fa6/pydantic_ai_slim/pydantic_ai/_agent_graph.py#L861) during tool call processing in the graph execution, and the second call happens at [result.py:179](https://github.com/pydantic/pydantic-ai/blob/88f6e92ce863995eeceeed5becf69434c5eb5fa6/pydantic_ai_slim/pydantic_ai/result.py#L179) when validating the response in `get_output()`.

### Details
Output tool processors are invoked:
- **2 times** when using `get_output()` alone
- **2 times** when using `stream_output()` alone
- **4 times** when using both `stream_output()` + `get_output()`
- **0 times** when the stream isn't consumed at all (IMHO think the expected behavior is to get called anyway)

https://gist.github.com/Danipulok/f27bedddfc2b7fc476b1836bbf3af14d

```Text
Python 3.13.4
pydantic==2.12.4
pydantic-ai==1.25.1
```

Danipulok:
Closes #3624

### DIFF
--- a/docs/output.md
+++ b/docs/output.md
@@ -266,7 +266,7 @@ _(This example is complete, it can be run "as is")_
 
 When streaming with `run_stream()` or `run_stream_sync()`, output functions are called **multiple times** — once for each partial output received from the model, and once for the final complete output.
 
-You should check the [`RunContext.partial_output`][pydantic_ai.tools.RunContext.partial_output] flag when your output function has **side effects** (e.g., sending notifications, logging, database updates) that should only execute on the final output
+You should check the [`RunContext.partial_output`][pydantic_ai.tools.RunContext.partial_output] flag when your output function has **side effects** (e.g., sending notifications, logging, database updates) that should only execute on the final output.
 
 When streaming, `partial_output` is `True` for each partial output and `False` for the final complete output.
 For all [other run methods](agents.md#running-agents), `partial_output` is always `False` as the function is only called once with the complete output.


Danipulok:
When I was working on tests, I discovered that `agent.run_stream` + `result.stream_text(delta=False)` (default behavior) does not call `TextOutput(my_tool)`

It seems this is what was discussed in https://github.com/pydantic/pydantic-ai/issues/3393#issuecomment-3518782057, so it does not deserve a new issue?

Here are the tests that demonstrate bugged behavior: [link](https://github.com/Danipulok/pydantic-ai/commit/e68a9de7b577fb7b0c5769e7247e81ba383a21e8)

I think I would like to include the failing streaming test with `xfail`, but what issue/comment should I link? Or maybe don't commit it at all?

Danipulok:
About tests: I created a new class designed to test `partial_output` and separated tests into tests for sync and streaming mode.
The tests were **not** deleted, just moved and formatted to look alike

Here are the changes:

**test_output_validator_partial_sync**

before: `test_agent.py::test_output_validator_partial_sync`
https://github.com/pydantic/pydantic-ai/blob/ccefddceb68eedab04619e7ab1fd12117e0c1153/tests/test_agent.py#L363-L377
after: `test_agent.py::TestPartialOutput::test_output_validator_text`
https://github.com/Danipulok/pydantic-ai/blob/d08191ee0e84ea2ae54fd4c28e8d5076986c281e/tests/test_agent.py#L370-L387

**test_output_validator_partial_stream_text**
before: `test_agent.py::test_output_validator_partial_stream_text`
https://github.com/pydantic/pydantic-ai/blob/ccefddceb68eedab04619e7ab1fd12117e0c1153/tests/test_agent.py#L380-L409
after: `test_streaming.py::TestPartialOutput::test_output_validator_text`
https://github.com/Danipulok/pydantic-ai/blob/d08191ee0e84ea2ae54fd4c28e8d5076986c281e/tests/test_streaming.py#L762-L789

**test_output_validator_partial_stream_output**

before: `test_agent.py::test_output_validator_partial_stream_output`
https://github.com/pydantic/pydantic-ai/blob/ccefddceb68eedab04619e7ab1fd12117e0c1153/tests/test_agent.py#L412-L439
after: `test_streaming.py::TestPartialOutput::test_output_validator_structured`
https://github.com/Danipulok/pydantic-ai/blob/d08191ee0e84ea2ae54fd4c28e8d5076986c281e/tests/test_streaming.py#L791-L818

docs/output.md
```diff
@@ -258,6 +258,90 @@ print(result.output)
 
 _(This example is complete, it can be run "as is")_
 
+#### Handling partial output in output functions
```
DouweM, line 265: Let's rename this `Handling partial output` and also drop `in output validators` from the other section as it's already under "Output validators"

docs/output.md
```diff
@@ -258,6 +258,90 @@ print(result.output)
...
+
+!!! warning "Output functions are called multiple times during streaming"
```
DouweM: I'd prefer for this to be top-level text rather than the entire docs section to be an indented warning block

docs/output.md
```diff
@@ -258,6 +258,90 @@ print(result.output)
...
+    When using streaming mode (`run_stream()`), output functions are called **multiple times** — once for each partial output received from the model, and once for the final complete output.
```
DouweM: I think we should say this in the "Handling partial output in output validators" section as well. Maybe the 2 sections can be merged as output validators and functions are very similar?

docs/output.md
```diff
@@ -258,6 +258,90 @@ print(result.output)
...
+
+    For output functions with **side effects** (e.g., sending notifications, logging, database updates), you should check the [`RunContext.partial_output`][pydantic_ai.tools.RunContext.partial_output] flag to avoid executing side effects on partial data.
+
+**How `partial_output` works:**
+
+- **In sync mode** (`run_sync()`):
```
DouweM: As I mentioned on the other PR, it's just specific to sync mode or `run_sync`: `run` (async), `iter` and `run_stream_events` have the same behavior. It's really just "`run_stream` vs the rest"

docs/output.md
```diff
@@ -258,6 +258,90 @@ print(result.output)
...
+agent = Agent('openai:gpt-5', output_type=save_to_database)
+
+result = agent.run_sync('Create a record with name "test" and value 42')
+print(result.output)
+#> name='test' value=42
+```
+
+_(This example is complete, it can be run "as is")_
+
+**Example without side effects (transformation only):**
```
DouweM: We don't need this example as it's the "base case" already covered above

docs/output.md
```diff
@@ -258,6 +258,90 @@ print(result.output)
...
+
+```python {title="output_function_with_side_effects.py"}
+from pydantic import BaseModel
+
+from pydantic_ai import Agent, RunContext
+
+
+class DatabaseRecord(BaseModel):
+    name: str
+    value: int
```
DouweM: This is actually not a great example because both of these fields are required, meaning the partial data wouldn't validate anyway until the output is complete, so there would not be any partial output in this case.

docs/output.md
```diff
@@ -258,6 +258,90 @@ print(result.output)
...
+
+    # Only execute side effect for the final output
+    print(f'Saving to database: {record.name} = {record.value}')
+    #> Saving to database: test = 42
+    return record
+
+
+agent = Agent('openai:gpt-5', output_type=save_to_database)
+
+result = agent.run_sync('Create a record with name "test" and value 42')
```
DouweM: We're not using `run_stream` here 😄 

This example needs some work to actually show the problem + use case

docs/output.md
DouweM: I'd rather this be a paragraph than a list btw

docs/output.md
```diff
@@ -258,6 +258,90 @@ print(result.output)
...
 _(This example is complete, it can be run "as is")_
 
+#### Handling partial output in output functions
+
+!!! warning "Output functions are called multiple times during streaming"
+    When using streaming mode (`run_stream()`), output functions are called **multiple times** — once for each partial output received from the model, and once for the final complete output.
+
+    For output functions with **side effects** (e.g., sending notifications, logging, database updates), you should check the [`RunContext.partial_output`][pydantic_ai.tools.RunContext.partial_output] flag to avoid executing side effects on partial data.
+
+**How `partial_output` works:**
```
DouweM: No "fake" subheadings please, I'd rather have paragraphs

docs/output.md
```diff
@@ -258,6 +258,90 @@ print(result.output)
...
+print(result.output)
+#> username='johndoe' email='john@example.com'
+```
+
+_(This example is complete, it can be run "as is")_
+
+**Best practices:**
+
+- If your output function **has** side effects (database writes, API calls, notifications) → use `ctx.partial_output` to guard them
+- If your output function only **transforms** data (formatting, validation, normalization) → no need to check the flag
```
DouweM: This is duplicative with `For output functions with **side effects** (e.g., sending notifications, logging, database updates), you should check the [`RunContext.partial_output`][pydantic_ai.tools.RunContext.partial_output] flag to avoid executing side effects on partial data.` above. I'd rather have one authoritative paragraph

tests/test_streaming.py
```diff
@@ -747,6 +752,100 @@ async def ret_a(x: str) -> str:  # pragma: no cover
...
+        assert outputs[-1] == Foo(a=42, b='foo')
+        assert call_log == snapshot(
+            [
+                (Foo(a=42, b='f'), True),
+                (Foo(a=42, b='foo'), True),
+                (Foo(a=42, b='foo'), False),
+            ]
+        )
+
+    async def test_output_function_structured(self):
```
DouweM, line 868: Let's add that `test_output_function_text` test with `TextOutput(func)`, create a new issue for that bug, and then skip the test

DouweM (changes requested):

Danipulok:
@DouweM I think the PR is ready to review. I've updated the docs and created and linked a new issue #3804

docs/output.md
```diff
@@ -258,6 +258,91 @@ print(result.output)
 
 _(This example is complete, it can be run "as is")_
 
+#### Handling partial output
+
+When streaming with `run_stream()`, output validators and output functions are called **multiple times** — once for each partial output received from the model, and once for the final complete output.
```
DouweM: We'll need to have something in the output validators section as well. I think it will make sense to split this up into a separate subsection for each feature (with some duplicated text about `run_stream` etc)

docs/output.md
```diff
@@ -258,6 +258,91 @@ print(result.output)
...
+
+For output validators or functions with **side effects** (e.g., sending notifications, logging, database updates), you should check the [`RunContext.partial_output`][pydantic_ai.tools.RunContext.partial_output] flag to avoid executing side effects on partial data.
```
DouweM: We should also mention you can use this when only the complete result should be validated, not intermediate values

docs/output.md
```diff
@@ -262,6 +262,51 @@ print(result.output)
 
 _(This example is complete, it can be run "as is")_
 
+#### Handling partial output in output functions
+
+When streaming with `run_stream()` or `run_stream_sync()`, output functions are called **multiple times** — once for each partial output received from the model, and once for the final complete output.
+
+You should check the [`RunContext.partial_output`][pydantic_ai.tools.RunContext.partial_output] flag when your output function has **side effects** (e.g., sending notifications, logging, database updates) that should only execute on the final output
```
DouweM:
```suggestion
You should check the [`RunContext.partial_output`][pydantic_ai.tools.RunContext.partial_output] flag when your output function has **side effects** (e.g., sending notifications, logging, database updates) that should only execute on the final output.
```

PR #3732, opened 2025-12-15
Allow typed `RunContext[Deps]` in `TextOutput` signature

dsfaccini:
`TextOutput` supports passing in functions that take in the agent's `RunContext[Deps]`, but pyright complains when the `Deps` type is specified in the signature. This PR adds support for properly typing the `RunContext`

Context (__badum tss__): https://pydantic.slack.com/archives/C081LUCJ4KX/p1765814128383649?thread_ts=1765812544.296369&cid=C081LUCJ4KX

Note: sorry about the change in the `uv.lock`, that seems to have escaped the `google-genai bump` PR

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/output.py
+++ b/pydantic_ai_slim/pydantic_ai/output.py
@@ -34,7 +34,6 @@ __all__ = (
 
 T = TypeVar('T')
 T_co = TypeVar('T_co', covariant=True)
-TextOutputAgentDepsT = TypeVar('TextOutputAgentDepsT', default=None, contravariant=True)
 
 OutputDataT = TypeVar('OutputDataT', default=str, covariant=True)
 """Covariant type variable for the output data type of a run."""
@@ -62,8 +61,8 @@ See [output docs](../output.md) for more information.
 
 TextOutputFunc = TypeAliasType(
     'TextOutputFunc',
-    Callable[[RunContext[TextOutputAgentDepsT], str], Awaitable[T_co] | T_co] | Callable[[str], Awaitable[T_co] | T_co],
-    type_params=(T_co, TextOutputAgentDepsT),
+    Callable[[RunContext[Any], str], Awaitable[T_co] | T_co] | Callable[[str], Awaitable[T_co] | T_co],
+    type_params=(T_co,),
 )
 """Definition of a function that will be called to process the model's plain text output. The function must take a single string argument.
 
@@ -261,7 +260,7 @@ class OutputObjectDefinition:
 
 
 @dataclass
-class TextOutput(Generic[OutputDataT, TextOutputAgentDepsT]):
+class TextOutput(Generic[OutputDataT]):
     """Marker class to use text output for an output function taking a string argument.
 
     Example:
@@ -283,7 +282,7 @@ class TextOutput(Generic[OutputDataT, TextOutputAgentDepsT]):
     ```
     """
 
-    output_function: TextOutputFunc[OutputDataT, TextOutputAgentDepsT]
+    output_function: TextOutputFunc[OutputDataT]
     """The function that will be called to process the model's plain text output. The function must take a single string argument."""
 
 
@@ -356,7 +355,7 @@ def StructuredDict(
 
 _OutputSpecItem = TypeAliasType(
     '_OutputSpecItem',
-    OutputTypeOrFunction[T_co] | ToolOutput[T_co] | NativeOutput[T_co] | PromptedOutput[T_co] | TextOutput[T_co, Any],
+    OutputTypeOrFunction[T_co] | ToolOutput[T_co] | NativeOutput[T_co] | PromptedOutput[T_co] | TextOutput[T_co],
     type_params=(T_co,),
 )
 
--- a/tests/typed_agent.py
+++ b/tests/typed_agent.py
@@ -287,14 +287,13 @@ Agent('test', tools=[Tool(foobar_ctx)])  # pyright: ignore[reportArgumentType,re
 # since deps are not set, they default to `None`, so can't be `int`
 Agent('test', tools=[Tool(foobar_plain)], deps_type=int)  # pyright: ignore[reportArgumentType,reportCallIssue]
 
-# TextOutput with RunContext
+# TextOutput with RunContext uses RunContext[Any], so deps_type is not checked.
+# This is intentional: type checking deps in output functions isn't feasible because
+# ToolOutput and plain output functions take arbitrary args, so the type checker
+# treats RunContext as just another arg rather than enforcing deps_type compatibility.
 text_output_with_ctx = TextOutput(str_to_regex_with_ctx)
-assert_type(text_output_with_ctx, TextOutput[re.Pattern[str], int])
+assert_type(text_output_with_ctx, TextOutput[re.Pattern[str]])
 Agent('test', output_type=text_output_with_ctx, deps_type=int)
-Agent('test', output_type=text_output_with_ctx, deps_type=bool)  # bool is subclass of int, works with contravariant
-# NOTE: The following don't produce type errors because _OutputSpecItem uses TextOutput[T_co, Any]
-# which erases the deps type constraint. This is a known limitation.
-# https://github.com/pydantic/pydantic-ai/pull/3732#discussion_r2628741424
 Agent('test', output_type=text_output_with_ctx, deps_type=str)
 Agent('test', output_type=text_output_with_ctx)
 


docs/output.md
```diff
@@ -240,9 +240,12 @@ RouterFailure(explanation='I am not equipped to provide travel information, such
 If you provide an output function that takes a string, Pydantic AI will by default create an output tool like for any other output function. If instead you'd like the model to provide the string using plain text output, you can wrap the function in the [`TextOutput`][pydantic_ai.output.TextOutput] marker class. If desired, this marker class can be used alongside one or more [`ToolOutput`](#tool-output) marker classes (or unmarked types or functions) in a list provided to `output_type`.
```
DouweM: I'd rather remove the example below, and explicitly state that `RunContext` is supported like so: (that new paragraph was mostly copied from the section above)

```suggestion
If you provide an output function that takes a string, Pydantic AI will by default create an output tool like for any other output function. If instead you'd like the model to provide the string using plain text output, you can wrap the function in the [`TextOutput`][pydantic_ai.output.TextOutput] marker class. 

If desired, this marker class can be used alongside one or more [`ToolOutput`](#tool-output) marker classes (or unmarked types or functions) in a list provided to `output_type`.

Like other output functions, text output functions can optionally take [`RunContext`][pydantic_ai.tools.RunContext] as the first argument, and can raise [`ModelRetry`][pydantic_ai.exceptions.ModelRetry] to ask the model to try again with modified arguments (or with a different output type).
```

pydantic_ai_slim/pydantic_ai/output.py
```diff
@@ -60,8 +61,8 @@
 
 TextOutputFunc = TypeAliasType(
     'TextOutputFunc',
-    Callable[[RunContext, str], Awaitable[T_co] | T_co] | Callable[[str], Awaitable[T_co] | T_co],
-    type_params=(T_co,),
+    Callable[[RunContext[AgentDepsT], str], Awaitable[T_co] | T_co] | Callable[[str], Awaitable[T_co] | T_co],
```
DouweM: See https://github.com/pydantic/pydantic-ai/pull/3319 and https://github.com/pydantic/pydantic-ai/pull/3721; we shouldn't be reusing generic type vars across different files, but instead need a new one here. 
And we should think about covariant vs contravariant etc.

The easiest way to ensure it works as expected is to add some cases that should be valid to `typed_agent.py`, which is automatically type checked

tests/test_agent.py
```diff
@@ -1167,6 +1167,24 @@ def call_tool(messages: list[ModelMessage], info: AgentInfo) -> ModelResponse:
     )
 
 
+def test_output_type_text_output_function_with_deps():
```
DouweM: I'm pretty sure we've already tested somewhere that it _works_, this PR is just for fixing the type checking, so let's remove the new tests and add some cases to `typed_agent.py` instead.

uv.lock
```diff
@@ -5666,7 +5666,7 @@ requires-dist = [
     { name = "fastmcp", marker = "extra == 'fastmcp'", specifier = ">=2.12.0" },
     { name = "genai-prices", specifier = ">=0.0.40" },
     { name = "google-auth", marker = "extra == 'vertexai'", specifier = ">=2.36.0" },
-    { name = "google-genai", marker = "extra == 'google'", specifier = ">=1.51.0" },
```
DouweM: Not part of this PR!

docs/output.md
```diff
@@ -237,26 +237,11 @@ RouterFailure(explanation='I am not equipped to provide travel information, such
 
 #### Text output
 
-If you provide an output function that takes a string, Pydantic AI will by default create an output tool like for any other output function. If instead you'd like the model to provide the string using plain text output, you can wrap the function in the [`TextOutput`][pydantic_ai.output.TextOutput] marker class. If desired, this marker class can be used alongside one or more [`ToolOutput`](#tool-output) marker classes (or unmarked types or functions) in a list provided to `output_type`.
+If you provide an output function that takes a string, Pydantic AI will by default create an output tool like for any other output function. If instead you'd like the model to provide the string using plain text output, you can wrap the function in the [`TextOutput`][pydantic_ai.output.TextOutput] marker class.
 
-```python {title="text_output_function.py"}
```
DouweM, line 242: Sorry I didn't mean to remove the example entirely; just the extra `RunContext`-based on you'd added 😄

tests/test_examples.py
```diff
@@ -344,6 +344,7 @@ async def call_tool(
     'What is the capital of the UK?': 'The capital of the UK is London.',
     'What is the capital of Mexico?': 'The capital of Mexico is Mexico City.',
     'Who was Albert Einstein?': 'Albert Einstein was a German-born theoretical physicist.',
+    'Hello world': 'Hello world',
```
DouweM: Not needed anymore

tests/typed_agent.py
```diff
@@ -281,6 +285,11 @@ def my_method(self) -> bool:
 # since deps are not set, they default to `None`, so can't be `int`
 Agent('test', tools=[Tool(foobar_plain)], deps_type=int)  # pyright: ignore[reportArgumentType,reportCallIssue]
 
+# TextOutput with RunContext
+Agent('test', output_type=TextOutput(str_to_regex_with_ctx), deps_type=int)
```
DouweM: To verify that `contravariant=True` is the right option, we should also have an example where the `deps_type` is a subclass of the type on the function's ctx, e.g. `bool`. That should work, because a function that takes an int is also able to handle a bool (which is essentially 1 or 0.

DouweM: Can we move `TextOutput(str_to_regex_with_ctx)` to a variable and add an `assert_type` so we can verify its type is inferred correctly?

tests/typed_agent.py
```diff
@@ -281,6 +285,11 @@ def my_method(self) -> bool:
...
+Agent('test', output_type=TextOutput(str_to_regex_with_ctx), deps_type=str)  # pyright: ignore[reportArgumentType,reportCallIssue]
+Agent('test', output_type=TextOutput(str_to_regex_with_ctx))  # pyright: ignore[reportArgumentType,reportCallIssue]
```
DouweM: We can then reuse the var here

DouweM: In `_OutputSpecItem`, we fill in the generic param as `Any`, so I'm curious if that could cause any issues here because maybe it'll accept any type instead of just same one as `deps_type`? It'd be interesting to see the typing errors you got here before you add the `pyright: ignore`s. (just pasting them into a GH comment here is fine)

DouweM (changes requested):

docs/output.md
dsfaccini: lmao I was like "okay..."

tests/typed_agent.py
dsfaccini: no type errors, I added a NOTE for it

docs/output.md
```diff
@@ -258,6 +258,8 @@ print(result.output)
 
 _(This example is complete, it can be run "as is")_
 
+Like other output functions, text output functions can optionally take [`RunContext`][pydantic_ai.tools.RunContext] as the first argument, and can raise [`ModelRetry`][pydantic_ai.exceptions.ModelRetry] to ask the model to try again with modified arguments (or with a different output type).
```
DouweM: Above the example please :) I prefer to explain, then show

tests/test_agent.py
DouweM: I think we can remove this test

tests/typed_agent.py
```diff
@@ -283,6 +287,16 @@ def my_method(self) -> bool:
 # since deps are not set, they default to `None`, so can't be `int`
 Agent('test', tools=[Tool(foobar_plain)], deps_type=int)  # pyright: ignore[reportArgumentType,reportCallIssue]
 
+# TextOutput with RunContext
+text_output_with_ctx = TextOutput(str_to_regex_with_ctx)
+assert_type(text_output_with_ctx, TextOutput[re.Pattern[str], int])
+Agent('test', output_type=text_output_with_ctx, deps_type=int)
+Agent('test', output_type=text_output_with_ctx, deps_type=bool)  # bool is subclass of int, works with contravariant
+# NOTE: The following don't produce type errors because _OutputSpecItem uses TextOutput[T_co, Any]
+# which erases the deps type constraint.
```
DouweM: If we can't make those give a typing error, then maybe it's not worth adding `TextOutputAgentDepsT` and we can just make it `RunContext[Any]`. 

Ideally, we would go a step further and add `DepsT` to `_OutputSpecItem`, `OutputSpec` etc, so that the user _is_ warned if their `ctx: RunContext[Foo]` arg doesn't match the agent's `deps_type`, which would work for `TextOutput(func)`, but unfortunately there's now way to make that work for output functions in general, e.g. `output_type=func` or `output_type=ToolOutput(func)`, as those functions that arbitrary args, so if the first arg `ctx: RunContext[Foo]` while `deps_type=Bar`, the type checker will just treat it like an arbitrary arg, instead of complaining. (ping me if that doesn't make sense; typing is tricky)

So I think the best we can do is just make it `RunContext[Any]` and accept that this part isn't type checked.

dsfaccini: I think I was able to enforce this

pydantic_ai_slim/pydantic_ai/output.py
```diff
@@ -60,8 +64,8 @@
 
 TextOutputFunc = TypeAliasType(
     'TextOutputFunc',
-    Callable[[RunContext, str], Awaitable[T_co] | T_co] | Callable[[str], Awaitable[T_co] | T_co],
-    type_params=(T_co,),
+    Callable[[RunContext[TextOutputAgentDepsT], str], Awaitable[T_co] | T_co] | Callable[[str], Awaitable[T_co] | T_co],
```
DouweM: Yeah looking at this now, since this will only work for `TextOutput` and not output functions in general, I don't think it's worth the extra (out of order) type var and would prefer to just have `RunContext[Any]` here and leave it at that

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/output.py
```diff
@@ -259,7 +261,7 @@ class OutputObjectDefinition:
 
 
 @dataclass
-class TextOutput(Generic[OutputDataT]):
+class TextOutput(Generic[OutputDataT, TextOutputAgentDepsT]):
```
DouweM: Let's drop `TextOutputAgentDepsT` entirely and use `RunContext[Any]` in `TextOutputFunc`, as we don't actually use the generic param and it bothers me that it's out of order compared to everything else that takes Deps followed by Output 😄

tests/typed_agent.py
```diff
@@ -283,6 +287,17 @@ def my_method(self) -> bool:
 # since deps are not set, they default to `None`, so can't be `int`
 Agent('test', tools=[Tool(foobar_plain)], deps_type=int)  # pyright: ignore[reportArgumentType,reportCallIssue]
 
+# TextOutput with RunContext
+text_output_with_ctx = TextOutput(str_to_regex_with_ctx)
+assert_type(text_output_with_ctx, TextOutput[re.Pattern[str], int])
+Agent('test', output_type=text_output_with_ctx, deps_type=int)
+Agent('test', output_type=text_output_with_ctx, deps_type=bool)  # bool is subclass of int, works with contravariant
+# NOTE: The following don't produce type errors because _OutputSpecItem uses TextOutput[T_co, Any]
```
DouweM: I agree it's worth having a comment like this, but this'll have to be rewritten if my comment above is addressed

Issue #3722, opened 2025-12-13
Support `CachePoint` in `ModelResponse` for Few-Shot AI Response Caching in Bedrock


mshavliuk:
**Description:**
First off, thanks for the great work in \#3438 to add Bedrock prompt caching support\! 👍

### The Issue

While the current implementation covers many scenarios, it does not yet support **AI response caching** (caching `ModelRequest` or model output messages), which is crucial for efficient few-shot prompting.

### Use Case

I am implementing a few-shot workflow where the prompt structure looks like this:

1.  System Prompt
2.  Few-shot user input example
3.  Few-shot AI output example **\<-- I need to place a Cache Point here**
4.  Actual user input (always unique, should not be cached)

Currently, it seems we cannot attach a `CachePoint` specifically to the AI's output example (a `TextPart` within a model message). This prevents caching the static portion of the few-shot examples effectively.

### Proposed Solution

I believe `TextPart` (and potentially other message parts) should be updated to allow providing cache points. This would enable users to mark the end of the static few-shot context—including the AI's example response—so that the entire block can be cached before the unique user input is appended.

**Example desired behavior:**

```python
# Pseudo-code illustrating the need
message = ModelResponse(
    parts=[
        TextPart(
            content=["This is the static AI example output.", CachePoint()],
        )
    ]
)
```

Wh1isper:
👍🏻Anthropic models can also benefit from this

mshavliuk:
One simple workaround I found is to add a short user input prefix and a cache point after that, followed by the real user prompt:
```
  def render_messages(msgs: list[ModelMessage])
    few_shot_messages = [...]
    user_prefix = ModelRequest(
        parts=[
            UserPromptPart(
                content=[
                    "Now consider this request:\n",
                    CachePoint(),
                ]
            )
        ]
    )

    return [*few_shot_messages, user_prefix, *msgs]

agent = Agent(
        ...
        history_processors=[render_messages],
    )
```

DouweM:
@mshavliuk Thanks for the request, feel free to submit a PR that adds `CachePoint` to the `ModelResponsePart` union -- I think it should be directly in `ModelResponse.parts`, not on `TextPart`.

vricciardulli:
I've opened a **draft** PR here https://github.com/pydantic/pydantic-ai/pull/3738 (no tests and no updates to the docs yet).

[...]

PR #3738, opened 2025-12-16
Support `CachePoint` for few-shot prompting for assistant messages (Bedrock and Anthropic)

vricciardulli:
Aiming to support `CachePoint` for assistant messages (i.e. `ModelResponsePart`) for `Anthropic` and `Bedrock`.

I still need to:
* Update docs.
* Add/fix tests.

I'm opening this as draft as, while developing, I've realized that this is probably too much complexity for little gain, while adding a short, cached, user prompt after the last assistant is easy and intuitive instead (as pointed out also in https://github.com/pydantic/pydantic-ai/issues/3722#issuecomment-3654808465).

So my question is whether you think I should keep on working on this or not?

Relevant links:
* Closes https://github.com/pydantic/pydantic-ai/issues/3722
* [Slack](https://pydanticlogfire.slack.com/archives/C083V7PMHHA/p1764773692581329)

### DIFF
--- a/tests/models/test_anthropic.py
+++ b/tests/models/test_anthropic.py
@@ -346,6 +346,62 @@ async def test_cache_point_adds_cache_control(allow_model_requests: None):
     )
 
 
+async def test_cache_point_few_shot_prompting_caches_assistant_response(allow_model_requests: None):
+    """Test that CachePoint correctly adds cache_control to content blocks of an assistant message.
+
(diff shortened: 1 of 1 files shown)


DouweM:
Closing because of https://github.com/pydantic/pydantic-ai/issues/3722#issuecomment-3658317604

PR #3741, opened 2025-12-16
Better follow otel semconv

dmontagu:
@alexmojaki not sure if you wanted to use `response` instead of `result` today; also not sure if we should wait to introduce these changes for "version 4" of the observability or similar. Or if we should keep this open and just make the changes when we release the next major version of pydantic-ai..?


pydantic_ai_slim/pydantic_ai/_otel_messages.py
```diff
@@ -25,36 +31,57 @@ class ToolCallPart(TypedDict):
...
-    type: Literal['thinking']
+    type: Literal['blob']
+    blob: NotRequired[str]
+    modality: NotRequired[str]
+
+
+class ReasoningPart(TypedDict):
+    """A reasoning/thinking part in a GenAI message.
+
+    Per the semantic conventions, uses 'reasoning' type.
```
dmontagu, line 69: You can find `UriPart`, `BlobPart`, and `ReasoningPart` in https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-input-messages.json with this schema. Not sure how practical it is for us to change this in the immediate future but I think we'll want to comply with those conventions asap..

pydantic_ai_slim/pydantic_ai/_otel_messages.py
```diff
@@ -25,36 +31,57 @@ class ToolCallPart(TypedDict):
 
 
 class ToolCallResponsePart(TypedDict):
+    """A tool call response part in a GenAI message."""
+
     type: Literal['tool_call_response']
     id: str
     name: str
+    # TODO: This should be `response` not `result`
```
dmontagu, line 39: This is also noted in https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-input-messages.json, I can update the PR to make this change you want, or I can revert the changes to the other types below.

Kludex:
@dmontagu @alexmojaki please sort this out, or close this PR.

dmontagu:
pretty sure we already merged something else that replaced this.

Issue #3734, opened 2025-12-15
AGUIEventStream doesn't set timestamps on events


jens-gilion:
### Initial Checks

- [x] I confirm that I'm using the latest version of Pydantic AI
- [x] I confirm that I searched for my issue in https://github.com/pydantic/pydantic-ai/issues before opening this issue

### Description

The `AGUIEventStream` class doesn't set timestamps when creating events, even though the AG-UI protocol defines `timestamp?: number` as an optional field on `BaseEvent`.

This means all events come through with `timestamp=None`, which makes it hard for frontends to:
- Order events that arrive out of sequence
- Show when messages/tools were actually executed
- Debug timing issues

I've been working around it by setting timestamps after receiving events:

```python
async for agui_event in agui_stream.transform_stream(events):
    agui_event.timestamp = agui_event.timestamp or int(datetime.now(UTC).timestamp() * 1000)
    # ... rest of processing
```

But every consumer has to do this manually.

### Location

In `pydantic_ai/ui/ag_ui/_event_stream.py`, all the event creation sites are missing timestamp parameters. For example:

```python
# Line 90-93
yield RunStartedEvent(
    thread_id=self.run_input.thread_id,
    run_id=self.run_input.run_id,
)  # Missing timestamp
```

Same issue in `handle_text_start()`, `handle_tool_call_start()`, etc. - about 19 places total.

### Fix

Add timestamps when creating events:

```python
def _now_ms() -> int:
    import time
    return int(time.time() * 1000)

yield RunStartedEvent(
    thread_id=self.run_input.thread_id,
    run_id=self.run_input.run_id,
    timestamp=_now_ms(),
)
```

### Python, Pydantic AI & AG-UI version

```
pydantic-ai: 1.27.0+
ag-ui-protocol: 0.1.10
python: 3.12
```

DouweM:
@jens-gilion Good catch; feel free to submit a PR to fix this!

jens-gilion:
@DouweM would you have any opinion on the implementation? We could give the events' optional timestamp field a default being int(now().seconds * 1_000)  --> or set it explicitly for each object instantiation.

Also curious whether you think system-timestamp or UTC is the way to go here

DouweM:
> We could give the events' optional timestamp field a default being int(now().seconds * 1_000) --> or set it explicitly for each object instantiation.

[...]

PR #3742, opened 2025-12-16
Set timestamps on AG-UI events

thejens:
Addresses https://github.com/pydantic/pydantic-ai/issues/3734

All AG-UI events now include timestamps to help frontends order events and show when things actually happened.

## Changes
- Added `handle_event()` override that sets timestamps on all transformed events if not already set
- Explicitly set timestamps on lifecycle events (RunStarted, RunFinished, RunError)
- Updated all test assertions to expect timestamp fields
- Added `test_timestamps_are_set()` to verify the behavior

Timestamps are in milliseconds (JS format) using `now_utc().timestamp() * 1000`.

The two-layer approach (explicit + override) ensures no events slip through without timestamps while avoiding double-setting.

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
+++ b/pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
@@ -11,6 +11,7 @@ from collections.abc import AsyncIterator, Iterable
 from dataclasses import dataclass, field
 from typing import Final
 
+from ..._utils import now_utc
 from ...messages import (
     BuiltinToolCallPart,
     BuiltinToolReturnPart,
@@ -26,7 +27,7 @@ from ...messages import (
 )
 from ...output import OutputDataT
 from ...tools import AgentDepsT
-from .. import SSE_CONTENT_TYPE, UIEventStream
+from .. import SSE_CONTENT_TYPE, NativeEvent, UIEventStream
 
 try:
     from ag_ui.core import (
@@ -86,10 +87,22 @@ class AGUIEventStream(UIEventStream[RunAgentInput, BaseEvent, AgentDepsT, Output
     def encode_event(self, event: BaseEvent) -> str:
         return self._event_encoder.encode(event)
 
+    @staticmethod
+    def _get_timestamp() -> int:
+        return int(now_utc().timestamp() * 1_000)
+
+    async def handle_event(self, event: NativeEvent) -> AsyncIterator[BaseEvent]:
+        """Override to set timestamps on all AG-UI events."""
+        async for agui_event in super().handle_event(event):
+            if agui_event.timestamp is None:
+                agui_event.timestamp = self._get_timestamp()
+            yield agui_event
+
     async def before_stream(self) -> AsyncIterator[BaseEvent]:
         yield RunStartedEvent(
             thread_id=self.run_input.thread_id,
             run_id=self.run_input.run_id,
+            timestamp=self._get_timestamp(),
         )
 
     async def before_response(self) -> AsyncIterator[BaseEvent]:
@@ -104,11 +117,12 @@ class AGUIEventStream(UIEventStream[RunAgentInput, BaseEvent, AgentDepsT, Output
             yield RunFinishedEvent(
                 thread_id=self.run_input.thread_id,
                 run_id=self.run_input.run_id,
+                timestamp=self._get_timestamp(),
             )
 
     async def on_error(self, error: Exception) -> AsyncIterator[BaseEvent]:
         self._error = True
-        yield RunErrorEvent(message=str(error))
+        yield RunErrorEvent(message=str(error), timestamp=self._get_timestamp())
 
     async def handle_text_start(self, part: TextPart, follows_text: bool = False) -> AsyncIterator[BaseEvent]:
         if follows_text:
--- a/tests/test_ag_ui.py
+++ b/tests/test_ag_ui.py
@@ -52,7 +52,7 @@ from pydantic_ai.models.test import TestModel
 from pydantic_ai.output import OutputDataT
 from pydantic_ai.tools import AgentDepsT, ToolDefinition
 
-from .conftest import IsDatetime, IsSameStr, try_import
+from .conftest import IsDatetime, IsInt, IsSameStr, try_import
 
 with try_import() as imports_successful:
     from ag_ui.core import (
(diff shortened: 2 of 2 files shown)


pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
```diff
@@ -104,11 +113,12 @@ async def after_stream(self) -> AsyncIterator[BaseEvent]:
             yield RunFinishedEvent(
                 thread_id=self.run_input.thread_id,
                 run_id=self.run_input.run_id,
+                timestamp=int(now_utc().timestamp() * 1000),
             )
 
     async def on_error(self, error: Exception) -> AsyncIterator[BaseEvent]:
         self._error = True
-        yield RunErrorEvent(message=str(error))
+        yield RunErrorEvent(message=str(error), timestamp=int(now_utc().timestamp() * 1000))
```
DouweM: Can this be a private helper method as we now repeat it 4 times?

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
```diff
@@ -86,10 +87,18 @@ def content_type(self) -> str:
     def encode_event(self, event: BaseEvent) -> str:
         return self._event_encoder.encode(event)
 
+    async def handle_event(self, event: NativeEvent) -> AsyncIterator[BaseEvent]:
+        """Override to set timestamps on all AG-UI events."""
+        async for agui_event in super().handle_event(event):
+            if agui_event.timestamp is None:
+                agui_event.timestamp = int(now_utc().timestamp() * 1000)
```
DouweM: Does AG-UI document this should be in ms precision?

thejens: Can't find any spec for this, but milliseconds is what the javascript date type expects and seems to be what the work-in-progress ag-ui Go SDK uses ([link](https://github.com/ag-ui-protocol/ag-ui/blob/10e4dea5379c315de6a09ec090227530f16290a1/sdks/community/go/pkg/core/events/events.go#L83))

DouweM: @thejens OK, that's pretty good evidence but that doesn't look like an official repo right? If you can verify it in the CopilotKit repo or in some other agent framework integration I'd feel more confident about it!

thejens: Yeah, I found it in a fork, updated the link to the official repo -- still a community provided SDK

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
thejens: Amended the commit using a private helper

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
thejens: The [community java sdk](https://github.com/ag-ui-protocol/ag-ui/blob/10e4dea5379c315de6a09ec090227530f16290a1/sdks/community/java/packages/core/src/main/java/com/agui/core/event/BaseEvent.java#L60) and [kotlin sdk](https://github.com/ag-ui-protocol/ag-ui/blob/10e4dea5379c315de6a09ec090227530f16290a1/sdks/community/kotlin/library/core/src/commonMain/kotlin/com/agui/core/types/Events.kt#L134) uses milliseconds as well it seems

DouweM: @thejens Ok that's good enough for me, thanks for double checking!

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
```diff
@@ -86,10 +87,22 @@ def content_type(self) -> str:
...
         return self._event_encoder.encode(event)
 
+    @staticmethod
+    def _get_now_utc_milliseconds() -> int:
+        return int(now_utc().timestamp() * 1_000)
+
+    async def handle_event(self, event: NativeEvent) -> AsyncIterator[BaseEvent]:
+        """Override to set timestamps on all AG-UI events."""
+        async for agui_event in super().handle_event(event):
+            if agui_event.timestamp is None:
```
DouweM, line 97: This is causing coverage to fail because the timestamp is _always_ none :)  I think we can drop the `if` here

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
```diff
@@ -86,10 +87,22 @@ def content_type(self) -> str:
     def encode_event(self, event: BaseEvent) -> str:
         return self._event_encoder.encode(event)
 
+    @staticmethod
+    def _get_now_utc_milliseconds() -> int:
```
DouweM: Small thing, but I like `_get_timestamp` better

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
jens-gilion: :) well it's always none in the current tests -- but I guess I'm not fluent enough in the ag-ui standard to understand if there are scenarios where events come on the stream from sources where the timestamp might be set -- I know it has quite some flexibility with custom events

Your call, I can remove the if (will solve my need) or furnish a test that streams an event with a pre-existing timestamp set

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
jens-gilion: _get_timestamp it is

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
DouweM: Yeah I suppose since we let the user stream their own AG-UI events from the `on_complete` handler, it's better to not overwrite that here. If you can add a test that'd be great!

jens-gilion: Added the test (sorry if me using two different github accounts (albeit w similar names) cause confusion in the comments)

DouweM:
@thejens Thanks Jens!

PR #3744, opened 2025-12-17
Add model profile flag for APIs that support native output but still require JSON schema in instructions

### LINKED ISSUE #3490: Some models ran through OpenAI-compatible APIs need JSON schema in instructions even when native structured output is used

zhaob1n:
closes #3490

Add a field "inject_schema_instructions_in_native_mode" in ModelProfile

The implementation is basically as per @DouweM , with the following improvement:

`NativeOutput` now can accept a template and json schema instruction is injected regardless of `inject_schema_instructions_in_native_mode` in that case. For example:
```python
result = await agent.run("how can I solve 8x + 7 = -23", output_type=NativeOutput(MathReasoning, template="{schema}"))
```

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/output.py
+++ b/pydantic_ai_slim/pydantic_ai/output.py
@@ -167,8 +167,7 @@ class NativeOutput(Generic[OutputDataT]):
     template: str | None
     """Template for the prompt passed to the model.
     The '{schema}' placeholder will be replaced with the output JSON schema.
-    If not specified, the default template specified on the model's profile will be used only if `native_output_requires_schema_in_instructions` on the profile is True.
-    If explicitly provided, it's used regardless of `native_output_requires_schema_in_instructions`.
+    If no template is specified but the model's profile indicates that it requires the schema to be sent as a prompt, the default template specified on the profile will be used.
     """
 
     def __init__(


pydantic_ai_slim/pydantic_ai/output.py
```diff
@@ -164,6 +164,11 @@ class NativeOutput(Generic[OutputDataT]):
     """The description of the structured output that will be passed to the model. If not specified and only one output is provided, the docstring of the output type or function will be used."""
     strict: bool | None
     """Whether to use strict mode for the output, if the model supports it."""
+    template: str | None
+    """Template for the prompt passed to the model.
+    The '{schema}' placeholder will be replaced with the output JSON schema.
+    If not specified, the default template specified on the model's profile will be used.
```
DouweM: We need to clarify that this is only used if 1) explicitly provided or 2) the model requires it

pydantic_ai_slim/pydantic_ai/profiles/__init__.py
```diff
@@ -51,6 +51,8 @@ class ModelProfile:
         """
     )
     """The instructions template to use for prompted structured output. The '{schema}' placeholder will be replaced with the JSON schema for the output."""
+    inject_schema_instructions_in_native_mode: bool = False
```
DouweM: I have a mild preference for describing traits of the models/providers, rather than our behavior:

```suggestion
    json_schema_output_requires_schema_in_instructions: bool = False
```

DouweM: Should we we set this to `True` on any profiles already?

DouweM: we may be able to remove this:

https://github.com/pydantic/pydantic-ai/blob/c1213e4a7a978099b64c642b38e1e041d18afad6/pydantic_ai_slim/pydantic_ai/models/outlines.py#L528-L534

if we set the new field here:

https://github.com/pydantic/pydantic-ai/blob/c1213e4a7a978099b64c642b38e1e041d18afad6/pydantic_ai_slim/pydantic_ai/providers/outlines.py#L33-L40

DouweM (changes requested):
@xcpky Thanks! Can you please add a test for the model profile behavior + new `template` field? If we update `OutlinesModel/Provider` to use the new model profile field, it may be sufficient to add a test to `test_outlines.py`.

pydantic_ai_slim/pydantic_ai/output.py
zhaob1n: I've added more clarification.

pydantic_ai_slim/pydantic_ai/profiles/__init__.py
zhaob1n: > I have a mild preference for describing traits of the models/providers, rather than our behavior:

I think `native_output_requires_schema_in_instructions` is better



> we may be able to remove this:
> 
> https://github.com/pydantic/pydantic-ai/blob/c1213e4a7a978099b64c642b38e1e041d18afad6/pydantic_ai_slim/pydantic_ai/models/outlines.py#L528-L534
> 
> if we set the new field here:
> 
> https://github.com/pydantic/pydantic-ai/blob/c1213e4a7a978099b64c642b38e1e041d18afad6/pydantic_ai_slim/pydantic_ai/providers/outlines.py#L33-L40

Done

zhaob1n:
> a

I've added the test, plz review.

pydantic_ai_slim/pydantic_ai/output.py
```diff
@@ -164,6 +164,12 @@ class NativeOutput(Generic[OutputDataT]):
     """The description of the structured output that will be passed to the model. If not specified and only one output is provided, the docstring of the output type or function will be used."""
     strict: bool | None
     """Whether to use strict mode for the output, if the model supports it."""
+    template: str | None
+    """Template for the prompt passed to the model.
+    The '{schema}' placeholder will be replaced with the output JSON schema.
+    If not specified, the default template specified on the model's profile will be used only if `native_output_requires_schema_in_instructions` on the profile is True.
+    If explicitly provided, it's used regardless of `native_output_requires_schema_in_instructions`.
```
DouweM: Users mostly don't deal with profile fields directly, and should be able to assume that we've done our best to make them correct out of the box so that they don't have to worry about them, so I'd rather say:

```suggestion
    If no template is specified but the model's profile indicates that it requires the schema to be sent as a prompt, the default template specified on the profile will be used.
```

Issue #3561, opened 2025-11-26
Add unified typed schemas for builtin tool return content


sarth6:
### Description

Currently, `BuiltinToolReturnPart.content` is typed as `Any` and simply passes through the raw dict/structure returned by each provider. This means users working with builtin tool results have no type hints or IDE support for accessing the return data.

For example, when using `CodeExecutionTool`, the content structure differs by provider:

```python
# Anthropic returns:
{
    'content': [],
    'return_code': 0,
    'stderr': '',
    'stdout': '15! = 1307674368000',
    'type': 'code_execution_result',
}

# OpenAI/Google likely return different structures
```

Current behavior

- `BuiltinToolReturnPart.content: Any` - no type safety
- Users must inspect raw responses or check provider docs to understand the structure
- No IDE autocompletion or type checking when accessing return data

Potential improvement

Define unified TypedDict or dataclass schemas for each builtin tool's return content that normalize across providers:

```python
class CodeExecutionResult(TypedDict):
    stdout: str
    stderr: str
    return_code: int
    # ... other normalized fields

class WebSearchResult(TypedDict):
    # ... normalized search result fields
```

This could either:

1. **Normalize at parse time** - Convert provider-specific formats into unified schemas when building `BuiltinToolReturnPart`
2. **Provide typed accessors** - Keep raw `content` but add typed properties/methods like `result.as_code_execution_result()` that parse and validate
3. **Generic typing** - Make `BuiltinToolReturnPart` generic: `BuiltinToolReturnPart[CodeExecutionResult]`

Questions

- Is this something the maintainers would consider valuable?
- Would a normalization layer add too much maintenance burden given provider format changes?
- Should raw provider data still be accessible alongside normalized types?

### References

- [`BaseToolReturnPart` with `content: Any`](https://github.com/pydantic/pydantic-ai/blob/main/pydantic_ai_slim/pydantic_ai/messages.py#L793-L800)
- [Builtin tool definitions](https://github.com/pydantic/pydantic-ai/blob/main/pydantic_ai_slim/pydantic_ai/builtin_tools.py)
- [Documentation showing raw dict examples](https://ai.pydantic.dev/builtin-tools/)

DouweM:
> * Is this something the maintainers would consider valuable?

@sarth6 Yes, definitely. The same goes for `BuiltinToolCallPart.args`, which for `CodeExecutionTool` should always contain a `'code'` key, etc.

[...]

PR #3754, opened 2025-12-17
Ensure thought signatures and other provider metadata survive a round trip through a Vercel AI frontend

### LINKED ISSUE #3748: Streaming with tools using VercelAIEventStream with Anthropic extended thinking models raises error
### Initial Checks

### Description

Sorry that this might be hard to reproduce. I've tried a few ways but putting this up first in case anyone runs into a similar error. It seems like there is a missing signature which causes the following error

## Error
```python
    status_code: 400
    body: {
        'type': 'error',
        'error': {
            'message': "messages.1.content.0.type: Expected `thinking` or `redacted_thinking`,
                       but found `text`. When `thinking` is enabled, a final `assistant` message
                       must start with a thinking block..."
        }
    }
```

## Workaround

 To fix this, I modified `pydantic_ai/ui/vercel_ai/_event_stream.py` to pass the signature in provider_metadata
```python
    async def handle_thinking_end(
        self, part: ThinkingPart, followed_by_thinking: bool = False
    ) -> AsyncIterator[BaseChunk]:
        yield ReasoningEndChunk(id=self.message_id)  # <-- part.signature not passed
```

```python
    async def handle_thinking_end(
        self, part: ThinkingPart, followed_by_thinking: bool = False
    ) -> AsyncIterator[BaseChunk]:
        provider_metadata = None
        if part.signature:
            pydantic_ai_meta = {"signature": part.signature}
            if part.provider_name:
                pydantic_ai_meta["provider_name"] = part.provider_name
            if part.id:
                pydantic_ai_meta["id"] = part.id
            provider_metadata = {"pydantic_ai": pydantic_ai_meta}

        yield ReasoningEndChunk(
            id=self.message_id, provider_metadata=provider_metadata
        )
```

### Example Code

```Python

```

```Text
python 3.13
pydantic 1.34.0
```

Light2Dark:
Fixes #3748 . Pulls provider details from reasoning end blocks, and passes them as provider_metadata.

Fixes anthropic issues: https://platform.claude.com/docs/en/build-with-claude/extended-thinking#extended-thinking-with-tool-use

Affected chunks (event-stream)
- TextStartChunk
- TextDeltaChunk
- TextEndChunk
- ReasoningStartChunk
- ReasoningDeltaChunk
- ReasoningEndChunk
- ToolInputAvailableChunk

Affected parts (adapter)
- assistant's TextUIPart
- assistant's BuiltInToolCallPart
- assistant's ToolCallPart
- assistant's FileUIPart

These messages have been modified to add provider_name
- TextPart
- BaseToolCallPart (only ToolCallPart affected)
- TextDeltaPart
- ToolCallPartDelta

### DIFF
--- a/tests/models/test_google.py
+++ b/tests/models/test_google.py
@@ -5660,22 +5660,3 @@ async def test_google_splits_tool_return_from_user_prompt(google_provider: Googl
             }
         ]
     )
-
-
-def test_process_part_with_provider_details_but_empty_provider_name():
-    """Test that _process_part handles provider_details without provider_name.
-
-    This tests the branch where thought_signature creates provider_details,
-    but provider_name is empty string (falsy), so provider_name should not be set on the item.
-    """
-    # Create a Part with thought_signature (which sets provider_details) and text
-    part = Part(text='test response', thought_signature=b'test_signature')
-
-    # Call _process_part with empty provider_name
-    item, code_execution_tool_call_id = _process_part(part, None, '')
-
-    # Verify provider_details is set but provider_name is not
-    assert item == snapshot(
-        TextPart(content='test response', provider_details={'thought_signature': 'dGVzdF9zaWduYXR1cmU='})
-    )
-    assert code_execution_tool_call_id is None
--- a/tests/models/test_xai.py
+++ b/tests/models/test_xai.py
@@ -1868,8 +1868,7 @@ async def test_xai_builtin_web_search_tool_stream(allow_model_requests: None, xa
             PartDeltaEvent(
                 index=0,
                 delta=ThinkingPartDelta(
-                    signature_delta='ech7t8FxCgEqI/eDeHo8Xd6VMDvsoYAuJZ7CwcXDJzLoflFyJRVo6l+sVq0kcryWOARaEtn8D8CTYhIH0NBDmnzAyPE25xNQdqkdVOQTxeQnRPFrbnBmjgXcEJmiIF3Ym2wuQm5t2c02XDd/O2va9ZjBMwGz/OR0SWUg7GlohlTMvpxvDmYbWl1/Co21yBuxk/7ghDX+E1WP3wlML7G6aAkQYM8GFUmk3Nz4Y+xM8JywizY8PPsbF7gsRs5vGqho3z5HAddTOKANUJzLcPPbeKips2K9ctbS2Wj6yYyD/yODpFzyf77fMk2w4UxMHfugzOmYej8GTRFHjMw5VDM9LGWNOYLVnE/08GTsuCSYgzww+iWdQ+Z1+BkjkFQKlOB9hTYmybKfhw8nvTi7DJEMIoRPaYn725I181Q8jHNwZmdRQcXAu+Gdnee9vdFBUILoB4zvKN+gn5KvtYFnFQ3frIURiJZXTy73g7CIQdhXGNRhJKbsQASvtyINBFTscbmnkX871dWA7UYqLrxy5ePhnXjrhyQd84Bco0JwUqaPt8Rqw44EBkeF5fC+zj7lzsjoR2NjvwXI3xkjJQ+uBnbZ+4uGGMvW1vwmi+aAoNFlYDIaXP0KIiyyGrZoGmjm/X+B1O614OkrZK6cisdbvtu9BHogQ8bzXAKB6kIIJWWt6j6j7oASgb5yAtiS7yCkYg3+L+EJUCOjaeecfBNtSCzIDb3i8RTrvquW/ZiEJj7ZSgiZL134zzEWO6ENL5tZWcE3kWm2LeFb3mOOuQnmzIDT0iDK0LxepAwnv4L8FgE7JD/qVZvMaVGUs0e2qnUsmbw7P4rKMcU0NtVFNLGYVRJaNSqZxxB/l0xfzcv0trmBVhw/YzjSlJUt77T2ZMZm0Y9KQJATd9EdVifzuZrjHUr1DPDRz3cfnBifTZTAIr/UcysQUGr6dAB6wEBRiQK0KoT/F2rhgIyJ2N0sqE0sUpKLGN1wcWkO+cbNER0rtyq2tapCREaZPsDgjZ4T+SLMxMraePnJodzfpDB1mLC3sBmYP66gx0Ay9iTo+2CwnueAFNjsq+fEtTTKu/WAGEEl/fVBvLDZnaNstXqxpBskHK6RKc+y+1TlXXfmWC38grI1oGC7VvzY+TJRGVBvtxU0YJ17IwWdtHgbjqCAx2btanfTWvigsObzRztA10ifWVLQwHKW5GYCn3YGjVzhwFwQJp7jj1/hmwCWwFM4Ijivolc5kPsWVUylbB6TUjIg4Y7zZO6ZryMXKNIJhbA7t6Vaej/E+0cfTFn6ACzFGnqqXxu62GLyte8NBGCEgvh4Bxhc+DoenauWKlCNktSR+i1M+mjg3pkgDcR0J5QiLNHvKhRblh5NGdHapYOphUDbWcwqdJ07VOwo9CHZ1WTox4VuVVSVz84Siag9ibuf18Z7XKd56MortTH7WAATHE6LB6RY8OG5+n1ZY9TWiHE3Mo6vKbHRKcEkbB2k/S5L28CL1a4C4yuAy5nNERL/+IRFhRpEUSH6GFQ4uIWcOcD62Zc',
-                    provider_name='xai',
+                    signature_delta='ech7t8FxCgEqI/eDeHo8Xd6VMDvsoYAuJZ7CwcXDJzLoflFyJRVo6l+sVq0kcryWOARaEtn8D8CTYhIH0NBDmnzAyPE25xNQdqkdVOQTxeQnRPFrbnBmjgXcEJmiIF3Ym2wuQm5t2c02XDd/O2va9ZjBMwGz/OR0SWUg7GlohlTMvpxvDmYbWl1/Co21yBuxk/7ghDX+E1WP3wlML7G6aAkQYM8GFUmk3Nz4Y+xM8JywizY8PPsbF7gsRs5vGqho3z5HAddTOKANUJzLcPPbeKips2K9ctbS2Wj6yYyD/yODpFzyf77fMk2w4UxMHfugzOmYej8GTRFHjMw5VDM9LGWNOYLVnE/08GTsuCSYgzww+iWdQ+Z1+BkjkFQKlOB9hTYmybKfhw8nvTi7DJEMIoRPaYn725I181Q8jHNwZmdRQcXAu+Gdnee9vdFBUILoB4zvKN+gn5KvtYFnFQ3frIURiJZXTy73g7CIQdhXGNRhJKbsQASvtyINBFTscbmnkX871dWA7UYqLrxy5ePhnXjrhyQd84Bco0JwUqaPt8Rqw44EBkeF5fC+zj7lzsjoR2NjvwXI3xkjJQ+uBnbZ+4uGGMvW1vwmi+aAoNFlYDIaXP0KIiyyGrZoGmjm/X+B1O614OkrZK6cisdbvtu9BHogQ8bzXAKB6kIIJWWt6j6j7oASgb5yAtiS7yCkYg3+L+EJUCOjaeecfBNtSCzIDb3i8RTrvquW/ZiEJj7ZSgiZL134zzEWO6ENL5tZWcE3kWm2LeFb3mOOuQnmzIDT0iDK0LxepAwnv4L8FgE7JD/qVZvMaVGUs0e2qnUsmbw7P4rKMcU0NtVFNLGYVRJaNSqZxxB/l0xfzcv0trmBVhw/YzjSlJUt77T2ZMZm0Y9KQJATd9EdVifzuZrjHUr1DPDRz3cfnBifTZTAIr/UcysQUGr6dAB6wEBRiQK0KoT/F2rhgIyJ2N0sqE0sUpKLGN1wcWkO+cbNER0rtyq2tapCREaZPsDgjZ4T+SLMxMraePnJodzfpDB1mLC3sBmYP66gx0Ay9iTo+2CwnueAFNjsq+fEtTTKu/WAGEEl/fVBvLDZnaNstXqxpBskHK6RKc+y+1TlXXfmWC38grI1oGC7VvzY+TJRGVBvtxU0YJ17IwWdtHgbjqCAx2btanfTWvigsObzRztA10ifWVLQwHKW5GYCn3YGjVzhwFwQJp7jj1/hmwCWwFM4Ijivolc5kPsWVUylbB6TUjIg4Y7zZO6ZryMXKNIJhbA7t6Vaej/E+0cfTFn6ACzFGnqqXxu62GLyte8NBGCEgvh4Bxhc+DoenauWKlCNktSR+i1M+mjg3pkgDcR0J5QiLNHvKhRblh5NGdHapYOphUDbWcwqdJ07VOwo9CHZ1WTox4VuVVSVz84Siag9ibuf18Z7XKd56MortTH7WAATHE6LB6RY8OG5+n1ZY9TWiHE3Mo6vKbHRKcEkbB2k/S5L28CL1a4C4yuAy5nNERL/+IRFhRpEUSH6GFQ4uIWcOcD62Zc'
                 ),
             ),
             PartStartEvent(
@@ -1906,8 +1905,7 @@ async def test_xai_builtin_web_search_tool_stream(allow_model_requests: None, xa
             PartDeltaEvent(
                 index=0,
                 delta=ThinkingPartDelta(
-                    signature_delta='DCSdJov0KqNTAiL+A6ioyn70PGRCLsIt3PRq4htAiEqlpTrAZvYxgY3Z22z69UNf30XmThQ6i8OzHILTc5t260s0YS8mrTBzlwtf60lp1o+5SE3BrAtT/iA9DMmMUfgvSe75iVaqrk54sGmxmpbUKzvnpcqBU01Nl1l3l+lssigniKZgD0VB7W/7fGSasp/ysO9BVAgrVTfn8aDGMYh7FOH8ItJCW5AdzPERnITXiL8YmiaeieqdlZBGCLg2datmkj4IldOyhIjF4AAfv+0p8Lv1vcWVAEv35ZI1PF7NMDMyxmyANUBDS+6ZanmMMeQB4hfFFf86d5cQUIF6VItRf4uahuDnmczDMo4W7Ho2xCFdPU8AEKOMndXA8yNeq8pwX3VRguYPzKCTDgaCIn3zBX+YWIfdXujB87L6rZ04FqlLoN1BPtoC+hal6O4OsyfZj3NVh6/P2nwJlgi7ntop4j/S7FxnttWDCtxWxSKMnrBrAO4V+fDaitEtokkxAnID8sPqdWXqN4vk49ZuBufUAG62ASqg88sfZq9up6afYkfONwnhRgv8kqmpqoSDABG79ZRLAvb/ipDrDkSjkfGd/jB6dGQAesTUGyzVLLC5v/NAkiLxVQQP9ADTymxSdJ/MlmScf6xlEIH1RhVsR2XdAst0aJENkWjtH5HjBJIemghkd4LQeIX9JFEd6XWqR5mjA9wMKHKAez7P/uQgD4SU4Yq1HFGHpync4NAOwD1/dLlNp1/qrrEUhGBMXM6uZokb2PYxCBVK4zPRinHfb+DnIvxjFQ6aSAtD88LZDeTpQYgGgflq9o8seGYnMGiLyv6faHyz4TUtmKE0X5T0PtS2iNqGDKn4xPqVxPZc5ErRm2JglnUs6XVkFAo',
-                    provider_name='xai',
+                    signature_delta='DCSdJov0KqNTAiL+A6ioyn70PGRCLsIt3PRq4htAiEqlpTrAZvYxgY3Z22z69UNf30XmThQ6i8OzHILTc5t260s0YS8mrTBzlwtf60lp1o+5SE3BrAtT/iA9DMmMUfgvSe75iVaqrk54sGmxmpbUKzvnpcqBU01Nl1l3l+lssigniKZgD0VB7W/7fGSasp/ysO9BVAgrVTfn8aDGMYh7FOH8ItJCW5AdzPERnITXiL8YmiaeieqdlZBGCLg2datmkj4IldOyhIjF4AAfv+0p8Lv1vcWVAEv35ZI1PF7NMDMyxmyANUBDS+6ZanmMMeQB4hfFFf86d5cQUIF6VItRf4uahuDnmczDMo4W7Ho2xCFdPU8AEKOMndXA8yNeq8pwX3VRguYPzKCTDgaCIn3zBX+YWIfdXujB87L6rZ04FqlLoN1BPtoC+hal6O4OsyfZj3NVh6/P2nwJlgi7ntop4j/S7FxnttWDCtxWxSKMnrBrAO4V+fDaitEtokkxAnID8sPqdWXqN4vk49ZuBufUAG62ASqg88sfZq9up6afYkfONwnhRgv8kqmpqoSDABG79ZRLAvb/ipDrDkSjkfGd/jB6dGQAesTUGyzVLLC5v/NAkiLxVQQP9ADTymxSdJ/MlmScf6xlEIH1RhVsR2XdAst0aJENkWjtH5HjBJIemghkd4LQeIX9JFEd6XWqR5mjA9wMKHKAez7P/uQgD4SU4Yq1HFGHpync4NAOwD1/dLlNp1/qrrEUhGBMXM6uZokb2PYxCBVK4zPRinHfb+DnIvxjFQ6aSAtD88LZDeTpQYgGgflq9o8seGYnMGiLyv6faHyz4TUtmKE0X5T0PtS2iNqGDKn4xPqVxPZc5ErRm2JglnUs6XVkFAo'
                 ),
             ),
             PartStartEvent(index=5, part=TextPart(content='Today'), previous_part_kind='builtin-tool-return'),
@@ -2757,10 +2755,7 @@ View this search on DeepWiki: https://deepwiki.com/search/provide-a-short-summar
             ),
             PartDeltaEvent(
                 index=0,
-                delta=ThinkingPartDelta(
-                    signature_delta=IsStr(),
-                    provider_name='xai',
-                ),
+                delta=ThinkingPartDelta(signature_delta=IsStr()),
             ),
             PartStartEvent(index=3, part=TextPart(content='P'), previous_part_kind='builtin-tool-return'),
             FinalResultEvent(tool_name=None, tool_call_id=None),


pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_event_stream.py
```diff
@@ -124,7 +124,17 @@ async def handle_thinking_delta(self, delta: ThinkingPartDelta) -> AsyncIterator
...
+        if part.signature is not None:
+            metadata: dict[str, Any] = {'signature': part.signature}
+            if part.provider_name is not None:
+                metadata['provider_name'] = part.provider_name
+            if part.provider_details is not None:
+                metadata['provider_details'] = part.provider_details
+            if part.id is not None:
+                metadata['id'] = part.id
+            provider_metadata = {'pydantic_ai': metadata}
+        yield ReasoningEndChunk(id=self.message_id, provider_metadata=provider_metadata)
```
DouweM, line 177: - I don't think just emitting this would be sufficient -- we should also read these values in `VercelAIAdapter.load_messages` and restore them on the newly build parts.
- We should also do this in `VercelAIAdapter.dump_messages` and ensure that they survive a roundtrip verbatim
- We should also do this on _all other parts_ that have `provider_details`, `provider_name`, and/or `id`, because e.g. Google will store `provider_details['thought_signature']` on there and also require it to be sent back verbatim.

So there's a little bit more work to do here to really eliminate this type of error :)

We'll likely want a helper method so that we don't repeat these same lines each time

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_event_stream.py
```diff
@@ -124,7 +124,17 @@ async def handle_thinking_delta(self, delta: ThinkingPartDelta) -> AsyncIterator
     async def handle_thinking_end(
         self, part: ThinkingPart, followed_by_thinking: bool = False
     ) -> AsyncIterator[BaseChunk]:
-        yield ReasoningEndChunk(id=self.message_id)
+        provider_metadata: dict[str, dict[str, Any]] | None = None
+        if part.signature is not None:
```
DouweM: I think we should store the other fields even if there is no `signature`

tests/test_vercel_ai.py
```diff
@@ -286,7 +286,17 @@ async def test_run(allow_model_requests: None, openai_api_key: str):
...
...TI4Tgj95q0SfkCzy_DY-wD6IIHjEyL4D7JO0Y7U_UADJN3ZAoKHWlZl3OSO7nw_uskVM5FB7jqqDfr5P0YBMQMwWjVuAibgMnt-9rO4NPBG1xifL1GwcBvFW30cgxE2SGFrOcPqyc8HS6vc55DgUS4kpPOkN8sTOP8wBT2_jEKzRde3ByBDa7Dyx1hAuet_huuNmYU6mGRcmKCqWc6qW6FJH4sjLSvUW9UoRDMn3ZpVBsHVtdiQAB8RlegtJNyr61gexss0Qla4InT65EbkB1carYnCdeQGyyMUKk-YD79CapD_BtRnMJ5twCFampVIiQn_tANwdMEX9NFGgQgnczJ3MzbrEHHA3iBhapWxWD65FEm2zYrDUaeMNm8yH60RuNKyZ7aZRo-sGKrYWMU6bZ4YwuRbqiTNF1NPizcpGrj_2k_hT-TcbtPnPNAMdXzTK9EB2QHfsQqmsqOImtx8q0vekrqCSWjN4H1xRHrYPZMvlvTdtDdhiWI8WSatsPludVXeYxqXxOy2XulPjzkiVtpgvMUwTGH_Yh-kUP5C94ZuZ3WDM93Gy9AsDdaNScSC_m0Ho4inuGpe5b_ctx7MNCdhdzl3PXwl67Fj_0E3qkdmDS6McKk0S2fOmzbvoJmaL7g2g_QVCCpxA-4vmmecak3J5Me_2EW6sHd5Zn1tvcFezrcmQfcZQbax7yoyZk20egkaQqpkli-AKjSUKBmP0aSR7qrkvAe2z4mgl4yP062GlC2YI3r_8xedjsg3jWRHyyJ4D_7Y0rtmAjVwhi1tfl0XK0JCPKVkwTTCBSbwVQo2yR_zc7zIazj9TDm8qFK2G_Q3QuEqz-Cz9VGQu_BPhrMaH1WaB-2ays6IUf3s7_gUavAAynaEM9kGGHj_t8zkyqDTjalRl0JE5T5PxxYdv1G9s3MU4B0yB3H_LhcDyjVAb1WzPZpcAxxrcBdpLVFa36ewDzKRmVmSONJd3fLtB9_HJhXgtFIIRgAKuhmRf1hMAG6qaJSsJNA3HYRZigy15OMe4RgkFSl1aj3ScrJc1-N3Si44OwqRIJGrVV2ihh_euAE0l_ImdW_vEgpRbgfMqBiDmFbxgdcv2cT_MaOMoFX3n_V1GrAuBzjrVQSBbhgau7nWsynSK4h9781waWYMAZK4LrOK0UQ4iyugcpj0JNO9k4WvB-yRR9E2rtIoEC6Nz0=',
```
DouweM: Please replace these signatures with `IsStr()` to make the test files easier to read

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_event_stream.py
Light2Dark: Thanks for the review! Fortunately, the adapter seems to do this for a [lot of parts already](https://github.com/pydantic/pydantic-ai/blob/a6e55ca25504d833dd9443c62454b3126ab529b1/pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py#L140) (not all) hence why it worked locally, I can do a pass to do it for more 🤔. I see that a robust solution involves pushing needed data through provider_metadata, and then pulling it in the adapter.

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
```diff
@@ -185,9 +217,9 @@ def load_messages(cls, messages: Sequence[UIMessage]) -> list[ModelMessage]:  #
                                 else:
                                     output = {'error_text': part.error_text, 'is_error': True}
 
-                                provider_name = (
-                                    (part.call_provider_metadata or {}).get('pydantic_ai', {}).get('provider_name')
-                                )
+                                pydantic_ai_meta = cls._get_pydantic_ai_meta(part.call_provider_metadata)
+                                provider_name = pydantic_ai_meta.get('provider_name')
+                                provider_details = pydantic_ai_meta.get('provider_details')
                                 call_part.provider_name = provider_name
```
DouweM: This is duplicated with what we're doing above right

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
```diff
@@ -196,10 +228,25 @@ def load_messages(cls, messages: Sequence[UIMessage]) -> list[ModelMessage]:  #
                                         tool_call_id=tool_call_id,
                                         content=output,
                                         provider_name=provider_name,
+                                        provider_details=provider_details,
```
DouweM: Hmm, in Pydantic AI we can have separate provider metadata on the call and return parts, but Vercel represents them as a single part. So we need to make sure they don't accidentally get merged, somehow, so that all the fields make it back to the correct part.

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_event_stream.py
```diff
@@ -193,3 +220,9 @@ def _tool_return_output(self, part: BaseToolReturnPart) -> Any:
         output = part.model_response_object()
         # Unwrap the return value from the output dictionary if it exists
         return output.get('return_value', output)
+
+
+def form_provider_metadata(**kwargs: ProviderDetailsDelta | str) -> ProviderMetadata | None:
```
DouweM: Can we make this a method on `VercelAIAdapter` like `_get_pydantic_ai_meta`? To match the load/dump_messages methods, I'd call them `_load_part_metadata` and `_dump_part_metadata`

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_event_stream.py
Light2Dark: I created those 2 methods under VercelAIAdapter, `_get_pydantic_ai_meta -> _load_part_metadata` and `_dump_part_metadata`.

However, does it mean we should remove this `form_provider_metadata` function? I think it'd be good to reuse the same function to standardize. If so, I need to make `_dump_part_metadata` function public and use it in `_event_stream.py`

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
```diff
@@ -277,33 +326,39 @@ def _dump_response_message(  # noqa: C901
...
+                call_provider_metadata = VercelAIAdapter._dump_part_metadata(
+                    id=part.id, provider_name=part.provider_name, provider_details=part.provider_details
                 )
 
                 if builtin_return := local_builtin_returns.get(part.tool_call_id):
+                    builtin_return_meta = VercelAIAdapter._dump_part_metadata(
+                        id=part.id,
+                        provider_name=builtin_return.provider_name,
+                        provider_details=builtin_return.provider_details,
+                    )
```
Light2Dark: I made this change because I think it's more accurate, but it does differ from the old behaviour. Can refer to [9748bbb](https://github.com/pydantic/pydantic-ai/pull/3754/commits/9748bbbb69fad310c96fb480f6080d2f909141ae) for the change on the test too.

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
Light2Dark: Not sure I'm getting it right but I did move getting this details to the top, so both call and return parts should be the same.

Light2Dark: I might be confused, I can also remove this..

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
```diff
@@ -185,9 +218,6 @@ def load_messages(cls, messages: Sequence[UIMessage]) -> list[ModelMessage]:  #
                                 else:
                                     output = {'error_text': part.error_text, 'is_error': True}
 
-                                provider_name = (
-                                    (part.call_provider_metadata or {}).get('pydantic_ai', {}).get('provider_name')
-                                )
                                 call_part.provider_name = provider_name
```
DouweM: I don't think we need this line anymore, as we already set this on the `call_part` above

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
DouweM: The problem is that in Pydantic AI, `(Builtin)ToolCallPart` and `(Builtin)ToolReturnPart` both have their _own_ `provider_details`, but Vercel AI combines both into one part (`ToolOutputAvailablePart` etc).

We need to make sure that when we translate the 2 Pydantic AI parts into 1 Vercel AI part and back into 2 Pydantic AI parts, they end up with the original `provider_details`, _not_ all details on one of the 2 parts, or both parts duplcating the same details. So I think in this case we should store 2 dicts on the Vercel AI part, so that we can separately store and extract the details for the call and those for the result

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
```diff
@@ -196,10 +226,19 @@ def load_messages(cls, messages: Sequence[UIMessage]) -> list[ModelMessage]:  #
...
                                 )
                         else:
-                            builder.add(ToolCallPart(tool_name=tool_name, tool_call_id=tool_call_id, args=args))
+                            builder.add(
+                                ToolCallPart(
+                                    tool_name=tool_name,
+                                    tool_call_id=tool_call_id,
+                                    args=args,
+                                    id=part_id,
+                                    provider_details=provider_details,
```
DouweM, line 250: Hmm `ToolCallPart` doesn't have `provider_name`, but in some places we only use `ToolCallPart.provider_details` if `ModelResponse.provider_name` has the expected value:

https://github.com/pydantic/pydantic-ai/blob/b5dee9b6fb0805428d7ecf5a70f76bd451361c65/pydantic_ai_slim/pydantic_ai/models/google.py#L914-L918

So to make this work properly with Google though signatures, we need to make sure the `provider_name` is also stored on the `ModelResponse` we build here from the Vercel AI data. Can you see how we would get that from Pydantic AI to Vercel AI format and back on the `ModelResponse`?

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
```diff
@@ -277,33 +326,39 @@ def _dump_response_message(  # noqa: C901
                 if ui_parts and isinstance(ui_parts[-1], TextUIPart):
                     ui_parts[-1].text += part.content
                 else:
-                    ui_parts.append(TextUIPart(text=part.content, state='done'))
+                    provider_metadata = VercelAIAdapter._dump_part_metadata(
```
DouweM: If we make this a `classmethod`, we can use `cls` here like we do above

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
DouweM: See above; we should store the data of both the call _and_ the return, we can't lose any piece of data (like the `'tool_type': 'web_search_preview'`)

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_event_stream.py
DouweM: Yeah I'd rather use the same `_dump_part_metadata` here. I'm OK with the event stream using private methods from the adapter; I'd rather not have this be user-facing public

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_event_stream.py
```diff
@@ -187,7 +214,7 @@ async def handle_builtin_tool_call_end(self, part: BuiltinToolCallPart) -> Async
             tool_name=part.tool_name,
             input=part.args_as_dict(),
             provider_executed=True,
-            provider_metadata={'pydantic_ai': {'provider_name': part.provider_name}},
+            provider_metadata=form_provider_metadata(provider_name=part.provider_name),
```
DouweM: We also need the provider details here right?

We should have exactly the same behavior here as in `dump_messages`

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_event_stream.py
Light2Dark: Thanks for the thorough review, I'll try to get it all right.

> Yeah I'd rather use the same _dump_part_metadata here. I'm OK with the event stream using private methods from the adapter; I'd rather not have this be user-facing public

Just realized this would cause a circular import (adapter imports `VercelAIEventStream`). I could move it to `_utils.py`.

Could also keep this function in _event_stream.py. Idt it's user facing since not exported in `__all__` declaration.

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
Light2Dark: I don't think we are passing this data in `_message_builder` to convert to pydantic messages.
```python
@dataclass
class MessagesBuilder:
    """Helper class to build Pydantic AI messages from request/response parts."""

    messages: list[ModelMessage] = field(default_factory=list)

    def add(self, part: ModelRequestPart | ModelResponsePart) -> None:
        """Add a new part, creating a new request or response message if necessary."""
        last_message = self.messages[-1] if self.messages else None
        if isinstance(part, get_union_args(ModelRequestPart)):
            part = cast(ModelRequestPart, part)
            if isinstance(last_message, ModelRequest):
                last_message.parts = [*last_message.parts, part]
            else:
                self.messages.append(ModelRequest(parts=[part])) ## <-- We don't pass additional data here
        else:
            part = cast(ModelResponsePart, part)
            if isinstance(last_message, ModelResponse):
                last_message.parts = [*last_message.parts, part]
            else:
                self.messages.append(ModelResponse(parts=[part])) ## <-- And here

```
Could infer details from the part
```python
    def add(self, part: ModelRequestPart | ModelResponsePart) -> None:
        """Add a new part, creating a new request or response message if necessary."""
        ...
        else:
            part = cast(ModelResponsePart, part)
            if isinstance(last_message, ModelResponse):
                last_message.parts = [*last_message.parts, part]
            else:
                self.messages.append(ModelResponse(parts=[part], provider_name=part.provider_name)) ## <-- here
```
and build the metadata as parts come in.

Another option is to modify the `add` function to accept more arguments and pass it to ModelResponse 🤔. Anyway, I'd prefer this to be another PR

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_event_stream.py
Light2Dark: I've decided to move both functions to `_utils.py`, and move around some code to `_models.py` to prevent circular import errors.

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
DouweM: I think it's worth fixing this in this PR, because the goal here is to make encrypted thinking work reliably when data goes from Pydantic AI to Vercel AI and back to Pydantic AI, which means making all fields that matter to thinking survive that round-trip, which includes this ones :(

One way to do that would be to see if we can add extra fields to `add`, making sure it works even if there's no part on `ModelResponse.parts` that actually has the `provider_name`, so it will need to be read off `ModelResponse.provider_name` itself.

But it may be better to add the `provider_name` field everywhere we have currently have `provider_details` (`ToolCallPart`, `TextPart` etc), and then making sure it's always set when where set `provider_details`. That may touch more code but is a lot more clean, so please give that a try.

Light2Dark: @DouweM , I assume you mean modifying the part like
```python
class ToolCallPart(BaseToolCallPart):
    """A tool call from a model."""

    _: KW_ONLY

    provider_name: str | None = None.  <---- NEW
    """The name of the provider that generated the response."""
    ...
```
I can make the changes for _event_stream.py, _adapter.py and for Vercel AI related changes. The surface area is not large. But doing the wiring for all the providers is imo going to be very large.

Example for the issue you mentioned
> ToolCallPart doesn't have provider_name, but in some places we only use ToolCallPart.provider_details

I'm guessing we would modify `google.py` like this or something similar:
```python
        if (
            item.provider_details
            and (thought_signature := item.provider_details.get('thought_signature'))
            and (m.provider_name == provider_name or item.provider_name == provider_name) ## <-- check item provider
        ):
```

DouweM: @Light2Dark Yep, adding `provider_name` to `ToolCallPart` and `TextPart` and any others that currently have `provider_details`. 

> But doing the wiring for all the providers is imo going to be very large.

I believe it's just one instance in `OpenAIResponsesModel`:
- https://github.com/pydantic/pydantic-ai/blob/fbe0dfe4f4b4dd14eb786e8651675316beeef71e/pydantic_ai_slim/pydantic_ai/models/openai.py#L1290

And a bunch in `GoogleModel`: (some of which may require changes to methods like `handle_tool_call_delta`, but that should be straightforward):

- https://github.com/pydantic/pydantic-ai/blob/fbe0dfe4f4b4dd14eb786e8651675316beeef71e/pydantic_ai_slim/pydantic_ai/models/google.py#L858
- https://github.com/pydantic/pydantic-ai/blob/fbe0dfe4f4b4dd14eb786e8651675316beeef71e/pydantic_ai_slim/pydantic_ai/models/google.py#L842
- https://github.com/pydantic/pydantic-ai/blob/fbe0dfe4f4b4dd14eb786e8651675316beeef71e/pydantic_ai_slim/pydantic_ai/models/google.py#L827-L834
- https://github.com/pydantic/pydantic-ai/blob/fbe0dfe4f4b4dd14eb786e8651675316beeef71e/pydantic_ai_slim/pydantic_ai/models/google.py#L1101

Really appreciate you working on this, we're almost there!

pydantic_ai_slim/pydantic_ai/models/google.py
```diff
@@ -1029,6 +1040,59 @@ def _content_model_response(m: ModelResponse, provider_name: str) -> ContentDict
     return ContentDict(role='model', parts=parts)
 
 
+def _process_part(
```
Light2Dark, line 1067: separated this out because the function was too long & raising lint errors

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
```diff
@@ -277,49 +342,87 @@ def _dump_response_message(  # noqa: C901
...
+                            ToolOutputAvailablePart(
+                                type=tool_name,
+                                tool_call_id=part.tool_call_id,
+                                input=part.args_as_json_str(),
+                                output=content,
+                                state='output-available',
+                                provider_executed=True,
+                                call_provider_metadata=combined_provider_meta,
+                            )
+                        )
```
Light2Dark, line 420: I think the error'ed output was missed on the first pass of this, this is a new change.

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
Light2Dark: Made fixes, ready for review!

Light2Dark:
> This PR is stale, and will be closed in 3 days if no reply is received.

keeping this alive

Light2Dark:
hi @DouweM , this is ready for review, thanks!

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -1032,6 +1032,9 @@ class TextPart:
     id: str | None = None
     """An optional identifier of the text part."""
 
+    provider_name: str | None = None
+    """The name of the provider that generated the response."""
```
DouweM: Let's specify that this is only required to be set when `provider_details` is.

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -1148,6 +1151,12 @@ class BaseToolCallPart:
 
     This is used by some APIs like OpenAI Responses."""
 
+    provider_name: str | None = None
+    """The name of the provider that generated the response.
+
+    Tool calls are only sent back to the same provider.
```
DouweM: This is only the case for built-in tool calls.

And as above (and in all the other optional `provider_name` fields), we should specify this is only required when `provider_details` is set.

pydantic_ai_slim/pydantic_ai/models/google.py
```diff
@@ -1032,6 +1043,59 @@ def _content_model_response(m: ModelResponse, provider_name: str) -> ContentDict
...
+        mime_type = inline_data.mime_type
+        assert data and mime_type, 'Inline data must have data and mime type'
+        content = BinaryContent(data=data, media_type=mime_type)
+        item = FilePart(content=BinaryContent.narrow_type(content))
+    else:  # pragma: no cover
+        raise UnexpectedModelBehavior(f'Unsupported response from Gemini: {part!r}')
+
+    if provider_details:
+        item.provider_details = {**(item.provider_details or {}), **provider_details}
+        if provider_name:
```
DouweM: provider_name is always set right?

pydantic_ai_slim/pydantic_ai/models/google.py
```diff
@@ -1032,6 +1043,59 @@ def _content_model_response(m: ModelResponse, provider_name: str) -> ContentDict
...
+            item.provider_name = item.provider_name or provider_name
```
DouweM: I think we can just override it without the `or`, since we know it won't be set yet (and if it is, it would be to the same value!)

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
```diff
@@ -175,31 +195,63 @@ def load_messages(cls, messages: Sequence[UIMessage]) -> list[ModelMessage]:  #
...
+                            # For builtin tools, we need to create 2 parts (BuiltinToolCall & BuiltinToolReturn) for a single Vercel ToolOutput
+                            # The call and return metadata are combined in the output part.
+                            # So we extract and return them to the respective parts
+                            call_meta = return_meta = {}
+                            has_tool_output = isinstance(part, (ToolOutputAvailablePart, ToolOutputErrorPart))
+
+                            if has_tool_output:
+                                loaded_call_meta, loaded_return_meta = cls._load_builtin_tool_meta(provider_meta)
+                                call_meta = loaded_call_meta or {}
+                                return_meta = loaded_return_meta or {}
```
DouweM: let's move the `or {}` to `_load_builtin_tool_meta` so we can remove these extra variables

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
DouweM: @Light2Dark THanks! One more thing to do here:
> in some places we only use `ToolCallPart.provider_details` if `ModelResponse.provider_name` has the expected value:
>
> https://github.com/pydantic/pydantic-ai/blob/b5dee9b6fb0805428d7ecf5a70f76bd451361c65/pydantic_ai_slim/pydantic_ai/models/google.py#L914-L918

We need to look for those patterns where we compare against `ModelResponse.provider_name`, and make it `(m.provider_name == provider_name or item.provider_name == provider_name)`

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
```diff
@@ -277,49 +342,87 @@ def _dump_response_message(  # noqa: C901
...
                     )
+                    return_meta = dump_provider_metadata(
+                        wrapper_key=None,
+                        provider_name=builtin_return.provider_name,
+                        provider_details=builtin_return.provider_details,
+                    )
+                    combined_provider_meta = cls._dump_builtin_tool_meta(call_meta, return_meta)
+
+                    response_object = builtin_return.model_response_object()
+                    is_error = response_object.get('is_error') is True and 'error_text' in response_object
```
DouweM: Let's add a comment that this is specific to `BuiltinToolReturnPart`s that themselves came from VercelAI, as we can't generally rely on any consistent fields for errors existing until we have https://github.com/pydantic/pydantic-ai/issues/3561.

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
```diff
@@ -277,49 +342,87 @@ def _dump_response_message(  # noqa: C901
...
+                    if is_error:
+                        ui_parts.append(
+                            ToolOutputErrorPart(
+                                type=tool_name,
+                                tool_call_id=part.tool_call_id,
+                                input=part.args_as_json_str(),
+                                error_text=response_object.get('error_text', ''),
```
DouweM: this can be merged into the if using `and (error_text := response_object.get('error_text')`

tests/models/test_google.py
```diff
@@ -5307,3 +5324,22 @@ async def test_google_system_prompts_and_instructions_ordering(google_provider:
         }
     )
     assert contents == snapshot([{'role': 'user', 'parts': [{'text': 'Hello'}]}])
+
+
+def test_process_part_with_provider_details_but_empty_provider_name():
+    """Test that _process_part handles provider_details without provider_name.
+
+    This tests the branch where thought_signature creates provider_details,
+    but provider_name is empty string (falsy), so provider_name should not be set on the item.
```
DouweM: Why do we need this scenario? Why would provider name be an empty string?

tests/test_messages.py
```diff
@@ -678,6 +680,142 @@ def test_binary_content_validation_with_optional_identifier():
     )
 
 
+def test_tool_call_part_delta_apply_to_tool_call_part_delta():
```
DouweM: I think the other fields of `ToolCallPartDelta` are already tested in `test_parts_manager`; let's update those tests to cover the new `provider_name` as well.

tests/test_vercel_ai.py
```diff
@@ -445,16 +586,41 @@ async def test_run(allow_model_requests: None, openai_api_key: str):
...
+                        'signature': IsStr(),
+                        'provider_name': 'openai',
+                        'id': 'rs_00e767404995b9950068e648130f0481918dc71103fbd6a486',
+                    }
+                },
+            },
+            {
+                'type': 'text-start',
+                'id': IsStr(),
+                'providerMetadata': {'pydantic_ai': {'id': 'msg_00e767404995b9950068e6482f25e0819181582a15cdd9207f'}},
```
DouweM: This one should also have a `provider_name`! Will need updating in openai.py.

tests/test_vercel_ai.py
```diff
@@ -1810,7 +2104,12 @@ async def test_adapter_load_messages():
                         input={'query': 'What is Logfire?'},
                         output="[Scrubbed due to 'Auth']",
                         provider_executed=True,
-                        call_provider_metadata={'pydantic_ai': {'provider_name': 'openai'}},
+                        call_provider_metadata={
+                            'pydantic_ai': {
+                                'call_meta': {'provider_name': 'openai'},
+                                'return_meta': {'provider_name': 'openai_return'},
```
DouweM, line 2088: Should be `openai`, not `openai_return`?

tests/test_vercel_ai.py
```diff
@@ -2884,3 +3195,1108 @@ async def test_adapter_dump_messages_with_cache_point():
...
+                RetryPromptPart(
+                    content='Tool execution failed',
+                    tool_name='failing_tool',
+                    tool_call_id='tc_fail',
+                )
+            ]
+        ),
+    ]
+
+    ui_messages = VercelAIAdapter.dump_messages(messages)
```
DouweM, line 3882: I think all these tests that test `dump_messages` should test the full roundtrip as well: `assert VercelAIAdapter.load_messages(ui_messages) == messages`

Since the goal is to make the conversion lossless.

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -2355,6 +2362,7 @@ async def _get_event_iterator(self) -> AsyncIterator[ModelResponseStreamEvent]:
                 for event in self._parts_manager.handle_thinking_delta(
                     vendor_part_id=chunk.item_id,
                     id=chunk.item_id,
+                    provider_name=self.provider_name,
```
DouweM, line 2442: On `ThinkingPart`s, I think we should also set `provider_name` if only `id` is set (not just when signature or provider_details is set), because that's also provider-specific information. So I think we have a few more `handle_thinking_delta`/`ThinkingPart` spots to update, including the one a few lines above.

Also the non-streaming version of this one.

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -1247,7 +1247,14 @@ def _process_response(  # noqa: C901
                         part_provider_details: dict[str, Any] | None = None
                         if content.logprobs:
                             part_provider_details = {'logprobs': _map_logprobs(content.logprobs)}
-                        items.append(TextPart(content.text, id=item.id, provider_details=part_provider_details))
+                        items.append(
+                            TextPart(
+                                content.text,
+                                id=item.id,
+                                provider_name=self.system if part_provider_details else None,
```
DouweM: This should also be added to the streaming version of this, right?

pydantic_ai_slim/pydantic_ai/models/openai.py
DouweM: Basically every check in a model that currently compares a against model.provider_name, should now (also) check  against part.provider_name, and we should make sure that when any of the fields are set that are only used when the name matches, we also set the provider_name.

In this case the relevant check is here:

https://github.com/pydantic/pydantic-ai/blob/9aa6dd40efafd93c04c19c2ef5596a454906ca53/pydantic_ai_slim/pydantic_ai/models/openai.py#L1663

So everywhere we set `id` on a part, we have to set `provider_name` as well.

tests/test_vercel_ai.py
Light2Dark: I intentionally set this to something different so the test will fail if return_meta is not `openai_return`. In practice, it's probably the same.

tests/test_vercel_ai.py
Light2Dark: sorry, why should this have provider_name? I thought we wanted to add provider_name for **thinking parts** with `id`. This doesn't provider_details nor is a thinking part

Light2Dark: Can I make a change where we always set provider_name? I think setting it conditionally seems more confusing imo

DouweM: @Light2Dark Any part that has provider-specific values on it, including `id`, should have the name of the provider its specific to, so that we only send the value back to that same provider. 

I'd prefer not to set it unless we have one of those provider-specific values (`id`, `signature`, `provider_details`), because it could also be confusing to have a `provider_name` on a simple `TextPart` with nothing provider-specific going on, but if that simplifies things significantly I'm OK with setting it in a few cases where we don't actually need it. But I wouldn't want it to be on _every_ `TextPart` just because _some_ `TextPart`s have an `id`.

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -1320,14 +1320,18 @@ def _process_response(  # noqa: C901
...
+                            )
+                        )
             elif isinstance(item, responses.ResponseFunctionToolCall):
                 items.append(
                     ToolCallPart(
-                        item.name,
-                        item.arguments,
-                        tool_call_id=item.call_id,
-                        id=item.id,
+                        item.name, item.arguments, tool_call_id=item.call_id, id=item.id, provider_name=self.system
```
Light2Dark, line 1346: i do wonder if we should add provider_details to all these parts now.

tests/models/test_google.py
```diff
@@ -5209,6 +5237,43 @@ def test_google_thought_signature_on_thinking_part():
     )
     assert old_google_response == new_google_response
 
+    # Test that thought_signature is used when item.provider_name matches even if ModelResponse.provider_name doesn't
+    # This happens when responses from other providers contain parts that originated from Google
```
DouweM: This isn't right: this is for the Vercel AI scenario were implementing here, where fields on the model response object don't survive a roundtrip because Vercel doesn't have message metadata, only part metadata.

tests/models/test_openai_responses.py
```diff
@@ -3421,286 +3686,930 @@ async def test_openai_responses_thinking_with_code_execution_tool_stream(
...
+                index=7,
+                delta=TextPartDelta(content_delta='140', provider_name='openai'),
+            ),
+            PartDeltaEvent(
+                index=7,
+                delta=TextPartDelta(content_delta='927', provider_name='openai'),
+            ),
+            PartDeltaEvent(
+                index=7,
+                delta=TextPartDelta(content_delta='865', provider_name='openai'),
```
DouweM: Looking through the snapshots in the tests, most of the extra `provider_name` fields make sense, but this one I think we should be able to could avoid as it was already on the initial `PartStartEvent(TextPart)`.

tests/test_vercel_ai.py
```diff
@@ -445,1731 +586,3230 @@ async def test_run(allow_model_requests: None, openai_api_key: str):
...
             {
                 'type': 'text-delta',
                 'delta': """\
+ limits).
 
-from opente\
+How\
 """,
                 'id': IsStr(),
+                'providerMetadata': {'pydantic_ai': {'provider_name': 'openai'}},
```
DouweM: As in the comment on test_openai_responses.py, I don't think we should need the provider name on every delta -- I assume that if it's on the initial text-start chunk, we're good

tests/test_vercel_ai.py
```diff
@@ -445,1731 +586,3230 @@ async def test_run(allow_model_requests: None, openai_api_key: str):
...
+    assert events == snapshot(
+        [
+            {'type': 'start'},
+            {'type': 'start-step'},
+            {'type': 'reasoning-start', 'id': IsStr()},
+            {'type': 'reasoning-delta', 'id': IsStr(), 'delta': 'Let me think...'},
+            {
+                'type': 'reasoning-end',
+                'id': IsStr(),
+                'providerMetadata': {'pydantic_ai': {'signature': 'sig_abc123', 'provider_name': 'function'}},
```
DouweM, line 1289: I assume it's fine for the `providerMetadata` to only be on the end event, or only on the start event, or only on any delta event, assuming the part would just take the last metadata seen, but it would be worth verifying that.

tests/test_vercel_ai.py
DouweM: @Light2Dark Did you see this comment? The most important thing we have to ensure in this PR is that when messages go from Pydantic AI to Vercel AI and back to Pydantic AI, no data is lost, i.e. `assert VercelAIAdapter.load_messages(VercelAIAdapter.dump_messages(messages)).parts == messages.parts`. Can you make sure we verify that in all the new tests related to `provider_details`?

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -883,7 +883,7 @@ class BuiltinToolReturnPart(BaseToolReturnPart):
     _: KW_ONLY
 
     provider_name: str | None = None
-    """The name of the provider that generated the response."""
+    """The name of the provider that generated the response. Set this when provider_details exists."""
```
DouweM: It's a bit nitpicky, but these docstrings are starting to matter more as they act as  rules for agents as well:
- Make it "Required to be set when `provider_details` is set.", with 2 newlines preceding it, and the backticks because these are rendered as Markdown in the API docs
- In the cases below that also have an `id`, we should of course say `when `provider_details` or `id` is set.`
- On the provider-specific fields, add (after 2 newlines) "When this field is set, `provider_name` is required to identify the provider that generated this data." or something like that

pydantic_ai_slim/pydantic_ai/models/openai.py
DouweM: All parts that are generated by providers have the `provider_details` field now, right? But I don't think that in the case of the ToolCallPart there are any details to add.

pydantic_ai_slim/pydantic_ai/models/openai.py
DouweM: @Light2Dark I think the second comment here is still TBD

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -2430,7 +2438,10 @@ async def _get_event_iterator(self) -> AsyncIterator[ModelResponseStreamEvent]:
 
             elif isinstance(chunk, responses.ResponseTextDeltaEvent):
                 for event in self._parts_manager.handle_text_delta(
-                    vendor_part_id=chunk.item_id, content=chunk.delta, id=chunk.item_id
+                    vendor_part_id=chunk.item_id,
+                    content=chunk.delta,
+                    id=chunk.item_id,
+                    provider_name=self.provider_name,
```
DouweM, line 2459: Related to the comment I left on the tests, I think we don't need this here as it would've aleady been included on the start event

DouweM: Same for the thinking delta and possibly tool call args delta, as we included it on the start event already

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/models/openai.py
Light2Dark: By that logic, why are there provider_name attributes on the delta/end parts? 🤔 Since they should be included in the start events.

tests/models/test_google.py
Light2Dark: Hmm, I thought it was for this change that you requested https://github.com/pydantic/pydantic-ai/pull/3754#discussion_r2677486918

> We need to look for those patterns where we compare against ModelResponse.provider_name, and make it (m.provider_name == provider_name or item.provider_name == provider_name)

Light2Dark: maybe the comment is inaccurate to descibe why this could happen, i will udpate that

pydantic_ai_slim/pydantic_ai/models/openai.py
Light2Dark: Anyways, updated to remove provider_name on some parts. Updated this in the handle_text_delta function instead of here

 Some like tool call delta can sometimes be upgraded to a full part (more logic required to handle this).  Thinking deltas seem to already pass in provider_name, I'm afraid to remove this

tests/test_vercel_ai.py
Light2Dark: I did! missed this test, updated. looks like this
```python
    # Verify roundtrip
    reloaded_messages = VercelAIAdapter.load_messages(ui_messages)
    _sync_timestamps(messages, reloaded_messages)
    assert reloaded_messages == messages
```

pydantic_ai_slim/pydantic_ai/_parts_manager.py
```diff
@@ -149,9 +149,7 @@ def handle_text_delta(
         else:
             # Update the existing TextPart with the new content delta
             existing_text_part, part_index = existing_text_part_and_index
-            part_delta = TextPartDelta(
-                content_delta=content, provider_name=provider_name, provider_details=provider_details
-            )
+            part_delta = TextPartDelta(content_delta=content, provider_details=provider_details)
```
DouweM: I don't think this is right: it's possible for a streaming text part (or thinking part etc) to not immediately have provider-specific data and thus to be created without a `provider_name` initially, and then later provider-specific data is received on the stream (e.g. https://github.com/pydantic/pydantic-ai/pull/3787) and then we pass `provider_details` and `provider_name` to `handle_text_delta`. So in that case, the `PartStartEvent` should not have either field, some later `PartDeltaEvent` _will_, and then the final `PartEndEvent` will have it as well because it has the complete event.

So we should revert this change, and trust that if the user passed the provider name into this method, they meant for it to end up on the delta. It's when calling `handle_text_delta` that we should try not to pass the `provider_name` if we know it's unnecessary.

But I'd rather send it more than needed, than less than needed :) So if it's really hard to make it work they way I'd want it to ideally, then I'm OK with a few extra `provider_name`s.

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -1559,7 +1587,7 @@ class ThinkingPartDelta:
     """Optional provider name for the thinking part.
 
     Signatures are only sent back to the same provider.
-    Set this when provider_details or id exists as they are provider-specific information.
+    Required to be set when provider_details is set.
```
DouweM: These docstrings are rendered as Markdown in the API docs, so all code snippets (i.e. field names) should be wrapped in backticks.

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -1084,13 +1098,15 @@ class ThinkingPart:
     """The name of the provider that generated the response.
 
     Signatures are only sent back to the same provider.
-    Set this when provider_details or id exists as they are provider-specific information.
+    Required to be set when provider_details or id is set.
```
DouweM: Or signature right?

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -1508,7 +1536,7 @@ class TextPartDelta:
     _: KW_ONLY
 
     provider_name: str | None = None
-    """The name of the provider that generated the response. Set this when provider_details exists."""
+    """The name of the provider that generated the response."""
```
DouweM: This field and the next one need the extra line of text about when they are required to be set

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -1670,12 +1699,13 @@ class ToolCallPartDelta:
     non-matching value is provided an error will be raised."""
 
     provider_name: str | None = None
-    """The name of the provider that generated the response. Set this when provider_details exists."""
+    """The name of the provider that generated the response."""
```
DouweM: This needs the extra line of text as well

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_event_stream.py
```diff
@@ -127,17 +127,15 @@ async def handle_text_start(self, part: TextPart, follows_text: bool = False) ->
             yield TextDeltaChunk(id=message_id, delta=part.content, provider_metadata=provider_metadata)
 
     async def handle_text_delta(self, delta: TextPartDelta) -> AsyncIterator[BaseChunk]:
+        # Only provider_details is needed for deltas as provider name is sent on the start event
```
DouweM: As mentioned above, this is not guaranteed. I'd rather include all fields from the `TextPartDelta`, and then make it the responsibility of the caller to `handle_text_delta` to only include the minimum required fields.

But as I said, it's much better to include too much than to include too little so this is too risky.

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_event_stream.py
```diff
@@ -127,17 +127,15 @@ async def handle_text_start(self, part: TextPart, follows_text: bool = False) ->
...
         if delta.content_delta:  # pragma: no branch
-            provider_metadata = dump_provider_metadata(
-                provider_name=delta.provider_name, provider_details=delta.provider_details
-            )
+            provider_metadata = dump_provider_metadata(provider_details=delta.provider_details)
             yield TextDeltaChunk(id=self.message_id, delta=delta.content_delta, provider_metadata=provider_metadata)
 
     async def handle_text_end(self, part: TextPart, followed_by_text: bool = False) -> AsyncIterator[BaseChunk]:
+        # Only provider_details is needed for end events as provider name is sent on the start event
```
DouweM: I'm fine with the end event always including the complete metadata, as we don't know if end metadata is merged with earlier metadata, or overwrites it entirely.

tests/test_vercel_ai.py
```diff
@@ -4672,6 +3868,16 @@ async def test_adapter_dump_messages_tool_error_with_provider_metadata():
         ]
     )
 
+    # Verify roundtrip
```
DouweM, line 3922: Are there any other tests where we should be doing this?

pydantic_ai_slim/pydantic_ai/_parts_manager.py
DouweM: So I think we should figure out in which scenario we were passing a `provider_name` into `handle_text_delta` when we didn't need to because we were not pssing `provider_details` or `id`.

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_event_stream.py
Light2Dark: ive added a util funciton to check the existing part and set the provider_name.

pydantic_ai_slim/pydantic_ai/models/openai.py
Light2Dark: I've updated to include openai, I think only google and openai does this.

tests/test_vercel_ai.py
Light2Dark: I've added it for most new tests added (where we dump_messages)

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -1748,8 +1748,9 @@ async def _map_messages(  # noqa: C901
                 file_search_item: responses.ResponseFileSearchToolCallParam | None = None
                 code_interpreter_item: responses.ResponseCodeInterpreterToolCallParam | None = None
                 for item in message.parts:
+                    item_provider_matches = item.provider_name == self.system
                     if isinstance(item, TextPart):
-                        if item.id and send_item_ids:
+                        if item.id and (send_item_ids or item_provider_matches):
```
DouweM: `send_item_ids` is originally set to `model_settings.get('openai_send_reasoning_ids', profile.openai_supports_encrypted_reasoning_content)`; if that is _false_ we should not be sending the ID, even if the item provider matches. So this logic is now too permissive. I think we need to drop `send_item_ids = send_item_ids and message.provider_name == self.system` under `isinstance(message, ModelResponse)` above and then inside `for item in message.parts:` create a new var like `send_item_id = send_item_ids and (item.provider_name == self.system or (item.provider_name is None and message.provider_name == self.system))`. then we can use that variable in all the branches below.

pydantic_ai_slim/pydantic_ai/_parts_manager.py
```diff
@@ -453,11 +458,21 @@ def _handle_embedded_thinking_content(
...
 
     def _handle_embedded_thinking_end(self, vendor_part_id: VendorId) -> None:
         """Handle </think> tag - stop tracking so next delta creates new part."""
         self._stop_tracking_vendor_id(vendor_part_id)
+
+    def _resolve_provider_name(
+        self, existing_part: ModelResponsePart | ToolCallPartDelta, provider_name: str | None
+    ) -> str | None:
+        """Return the provider name if it has not been set on previous parts."""
+        if existing_part.provider_name is None or provider_name != existing_part.provider_name:
```
DouweM, line 476: We couldn't address this by "just" not passing `provider_name` to `handle_text_delta` unless necessary (i.e. when we're also passing `id` or `provider_details`)? I don't mind this optimization, just unsure if it's actually necessary or if it would be cleaner to handle this at the call site.

tests/test_vercel_ai.py
```diff
@@ -2716,6 +2726,21 @@ async def test_adapter_dump_messages_with_retry_no_tool_name():
...
+    # When loaded back, it creates a UserPromptPart instead of RetryPromptPart.
+    reloaded_messages = VercelAIAdapter.load_messages(ui_messages)
+    assert len(reloaded_messages) == 3
+    _sync_timestamps(messages, reloaded_messages)
+    # First two messages are unchanged
+    assert reloaded_messages[:2] == messages[:2]
+    # Third message: RetryPromptPart becomes UserPromptPart (lossy conversion)
+    assert isinstance(reloaded_messages[2], ModelRequest)
+    assert isinstance(reloaded_messages[2].parts[0], UserPromptPart)
+    assert reloaded_messages[2].parts[0].content == (
```
DouweM: Can we "surgically" replace the part in question in `reloaded_messages` and then do the full `assert reloaded_messages == messages`? E.g:
```py
assert reloaded_messages[2] == snapshot()
reloaded_messages[2].parts[0] = RetryPromptPart(...)
assert reloaded_messages == messages
```

Similar to what we do above where we set `retry_prompt_part.content`.

tests/test_vercel_ai.py
```diff
@@ -2661,6 +2661,16 @@ async def test_adapter_dump_messages_with_retry():
         ]
     )
 
+    # Verify roundtrip
+    reloaded_messages = VercelAIAdapter.load_messages(ui_messages)
+    # Content will have changed for retry prompt part, so we set it back to the original value
+    retry_prompt_part = reloaded_messages[2].parts[0]
+    assert isinstance(retry_prompt_part, RetryPromptPart)
+    assert retry_prompt_part.content == 'Tool failed with error\n\nFix the errors and try again.'
```
DouweM: Let's use `assert retry_prompt_part == snapshot()` before swapping out the content. Similar in spirit to the request I put below.

pydantic_ai_slim/pydantic_ai/_parts_manager.py
Light2Dark: I think it's harder to track the parent part / existing part at the call site. Eg at 
https://github.com/pydantic/pydantic-ai/blob/ff07cc1c0434e73d6d33a20b26df7823ede7900a/pydantic_ai_slim/pydantic_ai/models/openai.py#L2440-L2447

We'd have to track whether we have already passed in `provider_name` to previous parts. Parts manager already keeps track of this state and decides when to pass in `provider_name` or not.

pydantic_ai_slim/pydantic_ai/models/openai.py
Light2Dark: Thank you, I missed that logic.

DouweM:
@Light2Dark Thanks for this Shahmir!

Light2Dark:
> @Light2Dark Thanks for this Shahmir!

Tq for all the reviews, really appreciate getting this in!

PR #3758, opened 2025-12-17
Operate on a deepcopy of `$defs` in `JsonSchemaTransformer` instead of the original schema

thomasst:
### Summary

This PR fixes `JsonSchemaTransformer` mutating input schemas by making it operate on a deep-copied version of `$defs` instead of the original.

### Problem

`JsonSchemaTransformer.walk()` deep-copies the root schema but continues to traverse `$defs` from the original schema. This can corrupt the caller’s schema, e.g. when handling a union.

### Fix

Initialize the transformer with a deep-copy of `$defs`, ensuring it is side-effect-free.


DouweM:
@thomasst Thank you!

PR #3759, opened 2025-12-17
Add support for `ImageGenerationTool` `output_compression` and `output_format` for Vertex AI Gemini image models

h0rv:
As of now, these options are only available via vertex-ai, not google-gla

### DIFF
--- a/docs/builtin-tools.md
+++ b/docs/builtin-tools.md
@@ -381,7 +381,7 @@ For more details, check the [API documentation][pydantic_ai.builtin_tools.ImageG
 | `input_fidelity` | ✅ | ❌ |
 | `moderation` | ✅ | ❌ |
 | `output_compression` | ✅ (100 (default), jpeg or webp only) | ✅ (75 (default), jpeg only, Vertex AI only) |
-| `output_format` | ✅ (png (default), webp, jpeg) | ✅ (png (default), jpeg, webp, Vertex AI only) |
+| `output_format` | ✅ | ✅ (Vertex AI only) |
 | `partial_images` | ✅ | ❌ |
 | `quality` | ✅ | ❌ |
 | `size` | ✅ (auto (default), 1024x1024, 1024x1536, 1536x1024) | ✅ (1K (default), 2K, 4K) |


pydantic_ai_slim/pydantic_ai/builtin_tools.py
```diff
@@ -309,6 +310,7 @@ class ImageGenerationTool(AbstractBuiltinTool):
     Supported by:
 
     * OpenAI Responses. Default: 'png'.
+    * Google (Vertex AI only).
```
DouweM: Can you add in the default, which I'm assuming is also PNG?

tests/models/test_google.py
```diff
@@ -3694,6 +3694,110 @@ async def test_google_image_generation_auto_size_raises_error(google_provider: G
...
+
+async def test_google_image_generation_rejected_by_gemini_api(
+    allow_model_requests: None, google_provider: GoogleProvider
+) -> None:
+    """Test that output_format and compression are rejected by Gemini API (google-gla)."""
+    model = GoogleModel('gemini-2.5-flash-image', provider=google_provider)
+
+    # Test output_format rejection
+    agent = Agent(model, builtin_tools=[ImageGenerationTool(output_format='png')], output_type=BinaryImage)
+    with pytest.raises(ValueError, match='output_mime_type parameter is not supported in Gemini API'):
```
DouweM: Instead of hitting errors, let's ignore the field silently as we would've done before. We've already documented it only works with Vertex, so we can check `self.system == 'google-vertex'` inside `GoogleModel`

pydantic_ai_slim/pydantic_ai/builtin_tools.py
```diff
@@ -301,6 +301,7 @@ class ImageGenerationTool(AbstractBuiltinTool):
     Supported by:
 
     * OpenAI Responses. Only supported for 'png' and 'webp' output formats.
+    * Google (Vertex AI only). Only supported for 'jpeg' output format.
```
DouweM: Please update the docs as well

DouweM (changes requested):

h0rv:
Updated to match Vertex AI documentation for compression quality:
- Changed `ImageGenerationTool.output_compression` default to `None`.
- `GoogleModel` now omits the compression parameter when `None`, allowing Vertex AI to use its documented default of `75`.
- `OpenAIChatModel` preserves its existing behavior by defaulting to `100` when the tool's value is `None`.
- Updated docstrings and tests to reflect these provider-specific defaults.

h0rv:
https://platform.openai.com/docs/guides/image-generation#customize-image-output

also fixed openai doc:

image

pydantic_ai_slim/pydantic_ai/models/google.py
```diff
@@ -358,6 +361,46 @@ async def request_stream(
...
+                image_config['output_mime_type'] = f'image/{tool.output_format}'
+
+            output_compression = tool.output_compression
+            if output_compression is not None:
+                if not (0 <= output_compression <= 100):
+                    raise UserError(
+                        f'Google image generation `output_compression` must be between 0 and 100. '
+                        f'Got: {output_compression}.'
+                    )
+                if tool.output_format not in (None, 'jpeg'):
```
DouweM, line 395: Does the `None` here mean that if `output_compression` is set but not `output_format`, the used output format automatically becomes JPEG? Then we should update the "default" wording in the docstring above.

pydantic_ai_slim/pydantic_ai/builtin_tools.py
```diff
@@ -295,12 +295,13 @@ class ImageGenerationTool(AbstractBuiltinTool):
...
 
-    output_compression: int = 100
+    output_compression: int | None = None
     """Compression level for the output image.
 
     Supported by:
 
-    * OpenAI Responses. Only supported for 'png' and 'webp' output formats.
+    * OpenAI Responses. Only supported for 'jpeg' and 'webp' output formats. Default: 100.
+    * Google (Vertex AI only). Only supported for 'jpeg' output format. Default: 75.
```
DouweM, line 304: Please update this section: https://ai.pydantic.dev/builtin-tools/#provider-support_4

docs/builtin-tools.md
```diff
@@ -380,14 +380,16 @@ For more details, check the [API documentation][pydantic_ai.builtin_tools.ImageG
 | `background` | ✅ | ❌ |
 | `input_fidelity` | ✅ | ❌ |
 | `moderation` | ✅ | ❌ |
-| `output_compression` | ✅ | ❌ |
-| `output_format` | ✅ | ❌ |
+| `output_compression` | ✅ (100 (default), jpeg or webp only) | ✅ (75 (default), jpeg only, Vertex AI only) |
+| `output_format` | ✅ (png (default), webp, jpeg) | ✅ (png (default), jpeg, webp, Vertex AI only) |
```
DouweM: Since they both the same values with the same default, I don't think we have to repeat the options here; just "Vertex AI onlyt"

PR #3760, opened 2025-12-18
Add AI SDK data chunk ID and tool approval types

bendrucker:
Adds missing fields to Vercel AI adapter types for full AI SDK compatibility, addressing data part reconciliation and other protocol gaps.

## Changes

- [`DataChunk`](https://github.com/vercel/ai/blob/c7d89b3601dc42bff4903339973da6e73c6c43a4/packages/ai/src/ui-message-stream/ui-message-chunks.ts#L135-L143): Adds optional `id` field for data part reconciliation and `transient` field for non-persistent data
- [`FinishChunk`](https://github.com/vercel/ai/blob/c7d89b3601dc42bff4903339973da6e73c6c43a4/packages/ai/src/ui-message-stream/ui-message-chunks.ts#L155-L168): Adds `finish_reason` field with standard stop reasons (`stop`, `length`, `content-filter`, `tool-calls`, `error`, `other`, `unknown`)
  - Populated from `ModelResponse.finish_reason` in `handle_run_result`
  - Set to `error` in `on_error` handler
- Adds [`ToolApprovalRequestChunk`](https://github.com/vercel/ai/blob/c7d89b3601dc42bff4903339973da6e73c6c43a4/packages/ai/src/ui-message-stream/ui-message-chunks.ts#L74-L77) and [`ToolOutputDeniedChunk`](https://github.com/vercel/ai/blob/c7d89b3601dc42bff4903339973da6e73c6c43a4/packages/ai/src/ui-message-stream/ui-message-chunks.ts#L94-L96) types for human-in-the-loop tool approval workflows

## Future Work

Tool approval integration will be addressed in a separate PR. This requires:

1. Emitting `ToolApprovalRequestChunk` when agent output is `DeferredToolRequests`
2. Exploring how Vercel AI SDK sends back approval/denial responses to map them to `DeferredToolResults`
3. Handling `ToolOutputDeniedChunk` — currently Pydantic AI doesn't distinguish denied tool results from normal results at the event stream level (denied results flow as `FunctionToolResultEvent` with `ToolReturnPart`)

## Testing

- Added test for `DataChunk` with `id` field to verify reconciliation IDs serialize correctly in the SSE stream
- Added test for `DataChunk` with `transient: true` to verify non-persistent data chunks are encoded properly
- Updated existing error handling tests to verify `finishReason: 'error'` is set correctly

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_event_stream.py
+++ b/pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_event_stream.py
@@ -5,7 +5,6 @@ from __future__ import annotations
 from collections.abc import AsyncIterator, Mapping
 from dataclasses import dataclass
 from typing import Any
-from uuid import uuid4
 
 from pydantic_core import to_json
 
@@ -16,7 +15,6 @@ from ...messages import (
     FilePart,
     FinishReason as PydanticFinishReason,
     FunctionToolResultEvent,
-    ModelResponse,
     RetryPromptPart,
     TextPart,
     TextPartDelta,
@@ -27,7 +25,7 @@ from ...messages import (
 )
 from ...output import OutputDataT
 from ...run import AgentRunResultEvent
-from ...tools import AgentDepsT, DeferredToolRequests
+from ...tools import AgentDepsT
 from .. import UIEventStream
 from .request_types import RequestData
 from .response_types import (
@@ -46,7 +44,6 @@ from .response_types import (
     TextDeltaChunk,
     TextEndChunk,
     TextStartChunk,
-    ToolApprovalRequestChunk,
     ToolInputAvailableChunk,
     ToolInputDeltaChunk,
     ToolInputStartChunk,
@@ -105,22 +102,14 @@ class VercelAIEventStream(UIEventStream[RequestData, BaseChunk, AgentDepsT, Outp
         yield DoneChunk()
 
     async def handle_run_result(self, event: AgentRunResultEvent) -> AsyncIterator[BaseChunk]:
-        messages = event.result.all_messages()
-        if messages and isinstance(messages[-1], ModelResponse):
-            pydantic_reason = messages[-1].finish_reason
-            if pydantic_reason:
-                self._finish_reason = _FINISH_REASON_MAP.get(pydantic_reason)
-
-        # Emit tool approval requests for deferred approvals
-        output = event.result.output
-        if isinstance(output, DeferredToolRequests):
-            for tool_call in output.approvals:
-                yield ToolApprovalRequestChunk(
-                    approval_id=str(uuid4()),
-                    tool_call_id=tool_call.tool_call_id,
-                )
+        pydantic_reason = event.result.response.finish_reason
+        if pydantic_reason:
+            self._finish_reason = _FINISH_REASON_MAP.get(pydantic_reason)
+        return
+        yield
 
     async def on_error(self, error: Exception) -> AsyncIterator[BaseChunk]:
+        self._finish_reason = 'error'
         yield ErrorChunk(error_text=str(error))
 
     async def handle_text_start(self, part: TextPart, follows_text: bool = False) -> AsyncIterator[BaseChunk]:
--- a/tests/test_vercel_ai.py
+++ b/tests/test_vercel_ai.py
@@ -1488,7 +1488,7 @@ Fix the errors and try again.\
             {'type': 'tool-input-available', 'toolCallId': IsStr(), 'toolName': 'unknown_tool', 'input': {}},
             {'type': 'error', 'errorText': 'Exceeded maximum retries (1) for output validation'},
             {'type': 'finish-step'},
-            {'type': 'finish'},
+            {'type': 'finish', 'finishReason': 'error'},
             '[DONE]',
         ]
     )
@@ -1531,7 +1531,7 @@ async def test_run_stream_request_error():
             },
             {'type': 'error', 'errorText': 'Unknown tool'},
             {'type': 'finish-step'},
-            {'type': 'finish'},
+            {'type': 'finish', 'finishReason': 'error'},
             '[DONE]',
         ]
     )
@@ -1572,7 +1572,7 @@ async def test_run_stream_on_complete_error():
             {'type': 'text-end', 'id': IsStr()},
             {'type': 'error', 'errorText': 'Faulty on_complete'},
             {'type': 'finish-step'},
-            {'type': 'finish'},
+            {'type': 'finish', 'finishReason': 'error'},
             '[DONE]',
         ]
     )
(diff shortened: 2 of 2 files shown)


pydantic_ai_slim/pydantic_ai/ui/vercel_ai/response_types.py
```diff
@@ -205,6 +227,8 @@ class FinishChunk(BaseChunk):
     """Finish chunk."""
 
     type: Literal['finish'] = 'finish'
+    finish_reason: FinishReason = None
```
DouweM, line 228: Can you please update the `after_stream` method that emits `FinishChunk` to set this field based on `ModelResponse.finish_reason`? That method doesn't take the model response directly, but is guaranteed to happen _after_ the `AgentRunResultEvent` handler is called, so we can set an instance var there and read it from `after_stream`.

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/response_types.py
```diff
@@ -145,6 +148,21 @@ class ToolOutputErrorChunk(BaseChunk):
     dynamic: bool | None = None
 
 
+class ToolApprovalRequestChunk(BaseChunk):
```
DouweM, line 151: Do you know if it's feasible for us to yield this automatically when https://ai.pydantic.dev/deferred-tools/#human-in-the-loop-tool-approval is used? I haven't looked into how approval requests work with Vercel AI, yet.

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/response_types.py
```diff
@@ -145,6 +148,21 @@ class ToolOutputErrorChunk(BaseChunk):
...
+    """Tool approval request chunk for human-in-the-loop approval."""
+
+    type: Literal['tool-approval-request'] = 'tool-approval-request'
+    approval_id: str
+    tool_call_id: str
+
+
+class ToolOutputDeniedChunk(BaseChunk):
```
DouweM, line 159: Do we ever receive (the part version) of this, and should we handle it? We'll likely need it for https://ai.pydantic.dev/deferred-tools/#human-in-the-loop-tool-approval

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/response_types.py
bendrucker: Will do!

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/response_types.py
bendrucker: Let me look into that, I originally set out to support `id` for data parts but will be working on tool approval soon. I compared the types in the AI SDK to the ones here and found the tool chunks missing. But I'll keep digging on what's needed to actually make this work end to end.

bendrucker: Turns out yes, easy enough to do this in `handle_run_result` after saving the finish reason. For any `DeferredToolRequests`, emit approval request chunks.

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_event_stream.py
```diff
@@ -85,9 +101,25 @@ async def before_response(self) -> AsyncIterator[BaseChunk]:
     async def after_stream(self) -> AsyncIterator[BaseChunk]:
         yield FinishStepChunk()
 
-        yield FinishChunk()
+        yield FinishChunk(finish_reason=self._finish_reason)
         yield DoneChunk()
 
+    async def handle_run_result(self, event: AgentRunResultEvent) -> AsyncIterator[BaseChunk]:
+        messages = event.result.all_messages()
+        if messages and isinstance(messages[-1], ModelResponse):
```
DouweM: We can directly read `event.result.response`

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_event_stream.py
```diff
@@ -85,9 +101,25 @@ async def before_response(self) -> AsyncIterator[BaseChunk]:
...
+            pydantic_reason = messages[-1].finish_reason
+            if pydantic_reason:
+                self._finish_reason = _FINISH_REASON_MAP.get(pydantic_reason)
+
+        # Emit tool approval requests for deferred approvals
+        output = event.result.output
+        if isinstance(output, DeferredToolRequests):
+            for tool_call in output.approvals:
+                yield ToolApprovalRequestChunk(
```
DouweM: Do you know how Vercel AI will send back the approval/denial? We'd also have to map that to `DeferredToolResults` for the entire flow to work automatically. Let's do that in a separate PR (if you're up for it!) and limit this one again to the types + finish reason.

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_event_stream.py
```diff
@@ -85,9 +101,25 @@ async def before_response(self) -> AsyncIterator[BaseChunk]:
...
-        yield FinishChunk()
+        yield FinishChunk(finish_reason=self._finish_reason)
         yield DoneChunk()
 
+    async def handle_run_result(self, event: AgentRunResultEvent) -> AsyncIterator[BaseChunk]:
+        messages = event.result.all_messages()
+        if messages and isinstance(messages[-1], ModelResponse):
+            pydantic_reason = messages[-1].finish_reason
+            if pydantic_reason:
+                self._finish_reason = _FINISH_REASON_MAP.get(pydantic_reason)
```
DouweM: Should we set this to `error` in `on_error`?

bendrucker:
Forgot to post this comment, splitting off deferred tool <> approval integration definitely makes sense based on my epxloration. I will slim this down to types + finish reason and then kick off a draft for tool approval logic.

Dug into denials a bit, approvals are closer to being easily supported.

When a user denies a tool execution, Pydantic AI doesn't have a distinct event type for it. Denied tool results flow as regular `FunctionToolResultEvent` with `ToolReturnPart`, with the denial message as the `content` field. There's no way to distinguish a denial from a normal tool result at the event stream level.

So we'd need something like:

- A new event type like `DeniedToolResultEvent`
- A `ToolDeniedPart` message type (vs `ToolReturnPart`)

DouweM:
> When a user denies a tool execution, Pydantic AI doesn't have a distinct event type for it.

@bendrucker Hmm, Pydantic AI doesn't have an event for denials because denials never originate in Pydantic AI itself, they come from the user in the form of `DeferredToolResults` passed into `agent.run`, so I'm not sure why Vercel AI expects us to yield denial chunks to them, rather than them sending approval/denial parts or metadata to us?

bendrucker:
Yeah you're right, I need to keep reading. The only thing I can think of is that the client is expecting an acknowledgement from the server of the denial. I've opened #3772 with half-working code and some notes for discussion. Really an issue with illustrative code attached right now, not close to reviewable/mergeable. If you want to dig further/comment, I'm certainly interested, but otherwise will spend some more time validating the client-side behavior and getting the PR to a reviewable state.

Issue #3710, opened 2025-12-11
lowest-versions CI jobs are not actually testing lowest package versions


DouweM:
### Initial Checks

- [x] I confirm that I'm using the latest version of Pydantic AI
- [x] I confirm that I searched for my issue in https://github.com/pydantic/pydantic-ai/issues before opening this issue

### Description

It appears the `lowest-version` CI jobs are broken.

They should've caught this issue: https://github.com/pydantic/pydantic-ai/issues/3707, but the job succeeded: https://github.com/pydantic/pydantic-ai/actions/runs/20104362057/job/57683809149?pr=3678.

Locally, this fails as expected on the v1.30.0 tag:  

```py
uv run --with openai==1.107.0 pytest tests/models/test_openai.py::test_image_url_tool_response
```

But in CI `uv run --all-extras --resolution lowest-direct coverage run -m pytest --durations=100 -n auto --dist=loadgroup` appears to not have installed the oldest version ( `openai==1.107.0`) but rather the latest version.

### Example Code

```Python

```

### Python, Pydantic AI & LLM client version

```Text
latest
```

PR #3765, opened 2025-12-18
Install lowest direct in CI

Kludex:
Closes https://github.com/pydantic/pydantic-ai/issues/3710

---

> [!NOTE]
> Adjust CI to use lowest-direct installs and expand Python matrix, bump a few dependency pins, and harden tests with guarded imports and minor fixes.
> 
> - **CI**:
>   - Use `uv sync --group dev` first and `--resolution lowest-direct` for lowest-versions; run pytest with `--no-sync`.
>   - Expand Python matrix to include `3.11`; increase `test-lowest-versions` timeout to `35m`; set `UV_FROZEN="0"` for lowest runs.
> - **Dependencies**:
>   - Pin `exceptiongroup>=1.2.2` (py<3.11).
>   - Tighten optional deps: `pillow>11.0.0`, `torch>=2.8.0`, `vllm>=0.8.0`, and `boto3-stubs[bedrock-runtime]>=1.40.64`.
> - **Tests/Typing**:
>   - Guard vendor imports with `try_import()` and move some imports under TYPE_CHECKING.
>   - Bedrock tests parametrize geo prefixes safely when package missing.
>   - Google error test now parameterizes error class by name via `getattr`.
>   - Minor refactors in HuggingFace/Anthropic/UI tests and small formatting in `create_web_app` HTML response.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 48b6562ffe3fedf11ec83c6d0a565f0236872cc4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -194,7 +194,7 @@ jobs:
   test-lowest-versions:
     name: test on ${{ matrix.python-version }} (lowest-versions)
     runs-on: ubuntu-latest
-    timeout-minutes: 25
+    timeout-minutes: 35
     strategy:
       fail-fast: false
       matrix:


.github/workflows/ci.yml
```diff
@@ -171,8 +171,6 @@ jobs:
 
       - run: uv sync --only-dev
 
-      - run: uv run mcp-run-python example --deps=numpy
```
Kludex, line 174: @DouweM is this still needed now that we have the `mcp-run-python` in another repo? I've already checked that this test is also ran on that repo.

DouweM: @Kludex I'm fine removing it. 

Note that we also test that it `mcp-run-python` be started here:

https://github.com/pydantic/pydantic-ai/blob/c1213e4a7a978099b64c642b38e1e041d18afad6/docs/mcp/client.md#L140-L152

That makes the test suite require deno. I actually think we should drop that entirely and really treat it as a separate package.

.github/workflows/ci.yml
```diff
@@ -171,8 +171,6 @@ jobs:
 
       - run: uv sync --only-dev
```
DouweM: Should we also move the `${{ matrix.install.command }}` here and make the test command `uv run --no-sync ...`?

DouweM:
Some warnings from the lowest-versions job:

```
warning: The direct dependency `boto3-stubs` is unpinned. Consider setting a lower bound when using `--resolution lowest` or `--resolution lowest-direct` to avoid using outdated versions.
warning: The direct dependency `exceptiongroup` is unpinned. Consider setting a lower bound when using `--resolution lowest` or `--resolution lowest-direct` to avoid using outdated versions.
warning: The direct dependency `torch` is unpinned. Consider setting a lower bound when using `--resolution lowest` or `--resolution lowest-direct` to avoid using outdated versions.
warning: The package `boto3-stubs==0.1.8` does not have an extra named `bedrock-runtime`
```

.github/workflows/ci.yml
```diff
@@ -169,9 +169,8 @@ jobs:
 
       - run: mkdir .coverage
 
-      - run: uv sync --only-dev
-
-      - run: uv run mcp-run-python example --deps=numpy
+      # We install the `--group dev` first because if we use `--package` together, it will try to install the `dev` group from the package.
+      - run: uv sync --group dev
```
DouweM, line 173: Adding `matrix.install.command` here didn't work?

.github/workflows/ci.yml
```diff
@@ -195,16 +194,16 @@ jobs:
   test-lowest-versions:
     name: test on ${{ matrix.python-version }} (lowest-versions)
     runs-on: ubuntu-latest
-    timeout-minutes: 20
+    timeout-minutes: 35
```
dsfaccini, line 197: needs to be reduced to 12 actualyl

PR #3766, opened 2025-12-18
Re-add `clai --help` output to clai README

dsfaccini:


clai/README.md
```diff
@@ -51,7 +51,35 @@ Either way, running `clai` will start an interactive session where you can chat
...
 
-For full CLI documentation, see the [CLI documentation](https://ai.pydantic.dev/cli/).
+## Help
+
+```
+usage: clai [-h] [-l] [--version] [-m MODEL] [-a AGENT] [-t CODE_THEME] [--no-stream] [prompt]
+
+Pydantic AI CLI v...
+
+Commands:
```
DouweM: The other sections are not capitalized, so let's make this one `subcommands`

clai/README.md
```diff
@@ -51,7 +51,35 @@ Either way, running `clai` will start an interactive session where you can chat
...
+  web         Start a web-based chat interface for an agent
+
+Use "clai web --help" for more information on the web command.
```
DouweM: "subcommands" implies there may be more, so then we should indent all the info relevant to each specific subcommand:

```suggestion
  web         Start a web-based chat interface for an agent
              Run "clai web --help" for more information
```

pydantic_ai_slim/pydantic_ai/_cli/__init__.py
```diff
@@ -180,8 +180,8 @@ def _cli_web(args_list: list[str], prog_name: str, default_model: str, qualified
     )
     parser.add_argument('--host', default='127.0.0.1', help='Host to bind server (default: 127.0.0.1)')
     parser.add_argument('--port', type=int, default=7932, help='Port to bind server (default: 7932)')
-
-    argcomplete.autocomplete(parser)
+    if 'pytest' not in sys.modules:  # pragma: no cover
```
DouweM: Why is this needed now?

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/_cli/__init__.py
dsfaccini: `uv run pytest` was hanging, ended up being an orphaned `uv` lock in the cache. I left this in though because it makes sense to me that this would me the loading faster during testing (which doesn't need tab-completion).

Edit: ok no it doesn't make a difference

clai/update_readme.py
```diff
@@ -0,0 +1,30 @@
...
+
+@pytest.mark.skipif(sys.version_info >= (3, 13), reason='slightly different output with 3.13')
+def test_cli_help(capfd: pytest.CaptureFixture[str]):
+    """Check README.md help output matches `clai --help`."""
+    os.environ['COLUMNS'] = '150'
+    with pytest.raises(SystemExit):
+        cli(['--help'], prog_name='clai')
+
+    help_output = capfd.readouterr().out.strip()
+    # TODO change when we reach v1
```
DouweM: we're already at v1 lol, so this can be removed?

docs/mcp/client.md
```diff
@@ -136,7 +136,7 @@ MCP also offers [stdio transport](https://spec.modelcontextprotocol.io/specifica
 
 In this example [mcp-run-python](https://github.com/pydantic/mcp-run-python) is used as the MCP server.
 
-```python {title="mcp_stdio_client.py"}
+```python {title="mcp_stdio_client.py" test="skip"}
```
DouweM: let's not include this in this PR, and in the PR where we do this, we should also drop the deno requirement

clai/update_readme.py
dsfaccini: right lol I just restored the file from the commit

docs/mcp/client.md
dsfaccini: I pushed the fix for the mcp error here https://github.com/pydantic/pydantic-ai/pull/3779

PR #3772, opened 2025-12-19
Add tool approval integration for Vercel AI adapter

bendrucker:
Closes https://github.com/pydantic/pydantic-ai/issues/4279

Adds tool approval integration for the Vercel AI adapter, enabling human-in-the-loop workflows with AI SDK UI.

## Summary

- Setting `sdk_version=6` enables tool approval
- Emit `tool-approval-request` chunks for deferred tool approvals
- Emit `tool-output-denied` chunks when user denies tool execution
- Auto-extract approval responses from follow-up requests
- Add `approval` field to tool parts with `ToolApprovalRequested`/`ToolApprovalResponded` states

## Usage

```python
from pydantic_ai import Agent
from pydantic_ai.ui.vercel_ai import VercelAIAdapter

agent = Agent('openai:gpt-5.2')

@agent.tool_plain(requires_approval=True)
def delete_file(path: str) -> str:
    """Delete a file from the filesystem."""
    os.remove(path)
    return f'Deleted {path}'

@app.post('/chat')
async def chat(request: Request) -> Response:
    adapter = await VercelAIAdapter.from_request(request, agent=agent, sdk_version=6)
    return adapter.streaming_response(adapter.run_stream())
```

When `sdk_version=6`, the adapter will:
1. Emit `tool-approval-request` chunks when tools with `requires_approval=True` are called
2. Automatically extract approval responses from follow-up requests
3. Emit `tool-output-denied` chunks for rejected tools, passing the denial `reason` through as `ToolDenied`

## AI SDK Tool Approval Protocol

Tool approval is an [AI SDK v6 feature](https://ai-sdk.dev/cookbook/next/human-in-the-loop). Here's how the protocol works:

### Protocol Flow

```mermaid
sequenceDiagram
    participant Model
    participant Server
    participant Client

    Model->>Server: tool call
    Server->>Client: tool-input-start
    Server->>Client: tool-input-available
    Server->>Client: tool-approval-request

    Note over Client: User approves/denies

    Client->>Server: approval response (next request)

    alt Approved
        Server->>Server: Execute tool
        Server->>Client: tool-output-available
    else Denied
        Server->>Client: tool-output-denied
        Server->>Model: denial info
    end
```

#### `tool-approval-request`

Server → client:

```json
{
  "type": "tool-approval-request",
  "approvalId": "<uuid>",
  "toolCallId": "<tool-call-id>"
}
```

#### `tool-output-denied`

Server → client, when denied:

```json
{
  "type": "tool-output-denied",
  "toolCallId": "<tool-call-id>"
}
```

#### Why the Server Emits This Chunk

> "Pydantic AI doesn't have an event for denials because denials never originate in Pydantic AI itself, they come from the user in the form of DeferredToolResults passed into agent.run" — [#3760 comment](https://github.com/pydantic/pydantic-ai/pull/3760#issuecomment-3672800411)

This is correct—the denial decision originates from the user via the `ToolApprovalResponded` field in tool parts. However, the AI SDK protocol expects the server to emit `tool-output-denied` for two reasons:

1. The client needs confirmation that the tool lifecycle is complete. When the server receives a denial and emits `tool-output-denied`, the client can transition its UI from "awaiting result" to "denied".

2. Just as the server emits `tool-output-available` when a tool executes successfully, it emits `tool-output-denied` when execution is skipped due to denial. This gives the client a consistent signal for each tool call's final state.

The flow is:
1. Client sends denial via `approval: { id, approved: false, reason }` in the tool part
2. Server extracts this from `deferred_tool_results` and passes it to the agent
3. When processing the denied tool result, server emits `tool-output-denied` to the client
4. Client updates UI to show the tool was denied

### Tool Part Approval States

Tool parts include an `approval` field tracking the approval lifecycle:

#### Awaiting Response

```json
{ "id": "<approval-id>" }
```

#### User Responded

```json
{ "id": "<approval-id>", "approved": true/false, "reason": "optional" }
```

### Two-Step Flow

Unlike regular tool calls, approved tools require two model interactions:

1. Model requests tool → server returns with `tool-approval-request` → awaits user
2. User decision sent → server executes (if approved) or informs model (if denied)

## Changes

- Add `ToolApprovalRequested`/`ToolApprovalResponded` types and `approval` field to tool UI parts
- Add `ToolApprovalRequestChunk` and `ToolOutputDeniedChunk` response chunks to event stream
- Gate tool approval behavior on `sdk_version=6` (added in #4166), enabling auto-extraction of approval responses via `deferred_tool_results` cached property
- Pass denial `reason` through as `ToolDenied` when provided
- Add tool approval section to Vercel AI docs

## Testing

- Approval request/denied chunk emission, opt-out behavior, and approval extraction (approved, denied with reason, no approval)
- Snapshot updates for `approval` field on tool parts

## References

- Builds on #3760 (merged)
- [AI SDK Human-in-the-Loop Cookbook](https://ai-sdk.dev/cookbook/next/human-in-the-loop)
- [AI SDK Stream Protocol](https://ai-sdk.dev/docs/ai-sdk-ui/stream-protocol)
- [Pydantic AI Deferred Tools](https://ai.pydantic.dev/deferred-tools/)

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/ui/_adapter.py
+++ b/pydantic_ai_slim/pydantic_ai/ui/_adapter.py
@@ -17,7 +17,7 @@ from typing import (
 )
 
 from pydantic import BaseModel, ValidationError
-from typing_extensions import TypeVar
+from typing_extensions import Self, TypeVar
 
 from pydantic_ai import DeferredToolRequests, DeferredToolResults
 from pydantic_ai.agent import AbstractAgent
@@ -128,7 +128,7 @@ class UIAdapter(ABC, Generic[RunInputT, MessageT, EventT, AgentDepsT, OutputData
     @classmethod
     async def from_request(
         cls, request: Request, *, agent: AbstractAgent[AgentDepsT, OutputDataT], **kwargs: Any
-    ) -> UIAdapter[RunInputT, MessageT, EventT, AgentDepsT, OutputDataT]:
+    ) -> Self:
         """Create an adapter from a request.
 
         Extra keyword arguments are forwarded to the adapter constructor, allowing subclasses
--- a/pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
+++ b/pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
@@ -109,21 +109,8 @@ class VercelAIAdapter(UIAdapter[RequestData, UIMessage, BaseChunk, AgentDepsT, O
         sdk_version: Literal[5, 6] = 5,
         **kwargs: Any,
     ) -> VercelAIAdapter[AgentDepsT, OutputDataT]:
-        """Create a Vercel AI adapter from a request.
-
-        Args:
-            request: The incoming Starlette/FastAPI request.
-            agent: The Pydantic AI agent to run.
-            sdk_version: Vercel AI SDK version. Set to 6 to enable tool approval streaming.
-            **kwargs: Additional keyword arguments forwarded to the adapter constructor.
-        """
-        return cls(
-            agent=agent,
-            run_input=cls.build_run_input(await request.body()),
-            accept=request.headers.get('accept'),
-            sdk_version=sdk_version,
-            **kwargs,
-        )
+        """Extends [`from_request`][pydantic_ai.ui.UIAdapter.from_request] with the `sdk_version` parameter."""
+        return await super().from_request(request, agent=agent, sdk_version=sdk_version, **kwargs)
 
     @classmethod
     async def dispatch_request(


docs/ui/vercel-ai.md
```diff
@@ -2,6 +2,9 @@
 
 Pydantic AI natively supports the [Vercel AI Data Stream Protocol](https://ai-sdk.dev/docs/ai-sdk-ui/stream-protocol#data-stream-protocol) to receive agent run input from, and stream events to, a [Vercel AI Elements](https://ai-sdk.dev/elements) frontend.
 
+!!! note "AI SDK Version Compatibility"
+    The base protocol is compatible with AI SDK v5 and later. However, [tool approval](#tool-approval) (human-in-the-loop) features require **AI SDK v6 or later**.
```
DouweM: - I don't think AI SDK is relevant here, as on the frontend the user would be using AI Elements, not the SDK. (Note that this could be wrong) Does AI Elements also only support this from a specific version?
- I think we should make this feature opt-in through a flag on `VercelAIAdapter`, both because of the people using older versions of Vercel AI on the frontend, _and_ because they may have already implemented their own `DeferredToolRequests` handling from before this was natively supported in AI Elements / Pydantic AI.

docs/ui/vercel-ai.md
```diff
@@ -81,3 +84,58 @@ async def chat(request: Request) -> Response:
     sse_event_stream = adapter.encode_stream(event_stream)
     return StreamingResponse(sse_event_stream, media_type=accept)
 ```
+
+## Tool Approval
+
+Pydantic AI supports [AI SDK's human-in-the-loop tool approval](https://ai-sdk.dev/cookbook/next/human-in-the-loop) workflow, allowing users to approve or deny tool executions before they run.
```
DouweM: Pydantic AI is an alternative to Vercel AI SDK, so we shouldn't need to link there. Us supporting tis feature is because of the Vercel AI Data Stream Protocol or Elements, not of AI SDK

DouweM: We should also link to our own HITL tool approval docs

docs/ui/vercel-ai.md
```diff
@@ -81,3 +84,58 @@ async def chat(request: Request) -> Response:
...
+```
+
+When processing a follow-up request with approval responses, extract and pass the deferred tool results:
+
+```py {test="skip"}
+@app.post('/chat')
+async def chat(request: Request) -> Response:
+    adapter = await VercelAIAdapter.from_request(request, agent=agent)
+    return adapter.streaming_response(
+        adapter.run_stream(deferred_tool_results=adapter.deferred_tool_results)
```
DouweM: We should be able to do this automatically, I'll comment below on how.

I don't think we need a code sample here at all, as we can link to our existing HITL tool approval docs and everything should just work.

docs/ui/vercel-ai.md
```diff
@@ -81,3 +84,58 @@ async def chat(request: Request) -> Response:
...
+    )
+```
+
+### Client-Side Setup
+
+On the frontend, use AI SDK v6's [`useChat`](https://v6.ai-sdk.dev/docs/ai-sdk-ui/use-chat) hook with the [`Confirmation`](https://ai-sdk.dev/elements/components/confirmation) component or the `addToolApprovalResponse` function to handle approval UI.
```
DouweM: I think we only need to document the client-side/frontend setup needs here, plus enabling the flag on the backend.

docs/ui/vercel-ai.md
```diff
@@ -81,3 +84,58 @@ async def chat(request: Request) -> Response:
...
+
+See the [AI SDK Human-in-the-Loop Cookbook](https://ai-sdk.dev/cookbook/next/human-in-the-loop) for complete frontend examples.
```
DouweM: I don't know how relevant this is, as it's for their agent SDK, not the frontend component library right?

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
```diff
@@ -87,6 +88,57 @@ def messages(self) -> list[ModelMessage]:
         """Pydantic AI messages from the Vercel AI run input."""
         return self.load_messages(self.run_input.messages)
 
+    @cached_property
+    def deferred_tool_results(self) -> DeferredToolResults | None:
```
DouweM, line 165: Please make this a field on the superclass, like `messages`, `toolset`, etc.

DouweM: We should then also update `run_stream_native` to automatically pass through to `agent.run` in if it's set

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
```diff
@@ -87,6 +88,57 @@ def messages(self) -> list[ModelMessage]:
...
+        the approval decision in the tool part's `approval` field. This method extracts
+        those responses and converts them to Pydantic AI's `DeferredToolResults` format.
+
+        Returns:
+            DeferredToolResults if any tool parts have approval responses, None otherwise.
+        """
+        return self.extract_deferred_tool_results(self.run_input.messages)
+
+    @classmethod
+    def extract_deferred_tool_results(cls, messages: Sequence[UIMessage]) -> DeferredToolResults | None:
```
DouweM: No need to have a separate method for this, this can be inlined into the method above

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
```diff
@@ -87,6 +88,57 @@ def messages(self) -> list[ModelMessage]:
...
+        for msg in messages:
+            if msg.role != 'assistant':
+                continue
+
+            for part in msg.parts:
+                if not isinstance(part, ToolUIPart | DynamicToolUIPart):
+                    continue
+
+                approval = part.approval
+                if approval is None or not isinstance(approval, ToolApprovalResponded):
```
DouweM: We shouldn't need the `isinstance` check as `part.approval` is already properly typed

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
```diff
@@ -87,6 +88,57 @@ def messages(self) -> list[ModelMessage]:
...
+                    continue
+
+                tool_call_id = part.tool_call_id
+                if not tool_call_id:
```
DouweM: Isn't there always a tool call ID?

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
```diff
@@ -87,6 +88,57 @@ def messages(self) -> list[ModelMessage]:
...
+                    continue
+
+                if approval.approved:
+                    approvals[tool_call_id] = ToolApproved()
+                else:
+                    approvals[tool_call_id] = ToolDenied(message=approval.reason or 'The tool call was denied.')
```
DouweM: If there's no message, we can pass `False`, and Pydantic AI will automatically pass a default message to the model

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
```diff
@@ -87,6 +88,57 @@ def messages(self) -> list[ModelMessage]:
...
+                approval = part.approval
+                if approval is None or not isinstance(approval, ToolApprovalResponded):
+                    continue
+
+                tool_call_id = part.tool_call_id
+                if not tool_call_id:
+                    continue
+
+                if approval.approved:
+                    approvals[tool_call_id] = ToolApproved()
```
DouweM: Let's use `True` as we're not setting `override_args`

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_event_stream.py
```diff
@@ -71,12 +80,37 @@ def _json_dumps(obj: Any) -> str:
     return to_json(obj).decode('utf-8')
 
 
+def _extract_denied_tool_ids(messages: list[UIMessage]) -> set[str]:
```
DouweM: I think we can inline this into the method on the class, if we make it a `cached_property`

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_event_stream.py
```diff
@@ -71,12 +80,37 @@ def _json_dumps(obj: Any) -> str:
...
+    """Extract tool_call_ids that were denied from UI messages."""
+    denied_ids: set[str] = set()
+    for msg in messages:
+        if msg.role != 'assistant':
+            continue
+        for part in msg.parts:
+            if not isinstance(part, ToolUIPart | DynamicToolUIPart):
+                continue
+            approval = part.approval
+            if isinstance(approval, ToolApprovalResponded) and not approval.approved:
```
DouweM: There's some duplication here with `extract_deferred_tool_results` above. Maybe the `run_input` object could get a helper method that returns the approvals/denials from the message history in a way that can be used in both these places?

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_event_stream.py
```diff
@@ -71,12 +80,37 @@ def _json_dumps(obj: Any) -> str:
...
+                tool_call_id = part.tool_call_id
+                if tool_call_id:
```
DouweM: Isn't there always a tool call ID?

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_event_stream.py
```diff
@@ -71,12 +80,37 @@ def _json_dumps(obj: Any) -> str:
...
 @dataclass
 class VercelAIEventStream(UIEventStream[RequestData, BaseChunk, AgentDepsT, OutputDataT]):
     """UI event stream transformer for the Vercel AI protocol."""
 
     _step_started: bool = False
     _finish_reason: FinishReason = None
+    _denied_tool_ids: set[str] | None = None
+
+    @property
+    def denied_tool_ids(self) -> set[str]:
```
DouweM: I think this should be a private cached property

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/response_types.py
```diff
@@ -1,7 +1,9 @@
 """Vercel AI response types (SSE chunks).
 
 Converted to Python from:
-https://github.com/vercel/ai/blob/ai%405.0.59/packages/ai/src/ui-message-stream/ui-message-chunks.ts
+https://github.com/vercel/ai/blob/ai%406.0.0-beta.159/packages/ai/src/ui-message-stream/ui-message-chunks.ts
+
+Tool approval types (`ToolApprovalRequestChunk`, `ToolOutputDeniedChunk`) require AI SDK v6 or later.
```
DouweM: Let's add this to the docstrings for the types in question

bendrucker:
Thanks for the quick reviews! Will get the test issue fixed and reply to your comments shortly.

docs/ui/vercel-ai.md
bendrucker: The naming is a bit confusing, I can revise the comment, but:

https://ai-sdk.dev/docs/ai-sdk-ui

AI SDK (https://github.com/vercel/ai) is composed of:

* Core (server)
* UI (client)
* RSC (hybrid)

AI SDK UI offers data hooks for React/Svelte/Vue. It takes the streaming protocol and turns it into a list of messages the UI can render.

AI Elements (https://github.com/vercel/ai-elements) is React components for rendering those messages.

The adapter is integrating with AI SDK UI.

The AI Elements `<Tool>` component does have support for approval states and actions:

https://ai-sdk.dev/elements/components/tool

docs/ui/vercel-ai.md
DouweM, line 1: I'll take a closer look at docs later, will focus primarily on the code for now

pydantic_ai_slim/pydantic_ai/ui/_adapter.py
```diff
@@ -125,6 +125,12 @@ class UIAdapter(ABC, Generic[RunInputT, MessageT, EventT, AgentDepsT, OutputData
     accept: str | None = None
     """The `Accept` header value of the request, used to determine how to encode the protocol-specific events for the streaming response."""
 
+    tool_approval: bool = False
+    """Whether to enable tool approval streaming for human-in-the-loop workflows."""
+
+    deferred_tool_results: DeferredToolResults | None = None
```
DouweM: This should be an overridable method like `toolset`

pydantic_ai_slim/pydantic_ai/ui/_adapter.py
```diff
@@ -125,6 +125,12 @@ class UIAdapter(ABC, Generic[RunInputT, MessageT, EventT, AgentDepsT, OutputData
     accept: str | None = None
     """The `Accept` header value of the request, used to determine how to encode the protocol-specific events for the streaming response."""
 
+    tool_approval: bool = False
```
DouweM: Please add an `enable_` prefix so it's clear at a glance that this is a "feature flag"

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
```diff
@@ -78,9 +81,51 @@ def build_run_input(cls, body: bytes) -> RequestData:
...
+
+        Args:
+            request: The incoming Starlette/FastAPI request.
+            agent: The Pydantic AI agent to run.
+            tool_approval: Whether to enable tool approval streaming for human-in-the-loop workflows.
+        """
+        run_input = cls.build_run_input(await request.body())
+
+        # Extract deferred tool results from messages when tool_approval is enabled
+        deferred_tool_results: DeferredToolResults | None = None
```
DouweM: As mentioned above, let's not override `from_request`, just a new `deferred_tool_results` method

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
```diff
@@ -78,9 +81,51 @@ def build_run_input(cls, body: bytes) -> RequestData:
...
+        if tool_approval:
+            approvals: dict[str, bool | DeferredToolApprovalResult] = {}
```
DouweM: I don't think we need `DeferredToolApprovalResult` here as we only pass booleans

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
```diff
@@ -78,9 +81,51 @@ def build_run_input(cls, body: bytes) -> RequestData:
...
+            for msg in run_input.messages:
+                if msg.role != 'assistant':
+                    continue
+                for part in msg.parts:
+                    if not isinstance(part, ToolUIPart | DynamicToolUIPart):
+                        continue
+                    approval = part.approval
+                    if not isinstance(approval, ToolApprovalResponded):
+                        continue
+                    approvals[part.tool_call_id] = approval.approved
```
DouweM: I like dict comprehensions:

```py
approvals = {
    part.tool_call_id: part.approval.approved
    for msg in run_input.messages
    if msg.role == "assistant"
    for part in msg.parts
    if isinstance(part, (ToolUIPart, DynamicToolUIPart))
    and isinstance(part.approval, ToolApprovalResponded)
}
```

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_event_stream.py
```diff
@@ -78,6 +87,20 @@ class VercelAIEventStream(UIEventStream[RequestData, BaseChunk, AgentDepsT, Outp
     _step_started: bool = False
     _finish_reason: FinishReason = None
 
+    @cached_property
+    def _denied_tool_ids(self) -> set[str]:
+        """Get the set of tool_call_ids that were denied by the user."""
+        denied_ids: set[str] = set()
```
DouweM: Like above, this can likely be a set comprehension!

DouweM: Maybe this and the one above could be helper methods on `RunInput`, since they share so much logic it feels like they should be in the same place!

tests/test_vercel_ai.py
```diff
@@ -1651,6 +1652,359 @@ async def on_complete(run_result: AgentRunResult[Any]) -> AsyncIterator[BaseChun
...
+        ],
+    )
+
+    adapter = VercelAIAdapter(agent, request, tool_approval=True)
+    events: list[str | dict[str, Any]] = [
+        '[DONE]' if '[DONE]' in event else json.loads(event.removeprefix('data: '))
+        async for event in adapter.encode_stream(adapter.run_stream())
+    ]
+
+    # Verify tool-approval-request chunk is emitted with UUID approval_id
```
DouweM: Can we please use snapshots, as we do in all the existing tests?

tests/test_vercel_ai.py
```diff
@@ -1651,6 +1652,359 @@ async def on_complete(run_result: AgentRunResult[Any]) -> AsyncIterator[BaseChun
...
+
+    adapter = await VercelAIAdapter[None, str | DeferredToolRequests].from_request(
+        mock_request, agent=agent, tool_approval=True
+    )
+    events: list[str | dict[str, Any]] = [
+        '[DONE]' if '[DONE]' in event else json.loads(event.removeprefix('data: '))
+        async for event in adapter.encode_stream(adapter.run_stream())
+    ]
+
+    # Verify tool-output-denied chunk is emitted
```
DouweM: Here and in the other cases, snapshots please! I'd also love to see the tool output available event as well for example

tests/test_vercel_ai.py
```diff
@@ -1651,6 +1652,359 @@ async def on_complete(run_result: AgentRunResult[Any]) -> AsyncIterator[BaseChun
     assert {'type': 'data-progress', 'data': {'percent': 100}, 'transient': True} in events
 
 
+async def test_tool_approval_request_emission():
+    """Test that ToolApprovalRequestChunk is emitted when tools require approval."""
+    from pydantic_ai.tools import DeferredToolRequests
```
DouweM: Imports up top please

tests/test_vercel_ai.py
```diff
@@ -1651,6 +1652,359 @@ async def on_complete(run_result: AgentRunResult[Any]) -> AsyncIterator[BaseChun
...
+    """Test that ToolOutputDeniedChunk is emitted when a tool call is denied.
+
+    This test verifies the full public interface: from_request() extracts approval
+    data from messages, and the adapter emits tool-output-denied chunks for denied tools.
+    """
+    from unittest.mock import AsyncMock
+
+    from starlette.requests import Request
+
+    from pydantic_ai.tools import DeferredToolRequests
```
DouweM: Imports up top

tests/test_vercel_ai.py
```diff
@@ -1651,6 +1652,359 @@ async def on_complete(run_result: AgentRunResult[Any]) -> AsyncIterator[BaseChun
...
+        ],
+    )
+
+    def mock_header_get(key: str) -> str | None:
+        return None
+
+    request_body = request.model_dump_json().encode()
+    mock_request = AsyncMock(spec=Request)
+    mock_request.body = AsyncMock(return_value=request_body)
+    mock_request.headers.get = mock_header_get
```
DouweM: It shouldn't be necessary to mock a request, we can build a `VercelAIAdapter` directly with a `SubmitMessage`

tests/test_vercel_ai.py
```diff
@@ -1651,6 +1652,359 @@ async def on_complete(run_result: AgentRunResult[Any]) -> AsyncIterator[BaseChun
...
```
DouweM: No mocking please

tests/test_vercel_ai.py
```diff
@@ -1651,6 +1652,359 @@ async def on_complete(run_result: AgentRunResult[Any]) -> AsyncIterator[BaseChun
...
+    # Verify tool-output-denied chunk is emitted
+    denied_event: dict[str, Any] | None = next(
+        (e for e in events if isinstance(e, dict) and e.get('type') == 'tool-output-denied'),
+        None,
+    )
+    assert denied_event is not None
+    assert denied_event['toolCallId'] == 'delete_1'
+
+
+@pytest.mark.skipif(not starlette_import_successful, reason='Starlette is not installed')
```
DouweM: These tests shouldn't need to require Starlette

bendrucker:
Have been on a break for a bit, will get the comments addressed early this week if not today.

DouweM:
@bendrucker Thanks Ben, I should probably have disabled the stale bot before the holidays :)

pydantic_ai_slim/pydantic_ai/ui/_adapter.py
```diff
@@ -125,6 +125,9 @@ class UIAdapter(ABC, Generic[RunInputT, MessageT, EventT, AgentDepsT, OutputData
     accept: str | None = None
     """The `Accept` header value of the request, used to determine how to encode the protocol-specific events for the streaming response."""
 
+    enable_tool_approval: bool = False
```
DouweM: Let's move this to `VercelAI` as it's the only one that implements it for now.

pydantic_ai_slim/pydantic_ai/ui/_adapter.py
```diff
@@ -237,6 +245,10 @@ def run_stream_native(
             toolsets: Optional additional toolsets for this run.
             builtin_tools: Optional additional builtin tools to use for this run.
         """
+        # Use cached property as fallback if not explicitly passed
```
DouweM: Please remove the comment and move this after the `if toolset` bit

pydantic_ai_slim/pydantic_ai/ui/_event_stream.py
```diff
@@ -70,6 +70,9 @@ class UIEventStream(ABC, Generic[RunInputT, EventT, AgentDepsT, OutputDataT]):
     accept: str | None = None
     """The `Accept` header value of the request, used to determine how to encode the protocol-specific events for the streaming response."""
 
+    enable_tool_approval: bool = False
```
DouweM: Only on VercelAIEventStream for now, please

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
```diff
@@ -61,8 +62,9 @@
 from .response_types import BaseChunk
 
 if TYPE_CHECKING:
-    pass
+    from starlette.requests import Request
 
+    from ...agent import AbstractAgent
```
DouweM, line 69: Doesn't depend on an optional dependency so should be at the top

tests/test_vercel_ai.py
```diff
@@ -1651,6 +1655,313 @@ async def on_complete(run_result: AgentRunResult[Any]) -> AsyncIterator[BaseChun
...
+                            reason='User cancelled the deletion',
+                        ),
+                    ),
+                ],
+            ),
+        ],
+    )
+
+    adapter = VercelAIAdapter(agent, request, enable_tool_approval=True)
+    events: list[str | dict[str, Any]] = [
```
DouweM, line 2385: Please include a snapshot of the resulting `all_messages()` on the result

tests/test_vercel_ai.py
```diff
@@ -1651,6 +1655,313 @@ async def on_complete(run_result: AgentRunResult[Any]) -> AsyncIterator[BaseChun
...
+        '[DONE]' if '[DONE]' in event else json.loads(event.removeprefix('data: '))
+        async for event in adapter.encode_stream(adapter.run_stream())
+    ]
+
+    assert events == snapshot(
+        [
+            {'type': 'start'},
+            {'type': 'tool-output-denied', 'toolCallId': 'delete_1'},
+            {'type': 'tool-output-available', 'toolCallId': 'delete_approved', 'output': 'Deleted approved.txt'},
+            {'type': 'start-step'},
```
DouweM, line 2395: Should the two tool output events be in their own start-step/finish-step? or in this one? Would be good to verify that the Vercel SDK does.

tests/test_vercel_ai.py
```diff
@@ -1651,6 +1655,313 @@ async def on_complete(run_result: AgentRunResult[Any]) -> AsyncIterator[BaseChun
...
+                ],
+            ),
+        ],
+    )
+
+    adapter = VercelAIAdapter(agent, request, enable_tool_approval=True)
+
+    # Verify that only the responded approval was extracted
+    assert adapter.deferred_tool_results is not None
+    assert adapter.deferred_tool_results.approvals == {'approved_tool': True}
```
DouweM, line 2503: In all these cases I'd also like to see the Pydantic AI messages this gets turned into, i.e. `adapter.message_history`, or after a real run, `result.all_messages()`

DouweM (changes requested):

tests/test_vercel_ai.py
bendrucker: AI SDK test showing tool output denied before start step:

https://github.com/vercel/ai/blob/483dc52858612c29552a741e1be6210fef249171/packages/ai/src/generate-text/stream-text.test.ts#L19283-L19320

bendrucker:
Thanks for your patience! Made the requested changes and provided a reference to relevant AI SDK test snapshots for the behavioral question.

bendrucker:
Squashing another upstream cause of a test flake: https://github.com/pytest-dev/pytest-xdist/pull/1299

DouweM:
@bendrucker I merged some other changes for the Vercel AI event stream that had been in the works for a few weeks -- can you resolve the conflicts please?

bendrucker:
Resolved!

docs/ui/vercel-ai.md
```diff
@@ -81,3 +81,27 @@ async def chat(request: Request) -> Response:
     sse_event_stream = adapter.encode_stream(event_stream)
     return StreamingResponse(sse_event_stream, media_type=accept)
 ```
+
+## Tool Approval
+
+!!! note
+    Tool approval requires AI SDK UI v6 or later on the frontend.
+
+Pydantic AI supports human-in-the-loop tool approval workflows with AI SDK UI, allowing users to approve or deny tool executions before they run. See the [deferred tool calls documentation](../deferred-tools.md) for details on setting up tools that require approval.
```
DouweM:
```suggestion
Pydantic AI supports human-in-the-loop tool approval workflows with AI SDK UI, allowing users to approve or deny tool executions before they run. See the [deferred tool calls documentation](../deferred-tools.md#human-in-the-loop-tool-approval) for details on setting up tools that require approval.
```

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
```diff
@@ -77,14 +78,54 @@
...
+
+        Args:
+            request: The incoming Starlette/FastAPI request.
+            agent: The Pydantic AI agent to run.
+            enable_tool_approval: Whether to enable tool approval streaming for human-in-the-loop workflows.
+        """
+        return cls(
+            agent=agent,
+            run_input=cls.build_run_input(await request.body()),
+            accept=request.headers.get('accept'),
```
DouweM: We shouldn't be repeating super class implementation here; can we call the super method and then set `enable_tool_approval` directly on the returned instance?

We could also choose not to have this method override at all, and just tell the user to set the flag like that directly.

DouweM: Note that if we override methods, we should also support this flag on `dispatch_request`, as that's the main way people use these UI adapters. If that's hard to do, perhaps it'd be better to always have the user build the adapter, set the flag, and then call `streaming_response` as in your example.

I also wonder if there's a way we can detect from the incoming API request whether the frontend is using AI SDK UI v6 or later? In that case we may not need the flag at all.

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
```diff
@@ -77,14 +78,54 @@
...
         """Build a Vercel AI event stream transformer."""
-        return VercelAIEventStream(self.run_input, accept=self.accept)
+        return VercelAIEventStream(self.run_input, accept=self.accept, enable_tool_approval=self.enable_tool_approval)
+
+    @cached_property
+    def deferred_tool_results(self) -> DeferredToolResults | None:
+        """Extract deferred tool results from Vercel AI messages with approval responses."""
+        if not self.enable_tool_approval:
+            return None
+        approvals: dict[str, bool | DeferredToolApprovalResult] = {}
```
DouweM, line 169: https://github.com/pydantic/pydantic-ai/pull/3772#discussion_r2641223390

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
```diff
@@ -77,14 +78,54 @@
...
+        for msg in self.run_input.messages:
```
DouweM: https://github.com/pydantic/pydantic-ai/pull/3772#discussion_r2641227984

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_event_stream.py
```diff
@@ -76,9 +85,27 @@ def _json_dumps(obj: Any) -> str:
...
+        denied: set[str] = set()
+        for msg in self.run_input.messages:
+            if msg.role == 'assistant':
+                for part in msg.parts:
+                    if (
+                        isinstance(part, ToolUIPart | DynamicToolUIPart)
+                        and isinstance(part.approval, ToolApprovalResponded)
+                        and not part.approval.approved
+                    ):
+                        denied.add(part.tool_call_id)
```
DouweM: https://github.com/pydantic/pydantic-ai/pull/3772#discussion_r2641232257

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/request_types.py
```diff
@@ -1,7 +1,9 @@
 """Vercel AI request types (UI messages).
 
 Converted to Python from:
-https://github.com/vercel/ai/blob/ai%405.0.59/packages/ai/src/ui/ui-messages.ts
+https://github.com/vercel/ai/blob/ai%406.0.0-beta.159/packages/ai/src/ui/ui-messages.ts
```
DouweM: Is it still beta?

tests/test_vercel_ai.py
```diff
@@ -1923,6 +1928,399 @@ async def on_complete(run_result: AgentRunResult[Any]) -> AsyncIterator[BaseChun
...
+                        approval=ToolApprovalResponded(id='approval-456', approved=True),
+                    ),
+                    DynamicToolInputAvailablePart(
+                        tool_name='delete_file',
+                        tool_call_id='delete_1',
+                        input={'path': 'test.txt'},
+                        approval=ToolApprovalResponded(
+                            id='approval-123',
+                            approved=False,
+                            reason='User cancelled the deletion',
```
DouweM, line 2369: Should this be used where it says "The tool call was denied." below?

tests/test_vercel_ai.py
```diff
@@ -1923,6 +1928,399 @@ async def on_complete(run_result: AgentRunResult[Any]) -> AsyncIterator[BaseChun
...
+        '[DONE]' if '[DONE]' in event else json.loads(event.removeprefix('data: '))
+        async for event in adapter.encode_stream(adapter.run_stream(deferred_tool_results=explicit_results))
+    ]
+
+    assert events == snapshot(
+        [
+            {'type': 'start'},
+            {
+                'type': 'error',
+                'errorText': 'Tool call results were provided, but the message history does not contain a `ModelResponse`.',
```
DouweM, line 2611: Let's test in a scenario where this doesn't generate an error :)

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
bendrucker: > I also wonder if there's a way we can detect from the incoming API request whether the frontend is using AI SDK UI v6 or later? In that case we may not need the flag at all.

Good call, I'll look into whether they have a protocol version or whether they do feature detection based on the requests.

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/request_types.py
bendrucker: Released! 🎉 

https://vercel.com/blog/ai-sdk-6

AFAICT there are no relevant changes from the beta, I'll get the docs updated

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
bendrucker: The extent of explicit versioning is the `x-vercel-ai-ui-message-stream: v1` header. It comes from the server. There's no `Accept` header or similar negotiation from client to server.

The client is strict about what the server sends, only known chunk types, only known keys:

https://github.com/vercel/ai/blob/7839e79ddab8e095691a395853d97cbc6df85073/packages/ai/src/ui-message-stream/ui-message-chunks.ts#L16-L180

Any validation error against this schema interrupts streaming. 

There's nothing we can detect in the client requests on the initial interaction. We'd send a tool approval request to the client and if it's v5 it just would never be able to provide an `approval-responded` part.

DouweM:
@bendrucker Please have a look at the conflicts

DouweM:
@bendrucker A few of my earlier comments remain unresolved, right?
- https://github.com/pydantic/pydantic-ai/pull/3772#discussion_r2733930345
- https://github.com/pydantic/pydantic-ai/pull/3772#discussion_r2641223390
- https://github.com/pydantic/pydantic-ai/pull/3772#discussion_r2641232257

bendrucker:
Hmm I thought I remember addressing those but I will take a look later and see what happened to that work.

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
bendrucker: At the time this was written, you were right — only booleans were being passed because I wasn't propagating denial reasons. Now when a denial includes a `reason`, we pass `ToolDenied(message=reason)` instead of a bare `False`. Approvals and reason-less denials still use plain booleans.

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
bendrucker: Went with this — removed the `from_request` override entirely so there's no Vercel-specific signature. Users set `enable_tool_approval` on the adapter instance after construction.

bendrucker:
Resolved:

1. https://github.com/pydantic/pydantic-ai/pull/3772#discussion_r2641223390 — We pass `ToolDenied(message=...)` when there's a denial reason, `bool` otherwise. `DeferredToolApprovalResult` was unused before because denial reasons weren't propagated.
2. https://github.com/pydantic/pydantic-ai/pull/3772#discussion_r2641227984 — `deferred_tool_results` and `iter_tool_approvals` use comprehensions.
3. https://github.com/pydantic/pydantic-ai/pull/3772#discussion_r2641232257 — Extracted `iter_tool_approvals` into `request_types.py`, shared by both adapter and event stream.
4. https://github.com/pydantic/pydantic-ai/pull/3772#discussion_r2733930345 — Removed the `from_request` override entirely. Users set `enable_tool_approval` on the instance directly.

DouweM:
@bendrucker Once https://github.com/pydantic/pydantic-ai/pull/4166 is merged (should be shortly), please update to use the sdk version flag to auto-enable this instead of the new boolean. You can mention in the docstring for the field that it enables approval handling.

bendrucker:
Merged! Totally forgot that somewhere in this PR I investigated the frontend side and found strict schema handling via zod, with errors on unknown keys. Not sure that's sensible/necessary behavior on AI SDK's part, but given that reality, a version attribute is definitely the right move for Pydantic AI. Certainly simplifies things here.

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
```diff
@@ -77,17 +78,62 @@ class VercelAIAdapter(UIAdapter[RequestData, UIMessage, BaseChunk, AgentDepsT, O
...
+            request: The incoming Starlette/FastAPI request.
+            agent: The Pydantic AI agent to run.
+            sdk_version: Vercel AI SDK version. Set to 6 to enable tool approval streaming.
+        """
+        return cls(
+            agent=agent,
+            run_input=cls.build_run_input(await request.body()),
+            accept=request.headers.get('accept'),
+            sdk_version=sdk_version,
+        )
```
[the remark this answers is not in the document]
bendrucker: Non-issue. `dispatch_request` is a convenience shorthand for simple one-shot runs. Tool approval is inherently multi-turn. The documented pattern is `from_request()` + `run_stream()`, which already accepts `sdk_version`. The alternatives (overriding `dispatch_request` to duplicate ~15 params, `**kwargs` forwarding, base class refactor) are all worse than the status quo.

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/request_types.py
```diff
@@ -110,6 +112,30 @@ class DataUIPart(BaseUIPart):
...
+
+    approved: bool
+    """Whether the user approved the tool call."""
+
+    reason: str | None = None
+    """Optional reason for the approval or denial."""
+
+
+ToolApproval = ToolApprovalRequested | ToolApprovalResponded
+"""Union of tool approval states."""
```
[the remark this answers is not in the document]
bendrucker: Non-issue. `CamelBaseModel` uses `extra='forbid'`, so `{"id": "x", "approved": true}` will fail validation against `ToolApprovalRequested` (extra field rejected) and fall through to `ToolApprovalResponded`. The discrimination is correct and robust without an explicit discriminator. Adding a `type` field would deviate from the upstream Vercel AI SDK types.

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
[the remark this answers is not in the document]
DouweM: > dispatch_request is a convenience shorthand for simple one-shot runs

@bendrucker That's not right; dispatch_request is the primary method through which these classes are used: https://ai.pydantic.dev/ui/vercel-ai/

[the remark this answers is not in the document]
bendrucker: 🤦‍♂️ was definitely up too late on this one

I believe we went back and forth on manually constructing and overriding vs re-implementing superclass implementation details. Now that it's `sdk_version` and not coupled to this specific feature gate it makes more sense to keep the high level interface, it'll be stable.

04ecd93 restores `dispatch_request`

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
```diff
@@ -77,17 +88,154 @@ class VercelAIAdapter(UIAdapter[RequestData, UIMessage, BaseChunk, AgentDepsT, O
...
+                model_settings=model_settings,
+                usage_limits=usage_limits,
+                usage=usage,
+                metadata=metadata,
+                infer_name=infer_name,
+                toolsets=toolsets,
+                builtin_tools=builtin_tools,
+                on_complete=on_complete,
+            ),
+        )
```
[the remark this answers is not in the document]
DouweM: Hmm yeah, `**kwargs` on `UIAdapter.dispatch_request` seems like the least-bad option then. I'd still want to repeat the method definition at this level so that we get proper IDE autocomplete for `sdk_version`, but at least the method body could be a simple `super()` call.

akshan-main:
Hey, I worked on a similar fix in #4283 (closed as a duplicate of this) and found a different approach to the dispatch_request duplication issue.

Instead of overriding dispatch_request (~95 lines), I override run_stream_native() to inject deferred_tool_results:


def run_stream_native(self, *, deferred_tool_results=None, **kwargs):
    if deferred_tool_results is None:
        deferred_tool_results = self.deferred_tool_results
    return super().run_stream_native(deferred_tool_results=deferred_tool_results, **kwargs)

This avoids duplicating the base class logic entirely. Happy to open a separate PR for just this pattern if it's useful.

bendrucker:
Ok, all comments are addressed! Back to `kwargs` + `super` for the method overrides.

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
```diff
@@ -77,17 +87,132 @@ class VercelAIAdapter(UIAdapter[RequestData, UIMessage, BaseChunk, AgentDepsT, O
...
+            agent: The Pydantic AI agent to run.
+            sdk_version: Vercel AI SDK version. Set to 6 to enable tool approval streaming.
+            **kwargs: Additional keyword arguments (unused, accepted for forward compatibility).
+        """
+        return cls(
+            agent=agent,
+            run_input=cls.build_run_input(await request.body()),
+            accept=request.headers.get('accept'),
+            sdk_version=sdk_version,
+        )
```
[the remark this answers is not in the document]
DouweM: I want the IDE autocomplete

DouweM:
@bendrucker Did you mean to push your changes?

bendrucker:
Whoops, done!

DouweM:
@bendrucker We've got merge conflicts 😅

bendrucker:
No worries, saw em, fixing them now!

DouweM:
@bendrucker Thanks Ben! It took a minute but I'll get this out in a release later today :)

bendrucker:
Thank you! Appreciate your patience, Python is still somewhat new to me. Learned a bunch and will be eager to integrate this soon!

PR #3775, opened 2025-12-19
Allow dynamic TemporalWrapperToolset in Temporal workflows

### LINKED ISSUE #3768: Temporal: allow cross agent tool registration
### Description

In scenarios where agents need to be defined dynamically ([example](https://github.com/pydantic/pydantic-ai/issues/3506)), the list of agent tools cannot be registered by the agent when the temporal worker is created, since the agent is not yet created.

Please decouple tool registration and naming from agent definition and naming in order to allow multiple agents (and dynamically created agents) within a temporal workflow to use the same set of registered tools. A master list of tools would be registered with the temporal workflow at the time of workflow creation, and dynamically created agents could refer to and use the subset of those tools as needed. 

For example, if the tool naming (currently something like `agent__{name}__tool__{tool_name}`) were to not include the agent name and instead be a workflow-wide tool, then dynamic agents could refer to the tool by its name/slug.

It would be nice to have a helper function for temporal workflow tool registration as well.

### References

From the [slack conversation](https://pydanticlogfire.slack.com/archives/C083V7PMHHA/p1766035426002909) with @DouweM : 

> That would be pretty easy to make work -- you can already create a `FunctionToolset` and then manually wrap it in a `TemporalFunctionToolset` (currently private) and then pass the `temporal_function_toolset.temporal_activities`  to the worker. The only part that isn't done yet is letting you pass those `TemporalFunctionToolset`s to `TemporalAgent.run(toolsets=...)`, which would require replacing this error with a check that all the toolsets are properly wrapped for Temporal: https://github.com/pydantic/pydantic-ai/blob/c1213e4a7a978099b64c642b38e1e041d18afad6/pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py#L898-L901

Related to:
- https://github.com/pydantic/pydantic-ai/issues/3506
- https://github.com/pydantic/pydantic-ai/pull/3537

qyinm:
Closes #3768.
## Summary
This PR enables the use of dynamic toolsets in Temporal workflows by allowing `toolsets` to be passed to `agent.run()` at runtime, provided they are instances of `TemporalWrapperToolset`.

## Changes
- **Modified `TemporalAgent`**: Updated `_temporal_overrides` to accept runtime toolsets if they are properly wrapped.
- **Added Test**: Created `tests/test_temporal_dynamic.py` to verify that a `TemporalFunctionToolset` can be successfully passed and executed dynamically within a workflow.

## Rationale
As discussed in #3768, this change allows for:
- Cross-agent tool registration.
- More flexible dynamic agent patterns.
- Strict enforcement of Temporal's determinism requirements (by ensuring all tools are executed as Activities).

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
+++ b/pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
@@ -374,8 +374,11 @@ class TemporalAgent(WrapperAgent[AgentDepsT, OutputDataT]):
                 # Already a temporal wrapper: use as-is
                 resolved_toolsets.append(t)
             else:
-                # Original toolset instance: find its temporal wrapper using object id
-                # This works for CombinedToolset and other composite toolsets
+                # Toolset instance that's not a direct TemporalWrapperToolset.
+                # This includes:
+                # - Original toolsets that need to be looked up in our mapping
+                # - CombinedToolset with temporal children (validated later via _validate_temporal_toolsets)
+                # - User-constructed composite toolsets
                 temporalized = self._original_to_temporal.get(id(t))
                 if temporalized is not None:
                     resolved_toolsets.append(temporalized)
--- a/tests/test_temporal.py
+++ b/tests/test_temporal.py
@@ -3397,13 +3397,16 @@ def test_resolve_toolsets_logic():
     # 3. Already a wrapper
     assert agent._resolve_toolsets([w1]) == [w1]  # pyright: ignore[reportPrivateUsage]
 
-    # 4. Original instance auto-resolution - wrapper found
-    # w1 wraps t1, so passing t1 should return w1 if it's in named_toolsets
+    # 4. Original instance auto-resolution
+    # When w1 (already a TemporalWrapperToolset) is registered, t1 is NOT tracked in _original_to_temporal
+    # because w1 is what gets stored, not t1's id -> w1 mapping
+    # This is the expected behavior: only toolsets passed directly to TemporalAgent are tracked
     result = agent_with_tools._resolve_toolsets([t1])  # pyright: ignore[reportPrivateUsage]
     assert result is not None
-    assert result == [w1]
+    # t1 is returned as-is since it's not in _original_to_temporal
+    assert result == [t1]
     assert len(result) == 1
-    assert isinstance(result[0], MockWrapper)
+    assert result[0] is t1
 
     # 5. Original instance not found - returns as-is
     t2 = FunctionToolset(tools=[], id='t2')
@@ -3412,3 +3415,104 @@ def test_resolve_toolsets_logic():
     assert result == [t2]
     assert len(result) == 1
     assert result[0] is t2
+
+
+def test_get_toolset_types_requiring_temporal_wrapper():
+    """Test the centralized function that returns toolset types requiring temporal wrapping."""
+    from pydantic_ai.durable_exec.temporal._toolset import (
+        get_toolset_types_requiring_temporal_wrapper,
+    )
+
+    types = get_toolset_types_requiring_temporal_wrapper()
+
+    # Should return a tuple
+    assert isinstance(types, tuple)
+
+    # Should include FunctionToolset and DynamicToolset at minimum
+    from pydantic_ai.toolsets._dynamic import DynamicToolset
+
+    assert FunctionToolset in types
+    assert DynamicToolset in types
+
+    # MCPServer and FastMCPToolset should also be included (assuming they're importable)
+    try:
+        from pydantic_ai.mcp import MCPServer
+
+        assert MCPServer in types
+    except ImportError:
+        pass
+
+    try:
+        from pydantic_ai.toolsets.fastmcp import FastMCPToolset
+
+        assert FastMCPToolset in types
+    except ImportError:
+        pass
+
+
+def test_validate_temporal_toolsets_combined_toolset():
+    """Test that CombinedToolset with temporal children passes validation."""
+    from collections.abc import Callable
+    from typing import Any
+
+    from pydantic_ai import FunctionToolset
+    from pydantic_ai.durable_exec.temporal import TemporalWrapperToolset
+    from pydantic_ai.durable_exec.temporal._agent import (
+        _validate_temporal_toolsets,  # pyright: ignore[reportPrivateUsage]
+    )
+    from pydantic_ai.toolsets import CombinedToolset
+
+    class MockTemporalWrapper(TemporalWrapperToolset):
+        @property
+        def temporal_activities(self) -> list[Callable[..., Any]]:
+            return []  # pragma: no cover
+
+    def tool_func(x: int) -> int:
+        return x  # pragma: no cover
+
+    # Create wrapped toolsets
+    func_toolset_1 = FunctionToolset(tools=[tool_func], id='test_func_1')
+    func_toolset_2 = FunctionToolset(tools=[tool_func], id='test_func_2')
+    wrapper_1 = MockTemporalWrapper(func_toolset_1)
+    wrapper_2 = MockTemporalWrapper(func_toolset_2)
+
+    # CombinedToolset with all temporal children should pass
+    combined = CombinedToolset([wrapper_1, wrapper_2])
+    _validate_temporal_toolsets([combined])
+
+    # CombinedToolset with one unwrapped child should fail
+    func_toolset_unwrapped = FunctionToolset(tools=[tool_func], id='test_func_unwrapped')
+    combined_with_unwrapped = CombinedToolset([wrapper_1, func_toolset_unwrapped])
+    with pytest.raises(UserError, match='must be wrapped in a `TemporalWrapperToolset`'):
+        _validate_temporal_toolsets([combined_with_unwrapped])
+
+
+def test_resolve_toolsets_with_original_to_temporal_mapping():
+    """Test that _original_to_temporal mapping correctly tracks toolsets by object id."""
+    from pydantic_ai import Agent, FunctionToolset
+    from pydantic_ai.durable_exec.temporal import TemporalAgent
+    from pydantic_ai.toolsets import CombinedToolset
+
+    base_agent = Agent(TestModel(), name='test-mapping-agent')
+
+    # Create toolsets
+    t1 = FunctionToolset(tools=[], id='t1')
+    t2 = FunctionToolset(tools=[], id='t2')
+
+    # Create a CombinedToolset
+    combined = CombinedToolset([t1, t2])
+
+    # Create agent with the combined toolset registered by name
+    # Note: In real usage, temporalize_toolset would wrap the children
+    # For this test, we're testing the id-based mapping mechanism
+    agent = TemporalAgent(base_agent, toolsets={'my_combined': combined})
+
+    # The original combined toolset should be tracked by object id
+    assert id(combined) in agent._original_to_temporal  # pyright: ignore[reportPrivateUsage]
+
+    # Resolving the original combined toolset should return the temporalized version
+    result = agent._resolve_toolsets([combined])  # pyright: ignore[reportPrivateUsage]
+    assert result is not None
+    assert len(result) == 1
+    # The result should be the temporalized version stored in _original_to_temporal
+    assert result[0] is agent._original_to_temporal[id(combined)]  # pyright: ignore[reportPrivateUsage]


pydantic_ai_slim/pydantic_ai/durable_exec/temporal/__init__.py
```diff
@@ -14,6 +14,7 @@
 
 from ...exceptions import UserError
 from ._agent import TemporalAgent
+from ._function_toolset import TemporalFunctionToolset
```
DouweM: This feature should work for any toolset, so we need to export the other wrappers (MCP etc) as well

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
```diff
@@ -242,10 +242,21 @@ def temporal_activities(self) -> list[Callable[..., Any]]:
         return self._temporal_activities
 
     @contextmanager
-    def _temporal_overrides(self) -> Iterator[None]:
-        # We reset tools here as the temporalized function toolset is already in self._toolsets.
+    def _temporal_overrides(self, toolsets: Sequence[AbstractToolset[AgentDepsT]] | None = None) -> Iterator[None]:
+        in_workflow = workflow.in_workflow()
+
+        if toolsets:
+            if in_workflow and any(not isinstance(t, TemporalWrapperToolset) for t in toolsets):
```
DouweM: - `CombinedToolset`s and others that don't require temporal wrappers are fine, so we should look at the ones covered by `temporalize_toolset`
- We should look down the hierarchy of toolsets using `toolset.visit()`, if we find a temporal wrapper, we know we don't need to look at its children, but otherwise we error if we encountered an unwrapped toolset

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
```diff
@@ -895,9 +906,9 @@ async def main():
                 raise UserError(
                     'Model cannot be set at agent run time inside a Temporal workflow, it must be set at agent creation time.'
                 )
-            if toolsets is not None:
+            if toolsets is not None and any(not isinstance(t, TemporalWrapperToolset) for t in toolsets):
```
DouweM: Same as above; the check is not precise enough

tests/test_temporal_dynamic.py
DouweM, line 1: This should be in the existing test file

tests/test_temporal_dynamic.py
```diff
@@ -0,0 +1,96 @@
...
+# though here we pass the toolset object directly.
+toolset = FunctionToolset(tools=[echo], id='my_tools')
+
+# 3. Wrap Toolset for Temporal (DouweM pattern)
+wrapped_toolset = TemporalFunctionToolset(
+    toolset,
+    activity_name_prefix='shared_tools',
+    activity_config=ActivityConfig(start_to_close_timeout=timedelta(minutes=1)),
+    tool_activity_config={},
+    deps_type=type(None),
```
DouweM: These 4 args should not be required

tests/test_temporal_dynamic.py
```diff
@@ -0,0 +1,96 @@
...
+)
+
+# 4. Create base agent for model activity registration
+# This agent's activities will be registered in the Worker
+# We use a known name "test_agent" so the dynamic agent can share it.
+base_model = TestModel()
+base_agent = Agent(base_model, name='test_agent')
+base_temporal_agent = TemporalAgent(base_agent)
```
DouweM: I want to also support the same kind of pre-registering we're adding in https://github.com/pydantic/pydantic-ai/pull/3537 for models, so you'd pass `toolsets={'funcs': toolset, 'other': ...}` and then you can pass `run(toolsets=['funcs'])` or `run(toolsets=[toolset])` later, in addition to the instance you've pre-wrapped yourself.

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/__init__.py
DouweM, line 1: We'll need docs! See https://github.com/pydantic/pydantic-ai/pull/3537 for an example

tests/test_temporal_dynamic.py
DouweM: In this scenario, we should make sure that `TemporalAgent.toolset_activity_config` and `TemporalAgent.tools_activity_config` work for toolsets that are not manually wrapped like that, as otherwise the user wouldn't have a way of specifying activity config

qyinm:
@DouweM  I've addressed all review feedback:

- Toolset pre-registration implemented (similar to #3537)
- Validation improved with visit_and_replace
- Documentation and tests added

Ready for re-review.

qyinm:
@DouweM Update based on feedback

docs/durable_execution/temporal.md
DouweM, line 1: Note to self: will review docs later

docs/durable_execution/temporal.md
```diff
@@ -264,6 +264,179 @@ class MultiModelWorkflow:
...
+
+To use toolsets at runtime with `TemporalAgent`, you need to:
+
+1. Wrap the toolset using the appropriate `TemporalWrapperToolset` class
+2. Register its activities with the Temporal Worker
+3. Pass the wrapped toolset at runtime via `run(toolsets=[...])`
+
+This pattern allows multiple agents to share the same set of tools without duplicating activity registrations, enabling dynamic agent creation while maintaining proper tool registration.
+
+Alternatively, you can pre-register toolsets with the `TemporalAgent` constructor and reference them by name at runtime. This is similar to how models are handled.
```
DouweM: I would present this as the preferred approach as it's a lot more convenient, and then explain manual wrapping as an alternative.

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/__init__.py
```diff
@@ -26,9 +29,28 @@
...
+    TemporalMCPServer = _mcp_server.TemporalMCPServer
+    __all__.append('TemporalMCPServer')
+
+try:
+    from . import _fastmcp_toolset as _fastmcp_toolset
+except ImportError:
+    pass
+else:
+    TemporalFastMCPToolset = _fastmcp_toolset.TemporalFastMCPToolset
+    __all__.append('TemporalFastMCPToolset')
```
DouweM: Hmm I'm a bit uncomfortable with this. Instead of exposing the types directly, what if we tell the user to use the `temporalize_toolset` function that intelligently wraps any toolset?

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
```diff
@@ -45,6 +46,69 @@
 from ._toolset import TemporalWrapperToolset, temporalize_toolset
 
 
+def _validate_temporal_toolsets(toolsets: Sequence[AbstractToolset[AgentDepsT]], context: str = 'at runtime') -> None:
```
DouweM: Please implement this in a more scalable way, likely by introducing a mapping of toolset types to temporal wrapper toolset types, and then also modifying `temporalize_toolset` to use it

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
```diff
@@ -45,6 +46,69 @@
...
+            if isinstance(t, FastMCPToolset):
+                raise UserError(error_msg)
+
+        # For other toolsets (like CombinedToolset, WrapperToolset, etc.),
+        # we return them unchanged - visit_and_replace will handle recursion
+        return t  # pragma: no cover - defensive code for future toolset types
+
+    # Visit and validate each toolset recursively
+    for toolset in toolsets:
+        toolset.visit_and_replace(validate_toolset)
```
DouweM: Just `visit` should do right?

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
```diff
@@ -45,6 +46,69 @@
...
+        context: A string describing the context (e.g., 'at runtime', 'contextually').
+
+    Raises:
+        UserError: If an unwrapped leaf toolset is found that requires temporal wrapping.
+    """
+    if context == 'contextually':
+        error_msg = 'Toolsets cannot be contextually overridden inside a Temporal workflow, unless they are wrapped in a `TemporalWrapperToolset`.'
+    else:
+        error_msg = (
+            'Toolsets provided at runtime inside a Temporal workflow must be wrapped in a `TemporalWrapperToolset`.'
```
DouweM: I don't love the `context` var. Can we have this method raise an error that mentions which toolset is the issue, and then where we call it, catch it and reraise it with the appropriate message prepended?

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
```diff
@@ -45,6 +46,69 @@
...
+        )
+
+    def validate_toolset(t: AbstractToolset[AgentDepsT]) -> AbstractToolset[AgentDepsT]:
+        # If we encounter a TemporalWrapperToolset, we don't need to check its children
+        # since they're already wrapped
+        if isinstance(t, TemporalWrapperToolset):  # pragma: no cover
```
DouweM: We should test this scenario. I have a hunch it won't work because `TemporalWrapperToolset.visit` currently passes the visitor onto the wrapped toolset, so we wouldn't get here. So that method may need to be overridden

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
```diff
@@ -176,6 +247,18 @@ async def streamed_response():
         activities.extend(temporal_model.temporal_activities)
         self._temporal_model = temporal_model
 
+        if toolsets:
+            if isinstance(toolsets, Mapping):
+                # Flatten the mapping for temporalization, but keep track of names
+                additional_toolsets = list(toolsets.values())
+                self._named_toolsets = toolsets
+            else:  # pragma: no cover - branch is covered but coverage is confused by the isinstance check
```
DouweM: We shouldn't need this branch, or the iinstance check

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
```diff
@@ -85,6 +150,11 @@ def __init__(
         Args:
             wrapped: The agent to wrap.
             name: Optional unique agent name to use in the Temporal activities' names. If not provided, the agent's `name` will be used.
+            toolsets:
+                Optional additional toolsets to register with the agent, or a mapping of toolset names to toolset instances.
```
DouweM: I don't think we need to support additional toolsets here that are immediately registered; just the mapping like we do for model.

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
```diff
@@ -132,6 +202,7 @@ def __init__(
         ]
         activity_config['retry_policy'] = retry_policy
         self.activity_config = activity_config
+        self._named_toolsets: Mapping[str, AbstractToolset[AgentDepsT]] | None = None
```
DouweM: Let's move this to where we use/set it

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
```diff
@@ -195,9 +278,35 @@ def temporalize_toolset(toolset: AbstractToolset[AgentDepsT]) -> AbstractToolset
...
+        # But wait, `wrapped.toolsets` are first.
+        if self._named_toolsets:
+            # The additional toolsets are at the end of `temporal_toolsets`
+            num_wrapped_toolsets = len(wrapped.toolsets)
+            # The temporalized additional toolsets
+            temporal_additional_toolsets = temporal_toolsets[num_wrapped_toolsets:]
+
+            # create a new mapping pointing to the temporalized versions
+            new_named_toolsets: dict[str, AbstractToolset[AgentDepsT]] = {}
+            for name, temporal_toolset in zip(self._named_toolsets, temporal_additional_toolsets):
```
DouweM: I don't think we need to do all this, we should be able to process the toolsets dict just once, replacing toolsets using `visit_and_replace`

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
```diff
@@ -254,21 +363,37 @@ def temporal_activities(self) -> list[Callable[..., Any]]:
...
-        When called outside a workflow, this is a no-op.
-        When called inside a workflow, it overrides the model and toolsets.
-        """
-        if not workflow.in_workflow() and not force:
-            yield
+        if toolsets:
+            if workflow.in_workflow():
+                # If toolsets are provided as strings, we can't validate them directly here as they are resolved later.
+                # We only validate if they are already AbstractToolset instances.
+                _validate_temporal_toolsets([t for t in toolsets if not isinstance(t, str)])
```
DouweM: I think we can clean this up a lot if we do the resolution etc at the callsite, instead of inside here, so that here we can just pass the toolsets through to `override` without additional checks.

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
```diff
@@ -282,6 +407,28 @@ def _temporal_overrides(
...
+        if toolsets is None:
+            return None
+
+        resolved_toolsets: list[AbstractToolset[AgentDepsT]] = []
+        for t in toolsets:
+            if isinstance(t, str):
+                if (
+                    self._named_toolsets is None
+                ):  # pragma: no cover - defensive check, _named_toolsets is always initialized
+                    raise UserError(f"Unknown toolset name: '{t}'. No named toolsets registered.")
```
DouweM: We shouldn't need this

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
```diff
@@ -282,6 +407,28 @@ def _temporal_overrides(
...
+                if t not in self._named_toolsets:
+                    raise UserError(
+                        f"Unknown toolset name: '{t}'. Available toolsets: {list(self._named_toolsets.keys())}"
+                    )
+                resolved_toolsets.append(self._named_toolsets[t])
```
DouweM: It's more Pythonic to read optimistically, and then handle `KeyError`, than to do a check-then-read like this.

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
```diff
@@ -282,6 +407,28 @@ def _temporal_overrides(
...
+            else:
+                resolved_toolsets.append(t)
```
DouweM, line 375: For models, we also accept the original toolset instances, and automatically find the appropriate wrapper. We should do that here as well

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
```diff
@@ -297,7 +444,7 @@ async def run(
         usage_limits: _usage.UsageLimits | None = None,
         usage: _usage.RunUsage | None = None,
         infer_name: bool = True,
-        toolsets: Sequence[AbstractToolset[AgentDepsT]] | None = None,
+        toolsets: Sequence[AbstractToolset[AgentDepsT] | str] | None = None,
```
DouweM, line 415: We should explain in the docstring what this means.

docs/durable_execution/temporal.md
```diff
@@ -264,6 +264,179 @@ class MultiModelWorkflow:
...
+    tools=[web_search, calculate],
+    id='shared_tools',  # (1)!
+)
+
+# Wrap the toolset for Temporal
+wrapped_shared_toolset = TemporalFunctionToolset(
+    shared_toolset,
+    activity_name_prefix='shared',  # (2)!
+    activity_config=ActivityConfig(start_to_close_timeout=timedelta(minutes=2)),
+    deps_type=type(None),  # (3)!
```
DouweM: This should not be required

docs/durable_execution/temporal.md
```diff
@@ -264,6 +264,179 @@ class MultiModelWorkflow:
...
+)
+
+# Wrap agents for Temporal (without pre-registering toolsets)
+temporal_research_agent = TemporalAgent(research_agent)
+temporal_math_agent = TemporalAgent(math_agent)
+
+
+@workflow.defn
+class SharedToolsetWorkflow:
+    __pydantic_ai_agents__ = [temporal_research_agent, temporal_math_agent]  # (4)!
```
DouweM, line 377: We need to have a convenient way to register the temporal toolsets' activities as well

docs/durable_execution/temporal.md
```diff
@@ -264,6 +264,179 @@ class MultiModelWorkflow:
...
+```
+
+1. The toolset must have a unique `id` to be used with Temporal.
+2. The `activity_name_prefix` ensures activity names don't conflict across different toolset registrations.
+3. `deps_type` must be specified if your tools use dependencies. Use `type(None)` if no dependencies are needed.
+4. The `__pydantic_ai_agents__` pattern automatically registers agent activities with the workflow.
+5. Pass the wrapped toolset at runtime to any agent that needs it.
+6. Register the shared toolset's activities once with the worker. Agent activities are automatically registered via `__pydantic_ai_agents__`.
+
+6. Register the shared toolset's activities once with the worker. Agent activities are automatically registered via `__pydantic_ai_agents__`.
```
DouweM: Duplication

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
```diff
@@ -659,8 +806,9 @@ def run_stream_events(
...
         builtin_tools: Sequence[AbstractBuiltinTool | BuiltinToolFunc[AgentDepsT]] | None = None,
+        event_stream_handler: EventStreamHandler[AgentDepsT] | None = None,
```
DouweM: This shouldn't have been added, here and below

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_dynamic_toolset.py
```diff
@@ -47,15 +47,22 @@ def __init__(
...
+        deps_type: type[AgentDepsT] | None = None,
         run_context_type: type[TemporalRunContext[AgentDepsT]] = TemporalRunContext[AgentDepsT],
     ):
         super().__init__(toolset)
-        self.activity_config = activity_config
-        self.tool_activity_config = tool_activity_config
+        from datetime import timedelta
+
+        if activity_name_prefix is None:  # pragma: no cover
+            raise UserError('activity_name_prefix is required for TemporalDynamicToolset')
```
DouweM: It shouldn't be

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_dynamic_toolset.py
```diff
@@ -47,15 +47,22 @@ def __init__(
...
+        if deps_type is None:  # pragma: no cover
+            raise UserError('deps_type is required for TemporalDynamicToolset')
```
DouweM: It shouldn't be

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_mcp.py
```diff
@@ -61,7 +63,12 @@ async def get_tools_activity(params: _GetToolsParams, deps: AgentDepsT) -> dict[
             return {name: tool.tool_def for name, tool in tools.items()}
 
         # Set type hint explicitly so that Temporal can take care of serialization and deserialization
-        get_tools_activity.__annotations__['deps'] = deps_type
+        if deps_type is not None:
+            get_tools_activity.__annotations__['deps'] = deps_type
+        else:  # pragma: no cover - deps_type is always provided in practice by TemporalAgent
```
DouweM: This is unncessary. And I think we should support a lacking `deps_type` by typing as None or maybe Any.

DouweM:
@qyinm Please have a look at my review, and please human-review the AI help; there were a couple of weird things in there, like `else: pass` snippets, unnecessary code comments, and lines that shouldn't have been added/changed. Using AI agents is fine of course, but we can work more efficiently if I'm not the first one to actually verify the AI output line by line.

qyinm:
Sorry for wasting your time. I'll make sure this doesn't happen again.

DouweM:
@qyinm Thanks for working on this some more, let me know when all the feedback is addressed, CI is green and it's ready for review again!

qyinm:
@DouweM Thanks! I'm still touching up temporal.md, but the rest is ready for review. Feel free to start with those, and I'll let you know when the whole thing is finalized.

qyinm:
@DouweM Updated based on feedback! Not sure why the tests failed though — I only touched the docs, so it seems unrelated. Merry Christmas! 🎄

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
```diff
@@ -45,6 +46,63 @@
...
+
+    Args:
+        toolsets: The toolsets to validate.
+
+    Raises:
+        UserError: If an unwrapped leaf toolset is found that requires temporal wrapping.
+            The error message includes the toolset label for identification.
+    """
+
+    def validate_toolset(t: AbstractToolset[AgentDepsT]) -> None:
```
DouweM, line 64: Can we dedupe this with the `temporalize_toolset` helper function? I'd like for the list of leaf toolsets that have Temporal wrappers to be defined in just one place, possibly as a dictionary mapping unwrapped classes to wrappers

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
```diff
@@ -45,6 +46,63 @@
...
+                raise UserError(f'Toolset {t.label} must be wrapped in a `TemporalWrapperToolset`.')
+
+        # Check if this is a FastMCPToolset that needs wrapping
+        try:
+            from pydantic_ai.toolsets.fastmcp import FastMCPToolset
+        except ImportError:
+            pass
+        else:
+            if isinstance(t, FastMCPToolset):
+                raise UserError(f'Toolset {t.label} must be wrapped in a `TemporalWrapperToolset`.')
```
DouweM: It should be possible to not repeat this

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
```diff
@@ -254,21 +324,30 @@ def temporal_activities(self) -> list[Callable[..., Any]]:
...
 
-        When called outside a workflow, this is a no-op.
-        When called inside a workflow, it overrides the model and toolsets.
-        """
-        if not workflow.in_workflow() and not force:
-            yield
+        if toolsets:
+            overridden_toolsets = [*self._toolsets, *toolsets]
+        else:
+            overridden_toolsets = list(self._toolsets)
```
DouweM: These 5 lines can be simplified as `[*..., *(toolsets or [])]`

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
```diff
@@ -254,21 +324,30 @@ def temporal_activities(self) -> list[Callable[..., Any]]:
...
     @contextmanager
     def _temporal_overrides(
-        self, *, model: models.Model | models.KnownModelName | str | None = None, force: bool = False
+        self,
+        toolsets: Sequence[AbstractToolset[AgentDepsT]] | None = None,
+        model: models.Model | models.KnownModelName | str | None = None,
+        force: bool = False,
     ) -> Iterator[None]:
-        """Context manager for workflow-specific overrides.
+        in_workflow = workflow.in_workflow()
```
DouweM: This doesn't need to be in a variable does it?

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
```diff
@@ -282,6 +361,45 @@ def _temporal_overrides(
...
+                # Original toolset instance: find its temporal wrapper
+                # Check if this toolset instance is wrapped in any of our named toolsets
+                wrapper = next(
+                    (
+                        wrapper
+                        for wrapper in self._named_toolsets.values()
+                        if isinstance(wrapper, TemporalWrapperToolset) and wrapper.wrapped is t
+                    ),
+                    None,
+                )
```
DouweM: This won't work if someone provides `TemporalAgent(toolsets={'foo': CombinedToolset(mcp_server_1, mcp_server_2)})`, because temporalized that'll look like `CombinedToolset(TemporalMCPServer(mcp_server_1), TemporalMCPServer(mcp_server_2))`, not `SomeTemporalWrapperToolset(CombinedToolset(...))`. So I think we need to keep track of a mapping from original to temporalized toolsets, based on the original toolset object ID or something as a key (as the object is not hashable).

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
```diff
@@ -282,6 +361,45 @@ def _temporal_overrides(
...
+                if wrapper is not None:
+                    resolved_toolsets.append(wrapper)
+                else:
+                    # Not found in named toolsets, use as-is (will be validated later)
+                    resolved_toolsets.append(t)
```
DouweM, line 387: It feels more natural to do the validation here, or would that not work right?

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
```diff
@@ -282,6 +361,45 @@ def _temporal_overrides(
...
+                    resolved_toolsets.append(self._named_toolsets[t])
+                except KeyError as e:
+                    if not self._named_toolsets:
+                        raise UserError(f"Unknown toolset name: '{t}'. No named toolsets registered.") from e
+                    raise UserError(
+                        f"Unknown toolset name: '{t}'. Available toolsets: {list(self._named_toolsets.keys())}"
+                    ) from e
+            elif isinstance(t, TemporalWrapperToolset):
+                # Already a temporal wrapper: use as-is
+                resolved_toolsets.append(t)
```
DouweM, line 375: Similar to the comment a few lines down, I think we should also support passing a `CombinedToolset(TemporalMCPServer(mcp_server_1), TemporalMCPServer(mcp_server_2))` here. Likely be just validating it?

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
```diff
@@ -372,13 +490,27 @@ async def main():
...
+        if toolsets:
+            if workflow.in_workflow():
+                # Validate toolsets in workflow context
+                try:
+                    _validate_temporal_toolsets([t for t in toolsets if not isinstance(t, str)])
+                except UserError as e:
+                    raise UserError(
+                        f'Toolsets provided at runtime inside a Temporal workflow must be wrapped in a `TemporalWrapperToolset`. {e}'
+                    ) from e
+            resolved_toolsets = self._resolve_toolsets(toolsets)
```
DouweM: Should we do this before validation?

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
```diff
@@ -979,9 +1117,12 @@ def override(
...
                 )
             if _utils.is_set(toolsets):
-                raise UserError(
-                    'Toolsets cannot be contextually overridden inside a Temporal workflow, they must be set at agent creation time.'
-                )
+                try:
+                    _validate_temporal_toolsets(toolsets)
+                except UserError as e:
+                    raise UserError(
+                        f'Toolsets cannot be contextually overridden inside a Temporal workflow, unless they are wrapped in a `TemporalWrapperToolset`. {e}'
```
DouweM: I don't think it should be necessary to repeat this validation + error 3 times; I'll put some thought into how this could be improved after the next iteration (if that doesn't improve it already!)

tests/test_temporal.py
```diff
@@ -96,6 +97,11 @@
 except ImportError:  # pragma: lax no cover
     pytest.skip('fastmcp not installed', allow_module_level=True)
 
+try:
+    from pydantic_ai.toolsets._dynamic import DynamicToolset
+except ImportError:  # pragma: lax no cover
+    pytest.skip('dynamic toolset not available', allow_module_level=True)
```
DouweM: We shouldn't need this check

tests/test_temporal.py
```diff
@@ -1654,7 +1660,7 @@ async def test_temporal_agent_run_in_workflow_with_toolsets(allow_model_requests
         with workflow_raises(
             UserError,
             snapshot(
-                'Toolsets cannot be set at agent run time inside a Temporal workflow, it must be set at agent creation time.'
+                'Toolsets provided at runtime inside a Temporal workflow must be wrapped in a `TemporalWrapperToolset`. Toolset FunctionToolset must be wrapped in a `TemporalWrapperToolset`.'
```
DouweM: - Error message has some duplication
- We should have a clearer suggestion on how to fix this issue. Can we recommend pre-registering the toolset instead?

tests/test_temporal.py
```diff
@@ -2970,3 +2976,543 @@ async def test_temporal_model_request_stream_outside_workflow():
...
+    combined_with_unwrapped = CombinedToolset([wrapper_1, func_toolset_unwrapped])
+    with pytest.raises(UserError, match='must be wrapped in a `TemporalWrapperToolset`'):
+        _validate_temporal_toolsets([combined_with_unwrapped])
+
+
+def test_resolve_toolsets_with_original_to_temporal_mapping():
+    """Test that _original_to_temporal mapping correctly tracks toolsets by object id."""
+    from pydantic_ai import Agent, FunctionToolset
+    from pydantic_ai.durable_exec.temporal import TemporalAgent
+    from pydantic_ai.toolsets import CombinedToolset
```
DouweM: All imports at the top of the file please

tests/test_temporal.py
```diff
@@ -2970,3 +2976,543 @@ async def test_temporal_model_request_stream_outside_workflow():
...
+
+    # Create agent with the combined toolset registered by name
+    # Note: In real usage, temporalize_toolset would wrap the children
+    # For this test, we're testing the id-based mapping mechanism
+    agent = TemporalAgent(base_agent, toolsets={'my_combined': combined})
+
+    # The original combined toolset should be tracked by object id
+    assert id(combined) in agent._original_to_temporal  # pyright: ignore[reportPrivateUsage]
+
+    # Resolving the original combined toolset should return the temporalized version
```
DouweM: Instead of unit testing the `_original_to_temporal` method, I think we have an integration test to verify that the scenario works.

tests/test_temporal.py
```diff
@@ -2970,3 +2976,543 @@ async def test_temporal_model_request_stream_outside_workflow():
...
+
+    # CombinedToolset with all temporal children should pass
+    combined = CombinedToolset([wrapper_1, wrapper_2])
+    _validate_temporal_toolsets([combined])
+
+    # CombinedToolset with one unwrapped child should fail
+    func_toolset_unwrapped = FunctionToolset(tools=[tool_func], id='test_func_unwrapped')
+    combined_with_unwrapped = CombinedToolset([wrapper_1, func_toolset_unwrapped])
+    with pytest.raises(UserError, match='must be wrapped in a `TemporalWrapperToolset`'):
+        _validate_temporal_toolsets([combined_with_unwrapped])
```
DouweM: As below, I'd rather have an integration test for this scenario, not a unit test

tests/test_temporal.py
DouweM, line 1: I left a couple of comments, but at a higher level: I want to make sure that we have integration tests for every scenario we support that verify that the runtime-specified toolset is actually executed, ideally 0 unit tests that directly test private methods

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
```diff
@@ -195,9 +236,24 @@ def temporalize_toolset(toolset: AbstractToolset[AgentDepsT]) -> AbstractToolset
...
             return toolset
 
-        temporal_toolsets = [toolset.visit_and_replace(temporalize_toolset) for toolset in wrapped.toolsets]
+        # Temporalize wrapped agent's toolsets
+        self._toolsets = [toolset.visit_and_replace(temporalize_toolset) for toolset in wrapped.toolsets]
+
+        # Process additional toolsets (if provided)
+        # Temporalize named toolsets and store the mapping
+        # Also track original toolset id -> temporalized toolset for CombinedToolset support
+        self._original_to_temporal: dict[int, AbstractToolset[AgentDepsT]] = {}
```
DouweM: Variable name is not descriptive enough

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
```diff
@@ -282,6 +342,61 @@ def _temporal_overrides(
...
+                    resolved_toolsets.append(t)
+
+        # Validate after resolving if requested
+        if validate_in_workflow and resolved_toolsets:
+            try:
+                _validate_temporal_toolsets(resolved_toolsets)
+            except UserError as e:
+                raise UserError(
+                    f'Toolsets provided at runtime inside a Temporal workflow must be wrapped in a `TemporalWrapperToolset`. {e}'
+                ) from e
```
DouweM: This is only need in the "not found in mapping" branch, right? Can we move the validation in there?

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
```diff
@@ -282,6 +342,61 @@ def _temporal_overrides(
             finally:
                 self._temporal_overrides_active.reset(temporal_active_token)
 
+    def _resolve_toolsets(
+        self,
+        toolsets: Sequence[AbstractToolset[AgentDepsT] | str] | None,
+        *,
+        validate_in_workflow: bool = False,
```
DouweM: Since we don't do the in_workflow check here, should this just be validate?

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
```diff
@@ -42,7 +42,42 @@
...
+        if isinstance(t, TemporalWrapperToolset):
+            return
+
+        # Check if this is a toolset type that requires temporal wrapping
+        if isinstance(t, types_requiring_wrapper):
+            raise UserError(f'Toolset {t.label} must be wrapped in a `TemporalWrapperToolset`.')
+
+        # For other toolsets (like CombinedToolset, WrapperToolset, etc.),
+        # we return them unchanged - apply will handle recursion
+        return  # pragma: no cover - defensive code for future toolset types
```
DouweM: this line seems entirely unnecessary

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
```diff
@@ -42,7 +42,42 @@
...
+        UserError: If an unwrapped leaf toolset is found that requires temporal wrapping.
+            The error message includes the toolset label for identification.
+    """
+    types_requiring_wrapper = get_toolset_types_requiring_temporal_wrapper()
+
+    def validate_toolset(t: AbstractToolset[AgentDepsT]) -> None:
+        # If we encounter a TemporalWrapperToolset, we don't need to check its children
+        # since they're already wrapped
+        if isinstance(t, TemporalWrapperToolset):
+            return
```
DouweM: Do we need this check still since we already have a limited isinstance below?

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
```diff
@@ -254,21 +310,25 @@ def temporal_activities(self) -> list[Callable[..., Any]]:
...
+        overridden_toolsets = [*self._toolsets, *(toolsets or [])]
 
-        When called outside a workflow, this is a no-op.
-        When called inside a workflow, it overrides the model and toolsets.
-        """
+        # Outside workflow, only apply toolsets override (model is passed directly to run)
         if not workflow.in_workflow() and not force:
-            yield
+            if toolsets:
+                with super().override(toolsets=overridden_toolsets, tools=[]):
```
DouweM: In the not-in-workflow scenario, I think it'll lead to easier-to-understand code if we pass the toolsets into the `super().run` call, similar to how we do for model.

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
```diff
@@ -401,7 +521,7 @@ async def main():
                 usage_limits=usage_limits,
                 usage=usage,
                 infer_name=infer_name,
-                toolsets=toolsets,
+                toolsets=None,  # Toolsets are set via _temporal_overrides
```
DouweM: See above; I'd like to see if we can pass `resolved_toolsets` here in the not-in-workflow scenario

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_dynamic_toolset.py
```diff
@@ -69,8 +71,10 @@ async def get_tools_activity(params: _GetToolsParams, deps: AgentDepsT) -> dict[
                     for name, tool in tools.items()
                 }
 
-        get_tools_activity.__annotations__['deps'] = deps_type
+        # Set type hint explicitly so that Temporal can take care of serialization and deserialization
+        get_tools_activity.__annotations__['deps'] = deps_type or Any
```
DouweM: Why `or Any` instead of using `None`?

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_function_toolset.py
```diff
@@ -49,8 +50,9 @@ async def call_tool_activity(params: CallToolParams, deps: AgentDepsT) -> CallTo
             return await self._call_tool_in_activity(name, params.tool_args, ctx, tool)
 
         # Set type hint explicitly so that Temporal can take care of serialization and deserialization
-        call_tool_activity.__annotations__['deps'] = deps_type
+        call_tool_activity.__annotations__['deps'] = deps_type or Any
 
+        activity_name_prefix = activity_name_prefix or ''
```
DouweM: This will result in teh toolset name `__toolset__....`, but `toolset__...` would look better, so we should only add the leading `__` if there is in fact a prefix

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_mcp.py
```diff
@@ -33,16 +34,17 @@ def __init__(
...
-        deps_type: type[AgentDepsT],
+        activity_name_prefix: str | None = None,
+        activity_config: ActivityConfig | None = None,
+        tool_activity_config: dict[str, ActivityConfig | Literal[False]] | None = None,
+        deps_type: type[AgentDepsT] | None = None,
         run_context_type: type[TemporalRunContext[AgentDepsT]] = TemporalRunContext[AgentDepsT],
     ):
         super().__init__(toolset)
-        self.activity_config = activity_config
+        self.activity_config = activity_config or ActivityConfig(start_to_close_timeout=timedelta(minutes=1))
```
DouweM: This is now duplicated between here and TemporalFunctionToolset. Would it make sense to move up to the wrapper?

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_mcp.py
```diff
@@ -61,7 +63,9 @@ async def get_tools_activity(params: _GetToolsParams, deps: AgentDepsT) -> dict[
             return {name: tool.tool_def for name, tool in tools.items()}
 
         # Set type hint explicitly so that Temporal can take care of serialization and deserialization
-        get_tools_activity.__annotations__['deps'] = deps_type
+        get_tools_activity.__annotations__['deps'] = deps_type or Any
+
+        activity_name_prefix = activity_name_prefix or ''
```
DouweM: Same questions as above apply here as well

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_toolset.py
```diff
@@ -18,6 +18,36 @@
 from ._run_context import TemporalRunContext
 
 
+def get_toolset_types_requiring_temporal_wrapper() -> tuple[type, ...]:
+    """Get the tuple of toolset types that require temporal wrapping.
+
+    This function lazily imports and returns all toolset types that need to be
+    wrapped in a TemporalWrapperToolset when used inside a Temporal workflow.
+    The list is defined in one place to avoid duplication between validation
+    and temporalization logic.
```
DouweM: Did you consider the approach I suggested of building a dict that maps unwrapped types to wrapper types? That we can then also use in `temporalize_toolset`?

PR #3778, opened 2025-12-19
Relock deps

dsfaccini:
- Fix `outlines-transformers` and `outlines-vllm-offline` markers to use
  `sys_platform == 'linux'` since vllm/torch only have Linux wheels
- Remove outdated `python_version < '3.12'` restriction from vllm
  (vllm 0.10.0+ supports Python 3.12+)
- Add `vllm>=0.10.0` minimum version to ensure proper wheel availability
- Add `[tool.uv.environments]` to restrict lockfile resolution to
  Linux and macOS platforms (excludes Windows which has no wheels)
- Regenerate uv.lock with updated dependencies

The previous marker `(sys_platform != 'darwin' or platform_machine != 'x86_64')`
incorrectly included macOS ARM64 and Windows, causing uv to fall back to
ancient vllm versions (0.1.3) that require CUDA to build from source.

Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -2,25 +2,24 @@
   "permissions": {
     "allow": [
       "Bash(rg:*)",
-      "Bash(make:*)",
-      "Bash(uv run pytest:*)",
-      "Bash(uv run:*)",
-      "Bash(git push:*)",
-      "Bash(git checkout:*)",
-      "Bash(gh pr create:*)",
-      "Bash(gh pr view:*)",
-      "Bash(git add:*)",
-      "Bash(git commit:*)",
-      "Bash(gh pr comment:*)",
-      "Bash(gh pr checks:*)",
-      "Bash(gh run view:*)",
-      "Bash(gh pr list:*)",
-      "Bash(gh api:*)",
-      "Bash(pre-commit run:*)",
+      "Bash(ls:*)",
+      "Bash(tree:*)",
       "Bash(grep:*)",
-      "Bash(find:*)",
-      "Bash(coverage:*)"
-    ],
-    "deny": []
+      "Bash(git log:*)",
+      "Bash(git show:*)",
+      "Bash(git diff:*)",
+      "Bash(git grep:*)",
+      "Bash(git status:*)",
+      "Bash(gh pr view:*)",
+      "Bash(gh pr checks:*)",
+      "Bash(gh pr list:*)",
+      "Bash(gh pr diff:*)",
+      "Bash(gh issue view:*)",
+      "Bash(gh issue list:*)",
+      "Bash(gh run view:*)",
+      "Bash(gh run list:*)",
+      "Bash(uv run:*)",
+      "Bash(make:*)"
+    ]
   }
 }
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -147,8 +147,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # TODO(Marcelo): Enable 3.11 again.
-        python-version: ["3.10", "3.12", "3.13"]
+        python-version: ["3.10", "3.11", "3.12", "3.13"]
     env:
       CI: true
       COVERAGE_PROCESS_START: ./pyproject.toml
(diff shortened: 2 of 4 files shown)


pyproject.toml
```diff
@@ -88,6 +88,12 @@ members = [
 
 [tool.uv]
 default-groups = ["dev", "lint", "docs"]
+environments = [
+    "sys_platform == 'linux' and platform_machine == 'x86_64'",
+    "sys_platform == 'linux' and platform_machine == 'aarch64'",
+    "sys_platform == 'darwin' and platform_machine == 'arm64'",
+    "sys_platform == 'darwin' and platform_machine == 'x86_64'",
+]
```
Kludex: Why do we have this?

pydantic_ai_slim/pyproject.toml
```diff
@@ -77,11 +77,11 @@ openrouter = ["openai>=2.8.0"]
 mistral = ["mistralai>=1.9.10"]
 bedrock = ["boto3>=1.40.14"]
 huggingface = ["huggingface-hub[inference]>=0.33.5,<1.0.0"]
-outlines-transformers = ["outlines[transformers]>=1.0.0, <1.3.0; (sys_platform != 'darwin' or platform_machine != 'x86_64')", "transformers>=4.0.0", "pillow", "torch; (sys_platform != 'darwin' or platform_machine != 'x86_64')"]
+outlines-transformers = ["outlines[transformers]>=1.0.0, <1.3.0; sys_platform == 'linux'", "transformers>=4.0.0", "pillow", "torch; sys_platform == 'linux'"]
```
Kludex: why did those change?

Kludex: ok, I got why we need this one, why do we need the `environments`?

pyproject.toml
Kludex: The reason claude gave is mistaken, please revert this one.

pydantic_ai_slim/pyproject.toml
dsfaccini: I restablished this part because was getting a type error on the `from outlines.models.transformers import from_transformers` in `tests/conftest.py`

I guess if someone is contributing from a windows PC they'll also have that type error

DouweM:
@dsfaccini @Kludex Is this one still relevant?

Kludex:
> @dsfaccini @Kludex Is this one still relevant?

Leave it to me.

uv.lock
```diff
@@ -3377,45 +3509,92 @@ wheels = [
...
+    { url = "https://files.pythonhosted.org/packages/fb/df/5bd7a48c256faecd1d36edc13133e51397e41b73bb77e1a69deab746ebac/markupsafe-3.0.3-cp314-cp314t-win32.whl", hash = "sha256:915c04ba3851909ce68ccc2b8e2cd691618c4dc4c4232fb7982bca3f41fd8c3d", size = 14819, upload-time = "2025-09-27T18:37:26.285Z" },
+    { url = "https://files.pythonhosted.org/packages/1a/8a/0402ba61a2f16038b48b39bccca271134be00c5c9f0f623208399333c448/markupsafe-3.0.3-cp314-cp314t-win_amd64.whl", hash = "sha256:4faffd047e07c38848ce017e8725090413cd80cbc23d86e55c587bf979e579c9", size = 15426, upload-time = "2025-09-27T18:37:27.316Z" },
+    { url = "https://files.pythonhosted.org/packages/70/bc/6f1c2f612465f5fa89b95bead1f44dcb607670fd42891d8fdcd5d039f4f4/markupsafe-3.0.3-cp314-cp314t-win_arm64.whl", hash = "sha256:32001d6a8fc98c8cb5c947787c5d08b0a50663d139f1305bac5885d98d9b40fa", size = 14146, upload-time = "2025-09-27T18:37:28.327Z" },
 ]
 
 [[package]]
 name = "mcp"
-version = "1.18.0"
+version = "1.25.0"
```
dsfaccini, line 3597: update mcp

dsfaccini:
lets merge this asap @Kludex

Kludex:
This PR seems to have a mix of changes. Didn't we relock already?

dsfaccini:
I don't think we did, there would otherwise not be much of a diff on the `uv.lock`, relocking also surfaced a couple bugs/deprecations on e.g. `mcp.py` which is why those changes are there. There's admittedly a bit of an `openai.py` refactor that I can roll back but otherwise, the rest of the changes just come from relocking and satisfying `make typecheck`

Kludex:
> I don't think we did, there would otherwise not be much of a diff on the `uv.lock`, relocking also surfaced a couple bugs/deprecations on e.g. `mcp.py` which is why those changes are there. There's admittedly a bit of an `openai.py` refactor that I can roll back but otherwise, the rest of the changes just come from relocking and satisfying `make typecheck`

I don't think we should merge this PR as is. Please create a new PR with the specific issues e.g. mcp.

Also, what you said about "typecheck" is because you bumped pyright, I don't think we should bump in that state.

PR #3785, opened 2025-12-20
Add `examples` field to `ToolDefinition` and send it to Anthropic

### LINKED ISSUE #4168: Anthropic tool use examples
### Description

- Context: https://github.com/pydantic/pydantic-ai/issues/3666

### References

_No response_

AlanPonnachan:
- Closes #4168

Adds structured tool input examples across `ToolDefinition`, function tools, agent decorators, function toolsets, and `ToolOutput`.

Anthropic transports that support `input_examples` receive examples natively. Legacy Amazon Bedrock and other providers receive a JSON fallback in the tool description. Function-tool examples are caller-authored wire objects, while `ToolOutput` examples are semantic output values serialized into the output tool input shape.

Includes public API compatibility tests, transport-specific profile coverage, an end-to-end Anthropic cassette assertion, fallback tests, documentation, and bundled agent guidance.

### Checklist


pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -459,6 +460,14 @@ def _get_betas_and_extra_headers(
         if has_strict_tools or model_request_parameters.output_mode == 'native':
             betas.add('structured-outputs-2025-11-13')
 
+        # Check if any tools use input_examples (tool use examples feature)
+        has_input_examples = any('input_examples' in tool for tool in tools)
+        if has_input_examples:
+            if isinstance(self.client, (AsyncAnthropicBedrock, AsyncAnthropicVertex)):
```
DouweM: We can use `|`

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -1125,6 +1134,8 @@ def _map_tool_definition(self, f: ToolDefinition) -> BetaToolParam:
         }
         if f.strict and self.profile.supports_json_schema_output:
             tool_param['strict'] = f.strict
+        if f.examples:
+            tool_param['input_examples'] = f.examples
```
DouweM, line 2278: I believe that if you have a tool with a single argument that's a `BaseModel` or `dataclass` (e.g. `def foo(bar: Baz)`), Pydantic/Pydantic AI automatically flattens the JSON schema so that the model generates a dict with the keys of `Baz`, rather than a dict with key `bar` pointing to a `Baz` dict. 

So in this case, if the user includes the `bar` key when passing `examples` to `agent.tool()`, those examples will not actually be valid from the perspective of the LLM/API, as it doesn't know about the outer key. 

So I think we should apply that same flattening processing on examples when the `ToolDefinition` is built.

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/models/anthropic.py
AlanPonnachan: good point

Kludex (changes requested):
I think we should be able to pass from the docstring, since Google docstring style has the `Examples:`, e.g.: https://gist.github.com/redlotus/3bc387c2591e3e908c9b63b97b11d24e#file-docstrings-py-L142-L148

pydantic_ai_slim/pydantic_ai/tools.py
```diff
@@ -516,6 +530,18 @@ class ToolDefinition:
         See the [tools documentation](../deferred-tools.md#human-in-the-loop-tool-approval) for more info.
     """
 
+    examples: list[dict[str, Any]] | None = None
+    """Example inputs demonstrating correct tool usage patterns.
+
+    Provide 1-5 realistic examples showing parameter conventions, optional field patterns,
```
dsfaccini: where does the `1-5` come from? if this is anthropic specific it should have a respective profile setting (e.g. `max_nr_of_examples_supported`

dsfaccini: otherwise simply change the wording to remove the `1-5`

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -459,6 +460,14 @@ def _get_betas_and_extra_headers(
...
             betas.add('structured-outputs-2025-11-13')
 
+        # Check if any tools use input_examples (tool use examples feature)
+        has_input_examples = any('input_examples' in tool for tool in tools)
+        if has_input_examples:
+            if isinstance(self.client, AsyncAnthropicBedrock | AsyncAnthropicVertex):
+                betas.add('tool-examples-2025-10-29')
+            else:
+                betas.add('advanced-tool-use-2025-11-20')
+
```
dsfaccini: this is missing a model check
image

dsfaccini: or more likely a profile check, we probably need a `supports_tool_examples` field

tests/models/test_anthropic.py
```diff
@@ -8292,3 +8292,83 @@ async def test_anthropic_system_prompts_and_instructions_ordering():
...
+    from anthropic import AsyncAnthropicBedrock
+
+    # Mock client to pass isinstance check
+    mock_client = MagicMock(spec=AsyncAnthropicBedrock)
+    mock_client.base_url = 'https://bedrock.amazonaws.com'
+
+    m = AnthropicModel('claude-3-5-sonnet-latest', provider=AnthropicProvider(anthropic_client=mock_client))
+
+    from pydantic_ai.models import ModelRequestParameters
+    from pydantic_ai.models.anthropic import AnthropicModelSettings
```
dsfaccini: please review any generated tests more thoroughly because you're importing the same modules in every test instead of importing them once at the top of the file

tests/models/test_anthropic.py
```diff
@@ -8292,3 +8292,83 @@ async def test_anthropic_system_prompts_and_instructions_ordering():
...
+    @agent.tool_plain(examples=examples)
+    def my_tool(x: int) -> int:
+        return x
+
+    await agent.run('hi')
+
+    completion_kwargs = get_mock_chat_completion_kwargs(mock_client)[0]
+    betas = completion_kwargs['betas']
+    assert 'advanced-tool-use-2025-11-20' in betas
+    assert 'tool-examples-2025-10-29' not in betas
```
dsfaccini: about all these tests, `tool-examples-2025-10-29` is only supported by Opus 4.5 as per https://platform.claude.com/docs/en/agents-and-tools/tool-use/implement-tool-use#providing-tool-use-examples

so we need one test that expects an exception when trying to use tool examples via a provider/model that's not supported and 2 VCR tests, one for the tool examples beta and one for the advanced tool use beta

include the relevant assertions inside the VCR test, e.g. by patching the http client and asserting the betas header is sent or some other way, there's likely already an example of this in the codebase

tests/test_tools.py
```diff
@@ -2776,3 +2793,101 @@ def test_agent_tool_timeout_passed_to_toolset():
...
     assert agent._function_toolset.timeout == 30.0
+
+
+def test_tool_examples_init():
+    def my_tool(x: int) -> int:
+        return x
+
+    examples = [{'x': 1}]
+    tool = Tool(my_tool, examples=examples)
+    assert tool.examples == examples
```
dsfaccini, line 3374: `assert tool.examples == tool.tool_def.examples == examples`

dsfaccini (changes requested):

dsfaccini:
hey @AlanPonnachan thank you for working on this! an ask for the future: please make sure to have a passing test suite and having done a read-through of any generated tests before submitting for review

pydantic_ai_slim/pydantic_ai/_griffe.py
```diff
@@ -65,6 +65,22 @@ def doc_descriptions(
     if main := next((p for p in sections if p.kind == DocstringSectionKind.text), None):
         main_desc = main.value
 
+    # Handle Examples section
+    if examples := next((p for p in sections if p.kind == DocstringSectionKind.examples), None):
```
DouweM: What's all this for? We don't seem to have a test that does anything with docstrings

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -904,6 +906,10 @@ def build(
                 strict = default_strict
 
             processor = ObjectOutputProcessor(output=output, description=description, strict=strict)
+            # Flatten examples if the output function has a single argument that gets flattened in the schema
+            if examples and processor._function_schema and processor._function_schema.single_arg_name:  # pyright: ignore[reportPrivateUsage]
+                arg_name = processor._function_schema.single_arg_name  # pyright: ignore[reportPrivateUsage]
```
DouweM: - It think we should do something similar for `processor.outer_typed_dict_key` (which is populated for e.g. `output_type=float`): wrapping the example in it
- Can we do this without using private properties? The processor could have a public property
- We're reading the field two times

pydantic_ai_slim/pydantic_ai/models/anthropic.py
DouweM: Yeah we should have a `AnthropicModelProfile.amthropic_supports_tool_examples` that is dynamically set in `AnthropicProvider.model_profile` based on the combination of `self.client` + `model_name`, and then checked here (and where we populated `input_examples`)

pydantic_ai_slim/pydantic_ai/output.py
```diff
@@ -125,12 +127,14 @@ def __init__(
         description: str | None = None,
         max_retries: int | None = None,
         strict: bool | None = None,
+        examples: list[dict[str, Any]] | None = None,
```
DouweM: Specifically on `ToolOutput`, I think this should be `list[Any]`, as you could have `output_type=ToolOutput(list[Foo], examples=[Foo(1), Foo(2), Foo(3)])`. That's where wrapping in `outer_typed_dict_key` comes in, as I mentioned above.

pydantic_ai_slim/pydantic_ai/tools.py
```diff
@@ -378,6 +382,11 @@ async def prep_my_tool(
         self.strict = strict
         self.sequential = sequential
         self.requires_approval = requires_approval
+        self.examples = examples
+        # Flatten examples if the tool has a single argument that gets flattened in the schema
+        if self.examples and self.function_schema.single_arg_name:
+            arg_name = self.function_schema.single_arg_name
+            self.examples = [ex[arg_name] if isinstance(ex, dict) and arg_name in ex else ex for ex in self.examples]
```
DouweM: Can we somehow not duplicate this with the output code?

tests/models/test_anthropic.py
```diff
@@ -8293,3 +8302,73 @@ async def test_anthropic_system_prompts_and_instructions_ordering():
...
+
+async def test_anthropic_bedrock_beta_headers_with_examples(allow_model_requests: None):
+    # Mock client to pass isinstance check
+    mock_client = MagicMock(spec=AsyncAnthropicBedrock)
+    mock_client.base_url = 'https://bedrock.amazonaws.com'
+
+    m = AnthropicModel('claude-3-5-sonnet-latest', provider=AnthropicProvider(anthropic_client=mock_client))
+
+    from pydantic_ai.models import ModelRequestParameters
+    from pydantic_ai.models.anthropic import AnthropicModelSettings
```
DouweM: Imports only at the top of the file please

tests/test_tools.py
```diff
@@ -2776,3 +2793,99 @@ def test_agent_tool_timeout_passed_to_toolset():
...
+def test_tool_output_examples():
+    examples = [{'x': 1}]
+    tool_output = ToolOutput(int, name='foo', examples=examples)
+    assert tool_output.examples == examples
+
+    agent = Agent('test', output_type=tool_output)
+    # Access the output toolset to verify definition
+    assert agent._output_toolset is not None
+    tool_def = agent._output_toolset._tool_defs[0]
+    assert tool_def.examples == examples
```
DouweM: This example is invalid! `{'x': 1}` does not actually match the schema, which would require `{'result': int}` because of the `outer_typed_dict_key` I mentioned above. So really we should pass `examples=[1]` above, and then here we should see `[{'result': 1}]`

tests/test_tools.py
```diff
@@ -2776,3 +2793,99 @@ def test_agent_tool_timeout_passed_to_toolset():
...
+        return arg.x
+
+    # User provides examples matching the python signature (wrapped in the argument name)
+    examples = [{'arg': {'x': 1}}, {'arg': {'x': 2}}]
+
+    # The Tool logic should flatten this to match the JSON schema sent to the LLM
+    tool = Tool(my_tool, examples=examples)
+
+    # Expect flattened examples
+    assert tool.examples == [{'x': 1}, {'x': 2}]
```
DouweM: I think we should probably only remove the single arg from the examples stored on the `tool_def`, where it needs to match the `parameter_json_schema`, while leaving the original examples on the `Tool`, which references the entire function.

tests/test_tools.py
```diff
@@ -2776,3 +2793,99 @@ def test_agent_tool_timeout_passed_to_toolset():
...
+        """
+        return x
+
+    tool = Tool(my_tool)
+    assert tool.description == snapshot(
+        """\
+My tool description.
+
+Examples:
+DocstringSectionKind.examples
```
DouweM: This is definitely not right :)

tests/test_tools.py
```diff
@@ -2776,3 +2793,99 @@ def test_agent_tool_timeout_passed_to_toolset():
...
+>>> my_tool(1)
+1\
```
DouweM: We lose indentation here it looks like

pydantic_ai_slim/pydantic_ai/_griffe.py
```diff
@@ -65,6 +65,22 @@ def doc_descriptions(
...
+        examples_content: list[str] = []
+        for source, output in examples.value:
+            if output:
+                examples_content.append(f'{source}\n{output}')
+            else:
+                examples_content.append(source)
+
+        if examples_content:
+            formatted_examples = '\n\n'.join(examples_content)
```
DouweM: Was it intentional that we're only including this in the description sent to the model, not trying to parse out JSON and send them as `input_examples`?

pydantic_ai_slim/pydantic_ai/tools.py
```diff
@@ -516,6 +530,18 @@ class ToolDefinition:
...
+    examples: list[dict[str, Any]] | None = None
+    """Example inputs demonstrating correct tool usage patterns.
+
+    Provide realistic examples showing parameter conventions, optional field patterns,
+    nested structures, and API-specific conventions. Each example must validate against
+    the tool's `parameters_json_schema`.
+
+    Supported by:
+
+    * [Anthropic](https://platform.claude.com/docs/en/agents-and-tools/tool-use/implement-tool-use#providing-tool-use-examples)
```
DouweM: Instead of only sending these to Anthropic, what do you think about including them in the description for other models? (like we're now parsing examples from the docstring)

If someone explicitly adds `examples`, they're already indicating they're OK with the extra token usage, so I think it'd more surprising if they _don't_ get sent to the model.

pydantic_ai_slim/pydantic_ai/tools.py
```diff
@@ -436,6 +449,7 @@ def tool_def(self):
             sequential=self.sequential,
             metadata=self.metadata,
             timeout=self.timeout,
+            examples=self.examples,
```
DouweM: When creating the tool def, I think we should serialize these examples using Pydantic `TypeAdapter[dict[str, Any]].dump_python(...)`, so that we don't accidentally send full `BaseModel`s etc into the Anthropic SDK

DouweM (changes requested):

dsfaccini:

```bash
git pull

deactivate

rm -rf .venv

make install
```

pydantic_ai_slim/pydantic_ai/_griffe.py
AlanPonnachan: Yes. Docstring examples are unstructured text or Python code (doctests), while `input_examples` requires strict JSON matching the schema. Appending them to the description is the safe way to ensure the model sees them as documentation without brittle parsing logic.

pydantic_ai_slim/pydantic_ai/tools.py
```diff
@@ -428,18 +437,42 @@ def from_schema(
...
+                TypeAdapter(Any).dump_python(ex, mode='json')  # pyright: ignore[reportUnknownMemberType]
+                for ex in examples  # pyright: ignore[reportUnknownVariableType]
+            ]
+
+            # Append examples to description so they are visible to models that don't support input_examples
+            examples_str = json.dumps(examples, indent=2)
+            if description:
+                description = f'{description}\n\nExamples:\n{examples_str}'
+            else:
+                description = f'Examples:\n{examples_str}'
```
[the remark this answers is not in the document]
DouweM: Right, we should make sure we only add them to the description in `Model.prepare_request`, once it's been verified that the model profile does not natively support this feature. We're doing the same thing for `return_schema` (which Google supports natively) in https://github.com/pydantic/pydantic-ai/pull/3865

AlanPonnachan:
@DouweM 

While working on the next batch of updates to address the review comments, I checked the latest docs: https://platform.claude.com/docs/en/agents-and-tools/tool-use/implement-tool-use#providing-tool-use-examples
and noticed a some changes: input_examples is now documented as a standard parameter and no longer seems to require beta headers.

Specifically:
1. `input_examples` is listed as a standard optional field in the client tools parameter table.
2. The code examples in doc pass `input_examples` directly into the `tools` definition, and none of them inject any beta headers (like `tool-examples-2025-10-29` or `advanced-tool-use-2025-11-20`) into the API call.



**Based on new doc and previous review comments, thought of doing following changes:**

1. Instead of an Anthropic-specific beta header string, I will add a simple `supports_tool_examples: bool = False` to the base `ModelProfile`.
2.  In `anthropic_model_profile`, I will set `supports_tool_examples = True` for the specific models that support it (e.g., the 4.x family).
3.  I will strip out all the Bedrock/Vertex/Standard checking logic from `AnthropicProvider` and `AnthropicModel`. We just unconditionally map `f.examples` to `input_examples` in `_map_tool_definition` if `self.profile.supports_tool_examples` is True.
4. To address Fix the "Double Token" Issue, I will move the logic that appends examples to the text `description` into `Model.prepare_request`. It will *only* append the text if `not self.profile.supports_tool_examples`. This prevents Anthropic from paying the token cost twice while preserving the fallback for other models.
5. I'll extract the flattening/serialization logic into a shared helper in `tools.py` (used by both `Tool` and `OutputToolset`), and fix the `list[dict[str, Any]]` to `list[Any]` type hints.
6. adjust the docstring parser so `Examples:` is strictly appended *after* the `<summary>`/`<returns>` XML wrapper to avoid breaking the XML layout.
7. add a full end-to-end `@pytest.mark.vcr()` test using `claude-opus-4-6` to verify `input_examples` reaches the network payload correctly.

Does this revised plan sound good to you?

DouweM:
@AlanPonnachan That all makes sense to me! @adtyavrdhn Can you have a look here please as this is similar to #3865, where we're sending the return schema in the description or a Google API field?

pydantic_ai_slim/pydantic_ai/profiles/anthropic.py
```diff
@@ -17,8 +17,13 @@ def anthropic_model_profile(model_name: str) -> ModelProfile | None:
     # TODO update when new models are released that support structured outputs
     # https://docs.claude.com/en/docs/build-with-claude/structured-outputs#example-usage
 
+    models_that_support_tool_examples = ('claude-opus-4-6',)
+
     supports_json_schema_output = model_name.startswith(models_that_support_json_schema_output)
+    supports_tool_examples = model_name.startswith(models_that_support_tool_examples)
```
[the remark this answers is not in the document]
AlanPonnachan: According to https://platform.claude.com/docs/en/agents-and-tools/tool-use/implement-tool-use#providing-tool-use-examples, I think 'claude-opus-4-6' is the only supported model

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -1147,6 +1147,9 @@ def _map_tool_definition(self, f: ToolDefinition) -> BetaToolParam:
         }
         if f.strict and self.profile.supports_json_schema_output:
             tool_param['strict'] = f.strict
+
+        if f.examples and self.profile.supports_tool_examples:
+            tool_param['input_examples'] = f.examples
```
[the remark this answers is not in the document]
AlanPonnachan: `input_examples` is now natively supported without beta headers for the latest models. Update PR description

tests/models/test_anthropic.py
```diff
@@ -8409,3 +8415,28 @@ async def test_anthropic_system_prompts_and_instructions_ordering():
...
+    tool_def = ToolDefinition(name='foo', description='bar', examples=examples)
+
+    mock_client = MockAnthropic.create_mock(completion_message([], BetaUsage(input_tokens=0, output_tokens=0)))
+    m = AnthropicModel('claude-opus-4-6', provider=AnthropicProvider(anthropic_client=mock_client))
+
+    mapped = m._map_tool_definition(tool_def)  # pyright: ignore[reportPrivateUsage]
+    assert mapped.get('input_examples') == examples
+
+
+@pytest.mark.vcr()
```
AlanPonnachan: @DouweM ,
Could you please run the test on your end  to generate the cassette?

DouweM: @AlanPonnachan Done and pushed!

tests/models/test_anthropic.py
```diff
@@ -8409,3 +8415,27 @@ async def test_anthropic_system_prompts_and_instructions_ordering():
...
+    tool_def = ToolDefinition(name='foo', description='bar', examples=examples)
+
+    mock_client = MockAnthropic.create_mock(completion_message([], BetaUsage(input_tokens=0, output_tokens=0)))
+    m = AnthropicModel('claude-opus-4-6', provider=AnthropicProvider(anthropic_client=mock_client))
+
+    mapped = m._map_tool_definition(tool_def)  # pyright: ignore[reportPrivateUsage]
+    assert mapped.get('input_examples') == examples
+
+
+async def test_anthropic_tool_examples(allow_model_requests: None, anthropic_api_key: str):
```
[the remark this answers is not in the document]
DouweM: Untrue, there's already vcr at the top.

tests/models/test_anthropic.py
```diff
@@ -8409,3 +8415,27 @@ async def test_anthropic_system_prompts_and_instructions_ordering():
...
+
+
+async def test_anthropic_tool_definition_examples(allow_model_requests: None):
+    examples = [{'x': 1}]
+    tool_def = ToolDefinition(name='foo', description='bar', examples=examples)
+
+    mock_client = MockAnthropic.create_mock(completion_message([], BetaUsage(input_tokens=0, output_tokens=0)))
+    m = AnthropicModel('claude-opus-4-6', provider=AnthropicProvider(anthropic_client=mock_client))
+
+    mapped = m._map_tool_definition(tool_def)  # pyright: ignore[reportPrivateUsage]
```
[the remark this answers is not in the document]
DouweM: I'm Ok with this here.

pydantic_ai_slim/pydantic_ai/tools.py
```diff
@@ -198,6 +198,27 @@ class ToolDenied:
...
+    single_arg_name: str | None = None,
+    outer_typed_dict_key: str | None = None,
+) -> list[dict[str, Any]]:
+    """Process tool examples by flattening, serializing, and optionally wrapping them."""
+    if single_arg_name:
+        examples = [ex[single_arg_name] if isinstance(ex, dict) and single_arg_name in ex else ex for ex in examples]
+
+    # Serialize examples to JSON-compatible dicts
+    examples = [
+        TypeAdapter(Any).dump_python(ex, mode='json')  # pyright: ignore[reportUnknownMemberType]
```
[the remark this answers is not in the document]
DouweM: We may also be able to use a method from `pydantic_core` instead of a `TypeAdapter`

pydantic_ai_slim/pydantic_ai/tools.py
```diff
@@ -535,6 +570,20 @@ class ToolDefinition:
...
+    Provide realistic examples showing parameter conventions, optional field patterns,
+    nested structures, and API-specific conventions. Each example must validate against
+    the tool's `parameters_json_schema`.
+
+    Supported by:
+
+    * [Anthropic](https://platform.claude.com/docs/en/agents-and-tools/tool-use/implement-tool-use#providing-tool-use-examples)
+
+    If provided, these examples are also appended to the tool description for other models that don't support structured examples directly.
+    """
```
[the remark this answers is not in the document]
DouweM: Yep we need docs.

pydantic_ai_slim/pydantic_ai/profiles/anthropic.py
```diff
@@ -17,8 +17,13 @@ def anthropic_model_profile(model_name: str) -> ModelProfile | None:
     # TODO update when new models are released that support structured outputs
     # https://docs.claude.com/en/docs/build-with-claude/structured-outputs#example-usage
 
+    models_that_support_tool_examples = ('claude-opus-4-6',)
```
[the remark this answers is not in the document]
AlanPonnachan: In docs, there is no explicit mention of supported models.
As of now i given 
('claude-sonnet-4-5', 'claude-sonnet-4-6', 'claude-opus-4-5', 'claude-opus-4-6')

adtyavrdhn:
Sorry about the delay, I missed this notification 😞 

This looks inline with what we did for Tool `return_schema`. The design is sound.

AlanPonnachan:
conflicts resolved, will check the ci and update

DouweM:
@AlanPonnachan Appreciate it Alan, it's on my list to get back to next week.

DouweM:
@AlanPonnachan Any chance you can get CI green for me? 😇

AlanPonnachan:
@DouweM Sorry for the delay, it’s been a busy week. CI is green now.

DouweM:
I'm about to go on parental leave (back on June 29), I'll continue on this when I'm back, appreciate the patience!

pydantic_ai_slim/pydantic_ai/providers/anthropic.py
```diff
@@ -2,7 +2,7 @@
 
 import os
 import warnings
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
```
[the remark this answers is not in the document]
adtyavrdhn: Valid and fixed. The provider no longer calls `dataclasses.replace()` on a `TypedDict`; profile overrides use `merge_profile()` and `AnthropicModelProfile(...)`.

tests/models/test_anthropic.py
adtyavrdhn: Valid for the earlier beta API, but superseded by Anthropic making `input_examples` a standard tool field. The current VCR test asserts `input_examples` in the wire payload, and transport tests cover native support versus legacy Bedrock fallback.

pydantic_ai_slim/pydantic_ai/output.py
```diff
@@ -121,6 +121,8 @@ class Vehicle(BaseModel):
     """
     strict: bool | None
     """Whether to use strict mode for the tool."""
+    examples: list[Any] | None
+    """Example inputs demonstrating correct tool usage."""
```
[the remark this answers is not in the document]
adtyavrdhn: The dedicated output docs now explain the `ToolOutput` contract in detail. We intentionally did not cross-reference `ToolDefinition.examples` from the attribute because `ToolOutput` accepts semantic output values, while `ToolDefinition` holds serialized JSON object inputs; conflating those contracts would be misleading.

pydantic_ai_slim/pydantic_ai/providers/anthropic.py
```diff
@@ -62,7 +62,19 @@ def model_profile(model_name: str) -> ModelProfile | None:
...
-        return AnthropicModelProfile(json_schema_transformer=AnthropicJsonSchemaTransformer).update(profile)
+        # TODO update when new models are released that support tool examples
+        # https://platform.claude.com/docs/en/agents-and-tools/tool-use/implement-tool-use#providing-tool-use-examples
+        models_that_support_tool_examples = (
+            'claude-sonnet-4-5',
+            'claude-sonnet-4-6',
+            'claude-opus-4-5',
+            'claude-opus-4-6',
+        )
+        supports_tool_examples = model_name.startswith(models_that_support_tool_examples)
```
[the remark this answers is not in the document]
adtyavrdhn: Valid. The stale model allowlist was removed. Anthropic provider profiles now enable standard tool examples without enumerating model IDs, including current and future models, with explicit coverage for a future-style model ID.

pydantic_ai_slim/pydantic_ai/_griffe.py
```diff
@@ -65,18 +66,38 @@ def doc_descriptions(
     if main := next((p for p in sections if p.kind == DocstringSectionKind.text), None):
         main_desc = main.value
 
+    formatted_examples = ''
+    if examples := next((p for p in sections if p.kind == DocstringSectionKind.examples), None):
+        examples_content: list[str] = []
+        for source, output in examples.value:
+            if isinstance(source, DocstringSectionKind):  # pragma: no branch
```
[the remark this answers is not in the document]
adtyavrdhn: The Griffe examples-section change was removed from this PR, including this branch pragma. Structured tool examples are now supplied only through the explicit `examples` API.

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -32,7 +32,12 @@
     ToolOutput,
     _OutputSpecItem,  # type: ignore[reportPrivateUsage]
 )
-from .tools import GenerateToolJsonSchema, ObjectJsonSchema, ToolDefinition
+from .tools import (
+    GenerateToolJsonSchema,
+    ObjectJsonSchema,
+    ToolDefinition,
+    _process_examples,  # pyright: ignore[reportPrivateUsage]
```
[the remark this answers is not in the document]
adtyavrdhn: Valid and fixed in `a4267c5d9`. Serialization now lives in `_utils.process_tool_examples`; both tool and output paths use the shared private module, and the cross-module private import plus Pyright suppression are gone.

tests/models/test_anthropic.py
adtyavrdhn: The cassette has been recorded and replaced by the current `claude-opus-4-6` VCR coverage, which asserts `input_examples` in each recorded request body.

pydantic_ai_slim/pydantic_ai/_griffe.py
adtyavrdhn: The docstring parsing change was removed from this PR to keep the feature scoped. Native `input_examples` now comes only from the explicit structured `examples` API, so there is no brittle conversion from prose or doctest content.

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -340,6 +340,14 @@ def prepare_request(
             stripped = {k: v for k, v in model_settings.items() if k != 'thinking'}
             model_settings = cast(ModelSettings, stripped) if stripped else None
 
+        # Apply examples to description if not natively supported by the model
+        if not self.profile.supports_tool_examples:
+            params = replace(
+                params,
+                function_tools=[self._append_tool_examples(t) for t in params.function_tools],
+                output_tools=[self._append_tool_examples(t) for t in params.output_tools],
+            )
```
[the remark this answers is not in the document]
adtyavrdhn: Valid and fixed. All profile reads use dictionary access with `.get('supports_tool_examples', False)`, with runtime and type-check coverage.

pydantic_ai_slim/pydantic_ai/profiles/__init__.py
```diff
@@ -108,6 +108,9 @@ class ModelProfile(TypedDict, total=False):
     thinking_tags: tuple[str, str]
     """The tags used to indicate thinking parts in the model's output. Default: [`DEFAULT_THINKING_TAGS`][pydantic_ai.profiles.DEFAULT_THINKING_TAGS]."""
 
+    supports_tool_examples: bool = False
+    """Whether the model natively supports providing structured tool examples."""
+
```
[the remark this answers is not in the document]
adtyavrdhn: Valid and fixed. `supports_tool_examples` is now a bare optional `TypedDict` item, with defaults handled at profile merge and access sites.

docs/output.md
```diff
@@ -357,6 +357,36 @@ print(repr(result.output))
 
 _(This example is complete, it can be run "as is")_
 
+##### Output Examples
+
+You can provide examples of correct tool usage to improve the model's understanding of your tool's expected input. Examples can be provided in two ways:
+
+1. **Structured Examples**: Pass a list of examples to the `examples` parameter in [`@agent.tool`][pydantic_ai.agent.Agent.tool], [`@agent.tool_plain`][pydantic_ai.agent.Agent.tool_plain], or the [`Tool`][pydantic_ai.tools.Tool] class directly.
+2. **Docstring Examples**: Include an `Examples:` section in the tool's docstring.
```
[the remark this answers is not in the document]
adtyavrdhn: Valid and fixed. The output documentation now uses `ToolOutput(..., examples=...)`, explains semantic output values, and uses the distinct `output_examples.py` title.

pydantic_ai_slim/pydantic_ai/providers/anthropic.py
```diff
@@ -62,7 +62,13 @@ def model_profile(model_name: str) -> ModelProfile | None:
         if bedrock_provider == 'anthropic':
             model_name = base_model_name
         profile = anthropic_model_profile(model_name)
-        return AnthropicModelProfile(json_schema_transformer=AnthropicJsonSchemaTransformer).update(profile)
+        result = AnthropicModelProfile(json_schema_transformer=AnthropicJsonSchemaTransformer).update(profile)
+        if bedrock_provider == 'anthropic':
+            # `anthropic_model_profile` enables `supports_tool_examples` for the direct Anthropic API, where
+            # `input_examples` is a standard tool field. Via Bedrock it's still beta-gated behind the
+            # `tool-examples-2025-10-29` header (which we don't send), so fall back to the description instead.
+            result = replace(result, supports_tool_examples=False)
```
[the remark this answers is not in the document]
adtyavrdhn: Valid concern and addressed with client-based transport gating. Direct Anthropic, Vertex, Foundry, and Bedrock Mantle use the standard `input_examples` field; legacy `AsyncAnthropicBedrock` falls back to descriptions. Transport tests cover the legacy and Mantle distinction, and the direct VCR asserts the wire payload.

docs/tools.md
```diff
@@ -375,6 +375,36 @@ _(This example is complete, it can be run "as is")_
...
+
+1. **Structured Examples**: Pass a list of examples to the `examples` parameter in [`@agent.tool`][pydantic_ai.agent.Agent.tool], [`@agent.tool_plain`][pydantic_ai.agent.Agent.tool_plain], or the [`Tool`][pydantic_ai.tools.Tool] class directly.
+2. **Docstring Examples**: Include an `Examples:` section in the tool's docstring.
+
+Anthropic models natively support structured examples via the `input_examples` parameter in their API payload. For models that do not natively support this feature, Pydantic AI will gracefully fall back to serializing the structured examples as JSON and appending them to the tool's text description.
+
+```python {title="tool_examples.py"}
+from pydantic_ai import Agent
+
+agent = Agent('anthropic:claude-opus-4-6')
```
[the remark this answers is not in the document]
adtyavrdhn: Not applicable to the current docs test setup. The documentation suite installs all provider extras, this snippet constructs but does not run the Anthropic model, and the example matrix passes on Python 3.11 through 3.14 without credentials.

pydantic_ai_slim/pydantic_ai/toolsets/function.py
```diff
@@ -188,6 +189,7 @@ def tool(
         strict: bool | None = None,
         sequential: bool | None = None,
         requires_approval: bool | None = None,
+        examples: list[Any] | None = None,
```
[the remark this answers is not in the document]
adtyavrdhn: Valid and fixed. `FunctionToolset.tool()` now forwards `examples` to `add_function()`, with a regression test covering decorator forwarding and positional API compatibility.

DouweM:
Closed, see issue.

PR #3786, opened 2025-12-20
Add response-based fallback support for `FallbackModel`

### LINKED ISSUE #3640: Response-Based Fallback for FallbackModel
### Description

# Feature Request: Response-Based Fallback for FallbackModel

## Summary

Add a `fallback_on_response` parameter to `FallbackModel` that allows fallback decisions based on inspecting the `ModelResponse` content, not just exceptions. This enables fallback when a model returns a successful HTTP response but the semantic content indicates failure (e.g., a builtin tool like `web_fetch` failed to retrieve a URL).

## Motivation

Currently, `FallbackModel` only supports exception-based fallback via the `fallback_on` parameter:

```python
fallback_on: Callable[[Exception], bool] | tuple[type[Exception], ...] = (ModelAPIError,)
```

This works well for:
- API errors (rate limits, 5xx responses)
- Network failures
- Authentication issues

However, it **cannot handle semantic failures** where:
- The model returns HTTP 200 (no exception raised)
- But the response content indicates the operation failed

### Real-World Example: `WebFetchTool` with Google Models

When using `WebFetchTool` with Google's Gemini models, the model may successfully return a response, but the `BuiltinToolReturnPart` indicates the URL fetch failed:

```python
BuiltinToolReturnPart(
    tool_name='web_fetch',
    content=[{
        'uri': 'https://example.com',
        'url_retrieval_status': 'URL_RETRIEVAL_STATUS_FAILED'  # Not SUCCESS!
    }]
)
```

In this case, I want to fallback to Anthropic's model (which has different web fetching capabilities), but there's no exception to catch.

### Current Workaround

Today, I have to implement manual fallback logic outside of `FallbackModel`:

```python
class BrandURLExtractionAgent:
    def __init__(self):
        self._google_agent = Agent(model=google_model, builtin_tools=[WebFetchTool()])
        self._anthropic_agent = Agent(model=anthropic_model, builtin_tools=[WebFetchTool()])

    async def get_brand_summary(self, url: str) -> str | None:
        # Try Google first
        try:
            result = await self._google_agent.run(user_prompt=prompt)

            # Manual inspection of response parts
            if self._check_web_fetch_success(result.all_messages(), url):
                return result.output
            # Fall through to fallback...
        except Exception:
            pass

        # Manual fallback to Anthropic
        result = await self._anthropic_agent.run(user_prompt=prompt)
        return result.output

    def _check_web_fetch_success(self, messages: list[ModelMessage], url: str) -> bool:
        for message in messages:
            if isinstance(message, ModelRequest):
                for part in message.parts:
                    if isinstance(part, BuiltinToolReturnPart) and part.tool_name == "web_fetch":
                        # Check if URL was successfully retrieved...
                        pass
        return False
```

This is verbose, error-prone, and doesn't benefit from `FallbackModel`'s clean abstraction.

## Proposed API

Add a new `fallback_on_response` parameter to `FallbackModel.__init__`:

```python
from collections.abc import Callable
from pydantic_ai.messages import ModelMessage, ModelResponse

class FallbackModel(Model):
    def __init__(
        self,
        default_model: Model | KnownModelName | str,
        *fallback_models: Model | KnownModelName | str,
        fallback_on: Callable[[Exception], bool] | tuple[type[Exception], ...] = (ModelAPIError,),
        # NEW PARAMETER:
        fallback_on_response: Callable[[ModelResponse, list[ModelMessage]], bool] | None = None,
    ):
        ...
```

### Parameter Signature

```python
fallback_on_response: Callable[[ModelResponse, list[ModelMessage]], bool] | None = None
```

- **`response: ModelResponse`** - The model's response to inspect
- **`messages: list[ModelMessage]`** - Full message history (needed because `BuiltinToolReturnPart` lives in `ModelRequest`, not `ModelResponse`)
- **Returns `bool`** - `True` to trigger fallback, `False` to accept the response

### Usage Example

```python
    BuiltinToolReturnPart,
    ModelMessage,
    ModelRequest,
    ModelResponse,
)
from pydantic_ai.models.fallback import FallbackModel
from pydantic_ai.models.google import GoogleModel


def web_fetch_failed(response: ModelResponse, messages: list[ModelMessage]) -> bool:
    """Return True if web_fetch tool failed to retrieve content."""
    for message in messages:
        if not isinstance(message, ModelRequest):
            continue
        for part in message.parts:
            if isinstance(part, BuiltinToolReturnPart) and part.tool_name == "web_fetch":
                content = part.content
                if isinstance(content, list):
                    for item in content:
                        status = item.get("url_retrieval_status")
                        if status and status != "URL_RETRIEVAL_STATUS_SUCCESS":
                            return True  # Trigger fallback
    return False  # Accept response


google_model = GoogleModel('gemini-2.0-flash')
anthropic_model = AnthropicModel('claude-3-5-haiku-latest')

fallback_model = FallbackModel(
    google_model,
    anthropic_model,
    fallback_on_response=web_fetch_failed,
)

agent = Agent(
    model=fallback_model,
    builtin_tools=[WebFetchTool()],
)

# Now if Google's web_fetch fails, automatically falls back to Anthropic!
result = await agent.run("Summarize https://example.com")
```

### Changes to `FallbackModel.request()`

```python
async def request(
    self,
    messages: list[ModelMessage],
    model_settings: ModelSettings | None,
    model_request_parameters: ModelRequestParameters,
) -> ModelResponse:
    exceptions: list[Exception] = []

    for model in self._models:
        try:
            response = await model.request(messages, model_settings, model_request_parameters)

            # NEW: Check response-based fallback condition
            if self._fallback_on_response is not None:
                if self._fallback_on_response(response, messages):
                    # Optionally log: "Fallback triggered by response inspection"
                    continue  # Try next model

            return response

        except Exception as e:
            if self._should_fallback(e):
                exceptions.append(e)
            else:
                raise

    raise FallbackExceptionGroup("All models failed", exceptions)
```

### Streaming Considerations

For `request_stream()`, response-based fallback is more complex since the response arrives incrementally. Options:

1. **Don't support for streaming initially** - Document that `fallback_on_response` only works with non-streaming requests
2. **Inspect after stream completes** - Buffer the full response, then check (but this defeats the purpose of streaming for the fallback check)
3. **Support a streaming-aware callback** - More complex API

I'd recommend starting with option 1 (non-streaming only) and expanding later if there's demand.

### Alternative 1: Extend `fallback_on` to accept response

```python
fallback_on: Callable[[Exception | ModelResponse], bool] | ...
```

**Rejected because:** Mixing exception and response handling in one callback is confusing and breaks existing type signatures.

### Alternative 2: Custom exception wrapping

Wrap response inspection failures in a custom exception that `fallback_on` can catch.

**Rejected because:** Requires users to raise synthetic exceptions, which is awkward and doesn't fit the "successful response with bad content" mental model.

### Alternative 3: Output validator with retry

Use `@agent.output_validator` to raise `ModelRetry`.

**Rejected because:** This triggers retry with the *same* model, not fallback to a *different* model. Also, the inspection often needs to happen at the message/part level, not the final output level.

## Additional Context

This pattern is useful beyond `WebFetchTool`. Other use cases:

- **Citation validation**: Fallback if the model's response doesn't include expected citations
- **Tool call validation**: Fallback if a required tool wasn't called
- **Content quality checks**: Fallback if response is too short, contains refusals, or lacks required structure
- **Provider-specific quirks**: Handle cases where one provider returns empty content while another handles the same prompt fine

## Checklist

- [ ] Add `fallback_on_response` parameter to `FallbackModel.__init__`
- [ ] Implement response inspection in `FallbackModel.request()`
- [ ] Add logging when response-based fallback is triggered
- [ ] Document the feature in the FallbackModel docs
- [ ] Add unit tests for response-based fallback
- [ ] Document streaming limitation (if not supported initially)

If this is something you would accept, I'd be interested in contributing it :)

### References

- #516 - Original FallbackModel feature request
- #2837 - FallbackModel doesn't handle `UnexpectedModelBehavior` during response handling (related: response-level issues not caught)
- #2119 - Per-model settings for FallbackModel (resolved, shows precedent for extending FallbackModel API)

sarth6:
## Summary

Extends `FallbackModel`'s `fallback_on` parameter to support response-based fallback (non-streaming).

Closes #3640

## Motivation

Some models return successful HTTP responses but indicate failure in the response content. For example, Google's `WebFetchTool` returns `URL_RETRIEVAL_STATUS_FAILED` instead of throwing an exception. The existing exception-based fallback can't handle these cases.

## Changes

**Unified `fallback_on` API:**
The `fallback_on` parameter now accepts:
- A tuple of exception types: `(ModelAPIError, RateLimitError)` (backward compatible)
- A single exception type: `ModelAPIError` (backward compatible)
- An exception handler (sync or async): `lambda exc: isinstance(exc, MyError)` (backward compatible)
- A response handler (sync or async): `def check(r: ModelResponse) -> bool` (new)
- A list mixing all of the above: `[ModelAPIError, exc_handler, response_handler]` (new)

Handler type is auto-detected by inspecting type hints on the first parameter:
- If the first parameter is hinted as `ModelResponse`, it's a response handler
- Otherwise (including untyped handlers and lambdas), it's an exception handler

**Scope:**
- Non-streaming: Full support for both exception and response-based fallback
- Streaming: Exception-based fallback for initialization errors, plus response-based fallback after stream consumption (raises `ResponseRejected` in `FallbackExceptionGroup`; no transparent retry since events are already consumed)

**Documentation:**
- Added finish reason example showing response-based fallback on bad `FinishReason`
- Added built-in tool failure example with Google's `WebFetchTool`
- Added mixed handler example combining exception types and response handlers
- Added warnings explaining that a solo response handler replaces the default `(ModelAPIError,)` exception fallback
- Added per-model settings example showing provider-specific `ModelSettings`

**Code organization:**
- Moved `get_first_param_type` helper from `_utils.py` to `_function_schema.py` to keep Pydantic internal API usage contained

**Error handling improvements:**
- `FallbackExceptionGroup` now includes rejected responses in its error message
- Warning when `fallback_on` is empty (prevents silent misconfiguration)

## Example

```python
from pydantic_ai.messages import ModelResponse
from pydantic_ai.models.fallback import FallbackModel
from pydantic_ai.exceptions import ModelAPIError


def web_fetch_failed(response: ModelResponse) -> bool:
    """Check if a web_fetch built-in tool failed."""
    for call, result in response.builtin_tool_calls:
        if call.tool_name == 'web_fetch':
            # inspect result.content for failure indicators
            return True
    return False


# Auto-detected as response handler via ModelResponse type hint
fallback_model = FallbackModel(
    'google:gemini-2.5-flash',
    'anthropic:claude-sonnet-4-5',
    fallback_on=web_fetch_failed,
)

# Or mix with exception types for both exception and response fallback
fallback_model = FallbackModel(
    'google:gemini-2.5-flash',
    'anthropic:claude-sonnet-4-5',
    fallback_on=[
        ModelAPIError,
        lambda exc: 'rate limit' in str(exc).lower(),
        web_fetch_failed,  # Auto-detected via type hint
    ],
)
```

## Test plan

- 44 tests covering exception, response, and mixed fallback
- Both streaming and non-streaming scenarios
- Type hint auto-detection tests (including callable classes and forward references)
- Edge cases: exception propagation, empty handler warning, rejected response error messages


pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -31,19 +36,36 @@ class FallbackModel(Model):
...
+                fallback. Use `fallback_on_part` if you need to handle mid-stream failures.
+            fallback_on_response: A callable that inspects the model response and message history,
+                returning `True` if fallback should be triggered. This enables fallback based on
+                response content (e.g., a builtin tool indicating failure) rather than exceptions.
+            fallback_on_part: A callable that inspects each model response part during streaming,
+                returning `True` if fallback should be triggered. This enables early abort when
+                a failure condition is detected (e.g., a builtin tool failure), saving tokens by
+                not consuming the rest of a doomed response. Only applies to streaming requests.
+                Note: The response is buffered until validation completes, so the caller receives
+                events after the full response is validated, not progressively.
```
sarth6: `fallback_on_part` isn't relevent to my use case but I felt it would be useful for users looking to save tokens when using response-based fallback with large responses. However, if it's undesirable I can remove it 👍

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -31,19 +36,36 @@ class FallbackModel(Model):
...
             fallback_models: The names or instances of the fallback models to use upon failure.
             fallback_on: A callable or tuple of exceptions that should trigger a fallback.
+                For streaming requests, this only catches exceptions during stream initialization
+                (e.g., connection errors, authentication failures). Exceptions that occur mid-stream
+                after events have started flowing will propagate to the caller without triggering
+                fallback. Use `fallback_on_part` if you need to handle mid-stream failures.
+            fallback_on_response: A callable that inspects the model response and message history,
+                returning `True` if fallback should be triggered. This enables fallback based on
+                response content (e.g., a builtin tool indicating failure) rather than exceptions.
+            fallback_on_part: A callable that inspects each model response part during streaming,
```
sarth6: I do wonder if this is the best API, or if we should somehow make `fallback_on` support exceptions, responses, and parts, all in one.

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -156,3 +223,62 @@ def fallback_condition(exception: Exception) -> bool:
...
     return fallback_condition
+
+
+def _raise_fallback_exception_group(
+    exceptions: list[Exception], response_rejections: int, part_rejections: int = 0
+) -> NoReturn:
+    """Raise a FallbackExceptionGroup combining exceptions and rejections."""
+    all_errors: list[Exception] = list(exceptions)
+    if part_rejections > 0:
+        all_errors.append(RuntimeError(f'{part_rejections} model(s) rejected by fallback_on_part during streaming'))
```
sarth6: should we add more detail here on which part failed?

docs/models/overview.md
DouweM, line 1: Note to self to review docs later

pydantic_ai_slim/pydantic_ai/models/fallback.py
DouweM: The use case is valid but I'm a lot more hesitant to introduce 2 new fields than 1 😅 

Note that the `ModelResponsePartsManager.get_parts()` method can already be used to build a `ModelResponse` at any stage of the stream, so it should be possible to make the same `fallback_on_response` work with streaming as well. We would want to add an `incomplete` field or something in that case, so that the handler can tell the difference.

That seems like enough of a rabbit hole that it shouldn't block the release of the main feature, but feel free to look into it in a separate issue/PR.

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -31,19 +36,36 @@ class FallbackModel(Model):
...
             default_model: The name or instance of the default model to use.
             fallback_models: The names or instances of the fallback models to use upon failure.
             fallback_on: A callable or tuple of exceptions that should trigger a fallback.
+                For streaming requests, this only catches exceptions during stream initialization
+                (e.g., connection errors, authentication failures). Exceptions that occur mid-stream
+                after events have started flowing will propagate to the caller without triggering
+                fallback. Use `fallback_on_part` if you need to handle mid-stream failures.
+            fallback_on_response: A callable that inspects the model response and message history,
+                returning `True` if fallback should be triggered. This enables fallback based on
+                response content (e.g., a builtin tool indicating failure) rather than exceptions.
```
DouweM: Let's also mention the finish reason as an example

pydantic_ai_slim/pydantic_ai/models/fallback.py
DouweM: That would be ideal, if we can figure it out in a backward compatible way :)

If we change `Callable[[Exception], bool]` to `Callable[[Exception | ModelResponse], bool]`, users' existing handlers would likely start raising errors, so we'd have to check the function signature type hint. But that wouldn't work with lambdas, and type hints are of course not required, so basically we would only send in `ModelResponse` if it is an explicitly hinted function.

That's not the end of the world, but a little awkward, also because the user would still need to check `isinstance(...)` etc.

So perhaps it'd be better to support a _list_ of exception types, exception handler functions, and response handler functions? And if the type of a function is not specified, we assume it's an exception handler function?

That way it'd also be easier to reuse checks like your `builtin_tool_error`.

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -1,24 +1,118 @@
...
+ResponseHandler = Callable[[ModelResponse], bool]
+
+
+@dataclass
+class OnResponse:
+    """Wrapper to explicitly mark a callable as a response handler for fallback_on.
+
+    Most response handlers are auto-detected via type hints. Use this wrapper when:
+    - Your handler is a lambda (can't have type hints)
+    - Your handler is untyped and you want it treated as a response handler
```
sarth6: if this is overkill we could remove it and not support untyped response handlers

docs/models/overview.md
```diff
@@ -235,3 +235,101 @@ passing a custom `fallback_on` argument to the `FallbackModel` constructor.
 
 !!! note
     Validation errors (from [structured output](../output.md#structured-output) or [tool parameters](../tools.md)) do **not** trigger fallback. These errors use the [retry mechanism](../agents.md#reflection-and-self-correction) instead, which re-prompts the same model to try again. This is intentional: validation errors stem from the non-deterministic nature of LLMs and may succeed on retry, whereas API errors (4xx/5xx) generally indicate issues that won't resolve by retrying the same request.
+
+!!! note "Streaming limitation"
+    For streaming requests, only exception-based fallback is currently supported, and only for errors during stream **initialization** (e.g., connection errors, authentication failures). If an exception occurs mid-stream after events have started flowing, it will propagate to the caller without triggering fallback. Response-based fallback for streaming is planned for a future release.
```
sarth6: I intend to put up a follow-up PR to support streaming using `ModelResponsePartsManager.get_parts()` as @DouweM mentioned, but decided not to block this feature on that. We can always support an optional `incomplete` `bool` parameter to the response handler later, will be backwards compatible.

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -1,24 +1,128 @@
...
+    from ..messages import ModelMessage
     from ..settings import ModelSettings
 
+# Type aliases for handlers
+ExceptionHandler = Callable[[Exception], bool]
+ResponseHandler = Callable[[ModelResponse], bool]
+
+
+@dataclass
+class OnResponse:
```
sarth6: if `OnResponse` adds cognitive overhead with little utility (I'm not sure how often users will prefer to use it over type annotated functions) we could remove it and not support untyped response handlers, I don't have a strong preference either way.

sarth6:
The CI failure will require a maintainer re-run, see https://github.com/pydantic/pydantic-ai/pull/3846

sarth6:
bump, not stale, just didn't want to ping Douwe on vacation :D

pydantic_ai_slim/pydantic_ai/models/fallback.py
DouweM: Yeah let's remove this as we don't have it anywhere else.

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -1,24 +1,128 @@
...
+)
+
+
+def _is_response_handler(handler: Callable[..., Any]) -> bool:
+    """Check if a callable is a response handler based on type hints.
+
+    Returns True if the first parameter is type-hinted as ModelResponse.
+    Returns False otherwise (including if there are no type hints).
+
+    Note: Async response handlers are not supported. If an async callable is detected
```
DouweM: Could we support async handlers? I'd prefer to support both sync and async callables everywhere we can.

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -1,24 +1,128 @@
...
+        if not params:
+            return False
+
+        first_param = params[0]
+
+        # Try to get type hints - this resolves forward references automatically
+        hints = get_type_hints(handler)
+        param_type = hints.get(first_param.name)
+
+        if param_type is None:
```
DouweM: Check out the `pydantic_ai._function_schema._takes_ctx` helper, we may be able to use parts of it here and/or turn it into a single `takes(func, <type>)` helper

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -1,24 +1,128 @@
...
+            return False
+
+        # Handle parameterized generic types (e.g., Optional[ModelResponse])
```
DouweM: I think we should only support `ModelResponse` as that's what the typing requires. So no optionals or subclasses.

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -1,24 +1,128 @@
...
+        return False
+
+
+def _is_exception_type(value: Any) -> bool:
+    """Check if value is a single exception type."""
+    return isinstance(value, type) and issubclass(value, BaseException)
+
+
+def _is_exception_types_tuple(value: Any) -> bool:
+    """Check if value is a tuple of exception types (the old API format)."""
```
DouweM: "old" implies too much, it's still a perfectly fine supported value

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -30,28 +134,110 @@ class FallbackModel(Model):
...
+
+    def _parse_fallback_on(self, fallback_on: FallbackOn) -> None:
+        """Parse the fallback_on parameter into exception and response handlers."""
+        # Case 1: Tuple of exception types (backward compatible)
+        if _is_exception_types_tuple(fallback_on):
+            self._exception_handlers.append(_exception_types_to_handler(fallback_on))  # type: ignore[arg-type]
+            return
+
+        # Case 2: Single OnResponse wrapper (explicit response handler)
+        if isinstance(fallback_on, OnResponse):
```
DouweM: Please use `elif`s

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -30,28 +134,110 @@ class FallbackModel(Model):
...
+        if isinstance(fallback_on, Sequence) and not isinstance(fallback_on, (str, bytes)):
+            for item in fallback_on:
+                if isinstance(item, OnResponse):
+                    self._response_handlers.append(item.handler)
+                elif _is_exception_type(item):
+                    self._exception_handlers.append(_exception_types_to_handler((item,)))  # type: ignore[arg-type]
+                elif callable(item):
+                    self._add_handler(item)
+                else:
+                    raise TypeError(
```
DouweM: No need to raise explicit errors for things the types wouldn't allow anyway. We can use `assert_never` in `else` to ensure at a typing level that we should never get there because all the options have been exhaustively covered in `if`/`elif` branches, and error if we do.

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -77,21 +263,26 @@ async def request(
         In case of failure, raise a FallbackExceptionGroup with all exceptions.
         """
         exceptions: list[Exception] = []
+        rejected_models: list[str] = []
```
DouweM: It's unclear from the variable name that this is about models that failed response validation. Maybe we should store the entire responses, since we can read the model name off it later anyway?

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -149,10 +347,33 @@ def _set_span_attributes(self, model: Model, model_request_parameters: ModelRequ
...
+
+    Args:
+        exceptions: List of exceptions raised by models.
+        rejected_models: List of model names whose responses were rejected by fallback_on handlers.
+    """
+    all_errors: list[Exception] = list(exceptions)
+    if rejected_models:
+        models_str = ', '.join(f"'{m}'" for m in rejected_models)
+        all_errors.append(
+            RuntimeError(f'{len(rejected_models)} model response(s) rejected by fallback_on handler: {models_str}')
```
DouweM: Do we need to list the models? We only get here if all responses failed validation (same as if all models raised exceptions), so I don't think it's worth basically repeating all the wrapped models' names.

pydantic_ai_slim/pydantic_ai/models/fallback.py
sarth6: thanks didn't know about `assert_never`!

docs/models/overview.md
```diff
@@ -236,3 +235,118 @@ passing a custom `fallback_on` argument to the `FallbackModel` constructor.
 
 !!! note
     Validation errors (from [structured output](../output.md#structured-output) or [tool parameters](../tools.md)) do **not** trigger fallback. These errors use the [retry mechanism](../agents.md#reflection-and-self-correction) instead, which re-prompts the same model to try again. This is intentional: validation errors stem from the non-deterministic nature of LLMs and may succeed on retry, whereas API errors (4xx/5xx) generally indicate issues that won't resolve by retrying the same request.
+
```
sarth6, line 315: I see why the FallbackModel docs were put here in the first place - it is a model after all! But I predict the average user looking to use it won't be thinking about models and will simply be thinking about reliability and error handling.

That begs the question, should docs be organized by code concept or by user intent?

If the latter, should we reorganize FallbackModel into a separate doc designed to help users make their agents more reliable? If so, perhaps there are other pydantic ai features we'd want moved to that doc.

pydantic_ai_slim/pydantic_ai/_utils.py
```diff
@@ -414,6 +415,45 @@ def is_async_callable(obj: Any) -> Any:
     return inspect.iscoroutinefunction(obj) or (callable(obj) and inspect.iscoroutinefunction(obj.__call__))  # type: ignore
 
 
+def get_first_param_type(callable_obj: Callable[..., Any]) -> Any | None:
```
sarth6, line 429: Refactored this out so it could be re-used in _function_schema.py and fallback.py

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -30,28 +73,104 @@ class FallbackModel(Model):
...
+        if _is_response_handler(handler):
+            self._response_handlers.append(handler)
         else:
-            self._fallback_on = fallback_on
+            self._exception_handlers.append(handler)
+
+    async def _should_fallback_on_exception(self, exc: Exception) -> bool:
+        """Check if any exception handler wants to trigger fallback."""
+        for handler in self._exception_handlers:
+            if is_async_callable(handler):
```
sarth6: Before I was using `inspect` to check if the handler was awaitable but I found `is_async_callable` in `_utils.py` while doing the `get_first_param_type` refactor, let me know if I should switch back

dsfaccini:
same comment as the others :) please resolve and sync deps locally

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -30,28 +73,104 @@ class FallbackModel(Model):
...
 
         Args:
             default_model: The name or instance of the default model to use.
             fallback_models: The names or instances of the fallback models to use upon failure.
-            fallback_on: A callable or tuple of exceptions that should trigger a fallback.
+            fallback_on: Conditions that trigger fallback to the next model. Accepts:
+
+                - A tuple of exception types: `(ModelAPIError, RateLimitError)`
+                - An exception handler: `lambda exc: isinstance(exc, MyError)`
+                - A response handler: `def check(r: ModelResponse) -> bool`
```
DouweM: Worth specifying explicitly that these can be async

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -30,28 +73,104 @@ class FallbackModel(Model):
...
+                - A sequence mixing all of the above: `[ModelAPIError, exc_handler, response_handler]`
+
+                Handler type is auto-detected by inspecting type hints on the first parameter.
+                If the first parameter is hinted as `ModelResponse`, it's a response handler.
+                Otherwise (including untyped handlers and lambdas), it's an exception handler.
+
+                Note: For streaming requests, only exception-based fallback is supported, and only for
+                errors during stream initialization. Response handlers are ignored for streaming.
+                A future release will add streaming support for response-based fallback.
```
DouweM: We should link to an issue; I suggest creating one already

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -30,28 +73,104 @@ class FallbackModel(Model):
...
-        if isinstance(fallback_on, tuple):
-            self._fallback_on = _default_fallback_condition_factory(fallback_on)
+        # Parse fallback_on into exception handlers and response handlers
+        self._exception_handlers = []
+        self._response_handlers = []
+        self._parse_fallback_on(fallback_on)
+
+    def _parse_fallback_on(self, fallback_on: FallbackOn) -> None:
+        """Parse the fallback_on parameter into exception and response handlers."""
+        if _is_exception_types_tuple(fallback_on):
```
DouweM: This can just be `isinstance(tuple)` right? Since the typing doesn't allow any tuples other than those containing exceptions, we don't necessarily have to check the types of the tuple items

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -30,28 +73,104 @@ class FallbackModel(Model):
...
+            self._response_handlers.append(handler)
         else:
-            self._fallback_on = fallback_on
+            self._exception_handlers.append(handler)
+
+    async def _should_fallback_on_exception(self, exc: Exception) -> bool:
+        """Check if any exception handler wants to trigger fallback."""
+        for handler in self._exception_handlers:
+            if is_async_callable(handler):
+                result = await cast(Awaitable[bool], handler(exc))
```
DouweM: Do we need the cast here? I don't see it in other places where we use `is_async_callable`. Maybe the type hint needs to be changed from `Callable[..., bool | Awaitable[bool]]` to `Callable[..., Awaitable[bool]] | Callable[..., bool]`

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -30,28 +73,104 @@ class FallbackModel(Model):
...
+            else:
+                result = cast(bool, handler(exc))
```
DouweM: cast shouldn't be necessary here either

docs/models/overview.md
DouweM: Yeah I think a doc on reliability / error handling etc would be better. No need to do that in this PR though.

docs/models/overview.md
```diff
@@ -235,3 +235,118 @@ passing a custom `fallback_on` argument to the `FallbackModel` constructor.
...
+!!! note "Streaming limitation"
+    For streaming requests, only exception-based fallback is currently supported, and only for errors during stream **initialization** (e.g., connection errors, authentication failures). If an exception occurs mid-stream after events have started flowing, it will propagate to the caller without triggering fallback. Response-based fallback for streaming is planned for a future release.
```
DouweM: We haven't introduced the concept of response-based fallbacks at this point yet, so this will read confusingly. 

Can you introduce the concept earlier, or move this down?

docs/models/overview.md
```diff
@@ -235,3 +235,118 @@ passing a custom `fallback_on` argument to the `FallbackModel` constructor.
...
+
+### Response-Based Fallback
+
+In addition to exception-based fallback, you can also trigger fallback based on the **content** of a model's response. This is useful when a model returns a successful HTTP response (no exception), but the response content indicates a semantic failure — for example, an unexpected finish reason or a built-in tool reporting failure.
+
+The `fallback_on` parameter accepts:
```
DouweM, line 324: I'd move this much further up, we should introduce that FallbackModel can fall back on both exceptions and semantically invalid responses up front.

pydantic_ai_slim/pydantic_ai/models/fallback.py
sarth6: That's true, I just thought it might be safer to be specific in case that invariant changes with a later code change, but `isinstance(tuple)` works

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -30,28 +66,104 @@ class FallbackModel(Model):
...
+        for handler in self._exception_handlers:
+            if is_async_callable(handler):
+                result = await handler(exc)
+            else:
+                result = handler(exc)
+            if result:
+                return True
+        return False
+
+    async def _should_fallback_on_response(self, response: ModelResponse) -> bool:
```
DouweM: These 2 methods could be turned into one with 2 overloads for the Exception and ModelResponse cases :)

docs/models/overview.md
```diff
@@ -236,3 +239,118 @@ passing a custom `fallback_on` argument to the `FallbackModel` constructor.
...
+```python {title="fallback_on_finish_reason.py"}
+from pydantic_ai import Agent
+from pydantic_ai.messages import FinishReason, ModelResponse
+from pydantic_ai.models.fallback import FallbackModel
+
+
+def bad_finish_reason(response: ModelResponse) -> bool:
+    """Fallback if the model stopped due to length limit, content filter, or error."""
+    reason: FinishReason | None = response.finish_reason
+    # Trigger fallback for problematic finish reasons
```
DouweM, line 346: This raises the question of why this is not the default behavior, so it's worth specifying that Pydantic AI will currently on "reject" a `ModelResponse` by if it has 0 parts, not if it just has an error finish reason. The logic here: https://github.com/pydantic/pydantic-ai/blob/3cd52974fd2b4c0a5938d017712e92842144676e/pydantic_ai_slim/pydantic_ai/_agent_graph.py#L619-L673. We raise exceptions if the finish reason was length or content_filter, otherwise we retry.

That's not currently documented at all so worth doing around here!

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/models/fallback.py
sarth6: good call missed that

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -149,10 +273,32 @@ def _set_span_attributes(self, model: Model, model_request_parameters: ModelRequ
...
+
+    Args:
+        exceptions: List of exceptions raised by models.
+        rejected_responses: List of responses that were rejected by fallback_on handlers.
+    """
+    if rejected_responses:
+        rejection_error = RuntimeError(f'{len(rejected_responses)} model response(s) rejected by fallback_on handler')
+        all_errors: list[Exception] = [*exceptions, rejection_error]
+    else:
+        all_errors = exceptions
```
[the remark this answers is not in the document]
sarth6: no we had this before, we decided the number of responses is fine as the user already has the model info available!

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -1,24 +1,60 @@
...
+    first_param_type = get_first_param_type(handler)
+    if first_param_type is None:
+        return False
+    # Only support exact ModelResponse type (no Optional, no subclasses)
+    return first_param_type is ModelResponse or get_origin(first_param_type) is ModelResponse
+
+
+def _is_exception_type(value: Any) -> TypeGuard[type[Exception]]:
+    """Check if value is a single exception type."""
+    return isinstance(value, type) and issubclass(value, BaseException)
```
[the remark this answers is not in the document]
sarth6: @DouweM the existing pattern is to use `Exception` here, but we could also open up fallback exceptions to any subclass of `BaseException`, thoughts?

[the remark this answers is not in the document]
sarth6: For now I replaced with `issubclass(value, Exception)`

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -30,28 +66,110 @@ class FallbackModel(Model):
...
+                    # Types guarantee all items are exception types or callables
+                    assert_never(item)
+            # Warn if empty sequence was provided
+            if not self._exception_handlers and not self._response_handlers:
+                warnings.warn(
+                    'FallbackModel created with empty fallback_on list. '
+                    'All exceptions will propagate and all responses will be accepted. '
+                    'Consider using fallback_on=(ModelAPIError,) for default behavior.',
+                    UserWarning,
+                    stacklevel=4,
```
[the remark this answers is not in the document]
sarth6: I went along with this but honestly `model = create_model()` might be more confusing to me

sarth6:
Usually I wait until CI is green to request review but seems CI is not running, if I'm doing smth wrong please let me know, thanks!

docs/models/overview.md
```diff
@@ -304,3 +307,125 @@ passing a custom `fallback_on` argument to the `FallbackModel` constructor.
...
+    return reason in ('length', 'content_filter', 'error')
+
+
+fallback_model = FallbackModel(
+    'openai:gpt-4o-mini',
+    'anthropic:claude-sonnet-4-5',
+    fallback_on=bad_finish_reason,
+)
```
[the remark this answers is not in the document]
DouweM: We should stick to the old behavior of not auto-including `ModelAPIError`, but I agree the docs examples could make it clearer that this drops the original error fallback

DouweM:
@sarth6 Let me know when the auto-review bot is done finding issues and you want me to have a look again :)

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -30,28 +66,112 @@ class FallbackModel(Model):
     models: list[Model]
 
     _model_name: str = field(repr=False)
```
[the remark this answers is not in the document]
sarth6: It is a hallmark robot skill to be able to come up with arbitrary feedback no matter how many times the robot feedback has been addressed 😂

docs/models/overview.md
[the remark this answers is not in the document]
sarth6: PR description should just be updated

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -149,10 +281,32 @@ def _set_span_attributes(self, model: Model, model_request_parameters: ModelRequ
...
+        rejection_error = RuntimeError(f'{len(rejected_responses)} model response(s) rejected by fallback_on handler')
+        all_errors: list[Exception] = [*exceptions, rejection_error]
+    else:
+        all_errors = exceptions
+
+    if all_errors:
+        raise FallbackExceptionGroup('All models from FallbackModel failed', all_errors)
+    else:
+        raise FallbackExceptionGroup(
+            'All models from FallbackModel failed',
```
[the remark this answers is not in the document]
sarth6: I have already received feedback from the maintainer that we shouldn't do this. This is now the second time I've had to tell you you're wrong and given a thumbs down ☹️

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -1,24 +1,60 @@
...
+    """Check if a callable is a response handler based on type hints.
+
+    Returns True if the first parameter is type-hinted as ModelResponse.
+    Returns False otherwise (including if there are no type hints).
+    """
+    first_param_type = get_first_param_type(handler)
+    if first_param_type is None:
+        return False
+    # Only support exact ModelResponse type (no Optional, no subclasses)
+    return first_param_type is ModelResponse
```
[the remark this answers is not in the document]
sarth6: Now you're just making stuff up 😂

sarth6:
@DouweM ready for another review whenever you have time, thanks!

sarth6:
thanks @DouweM! Ready for another review when you have time

sarth6:
ci failure is unrelated

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -1,24 +1,61 @@
...
 from contextlib import AsyncExitStack, asynccontextmanager, suppress
 from dataclasses import dataclass, field
 from functools import cached_property
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, NoReturn, TypeGuard, overload
 
 from opentelemetry.trace import get_current_span
+from typing_extensions import assert_never
 
+from pydantic_ai._function_schema import _get_first_param_type  # pyright: ignore[reportPrivateUsage]
```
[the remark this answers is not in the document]
DouweM: Moving to `_utils` seems reasonable

docs/models/overview.md
```diff
@@ -304,3 +307,132 @@ passing a custom `fallback_on` argument to the `FallbackModel` constructor.
...
+            continue
+        # Content is a list of UrlMetadata dicts from Google
+        if not isinstance(result.content, list):
+            continue
+        for item in result.content:
+            meta = UrlMetadata.model_validate(item)
+            if meta.url_retrieval_status != UrlRetrievalStatus.URL_RETRIEVAL_STATUS_SUCCESS:
+                return True
+    return False
+
```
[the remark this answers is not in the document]
DouweM: This may be wrong, it doesn't have a great understanding yet of how the example tests work

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -149,10 +269,32 @@ def _set_span_attributes(self, model: Model, model_request_parameters: ModelRequ
...
+    else:
+        all_errors = exceptions
+
+    if all_errors:
+        raise FallbackExceptionGroup('All models from FallbackModel failed', all_errors)
+    else:
+        raise FallbackExceptionGroup(
+            'All models from FallbackModel failed',
+            [RuntimeError('No models available')],
+        )  # pragma: no cover
```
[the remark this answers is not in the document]
DouweM: I'm sure you've mentioned 1 before...

docs/models/overview.md
```diff
@@ -304,3 +307,132 @@ passing a custom `fallback_on` argument to the `FallbackModel` constructor.
...
+    To keep exception-based fallback alongside a response handler, pass them together as a list — see the [mixed example below](#combining-handlers).
+
+!!! note "Why check finish reasons explicitly?"
+    By default, Pydantic AI only rejects responses with **0 parts** (no content) and finish reasons `'length'` or `'content_filter'` — these raise exceptions immediately. Empty responses with other finish reasons (including `'error'`) trigger a retry instead.
+
+    For non-empty responses, **all finish reasons are accepted** — even `'error'` — because the response may contain partial results or useful information. This is why explicit checking via `fallback_on` is useful: it allows you to inspect the response content and make semantic decisions about whether to accept it or fallback to another model.
+
+    For details on default finish reason handling and retries, see [Reflection and Self-Correction](../agent.md#reflection-and-self-correction).
```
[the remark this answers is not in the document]
DouweM: I agree some of this is worth mentioning, but this is perhaps indeed not the right section.

sarth6:
@DouweM all bot review feedback addressed (aside from the ones that were already addressed or the ones you disagreed with)

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -30,28 +66,93 @@ class FallbackModel(Model):
     models: list[Model]
 
     _model_name: str = field(repr=False)
```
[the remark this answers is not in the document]
sarth6: unrelated

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -30,28 +66,93 @@ class FallbackModel(Model):
...
-            self._fallback_on = _default_fallback_condition_factory(fallback_on)  # pyright: ignore[reportUnknownArgumentType]
+            if fallback_on:
+                # Tuple of exception types (typing guarantees tuple contents are exception types)
+                self._exception_handlers.append(_exception_types_to_handler(fallback_on))  # type: ignore[arg-type]
+        elif _is_exception_type(fallback_on):
+            # Single exception type
+            self._exception_handlers.append(_exception_types_to_handler((fallback_on,)))
+        elif callable(fallback_on):
+            # Single callable - auto-detect by type hints
+            self._add_handler(fallback_on)
```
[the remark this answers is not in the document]
sarth6: this is an intentional design choice, I've documented it well in the docs as per maintainer feedback

sarth6:
@DouweM I think this one is close, devin is repeating old feedback that we addressed or dismissed - ready for another auto bot review

sarth6:
Looks like CI is flaky

sarth6:
Looks like the auto-bot timed out, hopefully not because my PR has so many issues it crashed 😂

sarth6:
@DouweM could you try adding the auto-review label again?

DouweM:
@sarth6 auto-review keeps failing 🤦🏻 I'll have a manual look tomorrow, appreciate the patience.

pydantic_ai_slim/pydantic_ai/_utils.py
```diff
@@ -425,6 +426,45 @@ def is_async_callable(obj: Any) -> Any:
...
+        call_func = getattr(type(callable_obj), '__call__', None)
+        if call_func is not None:
+            callable_for_hints = call_func
+        else:
+            return None  # pragma: no cover
+
+    try:
+        type_hints = _typing_extra.get_function_type_hints(_decorators.unwrap_wrapped_function(callable_for_hints))
+    except (NameError, TypeError, AttributeError):
+        return None
```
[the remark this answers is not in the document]
DouweM: No need to change it, I'm fine with it as is.

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -154,10 +268,23 @@ def _set_span_attributes(self, model: Model, model_request_parameters: ModelRequ
...
+def _raise_fallback_exception_group(exceptions: list[Exception], rejected_responses: list[ModelResponse]) -> NoReturn:
+    """Raise a FallbackExceptionGroup combining exceptions and response rejections.
+
+    Args:
+        exceptions: List of exceptions raised by models.
+        rejected_responses: List of responses that were rejected by fallback_on handlers.
+    """
+    all_errors = list(exceptions)
+    if rejected_responses:
+        all_errors.append(RuntimeError(f'{len(rejected_responses)} model response(s) rejected by fallback_on handler'))
```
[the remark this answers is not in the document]
DouweM: I agree `RuntimeError` is too broad; let me look at the code and suggest something better.

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -1,24 +1,60 @@
...
+    from ..messages import ModelMessage
     from ..settings import ModelSettings
 
+# Type aliases for handlers (support both sync and async)
+_ExceptionHandler = Callable[[Exception], Awaitable[bool]] | Callable[[Exception], bool]
+_ResponseHandler = Callable[[ModelResponse], Awaitable[bool]] | Callable[[ModelResponse], bool]
+
+
+# The unified fallback_on type
+_FallbackOn = (
```
DouweM: For API docs reasons, let's make this and the 2 handlers above public (no leading underscore) and add a docstring.

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -30,28 +66,94 @@ class FallbackModel(Model):
...
+            assert_never(fallback_on)  # type: ignore[arg-type]  # pyright can't narrow str/bytes exclusion
+
+        # Warn if no handlers were registered (empty tuple, empty list, etc.)
+        if not self._exception_handlers and not self._response_handlers:
+            warnings.warn(
+                'FallbackModel created with empty fallback_on. '
+                'All exceptions will propagate and all responses will be accepted. '
+                'Consider using fallback_on=(ModelAPIError,) for default behavior.',
+                UserWarning,
+                stacklevel=3,  # user(3) -> __init__(2) -> _parse_fallback_on(1)
```
DouweM: I'd rather have this be a `UserError` as it clearly indicates something wrong.

pydantic_ai_slim/pydantic_ai/models/fallback.py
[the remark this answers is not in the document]
DouweM: I like a new local `ResponseRejected` exception.

docs/models/overview.md
```diff
@@ -153,8 +153,11 @@ attributes showing the queue depth and configured limits. The `name` parameter o
 ## Fallback Model
 
 You can use [`FallbackModel`][pydantic_ai.models.fallback.FallbackModel] to attempt multiple models
-in sequence until one successfully returns a result. Under the hood, Pydantic AI automatically switches
-from one model to the next if the current model returns a 4xx or 5xx status code.
+in sequence until one succeeds. Pydantic AI can switch to the next model when the current model
+raises an exception (like a 4xx/5xx API error) **or** when the response content indicates a semantic
+failure (like a truncated response or a failed built-in tool call). This behavior is controlled by the
```
DouweM: Please add an empty line before "This behavior is controlled by ...", and make sure we mention the default behavior here, so people don't think they _have_ to use `fallback_on`.

docs/models/overview.md
```diff
@@ -305,6 +308,133 @@ passing a custom `fallback_on` argument to the `FallbackModel` constructor.
...
+    for call, result in response.builtin_tool_calls:
+        if call.tool_name != 'web_fetch':
+            continue
+        # Content is a list of UrlMetadata dicts from Google
+        if not isinstance(result.content, list):
+            continue
+        for item in result.content:
+            meta = UrlMetadata.model_validate(item)
+            if meta.url_retrieval_status != UrlRetrievalStatus.URL_RETRIEVAL_STATUS_SUCCESS:
+                return True
```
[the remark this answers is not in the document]
DouweM: I agree with simplifying this a bit

docs/models/overview.md
```diff
@@ -305,6 +308,133 @@ passing a custom `fallback_on` argument to the `FallbackModel` constructor.
...
+result = agent.run_sync('Summarize https://ai.pydantic.dev')
+print(result.output)
+"""
+Pydantic AI is a Python agent framework for building production-grade LLM applications.
+"""
+```
+
+Response handlers receive the [`ModelResponse`][pydantic_ai.messages.ModelResponse] returned by the model and should return `True` to trigger fallback to the next model, or `False` to accept the response.
+
+#### Combining Handlers { #combining-handlers }
```
DouweM: We shouldn't need the explicit  anchor, nor do I think that is actually supported syntax

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -117,16 +230,17 @@ async def request_stream(
                         model.request_stream(messages, model_settings, model_request_parameters, run_context)
                     )
                 except Exception as exc:
-                    if self._fallback_on(exc):
+                    if await self._should_fallback(exc):
                         exceptions.append(exc)
                         continue
                     raise exc  # pragma: no cover
 
+                # For streaming, we yield the response directly (no response-based fallback)
```
DouweM: I have a feeling this wouldn't be that complicated to support. the wrapped model's `request_stream` method yields a `StreamedResponse` object, which is iterated over using `__aiter__`, and has a `get()` method to get the `ModelResponse`. Instead of `yield`ing `response` directly, could we wrap it in a `StreamedResponse` object of our own that passes through all methods, but modifies `__aiter__` to call `get()` at the end and pass that through to the response fallback handler?

I think that's worth spending 15 minutes on; it'd be nice to not have to do this in a separate followup PR.

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/models/fallback.py
sarth6: @DouweM Thanks for the feedback! I had version of this implemented earlier in the PR but it buffered the streaming result which wasn't desirable, glad you think this won't be much effort - I'll investigate.

What do you think of also supporting mid-stream exception fallback? A couple options:

1. Detect-only: Wrapper catches mid-stream exceptions and checks response handlers after completion. Mid-stream exceptions that match fallback_on get collected into the `FallbackExceptionGroup` (same as init errors today). No actual fallback happens with this, just there to provide `FallbackExceptionGroup` "parity".

2. Transparent retry: Wrapper manages the model loop internally in `__aiter__`. On mid-stream failure or response rejection, it opens the next model’s stream from scratch and keeps yielding. Caller sees mixed events, but `get()` returns only the last successful model’s response. Good fit for workloads that use streaming for network stability but only care about the final result.

We could do this in a different PR too, let me know what you think!

DouweM: @sarth6 I think 2 would be useful in some cases, but it'd need to have a flag on `FallbackModel` as it could also be very unexpected when streaming to a frontend. I suppose that's a reason to not consider response-based fallback for streaming yet either (or you'd get 2 responses streamed back to back...), and tackle both in a separate issue!

DouweM:
@sarth6 Per my last comment, let's not complicate this PR further with streaming support. Can you please expand the issue you had created to cover fallback for streaming more broadly, pull the new streaming stuff out of this PR into a new draft one, and get this one green again so we can merge it? :)

sarth6:
Thanks @DouweM - this one is ready for another review then. Confirmed docs are rendering well and look good to me
image
image
image

Issue #3126, opened 2025-10-09
Support structured citations


DouweM:
### Description

Being implemented in ~~https://github.com/pydantic/pydantic-ai/pull/2657~~ https://github.com/pydantic/pydantic-ai/pull/3729

To do
- [ ] Providers
    - [ ] Sub issues
    - [ ] `OpenRouterModel`

aristide1997:
Would this include AWS Bedrock citations for documents (eg. pdfs)? 

https://docs.aws.amazon.com/bedrock/latest/APIReference/API_agent-runtime_Citation.html

DouweM:
@aristide1997 I intend to cover that one as well, yes!

JoaoLages:
Please make this happen 🙏 
Right now I can't use Pydantic AI whenever Google search is needed. As @vinnymeller said [here](https://github.com/pydantic/pydantic-ai/issues/2890), getting links from grounding metadata is the only way to get proper citations.

DouweM:
@JoaoLages Work is underway in https://github.com/pydantic/pydantic-ai/pull/3729 :) In the meantime, same as in https://github.com/pydantic/pydantic-ai/issues/3698, I'd be open to a PR that stores the grounding metadata on `ModelResponse/TextPart.provider_details` if enabled, while we wait for full native structured citations

vinnymeller:
FWIW for anyone using gemini, as of the `gemini-3-*` model releases, just asking it to return the citations in the output works surprisingly well. This was _terrible_ in the `gemini-2.5-*` model family. If you don't need it to work 100% perfectly (just depends on your use case obviously), it works reasonably well

E.g.

```markdown
...
<CitationInstructions>
For all sources cited in your response, please provide a `## Sources` section at the end as a numbered list,
formatted with standard markdown links like `[Source Title](source url)`

E.g.

## Sources

1. [Pydantic-AI Github](https://github.com/pydantic/pydantic-ai)
2. ...
</CitationInstructions>
```

Not sure about other providers.

If you need all the extra metadata provided then of course this won't work

DouweM:
FYI @certainly-param has been working on this in https://github.com/pydantic/pydantic-ai/issues/3885 / https://github.com/pydantic/pydantic-ai/pull/3729

PR #3787, opened 2025-12-21
Add setting to include OpenAI raw text annotations in `TextPart.provider_details`

### LINKED ISSUE #3698: Expose OpenAI text annotations on `TextPart.provider_details`
### Description

I believe that this `TODO` is blocking support for web search via Azure OpenAI as they have decided to deviate from OpenAI on [implementation](https://learn.microsoft.com/en-us/azure/ai-foundry/openai/how-to/web-search?view=foundry-classic) by burying source citations under annotations in the streaming response. Meaning that web search "works" but without annotations it's not possible to extract the citations from the response.

Image

There is an open issue for broader support of annotations here https://github.com/pydantic/pydantic-ai/issues/3126, but implementation is not trivial due to variations in different providers handling of annotations.

As an interim workaround could the raw output format be included  in the `TextPart.provider_details` object? This would support accessing the details while full annotations support is being worked out.

cc @DouweM 

### References

_No response_

corytomlinson:
Fixes #3698 

Adds OpenAI raw text annotations to `TextPart.provider_details` as an interim workaround until #3126 can be implemented.

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/openai.py
+++ b/pydantic_ai_slim/pydantic_ai/models/openai.py
@@ -2732,29 +2732,20 @@ def _make_raw_content_updater(delta: str, index: int) -> Callable[[dict[str, Any
 
 # Convert logprobs to a serializable format
 def _map_logprobs(
-    logprobs: Sequence[chat_completion_token_logprob.ChatCompletionTokenLogprob]
-    | Sequence[responses.response_output_text.Logprob]
-    | Sequence[responses.response_text_done_event.Logprob],
+    logprobs: list[chat_completion_token_logprob.ChatCompletionTokenLogprob]
+    | list[responses.response_output_text.Logprob],
 ) -> list[dict[str, Any]]:
-    mapped: list[dict[str, Any]] = []
-    for lp in logprobs:
-        top_logprobs = lp.top_logprobs or []
-        mapped.append(
-            {
-                'token': lp.token,
-                'bytes': getattr(lp, 'bytes', None),
-                'logprob': lp.logprob,
-                'top_logprobs': [
-                    {
-                        'token': tlp.token,
-                        'bytes': getattr(tlp, 'bytes', None),
-                        'logprob': tlp.logprob,
-                    }
-                    for tlp in top_logprobs
-                ],
-            }
-        )
-    return mapped
+    return [
+        {
+            'token': lp.token,
+            'bytes': lp.bytes,
+            'logprob': lp.logprob,
+            'top_logprobs': [
+                {'token': tlp.token, 'bytes': tlp.bytes, 'logprob': tlp.logprob} for tlp in lp.top_logprobs
+            ],
+        }
+        for lp in logprobs
+    ]
 
 
 def _map_usage(


pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -329,6 +329,12 @@ class OpenAIResponsesModelSettings(OpenAIChatModelSettings, total=False):
     Corresponds to the `web_search_call.action.sources` value of the `include` parameter in the Responses API.
     """
 
+    openai_include_web_search_content_annotations: bool
```
DouweM: I'd like to include the word `raw` here :)

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -1279,10 +1285,22 @@ def _process_response(  # noqa: C901
...
                     if isinstance(content, responses.ResponseOutputText):  # pragma: no branch
-                        part_provider_details: dict[str, Any] | None = None
+                        part_provider_details: dict[str, Any] = {}
                         if content.logprobs:
-                            part_provider_details = {'logprobs': _map_logprobs(content.logprobs)}
-                        items.append(TextPart(content.text, id=item.id, provider_details=part_provider_details))
+                            part_provider_details['logprobs'] = _map_logprobs(content.logprobs)
+                        # NOTE: can be removed after https://github.com/pydantic/pydantic-ai/issues/3126
+                        # TODO: discuss gate via model settings
+                        # if content.annotations and model_settings.openai_include_web_search_content_annotations:
```
DouweM: Feel free to pass model settings into this method

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -2387,8 +2405,22 @@ async def _get_event_iterator(self) -> AsyncIterator[ModelResponseStreamEvent]:
...
 
             elif isinstance(chunk, responses.ResponseOutputTextAnnotationAddedEvent):
-                # TODO(Marcelo): We should support annotations in the future.
-                pass  # there's nothing we need to do here
+                # NOTE: can be removed after https://github.com/pydantic/pydantic-ai/issues/3126
+                # TODO: discuss gate via model settings
+                # if not model_settings.openai_include_web_search_content_annotations:
+                #     continue
+
+                annotations_by_item_id = getattr(self, '_annotations_by_item_id', {})
```
DouweM: We've gotta find a better way to do this :)

Since we need to be able to update the existing `provider_details['annotations']`, look at how `handle_thinking_delta` accepts a callable that modifies the existing provider details as an alternative to a dict. Can we use that method here as well?

pydantic_ai_slim/pydantic_ai/models/openai.py
corytomlinson: done

pydantic_ai_slim/pydantic_ai/models/openai.py
corytomlinson: perfect

pydantic_ai_slim/pydantic_ai/models/openai.py
corytomlinson: will do

dsfaccini:
reopening assuming @corytomlinson 's been out for the holidays, let us know if you'd like to keep working on this !

corytomlinson:
> reopening assuming @corytomlinson 's been out for the holidays, let us know if you'd like to keep working on this !

ty

corytomlinson:
Example `TextPart` from streaming response which now contains annotations in `provider_details` when `openai_include_web_search_content_annotations_raw` property is set on `OpenAIResponsesModelSettings`...

```python
TextPart(
    content="The most prominent recent story tied to **2026** is NextEra Energy expanding its partnership with **Google Cloud** to develop new large-scale US data center campuses paired with new generation, plus plans to launch an **AI-powered field and grid-operations product by mid-2026**. ([finance.yahoo.com](https://finance.yahoo.com/news/nextera-energy-google-cloud-expand-120933899.html/?utm_source=openai))",
    id="msg_062b2851cf4f2a4d00695d2f6fe2ac8196a8a6fb1acf2c990e",
    provider_details={
        "annotations": [
            {
                "type": "url_citation",
                "end_index": 402,
                "start_index": 280,
                "title": "NextEra, Google accelerate US data center build-out with new deals",
                "url": "https://finance.yahoo.com/news/nextera-energy-google-cloud-expand-120933899.html/?utm_source=openai",
            }
        ]
    },
)
```

corytomlinson:
@DouweM ready for another look whenever you are.

tests/models/test_openai_responses.py
```diff
@@ -8753,3 +8753,82 @@ async def test_reasoning_summary_auto(allow_model_requests: None, openai_api_key
 
 I need to respond with a Python function for calculating the factorial. The user wants me to think step-by-step, but I need to keep my reasoning brief. I'll provide a brief explanation of how the function works and include some input validation. I could choose either an iterative or recursive approach. I'll keep the details high-level, showing only the essential steps before presenting the final code to the user.\
 """)
+
+
+async def test_openai_responses_text_annotations_non_streaming(allow_model_requests: None):
+    """Test that text annotations are included in provider_details when the setting is enabled."""
+    from openai.types.responses.response_output_text import AnnotationURLCitation
```
DouweM: Imports at the top please (wrapped by the check for whether the openai package is available)

tests/models/test_openai_responses.py
```diff
@@ -8753,3 +8753,82 @@ async def test_reasoning_summary_auto(allow_model_requests: None, openai_api_key
...
+
+    # Create a mock annotation
+    mock_annotation = AnnotationURLCitation(
```
DouweM: Can we test this against a real request, using vcr, instead of mocking?

tests/models/test_openai_responses.py
```diff
@@ -8753,3 +8753,82 @@ async def test_reasoning_summary_auto(allow_model_requests: None, openai_api_key
...
+    assert len(result.all_messages()) == 2
+    response_msg = result.all_messages()[1]
+    assert isinstance(response_msg, ModelResponse)
+    assert len(response_msg.parts) == 1
+    text_part = response_msg.parts[0]
+    assert isinstance(text_part, TextPart)
+    assert text_part.content == 'Hello world [1]'
+    assert text_part.provider_details is not None
+    assert 'annotations' in text_part.provider_details
+    assert text_part.provider_details['annotations'] == [mock_annotation]
```
DouweM: please use `snapshot()` for comparing message histories

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -335,6 +335,14 @@ class OpenAIResponsesModelSettings(OpenAIChatModelSettings, total=False):
     Corresponds to the `file_search_call.results` value of the `include` parameter in the Responses API.
     """
 
+    openai_include_web_search_content_annotations_raw: bool
```
DouweM: Are these only web search? Or could they be from other sources as well? I like `include_raw_annotations`

tests/models/test_openai_responses.py
DouweM: Also we want only JSON-able stuff on `provider_details`, not full annotation objects, so this should be serialized first (using e.g. `TypeAdapter[...].dump_python`)

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -2397,7 +2428,22 @@ async def _get_event_iterator(self) -> AsyncIterator[ModelResponseStreamEvent]:
...
             elif isinstance(chunk, responses.ResponseTextDoneEvent):
-                pass  # there's nothing we need to do here
+                # When text is done, add logprobs and annotations to provider_details if available
+                part = self._parts_manager.get_part_by_vendor_id(chunk.item_id)
+                if part is not None and isinstance(part, TextPart):
+                    # Add logprobs if available
+                    if chunk.logprobs:
+                        if part.provider_details is None:
+                            part.provider_details = {}
+                        part.provider_details['logprobs'] = _map_logprobs(chunk.logprobs)
```
DouweM: Hmm were we not adding these before when streaming? May need a test now that we do!

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -2553,8 +2599,11 @@ def update_provider_details(existing: dict[str, Any] | None) -> dict[str, Any]:
 
 # Convert logprobs to a serializable format
 def _map_logprobs(
-    logprobs: list[chat_completion_token_logprob.ChatCompletionTokenLogprob]
-    | list[responses.response_output_text.Logprob],
+    logprobs: Sequence[chat_completion_token_logprob.ChatCompletionTokenLogprob]
+    | Sequence[responses.response_output_text.Logprob]
+    | Sequence[
+        Any
+    ],  # Accept duck-typed Logprob variants (e.g., from response_text_done_event) that share the same structure
```
DouweM: Is this needed? If we know it's the right type, I'd rather do a cast at the call site

tests/models/test_openai_responses.py
corytomlinson: done

pydantic_ai_slim/pydantic_ai/models/openai.py
corytomlinson: done

pydantic_ai_slim/pydantic_ai/models/openai.py
corytomlinson: done

tests/models/test_openai_responses.py
corytomlinson: done

tests/models/test_openai_responses.py
corytomlinson: done

pydantic_ai_slim/pydantic_ai/models/openai.py
corytomlinson: yes this was missing. test added.

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -1320,6 +1335,11 @@ def _process_response(  # noqa: C901
                         part_provider_details: dict[str, Any] | None = None
                         if content.logprobs:
                             part_provider_details = {'logprobs': _map_logprobs(content.logprobs)}
+                        if model_settings.get('openai_include_raw_annotations') and content.annotations:
+                            part_provider_details = part_provider_details or {}
+                            part_provider_details['annotations'] = TypeAdapter(
+                                list[responses.response_output_text.Annotation]
+                            ).dump_python(list(content.annotations), warnings=False)
```
DouweM: TypeAdapters are somewhat expensive to build, so please move it out of the loop and to the module level, like `web_search_tool_result_content_ta` in `anthropic.py`.

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -2425,8 +2451,11 @@ async def _get_event_iterator(self) -> AsyncIterator[ModelResponseStreamEvent]:
                 pass  # content already accumulated via delta events
 
             elif isinstance(chunk, responses.ResponseOutputTextAnnotationAddedEvent):
-                # TODO(Marcelo): We should support annotations in the future.
-                pass  # there's nothing we need to do here
+                # Collect annotations if the setting is enabled
+                if self._model_settings.get('openai_include_raw_annotations'):
+                    _annotations_by_item.setdefault(chunk.item_id, []).append(
+                        cast(responses.response_output_text.Annotation, chunk.annotation)
```
DouweM: Why do we need to cast?

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -2435,7 +2464,24 @@ async def _get_event_iterator(self) -> AsyncIterator[ModelResponseStreamEvent]:
...
             elif isinstance(chunk, responses.ResponseTextDoneEvent):
-                pass  # there's nothing we need to do here
+                # When text is done, add logprobs and annotations to provider_details if available
+                part = self._parts_manager.get_part_by_vendor_id(chunk.item_id)
+                if part is not None and isinstance(part, TextPart):
+                    # Add logprobs if available
+                    if chunk.logprobs:
+                        part.provider_details = part.provider_details or {}
+                        part.provider_details['logprobs'] = _map_logprobs(
+                            cast(Sequence[responses.response_output_text.Logprob], chunk.logprobs)
```
DouweM: Why cast?

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -2435,7 +2464,24 @@ async def _get_event_iterator(self) -> AsyncIterator[ModelResponseStreamEvent]:
...
+                        )
+
+                    # Add annotations if the setting is enabled
+                    if self._model_settings.get('openai_include_raw_annotations'):
```
DouweM: We don't really need to check this again, as the list wouldn't have been populated unless this setting was enabled, right?

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -2435,7 +2464,24 @@ async def _get_event_iterator(self) -> AsyncIterator[ModelResponseStreamEvent]:
...
+                        annotations = _annotations_by_item.get(chunk.item_id)
+                        if annotations:
+                            part.provider_details = part.provider_details or {}
+                            part.provider_details['annotations'] = TypeAdapter(
+                                list[responses.response_output_text.Annotation]
```
DouweM: See above; please reuse a constant type adapter.

tests/models/test_openai_responses.py
```diff
@@ -567,6 +574,121 @@ async def get_capital(country: str) -> str:
...
 
+async def test_openai_responses_stream_logprobs(allow_model_requests: None):
+    annotation = AnnotationURLCitation(
+        type='url_citation',
+        start_index=0,
+        end_index=5,
+        title='Example',
+        url='https://example.com',
+    )
+    logprob = Logprob.model_construct(
```
DouweM: Could we use a real request here as well, instead of mocking?

tests/models/test_openai_responses.py
```diff
@@ -567,6 +574,121 @@ async def get_capital(country: str) -> str:
...
+
+    settings = OpenAIResponsesModelSettings(openai_include_raw_annotations=True)
+
+    async with agent.run_stream(prompt, model_settings=settings) as result:
+        async for response, is_last in result.stream_responses(debounce_by=None):
+            if is_last:
+                text_part = next(part for part in response.parts if isinstance(part, TextPart))
+                assert text_part.provider_details is not None
+                assert text_part.provider_details['annotations'] == IsInstance(list)
+                json.dumps(text_part.provider_details)
```
DouweM: This is weird :) Snapshot please!

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -2435,7 +2464,24 @@ async def _get_event_iterator(self) -> AsyncIterator[ModelResponseStreamEvent]:
                     yield event
 
             elif isinstance(chunk, responses.ResponseTextDoneEvent):
-                pass  # there's nothing we need to do here
+                # When text is done, add logprobs and annotations to provider_details if available
+                part = self._parts_manager.get_part_by_vendor_id(chunk.item_id)
```
DouweM: This will not result in a new event being omitted. Maybe we can use `handle_text_delta` and pass `provider_details` in there?

tests/models/test_openai_responses.py
```diff
@@ -8964,3 +9086,123 @@ async def test_reasoning_summary_auto(allow_model_requests: None, openai_api_key
...
+                        tool_name='web_search',
+                        content={'status': 'completed'},
+                        tool_call_id=IsStr(),
+                        timestamp=IsDatetime(),
+                        provider_name='openai',
+                    ),
+                    TextPart(
+                        content=IsStr(),
+                        id=IsStr(),
+                        provider_details={'annotations': IsInstance(list)},
```
DouweM: Please include the actual value

tests/models/test_openai_responses.py
```diff
@@ -8964,3 +9086,123 @@ async def test_reasoning_summary_auto(allow_model_requests: None, openai_api_key
...
+                run_id=IsStr(),
+            ),
+        ]
+    )
+
+    response_msg = messages[1]
+    assert isinstance(response_msg, ModelResponse)
+    text_part = next(part for part in response_msg.parts if isinstance(part, TextPart))
+    assert text_part.provider_details is not None
+    json.dumps(text_part.provider_details)
```
DouweM: Not sure what this is needed for :) We can use a snapshot, or skip this as it's already in the snapshot above.

tests/models/test_openai_responses.py
```diff
@@ -567,6 +574,121 @@ async def get_capital(country: str) -> str:
...
+                assert text_part2.provider_details is None or 'annotations' not in text_part2.provider_details
+
+    model3 = OpenAIResponsesModel('gpt-5.2', provider=OpenAIProvider(api_key=openai_api_key))
+    agent3 = Agent(model3, instructions='Answer directly.')
+    settings3 = OpenAIResponsesModelSettings(openai_include_raw_annotations=True)
+    async with agent3.run_stream('What is 2+2?', model_settings=settings3) as result3:
+        async for response, is_last in result3.stream_responses(debounce_by=None):
+            if is_last:
+                text_part3 = next(part for part in response.parts if isinstance(part, TextPart))
+                assert text_part3.provider_details is None or 'annotations' not in text_part3.provider_details
```
DouweM: When streaming, I want to verify that provider details are included on the `PartEndEvent` (or an earlier `PartDeltaEvent`). So please have a look at other tests that test against events; we can use `run_stream_events` instead of `run_stream`.

pydantic_ai_slim/pydantic_ai/models/openai.py
corytomlinson: Because `chunk.logprobs` in streaming is typed as `response_text_done_event.Logprob`, which isn’t the same type as `_map_logprobs()` expects (`response_output_text.Logprob`). The cast was a quick way to satisfy the type checker without changing `_map_logprobs()`. I can widen `_map_logprobs()` to accept the new type instead if that is preferrable?

pydantic_ai_slim/pydantic_ai/models/openai.py
corytomlinson: done.

pydantic_ai_slim/pydantic_ai/models/openai.py
corytomlinson: done.

tests/models/test_openai_responses.py
corytomlinson: done.

tests/models/test_openai_responses.py
corytomlinson: done.

pydantic_ai_slim/pydantic_ai/models/openai.py
corytomlinson: Right. Unnecessary check. Fixed.

pydantic_ai_slim/pydantic_ai/models/openai.py
corytomlinson: done.

pydantic_ai_slim/pydantic_ai/models/openai.py
corytomlinson: Perhaps it's not worth trying to add log probs to streaming events at all? My initial thought was just to close the gap to keep parity with non-streaming events. But this above change will also lead to a test coverage issue which then requires another new test. Maybe the scope creep should be avoided altogether. Let me know what you prefer.

tests/models/test_openai_responses.py
corytomlinson: done. using `run_stream_events` instead.

pydantic_ai_slim/pydantic_ai/models/openai.py
DouweM: Yeah I'd rather split the PR up, although I do think it's a worth fix :) And regarding the specific issue, in a type-safe library like ours it's much much better to update the methods to take more types, than to "assume" one type matches another via cast.

DouweM:
@corytomlinson Please have a look at the failing test

dsfaccini:

```bash
git pull

deactivate

rm -rf .venv

make install
```

pydantic_ai_slim/pydantic_ai/models/openai.py
corytomlinson: ok going to pull this out and will submit another PR to add logprobs to the streaming response.

tests/models/test_openai_responses.py
corytomlinson: fixed test

pydantic_ai_slim/pydantic_ai/models/openai.py
corytomlinson: fixed.

DouweM:
@corytomlinson Thanks Cory!

Issue #3349, opened 2025-11-06
Add multimodel custom fileextension support


zhcn000000:
### Description

Add support for custom extensions for multimodal language model FileURLs and BinaryContents or add parameters that do not perform file extension verification. For example, I want to be able to pass in extensions such as .ppt and .pptx for DocumentUrl when sending documents to the model or sending cad file as ImageUrl to model which is rendered to image in server and  The model server may have a custom multimedia decoder or document parser that allows the input of multimedia files of types other than predefined ones

DouweM:
@zhcn000000 I agree the current lists of extensions and media types are too restrictive. We could use https://docs.python.org/3.14/library/mimetypes.html to support essentially all file types, and rely on the error from the model if it's sent something it doesn't support.

Would you be up for submitting a PR, as you know exactly what types and model APIs you're dealing with and can verify the new code works for you?

DouweM:
Related to https://github.com/pydantic/pydantic-ai/issues/3776

Issue #3776, opened 2025-12-19
BinaryContent: Add support for is_text


gaby:
### Description

When using `BinaryContent.from_path` with a text file, pydantic-ai defaults to `application/octet-stream` for the mimetype. This causes the request to fail if the backend/model does not support it. Instead similar to what `is_image` does, it should check common text mime types and return true via `is_text`.

```python
from pathlib import Path
from pydantic_ai import Agent, BinaryContent

my_compose_file = Path('docker-compose.yml')
binary_content = BinaryContent.from_path(my_compose_file )

# ^mime type in binary_content is now `application/octet-stream` instead of `application/yaml`
```

Right now `pydantic-ai` defaults to `application/octet-stream` for all these file types:
- text
- json
- yaml
- xml
- toml
- or any other text based file type.

DouweM:
@gaby Thanks for the report, looks like these extensions are not recognized by default by the `mimetypes` library that `from_path` uses: https://github.com/python/cpython/blob/3.13/Lib/mimetypes.py#L456, even if we turn off strict mode: https://github.com/python/cpython/blob/ae3834c97eec709d7ceda9d1fcf6f5f09cce7223/Lib/mimetypes.py#L613-L625

So I suppose we should use `mimetypes.add_type` to register them, as we're also doing in https://github.com/pydantic/pydantic-ai/pull/3501 (cc @Viicos).

Note that there is some debate on what the best MIME type for e.g. `.yaml` is: https://stackoverflow.com/questions/332129/yaml-media-type

Would you be up for submitting a PR?

Viicos:
Yes, let's wait until https://github.com/pydantic/pydantic-ai/pull/3501 gets merged (just need one small change), and then have a PR to add the necessary extra mime type.

gaby:
> Yes, let's wait until [#3501](https://github.com/pydantic/pydantic-ai/pull/3501) gets merged (just need one small change), and then have a PR to add the necessary extra mime type.

Makes sense, I started looking at implementing this but realized that #3501 needs to be merged first.

PR #3789, opened 2025-12-21
Improve the way of selecting file extensions and support automatic selection of file extensions

zhcn000000:
Improve the way of selecting file extension or mediatype and support automatic selection of file mediatype for FileUrl or BinaryContent 
Fix [#3349](https://github.com/pydantic/pydantic-ai/issues/3349)

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/messages.py
+++ b/pydantic_ai_slim/pydantic_ai/messages.py
@@ -109,7 +109,7 @@ VideoMediaType: TypeAlias = Literal[
 
 AudioFormat: TypeAlias = Literal['wav', 'mp3', 'oga', 'flac', 'aiff', 'aac']
 ImageFormat: TypeAlias = Literal['jpeg', 'png', 'gif', 'webp', 'heic', 'heif']
-DocumentFormat: TypeAlias = Literal['doc', 'docx', 'pdf', 'txt', 'xls', 'xlsx', 'ppt', 'pptx']
+DocumentFormat: TypeAlias = Literal['doc', 'docx', 'pdf', 'xls', 'xlsx', 'ppt', 'pptx']
 VideoFormat: TypeAlias = Literal['mkv', 'mov', 'mp4', 'webm', 'flv', 'mpeg', 'mpg', 'wmv', 'three_gp']
 TextFormat: TypeAlias = Literal['csv', 'html', 'md', 'yaml', 'toml', 'json', 'xml', 'txt']
 FinishReason: TypeAlias = Literal[


DouweM:
@Viicos Can you review this please as you were already in MIME land recently?

tests/test_messages.py
```diff
@@ -668,6 +678,46 @@ def test_binary_content_validation_with_optional_identifier():
...
+def test_binary_content_magika_not_installed(monkeypatch: pytest.MonkeyPatch):
+    """Test BinaryContent._infer_media_type when Magika is not installed (lines 611-614)."""
+    binary_content = BinaryContent(data=b'some data')
+
+    # Accessing media_type should raise ImportError with a helpful message
+    with pytest.raises(ImportError, match='Magika is not installed. Please install magika'):
+        _ = binary_content.media_type
+
+
 def test_binary_content_from_path(tmp_path: Path):
```
DouweM, line 751: Please see if we can also address https://github.com/pydantic/pydantic-ai/issues/3776 at the same time.

zhcn000000:
If magika is installed, it will be used to select extensions instead of guess_type when BinaryContent.from_path called. Is it necessary to mark magika as mandatory? Currently, magika does not support Python 3.14 due to incompatibilities caused by onnxruntime and its dependencies, and magika and its dependencies are too large

Viicos:
> @Viicos Can you review this please as you were already in MIME land recently?

I'm out of capacity until next month, FYI if you want to get this merged sooner.

DouweM:
@zhcn000000 Do we need magika at all? I think we can fix https://github.com/pydantic/pydantic-ai/issues/3349 and https://github.com/pydantic/pydantic-ai/issues/3776 just by adding additional types like we do here:

https://github.com/pydantic/pydantic-ai/blob/fbe0dfe4f4b4dd14eb786e8651675316beeef71e/pydantic_ai_slim/pydantic_ai/messages.py#L39-L58

I'd strongly prefer not to add a package that will make it harder to support 3.14, or to change behavior depending on whether a package is available or not.

zhcn000000:
I have already set this package as optional. It is only used to automatically select the media type of unknown files from BinaryContent,If you expect users to always input the media type and ensure the correct file extension when uploading from a path, then it is unnecessary and can be removed, or we can also use python-magic which is lighter than magika to automatically select the type, but this may reduce the accuracy of document files and text files

gaby:
> If magika is installed, it will be used to select extensions instead of guess_type when BinaryContent.from_path called. Is it necessary to mark magika as mandatory? Currently, magika does not support Python 3.14 due to incompatibilities caused by onnxruntime and its dependencies, and magika and its dependencies are too large

- Related: https://github.com/google/magika/issues/1258
- Related: https://github.com/microsoft/onnxruntime/issues/26309

gaby:
Why are stale PR's closed so fast? Should be at least a month.

dsfaccini:
sorry for that @gaby , I've reopened and I think we'll be extending it soon, we're generally more on top of this but I didn't notice this one being labeled as stale during the holidays

dsfaccini:
to add my 2 cents: I'm not thrilled about switching to magika, would much prefer to keep using mimetypes, any issues that arise from mimetypes are easily fixable via different channels.

gaby:
> to add my 2 cents: I'm not thrilled about switching to magika, would much prefer to keep using mimetypes, any issues that arise from mimetypes are easily fixable via different channels.

Agree, not at least until they fix the dependencies issue and onnx adds 3.14 support. 

I'm using magika with my projects since I kept running into issues matching mimetypes with pydantic-ai.

dsfaccini:
@gaby appreciate your input, would love to see what you've been using to fill in those gaps for your use cases, if you'd like to share a gist or something like that

gaby:
> @gaby appreciate your input, would love to see what you've been using to fill in those gaps for your use cases, if you'd like to share a gist or something like that

@dsfaccini I call magika then use the identified mimetype to call pydantic-ai.

```python
from magika import Magika
m = Magika()
example_data = b'function log(msg) {console.log(msg);}'
res = m.identify_bytes(example_data)

agent = Agent(model='gateway/openai:gpt-5')
result = agent.run_sync(
    [
        'Summarize the content',
        BinaryContent(data=example_data, media_type=res.output.mime_type),
    ]
)
```

This works for most types since Magika is pretty good at detecting mimetypes.

pydantic_ai_slim/pyproject.toml
```diff
@@ -125,6 +125,8 @@ dbos = ["dbos>=1.14.0"]
 # Prefect
 prefect = ["prefect>=3.4.21"]
 
+magika = ["magika>=1.0.1"]
```
DouweM: Please remove magika from this PR, I think we should be able to accomplish everything we want with `mimetypes`, and I don't want different behavior when a package is present vs not.

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -98,6 +101,9 @@
 ]
 """Reason the model finished generating the response, normalized to OpenTelemetry values."""
 
+# Shared instances for media type detection to avoid repeated initialization overhead
```
DouweM: odd place for this comment?

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -192,6 +198,36 @@ def __init__(
         self.force_download = force_download
         self.vendor_metadata = vendor_metadata
 
+    @staticmethod
+    def create(
```
DouweM: Is this necessary for what you're trying to address in this PR?

Instead of `create`, could it `from_url` similar to the `from` methods on `BinaryContent`?

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -560,6 +575,35 @@ def identifier(self) -> str:
...
+
+        Automatically detects the media type using Magika if not provided.
+        """
+        return self._media_type or self._infer_media_type()
+
+    def _infer_media_type(self) -> str:
+        """Infer the media type of the binary content."""
+        global _magika_instance
+
+        # Try Magika first (more accurate, especially for documents)
```
DouweM: Is there a specific common file type you have in mind that mimetypes can't detect?

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -560,6 +575,35 @@ def identifier(self) -> str:
...
+            if _magika_instance is None:
+                _magika_instance = Magika()
+            result = _magika_instance.identify_bytes(self.data).output
+            self._media_type = result.mime_type
+            self._type = result.group
+            self._extension = result.extensions[0] if result.extensions else None
+            return self._media_type
+        except ImportError as e:
+            raise ImportError(
+                'Magika is not installed. Please install magika to enable media type detection for BinaryContent.'
```
DouweM: This would very much be a breaking change

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -537,6 +550,8 @@ def from_path(cls, path: PathLike[str]) -> BinaryContent:
         path = Path(path)
         if not path.exists():
             raise FileNotFoundError(f'File not found: {path}')
+        if find_spec('magika'):
+            return cls.narrow_type(cls(data=path.read_bytes()))
```
DouweM: This is confusing

pydantic_ai_slim/pydantic_ai/messages.py
zhcn000000: If not use magica we should use mimetypes to infer mediatype from filename instead of from binary data

zhcn000000:
Magika has been removed, and mimetypes can only detect media_type using filenames. If a more lightweight solution is needed that can identify file types from binary data, consider using python-magic

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -595,22 +617,32 @@ def is_video(self) -> bool:
...
+
+        Returns the file extension (without leading dot) based on the media type.
+        Uses cached extension from Magika if available, otherwise maps from media_type.
+
+        Raises:
+            ValueError: If file format cannot be inferred from media type.
+        """
+        ext = _all_format_lookup.get(self.media_type) or _mime_types.guess_extension(self.media_type)
+
+        # Fallback to mimetypes.guess_extension
```
DouweM: Unnecssary comment

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -595,22 +617,32 @@ def is_video(self) -> bool:
...
     def is_document(self) -> bool:
         """Return `True` if the media type is a document type."""
-        return self.media_type in _document_format_lookup
+        return self.media_type in _document_format_lookup or self.is_text
+        # text files send as documents until we have a separate TextContent type
+
+    @property
+    def is_text(self) -> bool:
+        """Return `True` if the media type is a text type."""
+        # TODO: consider to read text content and send as text in user_prompt to the model?
```
DouweM: Remove comment please

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -595,22 +617,32 @@ def is_video(self) -> bool:
     @property
     def is_document(self) -> bool:
         """Return `True` if the media type is a document type."""
-        return self.media_type in _document_format_lookup
+        return self.media_type in _document_format_lookup or self.is_text
+        # text files send as documents until we have a separate TextContent type
```
DouweM: unnecessary comment

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -565,7 +564,30 @@ def identifier(self) -> str:
         It's also included in inline-text delimiters for providers that require inlining text documents, so the model can
         distinguish multiple files.
         """
-        return self._identifier or _multi_modal_content_identifier(self.data)
+        return self._identifier or (
+            _multi_modal_content_identifier(self.data) + ('-' + self.file_name if self.file_name else '')
```
DouweM: Unrelated change? I don't think we should do this.

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -472,10 +468,10 @@ class BinaryContent:
     Use `.base64` to get the base64-encoded string.
     """
 
-    _: KW_ONLY
+    file_name: str | None = None
```
DouweM: Please require this to be a kwarg, i.e. move it down

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -489,22 +485,28 @@ class BinaryContent:
         compare=False, default=None
     )
 
+    _media_type: Annotated[str | None, pydantic.Field(alias='media_type', default=None, exclude=True)] = field(
+        compare=False, default=None
```
DouweM: This caused a lot of examples and snapshots to be updated with the extra underscored key. I never liked how `FileUrl.media_type` worked, which also has that issue, so I'd prefer to not introduce it here.

Since `media_type` has always be a required field, I don't think we need to do the optional field + `computed_field` dance here. Could we leave the `media_type` field as it was, make it optional on `__init__` (or rather: require one of `media_type` and `file_name` to be provided by using overlaads), and then set it based on `file_name` / `self._infer_media_type()`?

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -199,6 +199,36 @@ def __init__(
...
+        vendor_metadata: dict[str, Any] | None = None,
+    ) -> FileUrl:
+        mime_type = media_type or _mime_types.guess_type(url)[0]
+        kwargs: dict[str, Any] = {
+            'url': url,
+            'media_type': media_type,
+            'identifier': identifier,
+            'force_download': force_download,
+            'vendor_metadata': vendor_metadata,
+        }
```
DouweM: This isn't type safe, so I'd rather not go the kwargs route. We can add a `url_cls` var, and then construct it as `url_cls(url=..., etc)`

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -293,12 +324,12 @@ def is_youtube(self) -> bool:
         return hostname in ('youtu.be', 'youtube.com', 'www.youtube.com')
 
     @property
-    def format(self) -> VideoFormat:
+    def format(self) -> str:
```
DouweM: For consistency with the others, `VideoFormat | str`

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/messages.py
zhcn000000: Provide original filename to model if available  may be useful

DouweM:
@zhcn000000 Please have a look at coverage

zhcn000000:
Added

dsfaccini:

```bash
git pull

deactivate

rm -rf .venv

make install
```

DouweM:
@zhcn000000 Please have a look at the conflicts

DouweM:
@zhcn000000 Please have a look at the conflicts

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -518,20 +525,15 @@ def from_data_uri(cls, data_uri: str) -> BinaryContent:
...
         """
         path = Path(path)
         if not path.exists():
             raise FileNotFoundError(f'File not found: {path}')
-        media_type, _ = _mime_types.guess_type(path)
-        if media_type is None:
-            media_type = 'application/octet-stream'
 
-        return cls.narrow_type(cls(data=path.read_bytes(), media_type=media_type))
+        return cls.narrow_type(cls(data=path.read_bytes(), file_name=path.name))
```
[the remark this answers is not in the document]
DouweM: Yeah I'd rather keep the `application/octet-stream` fallback

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -614,12 +640,15 @@ def __init__(
         self,
         data: bytes,
         *,
-        media_type: ImageMediaType | str,
+        media_type: ImageMediaType | str | None = None,
         identifier: str | None = None,
         vendor_metadata: dict[str, Any] | None = None,
+        file_name: str | None = None,
+        # Required for inline-snapshot which expects all dataclass `__init__` methods to take all field names as kwargs.
```
DouweM: Duplicated comment

pydantic_ai_slim/pydantic_ai/messages.py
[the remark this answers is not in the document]
DouweM: And it makes sense to have a `media_type` arg here as we have on `FileUrl.from_url`

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -274,28 +311,23 @@ def _infer_media_type(self) -> str:
...
     def is_youtube(self) -> bool:
         """True if the URL has a YouTube domain."""
         parsed = urlparse(self.url)
         hostname = parsed.hostname
-        return hostname in ('youtu.be', 'youtube.com', 'www.youtube.com')
+        return hostname in {'youtu.be', 'youtube.com', 'www.youtube.com'}
 
     @property
-    def format(self) -> VideoFormat:
+    def format(self) -> VideoFormat | str:
```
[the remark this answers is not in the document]
DouweM: This is a real concern, but I'm OK with it

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -484,12 +488,14 @@ def __init__(
         self,
         data: bytes,
         *,
-        media_type: AudioMediaType | ImageMediaType | DocumentMediaType | str,
+        media_type: str | None = None,
```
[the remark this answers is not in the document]
DouweM: We can use `@overloads` to require either `media_type` or `file_name` or both, but not have both be optional

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -202,6 +202,37 @@ def __init__(
...
+            url_cls = DocumentUrl
+        else:
+            raise ValueError(f'Could not classify file from URL: {url}.')
+        return url_cls(
+            url=url,
+            media_type=media_type,
+            identifier=identifier,
+            force_download=force_download,
+            vendor_metadata=vendor_metadata,
+        )
```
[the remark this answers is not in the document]
DouweM: I like the new method, bit it should indeed handle text (including e.g. `application/yaml`) as documents

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -593,22 +611,28 @@ def is_video(self) -> bool:
     @property
     def is_document(self) -> bool:
         """Return `True` if the media type is a document type."""
-        return self.media_type in _document_format_lookup
+        return self.media_type in _document_format_lookup or self.is_text
```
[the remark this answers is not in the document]
zhcn000000: This issue cannot be resolved unless TextUrl and processing text are added to TextUrl and binarycontent, and the text is processed differently from DocumentUrl or text  not processed  at all

PR #3252, opened 2025-10-24
Support embeddings models

dmontagu:
Started this in collaboration with @DouweM, I'd like to ensure consensus on the API design before adding the remaining-providers/logfire-instrumentation/docs/tests.

This is inspired by the approach in haiku.rag, though we adapted it to be a bit closer to the `Agent` APIs are used (and how you can override model, settings, etc.).

Closes https://github.com/pydantic/pydantic-ai/issues/58

Example:

```py
import asyncio

from pydantic_ai.embeddings import Embedder

embedder = Embedder("openai:text-embedding-3-large")


async def main():
    result = await embedder.embed_documents(["hello", "world"])
    print(result)
    # (IsList, snapshot, and IsDatetime are testing helpers, but you get the point)
    # EmbeddingResult(
    #     embeddings=IsList(
    #         IsList(
    #             snapshot(0.01681816205382347),
    #             snapshot(-0.05579638481140137),
    #             snapshot(0.005661087576299906),
    #             length=1536,
    #         ),
    #         IsList(
    #             snapshot(-0.010592407546937466),
    #             snapshot(-0.03599696233868599),
    #             snapshot(0.030227113515138626),
    #             length=1536,
    #         ),
    #         length=2,
    #     ),
    #     inputs=['hello', 'world'],
    #     input_type='document',
    #     usage=RequestUsage(input_tokens=2),
    #     model_name='text-embedding-3-small',
    #     timestamp=IsDatetime(),
    #     provider_name='openai',
    # )



if __name__ == "__main__":
    asyncio.run(main())
```

To do:
- [x] `Embedder.embed_sync`
- [x] Distinguish between encoding query and document(s) https://github.com/pydantic/pydantic-ai/pull/3463
- [x] Help with token counting (https://github.com/pydantic/pydantic-ai/pull/3252#issuecomment-3460220162)
	- [x] Add `count_tokens`
    - [x] Add `max_input_tokens`
- [x] Expose token usage and cost
- [x] Test known model names completeness
- [x] Instrumentation
	- [x] Cost tracking
	- [x] Verify Claude-generated attrs and metrics
	- [x] Update `logfire.instrument_pydantic_ai()`: https://github.com/pydantic/logfire/pull/1575
- [x] Error handling (`ModelAPIError`)
- [x] Test coverage
- [x] Add prices for Cohere: https://github.com/pydantic/genai-prices/pull/240
[...]

PR #3799, opened 2025-12-22
Add SurrealDB examples

martinschaer:
This PR:
- Adds examples rag_surrealdb.py, clone of rag.py, but using SurrealDB embedded
- Adds a new RAG section in the docs

### DIFF
--- a/examples/pydantic_ai_examples/rag_surrealdb.py
+++ b/examples/pydantic_ai_examples/rag_surrealdb.py
@@ -304,7 +304,7 @@ if __name__ == '__main__':
         uvicorn.run(app, host='127.0.0.1', port=7932)
     else:
         print(
-            'uv run --extra examples -m pydantic_ai_examples.rag_surrealdb build|search',
+            'uv run --extra examples -m pydantic_ai_examples.rag_surrealdb build|search|web',
             file=sys.stderr,
         )
         sys.exit(1)


DouweM:
@martinschaer Thanks for working on this! 

We're about to release https://github.com/pydantic/pydantic-ai/pull/3252, which would be great to use in this example, as well as https://ai.pydantic.dev/web/ for the web UI. That'll make this a bit more differentiated/modern compared to the existing example it's based on.

Note that we'll also need a doc like https://ai.pydantic.dev/examples/rag/, so that people can find it in the sidebar. I suggesting creating a new dedicated RAG category (separate from the existing "Data & Analytics"), so that it's more natural to have separate examples for different vector DBs. You can then move the existing Postgres/pgvector example there, and mention "SurrealDB" in the title of your new page.

Can you please make those changes (once that feature is released)?

Kludex:
Why do you folks use `cerberus` and not `pydantic`?

martinschaer:
> Why do you folks use `cerberus` and not `pydantic`?

@Kludex good question. I think we should consider pydantic-core

martinschaer:
@Kludex I just updated this PR with the latest surrealdb sdk which has pydantic-core instead of cerberus

examples/pydantic_ai_examples/rag_surrealdb.py
```diff
@@ -0,0 +1,298 @@
...
+        content: str
+        dist: float
+
+    result_ta = TypeAdapter(list[RetrievalQueryResult])
+
+    with logfire.span(
+        'create embedding for {search_query=}', search_query=search_query
+    ):
+        result = await embedder.embed_query(search_query)
+        embedding = result.embeddings
```
Kludex: There's no inner span on `embed_query` itself? @DouweM

mkdocs.yml
```diff
@@ -116,7 +116,9 @@ nav:
       - Data & Analytics:
           - examples/sql-gen.md
           - examples/data-analyst.md
+      - RAG:
           - examples/rag.md
```
DouweM: Can you please rename this file to `rag-pgvector`, update the page title, and set up a redirect (in docs-site/src/index.ts)?

examples/pydantic_ai_examples/chat_app_surreal.py
```diff
@@ -0,0 +1,243 @@
+"""Simple chat app example build with FastAPI using SurrealDB embedded.
```
DouweM: I don't think we need this anymore? At least it's not mention in the doc, and we have the native web app feature now

examples/pydantic_ai_examples/rag_surrealdb.py
```diff
@@ -0,0 +1,298 @@
...
+    """
+
+    @dataclass
+    class RetrievalQueryResult:
+        url: str
+        title: str
+        content: str
+        dist: float
+
+    result_ta = TypeAdapter(list[RetrievalQueryResult])
```
DouweM: Let's move the type and type adapter out of the tool

examples/pydantic_ai_examples/rag_surrealdb.py
```diff
@@ -0,0 +1,298 @@
...
+    with logfire.span(
+        'create embedding for {search_query=}', search_query=search_query
+    ):
+        result = await embedder.embed_query(search_query)
+        embedding = result.embeddings
+
+    assert len(embedding) == 1, (
+        f'Expected 1 embedding, got {len(embedding)}, doc query: {search_query!r}'
+    )
+    embedding_vector = list(embedding[0])
```
DouweM: Why do we need the `list` here?

examples/pydantic_ai_examples/rag_surrealdb.py
```diff
@@ -0,0 +1,298 @@
...
+    # Process SurrealDB query result
+    try:
+        rows = result_ta.validate_python(result)
+        logfire.info('Retrieved {len} results', len=len(rows))
+    except Exception as e:
+        logfire.error('Failed to validate JSON response: {error}', error=e)
+        raise
+
+    return '\n\n'.join(
+        f'# {row.title}\nDocumentation URL:{row.url}\n\n{row.content}\n' for row in rows
```
DouweM: We don't need the final `\n` if we're already joining these by 2 newlines!

examples/pydantic_ai_examples/rag_surrealdb.py
```diff
@@ -0,0 +1,298 @@
...
+
+    async with database_connect(True) as db:
+        with logfire.span('create schema'):
+            await db.query(DB_SCHEMA)
+
+        embedding_sem = asyncio.Semaphore(10)
+        db_sem = asyncio.Semaphore(1)
+        async with create_task_group() as tg:
+            for section in sections:
+                tg.start_soon(insert_doc_section, embedding_sem, db_sem, db, section)
```
DouweM: Instead of doing this separately for each document, can we use the fact that many apis support batch, and do just a single embedder.embed_documents call?

DouweM: I realize most of this code was taken from the existing rag.py, but I'd appreciate a bit more work into cleaning it up to match best practices as people will be using this as a starting point to use both surrealdb and pydantic ai embeddings!

examples/pydantic_ai_examples/rag_surrealdb.py
```diff
@@ -0,0 +1,298 @@
...
+            ORDER BY dist ASC
+            """,
+            {'vector': cast(Value, embedding_vector)},
+        )
+
+    # Process SurrealDB query result
+    try:
+        rows = result_ta.validate_python(result)
+        logfire.info('Retrieved {len} results', len=len(rows))
+    except Exception as e:
```
DouweM: `ValidationError`? `except Exception` is not best practice

examples/pydantic_ai_examples/rag_surrealdb.py
```diff
@@ -0,0 +1,298 @@
...
+    Value,
+)
+from typing_extensions import AsyncGenerator
+
+from pydantic_ai import Agent, Embedder
+
+# 'if-token-present' means nothing will be sent (and the example will work) if you don't have logfire configured
+logfire.configure(send_to_logfire='if-token-present')
+logfire.instrument_pydantic_ai()
+logfire.instrument_openai()
```
DouweM: unnecessary since pydantic ai is instrumented right?

examples/pydantic_ai_examples/rag_surrealdb.py
```diff
@@ -0,0 +1,298 @@
...
+    result_ta = TypeAdapter(list[RetrievalQueryResult])
+
+    with logfire.span(
+        'create embedding for {search_query=}', search_query=search_query
+    ):
+        result = await embedder.embed_query(search_query)
+        embedding = result.embeddings
+
+    assert len(embedding) == 1, (
+        f'Expected 1 embedding, got {len(embedding)}, doc query: {search_query!r}'
```
DouweM: This makes it seem like this is something the user needs to watch out for themselves, but the embedder methods are guaranteed to return the same number of embeddings as inputs

examples/pydantic_ai_examples/rag_surrealdb.py
```diff
@@ -0,0 +1,298 @@
...
+            FROM doc_sections
+            WHERE embedding <|8, 40|> $vector
+            ORDER BY dist ASC
+            """,
+            {'vector': cast(Value, embedding_vector)},
+        )
+
+    # Process SurrealDB query result
+    try:
+        rows = result_ta.validate_python(result)
```
DouweM: can we extract the surreal db query + result validation into a separate method with clean typing?

examples/pydantic_ai_examples/rag_surrealdb.py
```diff
@@ -0,0 +1,298 @@
...
+                tg.start_soon(insert_doc_section, embedding_sem, db_sem, db, section)
+
+
+async def insert_doc_section(
+    embedding_sem: asyncio.Semaphore,
+    db_sem: asyncio.Semaphore,
+    db: AsyncWsSurrealConnection | AsyncHttpSurrealConnection,
+    section: DocsSection,
+) -> None:
+    async with embedding_sem:
```
DouweM: why do we need this?

examples/pydantic_ai_examples/rag_surrealdb.py
```diff
@@ -0,0 +1,298 @@
...
+
+sections_ta = TypeAdapter(list[DocsSection])
+
+
+@asynccontextmanager
+async def database_connect(create_db: bool = False) -> AsyncGenerator[Any, None]:
+    namespace = 'pydantic_ai_examples'
+    database = 'rag_surrealdb'
+    username = 'root'
+    password = 'root'
```
DouweM: should these be constants at the top level?

examples/pydantic_ai_examples/rag_surrealdb.py
```diff
@@ -0,0 +1,298 @@
...
+    action = sys.argv[1] if len(sys.argv) > 1 else None
+    if action == 'build':
+        asyncio.run(build_search_db())
+    elif action == 'search':
+        if len(sys.argv) == 3:
+            q = sys.argv[2]
+        else:
+            q = 'How do I configure logfire to work with FastAPI?'
+        asyncio.run(run_agent(q))
+    else:
```
DouweM, line 305: Should we support a `web` subcommand?

examples/pydantic_ai_examples/rag_surrealdb.py
martinschaer: fixed in e35eab3a

examples/pydantic_ai_examples/chat_app_surreal.py
martinschaer: removed in e35eab3a

examples/pydantic_ai_examples/rag_surrealdb.py
martinschaer: because the `query` vars don't like `Sequence[float]`. Doing `cast(Value, embedding_vector)` is enough. Changed in e35eab3a

examples/pydantic_ai_examples/rag_surrealdb.py
martinschaer: Fixed in e35eab3a

examples/pydantic_ai_examples/rag_surrealdb.py
martinschaer: Fixed in e35eab3a

examples/pydantic_ai_examples/rag_surrealdb.py
martinschaer: Removed in e35eab3a

examples/pydantic_ai_examples/rag_surrealdb.py
martinschaer: Fixed in e35eab3a

mkdocs.yml
martinschaer: Changed in 14575c6a

examples/pydantic_ai_examples/rag_surrealdb.py
martinschaer: Something like this? Is there more efficient way?

```python
RecordType = TypeVar('RecordType')


class RetrievalQueryResult(BaseModel):
    url: str
    title: str
    content: str
    dist: float


async def query(
    conn: SurrealConn, query_: str, vars: dict[str, Value], record_type: RecordType
) -> list[RecordType]:
    result = await conn.query(query_, vars)
    result_ta = TypeAdapter(list[record_type])
    rows = result_ta.validate_python(result)
    return rows
```

and then

```python
            rows = await query(
                db,
                """
                SELECT url, title, content, vector::distance::knn() AS dist
                FROM doc_sections
                WHERE embedding <|8, 40|> $vector
                ORDER BY dist ASC
                """,
                {'vector': cast(Value, embedding_vector)},
                RetrievalQueryResult,
            )

```

martinschaer: Changed in 7d818f3d

examples/pydantic_ai_examples/rag_surrealdb.py
martinschaer: removed in 7d818f3d

examples/pydantic_ai_examples/rag_surrealdb.py
martinschaer: Added batch embedding generation in 033f7940 and bde1535e

examples/pydantic_ai_examples/rag_surrealdb.py
martinschaer: Removed, and refactored to do batch embedding

examples/pydantic_ai_examples/rag_surrealdb.py
martinschaer: Added in 9ea78d9a

dsfaccini:

```bash
git pull

deactivate

rm -rf .venv

make install
```

examples/pydantic_ai_examples/rag_surrealdb.py
```diff
@@ -0,0 +1,322 @@
...
+async def build_search_db():
+    """Build the search database."""
+    async with httpx.AsyncClient() as client:
+        response = await client.get(DOCS_JSON)
+        response.raise_for_status()
+    sections = sections_ta.validate_json(response.content)
+
+    async with database_connect(True) as db:
+        with logfire.span('create schema'):
+            await db.query(DB_SCHEMA)
```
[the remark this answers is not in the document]
martinschaer: fixed in 42c3a680

examples/pydantic_ai_examples/rag_surrealdb.py
```diff
@@ -0,0 +1,322 @@
...
+        if len(sys.argv) == 3:
+            q = sys.argv[2]
+        else:
+            q = 'How do I configure logfire to work with FastAPI?'
+        asyncio.run(run_agent(q))
+    elif action == 'web':
+        uvicorn.run(app, host='127.0.0.1', port=7932)
+    else:
+        print(
+            'uv run --extra examples -m pydantic_ai_exampls.rag_surrealdb build|search',
```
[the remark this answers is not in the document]
martinschaer: fixed in 42c3a680

examples/pydantic_ai_examples/rag_surrealdb.py
```diff
@@ -0,0 +1,322 @@
...
+    dist: float
+
+
+async def query(
+    conn: SurrealConn, query_: str, vars: dict[str, Value], record_type: RecordType
+) -> list[RecordType]:
+    result = await conn.query(query_, vars)
+    result_ta = TypeAdapter(list[record_type])
+    rows = result_ta.validate_python(result)
+    return rows
```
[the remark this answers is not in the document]
martinschaer: `SurrealConn` is a wrapper, so in this case `conn.query` returns a `list[dict]`, not a `list[list[dict]]` which is what the SDK returns

examples/pydantic_ai_examples/rag_surrealdb.py
```diff
@@ -0,0 +1,319 @@
...
+        try:
+            rows = await query(
+                db,
+                """
+                SELECT url, title, content, vector::distance::knn() AS dist
+                FROM doc_sections
+                WHERE embedding <|8, 40|> $vector
+                ORDER BY dist ASC
+                """,
+                {'vector': cast(Value, embedding_vector)},
```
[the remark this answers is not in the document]
martinschaer: fixed in f465f20d

examples/pydantic_ai_examples/rag_surrealdb.py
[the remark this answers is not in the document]
martinschaer: Updated PR description

PR #3800, opened 2025-12-22
Use in-memory SQLite for DBOS tests

qianl15:
For discussion with @Kludex

### DIFF
--- a/tests/test_dbos.py
+++ b/tests/test_dbos.py
@@ -128,7 +128,6 @@ in_memory_engine = create_engine(
 
 DBOS_CONFIG: DBOSConfig = {
     'name': 'pydantic_dbos_tests',
-    'system_database_url': 'sqlite:///:memory:?cache=shared',
     'run_admin_server': False,
     # enable_otlp requires dbos>1.14
     'enable_otlp': True,


qianl15:
Looks like using in-memory SQLite caused some (flaky) failures in lower Python versions. I was following the instructions here: https://docs.sqlalchemy.org/en/20/dialects/sqlite.html#using-a-memory-database-in-multiple-threads

> Note that using a :memory: database in multiple threads requires a recent version of SQLite.

So it might be the case where Python 3.10 and 3.11 didn't have good support for this.

**Update**: I'm not sure if the error was caused by in-memory SQLite or something else, because attempt 1 also had errors not related to DBOS.

tests/test_dbos.py
```diff
@@ -120,13 +122,17 @@ def workflow_raises(exc_type: type[Exception], exc_message: str) -> Iterator[Non
...
 
-DBOS_SQLITE_FILE = 'dbostest.sqlite'
+in_memory_engine = create_engine(
+    'sqlite:///:memory:?cache=shared', connect_args={'check_same_thread': False}, poolclass=StaticPool
+)
+
 DBOS_CONFIG: DBOSConfig = {
     'name': 'pydantic_dbos_tests',
-    'system_database_url': f'sqlite:///{DBOS_SQLITE_FILE}',
+    'system_database_url': 'sqlite:///:memory:?cache=shared',
```
Kludex: do you need to provide the `system_database_url` if the engine is provided?

qianl15: Good catch. Removed the unnecessary `system_database_url`.

Kludex (approved):

dsfaccini:
hey guys I think you broke CI 
- https://github.com/pydantic/pydantic-ai/actions/runs/20736280768/job/59534102308?pr=3206
- https://github.com/pydantic/pydantic-ai/actions/runs/20733788554/job/59527049075?pr=3835
- https://github.com/pydantic/pydantic-ai/actions/runs/20736547063/job/59534859530?pr=3910

PR #3803, opened 2025-12-22
Silently ignore AG-UI `ActivityMessage` instead of raising error

### LINKED ISSUE #3802: AG-UI adapter should not raise on ActivityMessage present in 'RunAgentInput'
### Initial Checks

### Description

PR #3715 added [new code](https://github.com/pydantic/pydantic-ai/blob/c9f5410d314bcda3c00f7696dbea79fd64c4a6a0/pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py#L231-L232) which introduces spurious errors when the `messages` field of the initiating request's `RunAgentInput` holds an `ActivityMessage`.

While the intent of the change was to avoid silently dropping such a message, raising an exception does not seem appropriate:  such messages are valid within the AG-UI spec, so even if the adapter cannot do anything in particular with the message, it should just ignore it.

The usecase I have is that a graph or other functional agent might emit `ActivitySnapshot` or `ActivityDelta` events in a prior run:  the `ActivityMessage` present in a subsequent run could be used to resume activity.

### Example Code

```Python

```

```Text
Python 3.13, `pydantic-ai-slim 1.37.0` (LLM not relevant to this issue).
```

tseaver:
Closes #3802.

/cc @DouweM, @bojan2501

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
+++ b/pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
@@ -228,7 +228,7 @@ class AGUIAdapter(UIAdapter[RunAgentInput, Message, BaseEvent, AgentDepsT, Outpu
                             )
                         )
 
-                case ActivityMessage():  # pragma: no cover
+                case ActivityMessage():
                     pass
 
         return builder.messages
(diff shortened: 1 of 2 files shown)


DouweM:
@tseaver Thanks Tres, please have a look at the failing CI steps

tseaver:
@DouweM fixed.

PR #3806, opened 2025-12-22
fix(google): `VideoUrl.vendor_metadata` for GCS URIs on google-vertex

### LINKED ISSUE #3805: Google: VideoUrl.vendor_metadata ignored for GCS URIs (only passed for YouTube)
### Initial Checks

### Description

**Observed behavior:** `VideoUrl.vendor_metadata` is only passed to Gemini as `video_metadata` when the URL is a YouTube URL. For GCS URIs (`gs://...`), the metadata is silently ignored.

**Expected behavior:** `vendor_metadata` should be passed for any `VideoUrl`, not just YouTube. Gemini fully supports `video_metadata` for GCS files—I verified this works with the native `google-genai` SDK:

- `start_offset="300s"` → Response: "segment begins at the 05:01 mark" ✓
- `start_offset="1800s"` → Response: "segment begins at the 30:30 mark" ✓

**Root cause:** In `pydantic_ai/models/google.py`, line 689:

```python
elif isinstance(item, VideoUrl) and item.is_youtube:
```

The `and item.is_youtube` check means non-YouTube `VideoUrl` objects fall through to the `FileUrl` branch (since `VideoUrl` inherits from `FileUrl`), which doesn't pass `vendor_metadata`.

**Suggested fix:** Remove the `is_youtube` restriction - let any argument provided by the caller get through. If you don't like that, you can add a cloud storage check similar to the youtube check:

```python
elif isinstance(item, VideoUrl):
    file_data_dict: FileDataDict = {'file_uri': item.url, 'mime_type': item.media_type}
    part_dict: PartDict = {'file_data': file_data_dict}
    if item.vendor_metadata:
        part_dict['video_metadata'] = cast(VideoMetadataDict, item.vendor_metadata)
    content.append(part_dict)
```

### Example Code

```Python
from pydantic_ai import Agent
from pydantic_ai.messages import VideoUrl

agent = Agent("google-vertex:gemini-3-flash-preview", output_type=str)

# YouTube - vendor_metadata IS passed ✓
youtube = VideoUrl(
    url="https://www.youtube.com/watch?v=VIDEO_ID",
    media_type="video/mp4",
    vendor_metadata={"start_offset": "300s", "end_offset": "330s"},
)
result = await agent.run([youtube, "Describe this segment"])
# Works - returns content from 5:00-5:30

# GCS - vendor_metadata is IGNORED ✗
gcs = VideoUrl(
    url="gs://my-bucket/video.mp4",
    media_type="video/mp4",
    vendor_metadata={"start_offset": "300s", "end_offset": "330s"},
)
result = await agent.run([gcs, "Describe this segment"])
# Broken - returns content from 0:00, ignoring the offset
```

```Text
Python: 3.12
pydantic-ai: latest (checked 2025-12-22)
google-genai: latest
Provider: google-vertex (Vertex AI)
```

barapa:
## Summary

`VideoUrl.vendor_metadata` (for Gemini's `video_metadata` with `start_offset`/`end_offset`) was only passed for YouTube URLs. GCS URIs fell through to `FileUrl` handling which doesn't support video offsets.

This adds GCS URI support on **google-vertex** (Vertex AI can access GCS buckets directly).

## The Problem

GCS URIs (`gs://bucket/video.mp4`) require google-vertex because:
- **google-vertex** (Vertex AI) runs in Google Cloud and can access GCS buckets directly
- **google-gla** (Gemini API) is a public API that cannot access your private GCS buckets

Previously, GCS URIs on google-gla would produce a request that failed with a cryptic API error. Now it fails early with a clear `UserError`.

## Changes

```python
elif isinstance(item, VideoUrl) and (
    item.is_youtube or (item.url.startswith('gs://') and self.system == 'google-vertex')
):
```

### URL Access by Provider

| URL Type | google-gla | google-vertex |
|----------|------------|---------------|
| YouTube | ✓ Google fetches it | ✓ Google fetches it |
| GCS (`gs://`) | ✗ Cannot access | ✓ Same GCP ecosystem |
| HTTP | ✓ Downloads first | ✓ Direct access |

### What Happens

| URL Type | google-gla | google-vertex |
|----------|------------|---------------|
| YouTube | `file_uri` | `file_uri` |
| GCS (`gs://`) | `UserError` (clear failure) | `file_uri` |
| HTTP | `inline_data` (downloaded) | `file_uri` |

**Note:** `vendor_metadata` → `video_metadata` is passed through whenever present. The tables above show URL access patterns, which is orthogonal to metadata handling.

## Tests

| Test | Verifies |
|------|----------|
| `test_gcs_video_url_with_vendor_metadata_on_google_vertex` | GCS + video_metadata works on Vertex AI |
| `test_gcs_video_url_raises_error_on_google_gla` | GCS fails clearly on Gemini API |
| `test_http_video_url_downloads_on_google_gla` | HTTP fallback preserved |
| `test_http_video_url_uses_file_uri_on_google_vertex` | HTTP behavior preserved |

## Related Issue

Fixes #3805

### DIFF
--- a/tests/models/test_google.py
+++ b/tests/models/test_google.py
@@ -4706,7 +4706,7 @@ async def test_http_video_url_downloads_on_google_gla(mocker: MockerFixture):
     assert 'inline_data' in content[0]
     assert 'file_data' not in content[0]
     # video_metadata is preserved even when video is downloaded
-    assert content[0]['video_metadata'] == {'start_offset': '10s', 'end_offset': '20s'}
+    assert content[0].get('video_metadata') == {'start_offset': '10s', 'end_offset': '20s'}
 
 
 async def test_http_video_url_uses_file_uri_on_google_vertex(mocker: MockerFixture):


barapa:
Good catch! Updated the fix to preserve the google-gla fallback behavior.

The condition is now:
```python
elif isinstance(item, VideoUrl) and (item.is_youtube or item.url.startswith('gs://')):
```

This ensures:
- **YouTube URLs**: Use `file_uri` (works on both providers) ✓
- **GCS URIs (`gs://...`)**: Use `file_uri` (works on google-vertex) ✓  
- **Other HTTP URLs**: Fall through to `FileUrl` handling (download/inline for google-gla) ✓

The original regression concern is addressed - non-YouTube/non-GCS VideoUrls will still get downloaded for google-gla users.

barapa:
@codex pls review

barapa:
This mirrors the existing YouTube behavior. The original code also bypasses `force_download` for YouTube URLs:

```python
# Original code - no force_download check
elif isinstance(item, VideoUrl) and item.is_youtube:
    file_data_dict: FileDataDict = {'file_uri': item.url, ...}
```

Additionally, `force_download=True` on a GCS URI would fail anyway with `UserError('Downloading from protocol "gs://" is not supported.')` - GCS URIs cannot be downloaded, only accessed via `file_uri` on Vertex AI.

So we're extending the existing YouTube pattern to GCS, and the edge case (`force_download=True` on GCS) would error either way.

I'll add a code comment noting this mirrors YouTube behavior for clarity.

barapa:
@codex pls review

pydantic_ai_slim/pydantic_ai/models/google.py
```diff
@@ -686,10 +686,18 @@ async def _map_user_prompt(self, part: UserPromptPart) -> list[PartDict]:
...
                         part_dict['video_metadata'] = cast(VideoMetadataDict, item.vendor_metadata)
                     content.append(part_dict)
-                elif isinstance(item, VideoUrl) and item.is_youtube:
+                elif isinstance(item, VideoUrl) and (
+                    item.is_youtube or (item.url.startswith('gs://') and self.system == 'google-vertex')
+                ):
+                    # YouTube URLs work on both google-gla and google-vertex
+                    # GCS URIs (gs://...) only work on google-vertex (Vertex AI can access GCS buckets)
+                    # GCS on google-gla falls through to FileUrl which raises clear error on download attempt
+                    # Other URLs fall through to FileUrl handling (download for google-gla)
```
DouweM, line 695: I think we should also set `video_metadata` for `VideoUrl` that fall back to downloading instead of sending the URL directly!

barapa: Done. Tested and confirmed the API respects the offsets even with inline_data.

PR #3807, opened 2025-12-22
Add `FileStore` for storing binary content in message history

### LINKED ISSUE #3073: Automatically store `BinaryContent`s in message history somewhere and turn them into `FileUrl`s
### Description

`BinaryContent` returned by the model (as in https://github.com/pydantic/pydantic-ai/pull/2970) or uploaded by a user (as in https://ai.pydantic.dev/input/) is typically going to be too large to store in a DB or send to an OpenTelemetry backend, so it'd be good to have a way to specify a `store_blob` function that turns `BinaryContent` into a `FileUrl`, and have that be in the stored message history. 

If the file is stored somewhere the model API can access, like a public bucket, the `FileUrl` will not need be converted back into a `BinaryContent` when using the message history, but if that's not the case you'd need a `load_blob` function as well that is given the `FileUrl` and can check the URL (e.g. starting with `file:` or `x-custom:` or `https://my-bucket.s3`), and download it if necessary. 

The functions should probably take the run context so they can access deps to know where to store the data, e.g. based on the user or a region.

https://github.com/pydantic/pydantic-ai/pull/2970 currently adds support for `output_type=BinaryImage`, and with this feature it could support `output_type=ImageUrl` as well, with the binary content returned by the model automatically uploaded to blob storage.

Inspired by https://github.com/nwcell/pydantic-ai-stash by @nwcell.

### References

_No response_

dsfaccini:
Closes #3073

## Problem

We generally prefer URLs (when providers support them) in the message history instead of having long byte strings in-memory. This also aids in reducing latency as media doesn't need to be constantly sent through the wire.

## Solution

`file_store_processor` - a history processor that:
- Uploads BinaryContent/FilePart to external storage (S3)
- Converts to URL types for models that support them
- Converts back to bytes for models that don't (bidirectional)
- Deduplicates uploads (same content = one upload)
- Refreshes expiring presigned URLs automatically

## Usage

```python
from pydantic_ai import Agent
from pydantic_ai.file_store import S3FileStore, file_store_processor

# Uses env vars: S3_ENDPOINT, S3_ACCESS_KEY_ID, S3_SECRET_ACCESS_KEY, S3_REGION
store = S3FileStore(bucket='my-bucket')
agent = Agent('openai:gpt-4o', history_processors=[file_store_processor(store)])
```

Custom implementation:

```python
from pydantic_ai.file_store import FileStore, file_store_processor

class MyFileStore(FileStore):
    async def store(self, key: str, data: bytes) -> str:
        # upload to your storage
        return key

    async def retrieve(self, key: str) -> bytes:
        # download from your storage
        return data

    def get_download_uri(self, key: str) -> str:
        return f'https://my-cdn.com/{key}'

    async def exists(self, key: str) -> bool:
        return True  # check if exists
```

## Changes

- `pydantic_ai/_file_store/` - FileStore protocol + processor
- `pydantic_ai/_file_store/s3/` - S3FileStore implementation
- `pydantic_ai/messages.py` - Added URL types to ModelResponsePart
- `docs/message-history.md` - "Storing Binary Content" section
- 80 tests (unit + VCR integration)

## Docs

- [Storing Binary Content](https://ai.pydantic.dev/message-history/#storing-binary-content) (TODO replace with preview link)
- [API Reference](https://ai.pydantic.dev/api/file_store/) (TODO replace with preview link)

## Notes

- Uses history_processors (not a separate agent param) - more flexible, follows existing patterns
- FileStore protocol doesn't have RunContext access - could be added later if needed


docs/message-history.md
```diff
@@ -552,6 +552,227 @@ agent = Agent('openai:gpt-5', history_processors=[filter_responses, summarize_ol
...
+)
+
+# Binary content is automatically stored and converted
+result = agent.run_sync('Generate an image of a sunset')
+# The FilePart in the response is uploaded to S3 and converted to ImageUrl
+```
+
+### How It Works
+
+The `file_store_processor` performs bidirectional conversion based on the model being used:
```
DouweM, line 585: How is the bidirectionality actually implemented? `history_processors` only transform messages before they're sent to the model, they don't transform messages coming back from the model, so I'm not sure this feature can/should be built (entirely) on `history_processors`.

pydantic_ai_slim/pydantic_ai/_file_store/__init__.py
```diff
@@ -0,0 +1,426 @@
...
+    # OpenAI: images via URL ok (unless force_download), audio/docs need bytes
+    if system == 'openai':
+        return {'image': True, 'audio': False, 'video': False, 'document': False}
+
+    # Anthropic: images via URL ok
+    if system == 'anthropic':
+        return {'image': True, 'audio': False, 'video': False, 'document': True}
+
+    # Google Vertex: most URLs ok
+    if system == 'google-vertex':
```
DouweM, line 166: You know I would never accept hard-coded checks like this, rather than some generic way to specify this using profile or something 😄 So I assume you haven't actually given this PR the human touch yet (which is fine as it's not a priority this month anyway), so  i'll put it in back in draft.

dsfaccini: Yeah I haven't worked on this since I put it up, thought it would be a good one to have up our pocket. Also when I started work on this I tried to make it as much a separate feature from the codebase as possible (i.e. not touch internals) which is why I didn't design an abstraction to use directly in models.

I'll keep this one inactive until we wanna prio it

dsfaccini:
this is now achievable with capabilities

PR #3810, opened 2025-12-23
Fix `BaseToolCallPart.has_content()` to always return `True`

### LINKED ISSUE #3793: Bug: `ToolCallPart.has_content()` returns False for falsy argument values
### Initial Checks

## Description

The `has_content()` method in `ToolCallPart` (and `BaseToolCallPart`) incorrectly returns `False` when the tool arguments contain falsy values like `False`, `0`, `""`, `[]`, `{}`, or `None`.

## Location

`pydantic_ai_slim/pydantic_ai/messages.py:1189-1196`

```python
def has_content(self) -> bool:
    """Return `True` if the arguments contain any data."""
    if isinstance(self.args, dict):
        # TODO: This should probably return True if you have the value False, or 0, etc.
        #   It makes sense to me to ignore empty strings, but not sure about empty lists or dicts
        return any(self.args.values())  # <-- Bug is here
    else:
        return bool(self.args)
```

The issue is that `any(self.args.values())` evaluates the truthiness of the values, not whether they exist.

## Reproduction

```python

# These should all return True (args dict is not empty)
print(ToolCallPart(tool_name="test", args={"flag": False}).has_content())  # Returns: False (BUG)
print(ToolCallPart(tool_name="test", args={"count": 0}).has_content())     # Returns: False (BUG)
print(ToolCallPart(tool_name="test", args={"name": ""}).has_content())     # Returns: False (BUG)
print(ToolCallPart(tool_name="test", args={"items": []}).has_content())    # Returns: False (BUG)
print(ToolCallPart(tool_name="test", args={"data": {}}).has_content())     # Returns: False (BUG)
print(ToolCallPart(tool_name="test", args={"value": None}).has_content())  # Returns: False (BUG)

# This correctly returns False (args dict is empty)
print(ToolCallPart(tool_name="test", args={}).has_content())  # Returns: False (CORRECT)
```

## Impact

This bug affects `stream_responses()` in `pydantic_ai/result.py:85-92`:

```python
async def stream_responses(self, *, debounce_by: float | None = 0.1) -> AsyncIterator[_messages.ModelResponse]:
    msg = self.response
    for part in msg.parts:
        if part.has_content():  # <-- Bug causes valid responses to not be yielded
            yield msg
            break
```

When a tool call has arguments with falsy values, the response might not be yielded at the start of streaming.

## Proposed Fix

Replace `any(self.args.values())` with `bool(self.args)`:

```python
def has_content(self) -> bool:
    """Return `True` if the arguments contain any data."""
    if isinstance(self.args, dict):
        return bool(self.args)  # True if dict has any keys
    else:
        return bool(self.args)
```

Or simplified:

```python
def has_content(self) -> bool:
    """Return `True` if the arguments contain any data."""
    return bool(self.args)
```

## Example Code

```python

# Bug demonstration
part = ToolCallPart(tool_name="set_flag", args={"enabled": False})
print(f"args: {part.args}")
print(f"has_content(): {part.has_content()}")  # Returns False, should be True
```

## Version

Latest main branch (commit ad5ba4f1)

### Example Code

```Python

```

```Text
Python: 3.13.5
pydantic-ai: 1.37.0
```

sotopelaez092-star:
## Summary
- Fix `BaseToolCallPart.has_content()` to always return `True`
- A tool call is always considered to have content, even without arguments

Fixes #3793

## Test plan
- [x] Existing tests pass
- [x] Verified the fix with manual testing


pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -1177,12 +1177,7 @@ def args_as_json_str(self) -> str:
 
     def has_content(self) -> bool:
         """Return `True` if the arguments contain any data."""
-        if isinstance(self.args, dict):
-            # TODO: This should probably return True if you have the value False, or 0, etc.
-            #   It makes sense to me to ignore empty strings, but not sure about empty lists or dicts
-            return any(self.args.values())
-        else:
-            return bool(self.args)
+        return bool(self.args)
```
DouweM: This doesn't always return `True` yet because `{}` is falsy :) And let's have a test for this

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/messages.py
tomsen02: Added a test for ToolCallPart.has_content(). I kept return bool(self.args) so that {} returns False (to avoid yielding empty values at the start of streaming), while args with falsy values like {'enabled': False} now correctly return True.

tomsen02:
The coverage drop on line 83 of result.py is because our fix to has_content() now yields more intermediate states during streaming (e.g., {'first': ''} now returns True). This causes the final response to match the last intermediate response, so line 83 (the final yield after the loop) is no longer executed.

  Should I add a specific test to cover this line, or is there another approach you'd prefer?

pydantic_ai_slim/pydantic_ai/messages.py
DouweM: I'm confused, did you change your mind from https://github.com/pydantic/pydantic-ai/issues/3793#issuecomment-3684723697?

tomsen02: I initially tried return True as you suggested, but it caused 9 test failures because streaming would yield empty {} at the start.

  So return bool(self.args) is a compromise:
  - {} → False (preserves existing streaming behavior)
  - {'enabled': False} → True (fixes the bug)

  If you'd prefer return True, I can update all the affected streaming tests. Let me know which approach you prefer!

DouweM: Yeah I'd like to see the tests that would need fixing, because it feels like we should treat any tool call as "having content", even if the tool call has no args or it does but no args have been streamed yet. 

But since `stream_responses` is used by `stream_output`, I could see the argument that such a tool call would have no content _that could be output_. But then the docstring is unclear about the real purpose of this method, and `ThinkingPart` also possibly has the wrong behavior.

So I'd like to see what happens with `True`, to figure whether it breaks `stream_output` in an unacceptable way, in which case we can go back depending on `self.args` but also make some more changes to docstrings etc.

I could also imagine that `self.args in ('', {})` will be better, so we don't ignore a `False` or `0` return value, even if not nested in a dict.

tomsen02: I've pushed the return True version so you can see the full CI results. 

Local testing shows 15 test failures, mainly in streaming tests (test_mistral.py and test_streaming.py). The failures are because yielding empty {} at the start of streaming causes validation errors.

Let me know how you'd like to proceed - I can update the affected tests if this behavior change is acceptable, or we can explore the self.args in ('', {}) alternative you mentioned.

DouweM: @sotopelaez092-star Hmm, code using `validate_response_output` _should_ handle `ValidationError`s, but since our tests worked fine without it as well, introducing this error would be fairly described as a breaking change.

So let's try `self.args in ('', {})`. 

Note that I'm about to do the last release of the year and sign off for the holiday break, so I'll get back to this in the new year!

tomsen02: Thanks for the suggestion! I implemented self.args not in ('', {}) but found it also needed None — because args defaults to None, and during streaming, the first DeltaToolCall (with only name, no json_args yet) creates a ToolCallPart with args=None.

Final fix: return self.args not in ('', {}, None)

Also added a test covering the original bug (falsy values like {'flag': False}, {'count': 0}).

Happy holidays and enjoy the break!

tomsen02:
Hi @dmontagu,

  I've updated the fix based on further investigation. The simple `self.args not in ('', {}, None)` approach was too broad and broke some Mistral streaming tests.

  **The updated fix:**
  - Returns `False` for `None`, `''`, `{}` (no args)
  - Returns `True` for `{'flag': False}`, `{'count': 0}` (the original bug fix)
  - Returns `False` for `{'text': ''}`, `{'response': []}` (empty strings/lists are treated as streaming placeholders)
  - Returns `True` for mixed values like `{'text': '', 'flag': False}`

  This preserves the original streaming behavior while still fixing the falsy values bug.

  **Regarding CI failures:** The 2 failing tests (`test_mistral_model_thinking_part`, `test_instrumented_model`) are unrelated to this PR. They fail due to `PytestUnraisableExceptionWarning` about unclosed SSL sockets - a known flaky test issue. All 3380+ other tests pass, including the Mistral streaming tests that were affected by the initial fix.

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -1179,12 +1179,14 @@ def args_as_json_str(self) -> str:
...
+        if self.args is None or self.args == '' or self.args == {}:
+            return False
         if isinstance(self.args, dict):
-            # TODO: This should probably return True if you have the value False, or 0, etc.
-            #   It makes sense to me to ignore empty strings, but not sure about empty lists or dicts
-            return any(self.args.values())
-        else:
-            return bool(self.args)
+            # Check if any value is non-empty. Empty strings and empty lists are placeholders
+            # during streaming, but False, 0, None, etc. ARE meaningful values (the bug fix).
```
DouweM, line 1236: The streaming placeholder thing seems very arbitrary to me, more focused on preserving the current tests than identifying the most correct behavior. Can you make it `return self.args not in ('', {}, None)` and update the tests?

dsfaccini:

```bash
git pull

deactivate

rm -rf .venv

make install
```

Issue #2906, opened 2025-09-15
AG-UI - How to force a StateSnapshotEvent at the end of the agent execution?


danielb-gyde:
### Question

I'm defining a test state and a route inside our FastAPI backend via ```pydantic_ai.ag_ui.run_ag_ui```
```python
class TestState(BaseModel):
    userName: str
    counter: int


class TestStateSnapshot(BaseModel):
    testState: TestState = Field(
        default_factory=lambda: TestState(userName="Test User", counter=1),
        description="The current test state",
    )
```

```python
@router.post("/chat")
async def run_agent(request: Request) -> Response:
    # [boilerplate]

    state = (
        TestStateSnapshot.model_validate(run_input.state)
        if run_input.state
        else TestStateSnapshot()
    )

    event_stream = run_ag_ui(
        main_agent,
        run_input,
        accept=accept,
        on_complete=on_request_complete,
        deps=StateDeps(state),
    )

    return StreamingResponse(event_stream, media_type=accept)
```

and I have a dummy-agent defining some tools and a sub-agent for agent orchestration:

```python
main_agent = Agent(
    name="Routing Agent",
    model=model,
    system_prompt=MAIN_AGENT_PROMPT,
    deps_type=StateDeps[TestStateSnapshot],
)


@main_agent.instructions
def beg_llm_to_update_state(ctx: RunContext[StateDeps[TestStateSnapshot]]) -> str:
    return "always call update_state when you're done"


@main_agent.tool
async def roll_dice(ctx: RunContext[StateDeps[TestStateSnapshot]]) -> DiceAgentResponse:
    """Roll a six-sided die and return the random result."""
    result = 6
    ctx.deps.state.testState.counter = result
    return DiceAgentResponse(roll=result)


@main_agent.tool
async def update_state(
    ctx: RunContext[StateDeps[TestStateSnapshot]],
) -> StateSnapshotEvent:
    return StateSnapshotEvent(type=EventType.STATE_SNAPSHOT, snapshot=ctx.deps.state)


sub_agent = Agent(
    name="Pink Fluffy Unicorn agent",
    model=model, 
    deps_type=StateDeps[TestStateSnapshot],
)

@main_agent.tool
async def pink_fluffy_unicorns(ctx: RunContext[StateDeps[TestStateSnapshot]]):
'''Test agent that changes the count to 42 in one of it's tools'''
    response = await sub_agent.run(
        message_history=ctx.messages,
        deps=ctx.deps,
    )
[...]

PR #3811, opened 2025-12-23
Add `metadata` support for `DeferredToolResults`

Wh1isper:
I know we can override parameters using `override_args`, but for non-existent parameters (like api_key) or scenarios where user input requires preprocessing, I'd like to have `metadata` that passes HITL preprocessing results to the tool. This could be the user's raw input, flags, etc. When the tool actually executes, it can respond based on this metadata rather than relying on parameters only.

```python
"""Example: Secure API Key via DeferredToolResults.metadata

Demonstrates passing sensitive data (API key) to a tool WITHOUT exposing it to LLM.
The call_api tool has NO parameters - the API key comes entirely from user input
via ToolApproved.metadata, completely bypassing the LLM.
"""

from pydantic_ai import (
    Agent,
    ApprovalRequired,
    DeferredToolRequests,
    DeferredToolResults,
    RunContext,
    ToolApproved,
)

agent = Agent(TestModel(), output_type=[str, DeferredToolRequests])


@agent.tool
def call_api(ctx: RunContext[None]) -> str:
    """Call the API. No parameters needed - API key is provided securely via HITL."""
    if not ctx.tool_call_approved:
        raise ApprovalRequired(metadata={"message": "Please provide your API key"})

    # Get API key from approval metadata - never seen by LLM
    api_key = (
        ctx.tool_call_metadata.get("api_key", "") if ctx.tool_call_metadata else ""
    )
    if not api_key:
        return "Error: No API key provided"

    # Use the API key (masked in output for demo)
    masked = f"{api_key[:4]}...{api_key[-4:]}" if len(api_key) > 8 else "****"
    return f"API called successfully with key: {masked}"


if __name__ == "__main__":
    # Step 1: User request triggers tool call
    result = agent.run_sync("Please call the API")

    if isinstance(result.output, DeferredToolRequests):
        # Step 2: Tool needs approval - get API key from user
        print("Tool requires API key input.")
        api_key = input("Enter API key: ")

        # Step 3: Get the actual tool_call_id from the approval request
        tool_call_id = result.output.approvals[0].tool_call_id

        # Step 4: Continue with API key in metadata (bypasses LLM entirely)
        result = agent.run_sync(
            deferred_tool_results=DeferredToolResults(
                approvals={tool_call_id: ToolApproved()},
                metadata={tool_call_id: {"api_key": api_key}},
            ),
        )

```


docs/deferred-tools.md
```diff
@@ -25,7 +25,7 @@ To require approval for calls to tools provided by a [toolset](toolsets.md) (lik
...
+Once you've gathered the user's approvals or denials, you can build a [`DeferredToolResults`][pydantic_ai.tools.DeferredToolResults] object with an `approvals` dictionary that maps each tool call ID to a boolean, a [`ToolApproved`][pydantic_ai.tools.ToolApproved] object (with optional `override_args` and `metadata`), or a [`ToolDenied`][pydantic_ai.tools.ToolDenied] object (with an optional custom `message` to provide to the model). When using `ToolApproved` with `metadata`, the metadata will be available in the tool's [`RunContext.tool_call_metadata`][pydantic_ai.tools.RunContext.tool_call_metadata] attribute. This `DeferredToolResults` object can then be provided to one of the agent run methods as `deferred_tool_results`, alongside the original run's [message history](message-history.md).
```
DouweM: Mirroring `DeferredToolRequests.metadata`, what do you think of the way to pass this data in being `DeferredToolResults.metadata: dict[str, Any]`? 

That would help with https://github.com/pydantic/pydantic-ai/issues/2906#issuecomment-3470804847:

> I have plans to make [ai.pydantic.dev/deferred-tools](https://ai.pydantic.dev/deferred-tools/) work cleanly across handoffs too, bubbling up to the top to present to the user, with the context necessary (i.e. message history) to push it down the tool call stack upon resumption and have all the agents pick up where they left off.

Wh1isper: Makes sense. Moved metadata to `DeferredToolResults.metadata: dict[str, dict[str, Any]]` (keyed by `tool_call_id`), mirroring `DeferredToolRequests.metadata`.

Not sure if this fully covers the handoff case you mentioned, let me know if anything else is needed

pydantic_ai_slim/pydantic_ai/_tool_manager.py
```diff
@@ -103,6 +104,7 @@ async def handle_call(
             allow_partial: Whether to allow partial validation of the tool arguments.
             wrap_validation_errors: Whether to wrap validation errors in a retry prompt part.
             approved: Whether the tool call has been approved.
+            metadata: Additional metadata from ToolApproved.metadata.
```
DouweM: Outdated now right

DouweM (changes requested):
@Wh1isper Thanks, this looks good, but I'm about to sign off for the holidays and do the last (minor) release of the year, and I don't want to merge this too quickly and then realize I needed something else for the handoffs use case, so I'll get back to contemplate and merge this in the new year!

Wh1isper:
@DouweM Thank you for the quick response. For now, I can implement a workaround via ctx.

Wh1isper:
ping

DouweM:
@Wh1isper Thank you!

PR #3817, opened 2025-12-23
ci: drop lint on higher python version

Kludex:


.github/workflows/ci.yml
Kludex, line 1: CI doesn't need to run on any specific python version. The linter/type checker tools need to know they are meant to run in the lowest supported version.

PR #3831, opened 2025-12-23
WIP: defer_loading=True and tool search

t0yv0:
- Closes #<issue>

### Pre-Review Checklist

- [ ] **PR title** is fit for the release changelog.

### Pre-Merge Checklist

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/agent/__init__.py
+++ b/pydantic_ai_slim/pydantic_ai/agent/__init__.py
@@ -62,10 +62,10 @@ from ..toolsets._dynamic import (
     DynamicToolset,
     ToolsetFunc,
 )
+from ..toolsets._searchable import SearchableToolset
 from ..toolsets.combined import CombinedToolset
 from ..toolsets.function import FunctionToolset
 from ..toolsets.prepared import PreparedToolset
-from ..toolsets.searchable import SearchableToolset
 from .abstract import AbstractAgent, AgentMetadata, EventStreamHandler, Instructions, RunOutputDataT
 from .wrapper import WrapperAgent
 
--- a/pydantic_ai_slim/pydantic_ai/models/__init__.py
+++ b/pydantic_ai_slim/pydantic_ai/models/__init__.py
@@ -50,7 +50,7 @@ from ..profiles import DEFAULT_PROFILE, ModelProfile, ModelProfileSpec
 from ..providers import Provider, infer_provider
 from ..settings import ModelSettings, merge_model_settings
 from ..tools import ToolDefinition
-from ..toolsets.searchable import is_active
+from ..toolsets._searchable import is_active
 from ..usage import RequestUsage
 
 KnownModelName = TypeAliasType(
--- a/pydantic_ai_slim/pydantic_ai/toolsets/_searchable.py
+++ b/pydantic_ai_slim/pydantic_ai/toolsets/_searchable.py
@@ -40,8 +40,8 @@ class SearchableToolset(WrapperToolset[AgentDepsT]):
 
         if search_tool.name in all_tools:
             raise UserError(
-                f"Tool name {search_tool.name!r} is reserved by Pydantic AI for implementing defer_loading=True tools. "
-                f"Rename the tool or wrap the toolset in a `PrefixedToolset` to avoid the conflict."
+                f'Tool name {search_tool.name!r} is reserved by Pydantic AI for implementing defer_loading=True tools. '
+                f'Rename the tool or wrap the toolset in a `PrefixedToolset` to avoid the conflict.'
             )
         all_tools[search_tool.name] = search_toolset_tool
 
--- a/pydantic_ai_slim/pydantic_ai/toolsets/__init__.py
+++ b/pydantic_ai_slim/pydantic_ai/toolsets/__init__.py
@@ -1,4 +1,5 @@
 from ._dynamic import ToolsetFunc
+from ._searchable import SearchableToolset
 from .abstract import AbstractToolset, ToolsetTool
 from .approval_required import ApprovalRequiredToolset
 from .combined import CombinedToolset
@@ -8,7 +9,6 @@ from .function import FunctionToolset
 from .prefixed import PrefixedToolset
 from .prepared import PreparedToolset
 from .renamed import RenamedToolset
-from .searchable import SearchableToolset
 from .wrapper import WrapperToolset
 
 __all__ = (
--- a/tests/test_toolsets.py
+++ b/tests/test_toolsets.py
@@ -28,7 +28,7 @@ from pydantic_ai.exceptions import ModelRetry, ToolRetryError, UnexpectedModelBe
 from pydantic_ai.models.test import TestModel
 from pydantic_ai.tools import ToolDefinition
 from pydantic_ai.toolsets._dynamic import DynamicToolset
-from pydantic_ai.toolsets.searchable import is_active, is_search_tool
+from pydantic_ai.toolsets._searchable import is_active, is_search_tool
 from pydantic_ai.usage import RunUsage
 
 pytestmark = pytest.mark.anyio
@@ -1003,6 +1003,11 @@ async def test_searchable_toolset_load_tools_name_conflict():
         """A tool that conflicts with the reserved name"""
         return x
 
+    @base_toolset.tool(defer_loading=True)
+    def some_deferred_tool(y: str) -> str:
+        """A deferred tool to trigger search tool creation"""
+        return y
+
     searchable = SearchableToolset[None](base_toolset)
     ctx = build_run_context(None)
 
@@ -1010,7 +1015,7 @@ async def test_searchable_toolset_load_tools_name_conflict():
         UserError,
         match=re.escape(
             "Tool name 'load_tools' is reserved by Pydantic AI for implementing defer_loading=True tools. "
-            "Rename the tool or wrap the toolset in a `PrefixedToolset` to avoid the conflict."
+            'Rename the tool or wrap the toolset in a `PrefixedToolset` to avoid the conflict.'
         ),
     ):
         await searchable.get_tools(ctx)


pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -1505,7 +1506,7 @@ def _get_toolset(
         toolset = CombinedToolset(toolsets)
 
         def copy_dynamic_toolsets(toolset: AbstractToolset[AgentDepsT]) -> AbstractToolset[AgentDepsT]:
-            if isinstance(toolset, DynamicToolset):
+            if isinstance(toolset, DynamicToolset) or isinstance(toolset, SearchableToolset):
```
t0yv0, line 1517: TODO: a Pythonic way to add a marker (protocol?) for toolsets that can prefer to copy their instance state for each id so the instance checks do not grow here.

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -701,6 +702,11 @@ def prepare_request(
                     f'Builtin tool(s) {unsupported_names} not supported by this model. Supported: {supported_names}'
                 )
 
+        # Filter out defer_loading=True definitions that have not yet been activated by tool search.
```
t0yv0, line 705: This will change subsequently as models with built-in tool search are supported natively.

pydantic_ai_slim/pydantic_ai/toolsets/_searchable.py
```diff
@@ -0,0 +1,180 @@
...
+    return tool, ToolsetTool(
+        toolset=toolset,
+        tool_def=tool.tool_def,
+        max_retries=tool.max_retries or 3,
+        args_validator=tool.function_schema.validator,
+    )
+
+
+@dataclass(kw_only=True)
+class _SearchToolsetToolWrapper(ToolsetTool[AgentDepsT]):
```
t0yv0, line 111: This wrapper may be unnecessary if get_tools runs sufficiently often and its results are not cached somewhere in the system. In this case we can simplify and get_tools can avoid a wrapper class, calculate the active flag for a tool immediately, and return it in the tool's metadata marker.

pydantic_ai_slim/pydantic_ai/toolsets/_searchable.py
```diff
@@ -0,0 +1,180 @@
...
+
+    def is_active(self, tool_def: ToolDefinition) -> bool:
+        return tool_def.name in self._active_tool_names
+
+
+def _search_tool(toolset: SearchableToolset) -> tuple[Tool, ToolsetTool[AgentDepsT]]:
+    async def search(ctx: RunContext[AgentDepsT], regex: str) -> list[str]:
+        return await toolset._search_tools(ctx, regex)
+
+    # TODO Check if pattern is a better name than regex.
```
t0yv0, line 83: One of the important considerations here is how good models are at actually finding tools with the search_tool, some more prompt tuning could be beneficial. I think examples could help the LLM, but I understand there is a strict token budget.

pydantic_ai_slim/pydantic_ai/toolsets/searchable.py
```diff
@@ -0,0 +1,180 @@
...
+
+        all_tools: dict[str, ToolsetTool[AgentDepsT]] = {}
+
+        for tool_name, tool in tools.items():
+            defer = tool.tool_def.defer_loading
+            all_tools[tool_name] = _SearchToolsetToolWrapper[AgentDepsT](tool, self) if defer else tool
+
+        search_tool, search_toolset_tool = _search_tool(toolset=self)
+
+        # TODO how to handle this error, or should we automatically disambiguate this name?
```
t0yv0: Need to look is it idiomatic in the codebase to generate unique tool names, probably this needs to be done deterministically.

pydantic_ai_slim/pydantic_ai/toolsets/_searchable.py
```diff
@@ -0,0 +1,180 @@
...
+        searchable_toolset: SearchableToolset,
+    ) -> None:
+        self.wrapped = tool
+        self._searchable_toolset = searchable_toolset
+        self.toolset = tool.toolset
+        self._tool_def = tool.tool_def
+        self.max_retries = tool.max_retries
+        self.args_validator = tool.args_validator
+
+    # TODO Mypy: Dataclass attribute may only be overridden by another attribute [misc]
```
t0yv0, line 130: Reminder to catch up on what is pythonic, if we still need a wrapper class.

t0yv0:
One design question is whether we want to expose SearchableToolset to the users, or else keep it hidden and auto-injected in the agent just once. An exposed Toolset must support nesting. The advantage of doing so would arise if we add more settings to control how the search is performed (regex vs BM or other strategies), and users would be able to control more specifically which part of their application uses the search strategy. 

Nesting could be supported as follows:

- SearchableToolset filters search tools from get_tools; since there is a root SearchableToolset, this will be the only one creating a search tool instance, of which there will be at most one.

- SearchableToolset dispatches down to the appropriate SearchableToolset instance to implement the correct instance-specific settings (regex or BM25), but stores the search results in its own (root instance) state

Opting to hide SearchableToolset for simplicity but can follow up with another PR to expose it.

Good note: the interface needs to be uniform.

t0yv0:
Next up: tests for searchable toolset, doc section.

pydantic_ai_slim/pydantic_ai/toolsets/_searchable.py
```diff
@@ -0,0 +1,181 @@
...
+    """Tracks activated tool names."""
+
+    async def get_tools(self, ctx: RunContext[AgentDepsT]) -> dict[str, ToolsetTool[AgentDepsT]]:
+        tools = await self.wrapped.get_tools(ctx)
+
+        # No need to add a search tool if there are no defer_loading tools.
+        if not any(t.tool_def.defer_loading for t in tools.values()):
+            return tools
+
+        # If any nested SearchableToolset instances have added their own search tools, drop those so that only one
```
t0yv0, line 29: This is only needed if we do allow nesting.

tests/models/test_openai.py
```diff
@@ -3845,3 +3846,97 @@ async def test_openai_chat_audio_url_uri_encoding(allow_model_requests: None):
     # Expect Data URI with correct MIME type for mp3
     assert audio_part['input_audio']['data'] == data_uri
     assert audio_part['input_audio']['format'] == 'mp3'
+
+
+async def test_defer_loading_tools_activated_explicitly(allow_model_requests: None):
```
t0yv0, line 3851: This tests survives from an initial differential test suite for openai vs anthropic, but not assuming anything about OpenAI specific anymore in the last implementation, possibly needs a new home.

pydantic_ai_slim/pydantic_ai/agent/__init__.py
t0yv0: I would also like to cover test this line, namely that defer_loading tools get deferred again on next run_id.

t0yv0: Ah I see why initial attempts fail here, the root SearchableToolset is constructed already, so to cover this we need to have the nested SearchableToolset instances.

pydantic_ai_slim/pydantic_ai/toolsets/function.py
```diff
@@ -47,6 +47,7 @@ class FunctionToolset(AbstractToolset[AgentDepsT]):
     docstring_format: DocstringFormat
     require_parameter_descriptions: bool
     schema_generator: type[GenerateJsonSchema]
+    defer_loading: bool
```
t0yv0, line 50: Need to cover test defaulting here, that defer_loading on FunctionToolset applies to inner tools unless they override this setting.

t0yv0:
Some context management optimizations come to mind on the interface between the model and the search tool. Currently the model can only search by a regex. If it makes unlucky guesses it does not know how many tools are present. There are two situations: perhaps there's too many tools to dump into context at once so `.*` regex defeats the purpose, and the other situation is when there are fewer tools available and it would be more efficient to list all their names at once to the model. The second situation is more rare if the users follow recommendations and only use the feature for large tool collections. We could tweak something here by returning not just the matching tool names to the model, but also a count of all available extra tools as well as a first 5 or so tool names even when they do not match, similar how Google Search returns results even if there are no explicit matches.

t0yv0:
Should this be a beta feature at first?

```
At Pydantic, we like to move quickly and innovate! To that end, minor releases may introduce beta features (indicated by a beta module) that are active works in progress.
```

pydantic_ai_slim/pydantic_ai/tools.py
```diff
@@ -275,6 +275,7 @@ class Tool(Generic[ToolAgentDepsT]):
     metadata: dict[str, Any] | None
     timeout: float | None
     function_schema: _function_schema.FunctionSchema
+    defer_loading: bool
```
t0yv0, line 278: Since the changes in this file are public API changes they deserve some light testing to avoid regressing accidentally, similar to how function.py changes are covered.

docs/tools-advanced.md
```diff
@@ -162,6 +162,67 @@ print(result.output)
 
 Please note that validation of the tool arguments will not be performed, and this will pass all arguments as keyword arguments.
 
+## Deferred Tool Loading {#deferred-tool-loading}
```
t0yv0, line 165: We have a "deferred execution" feature, so the name is not the best.

t0yv0: Ideas: Progressive Discovery.

docs/tools-advanced.md
```diff
@@ -162,6 +162,67 @@ print(result.output)
...
+@agent.tool_plain
+def get_current_time() -> datetime:
+    return datetime.now()
+
+@agent.tool_plain(defer_loading=True)
+def specialized_database_tool(table: str, query: str) -> str:
+    """Search a specialized database table."""
+    return f'Database result from {table}: {query}'
+
+result1 = await await agent.run('What time is it now?')
```
t0yv0, line 187: Run the examples.

pydantic_ai_slim/pydantic_ai/toolsets/_searchable.py
```diff
@@ -0,0 +1,185 @@
...
+
+_TOOL_SEARCH_METADATA_KEY = '__tool_search__'
+"""ToolDefinition.metadata key to store runtime metadata."""
+
+
+@dataclass
+class SearchableToolset(WrapperToolset[AgentDepsT]):
+    """A toolset that implements tool search and deferred tool loading."""
+
+    _active_tool_names: set[str] = field(default_factory=set)
```
t0yv0, line 19: Name is not a good identity possibly? What if new tools arise?

pydantic_ai_slim/pydantic_ai/toolsets/_searchable.py
```diff
@@ -0,0 +1,185 @@
...
+    """Tracks activated tool names."""
+
+    async def get_tools(self, ctx: RunContext[AgentDepsT]) -> dict[str, ToolsetTool[AgentDepsT]]:
+        tools = await self.wrapped.get_tools(ctx)
```
t0yv0, line 23: What if new tools show up here that match the search...

tests/models/test_openai.py
t0yv0: test_agent or test_toolset.

pydantic_ai_slim/pydantic_ai/models/__init__.py
t0yv0: Builtin fool fall-backs here perhaps.

Anthropic with native tool search:
  detect there is a built-in tool
  expose all the defer_loading tools

OpenAI without native tool search:
  expose pydantic search tool
  expose a subset of defer_loading_tools

No model-side filtering.

pydantic_ai_slim/pydantic_ai/toolsets/_searchable.py
```diff
@@ -0,0 +1,185 @@
...
+        'properties': {
+            'regex': {
+                'type': 'string',
+                'description': 'Regex pattern to search for relevant tools',
+            }
+        },
+        'required': ['regex'],
+    }
+
+    desc = 'Search and load additional tools that defer tool loading'
```
t0yv0, line 96: Need to be able to change based on how intelligent the model is.  Per model defaults?

pydantic_ai_slim/pydantic_ai/toolsets/_searchable.py
```diff
@@ -0,0 +1,185 @@
...
+    tool = Tool.from_schema(search, name='load_tools', description=desc, json_schema=schema, takes_ctx=True)
```
t0yv0, line 97: model tool is also load bearing.

Danipulok:
@t0yv0, hey here!
Just curious, how does this workflow strategy work?
Do you generate the code with AI help, then leave comments in this PR and make AI assistant fix what you mentioned in your comments?
It just seems there are some thoughts and comments that matter only in progress and would not help the final reviewer

DouweM:
Anton indicated he won't be able to work on this anymore so I'll close it.

PR #3833, opened 2025-12-23
tests: split tests into durable and not durable

Kludex:


DouweM:
@Kludex What's the plan here?

Kludex:
> @Kludex What's the plan here?

This PR is ready, the failure on the pipeline was flakiness.

Kludex:
I didn't want to merge this because I'm not sure this is what we want. I think what we really want is to drop support for Python 3.10... 😅 

The real reason the pipeline is slow seems to be the snapshots on 3.10.

DouweM:
Yeah I don't think this is what we want either, so I'll close it for now...

Issue #3783, opened 2025-12-20
Support Optional System Prompts


mrdkucher:
### Description

There are times when a system prompt decorator decorates a function which may provide part of a system prompt, but fail during execution, or otherwise be invalid, forcing a developer to supply an empty string as a part of the payload.

Instead of adding a message with contents equal to the return of the function, simply check if the content evaluates to a non-None value.

```python
from types import NoneType

from pydantic_ai import Agent, RunContext

agent = Agent("test", deps_type=NoneType)


@agent.system_prompt
def simple_system_prompt() -> str:
    return "foobar"


@agent.system_prompt
def none_system_prompt() -> None:
    return None


@agent.system_prompt(dynamic=True)
async def async_system_prompt(ctx: RunContext[str]) -> str:
    return f"{ctx.deps} is the best"


result_sync = agent.run_sync(user_prompt="hello")
print(
    type(result_sync.all_messages()[0].parts[0].content),
    result_sync.all_messages()[0].parts[0].content,
)
print(
    type(result_sync.all_messages()[0].parts[1].content),
    result_sync.all_messages()[0].parts[1].content,
)

print(
    type(result_sync.all_messages()[0].parts[2].content),
    result_sync.all_messages()[0].parts[2].content,
)
```
Output:
```
<class 'str'> foobar
<class 'NoneType'> None
<class 'str'> None is the best
```

The messages look like:
```json
[
  {
    "parts": [
      {
        "content": "foobar",
        "timestamp": "2025-12-20T01:25:20.462142+00:00",
        "dynamic_ref": null,
        "part_kind": "system-prompt"
      },
      {
        "content": null,
        "timestamp": "2025-12-20T01:25:20.462471+00:00",
        "dynamic_ref": null,
        "part_kind": "system-prompt"
      },
      {
        "content": "None is the best",
        "timestamp": "2025-12-20T01:25:20.462477+00:00",
        "dynamic_ref": "async_system_prompt",
        "part_kind": "system-prompt"
      },
      {
        "content": "hello",
        "timestamp": "2025-12-20T01:25:20.462480+00:00",
        "part_kind": "user-prompt"
      }
    ],
    "instructions": null,
    "kind": "request",
    "run_id": "61dc9736-8f1d-4666-abc4-5c31f6e3fd57",
    "metadata": null
  },
  {
    "parts": [
      {
        "content": "success (no tool calls)",
        "id": null,
        "provider_details": null,
        "part_kind": "text"
      }
[...]

PR #3834, opened 2025-12-23
Support system prompts functions returning `None`

mrdkucher:
If a system prompt function encounters an error or otherwise evaluates to None or an empty string, it should be omitted from model requests. This matches the behavior of Instructions functions.

NOTE: Dynamic system prompts which are re-evaluated use a placeholder empty string upon a None return value (maintaining existing behavior). This is necessary to ensure we re-evaluate the dynamic function on the next `run` call.

Closes: #3783


DouweM:
@mrdkucher Thank you!

PR #3835, opened 2025-12-23
qol: Small refactors

dsfaccini:


pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -1835,28 +1835,18 @@ async def _map_messages(  # noqa: C901
...
                         if item.provider_name == self.system and send_item_ids:  # pragma: no branch
+                            content_is_dict = isinstance(item.content, dict)
+                            status = item.content.get('status') if content_is_dict else None
                             if (
                                 item.tool_name == CodeExecutionTool.kind
                                 and code_interpreter_item is not None
-                                and isinstance(item.content, dict)
-                                and (content := cast(dict[str, Any], item.content))  # pyright: ignore[reportUnknownMemberType]
-                                and (status := content.get('status'))
+                                and status
```
DouweM: Can we pull `if status:` up then? Then we could also go back to the walrus operator (`:=`) approach

dsfaccini: I shortened it a bit more :)

```python
kind_to_item = {
    CodeExecutionTool.kind: code_interpreter_item,
    WebSearchTool.kind: web_search_item,
    FileSearchTool.kind: file_search_item,
}
```

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -371,6 +371,7 @@ class MCPServer(AbstractToolset[Any], ABC):
     _cached_tools: list[mcp_types.Tool] | None
     _cached_resources: list[Resource] | None
 
+    # TODO(David) v2: enforce the arguments to be passed as keyword arguments only
```
dsfaccini: @dmontagu https://pydanticlogfire.slack.com/archives/C083V7PMHHA/p1767126476473499?thread_ts=1767125068.885079&cid=C083V7PMHHA

tests/models/test_openai.py
```diff
@@ -839,21 +839,6 @@ async def test_openai_pass_custom_system_prompt_role(allow_model_requests: None,
...
-) -> None:
-    model = OpenAIChatModel(  # type: ignore[reportDeprecated]
-        'o1-mini', provider=OpenAIProvider(api_key=openai_api_key), system_prompt_role=system_prompt_role
-    )
-    agent = Agent(model=model, system_prompt='You are a helpful assistant.')
-
-    with pytest.raises(ModelHTTPError, match=r".*Unsupported value: 'messages\[0\]\.role' does not support.*"):
-        await agent.run('Hello')
-
-
```
dsfaccini, line 883: deleted this test because 
1. o1-mini is deprecated
2. upgrading to o4-mini makes the test pass
3. PR's title was [Add test to document OpenAI o1-mini behavior on system role #977](https://github.com/pydantic/pydantic-ai/pull/977) but like, this is probably something users just learn by themselves using the models, and `test_openai` is massive so "documentation" is not realistic in this case

say the word and I'll go ahead and delete all other o1-mini tests or upgrade them where relevant

pydantic_ai_slim/pydantic_ai/mcp.py
DouweM: Use DavidM or DavidSF for clarity :) And for v2 let's do `TODO (v2)`, we already have that in some places

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -1795,30 +1795,15 @@ async def _map_messages(  # noqa: C901
...
-                                file_search_item['status'] = status
+                            content_is_dict = isinstance(item.content, dict)
+                            status = item.content.get('status') if content_is_dict else None
+                            kind_to_item = {
+                                CodeExecutionTool.kind: code_interpreter_item,
+                                WebSearchTool.kind: web_search_item,
+                                FileSearchTool.kind: file_search_item,
+                            }
+                            if status and (_item := kind_to_item.get(item.tool_name)) is not None:
+                                _item['status'] = status
```
DouweM: Underscore prefixes imply a variable is "hidden", so it should just be for private things, and for unused variables. So this one needs a real name!

pydantic_ai_slim/pydantic_ai/providers/google.py
```diff
@@ -100,6 +100,9 @@ def __init__(
             base_url: The base URL for the Google API.
         """
         if client is None:
+            # Track if api_key was explicitly passed (for Vertex AI express mode)
+            api_key_explicitly_passed = api_key is not None
```
DouweM: This looks like a bugfix, not a refactor? What scenario was previously broken?

tests/models/test_openai.py
DouweM: if we remove the test, we should also remove the logic it (indirectly) tests:

https://github.com/pydantic/pydantic-ai/blob/1ad834b5b824c67e38dd19cf2ceb1b37ebe6cac0/pydantic_ai_slim/pydantic_ai/profiles/openai.py#L123-L125

but I don't want to remove that code until we are sure no one is using us with o1. so until it's fully removed, i think we need to keep this

pydantic_ai_slim/pydantic_ai/providers/google.py
dsfaccini: yeah it was bugging me: I have proper vertex credentials locally but the google provider automatically picks up the `gemini_api_key` var from my env and uses that for vertex auth instead, so what this is doing is ignoring that when using vertex

dsfaccini: to be clear, when auto uses the env key auth fails, which is why it bugs me, so I needed to unset it every time, and yeah, the change is so it doesn't take it into account when we're using vertex

tests/models/test_openai.py
dsfaccini: I was waiting for CI to see if a gap appeared from me removing the test but because no gap appeared I assumed it was fine

pydantic_ai_slim/pydantic_ai/providers/google.py
DouweM: so this is a bugfix / change in behavior, not a refactor? then it should have its own PR and a test right? is there any scenario thinkable where someone was relying on the original behavior, and we now broke it for them?

tests/models/test_openai.py
DouweM: yeah one-line conditionals that have only one branch tested won't show up as missing coverage :)

DouweM (changes requested):

tests/models/test_openai.py
DouweM: so we should remove this deletion right

PR #3841, opened 2025-12-24
Support Google embedding models

tboser:
- Closes https://github.com/pydantic/pydantic-ai/issues/3843

also fixes an existing bug in tests/docs which weren't actually using dimension controls.

### Pre-Review Checklist

- [x] Any **AI generated code** has been reviewed line-by-line by the human PR author, who stands by it.
- [x] No **breaking changes** in accordance with the [version policy](https://github.com/pydantic/pydantic-ai/blob/main/docs/version-policy.md).
- [x] **Linting and type checking** pass per `make format` and `make typecheck`.
- [ ] **PR title** is fit for the release changelog.

### Pre-Merge Checklist

- [ ] New **tests** for any fix or new behavior, maintaining 100% coverage.
- [ ] Updated **documentation** for new features and behaviors, including docstrings for API docs.

PR #3846, opened 2025-12-25
Fix flaky test_mistral_model_thinking_part test

sarth6:
## Summary

Fixes a flaky test in `test_mistral.py::test_mistral_model_thinking_part` by suppressing `PytestUnraisableExceptionWarning`.

## Problem

The test `test_mistral_model_thinking_part` intermittently fails with:
```
PytestUnraisableExceptionWarning: Exception ignored in: <ssl.S...
```

This occurs because the test creates both `OpenAIResponsesModel` and `MistralModel` providers, each with their own SSL-backed HTTP clients. When the test completes, Python's garbage collector tries to clean up these connections synchronously, which triggers an exception in the SSL socket's `__del__` method.

## Solution

Add `@pytest.mark.filterwarnings('ignore::pytest.PytestUnraisableExceptionWarning')` to suppress this known issue. This is a targeted fix that only affects this specific test, following the same pattern used elsewhere in the codebase (e.g., `test_huggingface.py` uses `ignore::ResourceWarning`).

## Test plan

- [x] Test passes consistently (verified 5+ runs)
- [x] Ruff lint and format checks pass

### DIFF
--- a/tests/models/test_mistral.py
+++ b/tests/models/test_mistral.py
@@ -2351,6 +2351,7 @@ async def test_mistral_model_instructions(allow_model_requests: None, mistral_ap
 
 
 @pytest.mark.vcr()
+@pytest.mark.filterwarnings('ignore::pytest.PytestUnraisableExceptionWarning')
 async def test_mistral_model_thinking_part(allow_model_requests: None, openai_api_key: str, mistral_api_key: str):
     openai_model = OpenAIResponsesModel('o3-mini', provider=OpenAIProvider(api_key=openai_api_key))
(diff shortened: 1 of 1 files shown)


Danipulok:
May close #3847 
But the implementation must be changed IMHO

dsfaccini:
hey guys @sarth6 @Danipulok I appreciate your initiative but I think this is a deeper issue hidden somewhere in a recent **merge** and it probably has something to do with http client not closing properly, see https://github.com/pydantic/pydantic-ai/pull/3841#discussion_r2646240638
image

we'll take care of it but in the meantime we'll deal with this by re-running CI

Issue #3847, opened 2025-12-25
Unclosed `SSLSocket` warning in tests


Danipulok:
### Initial Checks

- [x] I'm using the [latest version](https://github.com/pydantic/pydantic-ai/releases/latest) of Pydantic AI
- [x] I've searched for my issue in [the issue tracker](https://github.com/pydantic/pydantic-ai/issues) before opening this issue

### Description

CI/CD errors as in:
- https://github.com/pydantic/pydantic-ai/actions/runs/20493583661/job/58891435907
- https://github.com/pydantic/pydantic-ai/actions/runs/20501417326/job/58909215013?pr=3845
- https://github.com/pydantic/pydantic-ai/actions/runs/20501196804/job/58908698731

Danipulok:
Fixed in #3862

PR #3849, opened 2025-12-25
tests: only close unclosed http clients

Kludex:


Danipulok:
@Kludex, hey here!

I've created an issue for this
This PR:
Closes #3847

I didn't see you were working on this problem
Tried to fix something myself, but wasn't successful
Maybe this will give you some ideas, I tested a bunch of things: #3852

dsfaccini:
hey @Danipulok just fyi even though this seems like it's aiming for a fix for the current CI error it really is about a test optimization we talked about, which you can find in https://github.com/pydantic/pydantic-ai/pull/3778

idk inhowfar it could fix the current error but just wanted to note that so you don't get the wrong idea

Danipulok:
@dsfaccini got it, thank you very much! I though this is about that issue, yes

PR #3851, opened 2025-12-25
qol: Refactor openai

dsfaccini:
the `pydantic_ai/models/openai.py` file is 35000 tokens long which makes it inefficient for ai coding assistants to deal with, e.g. when there's a bug with completions but not for responses, or when responses releases support for something but no change is needed in completions.

breaking up the file will help out with better maintainability and usability and will improve the performance of ai coding assistants.

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/openai/chat.py
+++ b/pydantic_ai_slim/pydantic_ai/models/openai/chat.py
@@ -428,6 +428,7 @@ class OpenAIChatModel(Model):
             raise UnexpectedModelBehavior(f'Invalid response from {self.system} chat completions endpoint: {e}') from e
 
         choice = response.choices[0]
+
         items: list[ModelResponsePart] = []
 
         if thinking_parts := self._process_thinking(choice.message):
@@ -816,7 +817,6 @@ class OpenAIChatModel(Model):
                 elif isinstance(item, DocumentUrl):
                     # OpenAI Chat API's FileFile only supports base64-encoded data, not URLs.
                     # Some providers (e.g., OpenRouter) support URLs via the profile flag.
-                    profile = OpenAIModelProfile.from_profile(self.profile)
                     if not item.force_download and profile.openai_chat_supports_file_urls:
                         content.append(
                             File(
--- a/pydantic_ai_slim/pydantic_ai/models/openai/responses.py
+++ b/pydantic_ai_slim/pydantic_ai/models/openai/responses.py
@@ -64,6 +64,7 @@ from .. import (
     download_item,
     get_user_agent,
 )
+from ._azure import check_azure_content_filter
 from ._shared import (
     MCP_SERVER_TOOL_CONNECTOR_URI_SCHEME,
     OpenAIModelName,
@@ -183,6 +184,11 @@ class OpenAIResponsesModel(Model):
         response = await self._responses_create(
             messages, False, cast(OpenAIResponsesModelSettings, model_settings or {}), model_request_parameters
         )
+
+        # Handle ModelResponse
+        if isinstance(response, ModelResponse):
+            return response
+
         return self._process_response(response, model_request_parameters)
 
     @asynccontextmanager
@@ -253,16 +259,14 @@ class OpenAIResponsesModel(Model):
                             TextPart(
                                 content.text,
                                 id=item.id,
+                                provider_name=self.system,
                                 provider_details=part_provider_details,
                             )
                         )
             elif isinstance(item, responses.ResponseFunctionToolCall):
                 items.append(
                     ToolCallPart(
-                        item.name,
-                        item.arguments,
-                        tool_call_id=item.call_id,
-                        id=item.id,
+                        item.name, item.arguments, tool_call_id=item.call_id, id=item.id, provider_name=self.system
                     )
                 )
             elif isinstance(item, responses.ResponseCodeInterpreterToolCall):
@@ -374,7 +378,7 @@ class OpenAIResponsesModel(Model):
         stream: bool,
         model_settings: OpenAIResponsesModelSettings,
         model_request_parameters: ModelRequestParameters,
-    ) -> responses.Response | AsyncStream[responses.ResponseStreamEvent]:
+    ) -> responses.Response | AsyncStream[responses.ResponseStreamEvent] | ModelResponse:
         tools = (
             self._get_builtin_tools(model_request_parameters)
             + list(model_settings.get('openai_builtin_tools', []))
@@ -476,6 +480,9 @@ class OpenAIResponsesModel(Model):
                 extra_body=model_settings.get('extra_body'),
             )
         except APIStatusError as e:
+            if model_response := check_azure_content_filter(e, self.system, self.model_name):
+                return model_response
+
             if (status_code := e.status_code) >= 400:
                 raise ModelHTTPError(status_code=status_code, model_name=self.model_name, body=e.body) from e
             raise  # pragma: lax no cover
@@ -674,15 +681,10 @@ class OpenAIResponsesModel(Model):
                 file_search_item: responses.ResponseFileSearchToolCallParam | None = None
                 code_interpreter_item: responses.ResponseCodeInterpreterToolCallParam | None = None
                 for item in message.parts:
-                    # Check provider_name for parts that have it
-                    if isinstance(item, (BuiltinToolCallPart, BuiltinToolReturnPart, ThinkingPart, FilePart)):
-                        should_send_item_id = send_item_ids and (
-                            item.provider_name == self.system
-                            or (item.provider_name is None and message.provider_name == self.system)
-                        )
-                    else:
-                        # For parts without provider_name (TextPart, ToolCallPart), check message provider
-                        should_send_item_id = send_item_ids and message.provider_name == self.system
+                    should_send_item_id = send_item_ids and (
+                        item.provider_name == self.system
+                        or (item.provider_name is None and message.provider_name == self.system)
+                    )
 
                     if isinstance(item, TextPart):
                         if item.id and should_send_item_id:
@@ -1003,6 +1005,7 @@ class OpenAIResponsesStreamedResponse(StreamedResponse):
                 raw_finish_reason = (
                     details.reason if (details := chunk.response.incomplete_details) else chunk.response.status
                 )
+
                 if raw_finish_reason:  # pragma: no branch
                     self.provider_details = {**(self.provider_details or {}), 'finish_reason': raw_finish_reason}
                     self.finish_reason = _RESPONSES_FINISH_REASON_MAP.get(raw_finish_reason)
@@ -1024,7 +1027,6 @@ class OpenAIResponsesStreamedResponse(StreamedResponse):
                 maybe_event = self._parts_manager.handle_tool_call_delta(
                     vendor_part_id=chunk.item_id,
                     args=chunk.delta,
-                    provider_name=self.provider_name,
                 )
                 if maybe_event is not None:  # pragma: no branch
                     yield maybe_event
@@ -1046,6 +1048,7 @@ class OpenAIResponsesStreamedResponse(StreamedResponse):
                         args=chunk.item.arguments,
                         tool_call_id=chunk.item.call_id,
                         id=chunk.item.id,
+                        provider_name=self.provider_name,
                     )
                 elif isinstance(chunk.item, responses.ResponseReasoningItem):
                     pass
@@ -1075,7 +1078,6 @@ class OpenAIResponsesStreamedResponse(StreamedResponse):
                     maybe_event = self._parts_manager.handle_tool_call_delta(
                         vendor_part_id=f'{chunk.item.id}-call',
                         args=args_json_delta,
-                        provider_name=self.provider_name,
                     )
                     if maybe_event is not None:  # pragma: no branch
                         yield maybe_event
@@ -1098,7 +1100,6 @@ class OpenAIResponsesStreamedResponse(StreamedResponse):
                     maybe_event = self._parts_manager.handle_tool_call_delta(
                         vendor_part_id=f'{chunk.item.id}-call',
                         args=args_json_delta,
-                        provider_name=self.provider_name,
                     )
                     if maybe_event is not None:  # pragma: no branch
                         yield maybe_event
@@ -1135,7 +1136,6 @@ class OpenAIResponsesStreamedResponse(StreamedResponse):
                     maybe_event = self._parts_manager.handle_tool_call_delta(
                         vendor_part_id=f'{chunk.item.id}-call',
                         args=call_part.args,
-                        provider_name=self.provider_name,
                     )
                     if maybe_event is not None:  # pragma: no branch
                         yield maybe_event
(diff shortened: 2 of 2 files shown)


dsfaccini:
I refactored `test_openai_responses.py` into many files as well because it was 175k tokens long

- tests/models/openai_responses/test_responses_basic.py
- tests/models/openai_responses/test_responses_code_execution.py
- tests/models/openai_responses/test_responses_cot.py
- tests/models/openai_responses/test_responses_file_search.py
- tests/models/openai_responses/test_responses_history.py
- tests/models/openai_responses/test_responses_image_generation.py
- tests/models/openai_responses/test_responses_inputs.py
- tests/models/openai_responses/test_responses_mcp.py
- tests/models/openai_responses/test_responses_output_types.py
- tests/models/openai_responses/test_responses_streaming.py
- tests/models/openai_responses/test_responses_thinking.py
- tests/models/openai_responses/test_responses_tools.py (longest at 50k tokens)
- 
image

Kludex:
Please don't create 10k-line diff PRs.

dsfaccini:
this isn't meant to be reviewed by hand, I just moved the tests that were in `tests/models/test_openai_responses.py` into different files. the only thing that needs to be checked is that we kept all tests, which you can do by running:

```shell
uv run pytest tests/models/test_openai_responses.py --collect-only -q 2>/dev/null | tail -2
```

on main and checking out this PR and running 

```shell
uv run pytest tests/models/openai_responses/ --collect-only -q 2>/dev/null | tail -2
```

and checking they both show the same number (`87 tests collected`)

if you check the `files changed tab` you'll see no cassettes were deleted or anything, just moved
image

pydantic_ai_slim/pydantic_ai/models/openai/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+    'OpenAIResponsesModel',
+    'OpenAIModelSettings',
+    'OpenAIChatModelSettings',
+    'OpenAIResponsesModelSettings',
+    'OpenAIModelName',
+    'OpenAIStreamedResponse',
+    'OpenAIResponsesStreamedResponse',
+    'OpenAISystemPromptRole',
+    'NOT_GIVEN',
+    'OMIT',
```
Kludex: Those don't need to be exposed.
```suggestion
```

pydantic_ai_slim/pydantic_ai/models/openai/chat.py
Kludex, line 1: Given that the class is called `OpenAIChatModel`, I think the name here should be `chat.py`.

pydantic_ai_slim/pydantic_ai/models/openai/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+    OpenAIResponsesModel,
+    OpenAIResponsesStreamedResponse,
+)
+
+try:
+    from openai import NOT_GIVEN, omit
+
+    OMIT = omit
+except ImportError:
+    pass
```
Kludex:
```suggestion
```

pydantic_ai_slim/pydantic_ai/models/openai/_shared.py
```diff
@@ -0,0 +1,343 @@
...
+
+from typing import Any, Literal
+
+from typing_extensions import deprecated
+
+from ... import _utils, usage
+from ...builtin_tools import (
+    ImageAspectRatio,
+    ImageGenerationTool,
+)
```
Kludex:
```suggestion
from ...builtin_tools import ImageAspectRatio, ImageGenerationTool
```

pydantic_ai_slim/pydantic_ai/models/openai/_shared.py
```diff
@@ -0,0 +1,343 @@
...
+    'OpenAIResponsesModelSettings',
+    'MCP_SERVER_TOOL_CONNECTOR_URI_SCHEME',
+    '_OPENAI_ASPECT_RATIO_TO_SIZE',
+    '_OPENAI_IMAGE_SIZE',
+    '_OPENAI_IMAGE_SIZES',
+    '_resolve_openai_image_generation_size',
+    '_make_raw_content_updater',
+    '_map_logprobs',
+    '_map_usage',
+    '_map_provider_details',
```
Kludex: It doesn't make sense to expose private functions.


```suggestion
```

Kludex: Ah, I see that your idea is to use those in the chat and responses implementations.

You can drop the underscore, given that they are already in a private module.

Kludex: You don't need the `__all__` here as well.

pydantic_ai_slim/pydantic_ai/models/openai/responses.py
```diff
@@ -1387,6 +393,7 @@ async def _responses_create(  # noqa: C901
         instructions, openai_messages = await self._map_messages(messages, model_settings, model_request_parameters)
         reasoning = self._get_reasoning(model_settings)
 
+        omit = OMIT()
```
Kludex: Why is this being done?

pydantic_ai_slim/pydantic_ai/models/openai/responses.py
```diff
@@ -1446,24 +453,24 @@ async def _responses_create(  # noqa: C901
                 input=openai_messages,
                 model=self.model_name,
                 instructions=instructions,
-                parallel_tool_calls=model_settings.get('parallel_tool_calls', OMIT),
-                tools=tools or OMIT,
-                tool_choice=tool_choice or OMIT,
-                max_output_tokens=model_settings.get('max_tokens', OMIT),
+                parallel_tool_calls=model_settings.get('parallel_tool_calls', omit),
```
Kludex: Why are those being changed?

Kludex (changes requested):

pydantic_ai_slim/pydantic_ai/models/openai/__init__.py
dsfaccini: there are places using these, and I kept the exports in the `__init__.py` in order to not have to change imports

```
PYRIGHT_PYTHON_IGNORE_WARNINGS=1 uv run pyright 
Uninstalled 2 packages in 2ms
Installed 2 packages in 1ms
/Users/david/projects/forks/refactor/pydantic_ai_slim/pydantic_ai/embeddings/openai.py
  /Users/david/projects/forks/refactor/pydantic_ai_slim/pydantic_ai/embeddings/openai.py:21:43 - error: "OMIT" is unknown import symbol (reportAttributeAccessIssue)
  /Users/david/projects/forks/refactor/pydantic_ai_slim/pydantic_ai/embeddings/openai.py:21:43 - error: Type of "OMIT" is unknown (reportUnknownVariableType)
  /Users/david/projects/forks/refactor/pydantic_ai_slim/pydantic_ai/embeddings/openai.py:128:28 - error: Argument type is partially unknown
    Argument corresponds to parameter "dimensions" in function "create"
    Argument type is "int | Unknown" (reportUnknownArgumentType)
/Users/david/projects/forks/refactor/tests/models/mock_openai.py
  /Users/david/projects/forks/refactor/tests/models/mock_openai.py:20:43 - error: "NOT_GIVEN" is unknown import symbol (reportAttributeAccessIssue)
  /Users/david/projects/forks/refactor/tests/models/mock_openai.py:20:43 - error: Type of "NOT_GIVEN" is unknown (reportUnknownVariableType)
  /Users/david/projects/forks/refactor/tests/models/mock_openai.py:20:54 - error: "OMIT" is unknown import symbol (reportAttributeAccessIssue)
  /Users/david/projects/forks/refactor/tests/models/mock_openai.py:20:54 - error: Type of "OMIT" is unknown (reportUnknownVariableType)
7 errors, 0 warnings, 0 informations 
make: *** [typecheck-pyright] Error 1
```

pydantic_ai_slim/pydantic_ai/models/openai/__init__.py
dsfaccini: see above https://github.com/pydantic/pydantic-ai/pull/3851#discussion_r2653247718

pydantic_ai_slim/pydantic_ai/models/openai/_shared.py
dsfaccini: that's true the `__all__` is unnecessary, I'm leaving the underscores though because otherwise I need to re-add them in `chat.py` and `responses.py` so people can't import them as public functions from there

pydantic_ai_slim/pydantic_ai/models/openai/responses.py
dsfaccini: yeahp that was unnecessary, I moved the import to `_shared.py` and put this one and `chat.py` back to how it was before

pydantic_ai_slim/pydantic_ai/models/openai/__init__.py
Kludex: This was a mistake. Import from `openai` directly, please.

pydantic_ai_slim/pydantic_ai/models/openai/_shared.py
Kludex: > because otherwise I need to re-add them in chat.py and responses.py so people can't import them as public functions from there

If it's defined in a private module, then it's not public.

dsfaccini: but the user will not know that it's a private module if they can:

`from pydantic_ai.models.openai.responses import resolve_openai_image_generation_size`

https://github.com/user-attachments/assets/9b4cd534-9e95-4a78-a73d-37c8348410a0

pydantic_ai_slim/pydantic_ai/models/openai/responses.py
```diff
@@ -56,1051 +50,58 @@
...
 )
+from ._shared import (
+    MCP_SERVER_TOOL_CONNECTOR_URI_SCHEME,
+    OMIT,
+    OpenAIModelName,
+    OpenAIResponsesModelSettings,
+    _make_raw_content_updater,  # pyright: ignore[reportPrivateUsage]
+    _map_logprobs,  # pyright: ignore[reportPrivateUsage]
+    _map_usage,  # pyright: ignore[reportPrivateUsage]
+    resolve_openai_image_generation_size as _resolve_openai_image_generation_size,
```
dsfaccini: apropos private/public functions/modules

pydantic_ai_slim/pydantic_ai/models/openai/_shared.py
DouweM: Yeah it's unfortunate that the IDE wouldn't recognize it as private... We can solve that by using `__all__` in `chat.py` and `responses.py` to only export what we really want to, though, right?

pydantic_ai_slim/pydantic_ai/models/openai/_shared.py
```diff
@@ -0,0 +1,325 @@
...
+
+from typing_extensions import deprecated
+
+from ... import _utils, usage
+from ...builtin_tools import ImageAspectRatio, ImageGenerationTool
+from ...exceptions import UserError
+from ...settings import ModelSettings
+
+try:
+    from openai import omit as OMIT  # pyright: ignore[reportUnusedImport] # noqa: F401
```
DouweM: Shouldn't need this anymore if other files now import this directly from openai

pydantic_ai_slim/pydantic_ai/models/openai/_shared.py
```diff
@@ -0,0 +1,325 @@
...
+    """Whether to include the file search results in the response.
+
+    Corresponds to the `file_search_call.results` value of the `include` parameter in the Responses API.
+    """
+
+
+# Helper functions used by both chat (completions) and responses modules
+
+
+def _make_raw_content_updater(  # pyright: ignore[reportUnusedFunction]
```
DouweM: Drop the underscore please

pydantic_ai_slim/pydantic_ai/models/openai/_shared.py
```diff
@@ -0,0 +1,325 @@
...
+    delta: str, index: int
+) -> Any:  # Returns Callable[[dict[str, Any] | None], dict[str, Any]]
```
DouweM: Then why does it use `Any`? 😄

pydantic_ai_slim/pydantic_ai/models/openai/_shared.py
```diff
@@ -0,0 +1,325 @@
...
+        response_data,
+        provider=provider,
+        provider_url=provider_url,
+        provider_fallback='openai',
+        api_flavor=api_flavor,
+        details=details,
+    )
+
+
+def _map_provider_details(  # pyright: ignore[reportUnusedFunction]
```
DouweM: No private methods in private modules please :)

pydantic_ai_slim/pydantic_ai/models/openai/chat.py
```diff
@@ -0,0 +1,1031 @@
...
+from typing import Any, Literal, cast, overload
+
+from pydantic import ValidationError
+from typing_extensions import assert_never, deprecated
+
+from ... import ModelAPIError, ModelHTTPError, UnexpectedModelBehavior, _utils, usage
+from ..._output import DEFAULT_OUTPUT_TOOL_NAME, OutputObjectDefinition
+from ..._run_context import RunContext
+from ..._thinking_part import split_content_into_text_and_thinking
+from ..._utils import guard_tool_call_id as _guard_tool_call_id, now_utc as _now_utc, number_to_datetime
```
DouweM: If we're going to be better about the "exposing private methods through imports in public modules" thing (which I think is a good idea), we should either give `number_to_datetime` an alias as well (and `split_content_into_text_and_thinking`, and `OutputObjectDefinition`, etc, etc, etc...), _or_ use `__all__` in all public modules. I think the latter will be a lot easier to maintain.

cc @Kludex

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/models/openai/chat.py
Kludex: I don't think neither of the alias are necessary... If they are defined in a private module, they are private. No alias needed. The reimport doesn't matter.

pydantic_ai_slim/pydantic_ai/models/openai/chat.py
```diff
@@ -0,0 +1,1032 @@
...
+        WebSearchOptionsUserLocationApproximate,
+    )
+
+except ImportError as _import_error:
+    raise ImportError(
+        'Please install `openai` to use the OpenAI model, '
+        'you can use the `openai` optional group — `pip install "pydantic-ai-slim[openai]"`'
+    ) from _import_error
+
+__all__ = ['OpenAIChatModel', 'OpenAIModel', 'OpenAIStreamedResponse']
```
dsfaccini, line 112: just to be clear, this is the approach we want, right?

Kludex: To maintain `__all__` on each module is a lot of work, do we really need it? @DouweM

PR #3856, opened 2025-12-26
VoyageAI embeddings support

### LINKED ISSUE #3855: Support for VoyageAI embeddings
### Description

Given we now have embeddings support built-in it would be nice to also support the popular [VoyageAI](https://docs.voyageai.com/docs/embeddings)

### References

_No response_

ggozad:
- Add VoyageAIEmbeddingModel for VoyageAI's embedding API
- Support all current VoyageAI models including domain-specific ones (code, finance, law)
- Add voyageai optional dependency group

- Closes #3855 

### Pre-Review Checklist

- [X] **PR title** is fit for the release changelog.

### Pre-Merge Checklist

- [X] New **tests** for any fix or new behavior, maintaining 100% coverage.
- [X] Updated **documentation** for new features and behaviors, including docstrings for API docs.


pydantic_ai_slim/pydantic_ai/embeddings/__init__.py
```diff
@@ -44,6 +44,13 @@
         'cohere:embed-english-light-v3.0',
         'cohere:embed-multilingual-v3.0',
         'cohere:embed-multilingual-light-v3.0',
+        'voyageai:voyage-3-large',
+        'voyageai:voyage-3.5',
+        'voyageai:voyage-3.5-lite',
+        'voyageai:voyage-code-3',
+        'voyageai:voyage-finance-2',
+        'voyageai:voyage-law-2',
+        'voyageai:voyage-code-2',
```
Kludex, line 60: Is there an API to know which models are available?

ggozad: Not that I know of :( 
According to this [discussion](https://docs.voyageai.com/discuss/66d697827e0adb0018cc9d07) making an api call with a non-existing model name returns also the list of valid models in the error message, but that sounds very fragile.

pydantic_ai_slim/pydantic_ai/embeddings/voyageai.py
```diff
@@ -0,0 +1,184 @@
...
+    All fields from [`EmbeddingSettings`][pydantic_ai.embeddings.EmbeddingSettings] are supported,
+    plus VoyageAI-specific settings prefixed with `voyageai_`.
+    """
+
+    # ALL FIELDS MUST BE `voyageai_` PREFIXED SO YOU CAN MERGE THEM WITH OTHER MODELS.
+
+    voyageai_truncation: bool
+    """Whether to truncate inputs that exceed the model's context length.
+
+    Defaults to True. If False, an error is raised for inputs that are too long.
```
DouweM: With Cohere, I decided to default to False for consistency with OpenAI. Can that be the default here as well? Or do you think that was the wrong call?

pydantic_ai_slim/pydantic_ai/embeddings/voyageai.py
```diff
@@ -0,0 +1,184 @@
...
+    """
+
+    voyageai_output_dtype: Literal['float', 'int8', 'uint8', 'binary', 'ubinary']
+    """The output data type for embeddings.
```
DouweM: Our types currently require embeddings to be floats though

pydantic_ai_slim/pydantic_ai/embeddings/voyageai.py
```diff
@@ -0,0 +1,184 @@
...
+                for available options.
+            api_key: The VoyageAI API key. If not provided, uses the
+                `VOYAGE_API_KEY` environment variable.
+            max_retries: Maximum number of retries for failed requests.
+            timeout: Request timeout in seconds.
+            settings: Model-specific [`EmbeddingSettings`][pydantic_ai.embeddings.EmbeddingSettings]
+                to use as defaults for this model.
+        """
+        self._model_name = model_name
+        self._client = AsyncClient(
```
DouweM: Please do follow the existing provider class / model class pattern for consistency.

pydantic_ai_slim/pydantic_ai/embeddings/voyageai.py
```diff
@@ -0,0 +1,184 @@
...
+        settings = cast(VoyageAIEmbeddingSettings, settings)
+
+        voyageai_input_type = 'document' if input_type == 'document' else 'query'
+
+        try:
+            response = await self._client.embed(
+                texts=list(inputs),
+                model=self.model_name,
+                input_type=voyageai_input_type,
+                truncation=settings.get('voyageai_truncation', True),
```
DouweM: See above, I think this should be `False`

pydantic_ai_slim/pydantic_ai/embeddings/voyageai.py
```diff
@@ -0,0 +1,184 @@
...
+
+    async def max_input_tokens(self) -> int | None:
+        return _MAX_INPUT_TOKENS.get(self.model_name)
+
+
+def _map_usage(total_tokens: int, model: str) -> RequestUsage:
+    usage_data = {'total_tokens': total_tokens}
+    response_data = {'model': model, 'usage': usage_data}
+
+    return RequestUsage.extract(
```
DouweM: We should only use this if the models have data in genai-prices. In this case, it's better to build a `RequestUsage` object manually. As you can see in the tests snapshots, this will actually fail to extract anything.

pydantic_ai_slim/pydantic_ai/embeddings/voyageai.py
ggozad: Hmmm, it's hard to say, I did not notice to be honest in your original PR, I would have raised this. 
I would be inclined to have truncation as default primarily because we do not have usable and accurate tokenizers for all providers. Ollama for example has no tokenizer and truncates silently with no option. There is `tiktoken`, but that really only covers few providers.

Having said that, since that's the default elsewhere let's keep it like that, I will adapt.

pydantic_ai_slim/pydantic_ai/embeddings/voyageai.py
ggozad: Did not notice. Will look thoroughly and adapt.

ggozad:
Hi, with regards to the coverage issue that's blocking CI:

The `sentence-transformers` case in `infer_provider_class()` (`providers/__init__.py`) appears to be dead code that was introduced in PR #3252 but never covered by tests.

- SentenceTransformerEmbeddingModel doesn't use a provider - it runs locally
- infer_embedding_model('sentence-transformers:...') handles this case directly without calling infer_provider_class

I would remove the sentence-transformers case from `infer_provider_class()` and delete `providers/sentence_transformers.py` since it serves no purpose, but this is your decision :)

DouweM:
@ggozad The precedent for a local-model provider that doesn't actually do much is `OutlinesProvider`, which only exists because of the `model_profile` method, and so that `OutlinesModel` can have consistent `__init__` signature with other models. 

So even though it's not strictly necessary, I'd prefer to keep the `SentenceTransformersProvider`. I could imagine us adding `EmbeddingModelProfile` for example, which would then need `SentenceTransformersProvider.model_profile`.

pydantic_ai_slim/pydantic_ai/embeddings/__init__.py
```diff
@@ -70,14 +77,21 @@ def infer_embedding_model(
...
 
-    provider = provider_factory(provider_name)
-
     model_kind = provider_name
     if model_kind.startswith('gateway/'):
         from ..providers.gateway import normalize_gateway_provider
 
         model_kind = normalize_gateway_provider(model_kind)
 
+    # Handle models that don't need a provider first
```
DouweM: If we revert this change, would we get test coverage again?

pydantic_ai_slim/pydantic_ai/embeddings/voyageai.py
```diff
@@ -0,0 +1,169 @@
...
+class VoyageAIEmbeddingSettings(EmbeddingSettings, total=False):
+    """Settings used for a VoyageAI embedding model request.
+
+    All fields from [`EmbeddingSettings`][pydantic_ai.embeddings.EmbeddingSettings] are supported,
+    plus VoyageAI-specific settings prefixed with `voyageai_`.
+    """
+
+    # ALL FIELDS MUST BE `voyageai_` PREFIXED SO YOU CAN MERGE THEM WITH OTHER MODELS.
+
+    voyageai_truncation: bool
```
DouweM: Since multiple models support truncation to be toggled now, let's move this to the `EmbeddingSettings` superclass. We should keep supporting `cohere_truncate` as well, but can prioritize the main `truncate`

tests/test_embeddings.py
```diff
@@ -318,6 +322,63 @@ async def test_embed_error(self, co_api_key: str):
...
+    async def test_documents(self, voyage_api_key: str):
+        model = VoyageAIEmbeddingModel('voyage-3.5', provider=VoyageAIProvider(api_key=voyage_api_key))
+        embedder = Embedder(model)
+        result = await embedder.embed_documents(['hello', 'world'])
+        assert result == snapshot(
+            EmbeddingResult(
+                embeddings=IsList(IsList(IsFloat(), length=1024), length=2),
+                inputs=['hello', 'world'],
+                input_type='document',
+                usage=RequestUsage(),
```
DouweM, line 430: No usage in this case?

pydantic_ai_slim/pydantic_ai/providers/voyageai.py
```diff
@@ -0,0 +1,74 @@
...
+                client to use. If provided, `api_key`, `base_url`, `max_retries`, and `timeout` must be `None`/default.
+            base_url: The base URL for the VoyageAI API. Defaults to `https://api.voyageai.com/v1`.
+            max_retries: Maximum number of retries for failed requests.
+            timeout: Request timeout in seconds.
+        """
+        if voyageai_client is not None:
+            assert api_key is None, 'Cannot provide both `voyageai_client` and `api_key`'
+            assert base_url is None, 'Cannot provide both `voyageai_client` and `base_url`'
+            assert max_retries == 0, 'Cannot provide both `voyageai_client` and `max_retries`'
+            assert timeout is None, 'Cannot provide both `voyageai_client` and `timeout`'
```
DouweM: Unless "most users" will need them, I'd prefer to _not_ expose all the arguments on `AsyncClient` as arguments here: users can just pass their own `voyageai_client` if they want this level of control.

pydantic_ai_slim/pydantic_ai/providers/voyageai.py
```diff
@@ -0,0 +1,74 @@
...
+            api_key = api_key or os.getenv('VOYAGE_API_KEY')
+            if not api_key:
+                raise UserError(
+                    'Set the `VOYAGE_API_KEY` environment variable or pass it via `VoyageAIProvider(api_key=...)` '
+                    'to use the VoyageAI provider.'
+                )
+
+            # Only pass base_url if explicitly set; otherwise use VoyageAI's default
+            base_url = base_url or os.getenv('VOYAGE_BASE_URL')
+            self._client = AsyncClient(
```
DouweM: If this takes a `http_client`, we should use a cached version like we do in the openai provider etc.

pydantic_ai_slim/pydantic_ai/embeddings/voyageai.py
```diff
@@ -0,0 +1,169 @@
...
+    def model_name(self) -> VoyageAIEmbeddingModelName:
+        """The embedding model name."""
+        return self._model_name
+
+    @property
+    def system(self) -> str:
+        """The embedding model provider."""
+        return self._provider.name
+
+    async def embed(
```
DouweM, line 147: It's a shame they don't (seem to) support counting tokens :(

pydantic_ai_slim/pydantic_ai/embeddings/voyageai.py
```diff
@@ -0,0 +1,169 @@
...
+        inputs, settings = self.prepare_embed(inputs, settings)
+        settings = cast(VoyageAIEmbeddingSettings, settings)
+
+        voyageai_input_type = 'document' if input_type == 'document' else 'query'
+
+        try:
+            response = await self._provider.client.embed(
+                texts=list(inputs),
+                model=self.model_name,
+                input_type=voyageai_input_type,
```
DouweM: If we they support more types, we should have a setting for it!

docs/embeddings.md
DouweM, line 1: We need to also update docs/api/embeddings for the API links to work!

pydantic_ai_slim/pydantic_ai/embeddings/__init__.py
ggozad: Yep, you are right, reverted.

pydantic_ai_slim/pydantic_ai/embeddings/voyageai.py
ggozad: I added `truncate` to `EmbeddingSettings` and kept the Cohere settings.

tests/test_embeddings.py
ggozad: The voyageAI API returns total_tokens=0 for document embeddings. I am not sure why this is, most likely a bug on their end.

pydantic_ai_slim/pydantic_ai/providers/voyageai.py
ggozad: The VoyageAI sdk does not support custom HTTP clients :(

pydantic_ai_slim/pydantic_ai/providers/voyageai.py
ggozad: Done, left only `api_key` & `voyageai_client`.

pydantic_ai_slim/pydantic_ai/embeddings/voyageai.py
ggozad: Added a `voyageai_input_type setting` & tests.

pydantic_ai_slim/pydantic_ai/embeddings/settings.py
```diff
@@ -22,6 +22,22 @@ class EmbeddingSettings(TypedDict, total=False):
     * OpenAI
     * Cohere
     * Sentence Transformers
+    * VoyageAI
+    """
+
+    truncate: bool
+    """Whether to truncate inputs that exceed the model's context length.
```
DouweM, line 30: - We should specify that the default is `False`
- I think it's worth explaining that you can use the `max_input_tokens` and `count_tokens` methods to implement your own (smarter) "truncation"

pydantic_ai_slim/pydantic_ai/embeddings/voyageai.py
DouweM: We don't need this field anymore then, right?

pydantic_ai_slim/pydantic_ai/embeddings/voyageai.py
```diff
@@ -0,0 +1,192 @@
...
+
+    # ALL FIELDS MUST BE `voyageai_` PREFIXED SO YOU CAN MERGE THEM WITH OTHER MODELS.
+
+    voyageai_truncation: bool
+    """Whether to truncate inputs that exceed the model's context length.
+
+    Defaults to False. If True, inputs that are too long will be truncated.
+    """
+
+    voyageai_input_type: VoyageAIEmbedInputType
```
DouweM, line 63: Hmm if it only supports `query` and `document` anyway I don't think a setting is warranted. If "direct embedding without prefix" is something users would want to do, I think we should make it `'none'` instead of `None` so that the difference is more clear between this field being omitted (and we should use the default input type implied by the embed_query/document method) and explicitly being set to none.

pydantic_ai_slim/pydantic_ai/providers/voyageai.py
```diff
@@ -0,0 +1,58 @@
...
+
+    @property
+    def base_url(self) -> str:
+        return self._client._params.get('base_url') or 'https://api.voyageai.com/v1'  # type: ignore
+
+    @property
+    def client(self) -> AsyncClient:
+        return self._client
+
+    def __init__(
```
DouweM, line 39: Can we use overloads to ensure either api_key OR voyageai_client is provided, but not both? We have that on other providers as well

pydantic_ai_slim/pydantic_ai/embeddings/__init__.py
DouweM: The error message in the cassette below has a few more names, should we include them all?

tests/test_embeddings.py
```diff
@@ -317,6 +321,145 @@ async def test_embed_error(self, co_api_key: str):
         with pytest.raises(ModelHTTPError, match='not found,'):
             await embedder.embed_query('Hello, world!')
 
+    async def test_query_with_cohere_truncate(self, co_api_key: str):
+        model = CohereEmbeddingModel('embed-v4.0', provider=CohereProvider(api_key=co_api_key))
+        embedder = Embedder(model)
+        result = await embedder.embed_query('Hello, world!', settings={'cohere_truncate': 'END'})  # pyright: ignore[reportArgumentType]
```
DouweM: Why do we need the pyright ignore? Maybe if we use the `CohereEmbeddingSettings` constructor we won't need it

pydantic_ai_slim/pydantic_ai/embeddings/voyageai.py
ggozad: The `None` option according to their docs does "raw" embedding. My guess is that that's useful say if you want to do clustering or classification. For retrieval one would use document or query.
Will change to `none` as you suggested.

tests/test_embeddings.py
ggozad: I added the series 4 models that were released I think yesterday. The other models I saw from series 1 & 2 are legacy and I would not include them.

ggozad:
@DouweM thank you for the thorough review, I think I addressed all your comments, merged the google embeddings changes. I also added the series 4 embedding models that were released yesterday I think.

docs/embeddings.md
```diff
@@ -342,6 +342,61 @@ embedder = Embedder(
...
+
+```python {title="voyageai_settings.py"}
+from pydantic_ai import Embedder
+from pydantic_ai.embeddings.voyageai import VoyageAIEmbeddingSettings
+
+embedder = Embedder(
+    'voyageai:voyage-3.5',
+    settings=VoyageAIEmbeddingSettings(
+        dimensions=512,  # Reduce output dimensions
+        truncate=True,  # Truncate input if it exceeds context length
```
DouweM: This is currently in the "VoyageAI-specific settings" section, but neither of these are actually VoyageAI-specific :) So I think we should mention `truncate` in the top-level `Settings` section (where we mention `dimensions`) already, and change this section to be about `voyageai_input_type`, similar to the one about `google_task_type`

pydantic_ai_slim/pydantic_ai/embeddings/cohere.py
```diff
@@ -159,14 +159,22 @@ async def embed(
...
+            truncate = 'NONE'
+
         try:
             response = await self._client.embed(
                 model=self.model_name,
                 texts=inputs,
                 output_dimension=settings.get('dimensions'),
                 input_type=cohere_input_type,
                 max_tokens=settings.get('cohere_max_tokens'),
-                truncate=settings.get('cohere_truncate', 'NONE'),
```
DouweM, line 169: Let's specify on the `cohere_truncate` docstring that it overrides the `truncate` boolean

pydantic_ai_slim/pydantic_ai/embeddings/voyageai.py
DouweM: Hmm thinking about this more, what do you think about changing our `EmbedInputType` type to accept `None` as well, plus having `Embedder.embed`'s `input_type` argument default to `None`? Then we would not need this custom setting at all anymore. I don't think any of the embeddings APIs require an input type, and if they do we can pick a reasonable default. OpenAI ignores the argument entirely anyway.

Then we would not need a new setting here at all, so even though it's kind of a separate task from this PR, I think it's worth trying it here so we don't introduce the new setting and then immediately deprecate it.

pydantic_ai_slim/pydantic_ai/providers/voyageai.py
```diff
@@ -0,0 +1,65 @@
...
+
+    @property
+    def client(self) -> AsyncClient:
+        return self._client
+
+    @overload
+    def __init__(self, *, voyageai_client: AsyncClient) -> None: ...
+
+    @overload
+    def __init__(self, *, api_key: str | None = None, voyageai_client: None = None) -> None: ...
```
DouweM: This one should just accept the api key right?

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/providers/voyageai.py
ggozad: With regards to the `EmbedInputType`:
I think the `None` choice in VoyageAI is a bit weird. Typically you would want to either embed for a query or documents (when the differentiation is available). The raw embeddings is an edge-case as I mentioned when you want to do something other than semantic search. 
This makes me think that None here is ambiguous, as it does not mean "use default behaviour" (query for most) but rather use no prefixes in the embedding. So if we used None as default should we then introduce a "raw" setting?
Given we have meaningful methods for the common RAG case, i.e. `embed_query()` and `embed_documents()` and that different vendors have a mix of possibilities, I would keep `input_type` as vendor-specific. 
But happy to discuss this, let me know what you think.

DouweM:
@ggozad Thanks Yiorgis!

ggozad:
Thanks @DouweM for the review, hope for more soon!

PR #3858, opened 2025-12-26
Default to native output for groq models that support it

dsfaccini:
motivated by https://pydanticlogfire.slack.com/archives/C083V7PMHHA/p1766580626469929


pydantic_ai_slim/pydantic_ai/profiles/groq.py
```diff
@@ -21,3 +21,19 @@ def groq_model_profile(model_name: str) -> ModelProfile:
...
+
+    GPT-OSS models (gpt-oss-20b, gpt-oss-120b) support strict native structured output
+    with 100% schema adherence.
+    """
+    from .openai import OpenAIJsonSchemaTransformer
+
+    return GroqModelProfile(
+        supports_json_schema_output=True,
+        supports_json_object_output=True,
+        default_structured_output_mode='native',
```
DouweM: On Slack, Rami [pointed out](https://pydanticlogfire.slack.com/archives/C083V7PMHHA/p1766592828406369?thread_ts=1766580626.469929&cid=C083V7PMHHA):

> Quick test results: Tool mode is actually the best option for GPT-OSS + tools, but still has ~10% failure rate. Native mode can't work because Groq explicitly rejects JSON mode + tool calling together. Prompted mode confuses the model into trying to call a "json" tool. (edited) 

So we should make sure we only use native mode when there are no function tools. 

Similar to how we dynamically determine the structured output mode based on various factors in (i believe) google and anthropic

pydantic_ai_slim/pydantic_ai/profiles/groq.py
```diff
@@ -21,3 +21,19 @@ def groq_model_profile(model_name: str) -> ModelProfile:
     return GroqModelProfile(
         groq_always_has_web_search_builtin_tool=model_name.startswith('compound-'),
     )
+
+
+def groq_gpt_oss_model_profile(model_name: str) -> ModelProfile:
```
DouweM: `profiles/groq.py` is for models _by groq_ (like compound). this function should go into the provider.

pydantic_ai_slim/pydantic_ai/profiles/groq.py
```diff
@@ -21,3 +21,19 @@ def groq_model_profile(model_name: str) -> ModelProfile:
...
+    """Get profile for OpenAI GPT-OSS models on Groq.
+
+    GPT-OSS models (gpt-oss-20b, gpt-oss-120b) support strict native structured output
+    with 100% schema adherence.
+    """
+    from .openai import OpenAIJsonSchemaTransformer
```
DouweM: No inline imports please

pydantic_ai_slim/pydantic_ai/providers/groq.py
```diff
@@ -29,16 +29,29 @@
...
-    return ModelProfile(supports_json_object_output=True, supports_json_schema_output=True).update(
-        moonshotai_model_profile(model_name)
-    )
+    from pydantic_ai.profiles.openai import OpenAIJsonSchemaTransformer
+
+    return ModelProfile(
+        supports_json_object_output=True,
+        supports_json_schema_output=True,
+        default_structured_output_mode='native',
+        json_schema_transformer=OpenAIJsonSchemaTransformer,
```
DouweM: did we verify we need to do these 2 lines?

DouweM: - i'd rather stick to just the case where we confirmed native works better than tool
- if it needs OpenAIJsonSchemaTransformer, should we set that inside the provider.model_profile method?

pydantic_ai_slim/pydantic_ai/providers/groq.py
```diff
@@ -29,16 +29,29 @@
...
     if model_name in {'llama-4-maverick-17b-128e-instruct', 'llama-4-scout-17b-16e-instruct'}:
-        return ModelProfile(supports_json_object_output=True, supports_json_schema_output=True).update(
-            meta_model_profile(model_name)
+        from pydantic_ai.profiles.openai import OpenAIJsonSchemaTransformer
+
+        base = meta_model_profile(model_name) or ModelProfile()
+        return base.update(
+            ModelProfile(
+                supports_json_object_output=True,
+                supports_json_schema_output=True,
```
DouweM: the else branch doesn't need these changes?

DouweM: if all models now support these 2 things, should they be in provider.model_profile?

tests/models/test_groq.py
```diff
@@ -5733,3 +5733,74 @@ class CityLocation(BaseModel):
...
+        country: str
+
+    # No explicit NativeOutput/ToolOutput wrapper - uses profile default
+    agent = Agent(m, output_type=CityLocation)
+
+    result = await agent.run('What is the capital of France?')
+    assert result.output == snapshot(CityLocation(city='Paris', country='France'))
+
+    # Verify it used native output (TextPart with JSON) not tool calling (ToolCallPart)
+    response = result.all_messages()[-1]
```
DouweM: Use `result.response`

tests/models/test_groq.py
```diff
@@ -5733,3 +5733,74 @@ class CityLocation(BaseModel):
...
+    assert isinstance(response, ModelResponse)
+    # Native output returns TextPart with JSON content, not ToolCallPart
+    part_types = [type(part).__name__ for part in response.parts]
+    assert 'ToolCallPart' not in part_types, f'Expected native output but got tool call: {part_types}'
+    assert 'TextPart' in part_types, f'Expected TextPart for native output: {part_types}'
```
DouweM: checking string names is ugly, why not use isinstance?

DouweM (changes requested):

tests/models/test_groq.py
dsfaccini: should we have a pre-commit check for this?

tests/models/test_groq.py
dsfaccini: it's a bit more verbose but yeah I agree

pydantic_ai_slim/pydantic_ai/profiles/groq.py
dsfaccini: should we add this to a `profiles/CLAUDE.md` ?

pydantic_ai_slim/pydantic_ai/providers/groq.py
dsfaccini: I think we should because all of them are using it

Edit: actually I take that back: here it's formatted nicely, it's ugly and verbose in the `model_profile`

DouweM: I don't really understand your comment 😄 I wasn't talking about the formatting, but about these fields being set at all

tests/models/test_groq.py
DouweM: I'd rather put it in CLAUDE.md, I've made a note of that.

pydantic_ai_slim/pydantic_ai/providers/groq.py
dsfaccini: yeahp I created a constant to use for all

pydantic_ai_slim/pydantic_ai/providers/groq.py
dsfaccini: yeahp, moved to centralized constant and yes, verified they all support this

I'm not sure how to verify that the one would work better than the other.... other than running benchmarks

pydantic_ai_slim/pydantic_ai/providers/groq.py
```diff
@@ -69,6 +87,7 @@ def model_profile(self, model_name: str) -> ModelProfile | None:
             'mistral': mistral_model_profile,
             'moonshotai/': groq_moonshotai_model_profile,
             'compound-': groq_model_profile,
+            'openai/gpt-oss': groq_gpt_oss_model_profile,
```
DouweM: See the special logic for `endswith('/')` below; we now end up with `model_name` including the `openai/` bit, while we still should be stripping it. So we may need to use `'/' in ...` and `split`

pydantic_ai_slim/pydantic_ai/providers/groq.py
```diff
@@ -27,23 +27,41 @@
...
         return meta_model_profile(model_name)
 
 
+def groq_gpt_oss_model_profile(model_name: str) -> ModelProfile:
+    """Get profile for OpenAI GPT-OSS models on Groq.
+
+    GPT-OSS models (gpt-oss-20b, gpt-oss-120b) support strict native structured output
+    with 100% schema adherence.
+    """
+    return GroqModelProfile.from_profile(_GROQ_NATIVE_OUTPUT_PROFILE)
```
DouweM: Should we not use `harmony_model_profile` for gpt-oss?

pydantic_ai_slim/pydantic_ai/models/groq.py
```diff
@@ -177,6 +177,20 @@ def supported_builtin_tools(cls) -> frozenset[type[AbstractBuiltinTool]]:
...
         return frozenset({WebSearchTool})
 
+    def prepare_request(
+        self, model_settings: ModelSettings | None, model_request_parameters: ModelRequestParameters
+    ) -> tuple[ModelSettings | None, ModelRequestParameters]:
+        # Groq doesn't support native structured output with function tools
+        # Fall back to tool-based output when function tools are present
+        # This must happen BEFORE super().prepare_request() because the base class
+        # clears output_tools when output_mode != 'tool'
+        if model_request_parameters.function_tools and model_request_parameters.output_mode in ('native', 'auto'):
```
DouweM: If `native` was set intentionally, we should not  ignore it.

See how we handle this for Anthropic:

https://github.com/pydantic/pydantic-ai/blob/71b5027e1738b6c186127909d55c95d86d13dc2d/pydantic_ai_slim/pydantic_ai/models/anthropic.py#L346-L356

DouweM: And Google, also similar:

https://github.com/pydantic/pydantic-ai/blob/71b5027e1738b6c186127909d55c95d86d13dc2d/pydantic_ai_slim/pydantic_ai/models/google.py#L264-L272

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/providers/groq.py
dsfaccini: yeah

pydantic_ai_slim/pydantic_ai/providers/groq.py
```diff
@@ -27,23 +28,42 @@
...
+Groq's API requires OpenAI-compatible JSON schema format with strict mode
+constraints (additionalProperties: false on all objects) for structured output."""
+
+
 def groq_moonshotai_model_profile(model_name: str) -> ModelProfile | None:
     """Get the model profile for an MoonshotAI model used with the Groq provider."""
-    return ModelProfile(supports_json_object_output=True, supports_json_schema_output=True).update(
-        moonshotai_model_profile(model_name)
-    )
+    return _GROQ_NATIVE_OUTPUT_PROFILE.update(moonshotai_model_profile(model_name))
```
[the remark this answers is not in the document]
DouweM: @dsfaccini I think this is worth doing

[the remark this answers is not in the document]
dsfaccini: done!

pydantic_ai_slim/pydantic_ai/providers/groq.py
```diff
@@ -69,15 +90,19 @@ def model_profile(self, model_name: str) -> ModelProfile | None:
...
-                if prefix.endswith('/'):
-                    model_name = model_name[len(prefix) :]
-                return profile_func(model_name)
+            model_name_lower = model_name.lower()
+            if model_name_lower.startswith(prefix):
+                # Strip provider prefix (e.g., 'openai/gpt-oss-120b' -> 'gpt-oss-120b')
+                if '/' in model_name_lower:
+                    model_name_for_profile = model_name_lower.split('/', 1)[-1]
+                else:
+                    model_name_for_profile = model_name_lower
```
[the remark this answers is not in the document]
dsfaccini: this doesn't make sense to me, we had a bug before, now it's fixed... with this change we'll end up passing `qwen/qwen-3-32b` to `qwen_model_profile` which doesn't know what to do with that

I'm I cocos @DouweM ?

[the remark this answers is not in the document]
DouweM: @dsfaccini Devin is right that it changes behavior, but you're right that that change is desirable!

pydantic_ai_slim/pydantic_ai/providers/groq.py
```diff
@@ -69,15 +90,19 @@ def model_profile(self, model_name: str) -> ModelProfile | None:
             'mistral': mistral_model_profile,
             'moonshotai/': groq_moonshotai_model_profile,
             'compound-': groq_model_profile,
+            'openai/gpt-oss-': groq_gpt_oss_model_profile,
```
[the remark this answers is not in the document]
DouweM: @dsfaccini I guess it's complaining about this again because there was no reason stated for ignoring it on the last one. I'm fin keeping it like this,

pydantic_ai_slim/pydantic_ai/providers/groq.py
```diff
@@ -27,23 +28,43 @@
     ) from _import_error
 
 
+_GROQ_NATIVE_OUTPUT_PROFILE = ModelProfile(
+    supports_json_object_output=True,
+    supports_json_schema_output=True,
+    default_structured_output_mode='native',
+    json_schema_transformer=OpenAIJsonSchemaTransformer,
+)
```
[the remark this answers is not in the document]
DouweM: If David confirmed that works, it's OK by me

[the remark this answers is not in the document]
dsfaccini: yeah I just recorded some cassettes to verify

tests/models/test_groq.py
```diff
@@ -5748,3 +5751,109 @@ class CityLocation(BaseModel):
...
+
+    result = await agent.run('What is the capital of France?')
+    assert result.output == snapshot(CityLocation(city='Paris', country='France'))
+
+    # Verify it used tool output (ToolCallPart) not native output
+    response = result.response
+    assert isinstance(response, ModelResponse)
+    assert any(isinstance(part, ToolCallPart) for part in response.parts), (
+        f'Expected tool output fallback when native + tools: {response.parts}'
+    )
```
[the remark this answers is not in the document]
dsfaccini: no!

pydantic_ai_slim/pydantic_ai/models/groq.py
```diff
@@ -177,6 +177,24 @@ def supported_builtin_tools(cls) -> frozenset[type[AbstractBuiltinTool]]:
...
+        # clears output_tools when output_mode != 'tool'.
+        if model_request_parameters.function_tools:
+            if model_request_parameters.output_mode == 'native':
+                raise UserError(
+                    'Groq does not support native structured output (JSON mode) with function tools. '
+                    'Use `output_type=ToolOutput(...)` instead.'
+                )
+            elif model_request_parameters.output_mode == 'auto':
+                if self.profile.default_structured_output_mode == 'native':
+                    model_request_parameters = replace(model_request_parameters, output_mode='tool')
```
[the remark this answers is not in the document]
DouweM: That's an interesting catch that may exist in more places...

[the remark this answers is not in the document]
dsfaccini: yeah I started addressing it in anthropic and google as well, the bug exists but seems to get luckily suppressed by further logic in agent graph, but it's one of those instances where I'd prefer we didn't have so many layers upon layers of overlapping checks

tests/models/test_groq.py
```diff
@@ -5748,3 +5751,129 @@ class CityLocation(BaseModel):
...
+
+    # Verify it used native output (TextPart with JSON) not tool calling (ToolCallPart)
+    response = result.response
+    assert isinstance(response, ModelResponse)
+    assert not any(isinstance(part, ToolCallPart) for part in response.parts), (
+        f'Expected native output but got tool call: {response.parts}'
+    )
+    assert any(isinstance(part, TextPart) for part in response.parts), (
+        f'Expected TextPart for native output: {response.parts}'
+    )
```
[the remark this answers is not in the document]
dsfaccini: consciously decided against this

pydantic_ai_slim/pydantic_ai/providers/groq.py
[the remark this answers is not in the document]
dsfaccini: Claude here: This was confirmed and approved by DouweM. Native output works for all three model families (GPT-OSS, MoonshotAI, Meta Llama 4) on Groq.

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -676,12 +684,7 @@ def prepare_request(  # noqa: C901
             )
 
         if params.output_mode == 'auto':
-            output_mode = self.profile.default_structured_output_mode
-            params = replace(
-                params,
-                output_mode=output_mode,
-                allow_text_output=output_mode in ('native', 'prompted'),
-            )
+            params = params.resolve_auto_output_mode(self.profile.default_structured_output_mode)
```
DouweM: Can we move the `if == auto` check into this method? Because the `allow_text_output=output_mode in (native, prompted)` check is only valid if it was `auto` before.

pydantic_ai_slim/pydantic_ai/models/groq.py
```diff
@@ -197,6 +197,24 @@ def supported_builtin_tools(cls) -> frozenset[type[AbstractBuiltinTool]]:
...
+    def prepare_request(
+        self, model_settings: ModelSettings | None, model_request_parameters: ModelRequestParameters
+    ) -> tuple[ModelSettings | None, ModelRequestParameters]:
+        # Groq doesn't support native structured output with function tools.
+        # This must happen BEFORE super().prepare_request() because the base class
+        # clears output_tools when output_mode != 'tool'.
+        if model_request_parameters.function_tools:
+            if model_request_parameters.output_mode == 'native':
+                raise UserError(
+                    'Groq does not support native structured output (JSON mode) with function tools. '
```
DouweM, line 211: Really? Is there a doc we can link to as evidence?

pydantic_ai_slim/pydantic_ai/models/__init__.py
dsfaccini: Claude here: Resolved by merging upstream main which includes #4274 — renames to `with_default_output_mode()` with the auto guard inside the method, exactly as suggested.

pydantic_ai_slim/pydantic_ai/models/groq.py
dsfaccini: Claude here: Added doc link. From [Groq Structured Outputs docs](https://console.groq.com/docs/structured-outputs): "Streaming and tool use are not currently supported with Structured Outputs." The API returns `json mode cannot be combined with tool/function calling` (400 error) when both are sent together.

dsfaccini: added as a code comment https://console.groq.com/docs/structured-outputs
image

pydantic_ai_slim/pydantic_ai/providers/groq.py
```diff
@@ -70,15 +94,20 @@ def model_profile(model_name: str) -> ModelProfile | None:
...
-                if prefix.endswith('/'):
-                    model_name = model_name[len(prefix) :]
-                return profile_func(model_name)
+            if model_name_lower.startswith(prefix):
+                # Strip provider prefix (e.g., 'openai/gpt-oss-120b' -> 'gpt-oss-120b')
+                if '/' in model_name_lower:
+                    model_name_for_profile = model_name_lower.split('/', 1)[-1]
+                else:
+                    model_name_for_profile = model_name_lower
+                return profile_func(model_name_for_profile)
```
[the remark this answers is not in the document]
dsfaccini: Claude here: This was reviewed and approved by DouweM — the behavior change is intentional. Current compound models (`compound-beta`, `compound-beta-mini`) have no slashes, so the hypothetical `compound-search/v2` case does not apply. For `qwen/qwen-3-32b`, the stripped name is arguably more correct since profile functions should see the model name without the vendor prefix.

pydantic_ai_slim/pydantic_ai/providers/groq.py
```diff
@@ -27,23 +28,46 @@
...
+produce tool-based structured output on Groq."""
+
+
+def groq_moonshotai_model_profile(model_name: str) -> ModelProfile:
     """Get the model profile for an MoonshotAI model used with the Groq provider."""
-    return ModelProfile(supports_json_object_output=True, supports_json_schema_output=True).update(
-        moonshotai_model_profile(model_name)
-    )
+    base = moonshotai_model_profile(model_name) or ModelProfile()
+    return base.update(_GROQ_NATIVE_OUTPUT_PROFILE)
```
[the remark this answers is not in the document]
dsfaccini: Claude here: Verified — kimi-k2 does support native structured output on Groq (we have a recorded cassette with a successful `response_format: json_schema` request). The Groq docs simply haven't listed it yet. Since there's only one MoonshotAI model on Groq currently, the simplification is intentional. If Groq adds non-native-capable MoonshotAI models later, we'd add a guard like `meta_groq_model_profile` has.

PR #3862, opened 2025-12-27
tests: monkey-patch httpx clients

Kludex:


Danipulok:
Wow, I don't really think I deserve to be here, but thank you very much for marking me as co-author, @Kludex!

dsfaccini:
awesome! I removed my fix in #3778 in favour of this, thank you guys

Kludex:
> Wow, I don't really think I deserve to be here, but thank you very much for marking me as co-author, @Kludex!

The following line is copied from your code:

```py
    mocker.patch('pydantic_ai.providers.google_vertex.google.auth.default', return_value=return_value)
```

So... It's fair to put your name. :)

Danipulok:
@Kludex, then it sounds nice, thank you very much again!
Appreciate it!

PR #3826, opened 2025-12-23
Send multimodal tool results to APIs directly instead of splitting into user parts

dsfaccini:
Closes #3253
Closes #3902

## Problem

When a tool returns multimodal content (images, documents, etc.), PydanticAI currently:
1. Puts a text placeholder (`"See file {identifier}"`) in `ToolReturnPart.content`
2. Appends the actual multimodal content as a separate `UserPromptPart`

This breaks Bedrock with Anthropic models because they require tool results to be immediately followed by the next message - splitting multimodal content into a separate user part causes a `ValidationException`:
```
tool_use ids were found without tool_result blocks immediately following them
```

## Solution

Keep multimodal content directly in `ToolReturnPart.content` and let each model implementation decide how to handle it:
- **Anthropic/Bedrock/Google**: Pass multimodal directly in tool result (APIs support it)
- **OpenAI Chat API**: Split at model layer (emit tool message, then user message with multimodal)

This approach:
- Fixes Bedrock/Anthropic compatibility
- Keeps the message history cleaner (no placeholder text)
- Lets each provider optimize for their API capabilities

## Changes

- `pydantic_ai/_agent_graph.py` - Stop splitting multimodal in `_call_tool()`, keep content directly in `ToolReturnPart`
- `pydantic_ai/messages.py` - Add `model_response_parts()` helper to `BaseToolReturnPart` for extracting content parts
- `pydantic_ai/models/anthropic.py` - Handle multimodal in `ToolReturnPart` using content blocks array
- `pydantic_ai/models/bedrock.py` - Place multimodal inside `toolResult.content` array
- `pydantic_ai/models/google.py` - Extract `_map_tool_return()` method, emit `function_response` + multimodal parts
- `pydantic_ai/models/openai.py` - Split at model layer for Chat API (tool message + user message)
- `tests/test_agent.py` - Update test expectations for new message structure

## Before/After

**Before** (parallel tool calls returning images):
```python
ModelRequest([
    ToolReturnPart(tool_name='get_image_1', content='See file image1.png'),
    ToolReturnPart(tool_name='get_image_2', content='See file image2.png'),
    UserPromptPart(content=['This is file image1.png:', ImageUrl(...)]),
    UserPromptPart(content=['This is file image2.png:', ImageUrl(...)]),
])
```

**After**:
```python
ModelRequest([
    ToolReturnPart(tool_name='get_image_1', content=[ImageUrl(...)]),
[...]

PR #3865, opened 2025-12-27
Adds `return_schema` for tools.

### LINKED ISSUE #3122: Also include return (type) json schema for tools (similar how tool parameters are handled)
### Description

**Feature Request: Include Tool Return Type Schema in Tool Definitions (passed into LLM call as part of `tools` json)**

**Background**
We use pydantic-ai as an agent framework in a complex business domain. It's critical for us that in addition to the tool `parameters` schema and description, the **schema and description of tool return types** are also included in the JSON passed to the LLM as part of tool definitions.

**Current Problem:** The LLM doesn't know which properties/fields are available in a tool's return value without actually executing the tool call. This leads to:
- Unnecessary tool calls just to discover what data is available
- The LLM being unable to plan multi-step operations effectively
- Confusion about which tool provides which data points

**Use Cases**
1. **Rich DTOs as Tool Returns**
All of our tools return DTOs (Pydantic models) instead of plain strings or primitives. These models contain detailed information about domain entities like invoices, projects, and contracts. 

2. **Sequential (Chained) Tool Calls**
This is foundational for the sequential tool call improvements discussed in #2037. If the LLM knows what data a tool returns, it can:
   - Plan a sequence of tool calls where one tool's output feeds into another
   - Determine upfront if a requested data point is available
   - Chain operations more intelligently (e.g., "get invoice details, then get the contract details for that invoice")
 
3. **Field Descriptions in Complex Domains**
In complex business domains, field names alone are often insufficient for the LLM to understand what data represents. For example:
   - `audited_amount` vs `approved_amount` vs `paid_amount` on an invoice
   - `date_received` vs `date_posted` vs `date_due`
   - Domain-specific terminology that needs context

   By including field descriptions from Pydantic models in the return schema, the LLM can:
     - Understand the semantic meaning of each field
     - Provide more accurate answers about what data is available
    - Better explain differences between similar-sounding fields to users

4. **Better User Experience**
With return schemas available, the LLM can:
   - Politely decline requests for data that isn't available in any tool return
   - Suggest alternative data points that are available
   - Be more precise about what information it can provide

**Proposed Solution**
Similar to how `parameters` work today, each tool definition would include a `returns` field containing the JSON schema of the return type.

**How this could be implemented**
1. **Auto-generate from type annotations**: Since tools must return JSON-serializable data, we can extract the return type from function signatures and generate the schema
2. **Docstring parsing**: Parse return type descriptions from docstrings (similar to current parameter handling)
3. **Opt-in flag**: To avoid breaking changes and manage context length, introduce an opt-in flag (e.g., `include_return_schema=True`) in `ToolDefinition` or during tool preparation

I would be curious if you already thought about this and have a preferred way of how this should be implemented in mind.

**Example: Current vs. Proposed**

**Current tool definition:**
```json
{
    "type": "function",
    "function": {
        "name": "get_contract_details",
        "description": "Get details about a specific contract by ID.",
        "parameters": {
            "properties": {
                "contract_id": {
                    "type": "integer",
                    "description": "ID of the contract"
                }
            },
            "required": ["contract_id"],
            "type": "object",
        },
        "strict": true
    }
}
```

**Proposed with return schema:**
```json
{
    "type": "function",
    "function": {
        "name": "get_contract_details",
        "description": "Get details about a specific contract by ID.",
        "parameters": {
            "properties": {
                "contract_id": {
                    "type": "integer"
                }
            },
            "required": ["contract_id"],
            "type": "object"
        },
        "strict": true,
        "returns": {
            // Either with 'properties' or using #/$defs/... which is used for parameters
            "type": "object",
            "description": "Details about a contract within a project",
            "properties": {
                "id": {
                    "type": "integer",
                    "description": "Unique contract identifier"
                },
                "project_id": {
                    "type": "integer",
                    "description": "ID of the parent project"
                },
                "contractor_name": {
                    "type": "string",
                    "description": "Name of the contractor company"
                },
                "amount": {
                    "type": "number",
                    "description": "Total contract amount"
                },
                "signed_date": {
                    "type": "string",
                    "format": "date",
                    "description": "Date the contract was signed"
                }
            },
            "required": ["id", "project_id", "contractor_name", "amount", "signed_date"]
        }
    }
}
```

**Minimal Example**

Here's a minimal example demonstrating the need, especially for complex domain models with field descriptions:

```python
from datetime import date
from decimal import Decimal
import pydantic

class InvoiceDetails(pydantic.BaseModel):
    """Details about an invoice for a contract"""
    id: int = pydantic.Field(description="Unique invoice identifier")
    contract_id: int = pydantic.Field(description="ID of the associated contract")
    invoice_number: str = pydantic.Field(description="External invoice number")
    
    # Multiple amount fields - descriptions are crucial for LLM to understand differences
    initial_amount: Decimal = pydantic.Field(description="Initial invoice amount as submitted by contractor")
    audited_amount: Decimal = pydantic.Field(description="Amount verified after checking the invoice")
    approved_amount: Decimal = pydantic.Field(description="Amount released for payment after approval")
    paid_amount: Decimal = pydantic.Field(description="Amount actually paid out")
    
    # Multiple date fields - descriptions clarify the workflow
    date_received: date = pydantic.Field(description="Date invoice was received")
    date_posted: date = pydantic.Field(description="Date invoice was posted in accounting")
    status: str = pydantic.Field(description="Invoice status (new, approved, paid)")

def get_invoice_details(ctx: RunContext, invoice_id: int) -> InvoiceDetails:
    """Get details about a specific invoice by ID."""
    return InvoiceDetails(
        id=invoice_id,
        contract_id=1,
        invoice_number="INV-2024-001",
        initial_amount=Decimal("50000.00"),
        audited_amount=Decimal("48000.00"),
        approved_amount=Decimal("48000.00"),
        paid_amount=Decimal("48000.00"),
        date_received=date(2024, 3, 15),
        date_posted=date(2024, 3, 20),
        status="approved",
    )

agent = Agent("openai:gpt-4o", tools=[get_invoice_details])

# Without return schema: 
# - LLM doesn't know what fields InvoiceDetails contains
# - Can't distinguish between initial_amount, audited_amount, approved_amount, paid_amount
# - User asks "what's the invoice amount?" - which one do they mean?

# With return schema + field descriptions:
# - LLM knows exactly what fields are available
# - Understands the difference between amounts (initial vs audited vs approved vs paid)
# - Can ask clarifying questions or provide the most relevant amount based on context
```

**Related Issues**
- #2037 - Sequential tool calls (this feature would be foundational for that)

**Open Questions**
1. **Opt-in vs. opt-out**: Should this be enabled by default or require explicit configuration?
2. **Context length concerns**: Should there be automatic truncation/summarization for very large return schemas?
3. **Docstring parsing**: How should we handle return type descriptions from docstrings?

I'm happy to contribute to implementing this feature if there's interest! Let me know if you have preferences on the approach.

---

**Note:** This feature would significantly improve pydantic-ai's capabilities for complex, multi-step agentic workflows in production environments, especially in domains where field descriptions are essential for LLM comprehension.

### References

_No response_

adtyavrdhn:
- Closes #3122 

### Pre-Review Checklist

### Pre-Merge Checklist


adtyavrdhn:
I have to figure out a couple of things still, unclear about the deferred tools and how I want to optimize the token usage as well using $refs. It might take a while :)

adtyavrdhn:
We can take up the other suggestion to tack it onto the result content in another PR to keep this one short.

docs/toolsets.md
```diff
@@ -383,6 +386,88 @@ print(result.output)
 
 _(This example is complete, it can be run "as is")_
 
+### Including Return Schemas {#return-schema-toolset}
+
+[`ReturnSchemaToolset`][pydantic_ai.toolsets.ReturnSchemaToolset] wraps a toolset and appends each tool's return schema to its description. This helps LLMs understand what data a tool returns, enabling better planning for multi-step operations and tool chaining.
+
+When tools return complex types like Pydantic models or dataclasses, PydanticAI automatically infers the return schema from the function's return type annotation. By including this schema in the tool description, the LLM can:
```
dsfaccini: `Pydantic AI` with a space

docs/toolsets.md
```diff
@@ -383,6 +386,88 @@ print(result.output)
...
+                'name': {'type': 'string'},
+                'email': {'type': 'string'},
+            },
+            'required': ['id', 'name', 'email'],
+            'title': 'UserDetails',
+            'type': 'object',
+        },
+    )
+]
+"""
```
dsfaccini: we can remove the output from the code example, overall not sure we want to have much of a section on this but let's keep it for now, I'm just commenting on the surface stuff since the PR is still in draft

docs/toolsets.md
```diff
@@ -383,6 +386,88 @@ print(result.output)
...
+```
+
+1. We're using [`TestModel`][pydantic_ai.models.test.TestModel] here because it makes it easy to see which tools were available on each run.
```
dsfaccini: the clarification about testmodel we can remove methinks

pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -844,6 +855,7 @@ def _run_span_end_attributes(
             ),
         }
 
+    # TODO: Consider adding include_tool_return_schema to override() for testing scenarios
```
dsfaccini: overall not a good practice to add fields or variables just for testing's sake so lets not do this

pydantic_ai_slim/pydantic_ai/toolsets/abstract.py
```diff
@@ -189,3 +192,19 @@ def approval_required(
...
+        planning for multi-step operations and tool chaining.
+
+        Args:
+            include: Whether to include return schemas in tool descriptions. Defaults to True.
+
+        See [toolset docs](../toolsets.md#return-schema-toolset) for more information.
+        """
+        from .return_schema import ReturnSchemaToolset
+
+        return ReturnSchemaToolset(self, include_return_schema=include)
```
dsfaccini: I haven't read the linked issue so I won't comment on the code while the PR remains a draft

docs/toolsets.md
adtyavrdhn: Yeah that's not needed, need to clean up the docs.

pydantic_ai_slim/pydantic_ai/agent/__init__.py
adtyavrdhn: Lol that was a brain fade, I will not add this, I decided against it to add to run_* and override_* methods. I'll remove this TODO as well

pydantic_ai_slim/pydantic_ai/toolsets/abstract.py
adtyavrdhn: I'll clean it up and request you for a review when I have the stuff figured out :))

adtyavrdhn:
About `ToolReturn` I am skeptical of adding `str | Sequence[UserContent]` into the json schema when that information is sent via a `UserMessage` and not the Tool result itself. It feels semantically wrong, instead of latching that information onto the schema should we just explain it in the description that a message with `str | MultiModalContent` might follow this result?

The json schema for `UserContent` also feels bloated and I am not sure how useful that information will be for tool chaining which should we our first goal as soon as we wrap this up. 

Would appreciate you taking a look now @dsfaccini :))

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -593,6 +594,7 @@ def tool_for_tool_def(self, tool_def: ToolDefinition) -> ToolsetTool[Any]:
             tool_def=tool_def,
             max_retries=self.max_retries,
             args_validator=TOOL_SCHEMA_VALIDATOR,
+            include_return_schema=True,  # Setting it to True for MCP tools by default because it is a part of the MCP spec.
```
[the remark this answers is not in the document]
dsfaccini: on a second look, if the `ToolDefinition` has its own `include_return_schema` specified we should pass that along. you can still add the `TODO v2 (potentially set default to True to follow the MCP spec)`

pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -90,6 +92,7 @@
     'WrapperAgent',
     'AbstractAgent',
     'EventStreamHandler',
+    'ToolDefinition',
```
dsfaccini: we don't need this here

DouweM:
(For context to those reading along, Aditya and I discussed this PR on a call, so this is pending his addressing the updates we discussed then!)

adtyavrdhn:
@greptile

adtyavrdhn:
Yeah it doesn't take into account any of the important details of the issue at all, might as well just git diff and dump it into Claude. Not sure how much better it will get if we teach it but I would rather document that stuff in the repo so that we can swap out the providers or build our own. I would have liked it to discuss the implications of doing this and if not that then just talk about thinking through what these different schemas could be and how that would look like when we dump it into tool descriptions.

pydantic_ai_slim/pydantic_ai/_function_schema.py
```diff
@@ -212,6 +218,26 @@ def function_schema(  # noqa: C901
...
         description = json_schema.pop('description', None)
 
+    return_annotation = type_hints.get('return')
+    return_schema = None
+    # Check if origin of return annotation is ToolReturn which needs special handling
+    if get_origin(return_annotation) is ToolReturn or return_annotation is ToolReturn:
+        type_args = get_args(return_annotation)
+        inner_type = type_args[0] if type_args else Any
+        try:
+            return_schema = TypeAdapter(inner_type).json_schema(schema_generator=schema_generator, mode='serialization')
```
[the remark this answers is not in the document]
adtyavrdhn: yeah I missed that, fixed

pydantic_ai_slim/pydantic_ai/toolsets/return_schema.py
```diff
@@ -0,0 +1,54 @@
...
+    Defaults to True since the purpose of this wrapper is to include return schemas.
+    Individual tools can still opt-in independently via their own `include_return_schema` flag.
+    """
+
+    async def get_tools(self, ctx: RunContext[AgentDepsT]) -> dict[str, ToolsetTool[AgentDepsT]]:
+        original_tools = await super().get_tools(ctx)
+
+        def _build_description(tool: ToolsetTool[AgentDepsT]) -> str | None:
+            if (self.include_return_schema or tool.include_return_schema) and tool.tool_def.return_schema is not None:
+                # TODO: This should be overridable by PromptConfig when that lands
```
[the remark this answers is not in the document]
adtyavrdhn: Yeah it will and that's okay there are a couple of PRs which will need that

pydantic_ai_slim/pydantic_ai/tools.py
```diff
@@ -363,6 +365,7 @@ async def prep_my_tool(
             timeout: Timeout in seconds for tool execution. If the tool takes longer, a retry prompt is returned to the model.
                 Defaults to None (no timeout).
             function_schema: The function schema to use for the tool. If not provided, it will be generated.
+            include_return_schema: Whether to include the return schema in the tool definition. Defaults to False.
```
DouweM: "the tool definition _send to the model_", right? Since the return schema will always be on the `ToolDefinition`.

pydantic_ai_slim/pydantic_ai/profiles/__init__.py
```diff
@@ -25,6 +25,14 @@ class ModelProfile:
 
     supports_tools: bool = True
     """Whether the model supports tools."""
+    supports_tool_return_schema: bool = False
+    """Whether the model natively supports tool return schemas.
+
+    When True, the model's API accepts a structured return schema alongside each tool definition
+    (e.g., Google's ``response_json_schema`` in ``FunctionDeclarationDict``). The return_schema will be
```
DouweM: why the double backticks?

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -1337,9 +1347,17 @@ def _customize_tool_def(transformer: type[JsonSchemaTransformer], tool_def: Tool
     """
     schema_transformer = transformer(tool_def.parameters_json_schema, strict=tool_def.strict)
     parameters_json_schema = schema_transformer.walk()
+
+    # Also transform return_schema if present (e.g., strip unsupported fields for Google)
```
DouweM: Is this actually required? The limitations shouldn't apply because the LLM is never going to generate anything using this schema

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -1354,6 +1372,20 @@ def _customize_output_object(transformer: type[JsonSchemaTransformer], output_ob
...
+    """Inject the return schema as JSON text into the tool description.
+
+    This is a fallback for models that don't natively support tool return schemas.
+    """
+    if tool_def.return_schema is None:
+        return tool_def
+    import json
+
+    base_desc = tool_def.description or ''
+    description = '\n\n'.join([base_desc, 'Return schema:', json.dumps(tool_def.return_schema, indent=2)])
```
DouweM: Unnecessary newlines if `tool_def.description` is none :)

pydantic_ai_slim/pydantic_ai/mcp.py
DouweM: That feels too aggressive to me. Note that in https://github.com/pydantic/pydantic-ai/pull/4123 we're adding an opt-in flag to include MCP server instructions in the agent instructions. I think this should be a flag on the toolset as well.

pydantic_ai_slim/pydantic_ai/toolsets/fastmcp.py
```diff
@@ -141,6 +141,7 @@ async def get_tools(self, ctx: RunContext[AgentDepsT]) -> dict[str, ToolsetTool[
                             'annotations': mcp_tool.annotations.model_dump() if mcp_tool.annotations else None,
                             'output_schema': mcp_tool.outputSchema or None,
                         },
+                        return_schema=mcp_tool.outputSchema or None,
```
DouweM, line 154: The behavior of FastMCPServer should be consistent with MCPServer in terms of when this return schema is actually included

pydantic_ai_slim/pydantic_ai/tools.py
```diff
@@ -286,6 +286,7 @@ class Tool(Generic[ToolAgentDepsT]):
 
     This schema may be modified by the `prepare` function or by the Model class prior to including it in an API request.
     """
+    include_return_schema: bool
```
DouweM: All fields on `Tool` should also be on `FunctionToolset` as defaults for the `Tool`s it builds

pydantic_ai_slim/pydantic_ai/tools.py
```diff
@@ -441,6 +448,7 @@ def tool_def(self):
             metadata=self.metadata,
             timeout=self.timeout,
             kind='unapproved' if self.requires_approval else 'function',
+            return_schema=self.function_schema.return_schema,
```
DouweM, line 467: I'd like the `include` flag to live on the `ToolDefinition` so that it can be modified using `PreparedToolset`.

If the flag is set on the entire agent, that can also be applied on all `ToolDefinition`s using a `PreparedToolset` in `Agent._get_toolset`, instead of by passing the value to `ToolManager` directly. That `PreparedToolset` could then also nullify the field if the boolean is false at that stage.

Note that in that case I think the flag on `ToolDefinition` should be `None` by default, in which case the default from the agent is used, but supporting `False` to disable it explicitly.

pydantic_ai_slim/pydantic_ai/toolsets/abstract.py
```diff
@@ -57,6 +57,8 @@ class ToolsetTool(Generic[AgentDepsT]):
 
     For example, a [`pydantic.TypeAdapter(...).validator`](https://docs.pydantic.dev/latest/concepts/type_adapter/) or [`pydantic_core.SchemaValidator`](https://docs.pydantic.dev/latest/api/pydantic_core/#pydantic_core.SchemaValidator).
     """
+    include_return_schema: bool = False
```
DouweM: See above, I think this should be on ToolDef

pydantic_ai_slim/pydantic_ai/_function_schema.py
```diff
@@ -211,6 +217,28 @@ def function_schema(  # noqa: C901
...
+            if return_description and 'description' not in return_schema:
+                return_schema['description'] = return_description
+        except (PydanticSchemaGenerationError, PydanticUserError):
+            pass
+    elif return_annotation:  # ELIF!
```
DouweM: ELIF? :)

pydantic_ai_slim/pydantic_ai/_function_schema.py
```diff
@@ -211,6 +217,28 @@ def function_schema(  # noqa: C901
         # and set it on the tool
         description = json_schema.pop('description', None)
 
+    return_annotation = type_hints.get('return')
+    return_schema = None
+    # Check if origin of return annotation is ToolReturn which needs special handling.
+    if get_origin(return_annotation) is ToolReturn or return_annotation is ToolReturn:
```
DouweM: `if return_annotation is ToolReturn:`, we can skip all the `get_args` logic right?

Also if `inner_type=Any`, we don't need to do the whole `TypeAdapter` thing

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -654,9 +654,11 @@ class CachePoint:
 MultiModalContent = ImageUrl | AudioUrl | DocumentUrl | VideoUrl | BinaryContent
 UserContent: TypeAlias = str | MultiModalContent | CachePoint
 
+T = TypeVar('T', default=Any)
+
 
 @dataclass(repr=False)
-class ToolReturn:
+class ToolReturn(Generic[T]):
```
DouweM: We don't use `T`?

I suppose `ToolReturnContent` below should be `T`, and `default=Any` above should be `default=ToolReturnContent`

pydantic_ai_slim/pydantic_ai/_tool_manager.py
```diff
@@ -41,6 +41,13 @@ class ToolManager(Generic[AgentDepsT]):
     """Names of tools that failed in this run step."""
     default_max_retries: int = 1
     """Default number of times to retry a tool"""
+    include_return_schema: bool = False
```
DouweM: See above, may not need to be here if it can be set correctly on `ToolDefinition`s

pydantic_ai_slim/pydantic_ai/_tool_manager.py
```diff
@@ -79,11 +86,22 @@ async def for_run_step(self, ctx: RunContext[AgentDepsT]) -> ToolManager[AgentDe
...
 
+        tools = await self.toolset.get_tools(ctx)
+
+        # Gate return_schema: clear it unless opted in at agent or tool level
+        for name, tool in list(tools.items()):
+            opted_in = (
+                self.include_return_schema or tool.include_return_schema
+            ) and tool.tool_def.return_schema is not None
+            if not opted_in:
+                tools[name] = replace(tool, tool_def=replace(tool.tool_def, return_schema=None))
```
DouweM: I'd rather not do have logic this specific inside `ToolManager`. I think with the other suggestions, we can do this in a new `PreparedToolset`. `Model.prepare_request` also seems relevant, we could do the nullifying there, or maybe do the insert-into-description bit there, as we have access to `profile` at that point.

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -701,6 +701,16 @@ def customize_request_parameters(self, model_request_parameters: ModelRequestPar
                     output_object=_customize_output_object(transformer, output_object),
                 )
 
+        # For models that don't natively support tool return schemas,
+        # inject the return schema as JSON text into the tool description as a fallback
+        if not self.profile.supports_tool_return_schema:
```
DouweM: Let's move this to `Model.prepare_request`, that's where we do all the profile-based modifications to params.

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/tools.py
adtyavrdhn: Yep

pydantic_ai_slim/pydantic_ai/profiles/__init__.py
adtyavrdhn: > Blames Claude

pydantic_ai_slim/pydantic_ai/models/__init__.py
adtyavrdhn: I thought because parameters go through it return schema should as well just in case?

pydantic_ai_slim/pydantic_ai/models/__init__.py
adtyavrdhn: Yeah missed that, fixing

pydantic_ai_slim/pydantic_ai/mcp.py
adtyavrdhn: I was wondering the same, I added a comment to make it obvious and a note in the docs but I agree, will revert.

pydantic_ai_slim/pydantic_ai/toolsets/fastmcp.py
adtyavrdhn: Yep

pydantic_ai_slim/pydantic_ai/messages.py
adtyavrdhn: We were but when I merged it got removed because `ToolReturnContent` was added and I was unsure how to handle it cleanly. What you said makes sense though

pydantic_ai_slim/pydantic_ai/models/__init__.py
DouweM: @adtyavrdhn I wouldn't unless we have to.

pydantic_ai_slim/pydantic_ai/_function_schema.py
```diff
@@ -211,6 +217,34 @@ def function_schema(  # noqa: C901
...
+        type_args = get_args(return_annotation)
+        inner_type = type_args[0] if type_args else Any
+        if inner_type is not Any:
+            try:
+                return_schema = TypeAdapter(inner_type).json_schema(
+                    schema_generator=schema_generator, mode='serialization'
+                )
+                if return_description and 'description' not in return_schema:
+                    return_schema['description'] = return_description
+            except (PydanticSchemaGenerationError, PydanticUserError):
```
DouweM: These line are all duplicated with the branch below, should be able to clean that up a bit

pydantic_ai_slim/pydantic_ai/_griffe.py
```diff
@@ -33,10 +33,13 @@ def doc_descriptions(
             * Plain text if no returns section is present
             * XML-formatted if returns section exists, including <summary> and <returns> tags
         - dict[str, str]: Dictionary mapping parameter names to their descriptions
+        - str | None: Return description string, which may be either:
+            * Plain text if no returns section is present
```
DouweM: Inaccurate right?

pydantic_ai_slim/pydantic_ai/_griffe.py
```diff
@@ -77,7 +82,7 @@ def doc_descriptions(
         else:
             main_desc = return_xml
 
-    return main_desc, params
+    return main_desc, params, return_desc
```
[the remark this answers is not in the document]
DouweM: Yeah let's make sure it's not duplicated -- if it's already in the tool description via this route, we shouldn't include it in the return schema description a second time. If it's not in the tool description though, the return schema seems like the right place to put it

pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -319,6 +325,9 @@ def __init__(
             tool_timeout: Default timeout in seconds for tool execution. If a tool takes longer than this,
                 the tool is considered to have failed and a retry prompt is returned to the model (counting towards the retry limit).
                 Individual tools can override this with their own timeout. Defaults to None (no timeout).
+            include_tool_return_schema: Whether to include the return schema in tool descriptions. When True, all tools
```
DouweM: The "whether to send the return schema to the model" description better as it covers the Google case

pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -319,6 +325,9 @@ def __init__(
...
+                will have their return schemas appended to their descriptions. Individual tools can also opt-in
```
DouweM: Unless `include_return_schema` is specifically disabled on a tool right. We can soften this to be more about setting a default value

pydantic_ai_slim/pydantic_ai/agent/__init__.py
DouweM: I'm also going back and forth on whether this should be configurable on the `Agent` at all, because we already have so many arguments, or if a wrapper toolset (like `ApprovalRequiredToolset`) makes sense... For now I'm OK with the agent arg.

pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -1516,6 +1529,27 @@ def copy_dynamic_toolsets(toolset: AbstractToolset[AgentDepsT]) -> AbstractTools
...
+        ) -> list[ToolDefinition]:
+            resolved: list[ToolDefinition] = []
+            for td in tool_defs:
+                include = (
+                    td.include_return_schema if td.include_return_schema is not None else agent_include_return_schema
+                )
+                if not include and td.return_schema is not None:
+                    td = dataclasses.replace(td, return_schema=None, include_return_schema=False)
+                else:
+                    td = dataclasses.replace(td, include_return_schema=bool(include))
```
DouweM: I don't think we need to do this if `return_schema` is not set anyway. Creating a new dataclass is cheap but not free.

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -1354,6 +1363,24 @@ def _customize_output_object(transformer: type[JsonSchemaTransformer], output_ob
...
 
 
+def _inject_return_schema_in_description(tool_def: ToolDefinition) -> ToolDefinition:
+    """Inject the return schema as JSON text into the tool description.
+
+    This is a fallback for models that don't natively support tool return schemas.
+    """
+    if tool_def.return_schema is None:
+        return tool_def
+    import json
```
DouweM: at the top please :)

docs/toolsets.md
```diff
@@ -383,6 +383,78 @@ print(result.output)
 
 _(This example is complete, it can be run "as is")_
 
+### Including Return Schemas {#including-return-schemas}
```
DouweM: Seems like this should be on the Advanced Tools doc.

docs/toolsets.md
```diff
@@ -383,6 +383,78 @@ print(result.output)
...
+    'type': 'object',
+}
+"""
+```
+
+_(This example is complete, it can be run "as is")_
+
+The return schema is automatically generated from the function's return type annotation. If the return type's JSON schema doesn't already include a `description`, one is populated from the function's docstring `Returns` section, consistent with how parameter descriptions are extracted from docstrings.
+
+When a tool returns [`ToolReturn[T]`][pydantic_ai.messages.ToolReturn], the return schema is inferred from the inner type `T` rather than `ToolReturn` itself. For example, a tool annotated `-> ToolReturn[UserDetails]` will have its return schema derived from `UserDetails`.
```
DouweM: We should update the `ToolReturn` docs to make it clear that in order to get return schemas, you should specify the generic type

docs/toolsets.md
```diff
@@ -671,26 +743,26 @@ Pydantic AI provides two toolsets that allow an agent to connect to and call too
...
-- [`pydantic-deep`](https://github.com/vstorm-co/pydantic-deepagents) — Deep agent framework that includes a `FilesystemToolset` with multiple backends (in-memory, real filesystem, Docker sandbox).
+* [`pydantic-ai-filesystem-sandbox`](https://github.com/zby/pydantic-ai-filesystem-sandbox) - `FileSystemToolset` with a sandbox and LLM-friendly errors
+* [`pydantic-deep`](https://github.com/vstorm-co/pydantic-deepagents) — Deep agent framework that includes a `FilesystemToolset` with multiple backends (in-memory, real filesystem, Docker sandbox).
 
 ### Code Execution
 
 Toolsets for sandboxed code execution help agents run code in a sandboxed environment:
 
-- [`mcp-run-python`](https://github.com/pydantic/mcp-run-python) - MCP server by the Pydantic team that runs Python code in a sandboxed environment. Can be used as `MCPServerStdio('uv', args=['run', 'mcp-run-python', 'stdio'])`.
+* [`mcp-run-python`](https://github.com/pydantic/mcp-run-python) - MCP server by the Pydantic team that runs Python code in a sandboxed environment. Can be used as `MCPServerStdio('uv', args=['run', 'mcp-run-python', 'stdio'])`.
```
DouweM, line 694: Unnecessary changes :)

DouweM (changes requested):
Nice work, I think we can merge this in the next few days :)

pydantic_ai_slim/pydantic_ai/agent/__init__.py
adtyavrdhn: Yeah, maybe both can co exist either way? If Agent level is false but a toolset needs it to be enabled. 

I like it on the Agent personally but I absolutely agree it has too many args now.

pydantic_ai_slim/pydantic_ai/agent/__init__.py
adtyavrdhn: Yeah this might be misleading, I should mention `specifically disabled on a tool ` as well which would make sense for tools that can have bloated return schemas.

pydantic_ai_slim/pydantic_ai/_griffe.py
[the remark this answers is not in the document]
adtyavrdhn: Great catch,  completely missed that.

pydantic_ai_slim/pydantic_ai/_function_schema.py
```diff
@@ -211,6 +215,31 @@ def function_schema(  # noqa: C901
...
+    if return_annotation is ToolReturn:
+        # Bare ToolReturn without type parameter — no schema to generate
+        pass
+    elif get_origin(return_annotation) is ToolReturn:
+        type_args = get_args(return_annotation)
+        inner_type = type_args[0] if type_args else Any
+        if inner_type is not Any:
+            schema_type = inner_type
+    elif return_annotation is not None and return_annotation is not type(None):
+        schema_type = return_annotation
```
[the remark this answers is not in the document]
adtyavrdhn: Yeah we discussed but feel like we should let the user do something weird at the cost of being expensive rather than at the cost of being incorrect.

docs/tools-advanced.md
```diff
@@ -124,6 +124,81 @@ print(result.output)
...
+    'required': ['id', 'name', 'email'],
+    'title': 'UserDetails',
+    'type': 'object',
+}
+"""
+```
+
+_(This example is complete, it can be run "as is")_
+
+The return schema is automatically generated from the function's return type annotation. If the return type's JSON schema doesn't already include a `description`, one is populated from the function's docstring `Returns` section, consistent with how parameter descriptions are extracted from docstrings.
```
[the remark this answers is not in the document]
adtyavrdhn: Yeppers thanks

pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -252,6 +256,7 @@ def __init__(
         history_processors: Sequence[HistoryProcessor[AgentDepsT]] | None = None,
         event_stream_handler: EventStreamHandler[AgentDepsT] | None = None,
         tool_timeout: float | None = None,
+        include_tool_return_schema: bool = False,
```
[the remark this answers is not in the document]
DouweM: The agent tool decorators should definitely have it.

pydantic_ai_slim/pydantic_ai/_function_schema.py
```diff
@@ -211,6 +215,31 @@ def function_schema(  # noqa: C901
...
+
+    # Generate return schema if we have a type
+    if schema_type is not None:
+        try:
+            return_schema = TypeAdapter(schema_type).json_schema(
+                schema_generator=schema_generator, mode='serialization'
+            )
+        except (PydanticSchemaGenerationError, PydanticUserError):
+            pass
```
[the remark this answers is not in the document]
adtyavrdhn: @DouweM Any preferences for how this should be handled?

pydantic_ai_slim/pydantic_ai/agent/__init__.py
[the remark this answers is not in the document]
adtyavrdhn: Yep, I'll add it!

tests/test_tools.py
```diff
@@ -2779,6 +2823,974 @@ def test_agent_tool_timeout_passed_to_toolset():
...
+
+    @agent.tool_plain
+    def tool_with_return_tool_return() -> ToolReturn:  # pragma: no cover
+        return ToolReturn(return_value='hello', content=[ImageUrl(url='https://example.com/image.jpg')])
+
+    @agent.tool_plain
+    def tool_with_deferred_return() -> DeferredToolResult:  # pragma: no cover
+        return ToolApproved()
+
+    # DeferredToolResult has an insane token footprint, unsure if it is wise to do this but I will read more.
```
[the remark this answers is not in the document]
adtyavrdhn: @DouweM We maintain the same stance of user doing something weird let them or do you feel differently now?

pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -319,6 +324,9 @@ def __init__(
             tool_timeout: Default timeout in seconds for tool execution. If a tool takes longer than this,
                 the tool is considered to have failed and a retry prompt is returned to the model (counting towards the retry limit).
                 Individual tools can override this with their own timeout. Defaults to None (no timeout).
+            include_tool_return_schema: Whether to include the return schema in tool descriptions. When True, all tools
+                will have their return schemas appended to their descriptions. Individual tools can also opt-in
+                via their own `include_return_schema` flag. Defaults to False.
```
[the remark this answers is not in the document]
adtyavrdhn: This is so great @DouweM !! The iteration speed on our PRs will def be boosted with you not being bottlenecked, great work :))

tests/test_tools.py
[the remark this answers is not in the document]
DouweM: Yeah let's trust the user. I don't think there should ever be a reason to return `DeferredToolResult`, fwiw.

pydantic_ai_slim/pydantic_ai/_function_schema.py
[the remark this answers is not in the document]
DouweM: Hmm. If "include..." is enabled on eg the agent, it makes sense to me to print a warning for any tool that doesn't have a return schema. If the user doesn't want it on one specific tool, they can explicitly turn it to `False` on that tool right? 

If the default is set to True on the agent/functiontoolset, then could we push that value through to here so we can include the schema generation error? If not, we may only be able to say that there is none, as I wouldn't want to store the error on `ToolDefinition`.

pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -1516,6 +1537,34 @@ def copy_dynamic_toolsets(toolset: AbstractToolset[AgentDepsT]) -> AbstractTools
...
+                    td = dataclasses.replace(td, return_schema=None, include_return_schema=False)
+                elif td.include_return_schema is None and agent_include_return_schema:
+                    td = dataclasses.replace(td, include_return_schema=True)
+                if include and td.return_schema is None:
+                    warnings.warn(
+                        f'Tool {td.name!r} has `include_return_schema` enabled but no return schema was generated. '
+                        f'Set `include_return_schema=False` on this tool to suppress this warning.',
+                        UserWarning,
+                        stacklevel=2,
+                    )
```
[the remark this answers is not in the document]
adtyavrdhn: Claude: Python's default warning filter (`default`) deduplicates by `(message, category, module, lineno)`. Since the `warnings.warn` call is at the same line in the same module with the same message text for a given tool name, it only displays once per process even though `_resolve_return_schema` runs on every step.

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -1354,6 +1364,23 @@ def _customize_output_object(transformer: type[JsonSchemaTransformer], output_ob
...
+    This is a fallback for models that don't natively support tool return schemas.
+    """
+    if tool_def.return_schema is None:
+        return tool_def
+
+    parts: list[str] = []
+    if tool_def.description:
+        parts.append(tool_def.description)
+    parts.append('Return schema:')
+    parts.append(json.dumps(tool_def.return_schema, indent=2))
```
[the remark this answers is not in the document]
adtyavrdhn: I don't see the point of clearing it, models will not use it / send it apart from desc anyway?

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -1365,6 +1375,23 @@ def _customize_output_object(transformer: type[JsonSchemaTransformer], output_ob
...
+    description = '\n\n'.join(parts)
+    return replace(tool_def, description=description)
```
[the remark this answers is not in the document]
adtyavrdhn: I don't want to clear it smh, why would I clear it if non google models ignore the schema? ToolDef is not passed in verbatim?

pydantic_ai_slim/pydantic_ai/models/google.py
```diff
@@ -1172,6 +1172,8 @@ def _function_declaration_from_tool(tool: ToolDefinition) -> FunctionDeclaration
         description=tool.description or '',
         parameters_json_schema=json_schema,
     )
+    if tool.return_schema is not None:
+        f['response_json_schema'] = tool.return_schema
```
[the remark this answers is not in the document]
adtyavrdhn: @DouweM I remember you saying, It doesn't matter?

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -676,9 +676,22 @@ class CachePoint:
...
+    ToolReturnContent: TypeAlias = MultiModalContent | Sequence[Any] | Mapping[str, Any] | Any
+else:
+    # Recursive type for runtime Pydantic validation - enables automatic reconstruction of
+    # BinaryContent/FileUrl objects nested inside dicts/lists during deserialization
+    ToolReturnContent = TypeAliasType(
+        'ToolReturnContent',
+        MultiModalContent | Sequence['ToolReturnContent'] | Mapping[str, 'ToolReturnContent'] | Any,
+    )
+
+T = TypeVar('T', default=ToolReturnContent)
```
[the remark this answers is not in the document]
adtyavrdhn: Sure valid

pydantic_ai_slim/pydantic_ai/_function_schema.py
```diff
@@ -211,6 +218,36 @@ def function_schema(  # noqa: C901
...
+    if return_annotation is ToolReturn:
+        # Bare ToolReturn without type parameter — no schema to generate
+        pass
+    elif get_origin(return_annotation) is ToolReturn:
+        type_args = get_args(return_annotation)
+        inner_type = type_args[0] if type_args else Any
+        if inner_type is not Any:
+            schema_type = inner_type
+    elif return_annotation is not None and return_annotation is not type(None):
+        schema_type = return_annotation
```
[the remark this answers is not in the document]
adtyavrdhn: That is a low cost right? I have to read what it means by resolving it

tests/test_tools.py
```diff
@@ -2779,6 +2823,1028 @@ def test_agent_tool_timeout_passed_to_toolset():
...
```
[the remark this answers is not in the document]
adtyavrdhn: Hehe will remove, user do we allow decided!

pydantic_ai_slim/pydantic_ai/_function_schema.py
```diff
@@ -211,6 +218,36 @@ def function_schema(  # noqa: C901
...
+
+    # Generate return schema if we have a type and the tool hasn't explicitly opted out
+    if schema_type is not None and include_return_schema is not False:
```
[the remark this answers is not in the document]
DouweM: This is not that expensive to do if it's just at tool registration time.

pydantic_ai_slim/pydantic_ai/models/google.py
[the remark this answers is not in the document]
DouweM: I don't believe it has those same limitations, as they don't enforce that the tool return matches it, it's presumably just a hint to the LLM.

pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -1533,6 +1554,33 @@ def copy_dynamic_toolsets(toolset: AbstractToolset[AgentDepsT]) -> AbstractTools
...
+                if include and td.return_schema is None:
+                    warnings.warn(
+                        f'Tool {td.name!r} has `include_return_schema` enabled but no return schema was generated. '
+                        f'Set `include_return_schema=False` on this tool to suppress this warning.',
+                        UserWarning,
+                    )
+                resolved.append(td)
+            return resolved
+
+        toolset = PreparedToolset(toolset, _resolve_return_schema)
```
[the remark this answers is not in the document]
adtyavrdhn: No we also need to clear out the schema in case some tool opted out which would need to be via _resolve_return_schema otherwise we leak the schema anyway as pointed out by another of your comments.

adtyavrdhn:
Same comments again :(

pydantic_ai_slim/pydantic_ai/tools.py
```diff
@@ -533,6 +543,22 @@ class ToolDefinition:
     Defaults to None (no timeout).
     """
 
+    return_schema: ObjectJsonSchema | None = None
```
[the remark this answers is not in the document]
DouweM: `JsonSchemaValue` works for me

pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -1542,6 +1563,33 @@ def copy_dynamic_toolsets(toolset: AbstractToolset[AgentDepsT]) -> AbstractTools
...
```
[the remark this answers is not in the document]
DouweM: That sounds reasonable

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -676,9 +676,22 @@ class CachePoint:
...
+    ToolReturnContent: TypeAlias = MultiModalContent | Sequence[Any] | Mapping[str, Any] | Any
+else:
+    # Recursive type for runtime Pydantic validation - enables automatic reconstruction of
+    # BinaryContent/FileUrl objects nested inside dicts/lists during deserialization
+    ToolReturnContent = TypeAliasType(
+        'ToolReturnContent',
+        MultiModalContent | Sequence['ToolReturnContent'] | Mapping[str, 'ToolReturnContent'] | Any,
+    )
+
+ToolReturnT = TypeVar('ToolReturnT', default=ToolReturnContent)
```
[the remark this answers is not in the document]
DouweM: `ToolReturnT` would indeed be better as `ToolReturnValueT`

Using `ToolReturnContent` make some sense, and I requested this, but thinking about what it will mean for the LLM: functions currently annotated with `-> ToolReturn` without a generic type would ideally return a schema saying "this can be anything" (or no return schema at all?), rather than including the schemas for `BinaryContent` etc when it's not actually likely that'll be returned.

pydantic_ai_slim/pydantic_ai/_function_schema.py
```diff
@@ -211,6 +218,36 @@ def function_schema(  # noqa: C901
...
+
+    if return_annotation is ToolReturn:
+        # Bare ToolReturn without type parameter — no schema to generate
+        pass
+    elif get_origin(return_annotation) is ToolReturn:
+        type_args = get_args(return_annotation)
+        inner_type = type_args[0] if type_args else Any
+        if inner_type is not Any:
+            schema_type = inner_type
+    elif return_annotation is not None and return_annotation is not type(None):
```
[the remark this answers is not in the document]
DouweM: This is a good thing, the model needs to know if it can return None.

pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -1542,6 +1563,33 @@ def copy_dynamic_toolsets(toolset: AbstractToolset[AgentDepsT]) -> AbstractTools
...
+            ctx: RunContext[AgentDepsT], tool_defs: list[ToolDefinition]
+        ) -> list[ToolDefinition]:
+            resolved: list[ToolDefinition] = []
+            for td in tool_defs:
+                include = (
+                    td.include_return_schema if td.include_return_schema is not None else agent_include_return_schema
+                )
+                if not include and td.return_schema is not None:
+                    td = dataclasses.replace(td, return_schema=None)
+                elif td.include_return_schema is None and agent_include_return_schema:
```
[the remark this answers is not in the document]
DouweM: Unnecessary `dataclasses.replace` calls should definitely be prevented.

tests/test_tools.py
```diff
@@ -2779,6 +2823,1071 @@ def test_agent_tool_timeout_passed_to_toolset():
...
+                'name': 'tool_with_return_schema_binary_content',
+                'parameters_json_schema': {'additionalProperties': False, 'properties': {}, 'type': 'object'},
+                'description': None,
+                'outer_typed_dict_key': None,
+                'strict': None,
+                'sequential': False,
+                'kind': 'function',
+                'metadata': None,
+                'timeout': None,
+                'return_schema': {
```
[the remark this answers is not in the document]
DouweM: Hmm I think that for special types like this, that we actually return to the model directly (https://github.com/pydantic/pydantic-ai/pull/3826) rather than as JSON, sharing this schema in the description or in Google's arg is actually less than helpful. In CodeMode, the schema is valuable because the code may want to read attrs from these types (e.g. if it knows `BinaryContent` holds `bytes` as base64, it could actually parse those and do something with them).

But I'd rather keep this unfixed for now and move it to a followup issue.

docs/tools-advanced.md
```diff
@@ -124,6 +124,83 @@ print(result.output)
...
+toolset = FunctionToolset()
+
+
+@toolset.tool
+def get_user(user_id: int) -> UserDetails:
+    """Get user details by ID."""
+    return UserDetails(id=user_id, name='Alice', email='alice@example.com')
+
+
+test_model = TestModel()  # (1)!
```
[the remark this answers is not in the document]
DouweM: This is fine

pydantic_ai_slim/pydantic_ai/_function_schema.py
```diff
@@ -211,6 +218,36 @@ def function_schema(  # noqa: C901
...
+        schema_type = return_annotation
+
+    # Generate return schema if we have a type and the tool hasn't explicitly opted out
+    if schema_type is not None and include_return_schema is not False:
+        try:
+            return_schema = TypeAdapter(schema_type).json_schema(
+                schema_generator=schema_generator, mode='serialization'
+            )
+        except (PydanticSchemaGenerationError, PydanticUserError) as e:
+            if include_return_schema:
```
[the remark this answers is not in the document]
DouweM: I'm OK leaving this as is

pydantic_ai_slim/pydantic_ai/_function_schema.py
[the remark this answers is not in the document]
DouweM: I think this a good idea in principle, at least for the non-CodeMode case, but it'd also need to work with `list[BinaryContent]`, `MyOutput | DeferredToolResult` etc. OK leaving that for later.

pydantic_ai_slim/pydantic_ai/messages.py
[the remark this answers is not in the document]
DouweM: Yep, `default=Any` will be cleaner

pydantic_ai_slim/pydantic_ai/_function_schema.py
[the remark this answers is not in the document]
DouweM: It'd be worth seeing if we could use a `cached_property` here so we don't eagerly generate the `return_schema`, just when it's read. Similar to what the code mode PR is doing for `python_signature`.

DouweM:
@adtyavrdhn Please have a look at merge conflicts here!

PR #3890, opened 2025-12-31
Exclude `run_id` from cache key computation in `PrefectAgentInputs`

### LINKED ISSUE #3883: `run_id` not ignored in PrefectAgentInputs Prefect cache policy
### Initial Checks

### Description

https://github.com/pydantic/pydantic-ai/blob/b58e6e4cb64726c5be9938d4660038ec4a5900e2/pydantic_ai_slim/pydantic_ai/durable_exec/prefect/_cache_policies.py#L77
`PrefectAgentInputs` is a Prefect `CachePolicy` which computes a cache key based on the inputs ignoring timestamp fields. With the addition of `run_id` https://github.com/pydantic/pydantic-ai/pull/3366 to messages I think it should also exclude run_id as this is simple recreated for every new run invocation.

Sidenote: I think it would also be useful to make `PrefectAgentInputs` public? At the moment only `DEFAULT_PYDANTIC_AI_CACHE_POLICY` is.

```Python
from typing import Any
from pydantic_ai import Agent
from pydantic_ai.durable_exec.prefect import PrefectAgent, TaskConfig
from pydantic_ai.durable_exec.prefect._cache_policies import (
    _strip_timestamps,
    _replace_toolsets,
    _replace_run_context,
)
from prefect.cache_policies import TASK_SOURCE, CachePolicy, INPUTS
from prefect.context import TaskRunContext


class PrefectAgentInputs(CachePolicy):
    """Cache policy designed to handle input hashing for PrefectAgent cache keys.

    Computes a cache key based on inputs, ignoring nested 'timestamp' fields
    and serializing RunContext objects to only include hashable fields.
    """

    def compute_key(
        self,
        task_ctx: TaskRunContext,
        inputs: dict[str, Any],
        flow_parameters: dict[str, Any],
        **kwargs: Any,
    ) -> str | None:
        """Compute cache key from inputs with timestamps removed and RunContext serialized."""
        if not inputs:
            return None

        inputs_without_toolsets = _replace_toolsets(inputs)
        inputs_with_hashable_context = _replace_run_context(inputs_without_toolsets)
        filtered_inputs = _strip_timestamps(inputs_with_hashable_context)

        for msg in filtered_inputs.get("messages", []):
            print(f"run_id = {msg.get('run_id')}")

        return INPUTS.compute_key(task_ctx, filtered_inputs, flow_parameters, **kwargs)

agent = Agent("openai:gpt-5-mini", output_type=str, name="test")

# Use local caching (persist_result=True without result_storage uses local)
prefect_agent = PrefectAgent(
    agent,
    model_task_config=TaskConfig(
        cache_policy=PrefectAgentInputs(),
        persist_result=True,
    ),
)

prompt = "What is 2+2? Just the number."

r1 = prefect_agent.run_sync(prompt)
r2 = prefect_agent.run_sync(prompt)
```

### Logfire Trace

_No response_

- Python: 3.12.11
- Pydantic AI: 1.39.0
- LLM provider SDK: /

adtyavrdhn:
- Closes #3883 

### Pre-Review Checklist

### Pre-Merge Checklist


adtyavrdhn:
Prefect tests are being skipped at the moment, I am not sure if we can make this change with it still being skipped? I removed the skip and tested the change with the newly added test case(which works fine).

tests/test_prefect.py
```diff
@@ -1168,6 +1168,23 @@ async def test_cache_policy_empty_inputs():
...
+        task_ctx=mock_task_ctx,
+        inputs={'messages': [ModelRequest(parts=[], timestamp=time, run_id='run-1')]},
+        flow_parameters={},
+    ) == cache_policy.compute_key(
+        task_ctx=mock_task_ctx,
+        inputs={'messages': [ModelRequest(parts=[], timestamp=time, run_id='run-2')]},
+        flow_parameters={},
+    )
+
+
```
dsfaccini: to the extent that's possible, let's add a test that catches ephemeral values in the future so we're less likely to introduce cache-breaking fields without noticing

dsfaccini:
> Prefect tests are being skipped at the moment, I am not sure if we can make this change with it still being skipped? I removed the skip and tested the change with the newly added test case(which works fine).

Yeah CI was being flaky the past week but we solved it via https://github.com/pydantic/pydantic-ai/pull/3862, so we shouldn't be skipping the prefect tests any longer, as the CI checks just proved :)

tests/test_prefect.py
dsfaccini: by that I mean, having a normal execution run twice and check that the cache is acting on the second run, that way if we add an ephemeral field in the future we'll notice that the cache breaks

adtyavrdhn:
Oh no I meant locally, I just removed the skip and committed. Hopefully we see some greens again.

tests/test_prefect.py
adtyavrdhn: Yep, will add one. I'll remove the existing one as well, just added it to validate the cache key bug.

dsfaccini:
ahh yes, the good ol' **prefect hang**, see https://github.com/pydantic/pydantic-ai/pull/3778/files#diff-0e496998cde42ac9fc809c1e09b69890644494606ae02dbf0c33ae72bc0efdd7 and https://github.com/kevin1024/vcrpy/pull/965

don't put too much effort into fixing this, we'll either need to include this patch in this PR

```python
LOCALHOST_HOSTS = {"localhost", "127.0.0.1", "0.0.0.0"}


def _is_localhost_request(request: httpcore.Request) -> bool:
    return urlparse(bytes(request.url).decode("ascii")).hostname in LOCALHOST_HOSTS


@pytest.fixture(autouse=True, scope="module")
def patch_vcr_httpcore_for_localhost() -> Iterator[None]:
    try:
        import vcr.stubs.httpcore_stubs as httpcore_stubs
    except ImportError:  # pragma: no cover
        yield
        return

    orig_async = httpcore_stubs.vcr_handle_async_request  # pyright: ignore[reportUnknownMemberType, reportUnknownVariableType]
    orig_sync = httpcore_stubs.vcr_handle_request  # pyright: ignore[reportUnknownMemberType, reportUnknownVariableType]

    def _wrap(orig: Any) -> Any:
        def patched(cassette: Any, real_handle: Any) -> Any:
            vcr_wrapper: Any = orig(cassette, real_handle)

            def wrapper(self: Any, req: httpcore.Request) -> Any:
                return real_handle(self, req) if _is_localhost_request(req) else vcr_wrapper(self, req)

            return wrapper

        return patched

    httpcore_stubs.vcr_handle_async_request = _wrap(orig_async)
    httpcore_stubs.vcr_handle_request = _wrap(orig_sync)
    try:
        yield
    finally:
        httpcore_stubs.vcr_handle_async_request = orig_async
        httpcore_stubs.vcr_handle_request = orig_sync
```

or keep skipping the tests

cc: @Kludex

adtyavrdhn:
Ah this along with #3862 and related discussions seems like a good read. I would rather not add this patch tbh if we can wait it out but I'll leave it to you guys :)

adtyavrdhn:
@dsfaccini Seems like we are still running into the same?

dsfaccini:
yeah I think it's now reasonable to keep the tests in skipping

dsfaccini:
hey @adtyavrdhn we've decided internally to keep the patch, so I'll just add it in this PR

adtyavrdhn:
Sure thing @dsfaccini :)

dsfaccini:
@adtyavrdhn we've merged https://github.com/pydantic/pydantic-ai/pull/3930 

testsuite's not hanging anymore but you have some errors

adtyavrdhn:
@dsfaccini I'll take a look.

adtyavrdhn:
@dsfaccini Added a Cache Policy to allow for cross flow caching, the default cache policy has `RUN_ID`(From prefect not PyAI) which is different for each run so cache wasn't working, made it simpler for our purpose. 

Skeptical about it still being hung though seems to be taking a while :(

tests/test_prefect.py
```diff
@@ -74,8 +76,6 @@
     pytest.mark.anyio,
     pytest.mark.vcr,
     pytest.mark.xdist_group(name='prefect'),
-    # TODO(Marcelo): We are temporarily disabling it. We should enable them again.
-    pytest.mark.skip('This test suite is hanging with the latest versions of all packages.'),
```
DouweM, line 77: Let's go back to skipping the tests, just leave some evidence that you ran the new test locally and it passed

uv.lock
DouweM, line 1: Unrelated change?

DouweM (changes requested):

tests/test_prefect.py
adtyavrdhn: Yep will do that

uv.lock
adtyavrdhn: I think @dsfaccini added [954d2ee](https://github.com/pydantic/pydantic-ai/pull/3890/changes/954d2ee63f024c767370c7bad8bd4dfd7c6df21d) for: https://github.com/kevin1024/vcrpy/pull/965 but unsure if we need it now because doesn't seem like it helped.

dsfaccini: we don't need it any more, I inclucded the patch in a different PR but I think even that got reverted. I'll update the branch to see what happens with this. you'll need to `git pull` on your side

tests/test_prefect.py
```diff
@@ -1168,6 +1169,43 @@ async def test_cache_policy_empty_inputs():
...
 
 
+async def test_repeated_run_hits_cache():
+    """Test that running the same prompt twice hits the cache on the second run.
+
+    By default, the cache policy includes RUN_ID which means each flow run has
+    its own cache namespace. To enable cross-flow caching, we use CROSS_FLOW_CACHE_POLICY
+    which excludes RUN_ID.
+
+    We use tmp_path for result_storage to isolate this test from other runs.
```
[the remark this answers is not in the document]
adtyavrdhn: Valid miss

tests/test_prefect.py
```diff
@@ -1168,6 +1169,43 @@ async def test_cache_policy_empty_inputs():
     assert result is None
 
 
+async def test_repeated_run_hits_cache():
```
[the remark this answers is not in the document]
adtyavrdhn: Nah I checked and confirmed with Claude, run_sync wraps it in a flow on its own internally

tests/test_prefect.py
```diff
@@ -1168,6 +1169,43 @@ async def test_cache_policy_empty_inputs():
...
+
+    # Use unique prompt per test run to avoid cross-run cache hits
+    prompt = f'What is 2+2? {uuid.uuid4()}'
+    prefect_agent.run_sync(prompt)
+    assert call_count == 1
+
+    # Cache hit here, should not call the model again (cross-flow caching enabled)
+    prefect_agent.run_sync(prompt)
+    assert call_count == 1
+
```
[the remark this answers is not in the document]
adtyavrdhn: Yeah that is valid I should not have removed that one, will add back.

PR #3892, opened 2025-12-31
Add support for Bedrock Nova 2.0 built-in `Code Interpreter` tool

### LINKED ISSUE #3891: Support for Bedrock Nova2.0 System tools (Code Interpreter)
### Description

The AWS Bedrock Nova 2.0 model now includes several built-in tools. See [Built-in system tools](https://docs.aws.amazon.com/nova/latest/nova2-userguide/using-tools.html#builtin-tools)

This includes:
- Code Interpreter
- Web Grounding

I was wondering if you would accept a PR to add support for this?  I have a [branch](https://github.com/gmetzker/pydantic-ai/tree/bedrock-add-system-tools) that initially supports the **Code Interpreter** with the following:

- Model specific `supported_builtin_tools` in BedrockProvider.model_profile.  From what I can tell this is only supported on Nova2 and no other hosted models.
    - Allow `CodeExecutionTool` for Nova 2.0 model
    - Disallow support_builtin_tools for other models

- Mapping Pydantic `CodeExecutionTool` <-->  Bedrock `nova_code_interpreter` 
    - including BuiltinToolCallPart and BuiltinToolReturnPart
    - BedrockConverseModel._process_response
    - BedrockConverseModel._map_tool_config
    - BedrockConverseModel._map_messages
- Mapping of stream responses in BedrockStreamedResponse._get_event_iterator

WebGrounding might need a little more work but seems like this could be a future PR where it's maybe mapped to `WebSearchTool`?

### References

- [Built-in system tools](https://docs.aws.amazon.com/nova/latest/nova2-userguide/using-tools.html#builtin-tools)
- [Code Interpreter](https://docs.aws.amazon.com/nova/latest/nova2-userguide/using-tools.html#builtin-tools:~:text=the%20Converse%20API.-,Code%20Interpreter,-Code%20Interpreter%20allows)
- [Web Grounding](https://docs.aws.amazon.com/nova/latest/nova2-userguide/web-grounding.html)

gmetzker:
- Closes #3891 <issue>

### Pre-Review Checklist

- [X] **PR title** is fit for the release changelog.

### Pre-Merge Checklist

- [X] New **tests** for any fix or new behavior, maintaining 100% coverage.
- [X] Updated **documentation** for new features and behaviors, including docstrings for API docs.

### DIFF
--- a/tests/models/test_bedrock.py
+++ b/tests/models/test_bedrock.py
@@ -2382,7 +2382,7 @@ async def test_bedrock_map_messages_builtin_tool_provider_filtering(
                 BuiltinToolReturnPart(
                     provider_name='bedrock',
                     tool_name='foo',
-                    content=[{'text': 'success\n'}],
+                    content={'other': 'content'},
                     tool_call_id='call_4',
                     provider_details={'status': 'success'},
                 ),


pydantic_ai_slim/pydantic_ai/models/bedrock.py
```diff
@@ -159,6 +164,59 @@
     'tool_use': 'tool_call',
 }
 
+_BEDROCK_TO_PYDANTIC_BUILTIN_TOOL_MAP: dict[str, str] = {
+    'nova_code_interpreter': CodeExecutionTool.kind,
+}
+_BEDROCK_TO_PYDANTIC_TOOL_RESULT_TYPE_MAP: dict[str, str] = {
+    'nova_code_interpreter_result': CodeExecutionTool.kind,
+}
+
```
gmetzker: The mapping here is maybe overkill, but I was hoping to add the **Web Grounding** tool, and possibly future tools later on.

pydantic_ai_slim/pydantic_ai/models/bedrock.py
```diff
@@ -910,7 +1017,12 @@ async def _get_event_iterator(self) -> AsyncIterator[ModelResponseStreamEvent]:
             self.provider_response_id = self._provider_response_id
 
         chunk: ConverseStreamOutputTypeDef
-        tool_id: str | None = None
+        tool_ids: dict[int, str] = {}
+
+        # Bedrock has deltas for built-in tool returns, which aren't supported by parts manager.
+        # We accumulate the deltas here and yield the complete return part once the content block ends
+        builtin_tool_returns: dict[int, BuiltinToolReturnPart] = {}
```
gmetzker, line 977: Parts manager doesn't seem to support deltas for BuiltinToolReturnPart(s), and from what I can tell from the other model tests, they only emit a single `PartStartEvent` for a BuiltinToolReturnPart.  Other models didn't seem to emit any `PartDeltaEvent`(s), or `PartEndEvent`(s).

On the other hand, Bedrock/Nova2 seems to have multiple blocks for a  BuiltinToolReturnPart.  So here I'm just accumulating them and then fire one event in the `contentBlockStop` case (via `yield self._parts_manager.handle_part`)

pydantic_ai_slim/pydantic_ai/providers/bedrock.py
```diff
@@ -89,14 +96,20 @@ def model_profile(self, model_name: str) -> ModelProfile | None:
                 bedrock_send_back_thinking_parts=True,
                 bedrock_supports_prompt_caching=True,
                 bedrock_supports_tool_caching=True,
-            ).update(anthropic_model_profile(model_name)),
-            'mistral': lambda model_name: BedrockModelProfile(bedrock_tool_result_format='json').update(
-                mistral_model_profile(model_name)
-            ),
+            )
```
gmetzker: Only Bedrock Nova2.0 supports the built-in tools (for now), at least as far as I can tell.
I assume we want some way of marking these other models such that they don't have support for built-in tools.

docs/models/bedrock.md
```diff
@@ -243,6 +243,31 @@ print(result.output)
...
+Nova 2.0 models support built-in tools; see the [Nova 2.0 documentation](https://docs.aws.amazon.com/nova/latest/nova2-userguide/using-tools.html#builtin-tools) for details.
+
+| Tool | Pydantic-AI | Bedrock | Notes |
+|------|-------------------|----------|-------|
+| [Code Interpreter](https://docs.aws.amazon.com/nova/latest/nova2-userguide/using-tools.html#builtin-tools) | `CodeExecutionTool` | `nova_code_interpreter` | Execute Python code in isolated sandbox environments |
+| [Web Grounding](https://docs.aws.amazon.com/nova/latest/nova2-userguide/web-grounding.html) | N/A | `nova_grounding` | **Not implemented**: Access real-time information from the internet |
+
+### Example
+
+```python {title="bedrock_nova2_code_execution.py" test="skip"}
```
dsfaccini: what happens if we don't skip the test?

pydantic_ai_slim/pydantic_ai/models/bedrock.py
dsfaccini: if we need these I'd rather have them as a single dict, something like

```python
class _BedrockBuiltinTool(TypedDict):
    name: str  # if the library offers a literal type, use that instead
    result_type: str  # if the library offers a literal type, use that instead

_BUILTIN_TOOL_TO_BEDROCK: dict[str, _BedrockBuiltinTool] = {
    CodeExecutionTool.kind: {'name': 'nova_code_interpreter', 'result_type': 'nova_code_interpreter_result'}
}

# if needed for brevity
_RESULT_TYPE_TO_BUILTIN_TOOL = {bbt['result_type']: kind for kind, bbt in _BUILTIN_TOOL_TO_BEDROCK.items()}
```

pydantic_ai_slim/pydantic_ai/models/bedrock.py
```diff
@@ -159,6 +164,59 @@
...
+def map_bedrock_tool_result_type(pydantic_tool_name: str) -> str:
+    """Gets Bedrock toolResult.type given a Pydantic AI builtin tool name."""
+    if bedrock_tool_result_type := next(
+        (bd for bd, pd in _BEDROCK_TO_PYDANTIC_TOOL_RESULT_TYPE_MAP.items() if pd == pydantic_tool_name),
+        None,
+    ):
+        return bedrock_tool_result_type
+
+    raise NotImplementedError(
+        f"Pydantic AI builtin tool '{pydantic_tool_name}' is not supported yet. If it should be, please file an issue."
```
dsfaccini: these four methods seem a little bit verbose, do you think the above syntax could make it easier to inline some of these checks?

additionally, not sure we need the `map_pydantic_builtin_tool_name_by_result_type`, if we're receiving a `result_type` it's arguably because the user called a tool that's support by Pydantic AI; is there a realistic situation where we would need to handle a `result_type` for a `BuiltinTool` we don't support? Otherwise we can inline this without the `NotImplementedError`

pydantic_ai_slim/pydantic_ai/models/bedrock.py
```diff
@@ -527,12 +611,18 @@ def _map_inference_config(
...
+        for tool in model_request_parameters.builtin_tools:
+            tools.append({'systemTool': {'name': map_bedrock_builtin_tool_name(tool.kind)}})
+
     def _map_tool_config(
         self,
         model_request_parameters: ModelRequestParameters,
         model_settings: BedrockModelSettings | None,
     ) -> ToolConfigurationTypeDef | None:
         tools = self._get_tools(model_request_parameters)
+        self._add_builtin_tools(tools, model_request_parameters)
```
dsfaccini: this one can definitely be inlined unless we need to use it somewhere else

pydantic_ai_slim/pydantic_ai/models/bedrock.py
```diff
@@ -925,18 +1037,44 @@ async def _get_event_iterator(self) -> AsyncIterator[ModelResponseStreamEvent]:
...
-                            args=None,
+                        if tool_use_start.get('type') == 'server_tool_use':
+                            part = BuiltinToolCallPart(
+                                tool_name=map_pydantic_builtin_tool_name(tool_name),
+                                tool_call_id=tool_id,
+                                provider_name=self.provider_name,
+                            )
+                            yield self._parts_manager.handle_part(vendor_part_id=index, part=part)
+                        else:
+                            maybe_event = self._parts_manager.handle_tool_call_delta(
```
dsfaccini: `if maybe_event := self._parts_manager.handle_tool_call_delta(...): yield maybe_event` walruses are amazing

pydantic_ai_slim/pydantic_ai/models/bedrock.py
dsfaccini: also not sure if we're doing this in some other model but `map_bedrock_builtin_tool_name(pydantic_tool_name: str)` is a bit weird from the naming, I'd expect us to call this like `map_bedrock_builtin_tool_name(bedrock_builtin_tool)` , so semantically something like `bedrock_builtin_tool = _get_bedrock_builtin_tool(tool.kind)` reads more natural, where `_get_bedrock_builtin_tool` returns `_BedrockBuiltinTool`

pydantic_ai_slim/pydantic_ai/models/bedrock.py
```diff
@@ -398,13 +461,34 @@ async def _process_response(self, response: ConverseResponseTypeDef) -> ModelRes
                 if text := item.get('text'):
                     items.append(TextPart(content=text))
                 elif tool_use := item.get('toolUse'):
+                    if tool_use.get('type') == 'server_tool_use':
+                        items.append(
+                            BuiltinToolCallPart(
+                                provider_name=self.system,
+                                tool_name=map_pydantic_builtin_tool_name(tool_use['name']),
+                                args=tool_use.get('input'),
+                                tool_call_id=tool_use.get('toolUseId'),
```
dsfaccini: we're using `tool_use['input']` so, if possible, lets keep these consistent. If this is the proper way because the keys might not exist then disregard this comment.

tests/models/test_bedrock.py
```diff
@@ -2304,3 +2317,548 @@ async def test_bedrock_empty_model_response_skipped(bedrock_provider: BedrockPro
...
+                    )
+                ],
+                timestamp=IsDatetime(),
+                run_id=IsStr(),
+            ),
+        ]
+    )
+
+
+@pytest.mark.filterwarnings('ignore::pytest.PytestUnraisableExceptionWarning')
```
dsfaccini: why is this raising?

pydantic_ai_slim/pydantic_ai/providers/bedrock.py
dsfaccini: yeahp, this is a bit verbose though, maybe we can use a

```python
def _without_builtin_tools(profile: ModelProfile | None) -> ModelProfile:
    return replace(profile or ModelProfile(), supported_builtin_tools=frozenset())

'mistral': lambda model_name: BedrockModelProfile(bedrock_tool_result_format='json')
    .update(_without_builtin_tools(mistral_model_profile(model_name))),
```

gmetzker:
> hey @gmetzker thanks a lot for working on this. I've made a couple of comments but they're mostly on syntax, the implementation looks good to me so far, @DouweM will take a look when he's back to give this the green light

Thanks for the feedback.  I'll work through these and send a few commits.

pydantic_ai_slim/pydantic_ai/models/bedrock.py
gmetzker: Ok.  I've removed the mapping methods and changed to direct dict usage.  Its still a lot of dicts but I'm thinking it's maybe better than a ton of repeated conditionals everywhere.

pydantic_ai_slim/pydantic_ai/models/bedrock.py
gmetzker: Done

pydantic_ai_slim/pydantic_ai/models/bedrock.py
gmetzker: Ok.  I removed the add_builtins method, and moved it into _map_tool_config.  I was kind of thinking the exception might be nice so left it.

pydantic_ai_slim/pydantic_ai/models/bedrock.py
gmetzker: Done

pydantic_ai_slim/pydantic_ai/providers/bedrock.py
gmetzker: Updated per your suggestion.  Hopefully it's more concise.  The `cohere` mapping was returning null, but I still wrapped it in `_without_builtin_tools` so it will have a non-null ModelProfile with `supported_builtin_tools=frozenset()`.  I assume thats what we want.  I wasn't sure if there was any significance to having a null `ModelProfile `returned vs `BedrockModelProfile` with only `supported_builtin_tools` set.

docs/models/bedrock.md
gmetzker: I removed that, and updated the examples file so it should verify correctly now.

tests/models/test_bedrock.py
gmetzker: I can't 100% remember.  I think I was having some issue when I was running VCR but it seems resolved now.  I removed `pytest.mark.filterwarnings` and tests are passing.

gmetzker:
@dsfaccini I believe I've addressed the issues.  So I think this is ready when you are free to check it out.

dsfaccini:
hey @gmetzker I took a look and it LGTM, only thing I saw is we generally prefer snapshots to multi-line asserts in tests, but in this case I thought the asserts were appropriate. I've notified @DouweM so he can give you his opinion or mark this for merging.

Thanks for being on top of it!

docs/models/bedrock.md
```diff
@@ -243,6 +243,31 @@ print(result.output)
 - Additional `CachePoint` markers in messages are removed from oldest to newest when the limit is exceeded
 - This ensures critical caching (instructions/tools) is maintained while still benefiting from message-level caching
 
+## Built-in tools
```
DouweM: I'd rather drop this section; we don't have one on the other model pages that support builtin tools either. (except OpenAI Responses, but that was because we supported their builtin tools before we had the native feature)

docs/models/bedrock.md
```diff
@@ -243,6 +243,31 @@ print(result.output)
...
+
+Nova 2.0 models support built-in tools; see the [Nova 2.0 documentation](https://docs.aws.amazon.com/nova/latest/nova2-userguide/using-tools.html#builtin-tools) for details.
+
+| Tool | Pydantic-AI | Bedrock | Notes |
+|------|-------------------|----------|-------|
+| [Code Interpreter](https://docs.aws.amazon.com/nova/latest/nova2-userguide/using-tools.html#builtin-tools) | `CodeExecutionTool` | `nova_code_interpreter` | Execute Python code in isolated sandbox environments |
+| [Web Grounding](https://docs.aws.amazon.com/nova/latest/nova2-userguide/web-grounding.html) | N/A | `nova_grounding` | **Not implemented**: Access real-time information from the internet |
```
DouweM: Are you interested in working on this next? :)

pydantic_ai_slim/pydantic_ai/models/bedrock.py
```diff
@@ -398,13 +420,34 @@ async def _process_response(self, response: ConverseResponseTypeDef) -> ModelRes
                 if text := item.get('text'):
                     items.append(TextPart(content=text))
                 elif tool_use := item.get('toolUse'):
+                    if tool_use.get('type') == 'server_tool_use':
+                        items.append(
+                            BuiltinToolCallPart(
+                                provider_name=self.system,
+                                tool_name=_BEDROCK_TO_BUILTIN[tool_use['name']],
```
DouweM: I'd rather drop the constants and inline the values here, like we have in other models (specific code for each builtin tool), as we will likely want to map the input args etc (at some point: https://github.com/pydantic/pydantic-ai/issues/3561)

pydantic_ai_slim/pydantic_ai/models/bedrock.py
```diff
@@ -398,13 +420,34 @@ async def _process_response(self, response: ConverseResponseTypeDef) -> ModelRes
...
-                            tool_name=tool_use['name'],
-                            args=tool_use['input'],
-                            tool_call_id=tool_use['toolUseId'],
-                        ),
+                        BuiltinToolReturnPart(
+                            provider_name=self.system,
+                            tool_name=_RESULT_TYPE_TO_BUILTIN_TOOL[tool_result.get('type', '')],
+                            content=tool_result.get('content'),
+                            tool_call_id=tool_result.get('toolUseId'),
+                            provider_details={'status': tool_result['status']} if 'status' in tool_result else {},
```
DouweM: When would there be no status key?

pydantic_ai_slim/pydantic_ai/models/bedrock.py
```diff
@@ -533,6 +576,14 @@ def _map_tool_config(
         model_settings: BedrockModelSettings | None,
     ) -> ToolConfigurationTypeDef | None:
         tools = self._get_tools(model_request_parameters)
+        for tool in model_request_parameters.builtin_tools:
+            if bedrock_tool := _BUILTIN_TOOL_TO_BEDROCK.get(tool.kind):
```
DouweM: As above, I'd prefer to have the values inlined here.

I imagine that the converse API could add support for anthropic code interpreter / web search at some point, and we would need a new `systemTool` value depending on the model type etc.

pydantic_ai_slim/pydantic_ai/models/bedrock.py
```diff
@@ -925,18 +998,42 @@ async def _get_event_iterator(self) -> AsyncIterator[ModelResponseStreamEvent]:
...
-                        if maybe_event:  # pragma: no branch
+                        ):  # pragma: no branch
                             yield maybe_event
+                    elif 'toolResult' in start:  # pragma: no branch
+                        tool_result_start = start['toolResult']
+                        tool_id = tool_result_start['toolUseId']
+
+                        return_part = BuiltinToolReturnPart(
+                            provider_name=self.provider_name,
+                            tool_name=_RESULT_TYPE_TO_BUILTIN_TOOL[tool_result_start.get('type', '')],
```
DouweM: this will fail if there's no type key, right? because `''` won't be a key in the constant dict. So probably we should verify there's. type in the `elif` above

pydantic_ai_slim/pydantic_ai/models/bedrock.py
```diff
@@ -967,11 +1064,25 @@ async def _get_event_iterator(self) -> AsyncIterator[ModelResponseStreamEvent]:
                             vendor_part_id=index,
                             tool_name=tool_use.get('name'),
                             args=tool_use.get('input'),
-                            tool_call_id=tool_id,
+                            tool_call_id=tool_ids.get(index),
```
DouweM: is it a problem that this could return `None`?

pydantic_ai_slim/pydantic_ai/models/bedrock.py
```diff
@@ -967,11 +1064,25 @@ async def _get_event_iterator(self) -> AsyncIterator[ModelResponseStreamEvent]:
...
                         )
                         if maybe_event:  # pragma: no branch
                             yield maybe_event
-                case _:
+                    if 'toolResult' in delta:  # pragma: no branch
+                        if return_part := builtin_tool_returns.get(index):  # pragma: no branch
+                            return_part.content.extend(delta['toolResult'])
```
DouweM: Is every tool result we receive supposed to be a list item? If they're deltas, I'd expect them to be concatenated somehow

pydantic_ai_slim/pydantic_ai/models/bedrock.py
```diff
@@ -967,11 +1064,25 @@ async def _get_event_iterator(self) -> AsyncIterator[ModelResponseStreamEvent]:
...
+
+                            # Update state without emitting; we only emit once on contentBlockStop.
```
DouweM: Would it make sense to not pass it into the parts manager at all yet until we yield the final event? This feels like the parts manager and event stream are getting out of sync

pydantic_ai_slim/pydantic_ai/providers/bedrock.py
```diff
@@ -47,11 +48,17 @@ class BedrockModelProfile(ModelProfile):
 def bedrock_amazon_model_profile(model_name: str) -> ModelProfile | None:
     """Get the model profile for an Amazon model used via Bedrock."""
     profile = amazon_model_profile(model_name)
+    assert profile is not None
+    profile.supported_builtin_tools = frozenset()
```
DouweM: Let's make this an `else` branch that uses the same `_without_builtin_tools` helper?

tests/models/test_bedrock.py
```diff
@@ -2304,3 +2317,547 @@ async def test_bedrock_empty_model_response_skipped(bedrock_provider: BedrockPro
             {'role': 'user', 'content': [{'text': 'Hello'}, {'text': 'Follow up question'}]},
         ]
     )
+
+
+async def test_bedrock_map_tool_config_with_only_builtin_code_execution_tool(bedrock_provider: BedrockProvider):
```
DouweM: I think we can drop this test and the next two, as the lines in question will be covered in the test that actually uses the feature

tests/models/test_bedrock.py
```diff
@@ -2304,3 +2317,547 @@ async def test_bedrock_empty_model_response_skipped(bedrock_provider: BedrockPro
...
+        messages,
+        ModelRequestParameters(
+            function_tools=[],
+            allow_text_output=True,
+            builtin_tools=[CodeExecutionTool()],
+        ),
+        None,
+    )
+
+    assert len(bedrock_messages) == 1
```
DouweM: Please use `snapshot` so we get to see the entire structure

tests/models/test_bedrock.py
```diff
@@ -2304,3 +2317,547 @@ async def test_bedrock_empty_model_response_skipped(bedrock_provider: BedrockPro
...
+                parts=[
+                    BuiltinToolCallPart(
+                        tool_name='code_execution',
+                        args={'snippet': '1234 * 5678'},
+                        tool_call_id='tooluse_STF3ouy6vJ3qPvDgwnauKw',
+                        provider_name='bedrock',
+                    ),
+                    BuiltinToolReturnPart(
+                        tool_name='code_execution',
+                        content=[
```
DouweM: Instead of a list, could the content just be the dict with `exitCode` key etc?

tests/models/test_bedrock.py
```diff
@@ -2304,3 +2317,547 @@ async def test_bedrock_empty_model_response_skipped(bedrock_provider: BedrockPro
...
+    agent = Agent(model=model, system_prompt='You are a helpful chatbot.', builtin_tools=[CodeExecutionTool()])
+
+    class Response(TypedDict):
+        result: float
+
+    # First turn
+    result1 = await agent.run('What is 1234 * 5678?', output_type=Response)
+    assert result1.output == snapshot({'result': 7006652.0})
+
+    # Second turn
```
DouweM, line 2510: Let's merge this test with the previous one, so that we can be sure that the first turn actually used the code interpreter (by having the snapshot visible).

tests/models/test_bedrock.py
```diff
@@ -2304,3 +2317,547 @@ async def test_bedrock_empty_model_response_skipped(bedrock_provider: BedrockPro
...
+    result2 = await agent.run('Now multiply that by 2', message_history=result1.new_messages(), output_type=Response)
+    assert result2.output == snapshot({'result': 14013304.0})
+
+    assert result2.all_messages() == snapshot(
```
DouweM: here we can then check only new_messages(), so that we separately test turn 1 and turn 2

tests/models/test_bedrock.py
```diff
@@ -2304,3 +2317,547 @@ async def test_bedrock_empty_model_response_skipped(bedrock_provider: BedrockPro
...
+                    ),
+                ],
+                timestamp=IsDatetime(),
+                run_id=IsStr(),
+            ),
+            ModelResponse(
+                parts=[
+                    BuiltinToolCallPart(
+                        tool_name='code_execution',
+                        args={'snippet': '1234 * 5678'},
```
DouweM: Can we use the `code` key for consistency with the other providers that implement this builtin tool?

tests/models/test_bedrock.py
```diff
@@ -2304,3 +2317,547 @@ async def test_bedrock_empty_model_response_skipped(bedrock_provider: BedrockPro
...
+                    )
+                ],
+                timestamp=IsDatetime(),
+                run_id=IsStr(),
+            ),
+        ]
+    )
+
+
+async def test_bedrock_model_stream_with_code_execution_tool_iter(
```
DouweM: I think e can merge the 2 streaming tests and test both the all_messages() + event_parts in one place, like we do on the other providers' tests.

docs/models/bedrock.md
gmetzker: Done

docs/models/bedrock.md
gmetzker: Possibly.  I've played around with it a bit but when I looked at it I wasn't entirely sure how to map some of the output.  I probably need to look over the other providers and then open a discussion.

pydantic_ai_slim/pydantic_ai/models/bedrock.py
gmetzker: Ok.  All of those constant dicts are dropped, and I inlined the conditionals into _process_response, _map_messages, etc...

pydantic_ai_slim/pydantic_ai/models/bedrock.py
gmetzker: The typedefs have `status` as not required as does the [AWS documentation](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ToolResultBlock.html).  

The document
> "This field is only supported by Amazon Nova and Anthropic Claude 3 and 4 models.".  

So I was assuming its possible for the key to be absent, in which case wouldn't it be better to leave it unassigned vs `{status=None}`?

I've never seen it missing from the block for Nova2 tests. So I can change it to `provider_details={'status': tool_result.get('status')}` if you think thats cleaner.  It just means, blocks that don't have status will be `{'status': None}`

Let me know what you think and I'll update it in the other locations too.

pydantic_ai_slim/pydantic_ai/models/bedrock.py
gmetzker: Done

pydantic_ai_slim/pydantic_ai/models/bedrock.py
gmetzker: Ok.  Moved `tool_result_start.get('type')` as an inline conditional so should be good now.

pydantic_ai_slim/pydantic_ai/models/bedrock.py
gmetzker: Logically a `contentBlockDelta` should always have a prior `contentBlockStart` in the stream.  So the index should always be present in `tool_ids`.  I'll just remove the get.

pydantic_ai_slim/pydantic_ai/models/bedrock.py
gmetzker: [`ContentBlockDelta.toolResult`](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ContentBlockDelta.html) is documented as an Array of [`ToolResultBlockDelta`](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ToolResultBlockDelta.html)

Documentation of `ToolResultBlockDelta` seems to incorrectly document it as `{"text": "..."}`.

But in practice, I've never seen that it's always something like:
 `{'json': {'stdOut': '7006652', 'stdErr': '', 'exitCode': 0, 'isError': False}}`


The `converse_stream` api has it as an array, something like this:
```json
{
  "contentBlockDelta": {
    "toolResult": [
      {
        "json": {
          "stdOut": "7006652",
          "stdErr": "",
          "exitCode": 0,
          "isError": false
        }
      }
    ]
  }
}
``` 

The `converse` api has this for the [`ToolResultBlock`](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ToolResultBlock.html)
```json
{
  "toolResult": {
    "toolUseId": "tooluse_djfIsBHNsCnRhFK1G-4rvg",
    "content": [
      {
        "json": {
          "stdOut": "7006652",
          "stdErr": "",
          "exitCode": 0,
          "isError": false
        }
      }
    ],
    "type": "nova_code_interpreter_result",
    "status": "success"
  }
}
```

So both apis, for sure have it as an **array**. 

I wasn't really sure what to stuff in `BuiltinToolReturnPart.content` and as it's typed as `Any` I figured the easiest thing to round trip it was exactly what Bedrock returned.

Should I be mapping this somehow or do we want to just keep what the provider sent?

Looking at `anthropic.py` [`_map_code_execution_tool_result_block`](https://github.com/pydantic/pydantic-ai/blob/46b09809e0a1c793d6f3917c36739453811526c5/pydantic_ai_slim/pydantic_ai/models/anthropic.py#L1394) it looks like it just dumps `BetaCodeExecutionToolResultBlockContent` directly with no transforming.
```json
{
    "content": [],
    "return_code": 0,
    "stderr": "",
    "stdout": "3 * 12390 = 37170\n",
    "type": "code_execution_result",
}
```

Looking at `google.py` [`_map_code_execution_result`](https://github.com/pydantic/pydantic-ai/blob/9543e74cf1f90d97c1631fb064969c51da303148/pydantic_ai_slim/pydantic_ai/models/google.py#L1181) it looks like it dumps a `CodeExecutionResult` with no transforming.  
```json
{"outcome": "OUTCOME_OK", "output": "-428330955.97745\n"}
```

From what I can tell on the other providers `BuiltinToolReturnPart.content` is provider dependent content.  So if thats the case shouldn't we just keep what Bedrock gives us, a `list` in this case?

pydantic_ai_slim/pydantic_ai/models/bedrock.py
gmetzker: Done. 

After removing those the tests are passing fine and it has the same event parts.  So I'll just go ahead skip the parts manager call until the bloc end.  At least for the tool return parts.

pydantic_ai_slim/pydantic_ai/providers/bedrock.py
gmetzker: I'm not 100% certain what you are after on this one but I made a few modifications.  something like this:
`profile = _without_builtin_tools(amazon_model_profile(model_name))`

Hopefully cleaner now...

tests/models/test_bedrock.py
gmetzker: Done

tests/models/test_bedrock.py
gmetzker: Done.

I think I added that before I figured out the snapshots.  Now that I see how it works, the snapshots are really cool.

tests/models/test_bedrock.py
gmetzker: [AWS Documentation for ToolResultBlock](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ToolResultBlock.html) has this as a list of objects, and I confirmed on my own testing as well as the snapshots.

See my earlier question.  I need to figure out if we want to map this differently than what Bedrock provides.  Other providers don't seem to have a normalization step, although granted most have an object for `content`

tests/models/test_bedrock.py
gmetzker: Done

tests/models/test_bedrock.py
gmetzker: Done

tests/models/test_bedrock.py
gmetzker: For the `BuiltinToolCallPart.args` should we be transforming it to some normalized structure?  The type on the property is `str|dict` so I figured the easiest method to round-trip it would be to use the provider format.

Looking at the other providers it's not consistent.  Sometimes its a string sometimes its an object, and sometimes it has extra properties beyond `code` (e.g openai has `container_id`, google has `language`).

Its easy enough to map to `snippet <--> code`, I just wasn't sure what to do given the other providers have different values here.

tests/models/test_bedrock.py
gmetzker: Done

gmetzker:
@DouweM Thanks for the feedback.  

I've inlined the name/type mapping and consolidated the test to use snapshots.  I still have some questions above on the mapping of `args` and `content`.

pydantic_ai_slim/pydantic_ai/models/bedrock.py
DouweM: Yeah I was mostly asking if this was defined as optional on the type, in which case this makes sense!

pydantic_ai_slim/pydantic_ai/models/bedrock.py
DouweM: We'll be working on unifying the schemas in https://github.com/pydantic/pydantic-ai/issues/3561 so this is not all that urgent, but when you look at the other builtin tools like `web_search`, you'll see that I do often do a bit of normalization to make the data easier to process by end users, if there are superfluous keys or something. 

In this case it looks like this `ToolResult` type matches the one that we have to build for agent-side tool results that we send to the Converse API, and because we could be sending one or more of JSON and files, we have to pass it as a list with a `json` key if we just have JSON. 

So in this case, even though the `nova_code_interpreter` tool itself returns just JSON, in the API it has to be wrapped in the list and the `json` key. But the real tool result is `content[0]['json']`, not the full `content`, so let's normalize it to that.

tests/models/test_bedrock.py
DouweM: responded above!

tests/models/test_bedrock.py
DouweM: Yeah it hasn't been very consistent, which we'll be fixing in https://github.com/pydantic/pydantic-ai/issues/3561 😅 So OK to leave it like this for now. I do feel strongly that we should normalize the result, as the list and `json` key are just artifacts of the generic API schema for tool result, not something relevant to the code interpreter tool.

DouweM:
@gmetzker I've left some comments; will review again once the normalization is applied!

pydantic_ai_slim/pydantic_ai/models/bedrock.py
gmetzker: The tool result blocks should be normalized to an object form now (without the list or extra `{"json": {}}` wrapper).  I didn't normalize any of the inner  stdOut/stdError keys though.

tests/models/test_bedrock.py
gmetzker: Fixed

tests/models/test_bedrock.py
gmetzker: This should be normalized to `{"code": "..."}` now.

pydantic_ai_slim/pydantic_ai/models/bedrock.py
```diff
@@ -966,12 +1066,35 @@ async def _get_event_iterator(self) -> AsyncIterator[ModelResponseStreamEvent]:
                         maybe_event = self._parts_manager.handle_tool_call_delta(
                             vendor_part_id=index,
                             tool_name=tool_use.get('name'),
-                            args=tool_use.get('input'),
-                            tool_call_id=tool_id,
+                            args=re.sub(r'"snippet"', '"code"', tool_use.get('input'))
+                            if tool_use.get('input')
+                            else None,
+                            tool_call_id=tool_ids[index],
```
gmetzker, line 1060: For tool_use deltas, `input` actually comes in as a string, so I suppose it could stream-in across multiple parts...?
Not sure if this is the best way to normalize but I didn't want to trust any json parsing or anything here.

Open to other ideas if the regex isn't protective enough.

gmetzker:
> @gmetzker I've left some comments; will review again once the normalization is applied!

@DouweM I believe I have it normalized now.  Take a look at the snapshots and see if its what you are after.

Thanks!

pydantic_ai_slim/pydantic_ai/models/bedrock.py
gmetzker: Kind of sketchy to do the `snippet/code` key swap here when the deltas are strings.  I could imagine a case where the inner code content also contained the term `"snippet"`.  I could replace on `{"snippet":` but still feels kind of brittle.

Alternatively, this could:
1. try to parse as JSON
2. swap snippet -> code
3. convert back to a string

But that wouldn't solve for the cases where you had a super larger code block that was split over multiple deltas.

DouweM: Let's revert the snippet->code rename; I wasn't very clear in https://github.com/pydantic/pydantic-ai/pull/3892#discussion_r2688446935; I definitely want the result to be normalized because `[0]['json']` is not part of the `nova_code_interpreter` spec (it's an artifact of the Bedrock API), but the `snippet` key was intentionally chosen, so let's keep that -- especially because the rename is brittle as you point out.

We can deal with normalizing this in https://github.com/pydantic/pydantic-ai/pull/3904

pydantic_ai_slim/pydantic_ai/models/bedrock.py
```diff
@@ -398,13 +409,41 @@ async def _process_response(self, response: ConverseResponseTypeDef) -> ModelRes
...
+                                tool_name=tool_use['name'],
+                                args=tool_use['input'],
+                                tool_call_id=tool_use['toolUseId'],
+                            ),
+                        )
+                elif tool_result := item.get('toolResult'):
+                    if tool_result.get('type') == 'nova_code_interpreter_result':  # pragma: no branch
+                        # Assume Code Exe results only have a single json item
+                        # so we can convert to object form
+                        if tr_content := tool_result['content'] or None:  # pragma: no branch
```
DouweM: `or None` is not needed here

pydantic_ai_slim/pydantic_ai/models/bedrock.py
```diff
@@ -966,12 +1066,35 @@ async def _get_event_iterator(self) -> AsyncIterator[ModelResponseStreamEvent]:
...
+                    if 'toolResult' in delta:  # pragma: no branch
+                        if (
+                            return_part := builtin_tool_returns.get(index)
+                        ) and return_part.tool_name == CodeExecutionTool.kind:  # pragma: no branch
+                            # For now, only process `contentBlockDelta.toolResult` for Code Exe tool.
+
+                            if (tr_content := delta['toolResult']) and 'json' in tr_content[0]:  # pragma: no branch
+                                # Goal here is to convert to object form.
+                                # This assumes the first item is the relevant one.
+                                return_part.content = tr_content[0]['json']
```
DouweM: Not a huge deal, but let's make this code consistent with the non-streaming one, and use `tr_content[0].get('json')` instead of `key in dict` + `dict[key]`

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/models/bedrock.py
gmetzker: That was just to make sure `tr_content` was assigned none when `{"content": []}` (instead of empty list).  Its probably never empty list anyway.  I went ahead and simplified with an inline conditional that is more straight-up.

pydantic_ai_slim/pydantic_ai/models/bedrock.py
gmetzker: My bad.  I misread your comment.  I reverted the transform on the ToolCallPart, but kept the object transform o the ToolReturnPart.

pydantic_ai_slim/pydantic_ai/models/bedrock.py
gmetzker: Done

DouweM:
@gmetzker Thanks Gary, great work!

PR #3894, opened 2026-01-01
feat: Add unified thinking settings across all model providers

### LINKED ISSUE #3756: Unified thinking settings
### Initial Checks

### Description

What is a little confusing is when using `AnthropicModelSettings` and setting `anthropic_thinking`, e.g.:

```Python
AnthropicModelSettings(
    anthropic_thinking=BetaThinkingConfigEnabledParam(
        type="enabled",
        budget_tokens=2048,
    ),
    anthropic_metadata={},
)
```

The extended thinking is properly recognized by Claude on VertexAI because it seems the practice is to use `AnthropicModel` but use a VertexAI Client with `AnthropicProvider`, but the thinking config is not recognized when using `BedrockConverseModel` and we have to set 

```Python
...
bedrock_additional_model_requests_field={
    "thinking": {
        "type": "enabled",
        "budget_tokens": 2048,
    },
    "anthropic_beta": ["interleaved-thinking-2025-05-14"],
},
...
```

This is not obvious or clear in AWS bedrock docs and could use some help from the `pydantic-ai` side.

### Example Code

```Python

```

```Text
Python 3.13
Pydantic 2.11
pydantic-ai 1.33.0
```

sarth6:
## Summary

This PR implements a provider-agnostic unified thinking/reasoning API in `ModelSettings` that automatically translates to each provider's native format. Two flat fields replace provider-specific configuration for portable cross-provider code.

Closes #3756

### Unified API

```python
from pydantic_ai import Agent

# Works across all providers — no provider-specific imports needed
agent = Agent('anthropic:claude-sonnet-4-0')
result = agent.run_sync(
    'Solve this step by step',
    model_settings={'thinking': True, 'thinking_effort': 'high'},
)
```

- `thinking: bool` — enable/disable thinking
- `thinking_effort: Literal['low', 'medium', 'high']` — control thinking depth

Settings are silently ignored on models that don't support them, so the same code works when switching providers.

### Provider Translation

| Provider | `thinking=True, effort='high'` translates to |
|----------|----------------------------------------------|
| Anthropic (adaptive) | `thinking={'type': 'adaptive'}` + `output_config.effort='high'` |
| Anthropic (budget) | `thinking={'type': 'enabled', 'budget_tokens': 16384}` |
| Google (Gemini 3) | `thinking_config={'thinking_level': HIGH}` |
| Google (Gemini 2.5) | `thinking_config={'thinking_budget': 16384}` |
| OpenAI | `reasoning_effort='high'` |
| Bedrock (Claude) | `thinking={'type': 'enabled', 'budget_tokens': 16384}` |
| OpenRouter | `reasoning={'effort': 'high'}` |
| Groq | `reasoning_format='parsed'` (effort ignored) |
| Cerebras | `reasoning=True` (effort ignored) |
| Cohere | Enables thinking mode (effort ignored) |
| xAI | `reasoning_effort='high'` (`'medium'` maps to `'low'`) |

### Design

**Three-layer architecture:**
1. `ModelSettings` — user-facing flat fields (`thinking`, `thinking_effort`)
2. `resolve_thinking_config()` — pure normalization to `ResolvedThinkingConfig` (no validation)
3. `Model._resolve_*()` — per-provider translation with silent-drop for unsupported settings

**Key behaviors:**
- Provider-specific settings (e.g. `anthropic_thinking`) always take precedence
- Silent-drop semantics: unsupported settings are ignored (no errors), enabling portable code
- `thinking_effort` without `thinking` implicitly enables thinking
- `thinking=False` ignores effort (precedence rule)

### Profile enhancements

Two new capability flags in `ModelProfile`:
- `supports_thinking` — auto-detected from model names
- `thinking_always_enabled` — for models that can't disable thinking (o-series, DeepSeek R1, Magistral)

## Test plan

- [x] 100 tests (93 unit + 7 integration) covering all providers
- [x] Integration tests use mock clients (MockAnthropic, MockOpenAI, MockOpenAIResponses) to verify correct native API params through the full Agent pipeline
- [x] Tests cover: enable/disable, all effort levels, provider precedence, silent-drop, always-on models, profile detection, settings merge
- [x] Pyright type checking passes with 0 errors
- [x] Documentation updated (`docs/thinking.md`) with unified API section

## Files changed

**Core:**
- `settings.py` — Added `thinking: bool` and `thinking_effort` fields
- `thinking.py` — `resolve_thinking_config()` normalization (57 lines)
- `profiles/__init__.py` — Added `supports_thinking`, `thinking_always_enabled`
- `__init__.py` — Removed old `ThinkingConfig` export

**Models (provider translation):**
- `anthropic.py`, `google.py`, `openai.py`, `bedrock.py`, `openrouter.py`, `groq.py`, `cerebras.py`, `cohere.py`, `xai.py`

**Profiles (thinking detection):**
- `anthropic.py`, `google.py`, `cohere.py`, `grok.py`, `mistral.py`

**Docs:**
- `docs/thinking.md` — Added unified API section

**Tests:**
- `tests/test_unified_thinking.py` — 100 tests (unit + integration)
- `tests/models/test_bedrock.py` — Updated for new API

### DIFF
--- a/REFACTOR_PLAN.md
+++ b/REFACTOR_PLAN.md
@@ -1,0 +1,346 @@
+# Thinking Resolution Architecture Refactor
+
+> **Context**: DouweM's review on PR #3894 raised concerns that each model's `_resolve_thinking_config` carries too much knowledge about the thinking abstraction. This plan proposes a refactoring modeled on how pydantic-ai already handles its most successful cross-cutting abstraction: **output mode**.
+
+---
+
+## 1. The Established Patterns
+
+Before proposing changes, here's how pydantic-ai handles existing cross-cutting concerns. These patterns reveal clear maintainer preferences:
+
+### Output mode — the gold standard (profile-driven, base-class-resolved)
+
+```
+Profile declares:     supports_json_schema_output, default_structured_output_mode
+Base class resolves:  'auto' → concrete mode, clears irrelevant fields, validates
+Result stored in:     ModelRequestParameters.output_mode (resolved before models see it)
+Models translate:     Read params.output_mode at request time → native API format
+```
+
+**No model overrides `prepare_request()` for output mode.** The base class does 100% of resolution. Models just read the resolved value and translate.
+
+### Builtin tools — validation-only centralization
+
+```
+Profile declares:     supported_builtin_tools (frozenset of tool types)
+Model declares:       supported_builtin_tools() classmethod (what's implemented)
+Base class validates: Intersection of both, raises UserError for unsupported
+Models translate:     _add_builtin_tools() maps to native API format at request time
+```
+
+**Base class validates. Models translate.** No resolution hooks needed.
+
+### Basic settings (max_tokens, temperature, etc.) — direct pass-through
+
+```
+Declared in:          ModelSettings TypedDict
+Merged in:            Model.prepare_request() via merge_model_settings()
+Consumed:             Each model reads model_settings.get('field') at request time
+Translation:          Inline name mapping (e.g., stop_sequences → stop)
+```
+
+**No resolution layer.** Direct read-through with inline name translation.
+
+### Current thinking implementation — the outlier
+
+```
+Declared in:          ModelSettings (thinking, thinking_effort)
+Normalized in:        Each model calls resolve_thinking_config() itself
+Injected in:          Each model's prepare_request() override (9x boilerplate)
+Consumed:             model_settings.get('provider_thinking_field') at request time
+Strip pattern:        OpenAI subclasses strip parent's injection (Cerebras, OpenRouter)
+```
+
+**Every model overrides `prepare_request()` just for thinking.** This violates the patterns above — the base class should resolve, not each model.
+
+---
+
+## 2. The Refactoring: Follow the Output Mode Pattern
+
+### Core change: Add `resolved_thinking` to `ModelRequestParameters`
+
+```python
+@dataclass(repr=False, kw_only=True)
+class ModelRequestParameters:
+    # ... existing fields ...
+    output_mode: OutputMode = 'text'
+    output_object: OutputObjectDefinition | None = None
+    # ...
+
+    # NEW: Resolved thinking config (follows output_mode pattern)
+    resolved_thinking: ResolvedThinkingConfig | None = None
+```
+
+### Base class resolves thinking in `prepare_request()`
+
+```python
+# models/__init__.py — Model class
+
+class Model(ABC):
+    # ... existing methods ...
+
+    def _translate_thinking(self, resolved: ResolvedThinkingConfig) -> tuple[str, Any] | None:
+        """Translate resolved thinking config to provider-native format.
+
+        Override in models whose profile has supports_thinking=True.
+        Returns (provider_settings_key, native_value) or None.
+        The default returns None (thinking silently ignored).
+        """
+        return None
+
+    def prepare_request(self, model_settings, model_request_parameters):
+        model_settings = merge_model_settings(self.settings, model_settings)
+        params = self.customize_request_parameters(model_request_parameters)
+
+        # ... existing output mode resolution ...
+
+        # NEW: Resolve unified thinking settings (mirrors output mode resolution)
+        if model_settings:
+            resolved_thinking = resolve_thinking_config(model_settings, self.profile)
+            if resolved_thinking is not None:
+                params = replace(params, resolved_thinking=resolved_thinking)
+
+        return model_settings, params
+```
+
+The default `_translate_thinking()` returns `None`, following the same convention as `customize_request_parameters()` (default is a pass-through) and `supported_builtin_tools()` (default is empty frozenset). Non-thinking models don't need to override it.
+
+This mirrors how output mode resolution works:
+- Profile capabilities (`supports_thinking`, `thinking_always_enabled`) drive the guards
+- The canonical result is stored in `ModelRequestParameters`
+- Models receive an already-resolved config — they never call `resolve_thinking_config()` themselves
+
+### Models translate at request time — no `prepare_request()` overrides needed for thinking
+
+Each model reads `model_request_parameters.resolved_thinking` where it currently reads the provider-specific field, with provider-specific settings taking precedence:
+
+```python
+# Anthropic — in _messages_create():
+# BEFORE:
+#   thinking=model_settings.get('anthropic_thinking', OMIT),
+# AFTER:
+if 'anthropic_thinking' in model_settings:
+    thinking_config = model_settings['anthropic_thinking']
+elif model_request_parameters.resolved_thinking is not None:
+    thinking_config = self._translate_thinking(model_request_parameters.resolved_thinking)
+else:
+    thinking_config = OMIT
+```
+
+The `_translate_thinking()` method is a pure translator — no resolution, no profile checks:
+
+```python
+def _translate_thinking(self, resolved: ResolvedThinkingConfig) -> BetaThinkingConfigParam:
+    """Translate canonical thinking config to Anthropic native format."""
+    if not resolved.enabled:
+        return {'type': 'disabled'}
+    if AnthropicModelProfile.from_profile(self.profile).anthropic_supports_adaptive_thinking:
+        return {'type': 'adaptive'}
+    budget = EFFORT_TO_BUDGET.get(resolved.effort, DEFAULT_THINKING_BUDGET) if resolved.effort else DEFAULT_THINKING_BUDGET
+    return {'type': 'enabled', 'budget_tokens': budget}
+```
+
+---
+
+## 3. Per-Provider Changes
+
+### Simple providers — `prepare_request()` override eliminated entirely
+
+| Provider | Current | After |
+|---|---|---|
+| **Groq** | Overrides `prepare_request()` only for thinking | Override **removed** |
+| **Cohere** | Overrides `prepare_request()` only for thinking | Override **removed** |
+| **xAI** | Overrides `prepare_request()` only for thinking | Override **removed** |
+
+**Groq example (at request time):**
+
+```python
+# In _chat_completions_create():
+if 'groq_reasoning_format' in model_settings:
+    reasoning_format = model_settings['groq_reasoning_format']
+elif model_request_parameters.resolved_thinking is not None:
+    reasoning_format = self._translate_thinking(model_request_parameters.resolved_thinking)
+else:
+    reasoning_format = NOT_GIVEN
+```
+
+### OpenAI inheritance chain — strip pattern eliminated
+
+This is the biggest structural win. Currently:
+1. `OpenAIChatModel.prepare_request()` resolves thinking → injects `openai_reasoning_effort`
+2. `CerebrasModel.prepare_request()` **strips** `openai_reasoning_effort`, injects own
+3. `OpenRouterModel.prepare_request()` **strips** `openai_reasoning_effort`, injects own
+
+After the refactor:
+1. Base `Model.prepare_request()` resolves thinking → `params.resolved_thinking`
+2. `OpenAIChatModel._completions_create()` reads `params.resolved_thinking` at request time
+3. `CerebrasModel` and `OpenRouterModel` read `params.resolved_thinking` at request time
+4. **No injection, no stripping, no hidden parent dependency**
+
+### Anthropic — conflict check uses `resolved_thinking` directly
+
+Anthropic's `prepare_request()` currently checks thinking-vs-output-tools conflict and can mutate `output_mode` (from `'auto'` to `'native'`/`'prompted'`). This mutation must happen **before** the base class resolves output mode, so Anthropic's conflict check runs first:
+
+```python
+# AnthropicModel.prepare_request():
+
+# Merge settings early to check thinking state for the conflict check
+merged_settings = cast(AnthropicModelSettings, merge_model_settings(self.settings, model_settings) or {})
+
+# Determine if thinking is enabled (from provider-specific or unified settings)
+thinking_enabled = False
+if 'anthropic_thinking' in merged_settings:
+    thinking_enabled = merged_settings['anthropic_thinking'].get('type') in ('enabled', 'adaptive')
+else:
+    resolved = resolve_thinking_config(merged_settings, self.profile)
+    thinking_enabled = resolved is not None and resolved.enabled
+
+# Thinking-vs-output conflict check — may mutate output_mode before super() resolves it
+if thinking_enabled and model_request_parameters.output_tools:
+    if model_request_parameters.output_mode == 'auto':
+        output_mode = 'native' if self.profile.supports_json_schema_output else 'prompted'
+        model_request_parameters = replace(model_request_parameters, output_mode=output_mode)
+    elif not model_request_parameters.allow_text_output:
+        suggested = 'NativeOutput' if self.profile.supports_json_schema_output else 'PromptedOutput'
+        raise UserError(f'Anthropic does not support thinking and output tools...')
+
+# Now call super() — base class resolves output mode and thinking
+return super().prepare_request(merged_settings, model_request_parameters)
+```
+
+Note: Anthropic calls `merge_model_settings` itself and passes the merged result to `super()` (same pattern as today). The base class's `prepare_request` detects the already-merged settings and resolves thinking from them.
+
+Anthropic's `_build_output_config` reads effort from `params.resolved_thinking.effort` instead of a forwarded `anthropic_effort` field — no more effort forwarding. This requires passing `model_request_parameters` to `_build_output_config` (currently it only receives `model_settings`).
+
+### OpenRouter — set `supports_thinking=True` universally
+
+OpenRouter delegates capability detection to its backend, so its profiles should set `supports_thinking=True` for all models. This lets the base class resolve thinking normally — no re-resolution override needed. OpenRouter's `prepare_request()` still strips `openai_reasoning_effort` (injected by the OpenAI parent) and converts settings, but no longer needs thinking-specific logic:
+
+```python
+# OpenRouterModel.prepare_request():
+merged_settings, params = super().prepare_request(model_settings, model_request_parameters)
+merged_settings = cast(OpenRouterModelSettings, merged_settings or {})
+
+# Strip parent's openai_reasoning_effort (OpenRouter uses its own reasoning format)
+merged_settings.pop('openai_reasoning_effort', None)
+
+new_settings = _openrouter_settings_to_openai_settings(merged_settings)
+return new_settings, params
+```
+
+### Bedrock — reads at request time with model-family dispatch
+
+```python
+# BedrockConverseModel — at request time:
+if model_request_parameters.resolved_thinking is not None:
+    key, config = self._translate_bedrock_thinking(model_request_parameters.resolved_thinking)
+    if key and config:
+        additional[key] = config
+```
+
+---
+
+## 4. Why This Pattern Works
+
+### Follows the output mode precedent exactly
+
+| Aspect | Output mode | Thinking (proposed) |
+|---|---|---|
+| Profile capabilities | `supports_json_schema_output` | `supports_thinking`, `thinking_always_enabled` |
+| Resolved in | Base `Model.prepare_request()` | Base `Model.prepare_request()` |
+| Stored in | `ModelRequestParameters.output_mode` | `ModelRequestParameters.resolved_thinking` |
+| Models call resolver? | No | No |
+| Models override `prepare_request` for it? | No | No (except Anthropic for conflict check, OpenRouter for no-profile) |
+| Translation | Inline at request time | `_translate_thinking()` at request time |
+
+### Eliminates all three structural problems
+
+1. **9x boilerplate in `prepare_request()`** — gone. Base class resolves once.
+2. **Strip-and-reinject in OpenAI chain** — gone. No injection into `model_settings`.
+3. **Each model calling `resolve_thinking_config()` itself** — gone. Base class calls it.
+
+### What each model knows (before vs. after)
+
+**Before**: Each model knows how to resolve unified settings, when to apply profile guards, how to check provider-specific precedence, AND how to translate to native format.
+
+**After**: Each model knows how to translate `ResolvedThinkingConfig` to native format, and which provider-specific field takes precedence. That's it.
+
+---
+
+## 5. What This Does NOT Change
+
+- `resolve_thinking_config()` in `thinking.py` — stays as the centralized normalizer
+- `ResolvedThinkingConfig` dataclass — stays as the canonical form
+- Profile fields (`supports_thinking`, `thinking_always_enabled`) — stay on `ModelProfile`
+- Provider-specific settings (`anthropic_thinking`, `openai_reasoning_effort`, etc.) — still exist and take precedence
+- Silent-drop semantics — unchanged
+- Provider-specific profile extensions — unchanged
+
+---
+
+## 6. Naming Summary
+
+| Component | Name | Location |
+|---|---|---|
+| Centralized normalizer | `resolve_thinking_config()` | `thinking.py` (unchanged) |
+| Canonical representation | `ResolvedThinkingConfig` | `thinking.py` (unchanged) |
+| Resolved result on params | `resolved_thinking` | `ModelRequestParameters` (new field) |
+| Per-model translator | `_translate_thinking()` | Each model class (renamed from `_resolve_thinking_config`) |
+
+---
+
+## 7. Discoverability: Adding a New Thinking-Capable Provider
+
+Under this pattern, a new provider developer:
+
+1. Sets `supports_thinking=True` in their profile — the base class will then resolve thinking settings
+2. Overrides `_translate_thinking()` to map `ResolvedThinkingConfig` → provider-native format
+3. Reads `model_request_parameters.resolved_thinking` at request time, with provider-specific settings taking precedence
+4. If they forget steps 2-3, thinking settings are silently ignored (correct behavior — same as forgetting to read `output_mode`)
+
+The base `Model._translate_thinking()` returns `None` by default, so non-thinking models don't need to do anything. This follows the `customize_request_parameters` and `supported_builtin_tools` convention: optional hooks with sensible defaults, no abstract method enforcement for optional capabilities.
+
+`models/AGENTS.md` documents the pattern with a checklist.
+
+---
+
+## 8. Migration Checklist
+
+### Phase 1: Infrastructure
+- [ ] Add `resolved_thinking: ResolvedThinkingConfig | None = None` to `ModelRequestParameters`
+- [ ] Add thinking resolution to `Model.prepare_request()` (after output mode, before validation)
+
+### Phase 2: Migrate models (all can be done independently)
+For each of the 9 providers:
+- [ ] Add `_translate_thinking()` method (pure translator)
+- [ ] Move thinking consumption from `prepare_request()` injection → request-time read
+- [ ] Remove the `prepare_request()` thinking boilerplate
+- [ ] For Groq/Cohere/xAI: remove `prepare_request()` override entirely
+
+### Phase 3: Fix OpenAI inheritance chain
+- [ ] Remove thinking injection from `OpenAIChatModel.prepare_request()`
+- [ ] Remove `openai_reasoning_effort` stripping from CerebrasModel and OpenRouterModel
+- [ ] Each reads `params.resolved_thinking` at request time instead
+
+### Phase 4: Cleanup
+- [ ] Remove `anthropic_effort` forwarding from Anthropic's `prepare_request`
+- [ ] Update `_build_output_config` to read from `params.resolved_thinking.effort`
+- [ ] Rename all `_resolve_thinking_config` → `_translate_thinking`
+- [ ] Update `models/AGENTS.md` with new pattern documentation
+- [ ] Verify all tests pass
+
+---
+
+## 9. Design Decisions
+
+### `ModelRequestParameters.resolved_thinking` is public API
+
+`ModelRequestParameters` is already a public dataclass. `resolved_thinking` is read-only metadata analogous to `output_mode` — users who override `customize_request_parameters()` already interact with this dataclass. Hiding it would be inconsistent.
+
+### OpenRouter sets `supports_thinking=True` universally
+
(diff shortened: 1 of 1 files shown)


pydantic_ai_slim/pydantic_ai/models/_warnings.py
```diff
@@ -0,0 +1,169 @@
+"""Warning utilities for model providers.
+
+This module provides centralized warning helpers for emitting consistent
+warnings when unified settings don't translate directly to provider-specific
+configurations.
+"""
+
```
sarth6: As I began adding warnings everywhere I noticed a lot of duplicate logic, so made this utility. This should help reduce duplicate logic, and also standardize the help info we provide to users for all the different combinations of settings translation/support issues that can happen. let me know what you think. 👍

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -280,6 +281,55 @@ def supported_builtin_tools(cls) -> frozenset[type[AbstractBuiltinTool]]:
         """The set of builtin tool types this model can handle."""
         return frozenset({WebSearchTool, CodeExecutionTool, WebFetchTool, MemoryTool, MCPServerTool})
 
+    def _resolve_thinking_config(self, model_settings: AnthropicModelSettings) -> BetaThinkingConfigParam | None:
```
sarth6: Every model has its own _resolve_thinking_config method now, this is a design claude code came up with, it works but I'm not super happy with it:

* There is still too much burden on model code to understand how the thinking settings work (bool+dict, budget tokens vs effort, etc)
* This might lead to developer error over time as devs maintain models and add new models

I think the model profile should be responsible for declaring the model's capabilities, while the model settings subclass should look at the general thinking settings + the model profile and use it to provide simplified model-specific settings back to the model code. Then model code should pretty much be able to pass through those processed model-specific settings to the model API/client.

Here, the model code is just doing it all. I wanted to get this in review sooner or later but let me know your thoughts on this, I think there's more work for me to do here to improve maintainability.

pydantic_ai_slim/pydantic_ai/models/_warnings.py
```diff
@@ -0,0 +1,169 @@
+"""Warning utilities for model providers.
+
+This module provides centralized warning helpers for emitting consistent
+warnings when unified settings don't translate directly to provider-specific
+configurations.
+"""
```
sarth6: As I began adding warnings everywhere I noticed a lot of duplicate logic, so made this utility. This should help reduce duplicate logic, and also standardize the help info we provide to users for all the different combinations of settings translation/support issues that can happen. let me know what you think. 👍

pydantic_ai_slim/pydantic_ai/models/anthropic.py
sarth6: I think this is important enough I'm going to try to address it before asking for review

tests/test_model_warnings.py
```diff
@@ -0,0 +1,239 @@
+"""Tests for model warning utilities."""
```
sarth6: I could see the argument that some of warnings here should be errors instead, I wasn't sure what the philosophy is there so kept as warnings - happy to make that swap 👍

pydantic_ai_slim/pydantic_ai/models/_warnings.py
sarth6: Maybe there's a better place for the file though

Pavanmanikanta98:
Hey! @sarth6 
 I saw you raised the PR for this—no worries at all. I was actually working on the design for this in Obsidian (haven't even opened my code editor yet 😅) and decided to just review your PR instead.

I think I spotted a potential issue in `pydantic_ai/settings.py`. I noticed you didn't update `merge_model_settings`, but since you're adding `thinking` as a nested dictionary, the old `base | overrides` logic will now cause issues. It will completely replace the base `thinking` config with the override instead of merging them.

This technically affects existing fields like `extra_headers` too, but it's a bigger deal here because with `thinking`, users are much more likely to want to just "tweak" one value (like `budget_tokens`) at runtime while keeping the `enabled=True` flag from their Agent definition.

Updating that logic to handle shallow merging for dictionaries would ensure the base configuration is preserved when only specific fields are overridden.

sarth6:
@Pavanmanikanta98 thanks for the feedback! I was wondering if we'd want this, so good to know. I've update the PR - take a look at the `TestMergeModelSettingsThinking` tests I just added, they test various complex merge scenarios to ensure that settings are getting merged and not lost.

Pavanmanikanta98:
@sarth6  Awesome! The updated merge_model_settings logic and the new tests look perfect. Glad the feedback was helpful!

coccoinomane:
+1 for this 
I need to test my agentic workflow with arbitrary models, both thinking and non-thinking. I am ending up developing my own  `ThinkingConfig`-like abstraction, so it would be great if you ended up shipping it in Pydantic AI 
Thanks 🙏

pydantic_ai_slim/pydantic_ai/models/_warnings.py
DouweM: Let's move them to where `ModelSettings` is defined.

cc @dsfaccini as you were looking at adding some settings-being-ignored warnings as well

pydantic_ai_slim/pydantic_ai/models/anthropic.py
DouweM: @dsfaccini has been working on something similar in https://github.com/pydantic/pydantic-ai/pull/3611 for `tool_choice`.

Were you still planning to refactor here, or did that already happen?

pydantic_ai_slim/pydantic_ai/models/bedrock.py
```diff
@@ -303,6 +304,31 @@ def _map_tool_definition(f: ToolDefinition) -> ToolTypeDef:
 
         return {'toolSpec': tool_spec}
 
+    def _resolve_thinking_config(self, model_settings: BedrockModelSettings) -> dict[str, Any] | None:
+        """Resolve thinking configuration from unified settings for Bedrock.
+
+        For Claude models on Bedrock, thinking is passed via additionalModelRequestFields.
+        Returns the thinking config dict in Anthropic format, or None if not enabled.
```
DouweM: What if the current model is not Anthropic? Do we handle that properly?

DouweM: Do other models on Bedrock support thinking with different request fields?

pydantic_ai_slim/pydantic_ai/models/cerebras.py
```diff
@@ -92,9 +93,58 @@ def prepare_request(
...
+        if resolved is None:  # pragma: no cover
+            return None
+        if not resolved.enabled:
+            return True  # disable_reasoning=True
+
+        # Warn about ignored settings (Cerebras only supports enable/disable)
+        ignored: list[str] = []
+        for setting in ('budget_tokens', 'effort', 'include_in_response', 'summary'):
+            value = getattr(resolved, setting, None)
+            if value is not None and (setting != 'include_in_response' or value is not True):
```
DouweM: Could this be centralized in e.g. `Model.prepare_request` by checking on `profile` what fields are supported?

pydantic_ai_slim/pydantic_ai/models/cerebras.py
```diff
@@ -92,9 +93,58 @@ def prepare_request(
...
+                ignored.append(setting)
+
+        if ignored:
+            from ._warnings import warn_settings_ignored_batch
+
+            warn_settings_ignored_batch(
+                setting_names=ignored,
+                provider_name='Cerebras',
+                model_name=self._model_name,
+                reason='Cerebras reasoning can only be enabled or disabled. Reasoning will be enabled with default behavior',
```
DouweM: So far, we haven't warned for unsupported model settings fields because the `ModelSettings` field docstrings already state which models support it, and we want it to be possible to provide the "highest common denominator" of settings on the agent and then use it with multiple models. 

So, following that precedent, it's totally fine (i.e. not needing a warning) if you specify `enabled` + `budget`, and then a given model only uses `enabled`, as long as the docstring didn't claim that model supported `budget`.

Meaning that we should only warn if the docstring stated it was supported, but for some reason, we ignore it anyway (as in https://github.com/pydantic/pydantic-ai/issues/4011)

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -1489,13 +1508,93 @@ def _get_reasoning(self, model_settings: OpenAIResponsesModelSettings) -> Reason
...
+        reasoning_effort: ReasoningEffort | None,
+        reasoning_summary: Literal['detailed', 'concise', 'auto'] | None,
+    ) -> tuple[ReasoningEffort | None, Literal['detailed', 'concise', 'auto'] | None]:
+        """Apply unified thinking settings to reasoning effort and summary."""
+        thinking = model_settings.get('thinking')
+        if thinking is None:
+            return reasoning_effort, reasoning_summary
+
+        # Validate that the model supports thinking
+        if not self.profile.supports_thinking:
```
DouweM: I think this type of check should be in `Model.prepare_request`

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -1489,13 +1508,93 @@ def _get_reasoning(self, model_settings: OpenAIResponsesModelSettings) -> Reason
...
+            raise UserError(
+                f'Model {self.model_name!r} does not support thinking/reasoning. '
+                f'Remove the `thinking` setting or use a model that supports thinking.'
+            )
+
+        if thinking is False:
+            if self.profile.thinking_always_enabled:
+                raise UserError(
+                    f'Model {self.model_name!r} has reasoning always enabled and cannot be disabled. '
```
DouweM: Please consistently use thinking instead of "reasoning" unless we're specifically referring to the provider's name for the feature

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -1489,13 +1508,93 @@ def _get_reasoning(self, model_settings: OpenAIResponsesModelSettings) -> Reason
...
+                    f'Remove the `thinking=False` setting.'
+                )
+            return reasoning_effort, reasoning_summary
+
+        if thinking is True:
+            return reasoning_effort or 'medium', reasoning_summary
+
+        # Handle ThinkingConfig dict
+        config: ThinkingConfig = thinking
```
DouweM: shouldn't we be working off the resolved config?

pydantic_ai_slim/pydantic_ai/settings.py
```diff
@@ -1,9 +1,59 @@
...
+
+    Mapped to:
+    - Anthropic: budget_tokens
+    - Google: thinking_budget
+    - OpenAI: (not directly supported, use effort instead)
+    - Bedrock (Claude): budget_tokens
+    """
+
+    effort: Literal['low', 'medium', 'high']
+    """Thinking effort level. Alternative to budget_tokens.
```
DouweM: Since Gemini 3 and OpenAI both use effort instead of budget, I'd prefer for our unified interface to only support effort, not budget. If someone really wants to specify token budget specifically, they can use the provider-specific setting.

pydantic_ai_slim/pydantic_ai/settings.py
```diff
@@ -1,9 +1,59 @@
...
+    - Anthropic: budget_tokens (low=1024, medium=4096, high=16384)
+    - Google: thinking_budget (low=1024, medium=8192, high=32768)
+    """
+
+    include_in_response: bool
+    """Whether to include thinking content in the response. Defaults to True.
+
+    Mapped to:
+    - Google: include_thoughts
+    - Groq: reasoning_format ('parsed' vs 'hidden')
```
DouweM: For OpenAI, should this map to `openai_reasoning_summary`?

pydantic_ai_slim/pydantic_ai/settings.py
```diff
@@ -1,9 +1,59 @@
...
+    - Others: Always included when thinking is enabled
+    """
+
+    summary: Literal['none', 'concise', 'detailed', 'auto'] | bool
```
DouweM: Maybe this can be combined with `include_in_response`? Assuming that `include_thoughts` with Google also refers to including the _summaries_, while signatures are always included...?

pydantic_ai_slim/pydantic_ai/settings.py
```diff
@@ -1,9 +1,59 @@
...
+    """Request a summary of the thinking process.
+
+    Mapped to:
+    - OpenAI: reasoning_summary
```
DouweM: I'd prefer not to have dedicated fields for things supported by only one provider. If someone really wants to tweak this, they can use the OpenAI-specific dict.

(Of course if you can find one more provider that has this, we should list them too!)

pydantic_ai_slim/pydantic_ai/thinking.py
```diff
@@ -0,0 +1,114 @@
...
+                   or if thinking cannot be disabled for always-enabled models.
+    """
+    if thinking is None:
+        return None
+
+    # Validate model supports thinking
+    if not profile.supports_thinking:
+        raise UserError(
+            f'Model {model_name!r} does not support thinking/reasoning. '
+            f'Remove the `thinking` setting or use a model that supports thinking.'
```
DouweM: See above; we should not be raising errors if we never claimed in the model setting docstring that this model supported the setting anyway

In general, we prefer documenting lack of support + dropping silently over warnings or errors, because we want it to be very easy to use the same agent with different models.

pydantic_ai_slim/pydantic_ai/profiles/anthropic.py
```diff
@@ -16,7 +31,14 @@ def anthropic_model_profile(model_name: str) -> ModelProfile | None:
     # https://docs.claude.com/en/docs/build-with-claude/structured-outputs#example-usage
 
     supports_json_schema_output = model_name.startswith(models_that_support_json_schema_output)
+
+    # Check if model supports extended thinking
+    supports_thinking = model_name.startswith(_THINKING_MODELS)
```
DouweM: Instead of `startswith`, let's use `in` in all these types of checks (i.e. also in other model profile files), as some providers add a prefix ahead of the model name

pydantic_ai_slim/pydantic_ai/profiles/__init__.py
```diff
@@ -77,6 +77,22 @@ class ModelProfile:
...
+    """Whether thinking cannot be disabled (e.g., DeepSeek R1). If True, thinking=False raises UserError."""
+
+    supports_thinking_level: bool = False
+    """Whether the model uses thinking_level instead of thinking_budget (Gemini 3+)."""
+
+    default_thinking_budget: int | None = None
+    """Default thinking budget if not specified. For Anthropic, defaults to 4096."""
+
+    effort_to_budget_map: dict[str, int] | None = None
+    """Custom effort-to-budget mapping for this model (Gemini 2.5, Anthropic)."""
```
DouweM: - Are these mappings model-specific, or model _family_ specific, i.e. do we need this here, or could it be a constant dict in `AnthropicModel`, `GoogleModel`?
- Either way which fields we need here may change if we support only `effort` on the new setting

DouweM (changes requested):
@sarth6 Thanks for working on this Sarth! Note that I have yet to thoroughly review the code.

dsfaccini:

```bash
git pull

deactivate

rm -rf .venv

make install
```

sarth6:
@dsfaccini @DouweM I noticed anthropic is now moving way from token count based thinking settings to categorical based settings like openai - I think that will affect this PR a bit. When I pick this back up (or if someone beats me to it) we'll prob want to move further in that direction and we'll do another `PLAN.md` review before proceeding to implementation 👍

DouweM:
@sarth6 Makes sense, I'm eager to move forward with this one btw!

sarth6:
Hey @DouweM! I've done the redesign based on your review feedback + fresh research into where the provider APIs are heading (like the recent anthropic change). Just wanted to get your feedback:

First, removing `thinking: bool | ThinkingConfig` in favor of two flat `ModelSettings` fields:

```python
thinking: bool
thinking_effort: Literal['low', 'medium', 'high']
```

This drops `budget_tokens`, `include_in_response`, and `summary` from the unified API entirely:
- `budget_tokens` per your comment - Anthropic is deprecating `budget_tokens` in favor of `output_config.effort`, Gemini 3 recommends `thinking_level` over `thinking_budget`, and OpenAI never had budgets. Users who need precise token budgets can use `anthropic_thinking` or `google_thinking_config`.
- `summary` is only available for openai so too specific
- `include_in_response` also too specific - can stay in `google_thinking_config` / `groq_reasoning_format`.

Per your feedback on portability - removing all `UserError` raises from the unified thinking path. These settings are "best effort" now, silently ignored where unsupported (same pattern as `max_tokens`, `temperature`, etc.). Provider-specific settings can still error since those are explicit opt-ins.

Also per your comment on `effort_to_budget_map` - moving those constants out of `ModelProfile` and into the model classes themselves since they're provider-family-specific, not per-model. `ModelProfile` keeps just `supports_thinking` and `thinking_always_enabled`. And switching model name checks from `startswith` → `in` per your note about provider prefixes.

Full design doc with provider mapping tables and implementation checklist is in [`API_DESIGN.md`](https://github.com/sarth6/pydantic-ai/blob/unified-thinking-settings/API_DESIGN.md) on the branch.

Since you already left a lot of comments in this direction, I went ahead and implemented this plan so if you agree with it, the code is also ready for review. If not let me know and I'll adjust before you review, thanks!

pydantic_ai_slim/pydantic_ai/models/bedrock.py
```diff
@@ -367,6 +382,81 @@ def _map_tool_definition(f: ToolDefinition) -> ToolTypeDef:
...
+        """Dispatch thinking resolution to the appropriate model-family handler.
+
+        Returns (key, config) where key is the additionalModelRequestFields key
+        to use, or (None, None) if no thinking should be injected.
+        """
+        if 'anthropic' in self.model_name or 'deepseek' in self.model_name:
+            return 'thinking', self._resolve_thinking_config(model_settings)
+        elif 'amazon' in self.model_name:
+            return 'reasoningConfig', self._resolve_nova_thinking_config(model_settings)
+        return None, None
```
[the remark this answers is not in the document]
sarth6: it does set it, because it delegates to anthropic_model_profile() which does the substring matching against _THINKING_MODELS. The Bedrock provider correctly strips the Bedrock-specific name format (geo prefix, version suffix) before passing the clean model name like `claude-sonnet-4-5-20250514` to the Anthropic profile function.

pydantic_ai_slim/pydantic_ai/profiles/anthropic.py
```diff
@@ -1,9 +1,50 @@
...
+    'claude-haiku-4-5',
+)
+
+# Models that support adaptive thinking (type: "adaptive" with output_config.effort).
+# Older thinking models use type: "enabled" with budget_tokens.
+_ADAPTIVE_THINKING_MODELS = (
+    'claude-opus-4-5',
+    'claude-opus-4-6',
+    'claude-sonnet-4-5',
+    'claude-sonnet-4-6',
```
sarth6, line 42: @DouweM I saw you added sonnet 4.6 in https://github.com/pydantic/pydantic-ai/pull/4345

One important design goal I have in this PR is to make it hard for anyone adding a new model to accidentally have it not work with unified thinking settings (and unified settings in general for that matter).

@dsfaccini saw you also have a PR up that does some unified settings, looking for any advice either of you might have on that topic 🙏

sarth6: Thoughts on an addition to the `AGENTS.md`?

```
## Unified Thinking Settings

When adding or modifying a model that supports thinking/reasoning:

1. **Profile** (`profiles/{provider}.py`): Set `supports_thinking=True`. If thinking
   can't be disabled (always-on models like o-series, DeepSeek R1), also set
   `thinking_always_enabled=True`.
2. **Resolver**: Add a `_resolve_{provider}_thinking()` method that calls
   `_resolve_thinking_config(settings, self.profile)` and translates to the
   provider's native format. See existing providers for the pattern.
3. **OpenAIChatModel subclasses**: If inheriting from `OpenAIChatModel`, strip
   `openai_reasoning_effort` in `prepare_request()` after calling `super()` —
   the parent injects it, but your provider has its own mechanism.
4. **Tests** (`tests/test_unified_thinking.py`): Add tests for `thinking=True`,
   `thinking=False`, each effort level, silent drop on unsupported models, and
   provider-specific precedence.
```

I also wonder if pydantic ai should have a special checklist for PRs that add a new model? Or maybe that's an indication we need to better refactor the code, not sure

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -367,12 +395,28 @@ async def request_stream(
...
+            if thinking_config is not None:
+                merged_settings['anthropic_thinking'] = thinking_config
+                # For adaptive models, map unified thinking_effort → anthropic_effort
+                # so _build_output_config doesn't need to re-resolve thinking.
+                if (
+                    thinking_config.get('type') == 'adaptive'
+                    and 'anthropic_effort' not in merged_settings
+                    and (effort := merged_settings.get('thinking_effort'))
+                ):
+                    merged_settings['anthropic_effort'] = effort
```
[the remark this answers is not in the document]
sarth6: Fixed in 05878300. `merge_model_settings` now always returns a new dict (or None) via `base | {}` / `overrides | {}`, preventing `prepare_request` from mutating the stored settings.

pydantic_ai_slim/pydantic_ai/models/xai.py
```diff
@@ -489,6 +501,29 @@ async def _map_user_prompt(self, part: UserPromptPart) -> chat_types.chat_pb2.Me
...
+
+        # Only grok-3-mini has tunable effort; other models use model-name variants
+        if 'grok-3-mini' not in self.model_name:
+            return None
+
+        if not resolved.enabled:
+            return None  # grok-3-mini reasoning can't be disabled via API
+
+        if resolved.effort:
+            return _XAI_EFFORT_MAP.get(resolved.effort)
```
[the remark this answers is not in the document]
sarth6: Fixed in 05878300. grok-3-mini profile now sets `thinking_always_enabled=True` (reasoning can't be disabled, only effort-controlled). Removed the dead `if not resolved.enabled: return None` guard in xai.py that was unreachable with the correct profile.

pydantic_ai_slim/pydantic_ai/profiles/google.py
```diff
@@ -22,13 +22,18 @@ def google_model_profile(model_name: str) -> ModelProfile | None:
...
+    supports_thinking = is_2_5_or_newer and not is_image_model
+
     return GoogleModelProfile(
         json_schema_transformer=GoogleJsonSchemaTransformer,
         supports_image_output=is_image_model,
         supports_json_schema_output=is_3_or_newer or not is_image_model,
         supports_json_object_output=is_3_or_newer or not is_image_model,
         supports_tools=not is_image_model,
         google_supports_native_output_with_builtin_tools=is_3_or_newer,
+        supports_thinking=supports_thinking,
```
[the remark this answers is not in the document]
sarth6: Fixed in 3f9cb821. Google profile now sets `thinking_always_enabled=True` for Pro models (`-pro` in name). This means `resolve_thinking_config()` returns `None` for `thinking=False` on Pro models (silent drop), matching the design spec. Added regression tests for both Gemini 2.5 Pro and 3 Pro silent drop, plus effort-still-works-on-Pro.

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -590,6 +591,21 @@ def profile(self) -> ModelProfile:
...
+        """
+        resolved = resolve_thinking_config(model_settings, self.profile)
+        if resolved is None:
+            return None
+
+        if not resolved.enabled:
+            return None
+
+        # Direct 1:1 mapping — OpenAI's reasoning_effort aligns with our thinking_effort
+        return resolved.effort or 'medium'
```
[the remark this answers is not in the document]
sarth6: Not a bug — forward-compatibility speculation. For gpt-5.1+, `thinking=False` returns `None` (no `reasoning_effort` sent), which is correct because gpt-5.1+ defaults to reasoning off. No current OpenAI model has `thinking_always_enabled=False` with reasoning on by default. If such a model appears, we'd add handling then — not designing for hypothetical futures.

pydantic_ai_slim/pydantic_ai/profiles/groq.py
```diff
@@ -16,8 +16,24 @@ class GroqModelProfile(ModelProfile):
     """Whether the model always has the web search built-in tool available."""
 
 
+# Models that support reasoning on Groq
+_GROQ_REASONING_MODELS = (
+    'deepseek-r1',
+    'qwen-qwq',
+    'qwq',
+)
```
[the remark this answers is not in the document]
sarth6: Acknowledged — this is a profile completeness gap, not a logic bug. GPT-OSS on Groq is routed to `openai_model_profile()` which returns `supports_thinking=False`. Users can use `groq_reasoning_format` directly. Adding Groq-specific GPT-OSS profile routing is a follow-up enhancement, not needed for this PR.

pydantic_ai_slim/pydantic_ai/models/google.py
```diff
@@ -108,6 +110,20 @@
...
+    'medium': ThinkingLevel.MEDIUM,
+    'high': ThinkingLevel.HIGH,
+}
+
+# Effort-to-thinking-budget mapping for Gemini 2.5 models
+_GOOGLE_EFFORT_TO_BUDGET: dict[str, int] = {
+    'low': 1024,
+    'medium': 8192,
+    'high': 32768,
+}
```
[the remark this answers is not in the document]
sarth6: Fixed in 48dd7a98. Gemini 2.5 Flash max is 24576, Pro max is 32768. Lowered `'high'` from 32768 to 24576 so the same mapping works for both without model-specific branching. Added a guard test asserting the 'high' budget stays within the Flash ceiling.

sarth6:
@DouweM I’ve been working on a plan to address your feedback about centralizing/refactoring the thinking resolution. Wanted to run it by you before I start implementing.

The idea is to follow the same pattern as output mode: add a resolved_thinking field to ModelRequestParameters, have the base `Model.prepare_request()` call `resolve_thinking_config()` once and store the result there, and then each model just reads `params.resolved_thinking` at request time and translates to its native format. No model calls the resolver itself, no injection into model_settings, and the OpenAI/Cerebras/OpenRouter strip-and-reinject code goes away completely which is much cleaner. Groq, Cohere, and xAI lose their `prepare_request()` overrides entirely since they only existed for thinking, also much cleaner. Full plan is in `REFACTOR_PLAN.md`

I think this plan will also be good for reducing the chance of a future bug where a dev adds a model or provider and forgets to add unified thinking support since the base `Model` pushes the resolved thinking config onto params so the subclass just needs to wire that through, haven't thought of a safer way than that.

pydantic_ai_slim/pydantic_ai/providers/openrouter.py
```diff
@@ -130,12 +130,13 @@ def model_profile(self, model_name: str) -> ModelProfile | None:
...
         # As OpenRouterProvider is always used with OpenAIChatModel, which used to unconditionally use OpenAIJsonSchemaTransformer,
         # we need to maintain that behavior unless json_schema_transformer is set explicitly
-        return OpenAIModelProfile(
+        profile = OpenAIModelProfile(
             json_schema_transformer=OpenAIJsonSchemaTransformer,
             openai_chat_send_back_thinking_parts='field',
             openai_chat_thinking_field='reasoning',
             openai_chat_supports_file_urls=True,
         ).update(profile)
+        return replace(profile, supports_thinking=True, thinking_always_enabled=False)
```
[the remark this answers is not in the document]
sarth6: Looked into this against the current OpenRouter docs. OpenRouter exposes a normalized `reasoning` object across routed models, and unsupported parameters are ignored by default unless `provider.require_parameters=true`, so opting all OpenRouter models into unified thinking is intentional rather than inheriting each upstream provider's per-model support flags. I kept the behavior and added a clarifying code comment plus provider-profile coverage in `tests/providers/test_openrouter.py`.

pydantic_ai_slim/pydantic_ai/models/cerebras.py
```diff
@@ -92,9 +97,66 @@ def prepare_request(
...
         return new_settings, customized_parameters
 
+    @override
+    def _translate_thinking(self, resolved_thinking: ResolvedThinkingConfig) -> bool | None:
+        """Translate unified thinking settings to Cerebras `disable_reasoning`."""
+        if not resolved_thinking.enabled:
+            return True  # disable_reasoning=True
+
+        # Effort is silently ignored (Cerebras only supports enable/disable)
+        return None  # Don't set disable_reasoning (use default enabled behavior)
```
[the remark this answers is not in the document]
sarth6: Valid on the GPT-OSS part. The profile side was already live through `harmony_model_profile` / `zai_model_profile`, so the code was not dead, but unified `thinking_effort` was not flowing through to Cerebras GPT-OSS. I fixed that in `CerebrasModel._get_reasoning_effort()` so GPT-OSS now maps unified effort onto the OpenAI-compatible `reasoning_effort` field, while GLM remains enable/disable-only. I also added focused coverage in `tests/test_unified_thinking.py`.

pydantic_ai_slim/pydantic_ai/providers/openrouter.py
```diff
@@ -131,13 +131,14 @@ def model_profile(model_name: str) -> ModelProfile | None:
...
         # we need to maintain that behavior unless json_schema_transformer is set explicitly
-        return OpenAIModelProfile(
+        profile = OpenAIModelProfile(
             json_schema_transformer=OpenAIJsonSchemaTransformer,
             openai_chat_send_back_thinking_parts='field',
             openai_chat_thinking_field='reasoning',
             openai_chat_supports_file_urls=True,
             openai_chat_supports_web_search=True,
         ).update(profile)
+        return replace(profile, supports_thinking=True, thinking_always_enabled=False)
```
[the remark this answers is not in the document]
sarth6: I re-checked the always-on case against OpenRouter's reasoning docs as well. OpenRouter's own contract uses `effort: "none"` to disable reasoning, even when the routed upstream provider would otherwise treat reasoning as always-on, so preserving the upstream `thinking_always_enabled=True` here would suppress a capability that OpenRouter itself exposes. I kept `thinking_always_enabled=False`, added an explanatory comment in the provider, and kept coverage asserting that `openai/o3` can be explicitly disabled through OpenRouter.

DouweM:
Replaced by https://github.com/pydantic/pydantic-ai/pull/4640

PR #3897, opened 2026-01-01
ci: add check for orphaned cassettes

dsfaccini:
Today I had a case where I struggled to find the corresponding test for a cassette, which made me think that we should have an automated check that any cassette has a corresponding test :)

## Summary

- Add `scripts/check_cassettes.py` to detect orphaned VCR cassettes in CI
- Remove 13 orphaned cassettes found by the script

## How it works

The script collects all cassette files from `tests/**/cassettes/` and compares them against VCR-marked tests collected via `pytest --collect-only -m vcr`. It handles parametrized tests and class-based test naming conventions.

## Orphaned cassettes removed

- `test_dbos/test_complex_agent_run_stream_in_workflow.yaml`
- `test_dbos/test_dbos_agent_with_dataclass_deps_as_dict.yaml`
- `test_dbos/test_dbos_agent_with_non_dict_deps.yaml`
- `test_gemini/test_gemini_exclusive_minimum_and_maximum.yaml`
- `test_google/test_google_multiple_images.yaml`
- `test_huggingface/test_stream_completion.yaml`
- `test_mistral/test_mistral_model_instructions.yaml`
- `test_openai_responses/test_audio_as_binary_content_input.yaml`
- `test_openai_responses/test_openai_responses_model_builtin_tools.yaml`
- `test_prefect/test_prefect_agent_with_unserializable_deps.yaml`
- `test_temporal/test_complex_agent_run_stream_in_workflow.yaml`
- `test_temporal/test_temporal_agent_with_dataclass_deps_as_dict.yaml`
- `test_temporal/test_temporal_agent_with_non_dict_deps.yaml`

### DIFF
--- a/scripts/check_cassettes.py
+++ b/scripts/check_cassettes.py
@@ -69,7 +69,8 @@ def get_all_cassettes() -> dict[str, set[str]]:
         for subdir in cassette_dir.iterdir():
             if subdir.is_dir():
                 test_stem = subdir.name
-                cassette_names = {f.stem for f in subdir.glob('*.yaml')}
+                # Handle double extensions like .xai.yaml (xAI uses gRPC/protobuf, not HTTP)
+                cassette_names = {f.stem[:-4] if f.stem.endswith('.xai') else f.stem for f in subdir.glob('*.yaml')}
                 cassettes.setdefault(test_stem, set()).update(cassette_names)
 
     return cassettes


.github/workflows/ci.yml
```diff
@@ -41,6 +41,9 @@ jobs:
       - run: uv build --all-packages
       - run: ls -lh dist/
 
+      - name: Check orphaned cassettes
+        run: uv run python scripts/check_cassettes.py
```
DouweM: Could this instead be a pre-commit check, so that it also runs locally?

scripts/check_cassettes.py
```diff
@@ -0,0 +1,133 @@
...
+import subprocess
+import sys
+from pathlib import Path
+
+
+def get_all_cassettes() -> dict[str, set[str]]:
+    """Return {test_file_stem: set of cassette names (without .yaml)}."""
+    cassettes: dict[str, set[str]] = {}
+
+    for cassette_dir in Path('tests').rglob('cassettes'):
```
DouweM, line 66: vcrpy / pytest integration has no builtin way to get the cassettes etc?

scripts/check_cassettes.py
```diff
@@ -0,0 +1,133 @@
...
+                    print(f'  OK: {test_file}/{cassette}.yaml')
+            else:
+                orphans.append(f'{test_file}/{cassette}.yaml')
+
+    print()
+    print(f'Results: {matched} matched, {len(orphans)} orphaned')
+
+    if orphans:
+        print()
+        print('Orphaned cassettes (no matching test):')
```
DouweM, line 123: Can we also check for tests marked with vcr that don't have a cassette? those tests will fail in CI anyway, but if we make this a pre-commit check, then it'll be useful to have, since the `-a` in `git commit -am ...` doesn't automatically include new files, and it's happened to me before that I forgot to include/push them.

DouweM (changes requested):

scripts/check_cassettes.py
dsfaccini: 1. you need an alias
2. I'll check to see if I can add one without much issue, I explicitly left it out bc it seemed very realistic that this could be the case because of skip marks et al

.github/workflows/ci.yml
dsfaccini: it can be a pre-commit too, but I'd like to keep it in CI as well.

scripts/check_cassettes.py
dsfaccini: probably, I'll check out how

scripts/check_cassettes.py
dsfaccini: aahh the thing that I forgot: some files are marked as vcr as a whole even though they contain mocks

EDIT: nvm this I can find a solution, but may need to add a new marker to skip actual false positives (like tests that use api key fixtures but record no cassettes)

EDIT 2: nvm this is difficult af because there's no one criteria we can use to identify a test should have a cassette
1. checking that it's marked as vcr includes mocks that are in the same file as a global vcr marker
2. checking that has `allow_model_request` isn't precise because we're not using that fixture consistently
  - same goes for `api_key` 
4. checking that it uses a mocker produces false negatives because even vcr tests can use mockers for other things

scripts/check_cassettes.py
dsfaccini: I think I found a way

dsfaccini: I'll keep it as is, not sure what I'm doing wrong but I don't get the same result as with the CLI

.github/workflows/ci.yml
dsfaccini: I added it as pre commit :)

DouweM: now that it's in pre-commit, we shouldn't need this change anymore, right, as we already have `uses: pre-commit/action@v3.0.0` above

scripts/check_cassettes.py
DouweM: ok makes sense

scripts/check_cassettes.py
DouweM: can you show that approach and why it wasn't working? because I think that'll be a lot shorter/cleaner than this :)

DouweM (changes requested):

.github/workflows/ci.yml
dsfaccini: aahh sorry, that was me not knowing that when automatically ran pre-commit in CI

scripts/check_cassettes.py
```diff
@@ -0,0 +1,116 @@
...
+        ['uv', 'run', 'pytest', '--collect-only', '-q', '-m', 'vcr', 'tests/'],
+        capture_output=True,
+        text=True,
+    )
+
+    tests: dict[str, set[str]] = {}
+    for line in result.stdout.splitlines():
+        # Parse lines like: tests/models/test_bedrock.py::test_name[param]
+        # or: tests/models/test_bedrock.py::TestClass::test_method[param]
+        match = re.match(r'tests/.*/?(test_\w+)\.py::(.+)', line)
```
DouweM: There's no way to get this from the `pytest` module instead of running the CLI?

dsfaccini: yeahp finally got an alternative, but honestly the regex approach was more readable 😅

scripts/check_cassettes.py
```diff
@@ -0,0 +1,131 @@
...
+class _CollectCassettes:
+    """Pytest plugin to collect cassette names from VCR-marked tests."""
+
+    def __init__(self) -> None:
+        self.tests: dict[str, set[str]] = defaultdict(set)
+
+    def pytest_collection_modifyitems(
+        self, session: pytest.Session, config: pytest.Config, items: list[pytest.Item]
+    ) -> None:
+        from pytest_recording.plugin import get_default_cassette_name
```
DouweM, line 41: At the top please!

scripts/check_cassettes.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+def _remove_yaml_ext(s: str) -> str:
+    """Remove .yaml extension if present."""
+    if s.endswith('.yaml'):
+        return s[:-5]
+    return s
+
+
+class _CollectCassettes:
```
DouweM: Why one class with a method vs separate functions below? Can we organize this a bit more cleanly?

DouweM (changes requested):

scripts/check_cassettes.py
dsfaccini: had to look it up, pytest requires it to be that way, I expanded the docstring to include the explanation

scripts/check_cassettes.py
```diff
@@ -0,0 +1,133 @@
...
+from pytest_recording.plugin import get_default_cassette_name
+
+
+class _CollectVcrTests:
+    """Pytest plugin that collects cassette names referenced by VCR-marked tests.
+
+    This is a class (not functions) because pytest's plugin system requires objects
+    with hook methods, and we need to accumulate state across all test items.
+    """
+
```
dsfaccini, line 27: the explanation

docs/.model-prices-cache.json
DouweM, line 1: ?

scripts/check_cassettes.py
```diff
@@ -0,0 +1,135 @@
...
+    @staticmethod
+    def _remove_yaml_ext(s: str) -> str:
+        if s.endswith('.yaml'):
+            return s[:-5]
+        return s
+
+    def pytest_collection_modifyitems(
+        self, session: pytest.Session, config: pytest.Config, items: list[pytest.Item]
+    ) -> None:
+        # prevents pytest.PytestAssertRewriteWarning: Module already imported so cannot be rewritten; pytest_recording
```
DouweM, line 40: unsure what this means? is this related to the next line and why it's not a top-level import?

docs/.model-prices-cache.json
dsfaccini: https://pydantic.slack.com/archives/C081LUCJ4KX/p1767893863919019?thread_ts=1767792489.317659&cid=C081LUCJ4KX

DouweM: still has no right to be in this particular pr right/

dsfaccini: oh damn that's weird, sorry, I must have confused branches at some point 🤦 , I didn't pay attention and assumed this comment was on https://github.com/pydantic/pydantic-ai/pull/3941

scripts/check_cassettes.py
dsfaccini: yeah it explains why I'm not importing it at the top

Issue #793, opened 2025-01-28
How to include thoughts from the Gemini model in Pydantic output?


MarkusOdenthal:
I need help. Is there a way to include the thoughts from `gemini-2.0-flash-thinking-exp` in the response?

Here is how it works with the Gemini API.

```
from google import genai

client = genai.Client(api_key='GEMINI_API_KEY', http_options={'api_version':'v1alpha'})

config = {'thinking_config': {'include_thoughts': True}}
response = client.models.generate_content(
    model='gemini-2.0-flash-thinking-exp',
    contents='Explain how RLHF works in simple terms.',
    config=config
)

for part in response.candidates[0].content.parts:
    if part.thought:
        print(f"Model Thought:\n{part.text}\n")
    else:
        print(f"\nModel Response:\n{part.text}\n")
```

Thanks for help :)

sydney-runkle:
Hi!

Thanks for your inquiry. This isn't yet supported, but is something we can work to accommodate. I'm guessing we'll have to add  some info to the `parts` of the gemini response.

MarkusOdenthal:
Thanks for your response :)

I think this would be great, especially for tracking this thinking. I had a great experience using this thinking to improve prompts.

I also could try it but have no idea where to start.

I gemini you need to set the parameter `http_options={'api_version':'v1alpha'}` when creating the client.Where to set this in Pydantic AI?

Then you need to set the config when sending the generate_content request to: `config = {'thinking_config': {'include_thoughts': True}}`. Where so set this in pydantic AI?

And it would also be interesting if we could track this on Logfire :)

tranhoangnguyen03:
Similar to https://github.com/pydantic/pydantic-ai/issues/731?

Kludex:
I tried all the examples from https://github.com/GoogleCloudPlatform/generative-ai/blob/main/gemini/getting-started/intro_gemini_2_0_flash_thinking_mode.ipynb.

It seems none have `part.thought` being `True`.

Kludex:
Google seems to have removed thoughts from their API. See https://github.com/googleapis/python-genai/pull/231/.

I lost a lot of time on this. It seems their documentation is not in sync with their product.

Kludex:
Ah, I see what they did... They do show the thoughts in their Google AI Studio, but they don't return the thoughts via API anymore.

See https://discuss.ai.google.dev/t/thoughts-are-missing-cot-not-included-anymore/63653/16.

[...]

PR #3900, opened 2026-01-02
Add various cassette and snapshotting fixes

dsfaccini:
This PR encompasses a couple fixes collected from other PRs:
- #3826: collapsing long byte strings in snapshots into `IsBytes()`
- #3611: `normalize_smart_chars` in cassettes (so they're normalized in the snapshots too, otherwise locally normalized-snapshots fail in CI)
  - also here https://github.com/pydantic/pydantic-ai/pull/3904/files#r2659952162
- #3612: handle gzip and brotli decompression when recording cassettes. This issue occurred in https://github.com/pydantic/pydantic-ai/pull/2611#discussion_r2565806141 as well

- **add snapshot byte collapse**
- **normalize cassette characters**
- **add gzip and brotli decompression handling**
- **rewrite test_gemini_model_thinking_part after #793 closed**
- **update comment about caching**

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yaml
+++ b/.github/ISSUE_TEMPLATE/bug.yaml
@@ -1,6 +1,6 @@
 name: 🐛 Pydantic AI Bug
 description: Report a bug or unexpected behavior in Pydantic AI
-labels: ["need confirmation"]
+labels: ["bug"]
 
 body:
   - type: markdown
--- a/.github/ISSUE_TEMPLATE/feature-request.yaml
+++ b/.github/ISSUE_TEMPLATE/feature-request.yaml
@@ -1,6 +1,6 @@
 name: 🚀 Pydantic AI Feature Request
 description: "Suggest a new feature for Pydantic AI"
-labels: ["feature request"]
+labels: ["feature"]
 
 body:
   - type: markdown
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -2,6 +2,7 @@
 
 <!-- Please add the issue number that should be closed when this PR is merged. -->
 <!-- If there is no issue, please create one first, even for simple bug fixes. -->
+
 - Closes #<issue>
 
 ### Pre-Review Checklist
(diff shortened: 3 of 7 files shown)


pydantic_ai_slim/pydantic_ai/models/gemini.py
```diff
@@ -1,3 +1,12 @@
+"""Deprecated Gemini model implementation.
+
+This module is deprecated. Use `pydantic_ai.models.google.GoogleModel` instead.
```
DouweM: Link to API docs please

tests/json_body_serializer.py
```diff
@@ -1,10 +1,50 @@
...
 import urllib.parse
+import zlib
 from typing import TYPE_CHECKING, Any
 
+import brotli
 import yaml
 
+# Smart quote and special character normalization.
+# LLM APIs sometimes return smart quotes and special Unicode characters in responses.
+# When captured in cassettes, these can cause linter complaints about non-ASCII characters
```
DouweM: The linter will not complain about them being in cassettes, just about being in snapshots!

tests/conftest.py
```diff
@@ -112,6 +114,22 @@ def equals(self, other: Any) -> bool:
                 return other == self._first_other
 
 
+SNAPSHOT_BYTES_COLLAPSE_THRESHOLD = 50
+
+
+@customize_repr
+def _(value: bytes):  # pragma: no cover
```
DouweM, line 121: Would it make sense to do this for datetime as well?

tests/conftest.py
```diff
@@ -112,6 +114,22 @@ def equals(self, other: Any) -> bool:
...
+    """Use IsBytes() for large byte sequences in snapshots.
+    inline-snapshot serializes dataclass fields by calling repr() on each field value.
+    For BinaryContent.data (bytes), this would embed huge byte literals in snapshots.
+    By customizing bytes repr, we replace large byte sequences with IsBytes().
+    """
+    if len(value) > SNAPSHOT_BYTES_COLLAPSE_THRESHOLD:
+        return 'IsBytes()'
```
DouweM, line 124: I guess we can't auto-import this from the appropriate file...

dsfaccini: I mean... it'd be a bit more of a complicated hook but is certainly doable

tests/json_body_serializer.py
dsfaccini: yeah bad wording

tests/conftest.py
DouweM: I'm ok leaving that out for now

PR #3400, opened 2025-11-12
Add `XaiModel` that uses the xAI SDK and deprecate `GrokProvider` that used the OpenAI-compatible API

brightsparc:
## Summary

This PR adds first-class support for xAI via the native `xai-sdk` (**`XaiModel`** + **`XaiProvider`**) and deprecates the existing OpenAI-compatible **`GrokProvider`**. Because the xAI SDK is **gRPC/protobuf** (not HTTP), this PR also introduces a **protobuf “cassette” recorder/replayer** so we can run deterministic tests without live network calls.

### New: native xAI integration

- **`XaiModel`** (`pydantic_ai_slim/pydantic_ai/models/xai.py`)
  - Streaming + non-streaming
  - Reasoning/thinking support (including encrypted reasoning content where supported)
  - xAI built-in tools: web search, code execution, MCP server tool
  - Structured output via native JSON schema support
  - Multimodal inputs (e.g. images / documents where supported)

- **`XaiProvider`** (`pydantic_ai_slim/pydantic_ai/providers/xai.py`)
  - Wraps `xai_sdk.AsyncClient`
  - Reads `XAI_API_KEY` by default (or accepts `api_key=...` / `xai_client=...`)
  - Uses `grok_model_profile(...)` for capability detection (built-in tools + reasoning)

- **Grok profile updates** (`pydantic_ai_slim/pydantic_ai/profiles/grok.py`)
  - Introduces/extends `GrokModelProfile` with xAI-specific capability flags, including built-in tool support and reasoning support.

- **Known model names**
  - Adds `xai:`-prefixed Grok model strings to `KnownModelName` (`pydantic_ai_slim/pydantic_ai/models/__init__.py`).

### New: protobuf cassette recorder (test infra)

- **`tests/models/xai_proto_cassettes.py`**
  - **`XaiProtoRecorder`**: records request/response protobuf bytes from the real SDK into a YAML cassette.
  - **`XaiProtoCassetteClient`**: replays recorded protobuf responses and reconstructs streaming responses by applying recorded chunks (mirrors SDK `Response.process_chunk(...)` behavior).
  - Motivation: `pytest-recording`/VCR only records **HTTP**, but the xAI SDK uses **gRPC**.

- **`tests/conftest.py`**
  - Adds an `xai_provider` fixture wired to these proto cassettes under `tests/models/cassettes/test_xai/*.xai.yaml`.

### Deprecation: `GrokProvider`

[...]

PR #3904, opened 2026-01-04
Add normalized property for unified builtin tool content access

### LINKED ISSUE #3561: Add unified typed schemas for builtin tool return content

sarth6:
## Summary

This PR adds a **backward-compatible** `normalized` property to builtin tool return parts for provider-agnostic access to content.

### API Design

```python
# Existing API (unchanged - 100% backward compatible!)
part.content['stdout']     # Anthropic raw - STILL WORKS
part.content['outcome']    # Google raw - STILL WORKS
part.content['logs']       # OpenAI raw - STILL WORKS

# New unified API (additive)
part.normalized['output']  # "Hello\n" - unified output field
part.normalized['status']  # "completed" - unified status
```

## Changes

**What this PR adds:**
- `normalized` property on all 5 return part classes
- `Normalized*Content` TypedDicts for unified schema documentation
- Normalization functions that transform provider-specific fields to unified names

**What this PR does NOT change:**
- `content` field type stays as `Any` (inherited from base class)
- No changes to how providers populate `content`
- No breaking changes to existing code

## Normalized Field Mappings

| Provider | `output` Source | `status` Source |
|----------|-----------------|-----------------|
| Anthropic | `stdout` | `return_code == 0 ? 'completed' : 'failed'` |
| Google | `output` | `outcome` mapping |
| OpenAI | `'\n'.join(logs)` | `status` |

## Test Plan

- [x] All existing tests pass (raw content unchanged)
- [x] New tests for `normalized` property across all providers
- [x] Type checker passes

## Related

Closes #3561


.pre-commit-config.yaml
```diff
@@ -23,7 +23,8 @@ repos:
     rev: 0.6.8
     hooks:
       - id: fix-smartquotes
-        exclude: "cassettes/"
+        # Exclude cassettes and test files with API response snapshots that contain Unicode
+        exclude: "cassettes/|tests/models/test_anthropic\\.py|tests/models/test_openai_responses\\.py"
```
dsfaccini: sorry this isn't admissible lol, what you can do is set the values in the snapshots ` = IsStrI()`

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -28,6 +28,352 @@
...
 
+# =============================================================================
+# NORMALIZED SCHEMAS FOR BUILTIN TOOL RETURN CONTENT
+# =============================================================================
+# These TypedDicts define a provider-agnostic, normalized format for builtin tool
+# results. Use the `.normalized` property on return parts to access data in this
+# unified format, regardless of which provider generated the response.
+
+
+CodeExecutionStatus: TypeAlias = Literal['completed', 'failed', 'timeout', 'unknown']
```
DouweM, line 39: I'd rather encode unknown as None / omitting the key

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -28,6 +28,352 @@
...
+"""Status of a code execution result."""
+
+BuiltinToolStatus: TypeAlias = Literal['completed', 'failed', 'unknown']
```
DouweM, line 42: How about an optional `success: bool`?

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -28,6 +28,352 @@
...
+
+    Access via `part.normalized` for provider-agnostic field names.
+    """
+
+    status: CodeExecutionStatus
+    """Execution status: 'completed', 'failed', 'timeout', or 'unknown'."""
+    output: str
+    """Combined output from execution (stdout for Anthropic, output for Google, joined logs for OpenAI)."""
+    error: str
+    """Error output (stderr for Anthropic, not available for other providers)."""
```
DouweM, line 57: We recently received some new support for builtin tools, at least these ones but I suggest looking at the git history of builtin-tools.py/md as well:
- #3891
- #3400
- https://github.com/pydantic/pydantic-ai/pull/3599

We need to take those that have already been merged into account here, _and_ we should only include new fields for things that are supported by at least two providers. So no `error` field if only Anthropic has it. But I believe others (nova or xai?) have stderr as well.

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -28,6 +28,352 @@
...
+    exit_code: int
+    """Process exit code (return_code for Anthropic, not available for other providers)."""
```
DouweM, line 59: Same here, drop unless supported by at least 2 providers

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -28,6 +28,352 @@
...
+
+
+class NormalizedWebSearchSource(TypedDict, total=False):
+    """Unified schema for a single web search result source."""
+
+    title: str
+    """Title of the search result."""
+    url: str
+    """URL of the search result (normalized from 'uri' for Google)."""
```
DouweM, line 68: This applies to all of these:
- We don't need to specify what it's normalized from
- We do need to specify which providers support it; just like we have on builtin tool settings

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -28,6 +28,352 @@
...
+
+class NormalizedWebFetchPage(TypedDict, total=False):
+    """Unified schema for a single fetched web page."""
+
+    url: str
+    """URL of the fetched page (normalized from 'retrieved_url' for Google)."""
+    content: str
+    """The fetched page content/text."""
+    content_type: str
+    """MIME type of the content."""
```
DouweM, line 95: For all of these, double check that we only have fields that are used by at least 2 providers

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -28,6 +28,352 @@
...
+
+    Args:
+        raw: Raw provider-specific content.
+
+    Returns:
+        Normalized content with unified field names.
+    """
+    normalized: NormalizedCodeExecutionContent = {}
+
+    # Anthropic format: stdout, stderr, return_code
```
DouweM, line 174: The mapping should happen in the model classes, this key-based stuff is too brittle. The model classes are responsible for setting the normalized fields that they can.

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -887,6 +1233,144 @@ class BuiltinToolReturnPart(BaseToolReturnPart):
...
+    part_kind: Literal['code-execution-return'] = 'code-execution-return'  # pyright: ignore[reportIncompatibleVariableOverride]
+    """Part type identifier, this is available on all parts as a discriminator."""
+
+    @property
+    def normalized(self) -> NormalizedCodeExecutionContent:
+        """Get normalized content with unified field names.
+
+        Returns provider-agnostic data with fields: status, output, error, exit_code.
+        """
+        return normalize_code_execution_content(cast(dict[str, Any], self.content))
```
DouweM, line 1253: Doing this in the model classes as I suggested before means it's not going to apply to existing objects, but that's OK :)

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -887,6 +1233,144 @@ class BuiltinToolReturnPart(BaseToolReturnPart):
...
 
+@dataclass(repr=False)
+class CodeExecutionReturnPart(BuiltinToolReturnPart):
+    """A return part for code execution tool results.
+
+    The content field contains raw provider-specific data. Use the `normalized` property
+    for provider-agnostic access with unified field names.
+    """
+
+    part_kind: Literal['code-execution-return'] = 'code-execution-return'  # pyright: ignore[reportIncompatibleVariableOverride]
```
DouweM, line 1244: These types should live in `messages.py`

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -887,6 +1233,144 @@ class BuiltinToolReturnPart(BaseToolReturnPart):
...
+    """Part type identifier, this is available on all parts as a discriminator."""
+
+    @property
+    def normalized(self) -> NormalizedCodeExecutionContent:
```
DouweM, line 1248: What do you think about changing the type of `content` on this class to `CodeExecutionContent`, _not allow any nonstandard fields_, and update the model classes to set only the needed fields? Since `content` is currently not guaranteed to have any particular form, that wouldn't be a backward-incompatible change. Although I suppose the existing model class logic that reconstructs API parts from `BuiltinToolCallPart.content`s will need to remain for older parts (that don't have a specific `part_kind`)

All the other non-normalized data can then be stored on `provider_details`, both other things the user may want to use + things that need to be sent back to the API. That's the pattern we're starting to follow in other places as well.

pydantic_ai_slim/pydantic_ai/messages.py
DouweM: Also this should be a new field, not reusing the existing `part_kind`

Edit: or use the `tool_name`!

pydantic_ai_slim/pydantic_ai/messages.py
DouweM: Of course that goes for the entire design of the feature, not just this type.

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -887,6 +1233,144 @@ class BuiltinToolReturnPart(BaseToolReturnPart):
...
+    'web_search': WebSearchReturnPart,
+    'web_fetch': WebFetchReturnPart,
+    'url_context': WebFetchReturnPart,  # Deprecated alias
+    'file_search': FileSearchReturnPart,
+    'image_generation': ImageGenerationReturnPart,
+}
+
+
+def _migrate_builtin_tool_return_part(data: dict[str, Any]) -> dict[str, Any]:
+    """Migrate old BuiltinToolReturnPart data to specific subclass based on tool_name.
```
DouweM, line 1351: I'm not sure I want to; as mentioned above, we may change the content of `content`, so the old logic reading from that will still need to be able to tell the difference between old-style and new-style builtin tool results

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -887,6 +1233,144 @@ class BuiltinToolReturnPart(BaseToolReturnPart):
...
+    part_kind = data.get('part_kind')
+    tool_name = data.get('tool_name', '')
+
+    if part_kind == 'builtin-tool-return':
+        if tool_name in _TOOL_NAME_TO_RETURN_PART_CLASS:
+            new_class = _TOOL_NAME_TO_RETURN_PART_CLASS[tool_name]
+            data = dict(data)
+            data['part_kind'] = new_class.part_kind
+        # MCP and memory tools stay as base class (schemas not stable)
+        elif tool_name.startswith('mcp_server:') or tool_name in ('mcp_server', 'memory'):
```
DouweM, line 1368: Too hard-coded!

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -887,6 +1233,144 @@ class BuiltinToolReturnPart(BaseToolReturnPart):
...
+
+
+# Mapping from tool_name to specialized return part class for migration
+_TOOL_NAME_TO_RETURN_PART_CLASS: dict[str, type[BuiltinToolReturnPart]] = {
+    'code_execution': CodeExecutionReturnPart,
+    'web_search': WebSearchReturnPart,
+    'web_fetch': WebFetchReturnPart,
+    'url_context': WebFetchReturnPart,  # Deprecated alias
+    'file_search': FileSearchReturnPart,
+    'image_generation': ImageGenerationReturnPart,
```
DouweM, line 1346: We should normalize MCP as well!

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -1215,8 +1699,159 @@ class BuiltinToolCallPart(BaseToolCallPart):
     """Part type identifier, this is available on all parts as a discriminator."""
 
 
+@dataclass(repr=False)
+class CodeExecutionCallPart(BuiltinToolCallPart):
+    """A call part for code execution tool with normalized accessor properties."""
+
+    args: dict[str, Any] | str | None = None
```
DouweM, line 1706: Same suggestion as above: make `args` typed as a typed dict specifically for code execution, and store everything non-standard on `provider_details`

DouweM: And same as with returns, normalize in the model class instead of here

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -1215,8 +1699,159 @@ class BuiltinToolCallPart(BaseToolCallPart):
...
+    """The code execution call arguments."""
+
+    part_kind: Literal['code-execution-call'] = 'code-execution-call'  # pyright: ignore[reportIncompatibleVariableOverride]
```
DouweM, line 1709: Same, don't override :)

sarth6:
Thanks for the review! Will take a look at this one a bit later this week - the other PRs I have open seem to be getting linked more, or are getting more thumbs ups, or in the case of the fallback one seems closer to merging, so will prioritize those first.

But if you disagree with that priority let me know I'll adjust focus!

dsfaccini:
we merged a bump to pyright that causes changes in many places in the codebase and we want to prevent CI appearing green where it could fail for lacking the pyright bump.

please sync your setup locally after resolving the conflicts by running the following:

```bash
git pull

deactivate

rm -rf .venv

make install
```

sarth6:
still on my radar, just waiting for reviews on a few others and don't want to put up too many parallel PRs

PR #3905, opened 2026-01-05
tests: Close provider clients before their underlying httpx clients

dsfaccini:
we've a couple warnings that are being suppressed similar to the unraisable exception we were getting last week that we fixed by more precisely closing HTTP clients in https://github.com/dsfaccini/pydantic-ai/commit/1d01aa475e72d9a95b725ed1b95da3ec0ac6c84e

but I got the same error again in https://github.com/pydantic/pydantic-ai/actions/runs/20662173239/job/59326932015?pr=3611 so I started digging some more and looking for other ways of optimizing, so this is an excursion on that

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -345,7 +345,7 @@ def _wrap_provider_inits() -> Iterator[None]:
                 try:
                     _current_test_providers.get().add(self)
                 except LookupError:
-                    pass  # No test context (e.g., module-level imports)
+                    pass  # pragma: no cover
 
             return wrapper
 


dsfaccini:
I wanted to check whether I could either remove the need for the warning filters and/or speed up tests further but 0/2 seems to be the case, so I'll keep this PR for exploring approaches but nothing else

dsfaccini:
same as https://github.com/pydantic/pydantic-ai/pull/3910#issuecomment-3720220752

PR #3907, opened 2026-01-05
Add YAML and TOML media type support in `BinaryContent.from_path`

### LINKED ISSUE #3776: BinaryContent: Add support for is_text

majiayu000:
## Summary
- Add `application/yaml` mime type for `.yaml` and `.yml` extensions (RFC 9512)
- Add `application/toml` mime type for `.toml` extension (RFC 9519)
- Add tests for YAML and TOML file type detection

## Test plan
- [x] Added tests for `.yaml`, `.yml`, and `.toml` file extensions
- [x] All existing tests pass

Fixes #3776

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/messages.py
+++ b/pydantic_ai_slim/pydantic_ai/messages.py
@@ -59,6 +59,13 @@ _mime_types.add_type('audio/flac', '.flac')
 _mime_types.add_type('audio/ogg', '.oga')
 _mime_types.add_type('audio/wav', '.wav')
 
+# Text/data file types not recognized by default mimetypes
+# YAML: RFC 9512 (https://www.rfc-editor.org/rfc/rfc9512.html)
+_mime_types.add_type('application/yaml', '.yaml')
+_mime_types.add_type('application/yaml', '.yml')
+# TOML: RFC 9519 (https://www.rfc-editor.org/rfc/rfc9519.html)
+_mime_types.add_type('application/toml', '.toml')
+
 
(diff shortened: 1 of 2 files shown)


dsfaccini:
hey @majiayu000 appreciate your contribution to fix this

@DouweM I'm fine with this and they referenced the RFCs for this so there should be no discussion on whether `text/yaml` or `application/yaml`

should we add types for `text` `json` and `xml` while we're at this?

gaby:
> hey @majiayu000 appreciate your contribution to fix this
> 
> @DouweM I'm fine with this and they referenced the RFCs for this so there should be no discussion on whether `text/yaml` or `application/yaml`
> 
> should we add types for `text` `json` and `xml` while we're at this?

Agree, json, and xml would still fail

DouweM:
Feel free to do that in a followup PR :)

PR #3909, opened 2026-01-05
Add OTel span ID to each ModelMessage

### LINKED ISSUE #3906: Add OTEL span ID to each ModelMessage
### Description

To help with observability correlation, it would be great if the OTEL span ID were automatically included as a property on each `ModelMessage`, if any type of OTEL instrumentation is configured for the agent.

**Example Use Case**: This would simplify the workflow of allowing a user to provide feedback for a historical agent message, by allowing the feedback to be attached to the message's span in the OTEL system directly using the span reference, without having to deduce the correct span to annotate via fuzzy creation date matching against the message's timestamp.

**Potential Implementation**: This might be achieved by adding a new `span_id: str | None = None` field to `ModelRequest` and `ModelResponse`. Then when each `ModelMessage` is instantiated at runtime during trace execution, the current span ID could be queried and set on the message object. This would cause the span ID to be included in the message's serialized form (e.g. when persisting the message to a database), allowing for easy span lookup later in time.

epeters3:
- Closes #3906

### Pre-Review Checklist

### Pre-Merge Checklist

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/_agent_graph.py
+++ b/pydantic_ai_slim/pydantic_ai/_agent_graph.py
@@ -543,6 +543,11 @@ class ModelRequestNode(AgentNode[DepsT, NodeRunEndT]):
         response: _messages.ModelResponse,
     ) -> CallToolsNode[DepsT, NodeRunEndT]:
         response.run_id = response.run_id or ctx.state.run_id
+
+        # Copy span_id from response to request since both messages were created under the same span.
+        if response.span_id:
+            self.request.span_id = response.span_id
+
         # Update usage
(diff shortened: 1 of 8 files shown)


epeters3:
Hi @DouweM! I see a little review activity on this PR; that's exciting! FYI the real diff in this PR is only about 29 lines long; the rest is just new unit tests and a few snapshot updates.

DouweM:
@epeters3 Please check out my comment at https://github.com/pydantic/pydantic-ai/issues/3906#issuecomment-3725498697! I'd rather give each message an ID and store that on the span, than store the span ID on the message.

PR #3012, opened 2025-09-26
Replace `asyncio` to `anyio`

Kludex:

PR #1695, opened 2025-05-12
#748: don't share httpx.AsyncHTTPTransport between event loops

ewjoachim:
Fixes #748

In `cached_async_http_client` we cache the http client (...duh) and its transport. The transport keeps a pointer to the current event loop through:
```python
transport._pool._connections[0]._connection._network_stream._stream.transport_stream._transport._loop
```
and will try to reuse this event loop on subsequent calls. This causes `RuntimeError('Event loop is closed')` in case that loop was closed and we're using another one.

My fix makes it so that caches of the client **and** the transport are _per loop_, and on new event loops, we'll get fresh objects.

I've been using https://github.com/oscar-broman/pydantic-ai-gemini-issue-mre as a reproducer, and it fails on `main` and passes on my branch.

The reason we don't see the same test failure in this repo's tests surely is [this](https://anyio.readthedocs.io/en/stable/testing.html#technical-details): because there are fixtures with `scope='session'`, anyIO only creates a single event loop that each test share, so we can't see the issue (also, we mock HTTP calls with `pytest-recording` and I'm not 100% sure the mock occurs at a level that would make the test fail visible)

All that is to say: I'm not sure how best to implement a reproducer without impeding too much on the codebase. Opinions from maintainers welcome :)

> [!NOTE]
> I switched `@cache` to `@lru_cache(max_size=32)` because there's virtually infinitly many loops a program may start, so that could create a memory leak otherwise.

> [!WARNING]
> I'm afraid this makes the following fixture very brittle. It will work if the anyio single event loop issue continues to be, but if we manage to have an event loop for each test, then we won't be able to properly close the clients. That said: isn't that an issue with real code in the real life too ? Who's responsible for closing those cached clients ? (maybe they close on `__del__` but that's not ideal, is it ?

https://github.com/ewjoachim/pydantic-ai/blob/main/tests/conftest.py#L202-L217

PR #3910, opened 2026-01-05
Make cached_async_http_client event-loop-aware to fix Celery/Cloud Run issues

### LINKED ISSUE #3240: builtins.RuntimeError: <asyncio.locks.Event object at 0xffff3743a090 [unset]> is bound to a different event loop.
### Initial Checks

### Description

After upgrading from 1.1.0 to 1.4.0 without any code changes, I am now getting the following error:

```
  File "/usr/local/lib/python3.13/site-packages/google/genai/_api_client.py", line 1293, in _async_request
    return await self._async_retry(  # type: ignore[no-any-return]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        self._async_request_once, http_request, stream
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/usr/local/lib/python3.13/site-packages/tenacity/asyncio/__init__.py", line 111, in __call__
    do = await self.iter(retry_state=retry_state)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/tenacity/asyncio/__init__.py", line 153, in iter
    result = await action(retry_state)
             ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/tenacity/_utils.py", line 99, in inner
    return call(*args, **kwargs)
  File "/usr/local/lib/python3.13/site-packages/tenacity/__init__.py", line 420, in exc_check
    raise retry_exc.reraise()
          ~~~~~~~~~~~~~~~~~^^
  File "/usr/local/lib/python3.13/site-packages/tenacity/__init__.py", line 187, in reraise
    raise self.last_attempt.result()
          ~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/local/lib/python3.13/concurrent/futures/_base.py", line 449, in result
    return self.__get_result()
           ~~~~~~~~~~~~~~~~~^^
  File "/usr/local/lib/python3.13/concurrent/futures/_base.py", line 401, in __get_result
    raise self._exception
  File "/usr/local/lib/python3.13/site-packages/tenacity/asyncio/__init__.py", line 114, in __call__
    result = await fn(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/google/genai/_api_client.py", line 1266, in _async_request_once
    client_response = await self._async_httpx_client.request(
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<5 lines>...
    )
    ^
  File "/usr/local/lib/python3.13/site-packages/httpx/_client.py", line 1540, in request
    return await self.send(request, auth=auth, follow_redirects=follow_redirects)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/ddtrace/contrib/internal/httpx/patch.py", line 138, in _wrapped_async_send
    resp = await wrapped(*args, **kwargs)
  File "/usr/local/lib/python3.13/site-packages/httpx/_client.py", line 1629, in send
    response = await self._send_handling_auth(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<4 lines>...
    )
    ^
  File "/usr/local/lib/python3.13/site-packages/httpx/_client.py", line 1657, in _send_handling_auth
    response = await self._send_handling_redirects(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<3 lines>...
    )
    ^
  File "/usr/local/lib/python3.13/site-packages/httpx/_client.py", line 1694, in _send_handling_redirects
    response = await self._send_single_request(request)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/httpx/_client.py", line 1730, in _send_single_request
    response = await transport.handle_async_request(request)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/httpx/_transports/default.py", line 394, in handle_async_request
    resp = await self._pool.handle_async_request(req)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/httpcore/_async/connection_pool.py", line 256, in handle_async_request
    raise exc from None
  File "/usr/local/lib/python3.13/site-packages/httpcore/_async/connection_pool.py", line 236, in handle_async_request
    response = await connection.handle_async_request(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        pool_request.request
        ^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/usr/local/lib/python3.13/site-packages/httpcore/_async/connection.py", line 103, in handle_async_request
    return await self._connection.handle_async_request(request)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/httpcore/_async/http11.py", line 136, in handle_async_request
    raise exc
  File "/usr/local/lib/python3.13/site-packages/httpcore/_async/http11.py", line 106, in handle_async_request
    ) = await self._receive_response_headers(**kwargs)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/httpcore/_async/http11.py", line 177, in _receive_response_headers
    event = await self._receive_event(timeout=timeout)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/httpcore/_async/http11.py", line 217, in _receive_event
    data = await self._network_stream.read(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        self.READ_NUM_BYTES, timeout=timeout
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/usr/local/lib/python3.13/site-packages/httpcore/_backends/anyio.py", line 35, in read
    return await self._stream.receive(max_bytes=max_bytes)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/anyio/streams/tls.py", line 237, in receive
    data = await self._call_sslobject_method(self._ssl_object.read, max_bytes)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/anyio/streams/tls.py", line 180, in _call_sslobject_method
    data = await self.transport_stream.receive()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/anyio/_backends/_asyncio.py", line 1263, in receive
    await self._protocol.read_event.wait()
  File "/usr/local/lib/python3.13/asyncio/locks.py", line 210, in wait
    fut = self._get_loop().create_future()
          ~~~~~~~~~~~~~~^^
  File "/usr/local/lib/python3.13/asyncio/mixins.py", line 20, in _get_loop
    raise RuntimeError(f'{self!r} is bound to a different event loop')
RuntimeError: <asyncio.locks.Event object at 0xffff401736b0 [unset]> is bound to a different event loop
```

### Example Code

```Python
If required, I'll endeavour to create a minimal example and bisect, as many of the below packages were updated to latest at the same time.
```

```Text
Stack is (all latest):

fastapi==0.120.0
pydantic==2.12.3
uvicorn[standard]==0.38.0
#for database   #new
sqlalchemy==2.0.44
psycopg[binary,pool]==3.2.10
#for loading environment variables
python-dotenv==1.1.1
requests==2.32.5
# Specific version required by fastapi
starlette==0.48.0
Pillow==12.0.0
boto3==1.40.58
botocore==1.40.58
ddtrace==3.17.0
aiohttp==3.13.0
aiolimiter==1.2.1
JSON-log-formatter==1.1.1
pydantic-ai-slim[google,openai]==1.4.0
```

using gemini-2.5-flash-lite

dsfaccini:
This is an excursion with claude to try to fix #3240 but hasn't been confirmed in a Cloud Run or Celery environment.

## Summary

- Replace `@functools.cache` with a custom event-loop-aware cache that uses `id(asyncio.get_running_loop())` as part of the cache key
- Each event loop now gets its own cached `httpx.AsyncClient` instance
- This prevents the "RuntimeError: Event object is bound to a different event loop" error

## Problem

Users reported `RuntimeError: Event object is bound to a different event loop` when using pydantic-ai with Google models in concurrent environments like:
- Celery workers
- Google Cloud Run Functions with concurrency
- Other multi-event-loop environments

The root cause was that `httpx.AsyncClient` contains internal asyncio locks that are bound to the event loop at creation time. When a cached client was reused in a different event loop, those primitives failed.

## Solution

The cache key now includes `id(asyncio.get_running_loop())`, ensuring each event loop gets its own client instance. The implementation:

1. Uses a custom dict-based cache instead of `@functools.cache`
2. Thread-safe via `threading.Lock`
3. Handles edge cases (no running loop, closed clients)
4. Backwards compatible - no changes needed to providers

### DIFF
--- a/tests/test_dbos.py
+++ b/tests/test_dbos.py
@@ -51,8 +51,6 @@ try:
 
     from dbos import DBOS, DBOSConfig, SetWorkflowID
     from packaging.version import Version
-    from sqlalchemy import create_engine
-    from sqlalchemy.pool import StaticPool
 
     from pydantic_ai.durable_exec.dbos import DBOSAgent, DBOSMCPServer, DBOSModel
 
@@ -122,16 +120,13 @@ def workflow_raises(exc_type: type[Exception], exc_message: str) -> Iterator[Non
     assert str(exc_info.value) == exc_message
 
 
-in_memory_engine = create_engine(
-    'sqlite:///:memory:?cache=shared', connect_args={'check_same_thread': False}, poolclass=StaticPool
-)
-
+DBOS_SQLITE_FILE = 'dbostest.sqlite'
 DBOS_CONFIG: DBOSConfig = {
     'name': 'pydantic_dbos_tests',
+    'system_database_url': f'sqlite:///{DBOS_SQLITE_FILE}',
     'run_admin_server': False,
     # enable_otlp requires dbos>1.14
     'enable_otlp': True,
-    'system_database_engine': in_memory_engine,
 }
 
 
@@ -143,6 +138,18 @@ def dbos() -> Generator[DBOS, Any, None]:
         yield dbos
     finally:
         DBOS.destroy()
+
+
+# Automatically clean up old DBOS sqlite files
+@pytest.fixture(autouse=True, scope='module')
+def cleanup_test_sqlite_file() -> Iterator[None]:
+    if os.path.exists(DBOS_SQLITE_FILE):
+        os.remove(DBOS_SQLITE_FILE)  # pragma: lax no cover
+    try:
+        yield
+    finally:
+        if os.path.exists(DBOS_SQLITE_FILE):
+            os.remove(DBOS_SQLITE_FILE)  # pragma: lax no cover
 
 
 model = OpenAIChatModel(


adtyavrdhn:
I was going to look into this, seemed like an interesting issue but now I can just read through your fix :)

dsfaccini:
@adtyavrdhn my bad, yeah I was surprised that this was such a widespread issue so wanted to explore a fix. I don't like though that it's slowing down the test suite... and in any case merging will need to be discussed with @DouweM to rule out we're changing behavior for worse for existing users/uses.

I'll check out what's causing the slow down, we we're happy the last couple of weeks about CI being faster 😅 we don't wanna surrender that so easily

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -1161,37 +1163,82 @@ def infer_model(  # noqa: C901
...
 def _cached_async_http_client(provider: str | None, timeout: int = 600, connect: int = 5) -> httpx.AsyncClient:
-    return httpx.AsyncClient(
-        timeout=httpx.Timeout(timeout=timeout, connect=connect),
-        headers={'User-Agent': get_user_agent()},
-    )
+    """Get or create a cached httpx.AsyncClient for the current event loop."""
+    loop_id = _get_current_loop_id()
+
+    # If no loop is running, create a new uncached client
+    if loop_id is None:
```
adtyavrdhn, line 1212: This could be the reason for the slowdown, the previous @cache decorator did not care about loop_id but now if it is out of async context then it will always create a new client?

adtyavrdhn:
@dsfaccini No worries, I wouldn't have counted on myself to figure it out this soon :'))

pydantic_ai_slim/pydantic_ai/models/__init__.py
adtyavrdhn: Rather annoying if that is actually the reason, you could test it out by using a constant for when loop_id is None. (We would be exposed to the same bug though).

Kludex:
I think this solves a different issue than the one it's supposed to close.

It's more about when the `asyncio.Event`s are created. There's a PR about moving to `anyio` that we do that.

---

I don't think we should cache based on thread. I think the way to go is to move away from the cache. More context: https://github.com/pydantic/pydantic-ai/pull/1695

DouweM:
> It's more about when the asyncio.Events are created. There's a PR about moving to anyio that we do that.

@Kludex Would you expect github.com/pydantic/pydantic-ai/pull/3012 to fix https://github.com/pydantic/pydantic-ai/issues/3240 then?

> I don't think we should cache based on thread. I think the way to go is to move away from the cache. More context: https://github.com/pydantic/pydantic-ai/pull/1695

@Kludex I've created a new issue, curious about your thoughts there: https://github.com/pydantic/pydantic-ai/issues/3913

Kludex:
> @Kludex Would you expect github.com/[/pull/3012](https://github.com/pydantic/pydantic-ai/pull/3012) to fix #3240 then?

That's my expectation, I may be wrong, but I think we can get the lock changes from that PR and create a separate PR.

dsfaccini:
I can repurpose this branch for that.

dsfaccini:
I'm investigating this a bit and for one, httpx is moving __away__ from anyio bc of perf reasons https://github.com/agronholm/anyio/discussions/760

secondly httpx doesn't want to support AsyncClient being used in multiple event loops https://github.com/encode/httpx/issues/2058#issuecomment-1960608385

and thirdly, even though httpcore uses anyio internally, anyio's asyncio backend still creates actual `asyncio.Event()` objects that get bound to the event loop at creation time - see `StreamProtocol.connection_made()` in   https://github.com/agronholm/anyio/blob/master/src/anyio/_backends/_asyncio.py#L1183-L1184

so it doesn't seem to me like it will make a difference whether we implement `anyio.Locks` or do the full anyio migration because the event loop binding will still happen in anyio's asyncio backend when connections are established

so, in the face of that, at least until v2, maybe a fix in this PR should be introduced?

DouweM:
@Kludex Since https://github.com/encode/httpx/issues/2058 makes it clear clients shouldn't be used across event loops, we need something like this fix, right? I agree we should move away from caching entirely, but can't until v2.

@dsfaccini Did you verify this actually fixes the `asyncio.Event` error when running in Cloud Run? Because as Marcelo mention, t's not obvious that this change would fix that error.

dsfaccini:
@DouweM nope haven't made sure, will keep pursuing this

Kludex:
@DouweM I would try to move away from cache in a way that is not breaking - and can be done before v2. Or move towards v2 faster...

DouweM:
@Kludex Would love your thoughts on https://github.com/pydantic/pydantic-ai/issues/3913 :)

dsfaccini:
closing in favor of https://github.com/pydantic/pydantic-ai/issues/3913#issuecomment-3719282424

PR #3914, opened 2026-01-05
docs: add Bedrock application inference profiles documentation

psg2:
## Summary

Add documentation for using AWS Bedrock custom application inference profiles with `BedrockConverseModel`.

## Changes

- Added new "Using AWS ApplicationInference Profiles" section to `docs/models/bedrock.md`
- Explains how to use inference profile ARNs while maintaining model feature detection
- Includes code example showing proper usage of the `profile` parameter

## Context

When using custom application inference profiles for cost tracking, the profile ARN doesn't contain enough information for Pydantic AI to determine the underlying model. This documentation shows users how to:

1. Pass the inference profile ARN as the model name
2. Use the `profile` parameter to specify the logical model name for feature detection

This ensures that Pydantic AI can correctly identify model capabilities (streaming, tool use, caching, etc.) while still using the custom inference profile for cost tracking.

### DIFF
--- a/docs/models/bedrock.md
+++ b/docs/models/bedrock.md
@@ -284,11 +284,11 @@ agent = Agent(model)
 ...
 ```
 
-## Using AWS ApplicationInference Profiles
+## Using AWS Application Inference Profiles
 
-AWS Bedrock supports [custom application inference profiles](https://docs.aws.amazon.com/bedrock/latest/userguide/inference-profiles-create.html) for cost tracking and resource management. When using these profiles, you may want to specify the model name explicitly to maintain Pydantic AI's model and provider inference features.
+AWS Bedrock supports [custom application inference profiles](https://docs.aws.amazon.com/bedrock/latest/userguide/inference-profiles-create.html) for cost tracking and resource management. When using these profiles, you should specify a [model profile](https://ai.pydantic.dev/models/overview/#models-and-providers) to ensure Pydantic AI can correctly identify model capabilities (streaming, tool use, caching, etc.) while still using the custom inference profile for cost tracking.
 
-Without explicit configuration, an inference profile ARN like `arn:aws:bedrock:us-east-2:*****:application-inference-profile/****` doesn't contain enough information for Pydantic AI to determine the underlying model or provider. You can work around this by:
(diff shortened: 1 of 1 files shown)


docs/models/bedrock.md
```diff
@@ -284,6 +284,41 @@ agent = Agent(model)
 ...
 ```
 
+## Using AWS ApplicationInference Profiles
+
+AWS Bedrock supports [custom application inference profiles](https://docs.aws.amazon.com/bedrock/latest/userguide/inference-profiles-create.html) for cost tracking and resource management. When using these profiles, you may want to specify the model name explicitly to maintain Pydantic AI's model and provider inference features.
```
DouweM: - Instead of "maintain Pydantic AI's model and provider inference features", let's explain it's for "model profile" selection, with a link to https://ai.pydantic.dev/models/overview/#models-and-providers where it's explained what that means. 
	- It's a bit confusing that AWS and us both use the word profile here for very different things 😅 
- "you may want to" is not strong enough; to get the right behavior you really should be specifying a model profile in this case if the underlying model could not be identified from the model name.

docs/models/bedrock.md
```diff
@@ -284,6 +284,41 @@ agent = Agent(model)
...
+
+Without explicit configuration, an inference profile ARN like `arn:aws:bedrock:us-east-2:*****:application-inference-profile/****` doesn't contain enough information for Pydantic AI to determine the underlying model or provider. You can work around this by:
```
DouweM: "or provider" should be dropped, it's really about the model's capabilities.

docs/models/bedrock.md
```diff
@@ -284,6 +284,41 @@ agent = Agent(model)
...
+    profile=profile,  # Provides the logical model name for feature detection
+)
+
+agent = Agent(model)
+```
+
+This approach ensures that:
+
+- Pydantic AI correctly identifies the model capabilities (streaming, tool use, caching, etc.)
+- Your requests use the custom inference profile for cost tracking
```
DouweM: I prefer not to have additional text after the example, if it could be part of the explanation above. In this case I don't think this section is strictly necessary anyway; it could be merged with the text above.

PR #3915, opened 2026-01-05
ci: force labels on PRs

Kludex:
Also, create sections on the release notes.


DouweM:
Note to self: after merging this, making the new job required in GH settings "require status checks to pass before merging"

PR #3918, opened 2026-01-05
Ensure `stream_output()` always calls output function/validator with `partial_output=False` at the end of streaming

DouweM:
### Pre-Review Checklist

### Pre-Merge Checklist

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/result.py
+++ b/pydantic_ai_slim/pydantic_ai/result.py
@@ -84,9 +84,9 @@ class AgentStream(Generic[AgentDepsT, OutputDataT]):
             response = self.response
             # Final validation with allow_partial=False (the default).
             # We always yield the final result even if the content matches the last partial yield, because:
-            # 1. Output validators/functions receive partial_output=False only on this final call
-            # 2. Partial validation may succeed where full validation fails (e.g., missing required fields)
-            # 3. Users can rely on the last yielded item being the fully validated output
+            # 1. Output validators/functions receive partial_output=False only on this final call,
+            #    and may behave differently based on that flag
+            # 2. Users can rely on the last yielded item being the fully validated output
             self._cached_output = await self.validate_response_output(response)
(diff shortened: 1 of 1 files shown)


DouweM:
@dsfaccini Can you please add a test that verifies that in a scenario like the ones that had their snapshots updated, we call `partial_output=True` a few times, before a final `partial_output=False`? Following the same testing framework @Danipulok introduced for the rest of these

Danipulok:
Well, I got scared when I got a notification, and I thought that my PR had some bugged code, but it seems I just kept the bugged behavior
Glad it was not me, then :D

DouweM:
@Danipulok Sorry for scaring you 😄 This was indeed a pre-existing issue

dsfaccini:
sorry for the test failures, the folder I used for this branch is busy so I'll tackle this tomorrow

pydantic_ai_slim/pydantic_ai/result.py
```diff
@@ -80,13 +80,14 @@ async def stream_output(self, *, debounce_by: float | None = 0.1) -> AsyncIterat
...
-            not last_response or response.parts != last_response.parts
-        ):
-            # Final validation (allow_partial=False)
-            self._cached_output = await self.validate_response_output(response, allow_partial=False)
+        if self._raw_stream_response.final_result_event is not None:  # pragma: no branch
+            response = self.response
+            # Final validation with allow_partial=False (the default).
+            # We always yield the final result even if the content matches the last partial yield, because:
+            # 1. Output validators/functions receive partial_output=False only on this final call
+            # 2. Partial validation may succeed where full validation fails (e.g., missing required fields)
```
DouweM: This isn't right; partial validation will still only return valid objects, it may just have partial input JSON.

DouweM: "Partial validation may succeed where full validation fails" is true when the output validator/function depended on `partial_output` of course. So I think we can combine points 1 and 2

PR #3921, opened 2026-01-06
Revert "Use in-memory SQLite for DBOS tests"

Kludex:
Reverts pydantic/pydantic-ai#3800

@qianl15 it seems the CI has become flaky on dbos tests because of this, I'll revert just to not bother others. But we can work out later - it doesn't need to be in memory, we can also change the fixture to use `tmp_path` instead, so we don't need to check the existence, and pytest will automatically drop the file after the test is completed.


qianl15:
Got it! Sounds like a good plan.

PR #3923, opened 2026-01-06
Fix OTel attributes for `run_stream_sync` by adding context manager support

juanfabrega:
When using Agent.run_stream_sync(), OTel spans were broken because the async context manager lifecycle was split across multiple run_until_complete() calls, causing spans to be entered/exited in different contexts.

The fix adds proper context manager support to StreamedRunResultSync:
- __enter__ runs the entire async streaming lifecycle in a single run_until_complete() call, ensuring OTel spans close properly
- Streamed data is pre-collected and cached for later iteration
- Added asynccontextmanager_from_generator() utility to convert the async generator into a proper async context manager

Usage:
    with agent.run_stream_sync('Hello') as result:
        for chunk in result.stream_output():
            print(chunk)

- Builds on https://github.com/pydantic/pydantic-ai/pull/3716 (which is set to fix https://github.com/pydantic/pydantic-ai/issues/3714)

- Closes #<issue>

### Pre-Review Checklist

### Pre-Merge Checklist

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/agent/abstract.py
+++ b/pydantic_ai_slim/pydantic_ai/agent/abstract.py
@@ -726,9 +726,9 @@ class AbstractAgent(Generic[AgentDepsT, OutputDataT], ABC):
         agent = Agent('openai:gpt-4o')
 
         def main():
-            response = agent.run_stream_sync('What is the capital of the UK?')
-            print(response.get_output())
-            #> The capital of the UK is London.
+            with agent.run_stream_sync('What is the capital of the UK?') as response:
+                print(response.get_output())
+                #> The capital of the UK is London.
         ```
 
         Args:
--- a/pydantic_ai_slim/pydantic_ai/result.py
+++ b/pydantic_ai_slim/pydantic_ai/result.py
@@ -1,7 +1,9 @@
 from __future__ import annotations as _annotations
 
+import asyncio
 import inspect
 import threading
+import warnings
 from collections.abc import AsyncIterator, Awaitable, Callable, Iterable, Iterator
 from contextlib import asynccontextmanager
 from copy import deepcopy
@@ -44,13 +46,6 @@ __all__ = (
 
 T = TypeVar('T')
 """An invariant TypeVar."""
-
-
-async def _wait_briefly() -> None:
-    """Brief async sleep to allow polling without blocking."""
-    import asyncio
-
-    await asyncio.sleep(0.01)
 
 
 @dataclass(kw_only=True)
@@ -686,7 +681,7 @@ class StreamedRunResultSync(Generic[AgentDepsT, OutputDataT]):
         runs entirely in a background thread, ensuring proper OTel span handling.
         Streaming still works normally - chunks are yielded as they arrive.
         """
-        if not hasattr(self, '_stream'):
+        if self._streamed_run_result:
             # Already have a direct StreamedRunResult, nothing to do
             self._context_entered = True
             return self
@@ -721,6 +716,16 @@ class StreamedRunResultSync(Generic[AgentDepsT, OutputDataT]):
         if self._exception is not None and exc_type is None:
             raise self._exception  # pragma: no cover
 
+    def _warn_if_not_context_manager(self) -> None:
+        """Emit a deprecation warning if streaming methods are called without using the context manager."""
+        if not self._context_entered and not self._streamed_run_result:
+            warnings.warn(
+                'Using run_stream_sync() without the context manager pattern is deprecated. '
+                'Use `with agent.run_stream_sync(...) as result:` instead for proper OpenTelemetry span handling.',
+                DeprecationWarning,
+                stacklevel=3,
+            )
+
     def _run_async_stream(self) -> None:
         """Run the async stream lifecycle in a background thread.
 
@@ -739,7 +744,7 @@ class StreamedRunResultSync(Generic[AgentDepsT, OutputDataT]):
                 # Wait for the sync consumer to signal they're done
                 # We use a polling approach since we can't await a threading.Event
                 while not self._stream_done.is_set():
-                    await _wait_briefly()
+                    await asyncio.sleep(0.01)
             except BaseException as e:
                 self._exception = e
                 self._stream_ready.set()  # Signal ready even on error
@@ -833,6 +838,7 @@ class StreamedRunResultSync(Generic[AgentDepsT, OutputDataT]):
         Returns:
             An iterable of the response data.
         """
+        self._warn_if_not_context_manager()
         return self._async_iterator_to_sync(lambda result: result.stream_output(debounce_by=debounce_by))
 
     def stream_text(self, *, delta: bool = False, debounce_by: float | None = 0.1) -> Iterator[str]:
@@ -848,6 +854,7 @@ class StreamedRunResultSync(Generic[AgentDepsT, OutputDataT]):
                 Debouncing is particularly important for long structured responses to reduce the overhead of
                 performing validation as each token is received.
         """
+        self._warn_if_not_context_manager()
         return self._async_iterator_to_sync(lambda result: result.stream_text(delta=delta, debounce_by=debounce_by))
 
     def stream_responses(self, *, debounce_by: float | None = 0.1) -> Iterator[tuple[_messages.ModelResponse, bool]]:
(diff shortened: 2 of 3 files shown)


pydantic_ai_slim/pydantic_ai/result.py
```diff
@@ -622,9 +622,36 @@ async def _marked_completed(self, message: _messages.ModelResponse | None = None
...
+    with agent.run_stream_sync('Hello') as result:
+        for chunk in result.stream_output():
+            print(chunk)
+            #> success
+            #> success (no
+            #> success (no tool
+            #> success (no tool calls)
+    ```
+
+    When used as a context manager, all streamed data is collected when entering the context.
```
DouweM: Unfortunately I think this defeats the point of streaming, as you wouldn't be able to output the chunks as they come in, you'll just see nothing for a while and then all the chunks printed in one go. So in that case, why not just use `run_sync`?

DouweM: This sync/async code is all quite cursed, but I feel like it should be possible to have a sync contextmanager that wraps the async `run_stream` contextmanager and calls its `__aexit__` on `__exit__`, possibly in the same event loop by holding onto it in an instance variable...

juanfabrega: Ok, appreciate the feedback. Let me think more on this and see if we can get something that works.

pydantic_ai_slim/pydantic_ai/result.py
```diff
@@ -635,6 +673,84 @@ def __init__(
...
+        self._exception = None
+
+    def __enter__(self) -> StreamedRunResultSync[AgentDepsT, OutputDataT]:
+        """Enter the context manager, starting the async stream in a background thread.
+
+        When using the context manager pattern, the async context manager lifecycle
+        runs entirely in a background thread, ensuring proper OTel span handling.
+        Streaming still works normally - chunks are yielded as they arrive.
+        """
+        if not hasattr(self, '_stream'):
```
DouweM: I'd rather this be `if self._streamed_run_result`

pydantic_ai_slim/pydantic_ai/result.py
```diff
@@ -45,6 +46,13 @@
 """An invariant TypeVar."""
 
 
+async def _wait_briefly() -> None:
+    """Brief async sleep to allow polling without blocking."""
+    import asyncio
+
+    await asyncio.sleep(0.01)
```
DouweM: Let's inline this

pydantic_ai_slim/pydantic_ai/result.py
```diff
@@ -622,9 +630,39 @@ async def _marked_completed(self, message: _messages.ModelResponse | None = None
...
+    """Synchronous wrapper for [`StreamedRunResult`][pydantic_ai.result.StreamedRunResult] that only exposes sync methods.
+
+    This class should be used as a context manager to ensure proper cleanup of resources
+    and correct OpenTelemetry instrumentation:
+
+    ```python
+    from pydantic_ai import Agent
+
+    agent = Agent('test')
+    with agent.run_stream_sync('Hello') as result:
```
DouweM, line 637: Can you please update remaining `run_stream_sync` docs to use the contextmanager as well?

If possible, I'd also like for the type checker to complain if you use `run_stream_sync` without a contextmanager, _while keeping it working at runtime_, to nudge people to the improved approach without breaking their existing code.

pydantic_ai_slim/pydantic_ai/result.py
```diff
@@ -635,6 +673,84 @@ def __init__(
...
+        """Run the async stream lifecycle in a background thread.
+
+        This method runs the entire async context manager lifecycle in a single
+        run_until_complete() call, ensuring OTel context is properly maintained.
+        """
+
+        async def consume_stream() -> None:
+            try:
+                # Enter the async context manager by getting the result
+                result = await anext(self._stream)
```
DouweM, line 739: Do we need `anext` here and below, or could we use `async with self._stream` here?

DouweM (changes requested):
@juanfabrega Thanks for your help Juan, using a thread is obviously not ideal but I agree it may be the best we can do.

pydantic_ai_slim/pydantic_ai/result.py
juanfabrega: fixed

pydantic_ai_slim/pydantic_ai/result.py
juanfabrega: updated. I added a deprecation warning so users are warned at runtime but old code shouldn't be "broken" in any way.

pydantic_ai_slim/pydantic_ai/result.py
juanfabrega: done!

pydantic_ai_slim/pydantic_ai/result.py
```diff
@@ -635,6 +668,94 @@ def __init__(
...
+        self._exception = None
+
+    def __enter__(self) -> StreamedRunResultSync[AgentDepsT, OutputDataT]:
+        """Enter the context manager, starting the async stream in a background thread.
+
+        When using the context manager pattern, the async context manager lifecycle
+        runs entirely in a background thread, ensuring proper OTel span handling.
+        Streaming still works normally - chunks are yielded as they arrive.
+        """
+        if self._streamed_run_result:
```
DouweM, line 684: Should we also check `self._context_entered`?

pydantic_ai_slim/pydantic_ai/result.py
```diff
@@ -635,6 +668,94 @@ def __init__(
...
+        # Wait for the stream to be ready
+        self._stream_ready.wait()
+
+        # Check if an exception occurred during initialization
+        if self._exception is not None:
+            raise self._exception
+
+        return self
+
+    def __exit__(self, exc_type: type[BaseException] | None, exc_val: BaseException | None, exc_tb: Any) -> None:
```
DouweM, line 703: Let's validate `self._context_entered` is set here as well

pydantic_ai_slim/pydantic_ai/result.py
```diff
@@ -635,6 +668,94 @@ def __init__(
...
+        """
+        if self._thread is not None:
+            # Signal that we're done consuming (in case iteration didn't complete)
+            self._stream_done.set()
+            # Wait for the background thread to complete
+            self._thread.join()
+
+        # Re-raise any exception from the background thread
+        if self._exception is not None and exc_type is None:
+            raise self._exception  # pragma: no cover
```
DouweM, line 717: Could we make sure this is covered?

pydantic_ai_slim/pydantic_ai/result.py
```diff
@@ -717,6 +838,7 @@ def stream_output(self, *, debounce_by: float | None = 0.1) -> Iterator[OutputDa
         Returns:
             An iterable of the response data.
         """
+        self._warn_if_not_context_manager()
```
DouweM, line 841: Would it make sense for this be in `_async_iterator_to_sync`? Or `_async_to_sync`?

pydantic_ai_slim/pydantic_ai/result.py
DouweM: Looks like it's still relevant!

tests/test_streaming.py
```diff
@@ -3384,36 +3423,102 @@ def test_run_stream_sync_instrumentation(capfire: CaptureLogfire):
...
+        ]
+    )
+
+
+def test_run_stream_sync_context_manager_exception_in_stream():
+    """Test that exceptions during streaming are properly propagated.
+
+    This exercises the exception handling path in _run_async_stream and __enter__.
+    """
+    from pydantic_ai.models.function import DeltaToolCall
```
DouweM, line 3498: Imports at the top please

pydantic_ai_slim/pydantic_ai/agent/abstract.py
```diff
@@ -726,9 +726,9 @@ def run_stream_sync(
         agent = Agent('openai:gpt-4o')
 
         def main():
-            response = agent.run_stream_sync('What is the capital of the UK?')
-            print(response.get_output())
-            #> The capital of the UK is London.
+            with agent.run_stream_sync('What is the capital of the UK?') as response:
```
DouweM, line 729: Would it be possible to change the typing of `run_stream_sync` here to make it a type error to use it without `with`, while keeping it working? By hinting it as returning a contextmanager.

pydantic_ai_slim/pydantic_ai/result.py
```diff
@@ -635,6 +668,94 @@ def __init__(
...
+
+        # Re-raise any exception from the background thread
+        if self._exception is not None and exc_type is None:
+            raise self._exception  # pragma: no cover
+
+    def _warn_if_not_context_manager(self) -> None:
+        """Emit a deprecation warning if streaming methods are called without using the context manager."""
+        if not self._context_entered and not self._streamed_run_result:
+            warnings.warn(
+                'Using run_stream_sync() without the context manager pattern is deprecated. '
```
DouweM, line 723: Wrap the function name in backticks please

DouweM:
@juanfabrega Hey Juan, I'm still very interested in this fix so I've removed the Stale label!

Issue #3913, opened 2026-01-05
Move away from the http client cache


DouweM:
### Description

Causes many issues:
- https://github.com/pydantic/pydantic-ai/issues/3240
- https://github.com/pydantic/pydantic-ai/issues/748

Attempts at fixing (parts of) this without removing the cache:
- https://github.com/pydantic/pydantic-ai/pull/1695
- https://github.com/pydantic/pydantic-ai/pull/3910

Other related issues/PRs:
- https://github.com/pydantic/pydantic-ai/pull/3905

DouweM:
@Kludex We discussed this before. Was the idea that we'd let users manage the `httpx.AsyncClient` and pass it into the `run`? Because the provider/model/agent would've already been created at that point, so it can't be passed into there. 

And then if no client is provided we'd create a new one just for that request (accepting the extra SSL negotiation overhead etc)?

In a sync context like Google Cloud Run (https://github.com/pydantic/pydantic-ai/issues/3240), you couldn't do `async with httpx.AsyncClient() as http_client:...`, but I suppose you can just pass in the not-entered-yet client, and manually close it afterwards (since your code is sync anyway).

DouweM:
Talking to Marcelo, we should be able to:
- remove `cached_async_http_client` entirely
- always create a new `httpx.AsyncClient` in `Provider.__init__`
- in `Provider.__del__`, close the client _if it was created by us and not provided by the user_
- make Provider and Model asynccontextmanagers that cleanly enter and exit the httpclient when appropriate
- make `async with agent:` also enter the model (which enters the provider)
- because the same model/provider can be used multiple times in parallel (e.g. with subagents), and since the http client cannot be entered multiple times, we may need to create a new http + api client in `Provider.__aenter__` (unless of course the user provided their own http client, in which case we just pass that through and never open/close it). 
	- We can't use the `entered_count` "solution" from `MCPServer` as it has issues like https://github.com/pydantic/pydantic-ai/issues/2818
		- Actually maybe we can, because the http/SDK client may not have that cancel scope issue

Wh1isper:
> let users manage the httpx.AsyncClient 

Sounds good. I run hundreds of agents in a single process, and without a cached client, there might be performance issues.

DouweM:
@Wh1isper Yeah we want to make sure that in typical patterns performance doesn't suffer too much (honestly the amount of bugs this caching has caused is worth a small hit), and that it remains easy to use your own client.

[...]

PR #3936, opened 2026-01-07
Drop httpx cache

Kludex:
- Fixes https://github.com/pydantic/pydantic-ai/issues/3913 and related issues


DouweM:
I'll close this as the context of what needs to happen is in https://github.com/pydantic/pydantic-ai/issues/3913 and you said you wouldn't have time to continue on this PR so it'll be easier for whoever ges to it to submit a new one.

Issue #1771, opened 2025-05-19
support batch processing


darjeeling:
### Description

to reduce pricing, I think it's good to support batch process.

some models support that. how's your priority?

[anthropic batch process](https://docs.anthropic.com/en/docs/build-with-claude/batch-processing#message-batches-api)
[OpenAI Batch API](https://platform.openai.com/docs/guides/batch)

samuelcolvin:
From @ bjorkbjork on [slack](https://pydanticlogfire.slack.com/archives/C083V7PMHHA/p1751528009610749)

> We've built a dynamic factory that compiles Agents, Tools, and MCP servers from database records into the PydanticAI classes at runtime, and iterates the agent.
> Right now our main suite of agents that we'll be providing to the customer are a series of Root cause analysis agents - they ingest a lot of data from a clickhouse MCP server, and use this to inform their analysis and search for leads.
> There is a very big data source to look at, and we're getting rate limited by Anthropic. We reached out to them about raising the TPM limits, and they're trying to push us onto batch processing.
> If it were supported, what we would likely do is replace the ch_execute_query tool we have with an agent that has access to query the database (anthropic docs claim batches support tool usage) - we send the queries as batched runs that query the data, and then analyze and provide back insights (going through the huge token data sources as batched runs rather than sync), then wait for the completion.
> We'd probably run a lot of those batches in parallel, as the main agent jumps around the DB looking for clues that can inform new queries.
> If it's on the backburner I'll just continue without backed processes for now.

i suspect with a few hours of engineering and some async foo, you could do this now.

Ideally you'd make that implementation open source, and it would become the foundation for the built in implementation.

Here's my idea:
1. The idea relies on the fact that the `requests.params` matches the standard API format ([docs](https://docs.anthropic.com/en/api/creating-message-batches))
2. You create a new `BatchModel` implementation of `Model`: perhaps inheriting from [`WrapperModel`](https://github.com/pydantic/pydantic-ai/blob/49658f3e5fa701c8d6470872a097ae81d02ccf77/pydantic_ai_slim/pydantic_ai/models/wrapper.py#L16C7-L16C19) and wraps the `AnthropicModel`, or perhaps uses some methods from the anthropic model etc. but doesn't wrap it
3. Model has some (fairly complex) to create batches of messages and fire a request when enough messages come in, then wait for the response and yield the right responses to the right waiting tasks - this is the hard bit, might work best with anyio channels?
[...]

PR #3937, opened 2026-01-07
feat: Add Batch Processing API with OpenAI, Anthropic & Google support

sarth6:
## Summary

Add batch processing support to pydantic-ai for **OpenAI**, **Anthropic**, and **Google** providers. This enables users to submit multiple LLM requests as a single job for asynchronous processing, achieving **50% cost savings** on API calls.

**Partially addresses #1771** - implements core batch infrastructure, `BatchModel` wrapper, and shared utilities.

### High-Level API (Recommended)

```python
from pydantic_ai import ModelRequest
from pydantic_ai.direct import model_request_batch

results = await model_request_batch(
    'openai:gpt-4o-mini',
    [
        ('q1', [ModelRequest.user_text_prompt('What is 2+2?')]),
        ('q2', [ModelRequest.user_text_prompt('What is 3+3?')]),
    ],
    poll_interval=60.0,  # Auto-polls until complete
)

for result in results:
    if result.is_successful:
        print(f'{result.custom_id}: {result.response.parts[0].content}')
```

### BatchModel for Agent Integration

```python
from pydantic_ai import Agent
from pydantic_ai.models.batch import BatchModel

async with BatchModel('openai:gpt-4o-mini', batch_size=100) as batch_model:
    agent = Agent(model=batch_model)
    tasks = [agent.run(prompt) for prompt in prompts]
    await batch_model.submit()  # Or auto-submits at batch_size
    results = await asyncio.gather(*tasks)
```

**BatchModel features:**
- `batch_size` - auto-submit threshold
- `should_submit` - custom callback for submit logic
- `submit()` - manual submission
- Context manager support

### Check Batch Capability

```python
from pydantic_ai.models import supports_batch

model = OpenAIChatModel('gpt-4o-mini')
if supports_batch(model):
    batch = await model.batch_create(requests)
```

### Low-Level API (Advanced)

For custom polling logic or long-running jobs:

```python
from pydantic_ai.direct import batch_create, batch_status, batch_results

batch = await batch_create('openai:gpt-4o-mini', requests)
# ... store batch.id, resume later ...
batch = await batch_status('openai:gpt-4o-mini', batch)
results = await batch_results('openai:gpt-4o-mini', batch)
```

## Provider Support

| Provider | Status | Notes |
|----------|--------|-------|
| OpenAI | ✅ Supported | 50% discount, 24-hour processing window |
| Anthropic | ✅ Supported | 50% discount, 24-hour expiration |
| Google | ✅ Supported | Batch prediction API |

### Provider Differences

| Capability | OpenAI | Anthropic | Google |
|------------|--------|-----------|--------|
| Different model per request | ❌ | ✅ | ❌ |
| Different tools per request | ✅ | ✅ | ✅ |
| Different temperature per request | ✅ | ✅ | ✅ |

## Architecture

This PR includes shared utilities for extensibility:

- **`BatchCapable` protocol** - Type-safe interface for batch-capable models
- **`supports_batch()` function** - Runtime check for batch support
- **`_batch_utils.py` module** - Shared utilities:
  - `extract_batch_error` - Handles error extraction from various provider formats
  - `parse_batch_datetime` - Handles Unix timestamps, ISO strings, datetime objects
  - `BatchResultBuilder` - Tracks processed IDs to prevent duplicates
- **Implementer guide** - `docs/adding-batch-support.md` for adding new providers

### New Files
- `docs/batch-processing.md` - Dedicated documentation page
- `docs/adding-batch-support.md` - Guide for implementing batch support in new providers
- `pydantic_ai_slim/pydantic_ai/models/_batch_utils.py` - Shared batch utilities
- `pydantic_ai_slim/pydantic_ai/models/batch.py` - BatchModel for agent integration
- `tests/models/test_batch_utils.py` - Utility tests (37 tests)
- `tests/models/test_openai_batch.py` - OpenAI batch tests
- `tests/models/test_anthropic_batch.py` - Anthropic batch tests
- `tests/models/test_google_batch.py` - Google batch tests
- `tests/models/test_batch_model.py` - BatchModel tests

### Modified Files
- `pydantic_ai_slim/pydantic_ai/__init__.py` - Export BatchCapable, supports_batch
- `pydantic_ai_slim/pydantic_ai/models/__init__.py` - Batch abstractions, BatchCapable protocol
- `pydantic_ai_slim/pydantic_ai/models/openai.py` - OpenAI batch implementation
- `pydantic_ai_slim/pydantic_ai/models/anthropic.py` - Anthropic batch implementation
- `pydantic_ai_slim/pydantic_ai/models/google.py` - Google batch implementation
- `pydantic_ai_slim/pydantic_ai/models/wrapper.py` - Batch method forwarding
- `pydantic_ai_slim/pydantic_ai/direct.py` - High-level and low-level batch functions

## Future Roadmap

Items not yet implemented (potential follow-up PRs):

- [ ] **Cost tracking/limits** - Per-batch and per-request cost estimation
- [ ] **Dry-run mode** - Preview costs before submitting
- [ ] **State persistence** - Resume interrupted batches from stored batch IDs
- [ ] **Progress callbacks** - Real-time status updates during polling

## Test Plan

- [x] 153 tests covering all batch functionality
- [x] Unit tests for Batch/BatchStatus/BatchResult/BatchError
- [x] Shared utilities tests (37 tests)
- [x] OpenAI batch implementation with mock client
- [x] Anthropic batch implementation with mock client
- [x] Google batch implementation with mock client
- [x] BatchModel wrapper tests
- [x] Direct API functions (async and sync variants)
- [x] Wrapper model forwarding (InstrumentedModel, WrapperModel)
- [x] Error handling (API errors, validation errors, timeout)
- [x] Documentation examples tested via pytest-examples


docs/direct.md
```diff
@@ -6,11 +6,20 @@ These methods are thin wrappers around the [`Model`][pydantic_ai.models.Model] i
...
 The following functions are available:
 
+**Single requests:**
+
 - [`model_request`][pydantic_ai.direct.model_request]: Make a non-streamed async request to a model
 - [`model_request_sync`][pydantic_ai.direct.model_request_sync]: Make a non-streamed synchronous request to a model
 - [`model_request_stream`][pydantic_ai.direct.model_request_stream]: Make a streamed async request to a model
 - [`model_request_stream_sync`][pydantic_ai.direct.model_request_stream_sync]: Make a streamed sync request to a model
 
+**Batch processing (OpenAI only):**
```
DouweM: This feature deserves its own doc

docs/direct.md
```diff
@@ -6,11 +6,20 @@ These methods are thin wrappers around the [`Model`][pydantic_ai.models.Model] i
...
+
+- [`batch_create`][pydantic_ai.direct.batch_create]: Submit a batch of requests for asynchronous processing
+- [`batch_status`][pydantic_ai.direct.batch_status]: Check the status of a batch job
+- [`batch_results`][pydantic_ai.direct.batch_results]: Retrieve results from a completed batch
+- [`batch_cancel`][pydantic_ai.direct.batch_cancel]: Cancel an in-progress batch
```
DouweM: Instead of 4 methods, could we have just `model_request_batch()` that can be awaited, canceled via `asyncio.create_task(model_request_batch(...)).cancel()`, and eventually returns all results? These methods are too low-level for my taste; they could make sense on `Model` or a new `Batch` object, but this is not the API I'd want users to deal with directly.

docs/direct.md
```diff
@@ -149,3 +158,189 @@ print(model_response.parts[0].content)
...
+    questions = [
+        ('math-1', 'What is 2 + 2?'),
+        ('math-2', 'What is the square root of 144?'),
+        ('science-1', 'What is photosynthesis?'),
+        ('science-2', 'Explain gravity in one sentence.'),
+    ]
+
+    # Build batch requests - each is a (custom_id, messages) tuple
+    requests = [
+        (custom_id, [ModelRequest.user_text_prompt(question)])
```
DouweM: Beside model request, we need to be able to specify `model_settings` and `model_request_parameters

docs/direct.md
```diff
@@ -149,3 +158,189 @@ print(model_response.parts[0].content)
...
+        for custom_id, question in questions
+    ]
+
+    # Submit batch (50% cost savings!)
+    batch = await batch_create('openai:gpt-4o-mini', requests)
+    print(f'Batch {batch.id} submitted with {batch.request_count} requests')
+
+    # Poll for completion (batches typically complete within 24 hours)
+    while not batch.is_complete:
```
DouweM: I'd expect the status/sleep loop to be exposed either through a `wait()` method or something on `Batch`, most likely hidden inside the `model_request_batch` method I mentioned above.

docs/direct.md
```diff
@@ -149,3 +158,189 @@ print(model_response.parts[0].content)
...
+        print(f'Status: {batch.status} - waiting...')
+        await asyncio.sleep(60)  # Check every minute
+        batch = await batch_status('openai:gpt-4o-mini', batch)
```
DouweM: Why do we need to keep specifying the model name? What happens if it doesn't match the original name?

docs/direct.md
```diff
@@ -149,3 +158,189 @@ print(model_response.parts[0].content)
...
+    ]
+
+    # Build batch requests - each is a (custom_id, messages) tuple
+    requests = [
+        (custom_id, [ModelRequest.user_text_prompt(question)])
+        for custom_id, question in questions
+    ]
+
+    # Submit batch (50% cost savings!)
+    batch = await batch_create('openai:gpt-4o-mini', requests)
```
DouweM: Do all batch requests _need_ to match on model name? It'd be nice if we could vary every aspect of the request (model name, messages, settings, model request param) for each batch job

docs/direct.md
DouweM: I also want to make sure this sets us up for the `BatchModel` described in https://github.com/pydantic/pydantic-ai/issues/1771#issuecomment-3031440960 that will be much easier to use with agents. Can you please add that to this PR so we can be sure we're building the right primitives for it? I'd expect it to take:
1. a number of requests until it sends the batch
2. a function for determining whether to send the batch, which can check the size of the requests for example.
3. perhaps a way to never auto-submit and wait for the user to explicitly call `submit()` or something

DouweM: https://github.com/agamm/batchata is also worth checking out

docs/direct.md
```diff
@@ -149,3 +158,189 @@ print(model_response.parts[0].content)
...
+- [`batch_results`][pydantic_ai.direct.batch_results]: Retrieve results when complete (async)
+- [`batch_results_sync`][pydantic_ai.direct.batch_results_sync]: Retrieve results when complete (sync)
+- [`batch_cancel`][pydantic_ai.direct.batch_cancel]: Cancel an in-progress batch (async)
+- [`batch_cancel_sync`][pydantic_ai.direct.batch_cancel_sync]: Cancel an in-progress batch (sync)
+
+### Batch Status and Results
+
+The [`Batch`][pydantic_ai.Batch] object tracks job status:
+
+```python {lint="skip" test="skip"}
```
DouweM: Please don't ever skip linting or testing examples in docs!

docs/direct.md
```diff
@@ -149,3 +158,189 @@ print(model_response.parts[0].content)
...
+
+    requests = [
+        ('calc-1', [ModelRequest.user_text_prompt('What is 15 * 23?')]),
+        ('calc-2', [ModelRequest.user_text_prompt('What is 144 / 12?')]),
+    ]
+
+    batch = await batch_create(
+        'openai:gpt-4o-mini',
+        requests,
+        model_request_parameters=params,
```
DouweM: As mentioned, should be variable per job

docs/direct.md
```diff
@@ -149,3 +158,189 @@ print(model_response.parts[0].content)
...
+    )
+```
+
+!!! note "Tool Execution in Batches"
+    Unlike `Agent.run()`, batch processing does **not** automatically execute tools.
+    If the model returns tool calls, you must execute them yourself and submit
+    a follow-up batch with the tool results. For automatic tool handling, consider
+    using the Agent API for real-time requests or wait for future Agent-level batch support.
```
DouweM: - This applies to the direct API as well, so maybe worth moving up. Or maybe not needed if the rest of the content already makes it clear enough that this will do a single request?
- Since I suggested elsewhere to implement the smart `BatchModel` already, we may not need this, as we can clearly document which approach to use for "manual" batch requests, and which for agents.

docs/direct.md
```diff
@@ -149,3 +158,189 @@ print(model_response.parts[0].content)
...
+
+### Provider Support
+
+Currently, batch processing is supported for:
+
+| Provider | Status | Notes |
+|----------|--------|-------|
+| OpenAI | ✅ Supported | 50% discount, 24-hour processing window |
+| Anthropic | 🔮 Planned | Message Batches API support coming |
```
DouweM: I want to support at least one other provider in this PR, to ensure the abstraction is sufficient.

docs/direct.md
DouweM: Your PR description also mentioned `run_batch`, but I think that'll be tricky as you'd want to be able to vary every `agent.run` argument (assuming the batch APIs support that), which would be difficult to fit into a list passed to `run_batch`.

So perhaps an example with `BatchModel` + `asyncio.gather` + multiple `agent.run` calls will do the trick.

DouweM (changes requested):
@sarth6 I've started by only reviewing the docs and API as the code itself is downstream of that :)

sarth6:
@DouweM Thanks for the detailed feedback! I researched all three major batch APIs (OpenAI, Anthropic, Google Gemini) plus the batchata library you mentioned. Here's what I learned and how it shaped the final design:

### What I learned from provider research

**Key differences that affected API design:**

| Aspect | OpenAI | Anthropic | Google |
|--------|--------|-----------|--------|
| Submission | Upload JSONL file | JSON array in request body | Inline array or JSONL |
| Per-request model | ❌ Must be same | ✅ Can vary per request | ⚠️ Same per batch |
| Result retrieval | Download separate files | Stream JSONL from endpoint | Inline or GCS/BigQuery |

The biggest insight: **Anthropic allows different models per request** in a batch (e.g., mixing Sonnet and Haiku), while OpenAI requires all requests use the same model. This meant our abstraction needed to support per-request parameter flexibility.

**From batchata**, I adopted:
- High-level function with internal polling (exactly what you suggested)
- Graceful cancellation on `asyncio.CancelledError`
- Result categorization per request (success/error)

And avoided:
- Synchronous-only API (we're async-first)
- Fluent builder pattern (overkill for pydantic-ai's style)
- Global mutable state

### What we implemented

**1. High-level API (your suggestion)**

```python
from pydantic_ai.direct import model_request_batch

# Submit, poll, and get results in one call
results = await model_request_batch(
    'openai:gpt-4o-mini',
    requests,
    poll_interval=60.0,
    timeout=3600.0,  # Optional timeout
)

# Cancellation works exactly as you described
task = asyncio.create_task(model_request_batch(...))
task.cancel()  # Cancels the batch job too
```

**2. Low-level methods kept on Model class**

The 4 low-level methods (`batch_create`, `batch_status`, `batch_results`, `batch_cancel`) stay on the `Model` class for advanced use cases (e.g., submit now, resume later with stored `batch.id`). They're also exposed via `direct.py` for users who need custom polling logic.

**3. BatchModel for agent integration**

Added `BatchModel` wrapper as you described in #1771:

```python
from pydantic_ai.models.batch import BatchModel

# Configure batch behavior
batch_model = BatchModel(
    'openai:gpt-4o-mini',
    batch_size=100,                    # Auto-submit at 100 requests
    should_submit=lambda q: len(q) >= 50,  # Or custom logic
    auto_submit=False,                 # Manual submit mode
)

# Use with Agent + asyncio.gather
agent = Agent(model=batch_model)
tasks = [agent.run(prompt) for prompt in prompts]
await batch_model.submit()  # Explicit submit
results = await asyncio.gather(*tasks)
```

**4. Per-request flexibility**

Requests are 2-tuples or 3-tuples to support varying parameters:
```python
requests = [
    ('q1', messages),                    # Uses defaults
    ('q2', messages, custom_params),     # Custom tools/temperature/etc
]
```

For Anthropic, different models per request works. For OpenAI, model must match but other params can vary.

### Provider support

Both OpenAI and Anthropic are implemented and tested. ~~Google is left out, but may implement that in this PR as when choosing only 2 of the big 3 providers it feels a bit like picking a loser 😅~~ Ended up implementing google too 

Let me know if this addresses your concerns or if you'd like any changes to the API!

docs/direct.md
sarth6: Makes sense, anthropic actually allows the user to change the model per batch item - updated the API 👍

sarth6:
## ✨ Per-Request ModelSettings Support Added

This PR now includes full support for per-request settings override\!

### What's New

Each batch request can now specify its own `temperature`, `max_tokens`, and other model settings via `ModelRequestParameters.model_settings`:

```python
from pydantic_ai import ModelRequest, ModelRequestParameters

params = ModelRequestParameters()
requests = [
    ('precise', msgs, replace(params, model_settings={'temperature': 0.2})),
    ('creative', msgs, replace(params, model_settings={'temperature': 0.9})),
]
# Batch-wide settings provide fallback values
batch = await model.batch_create(requests, model_settings={'max_tokens': 100})
```

### Settings Merge Hierarchy

1. **Provider defaults** (lowest priority)
2. **Batch-wide** `model_settings` parameter  
3. **Per-request** `model_settings` in `ModelRequestParameters` (highest priority)

### Implementation Details

- ✅ All 3 providers (OpenAI, Anthropic, Google) support per-request settings
- ✅ Settings are merged using existing `merge_model_settings()` function
- ✅ Fixed `BatchModel` bug where only first request's settings were used
- ✅ 9 new tests (3 per provider) validating override, fallback, and merging
- ✅ Documentation updated with examples
- ✅ 0 type errors, all 107 batch tests pass

### Provider Support

| Provider | Per-Request Settings | Notes |
|----------|---------------------|-------|
| **Anthropic** | ✅ Full support | Different models + all settings per request |
| **OpenAI** | ✅ Supported | Same model required, but temperature/max_tokens vary |
| **Google** | ✅ Supported | Same model required, per-request settings work |

This was the last missing piece for feature parity across providers\! 🎉

sarth6:
Moving back to draft so this can bake a bit more

dsfaccini:

```bash
git pull

deactivate

rm -rf .venv

make install
```

dsfaccini:
hey @sarth6 I see you have a couple rather big PRs in the pipeline and you're heavily using claude

if you want some ideas for local setups for your claude development take a look at my own setup here https://github.com/dsfaccini/pydantic-ai-claude-tools

we want your claude to succeed as much as possible :) and let us know about any learnings you have so we can improve ai devx even more

sarth6:
Thanks @dsfaccini! Sorry for the late reply, just have been busy with some other work.

One thought I had: I could see pydantic ai mantaining a claude code marketplace for this which would contain various claude code plugins for different domains of pydantic ai work. Each plugin would then encapsulate a set of skills, agents, and commands. https://code.claude.com/docs/en/plugin-marketplaces

Here are a few example plugins my team maintains for our work on our own project:
* frontend
* backend (hooks into 4 other plugins: database expert, agent expert, job expert, and api expert)
* tech lead (a lot of linear mcp commands like /triage, /improve-issue, etc)
* oncall (emergency logfire mcp production debugger, retrospective, etc)
* many more

Most of these probably don't translate, and if you think only a single plugin would make sense then that what you have now works - but if you find there are multiple domains being worked on it could be useful (like a maintainer plugin vs a contributor plugin)

Some learnings I've had on my own claude code journey:
* Investing in good plugins is almost always worth it
* Requiring all PRs to update any relevant plugins keeps them up to date and is better for overall velocity (maybe this won't translate well to OSS though, but I have some thoughts below on how this could be done async instead)
* We have some codex users on my team so we use `AGENTS.md` and then have a claude code hook that creates a `CLAUDE.md` symlink for every `AGENTS.md` in the repo
* Claude code skills are fickle, we've found Claude is much more likely to call them when we add the relevant skills to various `AGENTS.md` (`CLAUDE.md`) files in our repo (ex. we have an `app/agents/` subfolder where all our agent work happens, and so in `app/agents/AGENTS.md` we mention a bunch of very useful skills
* If I complain about something twice it should probably be a claude code command
* sub-agents are very powerful for independent parallel tasks - unfortunately claude code doesn't allow recursive subagents, so the main pattern is a fan-out -> fan-in -> fan-out -> etc pattern. It's also important to know when to use an explore agent vs a task agent

I think the main challenge for an open-source repo like this, where most contributors are coming in with zero context or code culture, is maintainers like you will often find yourselves repeating the same review feedback over and over. So I think the #1 goal should be to find a way to take all historical review feedback and ongoing future review feedback and ingest it into offline lessons - I was experimenting with something that does this in https://github.com/sarth6/marketplace-of-ideas/blob/main/plugins/open-source-contributor/commands/review-lessons.md

Many of the code review tools can do this as well (I saw you added devin so maybe it will do that). The benefit of doing it outside of the code review tool is that you can also incorporate those lessons into AGENTS.md's, skills, etc - so that code made by coding agents are more likely to do it right the first time.

Perhaps a github action that uses the claude code action on commit to main could look at the PR and do a comprehensive learning analysis, and incorporate those learnings in a database or a set of markdown files (not sure what the best way is)

Hope this is helpful, I shared a list of lessons I'd compiled so far with @DouweM but it was just on 4-5 reviews I received from him, might need a massive ralph wiggum loop to do it on all historical prs, and will probably want to do it with many parallel task agents.

DouweM:
> So I think the #1 goal should be to find a way to take all historical review feedback and ongoing future review feedback and ingest it into offline lessons

@sarth6 Thanks Sarth, as you've seen I've started doing some of these things in https://github.com/pydantic/pydantic-ai/pull/4169 and https://github.com/pydantic/pydantic-ai/pull/4214 -- I started working on the `braindump` tool I used to generate the latter a couple days before you wrote your comment :)

sarth6:
Saw it's marked stale, happy to pick this up now but I felt best to wait to not overload too many parallel PRs

AlapinEnjoyer:
Any news on this? Wasn't this planned for the march milestone?

sarth6:
@AlapinEnjoyer Douwe and I met today to discuss priorities, this one is high priority but behind my unified thinking PR. I will post an update in the issue with any updates as soon I'm aware.

In terms of the milestones, a lot of the recent milestones were missed so I wouldn't anchor on them, but I'm confident after my meeting today that the team is forward looking on how to maintain velocity while maintaining quality, which is tough right now with coding agents.

If you can provide any data in the issue on what this feature would unlock for you / what having this implemented would change for you, it'd be instrumental in shaping the priorities!

AlapinEnjoyer:
> @AlapinEnjoyer Douwe and I met today to discuss priorities, this one is high priority but behind my unified thinking PR. I will post an update in the issue with any updates as soon I'm aware.
> 
> In terms of the milestones, a lot of the recent milestones were missed so I wouldn't anchor on them, but I'm confident after my meeting today that the team is forward looking on how to maintain velocity while maintaining quality, which is tough right now with coding agents.
> 
> If you can provide any data in the issue on what this feature would unlock for you / what having this implemented would change for you, it'd be instrumental in shaping the priorities!


@sarth6 for me personally this would unlock things like agentic data generation / revision / curation pipelines at a much more reasonable cost. For these types of workloads real-time interaction is not needed, and they are much more constrained by throughput/cost than latency.

It would also let me use the same framework for both online agent execution and offline jobs instead of handling those separately outside pydantic-ai.

A useful reference for the kind of workflow I have in mind is the recent [Tongyi](https://arxiv.org/pdf/2510.24701) paper!

ruaultadrien:
I agree with @AlapinEnjoyer, this feature would be amazing for data extraction tasks that are run in data pipelines. :slightly_smiling_face:

shun-liang:
Also want to voice support for the demand of this feature. It should be quite handy to have it to run evals. Would save a lot of cost.

PR #3941, opened 2026-01-07
Add model catalog

dsfaccini:
https://github.com/user-attachments/assets/43087dfd-162a-41ea-a9c1-5ea2a7b5a4f2

### DIFF
--- a/docs/models/vercel.md
+++ b/docs/models/vercel.md
@@ -17,7 +17,7 @@ You can set the `VERCEL_AI_GATEWAY_API_KEY` and `VERCEL_OIDC_TOKEN` environment 
 ```python
 from pydantic_ai import Agent
 
-agent = Agent('vercel:anthropic/claude-4-sonnet')
+agent = Agent('vercel:anthropic/claude-sonnet-4-5')
 ...
 ```
 
@@ -29,7 +29,7 @@ from pydantic_ai.models.openai import OpenAIChatModel
 from pydantic_ai.providers.vercel import VercelProvider
 
 model = OpenAIChatModel(
-    'anthropic/claude-4-sonnet',
+    'anthropic/claude-sonnet-4-5',
     provider=VercelProvider(api_key='your-vercel-ai-gateway-api-key'),
 )
 agent = Agent(model)
--- a/docs/models/xai.md
+++ b/docs/models/xai.md
@@ -17,7 +17,7 @@ You can set the `GROK_API_KEY` environment variable and use [`GrokProvider`][pyd
 ```python
 from pydantic_ai import Agent
 
-agent = Agent('grok:grok-2-1212')
+agent = Agent('grok:grok-4-fast')
 ...
 ```
 
@@ -29,7 +29,7 @@ from pydantic_ai.models.openai import OpenAIChatModel
 from pydantic_ai.providers.grok import GrokProvider
 
 model = OpenAIChatModel(
-    'grok-2-1212',
+    'grok-4-fast',
     provider=GrokProvider(api_key='your-xai-api-key'),
 )
 agent = Agent(model)


docs/models/openai.md
```diff
@@ -259,483 +259,22 @@ model = OpenAIChatModel(
...
+- [Heroku AI](heroku.md) - Heroku-hosted models
+- [LiteLLM](litellm.md) - Unified interface to 100+ LLM APIs
+- [MoonshotAI](moonshotai.md) - Kimi models
+- [Nebius AI Studio](nebius.md) - Various AI models
+- [Ollama](ollama.md) - Local open-source LLMs
+- [OVHcloud AI Endpoints](ovhcloud.md) - OVHcloud-hosted models
+- [Perplexity](perplexity.md) - Search-optimized models
+- [Together AI](together.md) - Fast inference for open-source models
+- [Vercel AI Gateway](vercel.md) - Unified gateway to multiple providers
+- [xAI (Grok)](xai.md) - Grok models
```
DouweM: Let's not have the descriptions, otherwise people will wonder why e.g. Fireworks and Together are called "fast", but Vercel isn't 😄 Some of htem are also pretty useless like "Models via GitHub".

docs/models/alibaba.md
```diff
@@ -0,0 +1,55 @@
+# Alibaba Cloud Model Studio (DashScope)
+
+Alibaba Cloud Model Studio (DashScope) provides access to Qwen models via an OpenAI-compatible API.
```
DouweM: If we're going to have this initial paragraph (note that OpenAI, Anthropic, etc do not), we should at least move the link from the configuration section here.

DouweM: And let's link "OpenAI-compatible API" to the section that explains those on the openai docs page.

docs/models/azure.md
```diff
@@ -0,0 +1,41 @@
+# Azure AI Foundry
+
+Azure AI Foundry provides access to OpenAI models and other AI models through Microsoft Azure's cloud platform.
+
+## Install
+
+To use Azure AI Foundry, you need to either install `pydantic-ai`, or install `pydantic-ai-slim` with the `openai` optional group:
```
DouweM: The "openai" will be confusing/ look likea typo here unless we mention teh "via OpenAI compat API" thing

docs/models/deepseek.md
```diff
@@ -0,0 +1,59 @@
+# DeepSeek
+
+DeepSeek provides high-performance AI models with an OpenAI-compatible API.
```
DouweM: Drop the "high-performance" (or the entire paragraph) please, that's too marketing-y for a docs page, and having it here but not elsewhere implies that the other models are not high performance.

docs/models/litellm.md
```diff
@@ -0,0 +1,39 @@
+# LiteLLM
+
+LiteLLM provides a unified interface to call 100+ LLM APIs using the OpenAI format.
```
DouweM: consistent wording please, we don't ay "OpenAI format" on the other pages

docs/models/heroku.md
```diff
@@ -0,0 +1,39 @@
+# Heroku AI
+
+Heroku AI provides access to AI models through Heroku's cloud platform.
+
+## Install
+
+To use Heroku AI, you need to either install `pydantic-ai`, or install `pydantic-ai-slim` with the `openai` optional group:
```
DouweM: see above; we need to somehow explain why this uses the openai optional group. If we drop the initial paragraph, we could add it as a parenthetical here (`(as it uses an OpenAI-compatible API)`). I think I prefer dropping the marketing sentence and always just mentioning this here.

docs/models/litellm.md
```diff
@@ -0,0 +1,39 @@
...
+
+To use LiteLLM, you need to either install `pydantic-ai`, or install `pydantic-ai-slim` with the `openai` optional group:
+
+```bash
+pip/uv-add 'pydantic-ai-slim[openai]'
+```
+
+## Configuration
+
+To use [LiteLLM](https://www.litellm.ai/), set the configs as outlined in the [doc](https://docs.litellm.ai/docs/set_keys). In `LiteLLMProvider`, you can pass `api_base` and `api_key`. The value of these configs will depend on your setup. For example, if you are using OpenAI models, then you need to pass `https://api.openai.com/v1` as the `api_base` and your OpenAI API key as the `api_key`. If you are using a LiteLLM proxy server running on your local machine, then you need to pass `http://localhost:<port>` as the `api_base` and your LiteLLM API key (or a placeholder) as the `api_key`.
```
DouweM: "`https://api.openai.com/v1` as the `api_base`" has got to be wrong right? That would just use the OpenAI API, not LiteLLM. I think we should remove this.

docs/models/overview.md
```diff
@@ -1,44 +1,66 @@
...
+| [Mistral](mistral.md) - Mistral | | [Groq](groq.md) - Fast inference |
+| [MoonshotAI](moonshotai.md) - Kimi | | [Heroku AI](heroku.md) |
+| [OpenAI](openai.md) - GPT | | [HuggingFace](huggingface.md) |
+| [Perplexity](perplexity.md) - Sonar | | [LiteLLM](litellm.md) - Unified interface |
+| [xAI](xai.md) - Grok | | [Nebius AI Studio](nebius.md) |
+| | | [Ollama](ollama.md) - Local models |
+| | | [OpenRouter](openrouter.md) - Multi-provider gateway |
+| | | [OVHcloud](ovhcloud.md) |
+| | | [Together AI](together.md) - Fast inference |
+| | | [Vercel AI Gateway](vercel.md) |
```
DouweM: I don't love the "Both" column, I'd rather list specific providers under both lists.

docs/models/overview.md
```diff
@@ -1,44 +1,66 @@
...
-- [Grok (xAI)](openai.md#grok-xai)
-- [Heroku](openai.md#heroku-ai)
-- [LiteLLM](openai.md#litellm)
-- [Nebius AI Studio](openai.md#nebius-ai-studio)
-- [Ollama](openai.md#ollama)
-- [OVHcloud AI Endpoints](openai.md#ovhcloud-ai-endpoints)
-- [Perplexity](openai.md#perplexity)
-- [Together AI](openai.md#together-ai)
-- [Vercel AI Gateway](openai.md#vercel-ai-gateway)
+# Models
```
DouweM: I think the "Model Providers" title was more accurate

docs/models/overview.md
```diff
@@ -1,44 +1,66 @@
...
+
+Pydantic AI is model-agnostic and has built-in support for multiple model providers.
+
+## Supported Providers
+
+Pydantic AI supports three types of providers:
+
+- **Model Providers**: Companies that develop and train their own AI models
+- **Inference Providers**: Services that host and serve models from various sources
```
DouweM, line 14: Inaccurate as some are gateways that don't host or serve

DouweM: Maybe we should split up "Gateways" and "Inference Providers"

DouweM: Claude recommends: First-Party Providers, Inference Providers, and Gateways

docs/models/overview.md
```diff
@@ -1,44 +1,66 @@
...
+- **Both**: Companies that train their own models AND provide inference for other models
+
+| Model Providers | Both | Inference Providers |
+|-----------------|------|---------------------|
+| [Anthropic](anthropic.md) - Claude | [Alibaba (DashScope)](alibaba.md) - Qwen + inference | [Azure AI Foundry](azure.md) |
+| [Cohere](cohere.md) - Command R | [Bedrock](bedrock.md) - Nova + AWS inference | [Cerebras](cerebras.md) - Fast inference |
+| [DeepSeek](deepseek.md) - DeepSeek | | [Fireworks AI](fireworks.md) - Fast inference |
+| [Google](google.md) - Gemini | | [GitHub Models](github-models.md) |
```
DouweM: Isn't Google both? :)  As they have Vertex model garden

docs/models/overview.md
DouweM: Also do you feel strongly that this should be a table? I'd personally prefer separate sections with leading paragraphs and lists, especially if we end up with 3 lists as I mentioned above, with some duplication

docs/models/overview.md
```diff
@@ -1,44 +1,66 @@
...
 Pydantic AI also comes with [`TestModel`](../api/models/test.md) and [`FunctionModel`](../api/models/function.md)
 for testing and development.
 
-To use each model provider, you need to configure your local environment and make sure you have the right
+To use each provider, you need to configure your local environment and make sure you have the right
 packages installed. If you try to use the model without having done so, you'll be told what to install.
 
+## Model Comparison
+
+Quick reference for popular models. Use the `Agent String` column to configure your agent:
```
DouweM: I don't think we've used the phrase "agent string" anywhere. Is there another phrase we've used?

docs/models/overview.md
```diff
@@ -1,44 +1,66 @@
...
+| | | [Vercel AI Gateway](vercel.md) |
 
 Pydantic AI also comes with [`TestModel`](../api/models/test.md) and [`FunctionModel`](../api/models/function.md)
 for testing and development.
 
-To use each model provider, you need to configure your local environment and make sure you have the right
+To use each provider, you need to configure your local environment and make sure you have the right
 packages installed. If you try to use the model without having done so, you'll be told what to install.
 
+## Model Comparison
```
DouweM: I don't think we need this anymore if we have the dedicated page

docs/models/popular-models.md
```diff
@@ -0,0 +1,506 @@
+# Popular Models
+
+This page provides detailed information about the most popular AI models supported by Pydantic AI, including pricing, specifications, and available inference providers.
+
+For a quick overview of all supported providers, see the [Models Overview](overview.md).
```
DouweM: drop "quick", that makes it seem like it's not exhaustive, which it is

docs/models/popular-models.md
```diff
@@ -0,0 +1,506 @@
...
+
+---
```
DouweM: ?

docs/models/popular-models.md
```diff
@@ -0,0 +1,506 @@
...
+</div>
+<div markdown>
+
+**Inference Providers:**
+
+- [Gateway](https://gateway.pydantic.dev):<br>
+    `'gateway/anthropic:claude-opus-4-5'`
+- [AWS Bedrock](bedrock.md), [OpenRouter](openrouter.md)
+
+**Pricing:** $5.00 / $25.00 per 1M tokens (in/out)
```
DouweM, line 57: Would it be possible to load this automatically from the genai prices repo JSON file or something?

docs/models/popular-models.md
```diff
@@ -0,0 +1,506 @@
...
+
+---
+
+## Anthropic Claude
+
+<div class="model-card" markdown>
+
+### Claude Opus 4.5
+
+ID: `anthropic:claude-opus-4-5`
```
DouweM, line 24: We should somehow explain what this ID is, how to use it, etc.

docs/models/popular-models.md
```diff
@@ -0,0 +1,506 @@
...
+- Context: 200K tokens
+- Modalities: Text
+- Release: November 2025
+
+**Performance:**
+
+<ul class="perf-list">
+<li><span>Cost</span> <span class="rating-badge rating-5">5</span></li>
+<li><span>Speed</span> <span class="rating-badge rating-2">2</span></li>
+<li><span>Intelligence</span> <span class="rating-badge rating-5">5</span></li>
```
DouweM, line 42: Where do these scores come from?

docs/models/popular-models.md
```diff
@@ -0,0 +1,506 @@
...
+</ul>
+
+</div>
+<div markdown>
+
+**Inference Providers:**
+
+- [Gateway](https://gateway.pydantic.dev):<br>
+    `'gateway/anthropic:claude-sonnet-4-5'`
+- [AWS Bedrock](bedrock.md), [OpenRouter](openrouter.md)
```
DouweM: We should list the ID for each provider

docs/models/popular-models.md
```diff
@@ -0,0 +1,506 @@
...
+<li><span>Speed</span> <span class="rating-badge rating-4">4</span></li>
+<li><span>Intelligence</span> <span class="rating-badge rating-4">4</span></li>
+</ul>
+
+</div>
+<div markdown>
+
+**Inference Providers:**
+
+- [Gateway](https://gateway.pydantic.dev):<br>
```
DouweM, line 93: Confusing unless it's "Pydantic AI Gateway"

DouweM: Why the br?

docs/models/popular-models.md
```diff
@@ -0,0 +1,506 @@
...
+Most cost-effective with ultra-low latency for high-volume apps.
+
+<div class="model-details" markdown>
+<div markdown>
+
+**Specifications:**
+
+- Context: 1M tokens
+- Modalities: Text, Audio
+- Release: 2025
```
DouweM, line 254: We should be consistent in listing the month

docs/models/popular-models.md
```diff
@@ -0,0 +1,506 @@
...
+Google's most capable model for complex multimodal tasks.
+
+<div class="model-details" markdown>
+<div markdown>
+
+**Specifications:**
+
+- Context: 1M tokens
+- Modalities: Text, Vision
+- Release: Preview 2025
```
DouweM, line 168: Above we list the month, so this is weird

docs/models/popular-models.md
```diff
@@ -0,0 +1,506 @@
...
+
+Google's most capable model for complex multimodal tasks.
+
+<div class="model-details" markdown>
+<div markdown>
+
+**Specifications:**
+
+- Context: 1M tokens
+- Modalities: Text, Vision
```
DouweM, line 167: Unclear if this is input or output modalities.

docs/models/popular-models.md
```diff
@@ -0,0 +1,506 @@
...
+- Context: 400K tokens
+- Modalities: Text
+- Release: December 2025
+
+**Performance:**
+
+<ul class="perf-list">
+<li><span>Cost</span> <span class="rating-badge rating-3">3</span></li>
+<li><span>Speed</span> <span class="rating-badge rating-4">4</span></li>
+<li><span>Intelligence</span> <span class="rating-badge rating-4">4</span></li>
```
DouweM, line 351: Why not 5?

docs/models/popular-models.md
```diff
@@ -0,0 +1,506 @@
...
+</ul>
+
+</div>
+<div markdown>
+
+**Inference Providers:**
+
+- [Gateway](https://gateway.pydantic.dev):<br>
+    `'gateway/openai:gpt-5.2-pro'`
+- [Azure](azure.md), [OpenRouter](openrouter.md)
```
DouweM: It's unclear to me how we selected the providers to list here. E.g. why not also have Azure for Anthropic? Or more gateways here?

docs/models/popular-models.md
```diff
@@ -0,0 +1,506 @@
...
+<ul class="perf-list">
+<li><span>Cost</span> <span class="rating-badge rating-4">4</span></li>
+<li><span>Speed</span> <span class="rating-badge rating-3">3</span></li>
+<li><span>Intelligence</span> <span class="rating-badge rating-5">5</span></li>
+</ul>
+
+</div>
+<div markdown>
+
+**Inference Providers:**
```
DouweM: I think this should be e.g. "Alternative Providers"

docs/models/popular-models.md
```diff
@@ -0,0 +1,506 @@
...
+- Context: 128K tokens
+- Modalities: Text, Vision
+- Release: 2024
+
+**Performance:**
+
+<ul class="perf-list">
+<li><span>Cost</span> <span class="rating-badge rating-2">2</span></li>
+<li><span>Speed</span> <span class="rating-badge rating-4">4</span></li>
+<li><span>Intelligence</span> <span class="rating-badge rating-3">3</span></li>
```
DouweM: If this model doesn't score 5 on anything and is more than a year old, I don't think we need to list it. As discussed, I thought it was only worth mentioning 4o if it still occupies a particular niche.

docs/models/popular-models.md
```diff
@@ -0,0 +1,506 @@
...
+</div>
+</div>
+
+---
+
+## xAI Grok
+
+<div class="model-card" markdown>
+
+### Grok 4
```
DouweM, line 378: We call this Popular Models and honestly Grok rarely comes up on Slack, so I'd rather stick to the big 3 providers. If we look at usage data, I wouldn't be surprised if Qwen is higher than Grok for example. Maybe we should include Qwen? Then the primary ID can be Alibaba, and we can point out it can be run with Ollama etc

mkdocs.yml
```diff
@@ -24,19 +24,36 @@ nav:
...
+          - models/deepseek.md
+          - models/google.md
           - models/mistral.md
-          - models/openrouter.md
-          - models/outlines.md
+          - models/moonshotai.md
+          - models/openai.md
+          - models/perplexity.md
+          - models/xai.md
+          - Inference Providers:
```
DouweM, line 40: See above, I think we may need "Gateways" and "Inference Providers" separately

mkdocs.yml
```diff
@@ -24,19 +24,36 @@ nav:
           - output.md
           - message-history.md
           - direct.md
-      - Models & Providers:
+      - Models:
           - Overview: models/overview.md
-          - models/openai.md
+          - Popular Models: models/popular-models.md
```
DouweM: We shouldn't need the prefix here as this is already the name of the page

DouweM (changes requested):
I think this PR is worth splitting up into 2 PRs because it looks like the page reorg is closer to ready than the popular model catalog

DouweM (changes requested):
I think this PR is worth splitting up into 2 PRs because it looks like the page reorg is closer to ready than the popular model catalog

docs/models/popular-models.md
dsfaccini: I mean, I know x-ai gives `grok code fast` away but yeah people use it a lot. Funnily enough Qwen isn't currently in the ranking. But `MiMo` (which I never heard of) is.
image

docs/models/popular-models.md
dsfaccini: TODO remove GPT 4o and mini

docs/models/openai.md
dsfaccini: That makes sense 😅

docs/models/overview.md
dsfaccini: there will be two entities reading docs, humans and LLMs, LLMs can read whatever humans can but not the other way around, so my argument for the table format is simply that it's more compact, so the person finds what they were looking for faster and don't need to scroll as much. The paragraph and list approach leaves a lot of whitespace...

docs/models/overview.md
dsfaccini: since I rename to `First-Party Providers, Inference Providers, and Gateways` google will remain a first party provider. Bedrock is in theory a first party provider (Nova) though it's more known for being an inference provider... so I'll leave it there

docs/models/overview.md
dsfaccini: I use `shorthand`, will change it to that if no one objects

docs/models/overview.md
dsfaccini: yeah

docs/models/overview.md
dsfaccini: though no longer relevant since I removed the section, but yeah, shorthand is my preferred

docs/models/overview.md
dsfaccini: I like "Models and Provider"

docs/models/popular-models.md
dsfaccini: pre-commit hook?

docs/models/popular-models.md
dsfaccini: `shorthand` to the rescue!

dsfaccini:
```
Each model card lists the **shorthand** (ID) that can be passed to the [`Agent`](../api/agent.md) class to use that model and provider combination:
```

docs/models/popular-models.md
dsfaccini: currently no algorithm, simply ranking them using the sources

docs/models/popular-models.md
dsfaccini: because with no line break I can click three times on the `shorthand` to select it, which makes copying easier. without the line break the provider name gets selected too

I could add some js so they're copyable just by clicking and we can remove the line break

docs/models/popular-models.md
dsfaccini: gonna add a clarification since all of these are only Text output AFAIK

docs/models/popular-models.md
dsfaccini: agree

docs/models/popular-models.md
dsfaccini: because gpt-5.2-pro is 5

docs/models/popular-models.md
dsfaccini: it starts getting long when we include all the gateways, but maybe it's fair to do that and maybe make the section expandable with a fade

docs/models/alibaba.md
dsfaccini: TODO: review all newly created model pages. Remove initial paragraph (we didn't have it before)

docs/models/popular-models.md
dsfaccini: https://github.com/pydantic/pydantic-ai/pull/3941/commits/6aea2cd38dfaefdad4dae529c61af88f6d402c4f

docs/models/azure.md
```diff
@@ -0,0 +1,39 @@
...
+model = OpenAIChatModel(
+    'gpt-5',
+    provider=AzureProvider(
+        azure_endpoint='your-azure-endpoint',
+        api_version='your-api-version',
+        api_key='your-api-key',
+    ),
+)
+agent = Agent(model)
+...
```
dsfaccini, line 38: should we link here to other models that can also be run in azure to not seem openai-fixated

PR #3957, opened 2026-01-08
Add pre-commit hook to enforce isinstance union syntax

dsfaccini:
## Summary
- Adds `scripts/check_isinstance_union.py` - an AST-based checker/fixer for isinstance/issubclass calls
- Adds pre-commit hook that auto-fixes tuple-style isinstance to union syntax
- Fixes 3 existing violations in the codebase

The hook detects and fixes:
- `isinstance(x, (A, B))` → `isinstance(x, A | B)` (tuple to union)
- `isinstance(x, (A | B))` → `isinstance(x, A | B)` (removes redundant parens)
- `isinstance(x, (A))` → `isinstance(x, A)` (removes redundant parens)

## Test plan
- [x] Run `uv run python scripts/check_isinstance_union.py --check pydantic_ai_slim/ tests/` to verify no issues remain
- [x] Run `pre-commit run isinstance-union --all-files` to verify hook passes
- [x] All pre-commit hooks pass

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/format_prompt.py
+++ b/pydantic_ai_slim/pydantic_ai/format_prompt.py
@@ -167,7 +167,7 @@ class _ToXml:
         path: str = '',
     ):
         """Parse data structures as dataclasses or Pydantic models to extract element names and attributes."""
-        if value is None or isinstance(value, (str | int | float | date | time | timedelta | bytearray | bytes | bool)):
+        if value is None or isinstance(value, str | int | float | date | time | timedelta | bytearray | bytes | bool):
             return
         elif isinstance(value, Mapping):
             for k, v in value.items():  # pyright: ignore[reportUnknownVariableType]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -39,6 +39,13 @@ repos:
           - tomli
   - repo: local
     hooks:
+      - id: isinstance-union
+        name: isinstance union syntax
+        entry: uv
+        args: [run, python, scripts/check_isinstance_union.py, --fix]
+        language: system
+        types: [python]
+        pass_filenames: true
       - id: clai-help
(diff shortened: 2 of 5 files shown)


pydantic_ai_slim/pydantic_ai/format_prompt.py
```diff
@@ -167,7 +167,7 @@ def _parse_data_structures(
         path: str = '',
     ):
         """Parse data structures as dataclasses or Pydantic models to extract element names and attributes."""
-        if value is None or isinstance(value, (str | int | float | date | time | timedelta | bytearray | bytes | bool)):
+        if value is None or isinstance(value, str | int | float | date | time | timedelta | bytearray | bytes | bool):
```
dsfaccini, line 170: naturally we had no instances of `isinstance(x, (Class1, Class2))` in the codebase, which this hook is mostly aimed at, but I also decided against the `(Class1 | Class2)` syntax (removing parentheses)

Let me know what you think

Kludex:
What's the motivation for this?

dsfaccini:
This is something Douwe enforces on his reviews so I thought it would be ideal to simply enforce as a pre-commit

Kludex:
https://docs.astral.sh/ruff/rules/non-pep604-isinstance/ (this rule was removed - read first paragraph)

Please don't create linting rules besides what ruff offers.

dsfaccini:
ok

DouweM:
Good to know that this is not actually recommended anymore! I started recommending it because I had noticed that when we dropped 3.9 support, we changed all the `isinstance` using tuples to use unions instead.

> Note that this results in slower code. Ignore this rule if the performance of an isinstance or issubclass check is a concern, e.g., in a hot loop.

This is really good to know as well; should we go back to using tuples across the board?

And strong agree that this would be linter territory, not a custom check.

Issue #3702, opened 2025-12-10
Temporal: Binary tool results fail serialization


dsfaccini:
### Initial Checks

- [x] I confirm that I'm using the latest version of Pydantic AI
- [x] I confirm that I searched for my issue in https://github.com/pydantic/pydantic-ai/issues before opening this issue

### Description

When MCP tools return binary data (bytes, images, audio), Temporal activity results fail with `PydanticSerializationError: invalid utf-8 sequence`.

**Root cause:**

`_ToolReturn.result` in `_toolset.py` stores the raw tool result:

```python
@dataclass
class _ToolReturn:
    result: Any
    kind: Literal['tool_return'] = 'tool_return'
```

Temporal serializes activity results via Pydantic JSON. When `result` contains `bytes`, JSON serialization fails because bytes are not JSON-serializable.

**Affected code path:**

1. `_wrap_call_tool_result()` stores raw bytes in `_ToolReturn(result=result)`
2. Temporal serializes the activity return value using `pydantic_core.to_json()`
3. Bytes trigger: `PydanticSerializationError: invalid utf-8 sequence`

**Proposed solution:**

Base64 encode bytes in `_wrap_call_tool_result()` and decode in `_unwrap_call_tool_result()`:

```python
import base64

async def _wrap_call_tool_result(self, coro: Awaitable[Any]) -> CallToolResult:
    try:
        result = await coro
        if isinstance(result, bytes):
            result = {'__bytes__': base64.b64encode(result).decode('ascii')}
        return _ToolReturn(result=result)
    # ...

def _unwrap_call_tool_result(self, result: CallToolResult) -> Any:
    if isinstance(result, _ToolReturn):
        r = result.result
        if isinstance(r, dict) and '__bytes__' in r:
            return base64.b64decode(r['__bytes__'])
        return r
    # ...
```

### Example Code

```python
from pydantic_ai import Agent
from pydantic_ai.mcp import MCPServerStdio
from pydantic_ai.durable_exec import TemporalAgent

# MCP server that returns binary data (e.g., image generation, audio synthesis)
mcp_server = MCPServerStdio('python', ['-m', 'my_mcp_server'])
agent = Agent('openai:gpt-4o', toolsets=[mcp_server])
temporal_agent = TemporalAgent(agent, deps_type=type(None))

# In workflow:
result = await temporal_agent.run('Generate an image')
# Fails with PydanticSerializationError if tool returns bytes
```

[...]

PR #3968, opened 2026-01-09
Fix serialization of `DocumentUrl` with custom `media_type` in Temporal

### LINKED ISSUE #3949: `DocumentUrl` with custom `media_type` (not derivable from URL) doesn't work with Temporal
### Initial Checks

### Description

As reported by a user:

> The `execute_activity` method does not serialize/deserialize the DocumentUrl pydantic-ai type correctly. It was dropping the `media_type` field, which is a computed field. If I recall correctly, I found the cause of this was that the pydantic_data_converter `to_payload` method does not consider the type when the object is being serialized. It just uses `to_json` which does not respect the computed field. Fix for this would be to do the inverse of `from_payload`: use a TypeAdapter to serialize if the type information is present. 
> media_type field: https://github.com/pydantic/pydantic-ai/blob/1ad834b5b824c67e38dd19cf2ceb1b37ebe6cac0/pydantic_ai_slim/pydantic_ai/messages.py#L195-L199
> pydantic convert to_payload: https://github.com/temporalio/sdk-python/blob/f0bd341450ef4f12b2aba584d4fb8f55e2b82fe0/temporalio/contrib/pydantic.py#L66-L83
> Workaround I used for this was to append the file extension as a fragment to the end of the download url (e.g. "#.jpg") so that pydantic-ai's auto-detection of the media type would kick in.
> 

```Python

```

### Logfire Trace

_No response_

latest

dsfaccini:
- **add `PydanticAIJSONPayloadConverter` and tests**
- **refactor**

- Closes #3949 

### Pre-Review Checklist

### Pre-Merge Checklist


dsfaccini:
new test makes CI hang :(

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_converter.py
```diff
@@ -0,0 +1,108 @@
...
+    EncodingPayloadConverter,
+    JSONPlainPayloadConverter,
+)
+
+from pydantic_ai.messages import FileUrl
+
+
+@lru_cache(maxsize=128)
+def _get_type_adapter(cls: type[Any]) -> TypeAdapter[Any]:
+    return TypeAdapter(cls)
```
dsfaccini: tests were taking too long without this, but idk if you have anything against leaving this in

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_converter.py
```diff
@@ -0,0 +1,108 @@
...
+        because it causes hangs in Temporal workflows.
+
+        For example, `CallToolParams` has `tool_args: dict[str, Any]` but at runtime contains
+        a Pydantic model. `TypeAdapter.dump_json()` tries to serialize through the dict[str, Any] schema,
+        which hangs in Temporal's threading context (works fine in isolation, hangs in workflows).
+
+        So we use `TypeAdapter` only for `FileUrl` (which needs it), and to_json() for everything else.
+        """
+        if isinstance(value, FileUrl):
+            data = _get_type_adapter(type(value)).dump_json(value)
```
DouweM: Does it have to be a type-specific type adapter, or would `TypeAdapter[Any]` work as well?

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_converter.py
```diff
@@ -0,0 +1,108 @@
...
+def make_data_converter(converter: DataConverter | None) -> DataConverter:
+    if converter is None:
+        return pydantic_ai_data_converter
+
+    # If the payload converter class is already a subclass of PydanticPayloadConverter,
+    # the converter is already compatible with Pydantic AI - return it as-is.
+    if issubclass(converter.payload_converter_class, PydanticAIPayloadConverter):
+        return converter
+
+    # Preserve custom PydanticPayloadConverter subclasses
```
DouweM: So in this case the bug would still exist silently?

Is there really no way we can make things work with the existing `PydanticPayloadConverter`? :/

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_converter.py
DouweM: Is there any fix we could upstream to Temporal?

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_converter.py
DouweM, line 1: Can you see if we can fix https://github.com/pydantic/pydantic-ai/issues/3702 here as well? Since it would likely also require changes to how we do data conversion, and I want to make sure we land on a workaround that works for both, not one now, and then find out we'll need another then.

tests/test_temporal.py
```diff
@@ -2374,6 +2377,62 @@ async def test_image_agent(allow_model_requests: None, client: Client):
...
+# ============================================================================
+# DocumentUrl Serialization Test - Verifies that DocumentUrl with custom
+# media_type is properly serialized through Temporal activities
+# ============================================================================
+
+
+def document_url_response(messages: list[ModelMessage], info: AgentInfo) -> ModelResponse:
+    """Return a DocumentUrl with a custom media_type that cannot be inferred from the URL."""
+    assert info.output_tools is not None
+    document_url_json = '{"url": "https://example.com/doc/12345", "media_type": "application/pdf"}'
```
DouweM: Why as stringified JSON instead of the object?

tests/test_temporal.py
```diff
@@ -2374,6 +2377,62 @@ async def test_image_agent(allow_model_requests: None, client: Client):
...
+    return ModelResponse(parts=[ToolCallPart(info.output_tools[0].name, document_url_json)])
```
DouweM: If we pass the `DocumentUrl` into `agent.run` instead of having a tool return it, it'll be easier to test this with `TestModel`.

tests/test_temporal.py
```diff
@@ -3024,26 +3083,55 @@ async def decode(
         return list(payloads)
 
 
-def test_pydantic_ai_plugin_no_converter_returns_pydantic_data_converter() -> None:
-    """When no converter is provided, PydanticAIPlugin uses the standard pydantic_data_converter."""
+def test_pydantic_ai_json_payload_converter_preserves_document_url_media_type() -> None:
+    """Test that `PydanticAIJSONPayloadConverter` preserves `DocumentUrl.media_type` during round-trip.
```
DouweM: I'd rather just have the higher level test, so we can potentially swap out the implementation while continuing to verify that using a DocumentUrl (with media type) works as expected.

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_converter.py
dsfaccini: I'm gonna push an update and then check if we could more easily do this in temporal (I agree that it would probably make more sense there)

drewhoskins-temporal: Let us know if you think it's better as a temporal fix and we can discuss.

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_converter.py
dsfaccini: `TypeAdapter[Any]` doesn't work because it falls back to generic serialization that ignores the model's schema. 
 
 ```python
@dataclass(init=False, repr=False)
class FileUrl(ABC):
    ...
    # specifically this
    _media_type: Annotated[str | None, pydantic.Field(alias='media_type', default=None, exclude=True)] = field(
        compare=False, default=None
    )
```
 
it outputs `{"_media_type":"application/json",...}` instead of `{"media_type":"application/json",...}`which we then can't deserialize

from my answer in slack

```
hey guys! haven't looked into it, let me get back to you on that. I'm a bit concerned that upstreaming the fix as it currently is in the PR would break validation for non pydantic ai users
[12:55](https://pydantic.slack.com/archives/C08JWSCRM88/p1768586115570459?thread_ts=1768583382.858809&cid=C08JWSCRM88)
as it stands now I'd prefer having a fix that we can merge into pydantic ai, upon which we can discuss upstreaming
```

this is where I'm thinking that if we were to change the behavior upstream it wouldn't be backwards compatible

File: pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_converter.py (hunk 3 above)
Comment (dsfaccini) [reply pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_converter.py]: yeah for users that have their own custom subclasses. we could add a warning (me and warnings, I know...)

and again, if we change the behavior of the `PydanticPayloadConverter` class itself it may change behavior for users, so any fix would need to include a flag to opt-in/out of it, but the logic starts getting messy in my mind and I don't feel confident enough with the temporal codebase yet to change that much stuff

File: tests/test_temporal.py (hunk 4 above)
Comment (dsfaccini) [reply tests/test_temporal.py]: I mean, `ToolCallPart` accepts both, and an LLM would return JSON, right? But it doesn't matter, I can change it.

Comment (DouweM) [reply tests/test_temporal.py]: But `DocumentUrl` never comes from the model, always from the user?

Comment (DouweM) [reply tests/test_temporal.py]: It being JSON here just makes it seem like there's a specific reason to do it like that

File: pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_converter.py (hunk 2 above)
Comment (DouweM) [reply pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_converter.py]: I wonder if using Pydantic dataclasses instead of plain dataclasses, would help

File: tests/test_temporal.py (hunk 4 above)
Comment (dsfaccini) [reply tests/test_temporal.py]: moved this to the `TestModel` implementation where it uses a dict now

File: pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_converter.py [hunk 7]
```diff
@@ -0,0 +1,108 @@
...
+        We can't use `TypeAdapter` for all BaseModel/dataclass types
+        because it causes hangs in Temporal workflows.
+
+        For example, `CallToolParams` has `tool_args: dict[str, Any]` but at runtime contains
+        a Pydantic model. `TypeAdapter.dump_json()` tries to serialize through the dict[str, Any] schema,
+        which hangs in Temporal's threading context (works fine in isolation, hangs in workflows).
+
+        So we use `TypeAdapter` only for `FileUrl` (which needs it), and to_json() for everything else.
+        """
+        if isinstance(value, FileUrl):
```
DouweM: By the way, this won't solve the case where a tool returns a list of `FileUrl`. But I don't think we should special case the list either. I'd love to find a way to make this use Pydantic's recursive approach...

DouweM (changes requested):
@dsfaccini I had one more thought that I added below. But I suggest adding `BinaryContent` support here as well and then we can see if we can clean up the solution.

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_converter.py
dsfaccini: I have a separate branch for the Binary serialization already, the issues seemed separate enough to do it that way.

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_converter.py
```diff
@@ -0,0 +1,122 @@
...
+@lru_cache(maxsize=128)
+def _get_type_adapter(cls: type[Any]) -> TypeAdapter[Any]:
+    return TypeAdapter(cls)
+
+
+def _prepare_for_serialization(value: Any) -> Any:
+    """Recursively convert `FileUrl` instances to dicts with computed fields preserved.
+
+    `pydantic_core.to_json()` doesn't use Pydantic schema information, so it doesn't
+    preserve computed fields or respect field aliases/exclusions on `FileUrl`.
```
DouweM: I don't believe "doesn't respect aliases" is exactly true, as https://docs.pydantic.dev/latest/api/pydantic_core/#pydantic_core.to_json shows it has a `by_alias` argument (that defaults to `True`).

The `serialize_as_any` argument also looks interesting, maybe that would make it respect Pydantic schema information? Then it may be easier to use our own subclass of the pydantic data converter that just sets that argument, and we can consider upstreaming that (as on by default or at least an option).

But honestly I don't see myself accepting this hard-coded support for `FileUrl` 😅

DouweM (changes requested):

PR #3971, opened 2026-01-09
Full support for AG-UI 0.1.13 and 0.1.15: reasoning, multi-modal, `dump_messages`

### LINKED ISSUE #3911: AGUI with Anthropic extended thinking models raises error
### Initial Checks

### Description

- Same as https://github.com/pydantic/pydantic-ai/issues/3748, but for `AGUIAdapter`

Came up [in Slack](https://pydanticlogfire.slack.com/archives/C083V7PMHHA/p1767039681535429), where I responded:

The underlying issue is the same as in https://github.com/pydantic/pydantic-ai/issues/3748, which was reported for the VercelAI integration: the round trip from Pydantic AI messages -> AG UI events -> AG UI messages -> Pydantic AI messages is lossy in terms of thinking signatures and other metadata, which Anthropic and Google now require to be included. A fix for Vercel is in the works at https://github.com/pydantic/pydantic-ai/pull/3754, which fixes this by storing Pydantic AI thinking metadata on the arbitrary provider metadata field that each Vercel AI event has. AG UI would need to do effectively the same thing, but doesn't have a arbitray metadata field, as far as I'm aware. The draft "reasoning events" at https://docs.ag-ui.com/concepts/events#reasoning-events would help, but don't cover the case where with Gemini, the thought_signature should be attached to the tool call or text part, rather than to a reasoning part.

```Python

```

### Logfire Trace

_No response_

main

### LINKED ISSUE #4161: Transforming Pydantic AI messages to AG-UI messges
### Description

The `AGUIAdapter` class provides a `load_messages` method to transform AG-UI messages into the Pydantic AI message representation. This is used, for instance, for parsing them from the run input.

However, there doesn't seem to be a way to go from Pydantic AI messages to the AG-UI format. This would be useful for cases like when messages have been persisted and should be sent to the client as a message snapshot instead of single events.

### References

_No response_

### LINKED ISSUE #4868: AGUI Thinking Events are deprecated
### Initial Checks

### Description

AG-UI has deprecated `THINKING_*` events in favor of `REASONING_*` events:
https://docs.ag-ui.com/concepts/events#deprecated-events

This might be tangentially related to #3911 

```Python

```

### Logfire Trace

_No response_

- Python: 3.14.0
- Pydantic AI: https://github.com/pydantic/pydantic-ai/blob/main/pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py#L47-L49

dsfaccini:
- Closes #3911
- Closes #4161 
- Fixes #4868
- Fixes #4871

### Pre-Review Checklist

### Pre-Merge Checklist

- [x] Updated **documentation**: don't think we need to

---

## Summary

AG-UI equivalent of #3754 (Vercel AI fix). Preserves thinking signatures, agent-generated files, uploaded files, and builtin tool calls/returns across the AG-UI round trip.

## Problem

Multi-turn conversations with Anthropic extended thinking models fail via AG-UI:
```
"messages.1.content.0.type: Expected `thinking` or `redacted_thinking`, but found `text`.
When `thinking` is enabled, a final `assistant` message must start with a thinking block..."
```

Root cause: the round trip Pydantic AI messages -> AG-UI events -> AG-UI messages -> Pydantic AI messages is lossy for thinking signatures required by Anthropic's API.

### Semver-based version gating

`ag_ui_version: str` accepts any version string (e.g. `'0.1.13'`). Defaults to the installed `ag-ui-protocol` version. Version comparison uses `parse_ag_ui_version()` which parses into a comparable tuple.

Known thresholds:

- **< 0.1.13**: `THINKING_*` events during streaming, `ThinkingPart` dropped from `dump_messages`
- **>= 0.1.13**: `REASONING_*` events with encrypted metadata, `ThinkingPart` preserved as `ReasoningMessage` in `dump_messages` for full round-trip fidelity
- **>= 0.1.15**: typed multimodal input content (`ImageInputContent`, `AudioInputContent`, etc.) instead of deprecated `BinaryInputContent`

`load_messages` always accepts `ReasoningMessage` and multimodal content types regardless of version (forward compatible).

### `preserve_file_data` opt-in

When `True`:
- `FilePart` round-trips via `ActivityMessage(activity_type='pydantic_ai_file')`
- `UploadedFile` round-trips via `ActivityMessage(activity_type='pydantic_ai_uploaded_file')`
- `BuiltinToolCallPart`/`BuiltinToolReturnPart` survive via prefixed tool call IDs

## Key files

- `_utils.py` — shared constants (`REASONING_VERSION`, `MULTIMODAL_VERSION`, activity type constants), `parse_ag_ui_version`, `detect_ag_ui_version`, `thinking_encrypted_metadata`, TypedDicts for ActivityMessage content schemas
- `_event_stream.py` — `AGUIEventStream` with version-branched event emission
- `_adapter.py` — `AGUIAdapter` with `dump_messages`/`load_messages`, version gating, multimodal support, `preserve_file_data`
- `_multimodal.py` — lazy-imported multimodal type conversion for ag-ui >= 0.1.15
- `_thinking_0_10.py` / `_thinking_0_13.py` — extracted thinking event handlers per version
- `ag_ui.py` — `handle_ag_ui_request`/`run_ag_ui` with `ag_ui_version` and `preserve_file_data` params

## Related

- Issue: #3911
- Vercel AI fix: #3754


DouweM:
@dsfaccini FYI I'd wait to do much more here until https://github.com/pydantic/pydantic-ai/pull/3754 lands

dsfaccini:
noted, there's honestly not much more to do here, I rewrote the logic a bit more cleanly as well as the tests, but this is now good to go from my side.

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
```diff
@@ -228,7 +231,21 @@ def load_messages(cls, messages: Sequence[Message]) -> list[ModelMessage]:  # no
                             )
                         )
 
-                case ActivityMessage():
-                    pass
+                case ActivityMessage() as activity_msg:
+                    if activity_msg.activity_type != 'pydantic_ai_thinking':
```
DouweM: I'm confused, where would these messages originate? Doesn't this imply we should be creating activity messages/events somewhere?

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
```diff
@@ -171,7 +171,16 @@ async def handle_thinking_end(
...
-            yield ThinkingEndEvent(type=EventType.THINKING_END)
+            _args = ['id', 'signature', 'provider_name', 'provider_details']
+            pydantic_ai_meta: dict[str, Any] = {
+                arg: getattr(part, arg) for arg in _args if getattr(part, arg) is not None
+            }
+
+            yield ThinkingEndEvent(
+                type=EventType.THINKING_END,
+                raw_event={'pydantic_ai': pydantic_ai_meta} if pydantic_ai_meta else None,
+                encryptedContent=part.signature,  # pyright: ignore[reportCallIssue]
```
DouweM: Are we allowed to add custom fields?

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
```diff
@@ -171,7 +171,16 @@ async def handle_thinking_end(
...
         if not followed_by_thinking:
-            yield ThinkingEndEvent(type=EventType.THINKING_END)
+            _args = ['id', 'signature', 'provider_name', 'provider_details']
+            pydantic_ai_meta: dict[str, Any] = {
+                arg: getattr(part, arg) for arg in _args if getattr(part, arg) is not None
+            }
+
+            yield ThinkingEndEvent(
+                type=EventType.THINKING_END,
+                raw_event={'pydantic_ai': pydantic_ai_meta} if pydantic_ai_meta else None,
```
DouweM: If we can add custom fields, you should look at https://github.com/pydantic/pydantic-ai/pull/3754 and do this for _every single `provider_details`_ as e.g. Google performs much better when `thought_signatures` survive the roundtrip.

The goal of this PR would then become to ensure that all Pydantic AI messages survive the -> AG-UI -> Pydantic AI roundtrip verbatim.

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
dsfaccini: yeahp, thank you for pointing that out, I needed to emit `ActivitySnapshotEvent` for this

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
dsfaccini: yeahp, I've added explanations on the new code blocks https://docs.ag-ui.com/concepts/events#activity-events

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
dsfaccini: oof let me push the new version first and if you're happy with it I'll start looking into integrating it with the rest of the `provider_details`

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
```diff
@@ -173,6 +175,23 @@ async def handle_thinking_end(
         if not followed_by_thinking:
             yield ThinkingEndEvent(type=EventType.THINKING_END)
 
+            # Emit ActivitySnapshotEvent to preserve thinking metadata for round-trip.
+            # Frontends receive this and send it back as ActivityMessage, which _adapter.py
+            # converts back to ThinkingPart. This preserves signature/id needed by providers
+            # like Anthropic for extended thinking.
+            # See: https://docs.ag-ui.com/concepts/events#activity-events
```
DouweM: I'm not sure if activity messages/events are the most appropriate one to use for this purpose, as app devs are encouraged to use them themselves, and would need to account for our `type`.

Maybe https://docs.ag-ui.com/concepts/events#raw is better? If you read the explanation for custom events that's follows it, it's made clear that custom events are for app devs to use, while raw events are not.

Although it looks like those don't have a message type so we can't trust that they'll be sent back to us :(

DouweM: @maxkorp Any suggestions here? :)

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
```diff
@@ -228,7 +231,23 @@ def load_messages(cls, messages: Sequence[Message]) -> list[ModelMessage]:  # no
                             )
                         )
 
-                case ActivityMessage():
-                    pass
+                case ActivityMessage() as activity_msg:
```
DouweM, line 425: We also have to update `dump_messages` right?

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
```diff
@@ -173,6 +175,23 @@ async def handle_thinking_end(
...
+            content: dict[str, Any] = {'content': part.content}
+            for field in ('id', 'signature', 'provider_name', 'provider_details'):
+                value = getattr(part, field)
+                if value is not None:
+                    content[field] = value
+
+            yield ActivitySnapshotEvent(
+                activity_type='pydantic_ai_thinking',
```
DouweM: We'll want to make this more generic for attaching any type of Pydantic Ai-specific fields to AG-UI events/messages, as in the Vercel PR.

tests/test_ag_ui.py
```diff
@@ -1109,6 +1128,153 @@ async def stream_function(
...
+                        signature='sig_abc123',
+                        provider_name='anthropic',
+                        provider_details={'some': 'details'},
+                    ),
+                    TextPart(content='Here is my response'),
+                ],
+                timestamp=IsDatetime(),
+            )
+        ]
+    )
```
DouweM, line 1341: The ultimate goal of this PR is to ensure that the Pydantic AI -> AG-UI -> Pydantic AI roundtrip is lossless. So we need to `assert load_messages(dump_messages(messages)) == messages`.

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
dsfaccini: we'll need to discuss this one, there's now two of these `activity_type`s

dsfaccini:
@DouweM did we get feedback on this one?

DouweM:
@dsfaccini They'll think about supporting arbitrary metadata on parts/events, but it'll be a while before we get it. They have a reasoning type that's further along in development, which we could use for reasoning at least, assuming that's already supported by their frontend libraries (worth checking). So we may need to use the reasoning part first, and then later expand to cover all provider metadata.

Or if we can document it clearer, a reserved activity type may not be the worst soliution.

pydantic_ai_slim/pydantic_ai/ui/_event_stream.py
```diff
@@ -360,7 +360,7 @@ async def handle_part_end(self, event: PartEndEvent) -> AsyncIterator[EventT]:
                 async for e in self.handle_text_end(part, followed_by_text=next_part_kind == 'text'):
                     yield e
             case ThinkingPart():
-                async for e in self.handle_thinking_end(part, followed_by_thinking=next_part_kind == 'thinking'):
```
DouweM, line 369: @dsfaccini Why are these arguments being removed at all? They're part of the public API

dsfaccini:
Re: [removal of `follows_thinking`/`followed_by_thinking`](https://github.com/pydantic/pydantic-ai/pull/3971#discussion_r2885588032) (also raised by [Devin](https://github.com/pydantic/pydantic-ai/pull/3971#discussion_r2879085067)):

The parameters were removed because each `ThinkingPart` now carries its own metadata (`id`, `signature`, `provider_details`) that must be emitted individually — one `REASONING_START`/`REASONING_ENCRYPTED_VALUE`/`REASONING_END` block per part. Merging consecutive thinking parts into a single reasoning block (which is what `follows_thinking=True` enabled) would lose that per-part metadata, breaking the round-trip with Anthropic's extended thinking API.

The only existing implementation — [Vercel's `handle_thinking_start`](https://github.com/pydantic/pydantic-ai/blob/165596f61562759043dfbfa51bfb082ff8c70dda/pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_event_stream.py#L170-L183) and [`handle_thinking_end`](https://github.com/pydantic/pydantic-ai/blob/165596f61562759043dfbfa51bfb082ff8c70dda/pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_event_stream.py#L195-L205) — already accepted but **never read** the parameter value. Both always emit a new `ReasoningStartChunk`/`ReasoningEndChunk` regardless of whether `follows_thinking` is `True` or `False`.

That said, the params have been in the [base class signature](https://github.com/pydantic/pydantic-ai/blob/165596f61562759043dfbfa51bfb082ff8c70dda/pydantic_ai_slim/pydantic_ai/ui/_event_stream.py#L463-L490) since v1.10.0, so an external subclass _could_ be using them. If that's a concern, happy to keep them in the base class with defaults and just stop passing them from the [dispatch sites](https://github.com/pydantic/pydantic-ai/blob/165596f61562759043dfbfa51bfb082ff8c70dda/pydantic_ai_slim/pydantic_ai/ui/_event_stream.py#L298) — that way existing overrides won't break, even though the value will always be `False`. Let me know which approach you'd prefer.

pydantic_ai_slim/pydantic_ai/ui/_event_stream.py
```diff
@@ -295,7 +295,7 @@ async def handle_part_start(self, event: PartStartEvent) -> AsyncIterator[EventT
                 async for e in self.handle_text_start(part, follows_text=previous_part_kind == 'text'):
                     yield e
             case ThinkingPart():
-                async for e in self.handle_thinking_start(part, follows_thinking=previous_part_kind == 'thinking'):
+                async for e in self.handle_thinking_start(part):
```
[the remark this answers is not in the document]
dsfaccini: see https://github.com/pydantic/pydantic-ai/pull/3971#issuecomment-4000380791

pydantic_ai_slim/pydantic_ai/ui/_event_stream.py
dsfaccini: see https://github.com/pydantic/pydantic-ai/pull/3971#issuecomment-4000380791

DouweM:
> That said, the params have been in the [base class signature](https://github.com/pydantic/pydantic-ai/blob/165596f61562759043dfbfa51bfb082ff8c70dda/pydantic_ai_slim/pydantic_ai/ui/_event_stream.py#L463-L490) since v1.10.0, so an external subclass _could_ be using them. If that's a concern, happy to keep them in the base class with defaults and just stop passing them from the [dispatch sites](https://github.com/pydantic/pydantic-ai/blob/165596f61562759043dfbfa51bfb082ff8c70dda/pydantic_ai_slim/pydantic_ai/ui/_event_stream.py#L298) — that way existing overrides won't break, even though the value will always be `False`. Let me know which approach you'd prefer.

I suggest keeping the `UIEventStream` behavior as-is, including passing the correct boolean from the dispatch site, and only stopping to use the params in `AGUIEventStream`.

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
```diff
@@ -228,7 +267,210 @@ def load_messages(cls, messages: Sequence[Message]) -> list[ModelMessage]:  # no
...
+                for attr in ['id', 'provider_name', 'provider_details']:
+                    if (value := getattr(part, attr)) is not None:
+                        file_content[attr] = value
+                result.append(
+                    ActivityMessage(
+                        id=_new_message_id(),
+                        activity_type='pydantic_ai_file',
+                        content=file_content,
+                    )
+                )
```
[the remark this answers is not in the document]
DouweM: I'm OK with it, as it's the best AG-UI lets us do right now.

But I think it should be behind an opt-in flag so we don't accidentally mess with people's own activity types.

tests/test_ag_ui.py
```diff
@@ -1110,6 +1149,725 @@ async def stream_function(
...
+            {'type': 'REASONING_MESSAGE_END', 'message_id': IsStr()},
+            {
+                'type': 'REASONING_ENCRYPTED_VALUE',
+                'subtype': 'message',
+                'entity_id': IsStr(),
+                'encrypted_value': '{"id": "thinking-123", "signature": "sig_xyz", "provider_name": "anthropic", "provider_details": {"model": "claude-sonnet-4-5"}}',
+            },
+            {'type': 'REASONING_END', 'message_id': IsStr()},
+        ]
+    )
```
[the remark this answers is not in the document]
dsfaccini: Claude here: Keeping these tests. `FunctionModel.DeltaThinkingPart` only supports `content`/`signature` — converting these to integration tests would require extending FunctionModel to also support `id`/`provider_name`/`provider_details`, which is disproportionate for 2 small edge-case tests. They verify exact JSON structure of `encrypted_value` and the redacted thinking path, which the integration tests only check loosely with `IsStr()`.

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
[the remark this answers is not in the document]
dsfaccini: Claude here: Implemented the opt-in flag. Added `include_file_parts: bool = False` field to `AGUIAdapter` (KW_ONLY, following the `VercelAIAdapter.sdk_version` pattern). When `False` (default), `FilePart` is silently dropped from `dump_messages` output and `ActivityMessage(pydantic_ai_file)` is ignored by `load_messages`. The flag threads through `from_request()`, `dump_messages()`, `load_messages()`, and `_dump_response_parts()`. Tests updated accordingly.

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
```diff
@@ -228,7 +292,225 @@ def load_messages(cls, messages: Sequence[Message]) -> list[ModelMessage]:  # no
...
 
-                case ActivityMessage():
-                    pass
+                case ReasoningMessage() as reasoning_msg:
+                    try:
+                        metadata: dict[str, Any] = (
+                            json.loads(reasoning_msg.encrypted_value) if reasoning_msg.encrypted_value else {}
+                        )
+                    except json.JSONDecodeError:
+                        metadata = {}
```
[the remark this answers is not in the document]
dsfaccini: Addressed in 762775e5 — added `isinstance(metadata, dict)` guard plus parametrized tests covering invalid JSON, non-dict string, list, and number values.

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
```diff
@@ -235,7 +298,225 @@ def load_messages(cls, messages: Sequence[Message]) -> list[ModelMessage]:  # no
...
+                                'ActivityMessage with activity_type=pydantic_ai_file must have a non-empty url.'
+                            )
+                        builder.add(
+                            FilePart(
+                                content=BinaryContent.from_data_uri(url),
+                                id=activity_content.get('id'),
+                                provider_name=activity_content.get('provider_name'),
+                                provider_details=activity_content.get('provider_details'),
+                            )
+                        )
```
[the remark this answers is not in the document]
dsfaccini: Intentionally not adding validation here — unlike `encrypted_value` which round-trips through an external frontend, `pydantic_ai_file` ActivityMessages are only produced by our own `dump_messages`. A non-dict content would indicate a fundamentally broken frontend, not a plausible edge case.

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
```diff
@@ -69,11 +70,27 @@
 BUILTIN_TOOL_CALL_ID_PREFIX: Final[str] = 'pyd_ai_builtin'
 
 
+def thinking_encrypted_metadata(part: ThinkingPart) -> dict[str, Any]:
```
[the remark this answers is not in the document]
dsfaccini: Good catch — renamed to `_thinking_encrypted_metadata` in the next commit. Not extracting to a separate `_utils.py` though, as it's only used in two places within the same package and doesn't warrant a new module.

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
```diff
@@ -235,7 +298,227 @@ def load_messages(cls, messages: Sequence[Message]) -> list[ModelMessage]:  # no
...
+                            id=metadata.get('id'),
+                            signature=metadata.get('signature'),
+                            provider_name=metadata.get('provider_name'),
+                            provider_details=metadata.get('provider_details'),
+                        )
+                    )
+
+                case ActivityMessage() as activity_msg:
+                    if activity_msg.activity_type == 'pydantic_ai_file' and include_file_parts:
+                        activity_content = activity_msg.content
```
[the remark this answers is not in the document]
dsfaccini: Same reasoning as https://github.com/pydantic/pydantic-ai/pull/3971#discussion_r2944392977 — `pydantic_ai_file` ActivityMessages are only produced by our own `dump_messages`, so we control the format. Not adding validation here.

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
```diff
@@ -235,7 +298,227 @@ def load_messages(cls, messages: Sequence[Message]) -> list[ModelMessage]:  # no
...
+                        )
+                    )
+            else:
+                assert_never(part)
+
+        flush()
+        return result
+
+    @classmethod
+    def dump_messages(cls, messages: Sequence[ModelMessage], *, include_file_parts: bool = False) -> list[Message]:
```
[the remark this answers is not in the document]
dsfaccini: Acknowledged — matches the Vercel AI precedent (`generate_message_id`). This is intentional.

tests/test_ag_ui.py
```diff
@@ -1110,6 +1151,800 @@ async def stream_function(
...
```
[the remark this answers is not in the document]
dsfaccini: `DeltaThinkingPart` only supports `content` and `signature`, so we can't test `id`/`provider_name`/`provider_details` metadata through FunctionModel integration tests. The VCR-backed `test_thinking_roundtrip_anthropic` validates the real end-to-end path. These unit tests cover the specific event stream edge cases (empty content + metadata, all metadata fields) that the integration path can't reach.

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
```diff
@@ -103,10 +154,22 @@ def build_event_stream(self) -> UIEventStream[RunAgentInput, BaseEvent, AgentDep
...
+    @classmethod
+    async def from_request(
+        cls,
+        request: Request,
+        *,
+        agent: AbstractAgent[AgentDepsT, OutputDataT],
+        include_file_parts: bool = False,
+        **kwargs: Any,
+    ) -> AGUIAdapter[AgentDepsT, OutputDataT]:
+        """Extends [`from_request`][pydantic_ai.ui.UIAdapter.from_request] with the `include_file_parts` parameter."""
```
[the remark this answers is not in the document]
dsfaccini: The `include_file_parts` field has a detailed docstring on the dataclass field itself. Adding more to `from_request` would duplicate it.

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
```diff
@@ -235,7 +298,227 @@ def load_messages(cls, messages: Sequence[Message]) -> list[ModelMessage]:  # no
...
+                            tool_call_id=prefixed_id,
+                        )
+                    )
+            elif isinstance(part, BuiltinToolReturnPart):
+                # Emitted when matching BuiltinToolCallPart is processed above.
+                pass
+            elif isinstance(part, FilePart):
+                if include_file_parts:
+                    flush()
+                    file_content: dict[str, Any] = {
```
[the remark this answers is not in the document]
dsfaccini: The two patterns aren't identical — `_thinking_encrypted_metadata` builds from `ThinkingPart` fields, while the file content dict includes `url` and `media_type` which don't exist on ThinkingPart. Extracting a shared helper would need awkward parameterization for a two-use pattern.

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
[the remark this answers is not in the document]
dsfaccini: Update: reverted the rename — pyright raises `reportPrivateUsage` since the function is imported from `_event_stream.py` into `_adapter.py`. Both modules are already private, so the unprefixed name is appropriate.

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
```diff
@@ -91,9 +107,44 @@ def label(self) -> str:
...
+    """Convert a user content item to AG-UI input content."""
+    if isinstance(item, str):
+        return TextInputContent(type='text', text=item)
+    elif isinstance(item, (ImageUrl, VideoUrl, AudioUrl, DocumentUrl)):
+        return BinaryInputContent(type='binary', url=item.url, mime_type=item.media_type or '')
+    elif isinstance(item, BinaryContent):
+        return BinaryInputContent(type='binary', data=item.base64, mime_type=item.media_type)
+    elif isinstance(item, UploadedFile):
+        # UploadedFile holds an opaque provider file_id (e.g. 'file-abc123'), not a URL or
+        # binary data, so it can't be mapped to AG-UI's BinaryInputContent. Skipped like CachePoint.
```
DouweM: That seems lossy :) Can we store them in an activity or smth?

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
```diff
@@ -91,9 +107,44 @@ def label(self) -> str:
...
+@dataclass
 class AGUIAdapter(UIAdapter[RunAgentInput, Message, BaseEvent, AgentDepsT, OutputDataT]):
     """UI adapter for the Agent-User Interaction (AG-UI) protocol."""
 
+    _: KW_ONLY
+    include_file_parts: bool = False
+    """Whether to include ``FilePart`` data in message conversion.
+
+    When ``True``, ``FilePart`` round-trips as ``ActivityMessage(activity_type='pydantic_ai_file')``.
+    When ``False`` (default), ``FilePart`` is silently dropped from ``dump_messages`` output
```
DouweM: No double backticks please!

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
```diff
@@ -91,9 +107,44 @@ def label(self) -> str:
...
+        assert_never(item)
+
+
+@dataclass
 class AGUIAdapter(UIAdapter[RunAgentInput, Message, BaseEvent, AgentDepsT, OutputDataT]):
     """UI adapter for the Agent-User Interaction (AG-UI) protocol."""
 
+    _: KW_ONLY
+    include_file_parts: bool = False
+    """Whether to include ``FilePart`` data in message conversion.
```
DouweM: This isn't super clear as a user may think it applies to files going from the user to LLM as well, but I believe this only applies to files generated by the agent.

So this is a case where the field name and docstring should be more user focused, instead of referring to the `FilePart` implementation detail

DouweM: We should also link to the activities doc, make it clear that that if they use activities themselves they should not handle this specific type.

DouweM: And "round trips as" below is not very clear to the end user.

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
```diff
@@ -146,33 +163,59 @@ async def handle_text_end(self, part: TextPart, followed_by_text: bool = False)
...
-        if not follows_thinking:
-            yield ThinkingStartEvent(type=EventType.THINKING_START)
+        self._reasoning_message_id = str(uuid4())
+        self._reasoning_started = False
 
         if part.content:
-            yield ThinkingTextMessageStartEvent(type=EventType.THINKING_TEXT_MESSAGE_START)
-            yield ThinkingTextMessageContentEvent(type=EventType.THINKING_TEXT_MESSAGE_CONTENT, delta=part.content)
-            self._thinking_text = True
+            yield ReasoningStartEvent(message_id=self._reasoning_message_id)
```
DouweM: As discussed on Slack, we need to make sure this is backward compatible with old frontends

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
```diff
@@ -235,7 +298,227 @@ def load_messages(cls, messages: Sequence[Message]) -> list[ModelMessage]:  # no
...
+                assert_never(part)
+
+        flush()
+        return result
+
+    @classmethod
+    def dump_messages(cls, messages: Sequence[ModelMessage], *, include_file_parts: bool = False) -> list[Message]:
+        """Transform Pydantic AI messages into AG-UI messages.
+
+        Note: The round-trip ``dump_messages`` -> ``load_messages`` is not fully lossless:
```
DouweM: Double back ticks!

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
dsfaccini: claude: no double backticks!

dsfaccini:
Claude here: Addressing all open review threads in this consolidated reply.

### Backward compatibility — THINKING_* vs REASONING_* events

[comment](https://github.com/pydantic/pydantic-ai/pull/3971#discussion_r2949615064), [comment](https://github.com/pydantic/pydantic-ai/pull/3971#discussion_r2954609197)

Added `ag_ui_version: Literal['0.1.10', '0.1.13'] = '0.1.10'` threaded through `AGUIAdapter` → `build_event_stream()` → `AGUIEventStream` (and `run_ag_ui()`).

- `'0.1.10'` (default): emits `THINKING_*` events, drops `ThinkingPart` from `dump_messages` — no breaking change for existing frontends.
- `'0.1.13'`: emits `REASONING_*` events with `ReasoningEncryptedValueEvent` for metadata, and `ThinkingPart` → `ReasoningMessage` in `dump_messages`.

`load_messages` always accepts `ReasoningMessage` regardless of version.

### Thinking metadata round-trip via ReasoningMessage

[comment](https://github.com/pydantic/pydantic-ai/pull/3971#discussion_r2699549723), [comment](https://github.com/pydantic/pydantic-ai/pull/3971#discussion_r2699552186), [comment](https://github.com/pydantic/pydantic-ai/pull/3971#discussion_r2700379136), [comment](https://github.com/pydantic/pydantic-ai/pull/3971#discussion_r2700383925), [comment](https://github.com/pydantic/pydantic-ai/pull/3971#discussion_r2700384521)

For thinking/reasoning, we now use AG-UI's first-class `ReasoningMessage.encrypted_value` (0.1.13) instead of activity messages. `thinking_encrypted_metadata()` collects all non-None fields (`id`, `signature`, `provider_name`, `provider_details`) into the JSON, surviving the full round-trip. `dump_messages` emits `ReasoningMessage` when `ag_ui_version='0.1.13'`.

Activity messages are still used for `FilePart` and `UploadedFile` (gated behind `preserve_file_data`), with `pydantic_ai_*` prefixed types.

### UploadedFile lossiness

[comment](https://github.com/pydantic/pydantic-ai/pull/3971#discussion_r2949602289)

`UploadedFile` now round-trips as `ActivityMessage(activity_type='pydantic_ai_uploaded_file')` when `preserve_file_data=True`, preserving `file_id`, `provider_name`, `media_type`, `identifier`, and `vendor_metadata`.

### Field naming and docstring quality

[comment](https://github.com/pydantic/pydantic-ai/pull/3971#discussion_r2949602609), [comment](https://github.com/pydantic/pydantic-ai/pull/3971#discussion_r2949606919), [comment](https://github.com/pydantic/pydantic-ai/pull/3971#discussion_r2949620826)

- Renamed `include_file_parts` → `preserve_file_data` with user-focused docstring (no `FilePart` references).
- Links to AG-UI activities docs, warns that `pydantic_ai_*` activity types should be ignored by frontend handlers.
- All double backticks → single backticks.

### Lossiness documentation

[comment](https://github.com/pydantic/pydantic-ai/pull/3971#discussion_r2954608800)

Added `BuiltinToolCallPart.id`/`.provider_details` and `BuiltinToolReturnPart.provider_details` to the `dump_messages` lossiness notes.

### Assert `_reasoning_message_id` non-None

[comment](https://github.com/pydantic/pydantic-ai/pull/3971#discussion_r2954609558)

Replaced `self._reasoning_message_id or ''` with `assert self._reasoning_message_id is not None` in both `handle_thinking_delta` and `handle_thinking_end`.

pydantic_ai_slim/pydantic_ai/ag_ui.py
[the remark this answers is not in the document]
dsfaccini: Claude here: Good catch — added `ag_ui_version` parameter to `handle_ag_ui_request` and forwarding it through to `dispatch_request`.

pydantic_ai_slim/pydantic_ai/run.py
```diff
@@ -201,6 +229,59 @@ def _task_to_node(
...
+        self,
+        node: _agent_graph.AgentNode[AgentDepsT, Any],
+    ) -> _agent_graph.AgentNode[AgentDepsT, Any] | End[FinalResult[Any]]:
+        """Execute a single graph step without firing capability hooks."""
+        task = [self._node_to_task(node)]
+        try:
+            task = await self._graph_run.next(task)
+        except StopAsyncIteration:
+            pass
+        return self._task_to_node(task)
```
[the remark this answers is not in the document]
dsfaccini: this file wasn't changed in this PR... idk where this issue comes from @DouweM

pydantic_ai_slim/pydantic_ai/profiles/openai.py
```diff
@@ -8,8 +8,20 @@
...
 
+OPENAI_REASONING_EFFORT_MAP: dict[ThinkingLevel, str] = {
+    True: 'medium',
+    False: 'none',
+    'minimal': 'minimal',
+    'low': 'low',
+    'medium': 'medium',
+    'high': 'high',
+    'xhigh': 'xhigh',
+}
```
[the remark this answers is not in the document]
dsfaccini: this file wasn't changed in this PR....

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
```diff
@@ -60,8 +66,12 @@
         'you can use the `ag-ui` optional group — `pip install "pydantic-ai-slim[ag-ui]"`'
     ) from e
 
+AGUIVersion = Literal['0.1.10', '0.1.13']
+"""Supported AG-UI protocol versions for thinking/reasoning event emission."""
+
```
dsfaccini: @DouweM is this okay or do you want full semver `>=`

pydantic_ai_slim/pydantic_ai/profiles/openai.py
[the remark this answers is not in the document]
dsfaccini: Claude here: This file is not modified by this PR — `git diff origin/main...HEAD -- pydantic_ai_slim/pydantic_ai/profiles/openai.py` is empty. The map was introduced in a0bc09c2 on main. If this is a valid concern, it should be filed as a separate issue.

pydantic_ai_slim/pydantic_ai/run.py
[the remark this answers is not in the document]
dsfaccini: Claude here: This file is not modified by this PR — `git diff origin/main...HEAD -- pydantic_ai_slim/pydantic_ai/run.py` is empty. The code was introduced in a0bc09c2 on main. If this is a valid concern, it should be filed as a separate issue.

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
DouweM: Ideally the user would be able to set the exact version they're using on the frontend (e.g. `0.1.14`) and we would figure out what that means for us, e.g. if the relevant boundaries are only 0.1.10, 0.1.13 etc. I don't think the user needs to set `>=`.

dsfaccini: claude: can we do full semver checks? i.e. `if version >= 0.1.13: ...`

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
```diff
@@ -147,33 +178,105 @@ async def handle_text_end(self, part: TextPart, followed_by_text: bool = False)
...
+        elif self.ag_ui_version == '0.1.13':
+            if part.content:
+                yield ReasoningStartEvent(message_id=self._reasoning_message_id)
+                self._reasoning_started = True
+                yield ReasoningMessageStartEvent(message_id=self._reasoning_message_id, role='assistant')
+                yield ReasoningMessageContentEvent(message_id=self._reasoning_message_id, delta=part.content)
+                self._reasoning_text = True
+        else:
+            # exhaustive branching protects against future additions of AG-UI versions
+            assert_never(self.ag_ui_version)
```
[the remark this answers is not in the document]
dsfaccini: Claude here: Correct, this is intentional. Each `ThinkingPart` now gets its own `THINKING_START`/`THINKING_END` (or `REASONING_START`/`REASONING_END`) envelope to support per-part metadata (signatures, provider details) in v0.1.13. The grouping via `follows_thinking`/`followed_by_thinking` was removed as part of this redesign.

pydantic_ai_slim/pyproject.toml
```diff
@@ -119,7 +119,7 @@ ui = ["starlette>=0.45.3"]
 # A2A
 a2a = ["fasta2a>=0.4.1"]
 # AG-UI
-ag-ui = ["ag-ui-protocol>=0.1.10", "starlette>=0.45.3"]
+ag-ui = ["ag-ui-protocol>=0.1.13", "starlette>=0.45.3"]
```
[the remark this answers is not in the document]
DouweM: Upgrading the Python package makes sense, as long as we would still work correctly with frontends using other versions.

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
```diff
@@ -91,22 +113,89 @@ def label(self) -> str:
...
+    else:
+        assert_never(item)
+
+
+@dataclass
 class AGUIAdapter(UIAdapter[RunAgentInput, Message, BaseEvent, AgentDepsT, OutputDataT]):
     """UI adapter for the Agent-User Interaction (AG-UI) protocol."""
 
+    _: KW_ONLY
+    ag_ui_version: str = '0.1.10'
```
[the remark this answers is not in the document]
DouweM: Vercel only makes breaking changes in major releases (presumably), so we don't need full version strings. But AG-UI is still in v0, so can make breaking changes at any version, and the usr shouldn't need to know that if they're using package version 0.1.14, they're really using "protocol version" 0.1.13. So (as long as these version numbers match frontend library versions!) I'd rather have the user put their exact version number, and we handle what to do.

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
```diff
@@ -67,15 +73,43 @@
...
     """UI event stream transformer for the Agent-User Interaction (AG-UI) protocol."""
 
-    _thinking_text: bool = False
+    ag_ui_version: str = '0.1.10'
+
+    _reasoning_message_id: str | None = None
+    _reasoning_started: bool = False
+    _reasoning_text: bool = False
     _builtin_tool_call_ids: dict[str, str] = field(default_factory=dict[str, str])
+    _ended_tool_call_ids: set[str] = field(default_factory=set[str])
```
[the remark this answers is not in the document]
DouweM: OK to keep here, but then the title / description need updates!

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
```diff
@@ -147,33 +181,98 @@ async def handle_text_end(self, part: TextPart, followed_by_text: bool = False)
...
+        else:
+            if not self._reasoning_started:
+                yield ThinkingStartEvent()
+                self._reasoning_started = True
+
+            if not self._reasoning_text:
+                yield ThinkingTextMessageStartEvent()
+                self._reasoning_text = True
+
+            yield ThinkingTextMessageContentEvent(delta=delta.content_delta)
```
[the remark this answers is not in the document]
DouweM: Good idea

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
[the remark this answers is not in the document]
dsfaccini: yeah makes sense, we can adjust the type to those literals

pydantic_ai_slim/pyproject.toml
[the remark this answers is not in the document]
dsfaccini: this is fair, lets explore how we can gate the code instead of raising the version

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
```diff
@@ -67,15 +73,43 @@
     'RunFinishedEvent',
 ]
 
+
+def parse_ag_ui_version(version: str) -> tuple[int, ...]:
+    """Parse an AG-UI version string into a comparable tuple."""
+    return tuple(int(x) for x in version.split('.'))
```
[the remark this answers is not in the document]
dsfaccini: this parsing is also not desired, why are we not using `semver.Version.parse`?

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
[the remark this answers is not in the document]
dsfaccini: we adopted #4733 but didn't update the PR description and title, claude: please update

pydantic_ai_slim/pydantic_ai/ag_ui.py
```diff
@@ -53,6 +53,7 @@ async def handle_ag_ui_request(
     agent: AbstractAgent[AgentDepsT, Any],
     request: Request,
     *,
+    ag_ui_version: str = '0.1.10',
```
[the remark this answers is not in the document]
dsfaccini: need to explore

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
```diff
@@ -235,7 +324,282 @@ def load_messages(cls, messages: Sequence[Message]) -> list[ModelMessage]:  # no
...
+
+        Note: The round-trip `dump_messages` -> `load_messages` is not fully lossless:
+
+        - `TextPart.id`, `.provider_name`, `.provider_details` are lost.
+        - `ToolCallPart.id`, `.provider_name`, `.provider_details` are lost.
+        - `BuiltinToolCallPart.id`, `.provider_details` are lost (only `.provider_name` survives
+          via the prefixed tool call ID).
+        - `BuiltinToolReturnPart.provider_details` is lost.
+        - `RetryPromptPart` becomes `ToolReturnPart` (or `UserPromptPart`) on reload.
+        - `CachePoint` and `UploadedFile` content items are dropped (unless `preserve_file_data=True`).
```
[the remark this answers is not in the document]
dsfaccini: good catch

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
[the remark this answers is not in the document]
dsfaccini: good idea

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
```diff
@@ -235,7 +324,282 @@ def load_messages(cls, messages: Sequence[Message]) -> list[ModelMessage]:  # no
...
+                        )
+                    )
+            else:
+                assert_never(part)
+
+        flush()
+        return result
+
+    @classmethod
+    def dump_messages(
```
DouweM, line 667: - This method is new, so I assume there are some other issues we can close?
- Also, here's a new type for you to consider: #4432

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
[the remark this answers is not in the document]
dsfaccini: correction we talked about with @DouweM
- we won't track frontend package versions
- we only track the python-side version
- thus, and for autocomplete purposes, we can reintroduce the AGUIVersion type instead of the general `str` type
- doesn't change any logic other than narrowing the type accepted and passed

one thing we didn't discuss but we can do for better UX:
- instead of defaulting to the lowest version always, we can default to the version that we identify as installed

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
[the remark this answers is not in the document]
dsfaccini: decided to keep simple parsing but fixed validation to catch bad formats

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
```diff
@@ -147,33 +216,98 @@ async def handle_text_end(self, part: TextPart, followed_by_text: bool = False)
...
+                yield ThinkingStartEvent()
+
+            if self._reasoning_text:
+                yield ThinkingTextMessageEndEvent()
+                self._reasoning_text = False
 
-        if not followed_by_thinking:
-            yield ThinkingEndEvent(type=EventType.THINKING_END)
+            yield ThinkingEndEvent()
+            self._reasoning_message_id = None
```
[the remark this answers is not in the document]
dsfaccini: Claude here: This is intentional. For the Reasoning path (>=0.1.13), per-part START/END is required because each ThinkingPart carries its own encrypted_metadata (thinking signature) that needs a 1:1 mapping to a ReasoningEncryptedValueEvent. Grouping would lose this association. The legacy THINKING_* path changed as a side effect to keep both paths uniform, but this is inconsequential — the AG-UI spec does not mandate grouping and frontends handle either approach correctly.

pydantic_ai_slim/pyproject.toml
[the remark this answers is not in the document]
dsfaccini: Claude here: Fixed — reverted `ag-ui-protocol>=0.1.13` back to `>=0.1.10`. Reasoning imports are now gated behind try/except and TYPE_CHECKING guards so older versions skip reasoning gracefully.

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
```diff
@@ -35,6 +36,12 @@
     from ag_ui.core import (
         BaseEvent,
         EventType,
+        ReasoningEncryptedValueEvent,
+        ReasoningEndEvent,
+        ReasoningMessageContentEvent,
+        ReasoningMessageEndEvent,
+        ReasoningMessageStartEvent,
+        ReasoningStartEvent,
```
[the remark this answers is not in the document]
dsfaccini: Claude here: Fixed — Reasoning event types are no longer imported in `_event_stream.py`. The version-specific thinking/reasoning logic is extracted into `_thinking_0_10.py` and `_thinking_0_13.py`, imported lazily based on `_use_reasoning`.

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
```diff
@@ -39,25 +50,38 @@
         BaseEvent,
         BinaryInputContent,
         DeveloperMessage,
+        FunctionCall,
         Message,
+        ReasoningMessage,
```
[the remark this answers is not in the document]
dsfaccini: Claude here: Fixed — `ReasoningMessage` is now imported behind a `TYPE_CHECKING` guard with a runtime stub fallback for ag-ui-protocol < 0.1.13. The match/case remains exhaustive for pyright.

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
```diff
@@ -60,8 +67,47 @@
...
+
+def _detect_ag_ui_version() -> AGUIVersion:
+    """Detect installed ag-ui-protocol version and map to the nearest supported `AGUIVersion`."""
+    try:
+        installed = importlib.metadata.version('ag-ui-protocol')
+        if parse_ag_ui_version(installed) >= REASONING_VERSION:
+            return '0.1.13'
+    except importlib.metadata.PackageNotFoundError:
+        pass
+    return '0.1.10'
```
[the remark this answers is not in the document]
dsfaccini: Claude here: Fixed — `parse_ag_ui_version` now uses a regex to extract the numeric prefix, stripping pre-release suffixes like `a1`, `b2`, `rc1`, `.dev0`. Additionally, `_detect_ag_ui_version` catches all exceptions (not just `PackageNotFoundError`) to gracefully fall back to `0.1.10`.

scripts/check_cassettes.py
```diff
@@ -11,52 +11,73 @@
...
+        elif isinstance(node, ast.ClassDef):
+            class_has_vcr = _has_vcr_marker(node.decorator_list)
+            for method in ast.iter_child_nodes(node):
+                if not isinstance(method, ast.FunctionDef | ast.AsyncFunctionDef):
+                    continue
+                if not method.name.startswith('test_'):
+                    continue
+                if module_has_vcr or class_has_vcr or _has_vcr_marker(method.decorator_list):
+                    cassette_names.add(_sanitize_cassette_name(f'{node.name}.{method.name}'))
+
```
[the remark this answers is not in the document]
dsfaccini: yeah I stole it from #4832 the original 1. took too long and 2. sometimes errored and caused memory issues

this approach is faster and we don't care about custom cassette names because we don't use them

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
```diff
@@ -235,7 +338,286 @@ def load_messages(cls, messages: Sequence[Message]) -> list[ModelMessage]:  # no
...
+                                        vendor_metadata=activity_content.get('vendor_metadata'),
+                                        media_type=activity_content.get('media_type'),
+                                        identifier=activity_content.get('identifier'),
+                                    )
+                                ]
+                            )
+                        )
+
+                case _:
+                    assert_never(msg)
```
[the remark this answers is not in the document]
dsfaccini: yeah but there's also a rule to end if/else chains with an exhaustiveness check which I think is more important, and if a user stumbles upon a crash, fixing the issue initially is as easy as adding a branch to acknowledge the type and skip it while actual implementation of the new feature is discussed. this is reflected in other files like models, where we explicitly skip over a list of potential classes that we don't handle.

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
```diff
@@ -60,8 +56,52 @@
...
+- `'0.1.10'`: emits `THINKING_*` events, drops `ThinkingPart` from `dump_messages`.
+- `'0.1.13'`: emits `REASONING_*` events with encrypted metadata, preserves `ThinkingPart`
+  as `ReasoningMessage` in `dump_messages` for full round-trip fidelity.
+"""
+
+REASONING_VERSION = (0, 1, 13)
+"""AG-UI version that introduced REASONING_* events (replacing THINKING_*)."""
+
+
+def parse_ag_ui_version(version: str) -> tuple[int, ...]:
```
DouweM: Since this is not event stream specific and would be useful in the adapter as well, let's move to shared _utils module.  I believe Vercel Ai already has one

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
```diff
@@ -60,8 +56,52 @@
...
+    """Parse an AG-UI version string (e.g. `'0.1.13'`) into a comparable tuple.
+
+    Pre-release suffixes like `a1`, `b2`, `rc1`, `.dev0` are stripped before parsing.
+    """
+    import re
+
+    from ...exceptions import UserError
```
DouweM: Wtf :) Up top please.

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
```diff
@@ -60,8 +56,52 @@
...
+    if not match:
+        raise UserError(f"Invalid AG-UI version {version!r}: expected a dotted numeric version like '0.1.13'")
+    return tuple(int(x) for x in match.group(1).split('.'))
+
+
+def _detect_ag_ui_version() -> AGUIVersion:
+    """Detect installed ag-ui-protocol version and map to the nearest supported `AGUIVersion`."""
+    try:
+        installed = importlib.metadata.version('ag-ui-protocol')
+        if parse_ag_ui_version(installed) >= REASONING_VERSION:
```
DouweM: This is weird, not generic enough to hard-code this to the reasoning constant. If we're going to do the regex tuple business further up, we should definitely make this actually find the nearest version in the `AGUIVersion` union.

Also, it's a bit weird that we only consider 0.1.10 and 0.1.13 "AGUI versions". What if we just detect the tuple, and then somewhere have a check like `reasoning_part_available = ag_ui_version >= (0, 1, 13)`?

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
```diff
@@ -147,33 +210,42 @@ async def handle_text_end(self, part: TextPart, followed_by_text: bool = False)
...
-        if part.content:
-            yield ThinkingTextMessageStartEvent(type=EventType.THINKING_TEXT_MESSAGE_START)
-            yield ThinkingTextMessageContentEvent(type=EventType.THINKING_TEXT_MESSAGE_CONTENT, delta=part.content)
-            self._thinking_text = True
+        if self._use_reasoning:
+            from ._thinking_0_13 import handle_thinking_start as _impl
+        else:
+            from ._thinking_0_10 import handle_thinking_start as _impl
+        async for event in _impl(self, part):
+            yield event
```
DouweM, line 160: I don't mind this that much, but what do you think of having `AGUIEventStream` subclasses for newer versions that override the appropriate methods? Then `AGUIAdapter.event_stream` can return the correct one based on its `ag_ui_version`.

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
```diff
@@ -203,16 +275,21 @@ async def handle_tool_call_delta(self, delta: ToolCallPartDelta) -> AsyncIterato
...
             tool_call_id = self._builtin_tool_call_ids[tool_call_id]
+        if tool_call_id in self._ended_tool_call_ids:
+            return
         yield ToolCallArgsEvent(
             tool_call_id=tool_call_id,
             delta=delta.args_delta if isinstance(delta.args_delta, str) else json.dumps(delta.args_delta),
         )
 
     async def handle_tool_call_end(self, part: ToolCallPart) -> AsyncIterator[BaseEvent]:
+        self._ended_tool_call_ids.add(part.tool_call_id)
```
DouweM: I don't think this makes sense, reading https://github.com/pydantic/pydantic-ai/issues/4733, the real root cause seems to be that `OpenAIResponsesModel` emits a tool call arg delta event after having already emitted a built-in tool call end event for that same builtin in tool call ID. So we should see if we can fix that at the root, before seeing if this hacky patch is necessary.

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
```diff
@@ -91,22 +124,92 @@ def label(self) -> str:
...
+def _user_content_to_input(
+    item: str | TextContent | ImageUrl | VideoUrl | AudioUrl | DocumentUrl | BinaryContent | UploadedFile | CachePoint,
+) -> TextInputContent | BinaryInputContent | None:
+    """Convert a user content item to AG-UI input content."""
+    if isinstance(item, str):
+        return TextInputContent(type='text', text=item)
+    elif isinstance(item, TextContent):
+        return TextInputContent(type='text', text=item.content)
+    elif isinstance(item, (ImageUrl, VideoUrl, AudioUrl, DocumentUrl)):
+        return BinaryInputContent(type='binary', url=item.url, mime_type=item.media_type or '')
```
DouweM, line 185: https://docs.ag-ui.com/concepts/messages#user-messages has multi-modal types like `DocumentInputContent` and says:

> In Python, the previous BinaryInputContent model is deprecated and remains temporarily available as a compatibility path.

So we should support that right away, while we're at it. Should be a single prompt :) And we'll likely have to add a new version boundary for this feature, which will make that stuff a bit better.

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
```diff
@@ -235,7 +338,288 @@ def load_messages(cls, messages: Sequence[Message]) -> list[ModelMessage]:  # no
...
+                            uploaded_content: dict[str, Any] = {
+                                'file_id': item.file_id,
+                                'provider_name': item.provider_name,
+                                'media_type': item.media_type,
+                                'identifier': item.identifier,
+                            }
+                            if item.vendor_metadata is not None:
+                                uploaded_content['vendor_metadata'] = item.vendor_metadata
+                            result.append(
+                                ActivityMessage(
```
DouweM, line 515: This needs more code comments about how hacky this is, that we abuse a feature, etc.

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
```diff
@@ -235,7 +338,288 @@ def load_messages(cls, messages: Sequence[Message]) -> list[ModelMessage]:  # no
...
+        user_content: list[TextInputContent | BinaryInputContent] = []
+
+        for part in msg.parts:
+            if isinstance(part, SystemPromptPart):
+                system_content.append(part.content)
+            elif isinstance(part, UserPromptPart):
+                if isinstance(part.content, str):
+                    user_content.append(TextInputContent(type='text', text=part.content))
+                else:
+                    for item in part.content:
```
DouweM, line 501: Would be nice to use a typeddict here :)

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
```diff
@@ -235,7 +338,288 @@ def load_messages(cls, messages: Sequence[Message]) -> list[ModelMessage]:  # no
...
+                            content=reasoning_msg.content,
+                            id=metadata.get('id'),
+                            signature=metadata.get('signature'),
+                            provider_name=metadata.get('provider_name'),
+                            provider_details=metadata.get('provider_details'),
+                        )
+                    )
+
+                case ActivityMessage() as activity_msg:
+                    if activity_msg.activity_type == 'pydantic_ai_file' and preserve_file_data:
```
DouweM: Magic strings should be constants!

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
```diff
@@ -235,7 +338,288 @@ def load_messages(cls, messages: Sequence[Message]) -> list[ModelMessage]:  # no
...
+                            )
+                        builder.add(
+                            FilePart(
+                                content=BinaryContent.from_data_uri(url),
+                                id=activity_content.get('id'),
+                                provider_name=activity_content.get('provider_name'),
+                                provider_details=activity_content.get('provider_details'),
+                            )
+                        )
+                    elif activity_msg.activity_type == 'pydantic_ai_uploaded_file' and preserve_file_data:
```
DouweM: Constant

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
dsfaccini: Claude here: We defined `FileActivityContent` and `UploadedFileActivityContent` TypedDicts in `_utils.py` to document the expected schema. However, `ActivityMessage.content` is typed as `dict[str, Any]`, and in pyright strict mode TypedDicts aren't assignable to `dict[str, Any]` (invariance). So we use `dict[str, Any]` at the construction sites with comments referencing the TypedDicts (e.g. "See `UploadedFileActivityContent`").

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
dsfaccini: Claude here: We investigated this. The stray deltas originate from OpenAI's `ResponseMcpCallArgumentsDeltaEvent` arriving after `ResponseOutputItemDoneEvent` for the same MCP tool call. However, filtering at the OpenAI model level isn't the right fix:

1. The OpenAI model's `_parts_manager` doesn't track tool call completion state
2. AG-UI owns the 'tool call end' semantics — other UI consumers (Vercel AI, plain streaming) don't have this concept
3. The model iterator processes events sequentially without context about downstream UI state

The AG-UI adapter is the correct layer for this: `handle_tool_call_delta` has direct visibility into which tool calls have ended (via `_ended_tool_call_ids`), and silently drops stray deltas. This is specific to AG-UI's protocol contract, not a general model issue.

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
dsfaccini: yeahp

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
dsfaccini: discussed in call and decided to keep the functionall approach

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
```diff
@@ -235,7 +392,305 @@ def load_messages(cls, messages: Sequence[Message]) -> list[ModelMessage]:  # no
...
+
+                    flush()
+                    encrypted = thinking_encrypted_metadata(part)
+                    result.append(
+                        ReasoningMessage(
+                            id=_new_message_id(),
+                            content=part.content,
+                            encrypted_value=json.dumps(encrypted) if encrypted else None,
+                        )
+                    )
```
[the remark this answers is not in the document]
dsfaccini: user would need to point the gun and shoot it into his foot for this to happen

dsfaccini:
claude on the `stray deltas` issue:

```
  Issue #4733 — Complete Analysis

  What the reporter saw

  User Ali-Aleph-Alpha uses an internal OpenAI Responses API-compatible endpoint (not the official OpenAI API) with MCPServerTool in streaming mode via AG-UI. The SSE stream produces:

  TOOL_CALL_ARGS   (delta: '{"action":"call_tool",...,"tool_args":')
  TOOL_CALL_ARGS   (delta: '}')
  TOOL_CALL_ARGS   (delta: '}')   ← stray, after end

  This violates AG-UI protocol ordering and crashes the @ag-ui/client verifier.

  Normal event flow (official OpenAI API)

  From the VCR cassette, the standard OpenAI API sends events in this order:

  1. ResponseOutputItemAddedEvent(McpCall) → PartStartEvent + first PartDeltaEvent
  2. ResponseMcpCallArgumentsDeltaEvent → more PartDeltaEvents
  3. ResponseMcpCallArgumentsDoneEvent → final PartDeltaEvent (closing })
  4. ResponseOutputItemDoneEvent(McpCall) → triggers PartEndEvent (TOOL_CALL_END) + PartStartEvent for return

  In this order, no stray delta occurs — all argument deltas finish before the item is "done."

  How the stray delta happens (non-standard API)

  The reporter's non-standard endpoint sends ResponseOutputItemDoneEvent before ResponseMcpCallArgumentsDoneEvent:

  1. ResponseOutputItemAddedEvent → PartStart + PartDelta (prefix)
  2. ResponseMcpCallArgumentsDeltaEvent → PartDelta (args)
  3. ResponseOutputItemDoneEvent → triggers PartEndEvent (TOOL_CALL_END) + PartStart for return
  4. ResponseMcpCallArgumentsDoneEvent → PartDelta (closing }) → STRAY — after end

  The iterator_with_part_end wrapper in models/__init__.py:1046-1081 injects PartEndEvent when a new PartStartEvent arrives for a different part. It has no way to know more deltas are still incoming for the old part.

  Why the parts manager can't filter it

  ModelResponsePartsManager has no concept of "done" state. It stores parts by vendor_part_id and applies deltas unconditionally. PartEndEvent is generated by the iterator_with_part_end wrapper, a separate layer the parts manager doesn't see.

  Where could we fix it?

  Option A: In _get_event_iterator (openai.py) — Track completed vendor_part_ids when ResponseOutputItemDoneEvent fires. Skip ResponseMcpCallArgumentsDeltaEvent/ResponseMcpCallArgumentsDoneEvent for completed IDs. This prevents stray deltas
  from reaching ANY downstream consumer (AG-UI, Vercel AI, plain streaming).

  Option B: In iterator_with_part_end (models/init.py) — After emitting PartEndEvent for a vendor_part_id, suppress any subsequent PartDeltaEvent for that same ID. This is the most generic fix — works for all models, all UI adapters.

  Option C: In each UI adapter — What we have now with _ended_tool_call_ids. Only protects AG-UI, Vercel AI has no such guard and is vulnerable.

  Douwe's position (from the call transcript)

  Douwe was very clear: this is a model-level bug, not an AG-UI issue. The fix belongs upstream. The _ended_tool_call_ids hack should be removed from this PR. The upstream fix should be a separate PR.
  ```

File: tests/test_ag_ui.py [hunk 54]
```diff
@@ -2575,6 +3379,96 @@ async def send(data: MutableMapping[str, Any]) -> None:
...
+            parts=[
+                ThinkingPart(content='Deep thoughts...', signature='sig_xyz'),
+                TextPart(content='Conclusion'),
+            ]
+        ),
+    ]
+
+    ag_ui_msgs = AGUIAdapter.dump_messages(messages, ag_ui_version='0.1.13')
+    reasoning_msgs = [m for m in ag_ui_msgs if isinstance(m, ReasoningMessage)]
+    assert len(reasoning_msgs) == 1
```
dsfaccini: we could snapshot this but I don't feel strongly enough about it, just noting it in case you do @DouweM

PR #3972, opened 2026-01-09
Add `AgentDepsT` to `AgentRunResult`

### LINKED ISSUE #3874: Give `on_complete` callback access to `deps`
### Description

When using `UIAdapter.run_stream()` the `on_complete()` callback only has direct access to the final AgentRunResult. In case we would want to use the agent dependencies (eg. to use a db connection and persist the messages) inside the `on_complete()` callback the only way to do it would be using the local variable deps (that is accessible to on_complete because is passed to the `run_stream()` method that calls it) referencing it with something like `global deps`.

Passing deps to `on_complete()` directly would be the optimal solution.

### References

_No response_

ferreroal:
Include deps to AgentRunResult to use it when needed, eg. inside `on_complete `hook for UI protocols

- Closes #3874 

### Pre-Review Checklist

### Pre-Merge Checklist


pydantic_ai_slim/pydantic_ai/agent/abstract.py
```diff
@@ -169,7 +169,7 @@ async def run(
         toolsets: Sequence[AbstractToolset[AgentDepsT]] | None = None,
         builtin_tools: Sequence[AbstractBuiltinTool | BuiltinToolFunc[AgentDepsT]] | None = None,
         event_stream_handler: EventStreamHandler[AgentDepsT] | None = None,
-    ) -> AgentRunResult[OutputDataT]: ...
+    ) -> AgentRunResult[AgentDepsT, OutputDataT]: ...
```
DouweM, line 172: While it's more consistent with everything else that takes these 2 params, we can't change the meaning of existing `AgentRunResult[Foo]` in user code, so we have to add the new param second, unfortunately.

pydantic_ai_slim/pydantic_ai/run.py
```diff
@@ -302,9 +303,12 @@ def __repr__(self) -> str:  # pragma: no cover
 
 
 @dataclasses.dataclass
-class AgentRunResult(Generic[OutputDataT]):
+class AgentRunResult(Generic[AgentDepsT, OutputDataT]):
     """The final result of an agent run."""
 
+    deps: AgentDepsT
```
DouweM, line 309: This is a public class so we can't:
- Reorder arguments (i.e. add a new one in front of an existing one)
- Add required arguments (i.e. this needs `= None`)

PR #3975, opened 2026-01-10
  Fix(evals): Rich `MarkupError` when error messages contain special characters

yongfenggu:
## Summary

  Fix `MarkupError` in evaluation failure reporting when error messages or stacktraces contain special characters like brackets.

  ## Problem

  When rendering evaluation failures in Rich tables, error messages containing special characters (e.g., `KeyError: 'user[0].name'`, `list[int]`, `dict[str, Any]`) are incorrectly parsed as Rich markup tags, resulting in:

  rich.errors.MarkupError: closing tag '[/?#]' at position 26 doesn't match any open tag
68e23dfe85b3f9a83dddc40e3c3bff4b

  ## Solution

  - Wrap `error_message` and `error_stacktrace` fields in `Text()` objects to prevent Rich from parsing them as markup
  - Update `build_failure_row()` return type from `list[str]` to `list[str | Text]` to reflect the actual types
  - Preserve `EMPTY_CELL_STR` for empty values to maintain consistent table rendering

  The `overflow='fold'` behavior is already handled at the column level (`table.add_column('Error Message', overflow='fold')`), so no additional configuration is needed on the `Text()` objects.

  ## Test Plan

  Tested with error messages containing:
  - Square brackets: `KeyError: 'user[0].name'`
  - Type annotations: `Expected list[int], got str`
  - Rich markup-like patterns: `Error in [section]`

  All cases now render correctly without `MarkupError`.


DouweM:
@GGyongfeng Thank you, fix looks good! Can you please add a test that would have failed before these changes?

dsfaccini:

```bash
git pull

deactivate

rm -rf .venv

make install
```

PR #3997, opened 2026-01-13
Update the default behavior for `OpenRouter` from downloading to sending `DocumentUrl`s

### LINKED ISSUE #3993: `DocumentUrl` ignores `force_download` flag, always downloads files (unlike `ImageUrl`)
## Bug Description

`DocumentUrl` always downloads files when using OpenAI/OpenRouter models, regardless of the `force_download` setting. This is inconsistent with `ImageUrl` behavior and contradicts the documentation.

## Expected Behavior

Per the [documentation](https://ai.pydantic.dev/input/#image-input):
> "you can instruct Pydantic AI to download the file content locally and send that instead of the URL by setting `force_download` on the URL object"

This implies `force_download=False` (the default) should pass URLs directly without downloading.

## Actual Behavior

`DocumentUrl` unconditionally downloads files, ignoring `force_download`.

## Code Evidence

**ImageUrl (correct)** - `models/openai.py` lines 936-943:
```python
elif isinstance(item, ImageUrl):
    image_url: ImageURL = {'url': item.url}  # Default: pass URL directly
    # ...
    if item.force_download:  # Only download if explicitly requested
        image_content = await download_item(...)
        image_url['url'] = image_content['data']
```

**DocumentUrl (bug)** - `models/openai.py` lines 983-1003:
```python
elif isinstance(item, DocumentUrl):
    if self._is_text_like_media_type(item.media_type):
        downloaded_text = await download_item(...)  # Always downloads!
    else:
        downloaded_item = await download_item(...)  # Always downloads!
```

No check for `item.force_download` - it unconditionally downloads.

## Impact

This is especially problematic with OpenRouter, which [supports PDF URLs directly](https://openrouter.ai/docs/guides/overview/multimodal/pdfs#using-pdf-urls):

```json
{
  "type": "file",
  "file": {
    "filename": "document.pdf",
    "file_data": "https://example.com/document.pdf"
  }
}
```

Currently, pydantic-ai downloads the entire file, converts to base64, and sends that instead of just passing the URL.

## Suggested Fix

Mirror the `ImageUrl` pattern:

```python
elif isinstance(item, DocumentUrl):
    if item.force_download:
        # Download and convert (existing logic)
    else:
        # Pass URL directly
        content.append(File(file=FileFile(file_data=item.url, filename=...)))
```

## Environment

- pydantic-ai version: 1.27.0
- Python version: 3.11
- Model: OpenRouter (anthropic/claude-sonnet-4.5)

dsfaccini:
## Summary
- Add `openai_chat_supports_file_urls` profile flag to `OpenAIModelProfile`
- Respect `force_download` flag in `DocumentUrl` for OpenRouter provider
- When `force_download=False` and provider supports file URLs: pass URL directly
- When `force_download=True` or provider doesn't support URLs: download and encode (existing behavior)
- Add comment explaining why native OpenAI Chat API always downloads (only supports base64)

**Note:** This fix is OpenRouter-specific because OpenAI's native Chat API `FileFile` type only supports `file_data` (base64) and `file_id`, not URLs. The Responses API already correctly handles this via `file_url` field.

Fixes #3993

- [x] New test

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/docs/input.md
+++ b/docs/input.md
@@ -111,13 +111,13 @@ Support for file URLs varies depending on type and provider:
 | Model | Send URL directly | Download and send bytes | Unsupported |
 |-------|-------------------|-------------------------|-------------|
 | [`OpenAIChatModel`][pydantic_ai.models.openai.OpenAIChatModel] | `ImageUrl` | `AudioUrl`, `DocumentUrl` | `VideoUrl` |
-| [`OpenRouterModel`][pydantic_ai.models.openrouter.OpenRouterModel] | `ImageUrl`, `DocumentUrl` | `AudioUrl` | `VideoUrl` |
 | [`OpenAIResponsesModel`][pydantic_ai.models.openai.OpenAIResponsesModel] | `ImageUrl`, `AudioUrl`, `DocumentUrl` | — | `VideoUrl` |
 | [`AnthropicModel`][pydantic_ai.models.anthropic.AnthropicModel] | `ImageUrl`, `DocumentUrl` (PDF) | `DocumentUrl` (`text/plain`) | `AudioUrl`, `VideoUrl` |
 | [`GoogleModel`][pydantic_ai.models.google.GoogleModel] (Vertex) | All URL types | — | — |
 | [`GoogleModel`][pydantic_ai.models.google.GoogleModel] (GLA) | [YouTube](models/google.md#document-image-audio-and-video-input), [Files API](models/google.md#document-image-audio-and-video-input) | All other URLs | — |
 | [`MistralModel`][pydantic_ai.models.mistral.MistralModel] | `ImageUrl`, `DocumentUrl` (PDF) | — | `AudioUrl`, `VideoUrl`, `DocumentUrl` (non-PDF) |
 | [`BedrockConverseModel`][pydantic_ai.models.bedrock.BedrockConverseModel] | S3 URLs (`s3://`) | `ImageUrl`, `DocumentUrl`, `VideoUrl` | `AudioUrl` |
+| [`OpenRouterModel`][pydantic_ai.models.openrouter.OpenRouterModel] | `ImageUrl`, `DocumentUrl` | `AudioUrl` | `VideoUrl` |
 
 A model API may be unable to download a file (e.g., because of crawling or access restrictions) even if it supports file URLs. For example, [`GoogleModel`][pydantic_ai.models.google.GoogleModel] on Vertex AI limits YouTube video URLs to one URL per request. In such cases, you can instruct Pydantic AI to download the file content locally and send that instead of the URL by setting `force_download` on the URL object:
 


pydantic_ai_slim/pydantic_ai/profiles/openai.py
```diff
@@ -70,6 +70,13 @@ class OpenAIModelProfile(ModelProfile):
     - `'uri'`: Data URI (e.g. `data:audio/wav;base64,...`).
     """
 
+    openai_chat_supports_file_urls: bool = False
+    """Whether the Chat API supports file URLs directly in the `file_data` field.
+
+    OpenAI's native Chat API only supports base64-encoded data, but some providers
+    like OpenRouter support passing URLs directly.
+    """
+
```
dsfaccini, line 79: what do you think about the name

tests/models/test_openrouter.py
```diff
@@ -648,3 +649,23 @@ async def test_openrouter_url_citation_annotation_validation(openrouter_api_key:
...
+    model = OpenRouterModel('openai/gpt-4.1-mini', provider=provider)
+    agent = Agent(model)
+
+    document_url = DocumentUrl(
+        url='https://www.w3.org/WAI/ER/tests/xhtml/testfiles/resources/pdf/dummy.pdf',
+        force_download=False,
+    )
+
+    result = await agent.run(['What is the main content of this document?', document_url])
+    assert 'dummy' in result.output.lower() or 'pdf' in result.output.lower()
```
DouweM, line 673: This test would have passed without the changes as well, as we're not testing that `force_download` is being respected / that no download is happening / that we're sending the URL. So we should probably test the actual args being sent to the API somehow.

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/profiles/openai.py
DouweM: Name is great, but the beautiful doc you wrote on how each model behaves with URLs needs updating ;)

Also the PR title doesn't make it clear that this is not just "respecting the field", but _changing the default behavior for OpenRouter from downloading to sending URLs_.

tests/models/test_openrouter.py
dsfaccini: yeahp good idea

tests/models/test_openrouter.py
```diff
@@ -648,3 +653,38 @@ async def test_openrouter_url_citation_annotation_validation(openrouter_api_key:
...
+    file_content = request_body['messages'][0]['content'][1]
+    assert file_content == snapshot(
+        {
+            'file': {
+                'file_data': 'https://www.w3.org/WAI/ER/tests/xhtml/testfiles/resources/pdf/dummy.pdf',
+                'filename': 'filename.pdf',
+            },
+            'type': 'file',
+        }
+    )
```
dsfaccini, line 688: added this check here

tests/models/test_openrouter.py
```diff
@@ -1,11 +1,15 @@
...
-from typing import Literal, cast
+from typing import TYPE_CHECKING, Literal, cast
 
 import pytest
 from inline_snapshot import snapshot
 from pydantic import BaseModel
 
+if TYPE_CHECKING:
+    from vcr.cassette import Cassette
+
```
dsfaccini: do you have something against this? I'm generally for marking type-hint-only classes explicitly like this

DouweM: Nah we should only do that when we're importing a type from an optional dependency that's not needed at runtime, only in type hints, because then without `if TYPE_CHECKING` it'd result in import errors, and if we guard against the import error, the type checker would complain that the type is not always available.

tests/models/test_openrouter.py
DouweM: Does this also work on the first run of the test when there is no cassette yet, or only on subsequent runs when there is?

dsfaccini: not sure I understand the Q, this will only pass if there's a cassette that fulfills the assertions. without a cassette it fails

tests/models/test_openrouter.py
dsfaccini: I removed it

pydantic_ai_slim/pydantic_ai/profiles/openai.py
dsfaccini: yeah title needs updating, gonna take care of the docs now

dsfaccini: updated the docs, btw thanks for the "beautiful" ✨

docs/input.md
```diff
@@ -110,14 +110,16 @@ Support for file URLs varies depending on type and provider:
...
+| [`OpenAIChatModel`][pydantic_ai.models.openai.OpenAIChatModel] | `ImageUrl` | `AudioUrl`, `DocumentUrl`[^1] | `VideoUrl` |
 | [`OpenAIResponsesModel`][pydantic_ai.models.openai.OpenAIResponsesModel] | `ImageUrl`, `AudioUrl`, `DocumentUrl` | — | `VideoUrl` |
 | [`AnthropicModel`][pydantic_ai.models.anthropic.AnthropicModel] | `ImageUrl`, `DocumentUrl` (PDF) | `DocumentUrl` (`text/plain`) | `AudioUrl`, `VideoUrl` |
 | [`GoogleModel`][pydantic_ai.models.google.GoogleModel] (Vertex) | All URL types | — | — |
 | [`GoogleModel`][pydantic_ai.models.google.GoogleModel] (GLA) | [YouTube](models/google.md#document-image-audio-and-video-input), [Files API](models/google.md#document-image-audio-and-video-input) | All other URLs | — |
 | [`MistralModel`][pydantic_ai.models.mistral.MistralModel] | `ImageUrl`, `DocumentUrl` (PDF) | — | `AudioUrl`, `VideoUrl`, `DocumentUrl` (non-PDF) |
 | [`BedrockConverseModel`][pydantic_ai.models.bedrock.BedrockConverseModel] | S3 URLs (`s3://`) | `ImageUrl`, `DocumentUrl`, `VideoUrl` | `AudioUrl` |
 
+[^1]: When using [`OpenRouterProvider`][pydantic_ai.providers.openrouter.OpenRouterProvider], `DocumentUrl` can be sent directly without downloading.
```
DouweM: Let's make this a row for `OpenRouterModel` instead

tests/models/test_openrouter.py
DouweM: Ok so this means that such an assertion based on `vcr: Cassette` can only be added to a test after it has already been run once without it and generated a cassette?

Would that mean that if you run this with `--record-mode=recreate`, the assertion above will run on the live response data, but these assertions will run on the _old_ recorded data? That'd be confusing :) Because you'd basically need to re-record _and then run it again_ to make sure the cassette-based part of the tests still succeeds.

But if `vcr.requests` refers to the cassette that's currently being recorded for this same test, we're fine.

dsfaccini: from the https://github.com/kiwicom/pytest-recording README I'd strongly assume that the assertions being made are always against the current cassette (the freshly recorded one)

image

> But if vcr.requests refers to the cassette that's currently being recorded for this same test, we're fine.

so of course I'm using it under that assumption

but to entertain the concern that assertions could be made against an old cassette:
- I locally run `--record-mode=rewrite`
- Cassette gets overwritten -> cassette file changes
- assertions are made against old cassette -> they pass
- new cassette has errors -> local tests pass anyway
- I commit and push to the PR
- CI fails cause it can only assert against the new cassette file

so it would be fine still, right?

DouweM: Yeah it's just be a bit unexpected if it only showed up at CI. But you're right that this is not an issue at all!

docs/input.md
```diff
@@ -111,6 +111,7 @@ Support for file URLs varies depending on type and provider:
 | Model | Send URL directly | Download and send bytes | Unsupported |
 |-------|-------------------|-------------------------|-------------|
 | [`OpenAIChatModel`][pydantic_ai.models.openai.OpenAIChatModel] | `ImageUrl` | `AudioUrl`, `DocumentUrl` | `VideoUrl` |
+| [`OpenRouterModel`][pydantic_ai.models.openrouter.OpenRouterModel] | `ImageUrl`, `DocumentUrl` | `AudioUrl` | `VideoUrl` |
```
DouweM: Move this down please, we either want OpenAI, Anthropic and Google (in no particular order) to be listed ahead of the rest, OR the entire list to be alphabetic.

dsfaccini: moved it down

PR #3999, opened 2026-01-13
Default to `developer` role for system prompts in OpenAI models

### LINKED ISSUE #3976: `OpenAIResponsesModel` should use `developer` role for `SystemPromptPart`s by default
### Question

Hi, I’m using Pydantic AI’s `OpenAIResponsesModel` and providing an input prompt via the `instructions` parameter. In the OpenAI Logs dashboard, requests generated with that code always have their instructions labeled as “System instructions” at the top.

But when I run the same instructions via the OpenAI Playground / Prompt dashboard, it shows up in the logs labeled as “Developer” at the top.

OpenAI’s latest docs also seem to emphasize “developer” instructions for app-level behavior (rather than "system"), which makes it seem like the best practice for their API is to set `"role": "developer"` for instructions like these, I think?

Is there a setting or hook in OpenAIResponsesModel to send these as developer instructions (or otherwise match the Playground behavior)? If not, would you consider adding one, or documenting the current behavior and recommended workaround?

Ultimately, I'd like to be able to test/optimize prompts in the playground, and then have my Pydantic AI agents in my application send requests to the API that have identical configuration as the ones I'm tweaking in the playground.

Thanks for reading!

### Additional Context

pydantic-ai v1.0.18
Python 3.13.3

danxshap:
### Changes

- Respect `OpenAIModelProfile.openai_system_prompt_role` setting when mapping system prompts in `OpenAIResponsesModel`, matching `OpenAIChatModel` behavior
- Default to `developer` role for OpenAI models per OpenAI recommendation & model spec
- Also apply the role setting when inserting instructions for JSON object output

- Closes #3976

### Pre-Review Checklist

### Pre-Merge Checklist


danxshap:
Will need to spend a bit more time looking at the tests that broke.

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -1631,7 +1632,15 @@ async def _map_messages(  # noqa: C901
...
-                        openai_messages.append(responses.EasyInputMessageParam(role='system', content=part.content))
+                        system_prompt_role = profile.openai_system_prompt_role
+                        if system_prompt_role == 'developer':
+                            openai_messages.append(
+                                responses.EasyInputMessageParam(role='developer', content=part.content)
+                            )
+                        elif system_prompt_role == 'user':
+                            openai_messages.append(responses.EasyInputMessageParam(role='user', content=part.content))
+                        else:
+                            openai_messages.append(responses.EasyInputMessageParam(role='system', content=part.content))
```
DouweM, line 1737: Let's make this (and the equiv code in `OpenAIChatModel`) a bit less repetitive as we're only changing the `role` arg

DouweM (changes requested):

dsfaccini:

```bash
git pull

deactivate

rm -rf .venv

make install
```

dsfaccini:
David's AICA here: Thanks for this, @danxshap — and sorry it sat so long. Since you opened it, the two parts of this change resolved separately:

- **Respecting `openai_system_prompt_role` in `OpenAIResponsesModel`** (the parity with `OpenAIChatModel`) shipped in #5430, merged 2026-05-14 — so `openai_system_prompt_role='developer'` already works on the Responses model on `main`.
- **Flipping the _default_ role to `developer`** was deliberately deferred by @DouweM as more controversial, to revisit for v2 ([his note on #5430](https://github.com/pydantic/pydantic-ai/pull/5430#discussion_r3244247345)). v2 has since shipped without it, and the original rationale has weakened: OpenAI's updated model spec (2025-12) now ranks `system` _above_ `developer` in the chain of command, which cuts against defaulting to `developer`.

Since the mechanical half is already on `main` and the default flip needs a fresh maintainer decision rather than a direct change, I'm closing this PR. If you'd like to make the case for the flip, the best spot is a comment on #3976 with the current model-spec reasoning. Thanks again for flagging it!

PR #4008, opened 2026-01-15
feat: add `BedrockEmbeddingModel` for Nova, Cohere and Titan endpoints

bitnahian:
- Adds to #3252 

### Pre-Review Checklist

### Pre-Merge Checklist


docs/embeddings.md
```diff
@@ -342,6 +342,204 @@ embedder = Embedder(
...
+Authentication with AWS Bedrock uses standard AWS credentials. You can configure credentials via:
+
+- **Environment variables**:
+  ```bash
+  export AWS_ACCESS_KEY_ID='your-access-key'
+  export AWS_SECRET_ACCESS_KEY='your-secret-key'
+  export AWS_REGION='us-east-1'
+  ```
+- **AWS credentials file** (`~/.aws/credentials`)
+- **IAM roles** (when running on AWS infrastructure)
```
DouweM: As we do in https://ai.pydantic.dev/embeddings/#vertex-ai, let's link to the existing Bedrock doc that is the canonical source on how to configure the provider, that also has more details like that we support the `AWS_BEARER_TOKEN_BEDROCK` env var. So I'd prefer for that to be the main place we explain all the options, and for this to have an example + link there for more details; if you want to specifically mention `~/.aws/credentials` for example, let's add it there, not here.

docs/embeddings.md
```diff
@@ -342,6 +342,204 @@ embedder = Embedder(
...
+
+Ensure your AWS account has access to the Bedrock embedding models you want to use. See [AWS Bedrock model access](https://docs.aws.amazon.com/bedrock/latest/userguide/model-access.html) for details.
+
+#### Basic Usage
+
+```python {title="bedrock_embeddings.py" test="skip"}
```
DouweM, line 420: Please don't skip testing examples unless we have no other option.

docs/embeddings.md
```diff
@@ -342,6 +342,204 @@ embedder = Embedder(
...
+    result = await embedder.embed_query('Hello world')
+    print(len(result.embeddings[0]))
+    #> 1024
+```
+
+_(This example requires AWS credentials configured)_
+
+#### Supported Models
+
+Bedrock supports three families of embedding models:
```
DouweM: This'll get outdated, let's link to Amazon's doc on this and have just 1 example here.

docs/embeddings.md
```diff
@@ -342,6 +342,204 @@ embedder = Embedder(
...
+from pydantic_ai import Embedder
+
+# Use US cross-region inference
+embedder = Embedder('bedrock:us.amazon.titan-embed-text-v2:0')
+
+# Use EU cross-region inference
+embedder = Embedder('bedrock:eu.cohere.embed-english-v3')
+```
+
+The model automatically normalizes these prefixes when looking up `max_input_tokens()`.
```
DouweM: I don't think this needs to be specified

docs/embeddings.md
```diff
@@ -342,6 +342,204 @@ embedder = Embedder(
...
+        },
+    ),
+)
+```
+
+The `extra_body` dict is deep-merged with the generated request body, allowing you to override or add any parameters.
+
+#### Using a Custom Provider
+
+For advanced configuration, you can create a [`BedrockProvider`][pydantic_ai.providers.bedrock.BedrockProvider] directly:
```
DouweM: I think we can have just one example here of building the provider yourself, and link to the existing bedrock docs for more details.

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
```diff
@@ -0,0 +1,662 @@
...
+
+import functools
+import json
+from abc import ABC, abstractmethod
+from collections.abc import Sequence
+from dataclasses import dataclass, field
+from typing import TYPE_CHECKING, Any, Literal, cast
+
+import anyio.to_thread
+from botocore.exceptions import ClientError
```
DouweM: Imports from optional dependencies should be inside `try`/`except ImportError` like on `embeddings/cohere.py`. We should make the same change in `models/bedrock.py`

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
```diff
@@ -0,0 +1,662 @@
...
+    # ==================== Cohere Settings ====================
+
+    bedrock_input_type: Literal['search_document', 'search_query', 'classification', 'clustering']
+    """The Cohere-specific input type for the embedding (Cohere models only).
+
+    Overrides the automatic mapping from `input_type` ('query' -> 'search_query',
+    'document' -> 'search_document').
+    """
+
+    bedrock_truncate: Literal['NONE', 'START', 'END']
```
DouweM: In https://github.com/pydantic/pydantic-ai/pull/3856 we're adding a `ModelSettings.truncate: bool` setting as well; we should support there here too.

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
```diff
@@ -0,0 +1,662 @@
...
+    """The truncation strategy to use (Cohere and Nova models).
+
+    - `'NONE'`: Raise an error if input exceeds max tokens.
+    - `'START'`: Truncate the start of the input text.
+    - `'END'`: Truncate the end of the input text.
+
+    For Cohere: optional (defaults to NONE).
+    For Nova: defaults to END.
```
DouweM: In line with the other models, we should always default to `NONE`

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
```diff
@@ -0,0 +1,662 @@
...
+    """
+
+    # ==================== Amazon Nova Settings ====================
+
+    bedrock_embedding_purpose: Literal[
```
DouweM: Should this also be determined automatically from `input_type=query|document`?

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
```diff
@@ -0,0 +1,662 @@
...
+        return response_body, input_tokens
+
+    async def max_input_tokens(self) -> int | None:
+        """Get the maximum number of tokens that can be input to the model."""
+        # Normalize model name by removing version suffix
+        normalized = self._normalize_model_name(self._model_name)
+        return _MAX_INPUT_TOKENS.get(normalized)
+
+    @staticmethod
+    def _remove_inference_geo_prefix(model_name: str) -> str:
```
DouweM: We already have logic like this somewhere, let's not duplicate it

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
```diff
@@ -0,0 +1,662 @@
...
+
+    @classmethod
+    def _normalize_model_name(cls, model_name: str) -> str:
+        """Normalize model name by removing regional prefix and version suffix."""
+        model_name = cls._remove_inference_geo_prefix(model_name)
+
+        # Remove version suffix like :0
+        if ':' in model_name:
+            return model_name.rsplit(':', 1)[0]
+        return model_name
```
DouweM: We have similar logic in `BedrockProvider.model_profile`, let's reuse it (possibly by splitting it into a helper function)

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
```diff
@@ -0,0 +1,662 @@
...
+                    body=json.dumps(body),
+                    contentType='application/json',
+                    accept='application/json',
+                )
+            )
+        except ClientError as e:
+            status_code = e.response.get('ResponseMetadata', {}).get('HTTPStatusCode')
+            if isinstance(status_code, int):
+                raise ModelHTTPError(status_code=status_code, model_name=self.model_name, body=e.response) from e
+            from pydantic_ai.exceptions import ModelAPIError
```
DouweM: Imports only at the top please

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
```diff
@@ -0,0 +1,662 @@
...
+
+            settings: Model-specific [`EmbeddingSettings`][pydantic_ai.embeddings.EmbeddingSettings]
+                to use as defaults for this model.
+        """
+        self._model_name = model_name
+
+        if isinstance(provider, str):
+            provider = infer_provider(provider)
+        self._provider = provider
+        self.client = cast('BedrockRuntimeClient', provider.client)
```
DouweM, line 542: Why do we need the cast? Will this not work with `BaseClient`? Should we change the generic param on the `BedrockProvider`?

docs/embeddings.md
DouweM: Edit: I see now that we specifically implement support these 3 families. So then it does make sense listing them, but I'd rather use subheadings so they show up in the ToC sidebar, and instead of one "Bedrock-Specific Settings", I think each model family section should have list its own settings, as they're model family specific right?

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
```diff
@@ -0,0 +1,662 @@
...
+    """Whether to normalize the embedding vectors (Amazon Titan only).
+
+    Normalized vectors can be used directly for similarity calculations.
+    If not specified, model default is used.
+    """
+
+    # ==================== Cohere Settings ====================
+
+    bedrock_input_type: Literal['search_document', 'search_query', 'classification', 'clustering']
+    """The Cohere-specific input type for the embedding (Cohere models only).
```
DouweM: Since these settings are specific to model families, let's include the model family in the setting name as well.

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
```diff
@@ -0,0 +1,662 @@
...
+    def supports_batch(self) -> bool:
+        return False  # Titan only supports single text per request
+
+    def prepare_request(
+        self,
+        texts: list[str],
+        input_type: EmbedInputType,
+        settings: BedrockEmbeddingSettings,
+    ) -> dict[str, Any]:
+        # Titan only supports single text, caller must handle batching
```
DouweM: Then we should assert that len == 1, right

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
```diff
@@ -0,0 +1,662 @@
...
+        response_body: dict[str, Any],
+        inputs: list[str],
+        input_type: EmbedInputType,
+        model_name: str,
+        provider_name: str,
+        input_tokens: int,
+    ) -> EmbeddingResult:
+        # Cohere returns embeddings in different formats based on embedding_types
+        # Default is float embeddings (when embedding_types not specified)
+        # When embedding_types is specified, returns a dict with keys like 'float', 'int8', etc.
```
DouweM: Is this relevant? We don't let the user specify `embeddings_types` right?

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
```diff
@@ -0,0 +1,662 @@
...
+        self,
+        texts: list[str],
+        input_type: EmbedInputType,
+        settings: BedrockEmbeddingSettings,
+    ) -> dict[str, Any]:
+        # Nova uses a task-based format for embeddings
+        # Only supports single text per request, caller must handle batching
+
+        # Get truncation mode - Nova requires this field
+        # Nova accepts: START, END, NONE (default: END to avoid errors on long text)
```
DouweM: - Should be NONE by default for consistency with `CohereEmbeddingModel`'s default.
- As mentioned above, should support the new `settings['truncate']: bool` as well

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
```diff
@@ -0,0 +1,662 @@
...
+        }
+
+        # Nova requires embeddingDimension (default: 1024)
+        dimensions = settings.get('dimensions', 1024)
+
+        # Nova requires embeddingPurpose - default based on input_type
+        # - queries default to GENERIC_RETRIEVAL (optimized for search)
+        # - documents default to GENERIC_INDEX (optimized for indexing)
+        default_purpose = 'GENERIC_RETRIEVAL' if input_type == 'query' else 'GENERIC_INDEX'
+        embedding_purpose = settings.get('bedrock_embedding_purpose', default_purpose)
```
DouweM: Setting docstring should be updated to specify how this maps to embed_query / embed_document, as we have on the cohere input type setting

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
```diff
@@ -0,0 +1,662 @@
...
+        # Nova uses a task-based format for embeddings
+        # Only supports single text per request, caller must handle batching
+
+        # Get truncation mode - Nova requires this field
+        # Nova accepts: START, END, NONE (default: END to avoid errors on long text)
+        truncate = settings.get('bedrock_truncate', 'END')
+
+        # Build text params - 'value' and 'truncationMode' are required by Nova
+        text_params: dict[str, Any] = {
+            'value': texts[0],
```
DouweM: assert len == 1

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
```diff
@@ -0,0 +1,662 @@
...
+
+        # Build singleEmbeddingParams - all fields are required by Nova
+        single_embedding_params: dict[str, Any] = {
+            'text': text_params,
+            'embeddingDimension': dimensions,
+            'embeddingPurpose': embedding_purpose,
+        }
+
+        body: dict[str, Any] = {
+            'taskType': 'SINGLE_EMBEDDING',
```
DouweM, line 432: This suggests there's also a batch embedding task type?

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
```diff
@@ -0,0 +1,662 @@
...
+
+        # Extract the embedding vector from the first item
+        embedding = embeddings_list[0].get('embedding')
+        if embedding is None:
+            raise UnexpectedModelBehavior(
+                'The Nova Bedrock embeddings response did not have an `embedding` field in the first item',
+                str(response_body),
+            )
+
+        return EmbeddingResult(
```
DouweM: Lets simplify this method by returning only the embeddings + tokens + in some cases the response ID, and then build the `EmbeddingResult` where we call `parse_response`. Right now when we combine single-document runs, we deconstruct the individual `EmbeddingResult`s and turn them into one combined one anyway, so let's skip the intermediate `EmbeddingResult`. That also means we no longer have to pass inputs, model_name etc into this method anymore.

tests/conftest.py
```diff
@@ -315,6 +315,19 @@ def mock_vcr_aiohttp_content(mocker: MockerFixture):
     mocker.patch('vcr.stubs.aiohttp_stubs.MockStream.set_exception', return_value=None)
 
 
+@pytest.fixture(autouse=True)
+def mock_vcr_botocore_content_length(mocker: MockerFixture):
+    # VCR doesn't properly handle botocore's content-length verification when replaying responses.
+    # This causes IncompleteReadError when the recorded response body length doesn't match the Content-Length header.
+    # This happens because VCR decodes compressed responses but doesn't update the Content-Length header.
```
DouweM: Should we instead fix this by updating the `Content-Length` in `tests/json_body_serializer.py`'s `serialize` method?

tests/test_embeddings.py
```diff
@@ -327,6 +331,269 @@ async def test_embed_error(self, co_api_key: str):
...
+        """Test Titan with minimal request body (just inputText)."""
+        model = BedrockEmbeddingModel('amazon.titan-embed-text-v2:0', provider=bedrock_provider)
+        embedder = Embedder(model)
+        result = await embedder.embed_query('Hello, world!')
+        assert result == snapshot(
+            EmbeddingResult(
+                embeddings=IsList(IsList(IsFloat(), length=1024), length=1),
+                inputs=['Hello, world!'],
+                input_type='query',
+                usage=RequestUsage(input_tokens=IsInt()),
```
DouweM: For usage, can we see the actual value rather than `IsInt`? I want to make sure it's not 0

tests/test_embeddings.py
```diff
@@ -599,6 +866,235 @@ def test_infer_model_error():
         infer_embedding_model('nonexistent')
 
 
+@pytest.mark.skipif(not bedrock_imports_successful(), reason='Bedrock not installed')
+class TestBedrockHandlers:
```
DouweM: I don't think we need to unit test this, assuming all the paths are covered by the integration test above?

tests/test_embeddings.py
```diff
@@ -599,6 +866,235 @@ def test_infer_model_error():
...
+                model_name='amazon.nova-2-multimodal-embeddings-v1:0',
+                provider_name='bedrock',
+                input_tokens=5,
+            )
+
+    async def test_invoke_model_client_error_with_status_code(self, bedrock_provider: BedrockProvider):
+        """Test error handling when ClientError is raised with HTTP status code."""
+        from botocore.exceptions import ClientError
+
+        from pydantic_ai.exceptions import ModelHTTPError
```
DouweM: All imports at the top please

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
bitnahian: We do something similar in https://github.com/pydantic/pydantic-ai/blob/3772a1e41344bd159dd359cb46d5aaaca925db01/pydantic_ai_slim/pydantic_ai/models/bedrock.py#L276 

I imagine changing it in `BedrockProvider` might work. I'll try and fix this after all the other changes.

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
bitnahian: Right now, the `extra_body` could be used to override this behaviour. We can restrict that but it'll require validating the `extra_body` settings. Right now, I'm allowing user's to use `extra_body` to override a lot of things but I'll probably change its usage in Bedrock it to keep the client opinionated in terms of embedding parameters.

bitnahian:
@DouweM Are we keeping the clients opinionated to work as text embedding models only? For instance, Nova has a tonne of [multi-modal specific parameters](https://docs.aws.amazon.com/nova/latest/userguide/embeddings-schema.html) and if we're restricting to text only, we shouldn't surface them.

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
bitnahian: [Allowed values: Must be "SINGLE_EMBEDDING" for synchronous calls.](https://docs.aws.amazon.com/nova/latest/userguide/embeddings-schema.html#:~:text=Allowed%20values%3A%20Must%20be%20%22SINGLE_EMBEDDING%22%20for%20synchronous%20calls.)

From my understanding, we're using the synchronous BaseClient.`invoke_model` method so this is the only allowed `taskType`.

bitnahian: To clarify further.. there is an async method with which you can pass longer contexts and use `SEGMENTED_EMBEDDING` as a task type, but that implementation only returns an invocation id.. and not the actual embeddings. So it didn't seem to fit here.

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
bitnahian: It does. But there are more than 2 options available. I've defaulted to a sensible value based on `query|document` but also allowing users to override this to something specific. 

I do find it icky that there are multi-modal purposes e.g. audio, image and video retrieval options as well. Perhaps we can remove those options from the literal?

bitnahian: Will restrict to `GENERIC_INDEX`, `GENERIC_RETRIEVAL` , `TEXT_RETRIEVAL`, `CLASSIFICATION` and `CLUSTERING` as allowed values.

bitnahian:
Hey @DouweM, just bumping this for re-review. Thanks!

docs/embeddings.md
```diff
@@ -342,6 +342,156 @@ embedder = Embedder(
...
+Bedrock supports cross-region inference using geographic prefixes like `us.`, `eu.`, or `apac.`:
+
+```python {title="bedrock_regional.py"}
+from pydantic_ai import Embedder
+
+embedder = Embedder('bedrock:us.amazon.titan-embed-text-v2:0')
+```
+
+!!! note "Token Counting"
+    Bedrock embedding models do not support the `count_tokens()` method because AWS Bedrock's token counting API only works with text generation models (Claude, Llama, etc.), not embedding models. Calling `count_tokens()` will raise `NotImplementedError`.
```
DouweM: This doesn't seem like it belongs in this section

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
```diff
@@ -0,0 +1,574 @@
...
+        )
+        ```
+    """
+
+    # ALL FIELDS MUST BE `bedrock_` PREFIXED SO YOU CAN MERGE THEM WITH OTHER MODELS.
+
+    # ==================== Amazon Titan Settings ====================
+
+    bedrock_titan_normalize: bool
+    """Whether to normalize embedding vectors for Titan models.
```
DouweM, line 107: Can we mention the default, if there is one? That goes for all of these.

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
```diff
@@ -0,0 +1,574 @@
...
+            body['max_tokens'] = max_tokens
+
+        if dimensions := settings.get('dimensions'):
+            body['output_dimension'] = dimensions
+
+        # Model-specific truncate takes precedence, then base truncate setting
+        if truncate := settings.get('bedrock_cohere_truncate'):
+            body['truncate'] = truncate
+        elif settings.get('truncate'):
+            body['truncate'] = 'END'
```
DouweM, line 355: I'd like to explicitly set this to `NONE` in an else branch just so we don't depend on the API default.

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
```diff
@@ -0,0 +1,574 @@
...
+        if truncate := settings.get('bedrock_nova_truncate'):
+            pass  # Use the model-specific setting
+        elif settings.get('truncate'):
+            truncate = 'END'
+        else:
+            truncate = 'NONE'
+
+        # Build text params based on input type
+        # Nova supports both direct text values and S3 URIs
+        if text.startswith('s3://'):
```
DouweM: This is a little too magic to me, I'd prefer to not support this for now.

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
DouweM: Ok then this is good enough for now I think. Note that we could use an async API endpoint and then just check the status in a loop, if that's faster than running multiple API requests in parallel.

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
```diff
@@ -0,0 +1,574 @@
...
+        input_type: EmbedInputType,
+        settings: BedrockEmbeddingSettings,
+    ) -> EmbeddingResult:
+        """Embed inputs one at a time and combine results."""
+        all_embeddings: list[Sequence[float]] = []
+        total_input_tokens = 0
+
+        for text in inputs:
+            body = self._handler.prepare_request([text], input_type, settings)
+            response, input_tokens = await self._invoke_model(body)
```
DouweM: Can we use `asyncio.gather` to do these requests in parallel?

pydantic_ai_slim/pydantic_ai/embeddings/settings.py
```diff
@@ -23,6 +23,27 @@ class EmbeddingSettings(TypedDict, total=False):
     * Cohere
     * Google
     * Sentence Transformers
+    * Bedrock
+    """
+
+    truncate: bool
```
DouweM, line 30: The Voyage AI PR just got merged, so please merge main and resolve any merge conflicts.

tests/test_embeddings.py
```diff
@@ -327,6 +331,433 @@ async def test_embed_error(self, co_api_key: str):
...
+                provider_name='bedrock',
+                timestamp=IsDatetime(),
+                usage=RequestUsage(input_tokens=4),
+                provider_response_id=IsStr(),
+            )
+        )
+
+    async def test_cohere_with_base_truncate(self, bedrock_provider: BedrockProvider):
+        """Test Cohere with base truncate=True setting (maps to END)."""
+        from pydantic_ai.embeddings.bedrock import BedrockEmbeddingSettings
```
DouweM: Move imports to the top please

tests/test_embeddings.py
```diff
@@ -327,6 +331,433 @@ async def test_embed_error(self, co_api_key: str):
...
+        """Test BedrockEmbeddingModel can be created with string provider."""
+        with patch('pydantic_ai.embeddings.bedrock.infer_provider', return_value=bedrock_provider) as mock_infer:
+            model = BedrockEmbeddingModel('amazon.titan-embed-text-v2:0', provider='bedrock')
+            mock_infer.assert_called_once_with('bedrock')
+            assert model.model_name == 'amazon.titan-embed-text-v2:0'
+
+    def test_normalize_model_name_variants(self):
+        """Test _normalize_model_name handles various model name formats."""
+        # Test with version suffix
+        result = BedrockEmbeddingModel._normalize_model_name('amazon.titan-embed-text-v2:0')  # pyright: ignore[reportPrivateUsage]
```
DouweM: I don't want to be unit testing this method like this; would rather have an integration test that uses one of the model names that need normalization, and wouldn't have worked if we didn't do the normalization. So that integration test passing proves that we are normalizing as we should.

tests/test_embeddings.py
DouweM: Still relevant.

bitnahian: My bad.. somehow missed this.

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
bitnahian: It also needs an s3 output config that we'd need to read from. I think this is beyond the scope of this ticket probably. Can support if needed in a different PR. 


image

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
bitnahian: Hmm.. I've been thinking about this. Do you think it makes sense to let users handle their own parallelism? We can probably allow up to a max batch size for each model as the only API interface. I feel like each model has its own caveats.. but I'll do a first pass with async gather implementation.

stuaxo:
Just want to say thanks @bitnahian and everyone here for looking at this.

bitnahian:
Hey @DouweM, I've addressed most of the comments so far. I haven't removed the `cast('BedrockRuntimeClient', provider.client)` since it touches a couple of files and this is already quite a big PR. 
  
  CI is failing on main  for **CI / test on 3.13 (all-extras) (push)** and it's also failing for me after merging main. 
  
  I've run coverage and seems to be hitting a 100% atm. 
  
  Keen to get your review again. Thanks!! 
  
    ============================= test session starts ==============================                                                                                      
     platform darwin -- Python 3.12.12, pytest-9.0.2, pluggy-1.6.0                                                                                                         
     rootdir: ----                                                                                                                     
     configfile: pyproject.toml                                                                                                                                            
     plugins: mock-3.15.1, recording-0.13.4, langsmith-0.6.4, pretty-1.3.0, xdist-3.8.0, devtools-0.12.2, anyio-4.8.0, inline-snapshot-0.31.1, examples-0.0.18,            
     logfire-4.16.0                                                                                                                                                        
     collected 42 items                                                                                                                                                    
                                                                                                                                                                           
     tests/test_embeddings.py ..........................................      [100%]                                                                                       
     Results (6.53s):                                                                                                                                                      
             42 passed                                                                                                                                                     
     Combined data file .coverage/.coverage.LM-GNP4V9G6NC.24805.XzczkvBx                                                                                                   
     Name    Stmts   Miss Branch BrPart    Cover   Missing                                                                                                                 
     -----------------------------------------------------                                                                                                                 
     TOTAL     204      0     38      0  100.00%

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
```diff
@@ -0,0 +1,695 @@
...
+
+    **Applies to:** `amazon.titan-embed-text-v1`, `amazon.titan-embed-text-v2:0`,
+    `amazon.nova-2-multimodal-embeddings-v1:0`
+
+    When embedding multiple texts with models that only support single-text requests,
+    this controls how many requests run in parallel. Defaults to 5.
+    """
+
+
+class BedrockEmbeddingHandler(ABC):
```
DouweM: Let's make this (and the subclasses) private as we don't want people importing them and using them directly

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
```diff
@@ -0,0 +1,695 @@
...
+
+    def __init__(self, model_name: str):
+        """Initialize the handler with the model name.
+
+        Args:
+            model_name: The normalized model name (e.g., 'amazon.titan-embed-text-v2:0').
+        """
+        self.model_name = model_name
+        config = _MODEL_CONFIG.get(model_name)
+        assert config is not None, f'No config found for model: {model_name}'
```
DouweM: If at all possible, I'd like to automatically support new models (in families that we have handlers) for, instead of erroring on unknown names. Would that be possible or is there nothing we can do without a model config? I think some model config are more like model _family_ config, and it's OK if e.g. `max_input_tokens` returns `None` when unknown.

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
```diff
@@ -0,0 +1,695 @@
...
+            # Titan v1 doesn't support dimensions or normalize parameters
+            if dimensions is not None:
+                warnings.warn(
+                    f'The `dimensions` setting is not supported by {self.model_name} and will be ignored. '
+                    'Only Titan v2 models support custom dimensions.',
+                    UserWarning,
+                )
+            if normalize is not None:
+                warnings.warn(
+                    f'The `bedrock_titan_normalize` setting is not supported by {self.model_name} and will be ignored. '
```
DouweM: No need to warn as long as the docstring already specifies a given model is not supported. Silently ignoring is consistent with how we handle other model settings.

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
```diff
@@ -0,0 +1,695 @@
...
+                    'Only Titan v2 models support the normalize parameter.',
+                    UserWarning,
+                )
+        else:
+            # Titan v2: Apply dimensions (default from config)
+            assert self.config.default_dimensions is not None, 'Titan v2 must have default_dimensions in config'
+            body['dimensions'] = dimensions if dimensions is not None else self.config.default_dimensions
```
DouweM: It's required to pass these in?

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
```diff
@@ -0,0 +1,695 @@
...
+                )
+            if normalize is not None:
+                warnings.warn(
+                    f'The `bedrock_titan_normalize` setting is not supported by {self.model_name} and will be ignored. '
+                    'Only Titan v2 models support the normalize parameter.',
+                    UserWarning,
+                )
+        else:
+            # Titan v2: Apply dimensions (default from config)
+            assert self.config.default_dimensions is not None, 'Titan v2 must have default_dimensions in config'
```
DouweM: I'd like to have fewer assertions, more "automatic self healing" and if we must have errors that the user can resolve (including by using a different model), make them `UserError`

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
DouweM: Agreed, out of scope for now

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
```diff
@@ -0,0 +1,695 @@
...
+
+def _get_handler_for_model(model_name: str) -> BedrockEmbeddingHandler:
+    """Get the appropriate handler for a Bedrock embedding model."""
+    normalized_name = remove_bedrock_geo_prefix(model_name)
+
+    if normalized_name.startswith('amazon.titan-embed'):
+        return TitanEmbeddingHandler(normalized_name)
+    elif normalized_name.startswith('cohere.embed'):
+        return CohereEmbeddingHandler(normalized_name)
+    elif normalized_name.startswith('amazon.nova'):
```
DouweM: Can we make this a dict of prefixes -> handler classes?

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
```diff
@@ -0,0 +1,695 @@
...
+
+    supports_batch: bool = False
+    """Whether the model supports batch embedding in a single request."""
+
+    default_dimensions: int | None = None
+    """Default embedding dimensions. None means fixed dimensions (not configurable)."""
+
+
+# Model configuration lookup (keys are normalized model names as returned by remove_bedrock_geo_prefix)
+_MODEL_CONFIG: dict[str, BedrockModelConfig] = {
```
DouweM: Constants before implementation please, i.e. if possible move this head of the handlers

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
bitnahian: Even within families, there are slight differences in the parameters in the API spec. I don't think model families are generalised enough, hence the different configs for each different family. It could be possible potentially to separate per family version, e.g. cohere-v3 multilingual and english can have the same config, but cohere v3 and v4 will almost certainly need separate configs. 

WDYT? Would you like me to break it up per family version? @DouweM

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
bitnahian: From your previous comment, I stopped relying on API defaults. So just coercing default values directly in the request.

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
bitnahian: These asserts are mainly there for type checking. Maybe I can move these into a `@property` method which should take care of the non-None type checks.

pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
bitnahian: We can probably relax the model configs a little bit if we start relying on API defaults instead of coercing them.

bitnahian:
I've removed the problematic _BedrockModelConfig class. Only kept a dictionary of `_MAX_INPUT_TOKENS` for the `max_token_count()` implementation.

- Resorting to API defaults for things like `max_tokens` and `dimensions` so as not to have to maintain a config.
- Kept a default request parameter for `truncate` only set to `NONE` explicitly as per prior comment.
- All helper classes are now private. The handlers are assigned based on model id **prefixes** and not the model ids directly. 
- Handlers are using a specific `_extract_version` method and a `match-case` style statement to silently ignore non-version specific settings.

Theoretically, if a `cohere.embed-v5:0` model came out tomorrow, it should at least make the API request. 

Thanks for reviewing so far @DouweM .. hopefully closer now.

bitnahian:
Heya @DouweM just pinging again in case this fell through the cracks

dsfaccini:

```bash
git pull

deactivate

rm -rf .venv

make install
```

DouweM:
@bitnahian Thanks Nahian, great work!

PR #4009, opened 2026-01-15
Have `OpenAIChatModel` send back reasoning content via field it's received in

### LINKED ISSUE #3867: Send back reasoning content via field if it's received in field
### Initial Checks

### Description

Currently, reasoning content is sent back by tags by default even if the reasoning content is received from field. I don't think this is a good default.

```Python

```

### Logfire Trace

_No response_

- Python: 3.13.*
- Pydantic AI: 1.39.0
- LLM provider SDK: openai

safina57:
- Closes #3867

### Pre-Review Checklist

### Pre-Merge Checklist

### Changes
@DouweM @xcpky, This is my attempt to resolve the issue where reasoning content is sent back in tags by default even when received from custom fields. I've implemented auto-detection to determine the appropriate send-back format based on where the thinking content originated.

Currently, `openai_chat_send_back_thinking_parts` defaults to `'tags'`, which causes thinking content to be wrapped in tags and sent in the message content field, even when the content was originally received in a custom field like `reasoning` or `reasoning_content`. This creates issues for providers like DeepSeek, MoonshotAI, and OpenRouter that expect thinking content to be sent back in the same custom field format for optimal performance and caching.

I've implemented an `'auto'` mode for `openai_chat_send_back_thinking_parts` that automatically detects the appropriate send-back format based on the thinking content's origin:

**Auto-detection logic:**

- If `ThinkingPart.id` exists and is not `'content'` (which indicates tag-based extraction) and `ThinkingPart.provider_name` matches the current model's system name, the thinking is sent back in the custom field specified by `ThinkingPart.id`

- Otherwise, thinking is sent back using tags (wrapped in `<think>...</think>` or custom tags)

**Default behavior change:**

- Changed `openai_chat_send_back_thinking_parts` default from `'tags'` to `'auto'`
- Removed Ollama's explicit `'tags'` override, allowing it to use the new auto-detection

Added tests in `tests/models/test_openai.py`:

- `test_openai_auto_mode_reasoning_field_sends_back_in_field`: Verifies thinking from `reasoning_content` field is sent back in the same field
- `test_openai_auto_mode_reasoning_field_different_provider_uses_tags`: Verifies fallback to tags when provider names don't match
- `test_openai_auto_mode_tags_sends_back_in_tags`: Verifies tag-based thinking (id='content') is sent back in tags
- `test_openai_auto_mode_multiple_thinking_parts_same_field`: Verifies multiple thinking parts are grouped correctly by field name
  

Updated `docs/thinking.md` with a new section "Sending thinking content back to models" that explains:

- The four modes: `'auto'` (default), `'tags'`, `'field'`, and `False`
- How auto-detection works
- Example configuration for overriding auto-detection
  

I'm happy to make any adjustments based on your feedback.

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/profiles/openai.py
+++ b/pydantic_ai_slim/pydantic_ai/profiles/openai.py
@@ -51,7 +51,7 @@ class OpenAIModelProfile(ModelProfile):
     This can be:
     * `'auto'` (default): Automatically detects how to send thinking content. If thinking was received in a custom field
     (tracked via `ThinkingPart.id` and `ThinkingPart.provider_name`), it's sent back in that same field. Otherwise,
-    it's sent using tags. only the `reasoning` and `reasoning_content` fields are checked by
+    it's sent using tags. Only the `reasoning` and `reasoning_content` fields are checked by
     default when receiving responses. If your provider uses a different field name, you must explicitly set
     `openai_chat_thinking_field` to that field name.
     * `'tags'`: The thinking content is included in the main `content` field, enclosed within thinking tags as


docs/thinking.md
```diff
@@ -13,7 +13,38 @@ You can customize the tags using the [`thinking_tags`][pydantic_ai.profiles.Mode
 
 Some [OpenAI-compatible model providers](models/openai.md#openai-compatible-models) might also support native thinking parts that are not delimited by tags. Instead, they are sent and received as separate, custom fields in the API. Typically, if you are calling the model via the `<provider>:<model>` shorthand, Pydantic AI handles it for you. Nonetheless, you can still configure the fields with [`openai_chat_thinking_field`][pydantic_ai.profiles.openai.OpenAIModelProfile.openai_chat_thinking_field].
 
-If your provider recommends to send back these custom fields not changed, for caching or interleaved thinking benefits, you can also achieve this with [`openai_chat_send_back_thinking_parts`][pydantic_ai.profiles.openai.OpenAIModelProfile.openai_chat_send_back_thinking_parts].
+### Sending thinking content back to models
```
DouweM: I don't really want to encourage people to tweak profile settings directly, so I'd prefer to keep just the old docs and explain the options in the profile setting docstring only. The user should be able to assume that we've set these as appropriate for all models, and they shouldn't need to think about them at all.

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -856,7 +856,9 @@ class _MapModelResponseContext:
         _model: OpenAIChatModel
 
         texts: list[str] = field(default_factory=list)
-        thinkings: list[str] = field(default_factory=list)
+        thinkings: list[str | tuple[str, str]] = field(
+            default_factory=list
+        )  # str for 'field' mode, tuple[field_name, content] for 'auto' mode
```
DouweM: I'd rather have just `tuple`s, is that an option?

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -888,10 +890,23 @@ def _into_message_param(self) -> chat.ChatCompletionAssistantMessageParam:
...
+                # Auto mode: thinkings are stored as tuples of (field_name, content)
+                # Group by field name and set each field
+                field_contents: dict[str, list[str]] = {}
+                for item in self.thinkings:
+                    if isinstance(item, tuple):
+                        field_name, content = item
+                        field_contents.setdefault(field_name, []).append(content)
+                    else:
+                        # Fallback for non-tuple items
+                        field_contents.setdefault('reasoning', []).append(item)  # pragma: no cover
```
DouweM: I don't think we should need to hard-code the `reasoning` field

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -888,10 +890,23 @@ def _into_message_param(self) -> chat.ChatCompletionAssistantMessageParam:
...
+                for field_name, contents in field_contents.items():
+                    message_param[field_name] = '\n\n'.join(contents)
+            elif profile.openai_chat_send_back_thinking_parts == 'field' and self.thinkings:
```
DouweM: The if and elif repeat `and self.thinkings`, so let's pull it up into its own `if`

pydantic_ai_slim/pydantic_ai/profiles/openai.py
```diff
@@ -92,6 +95,7 @@ def __post_init__(self):  # pragma: no cover
                 'If `openai_chat_send_back_thinking_parts` is "field", '
                 '`openai_chat_thinking_field` must be set to a non-None value.'
             )
+        # Note: 'auto' mode doesn't require openai_chat_thinking_field since it detects dynamically
```
DouweM, line 133: Would be worth specifying (in the docstring and/or here) that it will only look for `reasoning` and `reasoning_content` fields by default; if the reasoning is in another field, it does need to be configured specifically

tests/models/test_openai.py
```diff
@@ -3733,6 +3733,133 @@ async def test_openai_reasoning_in_thinking_tags(allow_model_requests: None):
...
+    """Test that auto mode detects reasoning from field and sends back in field."""
+    c = completion_message(
+        ChatCompletionMessage.model_construct(content='response', reasoning_content='reasoning', role='assistant')
+    )
+    m = OpenAIChatModel(
+        'foobar',
+        provider=OpenAIProvider(openai_client=MockOpenAI.create_mock(c)),
+        profile=OpenAIModelProfile(
+            openai_chat_thinking_field='reasoning_content',
+            openai_chat_send_back_thinking_parts='auto',  # Auto mode
```
DouweM, line 3786: We should test what happens if:
- the mode is set to auto, but `thinking_field` is not set
- the field is set to `reasoning_content` and the mode is set to `auto`, but the reasoning actually comes in e.g `reasoning` (which is checked by default)

tests/models/test_openai.py
```diff
@@ -3733,6 +3733,133 @@ async def test_openai_reasoning_in_thinking_tags(allow_model_requests: None):
...
+    assert m._map_model_response(resp) == snapshot(  # type: ignore[reportPrivateUsage]
+        {'role': 'assistant', 'reasoning_content': 'reasoning', 'content': 'response'}
+    )
+
+
+async def test_openai_auto_mode_reasoning_field_different_provider_uses_tags(allow_model_requests: None):
+    """Test that auto mode falls back to tags when provider_name doesn't match."""
+    # This test verifies behavior by checking that when thinking comes from a different provider,
+    # auto mode falls back to tags. We'll test this by creating a full message exchange.
+    from pydantic_ai.messages import ModelRequest, ModelResponse, ThinkingPart, UserPromptPart
```
DouweM: No inline imports please, only at the top

tests/models/test_openai.py
```diff
@@ -3733,6 +3733,133 @@ async def test_openai_reasoning_in_thinking_tags(allow_model_requests: None):
...
+        ),
+    ]
+
+    # Make a request - the thinking should be sent back in tags since provider doesn't match
+    settings = ModelSettings()
+    params = ModelRequestParameters()
+    await m.request(messages=messages, model_settings=settings, model_request_parameters=params)
+
+    # The test passes if no exception is raised and the mock receives the correct format
+    # (with tags, not in reasoning_content field)
```
DouweM: this confuses me, shouldn't we have `assert m._map_model_response(resp)` here?

tests/models/test_openai.py
```diff
@@ -3733,6 +3733,133 @@ async def test_openai_reasoning_in_thinking_tags(allow_model_requests: None):
...
+            openai_chat_send_back_thinking_parts='auto',
+        ),
+    )
+
+    # First request gets thinking in field
+    settings = ModelSettings()
+    params = ModelRequestParameters()
+    resp1 = await m.request(messages=[], model_settings=settings, model_request_parameters=params)
+
+    # Verify the thinking was extracted with the correct id
```
DouweM: please use a snapshot here and below

pydantic_ai_slim/pydantic_ai/providers/ollama.py
```diff
@@ -65,7 +65,6 @@ def model_profile(self, model_name: str) -> ModelProfile | None:
         return OpenAIModelProfile(
             json_schema_transformer=OpenAIJsonSchemaTransformer,
             openai_chat_thinking_field='reasoning',
-            openai_chat_send_back_thinking_parts='tags',
```
DouweM, line 68: Can you find out from the blame or something why we had this explicitly set in the first place?

safina57: I followed the change and found that the PR was merged, but it didn’t specifically explain why reasoning was forced to be sent using tags.
I tested both formats locally with DeepSeek, and both worked.
The original `tags` setting was likely added defensively

DouweM:
@safina57 Please have a look at the missing test coverage

tests/models/test_openai.py
```diff
@@ -3733,6 +3733,201 @@ async def test_openai_reasoning_in_thinking_tags(allow_model_requests: None):
...
+
+    thinking_parts = [p for p in resp1.parts if isinstance(p, ThinkingPart)]
+    assert len(thinking_parts) == 1
+    assert thinking_parts[0].id == 'reasoning'
+    assert thinking_parts[0].provider_name == 'openai'
+
+    mapped = m._map_model_response(resp1)  # type: ignore[reportPrivateUsage]
+    assert mapped == snapshot({'role': 'assistant', 'reasoning': 'first thought', 'content': 'response'})
+
+
```
SakshamKapoor2911: This test covers the missing branch 899->903 in _into_message_param method. The CI is failing because branch coverage for elif profile.openai_chat_send_back_thinking_parts == 'field': (line 899) isn't being exercised with multiple thinking parts that populate self.thinkings.

This test:
- Uses openai_chat_send_back_thinking_parts='field'
- Creates multiple ThinkingPart instances
- Verifies they're correctly grouped and joined with '\n\n'

This should bring coverage to 100% and fix the CI failure.

```suggestion
async def test_openai_field_mode_multiple_thinking_parts(allow_model_requests: None):
    """Test that field mode correctly groups multiple thinking parts."""
    c1 = completion_message(ChatCompletionMessage.model_construct(content='response', role='assistant'))
    m = OpenAIChatModel(
        'foobar',
        provider=OpenAIProvider(openai_client=MockOpenAI.create_mock(c1)),
        profile=OpenAIModelProfile(
            openai_chat_thinking_field='reasoning_content',
            openai_chat_send_back_thinking_parts='field',
        ),
    )

    # Create a response with multiple thinking parts manually
    messages = [
        ModelRequest(parts=[UserPromptPart(content='question')]),
        ModelResponse(
            parts=[
                ThinkingPart(
                    content='first thought',
                    id='reasoning_content',
                    provider_name='openai',
                ),
                ThinkingPart(
                    content='second thought',
                    id='reasoning_content',
                    provider_name='openai',
                ),
                TextPart(content='response'),
            ]
        ),
    ]

    mapped = m._map_model_response(messages[1])  # type: ignore[reportPrivateUsage]
    assert mapped == snapshot(
        {
            'role': 'assistant',
            'reasoning_content': 'first thought\n\nsecond thought',
            'content': 'response',
        }
    )
```

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -896,7 +896,7 @@ def _into_message_param(self) -> chat.ChatCompletionAssistantMessageParam:
                         field_contents.setdefault(field_name, []).append(content)
                     for field_name, contents in field_contents.items():
                         message_param[field_name] = '\n\n'.join(contents)
-                elif profile.openai_chat_send_back_thinking_parts == 'field':
+                elif profile.openai_chat_send_back_thinking_parts == 'field':  # pragma: no branch
```
SakshamKapoor2911: The CI is still failing because of the branch at line 899 (899->903). Since self.thinkings is only populated when the mode is 'auto' or 'field', if the mode is not 'auto', it must be 'field'. This makes the False path of the elif completely unreachable.

Here are two ways you can fix this:

Option 1: Logically refactor to else
By changing the elif to an else, you remove the branch entirely. This is the cleanest fix as it doesn't require configuration changes.
`else:
                    # mode must be 'field' here, and validated in OpenAIModelProfile.__post_init__
                    field = cast(str, profile.openai_chat_thinking_field)
                    message_param[field] = '\n\n'.join(content for _, content in self.thinkings)`

Option 2: Update Coverage Configuration
If you prefer keeping the elif and the pragma, you must add 'pragma: no branch' to the exclude_lines in pyproject.toml, otherwise the comment does nothing. I found out that 'pragma: no branch' is missing from that file and hence does not exclude coverage. 

`[tool.coverage.report]
exclude_lines = [
    'pragma: no cover', 
    'pragma: lax no cover',
    'pragma: no branch',  # Add this
    'raise NotImplementedError',`

safina57:
@DouweM that branch cannot practically be taken and the type checker doesn't know about the runtime validation in __post_init__ , so it still sees field as potentially None
i had to add # pragma: no branch to it

dsfaccini:

```bash
git pull

deactivate

rm -rf .venv

make install
```

SakshamKapoor2911:
Hey @safina57! The CI is failing due to a pyright type error. The `thinkings` field in `_MapModelResponseContext` is typed as `list[str]` but the code now appends tuples of `(field_name, content)`.

**Fix needed in `pydantic_ai_slim/pydantic_ai/models/openai.py` line 859:**

```diff
-        thinkings: list[str] = field(default_factory=list[str])
+        thinkings: list[tuple[str, str]] = field(default_factory=list[tuple[str, str]])
```

This matches the usage in `_map_response_thinking_part` where you append tuples like `(item.id, item.content)` and `(field, item.content)`.

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -890,10 +890,18 @@ def _into_message_param(self) -> chat.ChatCompletionAssistantMessageParam:
...
+                if profile.openai_chat_send_back_thinking_parts == 'auto':
+                    # Group by field name and set each field
+                    field_contents: dict[str, list[str]] = {}
+                    for field_name, content in self.thinkings:
+                        field_contents.setdefault(field_name, []).append(content)
+                    for field_name, contents in field_contents.items():
+                        message_param[field_name] = '\n\n'.join(contents)
+                elif profile.openai_chat_send_back_thinking_parts == 'field':  # pragma: no branch
+                    # Validated in OpenAIModelProfile.__post_init__, cast for type checker
+                    field = cast(str, profile.openai_chat_thinking_field)
```
DouweM: Could we use `assert ... is not None`?

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -890,10 +890,18 @@ def _into_message_param(self) -> chat.ChatCompletionAssistantMessageParam:
...
+                    message_param[field] = '\n\n'.join(content for _, content in self.thinkings)
```
DouweM: We can dedupe a bit here but setting `field_contents = {profile.openai_chat_thinking_field: [content for _, content in self.thinkings]}` and then doing this just once:

```py
for field_name, contents in field_contents.items():
    message_param[field_name] = '\n\n'.join(contents)
```

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -918,11 +926,33 @@ def _map_response_thinking_part(self, item: ThinkingPart) -> None:
...
+                    start_tag, end_tag = self._model.profile.thinking_tags
+                    self.texts.append('\n'.join([start_tag, item.content, end_tag]))
+            elif include_method == 'tags':
                 start_tag, end_tag = self._model.profile.thinking_tags
                 self.texts.append('\n'.join([start_tag, item.content, end_tag]))
             elif include_method == 'field':
-                self.thinkings.append(item.content)
+                field = profile.openai_chat_thinking_field
+                if field:  # pragma: no branch
+                    self.thinkings.append((field, item.content))
```
DouweM: Instead of `list[tuple[str, str]]`, could this be `dict[str, list[str]`?

pydantic_ai_slim/pydantic_ai/profiles/openai.py
```diff
@@ -30,16 +30,21 @@ class OpenAIModelProfile(ModelProfile):
...
 
-    openai_chat_send_back_thinking_parts: Literal['tags', 'field', False] = 'tags'
+    openai_chat_send_back_thinking_parts: Literal['auto', 'tags', 'field', False] = 'auto'
     """Whether the model includes thinking content in requests.
 
     This can be:
-    * `'tags'` (default): The thinking content is included in the main `content` field, enclosed within thinking tags as
+    * `'auto'` (default): Automatically detects how to send thinking content. If thinking was received in a custom field
+    (tracked via `ThinkingPart.id` and `ThinkingPart.provider_name`), it's sent back in that same field. Otherwise,
+    it's sent using tags. only the `reasoning` and `reasoning_content` fields are checked by
```
DouweM:
```suggestion
    it's sent using tags. Only the `reasoning` and `reasoning_content` fields are checked by
```

pydantic_ai_slim/pydantic_ai/models/openai.py
safina57: the pipeline still fails even after adding the `assert`, I thought adding asserts in business logic was a bad practice, but even with the assert in place, the failure persisted
considered adding explicit `else` branch to handle the fallback case
```python
else:
    field_contents = {}
```
but felt like adding `no branch` would be enough

pydantic_ai_slim/pydantic_ai/models/openai.py
safina57: I changed it to a `dict`
It may only have an impact if order matters, otherwise, it should be good to go

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -958,10 +958,19 @@ def _into_message_param(self) -> chat.ChatCompletionAssistantMessageParam:
...
             # shouldn't merge multiple texts into one unless you switch models between runs:
-            if profile.openai_chat_send_back_thinking_parts == 'field' and self.thinkings:
-                field = profile.openai_chat_thinking_field
-                if field:  # pragma: no branch (handled by profile validation)
-                    message_param[field] = '\n\n'.join(self.thinkings)
+            if self.thinkings:
+                field_contents: dict[str, list[str]] = {}
+                if profile.openai_chat_send_back_thinking_parts == 'auto':
+                    field_contents = self.thinkings
+                elif profile.openai_chat_send_back_thinking_parts == 'field':  # pragma: no branch
```
DouweM: I don't think we need a separate branch for `field` here, as we already ensure that `openai_chat_thinking_field` is used in the logic where we build `self.thinkings`, right?

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/models/openai.py
safina57: your're right, each message gets its own mapper instance with a fresh thinkings dict
when it's set to `field`, all thinking parts will use the same value, the flattening is not needed 
thanks for that!

PR #4014, opened 2026-01-16
Review `instructions` and `system_prompt` docs for clarity

dsfaccini:
from https://pydanticlogfire.slack.com/archives/C083V7PMHHA/p1768498541106069?thread_ts=1768485875.411099&cid=C083V7PMHHA


dsfaccini:
image

image

### expanded

image

docs/agents.md
```diff
@@ -890,17 +890,15 @@ Running `pyright` would identify the same issues.
 
 ## System Prompts
 
-System prompts might seem simple at first glance since they're just strings (or sequences of strings that are concatenated), but crafting the right system prompt is key to getting the model to behave as you want.
+System prompts are strings (or sequences of strings) that guide model behavior. Crafting the right system prompt is key to getting the model to behave as you want.
 
-!!! tip
-    For most use cases, you should use `instructions` instead of "system prompts".
+!!! warning "Behavior with `message_history`"
+    When you provide `message_history`, the current agent's `system_prompt` is **not added** to the request.
```
DouweM: Let's make it more explicit why this difference exists, and the semantic difference between system parts and instructions, rather than just framing this as a gotcha or maybe even bug. It's worth looking into this a bit yourself, but part of the story is that model APIs treat the system prompt as an explicit part of the message history (`SystemPromptPart`), so if you want to benefit from prompt caching, it cannot be changed between steps of an agent run, and since it's in the message history as a part, it carries over when you use one agent's message history with another, taking priority over that agent's own system prompt, as it'd be unexpected to modify the explicitly provided message history/parts. Instructions on the other hand are a property of a specific API request, meaning that they can be swapped out independent of message history caching, and it does _not_ carry over when you use one agent's message history with another, as the current instructions always take priority.

It's also worth mentioning that system prompt is ignored when message history is passed in, _even when the message history didn't itself have any system prompts_, because we want it to be consistent that the the agent's system prompt is only used on a new run (0 history), while instructions are used in every request.

Also worth noting that the instructions do not override existing system prompts in the history: you can end up with both if the history had system prompts and the new agent has instructions.

Basically, worth giving more background and stating the main difference clearly, and then laying out the implications.

docs/agents.md
```diff
@@ -945,16 +943,58 @@ _(This example is complete, it can be run "as is")_
...
-- `instructions` when you want your request to the model to only include system prompts for the _current_ agent
-- `system_prompt` when you want your request to the model to _retain_ the system prompts used in previous requests (possibly made using other agents)
-
-In general, we recommend using `instructions` instead of `system_prompt` unless you have a specific reason to use `system_prompt`.
+Instructions guide model behavior like system prompts, but handle `message_history` differently:
+
+| Feature | With `message_history` |
+|---------|------------------------|
+| `system_prompt` | Current agent's prompts are **skipped** (history's prompts remain) |
+| `instructions` | Current agent's instructions are **always included** (history's are ignored) |
```
DouweM: For just 2 rows an 2 columns I don't think a table is better than a list, but my comment above may result in more differences we want to call out that would make a table a good option.

docs/agents.md
```diff
@@ -945,16 +943,58 @@ _(This example is complete, it can be run "as is")_
 
 ## Instructions
```
DouweM: Let's take this opportunity to move the Instructions section before System Prompts, as it's now the recommended route. It may also be helpful to move the sections under a common heading, so that the differences can be laid out upfront, before going into detail on how to use/configure each separately.

docs/agents.md
```diff
@@ -945,16 +943,58 @@ _(This example is complete, it can be run "as is")_
...
+Instructions guide model behavior like system prompts, but handle `message_history` differently:
+
+| Feature | With `message_history` |
+|---------|------------------------|
+| `system_prompt` | Current agent's prompts are **skipped** (history's prompts remain) |
+| `instructions` | Current agent's instructions are **always included** (history's are ignored) |
+
+Use `instructions` by default. Use `system_prompt` only when you need prompts to accumulate across multiple agents or runs in a conversation.
+
+??? info "Example: payload structure across runs using both"
```
DouweM: "payload structure" undersells the difference I laid out above of "part of history vs property of request". So I think it's helpful to refer to how these show up in the API and `ModelRequest`, but not framed as a technical example, but part of the key explanation of the difference between the two.

dsfaccini:
https://github.com/user-attachments/assets/1bc7610e-c7bf-46eb-b954-77bbabb753c8

docs/agents.md
```diff
@@ -888,45 +888,47 @@ Found 2 errors in 1 file (checked 1 source file)
...
 
-## System Prompts
+## Instructions and System Prompts
 
-System prompts might seem simple at first glance since they're just strings (or sequences of strings that are concatenated), but crafting the right system prompt is key to getting the model to behave as you want.
+### Instructions
 
-!!! tip
-    For most use cases, you should use `instructions` instead of "system prompts".
+Both `instructions` and `system_prompt` guide model behavior through text, but they differ architecturally:
```
DouweM, line 932: This should be before the Instructions heading right?

docs/agents.md
```diff
@@ -888,45 +888,47 @@ Found 2 errors in 1 file (checked 1 source file)
...
 
-    If you know what you are doing though and want to preserve system prompt messages in the message history sent to the
-    LLM in subsequent completions requests, you can achieve this using the `system_prompt` argument/decorator.
+- **`instructions`**: A property of [`ModelRequest.instructions`][pydantic_ai.messages.ModelRequest.instructions], re-evaluated on every model turn
```
DouweM, line 934: "re-evaluated on every model turn" as a quality of instructions is a bit misleading, because we support dynamic system prompt functions as well, which are also re-evaluated.

docs/agents.md
```diff
@@ -888,45 +888,47 @@ Found 2 errors in 1 file (checked 1 source file)
...
+- **`system_prompt`**: Becomes a [`SystemPromptPart`][pydantic_ai.messages.SystemPromptPart] in [`ModelRequest.parts`][pydantic_ai.messages.ModelRequest.parts], part of message history
```
DouweM, line 935: I think you're overindexing on the architectural/implementation difference and mentioning classes and implementation details too quickly, instead of explaining first explaining the semantic difference of the 2 concepts, and the implications of the fact that system prompts are part of history (with caching implications etc) while instructions are passed to the model _alongside_ the history.

It's also confusing that we imply that instructions are not part of history, even though they show up in ModelRequest.instructions in the message history. So I think we need more non-Pydantic-AI specific explanation up front.

docs/agents.md
```diff
@@ -888,45 +888,47 @@ Found 2 errors in 1 file (checked 1 source file)
...
 
-    See the section below on [Instructions](#instructions) for more information.
+**Recommendation**: Use `instructions` by default. Use `system_prompt` only when you need prompts to persist across multiple agents or runs via `message_history`.
 
-Generally, system prompts fall into two categories:
+Instructions are re-evaluated on **every** model turn (including tool calls and retries) via the agent's `get_instructions()` method. When you pass `message_history`, the current agent's instructions are always used — any instructions from the previous agent are not carried over. This makes instructions ideal for multi-agent handoffs where each agent needs its own guidance.
```
DouweM, line 939: - system prompt functions are also sometimes re-evaluated on every turn, so that's not the relevant difference here
- `get_instructions()` is deeefinitely not appropriate to mention here, that's an impl detail the user should never need to care about.
- "current agent" and "previous agent" are confusing. If we're going to use those terms, we should state up front that we're talking about a multi-agent scenario. 

Note that https://github.com/pydantic/pydantic-ai/pull/4014#discussion_r2714182577  doesn't use the current/previous agent terms at all, so I don't know if they're relevant/necessary for the explanation.

docs/agents.md
```diff
@@ -994,15 +990,68 @@ print(result.output)
...
-   information like the dependencies used on that run.
-4. Another dynamic instruction, instructions don't have to have the `RunContext` parameter.
+2. Static system prompt defined at agent creation time.
+3. Dynamic system prompt defined via a decorator with [`RunContext`][pydantic_ai.tools.RunContext], this is called just after `run_sync`, not when the agent is created, so can benefit from runtime information like the dependencies used on that run.
+4. Another dynamic system prompt, system prompts don't have to have the `RunContext` parameter.
 
 _(This example is complete, it can be run "as is")_
 
-Note that returning an empty string will result in no instruction message added.
+### Key Differences
```
DouweM, line 1036: Shouldn't be a separate section, if we want to clearly present the differences, we should do that up front.

docs/agents.md
```diff
@@ -994,15 +990,68 @@ print(result.output)
...
+
+| Aspect | Instructions | System Prompts |
+|--------|--------------|----------------|
+| **Storage** | `ModelRequest.instructions` property | `SystemPromptPart` in `ModelRequest.parts` |
+| **Evaluated** | Every model turn | First turn only |
+| **With `message_history`** | Current agent's used | History's remain, current agent's skipped |
+| **Caching** | Via provider instructions field | Via provider prompt caching |
```
DouweM, line 1043: What does "caching via provider instructions field" even mean?

docs/agents.md
```diff
@@ -994,15 +990,68 @@ print(result.output)
...
+
+Both can coexist: an agent can have `instructions` while using `message_history` that contains system prompts from a previous agent.
```
DouweM, line 1045: But this will have weird results no?

docs/agents.md
```diff
@@ -994,15 +990,68 @@ print(result.output)
...
+    # - agent_b's instructions are added as request property
+    result2 = agent_b.run_sync('How are you?', message_history=result1.all_messages())
+    ```
+
+    **Internal structure:**
+
+    - Run 1 creates: `ModelRequest(parts=[SystemPromptPart('Be helpful'), UserPromptPart('Hello')], instructions=None)`
+    - Run 2 creates: `ModelRequest(parts=[UserPromptPart('How are you?')], instructions='Speak French')`
+
+    **What gets sent to the provider (Run 2):**
```
DouweM, line 1070: I don't think this part is worth including

docs/agents.md
```diff
@@ -994,15 +990,68 @@ print(result.output)
...
+
+    - **OpenAI**: System prompts and instructions become separate messages:
```
DouweM, line 1072: Not in OpenAI responses!

docs/agents.md
```diff
@@ -994,15 +990,68 @@ print(result.output)
...
+    # Run 1: agent_a's system_prompt is added to the request
+    result1 = agent_a.run_sync('Hello')
+
+    # Run 2: agent_b uses message_history from run 1
+    # - agent_a's system_prompt remains in history
+    # - agent_b's instructions are added as request property
+    result2 = agent_b.run_sync('How are you?', message_history=result1.all_messages())
+    ```
+
+    **Internal structure:**
```
DouweM, line 1065: "internal" is the wrong word, as we expose the model request to the user

docs/agents.md
```diff
@@ -994,15 +990,68 @@ print(result.output)
...
+
+    agent_a = Agent('test', system_prompt='Be helpful')
+    agent_b = Agent('test', instructions='Speak French')
+
+    # Run 1: agent_a's system_prompt is added to the request
+    result1 = agent_a.run_sync('Hello')
+
+    # Run 2: agent_b uses message_history from run 1
+    # - agent_a's system_prompt remains in history
+    # - agent_b's instructions are added as request property
```
DouweM, line 1061: This would be weird right? Because the model will have contradicting instructions.

We should probably recommend being consistent in which you use.

docs/agents.md
dsfaccini: yeah but what I didn't want to happen is that the user clicks on the link that has the `#instructions` anchor and misses this description, which is why I have the `##` header only to organize the other two underneath

docs/agents.md
dsfaccini: definitely but you can use both https://community.openai.com/t/response-endpoint-instructions-parameter-vs-developer-role/1312916

and if one turns out to weigh more heavily on the agent's behavior then the user will learn about it

PR #4015, opened 2026-01-16
  Allow sampling params for GPT-5.1+ when reasoning is off

### LINKED ISSUE #4011: Sampling params not passed to GPT-5.2
### Initial Checks

### Description

The samplings parameters (ex: temperature) are not passed in the request when using pydantic ai + litellm provider + litellm proxy.

Screenshot from the request in the litellm prox web view. 
Image

It works as expected when I use the litellm lib.
Image

```Python
import litellm

litellm.completion(
    api_key=settings.LITELLM_API_KEY,
    base_url=str(settings.LITELLM_PROXY_URL),
    model="gpt-5.2",
    messages=[
        {
            "role": "user",
            "content": "Hello, how are you?"
        }
    ],
    temperature=0.8
)
```

```Python
from app.conf.config import settings
from pydantic_ai import Agent, ModelSettings

lite_llm_provider = LiteLLMProvider(
    api_key=settings.LITELLM_API_KEY,
    api_base=str(settings.LITELLM_PROXY_URL),
)

llm_model = OpenAIChatModel(
    model_name=settings.LLM_MODEL,
    provider=lite_llm_provider,
)

agent = Agent(
    model=llm_model,
    name="Coach-Agent",
    system_prompt="You are an helpful assistant",
    model_settings=ModelSettings(
        temperature=0.8
    )
)

await agent.run("Hello, how are you?")
```

### Logfire Trace

_No response_

- Python:3.12
- Pydantic AI:1.42
- LLM provider SDK:

dsfaccini:
This is adding a new flag `openai_supports_sampling_with_reasoning_off` to the `OpenAIModelProfile`, not sure you're gonna agree on that @DouweM so let me know. The #3894 is only adding `thinking_always_enabled` but that doesn't help discriminate the handling of the params here.

- Closes #4011 

### Pre-Review Checklist

### Pre-Merge Checklist

  - should we doc this?


pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -1484,8 +1506,9 @@ async def _responses_create(  # noqa: C901
             text = text or {}
             text['verbosity'] = verbosity
 
-        unsupported_model_settings = profile.openai_unsupported_model_settings
-        for setting in unsupported_model_settings:
+        _warn_about_dropped_sampling_params(profile, model_settings)
+
+        for setting in profile.openai_unsupported_model_settings:
```
DouweM: I'd prefer to move this into the method, call it `drop_sampling_params`, and then both do the work + the warning in there

pydantic_ai_slim/pydantic_ai/profiles/openai.py
```diff
@@ -73,6 +88,12 @@ class OpenAIModelProfile(ModelProfile):
     openai_supports_encrypted_reasoning_content: bool = False
     """Whether the model supports including encrypted reasoning content in the response."""
 
+    openai_supports_sampling_with_reasoning_off: bool = False
+    """Whether the model supports sampling parameters (temperature, top_p, etc.) when reasoning_effort='none'.
```
DouweM, line 104: Don't all these models support sampling when reasoning is off? :) I think the specific thing we need to encode is something like `openai_reasoning_enabled_by_default`, which we'd enable for GPT-5 only (right?), and then we only drop the sampling params if reasoning is specifically enabled or if it is _not specifically disabled_ + reasoning is on by default.

Maybe we shouldn't be using `unsupported_model_settings` for this at all (because it only makes sense if reasoning is always enabled for a given model meaning it never supports these settings), and instead always drop all the sampling params if we can detect that reasoning is enabled (by checking `openai_reasoning_enabled_by_default` + the reasoning effort setting)

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/models/openai.py
dsfaccini: yeah agree, I actually left this as is cause I thought it was something more general but this is really only for the sake of the thinking-incompatible params, so if we have other things in the future lets also drop them in their own handler

DouweM: Hmm we also use openai_unsupported_model_settings in CerebrasProvider, so it's not quite reasoning-specific. So then it makes sense to keep the separate things. It should be clear from the `_warn_about_dropped_sampling_params` function name though that it actually drops, not just warns

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -263,6 +263,29 @@ def _check_azure_content_filter(e: APIStatusError, system: str, model_name: str)
...
 
 
+def _warn_about_dropped_sampling_params(profile: OpenAIModelProfile, model_settings: OpenAIChatModelSettings) -> None:
+    """Warn when sampling params will be dropped due to reasoning being enabled.
+
+    For models that support reasoning_effort='none' (GPT-5.1+), sampling params are only
+    supported when reasoning is off. This warns users when their params will be ignored.
+    """
+    reasoning_effort = model_settings.get('openai_reasoning_effort', 'none')
+    if not profile.openai_supports_reasoning_effort_none or reasoning_effort == 'none':
```
DouweM: It's a bit confusing now that for o series and GPT5, the dropping is handled via the other field, and this is only for 5.1+. Because reading this, it seems like in the case of `if not profile.openai_supports_reasoning_effort_none`, we should be dropping all fields instead of `returning` and not deleting anything.

DouweM: So maybe we should not use the unsupported... field for these reasoning fields, only for CerebrasProvider, and then have a new field `openai_supports_reasoning`. And we drop all sampling params if `openai_supports_reasoning and (not openai_supports_reasoning_effort_none or settings[openai_reasoning_effort] != 'none')`

pydantic_ai_slim/pydantic_ai/profiles/openai.py
dsfaccini: with `openai_reasoning_enabled_by_default` you still need to check whether the model even supports 'none', so I went with another flag `openai_supports_reasoning_effort_none` because if it doesn't we don't even need to check its `reasoning_effort`

tests/profiles/test_openai.py
```diff
@@ -0,0 +1,88 @@
...
+
+with try_import() as imports_successful:
+    from pydantic_ai.profiles.openai import OpenAIModelProfile, openai_model_profile
+
+pytestmark = [
+    pytest.mark.skipif(not imports_successful(), reason='openai not installed'),
+]
+
+
+class TestSamplingParamsSupport:
```
dsfaccini: I'm redoing this class a parametrized

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -263,6 +263,40 @@ def _check_azure_content_filter(e: APIStatusError, system: str, model_name: str)
...
+
+    Reasoning models (o-series, GPT-5, GPT-5.1+) don't support sampling parameters when
+    reasoning is active. For models that support reasoning_effort='none' (GPT-5.1+),
+    sampling params are allowed when reasoning is off.
+    """
+    if not profile.openai_supports_reasoning:
+        return
+
+    reasoning_effort = model_settings.get('openai_reasoning_effort', 'none')
+    if profile.openai_supports_reasoning_effort_none and reasoning_effort == 'none':
```
DouweM: The logic here is correct in terms of behavior, but by defaulting to `'none'` even when `openai_supports_reasoning_effort_none` is `False`, we end in the code below with `reasoning_effort == 'none'`, even though that's not actually the active reasoning effort and not actually supported by the model 😅 So I think we should have just `settings.get(...)` and then `if ... and reasoning_effort in (None, 'none')` with a comment saying "On these models, none is teh default."

That way, if we fall through this branch to the code below, and the setting is unset, we'll end up with `None` as a value, correctly indicating "unset" instead of the _unsupported_ `none`.

pydantic_ai_slim/pydantic_ai/profiles/openai.py
```diff
@@ -73,6 +88,17 @@ class OpenAIModelProfile(ModelProfile):
     openai_supports_encrypted_reasoning_content: bool = False
     """Whether the model supports including encrypted reasoning content in the response."""
 
+    openai_supports_reasoning: bool = False
+    """Whether the model supports reasoning (o-series, GPT-5, GPT-5.1+).
```
DouweM:
```suggestion
    """Whether the model supports reasoning (o-series, GPT-5+).
```

pydantic_ai_slim/pydantic_ai/profiles/openai.py
```diff
@@ -10,6 +10,21 @@
...
 from . import ModelProfile
 
+SAMPLING_PARAMS = (
+    'temperature',
+    'top_p',
+    'presence_penalty',
+    'frequency_penalty',
+    'logit_bias',
+    'logprobs',
+    'top_logprobs',
```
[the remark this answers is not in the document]
DouweM: This may have been an existing issue!

PR #4016, opened 2026-01-16
Add `args_validator` parameter to tools for pre-execution validation

### LINKED ISSUE #3893: Support for additional tool validation logic
### Description

Tool call validation is really valuable for steering proper agent behavior. Currently pydantic-ai only provides tool call argument schema validation, but it could be useful to attach custom validation logic.

For example, an agent may provide a hallucinated UUID which is of proper format, but does not correspond to actual data.

Today, this validation logic must happen in the tool logic and throw a `ModelRetry`, however, with deferred tools or tools that require approval, we end up requesting approval first to potentially run into invalid input. It would be better to have all validation occur up-front before any logic is run.

### References

_No response_

### LINKED ISSUE #3992: Allow way to run all validation logic before `FunctionToolCallEvent`
### Description

This is a new issue to track the specific request originally raised here: https://github.com/pydantic/pydantic-ai/issues/3893 . Request to open a new issue followed from discussion with @DouweM 

The request can be summarized as:

1. Allow a way to add custom validation logic that allow `async` calls that can run before the `FunctionToolCallEvent` is emitted
2. Add support for filtering `FunctionToolCallEvent`s where validation failed

Currently, custom validation logic typically happens inside of the tool definition, with potential `ModelRetry` being raised. This by nature always occurs after the original `FunctionToolCallEvent` is emitted so it's difficult to know if all validation has passed before we pass the event to the frontend for rendering.

Ideally we are able to add custom validation logic (e.g., checking if an ID is valid record in the DB) that occurs before `FunctionToolCallEvent` is emitted so it is much easier to filter out for the frontend. 

---

On the comment reguarding strong types for the `args_validator`, it appears at least for the `@agent.tool` decorator that the type can be `ToolFuncContext[AgentDepsT, ToolParams] | None` or similar, requiring the validator to take the exact same params at the tool function itself.

### References

_No response_

jerry-reevo:
Closes #3992
Closes #3893

### Summary

  - Add `args_validator` parameter to all tool entry points for custom pre-execution validation
  - Add `args_valid` field to `FunctionToolCallEvent` indicating validation status
  - Refactor `ToolManager` internals: split `handle_call` into `validate_tool_call()` + `execute_tool_call()` with a new `ToolCallValidation` dataclass, separating validation from execution
  - For deferred tools, validation now runs at deferral time — only tool calls with valid arguments are deferred, while failed validation triggers a retry like regular tools

### Test plan

  - Run new args_validator tests in tools: `uv run pytest tests/test_tools.py -k args_validator -v`
  - Run new args_validator tests in streaming: `uv run pytest tests/test_streaming.py -k args_validator -v`
  - Run existing tool tests: `uv run pytest tests/test_tools.py -v`
  - Run existing agent tests: `uv run pytest tests/test_agent.py -v`
  - Run existing streaming tests: `uv run pytest tests/test_streaming.py -v`
  - Type checking: `uv run pyright` to verify validator signature enforcement
  - Full test suite: `make test`

###  Design Decisions for Review

1. Strongly-typed validator signature

    The `args_validator` function has the exact same typed parameters as the tool function:

    ```Python
    def my_validator(ctx: RunContext[int], user_id: int) -> None:
        if user_id < 0:
            raise ModelRetry('user_id must be positive')
    

    @agent.tool(args_validator=my_validator)
    def get_user(ctx: RunContext[int], user_id: int) -> str:
        return f"User {user_id}"
    ```

    Trade-off: This provides IDE autocomplete and type safety, but requires the validator to duplicate the parameter list. An alternative would be `Callable[[RunContext, dict[str, Any]], None]` which is simpler but loses type safety.  

2. `args_valid` field semantics

    - True: pydantic schema validation AND custom validation pass (if configured)
    - False: pydantic schema validation OR custom validation failed

3. Validator error handling

    Validators should raise `ModelRetry` on failure. Other exceptions propagate as tool errors. This matches the existing pattern for tool execution errors.

### Pre-Review Checklist

### Pre-Merge Checklist

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/_tool_manager.py
+++ b/pydantic_ai_slim/pydantic_ai/_tool_manager.py
@@ -196,8 +196,8 @@ class ToolManager(Generic[AgentDepsT]):
             The validated arguments as a dictionary.
 
         Raises:
-            ValidationError: If Pydantic schema validation fails.
-            ModelRetry: If custom args_validator_func validation fails.
+            ValidationError: If argument validation fails.
+            ModelRetry: If argument validation fails with a retry request.
         """
         pyd_allow_partial = 'trailing-strings' if allow_partial else 'off'
         validator = tool.args_validator


pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -1256,6 +1264,7 @@ def tool_plain(
         description: str | None = None,
         retries: int | None = None,
         prepare: ToolPrepareFunc[AgentDepsT] | None = None,
+        args_validator: ArgsValidatorFunc[AgentDepsT, ToolParams] | None = None,
```
jerry-reevo, line 1296: Should the `tool_plain` methods take `AgentDepsT`?

pydantic_ai_slim/pydantic_ai/toolsets/abstract.py
```diff
@@ -57,6 +57,12 @@ class ToolsetTool(Generic[AgentDepsT]):
 
     For example, a [`pydantic.TypeAdapter(...).validator`](https://docs.pydantic.dev/latest/concepts/type_adapter/) or [`pydantic_core.SchemaValidator`](https://docs.pydantic.dev/latest/api/pydantic_core/#pydantic_core.SchemaValidator).
     """
+    args_validator_func: Callable[..., Any] | None = None
```
jerry-reevo, line 60: There is some assumption here that strong typing is enforced at entry points (e.g., decorators)

jerry-reevo:
@DouweM would you mind providing some feedback?

pydantic_ai_slim/pydantic_ai/_tool_manager.py
```diff
@@ -91,6 +92,74 @@ def get_tool_def(self, name: str) -> ToolDefinition | None:
         except KeyError:
             return None
 
+    async def validate_tool_args(
```
DouweM: Some duplication here with https://github.com/pydantic/pydantic-ai/pull/4049 :)

Can you please have a look there and make it so that this PR can close https://github.com/pydantic/pydantic-ai/issues/3893 as well? I'll close that PR pointing at this one.

pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -1206,6 +1209,10 @@ async def spam(ctx: RunContext[str], y: float) -> float:
             prepare: custom method to prepare the tool definition for each step, return `None` to omit this
                 tool from a given step. This is useful if you want to customise a tool at call time,
                 or omit it completely from a step. See [`ToolPrepareFunc`][pydantic_ai.tools.ToolPrepareFunc].
+            args_validator: custom method to validate tool arguments before execution. The validator receives
+                the same typed parameters as the tool function, with `RunContext` as the first argument.
+                Should raise [`ModelRetry`][pydantic_ai.exceptions.ModelRetry] on validation failure.
```
DouweM: We need to document the feature on the advanced tool features doc!

pydantic_ai_slim/pydantic_ai/agent/__init__.py
DouweM: You mean if they should take `RunContext`? I think hooks like this always should.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -922,7 +923,8 @@ async def process_tool_calls(  # noqa: C901
                 # This allows exhaustive strategy to complete successfully when at least one output tool is valid
                 if final_result:
                     # If output tool fails when we already have a final result, skip it without retrying
-                    yield _messages.FunctionToolCallEvent(call)
+                    # Validation already failed in handle_call, so args_valid=False
+                    yield _messages.FunctionToolCallEvent(call, args_valid=False)
```
DouweM: I'm not sure `args_valid=False` is correct here, because there could be other reasons that `handle_call` failed than argument validation. Maybe `args_validated=False` gets the point across that these args are not guaranteed to be valid, without saying that they are definitely invalid?

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -943,7 +945,8 @@ async def process_tool_calls(  # noqa: C901
                     ctx.state.increment_retries(
                         ctx.deps.max_result_retries, error=e, model_settings=ctx.deps.model_settings
                     )
-                yield _messages.FunctionToolCallEvent(call)
+                # ToolRetryError means validation failed in handle_call, so args_valid=False
```
DouweM: Same as above; the error could be coming from inside the function, so args could very well be valid.

Specifically breaking up the args validation + tool call steps could resolve this.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
jerry-reevo: @DouweM What do you recommend here? It looks like you're right that this branch can occur for other reasons not related to validation, although most cases seem to cover validation/`ModelRetry` type of errors. I think saying `False` means they are not guaranteed to be valid works for my case, but I'm not sure for others.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
jerry-reevo: @DouweM Would you be in favor of the flavor where `args_valid` only reflects the result from schema validation and custom validators? If you do additional validation in the tool function and raise a `ModelRetry` it will not reflect that.

I think this might be the cleanest semantics but do require a bit of quirkiness.

jerry-reevo: @DouweM I'm starting to realize that the reason validation and execution were a bit coupled (and maybe why they were missing for `approval_required` tools) is for tracing and tool return behaviors.

It appears that when there is a validation error today, we both trace the call as having an error and set the tool_return to the validation error.

It seems we probably need to do that for all other use cases where validation is run too. For some corner cases, like for output tools and `end_strategy == 'early'`, we can probably continue to set `args_valid = <validation result>` without a tool call span or information in the tool return.

What are your thoughts about how we handle other parts?

dsfaccini:

```bash
git pull

deactivate

rm -rf .venv

make install
```

pydantic_ai_slim/pydantic_ai/_agent_graph.py
DouweM: @jerry-reevo I'm OK with saying `args_validated=True` guarantees the args are valid, but `args_validated=False` does not guarantee that they are invalid, just that they're not validated.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
DouweM: > Would you be in favor of the flavor where args_valid only reflects the result from schema validation and custom validators? If you do additional validation in the tool function and raise a ModelRetry it will not reflect that.

Yeah I think of args validation is a specific thing separate from validation inside the tool function.

> It seems we probably need to do that for all other use cases where validation is run too. For some corner cases, like for output tools and `end_strategy == 'early'`, we can probably continue to set `args_valid = <validation result>` without a tool call span or information in the tool return.

That sounds right to me.

> What are your thoughts about how we handle other parts?

Sorry I'm doing a quick review without having the whole context ready in my head; what specific other parts are you thinking about? Easier for me to make a call on specific scenarios :)

jerry-reevo: @DouweM Specifically deferred tools. 

Latest revision now always runs validation on all deferred tool calls and will immediately create a tool call span and set the tool return to the validation error, if the call failed validation.

If the deferred tool call passes, no tool call span and no tool results are set. Previous behavior is preserved here.

---

In order to properly coordinate across the layers, the code got a bit messy. Open to restructuring it, but feels like a lot to manage at this time.

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -1970,6 +1970,13 @@ class FunctionToolCallEvent:
 
     _: KW_ONLY
 
+    args_validated: bool = False
```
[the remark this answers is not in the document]
DouweM: > adding any new required field (even with a default) to a public dataclass is a breaking change

I don't think so

> is a breaking change acceptable for the current version?

Definitely not

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -924,48 +925,92 @@ async def process_tool_calls(  # noqa: C901
...
+                        tool_call_id=call.tool_call_id,
+                    )
+                    output_parts.append(part)
 
-                # In both `early` and `exhaustive` modes, use the first output tool's result as the final result
-                if not final_result:
-                    final_result = result.FinalResult(result_data, call.tool_name, call.tool_call_id)
+                    # In both `early` and `exhaustive` modes, use the first output tool's result as the final result
+                    if not final_result:
+                        final_result = result.FinalResult(result_data, call.tool_name, call.tool_call_id)
```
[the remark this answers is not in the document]
jerry-reevo: I think this is intended behavior. @DouweM could you help confirm?

pydantic_ai_slim/pydantic_ai/messages.py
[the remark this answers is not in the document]
jerry-reevo: @DouweM I changed this to have default `False` to semantically mean validation may not have run. This should not be a breaking change, right?

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -924,48 +925,92 @@ async def process_tool_calls(  # noqa: C901
...
             else:
-                part = _messages.ToolReturnPart(
-                    tool_name=call.tool_name,
-                    content='Final result processed.',
-                    tool_call_id=call.tool_call_id,
-                )
-                output_parts.append(part)
+                # Validation passed - execute the tool
+                try:
+                    result_data = await tool_manager.execute_tool_call(validated)
```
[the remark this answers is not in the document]
jerry-reevo: This appears to have been intentional pre-existing behavior. Output tools are not traced or counted in output tool calls.

pydantic_ai_slim/pydantic_ai/messages.py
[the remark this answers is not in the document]
DouweM: I agree, this is fine

pydantic_ai_slim/pydantic_ai/_tool_manager.py
```diff
@@ -119,151 +144,334 @@ def get_tool_def(self, name: str) -> ToolDefinition | None:
...
-                tool_call_metadata=metadata,
-                partial_output=allow_partial,
+            if not allow_partial:  # pragma: no branch
+                self.failed_tools.add(name)
+
+            return ValidatedToolCall(
+                call=call,
+                tool=None,
+                ctx=self.ctx,
+                args_validated=False,
```
DouweM: Maybe here it should be `args_valid` since it indicates the actual validation status, right?

pydantic_ai_slim/pydantic_ai/_tool_manager.py
```diff
@@ -119,151 +144,334 @@ def get_tool_def(self, name: str) -> ToolDefinition | None:
...
-                )
-            else:
-                args_dict = validator.validate_python(
-                    call.args or {}, allow_partial=pyd_allow_partial, context=ctx.validation_context
-                )
+        ctx = self._build_tool_context(call, tool, allow_partial=allow_partial, approved=approved, metadata=metadata)
 
-            return await self.toolset.call_tool(name, args_dict, ctx, tool)
+        try:
+            validated_args = await self._validate_tool_args(call, tool, ctx, allow_partial=allow_partial)
```
DouweM, line 264: Should we skip this if `call.args_validated` is already `True`?

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -1970,6 +1970,13 @@ class FunctionToolCallEvent:
...
+    """Whether the tool arguments have been validated successfully.
+
+    - `True`: Schema validation and custom validation (if configured) both passed; args are guaranteed valid.
+    - `False`: Validation was not performed, or validation failed. No guarantees about argument validity.
```
DouweM: Would it simplify things if we made this `args_valid: bool | None = None`, so that we can communicate invalidity?

pydantic_ai_slim/pydantic_ai/_tool_manager.py
```diff
@@ -119,151 +144,334 @@ def get_tool_def(self, name: str) -> ToolDefinition | None:
...
+                validated.ctx,
+                validated.tool,
+            )
+        except ModelRetry as e:
+            # Tool raised ModelRetry during execution - check max retries
+            max_retries = validated.tool.max_retries
+            current_retry = self.ctx.retries.get(name, 0)
+
+            if current_retry == max_retries:
+                raise UnexpectedModelBehavior(f'Tool {name!r} exceeded max retries count of {max_retries}') from e
```
DouweM: We have this in multiple places now, is that really necessary? Same for the `RetryPromptPart` building...

pydantic_ai_slim/pydantic_ai/_agent_graph.py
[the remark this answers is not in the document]
DouweM: Yeah I think so

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/_tool_manager.py
jerry-reevo: It looks like there is never a `call.args_validated` var at this location.

pydantic_ai_slim/pydantic_ai/_tool_manager.py
jerry-reevo: Changed

pydantic_ai_slim/pydantic_ai/_tool_manager.py
DouweM: But isn't `call` a `ToolCallPart`? This may not be the right location, but it feels like we should be able to skip double validation if we store the validation state on the tool call.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -924,48 +925,92 @@ async def process_tool_calls(  # noqa: C901
         # Early strategy is chosen and final result is not yet set
         # Or exhaustive strategy is chosen
         else:
+            # Validate first to get accurate args_validated status
```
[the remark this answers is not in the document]
jerry-reevo: Since the validation did fail, I think it's fine to keep `args_valid=False` and we re-raise the error

pydantic_ai_slim/pydantic_ai/_tool_manager.py
jerry-reevo: I ended up propagating the tool validation results to avoid double validation. `ToolCallPart` was left unmodified for now.

pydantic_ai_slim/pydantic_ai/_tool_manager.py
jerry-reevo: I tried reducing code duplication by factoring out some of the common logic.

pydantic_ai_slim/pydantic_ai/tools.py
```diff
@@ -273,6 +287,7 @@ class Tool(Generic[ToolAgentDepsT]):
     name: str
     description: str | None
     prepare: ToolPrepareFunc[ToolAgentDepsT] | None
+    args_validator: ArgsValidatorFunc[ToolAgentDepsT, ...] | None
```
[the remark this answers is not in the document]
DouweM: 1. I'm inclined to prefer this approach because of type safety, but maybe it won't be flexible enough and we need to (also) have a way to dynamically validate the args dict. We could support both the params list, or `call: ToolCallPart` so you can test `call.args`. @jerry-reevo Wdyt?
2. No this is enough
3. Hmmm, I like it returning overridden args, since that's kind of what a "Pydantic validator" does as well, but that would not work nicely with the params list approach, although it'd work with `ToolCallPart`. @jerry-reevo Curious what you think here as well
4. Yes I like supporting `ApprovalRequired` and `CallDeferred` here as well.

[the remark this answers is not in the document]
jerry-reevo: > 1. I'm inclined to prefer this approach because of type safety, but maybe it won't be flexible enough and we need to (also) have a way to dynamically validate the args dict. We could support both the params list, or call: ToolCallPart so you can test call.args. @jerry-reevo Wdyt?
> 3. Hmmm, I like it returning overridden args, since that's kind of what a "Pydantic validator" does as well, but that would not work nicely with the params list approach, although it'd work with ToolCallPart. @jerry-reevo Curious what you think here as well

@DouweM The type safety and clarity of the explicit params list is really valuable for my use case. Some arguments have interdependencies that are much easier to reason about when all parameters are visible in one place.

A potential extension that accepts a `call: ToolCallPart` and returns transformed args would likely need to use `dict | None` as the return type, where None signals that the args should remain unmodified. Is this acceptable?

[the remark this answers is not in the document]
jerry-reevo: > 4. Yes I like supporting ApprovalRequired and CallDeferred here as well.

I agree, however, I would like to defer this. The current complexity level of the PR is fairly high, and _I think_ (please correct me if wrong) support for this could be added later without any breaking changes by adding exception handling during validation.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -913,7 +913,12 @@ async def process_tool_calls(  # noqa: C901
             output_parts.append(part)
         # Early strategy is chosen and final result is already set
         elif ctx.deps.end_strategy == 'early' and final_result:
-            yield _messages.FunctionToolCallEvent(call)
+            try:
+                validated = await tool_manager.validate_tool_call(call)
+                args_valid = validated.args_valid
+            except exceptions.UnexpectedModelBehavior:
+                args_valid = False
+            yield _messages.FunctionToolCallEvent(call, args_valid=args_valid)
```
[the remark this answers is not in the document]
jerry-reevo: I think I agree. Since function execution is actually being skipped, validation should probably be skipped too. I've reverted to original behavior.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -1038,11 +1087,47 @@ async def process_tool_calls(  # noqa: C901
...
+                        await tool_manager.execute_tool_call(
+                            validated,
+                            tracer=ctx.deps.tracer,
+                            include_content=include_content,
+                            instrumentation_version=instrumentation_version,
+                            usage=ctx.state.usage,
+                        )
+                    except ToolRetryError as e:
+                        output_parts.append(e.tool_retry)
+                        yield _messages.FunctionToolResultEvent(e.tool_retry)
```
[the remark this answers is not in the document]
jerry-reevo: Added comment for clarity.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -924,48 +924,92 @@ async def process_tool_calls(  # noqa: C901
...
+                        tool_call_id=call.tool_call_id,
+                    )
+                    output_parts.append(part)
+                    yield _messages.FunctionToolResultEvent(part)
+                else:
                     ctx.state.increment_retries(
-                        ctx.deps.max_result_retries, error=e, model_settings=ctx.deps.model_settings
+                        ctx.deps.max_result_retries,
+                        error=validated.validation_error,
+                        model_settings=ctx.deps.model_settings,
```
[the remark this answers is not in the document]
jerry-reevo: I think this new behavior is reasonable and preferred. Before it seems `RetryPromptPart` may have been used to signal validation failed, but since the run is ending and the output tool is not going to be retried, it is less semantically consistant than the new behavior where we return `TooReturnPart` and set `args_valid=False`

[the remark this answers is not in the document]
jerry-reevo: @DouweM This does change user visible message history behavior. Do you consider this a breaking change?

pydantic_ai_slim/pydantic_ai/_agent_graph.py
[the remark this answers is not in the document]
DouweM: Yep works for me

pydantic_ai_slim/pydantic_ai/_agent_graph.py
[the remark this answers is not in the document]
DouweM: The change makes sense to me

pydantic_ai_slim/pydantic_ai/tools.py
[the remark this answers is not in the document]
DouweM: @jerry-reevo Let's wait to implement both of those things then,  Iagree it can be done in a backward compatible way if/when users want it.

pydantic_ai_slim/pydantic_ai/tools.py
```diff
@@ -71,6 +72,19 @@
 
 Usage `ToolFuncEither[AgentDepsT, ToolParams]`.
 """
+ArgsValidatorFunc: TypeAlias = (
+    Callable[Concatenate[RunContext[AgentDepsT], ToolParams], Awaitable[None]]
+    | Callable[Concatenate[RunContext[AgentDepsT], ToolParams], None]
+)
```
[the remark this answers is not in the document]
DouweM: Yep, I've also responded there

pydantic_ai_slim/pydantic_ai/toolsets/abstract.py
```diff
@@ -57,6 +57,14 @@ class ToolsetTool(Generic[AgentDepsT]):
...
+    """Custom args validator function that runs after schema validation but before tool execution.
+
+    Called on every tool call, receiving the schema-validated arguments as keyword args.
+    The function should have the same typed parameters as the tool function,
+    with `RunContext` as the first argument.
+    Should raise [`ModelRetry`][pydantic_ai.exceptions.ModelRetry] on failure, return `None` on success.
+    """
```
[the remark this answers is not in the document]
DouweM: `ToolsetTool` is internal so as long as it's documented, I'm fine with a difference.

tests/test_streaming.py
```diff
@@ -3302,3 +3325,385 @@ async def test_get_output_after_stream_output():
...
+        return x + y
+
+    events: list[Any] = []
+    async for event in agent.run_stream_events('call add_numbers with x=1 and y=2', deps=42):
+        events.append(event)
+
+    tool_call_events: list[FunctionToolCallEvent] = [e for e in events if isinstance(e, FunctionToolCallEvent)]
+    assert len(tool_call_events) >= 1
+
+    add_number_events = [e for e in tool_call_events if e.part.tool_name == 'add_numbers']
```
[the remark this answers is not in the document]
jerry-reevo: This appears to have been on the wrong line number, which was confusing. The bare exception appears to be on line 3526.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -1038,11 +1075,48 @@ async def process_tool_calls(  # noqa: C901
...
+                        await tool_manager.execute_tool_call(
+                            validated,
+                            tracer=ctx.deps.tracer,
+                            include_content=include_content,
+                            instrumentation_version=instrumentation_version,
+                            usage=ctx.state.usage,
+                        )
+                    except ToolRetryError as e:
+                        output_parts.append(e.tool_retry)
+                        yield _messages.FunctionToolResultEvent(e.tool_retry)
```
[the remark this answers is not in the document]
jerry-reevo: I think this is intended behavior and thus not actionable.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -914,58 +923,86 @@ async def process_tool_calls(  # noqa: C901
...
+                        tool_call_id=call.tool_call_id,
+                    )
+                    output_parts.append(part)
 
-                # In both `early` and `exhaustive` modes, use the first output tool's result as the final result
-                if not final_result:
-                    final_result = result.FinalResult(result_data, call.tool_name, call.tool_call_id)
+                    # In both `early` and `exhaustive` modes, use the first output tool's result as the final result
+                    if not final_result:
+                        final_result = result.FinalResult(result_data, call.tool_name, call.tool_call_id)
```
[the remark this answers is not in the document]
DouweM: Yeah I'd like to clean this up to make it easier for me to understand the flow before I merge

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -1083,8 +1151,21 @@ async def _call_tools(  # noqa: C901
...
+        try:
+            validated = await tool_manager.validate_tool_call(call)
+        except exceptions.UnexpectedModelBehavior:
+            yield _messages.FunctionToolCallEvent(call, args_valid=False)
+            raise
+        validated_calls[call.tool_call_id] = validated
+        yield _messages.FunctionToolCallEvent(call, args_valid=validated.args_valid)
 
     with tracer.start_as_current_span(
         'running tools',
```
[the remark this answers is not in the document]
jerry-reevo: I am opting to leave it as-is.

The behavioral change only affects the max-retries-exceeded case (UnexpectedModelBehavior), which aborts the agent run regardless. In the old code with parallel execution, tool A might start before tool B's failure is detected, but the result would be discarded since the exception propagates up.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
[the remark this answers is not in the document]
jerry-reevo: Made the suggested refactor reducing duplication

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -1038,11 +1075,42 @@ async def process_tool_calls(  # noqa: C901
...
+            # Validate deferred tools before deferring them
+            # If validation fails, trace the error and add retry prompt
+            # If validation passes, just emit event and defer
             for call in calls:
-                yield _messages.FunctionToolCallEvent(call)
+                try:
+                    validated = await tool_manager.validate_tool_call(call)
+                except exceptions.UnexpectedModelBehavior:
+                    # Max retries exceeded - re-raise
+                    raise
```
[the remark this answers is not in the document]
jerry-reevo: Accepted suggested change to emit tool call event first

pydantic_ai_slim/pydantic_ai/_tool_manager.py
```diff
@@ -119,151 +144,293 @@ def get_tool_def(self, name: str) -> ToolDefinition | None:
...
-                wrap_validation_errors=wrap_validation_errors,
-                approved=approved,
-                metadata=metadata,
+        name = call.tool_name
+        tool = self.tools.get(name)
+
+        # Handle unknown tool
+        if tool is None:
+            if self.tools:
+                msg = f'Available tools: {", ".join(f"{name!r}" for name in self.tools.keys())}'
```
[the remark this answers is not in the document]
jerry-reevo: Accepted suggested change

pydantic_ai_slim/pydantic_ai/_tool_manager.py
```diff
@@ -119,151 +144,293 @@ def get_tool_def(self, name: str) -> ToolDefinition | None:
...
+            include_content: Whether to include tool arguments and results in the span.
+            instrumentation_version: The instrumentation version for span attribute names.
+            usage: Usage tracking object to increment tool_calls count.
+
+        Returns:
+            The tool result if validation passed and execution succeeded.
+
+        Raises:
+            ToolRetryError: If validation failed (contains the retry prompt).
+            RuntimeError: If trying to execute an external tool.
```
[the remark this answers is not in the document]
jerry-reevo: Changed to not require passing in `tracer`, `include_content` and `instrumentation_version` and just passing through the values from `self.ctx` and `tool_manager.ctx`. This reduced some boilerplate.

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -1981,6 +1981,14 @@ class FunctionToolCallEvent:
 
     _: KW_ONLY
 
+    args_valid: bool | None = None
```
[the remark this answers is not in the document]
jerry-reevo: After further discussion I think we agreed to revert to `args_valid` with tri-state semantics for clarity.

docs/tools-advanced.md
```diff
@@ -403,6 +403,43 @@ async def fast_tool() -> str:
...
+
+result = agent.run_sync('add 5 and 3', deps=100)
+print(result.output)
+#> {"add_numbers":0}
+```
+
+_(This example is complete, it can be run "as is")_
+
+When validation fails, the error message is sent back to the LLM as a retry prompt. This respects the `retries` setting on the tool. For [deferred tools](deferred-tools.md), validation runs at deferral time — only tool calls with valid arguments are deferred, while failed validation triggers a retry just like regular tools.
+
```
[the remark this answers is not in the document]
DouweM: Yep this is correct

docs/tools-advanced.md
```diff
@@ -403,6 +403,43 @@ async def fast_tool() -> str:
...
+The `args_validator` parameter is available on [`@agent.tool`][pydantic_ai.agent.Agent.tool], [`@agent.tool_plain`][pydantic_ai.agent.Agent.tool_plain], [`Tool`][pydantic_ai.tools.Tool], [`Tool.from_schema`][pydantic_ai.tools.Tool.from_schema], and [`FunctionToolset`][pydantic_ai.toolsets.function.FunctionToolset]. Validators can be sync or async functions. When used with [`prepare`](#tool-prepare), the prepare function runs first to customize the tool definition, then `args_validator` runs at call time.
```
DouweM: Note sure if the relationship between `prepare` and `args_validator` is worth specifying

docs/tools-advanced.md
```diff
@@ -403,6 +403,43 @@ async def fast_tool() -> str:
...
+agent = Agent('test', deps_type=int)
+
+
+def validate_sum_limit(ctx: RunContext[int], x: int, y: int) -> None:
+    """Validate that the sum doesn't exceed the limit from deps."""
+    if x + y > ctx.deps:
+        raise ModelRetry(f'Sum of x and y must not exceed {ctx.deps}')
+
+
+@agent.tool(args_validator=validate_sum_limit)
```
DouweM: I think the example will be more impactful if it's something that couldn't be done equivalently by inlining the check in the tool function, i.e. a tool that requires approval.

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -1981,6 +1981,14 @@ class FunctionToolCallEvent:
...
+    """Whether the tool arguments passed validation.
+
+    - `True`: Schema validation and custom validation (if configured) both passed; args are guaranteed valid.
```
DouweM, line 2033: Let's link to the custom validation docs here explicitly

pydantic_ai_slim/pydantic_ai/tools.py
```diff
@@ -71,6 +72,19 @@
...
+"""A function that validates tool arguments before execution.
+
+The validator receives the same typed parameters as the tool function,
+with [`RunContext`][pydantic_ai.tools.RunContext] as the first argument for dependency access.
+
+Should raise [`ModelRetry`][pydantic_ai.exceptions.ModelRetry] on validation failure, return `None` on success.
+
+Usage `ArgsValidatorFunc[AgentDepsT, ToolParams]`.
```
DouweM: Not worth mentioning

pydantic_ai_slim/pydantic_ai/tools.py
```diff
@@ -71,6 +72,19 @@
...
+ArgsValidatorFunc: TypeAlias = (
+    Callable[Concatenate[RunContext[AgentDepsT], ToolParams], Awaitable[None]]
+    | Callable[Concatenate[RunContext[AgentDepsT], ToolParams], None]
+)
+"""A function that validates tool arguments before execution.
+
+The validator receives the same typed parameters as the tool function,
+with [`RunContext`][pydantic_ai.tools.RunContext] as the first argument for dependency access.
+
+Should raise [`ModelRetry`][pydantic_ai.exceptions.ModelRetry] on validation failure, return `None` on success.
```
DouweM: "return None" makes it sounds like this is something to be done explicitly, I think we can drop it.

pydantic_ai_slim/pydantic_ai/tools.py
```diff
@@ -296,6 +311,7 @@ def __init__(
         name: str | None = None,
         description: str | None = None,
         prepare: ToolPrepareFunc[ToolAgentDepsT] | None = None,
+        args_validator: ArgsValidatorFunc[ToolAgentDepsT, ...] | None = None,
```
DouweM: Why `...` here and above? We can't make it match the signature of `function`?

pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -1158,6 +1159,7 @@ def tool(
         description: str | None = None,
         retries: int | None = None,
         prepare: ToolPrepareFunc[AgentDepsT] | None = None,
+        args_validator: ArgsValidatorFunc[AgentDepsT, ToolParams] | None = None,
```
DouweM, line 1171: Since we want to enforce matching arg signatures at the typing level, can you please add cases to `tests/typed_agent.py` to verify that correct usage typechecks, and incorrect usage doesn't (by adding an ignore directive)?

pydantic_ai_slim/pydantic_ai/toolsets/function.py
```diff
@@ -244,6 +254,7 @@ def add_function(
         description: str | None = None,
         retries: int | None = None,
         prepare: ToolPrepareFunc[AgentDepsT] | None = None,
+        args_validator: ArgsValidatorFunc[AgentDepsT, ...] | None = None,
```
DouweM: Should use `ToolParams`? We have a `...` in a couple places where I think we can be more strict

pydantic_ai_slim/pydantic_ai/_tool_manager.py
```diff
@@ -119,151 +144,263 @@ def get_tool_def(self, name: str) -> ToolDefinition | None:
...
+        if tool is None:
+            if self.tools:
+                msg = f'Available tools: {", ".join(f"{n!r}" for n in self.tools)}'
+            else:
+                msg = 'No tools available.'
+            error_msg = f'Unknown tool name: {name!r}. {msg}'
+
+            max_retries = self.default_max_retries
+            if self.ctx.retries.get(name, 0) == max_retries:
+                raise UnexpectedModelBehavior(f'Tool {name!r} exceeded max retries count of {max_retries}')
```
DouweM: This should use `_check_max_retries` right?

pydantic_ai_slim/pydantic_ai/_tool_manager.py
```diff
@@ -119,151 +144,263 @@ def get_tool_def(self, name: str) -> ToolDefinition | None:
...
+
+            validation_error = ToolRetryError(
+                _messages.RetryPromptPart(tool_name=name, content=error_msg, tool_call_id=call.tool_call_id)
```
DouweM: Could we move this inside the `try` below and raise a `ModelRetry`, so we go throug the `_wrap_error_as_retry` for this?

pydantic_ai_slim/pydantic_ai/_tool_manager.py
```diff
@@ -119,151 +144,263 @@ def get_tool_def(self, name: str) -> ToolDefinition | None:
...
-    async def _call_tool(
+    async def execute_tool_call(
         self,
-        call: ToolCallPart,
-        *,
-        allow_partial: bool,
-        wrap_validation_errors: bool,
-        approved: bool,
-        metadata: Any = None,
+        validated: ToolCallValidation[AgentDepsT],
```
DouweM: Minor, but let's call the new type `ValidatedToolCall`

pydantic_ai_slim/pydantic_ai/_tool_manager.py
```diff
@@ -119,151 +144,263 @@ def get_tool_def(self, name: str) -> ToolDefinition | None:
...
-                            tool_name=name,
-                            content=e.message,
-                            tool_call_id=call.tool_call_id,
-                        )
-                        e = ToolRetryError(m)
-                    else:
-                        assert_never(e)
-
-                if not allow_partial:
-                    # If we're validating partial arguments, we don't want to count this as a failed tool as it may still succeed once the full arguments are received.
```
DouweM, line 239: We lost this comment!

pydantic_ai_slim/pydantic_ai/_tool_manager.py
```diff
@@ -119,151 +144,263 @@ def get_tool_def(self, name: str) -> ToolDefinition | None:
...
         instrumentation_names = InstrumentationNames.for_version(instrumentation_version)
+        call = validated.call
 
         span_attributes = {
             'gen_ai.tool.name': call.tool_name,
-            # NOTE: this means `gen_ai.tool.call.id` will be included even if it was generated by pydantic-ai
             'gen_ai.tool.call.id': call.tool_call_id,
             **({instrumentation_names.tool_arguments_attr: call.args_as_json_str()} if include_content else {}),
             'logfire.msg': f'running tool: {call.tool_name}',
-            # add the JSON schema so these attributes are formatted nicely in Logfire
```
DouweM, line 266: Lost both of these comments

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -875,6 +875,24 @@ def build_validation_context(
         return validation_ctx
 
 
+def _skip_output_tool(
```
DouweM: This name makes it seem like this skips it, rather than emitting events for a skipped call

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -913,59 +931,80 @@ async def process_tool_calls(  # noqa: C901
...
-                    )
-                    output_parts.append(part)
-                    yield _messages.FunctionToolResultEvent(part)
-                else:
-                    # No valid result yet, so this is a real failure
-                    ctx.state.increment_retries(
-                        ctx.deps.max_result_retries, error=e, model_settings=ctx.deps.model_settings
-                    )
-                    raise e  # pragma: lax no cover
+                    for event in _skip_output_tool(
```
DouweM: It warrants a comment why this doesn't use the very similar logic below

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -913,59 +931,80 @@ async def process_tool_calls(  # noqa: C901
...
+                result_data = await tool_manager.execute_tool_call(validated)
+            except exceptions.UnexpectedModelBehavior as e:
+                if final_result:
+                    for event in _skip_output_tool(
+                        call, 'Output tool not used - tool execution failed.', output_parts, args_valid=True
+                    ):
+                        yield event
+                    continue
+                ctx.state.increment_retries(
+                    ctx.deps.max_result_retries, error=e, model_settings=ctx.deps.model_settings
```
DouweM, line 998: This is duplicated with the `UnexpectedModelBehavior` branch above, I wonder if it could be a single bigger `try/except`, with the message depending on whether `validated and validated.args_valid`

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -913,59 +931,80 @@ async def process_tool_calls(  # noqa: C901
...
         # Early strategy is chosen and final result is not yet set
         # Or exhaustive strategy is chosen
         else:
+            # Validate and execute the output tool call
             try:
-                result_data = await tool_manager.handle_call(call)
+                validated = await tool_manager.validate_tool_call(call)
             except exceptions.UnexpectedModelBehavior as e:
-                # If we already have a valid final result, don't fail the entire run
-                # This allows exhaustive strategy to complete successfully when at least one output tool is valid
```
DouweM, line 939: We lost some valuable comments here as well

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -913,59 +931,80 @@ async def process_tool_calls(  # noqa: C901
...
                 if final_result:
-                    # If output tool fails when we already have a final result, skip it without retrying
```
DouweM, line 941: And here

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -913,59 +931,80 @@ async def process_tool_calls(  # noqa: C901
...
+                    ):
+                        yield event
+                    continue
+                ctx.state.increment_retries(
+                    ctx.deps.max_result_retries, error=e, model_settings=ctx.deps.model_settings
+                )
+                raise  # pragma: lax no cover
             except ToolRetryError as e:
-                # If we already have a valid final result, don't increment retries for invalid output tools
-                # This allows the run to succeed if at least one output tool returned a valid result
```
DouweM, line 958: And here :)

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -913,59 +931,80 @@ async def process_tool_calls(  # noqa: C901
...
-            else:
-                part = _messages.ToolReturnPart(
-                    tool_name=call.tool_name,
-                    content='Final result processed.',
-                    tool_call_id=call.tool_call_id,
-                )
-                output_parts.append(part)
+                continue
 
-                # In both `early` and `exhaustive` modes, use the first output tool's result as the final result
```
DouweM, line 974: And this one :)

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -1083,8 +1139,20 @@ async def _call_tools(  # noqa: C901
         projected_usage.tool_calls += len(tool_calls)
         usage_limits.check_before_tool_call(projected_usage)
 
+    # Validate upfront and cache results; skip calls with deferred results as those
+    # will be validated in _call_tool via handle_call with the correct approval context.
```
DouweM: Would it make sense to move this validation into `process_tool_calls` as well, so that all the validations are there and the executions here?

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -1198,14 +1278,17 @@ async def _call_tool(
     tool_call: _messages.ToolCallPart,
     tool_call_result: DeferredToolResult | None,
     tool_call_metadata: dict[str, dict[str, Any]] | None,
+    validated: ToolCallValidation[DepsT] | None = None,
```
DouweM: Could we pass this in as `tool_call: ValidatedToolCall | ToolCallPart`, instead of a new arg?

tests/test_streaming.py
```diff
@@ -3302,3 +3325,385 @@ async def test_get_output_after_stream_output():
...
+    # First call: validation fails -> args_valid=False; retry: validation passes -> args_valid=True
+    args_valid_values = [e.args_valid for e in tool_call_events]
+    assert args_valid_values[0] is False
+    assert args_valid_values[1] is True
+
+    # Verify retry prompt contains the validation error message
+    result_events = [e for e in events if isinstance(e, FunctionToolResultEvent)]
+    retry_results = [e for e in result_events if isinstance(e.result, RetryPromptPart)]
+    assert retry_results
+    assert 'x must be positive' in str(retry_results[0].result.content)
```
DouweM: I'd like a `assert events == snapshot(...)` so we can see the full picture of what's emitted.

tests/test_streaming.py
```diff
@@ -3302,3 +3325,385 @@ async def test_get_output_after_stream_output():
...
+
+async def test_external_tool_validation_failure():
+    """Test that external tool validation failure is handled correctly."""
+    from dataclasses import replace as dataclass_replace
+
+    from pydantic_core import SchemaValidator, core_schema
+
+    from pydantic_ai._output import DeferredToolRequests
+    from pydantic_ai._run_context import RunContext as RC
+    from pydantic_ai.toolsets.abstract import AbstractToolset, ToolsetTool
```
DouweM: Imports at the top only please!

DouweM: Imports at the top only please!

tests/test_streaming.py
```diff
@@ -3302,3 +3325,385 @@ async def test_get_output_after_stream_output():
...
+
+    # Create a schema validator that requires 'value' field of type string
+    strict_schema = core_schema.typed_dict_schema(
+        {
+            'value': core_schema.typed_dict_field(core_schema.str_schema(), required=True),
+        },
+    )
+    STRICT_VALIDATOR = SchemaValidator(schema=strict_schema)
+
+    class StrictExternalToolset(AbstractToolset[None]):
```
DouweM: Do we need a custom external toolset for this test, or could we test it with a tool with `approval_required=True`?

tests/test_streaming.py
```diff
@@ -3302,3 +3325,385 @@ async def test_get_output_after_stream_output():
...
+    messages = result.all_messages()
+    events: list[Any] = []
+    async for event in agent.run_stream_events(
+        message_history=messages,
+        deferred_tool_results=DeferredToolResults(approvals={tool_call_id: ToolApproved()}),
+        deps=42,
+    ):
+        events.append(event)
+
+    # The FunctionToolCallEvent for the pre-supplied result should have args_valid=None
```
DouweM: Hmm, shouldn't it be `True` because we would've validated the args before?

tests/test_tools.py
```diff
@@ -2983,3 +2983,382 @@ def my_tool(ctx: RunContext[None], x: int) -> int:
...
+
+    # First attempt: retry=0, raises ModelRetry; Second attempt: retry=1, succeeds
+    assert retry_values == [0, 1]
+
+
+def test_args_validator_not_double_called_for_approved_tools():
+    """Test that args_validator is called exactly once for ToolApproved deferred tool calls.
+
+    When a tool requires approval and is later re-run with ToolApproved, the validator
+    should only be called once (during handle_call with approved=True), not twice.
```
DouweM: This is incorrect right? It would've run during the original handle call before there was approval

tests/test_streaming.py
```diff
@@ -2098,7 +2098,7 @@ async def stream_function(_: list[ModelMessage], info: AgentInfo) -> AsyncIterat
                         ),
                         ToolReturnPart(
                             tool_name='second_output',
-                            content='Output tool not used - output failed validation.',
+                            content='Output tool not used - tool execution failed.',
```
DouweM: Technically it's "output function execution", right?

pydantic_ai_slim/pydantic_ai/tools.py
jerry-reevo: @DouweM The issue is that I think we'd have to make `Tool` Generic over `ToolParams` for this to properly work. An alternative is to make the signatures match in the `__init__` so type consistency is enforced at entry point. I've added a `typed_agent.py` test for this.

pydantic_ai_slim/pydantic_ai/_tool_manager.py
jerry-reevo: @DouweM That was the original name but a Claude review pointed out it can be confusing because it can carry tool call info that was invalid. I don't mind changing, just wanted to get your confirmation.

pydantic_ai_slim/pydantic_ai/_tool_manager.py
jerry-reevo: Context: `_check_max_retries` takes an exception that we haven't caught. I've switched to `_check_max_retries` by constructing an exception for it.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -1041,11 +1080,28 @@ async def process_tool_calls(  # noqa: C901
...
+                    else:
+                        deferred_calls['unapproved'].append(call)
+                else:
+                    # Call execute_tool_call to raise the validation error inside a trace span;
+                    # retries are already tracked by validate_tool_call() via failed_tools.
+                    try:
+                        await tool_manager.execute_tool_call(validated)
+                    except ToolRetryError as e:
+                        output_parts.append(e.tool_retry)
+                        yield _messages.FunctionToolResultEvent(e.tool_retry)
```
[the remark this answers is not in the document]
jerry-reevo: This was intentional

tests/test_tools.py
jerry-reevo: The validation should run once before deferral, and again when re-run after approval. Before is needed to ensure validated args are presented for approval, after is needed to ensure and user overrides are validated.

tests/test_streaming.py
jerry-reevo: Yah I think that's more correct and clear. Updated it upstream and in relevant tests.

tests/test_streaming.py
jerry-reevo: Youre right. We also validate again when executing since users can modify the args. Updated to correctly reflect this.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
jerry-reevo: Yes I think so. It better locates related logic, but the `process_tool_calls` is already pretty bloated. Moving forward with suggested refactor.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
jerry-reevo: I tried but I think it increases complexity and makes exception origin and reason a little less clear. I am favoring leaving it like this.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -916,59 +934,87 @@ async def process_tool_calls(  # noqa: C901
...
+                if final_result:
+                    for event in _emit_skipped_output_tool(
+                        call, 'Output tool not used - output function execution failed.', output_parts, args_valid=True
+                    ):
+                        yield event
+                    continue
+                ctx.state.increment_retries(
+                    ctx.deps.max_result_retries, error=e, model_settings=ctx.deps.model_settings
+                )
+                raise  # pragma: lax no cover
```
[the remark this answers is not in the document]
jerry-reevo: I still would elect to keep this separate.

pydantic_ai_slim/pydantic_ai/tools.py
```diff
@@ -71,6 +72,17 @@
...
+    Callable[Concatenate[RunContext[AgentDepsT], ToolParams], Awaitable[None]]
+    | Callable[Concatenate[RunContext[AgentDepsT], ToolParams], None]
+)
+"""A function that validates tool arguments before execution.
+
+The validator receives the same typed parameters as the tool function,
+with [`RunContext`][pydantic_ai.tools.RunContext] as the first argument for dependency access.
+
+Should raise [`ModelRetry`][pydantic_ai.exceptions.ModelRetry] on validation failure.
+"""
```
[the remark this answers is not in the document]
jerry-reevo: Yes I think making `Tool` generic over `ToolParams` is currently out of scope and potentially something I can help with in a follow up.

PR #4022, opened 2026-01-16
Support `WebSearchTool` for `OpenRouterModel` through OpenRouter plugins

### LINKED ISSUE #3952: openrouter web search builtin tools support
### Description

Hi, can we support web search builtin tool for openrouter?

### References

https://openrouter.ai/docs/guides/features/plugins/web-search

mochow13:
- Closes #3952 

### Pre-Review Checklist

### Pre-Merge Checklist

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/openrouter.py
+++ b/pydantic_ai_slim/pydantic_ai/models/openrouter.py
@@ -574,7 +574,6 @@ class OpenRouterModel(OpenAIChatModel):
 
     @override
     def prepare_request(
-    def prepare_request(
         self,
         model_settings: ModelSettings | None,
         model_request_parameters: ModelRequestParameters,


docs/models/openrouter.md
```diff
@@ -73,3 +73,35 @@ model = OpenRouterModel('openai/gpt-5')
...
+
+## Web Search
+
+OpenRouter supports web search via its [plugins](https://openrouter.ai/docs/guides/features/plugins/web-search). You can enable it using the [`WebSearchTool`][pydantic_ai.builtin_tools.WebSearchTool].
+
+### Web Search Parameters
+
+You can customize the web search behavior using parameters on [`WebSearchTool`][pydantic_ai.builtin_tools.WebSearchTool]:
+
+```python {test="skip"}
```
mochow13: @DouweM I didn't intend to put `skip` here but I miserably failed to make this work in `test_examples.py`.

```
pydantic_ai.exceptions.UserError: Builtin tool(s) ['WebSearchTool'] not supported by this model. Supported: []
```

I did walk through the code step by step but couldn't reason why supported builtin tool is empty. Added an override for `supported_builtin_tools` in `models/openrouter.py` which in my understanding is needed but the test doesn't pass anyway.

dsfaccini: it's possible whatever `supported_builtin_tools` setting is used by the `OpenRouterModel` is being overridden by the model's own profile, can you check that please? in that case you'd need to deal with that logic in `profiles/openrouter.py`

mochow13: Hi @dsfaccini 
> by the model's own profile

do you mean the underlying model? For example, if I am using `openai/gpt-4o` then `OpenAI`'s own profile might be overriding the `OpenRouter` setting?

mochow13: @dsfaccini Hi, I have updated the MR. Looks like all I needed is to set `openai_chat_supports_web_search=True`. Also the pipeline is now green :')

dsfaccini:

```bash
git pull

deactivate

rm -rf .venv

make install
```

pydantic_ai_slim/pydantic_ai/builtin_tools.py
```diff
@@ -148,6 +150,35 @@ class WebSearchTool(AbstractBuiltinTool):
     * Anthropic
     """
 
+    engine: Literal['native', 'exa'] | None = None
```
DouweM: See https://github.com/pydantic/pydantic-ai/issues/4115#issuecomment-3807168810 on how we should support fields supported by only a single provider.

pydantic_ai_slim/pydantic_ai/models/openrouter.py
```diff
@@ -511,6 +558,14 @@ def _openrouter_settings_to_openai_settings(model_settings: OpenRouterModelSetti
     if usage := model_settings.pop('openrouter_usage', None):
         extra_body['usage'] = usage
 
+    if web_search_plugin is not None:
+        plugins = extra_body.get('plugins', [])
+        plugins.append(web_search_plugin)
```
DouweM: We shouldn't modify model settings in place, because we'd add a new item every time the model is used.

pydantic_ai_slim/pydantic_ai/models/openrouter.py
```diff
@@ -537,14 +592,52 @@ def __init__(
...
         super().__init__(model_name, provider=provider or OpenRouterProvider(), profile=profile, settings=settings)
 
+    @classmethod
+    @override
+    def supported_builtin_tools(cls) -> frozenset[type[AbstractBuiltinTool]]:
+        """Return the set of builtin tool types this model can handle.
+
+        OpenRouter supports web search via its plugins system.
+        """
+        return frozenset({WebSearchTool})
```
DouweM, line 573: This will conflict with https://github.com/pydantic/pydantic-ai/pull/3599

pydantic_ai_slim/pydantic_ai/models/openrouter.py
```diff
@@ -537,14 +592,52 @@ def __init__(
...
+                web_search_plugin['max_results'] = web_search_tool.max_results
+            if web_search_tool.search_prompt is not None:
+                web_search_plugin['search_prompt'] = web_search_tool.search_prompt
+
+            web_search_options = {'search_context_size': web_search_tool.search_context_size}
+
+            filtered_builtin_tools = [
+                t for t in model_request_parameters.builtin_tools if not isinstance(t, WebSearchTool)
+            ]
+            model_request_parameters = replace(model_request_parameters, builtin_tools=filtered_builtin_tools)
```
DouweM: Why do we filter?

DouweM: I don't think we should do this builtin tools parsing here in `prepare_request`, but rather when we build the payload for the request.

pydantic_ai_slim/pydantic_ai/models/openrouter.py
mochow13: But can be resolved by the one that's merged later?

pydantic_ai_slim/pydantic_ai/models/openrouter.py
mochow13: Well I thought this is the place where the request is built to send to OpenRouter. Where is it more suitable?

pydantic_ai_slim/pydantic_ai/builtin_tools.py
mochow13: So it's the number 3 you prefer:

> A provider_metadata: dict[ProviderName, dict[str, Any] would let you specify FileSearchTool(..., provider_metadata={'google': {'metadata_filter': '...'}, 'openai': {...}}, which solves those problems.

And is the idea to provide this `provider_metadata` inside `WebSearchTool`?

DouweM: @mochow13 Yep exacrtly

pydantic_ai_slim/pydantic_ai/models/openrouter.py
DouweM: Yep that was mostly an FYI

pydantic_ai_slim/pydantic_ai/models/openrouter.py
DouweM: Fair! Though I prefer the approach in https://github.com/pydantic/pydantic-ai/pull/3599/files where the `model_request_parameters` are passed into `_openrouter_settings_to_openai_settings`.

And I don't think it should be necessary to filter out unsupported tools at all, as the `super().prepare_request` already validates against `supported_builtin_tools`, and the user _should_ get an error if they select a tool that's not supported.

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/models/openrouter.py
mochow13: Well thinking about this, I am a bit confused :D I see `extra_body` is also modified in-place. Is this not the same?

Edit: Made an update to avoid in-place update.

pydantic_ai_slim/pydantic_ai/models/openrouter.py
mochow13: @DouweM I have updated the approach. Also there are changes added based on some feedback from Devin.

Perhaps a fresh look from you would be required! 🙇‍♂️

mochow13:
@DouweM @dsfaccini This MR is waiting for somewhat longer than usual! Could any of you help me land it to `main`?

mochow13:
Hello @dsfaccini, I see you have marked the MR as auto-review. But there are comments from the bots hanging that I don't think applicable and/or useful. Could it be reviewed by a human for the last mile? :)

pydantic_ai_slim/pydantic_ai/builtin_tools.py
```diff
@@ -35,6 +35,26 @@
 """Supported aspect ratios for image generation tools."""
 
 
+class WebSearchToolOpenRouterMetadata(TypedDict, total=False):
```
DouweM: @mochow13 Unless you have an immediate need for these OpenRouter-specific options, I'd prefer to remove them from this PR so we can land basic support immediately, without having to get confident about a new pattern for provider-specific builtin tool options. Note that OpenAI, Anthropic and Google also have provider-specific web search options that we don't currently expose. If you drop these types and the new metadata field, the PR will get a lot smaller, easier to review/merge, and still good enough for most uses.

mochow13:
@DouweM I have updated the MR—made it simpler as you suggested.

Question on AI review: any specific reason to have both Devin and Github review? More often than not, they are suggesting the same thing, and also not always useful. Two bots felt a little exhausting.

DouweM:
> I have updated the MR—made it simpler as you suggested.

@mochow13 Thanks, looks good!

> 
> Question on AI review: any specific reason to have both Devin and Github review? More often than not, they are suggesting the same thing, and also not always useful. Two bots felt a little exhausting.

Devin primarily looks for bugs and edge cases but doesn't have much Pydantic AI-specific knowledge, while the `auto-review` bot is specifically "trained" (see https://pydantic.dev/articles/scaling-open-source-with-ai) to act as more of a co-maintainer with Pydantic AI-specific knowledge. But I agree there's some overlap in what they report which seems to be increasing as Devin gets better...

PR #4023, opened 2026-01-17
Make `count_tokens()` work on `TemporalModel`

DouweM:
Created by Claude Code

TODO: Actually test that UsageLimits works as expected with a real model request


DouweM:
Will need to be re-explored after #4977

PR #4027, opened 2026-01-18
feat(tools): add optional JSON repair for malformed tool arguments

### LINKED ISSUE #3008: Allow raw tool call arguments to be pre-processed in case schema validation fails
### Question

When I set the output type of an agent to a TypedDict, sometimes the AI returns incorrect JSON (see attached screenshot).

Image

I’d like to know if there’s a way to preprocess or adjust the agent’s response after the AI generates it but before it’s passed to the final_result tool.
* Does output_validator handle this kind of preprocessing?
* Or do I need to use stream_responses and validate_response_output manually to intercept and fix responses?

### Additional Context

_No response_

mpfaffenberger:
When LLMs generate tool calls, they sometimes produce malformed JSON
    (missing braces, trailing commas, single quotes, etc.). This change
    adds optional integration with fast-json-repair to automatically fix
    these issues before validation.

    Changes:
    - Add _maybe_repair_json() to _tool_manager.py that repairs JSON when
      fast-json-repair is installed, otherwise passes through unchanged
    - Add 'json-repair' optional dependency group in pydantic-ai-slim
    - Include json-repair by default in the main pydantic-ai package
    - Add comprehensive TDD-style tests for various JSON malformations

    The feature gracefully degrades: if fast-json-repair is not installed,
    the original validation error behavior is preserved (allowing retry
    logic to kick in as usual).

    Note: fast-json-repair requires Python 3.11+, so the dependency has
    a version marker to handle Python 3.10 compatibility.

- Closes #3008 

- Note: this P/R was written entirely by [Code Puppy](https://github.com/mpfaffenberger/code_puppy) which is powered by Pydantic AI. WOOF!

### Pre-Review Checklist

### Pre-Merge Checklist

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/_tool_manager.py
+++ b/pydantic_ai_slim/pydantic_ai/_tool_manager.py
@@ -33,7 +33,7 @@ class ToolManager(Generic[AgentDepsT]):
     """The agent run context for a specific run step."""
     tools: dict[str, ToolsetTool[AgentDepsT]] | None = None
     """The cached tools for this run step."""
-    failed_tools: set[str] = field(default_factory=set)
+    failed_tools: set[str] = field(default_factory=set[str])
     """Names of tools that failed in this run step."""
     default_max_retries: int = 1
     """Default number of times to retry a tool"""
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -6,9 +6,9 @@
       "Bash(tree:*)",
       "Bash(grep:*)",
       "Bash(git log:*)",
-      "Bash(git show:*)",
       "Bash(git diff:*)",
       "Bash(git grep:*)",
+      "Bash(git show:*)",
       "Bash(git status:*)",
       "Bash(gh pr view:*)",
       "Bash(gh pr checks:*)",
@@ -20,6 +20,13 @@
       "Bash(gh run list:*)",
       "Bash(uv run:*)",
       "Bash(make:*)"
+    ],
+    "ask": [
+      "Bash(git add:*)",
+      "Bash(git checkout:*)",
+      "Bash(git commit:*)",
+      "Bash(git push:*)",
+      "Bash(gh pr create:*)"
     ]
   }
 }
--- a/examples/pydantic_ai_examples/ag_ui/api/agentic_generative_ui.py
+++ b/examples/pydantic_ai_examples/ag_ui/api/agentic_generative_ui.py
@@ -27,7 +27,9 @@ class Step(BaseModel):
 class Plan(BaseModel):
     """Represents a plan with multiple steps."""
 
-    steps: list[Step] = Field(default_factory=list, description='The steps in the plan')
+    steps: list[Step] = Field(
+        default_factory=list[Step], description='The steps in the plan'
+    )
 
 
 class JSONPatchOp(BaseModel):
--- a/examples/pydantic_ai_examples/ag_ui/api/shared_state.py
+++ b/examples/pydantic_ai_examples/ag_ui/api/shared_state.py
@@ -62,18 +62,18 @@ class Recipe(BaseModel):
         description='The skill level required for the recipe',
     )
     special_preferences: list[SpecialPreferences] = Field(
-        default_factory=list,
+        default_factory=list[SpecialPreferences],
         description='Any special preferences for the recipe',
     )
     cooking_time: CookingTime = Field(
         default=CookingTime.FIVE_MIN, description='The cooking time of the recipe'
     )
     ingredients: list[Ingredient] = Field(
-        default_factory=list,
+        default_factory=list[Ingredient],
         description='Ingredients for the recipe',
     )
     instructions: list[str] = Field(
-        default_factory=list, description='Instructions for the recipe'
+        default_factory=list[str], description='Instructions for the recipe'
     )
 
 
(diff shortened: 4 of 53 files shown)


mpfaffenberger:
Could a maintainer please add the `feature` label to this PR? Thanks! 🐶

mpfaffenberger:
@DouweM - I can't seem to figure out how to get that one failing test run to be stable, have you guys seen that before?

dsfaccini:
I think the issue is, sometimes CI gets assigned a runner that lacks a CPU instruction used by `llama_cpp` that crashes the python worker. I've tried solving this before but didn't land on any solution aside from simply turning the test completely off (`RUN_LLAMA_CPP_TESTS: false`) because it's a `Fatal Python exception` so it can't even be caught...

I hope if you do a new push it might solve itself

docs/install.md
```diff
@@ -67,6 +67,9 @@ pip/uv-add "pydantic-ai-slim[openai]"
 * `ag-ui` - installs [AG-UI Event Stream Protocol](ui/ag-ui.md) dependencies `ag-ui-protocol` [PyPI ↗](https://pypi.org/project/ag-ui-protocol){:target="_blank"} and `starlette` [PyPI ↗](https://pypi.org/project/starlette){:target="_blank"}
 * `dbos` - installs [DBOS Durable Execution](durable_execution/dbos.md) dependency `dbos` [PyPI ↗](https://pypi.org/project/dbos){:target="_blank"}
 * `prefect` - installs [Prefect Durable Execution](durable_execution/prefect.md) dependency `prefect` [PyPI ↗](https://pypi.org/project/prefect){:target="_blank"}
+* `retries` - installs HTTP retry utilities dependency `tenacity` [PyPI ↗](https://pypi.org/project/tenacity){:target="_blank"}
+* `temporal` - installs [Temporal Durable Execution](durable_execution/temporal.md) dependency `temporalio` [PyPI ↗](https://pypi.org/project/temporalio){:target="_blank"}
+* `json-repair` - installs automatic JSON repair for malformed tool arguments dependency `fast-json-repair` [PyPI ↗](https://pypi.org/project/fast-json-repair){:target="_blank"} (Python 3.11+ only)
```
DouweM: We need to document automatic JSON repairing on the advanced tool features page as well.

pydantic_ai_slim/pydantic_ai/_tool_manager.py
```diff
@@ -173,9 +189,23 @@ async def _call_tool(
...
-                    call.args or '{}', allow_partial=pyd_allow_partial, context=ctx.validation_context
-                )
+                args_str = call.args or '{}'
+                try:
+                    args_dict = validator.validate_json(
+                        args_str, allow_partial=pyd_allow_partial, context=ctx.validation_context
+                    )
+                except ValidationError:
+                    # Only attempt JSON repair on final (non-partial) calls,
+                    # as repairing partial JSON can interfere with streaming behavior
```
DouweM: How so? I can see the argument that we don't need to repair because Pydantic already has logic to accept incomplete JSON, but there may be some additional cases that `fast_json_repair` handles. And I don't _think_ it should interfere with streaming?

pydantic_ai_slim/pydantic_ai/_tool_manager.py
```diff
@@ -173,9 +189,23 @@ async def _call_tool(
...
+                    if allow_partial:
+                        raise
+                    repaired = _maybe_repair_json(args_str)
```
DouweM: We should also do this in `ObjectOutputProcessor.validate`, meaning we need a new helper used both here and there.

tests/test_json_repair.py
```diff
@@ -0,0 +1,166 @@
...
+    @toolset.tool
+    def add(a: int, b: int) -> int:
+        """Add two numbers."""
+        return a + b
+
+    context = build_run_context()
+    tool_manager = await ToolManager[None](toolset).for_run_step(context)
+
+    # Broken JSON: missing closing brace
+    broken_json = '{"a": 1, "b": 2'
```
DouweM: I'd prefer to test this at the agent level with a `FunctionModel` that returns a `ToolCallPart` with invalid args, and then validating that the agent returns the expected output without a model-retry round trip.

As I mentioned above, we should also make this work for `ObjectOutputProcessor`, which is used in native output mode, so that can also easily be tested using `FunctionModel`.

dsfaccini:
hey @mpfaffenberger you'll need to resolve the merge conflict by checking out the `uv.lock` from main and relocking your changes

note that you should delete your `.venv` and `make install` again so you don't keep outdated package versions in your dev environment

we've bumped pyright so you may have a couple new typecheck errors to fix, fyi

mpfaffenberger:
@DouweM @dsfaccini I goofed up syncing my fork :( - it was easier to just open a new PR #4111

pydantic_ai_slim/pydantic_ai/_tool_manager.py
mpfaffenberger: Ahhh yes... you are right to challenge this. I attempted to apply `json_repair` on the partial JSON, but it did give me errors. Maybe I'm doing it wrong?

Either way I think the goal is more about repairing the JSON to be valid when a model emits something silly. For example, what I saw commonly with GLM 4.7 is this:

```json
{
   args: {
     "reasoning": "I'm now going to invoke the edit_file tool so that I can update the code </think>
   }
}

It sometimes would throw a closing </think> tag in the middle of tool call args. Weird right?

pydantic_ai_slim/pydantic_ai/_tool_manager.py
mpfaffenberger: Will update!

PR #4032, opened 2026-01-19
fix(google): Set timeout in `HttpOptions` to prevent requests hanging indefinitely

### LINKED ISSUE #4031: GoogleProvider timeout is ineffective - google-genai SDK overrides httpx client timeout with None
## Description

When using `GoogleProvider`, API requests can hang indefinitely even though `pydantic_ai` configures a 600-second timeout on the httpx client. This is because the `google-genai` SDK explicitly passes `timeout=None` to httpx requests, which overrides the client-level timeout.

## Root Cause

1. `pydantic_ai` creates an httpx client with a 600s timeout via `cached_async_http_client()`:
   ```python
   # pydantic_ai/models/__init__.py:1170-1174
   httpx.AsyncClient(timeout=httpx.Timeout(timeout=600, connect=5))
   ```

2. `GoogleProvider` passes this client to `HttpOptions` but does **not** set the `timeout` parameter:
   ```python
   # pydantic_ai/providers/google.py:113-117
   http_options = HttpOptions(
       base_url=base_url,
       headers={'User-Agent': get_user_agent()},
       httpx_async_client=http_client,
       # timeout is not set, defaults to None
   )
   ```

3. `HttpOptions.timeout` defaults to `None` in google-genai (`google/genai/types.py`)

4. When google-genai makes requests, it explicitly passes `timeout=None` to httpx:
   ```python
   # google/genai/_api_client.py:1340-1346
   client_response = await self._async_httpx_client.request(
       method=http_request.method,
       url=http_request.url,
       headers=http_request.headers,
       content=data,
       timeout=http_request.timeout,  # This is None
   )
   ```

5. In httpx, explicitly passing `timeout=None` to a request **disables** the timeout entirely, overriding any client-level timeout.

## Reproduction

```python
import asyncio
from pydantic_ai import Agent

agent = Agent("google-vertex:gemini-2.0-flash")

async def main():
    # If the API hangs or is slow, this will wait forever
    # even though pydantic_ai intended a 600s timeout
    result = await agent.run("Hello, world!")
    print(result.output)

asyncio.run(main())
```

To verify the timeout is being overridden, you can add logging to see that `timeout=None` is passed:

```python
import asyncio
import httpx
from unittest.mock import patch
from pydantic_ai import Agent

agent = Agent("google-vertex:gemini-2.0-flash")

original_request = httpx.AsyncClient.request

async def logged_request(self, *args, **kwargs):
    print(f"httpx request called with timeout={kwargs.get('timeout')}")
    return await original_request(self, *args, **kwargs)

async def main():
    with patch.object(httpx.AsyncClient, 'request', logged_request):
        # Output: httpx request called with timeout=None

asyncio.run(main())
```

## Expected Behavior

API requests should respect the 600-second timeout configured by `pydantic_ai`, or allow users to configure a timeout that is actually enforced.

## Suggested Fix

Set `timeout` in `HttpOptions` when creating the `GoogleProvider`:

```python
# pydantic_ai/providers/google.py
http_options = HttpOptions(
    base_url=base_url,
    headers={'User-Agent': get_user_agent()},
    httpx_async_client=http_client,
    timeout=600_000,  # 600 seconds in milliseconds
)
```

Alternatively, this could be made configurable via a parameter to `GoogleProvider`.

## Related Issues

- googleapis/python-genai#1876 - `generate_content()` does not honor `httpx_client` override
- googleapis/python-genai#1893 - Requests hang indefinitely
- googleapis/python-genai#911 - Setting timeout in `genai.Client()` does not work

assafeisr:
## Summary

- Adds `DEFAULT_HTTP_TIMEOUT` constant (600 seconds) to `models/__init__.py`
- Uses this constant in `cached_async_http_client()` instead of hardcoded value
- Sets `timeout` in `HttpOptions` using the same constant (converted to milliseconds)

## Problem

The google-genai SDK's `HttpOptions.timeout` defaults to `None`. When making requests, the SDK passes this value to httpx:

```python
# google/genai/_api_client.py:1340-1346
client_response = await self._async_httpx_client.request(
    ...
    timeout=http_request.timeout,  # This is None
)
```

In httpx, explicitly passing `timeout=None` **overrides** any client-level timeout and disables timeouts entirely. This means the 600-second timeout configured by `cached_async_http_client()` is ignored, and requests can hang forever.

## Solution

1. Extract the timeout to a `DEFAULT_HTTP_TIMEOUT` constant
2. Use the constant in both `cached_async_http_client()` and `HttpOptions`
3. Convert from seconds to milliseconds for `HttpOptions` (google-genai expects milliseconds)

## Test plan

- [ ] Verify existing tests pass
- [ ] Manual testing: Confirm requests now respect the timeout instead of hanging indefinitely

Fixes #4031

### DIFF
--- a/tests/models/test_google.py
+++ b/tests/models/test_google.py
@@ -5470,3 +5470,125 @@ def test_google_provider_respects_custom_http_client_timeout(gemini_api_key: str
 
     http_options = provider._client._api_client._http_options  # pyright: ignore[reportPrivateUsage]
     assert http_options.timeout == custom_timeout * 1000
+
+
+async def test_google_splits_tool_return_from_user_prompt(google_provider: GoogleProvider):
+    """Test that ToolReturnPart and UserPromptPart are split into separate content objects.
+
+    TODO: Remove workaround when https://github.com/pydantic/pydantic-ai/issues/3763 is resolved
+    """
+    m = GoogleModel('gemini-2.5-flash', provider=google_provider)
+
+    # ToolReturn + UserPrompt
+    messages: list[ModelMessage] = [
+        ModelRequest(
+            parts=[
+                ToolReturnPart(tool_name='final_result', content='Final result processed.', tool_call_id='test_id'),
+                UserPromptPart(content="What's 2 + 2?"),
+            ]
+        )
+    ]
+
+    _, contents = await m._map_messages(messages, ModelRequestParameters())  # pyright: ignore[reportPrivateUsage]
+
+    assert contents == snapshot(
+        [
+            {
+                'role': 'user',
+                'parts': [
+                    {
+                        'function_response': {
+                            'name': 'final_result',
+                            'response': {'return_value': 'Final result processed.'},
+                            'id': 'test_id',
+                        }
+                    }
+                ],
+            },
+            {
+                'role': 'user',
+                'parts': [
+                    {
+                        'text': "What's 2 + 2?",
+                    }
+                ],
+            },
+        ]
+    )
+
+    # ToolReturn + Retry + UserPrompts
+    messages = [
+        ModelRequest(
+            parts=[
+                ToolReturnPart(tool_name='final_result', content='Final result processed.', tool_call_id='test_id_1'),
+                RetryPromptPart(content='Tool error occurred', tool_name='another_tool', tool_call_id='test_id_2'),
+                UserPromptPart(content="What's 2 + 2?"),
+                UserPromptPart(content="What's 3 + 3?"),
+            ]
+        )
+    ]
+
+    _, contents = await m._map_messages(messages, ModelRequestParameters())  # pyright: ignore[reportPrivateUsage]
+
+    assert contents == snapshot(
+        [
+            {
+                'role': 'user',
+                'parts': [
+                    {
+                        'function_response': {
+                            'name': 'final_result',
+                            'response': {'return_value': 'Final result processed.'},
+                            'id': 'test_id_1',
+                        }
+                    },
+                    {
+                        'function_response': {
+                            'name': 'another_tool',
+                            'response': {'error': 'Tool error occurred\n\nFix the errors and try again.'},
+                            'id': 'test_id_2',
+                        }
+                    },
+                ],
+            },
+            {
+                'role': 'user',
+                'parts': [
+                    {
+                        'text': "What's 2 + 2?",
+                    },
+                    {
+                        'text': "What's 3 + 3?",
+                    },
+                ],
+            },
+        ]
+    )
+
+    # ToolReturn only
+    messages = [
+        ModelRequest(
+            parts=[
+                ToolReturnPart(tool_name='final_result', content='Final result processed.', tool_call_id='test_id'),
+            ]
+        )
+    ]
+
+    _, contents = await m._map_messages(messages, ModelRequestParameters())  # pyright: ignore[reportPrivateUsage]
+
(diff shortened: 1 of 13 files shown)


pydantic_ai_slim/pydantic_ai/providers/google.py
```diff
@@ -110,10 +110,15 @@ def __init__(
...
             )
+            # Note: google-genai's HttpOptions.timeout defaults to None, which causes
+            # the SDK to explicitly pass timeout=None to httpx, overriding any timeout
+            # configured on the httpx client. We must set the timeout here to ensure
+            # requests actually time out. The value is converted from seconds to milliseconds.
             http_options = HttpOptions(
                 base_url=base_url,
                 headers={'User-Agent': get_user_agent()},
                 httpx_async_client=http_client,
+                timeout=DEFAULT_HTTP_TIMEOUT * 1000,
```
DouweM: Could we read the timeout off of `http_client`? As it stands, it looks like that field on a user's custom client would still be ignored.

DouweM (changes requested):
Add tests please

pydantic_ai_slim/pydantic_ai/providers/google.py
assafeisr: Added tests and implemented your suggestion, thanks.

pydantic_ai_slim/pydantic_ai/providers/google.py
```diff
@@ -110,10 +110,18 @@ def __init__(
             http_client = http_client or cached_async_http_client(
                 provider='google-vertex' if vertexai else 'google-gla'
             )
+            # Note: google-genai's HttpOptions.timeout defaults to None, which causes
+            # the SDK to explicitly pass timeout=None to httpx, overriding any timeout
+            # configured on the httpx client. We must set the timeout here to ensure
+            # requests actually time out. Read the timeout from the http_client if set,
+            # otherwise use the default. The value is converted from seconds to milliseconds.
+            timeout_seconds = http_client.timeout.read
```
DouweM: Could this have `or DEFAULT_HTTP_TIMEOUT`, and then we can drop the `if...` below?

tests/models/test_google.py
```diff
@@ -5444,3 +5444,71 @@ async def stream_iterator():
...
+        return mock_response
+
+    httpx_client = provider._client._api_client._async_httpx_client  # pyright: ignore[reportPrivateUsage]
+    mocker.patch.object(
+        httpx_client,
+        'request',
+        side_effect=capture_timeout_request,
+    )
+
+    await agent.run('Hello!')
```
DouweM: - Do we have to run the entire agent, or could we just call `GoogleModel.request`?
- Do we have to verify that timeout is passed to the request method, or could we "just" double check that e.g. `provider._client.http_options.timeout` is set? (That specific field probably won't work, but you get the idea). I'd prefer to do less mocking.

pydantic_ai_slim/pydantic_ai/providers/google.py
assafeisr: Done - simplified to `timeout_seconds = http_client.timeout.read or DEFAULT_HTTP_TIMEOUT`

tests/models/test_google.py
assafeisr: Good call - simplified the tests to just check `http_options.timeout` directly. No more agent runs or mocking needed.

PR #4037, opened 2026-01-19
feat(openai): add `allowed_domains` support for `WebSearchTool`

### LINKED ISSUE #4021: Add allowed_domains support for OpenAI Responses API WebSearchTool
## Summary
The OpenAI Responses API natively supports domain filtering via `filters.allowed_domains`, but pydantic-ai's `WebSearchTool` doesn't pass this through when using `OpenAIResponsesModel`.

## Use Case
We use Perplexity with domain filtering for specialized searches (academic sources, SEC filings). When Perplexity is unavailable, we fall back to `WebSearchTool`. Without `allowed_domains`, we lose the ability to constrain searches to specific domains like sec.gov or academic publishers.

Anthropic's `WebSearchTool` already supports `allowed_domains`, so parity with OpenAI would be valuable.

## Current Behavior
`_get_web_search_options` in `openai.py` only passes `search_context_size` and `user_location`:

```python
def _get_web_search_options(self, model_request_parameters: ModelRequestParameters) -> WebSearchOptions | None:
    for tool in model_request_parameters.builtin_tools:
        if isinstance(tool, WebSearchTool):
            if tool.user_location:
                return WebSearchOptions(
                    search_context_size=tool.search_context_size,
                    user_location=WebSearchOptionsUserLocation(...)
                )
            return WebSearchOptions(search_context_size=tool.search_context_size)
    return None
```

### This should work:
`WebSearchTool(allowed_domains=["sec.gov", "nature.com"])`

### And pass through to OpenAI as:
`filters: { allowed_domains: ["sec.gov", "nature.com"] }`

[Web Search Tool - Domain Filtering](https://platform.openai.com/docs/guides/tools-web-search#domain-filtering)

```python
tools=[{
    "type": "web_search_preview",
    "search_context_size": "medium",
    "user_location": {...},
    "filters": {
        "allowed_domains": ["sec.gov"]  # <-- not currently passed by pydantic-ai
    }
}]
```

## Proposed Change
Add `allowed_domains: list[str] | None = None` to `WebSearchTool`
Update `_get_web_search_options` to include `filters.allowed_domains` when set

Happy to submit a PR if this approach is acceptable.

### References

_No response_

eduardojdiniz:
## Summary

Wire up the existing `allowed_domains` parameter from WebSearchTool to the OpenAI Responses API via the `filters` parameter. This enables domain filtering for web searches when using OpenAI.

- Pass `allowed_domains` to `WebSearchToolParam` filters in `_get_builtin_tools`
- Update `allowed_domains` docstring to list OpenAI Responses as supported
- Update provider support table in docs
- Add unit test for `allowed_domains` parameter

Closes #4021

## Test plan

- [x] New unit test `test_openai_responses_web_search_tool_allowed_domains` passes
- [x] All existing `web_search` related tests pass (7/7)
- [x] Linting passes
- [x] Type checking passes for modified files

## Example usage

```python
from pydantic_ai import Agent
from pydantic_ai.builtin_tools import WebSearchTool

agent = Agent(
    model="openai:gpt-4o",
    builtin_tools=[WebSearchTool(allowed_domains=["sec.gov", "wikipedia.org"])],
)

result = await agent.run("Search for SEC filings")
```

### DIFF
--- a/tests/models/test_openai_responses.py
+++ b/tests/models/test_openai_responses.py
@@ -1719,7 +1719,8 @@ def test_model_profile_strict_not_supported():
 async def test_reasoning_model_with_temperature(allow_model_requests: None, openai_api_key: str):
     m = OpenAIResponsesModel('o3-mini', provider=OpenAIProvider(api_key=openai_api_key))
     agent = Agent(m, model_settings=OpenAIResponsesModelSettings(temperature=0.5))
-    result = await agent.run('What is the capital of Mexico?')
+    with pytest.warns(UserWarning, match='Sampling parameters.*temperature.*not supported when reasoning is enabled'):
+        result = await agent.run('What is the capital of Mexico?')
     assert result.output == snapshot(
         'The capital of Mexico is Mexico City. It serves as the political, cultural, and economic heart of the country and is one of the largest metropolitan areas in the world.'
     )
--- a/pydantic_ai_slim/pydantic_ai/models/openai.py
+++ b/pydantic_ai_slim/pydantic_ai/models/openai.py
@@ -57,7 +57,7 @@ from ..messages import (
     VideoUrl,
 )
 from ..profiles import ModelProfile, ModelProfileSpec
-from ..profiles.openai import OpenAIModelProfile, OpenAISystemPromptRole
+from ..profiles.openai import SAMPLING_PARAMS, OpenAIModelProfile, OpenAISystemPromptRole
 from ..providers import Provider, infer_provider
 from ..settings import ModelSettings
 from ..tools import ToolDefinition
@@ -261,6 +261,41 @@ def _check_azure_content_filter(e: APIStatusError, system: str, model_name: str)
         except ValidationError:
             pass
     return None
+
+
+def _drop_sampling_params_for_reasoning(profile: OpenAIModelProfile, model_settings: OpenAIChatModelSettings) -> None:
+    """Drop sampling params when reasoning is enabled on models that support it.
+
+    Reasoning models (o-series, GPT-5, GPT-5.1+) don't support sampling parameters when
+    reasoning is active. For models that support reasoning_effort='none' (GPT-5.1+),
+    sampling params are allowed when reasoning is off.
+    """
+    if not profile.openai_supports_reasoning:
+        return
+
+    reasoning_effort = model_settings.get('openai_reasoning_effort')
+    # On GPT-5.1+ models, 'none' is the default
+    if profile.openai_supports_reasoning_effort_none and reasoning_effort in (None, 'none'):
+        return
+
+    if dropped := [k for k in SAMPLING_PARAMS if k in model_settings]:
+        warnings.warn(
+            f'Sampling parameters {dropped} are not supported when reasoning is enabled. '
+            'These settings will be ignored.',
+            UserWarning,
+        )
+
+    for k in SAMPLING_PARAMS:
+        model_settings.pop(k, None)
+
+
+def _drop_unsupported_params(profile: OpenAIModelProfile, model_settings: OpenAIChatModelSettings) -> None:
+    """Drop unsupported parameters based on model profile.
+
+    Used currently only by Cerebras
+    """
+    for setting in profile.openai_unsupported_model_settings:
+        model_settings.pop(setting, None)
 
 
 class OpenAIChatModelSettings(ModelSettings, total=False):
@@ -614,12 +649,10 @@ class OpenAIChatModel(Model):
         tools = self._get_tools(model_request_parameters)
         web_search_options = self._get_web_search_options(model_request_parameters)
 
+        profile = OpenAIModelProfile.from_profile(self.profile)
         if not tools:
             tool_choice: Literal['none', 'required', 'auto'] | None = None
-        elif (
-            not model_request_parameters.allow_text_output
-            and OpenAIModelProfile.from_profile(self.profile).openai_supports_tool_choice_required
-        ):
+        elif not model_request_parameters.allow_text_output and profile.openai_supports_tool_choice_required:
             tool_choice = 'required'
         else:
             tool_choice = 'auto'
@@ -636,9 +669,9 @@ class OpenAIChatModel(Model):
         ):  # pragma: no branch
             response_format = {'type': 'json_object'}
 
-        unsupported_model_settings = OpenAIModelProfile.from_profile(self.profile).openai_unsupported_model_settings
-        for setting in unsupported_model_settings:
-            model_settings.pop(setting, None)
+        _drop_sampling_params_for_reasoning(profile, model_settings)
+
+        _drop_unsupported_params(profile, model_settings)
 
         try:
             extra_headers = model_settings.get('extra_headers', {})
@@ -1502,9 +1535,9 @@ class OpenAIResponsesModel(Model):
             text = text or {}
             text['verbosity'] = verbosity
 
-        unsupported_model_settings = profile.openai_unsupported_model_settings
-        for setting in unsupported_model_settings:
-            model_settings.pop(setting, None)
+        _drop_sampling_params_for_reasoning(profile, model_settings)
+
+        _drop_unsupported_params(profile, model_settings)
 
         include: list[responses.ResponseIncludable] = []
         if profile.openai_supports_encrypted_reasoning_content:
--- a/pydantic_ai_slim/pydantic_ai/profiles/openai.py
+++ b/pydantic_ai_slim/pydantic_ai/profiles/openai.py
@@ -10,6 +10,21 @@ from .._json_schema import JsonSchema, JsonSchemaTransformer
 from ..exceptions import UserError
 from . import ModelProfile
 
+SAMPLING_PARAMS = (
+    'temperature',
+    'top_p',
+    'presence_penalty',
+    'frequency_penalty',
+    'logit_bias',
+    'openai_logprobs',
+    'openai_top_logprobs',
+)
+"""Sampling parameter names that are incompatible with reasoning.
+
+These parameters are not supported when reasoning is enabled (reasoning_effort != 'none').
+See https://platform.openai.com/docs/guides/reasoning for details.
+"""
+
 OpenAISystemPromptRole = Literal['system', 'developer', 'user']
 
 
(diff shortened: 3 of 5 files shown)


tests/models/test_openai_responses.py
```diff
@@ -8964,3 +8964,36 @@ async def test_reasoning_summary_auto(allow_model_requests: None, openai_api_key
...
+
+
+async def test_openai_responses_web_search_tool_allowed_domains(allow_model_requests: None):
+    """Test that allowed_domains is properly passed to the OpenAI API via filters."""
+    c = response_message(
+        [
+            ResponseOutputMessage(
+                id='output-1',
+                content=cast(
+                    list[Content], [ResponseOutputText(text='Search result', type='output_text', annotations=[])]
```
dsfaccini: we're actually moving to having less (actually none) mock tests, opting instead for VCR tests. could you record on please? you could even modify whatever existing test there is for the webserachtool for openai responses and simply passing the `allowed_domains`, two birds... you'll need to rerecord anyway though, this is done using the `--record-mode=rewrite` flag when running the test

dsfaccini:
@eduardojdiniz thanks for working on this!

tests/models/test_openai_responses.py
eduardojdiniz: Happy to switch this to VCR. Just pushed a commit. I looked at the other tests, and they all use VCR already, except for `test_web_search_call_action_find_in_page`, which tests a specific edge case. This one is hard to reliably trigger the model into returning the exact response format, so I left it untouched.

eduardojdiniz: Out of curiosity, I’d love to hear a bit more about the decision to go full VCR/zero mocks. I haven't seen many AI libraries go 100% this direction yet. What specific pain points with mocks drove the decision, and what do you view as the biggest pros/cons now that you've settled into this workflow?

Any insights would be super helpful for my own learning (I am considering a similar shift for other repos I am working on)!

dsfaccini: well, mocks test that our internal logic is doing what we wanted to do, they don't test that what we actually send to the provider's is valid for them and doesn't showcase the behavior

having cassettes helps us both visualize what's happening at the HTTP level and validate that the providers actually accept it, plus that the LLM will largely do what we expect it to

they have the drawback -- which we want to address **soon**™ -- that we need to pick a cache key that's sensitive enough to prompt for rerecording if the payload changes in an important way, but not sensitive enough that random or minor variances will make the test suite fail in CI

so there's some work we need to do there but that'd be a step into a development cycle with increased confidence!

I personally see no pros vs cons of using mocks, VCRs are harder but better in all senses.

tests/models/cassettes/test_openai_responses/test_openai_responses_model_web_search_tool_with_allowed_domains.yaml
```diff
@@ -0,0 +1,131 @@
...
+          allowed_domains:
+          - wikipedia.org
+        search_context_size: medium
+        type: web_search
+        user_location:
+          city: null
+          country: US
+          region: null
+          timezone: null
+          type: approximate
```
dsfaccini, line 162: for instance, we'll want to leave this out of cassettes as to not accidentally dox our contributors

tests/models/test_openai_responses.py
```diff
@@ -997,6 +997,20 @@ async def test_openai_responses_model_web_search_tool_with_user_location(
...
+    agent = Agent(
+        m,
+        instructions='You are a helpful assistant.',
+        builtin_tools=[WebSearchTool(allowed_domains=['wikipedia.org'])],
+    )
+
+    result = await agent.run('What is the capital of France?')
+    assert result.output == snapshot('Paris.')
+
+
```
dsfaccini, line 1153: funnily enough the cassette doesn't seem to show that the LLM even used the `WebSearch` tool, i.e. we're not actually verifying that the LLM won't use a domain that is not whitelisted

if the openai API works as **they** expect that should always be the behavior though, and the fact that it replies does showcase that the params etc are alright

tests/models/cassettes/test_openai_responses/test_openai_responses_model_web_search_tool_with_allowed_domains.yaml
```diff
@@ -70,16 +70,80 @@ interactions:
...
+        summary: []
+        type: reasoning
+      - action:
+          queries:
+          - Tokyo Metropolis population Wikipedia
+          - Tokyo prefecture population Wikipedia
+          - Tokyo population 2024 Wikipedia
+          - Tokyo Metropolitan population statistics Wikipedia
+          query: Tokyo Metropolis population Wikipedia
+          type: search
```
dsfaccini, line 83: the new one does showcase it uses it :D

tests/models/test_openai_responses.py
```diff
@@ -997,6 +997,23 @@ async def test_openai_responses_model_web_search_tool_with_user_location(
...
+        m,
+        instructions='You are a helpful assistant.',
+        builtin_tools=[WebSearchTool(allowed_domains=['wikipedia.org'])],
+    )
+
+    result = await agent.run(
+        'What is the current population of Tokyo prefecture according to Wikipedia? Give me just the number.'
+    )
+    assert result.output
+    assert 'wikipedia.org' in result.output.lower()
```
dsfaccini: please replace with `result.output == snapshot()`

dsfaccini: note: you'll need to run the test to populate the snapshot

tests/models/cassettes/test_openai_responses/test_openai_responses_model_web_search_tool_with_allowed_domains.yaml
eduardojdiniz: Got it. I manually scrubbed some sensitive info. We could add the header scrubbing to the VCR config, but that would be out of scope for this PR

tests/models/test_openai_responses.py
eduardojdiniz: Changed to snapshot() :)

tests/models/cassettes/test_openai_responses/test_openai_responses_model_web_search_tool_with_allowed_domains.yaml
eduardojdiniz: Yes, my query was too generic on the first pass 😅

tests/models/test_openai_responses.py
eduardojdiniz: That makes sense - mocks test "did we call the SDK correctly" while VCR tests "did the SDK send the right HTTP request." Two different layers. Is the VCR stance for everything (e.g., a 429 response) or just for contracts?

I've seen some teams move away from VCR due to the maintenance burden and cassette drift.  How do you detect cassette drift? Periodic re-recording? CI job that hits real APIs? Re-recording at the scale of pydantic-ai probably requires rate limit management and manual review of diffs. What's the expected   
workflow when, say, OpenAI adds a new field to responses? 

Not arguing against VCR - the "validates what we actually send" point is compelling. Just trying to understand where these concerns land in pydantic-ai's testing philosophy.  Thanks for indulging me!

dsfaccini:
thanks @eduardojdiniz !

PR looks good to me, I'll let @DouweM take one more look and merge it

tests/models/test_openai_responses.py
```diff
@@ -997,6 +997,24 @@ async def test_openai_responses_model_web_search_tool_with_user_location(
...
+):
+    m = OpenAIResponsesModel('gpt-5', provider=OpenAIProvider(api_key=openai_api_key))
+    agent = Agent(
+        m,
+        instructions='You are a helpful assistant.',
+        builtin_tools=[WebSearchTool(allowed_domains=['wikipedia.org'])],
+    )
+
+    result = await agent.run(
+        'What is the current population of Tokyo prefecture according to Wikipedia? Give me just the number.'
```
DouweM: As it stands, this test doesn't actually validate that `allowed_domains` are passed along, or used at all. If we dropped "to Wikipedia", and the model chose to search Wikipedia because that was its only option, that'd be better. And/or ask it to search Google, and have it tell you it's not allowed to.

Either way, let's show the `assert result.all_messages() == snapshot()` so that we can verify that the web search tool was used at all.

eduardojdiniz: It all makes sense! I updated the test prompt to request a Google search instead of Wikipedia, regenerated the cassette, and updated the assert to all_messages

tests/models/test_openai_responses.py
```diff
@@ -1008,6 +1008,153 @@ async def test_openai_responses_model_web_search_tool_with_user_location(
...
...MzrSv-gvdCHt2q5DPNwVvSTDAU21qbjLEnEKsJU-FuQUr35hHcZ_ZT3U4_eQqu2Sb8fXwJQAD16xWggfaseZy5M_8Fv9XeGKpeXL5EFwlW-22fCPUsPREZtSVQFrEy64ysFJePcRqGCd_0NxZtCHndG_ySLvMuzEtle2lV6s0Yd2oFTaKYFDavEBKG4dT0ft1YFgQwdL3tefjMgvyhZX0diXpZa1MBDqFNKtWYtYbMDjucb960oLP0VvywA3oTBEHQy_CnMhct4ZRBtFEcUFp9-YPS7qc_Fnk7kXC5fZ6JXP7CNrBos3MlJrWkiGEeXvV2fZJqB-A5c7-rCGCB2rPO0RfZLMygRd3BvVYup-YWdHzyo58hNiH8FXOlqGDoMOtw0jbpGs7LprThKHvyooaucT0hec-8ENXFy-xDVZbO89NYptDTNK0nX-Y2N49IUQIRUl1kelobQg0WCQGrloyGdtr-MPpkk8Ge2FS7TZ32qf5BxHEw-JTFaNJFFRN2qnacKL9U1VvvYp_EuJFDUCOkSLJvAH552bDH61opbcGCrQDxr-nr4H8FKiqxmiqsNK94MligNgtFgE813H02hDnmHpxjp-8HFL40nAHJJzl6a3K_qPF5Sdr6-OqhujA_zwob5kgpgZ_L94Btag30G3t1iYX6zfK0Ab7Bql1Uy9c-azfD-hwSma3iFfFxxymb9dNNxpCXa1mbmw1Fr-IEWyFjZaoC_DGIT7iDrlRqiYjpJjcEuY8SPY5abmdFaDDuG96mAUKPPP2DY-SO3d6XWVT1Ni1KomYnLmq9Pd24N61q5b-W6owPcdrNh_zDn81zJwU91USm8PnHkeG7fVY0LzNs0ZQfwfq4l-9zxcJ9BL2cT7hbp0MeJy-7AzKhr0H2swUfYvFjXKjL4DrElqr7ZHwD_PSubL1Z_NIognp7pQonFA4gWJUOGR3Q32TSrjPrMdF3wCCfmm68Q_jH3_LeswHFcjMqjohn942uz43BairbW0jIqVYAnehAIFMdEwkUPyCxpDhkHc4Cg88GHCbCG2_LRGtFWGBtOgZrw25VwRAsX9Td2DA24Lw5oLdHnRkjdlN0Fmv7EpYAlzBn2XAYLr0bWDeHch2vc9RlOcRMaeOkbk9zEfiEAb4A88rklLXqOK2BgF7XtOWH-VjW_QesrJcQ_SCO4gSuAtxfXf_WDDfrQQ==',
```
dsfaccini: can you replace with `IsStr()`? same for all other instances below, that way we keep test files shorter

eduardojdiniz: Done! I changed those:
  - ThinkingPart.id and signature fields → IsStr()                              
  - tool_call_id values in BuiltinToolCallPart and BuiltinToolReturnPart →      
  IsStr()                                                                           
  - TextPart.id → IsStr()                                                           
  - provider_response_id → IsStr()
  Let me know if I should revert any of these

PR #4043, opened 2026-01-19
Fix serialization of `BinaryContent` returned by tool in Temporal

### LINKED ISSUE #3702: Temporal: Binary tool results fail serialization

dsfaccini:
- Closes #3702

### Pre-Review Checklist

### Pre-Merge Checklist

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/messages.py
+++ b/pydantic_ai_slim/pydantic_ai/messages.py
@@ -622,12 +622,33 @@ class BinaryContent:
     __repr__ = _utils.dataclasses_no_defaults_repr
 
 
+@pydantic_dataclass(
+    repr=False,
+    config=pydantic.ConfigDict(
+        ser_json_bytes='base64',
+        val_json_bytes='base64',
+    ),
+)
 class BinaryImage(BinaryContent):
     """Binary content that's guaranteed to be an image."""
 
+    # `pydantic_dataclass` replaces `__init__` so this method is never used.
+    # The signature is kept so that pyright/IDE hints recognize the `identifier` alias for the `_identifier` field.
+    def __init__(
+        self,
+        data: bytes,
+        *,
+        media_type: ImageMediaType | str,
+        identifier: str | None = None,
+        vendor_metadata: dict[str, Any] | None = None,
+        kind: Literal['binary'] = 'binary',
+        # Required for inline-snapshot which expects all dataclass `__init__` methods to take all field names as kwargs.
+        _identifier: str | None = None,
+    ) -> None: ...  # pragma: no cover
+
     def __post_init__(self):
         if not self.is_image:
-            raise ValueError('`BinaryImage` must be have a media type that starts with "image/"')  # pragma: no cover
+            raise ValueError('`BinaryImage` must have a media type that starts with "image/"')
 
 
 @dataclass
--- a/tests/models/test_google.py
+++ b/tests/models/test_google.py
@@ -75,7 +75,7 @@ from pydantic_ai.output import NativeOutput, PromptedOutput, TextOutput, ToolOut
 from pydantic_ai.settings import ModelSettings
 from pydantic_ai.usage import RequestUsage, RunUsage, UsageLimits
 
-from ..conftest import IsBytes, IsDatetime, IsInstance, IsNow, IsStr, try_import
+from ..conftest import IsDatetime, IsInstance, IsNow, IsStr, try_import
 from ..parts_from_messages import part_types_from_messages
 
 with try_import() as imports_successful:
@@ -3269,7 +3269,7 @@ async def test_google_image_generation(allow_model_requests: None, google_provid
     result = await agent.run('Generate an image of an axolotl.')
     messages = result.all_messages()
 
-    assert result.output == snapshot(BinaryImage(data=IsBytes(), media_type='image/jpeg', _identifier='b6e95a'))
+    assert result.output == snapshot(IsInstance(BinaryImage))
     assert messages == snapshot(
         [
             ModelRequest(
@@ -3285,11 +3285,7 @@ async def test_google_image_generation(allow_model_requests: None, google_provid
             ModelResponse(
                 parts=[
                     FilePart(
-                        content=BinaryImage(
-                            data=IsBytes(),
-                            media_type='image/jpeg',
-                            _identifier='b6e95a',
-                        ),
+                        content=IsInstance(BinaryImage),
                         provider_name='google-gla',
                         provider_details={'thought_signature': IsStr()},
                     )
@@ -3312,7 +3308,7 @@ async def test_google_image_generation(allow_model_requests: None, google_provid
     )
 
     result = await agent.run('Now give it a sombrero.', message_history=messages)
-    assert result.output == snapshot(BinaryImage(data=IsBytes(), media_type='image/jpeg', _identifier='14bec0'))
+    assert result.output == snapshot(IsInstance(BinaryImage))
     assert result.new_messages() == snapshot(
         [
             ModelRequest(
@@ -3328,11 +3324,7 @@ async def test_google_image_generation(allow_model_requests: None, google_provid
             ModelResponse(
                 parts=[
                     FilePart(
-                        content=BinaryImage(
-                            data=IsBytes(),
-                            media_type='image/jpeg',
-                            _identifier='14bec0',
-                        ),
+                        content=IsInstance(BinaryImage),
                         provider_name='google-gla',
                         provider_details={'thought_signature': IsStr()},
                     )
@@ -3365,14 +3357,7 @@ async def test_google_image_generation_stream(allow_model_requests: None, google
     agent = Agent(m, output_type=BinaryImage)
 
     async with agent.run_stream('Generate an image of an axolotl') as result:
-        assert await result.get_output() == snapshot(
-            BinaryImage(
-                data=IsBytes(),
-                media_type='image/png',
-                _identifier='9ff9cc',
-                identifier='9ff9cc',
-            )
-        )
+        assert await result.get_output() == snapshot(IsInstance(BinaryImage))
 
     event_parts: list[Any] = []
     async with agent.iter(user_prompt='Generate an image of an axolotl.') as agent_run:
@@ -3383,14 +3368,7 @@ async def test_google_image_generation_stream(allow_model_requests: None, google
                         event_parts.append(event)
 
     assert agent_run.result is not None
-    assert agent_run.result.output == snapshot(
-        BinaryImage(
-            data=IsBytes(),
-            media_type='image/png',
-            _identifier='2af2a7',
-            identifier='2af2a7',
-        )
-    )
+    assert agent_run.result.output == snapshot(IsInstance(BinaryImage))
     assert agent_run.result.all_messages() == snapshot(
         [
             ModelRequest(
@@ -3406,14 +3384,7 @@ async def test_google_image_generation_stream(allow_model_requests: None, google
             ModelResponse(
                 parts=[
                     TextPart(content='Here you go! '),
-                    FilePart(
-                        content=BinaryImage(
-                            data=IsBytes(),
-                            media_type='image/png',
-                            _identifier='2af2a7',
-                            identifier='2af2a7',
-                        )
-                    ),
+                    FilePart(content=IsInstance(BinaryImage)),
                 ],
                 usage=RequestUsage(
                     input_tokens=10,
@@ -3438,13 +3409,7 @@ async def test_google_image_generation_stream(allow_model_requests: None, google
             PartEndEvent(index=0, part=TextPart(content='Here you go! '), next_part_kind='file'),
             PartStartEvent(
                 index=1,
-                part=FilePart(
-                    content=BinaryImage(
-                        data=IsBytes(),
-                        media_type='image/png',
-                        _identifier='2af2a7',
-                    )
-                ),
+                part=FilePart(content=IsInstance(BinaryImage)),
                 previous_part_kind='text',
             ),
             FinalResultEvent(tool_name=None, tool_call_id=None),
@@ -3488,12 +3453,7 @@ A little axolotl named Archie lived in a beautiful glass tank, but he always won
                         provider_details={'thought_signature': IsStr()},
                     ),
                     FilePart(
-                        content=BinaryImage(
-                            data=IsBytes(),
-                            media_type='image/jpeg',
-                            _identifier='00f2af',
-                            identifier=IsStr(),
-                        ),
+                        content=IsInstance(BinaryImage),
                         provider_name='google-gla',
                         provider_details={'thought_signature': IsStr()},
                     ),
@@ -3527,14 +3487,7 @@ async def test_google_image_or_text_output(allow_model_requests: None, google_pr
     )
 
     result = await agent.run('Generate an image of an axolotl.')
-    assert result.output == snapshot(
-        BinaryImage(
-            data=IsBytes(),
-            media_type='image/png',
-            _identifier='f82faf',
-            identifier='f82faf',
-        )
-    )
+    assert result.output == snapshot(IsInstance(BinaryImage))
 
 
 async def test_google_image_and_text_output(allow_model_requests: None, google_provider: GoogleProvider):
@@ -3545,16 +3498,7 @@ async def test_google_image_and_text_output(allow_model_requests: None, google_p
     assert result.output == snapshot(
         'Once, in a hidden cenote, lived an axolotl named Pip who loved to collect shiny pebbles. One day, Pip found a pebble that glowed, illuminating his entire underwater world with a soft, warm light. '
     )
-    assert result.response.files == snapshot(
-        [
-            BinaryImage(
-                data=IsBytes(),
-                media_type='image/png',
-                _identifier='67b12f',
-                identifier='67b12f',
-            )
-        ]
-    )
+    assert result.response.files == snapshot([IsInstance(BinaryImage)])
 
 
 async def test_google_image_generation_with_tool_output(allow_model_requests: None, google_provider: GoogleProvider):
@@ -3600,11 +3544,7 @@ async def test_google_image_generation_with_native_output(allow_model_requests: 
             ModelResponse(
                 parts=[
                     FilePart(
-                        content=BinaryImage(
-                            data=IsBytes(),
-                            media_type='image/jpeg',
-                            _identifier='4e5b3e',
-                        ),
+                        content=IsInstance(BinaryImage),
                         provider_name='google-gla',
                         provider_details={'thought_signature': IsStr()},
                     )
@@ -3698,7 +3638,7 @@ async def test_google_image_generation_with_web_search(allow_model_requests: Non
     result = await agent.run(
         'Visualize the current weather forecast for the next 5 days in Mexico City as a clean, modern weather chart. Add a visual on what I should wear each day'
     )
-    assert result.output == snapshot(BinaryImage(data=IsBytes(), media_type='image/jpeg', _identifier='787c28'))
+    assert result.output == snapshot(IsInstance(BinaryImage))
     assert result.all_messages() == snapshot(
         [
             ModelRequest(
@@ -3743,11 +3683,7 @@ async def test_google_image_generation_with_web_search(allow_model_requests: Non
                         provider_name='google-gla',
                     ),
                     FilePart(
-                        content=BinaryImage(
-                            data=IsBytes(),
-                            media_type='image/jpeg',
-                            _identifier='787c28',
-                        ),
+                        content=IsInstance(BinaryImage),
                         provider_name='google-gla',
                         provider_details={'thought_signature': IsStr()},
                     ),
@@ -3960,9 +3896,7 @@ async def test_google_vertexai_image_generation(allow_model_requests: None, vert
     agent = Agent(model, output_type=BinaryImage)
 
     result = await agent.run('Generate an image of an axolotl.')
-    assert result.output == snapshot(
-        BinaryImage(data=IsBytes(), media_type='image/png', _identifier='b037a4', identifier='b037a4')
-    )
+    assert result.output == snapshot(IsInstance(BinaryImage))
 
 
 async def test_google_httpx_client_is_not_closed(allow_model_requests: None, gemini_api_key: str):
--- a/tests/models/test_openai_responses.py
+++ b/tests/models/test_openai_responses.py
@@ -52,7 +52,7 @@ from pydantic_ai.profiles.openai import openai_model_profile
 from pydantic_ai.tools import ToolDefinition
 from pydantic_ai.usage import RequestUsage, RunUsage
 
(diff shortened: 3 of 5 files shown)


pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_model.py
```diff
@@ -24,6 +25,21 @@
 from pydantic_ai.usage import RequestUsage
 
 from ._run_context import TemporalRunContext
+from ._toolset import rehydrate_binary_content
+
+
+def _rehydrate_messages(messages: list[ModelMessage]) -> list[ModelMessage]:
+    """Rehydrate `BinaryContent` in `ToolReturnPart.content` after Temporal deserialization.
```
DouweM: Could we skip this if we change the type of `ToolReturnPart.content` from `Any` to `BinaryContent | list[BinaryContent | Any] | Any`? That way Pydantic will know it should be trying to validate things as `BinaryContent`. We should also include `FileUrl`.

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -462,11 +465,22 @@ def format(self) -> DocumentFormat:
...
+
+@pydantic_dataclass(init=False, repr=False, config=pydantic.ConfigDict(arbitrary_types_allowed=True))
 class BinaryContent:
     """Binary content, e.g. an audio or image file."""
 
-    data: bytes
+    data: Annotated[
+        bytes,
+        WrapValidator(_validate_binary_data),
+        PlainSerializer(lambda v: base64.b64encode(v).decode('ascii'), when_used='json'),
```
DouweM: We shouldn't need to do any base64 stuff manually; Pydantic has features like `ser_json_bytes` and `val_json_bytes` for this.

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -462,7 +464,14 @@ def format(self) -> DocumentFormat:
...
 
-@dataclass(init=False, repr=False)
+@pydantic_dataclass(
+    init=False,
+    repr=False,
+    config=pydantic.ConfigDict(
+        ser_json_bytes='base64',
+        val_json_bytes='base64',
+    ),
+)
```
dsfaccini, line 474: worked like a charm, thanks for the suggestion

tests/test_temporal.py
```diff
@@ -3103,3 +3112,86 @@ def test_pydantic_ai_plugin_with_non_pydantic_converter_preserves_codec() -> Non
...
+    wrapped = _ToolReturn(result=bc)
+
+    # Serialize with to_json (what Temporal uses internally)
+    serialized = to_json(wrapped)
+
+    # Deserialize (simulating Temporal's deserialization)
+    deserialized = json.loads(serialized)
+
+    # The result field will be a dict after JSON deserialization
+    result = rehydrate_binary_content(deserialized['result'])
```
dsfaccini: this is a bit drawn out for my taste but I didn't find a better way to do it

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_toolset.py
```diff
@@ -63,6 +65,30 @@ class _ToolReturn:
...
+    """Recursively convert dicts with kind='binary' back to BinaryContent after Temporal deserialization.
+
+    When tool results contain BinaryContent (either directly or nested in dicts/lists),
+    Temporal deserializes them as plain dicts. This function walks the structure and
+    reconstructs BinaryContent objects where appropriate.
+    """
+    if isinstance(value, dict):
+        d = cast(dict[str, Any], value)
+        if d.get('kind') == 'binary' and 'data' in d and 'media_type' in d:
+            data = base64.b64decode(d['data']) if isinstance(d['data'], str) else d['data']
```
DouweM: Do we still need this now that we have the val/ser bytes stuff? I'd really like to get rid of this hard-coded mapping of the fields.

tests/test_temporal.py
```diff
@@ -3103,3 +3112,86 @@ def test_pydantic_ai_plugin_with_non_pydantic_converter_preserves_codec() -> Non
...
+            'kind': 'binary',
+            'identifier': '4caece',
+        }
+    )
+
+
+def test_binary_content_serialization_round_trip():
+    """Test full round-trip: BinaryContent -> to_json -> dict -> BinaryContent."""
+    # PNG magic bytes - non-UTF8 binary data that would fail direct JSON serialization
+    binary_data = bytes([0x89, 0x50, 0x4E, 0x47, 0x0D, 0x0A, 0x1A, 0x0A])
```
DouweM: This can be `b'\x89PNG\r\n\x1a\n'`, and then we can use the same var in the assert?

tests/test_temporal.py
```diff
@@ -3103,3 +3112,86 @@ def test_pydantic_ai_plugin_with_non_pydantic_converter_preserves_codec() -> Non
...
+    bc = BinaryContent(data=binary_data, media_type='image/png', identifier='test-file')
+
+    # Wrap in _ToolReturn like Temporal toolset does
+    wrapped = _ToolReturn(result=bc)
+
+    # Serialize with to_json (what Temporal uses internally)
+    serialized = to_json(wrapped)
```
DouweM: The fact that this does respect pydantic_dataclass serialization logic like val/ser bytes, suggests to me that we should be able to resolve https://github.com/pydantic/pydantic-ai/pull/3968 without moving away from `to_json` as well.

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_toolset.py
dsfaccini: I tried removing rehydration after the config fix and it didn't work, I think it was something having to do with nested things, because pydantic will handle `BinaryContent` but it won't assume `BinaryContent` when the type it's acting on is `dict[str, Any]`

I'll check again and leave an explanation explaining why more ind etail

tests/test_temporal.py
dsfaccini: TODO to be confirmed

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_toolset.py
dsfaccini: yeah so I confirmed this and left more of an explanation in the dosctring, but also added `TypeAdapter`s for the `FileUrl` subclasses, these will be needed if we want to apply the same nested fix to them

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_model.py
```diff
@@ -24,6 +25,21 @@
 from pydantic_ai.usage import RequestUsage
 
 from ._run_context import TemporalRunContext
+from ._toolset import rehydrate_binary_content
+
+
+def _rehydrate_messages(messages: list[ModelMessage]) -> list[ModelMessage]:
```
DouweM: Do we still need to rehydrate at all now that we have `content: BinaryContent | FileUrl | Sequence[BinaryContent | FileUrl | Any]`?

We may need to do that on `_ToolReturn` as well.

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_toolset.py
```diff
@@ -63,6 +64,42 @@ class _ToolReturn:
...
+}
+
+
+def rehydrate_binary_content(value: Any) -> Any:
+    """Recursively reconstruct `BinaryContent`/`FileUrl` objects from dicts after Temporal deserialization.
+
+    Why this is needed:
+    While `_ToolReturn.result` is typed as `BinaryContent | FileUrl | ...`, Pydantic only validates
+    top-level and list items against this union. When BinaryContent/FileUrl is nested inside a
+    `dict[str, Any]`, Pydantic doesn't recurse into the dict values to reconstruct typed objects.
```
DouweM: Then should `Any` in `dict[str, Any]` also be `BinaryContent | ...`? Using a recursive definition that always forces it to check whether the input is serializable as one of those types first? We can also use discriminated unions to have it explicitly check for the `kind` field.

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -801,7 +813,7 @@ class BaseToolReturnPart:
     tool_name: str
     """The name of the "tool" was called."""
 
-    content: Any
+    content: BinaryContent | FileUrl | Sequence[BinaryContent | FileUrl | Any] | dict[str, Any] | Any
```
DouweM: Let's make this a typevar

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -502,7 +511,7 @@ def __init__(
         kind: Literal['binary'] = 'binary',
         # Required for inline-snapshot which expects all dataclass `__init__` methods to take all field names as kwargs.
         _identifier: str | None = None,
-    ) -> None:
+    ) -> None:  # pragma: no cover
```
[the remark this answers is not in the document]
dsfaccini: bad suggestion, the pragma reveals a coverage gap. only way to know is actually removing it a running CI.

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_toolset.py
dsfaccini: managed to do this with recursive definition + `TypeAliasType`

pydantic_ai_slim/pydantic_ai/messages.py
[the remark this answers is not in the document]
DouweM: Still it makes me curious/doubt why we need this no-cover 😄

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -795,6 +807,41 @@ def otel_message_parts(self, settings: InstrumentationSettings) -> list[_otel_me
     Any, config=pydantic.ConfigDict(defer_build=True, ser_json_bytes='base64', val_json_bytes='base64')
 )
 
+if TYPE_CHECKING:
+    # Simpler non-recursive type for static analysis (pyright doesn't fully support recursive TypeAliasType)
```
DouweM: Can you link to an issue or smth as evidence? Something actionable like "TODO: Remove if we move to ty which does support this: [link]" would also be good

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -795,6 +807,41 @@ def otel_message_parts(self, settings: InstrumentationSettings) -> list[_otel_me
...
+    ToolReturnContent: TypeAlias = (
+        BinaryContent
+        | VideoUrl
+        | AudioUrl
+        | ImageUrl
+        | DocumentUrl
```
DouweM: Maybe use MultiModalContent or whatever that constant is called

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -795,6 +807,41 @@ def otel_message_parts(self, settings: InstrumentationSettings) -> list[_otel_me
...
+        | list[Any]
+        | dict[str, Any]
+        | str
+        | int
+        | float
+        | bool
+        | None
```
DouweM: - I don't think exhaustively list these?
- I think we definitely still should have `Any` here to allow returning BaseModels and any other type of serializable custom object

So remove everything we don't need recursion on and we should be good :)

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -795,6 +807,41 @@ def otel_message_parts(self, settings: InstrumentationSettings) -> list[_otel_me
...
+    # BinaryContent/FileUrl objects nested inside dicts/lists during deserialization
+    ToolReturnContent = TypeAliasType(
+        'ToolReturnContent',
+        BinaryContent
+        | VideoUrl
+        | AudioUrl
+        | ImageUrl
+        | DocumentUrl
+        | list['ToolReturnContent']
+        | dict[str, 'ToolReturnContent']
```
DouweM: See if we can Sequence, and the equivalent thing for dict, so that we catch all similar container objects we need to recurse through

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -803,7 +850,7 @@ class BaseToolReturnPart:
     tool_name: str
     """The name of the "tool" was called."""
 
-    content: Any
+    content: ToolReturnContent
```
DouweM, line 829: Also on `ToolReturn`?

tests/test_temporal.py
```diff
@@ -3103,3 +3109,76 @@ def test_pydantic_ai_plugin_with_non_pydantic_converter_preserves_codec() -> Non
...
+    )
+
+
+def test_binary_content_round_trip_in_list():
+    """Test round-trip when BinaryContent is inside a list."""
+    binary_data = b'\x89PNG'
+    bc = BinaryContent(data=binary_data, media_type='image/png')
+
+    wrapped = _ToolReturn(result=[bc, 'other-item'])
+    serialized = to_json(wrapped)
```
DouweM: We need a test that actually uses Temporal rather than just relying on our (current, possibly outdated in the future) knowledge that it uses `to_json`. Unit testing instead of integration testing is OK with our own code, but deeefinitely not with 2 external packages 😄 

So we need to actually set up a workflow and make sure that a real tool's return value of binary data makes it through from the activity to the workflow, by itself, in lists, dicts, etc.

DouweM: Basically, I suggest throwing these tests away and starting from scratch :)

dsfaccini: rewrote them :)

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -795,6 +808,17 @@ def otel_message_parts(self, settings: InstrumentationSettings) -> list[_otel_me
...
+    # Simpler type for static analysis - recursive TypeAliasType with Any produces spurious Unknown types
+    ToolReturnContent: TypeAlias = MultiModalContent | Sequence[Any] | Mapping[str, Any] | Any
+else:
+    # Recursive type for runtime Pydantic validation - enables automatic reconstruction of
+    # BinaryContent/FileUrl objects nested inside dicts/lists during deserialization
+    ToolReturnContent = TypeAliasType(
+        'ToolReturnContent',
+        MultiModalContent | Sequence['ToolReturnContent'] | Mapping[str, 'ToolReturnContent'] | Any,
+    )
+
```
dsfaccini, line 820: I expanded this comment to explain the need for this. I couldn't find a current pyright issue that I could link

pydantic_ai_slim/pydantic_ai/messages.py
dsfaccini: https://github.com/pydantic/pydantic-ai/pull/4043#discussion_r2729566811

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -502,9 +511,10 @@ def __init__(
         identifier: str | None = None,
         vendor_metadata: dict[str, Any] | None = None,
         kind: Literal['binary'] = 'binary',
-        # Required for inline-snapshot which expects all dataclass `__init__` methods to take all field names as kwargs.
         _identifier: str | None = None,
-    ) -> None:
+    ) -> None:  # pragma: no cover
+        # Pydantic dataclasses replace __init__ with a generated validator-wrapped version.
+        # This custom __init__ provides type hints for IDEs and accepts _identifier for inline-snapshot.
```
DouweM: It's confusing to me that it's both replaced and still needed for inline-snapshot? But also no-cover so it's never used in tests?

tests/test_temporal.py
```diff
@@ -3103,3 +3105,64 @@ def test_pydantic_ai_plugin_with_non_pydantic_converter_preserves_codec() -> Non
...
+        plugins=[AgentPlugin(binary_tool_temporal_agent)],
+    ):
+        output = await client.execute_workflow(
+            BinaryToolWorkflow.run,
+            args=['Call all binary tools'],
+            id='test_binary_content_serialization',
+            task_queue=TASK_QUEUE,
+        )
+        assert output == snapshot(
+            '{"get_binary_direct":"See file 4caece","get_binary_in_list":["See file 4effda","other-item"],"get_binary_in_dict":{"image":{"data":"iVBORw==","media_type":"image/png","vendor_metadata":null,"kind":"binary","identifier":"4effda"},"label":"test"}}'
```
DouweM: This isn't actually clearly testing that the `BinaryContent` made it through correctly. Can we check out `AgentRunResult.all_messages()`?

dsfaccini: I'm asserting against the function return directly now

pydantic_ai_slim/pydantic_ai/messages.py
dsfaccini: this is not coverable...

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -502,14 +511,8 @@ def __init__(
...
         kind: Literal['binary'] = 'binary',
-        # Required for inline-snapshot which expects all dataclass `__init__` methods to take all field names as kwargs.
         _identifier: str | None = None,
-    ) -> None:
-        self.data = data
-        self.media_type = media_type
-        self._identifier = identifier or _identifier
-        self.vendor_metadata = vendor_metadata
-        self.kind = kind
+    ) -> None: ...  # pragma: no cover
```
DouweM, line 517: This reads very confusing, we need at least some comment explaining why we have an init method that's not actually used.

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -502,14 +511,8 @@ def __init__(
         identifier: str | None = None,
         vendor_metadata: dict[str, Any] | None = None,
         kind: Literal['binary'] = 'binary',
-        # Required for inline-snapshot which expects all dataclass `__init__` methods to take all field names as kwargs.
```
DouweM, line 505: Still relevant?

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -631,9 +634,12 @@ def __init__(
         kind: Literal['binary'] = 'binary',
         _identifier: str | None = None,
     ):
-        super().__init__(
-            data=data, media_type=media_type, identifier=identifier or _identifier, vendor_metadata=vendor_metadata
-        )
+        # Directly set attributes to avoid Pydantic validation (allows inline-snapshot matchers)
```
DouweM: Why only on the subclass?

tests/test_temporal.py
```diff
@@ -3103,3 +3105,57 @@ def test_pydantic_ai_plugin_with_non_pydantic_converter_preserves_codec() -> Non
...
+# Tests for BinaryContent serialization in Temporal (#3702)
+
+
+binary_tool_agent = Agent(TestModel(), name='binary_tool_agent')
+
+
+@binary_tool_agent.tool
+def get_binary_in_dict(ctx: RunContext[None]) -> dict[str, Any]:
+    """Return BinaryContent nested in a dict."""
+    return {'image': BinaryContent(data=b'\x89PNG', media_type='image/png'), 'label': 'test'}
```
DouweM: We don't actually properly support `BinaryContent`s in dicts right? We don't have it in https://github.com/pydantic/pydantic-ai/pull/3826, and don't cover it in the current 'See file ...' implementation either.

So let's return a list of text + BC instead. And then in the test below, I'd rather use a snapshot than `== tool_func()`

pydantic_ai_slim/pydantic_ai/messages.py
dsfaccini: because it went from a `dataclass` to a `pydantic_dataclass` so the `__init__` isn't used any more, but pyright uses it for type hints and snapshots do need it to accept the `_identifier`

will leave a comment to that tune

don't ask me why the snapshots using it don't cover the gap lol

pydantic_ai_slim/pydantic_ai/messages.py
dsfaccini: I mean, yeah but it's partially superseded by the new comment. Will reintroduce

pydantic_ai_slim/pydantic_ai/messages.py
dsfaccini: `BinaryContent` is now a `pydantic_dataclass` (for `ser_json_bytes` config), which wraps `__init__` with Pydantic validation. `BinaryImage` is a plain class whose `__init__` actually runs — it can't call `super().__init__()` because Pydantic validation rejects inline-snapshot matchers (e.g. `IsBytes()`). On main this wasn't needed because `BinaryContent` was a stdlib `@dataclass(init=False)`.

Will update the comment to clarify.

dsfaccini: because it's a normal `dataclass`, not a `pydantic_dataclass`. I'll refine the comment

tests/test_temporal.py
dsfaccini: ok, I've updated it now

PR #4047, opened 2026-01-20
Bump `mcp` and other dependencies for Dependabot alerts

dsfaccini:
Bumping several packages.

`mcp.py` needs a bit of a refactor because of an http function that changes name
at the same time, `mcp 1.24` introduces a hang that has a (draft) PR open https://github.com/modelcontextprotocol/python-sdk/pull/1838 so we need to `pragma: no cover` a couple lines and wait for that to merge

## Summary

- mcp: 1.18.0 → 1.25.0
- fastmcp: 2.12.0 → 2.14.3
- gradio: 5.9.0 → 6.3.0
- aiohttp: 3.13.2 → 3.13.3
- urllib3: 2.6.2 → 2.6.3
- pyasn1: 0.6.1 → 0.6.2
- cbor2: 5.7.1 → 5.8.0

Updated MCP integration for breaking API changes.

## Test plan

- [x] MCP tests pass
- [x] Full CI

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/mcp.py
+++ b/pydantic_ai_slim/pydantic_ai/mcp.py
@@ -1157,7 +1157,8 @@ class MCPServerSSE(_MCPServerHTTP):
 
     # sse_client has a hang bug (https://github.com/modelcontextprotocol/python-sdk/issues/1811)
     # that prevents testing SSE transport in CI.
-    # Remove pragma once https://github.com/modelcontextprotocol/python-sdk/pull/1838 is released.
+    # TODO: Remove pragma and add a test
+    # once https://github.com/modelcontextprotocol/python-sdk/pull/1838 is released.
     @asynccontextmanager
     async def client_streams(  # pragma: no cover
         self,


pydantic_ai_slim/pydantic_ai/_mcp.py
```diff
@@ -33,11 +33,12 @@ def map_from_mcp_params(params: mcp_types.CreateMessageRequestParams) -> list[me
...
             if isinstance(content, mcp_types.TextContent):
                 user_part_content: str | Sequence[messages.UserContent] = content.text
-            else:
-                # image content
+            elif isinstance(content, mcp_types.ImageContent):
                 user_part_content = [
                     messages.BinaryContent(data=base64.b64decode(content.data), media_type=content.mimeType)
                 ]
+            else:
+                raise NotImplementedError(f'Unsupported user content type: {type(content).__name__}')
```
DouweM: What are the new types we could support here and below? Add a comment at lease please

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -1134,8 +1134,10 @@ def _transport_client(
         ],
     ]: ...
 
+    # Only used by MCPServerSSE which has a hang bug (https://github.com/modelcontextprotocol/python-sdk/issues/1811).
```
DouweM: Should this just be on that subclass then? We may not need the `_transport_client` thing anymore if each class will have its own implementation of this entire thing

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/_mcp.py
dsfaccini: the painful stuff is that `content: SamplingMessageContentBlock | list[SamplingMessageContentBlock]`

but I've moved the `NotImplementedError` into an explicit branch checking all rest-types and added an `assert_never` at the end

pydantic_ai_slim/pydantic_ai/mcp.py
dsfaccini: yeahp I wanted to keep this lean but I made that change now

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -1217,9 +1155,47 @@ def __get_pydantic_core_schema__(cls, _: Any, __: Any) -> CoreSchema:
...
+            # sse_client's McpHttpClientFactory type expects -> AsyncClient, but we use
+            # @asynccontextmanager to prevent closing the user's client. Works at runtime.
+            @asynccontextmanager
+            async def httpx_client_factory(
+                headers: dict[str, str] | None = None,
+                timeout: httpx.Timeout | None = None,
+                auth: httpx.Auth | None = None,
+            ) -> AsyncIterator[httpx.AsyncClient]:
+                assert self.http_client is not None
+                yield self.http_client
```
Kludex: Why this doesn't work?
```suggestion
            def httpx_client_factory(
                headers: dict[str, str] | None = None,
                timeout: httpx.Timeout | None = None,
                auth: httpx.Auth | None = None,
            ) -> AsyncIterator[httpx.AsyncClient]:
                assert self.http_client is not None
                return self.http_client
```

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -1281,9 +1257,33 @@ def __get_pydantic_core_schema__(cls, _: Any, __: Any) -> CoreSchema:
...
+                yield read_stream, write_stream
+        else:
+            timeout = httpx.Timeout(self.timeout, read=self.read_timeout)
+            async with httpx.AsyncClient(timeout=timeout, headers=self.headers) as client:
+                async with streamable_http_client(self.url, http_client=client) as (
+                    read_stream,
+                    write_stream,
+                    *_,
+                ):
+                    yield read_stream, write_stream
```
Kludex: To avoid duplication on `streamable_http_client`:
```suggestion
        aexit_stack = AsyncExitStack()
        if self.http_client is not None:
            http_client = self.http_client
        else:
            timeout = httpx.Timeout(self.timeout, read=self.read_timeout)
            http_client = await aexit_stack.enter_async_context(
                httpx.AsyncClient(timeout=timeout, headers=self.headers)
            )
        read_stream, write_stream, *_ = aexit_stack.enter_async_context(
            streamable_http_client(self.url, http_client=client)
        )
            try:
                yield read_stream, write_stream
            finally:
                # unwrap the aexit
```

dsfaccini: again, I wanted to keep this trim but yeah, I'll commit those and if they work great

dsfaccini: I rushingly wrote this comment so I'd like to clarify the "again" part was in reference to my above reply to Douwe's comment https://github.com/pydantic/pydantic-ai/pull/4047#discussion_r2710447071 where he also asked why I didn't fix the implementation, since I'm already doing this PR

the reality is I did try to fix it but things were failing and I wasn't sure "my fix" was even the right approach, which is why I left it as it was + the ignore, because I held the priority of bumping the packages higher than addressing the implementation.

the " I'll commit those and if they work great" is also in reference to my own experience trying to fix it and failing, absolutely not an opinion on the change you suggested or in doubt that it's the better approach

appreciate you making the suggestion, I merged it and filled out the gaps and CI seems to agree with it, so now we can bump the package and have good conscience that we left the implementation better than before

sorry if my above comment came off as dismissive! I sometimes prioritize leaving a quick acknowledgement and don't read myself how it might sound!

Kludex: all good

pydantic_ai_slim/pydantic_ai/_mcp.py
DouweM: - Could we support `AudioContent` easily in the `ImageContent` branch above?
- Let's make it clear in the error which types are not _yet_ supported (like in `map_from_sampling_content`) but relatively easily could be, vs anything that is fundamentally unsupported. Just so people know what their options are when they hit this error (e.g. contribute a fix)

pydantic_ai_slim/pydantic_ai/_mcp.py
```diff
@@ -47,7 +54,10 @@ def map_from_mcp_params(params: mcp_types.CreateMessageRequestParams) -> list[me
                 pai_messages.append(messages.ModelRequest(parts=request_parts))
                 request_parts = []
 
-            response_parts.append(map_from_sampling_content(content))
+            if isinstance(content, (mcp_types.TextContent, mcp_types.ImageContent, mcp_types.AudioContent)):
+                response_parts.append(map_from_sampling_content(content))
```
DouweM, line 58: We're passing in the Image/Audio content, but then this method will raise an error. That's fine, but I think now that we have `FilePart` it should be easy enough to support then. So worth adding a todo comment, and/or creating an issue so users can pick it up if they want to.

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -1217,9 +1155,45 @@ def __get_pydantic_core_schema__(cls, _: Any, __: Any) -> CoreSchema:
             ),
         )
 
-    @property
-    def _transport_client(self):
-        return sse_client  # pragma: no cover
+    # sse_client has a hang bug (https://github.com/modelcontextprotocol/python-sdk/issues/1811).
+    # Remove pragma once https://github.com/modelcontextprotocol/python-sdk/pull/1838 is released.
```
DouweM: It's not obvious why a hang bug results in a `pragma: no cover`, and why we can remove the pragma once the hang bug is fixed. I think we'd need to add a new test to actually cover SSE, but your point is that we can't because of the hang bug?

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -1281,9 +1255,27 @@ def __get_pydantic_core_schema__(cls, _: Any, __: Any) -> CoreSchema:
...
+        self,
+    ) -> AsyncIterator[
+        tuple[
+            MemoryObjectReceiveStream[SessionMessage | Exception],
+            MemoryObjectSendStream[SessionMessage],
+        ]
+    ]:
+        aexit_stack = AsyncExitStack()
+        http_client = self.http_client or await aexit_stack.enter_async_context(
+            httpx.AsyncClient(timeout=httpx.Timeout(self.timeout, read=self.read_timeout), headers=self.headers)
```
DouweM, line 1274: Should we have the `ValueError('`http_client` is mutually exclusive with `headers`.')` error here as well?

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -1281,9 +1255,27 @@ def __get_pydantic_core_schema__(cls, _: Any, __: Any) -> CoreSchema:
...
+        )
+        read_stream, write_stream, *_ = await aexit_stack.enter_async_context(
+            streamable_http_client(self.url, http_client=http_client)
+        )
+        try:
+            yield read_stream, write_stream
+        finally:
+            # unwrap the aexit
```
DouweM: unnecessary comment, this just literally explains the next line :)

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/mcp.py
dsfaccini: precisely

pydantic_ai_slim/pydantic_ai/_mcp.py
```diff
@@ -123,4 +133,5 @@ def map_from_sampling_content(
     if isinstance(content, mcp_types.TextContent):  # pragma: no branch
         return messages.TextPart(content=content.text)
     else:
+        # TODO(Marcelo): Add support for Image/Audio using FilePart.
```
DouweM: Why for Marcelo lol

pydantic_ai_slim/pydantic_ai/mcp.py
DouweM: Remove pragma "_and_ add a test", right? Also worth marking as "TODO:"

pydantic_ai_slim/pydantic_ai/_mcp.py
dsfaccini: lmao sorry

PR #4048, opened 2026-01-20
bug: Warn about Azure OpenAI completions file-incompatibility

### LINKED ISSUE #4035: `AzureOpenAI` models don't support files
### Initial Checks

### Description

Using OpenAI models with Azure does not allow files to be processed. However, the same implementation with the same model directly to OpenAI does work.
According to the official Azure documentation, these models do support file input as bytes:
https://learn.microsoft.com/en-us/azure/ai-foundry/openai/how-to/responses?view=foundry-classic&tabs=rest-api#convert-pdf-to-base64-and-analyze

Agents with OpenAI model can process files as bytes:
```Python
agent = Agent(model="openai:gpt-5-mini")
await agent.run(
    [
        SYSTEM_PROMPT,
        BinaryContent(data=doc_bytes, media_type='application/pdf'),
    ]
)
```
This works correctly.

However, the same code and model with an Azure OpenAI model generates an error:
```Python
agent = Agent(model="azure:gpt-5-mini")
await agent.run(
    [
        SYSTEM_PROMPT,
        BinaryContent(data=doc_bytes, media_type='application/pdf'),
    ]
)
```

Error:
```
pydantic_ai.exceptions.ModelHTTPError: status_code: 400, model_name: gpt-5-mini, body: {'message': "Invalid Value: 'file'. This model does not support file content types.", 'type': 'invalid_request_error', 'param': 'messages[0].content[1].type', 'code': 'invalid_value'}
```

### Logfire Trace

_No response_

- Python: 3.11.13
- Pydantic AI: 1.44.0

dsfaccini:
- Closes #4035

### Pre-Review Checklist

### Pre-Merge Checklist


docs/input.md
```diff
@@ -118,7 +118,31 @@ Support for file URLs varies depending on type and provider:
...
+    agent = Agent(model)
+    result = agent.run_sync([
+        'Summarize this document',
+        BinaryContent(data=pdf_bytes, media_type='application/pdf'),
+    ])
+    ```
+
+A model API may be unable to download a file (e.g., because of crawling or access restrictions) even if it supports file URLs. For example, [`GoogleModel`][pydantic_ai.models.google.GoogleModel] on Vertex AI limits YouTube video URLs to one URL per request.
+
+In such cases, you can instruct Pydantic AI to download the file content locally and send that instead of the URL by setting `force_download` on the URL object:
```
dsfaccini, line 158: added a line break here for personal taste

docs/input.md
```diff
@@ -118,7 +118,31 @@ Support for file URLs varies depending on type and provider:
 | [`MistralModel`][pydantic_ai.models.mistral.MistralModel] | `ImageUrl`, `DocumentUrl` (PDF) | — | `AudioUrl`, `VideoUrl`, `DocumentUrl` (non-PDF) |
 | [`BedrockConverseModel`][pydantic_ai.models.bedrock.BedrockConverseModel] | S3 URLs (`s3://`) | `ImageUrl`, `DocumentUrl`, `VideoUrl` | `AudioUrl` |
 
-A model API may be unable to download a file (e.g., because of crawling or access restrictions) even if it supports file URLs. For example, [`GoogleModel`][pydantic_ai.models.google.GoogleModel] on Vertex AI limits YouTube video URLs to one URL per request. In such cases, you can instruct Pydantic AI to download the file content locally and send that instead of the URL by setting `force_download` on the URL object:
+??? warning "`DocumentUrl` and `BinaryContent` documents are not supported when using `AzureProvider` with `OpenAIChatModel`."
+    Use [`OpenAIResponsesModel`][pydantic_ai.models.openai.OpenAIResponsesModel] with [`AzureProvider`][pydantic_ai.providers.azure.AzureProvider] instead:
```
dsfaccini: image

pydantic_ai_slim/pydantic_ai/profiles/openai.py
```diff
@@ -80,6 +80,13 @@ class OpenAIModelProfile(ModelProfile):
     See https://github.com/pydantic/pydantic-ai/issues/3245 for more details.
     """
 
+    openai_chat_supports_file_input: bool = True
+    """Whether the Chat Completions API supports file content parts (type='file').
+
+    Some OpenAI-compatible providers (e.g. Azure) do not support file input via the Chat Completions API.
+    For Azure, use `OpenAIResponsesModel` with `AzureProvider` instead.
```
dsfaccini: remove this line?

DouweM: yep

docs/input.md
DouweM: makes sense

docs/input.md
```diff
@@ -118,7 +118,33 @@ Support for file URLs varies depending on type and provider:
 | [`MistralModel`][pydantic_ai.models.mistral.MistralModel] | `ImageUrl`, `DocumentUrl` (PDF) | — | `AudioUrl`, `VideoUrl`, `DocumentUrl` (non-PDF) |
 | [`BedrockConverseModel`][pydantic_ai.models.bedrock.BedrockConverseModel] | S3 URLs (`s3://`) | `ImageUrl`, `DocumentUrl`, `VideoUrl` | `AudioUrl` |
 
-A model API may be unable to download a file (e.g., because of crawling or access restrictions) even if it supports file URLs. For example, [`GoogleModel`][pydantic_ai.models.google.GoogleModel] on Vertex AI limits YouTube video URLs to one URL per request. In such cases, you can instruct Pydantic AI to download the file content locally and send that instead of the URL by setting `force_download` on the URL object:
+??? warning "`DocumentUrl` and `BinaryContent` documents are not supported when using `AzureProvider` with `OpenAIChatModel`."
```
DouweM: I'd rather add a Notes column to the table, like we have on the builtin tools doc, where we can state on the `OpenAIChatModel` row that Azure doesn't support it, and any other notes for provider-specific support (I believe there are some profile flags for specific providers that enable file types that are not supported by OpenAI itself)

docs/input.md
DouweM: I don't think we need an example here, but I want to make sure the Azure doc (currently section on the OpenAI page) explains (in a generic way) that it can be used with the Responses API as well, and that that's actually recommend over `OpenAIChatModel`. Can you add it there instead please?

docs/input.md
dsfaccini: looking at my screen I'm not liking the idea of a fifth column 
image

docs/input.md
dsfaccini: for sure, TODO

dsfaccini: moved to azure, I'm thinking we should simply replace the default example with responses and leave the chat one in a second tab?
Edit: also can I remove those trailing `...` from examples?
image

docs/input.md
DouweM: Instead of adding a fifth column, what about mentioning the Azure case under "Unsupported", and then linking to the section that explains the responses workaround?

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -1151,6 +1155,16 @@ async def _map_user_prompt(self, part: UserPromptPart) -> chat.ChatCompletionUse
                     assert_never(item)
         return chat.ChatCompletionUserMessageParam(role='user', content=content)
 
+    def _raise_file_input_not_supported_error(self) -> Never:
```
DouweM: If this is specific to documents, as the error suggests, let's have that in the name instead of "file". Or does it really not support any files, like the profile flag suggets?

docs/models/openai.md
```diff
@@ -427,6 +427,34 @@ agent = Agent(model)
...
+        ),
+    )
+    agent = Agent(model)
+    result = agent.run_sync([
+        'Summarize this document',
+        BinaryContent(data=pdf_bytes, media_type='application/pdf'),
+    ])
+    ```
+
+The Responses API provides the same functionality as the Chat Completions API, with additional features and better support for advanced input types. When using Azure, consider using `OpenAIResponsesModel` for the most comprehensive feature support.
```
DouweM: Duplicated with the text above

tests/models/test_openai.py
```diff
@@ -1442,6 +1443,72 @@ async def test_text_document_as_binary_content_input(
     )
 
 
+async def test_yaml_document_as_binary_content_input(allow_model_requests: None, openai_api_key: str):
```
DouweM, line 1382: Why did we need these new tests?

dsfaccini: The YAML VCR tests (`test_yaml_document_as_binary_content_input`, `test_x_yaml_document_as_binary_content_input`, `test_yaml_document_url_input`) cover the YAML text-like media type handling added in this PR (the new `application/yaml` and `application/x-yaml` branches in `_is_text_like_media_type`).

`test_is_text_like_media_type` is a unit test for branch coverage of `_is_text_like_media_type`.

`test_video_url_not_supported` replaces a `pragma: no cover` that was on the `VideoUrl` raise — now that branch is tested directly.

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -1246,6 +1250,16 @@ async def _map_user_prompt(self, part: UserPromptPart) -> chat.ChatCompletionUse
                     assert_never(item)
         return chat.ChatCompletionUserMessageParam(role='user', content=content)
 
+    def _raise_document_input_not_supported_error(self) -> Never:
+        if self._provider.name == 'azure':
```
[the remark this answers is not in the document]
DouweM: I'm OK with it in this specific case.

tests/models/test_openai.py
dsfaccini: These tests cover YAML media-type edge cases I discovered while investigating how Azure handles different file types. They validate pre-existing behavior that was untested, ensuring the content-type classification works correctly for YAML variants (application/x-yaml, text/yaml, etc). Happy to split them out if you prefer, but they're directly related to the file handling work in this PR.

.claude/skills/gh-cli-best-practices/fetch-comment
DouweM, line 1: These don't belong in this PR :)

tests/models/test_openai.py
DouweM: OK that makes sense, happy to have them here then

pydantic_ai_slim/pydantic_ai/profiles/openai.py
```diff
@@ -118,6 +118,12 @@ class OpenAIModelProfile(ModelProfile):
     See https://github.com/pydantic/pydantic-ai/issues/3245 for more details.
     """
 
+    openai_chat_supports_document_input: bool = True
+    """Whether the Chat Completions API supports document content parts (type='file').
```
DouweM: backticks for code

docs/input.md
```diff
@@ -110,7 +110,7 @@ Support for file URLs varies depending on type and provider:
 
 | Model | Send URL directly | Download and send bytes | Unsupported |
 |-------|-------------------|-------------------------|-------------|
-| [`OpenAIChatModel`][pydantic_ai.models.openai.OpenAIChatModel] | `ImageUrl` | `AudioUrl`, `DocumentUrl` | `VideoUrl` |
+| [`OpenAIChatModel`][pydantic_ai.models.openai.OpenAIChatModel] | `ImageUrl` | `AudioUrl`, `DocumentUrl` | `VideoUrl`. `AudioUrl` and `DocumentUrl` [not supported with `AzureProvider`](models/openai.md#using-azure-with-the-responses-api) |
```
DouweM: In the rest of the code and docs we only talk about document support, but this suggests Audio is not supported either?

dsfaccini: actually does, my bad https://learn.microsoft.com/en-us/azure/foundry-classic/openai/concepts/audio

pydantic_ai_slim/pydantic_ai/models/openai.py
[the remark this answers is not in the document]
dsfaccini: IMO the `openai_chat_supports_file_urls` flag inherently implies the provider supports documents, so I don't think we need to explicitly handle the combination

for future reference, __if__ we found these inconsistencies, having a single dataclass that uses property to validate combinations would be a better approach than checking multiple flags, since adding a check branch here would bloat this out further

dsfaccini:
claude: TODO: remove the skills form this PR

PR #4049, opened 2026-01-20
fix(tools): Validate args before deferring for approval

### LINKED ISSUE #3893: Support for additional tool validation logic

dsfaccini:
- Closes #3893

### Pre-Review Checklist

### Pre-Merge Checklist

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/_tool_manager.py
+++ b/pydantic_ai_slim/pydantic_ai/_tool_manager.py
@@ -168,7 +168,7 @@ class ToolManager(Generic[AgentDepsT]):
             allow_partial: Whether to allow partial validation of the tool arguments.
             wrap_validation_errors: Whether to wrap validation errors in a retry prompt part.
             approved: Whether the tool call has been approved.
-            metadata: Additional metadata from `DeferredToolResults`.metadata.
+            metadata: Additional metadata from DeferredToolResults.metadata.
         """
         if self.tools is None or self.ctx is None:
             raise ValueError('ToolManager has not been prepared for a run step yet')  # pragma: no cover


pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -1029,10 +1029,18 @@ async def process_tool_calls(  # noqa: C901
...
         elif calls:
-            deferred_calls['external'].extend(tool_calls_by_kind['external'])
-            deferred_calls['unapproved'].extend(tool_calls_by_kind['unapproved'])
-
             for call in calls:
+                try:
+                    tool_manager.validate_args(call)
+                except ToolRetryError as e:
+                    ctx.state.increment_retries(ctx.deps.max_result_retries, model_settings=ctx.deps.model_settings)
+                    output_parts.append(e.tool_retry)
```
DouweM, line 1037: Should we be emitting function tool call events here like in the `ToolRetryError` branch further up?

pydantic_ai_slim/pydantic_ai/_tool_manager.py
```diff
@@ -91,6 +91,67 @@ def get_tool_def(self, name: str) -> ToolDefinition | None:
         except KeyError:
             return None
 
+    def validate_args(self, call: ToolCallPart) -> dict[str, Any]:
```
DouweM, line 94: - This definitely has way too much duplication with `handle_call`. Either pull things out, or use this method in that one.
- I think `validate_call` would be a better name, as `args` implies it takes a dict of `args`.

DouweM:
Very similar changes are being made in https://github.com/pydantic/pydantic-ai/pull/4016, so let's have that one solve https://github.com/pydantic/pydantic-ai/issues/3893 as well.

PR #4056, opened 2026-01-21
Add `continuous_usage_stats` model setting for OpenAI

kousun12:
## Summary

When `continuous_usage_stats` is enabled in `stream_options`, the OpenAI API returns **cumulative** usage data with each chunk rather than only at the end. Previously, pydantic-ai would add all these cumulative values together, resulting in inflated usage statistics.

## Problem

Example output with `continuous_usage_stats=True`:
```
chunk 1: completion_tokens=1, total_tokens=11
chunk 2: completion_tokens=4, total_tokens=14  
chunk 3: completion_tokens=5, total_tokens=15
...
final:   completion_tokens=99, total_tokens=109
```

The old code would sum all of these: 1+4+5+...+99 = inflated total
The correct behavior: use only the final value (99 completion tokens)

## Changes

1. **New setting**: Added `openai_continuous_usage_stats: bool` to `OpenAIChatModelSettings`
2. **API request**: When enabled, passes `continuous_usage_stats: True` in `stream_options` to the API
3. **Response handling**: `OpenAIStreamedResponse` now replaces (rather than accumulates) usage when `_continuous_usage_stats` is True

## Usage

```python
from pydantic_ai import Agent

agent = Agent(OpenAIChatModel('gpt-4'))

# Enable continuous usage stats
result = await agent.run(
    'Hello',
    model_settings={'openai_continuous_usage_stats': True}
)

# Usage will now be correct (not inflated)
print(result.usage())
```

## Notes

- Default behavior unchanged (`continuous_usage_stats=False`)
- Only affects OpenAI Chat Completions API streaming (not Responses API)


pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -827,6 +842,21 @@ def _streamed_response_cls(self) -> type[OpenAIStreamedResponse]:
...
+    def _get_stream_options(
+        self, stream: bool, model_settings: OpenAIChatModelSettings
+    ) -> dict[str, bool] | Omit:
+        """Build stream_options for the API request.
+
+        Returns OMIT for non-streaming requests, otherwise returns a dict with
+        include_usage=True and optionally continuous_usage_stats if configured.
+        """
+        if not stream:
+            return OMIT
```
DouweM: I'd rather not pass in `stream` and have `... if stream else OMIT` at the call site

DouweM (changes requested):
@kousun12 Please have a look at the failing CI steps!

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -827,6 +842,21 @@ def _streamed_response_cls(self) -> type[OpenAIStreamedResponse]:
...
+        options: dict[str, bool] = {'include_usage': True}
+        if model_settings.get('openai_continuous_usage_stats'):
+            options['continuous_usage_stats'] = True
```
DouweM, line 850: Should we support this setting in `OpenAIResponsesModel` as well?

kousun12: no i dont believe there is a similar option for that api

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -814,6 +826,9 @@ async def _process_streamed_response(
             _provider_name=self._provider.name,
             _provider_url=self._provider.base_url,
             _provider_timestamp=number_to_datetime(first_chunk.created) if first_chunk.created else None,
+            _continuous_usage_stats=bool(model_settings.get('openai_continuous_usage_stats'))
```
DouweM: I'd rather pass in the entire model settings, like we do model request parameters. I believe we already do that with other streamed response classes for other models.

tests/models/test_openai.py
```diff
@@ -4049,3 +4049,101 @@ def test_transformer_adds_properties_to_object_schemas():
...
+        chunk_with_usage([ChoiceDelta(content='!')], completion_tokens=15, prompt_tokens=10, total_tokens=25),
+        chunk_with_usage([], finish_reason='stop', completion_tokens=15, prompt_tokens=10, total_tokens=25),
+    ]
+    mock_client = MockOpenAI.create_mock_stream(stream)
+    m = OpenAIChatModel('gpt-4o', provider=OpenAIProvider(openai_client=mock_client))
+    agent = Agent(m)
+
+    settings = cast(OpenAIChatModelSettings, {'openai_continuous_usage_stats': True})
+    async with agent.run_stream('', model_settings=settings) as result:
+        chunks = [c async for c in result.stream_text(debounce_by=None)]
```
DouweM: Can we verify that `stream_responses` gives us updated usage at each step?

tests/models/test_openai.py
```diff
@@ -4049,3 +4049,101 @@ def test_transformer_adds_properties_to_object_schemas():
...
+    ]
+    mock_client = MockOpenAI.create_mock_stream(stream)
+    m = OpenAIChatModel('gpt-4o', provider=OpenAIProvider(openai_client=mock_client))
+    agent = Agent(m)
+
+    async with agent.run_stream('') as result:
+        chunks = [c async for c in result.stream_text(debounce_by=None)]
+        assert chunks == snapshot(['hello ', 'hello world', 'hello world!'])
+
+    # Without continuous_usage_stats, usage is accumulated: 5+5+5+0 = 15 output tokens, 10+0+0+0 = 10 input tokens
```
DouweM: Would we ever actually get a response with usage on each chunk _without_ continuous usage? If not, I don't think we need to test this case?

tests/models/test_openai.py
```diff
@@ -4049,3 +4049,101 @@ def test_transformer_adds_properties_to_object_schemas():
...
+    """Test that _get_stream_options includes continuous_usage_stats when set."""
+    mock_client = MockOpenAI.create_mock(completion_message(ChatCompletionMessage(content='test', role='assistant')))
+    m = OpenAIChatModel('gpt-4o', provider=OpenAIProvider(openai_client=mock_client))
+
+    # Without continuous_usage_stats
+    options = m._get_stream_options(cast(OpenAIChatModelSettings, {}))  # pyright: ignore[reportPrivateUsage]
+    assert options == {'include_usage': True}
+
+    # With continuous_usage_stats=False (explicit)
+    options = m._get_stream_options(cast(OpenAIChatModelSettings, {'openai_continuous_usage_stats': False}))  # pyright: ignore[reportPrivateUsage]
```
DouweM: No unit testing of this private method please, the integration tests should be enough

dsfaccini:

```bash
git pull

deactivate

rm -rf .venv

make install
```

kousun12:
@DouweM  would love to get this in!

DouweM:
@kousun12 Thanks Rob!

PR #4058, opened 2026-01-21
bug: Fix gateway snippet auto-generation for the `google-gla` shorthand

dsfaccini:
our docs hook was generating examples with gateway/google-gla shorthands that aren't actually valid


dsfaccini:
waiting on preview docs link add a screenshot

PR #4061, opened 2026-01-22
feat: preserve reasoning content field during multi-turn round-trips (#3867)

SakshamKapoor2911:
Currently, Pydantic-AI often defaults to wrapping reasoning/thinking content in XML-like tags (<think>) when sending assistant messages back to a model in subsequent turns. This PR implements 'field-aware round-tripping' for reasoning content.

### Key Changes
- **OpenAI Refactor**: 
  - Modified `_MapModelResponseContext` to use a `dict[str, list[str]]` for storing thoughts, keyed by their source field ID.
  - Updated mapping logic to utilize `ThinkingPart.id` as the target field name.
  - Implemented a safety fallback: if `id` is generic (`'content'`), it reverts to the profile's default thinking field.
- **Provider Standardization**: 
  - Assigned default IDs to reasoning parts in **Anthropic**, **Mistral**, and **Groq** models for seamless model switching.
- **Improved Extraction**: 
  - The OpenAI model now captures *all* available reasoning fields if a provider returns more than one.

### Testing
- Added comprehensive verification in `tests/models/test_openai.py` covering multiple fields, custom IDs, and streaming reconstruction.

Closes: #3867


DouweM:
@SakshamKapoor2911 There's already a PR for this issue at https://github.com/pydantic/pydantic-ai/pull/4009. If you have feedback on that implementation, please share it there.

PR #4075, opened 2026-01-23
Record `cache_read` tokens in token usage histograms

### LINKED ISSUE #4074: Capture cached token usage in otel histograms
### Description

Capture cache read tokens in the gen_ai token usage histogram as a new token type. This helps library users get a more accurate cost calculation. It also will enable aggregation of cache read tokens in logfire otel spans.

### References

_No response_

chasefarmer-pixee:
- Closes #4074 

### Pre-Review Checklist

### Pre-Merge Checklist


pydantic_ai_slim/pydantic_ai/models/instrumented.py
```diff
@@ -329,13 +329,13 @@ def record_metrics(
         price_calculation: PriceCalculation | None,
         attributes: dict[str, AttributeValue],
     ):
-        for typ in ['input', 'output']:
-            if not (tokens := getattr(response.usage, f'{typ}_tokens', 0)):  # pragma: no cover
+        for typ in ['input', 'output', 'cache_read']:
```
DouweM: If we're going to add more than just `input` and `output`, I think we should include all of the fields on `RequestUsage`.

@alexmojaki Any concerns about that / the rest of the PR?

dsfaccini:

```bash
git pull

deactivate

rm -rf .venv

make install
```

pydantic_ai_slim/pydantic_ai/models/instrumented.py
alexmojaki: This violates this:

> As a rule of thumb, either the sum() or the avg() over all dimensions of a given metric should be meaningful (though not necessarily useful). If it is not meaningful, split the data up into multiple metrics.

https://prometheus.io/docs/practices/naming/

which I think OTel also likes to stick to.

Since input+output includes all tokens, and all other token types like cached tokens are a subset of one of those, I think they need a separate metric name to not mess with adding up values.

chasefarmer-pixee: Thanks for the feedback here!  I agree a separate metric makes sense.

DouweM:
@chasefarmer-pixee Please check the failing CI

pydantic_ai_slim/pydantic_ai/models/instrumented.py
```diff
@@ -330,14 +343,20 @@ def record_metrics(
...
-                cost = float(getattr(price_calculation, f'{typ}_price'))
+            if price_calculation and (cost_attr := getattr(price_calculation, f'{typ}_price', None)) is not None:
+                cost = float(cost_attr)
                 self.cost_histogram.record(cost, token_attributes)
 
+        for cache_typ in ['cache_read', 'cache_write']:
+            if not (tokens := getattr(response.usage, f'{cache_typ}_tokens', 0)):
+                continue
+            cache_token_attributes = {**attributes, 'gen_ai.token.type': cache_typ}
+            self.cached_tokens_histogram.record(tokens, cache_token_attributes)
```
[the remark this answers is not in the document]
DouweM: Cache tokens only for now works for me.

PR #4077, opened 2026-01-24
Support parallel tool calls in `DBOSAgent`

### LINKED ISSUE #4076: Support parallel tool calls in DBOSAgent
### Description

Related to the discussion in #2853: running tools in parallel can significantly reduce end-to-end latency, especially when individual tool calls are long-running.

DBOS recently added support for parallel async steps (https://github.com/dbos-inc/dbos-transact-py/pull/559). We're working on a solution and planning to open a PR soon to support parallel tool calling in DBOSAgent.

### References

_No response_

qianl15:
This PR enables parallel tool calls for `DBOSAgent`. 
- [x] Update the DBOS version to 2.10.0. It's the version that supports parallel step execution.
- Update `dbos_test.py` to use `tmp_path` so the SQLite file is automatically cleaned up after tests. Solves #3921
- Add a new tool execution mode `parallel_ordered_events` (in addition to `parallel` (default) and `sequential`). If it's enabled, then tools are executed in parallel, but the agent would wait for all tools to finish before emitting events in order. This is essential to guarantee a deterministic invocation order (especially for the event_stream_handler) when replaying workflows.

- Closes #4076

### Pre-Review Checklist

### Pre-Merge Checklist

### DIFF
--- a/docs/builtin-tools.md
+++ b/docs/builtin-tools.md
@@ -150,7 +150,7 @@ _(This example is complete, it can be run "as is")_
 | `search_context_size` | ✅ | ❌ | ❌ | ❌ |
 | `user_location` | ✅ | ✅ | ❌ | ❌ |
 | `blocked_domains` | ❌ | ✅ | ✅ | ✅ |
-| `allowed_domains` | ❌ | ✅ | ✅ | ✅ |
+| `allowed_domains` | ✅ | ✅ | ✅ | ✅ |
 | `max_uses` | ❌ | ✅ | ❌ | ❌ |
 
 !!! note "Anthropic Domain Filtering"
--- a/pydantic_ai_slim/pydantic_ai/builtin_tools.py
+++ b/pydantic_ai_slim/pydantic_ai/builtin_tools.py
@@ -137,6 +137,7 @@ class WebSearchTool(AbstractBuiltinTool):
 
     * Anthropic, see <https://docs.anthropic.com/en/docs/build-with-claude/tool-use/web-search-tool#domain-filtering>
     * Groq, see <https://console.groq.com/docs/agentic-tooling#search-settings>
+    * OpenAI Responses, see <https://platform.openai.com/docs/guides/tools-web-search>
     * xAI, see <https://docs.x.ai/docs/guides/tools/search-tools#web-search-parameters>
     """
 
--- a/pydantic_ai_slim/pydantic_ai/embeddings/openai.py
+++ b/pydantic_ai_slim/pydantic_ai/embeddings/openai.py
@@ -169,11 +169,15 @@ class OpenAIEmbeddingModel(EmbeddingModel):
 
 
 def _map_usage(
-    usage: Usage,
+    usage: Usage | None,
     provider: str,
     provider_url: str,
     model: str,
 ) -> RequestUsage:
+    # OpenAI SDK types say CreateEmbeddingResponse.usage will always be set, in reality some OpenAI-compatible APIs omit it.
+    if usage is None:
+        return RequestUsage()
+
     usage_data = usage.model_dump(exclude_none=True)
     details = {k: v for k, v in usage_data.items() if k not in {'prompt_tokens', 'total_tokens'} if isinstance(v, int)}
     response_data = dict(model=model, usage=usage_data)
--- a/pydantic_ai_slim/pydantic_ai/models/openai.py
+++ b/pydantic_ai_slim/pydantic_ai/models/openai.py
@@ -351,6 +351,16 @@ class OpenAIChatModelSettings(ModelSettings, total=False):
     See the [OpenAI Prompt Caching documentation](https://platform.openai.com/docs/guides/prompt-caching#how-it-works) for more information.
     """
 
+    openai_continuous_usage_stats: bool
+    """When True, enables continuous usage statistics in streaming responses.
+
+    When enabled, the API returns cumulative usage data with each chunk rather than only at the end.
+    This setting correctly handles the cumulative nature of these stats by using only the final
+    usage values rather than summing all intermediate values.
+
+    See [OpenAI's streaming documentation](https://platform.openai.com/docs/api-reference/chat/create#stream_options) for more information.
+    """
+
 
 @deprecated('Use `OpenAIChatModelSettings` instead.')
 class OpenAIModelSettings(OpenAIChatModelSettings, total=False):
@@ -615,11 +625,10 @@ class OpenAIChatModel(Model):
             model_settings,
             model_request_parameters,
         )
-        response = await self._completions_create(
-            messages, True, cast(OpenAIChatModelSettings, model_settings or {}), model_request_parameters
-        )
+        model_settings_cast = cast(OpenAIChatModelSettings, model_settings or {})
+        response = await self._completions_create(messages, True, model_settings_cast, model_request_parameters)
         async with response:
-            yield await self._process_streamed_response(response, model_request_parameters)
+            yield await self._process_streamed_response(response, model_request_parameters, model_settings_cast)
 
     @overload
     async def _completions_create(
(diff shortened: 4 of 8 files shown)


docs/tools-advanced.md
```diff
@@ -407,6 +407,7 @@ When a timeout occurs, the tool is considered to have failed and the model recei
 
 When a model returns multiple tool calls in one response, Pydantic AI schedules them concurrently using `asyncio.create_task`.
 If a tool requires sequential/serial execution, you can pass the [`sequential`][pydantic_ai.tools.ToolDefinition.sequential] flag when registering the tool, or wrap the agent run in the [`with agent.sequential_tool_calls()`][pydantic_ai.agent.AbstractAgent.sequential_tool_calls] context manager.
+If you want to wait for all tools in a batch to finish before emitting events, you can wrap the agent run in the [`with agent.parallel_wait_all_tool_calls()`][pydantic_ai.agent.AbstractAgent.parallel_wait_all_tool_calls] context manager.
```
DouweM: If we can't imagine a specific scenario where a regular non-DBOS user would want to do this, let's not document it.

pydantic_ai_slim/pydantic_ai/_tool_manager.py
```diff
@@ -48,6 +49,21 @@ def sequential_tool_calls(cls) -> Iterator[None]:
         finally:
             _sequential_tool_calls_ctx_var.reset(token)
 
+    @classmethod
+    @contextmanager
+    def parallel_wait_all_tool_calls(cls) -> Iterator[None]:
```
DouweM: I'd rather deprecate `sequential_tool_calls` and add a new method that takes a `Literal` of tool execution modes: `parallel`, `sequential`, and some new value for the new mode. I don't love `parallel_wait_all_tool_calls` as a name though as it's not clear that it's really just about events. Any other ideas? 😄

DouweM: If we do that with `agent.sequential_tool_calls` as well, then you can ignore my "don't document the new mode" point from above, as it'd be more natural to document if it's just an option in a literal type.

pydantic_ai_slim/pydantic_ai/durable_exec/dbos/_agent.py
```diff
@@ -57,11 +58,15 @@ def __init__(
             event_stream_handler: Optional event stream handler to use instead of the one set on the wrapped agent.
             mcp_step_config: The base DBOS step config to use for MCP server steps. If no config is provided, use the default settings of DBOS.
             model_step_config: The DBOS step config to use for model request steps. If no config is provided, use the default settings of DBOS.
+            sequential_tool_calls: If True, force tool calls to execute sequentially instead of in parallel.
```
DouweM: This could be a flag that takes the same mode Literal I suggest above, without the regular `parallel` option.

docs/tools-advanced.md
qianl15: Switched to use literals and updated the document for `parallel`, `sequential`, and `parallel_ordered_events`.

pydantic_ai_slim/pydantic_ai/durable_exec/dbos/_agent.py
qianl15: Changed this parameter to accept a Literal of modes (without the `parallel` option)

pydantic_ai_slim/pydantic_ai/_tool_manager.py
qianl15: Updated this PR to deprecate `sequential_tool_calls` and added a new `tool_calls_mode()` function instead.

docs/tools-advanced.md
```diff
@@ -406,7 +406,11 @@ When a timeout occurs, the tool is considered to have failed and the model recei
 ### Parallel tool calls & concurrency
 
 When a model returns multiple tool calls in one response, Pydantic AI schedules them concurrently using `asyncio.create_task`.
-If a tool requires sequential/serial execution, you can pass the [`sequential`][pydantic_ai.tools.ToolDefinition.sequential] flag when registering the tool, or wrap the agent run in the [`with agent.sequential_tool_calls()`][pydantic_ai.agent.AbstractAgent.sequential_tool_calls] context manager.
+You can control tool call execution using [`with agent.tool_calls_mode(mode)`][pydantic_ai.agent.AbstractAgent.tool_calls_mode]:
```
DouweM: The recommended way should still be the `sequential` flag. I'd mention the context manager as "...or wrap in `agent.tool_calls_mode('sequential')`", and then have the docstring explain the 3 options.

pydantic_ai_slim/pydantic_ai/_tool_manager.py
```diff
@@ -20,7 +20,9 @@
 from .toolsets.abstract import AbstractToolset, ToolsetTool
 from .usage import RunUsage
 
-_sequential_tool_calls_ctx_var: ContextVar[bool] = ContextVar('sequential_tool_calls', default=False)
+ToolCallsMode = Literal['parallel', 'sequential', 'parallel_ordered_events']
```
DouweM: Since this only affects "parallel tool calls" made by the model in one step, I'd like to call it `ParallelExecutionMode`

pydantic_ai_slim/pydantic_ai/_tool_manager.py
```diff
@@ -77,10 +94,14 @@ def tool_defs(self) -> list[ToolDefinition]:
...
     def should_call_sequentially(self, calls: list[ToolCallPart]) -> bool:
         """Whether to require sequential tool calls for a list of tool calls."""
-        return _sequential_tool_calls_ctx_var.get() or any(
+        return _tool_calls_mode_ctx_var.get() == 'sequential' or any(
             tool_def.sequential for call in calls if (tool_def := self.get_tool_def(call.tool_name))
         )
 
+    def should_order_events(self) -> bool:
+        """Whether to wait for all parallel tool calls to complete before yielding events."""
+        return _tool_calls_mode_ctx_var.get() == 'parallel_ordered_events'
```
DouweM: I'd rather have a method that returns `_tool_calls_mode_ctx_var.get()` so we can directly check the value in the agent graph.

DouweM: That method can then also take into account the `tool_def.sequential` stuff and return `'sequential'` in that case

pydantic_ai_slim/pydantic_ai/_tool_manager.py
```diff
@@ -40,13 +42,28 @@ class ToolManager(Generic[AgentDepsT]):
 
     @classmethod
     @contextmanager
-    def sequential_tool_calls(cls) -> Iterator[None]:
-        """Run tool calls sequentially during the context."""
-        token = _sequential_tool_calls_ctx_var.set(True)
+    def tool_calls_mode(cls, mode: ToolCallsMode = 'parallel') -> Iterator[None]:
```
DouweM: Related to the above, I'd like `parallel_execution_mode`

pydantic_ai_slim/pydantic_ai/agent/abstract.py
```diff
@@ -1172,9 +1173,24 @@ def _infer_name(self, function_frame: FrameType | None) -> None:
 
     @staticmethod
     @contextmanager
+    def tool_calls_mode(mode: _tool_manager.ToolCallsMode = 'parallel') -> Iterator[None]:
```
DouweM: See above, `parallel_tool_call_execution_mode` is very verbose, but I kind of like the explicitness

docs/tools-advanced.md
qianl15: Updated the doc.

pydantic_ai_slim/pydantic_ai/_tool_manager.py
qianl15: Updated.

pydantic_ai_slim/pydantic_ai/_tool_manager.py
qianl15: Sounds good! Changed the logic here and added a `get_parallel_execution_mode()` method.

pydantic_ai_slim/pydantic_ai/_tool_manager.py
qianl15: Updated

pydantic_ai_slim/pydantic_ai/agent/abstract.py
qianl15: Updated

DouweM:
@qianl15 Thanks Qian!

PR #4080, opened 2026-01-24
Fix UsageLimits to preserve explicit `0` token limits

### LINKED ISSUE #4068: UsageLimits should preserve explicit 0 token limits
### Initial Checks

### Description

`UsageLimits` currently uses `or` when selecting between the new `input_tokens_limit` / `output_tokens_limit` arguments and the deprecated `request_tokens_limit` / `response_tokens_limit` arguments.

Because `0` is falsy in Python, setting `input_tokens_limit=0` (or `output_tokens_limit=0`) is not preserved and can be replaced by the deprecated fallback value (or end up as `None`). This makes it impossible to intentionally configure a strict “0 tokens allowed” limit, and is inconsistent with how numeric limits are normally interpreted (where `0` is a valid explicit value).

Expected behavior:
- Only fall back to deprecated args when the new args are `None`, not when they are `0`.

```Python
from pydantic_ai.usage import UsageLimits

# Expect: input_tokens_limit == 0 (explicitly set)
# Actual (before fix): may fall back / be replaced due to `or` semantics.
limits = UsageLimits(input_tokens_limit=0, request_tokens_limit=123)

print(limits.input_tokens_limit)
# Expected: 0
# Actual: 123 (or other unintended value)
```

### Logfire Trace

N/A

- Python: 3.12
- Pydantic AI: latest main
- LLM provider SDK: N/A

Goldokpa:
Fixes #4068

- `UsageLimits.init` previously used `or`, which treats `0` as falsy and could override explicit limits.
- Use `None` checks so `0` is respected.

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/usage.py
+++ b/pydantic_ai_slim/pydantic_ai/usage.py
@@ -344,8 +344,8 @@ class UsageLimits:
     ):
         self.request_limit = request_limit
         self.tool_calls_limit = tool_calls_limit
-        self.input_tokens_limit = input_tokens_limit or request_tokens_limit
-        self.output_tokens_limit = output_tokens_limit or response_tokens_limit
+        self.input_tokens_limit = request_tokens_limit if input_tokens_limit is None else input_tokens_limit
+        self.output_tokens_limit = response_tokens_limit if output_tokens_limit is None else output_tokens_limit
         self.total_tokens_limit = total_tokens_limit
         self.count_tokens_before_request = count_tokens_before_request
 


adtyavrdhn:
Thanks @Goldokpa 
Yeah, It makes sense although 0 would be the only explicitly set limit that would get overridden which doesn't change anything?
What sort of scenario were you facing or had in mind where this could be an issue?

DouweM:
Closing because https://github.com/pydantic/pydantic-ai/issues/4068#issuecomment-3790639290 wasn't answered first

PR #4085, opened 2026-01-25
Improve tool call order preservation behavior for `end_strategy=exhaustive` and support multi-round deferred tool handling

### LINKED ISSUE #3791: Output tools and function tools are not executed in parallel tool call order when end strategy is exhaustive
### Description

The `EndStrategy` parameter controls how the final result tool is handled. It supports two possible values, `early` and `exhaustive`.

https://github.com/pydantic/pydantic-ai/blob/ad5ba4f17834aaf1469bae8f87c7cb5316e65094/pydantic_ai_slim/pydantic_ai/agent/__init__.py#L122-L127

https://github.com/pydantic/pydantic-ai/blob/ad5ba4f17834aaf1469bae8f87c7cb5316e65094/pydantic_ai_slim/pydantic_ai/_agent_graph.py#L865-L866

Both options run output tools before all the other tools. This didn't match with my intuition of how things should work. Since `output_validator` and `output_type` are handled as output tools, this design makes agents run validation before handling any other tools.

I need an option to make output tools run in order or defer the output handling until all the other tool calls are complete. I'd like to learn what's the motivation behind the current design, since calling the output tools last feels like a better default to me.

## Example Use Case

```python

from pydantic_ai import Agent
from pydantic_ai.messages import ModelMessage, ModelResponse, ToolCallPart

events: list[str] = []


class Result(BaseModel):
    result: str


def model_fn(_: list[ModelMessage], info: AgentInfo) -> ModelResponse:
    assert info.output_tools is not None
    output_tool_name = info.output_tools[0].name
    return ModelResponse(
        parts=[
            ToolCallPart("tool_a", {}),
            ToolCallPart("tool_b", {}),
            ToolCallPart(output_tool_name, {"result": "done"}),
        ]
    )


# Need a way to specify new EndStrategy here
agent = Agent(FunctionModel(model_fn), output_type=Result)


@agent.tool_plain(sequential=True)
def tool_a() -> str:
    events.append("tool_a")
    return "a"


@agent.tool_plain(sequential=True)
def tool_b() -> str:
    events.append("tool_b")
    return "b"


@agent.output_validator
def record_output(data: Result) -> Result:
    print(events)
    # Verify a and b are both called before
    # In production code, a mismatch would be `raise ModelRetry(...)`
    assert events == ["tool_a", "tool_b"]
    return data


result = agent.run_sync("Call tool A and B before the final result")
assert result.output == Result(result="done")
```

(`events` is probably better to be included as part of agent context, but I used a shared variable for the sake of simplicity)

Execution Result:
```
[]
...
    assert events == ["tool_a", "tool_b"]
AssertionError
```

Imagine an agent that was instructed to call tool A and B before it finishes. If the agent returns `tool A`, `tool B`, `final_result` in order, the validation will fail.

### References

_No response_

qwaz-theori:
- Closes #3791

### Pre-Review Checklist

    - It has a breaking change for `end_strategy=exhaustive`, but it was discussed in the issue discussion to [treat the current behavior as a bug](https://github.com/pydantic/pydantic-ai/issues/3791#issuecomment-3729181000): "Yeah I agree. So let's treat the current behavior as a bug, change the behavior to respect the order, and update the docs. If anyone complains, we can introduce a new end_strategy option for the current output-then-functions behavior."

### Pre-Merge Checklist

    - EDIT: Working on 100% coverage.

## Summary

- **Exhaustive mode now executes tool calls in model response order** - Tools are processed sequentially in the order the model returned them, rather than grouping all output tools before other tools
- **Early mode better preserves tool order among non-output tools** - While output tools are still prioritized, the relative order of remaining tools is maintained
- **Support for multiple deferred tool request rounds** - When deferred tools are interspersed with output tools, the agent can now return `DeferredToolRequests` multiple times before delivering the final result
- **State tracking via message history** - The agent now infers execution state from the message history itself instead of relying on caller-provided `deferred_tool_results` field

I added several examples in the documentation to illustrate the changes and pre-existing edge cases. I wish they serve as a good starting point for reviewing the changes.

### 1. Tool Call Grouping

Introduces a new tool call grouping abstraction (`ToolCallGroup`) that organizes tool calls based on the end strategy:

- **`early` strategy**: Output tools are pulled to the front and executed first. If one succeeds, remaining tool calls are skipped. The order among non-output tools is preserved.
- **`exhaustive` strategy**: Tool order from the model response is preserved. Output tools act as group boundaries, splitting non-output tool runs into separate processing groups.

When `sequential=True` tools are present, each tool call becomes its own group, ensuring one-by-one execution.

### 2. Multi-Round Deferred Tool Handling

Previously, deferred tool calls could only pause execution once per `ModelResponse`. The new design processes tool calls group-by-group:

1. Agent encounters a deferred tool group → returns `DeferredToolRequests`
2. Caller provides results → agent resumes from the next group
3. If another deferred group exists → returns `DeferredToolRequests` again
4. Continues until all groups are processed

This enables scenarios like: `deferred_A → output_B → deferred_C` in exhaustive+sequential mode, where the agent must pause twice to preserve the tool call order.

Note: This is only supported for non-streaming mode. See the updated docs for details and examples.

### 3. State Detection via Message History

Previously, `ModelRequest` was appended to the message history only when there are one or more tool results. This design has a problem that two agent states, "the agent received the model response and stopped there" and "the agent executed tool calls, and returned deferred tool requests to the user" had the same representable state.

This PR generalizes on the mechanism for resuming the deferred tool cal. The agent now uses the message history structure to track execution state:

- **`ModelResponse` as last message**: Tool calls in that response have not been processed yet
- **`ModelRequest` as last message**: Tool calls have been partially or fully processed
  - Compares tool call IDs in the request vs response to determine which tools remain
  - Empty `ModelRequest` indicates the agent paused at the first deferred tool group

This PR changes the representation of the second case to have an empty model request at the end, and some test snapshots were updated to include this empty `ModelRequest` once the execution returns from the agent.


tests/test_streaming.py
```diff
@@ -1594,6 +1594,12 @@ def regular_tool(x: int) -> int:  # pragma: no cover
                             tool_call_id=IsStr(),
                             timestamp=IsNow(tz=datetime.timezone.utc),
                         ),
+                        ToolReturnPart(
+                            tool_name='external_tool',
+                            content='Tool not executed - a final result was already processed.',
+                            tool_call_id=IsStr(),
+                            timestamp=IsDatetime(),
+                        ),
```
qwaz-theori, line 1602: Hmm actually I think this tool return shouldn't appear here. I'll take a look later.

tests/models/test_outlines.py
```diff
@@ -649,15 +649,6 @@ def test_input_format(transformers_multimodal_model: OutlinesModel, binary_image
...
 
-    # unsupported: tool returns
-    tool_return_message_history: list[ModelMessage] = [
-        ModelRequest(
-            parts=[ToolReturnPart(tool_name='get_location', content='London', tool_call_id='1')], timestamp=IsDatetime()
-        )
-    ]
-    with pytest.raises(UserError, match='Tool calls are not supported for Outlines models yet.'):
-        agent.run_sync('How are you doing?', message_history=tool_return_message_history)
-
```
qwaz-theori, line 660: Note: I removed the test because the case doesn't really make sense (tool result is provided without corresponding model response), but let me know if this needs to be preserved. It raises a different error message now; it fails early at `UserPromptNode` with "Last model request indicates tool calls were made, but there is no corresponding model response.".

dsfaccini:

```bash
git pull

deactivate

rm -rf .venv

make install
```

DouweM:
@qwaz-theori I haven't reviewed this yet because https://github.com/pydantic/pydantic-ai/issues/3791#issuecomment-3801716699 may affect it; let me know what you think of my notes there.

qwaz-theori:
Thanks! I made a reply there.

DouweM:
@qwaz-theori I have yet to respond there but I'm going to have our new auto-review agent do a pass over the PR.

Edit: Heh, it's not running because of the protection against edits to `AGENTS.md` for security reasons, which is being tripped by the change to `docs/agents.md`. If you merge `main` (and resolve conflicts by moving your changes to the newly renamed `docs/agent.md`), auto-review should run again.

qwaz:
Hi, following up from my previous account as mentioned. Unfortunately, I haven't been able to find the time to finish the remaining work, and I don't think I'll be able to in the foreseeable future. I'm going to close this PR for now.

Since I can't close the PR from this account, could a maintainer close it when you get a chance? If anyone wants to pick up where I left off, feel free to use my branch as a starting point.

DouweM:
Thanks for your patience @qwaz! I'm addressing this in #5339, see https://github.com/pydantic/pydantic-ai/issues/3791#issuecomment-4410071528 for more context :)

The multi-round-deferred ambition is intentionally descoped. For that use case, `parallel_tool_calls=False` (or the new inline `HandleDeferredToolCalls` callback) give the same effect with less framework complexity, and we'd rather not bake the multi-round resume semantics in until there's clearer real-world demand for them.

PR #4087, opened 2026-01-25
feat(ui): add `UIAdapter.manage_system_prompt` + `ReinjectSystemPrompt` capability

### LINKED ISSUE #1646: Dynamic System prompts not being included on fresh runs with message_history
### Initial Checks

### Description

As mentioned in the title, the documentation implies that dynamic prompts are always evaluated even if message_history is present. So far, logging all_messages() and putting a bunch of print statemetns all over the place seems to imply that they don't get included

The LLM model responses also suggest that the content from dynamic system prompts is not being included
(I use Gemini models)

### Example Code

```Python

```

```Text
pydantic-ai 0.1.9
Gemini 2.0
```

### LINKED ISSUE #3315: `system_prompt` is ignored when using UI adapter (AG-UI, Vercel AI) or `to_web`
### Initial Checks

### Description

Following this example to implement the AG-UI protocol:
```
agent = Agent('openai:gpt-5')

app = FastAPI()


@app.post('/')
async def run_agent(request: Request) -> Response:
    accept = request.headers.get('accept', SSE_CONTENT_TYPE)
    try:
        run_input = AGUIAdapter.build_run_input(await request.body())  # 
    except ValidationError as e:
        return Response(
            content=json.dumps(e.json()),
            media_type='application/json',
            status_code=HTTPStatus.UNPROCESSABLE_ENTITY,
        )

    adapter = AGUIAdapter(agent=agent, run_input=run_input, accept=accept)
    event_stream = adapter.run_stream() # 

    sse_event_stream = adapter.encode_stream(event_stream)
    return StreamingResponse(sse_event_stream, media_type=accept)
```
If I pass system_prompt directly to the agent or if I use the decorator like this, the agent is executed without the system_prompt:
```
@agent.system_prompt
def get_agent_instructions(ctx: RunContext[str]) -> str:
    print('----LOADING PROMPT----')  # this is not printed
    return 'Be fun!'
```

Instead if I use the `instructions` parameter, agent return a compliant response with the prompt.

Is it possible that `AGUIAdapter` does not consider system_prompt parameter during execution?

Thanks.

dsfaccini:
- Closes #3315

## Summary

Two changes, both composing around a new capability.

### `manage_system_prompt` on `UIAdapter`

New `manage_system_prompt: Literal['server', 'client'] = 'server'` on [`UIAdapter`][pydantic_ai.ui.UIAdapter] (and the `handle_ag_ui_request` / `run_ag_ui` convenience functions, and `VercelAIAdapter`).

- **`'server'`** (default): any `SystemPromptPart` sent by the frontend is stripped with a warning, since a malicious client could otherwise inject arbitrary instructions via crafted API requests. The agent's own `system_prompt` is then reinjected at the head of the first request via the [`ReinjectSystemPrompt`][pydantic_ai.capabilities.ReinjectSystemPrompt] capability (added automatically in this mode).
- **`'client'`**: frontend `SystemPromptPart`s are preserved as-is. The agent's configured `system_prompt` is not injected — the caller is fully responsible for sending it on every turn if desired. To opt into the fallback behavior anyway, add `ReinjectSystemPrompt` to the agent's capabilities.

If you want per-request guidance that doesn't need to live in message history, prefer `instructions` over `system_prompt`.

### `ReinjectSystemPrompt` capability

New capability at [`pydantic_ai.capabilities.ReinjectSystemPrompt`][pydantic_ai.capabilities.ReinjectSystemPrompt]. Ensures the agent's configured `system_prompt` is at the head of the first `ModelRequest` on every model request. If any `SystemPromptPart` is already present anywhere in the history (preserved from a prior run, handed off from another agent, or sent by a client), the capability is a no-op.

Useful for any scenario where `message_history` comes from a source that doesn't round-trip system prompts — UI frontends, database persistence layers, conversation compaction pipelines. Community users have asked for this repeatedly on #1646; shipping it as an opt-in capability avoids breaking the documented contract of `agent.run(message_history=[...])` while giving people a single one-line opt-in.

Supporting infrastructure:

- New public `Agent.system_prompts(...)` method on [`AbstractAgent`][pydantic_ai.agent.AbstractAgent] that resolves configured static strings + runner functions into `SystemPromptPart`s, reusable by anyone who needs to resolve system prompts outside an agent run.
- New `capabilities=` run-time kwarg on `Agent.run` / `Agent.iter` / `Agent.run_stream` / `Agent.run_stream_events` — additive with the agent's configured capabilities (matches the `toolsets=` / `builtin_tools=` pattern). Used internally by `UIAdapter` to opt into the capability in server mode.

A `# TODO (v2):` comment marks the capability for potential default inclusion in v2 — which would close #1646 directly, at the cost of a documented behavior change to `agent.run(message_history=[...])`.

## Behavior change (security-relevant)

Frontend-originated `SystemPromptPart`s are now **stripped by default** in UI adapter flows and a warning is emitted. Rationale: a malicious client could otherwise inject arbitrary instructions via crafted API requests. The agent's configured `system_prompt` is authoritative in this default mode and is reinjected when missing.

Callers who intentionally let the frontend own the system prompt must opt in with `manage_system_prompt='client'`.

## Pre-Review Checklist

- [x] **Linting and type checking** pass per \`make format\` and \`make typecheck\`.

## Pre-Merge Checklist

- [x] Docs updated: \`docs/capabilities.md\` documents the capability with an example; \`docs/message-history.md\` points to the capability for histories missing sys_parts; \`docs/ui/ag-ui.md\` and \`docs/ui/vercel-ai.md\` reframed around the \`'server'\` / \`'client'\` ownership model.


pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -263,6 +263,14 @@ async def run(  # noqa: C901
 
         if next_message:
             await self._reevaluate_dynamic_prompts([next_message], run_context)
+            # Check if we should add system prompts for the initial message
+            # This happens forn instance when a UI adapter passes a first user message but no prior model responses
+            if not messages or not any(isinstance(msg, _messages.ModelResponse) for msg in messages):
```
DouweM: I think we can just have it be `if not messages`, same as in the branch below. Having multiple model requests, but no responses, that didn't get joined by `clean_message_history` seems like a weird edge case -- and would be inconsistent with the logic below.

DouweM: Note that this doesn't quite fully resolve the problem described in the issue, because we'll now be inserting system prompts for the initial user prompt request coming from the frontend, but when we stream back the assistant response, and the user does a followup, we'll get 3 parts: user, assistant, user, without system prompts as they were never sent to the frontend. So then we wouldn't insert them anymore...

So while this fix makes sense for the "single-message message history" scenario, I'm thinking now that it could be misleading in making people think things are working fine on the first prompt, and then highly surprisingly the system prompt is actually dropped on followups.

Any ideas? :)

We may have to say that UI adapters and `system_prompt` are fundamentally incompatible, and users should use `instructions` instead.

Or at the very least if we make a fix here that goes against existing system prompt/instructions behavior, we should make it happen only in UI adapters?

tests/test_ag_ui.py
```diff
@@ -2345,3 +2346,98 @@ async def send(data: MutableMapping[str, Any]) -> None:
...
+        tools=[],
+        forwarded_props=None,
+    )
+
+    with capture_run_messages() as messages:
+        async for _ in run_ag_ui(agent, run_input):
+            pass
+
+    # Check the last ModelRequest (which should be for "Second message")
+    last_request = next((msg for msg in reversed(messages) if hasattr(msg, 'parts')), None)
```
DouweM: the `hasattr` is very ugly.

tests/test_ag_ui.py
```diff
@@ -2345,3 +2346,98 @@ async def send(data: MutableMapping[str, Any]) -> None:
...
+    assert last_request is not None
+
+    # system prompt should only be in the initial history, not repeated
+    system_prompt_parts = [part for part in last_request.parts if isinstance(part, SystemPromptPart)]
+    assert len(system_prompt_parts) == 0, (
```
DouweM: I'd rather have a snapshot, that may also have shown us that the system prompt is not included at all in this case, which we could've identified as unexpected (as I did above)

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/_agent_graph.py
dsfaccini: Addressed both this and the followup concern (Comment B) via a `reinject_system_prompts` flag on `UserPromptNode`. 

The condition is now `if not messages or self.reinject_system_prompts` — when `reinject_system_prompts=False` (default for non-UI paths), it simplifies to just `if not messages` as you suggested. UI adapters set `reinject_system_prompts=True`, which ensures system prompts are reinjected on every request including followups, without affecting non-UI paths.

tests/test_ag_ui.py
dsfaccini: Changed to `isinstance(msg, ModelRequest)` check.

tests/test_ag_ui.py
dsfaccini: Converted all 6 system prompt tests (3 in test_ag_ui.py, 3 in test_vercel_ai.py) to snapshot assertions.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -267,6 +267,15 @@ async def run(  # noqa: C901
...
             await self._reevaluate_dynamic_prompts([next_message], run_context)
+            if not any(
+                isinstance(part, _messages.SystemPromptPart)
+                for msg in (*messages, next_message)
+                if isinstance(msg, _messages.ModelRequest)
+                for part in msg.parts
+            ):
+                sys_parts = await self._sys_parts(run_context)
+                if sys_parts:
+                    next_message.parts = [*sys_parts, *next_message.parts]
```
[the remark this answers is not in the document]
dsfaccini: Claude says:

Valid observation. When a frontend sends `SystemMessage`/`DeveloperMessage`, the adapter converts them to `SystemPromptPart` (`_adapter.py:176-177`), and the `any()` check here would detect them and skip injecting the agent's own system prompts.

However, this is a pre-existing design question rather than a regression from this PR — the original code didn't handle system prompts in the `next_message` path at all. A few options:

1. **Acknowledge, don't fix here** — this PR fixes the duplicate system prompt bug; the frontend-SystemMessage interaction is a separate concern that could be tracked in its own issue
2. **Fix in this PR** — change the check to only match agent-originated `SystemPromptPart`s (e.g. by comparing against `_sys_parts` output or checking `dynamic_ref`) so frontend system messages coexist with agent prompts
3. **Add a test only** — document the current behavior explicitly with a test, making the design choice visible without changing it

I'd lean towards (1) — keeping this PR focused on the duplicate fix and deferring the coexistence question to a separate issue where maintainers can weigh in on the intended semantics.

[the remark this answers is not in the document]
dsfaccini: so @DouweM, what do you think is the right behavior? I'm for fixing: the frontend shouldn't override the backend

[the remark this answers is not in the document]
DouweM: @dsfaccini Hmm, AG-UI users may have been working around this limitation by specifying the system prompts in the frontend, so I don't like breaking that, but we had a related discussion with the CopilotKit team: ([private Slack link](https://pydantic.slack.com/archives/C091V4LSFAP/p1770295254090509?thread_ts=1770068327.055779&cid=C091V4LSFAP)) where they recommended rejecting frontend-originated system prompt parts by default, logging a warning if any are seen, and then letting the user explicitly opt into that. There are some security implications (e.g. you usually wouldn't want users to be able to inject system prompts by crafting a special API request), so with that context I'm ok with changing the default behavior.

So combined with this feature, I suggest we add an `accept_frontend_system_prompt: bool = False` flag to `UIAdapter` and use it in both the AG-UI and Vercel implementations:

| accept_frontend_system_prompt | frontend system prompt parts exist | agent system prompt exists | result |
|--|--|--|--|
| False | False | False | nothing |
| False | False | True | agent system prompt |
| False | True | False | warning |
| False | True | True | warning + agent system prompt |
| True | False | False | nothing |
| True | False | True | agent system prompt |
| True | True | False | frontend system prompt |
| True | True | True | frontend system prompt |

So if `accept_frontend_system_prompt` is True _and_ there are frontend system prompts, they take precedence over agent system prompts. If there are no frontend system prompts, we still use the agent system prompt.

You can check with Claude if I missed some scenario there lol.

pydantic_ai_slim/pydantic_ai/ui/_adapter.py
```diff
@@ -125,6 +125,17 @@ class UIAdapter(ABC, Generic[RunInputT, MessageT, EventT, AgentDepsT, OutputData
...
+    accept_frontend_system_prompt: bool = False
+    """Whether to accept system prompts sent by the frontend.
+
+    When `False` (default), any `SystemPromptPart` found in the frontend messages will be stripped
+    and a warning emitted. The agent's own system prompt (if any) will be used instead.
+    This is a security measure to prevent users from injecting system prompts via crafted API requests.
+
+    When `True`, frontend system prompts are accepted and take precedence over the agent's
+    configured system prompt.
+    """
```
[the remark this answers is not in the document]
dsfaccini: yeah docs now definitely have a reason to be updated

tests/test_ag_ui.py
```diff
@@ -2399,3 +2403,373 @@ async def send(data: MutableMapping[str, Any]) -> None:
...
+            ModelResponse(
+                parts=[TextPart(content='success (no tool calls)')],
+                usage=RequestUsage(input_tokens=54, output_tokens=4),
+                model_name='test',
+                timestamp=IsDatetime(),
+                provider_name='test',
+                run_id=IsStr(),
+            ),
+        ]
+    )
```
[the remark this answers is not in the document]
dsfaccini: we can probably make a parametrized test out of this. single-turn isn't necessary if multi-turn is setup and assertions are made at the right place

tests/test_ag_ui.py
```diff
@@ -2399,3 +2403,373 @@ async def send(data: MutableMapping[str, Any]) -> None:
...
+    adapter = AGUIAdapter(agent=agent, run_input=run_input)
+
+    with capture_run_messages() as messages:
+        with pytest.warns(UserWarning, match='accept_frontend_system_prompt'):
+            async for _ in adapter.encode_stream(adapter.run_stream()):
+                pass
+
+    assert not any(
+        isinstance(part, SystemPromptPart) for msg in messages if isinstance(msg, ModelRequest) for part in msg.parts
+    )
```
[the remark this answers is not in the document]
dsfaccini: makes sense, thanks

docs/ui/ag-ui.md
```diff
@@ -282,6 +282,24 @@ Since `app` is an ASGI application, it can be used with any ASGI server:
 uvicorn ag_ui_tool_events:app --host 0.0.0.0 --port 9000
 ```
 
+### System prompts
+
+Agents configured with `system_prompt` or `@agent.system_prompt` work as expected with AG-UI — the agent's system prompts are injected into the request sent to the model.
```
DouweM: - We should link to the system prompt / instructions docs
- We should make it clear how the differences between system prompts and instructions matter here
- I don't think we need this "work as expected" line -- it's expected that they work as expected 😄  and this is only relevant to people who are aware of the non-expected buggy current behavior. So this section should be more about the considerations below, and less about the expected behavior.

docs/ui/vercel-ai.md
```diff
@@ -146,3 +146,21 @@ When `sdk_version=6`, the adapter will:
...
+
+## System Prompts
+
+Agents configured with `system_prompt` or `@agent.system_prompt` work as expected with the Vercel AI adapter — the agent's system prompts are injected into the request sent to the model.
+
+By default, system prompts sent by the frontend are **stripped** for security, since a malicious client could inject arbitrary instructions via crafted API requests. A warning is emitted when this happens.
+
+To accept frontend system prompts instead, pass `accept_frontend_system_prompt=True`:
+
+```python {test="skip" lint="skip"}
```
DouweM: Don't like the skipping here :)

pydantic_ai_slim/pydantic_ai/_agent_graph.py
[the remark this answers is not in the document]
DouweM: I think this logic should live inside the `UIAdapter` only. That also makes it easier to verify that the behavior matches the table here: https://github.com/pydantic/pydantic-ai/pull/4087#discussion_r2813772858

DouweM (changes requested):

dsfaccini:
Claude here: Addressed by gating the system prompt reinjection logic behind a `ContextVar` (`_reinject_system_prompts`) in `_agent_graph.py`. The UIAdapter sets it to `True` in `run_stream_native` before iterating the event stream; the default is `False`, so non-UI paths are completely unaffected.

This avoids threading a private parameter through the public `run_stream_events` / `run` / `iter` chain. The `_run_stream_events` method creates a task via `asyncio.create_task`, which copies the current context — so the ContextVar propagates correctly through the entire execution path.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -1748,6 +1756,8 @@ class _RunMessages:
 
 _messages_ctx_var: ContextVar[_RunMessages] = ContextVar('var')
 
+reinject_system_prompts: ContextVar[bool] = ContextVar('reinject_system_prompts', default=False)
```
dsfaccini: Claude here: nit: this should be `_reinject_system_prompts` (with leading underscore) to match the adjacent `_messages_ctx_var` naming convention. Both are internal-only symbols in a private module.

tests/test_ag_ui.py
```diff
@@ -40,6 +40,7 @@
     UserPromptPart,
     VideoUrl,
 )
+from pydantic_ai._agent_graph import capture_run_messages
```
dsfaccini: Claude here: `capture_run_messages` is re-exported from the public `pydantic_ai` package. Other test files (e.g. `test_agent.py`, `test_streaming.py`) import it from there:

```python
from pydantic_ai import capture_run_messages
```

Same applies to the import in `tests/test_vercel_ai.py:12`.

dsfaccini (commented):
Claude here: self-review notes — two minor suggestions, no blocking issues.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
dsfaccini: @DouweM claude wrote it as private before and I said `no private functions in private modules` but it does get confused because there are indeed other private functions in private modules

I'd leave as is, just wanted to raise this. Could be an improvement to our development rules, or we could just forget about it and let claude write private functions in private modules

tests/test_ag_ui.py
dsfaccini: not sure why claude thought this comment was important, leaving it here in case it is

dsfaccini:
Claude here: Branch review of `fix-toweb-system-prompt` against `main`.

**Note:** Most of the 76-file diff is from the upstream merge (HTTP client lifecycle #4421, AG-UI 0.1.13/0.1.15 #3971, force streaming #5028, contributing docs #5006). Review focuses on the 10 PR-specific commits.

### Pattern Issues — 2 actionable

1. **`pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py:467`** — Missing `stacklevel` in `warnings.warn(f'AG-UI message type {type(msg).__name__} is not yet implemented; skipping.', UserWarning)`. Without it, the warning points to internal `_adapter.py` rather than the caller.
   → Add `stacklevel=2`.

2. **`tests/test_vercel_ai.py`** — `test_system_prompt_with_vercel_adapter`, `test_dynamic_system_prompt_with_vercel_adapter`, and `test_system_prompt_reinjected_with_vercel_history` collect events into `events: list[dict[str, Any]]` but never assert on them. The AG-UI equivalents correctly discard events.
   → Either assert on `events` or discard with `async for _ in ...`.

### Pitfall Check — 0 violations

### Design — 0 blocking / 2 warnings

**ContextVar coupling between UI adapter and agent graph**
`reinject_system_prompts` ContextVar in `_agent_graph.py:1759` is set from `_adapter.py:315`. This is a side-channel between the UI layer and `UserPromptNode._run`. However, it follows the established pattern (`_messages_ctx_var` is defined right next to it), scope is well-contained with `set()`/`reset()` in `try`/`finally`, and the alternative (threading a parameter through the entire call stack) would be far more invasive.
→ No action needed. Worth a brief inline comment near the ContextVar definition noting it's set by UI adapters.

**`_set_http_client` accesses SDK private internals (upstream #4421)**
Multiple providers reach into `self._client._client` to swap HTTP clients. Any SDK version bump could silently break re-entry after close.
→ Not this PR's responsibility. No action needed here.

### Verdict: PASS after fixing the 2 pattern issues above.

dsfaccini:
Claude here: Pushed two fixes from the branch review:

1. **Added `stacklevel=2`** to `warnings.warn()` in `AGUIAdapter.load_messages` (`_adapter.py:467`) — the warning for unrecognized AG-UI message types now points to the caller instead of the internal adapter line.

2. **Removed unused event collection** in 3 Vercel tests (`test_system_prompt_with_vercel_adapter`, `test_dynamic_system_prompt_with_vercel_adapter`, `test_system_prompt_reinjected_with_vercel_history`) — these tests collected streamed events into a list but never asserted on them. Changed to `async for _ in ...` to just drain the stream, matching the AG-UI test equivalents. These tests only need to verify the messages sent to the model (via `capture_run_messages`), not the streamed output.

dsfaccini:
Claude here: addressed the three huddle items from 2026-04-10 in 575fb775.

1. **Renamed the field** from `accept_frontend_system_prompt: bool` to `manage_system_prompt: Literal['server', 'client']` on `UIAdapter` and the `handle_ag_ui_request` / `run_ag_ui` convenience functions. `'accept'` implied the prompts were optional for the opt-in case; `'server'` / `'client'` makes ownership explicit and reserves room for future modes.

2. **Wrapped the raw `ContextVar.set()`/`reset()`** in a `reinject_system_prompts()` context manager in `_agent_graph.py`. The adapter now does `with reinject_system_prompts() if self.manage_system_prompt == 'server' else nullcontext()`, so there's no raw set/reset anywhere outside the helper.

3. **`'client'` mode no longer silently falls back to reinjection.** When the caller opts into client ownership and the frontend omits a system prompt, the agent-configured `system_prompt` is no longer injected as a safety net — the caller asked to own it, so we don't second-guess. `'server'` mode (the default) is unchanged.

Behavior-change callout worth flagging to release notes: frontend-originated `SystemPromptPart`s are **stripped by default** and a warning is emitted, because a malicious client could otherwise inject arbitrary instructions. Callers who relied on implicit acceptance need to set `manage_system_prompt='client'`.

Docs at `docs/ui/ag-ui.md` and `docs/ui/vercel-ai.md` were reframed around the server/client ownership model and now link to `[system_prompt][pydantic_ai.Agent.system_prompt]` and `[instructions][pydantic_ai.Agent.instructions]`. New snapshot tests in `test_ag_ui.py` and `test_vercel_ai.py` cover the `'client'`-mode no-inject behavior.

pydantic_ai_slim/pydantic_ai/ui/_adapter.py
```diff
@@ -125,6 +128,21 @@ class UIAdapter(ABC, Generic[RunInputT, MessageT, EventT, AgentDepsT, OutputData
...
     """The `Accept` header value of the request, used to determine how to encode the protocol-specific events for the streaming response."""
 
+    manage_system_prompt: Literal['server', 'client'] = 'server'
+    """Who owns the system prompt.
+
+    `'server'` (default): the agent's configured `system_prompt` is authoritative.
+    Any `SystemPromptPart` found in frontend messages is stripped and a warning
+    is emitted, since a malicious client could otherwise inject arbitrary
+    instructions via crafted API requests. The agent's system prompt is
+    injected into the history whenever it is missing.
```
DouweM: This last sentence is incorrect right? "injected whenever it is missing" suggests that this is not the case if system prompts are already in the history, as if we would then respect those -- while we actually still filter them out.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -275,6 +275,15 @@ async def run(  # noqa: C901
 
         if next_message:
             await self._reevaluate_dynamic_prompts([next_message], run_context)
+            if _reinject_system_prompts_var.get() and not any(
+                isinstance(part, _messages.SystemPromptPart)
+                for msg in (*messages, next_message)
+                if isinstance(msg, _messages.ModelRequest)
+                for part in msg.parts
+            ):
```
DouweM: Do we need the `not any(...)` check? It means that we ignore the "reinject" notice if the message history already had system parts, respecting the incoming system parts, while the issue we're trying to solve is exactly that we should _not_ be respecting incoming system parts. It's fine because we filter those out in the UI adapter, but it's not obvious from the name `reinject_system_prompts` that this is _ignored_ if there are already system prompts. So either we do the filter + reinjection here (that does feel consistent), or we just always reinject here + keep the filter in the UI adapter, but at least we don't ignore the "reinject" setting.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -1774,6 +1782,17 @@ class _RunMessages:
 
 _messages_ctx_var: ContextVar[_RunMessages] = ContextVar('var')
 
+_reinject_system_prompts_var: ContextVar[bool] = ContextVar('_reinject_system_prompts', default=False)
+
+
+@contextmanager
+def reinject_system_prompts() -> Iterator[None]:
```
DouweM: Let's make this a static method on `UserPromptNode`

docs/ui/ag-ui.md
```diff
@@ -282,6 +282,35 @@ Since `app` is an ASGI application, it can be used with any ASGI server:
 uvicorn ag_ui_tool_events:app --host 0.0.0.0 --port 9000
 ```
 
+### System prompts and instructions
+
+When running via AG-UI, you choose who owns the system prompt with the `manage_system_prompt` parameter on [`AGUIAdapter`][pydantic_ai.ui.ag_ui.AGUIAdapter] (and [`handle_ag_ui_request`][pydantic_ai.ag_ui.handle_ag_ui_request] / [`run_ag_ui`][pydantic_ai.ag_ui.run_ag_ui]).
+
+- `'server'` (default): the agent's [`system_prompt`][pydantic_ai.Agent.system_prompt] is authoritative. Any `SystemMessage` found in the AG-UI message history is stripped and a warning is emitted, since a malicious client could otherwise inject arbitrary instructions via crafted API requests. The agent's configured system prompt is injected into the history whenever it is missing.
```
DouweM: Same as above, "whenever it missing" seems misleading here.

docs/ui/ag-ui.md
```diff
@@ -282,6 +282,35 @@ Since `app` is an ASGI application, it can be used with any ASGI server:
...
+- `'client'`: the frontend owns the system prompt. Frontend `SystemMessage`s are preserved, and the agent's configured system prompt is never injected as a fallback — the caller is fully responsible for sending it on every turn.
+
+If you want per-request guidance that doesn't need to live in the message history at all, use [`instructions`][pydantic_ai.Agent.instructions] instead of a system prompt: instructions are always injected fresh on each request and are the recommended default when you control the server side.
```
DouweM: Let's link to the sections on system prompts and instructions as early as possible so that people understand the difference as they read the rest. And then we can make it clear that the manage stuff really only applies to "system prompt" parts. If you're using instructions only, you don't have to think about the system prompt manage at all.

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/ui/_adapter.py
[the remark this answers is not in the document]
dsfaccini: Claude here: Good catch. I fixed this by making manage_system_prompt explicit in the public adapter entrypoints instead of forwarding it only through **kwargs. It is now visible on UIAdapter.from_request, UIAdapter.dispatch_request, AGUIAdapter.from_request, and VercelAIAdapter.from_request / dispatch_request, with regression tests covering the exposed signatures. Pushed in 0fd3e5b0.

PR #4091, opened 2026-01-26
Fix `embedding_type` requirement for cohere

### LINKED ISSUE #4044: cohere embeddings requires an embedding types value
### Initial Checks

### Description

Cohere embedding may require an "embedding_types" attribute when calling the "embed" method of their client in some cases.    It appears this value cannot be set in pydantic-ai interaction with the cohere client.

Of note: This is perhaps really a bug in the cohere client.  The bug appears as a bug in the cohere client code but it only become difficult to work around  because the pydantic-ai client isn't specifying the "embedding_types" value when making the request.

File cohere/core/unchecked_base_model.py line 304 in construct_type
TypeError: cohere.core.unchecked_base_model.UncheckedBaseModel.model_construct() argument after ** must be a mapping, not a list

In Cohere docs:

> embedding_typeslist of enumsOptional
> 
> Specifies the types of embeddings you want to get back. Can be one or more of the following types.
> 
>     "float": Use this when you want to get back the default float embeddings. Supported with all Embed models.
>     "int8": Use this when you want to get back signed int8 embeddings. Supported with Embed v3.0 and newer Embed models.
>     "uint8": Use this when you want to get back unsigned int8 embeddings. Supported with Embed v3.0 and newer Embed models.
>     "binary": Use this when you want to get back signed binary embeddings. Supported with Embed v3.0 and newer Embed models.
>     "ubinary": Use this when you want to get back unsigned binary embeddings. Supported with Embed v3.0 and newer Embed models.
>     "base64": Use this when you want to get back base64 embeddings. Supported with Embed v3.0 and newer Embed models.

A quick fix is the add embedding_types=["float"] to the cohere client embed call.   This is within the "embed" method of CohereEmbeddingModel
```
        try:
            response = await self._client.embed(
                model=self.model_name,
                texts=inputs,
                output_dimension=settings.get('dimensions'),
                input_type=cohere_input_type,
                max_tokens=settings.get('cohere_max_tokens'),
                truncate=settings.get('cohere_truncate', 'NONE'),
                request_options=request_options,
                embedding_types=["float"] # hack: added as workaround
            )
```

Proposed fix would be to allow the embedding_types value to be set. The "float" type may be the default in most cases.

```Python
# may not be easily reproducible 
#fix, add embedding types item.
        try:
            response = await self._client.embed(
                model=self.model_name,
                texts=inputs,
                output_dimension=settings.get('dimensions'),
                input_type=cohere_input_type,
                max_tokens=settings.get('cohere_max_tokens'),
                truncate=settings.get('cohere_truncate', 'NONE'),
                request_options=request_options,
                embedding_types=["float"]
            )
```

### Logfire Trace

_No response_

- Python: 3.11
- Pydantic AI: 1.40.0
- LLM provider SDK: cohere 5.20.1

R-Yash:
- Closes #4044

### Pre-Review Checklist

### Pre-Merge Checklist

### Changes
Added the `embedding_types` parameter to satisfy the updated Cohere client requirements. Set it to `['float']` by default.


DouweM:
@R-Yash Linting is failing and we need a test that would've failed before, and will succeed now that we've made this change. 

Also please investigate why this change was not necessary when I originally implemented this: as you can see, the existing cohere embeddings tests are passing. So it may be a recent change in their API or SDK. I can't merge this unless we understand the root cause of the issue and have a regression test.

dsfaccini:
@R-Yash 

```bash
git pull

deactivate

rm -rf .venv

make install
```

PR #4095, opened 2026-01-26
docs: Add internals section to output docs

dsfaccini:


docs/output.md
```diff
@@ -830,6 +837,103 @@ async def main():
 
 _(This example is complete, it can be run "as is" — you'll need to add `asyncio.run(main())` to run `main`)_
 
+## Internals
```
DouweM, line 840: From Slack:

> I wouldn't even consider this internals; "what part of a function tool / output model are sent to the LLM" is very much worth explaining. Also because it interacts with settings like https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.use_attribute_docstrings, that would be worth explaining

So I wouldn't make it seem like this is about internals at all, nor would I mention explicitly `ToolOutput`, `NativeOutput`, `PromptedOutput` or anything to do with `OutputObjectDefinition` here.

The goal should be to explain what is sent to the model, and how the user can affect that, which includes `use_attribute_docstrings` which is not currently documented.

docs/output.md
```diff
@@ -830,6 +837,103 @@ async def main():
...
+
+*Wrapper classes: [`ToolOutput`][pydantic_ai.output.ToolOutput], [`NativeOutput`][pydantic_ai.output.NativeOutput], [`PromptedOutput`][pydantic_ai.output.PromptedOutput]
+
+**Key points:**
+
+1. **`Field(description='...')` always works**: Pydantic's `TypeAdapter.json_schema()` preserves all field descriptions. These appear in the `properties` of the JSON schema sent to the model, regardless of output mode.
+
+2. **Class docstrings are used**: The model's class docstring becomes the `description` field in the JSON schema. For output types, if you also provide an explicit `description=` parameter, they are combined as `'{explicit}. {docstring}'`.
+
+3. **Constraints are preserved**: Field constraints like `ge`, `le`, `min_length`, `max_length` are also included in the JSON schema.
```
DouweM, line 862: Are you confident they're never filtered out?

docs/output.md
```diff
@@ -830,6 +837,103 @@ async def main():
...
+| Class docstring | ✅ Yes (in schema's `description`) | See below |
+| Explicit `description=` on wrapper* | N/A | ✅ Combined with docstring |
+
+*Wrapper classes: [`ToolOutput`][pydantic_ai.output.ToolOutput], [`NativeOutput`][pydantic_ai.output.NativeOutput], [`PromptedOutput`][pydantic_ai.output.PromptedOutput]
+
+**Key points:**
+
+1. **`Field(description='...')` always works**: Pydantic's `TypeAdapter.json_schema()` preserves all field descriptions. These appear in the `properties` of the JSON schema sent to the model, regardless of output mode.
+
+2. **Class docstrings are used**: The model's class docstring becomes the `description` field in the JSON schema. For output types, if you also provide an explicit `description=` parameter, they are combined as `'{explicit}. {docstring}'`.
```
DouweM, line 860: This list duplicates the table to a degree, I think it's confusing to have both

docs/output.md
```diff
@@ -830,6 +837,103 @@ async def main():
 
 _(This example is complete, it can be run "as is" — you'll need to add `asyncio.run(main())` to run `main`)_
 
+## Internals
+
+This section explains how Pydantic AI handles output types internally. Understanding these details can help you debug unexpected behavior and make informed decisions about how to structure your output types.
+
+### How Descriptions Work
+
+When you define a Pydantic model as an output type, various pieces of metadata are extracted and sent to the model. Here's how different description sources are handled:
```
DouweM, line 846: Just Pydantic model? Or also dataclass? TypedDict etc?

docs/output.md
```diff
@@ -31,27 +31,32 @@ _(This example is complete, it can be run "as is")_
...
+The [`Agent`][pydantic_ai.Agent] class constructor takes an `output_type` argument that accepts:
 
-By default, Pydantic AI leverages the model's tool calling capability to make it return structured data. When multiple output types are specified (in a union or list), each member is registered with the model as a separate output tool in order to reduce the complexity of the schema and maximise the chances a model will respond correctly. This has been shown to work well across a wide range of models. If you'd like to change the names of the output tools, use a model's native structured output feature, or pass the output schema to the model in its [instructions](agents.md#instructions), you can use an [output mode](#output-modes) marker class.
+- Simple scalar types (`int`, `str`, `bool`, etc.)
+- Collection types (`list`, `dict`, `TypedDict`, [`StructuredDict`](#structured-dict))
+- Dataclasses and Pydantic models
+- Type unions and lists of multiple types
```
DouweM, line 39: This misses the nuance that lists of _all of the options_ are supported

docs/output.md
```diff
@@ -311,9 +316,11 @@ _(This example is complete, it can be run "as is" — you'll need to add `asynci
 
 Pydantic AI implements three different methods to get a model to output structured data:
 
-1. [Tool Output](#tool-output), where tool calls are used to produce the output.
-2. [Native Output](#native-output), where the model is required to produce text content compliant with a provided JSON schema.
-3. [Prompted Output](#prompted-output), where a prompt is injected into the model instructions including the desired JSON schema, and we attempt to parse the model's plain-text response as appropriate.
+| Mode | Mechanism | Model Support | Best For |
+|------|-----------|---------------|----------|
+| [Tool Output](#tool-output) (default) | Schema as tool parameters | Nearly all models | General use, multiple output types |
+| [Native Output](#native-output) | `response_format` JSON schema | OpenAI, Gemini, others | Strict schema compliance |
```
DouweM, line 322: This implies tool output does not have strict schema compliance

docs/output.md
```diff
@@ -311,9 +316,11 @@ _(This example is complete, it can be run "as is" — you'll need to add `asynci
 
 Pydantic AI implements three different methods to get a model to output structured data:
 
-1. [Tool Output](#tool-output), where tool calls are used to produce the output.
-2. [Native Output](#native-output), where the model is required to produce text content compliant with a provided JSON schema.
-3. [Prompted Output](#prompted-output), where a prompt is injected into the model instructions including the desired JSON schema, and we attempt to parse the model's plain-text response as appropriate.
+| Mode | Mechanism | Model Support | Best For |
+|------|-----------|---------------|----------|
+| [Tool Output](#tool-output) (default) | Schema as tool parameters | Nearly all models | General use, multiple output types |
```
DouweM, line 321: - "Nearly all models"?
- "multiple output types" implies the others don't support that

docs/output.md
```diff
@@ -830,6 +837,103 @@ async def main():
...
 
+## Internals
+
+This section explains how Pydantic AI handles output types internally. Understanding these details can help you debug unexpected behavior and make informed decisions about how to structure your output types.
+
+### How Descriptions Work
+
+When you define a Pydantic model as an output type, various pieces of metadata are extracted and sent to the model. Here's how different description sources are handled:
+
+| Source | Always included in schema? | Used as tool/output description? |
```
DouweM, line 848: tool/output makes me think that we should explain a bunch of this on the Tool docs as well, as it also applies to types used in the arg hints

PR #3917, opened 2026-01-05
Add `Model.model_id` property that returns `provider:model`

tdevelope:
- Closes #3908

### Summary
This PR introduces a new **abstract property** `fully_qualified_model_name` in the base `Model` class, and implements it in all concrete Model subclasses.  
The property returns the model name **prefixed by its provider**, e.g., `"openai:gpt-4"` instead of just `"gpt-4"`.  

This helps unify model identification across different providers and simplifies workflows where fully qualified names are required.

### Changes
- Added `@property` `fully_qualified_model_name` in the base `Model` class.
- Implemented property in all concrete subclasses:
  - `OpenAIChatModel`, `OpenAIResponsesModel`
  - `AnthropicModel`, `CohereModel`
  - `MistralModel`, `GroqModel`, `GeminiModel`, `GoogleModel`
  - `BedrockConverseModel`, `HuggingFaceModel`
  - `TestModel`, `FunctionModel`, `MCPSamplingModel`, `OutlinesModel`
  - `WrapperModel` delegates to wrapped model
  - `FallbackModel` combines wrapped models
- Added unit tests to verify `fully_qualified_model_name` in each subclass.
- Added docstrings for clarity.

#### Pre-Review
- [x] AI-generated code reviewed line-by-line.
- [x] No breaking changes per [version policy](https://github.com/pydantic/pydantic-ai/blob/main/docs/version-policy.md).
- [x] Linting & type checks pass (`make format`, `make typecheck`).
- [x] PR title is clear and suitable for release notes.

#### Pre-Merge
- [x] New unit tests added for the property.
- [x] Documentation updated for new features and behaviors (docstrings included).

PR #4100, opened 2026-01-26
Fix builtin tool availability checking when using `TemporalAgent` with multiple models

### LINKED ISSUE #4099: Correct builtin tool validation during temporal workflow with multiple instances
### Initial Checks

### Description

Temporal with multiple models keeps the default models profile when it comes to evaluating the builtin tools. We need to update the model properties and extract out a pure function for getting the profile to ensure the builtins are evaluated correctly.

```Python

```

### Logfire Trace

_No response_

- Python:
- Pydantic AI:
- LLM provider SDK:

mattbrandman:
This is a bug fix to ensure that model names, provider, and profile names are set correctly when utilizing temporal. This also fixes a bug where Agent.instrument_all() was causing issues with temporal agents and builtin tools because the instrumented tools would call prepare_message and have the default wrapped model vs the current runtime one.

### Breaking Change

`Provider.model_profile` has been changed from an instance method to a `@staticmethod`. This is necessary so that `infer_model_profile()` can look up profiles without constructing a provider instance (which would require API credentials). Any custom `Provider` subclass that overrides `model_profile` as `def model_profile(self, model_name)` will need to update the signature to `@staticmethod def model_profile(model_name)`. Per the version policy, this is a necessary breaking change for the bug fix (acceptable under "some apparently safe changes and bug fixes will inevitably break some users' code").

- Closes #4099

### Pre-Review Checklist

  - `Provider.model_profile` changed from instance method to `@staticmethod` (see above)

### Pre-Merge Checklist

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/providers/__init__.py
+++ b/pydantic_ai_slim/pydantic_ai/providers/__init__.py
@@ -193,10 +193,10 @@ def infer_provider(provider: str) -> Provider[Any]:
 
         upstream_provider = provider.removeprefix('gateway/')
         return gateway_provider(upstream_provider)
-    elif provider in ('google-vertex', 'google-gla'):
+    elif provider in ('google-vertex', 'google-gla', 'vertexai'):
         from .google import GoogleProvider
 
-        return GoogleProvider(vertexai=provider == 'google-vertex')
+        return GoogleProvider(vertexai=provider in ('google-vertex', 'vertexai'))
     else:
         provider_class = infer_provider_class(provider)
         return provider_class()


pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_model.py
```diff
@@ -18,13 +18,52 @@
...
+    return None, model_string
+
+
+def _infer_profile(model_string: str) -> ModelProfile:
+    """Infer model profile from a provider string without constructing a provider.
+
+    This is safe to call inside Temporal workflows because it only does pure
+    computation (string parsing and profile function calls). No I/O occurs.
+    """
+    from pydantic_ai.providers import infer_profile_func
```
dsfaccini: this line can probably go at the top

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_model.py
```diff
@@ -299,6 +338,63 @@ def using_model(self, model: models.Model | models.KnownModelName | str | None)
...
+        Only used for `settings` and `base_url`. For unregistered strings,
+        falls back to the default model since these require provider construction.
+        """
+        model_id = self._current_model_id()
+        if model_id is None:
+            return self.wrapped
+        if model_id in self._models_by_id:
+            return self._models_by_id[model_id]
+        if workflow.in_workflow():
+            return self.wrapped
```
dsfaccini: lets pack `model_id` is None and `workflow.in_workflow` in the same if and turn the rest to elifs -> else

tests/test_temporal.py
```diff
@@ -2924,6 +3030,48 @@ async def test_temporal_agent_model_selection_by_instance(
         assert output == expected_output
 
 
+def test_temporal_model_profile_for_raw_strings():
+    """Test TemporalModel infers model_name, system, and profile from raw strings without constructing providers."""
+    from pydantic_ai.profiles.openai import openai_model_profile
```
dsfaccini: do we need to import this here?

dsfaccini: I'd imagine it needs to go behind a `with try_import...` guard but CI passes so I guess not

tests/test_temporal.py
```diff
@@ -2924,6 +3030,48 @@ async def test_temporal_agent_model_selection_by_instance(
...
+        activity_name_prefix='test__profile_inference',
+        activity_config={'start_to_close_timeout': timedelta(seconds=60)},
+        deps_type=type(None),
+    )
+
+    # Without using_model, properties come from default
+    assert temporal_model.profile == default_model.profile
+    assert temporal_model.model_name == default_model.model_name
+    assert temporal_model.system == default_model.system
+
```
dsfaccini, line 3276: would `assert temporal_model == default_model` work here? perhaps with minimal removal of potentially ephemeral fields? if it works here we could have something similar in the tests below, using snapshots.

the `WebSearchTool` check probably needs to stay as an assert because `supported_builtin_tools` is a property and not captured in the snapshot

dsfaccini:
hey @mattbrandman thanks for working on this. 

```bash
git pull

deactivate

rm -rf .venv

make install
```

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_model.py
```diff
@@ -18,13 +18,52 @@
...
+from pydantic_ai.profiles import DEFAULT_PROFILE, ModelProfile
 from pydantic_ai.providers import Provider
 from pydantic_ai.settings import ModelSettings
 from pydantic_ai.tools import AgentDepsT, RunContext
 from pydantic_ai.usage import RequestUsage
 
 from ._run_context import TemporalRunContext
 
+# Legacy model name prefixes (no provider: prefix)
+_LEGACY_PREFIXES = {'gpt': 'openai', 'o1': 'openai', 'o3': 'openai', 'claude': 'anthropic', 'gemini': 'google-gla'}
```
DouweM: This stuff should only live in/close to `infer_model`, so if we need additional helper methods, they should be defined where that method, and be used by it if possible.

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_model.py
```diff
@@ -299,6 +338,63 @@ def using_model(self, model: models.Model | models.KnownModelName | str | None)
...
+        return models.infer_model(model_id)
+
+    @property
+    def model_name(self) -> str:
+        """Get the model name, inferring from raw strings without provider construction."""
+        model_id = self._current_model_id()
+        if model_id is None:
+            return self.wrapped.model_name
+        if model_id in self._models_by_id:
+            return self._models_by_id[model_id].model_name
```
DouweM: Except for the `model_name` call vs `system` vs `profile`, the logic here is identical between the 3 methods. Could we use `_current_model_for_context` or smth?

pydantic_ai_slim/pydantic_ai/providers/__init__.py
```diff
@@ -187,3 +191,52 @@ def infer_provider(provider: str) -> Provider[Any]:
...
+        'groq': groq_model_profile,
+        'mistral': mistral_model_profile,
+        'cohere': cohere_model_profile,
+        'xai': grok_model_profile,
+        'grok': grok_model_profile,
+        'deepseek': deepseek_model_profile,
+        'moonshotai': moonshotai_model_profile,
+        'alibaba': qwen_model_profile,
+        # Bedrock uses amazon profile as base
+        'bedrock': amazon_model_profile,
```
DouweM: The logic here is all wrong -- model profiles are _model name_-specific, not provider-specific. I.e. So bedrock used with Anthropic should give `anthropic_model_profile`.

That's all defined in `Provider.model_profile`, so we need to use those actual methods. I can't see myself accepting a PR that adds a mapping of providers/models to profiles separate from that one.

mattbrandman:
@DouweM @dsfaccini appreciate the feedback, this is a pretty first pass PR so will let you know once its in a better place.

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_model.py
mattbrandman: I took that function out entirely as base_url and settings were not actually necessary to override from testing. these others access nested attributes early enough that the only abstractions I can think of are extremely small or based on getattr and callables which I try to avoid for something this straightforward

mattbrandman: Something like 
```
T = TypeVar('T')

  def _get_current_model_attr(self, attr: str, fallback: Callable[[str], T]) -> T:
      """Get an attribute from the current model, with fallback for raw string model IDs."""
      model_id = self._current_model_id()
      if model_id is None:
          return getattr(self.wrapped, attr)
      if model_id in self._models_by_id:
          return getattr(self._models_by_id[model_id], attr)
      return fallback(model_id)
```

tests/test_temporal.py
mattbrandman: I don't believe this would work as its a temporalmodel vs testmodel and they don't have eq methods so it would always be false?

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_model.py
mattbrandman: Found a cleaner way not entirely abstracted but a good deal more

mattbrandman:
@DouweM @dsfaccini got this working correctly now and cleaned up a good deal. Ready for review, no rush though!

dsfaccini:
hey @mattbrandman I may be out of the loop here but what's the reason why you turned all the `model_profile` methods into `@staticmethods`?

mattbrandman:
Instantiating a model to get its profile is an unsafe operation in temporal since it is an IO operation (normally done by the provider when the model is instantiated) this way we can get the model profile without needing a self on the model if that makes sense. @dsfaccini

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -1091,6 +1091,38 @@ def override_allow_model_requests(allow_model_requests: bool) -> Iterator[None]:
         ALLOW_MODEL_REQUESTS = old_value  # pyright: ignore[reportConstantRedefinition]
 
 
+LEGACY_MODEL_PREFIXES: dict[str, str] = {
```
DouweM: Private constant please

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -1091,6 +1091,38 @@ def override_allow_model_requests(allow_model_requests: bool) -> Iterator[None]:
...
+    'gemini': 'google-gla',
+}
+"""Mapping of legacy model name prefixes to provider names.
+
+These prefixes allow using model names without the `provider:` prefix for backwards compatibility.
+For example, `gpt-4` is equivalent to `openai:gpt-4`.
+"""
+
+
+def parse_model_string(model: str) -> tuple[str | None, str]:
```
DouweM: Let's call this `parse_model_id` in line with https://github.com/pydantic/pydantic-ai/pull/3917

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -1091,6 +1091,38 @@ def override_allow_model_requests(allow_model_requests: bool) -> Iterator[None]:
...
+    if ':' in model:
+        provider_name, model_name = model.split(':', maxsplit=1)
+        return provider_name, model_name
+
+    # Legacy model names without provider prefix
+    for prefix, provider_name in LEGACY_MODEL_PREFIXES.items():
+        if model.startswith(prefix):
+            return provider_name, model
+
+    return None, model
```
DouweM, line 1159: Why do we need this case?

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_model.py
```diff
@@ -16,16 +16,35 @@
...
+from pydantic_ai.profiles import DEFAULT_PROFILE, ModelProfile
+from pydantic_ai.providers import Provider, infer_provider_class
 from pydantic_ai.settings import ModelSettings
 from pydantic_ai.tools import AgentDepsT, RunContext
 from pydantic_ai.usage import RequestUsage
 
 from ._run_context import TemporalRunContext
 
 
+def _infer_profile(model_string: str) -> ModelProfile:
```
DouweM: This can be public on the models module as well

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_model.py
```diff
@@ -299,6 +318,69 @@ def using_model(self, model: models.Model | models.KnownModelName | str | None)
...
+        if model_id in self._models_by_id:
+            return self._models_by_id[model_id]
+        return model_id
+
+    @property
+    def model_name(self) -> str:
+        """Get the model name, inferring from raw strings without provider construction."""
+        current = self._current_model()
+        if isinstance(current, str):
+            return parse_model_string(current)[1]
```
DouweM: For tuples prefer `_, model_name = ...` over index-based lookup

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_model.py
DouweM: Would be named `infer_model_profile`

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_model.py
```diff
@@ -299,6 +318,69 @@ def using_model(self, model: models.Model | models.KnownModelName | str | None)
...
+        via `using_model()`, we use that model's profile for validation and
+        parameter preparation, not the default wrapped model's profile.
+        """
+        model_id = self._current_model_id()
+
+        # If using a registered model, delegate to its prepare_request directly
+        if model_id is not None and model_id in self._models_by_id:
+            return self._models_by_id[model_id].prepare_request(model_settings, model_request_parameters)
+
+        # If using default model (model_id is None), delegate to wrapped model
```
DouweM: Could we use `self._current_model()` in this method?

tests/test_temporal.py
```diff
@@ -2713,6 +2717,111 @@ async def run(self, prompt: str, model_id: str | None = None) -> str:
...
+    def supported_builtin_tools(cls) -> frozenset[type[AbstractBuiltinTool]]:
+        return cls.SUPPORTED_TOOLS
+
+    @property
+    def model_name(self) -> str:
+        return self._model_name
+
+    @property
+    def system(self) -> str:
+        return 'test'
```
DouweM: Could we subclass `TestModel`?

tests/test_temporal.py
```diff
@@ -2924,6 +3093,47 @@ async def test_temporal_agent_model_selection_by_instance(
...
+    assert temporal_model.profile == default_model.profile
+    assert temporal_model.model_name == default_model.model_name
+    assert temporal_model.system == default_model.system
+
+    # With raw string, all properties are inferred correctly
+    with temporal_model.using_model('openai:gpt-5'):
+        assert temporal_model.model_name == 'gpt-5'
+        assert temporal_model.system == 'openai'
+        expected_profile = openai_model_profile('gpt-5')
+        assert temporal_model.profile.supports_json_schema_output == expected_profile.supports_json_schema_output
```
DouweM: Can we compare the entire profile?

tests/test_temporal.py
```diff
@@ -2924,6 +3093,47 @@ async def test_temporal_agent_model_selection_by_instance(
...
+    # Registered models work correctly for all properties
+    temporal_model_with_registry = TemporalModel(
+        default_model,
+        activity_name_prefix='test__profile_registry',
+        activity_config={'start_to_close_timeout': timedelta(seconds=60)},
+        deps_type=type(None),
+        models={'alt': TestModel(custom_output_text='alt', model_name='alt-model')},
+    )
+    with temporal_model_with_registry.using_model('alt'):
+        assert temporal_model_with_registry.model_name == 'alt-model'
```
DouweM, line 3299: Also check the system and profile for registered models pelase

pydantic_ai_slim/pydantic_ai/models/__init__.py
mattbrandman: Given its a pure parser we want to let the caller decide on how to handle the unknown provider case. For infer model thats an error since we can't safely instantiate a model but a profile can be safely inferred since it can fall back to the default model profile. We could also return error for that and then move error into this method if you think thats "less surprising / less magic" behavior

mattbrandman:
@DouweM  tests failure seems transient as its socket related but I can't directly retry it

pydantic_ai_slim/pydantic_ai/providers/__init__.py
```diff
@@ -43,7 +43,8 @@ def client(self) -> InterfaceClient:
         """The client for the provider."""
         raise NotImplementedError()
 
-    def model_profile(self, model_name: str) -> ModelProfile | None:
+    @staticmethod
+    def model_profile(model_name: str) -> ModelProfile | None:
         """The model profile for the named model, if available."""
         return None  # pragma: no cover
```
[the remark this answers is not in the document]
DouweM: Classmethod sounds reasonable!

pydantic_ai_slim/pydantic_ai/providers/nebius.py
```diff
@@ -55,11 +56,11 @@ def model_profile(self, model_name: str) -> ModelProfile | None:
...
         profile = None
 
-        try:
-            model_name = model_name.lower()
-            provider, model_name = model_name.split('/', 1)
-        except ValueError:
-            raise UserError(f"Model name must be in 'provider/model' format, got: {model_name!r}")
+        model_name = model_name.lower()
+        if '/' not in model_name:
+            return OpenAIModelProfile(json_schema_transformer=OpenAIJsonSchemaTransformer)
```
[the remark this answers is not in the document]
mattbrandman: this was done on purpose to standardize these two were out of the pattern anyway

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_model.py
```diff
@@ -299,6 +300,72 @@ def using_model(self, model: models.Model | models.KnownModelName | str | None)
...
+    def system(self) -> str:
+        """Get the system (provider) name, inferring from raw strings without provider construction."""
+        current = self._current_model()
+        if isinstance(current, str):
+            provider_name, _ = parse_model_id(current)
+            return provider_name or self.wrapped.system
+        return current.system
+
+    @property
+    def profile(self) -> ModelProfile:  # type: ignore[override]
```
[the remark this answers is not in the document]
DouweM: This could cause bugs only if `self.profile` is used inside e.g. `prepare_request`, but if that's always the version from the wrapped model, we're good.

pydantic_ai_slim/pydantic_ai/providers/nebius.py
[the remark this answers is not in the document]
DouweM: This is OK, they'll get an API error if it doesn't exist.

pydantic_ai_slim/pydantic_ai/providers/__init__.py
```diff
@@ -43,7 +43,8 @@ def client(self) -> InterfaceClient:
         """The client for the provider."""
         raise NotImplementedError()
 
-    def model_profile(self, model_name: str) -> ModelProfile | None:
+    @classmethod
+    def model_profile(cls, model_name: str) -> ModelProfile | None:
```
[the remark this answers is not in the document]
mattbrandman: Going with option 1 — this is a necessary breaking change for the bug fix. Will document in the PR description.

_skill: address-feedback_

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -1100,6 +1100,59 @@ def override_allow_model_requests(allow_model_requests: bool) -> Iterator[None]:
...
+    if provider == 'vertexai':  # pragma: no cover
+        provider = 'google-vertex'
+
+    if provider.startswith('gateway/'):
+        from ..providers.gateway import normalize_gateway_provider
+
+        provider = normalize_gateway_provider(provider)
+
+    try:
+        provider_class = infer_provider_class(provider)
```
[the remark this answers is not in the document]
mattbrandman: Good catch. Added `google` -> `google-gla` normalization in `infer_model_profile`, mirroring how `infer_model` handles the `google` provider name at line ~1219. Also added test coverage for `google:gemini-2.0-flash`.

_skill: address-feedback_

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -1100,6 +1100,59 @@ def override_allow_model_requests(allow_model_requests: bool) -> Iterator[None]:
...
 
 
+_LEGACY_MODEL_PREFIXES: dict[str, str] = {
+    'gpt': 'openai',
+    'o1': 'openai',
+    'o3': 'openai',
+    'claude': 'anthropic',
+    'gemini': 'google-gla',
+}
+"""Backward compat: allows prefix-only model names like `gpt-4` without `provider:`."""
```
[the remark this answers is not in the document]
mattbrandman: This is already addressed — `infer_model` uses `parse_model_id()` (line 1176) which references `_LEGACY_MODEL_PREFIXES`. There is no duplication; the dict is the single source of truth.

_skill: address-feedback_

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -1100,6 +1100,59 @@ def override_allow_model_requests(allow_model_requests: bool) -> Iterator[None]:
...
+    # Legacy model names without provider prefix
+    for prefix, provider_name in _LEGACY_MODEL_PREFIXES.items():
+        if model.startswith(prefix):
+            return provider_name, model
+
+    # Unknown prefix: let callers decide how to handle this case.
+    return None, model
+
+
+def infer_model_profile(model: str) -> ModelProfile:
```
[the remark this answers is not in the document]
mattbrandman: Expanded both docstrings with `Args:`/`Returns:` sections. `infer_model_profile` now explicitly documents that it returns the raw provider profile without intersecting `Model.supported_builtin_tools()`, and explains why this is acceptable for best-effort scenarios.

_skill: address-feedback_

pydantic_ai_slim/pydantic_ai/providers/__init__.py
```diff
@@ -169,7 +170,7 @@ def infer_provider_class(provider: str) -> type[Provider[Any]]:  # noqa: C901
         from .voyageai import VoyageAIProvider
 
         return VoyageAIProvider
-    else:  # pragma: no cover
+    else:
         raise ValueError(f'Unknown provider: {provider}')
```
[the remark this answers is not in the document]
mattbrandman: Confirmed — `test_infer_model_profile[unknown-provider]` with `unknown-provider:some-model` hits the `ValueError` catch path and covers this branch.

_skill: address-feedback_

tests/models/test_model.py
```diff
@@ -267,3 +267,50 @@ def test_infer_model_with_provider():
...
+        pytest.param('anthropic:claude-sonnet-4-5', False, id='anthropic'),
+        pytest.param('gateway/openai:gpt-5', False, id='gateway-openai'),
+        pytest.param('unknown-provider:some-model', True, id='unknown-provider'),
+        pytest.param('unknown-model', True, id='unknown-no-prefix'),
+        pytest.param('nebius:model-without-slash', False, id='provider-unknown-model'),
+    ],
+)
+def test_infer_model_profile(model_id: str, is_default: bool):
+    profile = infer_model_profile(model_id)
+    if is_default:
```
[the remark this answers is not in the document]
mattbrandman: Parametrized `test_infer_model_profile_matches_provider` across OpenAI, Anthropic, Google GLA, and Google shorthand (`google:` prefix). Each now asserts the inferred profile matches the actual provider class's `model_profile()` return value.

_skill: address-feedback_

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_model.py
```diff
@@ -299,6 +300,76 @@ def using_model(self, model: models.Model | models.KnownModelName | str | None)
...
```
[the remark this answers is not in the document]
mattbrandman: Agreed this is the right tradeoff. The dynamic nature of `_current_model_id()` requires a regular property. Performance should be fine for now — `infer_model_profile` is lightweight (dict lookup + class method call). Will revisit if profiling shows it matters.

_skill: address-feedback_

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -1100,6 +1100,83 @@ def override_allow_model_requests(allow_model_requests: bool) -> Iterator[None]:
...
+    'gpt': 'openai',
+    'o1': 'openai',
+    'o3': 'openai',
+    'claude': 'anthropic',
+    'gemini': 'google-gla',
+}
+"""Backward compat: allows prefix-only model names like `gpt-4` without `provider:`."""
+
+
+def parse_model_id(model: str) -> tuple[str | None, str]:
```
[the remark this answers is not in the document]
DouweM: I'm OK with them being public

pydantic_ai_slim/pydantic_ai/models/wrapper.py
```diff
@@ -58,7 +58,7 @@ async def request_stream(
             yield response_stream
 
     def customize_request_parameters(self, model_request_parameters: ModelRequestParameters) -> ModelRequestParameters:
-        return self.wrapped.customize_request_parameters(model_request_parameters)  # pragma: no cover
+        return self.wrapped.customize_request_parameters(model_request_parameters)
```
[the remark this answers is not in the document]
DouweM: `InstrumentedModel` delegating makes sense to me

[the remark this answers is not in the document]
mattbrandman: addressed

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_model.py
```diff
@@ -299,6 +300,76 @@ def using_model(self, model: models.Model | models.KnownModelName | str | None)
...
+        """
+        current = self._current_model()
+
+        # For unregistered model strings, use Model.prepare_request (grandparent's method)
+        # with our overridden profile property. This allows validation to use the correct
+        # profile inferred from the model string, without constructing a full model instance.
+        if isinstance(current, str):
+            return Model.prepare_request(self, model_settings, model_request_parameters)
+
+        return current.prepare_request(model_settings, model_request_parameters)
```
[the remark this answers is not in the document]
mattbrandman: @DouweM should this be added?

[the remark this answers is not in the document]
DouweM: OK to leave out for now

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -1129,25 +1209,17 @@ def infer_model(  # noqa: C901
...
-                DeprecationWarning,
-            )
-        else:
-            raise UserError(f'Unknown model: {model}')
+    provider_name, model_name = parse_model_id(model)
+    if provider_name is None:
+        raise UserError(f'Unknown model: {model}')
+
+    # Warn if using legacy model name without provider prefix
+    if ':' not in model:
```
DouweM: Let's move this into `parse_model_id`

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -1111,6 +1111,86 @@ def override_allow_model_requests(allow_model_requests: bool) -> Iterator[None]:
...
+        The inferred `ModelProfile`, or `DEFAULT_PROFILE` if the provider is unknown.
+    """
+    provider, model_name = parse_model_id(model)
+    if provider is None:
+        return DEFAULT_PROFILE
+
+    if provider == 'vertexai':  # pragma: no cover
+        provider = 'google-vertex'
+    elif provider == 'google':
+        provider = 'google-gla'
```
DouweM: We can drop this if we handle `vertexai` and `google` in `infer_provider_class`, right?

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -1111,6 +1111,86 @@ def override_allow_model_requests(allow_model_requests: bool) -> Iterator[None]:
...
+
+    if provider.startswith('gateway/'):
+        from ..providers.gateway import normalize_gateway_provider
+
+        provider = normalize_gateway_provider(provider)
```
DouweM: This also seems like it could live inside `infer_provider_class`

pydantic_ai_slim/pydantic_ai/providers/__init__.py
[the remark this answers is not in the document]
DouweM: Heh, I had approved `classmethod` specifically because the bot suggested that it would be backward compatible: https://github.com/pydantic/pydantic-ai/pull/4100#discussion_r2776512125

But now it appears that it's not. 

@mattbrandman Could we have a test that verifies that the `provider.model_profile` pattern that's in all the model `__init__` methods still works with an outdated custom provider that defines this as an instance method rather than a class method? I wouldn't expect them to get a runtime error unless they're using it with `TemporalModel`, which is OK -- but if we can, we should make sure that non-Temporal usage with a custom provider still works.

I also like `staticmethod` better than `classmethod` if we're going to be calling it on both instances and classes.

pydantic_ai_slim/pydantic_ai/models/__init__.py
mattbrandman: Done — `parse_model_id` now emits the `DeprecationWarning` for legacy prefix-based model names, and the redundant warning was removed from `infer_model`.

_skill: address-feedback_

pydantic_ai_slim/pydantic_ai/models/__init__.py
mattbrandman: Yes — added `vertexai` -> `google-vertex` and `google` -> `google-gla` normalization to `infer_provider_class`, and dropped the special cases from `infer_model_profile`.

_skill: address-feedback_

pydantic_ai_slim/pydantic_ai/models/__init__.py
mattbrandman: Done — `infer_provider_class` now normalizes `gateway/*` prefixes via `normalize_gateway_provider`, so `infer_model_profile` no longer needs its own gateway handling.

_skill: address-feedback_

pydantic_ai_slim/pydantic_ai/providers/__init__.py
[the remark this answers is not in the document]
mattbrandman: Changed to `@staticmethod` across all 30 providers. Added `test_custom_provider_instance_method_model_profile` in `tests/models/test_model.py` — it creates a `LegacyCustomProvider` with the old `def model_profile(self, model_name)` signature and verifies that instance calls still work. (Class-level calls from `infer_model_profile` would break for such providers, but as you noted that only affects TemporalModel usage which is fine.)

_skill: address-feedback_

PR #4114, opened 2026-01-27
feat: Introduce `ToolFailed` exception

### LINKED ISSUE #2586: More flexible tool exception handling

rgbkrk:
Introduces `ToolFailed` exception to allow tools to fail without terminating the agent run. This is especially useful for parallel tool execution where partial failures should not stop the entire batch.

Key features:
- Errors are traced in telemetry (unlike `ModelRetry`)
- Agent continues execution (unlike arbitrary exceptions)

Three exception modes:

- `ModelRetry`: Expected retry behavior, not an error
- `ToolFailed(disable=False)`: System error that should be logged/monitored
- `ToolFailed(disable=True)`: Permanent failure, disable tool

This can coexist with `ToolPolicy` from https://github.com/pydantic/pydantic-ai/pull/3691. The main difference here is that `ToolFailed` is explicit inline in the user's tools (in Python) whereas `ToolPolicy` is declarative. Whenever the tool policy PR is merged, I'll want to adapt that to the policy's error modes.

- Closes #2586

## Problem

Currently, only `ModelRetry` and `ValidationError` are caught during tool execution. Any other exception terminates the agent run. When `ModelRetry` or `ValidationError` hits `max_retries`, it raises `UnexpectedModelBehavior` which kills the run.

This is particularly painful for parallel tool execution where some tools might fail but the agent should continue with successful results. There's no way to:

1. Mark a tool error in telemetry while letting the agent continue
2. Disable a broken tool without killing the entire agent run

With #2952, tools can check `ctx.last_attempt` to return custom messages on the final try. However, returning a string marks the call as successful and resets the retry count to 0. The model can immediately call the same tool again. That doesn't help if you want to get the model to know more detail as well as to disable the tool.

## Solution

This PR introduces the `ToolFailed` exception with explicit control over tool availability:

```python
# Tool error, keeps tool available (model can retry)
raise ToolFailed("Database busy, try again later", disable=False)

# Tool error, disables tool for this run in hopes the agent can continue with other tools
raise ToolFailed("Invalid credentials", disable=True)
```

## Coordinating with ToolPolicy PR (https://github.com/pydantic/pydantic-ai/pull/3691)

`ToolFailed` is complementary to `ToolPolicy`. Key differences:

- **`ToolFailed`**: Explicit, inline Python exceptions in tool code (runtime)
- **`ToolPolicy`**: Declarative, agent-level configuration (design-time)

Both can coexist. When `ToolPolicy` merges, we can adapt its error handling modes to work with `ToolFailed`.

### Pre-Review Checklist

### Pre-Merge Checklist

### DIFF
--- a/docs/tools-advanced.md
+++ b/docs/tools-advanced.md
@@ -377,9 +377,8 @@ While [`ModelRetry`][pydantic_ai.exceptions.ModelRetry] is intended for expected
 
 This is particularly useful in parallel tool execution where some tools may fail but you want the agent to proceed with the successful results.
 
-```python
+```python {lint="skip"}
 from pydantic_ai import Agent, RunContext, ToolFailed
-
 
 agent = Agent('test')
 
@@ -392,10 +391,10 @@ async def query_database(ctx: RunContext[None], query: str) -> str:
         return result
     except AuthenticationError:
         # Permanent failure - disable this tool for the rest of the run
-        raise ToolFailed("Database credentials invalid", disable=True)
+        raise ToolFailed('Database credentials invalid', disable=True)
     except TemporaryConnectionError:
         # Transient failure - keep tool available for retry
-        raise ToolFailed("Database temporarily unavailable", disable=False)
+        raise ToolFailed('Database temporarily unavailable', disable=False)
 ```
 
 Key differences between `ModelRetry` and `ToolFailed`:


tests/test_agent.py
```diff
@@ -7648,3 +7648,98 @@ async def filtered_response(messages: list[ModelMessage], info: AgentInfo) -> Mo
...
+            raise ToolFailed('First attempt failed', disable=False)
+        return f'Success: {value}'
+
+    result = agent.run_sync('Get data')
+    assert result.output == 'Got the data'
+    assert call_count == 3
+
+
+def test_tool_failed_disables_tool():
+    """Test that ToolFailed with disable=True disables the tool."""
```
DouweM, line 7686: I don't think this actually verifies that the tool is no longer available to the model on the second step. Can you double check that please?

docs/tools-advanced.md
```diff
@@ -371,6 +371,44 @@ def my_flaky_tool(query: str) -> str:
...
+
+- **`ModelRetry`**: Expected retry behavior, not traced as an error in telemetry
+- **`ToolFailed`**: Traced as an error in telemetry, indicates a genuine failure
+
+The `disable` parameter controls tool availability:
+
+- **`disable=False`** (default): Tool remains available. The model can retry with different arguments or try again later.
+- **`disable=True`**: Tool is disabled for the remainder of the agent run. Useful for permanent failures like missing credentials or unsupported operations.
+
+Like `ModelRetry`, `ToolFailed` generates a `RetryPromptPart` that informs the model of the failure, allowing it to adjust its approach or continue with other tools.
```
DouweM, line 410: It seems surprising that this creates `RetryPromptPart` since the point is _not_ to retry. :)

tests/test_agent.py
```diff
@@ -7648,3 +7648,98 @@ async def filtered_response(messages: list[ModelMessage], info: AgentInfo) -> Mo
...
+        else:
+            # Final response
+            return ModelResponse(parts=[TextPart('Got the data')])
+
+    agent = Agent(FunctionModel(call_tool))
+
+    @agent.tool_plain
+    def get_data(value: str) -> str:
+        if value == 'first':
+            raise ToolFailed('First attempt failed', disable=False)
```
DouweM, line 7677: I think we need to also validate that this message is sent to the model verbatim without the "Fix the errors and try again." text that's automatically appended when `ModelRetry` is used.

DouweM (changes requested):

PR #3438, opened 2025-11-15
Add prompt caching support for AWS Bedrock

DenysMoskalenko:
## Summary
 Implements AWS Bedrock prompt caching support (closes #3418) by fixing how cache points are sent, documenting the workflow, and extending test coverage to assert cache writes and reads. 

## Testing
  - uv run pytest tests/models/test_bedrock.py
  - uv run coverage run -m pytest tests/models/test_bedrock.py

PR #2560, opened 2025-08-14
Add Bedrock/Anthropic prompt caching

larryhudson:

PR #4122, opened 2026-01-27
bug(bedrock): stop injecting extra message when caching

### LINKED ISSUE #4121: bedrock_cache_messages results in an extraneous extra user message which prevents cache hits
### Initial Checks

### Description

pydantic's bedrock model has a workaround that adds an empty new line to the messages that are sent to the model if `bedrock_cache_messages: True` and the last message does not have a `text`. See:
https://github.com/pydantic/pydantic-ai/blob/main/pydantic_ai_slim/pydantic_ai/models/bedrock.py#L749

Since this new message is not persistent in pydantic's message list, each time a call to the LLM happens this new line message is moved around. Critically, it is removed from the location is was before. This results in always having cache misses and rendering `bedrock_cache_messages: True` useless. 

Here's a sample run. This is the first set of input messages (some info redacted)

```
{
  "messages": [
    {
      "content": "[redacted]",
      "role": "system"
    },
    { "content": "[Unsupported content type: cachePoint]", "role": "system" },
    { "content": "<some query>", "role": "user" },
    {
      "tool_calls": [
        {
          "name": "list_data_models",
          "tool_id": "tooluse_vTH0UddkSM-vMXRj3753Zg",
          "type": "toolUse"
        }
      ]
    },
    { "content": "\n", "role": "user" },
    { "content": "[Unsupported content type: cachePoint]", "role": "user" },
    {
      "role": "user",
      "tool_results": [
        {
          "result": "<some result>",
          "tool_id": "tooluse_vTH0UddkSM-vMXRj3753Zg",
          "type": "toolResult"
        }
      ]
    }
  ]
}
```
and the next llm call looks like this:
```
{
  "messages": [
    {
      "content": "[redacted]",
      "role": "system"
    },
    { "content": "[Unsupported content type: cachePoint]", "role": "system" },
    { "content": "<some query>", "role": "user" },
    {
      "tool_calls": [
        {
          "name": "list_data_models",
          "tool_id": "tooluse_vTH0UddkSM-vMXRj3753Zg",
          "type": "toolUse"
        }
      ]
    },
    {
      "role": "user",
      "tool_results": [
        {
          "result": "<some result>",
          "tool_id": "tooluse_vTH0UddkSM-vMXRj3753Zg",
          "type": "toolResult"
        }
      ]
    },
    {
      "tool_calls": [
        {
          "name": "get_data_model_schema",
          "arguments": { "model_slugs": ["case"] },
          "tool_id": "tooluse_5UjEI5aZTLCXb4irdpesyw",
          "type": "toolUse"
        }
      ]
    },
    { "content": "\n", "role": "user" },
    { "content": "[Unsupported content type: cachePoint]", "role": "user" },
    {
      "role": "user",
      "tool_results": [
        {
          "result": "<some result>",
          "tool_id": "tooluse_5UjEI5aZTLCXb4irdpesyw",
          "type": "toolResult"
        }
      ]
    }
  ]
}
```

notice how `{ "content": "\n", "role": "user" }` which was right after the `list_data_models` tool call in the first call is no longer there in the second call. This causes a cache miss.

I forked the repo and removed the linked code and see much better caching. This is the caching info on the last llm call of the agent with the current code. (38% caching)

Image

And this is with those lines removed: (94.5% caching)

Image

```Python

```

### Logfire Trace

_No response_

- Python: Python 3.11.14
- Pydantic: 2.11.9 
- Pydantic AI slim: 1.38.0
- LLM provider SDK: anthropic.claude-opus-4-5-20251101-v1:0

matthgt:
Current logic adds an extra user message when user message caching is enabled and the last message is not a user message. This injected message is not persisted to pydantic list of messagges, so it moved aroudn on each iteration, resulting in no cache hits.

- Closes #4121

### Pre-Review Checklist

### Pre-Merge Checklist

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/bedrock.py
+++ b/pydantic_ai_slim/pydantic_ai/models/bedrock.py
@@ -746,10 +746,14 @@ class BedrockConverseModel(Model):
         if processed_messages and settings.get('bedrock_cache_messages') and profile.bedrock_supports_prompt_caching:
             last_user_content = self._get_last_user_message_content(processed_messages)
             if last_user_content is not None:
-                # AWS currently rejects cache points that directly follow document content.
-                # only add cache point if last block is not a document
-                if 'document' not in last_user_content[-1]:
-                    last_user_content.append({'cachePoint': {'type': 'default'}})
+                # AWS currently rejects cache points that directly follow binary content (document, image, video).
+                # Insert a newline text block as a workaround only for these specific content types.
+                # We avoid adding this workaround for other non-text content like toolResult, as the
+                # non-persisted newline causes cache misses on subsequent calls (see #4121).
+                last_block = last_user_content[-1]
+                if any(key in last_block for key in ('document', 'image', 'video')):
+                    last_user_content.append({'text': '\n'})
+                last_user_content.append({'cachePoint': {'type': 'default'}})
 
         return system_prompt, processed_messages
 
@@ -859,10 +863,13 @@ class BedrockConverseModel(Model):
                             'CachePoint cannot be the first content in a user message - there must be previous content to cache when using Bedrock. '
                             'To cache system instructions or tool definitions, use the `bedrock_cache_instructions` or `bedrock_cache_tool_definitions` settings instead.'
                         )
-                    if 'text' not in content[-1]:
-                        # AWS currently rejects cache points that directly follow non-text content.
-                        # Insert an empty text block as a workaround (see https://github.com/pydantic/pydantic-ai/issues/3418
-                        # and https://github.com/pydantic/pydantic-ai/pull/2560#discussion_r2349209916).
+                    # AWS currently rejects cache points that directly follow binary content (document, image, video).
+                    # Insert a newline text block as a workaround only for these specific content types.
+                    # We avoid adding this workaround for other non-text content like toolResult, as the
+                    # non-persisted newline causes cache misses on subsequent calls (see #4121).
+                    # See also https://github.com/pydantic/pydantic-ai/issues/3418 and
+                    # https://github.com/pydantic/pydantic-ai/pull/2560#discussion_r2349209916.
+                    if any(key in content[-1] for key in ('document', 'image', 'video')):
                         content.append({'text': '\n'})
                     content.append({'cachePoint': {'type': 'default'}})
                 else:
--- a/tests/models/test_bedrock.py
+++ b/tests/models/test_bedrock.py
@@ -1967,7 +1967,6 @@ async def test_bedrock_cache_messages_with_document_as_last_content(
     usage = result.usage()
     assert usage.input_tokens > 0
     assert usage.output_tokens > 0
-    # Should have cache write tokens on first run
     assert usage.cache_write_tokens == 0 and usage.cache_read_tokens == 0
 
 
@@ -2153,7 +2152,7 @@ async def test_bedrock_cache_messages(allow_model_requests: None, bedrock_provid
 async def test_bedrock_cache_messages_with_binary_content(
     allow_model_requests: None, bedrock_provider: BedrockProvider
 ):
-    """Test that bedrock_cache_messages does not add cache point for document content."""
+    """Test that bedrock_cache_messages does add cache point for document content with newline workaround."""
     model = BedrockConverseModel('us.anthropic.claude-haiku-4-5-20251001-v1:0', provider=bedrock_provider)
     messages: list[ModelMessage] = [
         ModelRequest(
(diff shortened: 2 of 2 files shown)


pydantic_ai_slim/pydantic_ai/models/bedrock.py
```diff
@@ -746,10 +746,6 @@ async def _map_messages(  # noqa: C901
         if processed_messages and settings.get('bedrock_cache_messages') and profile.bedrock_supports_prompt_caching:
             last_user_content = self._get_last_user_message_content(processed_messages)
             if last_user_content is not None:
-                # AWS currently rejects cache points that directly follow non-text content.
```
DouweM, line 749: cc @DenysMoskalenko because you implemented this originally in https://github.com/pydantic/pydantic-ai/pull/3438.

I suppose AWS has fixed this, that's great news!

tests/models/test_bedrock.py
```diff
@@ -2121,43 +2121,6 @@ async def test_bedrock_cache_messages(allow_model_requests: None, bedrock_provid
...
-            {
-                'document': {
-                    'name': 'Document 1',
-                    'format': 'txt',
-                    'source': {'bytes': b'Test document content'},
-                }
-            },
-            {'text': '\n'},
-            {'cachePoint': {'type': 'default'}},
-        ]
```
DouweM, line 2157: @matthgt Have you verified that Bedrock now accepts a document followed by a cachepoint without the `\n` in between? 

I seem to recall that Bedrock was only requiring this for documents specifically, not for other file types followed by cache points, so it's possible that by removing the logic entirely we've introduced a bug, and we should instead only add the `\n` for cache points following documents.

I'd like to see a new tests that actually does a request to Bedrock (recorded using vcr) and verifies that the document + cache point scenario now works without issue.

matthgt: added the test, but I'm having trouble getting the recording of the request to work. I think I can't seem to get the test to run with  my aws credentials. Any guidance on that?

matthgt: figured it out the moment I added this comment. Looks like the error is still there when the last message is a document. I'm inclined to change the workaround to _not_ add the cache point in that case. WDYT?

pydantic_ai_slim/pydantic_ai/models/bedrock.py
```diff
@@ -746,11 +746,10 @@ async def _map_messages(  # noqa: C901
...
-                # Insert a newline text block as a workaround.
-                if 'text' not in last_user_content[-1]:
-                    last_user_content.append({'text': '\n'})
-                last_user_content.append({'cachePoint': {'type': 'default'}})
+                # AWS currently rejects cache points that directly follow document content.
+                # only add cache point if last block is not a document
+                if 'document' not in last_user_content[-1]:
```
DouweM: https://github.com/pydantic/pydantic-ai/pull/4128 suggests that this bug also exists for multi-modal content other than documents, although it's unclear if that was verified or assumed. Can you verify please?

I also think AI did a good job on the code comments in that PR, including finding the original thread where this bug was discussed: https://github.com/pydantic/pydantic-ai/pull/2560#discussion_r2319364458.

Also note that that PR also makes this change in `map_user_prompt`, I assume we should do that here as well.

pydantic_ai_slim/pydantic_ai/models/bedrock.py
```diff
@@ -746,11 +746,10 @@ async def _map_messages(  # noqa: C901
         if processed_messages and settings.get('bedrock_cache_messages') and profile.bedrock_supports_prompt_caching:
             last_user_content = self._get_last_user_message_content(processed_messages)
             if last_user_content is not None:
-                # AWS currently rejects cache points that directly follow non-text content.
-                # Insert a newline text block as a workaround.
-                if 'text' not in last_user_content[-1]:
-                    last_user_content.append({'text': '\n'})
-                last_user_content.append({'cachePoint': {'type': 'default'}})
+                # AWS currently rejects cache points that directly follow document content.
+                # only add cache point if last block is not a document
```
DouweM: Instead of not adding a cache point at all in this case, should we add it before the document?

tests/models/test_bedrock.py
```diff
@@ -1942,6 +1942,35 @@ def diagnostics() -> str:  # pragma: no cover - exercised via agent call
...
+
+    result = await agent.run(run_args)
+    assert 'document' in result.output.lower() or 'analysis' in result.output.lower()
+
+    # Verify cache tokens are not being used
+    usage = result.usage()
+    assert usage.input_tokens > 0
+    assert usage.output_tokens > 0
+    # Should have cache write tokens on first run
+    assert usage.cache_write_tokens == 0 and usage.cache_read_tokens == 0
```
DouweM, line 1970: `== 0` contradicts the `Should have cache write tokens on first run` comment :)

tests/models/test_bedrock.py
```diff
@@ -2142,7 +2171,7 @@ async def test_bedrock_cache_messages_with_binary_content(
         ModelRequestParameters(),
         BedrockModelSettings(bedrock_cache_messages=True),
     )
-    # Should add newline text block before cache point when last content is not text
+    # Should not add cache point for document content
```
DouweM, line 2173: I think that if there was text preceding the document, the cache point should be inserted after the text, instead of being ignored at all.

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/models/bedrock.py
matthgt: updated code to use logic from #4128 and added more tests.

pydantic_ai_slim/pydantic_ai/models/bedrock.py
matthgt: with the new code this is no longer an issue since we're reverting to the original "add a \n" workaround

tests/models/test_bedrock.py
matthgt: reverted to previous behavior so this is no longer needed

pydantic_ai_slim/pydantic_ai/models/bedrock.py
DouweM: @matthgt Are you sure the "inset newline before cache point" trick makes sense to keep at all? I thought the issue was that that cache point would never be used because the `\n` will not be in future message histories. So perhaps the better behavior would be to add the cache point at the last place where we know it'll still make a difference, i.e. before trailing documents?

matthgt: probably not, I just don't have more spare cycles right now to make it better. This at least fixes it for tool parts, which is what I had been seeing.

matthgt:
@DouweM any hope you'd want to merge this as is and table further fixes for later?

DouweM:
@matthgt I'd much prefer to fully fix the issue instead of just for tool calls, so I was planning to spend a minute looking into this tomorrow, but if I can't find time to improve it I'll merge and release this as-is tomorrow.

matthgt:
> @matthgt I'd much prefer to fully fix the issue instead of just for tool calls, so I was planning to spend a minute looking into this tomorrow, but if I can't find time to improve it I'll merge and release this as-is tomorrow.

Sounds good, thank you for your help getting this over the line!

DouweM:
OK I've made some tweaks in https://github.com/pydantic/pydantic-ai/pull/4152, let's continue there.

Issue #3781, opened 2025-12-19
Support Skills with any model


DouweM:
### Description

Related to https://github.com/pydantic/pydantic-ai/issues/3365

As I wrote there:

> It could leverage native support when available, but I'm not sure that actually make sense considering access to local files, running scripts etc. If we do go that route (under some conditions, like there being no scripts), note that it's also how we're planning to support deferred loading of tools and programmatic tool calling: https://github.com/pydantic/pydantic-ai/issues/3666#issuecomment-3629161845, by essentially falling back from the "builtin tool" to a Pydantic AI side implementation if the builtin tool is not supported by a given model: https://github.com/pydantic/pydantic-ai/issues/3212

jerry-reevo:
Some recent events worth noting as they could impact this work.

1. OpenAI has recently adopted skills in Codex: https://developers.openai.com/codex/skills/
2. Anthropic has created an open standard for skills: https://agentskills.io/home

DouweM:
@jerry-reevo Yep, @dougtrajano also mentioned VSCode on the PR: https://github.com/microsoft/vscode/pull/278445

Wh1isper:
I think we can achieve this through `Toolset`. 

Skills are fundamentally based on the file system, injecting into the system prompt to guide the model in further ingesting instructions via the file system. 

Currently, as long as we allow Toolset to inject into the system prompt, we can implement skills. (A file-reading Toolset can only access files within a pre-configured folder and inject the skill's metadata into the system prompt.)

> The above was written before I read #3780.

After reading #3780, it's great that we're getting Toolset to inject the system prompt. Based on my practical experience, the `list_skills` tool is unlikely to be noticed by agents without any system prompt guidance. Once we need to inject system prompts, the best approach is to always load metadata: https://platform.claude.com/docs/en/agents-and-tools/agent-skills/overview#level-1-metadata-always-loaded

linjer:
@DouweM I'm interested in taking a stab at this but would like your thoughts on the design direction first.

**The challenge**: Agent Skills are filesystem-based (directories with SKILL.md, scripts/, references/, assets/) while pydantic-ai is environment-agnostic with no built-in filesystem, shell, or sandbox.

**Proposed scope**:

Core: Skill parsing, discovery, and management APIs — the Skill dataclass, metadata parsing, and a SkillLoader protocol

[...]

PR #3780, opened 2025-12-19
Introduce support for agent skills

DougTrajano:
- Closes https://github.com/pydantic/pydantic-ai/issues/3781

This pull request introduces a new "Skills" system to Pydantic AI, enabling modular, progressive skill discovery and execution for agents. The main changes include the addition of the `SkillsToolset` and related types, updates to documentation and examples to demonstrate skill usage, and new skill definitions and scripts for practical use.

> **Warning!** The `pyYAML` dependency was added to `pydantic-ai-slim` because Agent Skills uses YAML frontmatter to structure skill metadata.

**References:**
- [DougTrajano/pydantic-ai-skills](https://github.com/DougTrajano/pydantic-ai-skills/tree/main)
- [vstorm-co/pydantic-deepagents](https://github.com/vstorm-co/pydantic-deepagents)

**Skills Toolset Integration**
* Added `SkillsToolset` and supporting types (`Skill`, `SkillMetadata`, `SkillResource`, `SkillScript`, etc.) to the main package exports, enabling agents to discover and use skills dynamically. [[1]](diffhunk://#diff-29ffd535de7b5dc891c25a56e837db2111c6950f1fe9e7570668f1022c34b72dR111) [[2]](diffhunk://#diff-29ffd535de7b5dc891c25a56e837db2111c6950f1fe9e7570668f1022c34b72dR216) [[3]](diffhunk://#diff-eb1e03202b42c9eb117dd0a94b0100d4766c5c6df7cf5fed7c52840119e3cc60R11-R24) [[4]](diffhunk://#diff-eb1e03202b42c9eb117dd0a94b0100d4766c5c6df7cf5fed7c52840119e3cc60R41-R53) [[5]](diffhunk://#diff-e47f91b4b6c770b2c10eca0ee7e09b1f97fdf474532c2e80cde003ea4f969086R1-R64)
* Created `pydantic_ai/toolsets/skills/__init__.py` with documentation and examples for building and managing agent skills.

**Documentation and API Reference Updates**
* Updated API docs (`docs/api/toolsets.md`) and navigation (`mkdocs.yml`) to include the new Skills toolset and its members, ensuring clear guidance for users. [[1]](diffhunk://#diff-e1476212f53dc1d85c19c94de787f8621e13e5995fb64871798c799e4802781bR15-R33) [[2]](diffhunk://#diff-98d0f806abc9af24e6a7c545d3d77e8f9ad57643e27211d7a7b896113e420ed2R44)
[...]

PR #4123, opened 2026-01-28
Add `AbstractToolset.get_instructions` method and `include_instructions` arg to MCP Servers

DougTrajano:
Related to https://github.com/pydantic/pydantic-ai/issues/3781#issuecomment-3807504837

### Pre-Review Checklist

> `make typecheck` is failing due to issues with unit tests.

### Pre-Merge Checklist

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/durable_exec/dbos/_mcp.py
+++ b/pydantic_ai_slim/pydantic_ai/durable_exec/dbos/_mcp.py
@@ -108,10 +108,20 @@ class DBOSMCPToolset(WrapperToolset[AgentDepsT], ABC):
         if result is not None:
             return result
         # If instructions are enabled but the server isn't initialized locally, fetch via step.
-        from pydantic_ai.mcp import MCPServer
-        from pydantic_ai.toolsets.fastmcp import FastMCPToolset
+        _mcp_types: tuple[type, ...] = ()
+        try:
+            from pydantic_ai.mcp import MCPServer
 
-        if isinstance(self.wrapped, (MCPServer, FastMCPToolset)) and self.wrapped.include_instructions:
+            _mcp_types += (MCPServer,)
+        except ImportError:
+            pass
+        try:
+            from pydantic_ai.toolsets.fastmcp import FastMCPToolset
+
+            _mcp_types += (FastMCPToolset,)
+        except ImportError:
+            pass
+        if _mcp_types and isinstance(self.wrapped, _mcp_types) and self.wrapped.include_instructions:  # type: ignore[union-attr]
             return await self._dbos_wrapped_get_instructions_step(ctx)
         return None
 
--- a/pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_mcp.py
+++ b/pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_mcp.py
@@ -104,10 +104,20 @@ class TemporalMCPToolset(TemporalWrapperToolset[AgentDepsT], ABC):
             return await super().get_instructions(ctx)
 
         # If instructions are enabled, fetch via activity (the server isn't initialized locally in workflows).
-        from pydantic_ai.mcp import MCPServer
-        from pydantic_ai.toolsets.fastmcp import FastMCPToolset
+        _mcp_types: tuple[type, ...] = ()
+        try:
+            from pydantic_ai.mcp import MCPServer
 
-        if isinstance(self.wrapped, (MCPServer, FastMCPToolset)) and self.wrapped.include_instructions:
+            _mcp_types += (MCPServer,)
+        except ImportError:
+            pass
+        try:
+            from pydantic_ai.toolsets.fastmcp import FastMCPToolset
+
+            _mcp_types += (FastMCPToolset,)
+        except ImportError:
+            pass
+        if _mcp_types and isinstance(self.wrapped, _mcp_types) and self.wrapped.include_instructions:  # type: ignore[union-attr]
             serialized_run_context = self.run_context_type.serialize_run_context(ctx)
             activity_config: ActivityConfig = {'summary': f'get instructions: {self.id}', **self.activity_config}
             return await workflow.execute_activity(


tests/test_mcp.py
```diff
@@ -2186,15 +2246,6 @@ async def test_resource_methods_without_capability(mcp_server: MCPServerStdio) -
             assert result == []
```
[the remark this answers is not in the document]
DougTrajano: @DouweM can you handle that when writing the release notes?

pydantic_ai_slim/pydantic_ai/toolsets/wrapper.py
```diff
@@ -49,3 +57,12 @@ def visit_and_replace(
...
         return replace(self, wrapped=self.wrapped.visit_and_replace(visitor))
+
+    def __getattr__(self, name: str) -> Any:
+        """Fallback delegation for custom attributes and methods from the wrapped toolset.
+
+        This enables runtime access to custom methods and attributes that may be defined
+        on the wrapped toolset but not on this wrapper. Note that static type checkers
+        may not recognize these delegated attributes.
+        """
+        return getattr(self.wrapped, name)
```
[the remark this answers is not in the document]
DougTrajano: @DouweM I will leave this comment for you. For me, the exception message seems good enough for debugging.

When I was developing the `SkillsToolset` and testing the toolset composition methods, I encountered challenges because the `WrappedToolset` lacked the same methods and attributes I defined in my custom toolset.

pydantic_ai_slim/pydantic_ai/toolsets/abstract.py
```diff
@@ -109,6 +109,21 @@ async def __aexit__(self, *args: Any) -> bool | None:
         """
         return None
 
+    async def get_instructions(self, ctx: RunContext[AgentDepsT]) -> str | None:
```
Kludex:
```suggestion
    async def instructions(self, ctx: RunContext[AgentDepsT]) -> str | None:
```

DouweM: For consistency with `get_tools`, I think it should still be `get_instructions`. And I don't think we should be removing the existing `MCPServer.instructions` method.

DouweM:
@Kludex Do you know if the MCP server's instructions should be included in the agent instructions by default?

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -465,14 +476,31 @@ def capabilities(self) -> ServerCapabilities:
             )
         return self._server_capabilities
 
-    @property
```
adtyavrdhn, line 473: This seems to be a breaking change, I don't think that is something we can do until v2? @DouweM

DouweM: Correct, I want to restore this field as I suggested in https://github.com/pydantic/pydantic-ai/pull/4123#discussion_r2737964746.

pydantic_ai_slim/pydantic_ai/toolsets/abstract.py
DougTrajano: +1 vote for `get_instructions` hahah

I usually prefer to have the verb in methods, but the final decision is your guys. I am ok with both ones.

About removing the `MCPServer.instructions` property, we discussed it in the previous PR (https://github.com/pydantic/pydantic-ai/pull/3780#discussion_r2695920671)

My proposal is to keep both, with a deprecation message that `instructions` will be removed in the future, since they have the same functionality.

DougTrajano:
> @Kludex Do you know if the MCP server's instructions should be included in the agent instructions by default?

Let me share my two cents here

I strongly believe that it SHOULD NOT be included by default. Currently, users must explicitly add it to the agent.

```python
from pydantic_ai import Agent
from pydantic_ai.mcp import MCPServerStreamableHTTP

server = MCPServerStreamableHTTP('http://localhost:8000/mcp')
agent = Agent('gateway/openai:gpt-5', toolsets=[server])

@agent.instructions
async def mcp_server_instructions():
    return server.instructions  
```

Ref: https://ai.pydantic.dev/mcp/client/#server-instructions

The new argument (`use_server_instructions`), btw I am think on renaming it to `add_server_instructions`) turns it simpler:

```python
from pydantic_ai import Agent
from pydantic_ai.mcp import MCPServerStreamableHTTP

server = MCPServerStreamableHTTP(
    'http://localhost:8000/mcp',
    use_server_instructions=True
)

agent = Agent('openai:gpt-5', toolsets=[server])
```

I’m a bit concerned about agents that use multiple MCP servers containing their own instructions. In those cases, the system prompt can become enormous without the developer explicitly requesting it.

I believe that it's a really open discussion. I didn't find a good reference in the https://modelcontextprotocol.io/

only this https://modelcontextprotocol.io/specification/2025-11-25/schema#initializeresult-instructions

adtyavrdhn:
Yeah I am inclined to agree with @DougTrajano as well, I don't think it should be added to the agent's instructions by default but unaware what the MCP spec outlines(if it does at all).

DouweM:
@DougTrajano @adtyavrdhn Yeah I agree we shouldn't include by default. I wonder if there are guidelines about that though. I'm specifically asking @Kludex because he's an MCP maintainer :)

DouweM:
@DougTrajano Please have a look at CI and the Devin comment!

olaservo:
> I didn't find a good reference in the https://modelcontextprotocol.io/
> 
> only this https://modelcontextprotocol.io/specification/2025-11-25/schema#initializeresult-instructions

Hi @DougTrajano I saw your comment about this - linking to a couple official resources about server instructions:

- MCP blog post: https://blog.modelcontextprotocol.io/posts/2025-11-03-using-server-instructions/
- There was also more documentation [added yesterday](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2184) to the spec docs here: https://modelcontextprotocol.io/specification/draft/schema#initializeresult-instructions

Hope that helps!

DougTrajano:
> @DougTrajano Please have a look at CI and the Devin comment!

I will review them and update this PR when I come back from vacation next week. Thank you

DougTrajano:
> > I didn't find a good reference in the https://modelcontextprotocol.io/
> 
> > 
> 
> > only this https://modelcontextprotocol.io/specification/2025-11-25/schema#initializeresult-instructions
> 
> 
> 
> Hi @DougTrajano I saw your comment about this - linking to a couple official resources about server instructions:
> 
> 
> 
> - MCP blog post: https://blog.modelcontextprotocol.io/posts/2025-11-03-using-server-instructions/
> 
> - There was also more documentation [added yesterday](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2184) to the spec docs here: https://modelcontextprotocol.io/specification/draft/schema#initializeresult-instructions
> 
> 
> 
> Hope that helps!

Thank you for sharing these links. I will reference them

pedroallenrevez:
Any help needed here? I did a PR that was exactly the same as this.

DougTrajano:
> Any help needed here? I did a PR that was exactly the same as this.

Hi @pedroallenrevez, I am on vacation and I will be back next Monday then I will continue the work pn this PR.

I think that it's under control, but I appreciate your support!

pydantic_ai_slim/pydantic_ai/toolsets/abstract.py
DougTrajano: @DouweM @Kludex Since I didn't receive any new comments here, I am deciding on `get_instructions` (2x1 😆)

Feel free to share any concerns.

DougTrajano:
Hi guys, I am back in the game.

I updated the code based on the CI and Devin's comments. @DouweM can you review again, pls?

I also incorporated some changes proposed in #4282 by @pedroallenrevez that were not present in this PR.

pydantic_ai_slim/pydantic_ai/toolsets/function.py
```diff
@@ -236,6 +251,37 @@ def tool_decorator(
...
+        toolset = FunctionToolset[int]()
+
+        @toolset.instructions
+        def my_instructions(ctx: RunContext[int]) -> str:
+            return 'Always use the search tool when looking for information.'
+
+        @toolset.tool
+        def search(ctx: RunContext[int], query: str) -> str:
+            return f'Results for: {query}'
+        ```
```
Kludex, line 435:
```suggestion
            ```python
	        from pydantic_ai import FunctionToolset, RunContext
	
	        toolset = FunctionToolset[int]()
	
	        @toolset.instructions
	        def my_instructions(ctx: RunContext[int]) -> str:
	            return 'Always use the search tool when looking for information.'
	
	        @toolset.tool
	        def search(ctx: RunContext[int], query: str) -> str:
	            return f'Results for: {query}'
	        ```
```

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -908,6 +955,8 @@ def __init__(
                 See [`MCPServer.cache_tools`][pydantic_ai.mcp.MCPServer.cache_tools].
             cache_resources: Whether to cache the list of resources.
                 See [`MCPServer.cache_resources`][pydantic_ai.mcp.MCPServer.cache_resources].
+            add_server_instructions: Whether to include the server's instructions in the agent's system prompt.
```
Kludex:
```suggestion
            add_server_instructions: Whether to include the server's instructions in the agent's instructions
```

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -466,14 +477,48 @@ def capabilities(self) -> ServerCapabilities:
...
+        'The `instructions` property is deprecated. '
+        "Use `add_server_instructions=True` when constructing the server to have instructions automatically injected into the agent's system prompt."
+    )
     def instructions(self) -> str | None:
-        """Access the instructions sent by the MCP server during initialization."""
+        """Access the instructions sent by the MCP server during initialization.
+
+        .. deprecated::
+            Use ``add_server_instructions=True`` when constructing the server to have instructions
+            automatically injected into the agent's system prompt.
```
Kludex: This is not google style, it's not needed. The documentation uses the decorator to render correctly in the docs.

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -354,6 +354,15 @@ class MCPServer(AbstractToolset[Any], ABC):
...
     """
 
+    add_server_instructions: bool
+    """Whether to include the server's instructions in the agent's system prompt.
+
+    When enabled, the instructions sent by the MCP server during initialization
+    will be automatically injected into the agent's system prompt via `instructions()`.
+
+    Defaults to `False` for backward compatibility.
+    """
```
Kludex, line 366: I think `add_server_instructions` is not a good name for this parameter. "Add" to where? We are already dealing with the "server" here anyway.

Shall we call it `include_instructions`?

pydantic_ai_slim/pydantic_ai/toolsets/abstract.py
```diff
@@ -109,6 +109,21 @@ async def __aexit__(self, *args: Any) -> bool | None:
...
+        Override this method to provide custom instructions that help the agent understand
+        how to use the tools in this toolset effectively.
+
+        Args:
+            ctx: The run context for this agent run.
+
+        Returns:
+            Instructions string to add to the system prompt, or None if no instructions.
+        """
+        del ctx  # unused in base implementation
```
Kludex: But why?

Kludex:
@DougTrajano Can you fix the pipeline please?

pydantic_ai_slim/pydantic_ai/mcp.py
DougTrajano: YEap, `include_instructions` is a better name. I will change it.

pydantic_ai_slim/pydantic_ai/toolsets/abstract.py
DougTrajano: lint issues, I will delete it haha

pydantic_ai_slim/pydantic_ai/mcp.py
DougTrajano: Okay, I am deleting it.

pydantic_ai_slim/pydantic_ai/toolsets/function.py
```diff
@@ -236,6 +251,37 @@ def tool_decorator(
...
+
+                @toolset.tool
+                def search(ctx: RunContext[int], query: str) -> str:
+                    return f'Results for: {query}'
+                ```
+
+        Args:
+            func: The instructions function to register.
+        """
+        self._instructions_funcs.append(SystemPromptRunner(func))
```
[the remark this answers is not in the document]
DougTrajano: The concern is valid, but this pattern already exists in the codebase.

`FunctionToolset.instructions` is intentionally mirroring the exact same pattern already established by `Agent.system_prompt`:

- `Agent.__init__` has a `system_prompt: str | Sequence[str]` parameter
- `Agent.system_prompt` is also a decorator method (line 1038)

Both share the same naming overlap. `agent.system_prompt` gives you the decorator method, not the constructor argument, so `FunctionToolset.instructions` is consistent with existing library codebase.

@Kludex @DouweM thoughts?

DougTrajano:
> @DougTrajano Can you fix the pipeline please?

done

pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -653,15 +653,30 @@ async def main():
...
+
+            async def collect_toolset_instructions(ts: AbstractToolset[AgentDepsT]) -> None:
+                instruction = await ts.get_instructions(run_context)
+                if instruction:
+                    toolset_instructions.append(instruction)
+
+            toolsets_to_visit: list[AbstractToolset[AgentDepsT]] = []
+            toolset.apply(toolsets_to_visit.append)
+            for ts in toolsets_to_visit:
+                await collect_toolset_instructions(ts)
```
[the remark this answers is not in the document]
DougTrajano: For me, it's not an issue.

docs/toolsets.md
```diff
@@ -652,10 +782,15 @@ print([t.name for t in test_model.last_model_request_parameters.function_tools])
 
 _(This example is complete, it can be run "as is")_
 
+!!! note "Dynamic toolset instructions are unavailable on the first model request"
+    Because a dynamic toolset is resolved inside [`get_tools()`][pydantic_ai.toolsets.AbstractToolset.get_tools] — which is called to build each model request — its [`get_instructions()`][pydantic_ai.toolsets.AbstractToolset.get_instructions] result is `None` until the toolset has been resolved at least once. This means that on the very first request of a run, any instructions defined on the inner toolset are silently omitted. If you need toolset instructions to be present from the very first request, define them on a `FunctionToolset` registered directly with the agent rather than inside a dynamic toolset factory.
```
DouweM: > This means that on the very first request of a run, any instructions defined on the inner toolset are silently omitted.

I would very much like to find a way around this. Can we not ensure that `toolset.get_instructions(ctx)` is called after `toolset.get_tools(ctx)`?

DouweM: Or we can have `get_instructions(ctx)` build the inner toolset, like `get_tools` does?

pydantic_ai_slim/pydantic_ai/toolsets/function.py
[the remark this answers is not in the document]
DouweM: I'm OK as-is

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -138,7 +138,7 @@ class GraphAgentDeps(Generic[DepsT, OutputDataT]):
     usage_limits: _usage.UsageLimits
     max_result_retries: int
     end_strategy: EndStrategy
-    get_instructions: Callable[[RunContext[DepsT]], Awaitable[str | None]]
```
DouweM, line 129: Why did we rename this? I liked having it very clear that this is a factory

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -479,6 +494,31 @@ def instructions(self) -> str | None:
...
+        Raises:
+            AttributeError: If ``include_instructions`` is ``True`` but the server
+                has not been initialized yet.
+        """
+        if self.include_instructions:
+            if not hasattr(self, '_instructions'):
+                raise AttributeError(
+                    f'The `{self.__class__.__name__}.get_instructions` is only instantiated after initialization.'
+                )
+            return self._instructions
```
DouweM: Can this use `self.instructions`?

DouweM:
> Can we make `get_instructions` support both sync and async modes? Just like [@agent.instructions](https://ai.pydantic.dev/agent/#instructions)
> 
> ```python
> def get_instructions(self, ctx: RunContext[AgentDepsT]) -> str | Awaitable[str | None] | None:
> ```

@Wh1isper The `FunctionToolset.instructions` decorator supports both, but I'd rather require the abstract method on `AbstractToolset` to be async, like `get_tools(ctx)` is.

DougTrajano:
> Can we make `get_instructions` support both sync and async modes? Just like [@agent.instructions](https://ai.pydantic.dev/agent/#instructions)
> 
> ```python
> def get_instructions(self, ctx: RunContext[AgentDepsT]) -> str | Awaitable[str | None] | None:
> ```

yeap, we can. Thank you for the catch-up.

pydantic_ai_slim/pydantic_ai/mcp.py
DougTrajano: We can, but if we want to deprecate the `self.instructions` in the future in favor of the `self.get_instructions()` and `add_server_instructions` parameter, I recommend that we avoid it.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
DougTrajano: Ops, I renamed it when I rolled back everything to `get_instructions`, it's my fault, sorry.

I rolled it back now.

docs/toolsets.md
DougTrajano: Okay, it's now fixed. The root cause was that `get_instructions` was being called in `UserPromptNode` before `for_run_step` had a chance to resolve dynamic toolsets. The fix moves the `get_instructions` call into `ModelRequestNode._prepare_request()`, immediately after `for_run_step`.

This is exactly your first suggestion: `get_instructions` is now always called after `get_tools`. All the redundant `get_instructions` calls scattered across `CallToolsNode` (empty response retry, tool retry error, `_handle_tool_calls`) have been removed — `_prepare_request` is the single call site.

The only remaining early call is the short-circuit guard in `UserPromptNode` that skips an unnecessary model request when resuming with a `ModelResponse` in history and no new user prompt. That path only fires when at least one prior model request has already completed, so `DynamicToolset._toolset` is already resolved — `for_run_step` doesn't need to run first.

DougTrajano: Based on my tests, it's working well, but I am a little bit afraid of breaking something here. Please review it carefully.

The specific commit is https://github.com/pydantic/pydantic-ai/pull/4123/changes/40c704dfc688b8803a368ed0d544612ce88f31d4

pydantic_ai_slim/pydantic_ai/_agent_graph.py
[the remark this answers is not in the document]
DougTrajano: The short-circuit check at line 259 is unchanged by this PR — it was already there before and is not in the diff. The only `get_instructions()` call removed from `UserPromptNode.run()` was in the main path (`if not run_context:` block).

The scenario you describe also can't happen in practice for this code path. The check at line 256 only fires when `last_message` is a `ModelResponse`, which means at least one prior `_prepare_request()` call has already run `for_run_step()` and resolved `DynamicToolset._toolset`. So `_toolset` is never `None` here in a normal agent run. The only exception would be passing pre-built `message_history` from a prior session — but that was true before this PR too.

docs/toolsets.md
DougTrajano: I ran some tests and it appears to work well with my agents, but I'm a bit concerned about causing issues. Please review it to ensure nothing is broken.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -267,7 +267,6 @@ async def run(  # noqa: C901
 
         if not run_context:
             run_context = build_run_context(ctx)
```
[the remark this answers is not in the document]
DougTrajano: You're right. The root cause is that the guard in `UserPromptNode.run()` runs before instructions are resolved, so it can never give a correct answer for the instructions dimension regardless of how its condition is written.

The fix is to remove the check from `UserPromptNode.run()` entirely and move it to `_prepare_request()`, after `self.request.instructions` has been set by `get_instructions()`. Prior message history is obtained via `ctx.state.message_history[:-1]` since `self.request` was already appended earlier in `_prepare_request()`:

```python
self.request.instructions = await ctx.deps.get_instructions(run_context)

# Validate after instructions are resolved; self.request was appended above so [:-1] is prior history
if not ctx.state.message_history[:-1] and not self.request.parts and not self.request.instructions:
    raise exceptions.UserError('No message history, user prompt, or instructions provided')
```

This preserves the original user-friendly error for truly empty runs while allowing instructions-only runs (e.g. `agent.run(None)` with a DynamicToolset providing instructions) to proceed correctly.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -528,6 +524,13 @@ async def _prepare_request(
         # This will raise errors for any tool name conflicts
         ctx.deps.tool_manager = await ctx.deps.tool_manager.for_run_step(run_context)
 
+        # Fetch instructions now that dynamic toolsets have been resolved by for_run_step
+        self.request.instructions = await ctx.deps.get_instructions(run_context)
+
```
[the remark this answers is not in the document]
DougTrajano: Yeap, I updated the snapshot step in `test_logfire.py`.

@DouweM do you think that we need to document it? In my opinion, it's too deep, very internal stuff.

pydantic_ai_slim/pydantic_ai/mcp.py
DouweM: I don't see a strong need to deprecate `.instructions`, in case people want to access  it directly for some reason.

DougTrajano: Okay, I remove the deprecation and update the `get_instructions()` to use `instructions` property.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -528,6 +524,13 @@ async def _prepare_request(
         # This will raise errors for any tool name conflicts
         ctx.deps.tool_manager = await ctx.deps.tool_manager.for_run_step(run_context)
 
+        # Fetch instructions now that dynamic toolsets have been resolved by for_run_step
+        self.request.instructions = await ctx.deps.get_instructions(run_context)
```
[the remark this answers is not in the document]
DougTrajano: @DouweM this is the price to have the `get_instructions()` resolved after `get_tools()`.

I don't know the entire impact of this change, based on my understanding of the codebase, that's okay, but the final decision is yours.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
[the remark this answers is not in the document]
DougTrajano: It's technically correct, but the practical impact is very narrow as it requires all of these conditions simultaneously:

- A brand-new agent instance resuming with a provided `message_history` (so `DynamicToolset._toolset` is `None`)
- The last message in history is a `ModelResponse` (not `ModelRequest`)
- No `user_prompt` is provided (resuming without a prompt)
- The only source of instructions is a `DynamicToolset` (no literal instructions or instruction functions on the agent)

However, I fixed it by calling `for_run_step()` before `get_instructions()` in this path, matching the pattern already used in `ModelRequestNode._prepare_request()`. No performance impact since `ToolManager.for_run_step()` caches by `run_step` — subsequent calls at the same step return self without invoking `get_tools()` again.

tests/test_logfire.py
DouweM: @DougTrajano I think the `run_step == 1` behavior is actually better than `run_step == 0` which we had before, which could be considered a bug. It's better for it to be consistent with tool prepare functions etc. 

But then we can drop the `+ 1` here to make the numbers below as expected!

pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -654,15 +654,30 @@ async def main():
...
+
+            parts.extend(toolset_instructions)
```
DouweM: Wouldn't it be enough to call `await toolset.get_instructions(ctx)`, assuming `CombinedToolset` and `WrappedToolset` appropriately forward the call to their child(ren)?

DouweM: We may want to allow `get_instructions()` to return `list[str]` as well, to make it easier to combine them and then only join by `\n\n` at the last moment.

DouweM (changes requested):

tests/test_logfire.py
DougTrajano: Ok, done!

pydantic_ai_slim/pydantic_ai/agent/__init__.py
DougTrajano: Ok, done! Good catch, the code is so much clearer and dev friendly.

DougTrajano:
Done @DouweM

DouweM:
@DougTrajano Please have a look at coverage.

DougTrajano:
> @DougTrajano Please have a look at coverage.

resolved.

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -909,6 +939,8 @@ def __init__(
                 See [`MCPServer.cache_tools`][pydantic_ai.mcp.MCPServer.cache_tools].
             cache_resources: Whether to cache the list of resources.
                 See [`MCPServer.cache_resources`][pydantic_ai.mcp.MCPServer.cache_resources].
+            include_instructions: Whether to include the server's instructions in the agent's instructions
```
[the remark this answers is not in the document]
DouweM: I prefer `instructions` in both places, but this is pretty minor.

pydantic_ai_slim/pydantic_ai/toolsets/abstract.py
```diff
@@ -117,6 +117,21 @@ async def __aexit__(self, *args: Any) -> bool | None:
         """
         return None
 
+    async def get_instructions(self, ctx: RunContext[AgentDepsT]) -> str | list[str] | None:
+        r"""Return instructions to inject into the agent's system prompt when this toolset is used.
```
DouweM: Should be "agent's instructions" here and everywhere

pydantic_ai_slim/pydantic_ai/toolsets/abstract.py
```diff
@@ -117,6 +117,21 @@ async def __aexit__(self, *args: Any) -> bool | None:
...
+
+        Override this method to provide custom instructions that help the agent understand
+        how to use the tools in this toolset effectively.
+
+        Args:
+            ctx: The run context for this agent run.
+
+        Returns:
+            Instructions string (or list of strings) to add to the system prompt, or None if
+            no instructions. A list of strings will be joined with ``\n\n`` by the agent.
```
DouweM: Single backticks please as we render these docstrings as Markdown. This is something Claude loves to do wrong

pydantic_ai_slim/pydantic_ai/toolsets/function.py
```diff
@@ -103,6 +108,16 @@ def __init__(
         self.requires_approval = requires_approval
         self.metadata = metadata
 
+        self._instructions_funcs: list[str | SystemPromptRunner[AgentDepsT]] = []
```
DouweM: Shouldn't be called `funcs` if it can also hold plain `str`s. Just `self._instructions` will do

pydantic_ai_slim/pydantic_ai/toolsets/function.py
```diff
@@ -352,6 +398,19 @@ def add_tool(self, tool: Tool[AgentDepsT]) -> None:
...
+            return None
+        parts: list[str] = []
+        for func in self._instructions_funcs:
+            if isinstance(func, str):
+                parts.append(func)
+            else:
+                result = await func.run(ctx)
+                if result:
+                    parts.append(result)
+        return '\n\n'.join(parts).strip() or None
```
DouweM: Can return an unjoined list now, right?

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -359,6 +359,15 @@ class MCPServer(AbstractToolset[Any], ABC):
     Set to `False` for servers that change resources dynamically without sending notifications.
     """
 
+    include_instructions: bool
+    """Whether to include the server's instructions in the agent's system prompt.
```
DouweM: "agent's instructions" please

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -359,6 +359,15 @@ class MCPServer(AbstractToolset[Any], ABC):
...
+
+    When enabled, the instructions sent by the MCP server during initialization
+    will be automatically injected into the agent's system prompt via `instructions()`.
```
DouweM: This whole paragraph seems unnecessary

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -479,6 +490,23 @@ def instructions(self) -> str | None:
             )
         return self._instructions
 
+    async def get_instructions(self, ctx: RunContext[Any]) -> str | None:
```
DouweM: I'd like this to match the super signature, meaning include `list[str]` even if we don't use it

docs/mcp/client.md
```diff
@@ -342,25 +342,25 @@ agent = Agent('openai:gpt-5.2', toolsets=[weather_server, calculator_server])
 
 MCP servers can provide instructions during initialization that give context about how to best interact with the server's tools. These instructions are accessible via the [`instructions`][pydantic_ai.mcp.MCPServer.instructions] property after the server connection is established.
 
+You can enable automatic instruction injection when creating the MCP server by setting the `include_instructions` parameter to `True`. This will fetch the server's instructions during connection and inject them into the agent's prompt:
```
DouweM: `agent's instructions`

docs/mcp/client.md
```diff
@@ -342,25 +342,25 @@ agent = Agent('openai:gpt-5.2', toolsets=[weather_server, calculator_server])
...
+
 ```python {title="mcp_server_instructions.py"}
 from pydantic_ai import Agent
 from pydantic_ai.mcp import MCPServerStreamableHTTP
 
-server = MCPServerStreamableHTTP('http://localhost:8000/mcp')
-agent = Agent('openai:gpt-5.2', toolsets=[server])
+server = MCPServerStreamableHTTP(
+    'http://localhost:8000/mcp',
+    include_instructions=True
```
DouweM: This example doesn't actually demonstrate in any way that the feature works :) I think we can remove the example entirely, and simplify the paragraphs above.

pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -1514,6 +1523,23 @@ def _get_instructions(
...
+        are collected from all leaf toolsets and combined into the final system prompt.
+
+        Args:
+            additional_instructions: Additional instructions to include for this run.
+
+        Returns:
+            A tuple of (literal_instructions, instruction_functions) where:
+            - literal_instructions: Combined literal string instructions or None
+            - instruction_functions: List of instruction functions that need to be evaluated at runtime
+        """
```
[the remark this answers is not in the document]
DouweM: It's a bit much, but I'm OK with it.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -256,6 +256,9 @@ async def run(  # noqa: C901
             elif isinstance(last_message, _messages.ModelResponse):
                 if self.user_prompt is None:
                     run_context = build_run_context(ctx)
+                    # Resolve dynamic toolsets before fetching instructions, so DynamicToolset.get_instructions()
+                    # can delegate to the resolved inner toolset instead of returning None
+                    ctx.deps.tool_manager = await ctx.deps.tool_manager.for_run_step(run_context)
```
DouweM, line 247: Would it make sense to move this into the `ctx.deps.get_instructions` function?

docs/toolsets.md
```diff
@@ -113,6 +113,136 @@ print([t.name for t in test_model.last_model_request_parameters.function_tools])
...
+"""
+Use weather tools for forecasts.
+
+Use calendar tools for scheduling.
+"""
+```
+
+_(This example is complete, it can be run "as is")_
+
+Custom toolsets that extend [`AbstractToolset`][pydantic_ai.toolsets.AbstractToolset] can also provide instructions by overriding the [`get_instructions()`][pydantic_ai.toolsets.AbstractToolset.get_instructions] method. Wrapper toolsets (like [`PrefixedToolset`][pydantic_ai.toolsets.PrefixedToolset]) automatically pass through instructions from the wrapped toolset.
```
DouweM: Not worth mentioning here as we already have it in the "custom toolset" section below

docs/toolsets.md
```diff
@@ -656,6 +786,8 @@ _(This example is complete, it can be run "as is")_
 
 To define a fully custom toolset with its own logic to list available tools and handle them being called, you can subclass [`AbstractToolset`][pydantic_ai.toolsets.AbstractToolset] and implement the [`get_tools()`][pydantic_ai.toolsets.AbstractToolset.get_tools] and [`call_tool()`][pydantic_ai.toolsets.AbstractToolset.call_tool] methods.
 
+You can also override the [`get_instructions()`][pydantic_ai.toolsets.AbstractToolset.get_instructions] method to provide instructions that will be injected into the agent's system prompt. This is useful for helping the model understand how to effectively use your toolset's tools.
```
DouweM: "agent's instructions"

tests/test_mcp.py
```diff
@@ -184,12 +184,72 @@ async def process_tool_call(
...
+
+
+def test_streamable_http_server_with_use_server_instructions():
+    """Test that StreamableHTTP server can be configured with use_server_instructions=True."""
+    http_server = MCPServerStreamableHTTP(
+        url='http://localhost:8000/mcp',
+        use_server_instructions=True
+    )
+    assert http_server.use_server_instructions is True
+
```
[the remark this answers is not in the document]
DouweM: This is a good point. We can add instructions to the `tests.mcp_server`.

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/agent/__init__.py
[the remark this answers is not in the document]
DougTrajano: Yeap, but I can make it shorter, while keeping it well-described. Changed it.

pydantic_ai_slim/pydantic_ai/mcp.py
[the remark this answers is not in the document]
DougTrajano: I will standardize it as `instructions`

pydantic_ai_slim/pydantic_ai/toolsets/abstract.py
DougTrajano: yeap ahah fixed

pydantic_ai_slim/pydantic_ai/toolsets/function.py
DougTrajano: Yeap, for sure. Good catch

pydantic_ai_slim/pydantic_ai/mcp.py
DougTrajano: yeap, I deleted it when reviewing the previous comment haha

pydantic_ai_slim/pydantic_ai/_agent_graph.py
DougTrajano: The `ctx.deps.get_instructions` captures toolset but has no reference to `ctx.deps.tool_manager`. It can call `toolset.get_instructions(run_context)`, but it can't call `tool_manager.for_run_step()` and write the result back to `ctx.deps.tool_manager`.

The only clean alternative I believe is to fix it in the `DynamicToolset`, but it causes a side effect of requesting instructions, which would be unexpected and could lead to double-resolution when `for_run_step` is called shortly afterward. I'd leave it as-is.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -528,6 +527,13 @@ async def _prepare_request(
         # This will raise errors for any tool name conflicts
         ctx.deps.tool_manager = await ctx.deps.tool_manager.for_run_step(run_context)
 
+        # Fetch instructions now that dynamic toolsets have been resolved by for_run_step
+        self.request.instructions = await ctx.deps.get_instructions(run_context)
+
+        # Validate after instructions are resolved; self.request was appended above so [:-1] is prior history
+        if not ctx.state.message_history[:-1] and not self.request.parts and not self.request.instructions:
+            raise exceptions.UserError('No message history, user prompt, or instructions provided')
```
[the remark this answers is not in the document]
DougTrajano: Yeap, it's intentional. I will add test cases to describe this edge case.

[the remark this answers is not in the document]
DougTrajano: @DouweM are you okay with this behavior? Just double-check

We will not raise an exception when at least the `toolset` has instructions.

```python
    from pydantic_ai import Agent

    toolset = FunctionToolset(instructions='Always use my_tool correctly.')

    @toolset.tool
    def my_tool() -> str:
        """A simple tool."""
        return 'done'

    agent = Agent(TestModel(), toolsets=[toolset])
    result = await agent.run()
    first_message = result.all_messages()[0]
    assert first_message.instructions == 'Always use my_tool correctly.'
    assert first_message.parts == []
```

tests/test_toolsets.py
```diff
@@ -892,3 +1032,318 @@ def other_tools(ctx: RunContext[None]) -> FunctionToolset[None]:
...
+    """The @toolset.instructions decorator registers instruction functions."""
+    toolset = FunctionToolset()
+
+    @toolset.instructions
+    def my_instructions() -> str:
+        return 'Use tool Y for data processing.'
+
+    ctx = build_run_context(None)
+    result = await toolset.get_instructions(ctx)
+    assert result == 'Use tool Y for data processing.'
```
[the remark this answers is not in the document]
DougTrajano: I am on int

docs/mcp/client.md
```diff
@@ -340,26 +340,7 @@ agent = Agent('openai:gpt-5.2', toolsets=[weather_server, calculator_server])
...
-    return server.instructions  # (1)!
-
-async def main():
-    result = await agent.run('What is 7 plus 5?')
-    print(result.output)
-    #> The answer is 12.
-```
-
-1. The server connection is guaranteed to be established by this point, so `server.instructions` is available.
+MCP servers can provide instructions during initialization that give context about how to best interact with the server's tools. These are accessible via the [`instructions`][pydantic_ai.mcp.MCPServer.instructions] property after the connection is established, and can be automatically injected into the agent's instructions by setting `include_instructions=True` when creating the server.
```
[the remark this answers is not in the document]
DougTrajano: No bro, actually the reviewer @DouweM suggested to remove it hahah

DougTrajano:
@DouweM done. I changed the 993 agent's system prompt to `instructions` hahah

Can you rerun the `CI / lint`, please? Looks like it's an intermittent issue

tests/test_toolsets.py
```diff
@@ -892,3 +1032,318 @@ def other_tools(ctx: RunContext[None]) -> FunctionToolset[None]:
...
+    @toolset.tool
+    def my_tool() -> str:
+        """A simple tool."""
+        return 'done'
+
+    agent = Agent(TestModel(), toolsets=[toolset])
+    result = await agent.run()
+    first_message = result.all_messages()[0]
+    assert first_message.instructions == 'Always use my_tool correctly.'  # type: ignore[union-attr]
+    assert first_message.parts == []  # type: ignore[union-attr]
```
DouweM: @DougTrajano Looks like `lint` is complaining here:

```
  /home/runner/work/pydantic-ai/pydantic-ai/tests/test_toolsets.py:1340:41 - error: Unnecessary "# type: ignore" comment (reportUnnecessaryTypeIgnoreComment)
`

File: pydantic_ai_slim/pydantic_ai/_agent_graph.py (hunk 34 above)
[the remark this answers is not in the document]
Comment (DouweM) [reply pydantic_ai_slim/pydantic_ai/_agent_graph.py:780]: @DougTrajano Hmm, it's a good point by Devin, and OpenAI's new [tool namespaces feature](https://developers.openai.com/api/docs/guides/function-calling#defining-namespaces) suggests that we should not just be dumping these into the main `instructions` at the agent graph level, but rather sending them into the model instance to decide to put them in the instructions, or in dedicated API fields.

This is similar to how the `PromptedOutput` instructions are "smuggled" through `ModelRequestParameters` and then added to the instructions at the last moment:

https://github.com/pydantic/pydantic-ai/blob/0b0ec325b651ece0f0087f8a059a91a262a5efdc/pydantic_ai_slim/pydantic_ai/models/__init__.py#L926-L930

And to how in https://github.com/pydantic/pydantic-ai/pull/3785, we're setting the tool call examples on the `ToolDefinition` object (which is included in `ModelRequestParameters`) and then the model's `prepare_request` method decides whether to insert them into the tool description, or use Anthropic's dedicated `input_examples` field.

I don't want to make this PR more complicated by supporting that entire feature, but I think the least we can do is rename `get_instructions` to `get_description` and make it clear that this will not always necessarily end up in the instructions. `MCPServer.include_instructions` can stay the same, as the MCP feature is called "server instructions".

Do you think there are other changes we'll want to make now?

I've created an issue for namespaces: https://github.com/pydantic/pydantic-ai/issues/4574

File: tests/test_toolsets.py (hunk 37 above)
Comment (DougTrajano) [reply tests/test_toolsets.py]: Fixed

File: pydantic_ai_slim/pydantic_ai/_agent_graph.py (hunk 34 above)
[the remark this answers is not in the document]
Comment (DougTrajano) [reply pydantic_ai_slim/pydantic_ai/_agent_graph.py:780]: Okay, I think this is a good way to approach it.

Here are the next steps:

This PR:

- Rename `AbstractToolset.get_instructions` to `get_description`, except for `MCPServer.include_instructions`.

Next PR for [#4574](https://github.com/pydantic/pydantic-ai/issues/4574):

- Move the `get_description` resolution from the agent graph level to the model level.
- Use `get_description` to populate the `namespace.description` in the OpenAPI API request.
  - Other providers will still be adding to the overall system prompt until they support a similar feature.
- Add `AbstractToolset.name` and populate it as `namespace.name` for OpenAI API requests.
  - I believe that it should NOT be added to the system prompt for model providers that do not have a specific definition for toolsets/namespaces, etc., as it may cause confusion with agent skills, which only add names and descriptions to the system prompt.

P.S.: I can work on that issue after finishing this one.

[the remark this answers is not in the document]
Comment (DougTrajano) [reply pydantic_ai_slim/pydantic_ai/_agent_graph.py:780]: Just a thought here.

Conceptually, I believe that `instructions` and `descriptions` are not the same thing.

`descriptions` sounds like short texts, while `instructions` can be more specific and larger. For example:

```yaml
name: jira
description: Jira tools for issue tracking and project management.
instructions: >
    There are tools that enable you to interact with Jira for managing issues and projects. Use them to create, update, and query Jira issues, as well as to retrieve information about projects and users.

    Some Jira projects may require custom fields for issue creation. If you get an error when creating an issue, you can check recently created issues in the project to see which fields they include, which might help you identify any required fields you need to add in your next attempt to create the issue.
```

It seems so specific, doesn't it? Well, it's a problem I'm facing today, hahaha. I solved it using Jira MCP Server + Agent Skill with special instructions for each project I touch.

I didn't find any documented limitations for `namespace.description` in the OpenAI Responses API or the Chat Completions API (tool description). However, some community posts say that there is a character limit being applied to tool descriptions in the OpenAI Chat Completions API:

- https://community.openai.com/t/function-call-description-max-length/529902
- https://community.openai.com/t/character-limit-for-tools-descriptions/1153112

Looks like it was relaxed in the Responses API, where it seems to be bounded only by the model's context window.

Anyway, I believe that model providers will never include a specific field for instructions in the toolsets/namespaces concept, so I agree that we should align on `description` and allow developers to use it interchangeably for both purposes.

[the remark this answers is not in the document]
DougTrajano: @DouweM renamed to `get_description`. Ready for review.

DouweM:
@DougTrajano Can you add a `# noqa: C901` for this one please?

```
C901 `iter` is too complex (16 > 15)
   --> pydantic_ai_slim/pydantic_ai/agent/__init__.py:529:15
    |
528 |     @asynccontextmanager
529 |     async def iter(
    |               ^^^^
530 |         self,
531 |         user_prompt: str | Sequence[_messages.UserContent] | None = None,
    |
```

DougTrajano:
> @DougTrajano Can you add a `# noqa: C901` for this one please?
> 
> ```
> C901 `iter` is too complex (16 > 15)
>    --> pydantic_ai_slim/pydantic_ai/agent/__init__.py:529:15
>     |
> 528 |     @asynccontextmanager
> 529 |     async def iter(
>     |               ^^^^
> 530 |         self,
> 531 |         user_prompt: str | Sequence[_messages.UserContent] | None = None,
>     |
> ```

Done

pydantic_ai_slim/pydantic_ai/_agent_graph.py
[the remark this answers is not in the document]
DouweM: @DougTrajano Hmm, looking at https://blog.modelcontextprotocol.io/posts/2025-11-03-using-server-instructions/ the MCP server instructions are very explicitly meant to be instructions, not descriptions. And if they had MCP server descriptions, that'd be a new field. So I'm changing my mind again (but at least we're learning things!) and we should rename `get_description` back to `get_instructions` again. If we add a new "toolset description" concept in the future, that would be separate from "toolset instructions".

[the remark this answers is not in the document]
DougTrajano: All good, this is the time to reflect and learn about these things. 🤗

I will rename it to `get_instructions` again.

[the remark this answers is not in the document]
DougTrajano: Done

CostcoFanboy:
Thank you for doing this PR @DougTrajano . 

Would you kindly add this to FastMCPToolset as well?
If too much of a scope creep, I can do it in a separate PR.

Looking forward to a merge 🥳

DougTrajano:
> Thank you for doing this PR @DougTrajano .
> 
> Would you kindly add this to FastMCPToolset as well? If too much of a scope creep, I can do it in a separate PR.
> 
> Looking forward to a merge 🥳

Yeap, I spoke with @DouweM 

I will add here tonight.

pydantic_ai_slim/pydantic_ai/toolsets/fastmcp.py
```diff
@@ -100,20 +109,34 @@ def __init__(
...
         return self._id
 
+    @property
+    def instructions(self) -> str | None:
+        """Access the instructions sent by the FastMCP server during initialization."""
+        if not hasattr(self, '_instructions'):
+            raise AttributeError(
+                f'The `{self.__class__.__name__}.instructions` is only available after initialization.'
+            )
+        return self._instructions
```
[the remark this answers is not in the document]
DougTrajano: Resolved by removing `self._instructions` from `__init__` (same as `MCPToolset`).

tests/test_fastmcp.py
```diff
@@ -227,6 +227,38 @@ async def test_context_manager_nested_enter_exit(
...
+
+    async def test_get_instructions_disabled_returns_none(
+        self,
+        fastmcp_client: Client[FastMCPTransport],
+        run_context: RunContext[None],
+    ):
+        """When include_instructions is disabled, get_instructions returns None."""
+        toolset = FastMCPToolset(fastmcp_client, include_instructions=False)
+
+        assert toolset.instructions is None
```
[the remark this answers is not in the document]
DougTrajano: Unit tests updated

tests/test_mcp.py
```diff
@@ -231,12 +231,65 @@ async def process_tool_call(
...
+        instructions = await server.get_instructions(run_context)
+        assert instructions is None
+
+
+async def test_server_instructions_enabled(run_context: RunContext[int]):
+    """Test that server instructions are returned when include_instructions=True."""
+    server = MCPServerStdio('python', ['-m', 'tests.mcp_server'], include_instructions=True)
+    async with server:
+        instructions = await server.get_instructions(run_context)
+        assert instructions == 'Be a helpful assistant.'
```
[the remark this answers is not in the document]
DougTrajano: Added it in `test_server_instructions_included_in_agent_request()`

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -256,6 +256,9 @@ async def run(  # noqa: C901
             elif isinstance(last_message, _messages.ModelResponse):
                 if self.user_prompt is None:
                     run_context = build_run_context(ctx)
+                    # Resolve dynamic toolsets before fetching instructions, so DynamicToolset.get_instructions()
+                    # can delegate to the resolved inner toolset instead of returning None
+                    ctx.deps.tool_manager = await ctx.deps.tool_manager.for_run_step(run_context)
                     instructions = await ctx.deps.get_instructions(run_context)
                     if not instructions:
                         # If there's no new prompt or instructions, skip ModelRequestNode and go directly to CallToolsNode
```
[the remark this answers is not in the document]
DougTrajano: Fixed.

pydantic_ai_slim/pydantic_ai/toolsets/wrapper.py
```diff
@@ -34,6 +34,14 @@ async def __aenter__(self) -> Self:
     async def __aexit__(self, *args: Any) -> bool | None:
         return await self.wrapped.__aexit__(*args)
 
+    async def get_instructions(self, ctx: RunContext[AgentDepsT]) -> str | list[str] | None:
+        """Delegate instructions to the wrapped toolset.
+
+        This explicit delegation ensures type safety and proper propagation of the
+        instructions from wrapped toolsets to the agent's system prompt.
+        """
+        return await self.wrapped.get_instructions(ctx)
```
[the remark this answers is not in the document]
DougTrajano: Good catch. We intentionally inherit instruction delegation from WrapperToolset, so durable wrappers already pass through get_instructions. I’m adding regression coverage in Temporal/DBOS to lock this in. If we later need instruction serialization/caching across activities, we can add an explicit durable-specific path in a follow-up.

tests/test_toolsets.py
```diff
@@ -894,6 +1035,339 @@ def other_tools(ctx: RunContext[None]) -> FunctionToolset[None]:
...
+
+    agent = Agent(TestModel(custom_output_text='done'), toolsets=[DynamicToolset(make_toolset)])
+    result = await agent.run(message_history=[ModelResponse(parts=[TextPart(content='previous')])])
+
+    # The resume pre-check and request preparation should use the same run_step.
+    assert run_steps == [1]
+
+    requests = [m for m in result.all_messages() if isinstance(m, ModelRequest)]
+    assert requests[-1].instructions == 'Dynamic instructions at step 1.'
+
```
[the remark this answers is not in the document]
DougTrajano: Added unit test for it

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -744,6 +772,13 @@ async def _prepare_request(
         # This will raise errors for any tool name conflicts
         ctx.deps.tool_manager = await ctx.deps.tool_manager.for_run_step(run_context)
 
+        # Fetch instructions now that dynamic toolsets have been resolved by for_run_step
+        self.request.instructions = await _get_instructions(ctx, run_context)
```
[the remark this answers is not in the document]
DouweM: Re: double `_get_instructions` evaluation in the resume-without-prompt flow (line 253 and 776):

This was already present in the PR design before the merge — the first call in `UserPromptNode` is a gate check (should we even proceed to `ModelRequestNode`?), and the second in `_prepare_request` sets the definitive value.

Since `for_run_step` was already called in `UserPromptNode` (line 247) and `ToolManager.for_run_step` returns self when the step has not changed, the toolset state is identical, so results are the same. Instructions functions should be idempotent, so this is a minor performance cost, not a correctness issue.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -242,10 +242,17 @@ async def run(  # noqa: C901
                         ctx.deps.prompt = combined_content
             elif isinstance(last_message, _messages.ModelResponse):
                 if self.user_prompt is None:
-                    run_context = build_run_context(ctx)
-                    instructions = await ctx.deps.get_instructions(run_context)
+                    # Align with the upcoming request step so we don't resolve dynamic toolsets twice.
+                    run_context = replace(build_run_context(ctx), run_step=ctx.state.run_step + 1)
+                    ctx.deps.tool_manager = await ctx.deps.tool_manager.for_run_step(run_context)
+                    if last_message.tool_calls:
```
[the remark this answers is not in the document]
DouweM: The bot suggests the closure could reference ctx.deps.tool_manager.toolset at call time, but the closure only receives run_context: RunContext — it does not have access to ctx.deps (which is GraphAgentDeps, internal to the graph execution).

The split exists because toolset instructions must come from the current post-for_run_step toolset (which lives on ctx.deps.tool_manager), while base/capability instructions are stable across steps (resolved once from the closure captures). Consolidating would require either passing tool_manager into the closure somehow, or moving all instruction assembly into _agent_graph.py.

Happy to restructure if you prefer a different approach — the current split keeps the closure simple and matches how capability instructions flow through the same ctx.deps.get_instructions path.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -273,9 +279,6 @@ async def run(  # noqa: C901
 
         next_message.instructions = instructions
```
DouweM: Does this still make sense here if we add instructions in `ModelRequestNode`?

pydantic_ai_slim/pydantic_ai/durable_exec/dbos/_mcp.py
```diff
@@ -89,6 +102,14 @@ async def get_tools(self, ctx: RunContext[AgentDepsT]) -> dict[str, ToolsetTool[
         tool_defs = await self._dbos_wrapped_get_tools_step(ctx)
         return {name: self.tool_for_tool_def(tool_def) for name, tool_def in tool_defs.items()}
 
+    async def get_instructions(self, ctx: RunContext[AgentDepsT]) -> str | list[str] | None:
+        # When instructions are disabled, base delegation returns `None` directly.
+        # If enabled but not yet initialized in this process, fetch inside a DBOS step.
+        try:
+            return await super().get_instructions(ctx)
+        except AttributeError:
+            return await self._dbos_wrapped_get_instructions_step(ctx)
```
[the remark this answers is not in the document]
DouweM: Sounds like the `AttributeError` should be handled inside the MCP server.

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_mcp.py
```diff
@@ -85,7 +97,30 @@ def tool_for_tool_def(self, tool_def: ToolDefinition) -> ToolsetTool[AgentDepsT]
...
+        if not workflow.in_workflow():  # pragma: no cover
+            return await super().get_instructions(ctx)
+
+        # When instructions are disabled, base delegation returns `None` directly.
+        # If enabled but not yet initialized in the workflow process, we must fetch
+        # instructions inside an activity where the wrapped MCP toolset can be entered.
+        try:
+            return await super().get_instructions(ctx)
+        except AttributeError:
+            pass
```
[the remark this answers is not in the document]
DouweM: Seems like we should handle this inside the MCP server itself, or just always use the activity

docs/toolsets.md
```diff
@@ -114,6 +114,134 @@ print([t.name for t in test_model.last_model_request_parameters.function_tools])
 
 _(This example is complete, it can be run "as is")_
 
+### Toolset Instructions
+
+A [`FunctionToolset`][pydantic_ai.toolsets.FunctionToolset] can provide instructions that are automatically included in the model request whenever the toolset is used. This lets each toolset carry its own usage guidance alongside its tools, so you don't need to duplicate instructions on every agent that uses the toolset.
```
DouweM: "whenever the toolset is used" implies it has something to do with the tools being used, which is not the case. So drop iths part.

DouweM (changes requested):

PR #4131, opened 2026-01-28
Handle non-primitive types in `BaseModel` in `format_as_xml`

### LINKED ISSUE #4130: Allow format_as_xml to handle non primitive types
### Description

I've noticed that when using the format_as_xml with pydantic `BaseModel`, it fails on non primitive types even when pydantic normally handles these well. Right now I am defining a field serializer in the model to get around this issue. 

There should be a way to pass which serialization mode to use when serializing the model. 
### References

_No response_

bohdanhr:
- Closes #4130 
- Adds support for Decimal and UUID.
### Pre-Review Checklist

### Pre-Merge Checklist


pydantic_ai_slim/pydantic_ai/format_prompt.py
```diff
@@ -125,7 +154,7 @@ def _to_xml(self, value: Any, path: str, tag: str | None = None) -> ElementTree.
                 element.tag = value.__class__.__name__
             # by dumping the model we loose all metadata in nested data structures,
             # but we have collected it when called _init_structure_info
-            self._mapping_to_xml(element, value.model_dump(), path)
+            self._mapping_to_xml(element, value.model_dump(mode=self.mode), path)
```
DouweM: Would it make sense to always use JSON mode here, as we know values will need to be serialized as text for XML? Or is there a specific reason you think it should be configurable?

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/format_prompt.py
bohdanhr: I'll make the change to just default to `json` since I dont think there would be a world where we would want to dump to `python`. I've mostly done it this way so we can keep the functionality backwards compatible since the default mode is `python`. I suppose the only thing that comes to mind is that it would attempt to serialize an unserializable value here and raise a different error. We can catch that with a try/ catch and reraise a TypeError to match existing behavior.

DouweM: I'm ok always using json as it seems like a strict improvement

pydantic_ai_slim/pydantic_ai/format_prompt.py
```diff
@@ -98,40 +99,54 @@ def to_xml(self, tag: str | None = None) -> ElementTree.Element:
...
             # by dumping the model we loose all metadata in nested data structures,
             # but we have collected it when called _init_structure_info
-            self._mapping_to_xml(element, value.model_dump(), path)
-        elif isinstance(value, Iterable):
+            try:
+                mapping = value.model_dump(mode='json')
+            except PydanticSerializationError as e:
+                raise TypeError(f'Unsupported type for XML formatting: {e}') from e
+            self._mapping_to_xml(element, mapping, path)
+            return element
```
bohdanhr, line 128: actual change.

pydantic_ai_slim/pydantic_ai/format_prompt.py
```diff
@@ -98,40 +99,54 @@ def to_xml(self, tag: str | None = None) -> ElementTree.Element:
...
+        elif isinstance(value, str):
+            element.text = value
+        elif isinstance(value, bytes | bytearray):
+            element.text = value.decode(errors='ignore')
+        elif isinstance(value, bool | int | float | Enum):
+            element.text = str(value)
+        elif isinstance(value, date | time):
+            element.text = value.isoformat()
+        elif isinstance(value, timedelta):
+            element.text = str(value)
```
bohdanhr, line 148: had to reformat a bit to reduce complexity, [C901](https://docs.astral.sh/ruff/rules/complex-structure/)

pydantic_ai_slim/pydantic_ai/format_prompt.py
```diff
@@ -98,40 +99,54 @@ def to_xml(self, tag: str | None = None) -> ElementTree.Element:
...
+
+    def _set_scalar_text(self, element: ElementTree.Element, value: Any) -> bool:
+        """Set element.text for scalar types. Return True if handled, False otherwise."""
+        if value is None:
+            element.text = self.none_str
+        elif isinstance(value, str):
+            element.text = value
+        elif isinstance(value, bytes | bytearray):
+            element.text = value.decode(errors='ignore')
+        elif isinstance(value, bool | int | float | Enum):
```
DouweM, line 143: @bohdanhr Should we list `Decimal` here as well?

bohdanhr: I can. Probably a good idea to be explicit. Thoughts on then also adding UUID? I think that will cover all scalar types.

pydantic_ai_slim/pydantic_ai/format_prompt.py
```diff
@@ -98,40 +101,58 @@ def to_xml(self, tag: str | None = None) -> ElementTree.Element:
...
+        if isinstance(value, BaseModel):
             self._init_structure_info()
             if tag is None:
                 element.tag = value.__class__.__name__
             # by dumping the model we loose all metadata in nested data structures,
             # but we have collected it when called _init_structure_info
-            self._mapping_to_xml(element, value.model_dump(), path)
-        elif isinstance(value, Iterable):
+            try:
+                mapping = value.model_dump(mode='json')
```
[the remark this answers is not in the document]
bohdanhr: https://github.com/pydantic/pydantic-ai/blob/0b06f0db2564adc95da9f722b1f0bb8befd1ef7e/tests/test_format_as_xml.py#L60

this behavior hasn't actually changed, we had testing for this in the past.

DouweM:
@bohdanhr Thanks!

PR #4152, opened 2026-01-30
Fix Bedrock caching when user message ended on non-text content

DouweM:
Replaces https://github.com/pydantic/pydantic-ai/pull/4122

Closes https://github.com/pydantic/pydantic-ai/issues/4121


matthgt:
@DouweM thank you for getting this over the line!

Issue #4116, opened 2026-01-27
Enable native structured output and streaming tool calls arguments with Ollama by using API directly


rgbkrk:
### Initial Checks

- [x] I'm using the [latest version](https://github.com/pydantic/pydantic-ai/releases/latest) of Pydantic AI
- [x] I've searched for my issue in [the issue tracker](https://github.com/pydantic/pydantic-ai/issues) before opening this issue

### Description

Buried within https://github.com/pydantic/pydantic-ai/issues/242 is a desire to tap into a "native" feature of Ollama for structured outputs. As the issue was closed due to more broad structured outputs approach (https://github.com/pydantic/pydantic-ai/issues/582), the details on adapting to the Ollama API are lost in the mix of that issue. I'm opening this one to motivate and track the work for bringing Ollama in natively. These are the specific features people want unlocked, including myself:

* [x] Structured Outputs
* [ ] Streaming Tool Call Arguments

In the past, a custom httpx client was needed (see https://github.com/ollama/ollama-python/pull/380 in `ollama-python`). As the PR is stale, I'd like to take a fresh look at it to see if it can be adapted as a provider in Pydantic AI again.

Related PRs:

* [ ] https://github.com/ollama/ollama-python/pull/618

### Python, Pydantic AI & LLM client version

- Python: 3.12
- Pydantic AI: 1.47.0
- LLM provider SDK: Ollama

DouweM:
@rgbkrk Do we need the native Ollama SDK/API for this? They don't support structured output or streaming tool call args through the Chat Completions API?

rgbkrk:
Bitten by the "can I reproduce it bug" I'll have to showcase it. While I was working on the jupyter pydantic support I noticed that OpenAI streaming tool calls args worked with OpenAI but _not_ with Groq or Ollama. The native [Ollama endpoint supports streaming tool calls](https://ollama.com/blog/streaming-tool).

Goldokpa:
Hi @DouweM and @rgbkrk , i encountered this same issue as well while cooking , the OpenAI streaming too call did not work with Ollama, would you be happy if i made a PR on this and you guide me ..

Goldokpa:
Hi @DouweM could you please go through this

adrianabasbous:
Hello, is there any foreseen timeline for implementing native structured output with Ollama?
thank you

rgbkrk:
[...]

PR #4160, opened 2026-02-01
Add `OllamaModel` subclass and correct Ollama capability flags

### LINKED ISSUE #4917: OllamaProvider incorrectly advertises JSON schema support via OpenAI API leading to silent structured output failure
### Initial Checks

### Description

```python
return OpenAIModelProfile(
).update(profile)
```

```json
"response_format": {
  "type": "json_schema",
  "json_schema": { ... }
}
```

```json
"format": {
  "type": "json",
  "schema": { ... }
}
```

As a result:

* Validation fails repeatedly

```python
from pydantic_ai import Agent

class Character(BaseModel):
    name: str
    role: str

class Output(BaseModel):
    characters: list[Character]

agent = Agent(
    model="qwen3.5",
    retries=3,
)

print(result)
```

### Logfire Trace

_No response_

- Python: 3.13.12
- PydanticAI: 1.73.0
- Pydantic: 2.12.5

Goldokpa:
## What
Add `OllamaModel` subclass and correct Ollama capability flags.

## Why
- `OllamaModel` gives users a first-class import (`from pydantic_ai.models.ollama import OllamaModel`) and docs page, consistent with `CerebrasModel`/`OpenRouterModel`.
- `openai_supports_strict_tool_definition=False` fixes tool-calling on Ollama (which silently drops the `strict` field).
- Ollama Cloud's `/v1/chat/completions` accepts `response_format: json_schema` but does not apply grammar-constrained decoding ([ollama/ollama#12362](https://github.com/ollama/ollama/issues/12362)), causing `NativeOutput` to silently fail in a retry loop (#4917). The fix auto-detects Cloud paths and disables the flag.

## Changes
- **`models/ollama.py`**: `OllamaModel(OpenAIChatModel)` with Cloud detection in `__init__`
- **`models/__init__.py`**: route `ollama:` → `OllamaModel` in `infer_model`
- **`providers/ollama.py`**: add `openai_supports_strict_tool_definition=False`, move `model_name.lower()` out of loop
- **`tests/models/test_ollama.py`**: 7 unit tests + 3 VCR cassettes (local + Cloud)
- **`docs/models/ollama.md`**: dedicated docs page with structured output caveats
- **`docs/api/models/ollama.md`** + `mkdocs.yml` nav entries

Closes #4917. Partially addresses #4116.

- [ ] AI generated code

### DIFF
--- a/docs/models/ollama.md
+++ b/docs/models/ollama.md
@@ -14,7 +14,7 @@ Pydantic AI supports both self-hosted [Ollama](https://ollama.com/) servers (run
 
 For servers running locally, use the `http://localhost:11434/v1` base URL. For Ollama Cloud, use `https://ollama.com/v1` and ensure an API key is set.
 
-[`OllamaModel`][pydantic_ai.models.ollama.OllamaModel] uses Ollama's OpenAI-compatible Chat Completions API (`/v1/chat/completions`), so existing Ollama users can start using Pydantic AI without any migration.
+For backward compatibility, [`OllamaModel`][pydantic_ai.models.ollama.OllamaModel] uses Ollama's OpenAI-compatible Chat Completions API (`/v1/chat/completions`).
 
 ## Environment variable
 
@@ -95,7 +95,3 @@ agent = Agent(model, output_type=NativeOutput(CityLocation))
     When [`OllamaModel`][pydantic_ai.models.ollama.OllamaModel] detects a Cloud path — either a `base_url` on `ollama.com` or a model name ending in `-cloud` — it automatically disables `supports_json_schema_output` on the profile.
 
     If you use [`NativeOutput`][pydantic_ai.output.NativeOutput] with an Ollama Cloud model, you'll get a clear [`UserError`][pydantic_ai.exceptions.UserError] instead of a silent retry loop. Use the default [`ToolOutput`][pydantic_ai.output.ToolOutput] or [`PromptedOutput`][pydantic_ai.output.PromptedOutput] instead — both work on Cloud.
-
-## Tool calling
-
-Tool calling works on both self-hosted and Cloud Ollama via the inherited OpenAI Chat Completions API path. Ollama does not support OpenAI's `strict` mode for tool schemas, so [`openai_supports_strict_tool_definition`][pydantic_ai.profiles.openai.OpenAIModelProfile.openai_supports_strict_tool_definition] is set to `False` in the profile.


DouweM:
> ### Streaming tool calls
> * No code change: existing `_map_tool_call_delta` and OpenAI-compatible endpoint are used; Ollama's `/v1/chat/completions` supports tools and streaming.

@Goldokpa https://github.com/pydantic/pydantic-ai/issues/4116 suggests that streaming tool calls were not working with Ollama previously. Have you confirmed it works?

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -640,6 +640,14 @@ async def _completions_create(
         for setting in unsupported_model_settings:
             model_settings.pop(setting, None)
 
+        extra_body: dict[str, Any] | None = model_settings.get('extra_body')
+        if self.system == 'ollama' and response_format is not None and isinstance(response_format, dict):
```
DouweM: Instead of directly checking the system, can we have an `OllamaModel` as a subclass of `OpenAIChatModel`, as we already have `OpenRouterModel` and `CerebrasModel`?

pydantic_ai_slim/pydantic_ai/providers/ollama.py
```diff
@@ -61,12 +62,16 @@ def model_profile(self, model_name: str) -> ModelProfile | None:
...
+        # we need to maintain that behavior unless json_schema_transformer is set explicitly.
+        # Ollama's native structured output uses `format` with a raw JSON schema; strict mode is not supported (issue #4116).
+        # Ensure native structured output is supported regardless of the model-specific profile default.
+        base = OpenAIModelProfile(
             json_schema_transformer=OpenAIJsonSchemaTransformer,
             openai_chat_thinking_field='reasoning',
             openai_chat_send_back_thinking_parts='tags',
+            openai_supports_strict_tool_definition=False,
         ).update(profile)
+        return replace(base, supports_json_schema_output=True)
```
DouweM: We can support `json_object_output` as well by setting `format="json"`

DouweM: Why replace instead of adding it to `base`?

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -640,6 +640,14 @@ async def _completions_create(
...
+            if response_format.get('type') == 'json_schema' and 'json_schema' in response_format:
+                js = response_format['json_schema']
```
DouweM: Should we be resetting the `response_format` var as otherwise the schema is added to the request in 2 places, and Ollama may trip on the `response_format`?

DouweM: This should probably be part of the `model_request_parameters.output_mode` branching logic above, so we don't set the data in 2 vars.

If we extract this into a subclass, we'll need a new hook where we can override how the response data is added to the request.

scripts/verify_ollama_structured_and_tools.py
DouweM, line 1: No verification script please, can you add a test that validates the args sent to the API?

Goldokpa: yes sure

Goldokpa:
@DouweM Thanks for the review. Here’s what I changed:

**1. Verification script** — Removed `scripts/verify_ollama_structured_and_tools.py`. Added tests that assert the args sent to the API:
- `test_ollama_native_output_sends_format_in_extra_body`: with `OllamaModel` + `NativeOutput`, asserts `extra_body['format']` contains the raw JSON schema and `response_format` is not sent.
- `test_ollama_prompted_json_object_sends_format_json_in_extra_body`: with `OllamaModel` + `PromptedOutput`, asserts `extra_body['format'] == 'json'` and `response_format` is not sent.

**2. OllamaModel subclass** — Introduced `OllamaModel` (subclass of `OpenAIChatModel`) in `pydantic_ai_slim/pydantic_ai/models/ollama.py`, following the same pattern as `OpenRouterModel` and `CerebrasModel`. The factory in `models/__init__.py` now returns `OllamaModel` for `model_kind == 'ollama'`. Removed the `self.system == 'ollama'` check from `openai.py`.

**3. Hook for response format / extra_body** — In `OpenAIChatModel` added `_prepare_response_format_and_extra_body(response_format, extra_body, model_request_parameters)` that by default returns `(response_format, extra_body)`. `OllamaModel` overrides it: for native json_schema it sets `extra_body['format'] = schema` and returns `(None, extra_body)` so the schema is only sent via `format`, not `response_format`; for prompted json_object it sets `extra_body['format'] = 'json'` and returns `(None, extra_body)`.

**4. Ollama provider profile** — Dropped `replace()` and set `supports_json_schema_output=True` and `supports_json_object_output=True` on the base `OpenAIModelProfile(...)` in `ollama.py` so both native and prompted json_object are supported.

All 160 tests in `tests/providers/test_ollama.py` and `tests/models/test_openai.py` pass.

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -750,6 +757,19 @@ def _validate_completion(self, response: chat.ChatCompletion) -> chat.ChatComple
...
+        *,
+        extra_body: dict[str, Any] | None,
+        response_format: chat.completion_create_params.ResponseFormat | None,
+        model_request_parameters: ModelRequestParameters,
+    ) -> tuple[dict[str, Any] | None, chat.completion_create_params.ResponseFormat | None]:
+        """Hook for subclasses to customize the request payload.
+
+        This is intended for OpenAI-compatible providers that need to map OpenAI fields into provider-specific fields.
+        """
+        return extra_body, response_format
```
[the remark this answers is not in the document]
DouweM: Keeping it makes sense if that's the pattern for other hook methods

tests/models/test_openai.py
```diff
@@ -2401,6 +2401,73 @@ class CityLocation(BaseModel):
...
+    @agent.tool_plain
+    async def get_weather(city: str) -> str:
+        return 'sunny'
+
+    async with agent.run_stream('What is the weather in Paris?'):
+        pass
+
+    kwargs = get_mock_chat_completion_kwargs(mock_client)[-1]
+    assert kwargs['stream_options'] == {'include_usage': True}
+    assert 'tools' in kwargs
```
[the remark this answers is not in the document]
Goldokpa: > @DouweM Yes, I've confirmed streaming tool calls work with Ollama. Tested against live Ollama v0.15.6 with qwen3:0.6b:
>
> - Streaming tool calls: Working. The existing _map_tool_call_delta logic handles Ollama's /v1/chat/completions tool call deltas correctly with no code changes needed.
> - Non-streaming tool calls: Also working.
> - Native structured output: Working. I found and fixed a bug in my original implementation — Ollama's OpenAI-compat endpoint (/v1/chat/completions) supports standard response_format with json_schema natively. The format top-level field I was mapping to via extra_body only works on the native /api/chat endpoint and is ignored by the compat endpoint. I've removed the _customize_request_payload override so the standard response_format flow inherited from OpenAIChatModel is used instead. Tests updated to match.
>
> The _customize_request_payload hook on OpenAIChatModel is kept as a clean extension point for other providers.

Goldokpa:
@DouweM Yes, I've confirmed streaming tool calls work with Ollama. Tested against live Ollama v0.15.6 with `qwen3:0.6b`:

- **Streaming tool calls**: Working. The existing `_map_tool_call_delta` logic handles Ollama's `/v1/chat/completions` tool call deltas correctly with no code changes needed.
- **Non-streaming tool calls**: Also working.
- **Native structured output**: Working. I found and fixed a bug in my original implementation — Ollama's OpenAI-compat endpoint (`/v1/chat/completions`) supports standard `response_format` with `json_schema` natively. The `format` top-level field I was mapping to via `extra_body` only works on the native `/api/chat` endpoint and is ignored by the compat endpoint. I've removed the `_customize_request_payload` override so the standard `response_format` flow inherited from `OpenAIChatModel` is used instead. Tests updated to match.

The `_customize_request_payload` hook on `OpenAIChatModel` is kept as a clean extension point for other providers.

pydantic_ai_slim/pydantic_ai/models/openai.py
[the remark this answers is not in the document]
DouweM: Yep let's drop it for now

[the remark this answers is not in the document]
Goldokpa: Ok sound @DouweM i'll do just that then. Thank you

Goldokpa:
Hi @DouweM how are you doing mate? could you please assist on reviewing the PR

Goldokpa:
Hey @DouweM 👋.      just checking in  all review comments have been addressed and CI is green. Could you take another look when you get a chance? Thanks!

DouweM:
@Goldokpa Thanks for all your work & patience here! The PR is in a good state so I'll take it over from here and will get it out in the next few days, so that I don't keep us going in a never ending (auto-)code/review loop!

DouweM:
I rescoped the PR after investigating how Ollama's OpenAI-compatible `/v1/chat/completions` endpoint actually behaves for `response_format: {type: "json_schema", ...}`:

- **Self-hosted Ollama** (v0.5.0+): works as intended. The source code plumbs `json_schema` → `llama.cpp`'s grammar-constrained decoder. Empirically verified against local `qwen3:0.6b` — even an adversarial "write me a poem" prompt produces structurally valid JSON.
- **Ollama Cloud**: accepts the `json_schema` payload without error but **does not enforce it**. Empirically verified against `gpt-oss:20b`, `gpt-oss:120b`, and `deepseek-v3.1:671b` — all fail or silently ignore the schema. Ollama maintainer `ParthSareen` acknowledged this in [ollama/ollama#12362](https://github.com/ollama/ollama/issues/12362): *"we don't currently support structured outputs on the cloud just yet."*

The new behavior: `OllamaModel` detects Cloud paths (`ollama.com` in `base_url` or model name ending in `-cloud`) and disables `supports_json_schema_output` on the profile. Users who try `NativeOutput` on Cloud get a clear `UserError` instead of the silent retry loops reported in #4917. `PromptedOutput` and the default `ToolOutput` both work on Cloud.

This PR now closes #4917 and provides the dedicated `OllamaModel` subclass requested in #4116 (full native `/api/chat` support is a separate, larger effort).

docs/models/ollama.md
```diff
@@ -0,0 +1,101 @@
...
+pip/uv-add "pydantic-ai-slim[openai]"
+```
+
+## Configuration
+
+Pydantic AI supports both self-hosted [Ollama](https://ollama.com/) servers (running locally or remotely) and [Ollama Cloud](https://ollama.com/cloud).
+
+For servers running locally, use the `http://localhost:11434/v1` base URL. For Ollama Cloud, use `https://ollama.com/v1` and ensure an API key is set.
+
+[`OllamaModel`][pydantic_ai.models.ollama.OllamaModel] uses Ollama's OpenAI-compatible Chat Completions API (`/v1/chat/completions`), so existing Ollama users can start using Pydantic AI without any migration.
```
DouweM: Call it "backward compatibility".

docs/models/ollama.md
```diff
@@ -0,0 +1,101 @@
...
+
+    When [`OllamaModel`][pydantic_ai.models.ollama.OllamaModel] detects a Cloud path — either a `base_url` on `ollama.com` or a model name ending in `-cloud` — it automatically disables `supports_json_schema_output` on the profile.
+
+    If you use [`NativeOutput`][pydantic_ai.output.NativeOutput] with an Ollama Cloud model, you'll get a clear [`UserError`][pydantic_ai.exceptions.UserError] instead of a silent retry loop. Use the default [`ToolOutput`][pydantic_ai.output.ToolOutput] or [`PromptedOutput`][pydantic_ai.output.PromptedOutput] instead — both work on Cloud.
+
+## Tool calling
+
+Tool calling works on both self-hosted and Cloud Ollama via the inherited OpenAI Chat Completions API path. Ollama does not support OpenAI's `strict` mode for tool schemas, so [`openai_supports_strict_tool_definition`][pydantic_ai.profiles.openai.OpenAIModelProfile.openai_supports_strict_tool_definition] is set to `False` in the profile.
```
DouweM: This is not worth explicitly documenting.

PR #4165, opened 2026-02-02
feat: Add `XSearchTool` and `FileSearch` support for xAI

### LINKED ISSUE #3896: Add XSearchTool built-in tool for X/Twitter search via Grok provider
### Description
Add `XSearchTool` as a new built-in tool that enables Grok-powered agents to search X/Twitter posts, user profiles, and threads natively. This is a unique capability only available through X AI's Grok models.

### Motivation
X AI's Grok models offer the only LLM provider with native access to real-time X/Twitter data. This enables use cases not possible with any other provider:

- **Real-time social media search** - Search posts, threads, and user content on X
- **Date-range filtering** - Query posts within specific time windows
- **Handle-based filtering** - Include or exclude specific X accounts
- **Multimedia understanding** - Analyze images and videos embedded in posts

### Proposed code

```python
from pydantic_ai import Agent, XSearchTool

agent = Agent(
    'grok:grok-4-1-fast',
    builtin_tools=[
        XSearchTool(
            allowed_x_handles=['pydantic'],  # Max 10 handles
            from_date='2025-01-01',
            to_date='2025-06-01',
            enable_image_understanding=True,
            enable_video_understanding=True,
        )
    ]
)

result = agent.run_sync('What are the latest Pydantic AI updates on X?')
print(result.output)
```

### Provider Support

Only XAI/Grok

### References

- [X AI Search Tools Documentation](https://docs.x.ai/docs/guides/tools/search-tools)
- [Existing Grok provider](https://github.com/pydantic/pydantic-ai/blob/main/pydantic_ai_slim/pydantic_ai/providers/grok.py)
- [Built-in tools pattern](https://ai.pydantic.dev/builtin-tools/)

### References

_No response_

colesmcintosh:
## Summary

- Add `XSearchTool` as a new built-in tool for searching X/Twitter posts with xAI models
  - Handle filtering (`allowed_x_handles`/`excluded_x_handles`, max 10 each, mutually exclusive)
  - Date range filtering (`from_date`/`to_date`)
  - Media analysis (`enable_image_understanding`/`enable_video_understanding`)
  - Raw search results opt-in via `include_x_search_output` (tool field) or `XaiModelSettings.xai_include_x_search_output` (model setting)
- Add `XSearch` capability in `models/xai.py` following the `BuiltinOrLocalTool` pattern
- Map `FileSearchTool` to xAI's collections search API (documented via `docs/builtin-tools.md`)
- Preserve xAI SDK function names across message round-trips via `BuiltinToolCallPart.provider_details['function_name']` (xAI sends `x_keyword_search`/`x_semantic_search` etc., not `x_search`)
- Treat grok-4 reasoning models as not supporting the unified `thinking` setting — they always reason but the API rejects `reasoning_effort`, so forwarding the parameter would produce `INVALID_ARGUMENT`
- Add `builtin_tools/AGENTS.md` with guidelines for adding new builtin tools

Closes #3896

### DIFF
--- a/.github/workflows/gateway-model-health.yml
+++ b/.github/workflows/gateway-model-health.yml
@@ -39,3 +39,28 @@ jobs:
 
       - name: Run gateway model catalog checks
         run: uv run pytest tests/providers/test_gateway_catalog.py -q --run-gateway-live
+
+      - name: Notify Slack on failure
+        if: failure() && !cancelled()
+        uses: slackapi/slack-github-action@af78098f536edbc4de71162a307590698245be95 # v3.0.1
+        with:
+          errors: true
+          webhook: ${{ secrets.SLACK_WEBHOOK_URL }}
+          webhook-type: incoming-webhook
+          payload: |
+            text: ":x: Gateway model health failed in ${{ github.repository }} on ${{ github.ref_name }}."
+            blocks:
+              - type: section
+                text:
+                  type: mrkdwn
+                  text: ":x: *Gateway model health failed*"
+              - type: section
+                fields:
+                  - type: mrkdwn
+                    text: "*Repository*\n<${{ github.server_url }}/${{ github.repository }}|${{ github.repository }}>"
+                  - type: mrkdwn
+                    text: "*Trigger*\n${{ github.event_name }}"
+                  - type: mrkdwn
+                    text: "*Ref*\n`${{ github.ref_name }}`"
+                  - type: mrkdwn
+                    text: "*Run*\n<${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}|View workflow run>"
--- a/docs/durable_execution/overview.md
+++ b/docs/durable_execution/overview.md
@@ -2,10 +2,11 @@
 
 Pydantic AI allows you to build durable agents that can preserve their progress across transient API failures and application errors or restarts, and handle long-running, asynchronous, and human-in-the-loop workflows with production-grade reliability. Durable agents have full support for [streaming](../agent.md#streaming-all-events) and [MCP](../mcp/client.md), with the added benefit of fault tolerance.
 
-Pydantic AI natively supports three durable execution solutions:
+Pydantic AI officially supports four durable execution solutions:
 
 - [Temporal](./temporal.md)
 - [DBOS](./dbos.md)
 - [Prefect](./prefect.md)
+- [Restate](./restate.md)
 
-These integrations only use Pydantic AI's public interface, so they also serve as a reference for integrating with other durable systems.
+These integrations are co-maintained by the Pydantic and vendor teams and only use Pydantic AI's public interface, so they also serve as a reference for integrating with other durable systems.
--- a/docs/durable_execution/restate.md
+++ b/docs/durable_execution/restate.md
@@ -1,0 +1,113 @@
+# Durable Execution with Restate
+
+[Restate](https://restate.dev) is a lightweight durable execution runtime with first-class support for AI agents. The Pydantic AI integration is provided via the [Restate Python SDK](https://github.com/restatedev/sdk-python/tree/main/python/restate/ext/pydantic).
+
+Visit the [Restate documentation](https://docs.restate.dev/ai/patterns/durable-agents) for more information.
+
+## Durable Execution
+
+Restate makes your agent **durable** by recording every step of its execution in a journal. If your process crashes mid-execution, Restate replays the journal, skips completed steps, and resumes from exactly where it left off.
+
+Your agent runs in a regular HTTP handler inside a Restate **service**. The Restate Server sits in front of your application and manages orchestration, journaling, and retries. Services run like regular Docker containers or serverless functions.
+
+A durable agent has three building blocks:
+
+1. The **handler**: your agent logic, exposed as an HTTP endpoint in a Restate service.
+2. **LLM calls**: persisted so responses are not re-fetched on recovery — saving cost and time.
+3. **Tool executions**: wrapped in durable steps so side effects are not duplicated.
+
+```text
+                  Clients
+              (HTTP, Kafka, etc.)
+                     |
+                     v
+            +---------------------+
+            |   Restate Server    |      (Journals execution,
+            +---------------------+       retries on failure,
+                     ^                    manages state)
+                     |
+        Journal      |   Replay on
+        steps,       |   recovery,
+        retries      |   schedule calls
+                     v
++------------------------------------------------------+
+|               Application Process                    |
+|   +----------------------------------------------+   |
+|   |         Restate Service Handler              |   |
+|   |           (Agent Run Loop)                   |   |
+|   |    [ Durable Steps (Tool, MCP, Model) ]      |   |
+|   +----------------------------------------------+   |
+|         |           |                |               |
++------------------------------------------------------+
+          |           |                |
+          v           v                v
+      [External APIs, services, databases, etc.]
+```
+
+See the [Restate documentation](https://docs.restate.dev/ai/patterns/durable-agents) for more information.
+
+## Durable Agent
+
+Any Pydantic AI agent can be made durable by wrapping it with `RestateAgent` from the Restate SDK and running it inside a Restate service handler.
+
+Install the Restate SDK:
+
+```bash
+pip/uv-add pydantic-ai restate_sdk[serde]
+```
+
+Here is a complete example of a durable Pydantic AI agent with Restate:
+
+```python {title="restate_agent.py" test="skip" lint="skip"}
+import restate
+from pydantic_ai import Agent, RunContext
+from restate.ext.pydantic import RestateAgent, restate_context
+
+weather_agent = Agent(  # (1)!
+    'openai:gpt-5.2',
+    system_prompt='You are a helpful agent that provides weather updates.',
+)
+
+
+@weather_agent.tool()
+async def get_weather(_run_ctx: RunContext[None], city: str) -> dict:
+    """Get the current weather for a given city."""
+
+    # Do durable tool steps using the Restate context
+    async def call_weather_api(city: str) -> dict:
+        return {'temperature': 23, 'description': 'Sunny and warm.'}
+
+    return await restate_context().run_typed(  # (2)!
+        f'Get weather {city}', call_weather_api, city=city
+    )
+
+
+restate_agent = RestateAgent(weather_agent)  # (3)!
+
+agent_service = restate.Service('WeatherAgent')
+
+
+@agent_service.handler()
+async def run(_ctx: restate.Context, prompt: str) -> str:  # (4)!
+    result = await restate_agent.run(prompt)
+    return result.output
+
+
+app = restate.app(services=[agent_service])  # (5)!
+
+if __name__ == "__main__":  # (6)!
+    import hypercorn
+    import asyncio
+    conf = hypercorn.Config()
+    conf.bind = ["0.0.0.0:9080"]
+    asyncio.run(hypercorn.asyncio.serve(app, conf))
+```
+
+1. Define your agent and tools as you normally would with Pydantic AI.
+2. Use `restate_context()` actions inside tools to make their execution durable. The result is persisted and retried until it succeeds. Side effects won't be duplicated on recovery.
+3. `RestateAgent` wraps the agent so every LLM response is saved in the Restate Server and replayed during recovery.
+4. The Restate service handler gives the agent a durable execution context and exposes it as an HTTP endpoint.
+5. `restate.app()` creates the application that can be served.
+6. Run the application with an ASGI server like Hypercorn.
+
+See the [Restate agent quickstart](https://docs.restate.dev/ai-quickstart) to learn how to run the agent.
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -110,6 +110,7 @@ nav:
           - Temporal: durable_execution/temporal.md
           - DBOS: durable_execution/dbos.md
           - Prefect: durable_execution/prefect.md
+          - Restate: durable_execution/restate.md
       - UI Event Streams:
           - Overview: ui/overview.md
           - AG-UI: ui/ag-ui.md
--- a/pydantic_ai_slim/pydantic_ai/_function_schema.py
+++ b/pydantic_ai_slim/pydantic_ai/_function_schema.py
@@ -31,7 +31,6 @@ from ._utils import (
     run_in_executor,
     takes_run_context,
 )
-from .function_signature import FunctionSignature
 from .messages import ToolReturn
 
 if TYPE_CHECKING:
@@ -58,8 +57,6 @@ class FunctionSchema:
     var_positional_field: str | None = None
     return_schema: ObjectJsonSchema = field(default_factory=dict[str, Any])
     """JSON schema for the function's return type. At minimum `{}` (equivalent to `Any`)."""
-    function_signature: FunctionSignature | None = None
-    """Function signature shape for this function. `None` for output tools."""
 
     async def call(self, args_dict: dict[str, Any], ctx: RunContext[Any]) -> Any:
         args, kwargs = self._call_args(args_dict, ctx)
@@ -246,13 +243,6 @@ def function_schema(  # noqa: C901
         )
         return_schema = {}
 
-    # Compute function signature eagerly alongside the other schemas
-    func_sig = FunctionSignature.from_schema(
-        name=name,
-        parameters_schema=checked_json_schema,
-        return_schema=return_schema,
-    )
-
     return FunctionSchema(
         name=name,
         description=description,
@@ -265,7 +255,6 @@ def function_schema(  # noqa: C901
         is_async=is_async_callable(function),
         function=function,
         return_schema=return_schema,
-        function_signature=func_sig,
     )
 
 
--- a/pydantic_ai_slim/pydantic_ai/_utils.py
+++ b/pydantic_ai_slim/pydantic_ai/_utils.py
@@ -5,6 +5,7 @@ import copy
 import functools
 import inspect
 import re
+import sys
 import time
 import uuid
 from collections.abc import AsyncIterable, AsyncIterator, Awaitable, Callable, Iterable, Iterator
@@ -27,7 +28,13 @@ from typing import (
     overload,
 )
 
+import anyio
 from anyio.to_thread import run_sync
+
+if sys.version_info < (3, 11):
+    from exceptiongroup import BaseExceptionGroup as BaseExceptionGroup  # pragma: lax no cover
+else:
+    BaseExceptionGroup = BaseExceptionGroup  # pragma: lax no cover
 from pydantic import BaseModel, TypeAdapter
 from pydantic._internal import _decorators, _typing_extra
 from pydantic.json_schema import JsonSchemaValue
@@ -181,6 +188,32 @@ class Some(Generic[T]):
 
 Option: TypeAlias = Some[T] | None
 """Analogous to Rust's `Option` type, usage: `Option[Thing]` is equivalent to `Some[Thing] | None`."""
+
+
+async def gather(*coros: Awaitable[T]) -> list[T]:
+    """Run awaitables concurrently via an `anyio` task group and return results in input order.
+
+    Unlike `asyncio.gather`, a failure in one coroutine cancels the rest instead of leaving them
+    as orphan background tasks. If exactly one task fails, its exception is re-raised directly to
+    match `asyncio.gather`'s shape; multi-failure cases propagate as an `ExceptionGroup`.
+    """
+    results: list[T] = [None] * len(coros)  # type: ignore[list-item]
+
+    async def _run(index: int, coro: Awaitable[T]) -> None:
+        results[index] = await coro
+
+    try:
+        async with anyio.create_task_group() as tg:
+            for i, coro in enumerate(coros):
+                tg.start_soon(_run, i, coro)
+    except BaseExceptionGroup as eg:
+        if len(eg.exceptions) == 1:
+            exc = eg.exceptions[0]
+            exc.__suppress_context__ = True
+            raise exc
+        raise
+
+    return results
 
 
 class Unset:
--- a/pydantic_ai_slim/pydantic_ai/capabilities/combined.py
+++ b/pydantic_ai_slim/pydantic_ai/capabilities/combined.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import asyncio
 from collections.abc import AsyncIterable, Awaitable, Callable, Sequence
 from dataclasses import dataclass, replace
 from typing import TYPE_CHECKING, Any
@@ -9,6 +8,7 @@ from pydantic import ValidationError
 
 from pydantic_ai import _system_prompt
 from pydantic_ai._instructions import AgentInstructions, normalize_instructions
+from pydantic_ai._utils import gather
 from pydantic_ai.exceptions import ModelRetry
 from pydantic_ai.messages import AgentStreamEvent, ModelResponse, ToolCallPart
 from pydantic_ai.settings import ModelSettings, merge_model_settings
@@ -50,7 +50,7 @@ class CombinedCapability(AbstractCapability[AgentDepsT]):
         return any(c.has_wrap_run_event_stream for c in self.capabilities)
 
     async def for_run(self, ctx: RunContext[AgentDepsT]) -> AbstractCapability[AgentDepsT]:
-        new_caps = await asyncio.gather(*(c.for_run(ctx) for c in self.capabilities))
+        new_caps = await gather(*(c.for_run(ctx) for c in self.capabilities))
         if all(new is old for new, old in zip(new_caps, self.capabilities)):
             return self
         return replace(self, capabilities=list(new_caps))
--- a/pydantic_ai_slim/pydantic_ai/models/openai.py
+++ b/pydantic_ai_slim/pydantic_ai/models/openai.py
@@ -1591,6 +1591,10 @@ class OpenAIResponsesModel(Model[AsyncOpenAI]):
             usage=_map_usage(response, self._provider.name, self._provider.base_url, self.model_name),
             model_name=self._model_name,
             provider_response_id=response.id,
+            # Marks this ModelResponse as coming from the stateless `/compact` endpoint.
+            # `_get_previous_response_id_and_new_messages` uses this to break the auto-chain,
+            # since compaction response ids cannot be used as `previous_response_id`.
+            provider_details={'compaction': True},
             timestamp=_now_utc(),
             provider_name=self._provider.name,
             provider_url=self._provider.base_url,
@@ -2109,6 +2113,12 @@ class OpenAIResponsesModel(Model[AsyncOpenAI]):
         trimmed_messages: list[ModelMessage] = []
         for m in reversed(messages):
             if isinstance(m, ModelResponse) and m.provider_name == self.system:
+                # Responses from the stateless `/compact` endpoint can't be used as
+                # `previous_response_id`, so the compaction acts as a hard chain boundary:
+                # the next request must pass the `CompactionPart` via the `input` array
+                # (handled by `_map_messages`) without a `previous_response_id`.
+                if m.provider_details and m.provider_details.get('compaction'):
+                    return None, messages
                 previous_response_id = m.provider_response_id
                 break
             else:
--- a/pydantic_ai_slim/pydantic_ai/tools.py
+++ b/pydantic_ai_slim/pydantic_ai/tools.py
@@ -3,6 +3,7 @@ from __future__ import annotations as _annotations
 import inspect
 from collections.abc import Awaitable, Callable, Sequence
 from dataclasses import KW_ONLY, dataclass, field
+from functools import cached_property
 from typing import Annotated, Any, Concatenate, Generic, Literal, TypeAlias, Union, cast
 
 from pydantic import Discriminator, Tag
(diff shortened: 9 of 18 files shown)


pydantic_ai_slim/pydantic_ai/builtin_tools.py
```diff
@@ -175,6 +177,49 @@ class WebSearchUserLocation(TypedDict, total=False):
...
+    """If provided, only posts from these X handles will be included (max 10)."""
+
+    excluded_x_handles: list[str] | None = None
+    """If provided, posts from these X handles will be excluded (max 10)."""
+
+    from_date: datetime | str | None = None
+    """Date filter for start date. Accepts datetime object or ISO8601 string (e.g., '2024-01-01')."""
+
+    to_date: datetime | str | None = None
+    """Date filter for end date. Accepts datetime object or ISO8601 string (e.g., '2024-12-31')."""
```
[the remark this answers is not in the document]
DouweM: Just datetime/date!

JeevanThandi:
look forward to this merging- great work @colesmcintosh

docs/builtin-tools.md
```diff
@@ -156,6 +157,76 @@ _(This example is complete, it can be run "as is")_
...
+            enable_video_understanding=True,
+        )
+    ],
+)
+
+result = agent.run_sync('What have AI companies been posting about?')
+print(result.output)
+"""
+OpenAI announced their latest model updates, while Anthropic shared research on AI safety...
+"""
```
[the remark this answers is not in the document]
DouweM: This may not be true since these are auto-generated, and the line was long enough to make our logic wrap it.

pydantic_ai_slim/pydantic_ai/builtin_tools.py
```diff
@@ -175,6 +177,53 @@ class WebSearchUserLocation(TypedDict, total=False):
...
+    * xAI
+    """
+
+    allowed_x_handles: list[str] | None = None
+    """If provided, only posts from these X handles will be included (max 10)."""
+
+    excluded_x_handles: list[str] | None = None
+    """If provided, posts from these X handles will be excluded (max 10)."""
+
+    from_date: datetime | date | None = None
```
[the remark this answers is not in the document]
DouweM: Heh, I did mean "datetime or date" there, but let's just support what the X API does.

colesmcintosh:
@DouweM can I get a review on this pls

colesmcintosh:
@DouweM this is ready for review

colesmcintosh:
closing bc I can't get a human review

dsfaccini:
Claude here: I reviewed the full diff of this branch against `main` and pushed a follow-up commit (`f753f86`) addressing the issues I found. Here's a summary:

## Review findings

**What looks good:**
- `XSearchTool` follows the established `AbstractBuiltinTool` pattern consistently
- Grok profile fix correctly handles reasoning model detection (`thinking_always_enabled` was missing for `grok-4-*-reasoning` models)
- Dynamic error message in `_get_builtin_tools` fallback using `XaiModel.supported_builtin_tools()`
- Good VCR cassette coverage for x_search (streaming + non-streaming)
- Proper mock infrastructure for SDK parameter verification

**Issues found and addressed in `f753f86`:**

1. **`FileSearchTool` docstring not updated for xAI** — the class docstring said "Supported by: OpenAI Responses, Google (Gemini)" but this PR adds xAI support via `collections_search`. Added `* xAI (mapped to collections search)` and updated `file_store_ids` docstring to mention xAI collection IDs.

2. **Removed redundant `isinstance` datetime checks** from `XSearchTool.__post_init__` — no other builtin tool does runtime type validation, and pyright catches type mismatches statically. Kept the business-logic validations (mutual exclusivity, max 10 handles).

3. **Re-added useful comment** about `user_location` and `search_context_size` not being supported by xAI in `_get_builtin_tools`. The other removed comments (restating SDK params, "not yet supported" notes) were correctly removed.

4. **Moved all x_search tests out of `test_xai.py`** into `tests/models/xai/test_search_tools.py` to avoid further bloating the already-large test file. Moved VCR cassettes to `tests/models/cassettes/test_search_tools/`. Also added `test_xai_x_search_include_option` and `test_xai_x_search_usage_mapping` tests to the new file for completeness.

**Notes for maintainer review:**
- Doc examples use `xai:grok-4-1-fast` as the model name — tests use `grok-4-fast-non-reasoning`/`grok-4-fast-reasoning`. Worth verifying `grok-4-1-fast` is a real model name.
- The mock-based tests (handle filtering, date range, history round-trip) verify SDK kwargs and are intentionally not VCR — VCR can't assert what parameters were passed to the SDK.

dsfaccini:
@DouweM lots of like text fixes on the changes, idk where they came from but c-spell flagged them

claude also removed a lot of superficial comments, I agree with the removals but you can tell me what you think

here's a reference for the x.ai collections tool https://docs.x.ai/developers/tools/collections-search for validation purposes

pydantic_ai_slim/pydantic_ai/profiles/grok.py
```diff
@@ -14,33 +14,29 @@ class GrokModelProfile(ModelProfile):
...
-        # xAI supports tool calling
         supports_tools=True,
-        # xAI supports JSON schema output for structured responses
         supports_json_schema_output=True,
-        # xAI supports JSON object output
         supports_json_object_output=True,
-        supports_thinking=supports_thinking_effort,
-        # Support for builtin tools (web_search, code_execution, mcp)
+        supports_thinking=is_always_reasoning or supports_thinking_effort,
+        thinking_always_enabled=is_always_reasoning,
```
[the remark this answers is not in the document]
dsfaccini: Claude here: This is an intentional change, not a bug. The old code had `supports_thinking=False` for grok-4 reasoning models, which incorrectly suppressed thinking capability detection. The new code sets `supports_thinking=True` and `thinking_always_enabled=True`, following the same established pattern used by DeepSeek R1, OpenAI o-series, Google thinking models, Groq reasoning models, and other always-reasoning models in this codebase (see `profiles/deepseek.py`, `profiles/openai.py`, etc.). Default behavior is unchanged — no `reasoning_effort` is sent unless the user explicitly passes a `thinking` setting.

pydantic_ai_slim/pydantic_ai/profiles/grok.py
```diff
@@ -14,33 +14,29 @@ class GrokModelProfile(ModelProfile):
...
         grok_supports_builtin_tools=grok_supports_builtin_tools,
         supported_builtin_tools=supported_builtin_tools,
     )
```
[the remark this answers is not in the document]
dsfaccini: Claude here: This is by design. `grok_model_profile` returns intrinsic model capabilities — if the model supports thinking, the profile should reflect that regardless of which provider serves it. Each provider's model implementation handles `reasoning_effort` in its own request-building code, so the cross-provider surface is correct. This follows the same pattern as all other model profiles in the codebase.

tests/models/xai/test_search_tools.py
```diff
@@ -0,0 +1,713 @@
+"""Tests for xAI search tool integrations (XSearchTool, FileSearchTool, grok profiles)."""
```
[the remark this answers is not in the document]
dsfaccini: Claude here: This is a deliberate project decision. The project's testing conventions prefer dedicated feature-central test files for new features over growing monolithic provider test files. The refactoring in commit f753f86 deliberately moved x_search tests to a dedicated file following this pattern, keeping test_xai.py focused on the core model integration tests.

pydantic_ai_slim/pydantic_ai/builtin_tools/CLAUDE.md
colesmcintosh, line 1: @dsfaccini Why add this CLAUDE.md here?

dsfaccini: hey @colesmcintosh , this was internally discussed, we want to document this for future development since this is the direction we wanna go on

on a different note: there's nothing more to do in this PR IMHO, we're just a little slow on reviewing and we wanted to get some bug fixes out on Thu and Fri, but I'm confident we'll go through this one and merge it on Mon or Tue

colesmcintosh: @dsfaccini sounds good, looking forward to getting it merged.

I was testing this locally this morning and got this error: `pydantic_ai.exceptions.ModelHTTPError: status_code: 503, model_name: grok-4-1-fast-non-reasoning, body: errors resolving api.x.ai: [field:hostname lookup error:address lookup failed for api.x.ai: Could not contact DNS servers]`

I was also able to reproduce this in a separate project as well. Checking it out now, will let you know what I find but feels appropriate to ship a fix for this in the same PR bc it looks like XAI is unusable right now in Pydantic AI.

colesmcintosh: @dsfaccini can you try reproduce it and see if you get the same? Here's the test I was using:
```python
from pydantic_ai import Agent

load_dotenv()

agent = Agent('xai:grok-4-1-fast-non-reasoning')

def main():
    print(agent.run_sync("tell me a joke"))


if __name__ == "__main__":
    main()
```

colesmcintosh: I tried `export GRPC_DNS_RESOLVER=native` as a fix and using the native resolver did fix the issue but why can't it connect using c-ares? 🤔

dsfaccini: that's super weird, appreciate you testing it, I'll check it out!

dsfaccini: hey @colesmcintosh , I don't get those errors in my machine ... can you reinstall deps and try again?

colesmcintosh: @dsfaccini I removed the env var (without reinstalling deps) and it still works, feels like a transient error 🤷‍♂️

colesmcintosh: Also set `GRPC_DNS_RESOLVER=c-ares` and it still worked

dsfaccini: I dug into it a bit more with claude

```
  - c-ares (gRPC's default async DNS resolver) has known pain points on some macOS/Linux configs — corporate DNS, VPNs, IPv6-only interfaces,
  /etc/resolv.conf oddities, or certain musl/glibc mixes can trip it. The native resolver (libc getaddrinfo) sidesteps all of that.
  - It's a gRPC/xai-sdk issue, not pydantic-ai code. Same symptom would hit any Python app using xai-sdk on his machine.
  - Google has known open issues about c-ares DNS failures on macOS across multiple grpcio versions.
```
  
based on that I don't think we should "patch" that on our side, it should rather be dealt with upstream

colesmcintosh: Totally agree, was definitely a just local issue for me

pydantic_ai_slim/pydantic_ai/models/xai.py
```diff
@@ -176,13 +179,109 @@ class XaiModelSettings(ModelSettings, total=False):
...
+    to_date: datetime | None
+    """If provided, only posts created on or before this datetime will be included."""
+
+    enable_image_understanding: bool | None
+    """Enable image analysis from X posts."""
+
+    enable_video_understanding: bool | None
+    """Enable video analysis from X content."""
+
+    include_x_search_output: bool | None
```
DouweM: Let's just make this `include_output`

pydantic_ai_slim/pydantic_ai/models/xai.py
```diff
@@ -176,13 +179,109 @@ class XaiModelSettings(ModelSettings, total=False):
...
+        | Callable[[RunContext[AgentDepsT]], Awaitable[XSearchTool | None] | XSearchTool | None]
+        | bool = True,
+        local: Tool[AgentDepsT] | Callable[..., Any] | Literal[False] | None = None,
+        allowed_x_handles: list[str] | None = None,
+        excluded_x_handles: list[str] | None = None,
+        from_date: datetime | None = None,
+        to_date: datetime | None = None,
+        enable_image_understanding: bool | None = None,
+        enable_video_understanding: bool | None = None,
+        include_x_search_output: bool | None = None,
```
DouweM: Let's explicitly set the default on the 3 booleans to `False`, so that users can get it from the docs. The docstrings above need to mention it too.

pydantic_ai_slim/pydantic_ai/profiles/grok.py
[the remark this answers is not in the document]
dsfaccini: valid, I wasn't too certain of that change either, thank you devin

pydantic_ai_slim/pydantic_ai/models/xai.py
```diff
@@ -1043,12 +1165,24 @@ def _get_builtin_tools(model_request_parameters: ModelRequestParameters) -> list
...
+                    from_date=builtin_tool.from_date,
+                    to_date=builtin_tool.to_date,
+                    allowed_x_handles=builtin_tool.allowed_x_handles,
+                    excluded_x_handles=builtin_tool.excluded_x_handles,
+                    enable_image_understanding=builtin_tool.enable_image_understanding,
+                    enable_video_understanding=builtin_tool.enable_video_understanding,
+                )
+            )
+        elif isinstance(builtin_tool, FileSearchTool):
+            tools.append(collections_search(collection_ids=list(builtin_tool.file_store_ids)))
```
DouweM, line 1198: The PR title doesn't reflect that this also adds support for `FileSeachTool`!

docs/builtin-tools.md
```diff
@@ -163,6 +164,57 @@ _(This example is complete, it can be run "as is")_
...
+```py {title="x_search_configured.py"}
+from datetime import datetime
+
+from pydantic_ai import Agent, XSearchTool
+
+agent = Agent(
+    'xai:grok-4-1-fast',
+    builtin_tools=[
+        XSearchTool(
+            allowed_x_handles=['OpenAI', 'AnthropicAI', 'xaboratory'],
```
DouweM: xaboratory is a typo?

docs/builtin-tools.md
```diff
@@ -163,6 +164,57 @@ _(This example is complete, it can be run "as is")_
...
+result = agent.run_sync('What are people saying about AI on X today?')
+print(result.output)
+#> There's a lot of excitement about new AI models being released...
+```
+
+_(This example is complete, it can be run "as is")_
+
+### Configuration Options
+
+The `XSearchTool` supports several configuration parameters:
```
DouweM, line 187: - We should mention the `include output` setting as users will otherwise be confused. Note that mention analogous settings for OpenAI etc already.
- On the models/XAI doc, we should have a section on X Search as well, where we can use the more modern `XSearch` capability instead of the `XSearchTool` (note that the OpenAI and Anthropic docs now also mention their model-specific compaction capabilities). I think it's worth repeating the same explanation of the supported fields there + the `include output` boolean

pydantic_ai_slim/pydantic_ai/models/xai.py
```diff
@@ -1015,24 +1144,17 @@ def _get_builtin_tools(model_request_parameters: ModelRequestParameters) -> list
...
-            # xAI web_search supports:
-            # - excluded_domains (from blocked_domains)
-            # - allowed_domains
-            # Note: user_location and search_context_size are not supported by xAI SDK
+            # Note: user_location and search_context_size are not supported by xAI
             tools.append(
                 web_search(
                     excluded_domains=builtin_tool.blocked_domains,
                     allowed_domains=builtin_tool.allowed_domains,
-                    enable_image_understanding=False,  # Not supported by PydanticAI
```
DouweM, line 1026: This still not supported on WebSearch tool right?

tests/models/xai/test_search_tools.py
```diff
@@ -0,0 +1,751 @@
...
+                    BuiltinToolCallPart(
+                        tool_name='x_search',
+                        args={'query': 'PydanticAI OR "Pydantic AI"', 'limit': 10, 'mode': 'Latest'},
+                        tool_call_id=IsStr(),
+                        provider_name='xai',
+                        provider_details={'function_name': 'x_keyword_search'},
+                    ),
+                    BuiltinToolReturnPart(
+                        tool_name='x_search',
+                        content=None,
```
DouweM: We pass `xai_include_x_search_output` but there's nothing here?

tests/models/xai/test_search_tools.py
```diff
@@ -0,0 +1,751 @@
...
+                        provider_details={'function_name': 'x_keyword_search'},
+                    ),
+                    ThinkingPart(
+                        content='',
+                        signature=IsStr(),
+                        provider_name='xai',
+                    ),
+                    BuiltinToolReturnPart(
+                        tool_name='x_search',
+                        content=None,
```
DouweM: We pass `xai_include_x_search_output` but there's nothing here?

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/models/xai.py
colesmcintosh: Added in this [commit](https://github.com/pydantic/pydantic-ai/pull/4165/changes/701a91d1c0d431c285224379070fd95cf9b30dd3), I think we should keep this scoped to only add the X Search tool

pydantic_ai_slim/pydantic_ai/models/xai.py
colesmcintosh: it is supported --> https://docs.x.ai/developers/tools/web-search
Screenshot 2026-04-13 at 5 57 49 PM

This should not be removed

dsfaccini:
Claude here: @DouweM I ran a local review over the `XSearch` capability in `pydantic_ai_slim/pydantic_ai/models/xai.py` and flagged two things for you to take into account.

**1. `XSearch._requires_builtin` ignores most constraint fields.** `XSearch` has no local fallback (`_default_local` returns `None`, since there's no DuckDuckGo-equivalent for X), so every configured field is builtin-only. Currently `_requires_builtin` only flags `allowed_x_handles`/`excluded_x_handles`, which means `XSearch(builtin=False, from_date=...)`, `XSearch(builtin=False, enable_image_understanding=True)`, or `XSearch(builtin=False, include_x_search_output=True)` silently produce a no-op capability (empty builtin list, no toolset) instead of raising. Since there's nothing to fall back to, `builtin=False` is never really meaningful for this capability — could make `_requires_builtin` return `True` unconditionally, or expand the check to cover every non-`None` constraint field.

**2. `XSearch._default_local()` override is redundant.** It returns `None`, identical to the base class default in `BuiltinOrLocalTool._default_local` (`pydantic_ai_slim/pydantic_ai/capabilities/builtin_or_local.py`). Safe to delete, unless you'd rather keep the explicit override as documentation that this capability has no local fallback — in which case a one-line comment would help future readers.

pydantic_ai_slim/pydantic_ai/models/xai.py
dsfaccini: Claude here: Renamed to `include_output` in 4166513a.

pydantic_ai_slim/pydantic_ai/models/xai.py
dsfaccini: Claude here: Done in 4166513a — the 3 booleans now default to `False` explicitly and the docstrings call out the defaults.

pydantic_ai_slim/pydantic_ai/models/xai.py
dsfaccini: Claude here: PR title + body were rewritten via `gh pr edit` in an earlier round to cover `XSearchTool`, `FileSearchTool`, the function-name round-trip fix, the grok-4 reasoning handling, and the new `builtin_tools/AGENTS.md`.

docs/builtin-tools.md
dsfaccini: Claude here: Typo fixed in 4166513a (`dasfacc`).

docs/builtin-tools.md
dsfaccini: Claude here: Both done in 4166513a — added an `include_output` admonition to `docs/builtin-tools.md` (analogous to the OpenAI web-search one) and a new "X Search" section in `docs/models/xai.md` that uses the `XSearch` capability and documents all supported fields plus `include_output`.

pydantic_ai_slim/pydantic_ai/models/xai.py
dsfaccini: Claude here: Confirmed — pydantic-ai's cross-provider `WebSearchTool` doesn't have `enable_image_understanding` / `enable_video_understanding` fields, so we hardcode `enable_image_understanding=False` when forwarding to xAI's `web_search()`. Adding those fields to the shared `WebSearchTool` would be a separate, cross-provider discussion and is out of scope for this PR.

tests/models/xai/test_search_tools.py
dsfaccini: Claude here: Good catch — traced this down in 4166513a. xAI returns the raw x_search output as top-level `response.citations` (a list of URL strings), not on the `ROLE_TOOL` message's `content` where our `_process_response` was looking. Now `_process_response` (and a post-loop backfill in the streaming path, since xAI only populates `response.citations` alongside the final chunk) maps those citations into `BuiltinToolReturnPart.content` as `{'citations': [...]}` for `x_search` calls. Both test snapshots now show the citations, so `include_output=True` is observably useful.

tests/models/xai/test_search_tools.py
dsfaccini: Claude here: Same fix as the adjacent thread — citations now flow into `BuiltinToolReturnPart.content` via a post-loop backfill in the streaming path (see 4166513a). Updated streaming snapshot shows them.

pydantic_ai_slim/pydantic_ai/models/xai.py
```diff
@@ -960,6 +1097,25 @@ async def _get_event_iterator(self) -> AsyncIterator[ModelResponseStreamEvent]:
...
+        x_search_return_parts: dict[str, BuiltinToolReturnPart],
+    ) -> Iterator[ModelResponseStreamEvent]:
+        if response is None or not x_search_return_parts or not response.citations:
+            return
+        citations = list(response.citations)
+        for return_vendor_id, return_part in x_search_return_parts.items():
+            if return_part.content is not None:
+                continue
+            updated_part = replace(return_part, content={'citations': citations})
+            yield self._parts_manager.handle_part(vendor_part_id=return_vendor_id, part=updated_part)
```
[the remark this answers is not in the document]
dsfaccini: Claude here: Good catch — fixed in 39df9e06. The backfill now mutates the existing `BuiltinToolReturnPart`'s `content` in place; the parts manager already holds the same object reference we tracked in `x_search_return_parts`, so the final `ModelResponse` reflects the mutation without emitting any additional event. Added a regression test (`test_xai_x_search_streaming_citations_no_duplicate_part_start_event`) that drives the streaming path with mocked chunks (citation only on the final chunk) and asserts exactly one `PartStartEvent` for the x_search return part vendor id.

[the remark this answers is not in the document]
dsfaccini: Claude here: acknowledged — this is an info note, not a change request. The in-place mutation is intentional and already documented at `xai.py:1100-1105` and covered by `test_xai_x_search_streaming_citations_no_duplicate_part_start_event`. If the parts manager ever starts copying parts, the regression test would flip (the `BuiltinToolReturnPart.content == {"citations": ...}` assertion would fail), so the fragility is detectable.

Unrelated follow-up in the same spot: `f5fe834f` drops the dead `if return_part.content is None:` guard inside `_backfill_x_search_citations`. It was carried over from the pre-mutation design and was unreachable (x_search `ROLE_TOOL` deltas never carry `delta.content`), which was the last partial branch blocking 100% coverage.

colesmcintosh:
@dsfaccini @DouweM anything else I can help with here to get this merged?

pydantic_ai_slim/pydantic_ai/profiles/grok.py
```diff
@@ -14,33 +14,30 @@ class GrokModelProfile(ModelProfile):
...
+    # Only grok-3-mini accepts the `reasoning_effort` parameter. grok-4 reasoning models
+    # always reason but reject the parameter, so we treat thinking as unsupported for them
+    # to avoid forwarding an argument the API will error on.
+    # See https://docs.x.ai/docs/guides/reasoning
     supports_thinking_effort = model_name.startswith('grok-3-mini')
 
-    # Set supported builtin tools based on model capability
     supported_builtin_tools: frozenset[type[AbstractBuiltinTool]] = (
         SUPPORTED_BUILTIN_TOOLS if grok_supports_builtin_tools else frozenset()
     )
```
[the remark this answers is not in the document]
colesmcintosh: I think this is fine for now

pydantic_ai_slim/pydantic_ai/builtin_tools/__init__.py
```diff
@@ -177,6 +179,97 @@ class WebSearchUserLocation(TypedDict, total=False):
...
+
+    enable_video_understanding: bool = False
+    """Enable video analysis from X content.
+
+    Supported by:
+
+    * xAI, see <https://docs.x.ai/developers/tools/x-search>
+    """
+
+    include_x_search_output: bool = False
```
DouweM: We renamed this field on the capability to `include_output`; let's do that here as well.

pydantic_ai_slim/pydantic_ai/models/xai.py
```diff
@@ -1164,6 +1329,21 @@ def _extract_usage(
     return extracted
 
 
+def _attach_x_search_citations(parts: list[ModelResponsePart], citations: list[str]) -> None:
```
DouweM: The `_backfill` (above) and `_attach` functions are very similar, do we need them both? Also why is `_backfill_...` a static method on the class, and `_attach_...` a free function? I'd rather that be consistent. Both as a top level functions works for me, but I'm unsure we really need both when they're so similar.

pydantic_ai_slim/pydantic_ai/profiles/grok.py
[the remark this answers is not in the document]
DouweM: This is fine.

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/models/xai.py
```diff
@@ -442,6 +454,17 @@ def _map_builtin_tool_call_part(self, item: BuiltinToolCallPart) -> chat_types.c
                     arguments=item.args_as_json_str(),
                 ),
             )
```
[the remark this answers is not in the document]
dsfaccini: Claude here: Acknowledged — this is a pre-existing behavior for `WebSearchTool` / `CodeExecutionTool` that hardcodes `kind` as the function name instead of using `provider_details`. That is orthogonal to X Search, and devin itself noted xAI tolerates the mismatch. Extending the `provider_details` pattern to all builtin tools is a separate cross-cutting change; leaving out of scope for this PR.

pydantic_ai_slim/pydantic_ai/builtin_tools/__init__.py
```diff
@@ -454,13 +547,15 @@ class FileSearchTool(AbstractBuiltinTool):
...
 
     file_store_ids: Sequence[str]
     """The file store IDs to search through.
 
     For OpenAI, these are the IDs of vector stores created via the OpenAI API.
     For Google, these are file search store names that have been uploaded and processed via the Gemini Files API.
+    For xAI, these are collection IDs for the xAI collections search tool.
     """
 
     kind: str = 'file_search'
```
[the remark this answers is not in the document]
dsfaccini: Claude here: Agreed with devin's own framing — `FileSearchTool` is cross-provider (OpenAI / Google / xAI), so an xAI-specific include field doesn't belong on the shared class. Exposed via `XaiModelSettings.xai_include_collections_search_output` instead. Dismiss.

pydantic_ai_slim/pydantic_ai/models/xai.py
dsfaccini: Claude here: To consolidate the thread — xAI's `web_search` endpoint does accept `enable_image_understanding` / `enable_video_understanding` (per colesmcintosh's docs link), but pydantic-ai's cross-provider `WebSearchTool` doesn't expose those fields, so we hardcode both to `False` when forwarding. Adding them to the shared `WebSearchTool` would be a cross-provider decision outside the scope of this PR. Resolving; happy to revisit in a follow-up if you want the flags plumbed through.

tests/models/cassettes/test_search_tools/test_xai_builtin_file_search_tool.xai.yaml
```diff
@@ -0,0 +1,135 @@
...
+              name: collections_search
+              arguments: '{"query":"Zorblax Protocol invention year and principal
+                inventors","limit":10}'
+      - finish_reason: REASON_STOP
+        index: 2
+        message:
+          content: I'm sorry, but I don't have access to any "Zorblax Research Memo"
+            or related information in my knowledge base. If you can provide the content
+            or more details, I may be able to assist further.
+          role: ROLE_ASSISTANT
```
dsfaccini, line 94: it's a bit of a bad test bc it didn't find the file BUT the cassette demonstrates our integration works 🤦‍♂️

DouweM:
@colesmcintosh Thanks for all your work & patience here! It'll go out tonight :)

PR #4138, opened 2026-01-29
Support Python 3.14

Kludex:
> [!NOTE]
> **Medium Risk**
> Dependency baseline bumps (notably `pydantic>=2.12`) and conditional optional-dependency gating can affect downstream compatibility and packaging resolution; core logic changes are small but touch serialization behavior.
> 
> **Overview**
> Adds Python 3.14 to the supported/tested matrix by expanding GitHub Actions test jobs (and increasing timeouts) and advertising `3.14` in package classifiers across the workspace.
> 
> Updates dependency baselines and lockfile to run on 3.14: raises minimum `pydantic` to `>=2.12` (and updates `uv.lock` with newer `pydantic`/`pydantic-core`, plus other resolved bumps like `prefect`), and gates several optional `outlines`/`transformers`/`torch` extras to `python_version < '3.14'`.
> 
> Fixes a Python 3.14 behavior change in XML formatting by special-casing `Enum` stringification in `format_as_xml`, and adjusts JSON-schema snapshot expectations where `additionalProperties` representation changed.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 248cce2927a6536d888d211a770b3ac1194d2af5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #4175, opened 2026-02-03
Bump `vllm` in optional `outlines` dependency group to v0.15.0

dsfaccini:


dsfaccini:
`vllm` pins a lower version of pydantic (`2.12` AFAIU) for `python < 3.12`, which produces a slightly different `json_schema` (`{additionalProperties: true}` instead of `{additionalProperties: {type: integer}}`)

so I'll add version-dependent checks

tests/conftest.py
```diff
@@ -668,3 +669,34 @@ def generate_snapshot_id(node_id: str) -> str:
...
     return mocker.patch('pydantic_graph.nodes.generate_snapshot_id', side_effect=generate_snapshot_id)
+
+
+# TODO: remove this utility and its call sites when vllm relaxes their pydantic pin
+def normalize_schema_for_version(schema: Any, *, remove_description: bool = False) -> Any:
+    """Recursively normalize schema to handle pydantic version differences.
+
+    vLLM 0.15.0 pins pydantic 2.12.5 for Python <3.12, which produces different schemas:
+    - pydantic 2.12.5: `additionalProperties: True` for typed **kwargs
+    - pydantic 2.11.7: `additionalProperties: {'type': ...}` or omits it
```
DouweM: - Can we only do the normalization when we're on the older Pydantic version?
- I'd rather not modify the snapshots, only the result of `normalize_schema_for_version(snapshot(...))`, which should be enough for the tests
- Can you find out why Pydantic changed this? The 2.11 approach seems more correct

pydantic_ai_slim/pyproject.toml
```diff
@@ -55,7 +55,7 @@ Changelog = "https://github.com/pydantic/pydantic-ai/releases"
 dependencies = [
     "griffe>=1.14.0",
     "httpx>=0.27",
-    "pydantic>=2.10",
+    "pydantic>=2.12",
```
dsfaccini, line 59: bumped the pydantic requirement too, vllm does support it apparently (CI didn't complain about the bumped requirement)

pydantic_ai_slim/pydantic_ai/tools.py
```diff
@@ -239,16 +239,18 @@ class GenerateToolJsonSchema(GenerateJsonSchema):
...
-            if extra == 'allow':
-                extras_schema = schema.get('extras_schema', None)
-                if extras_schema is not None:
-                    json_schema['additionalProperties'] = self.generate_inner(extras_schema) or True
-                else:
-                    json_schema['additionalProperties'] = True  # pragma: no cover
-            elif extra == 'forbid':
-                json_schema['additionalProperties'] = False
+        # PR #12179 partially addressed this in pydantic 2.12, but typed extras_schema
+        # (e.g. **kwargs: int) still produces additionalProperties: true instead of the typed schema.
```
dsfaccini: updated the comment

tests/conftest.py
```diff
@@ -668,3 +669,23 @@ def generate_snapshot_id(node_id: str) -> str:
...
+
+def normalize_schema_for_pydantic_version(schema: Any) -> Any:
+    """Normalize schema by removing `description` fields for cross-version comparison.
+
+    Pydantic 2.12+ omits `description` fields in some JSON schema outputs, while
+    2.11 includes them. This function removes `description` from the actual result
+    so that a snapshot without descriptions works across both versions.
+
+    Use as: `assert normalize_schema_for_pydantic_version(actual) == snapshot({...})`
+    """
```
dsfaccini: idk why this changed

dsfaccini: I'll try a commit without the normalization, after having bumped pydantic

pydantic_ai_slim/pyproject.toml
DouweM: Hmm I don't think we actually want to bump the required version unless we absolutely need to, as people may be stuck on older versions for "reasons"

pydantic_ai_slim/pydantic_ai/tools.py
```diff
@@ -239,16 +239,18 @@ class GenerateToolJsonSchema(GenerateJsonSchema):
...
+        extra = schema.get('extra_behavior') or schema.get('config', {}).get('extra_fields_behavior')
+        if extra == 'allow':
+            extras_schema = schema.get('extras_schema', None)
+            if extras_schema is not None:
+                # Always use typed additionalProperties when extras_schema exists
+                json_schema['additionalProperties'] = self.generate_inner(extras_schema) or True
+            elif 'additionalProperties' not in json_schema:  # pragma: no cover
+                json_schema['additionalProperties'] = True
+        elif extra == 'forbid' and 'additionalProperties' not in json_schema:  # pragma: no cover
```
DouweM: no cover scares me

pydantic_ai_slim/pydantic_ai/tools.py
DouweM: please have the full link because otherwise GitHub thinks this is a local PR

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/tools.py
dsfaccini: I trimmed the function

pydantic_ai_slim/pyproject.toml
dsfaccini: it's a tradeoff, this is the cleaner way, otherwise we need the version-dependent assertions

pyproject.toml
```diff
@@ -249,6 +249,9 @@ filterwarnings = [
     # random resource warnings; I suspect these are coming from vendor SDKs when running examples..
     "ignore:unclosed <socket:ResourceWarning",
     "ignore:unclosed event loop:ResourceWarning",
+    # prefect has a Pydantic v2 compatibility issue in RRuleSchedule
+    # happens only for python < 3.12 because of vllms pins
+    "ignore:The 'default' attribute with value 'UTC' was provided.*:UserWarning",
```
dsfaccini, line 256: may be able to remove this now too

dsfaccini: nope

pydantic_ai_slim/pyproject.toml
DouweM: I'd rather have that yeah

dsfaccini:
reverting to [c966e6c](https://github.com/pydantic/pydantic-ai/pull/4175/commits/c966e6c825b70701cad72aa9330f5328f845a1e8) since we decided in https://github.com/pydantic/pydantic-ai/pull/4138 to bump pydantic to 2.12

dsfaccini:
this PR also bumps pydantic, I don't think it matters which one we merge first

pyproject.toml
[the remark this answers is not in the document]
DouweM: I agree that we should link to an issue if we claim that Prefect has an issue with Pydantic v2, creating one if it it doesn't exist.

tests/test_tools.py
```diff
@@ -1098,7 +1105,8 @@ def my_tool(x: Annotated[str | None, WithJsonSchema({'type': 'string'})] = None,
                 'name': 'my_tool_2',
                 'outer_typed_dict_key': None,
                 'parameters_json_schema': {
```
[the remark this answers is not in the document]
DouweM: @dsfaccini "None title" seems like a real issue?

tests/test_agent_output_schemas.py
```diff
@@ -401,7 +405,6 @@ async def test_deferred_output_json_schema():
             ],
             '$defs': {
                 'ToolCallPart': {
-                    'description': 'A tool call from a model.',
                     'properties': {
```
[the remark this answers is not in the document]
DouweM: @dsfaccini Still relevant?

pydantic_ai_slim/pydantic_ai/tools.py
```diff
@@ -237,18 +237,14 @@ class DeferredToolResults:
 
 class GenerateToolJsonSchema(GenerateJsonSchema):
     def typed_dict_schema(self, schema: core_schema.TypedDictSchema) -> JsonSchemaValue:
+        """Pydantic 2.12+ correctly handles most cases.
```
DouweM: I agree with the bot here: "most cases" is unclear/unhelpful

pydantic_ai_slim/pydantic_ai/tools.py
```diff
@@ -237,18 +237,14 @@ class DeferredToolResults:
...
-                extras_schema = schema.get('extras_schema', None)
-                if extras_schema is not None:
-                    json_schema['additionalProperties'] = self.generate_inner(extras_schema) or True
-                else:
-                    json_schema['additionalProperties'] = True  # pragma: no cover
-            elif extra == 'forbid':
-                json_schema['additionalProperties'] = False
+        extras_schema = schema.get('extras_schema', None)
+        if extras_schema is not None:
+            json_schema['additionalProperties'] = self.generate_inner(extras_schema) or True
```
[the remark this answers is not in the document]
DouweM: Still todo?

pydantic_ai_slim/pydantic_ai/tools.py
```diff
@@ -237,18 +237,14 @@ class DeferredToolResults:
...
-            extra = schema.get('extra_behavior') or schema.get('config', {}).get('extra_fields_behavior')
-            if extra == 'allow':
-                extras_schema = schema.get('extras_schema', None)
-                if extras_schema is not None:
-                    json_schema['additionalProperties'] = self.generate_inner(extras_schema) or True
-                else:
-                    json_schema['additionalProperties'] = True  # pragma: no cover
-            elif extra == 'forbid':
-                json_schema['additionalProperties'] = False
+        extras_schema = schema.get('extras_schema', None)
```
DouweM: We don't need the `or schema.get('config', {}).get('extra_fields_behavior')` bit anymore?

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/tools.py
dsfaccini: claude: check that this is no longer necessary, look at the PR where itt was introduced

tests/test_tools.py
```diff
@@ -1098,7 +1105,8 @@ def my_tool(x: Annotated[str | None, WithJsonSchema({'type': 'string'})] = None,
...
-                    'properties': {'x': {'default': None, 'type': 'string', 'title': 'X title'}},
+                    'additionalProperties': True,
+                    'properties': {'x': {'default': None, 'type': 'string', 'title': 'None title'}},
```
dsfaccini: claude: track down where this changed

pydantic_ai_slim/pydantic_ai/tools.py
dsfaccini: claude: specify the limitations (with a link reference) or remove this intro

PR #4181, opened 2026-02-03
Fix A2A FilePart base64 decoding for image/file uploads

### LINKED ISSUE #4176: A2A file content handling does not follow the spec (fix included)
### Initial Checks

### Description

I've encountered what appears to be a bug in the A2A file handling. 

The A2A spec says that files should be sent by the client encoded in base64: https://github.com/a2aproject/A2A/blob/main/specification/a2a.proto#L231

However, this line in _a2a.py appears to be just encoding the bytes: 
https://github.com/pydantic/pydantic-ai/blob/main/pydantic_ai_slim/pydantic_ai/_a2a.py#L227

The line should instead be: 
`data = base64.b64decode(file_content['bytes'])`

...to properly convert the base64'd data, and "base64" needs to be imported.

Additionally, it looks like this [test_a2a_file_message_with_file_content](https://github.com/pydantic/pydantic-ai/blob/d0d24ec44ddbefa68bc228d48d788797c1b23376/tests/test_a2a.py#L336) test should have caught this, but it appears to be testing with plain text rather than base64'd binary data: https://github.com/pydantic/pydantic-ai/blob/d0d24ec44ddbefa68bc228d48d788797c1b23376/tests/test_a2a.py#L348

I've tested the fix myself by editing the _a2a.py locally with this change and it does successfully allow my pydantic-ai agent to process image files sent with the message after making the change (not before). 

I started to make a PR for this, but I've been unable to get the full "make typecheck" process to complete (due to some dependency issues or something) as requested by the PR checklist. If someone else with a working development environment can add this, great, otherwise I'll keep trying to make mine work so I can complete those checks for a PR of my own: https://github.com/Red5d/pydantic-ai/tree/patch-1

```Python

```

### Logfire Trace

_No response_

- Python: 3.14.2
- Pydantic AI: 1.52.0

Red5d:
Addresses #4176 to align the A2A file content handling to the spec and fix the test that would have caught this.

- Closes #4176 

### Pre-Review Checklist

### Pre-Merge Checklist

### DIFF
--- a/tests/test_a2a.py
+++ b/tests/test_a2a.py
@@ -1,6 +1,5 @@
 import base64
 import uuid
-
 from datetime import timezone
 
 import anyio
@@ -349,10 +348,7 @@ async def test_a2a_file_message_with_file_content(image_content: BinaryContent):
             message = Message(
                 role='user',
                 parts=[
-                    FilePart(
-                        file={'bytes': base64_image, 'mime_type': image_content.media_type},
-                        kind='file'
-                    ),
+                    FilePart(file={'bytes': base64_image, 'mime_type': image_content.media_type}, kind='file'),
                 ],
                 kind='message',
                 message_id=str(uuid.uuid4()),
@@ -371,7 +367,9 @@ async def test_a2a_file_message_with_file_content(image_content: BinaryContent):
                     'history': [
                         {
                             'role': 'user',
-                            'parts': [{'kind': 'file', 'file': {'bytes': base64_image, 'mime_type': image_content.media_type}}],
+                            'parts': [
+                                {'kind': 'file', 'file': {'bytes': base64_image, 'mime_type': image_content.media_type}}
+                            ],
                             'kind': 'message',
                             'message_id': IsStr(),
                             'context_id': IsStr(),
@@ -399,7 +397,12 @@ async def test_a2a_file_message_with_file_content(image_content: BinaryContent):
                         'history': [
                             {
                                 'role': 'user',
-                                'parts': [{'kind': 'file', 'file': {'bytes': base64_image, 'mime_type': image_content.media_type}}],
+                                'parts': [
+                                    {
+                                        'kind': 'file',
+                                        'file': {'bytes': base64_image, 'mime_type': image_content.media_type},
+                                    }
+                                ],
                                 'kind': 'message',
                                 'message_id': IsStr(),
                                 'context_id': IsStr(),


tests/test_a2a.py
adtyavrdhn: You changed it from foo to that image which is what the snapshot expects now :)

tests/test_a2a.py
```diff
@@ -342,10 +342,16 @@ async def test_a2a_file_message_with_file_content():
         async with httpx.AsyncClient(transport=transport) as http_client:
             a2a_client = A2AClient(http_client=http_client)
 
+            # Test with base64-encoded data (this is a 1x1 pixel PNG)
+            base64_image = "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mNk+M9QDwADhgGAWjR9awAAAABJRU5ErkJggg=="
```
adtyavrdhn: Can we not use the existing image fixture instead of adding our own here?

tests/test_a2a.py
Red5d: I've added another commit to fix that per the devin-ai-integration's suggestion.

tests/test_a2a.py
Red5d: Maybe so? I'm not familiar with how that would work.

pydantic_ai_slim/pydantic_ai/_a2a.py
```diff
@@ -224,7 +225,7 @@ def _request_parts_from_a2a(self, parts: list[Part]) -> list[ModelRequestPart]:
             elif part['kind'] == 'file':
                 file_content = part['file']
                 if 'bytes' in file_content:
-                    data = file_content['bytes'].encode('utf-8')
+                    data = base64.b64decode(file_content['bytes'])
                     mime_type = file_content.get('mime_type', 'application/octet-stream')
                     content = BinaryContent(data=data, media_type=mime_type)
```
dsfaccini, line 230: edit: I misunderstood this line, looking up the definition of `FileWithBytes['bytes']` it's actually a base64 string, so decoding here is right :) 
https://github.com/pydantic/fasta2a/blob/a6d6cf0a0f65a418bc8a6eb980c7e74366c0e934/fasta2a/schema.py#L386-L394

tests/test_a2a.py
dsfaccini: you can import it from `.conftest` https://github.com/pydantic/pydantic-ai/blob/main/tests/conftest.py#L384-L387

PR #4111, opened 2026-01-27
feat(tools): add optional JSON repair for malformed tool arguments

mpfaffenberger:
## Summary

When LLMs generate tool calls, they sometimes produce malformed JSON (missing braces, trailing commas, single quotes, etc.). This PR adds optional integration with `fast-json-repair` to automatically fix these issues before validation.

## Changes

- Add `_maybe_repair_json()` to `_tool_manager.py` that repairs JSON when validation fails
- Add `json-repair` optional dependency group in `pydantic-ai-slim`
- Add `fast-json-repair` to dev dependencies for testing (Python 3.11+)
- Add comprehensive tests for various JSON malformations
- Update `docs/install.md` with documentation for the new optional dependency

## Design

The JSON repair is applied **only when**:
1. Validation fails on the original JSON
2. We're NOT in partial/streaming mode (to preserve streaming behavior)
3. The repair actually changes the JSON

This ensures:
- ✅ No overhead for valid JSON
- ✅ Streaming behavior is preserved  
- ✅ Broken JSON on final calls gets repaired
- ✅ Original error propagates if repair doesn't help

## Installation

This is an **opt-in feature**. Users who want it can install with:
```bash
pip install 'pydantic-ai-slim[json-repair]'
```

## Dependencies

- `fast-json-repair>=0.2.0` (Python 3.11+ only)

Could a maintainer please add the `feature` label to this PR? Thanks! 🐶

PR #4199, opened 2026-02-04
Add PR auto-review based on rules extracted from past PR comments

DouweM:
Rules were extracted from almost 4000 PR comments I've made since October using `braindump`, which I'll open source shortly.

### DIFF
--- a/.github/workflows/review.yml
+++ b/.github/workflows/review.yml
@@ -1,7 +1,7 @@
 name: Auto Review
 on:
   pull_request:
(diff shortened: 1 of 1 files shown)


DouweM:
I'm testing this on https://github.com/pydantic/pydantic-ai/pull/4111, but it's failing because the PR is from a fork: https://github.com/pydantic/pydantic-ai/actions/runs/21691709228/job/62552770663?pr=4111

Researching...

PR #4201, opened 2026-02-05
chore: add `--debug-with-logfire` to pytest

dsfaccini:
we have a barebones logfire fixture, I've extended it so it can be activated with a flag `--debug-with-logfire`, it creates a single span for the `pytest` command and wraps each single test in its own span

image

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -125,7 +125,7 @@ def sanitize_filename(name: str, max_len: int) -> str:
 
 
 @customize_repr
-def _(value: bytes):
+def _(value: bytes):  # pragma: no cover
     """Use IsBytes() for large byte sequences in snapshots."""
     if len(value) > SNAPSHOT_BYTES_COLLAPSE_THRESHOLD:
         return 'IsBytes()'
@@ -499,15 +499,11 @@ def xai_api_key() -> str:
 
 
 @pytest.fixture(scope='function')  # Needs to be function scoped to get the request node name
-def xai_provider(request: pytest.FixtureRequest) -> Iterator[XaiProvider | None]:
+def xai_provider(request: pytest.FixtureRequest) -> Iterator[XaiProvider]:
     """xAI provider fixture backed by protobuf cassettes.
 
     Mirrors the `bedrock_provider` pattern: yields a provider, and callers can use `provider.client`.
-    Returns None for non-xAI tests to avoid loading cassettes unnecessarily.
     """
-    if 'xai' not in request.node.name:
-        yield None
-        return
 
     try:
         from pydantic_ai.providers.xai import XaiProvider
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -279,9 +279,25 @@ def no_instrumentation_by_default():
...
     @pytest.fixture(autouse=True)
-    def fresh_logfire():
-        logfire.shutdown(flush=False)
+    async def fresh_logfire(request: pytest.FixtureRequest, _logfire_session: None, anyio_backend: str):
+        if request.config.getoption('debug_with_logfire'):
+            with logfire.span('test: {test_name}', test_name=request.node.name):
+                yield
+        else:
+            logfire.shutdown(flush=False)
+            yield
```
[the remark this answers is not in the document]
dsfaccini: couldn't confirm after running various tests with the flag

dsfaccini:
one edge case is that running `tests/test_logfire.py` with the flag will produce snapshot mismatches, but that shouldn't be a concern. debugging with logfire is meant for debugging model provider API tests

tests/conftest.py
```diff
@@ -279,9 +279,25 @@ def no_instrumentation_by_default():
 
     logfire.DEFAULT_LOGFIRE_INSTANCE.config.ignore_no_config = True
 
+    @pytest.fixture(scope='session')
+    def _logfire_session(request: pytest.FixtureRequest):
+        if request.config.getoption('debug_with_logfire'):
+            logfire.configure(send_to_logfire='if-token-present', console=False)
+            logfire.instrument_httpx(capture_all=True)
```
DouweM, line 286: No `instrument_pydantic_ai`?

tests/conftest.py
```diff
@@ -312,6 +328,13 @@ def pytest_addoption(parser: Any) -> None:
         dest='xai_proto_include_json',
         help='Include JSON representations in xAI proto cassette YAML files.',
     )
+    parser.addoption(
+        '--debug-with-logfire',
```
DouweM, line 332: Why not just `--logfire`?

tests/conftest.py
```diff
@@ -331,14 +354,18 @@ def mock_vcr_aiohttp_content(mocker: MockerFixture):
...
-    return {
+@pytest.fixture(scope='function')
+def vcr_config(request: pytest.FixtureRequest):
+    config: dict[str, Any] = {
         'ignore_localhost': True,
         # Note: additional header filtering is done inside the serializer
         'filter_headers': ['authorization', 'x-api-key'],
         'decode_compressed_response': True,
     }
+    if request.config.getoption('debug_with_logfire'):
```
DouweM, line 365: Should have a constant for this string

DouweM (changes requested):

DouweM:
> one edge case is that running tests/test_logfire.py with the flag will produce snapshot mismatches, but that shouldn't be a concern. debugging with logfire is meant for debugging model provider API tests

@dsfaccini Could we have a fixture that we use in `test_logfire` to skip the extra layer of instrumentation?

tests/conftest.py
dsfaccini: I prefer explicit flags and I lowkey didn't want llms to confuse `--lf` with `--logfire`

tests/conftest.py
dsfaccini: when I'm looking for errors I look into the HTTP, which is why I thought default httpx instrumentation was better, but I'm not stubborn on this, lmk which default you want and whether we should have separate flags like `--instrument-pai` and `--instrument-httpx`

dsfaccini:
I'll look into it

tests/conftest.py
DouweM: I always prefer to instrument both, just gives more context into what's happening with tool calls etc

tests/conftest.py
DouweM: But all test running is basically debugging right? I think `--instrument` is at least as explicit

Kludex:
We have a proper integration in Logfire now: https://pydantic.dev/docs/logfire/integrations/pytest

I'll close this as stale, since it's not using that approach. Please discuss between yourselves if you want to add that pytest plugin to the repository.

PR #4207, opened 2026-02-05
fix(bedrock): CachePoint placement with multiple trailing documents

psg2:
## Summary

Fix bug where `CachePoint` was inserted before only the last document/video instead of before the entire trailing contiguous group of documents/videos.

## The Problem

AWS Bedrock rejects cache points that directly follow documents and videos (but not images). The previous code found the last document/video and inserted the cache point before it, but this failed when there were multiple consecutive trailing documents.

**Before (buggy):**
```
Input:  [text, Doc1, Doc2, CachePoint]
Result: [text, Doc1, cachePoint, Doc2]  ❌ AWS rejects: cachePoint after Doc1
```

**After (fixed):**
```
Input:  [text, Doc1, Doc2, CachePoint]
Result: [text, cachePoint, Doc1, Doc2]  ✅
```

## Changes

- Extract `_insert_cache_point_before_trailing_documents()` as a module-level helper function to eliminate code duplication between `_map_messages` and `_map_user_prompt`
- Fix logic to find the start of the trailing contiguous group of documents/videos instead of just the last one
- Add tests for multiple trailing documents scenarios

## Test Plan

- Added `test_bedrock_cache_point_with_multiple_trailing_documents`
- Added `test_bedrock_cache_point_with_mixed_content_and_trailing_documents`
- Added `test_bedrock_cache_messages_with_multiple_trailing_documents`
- All 81 existing Bedrock tests pass

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/bedrock.py
+++ b/pydantic_ai_slim/pydantic_ai/models/bedrock.py
@@ -169,7 +169,6 @@ def _insert_cache_point_before_trailing_documents(
     content: list[Any],
     *,
     raise_if_cannot_insert: bool = False,
-    skip_if_back_to_back: bool = False,
 ) -> bool:
     """Insert a cache point before trailing document/video content.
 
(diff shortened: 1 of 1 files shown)


pydantic_ai_slim/pydantic_ai/models/bedrock.py
```diff
@@ -165,6 +165,63 @@
 }
 
 
+def _insert_cache_point_before_trailing_documents(
+    content: list[Any],
+    *,
+    raise_if_cannot_insert: bool = False,
+    skip_if_back_to_back: bool = False,
```
DouweM: We don't need this flag if we always set it to `True`, right?

PR #4208, opened 2026-02-05
Update `UsageLimits` to preserve explicit 0 token limits instead of falling back on deprecated fields

### LINKED ISSUE #4068: UsageLimits should preserve explicit 0 token limits

Mr-Neutr0n:
## Summary

Fixes the bug where setting `input_tokens_limit=0` or `output_tokens_limit=0` would incorrectly fall back to deprecated argument values because `0` is falsy in Python.

## Problem

```python
from pydantic_ai.usage import UsageLimits

# Expected: input_tokens_limit == 0 (explicitly set)
# Actual (before fix): input_tokens_limit == 123 (deprecated fallback)
limits = UsageLimits(input_tokens_limit=0, request_tokens_limit=123)
print(limits.input_tokens_limit)  # Was: 123, Should be: 0
```

The `or` operator treats `0` as falsy:
```python
self.input_tokens_limit = input_tokens_limit or request_tokens_limit  # 0 or 123 = 123
```

## Solution

Check for `None` explicitly instead of using `or`:
```python
self.input_tokens_limit = input_tokens_limit if input_tokens_limit is not None else request_tokens_limit
```

## Changes

- `pydantic_ai_slim/pydantic_ai/usage.py`: Fixed both `input_tokens_limit` and `output_tokens_limit` assignments
- `tests/test_usage_limits.py`: Added `test_usage_limits_preserves_explicit_zero` with comprehensive test cases

## Test Plan

- [x] Added tests for:
  - Explicit `0` with deprecated fallback → should use `0`
  - Explicit `None` with deprecated fallback → should use fallback
  - Both `None` → should be `None`
  - Only new arg set → should use new arg

Fixes #4068

### DIFF
--- a/tests/test_usage_limits.py
+++ b/tests/test_usage_limits.py
@@ -464,10 +464,10 @@ def test_usage_limits_preserves_explicit_zero():
     # When input_tokens_limit=0 and deprecated request_tokens_limit is also set,
     # the explicit 0 should be preserved (not overwritten by the deprecated fallback).
     # We ignore type errors below because overloads don't allow mixing current and deprecated args.
-    limits = UsageLimits(input_tokens_limit=0, request_tokens_limit=123)  # pyright: ignore[reportDeprecated]
+    limits = UsageLimits(input_tokens_limit=0, request_tokens_limit=123)  # pyright: ignore[reportCallIssue]
     assert limits.input_tokens_limit == 0
 
-    limits = UsageLimits(output_tokens_limit=0, response_tokens_limit=456)  # pyright: ignore[reportDeprecated]
+    limits = UsageLimits(output_tokens_limit=0, response_tokens_limit=456)  # pyright: ignore[reportCallIssue]
     assert limits.output_tokens_limit == 0
 
     # When only deprecated arg is passed, should use it as fallback


tests/test_usage_limits.py
```diff
@@ -457,3 +457,28 @@ async def tool_c() -> str:
...
+    """Test that explicit 0 token limits are preserved and not replaced by deprecated fallbacks."""
+    # When new arg is 0 and deprecated arg is non-zero, should use 0 (not fallback)
+    limits = UsageLimits(input_tokens_limit=0, request_tokens_limit=123)
+    assert limits.input_tokens_limit == 0
+
+    limits = UsageLimits(output_tokens_limit=0, response_tokens_limit=456)
+    assert limits.output_tokens_limit == 0
+
+    # When new arg is None, should use deprecated fallback
+    limits = UsageLimits(input_tokens_limit=None, request_tokens_limit=123)
```
dsfaccini: lets not use `None` here, just leave empty. same for cases below

dsfaccini:
thanks for the fix, please address the comment above and fix type checking :)

Mr-Neutr0n:
Thanks for the review! I've addressed the feedback:

- Removed explicit `None` arguments in test cases, using empty arguments instead

Let me know if there are any other changes needed!

tests/test_usage_limits.py
```diff
@@ -457,3 +457,29 @@ async def tool_c() -> str:
 
 def test_usage_unknown_provider():
     assert RequestUsage.extract({}, provider='unknown', provider_url='', provider_fallback='') == RequestUsage()
+
+
+def test_usage_limits_preserves_explicit_zero():
+    """Test that explicit 0 token limits are preserved and not replaced by deprecated fallbacks."""
+    # When new arg is 0 and deprecated arg is non-zero, should use 0 (not fallback)
```
dsfaccini: why are we calling them "new" and "deprecated" args?

tests/test_usage_limits.py
```diff
@@ -457,3 +457,29 @@ async def tool_c() -> str:
...
+    # Using type: ignore since we're intentionally testing mixed deprecated/new args
+    limits = UsageLimits(input_tokens_limit=0, request_tokens_limit=123)  # type: ignore[call-overload]
+    assert limits.input_tokens_limit == 0
+
+    limits = UsageLimits(output_tokens_limit=0, response_tokens_limit=456)  # type: ignore[call-overload]
+    assert limits.output_tokens_limit == 0
```
dsfaccini: it's weird that we need ignores... the arg types are fine, right? if we really need them use `# pyright: ignore[specificError]`

dsfaccini:
hey @Mr-Neutr0n thanks for fixing, I've requested a couple more changes, which should also help with the type check issue

Mr-Neutr0n:
Thanks for the feedback! I've addressed the review comments:

1. **Changed `type: ignore` to `pyright: ignore[call-overload]`** as requested

2. **Updated terminology**: Replaced "new/deprecated" with "current/deprecated" for clarity

3. **Added clarification in comments** explaining why the ignore is needed: the type overloads don't support mixing the current arg with the deprecated arg (which is what these tests are specifically testing)

The `pyright: ignore` is necessary here because we're intentionally testing an edge case where both `input_tokens_limit=0` and `request_tokens_limit=123` are passed together - this shouldn't happen in normal usage (hence not having a type overload for it), but we need to test that when it does happen, the explicit `0` is preserved rather than being overwritten by the fallback logic.

dsfaccini:
hey @Mr-Neutr0n appreciate the explanation, I should've known the deprecated part lol

you'll need to fix the pyright ignore replacing `no-call` with `reportDeprecated`, then linting will pass

if the flaky 3.10 test fails again I'll rerun it

Mr-Neutr0n:
Done\! Changed to `reportDeprecated` - makes much more sense since we're testing the deprecated args. Thanks for catching that\!

Mr-Neutr0n:
Fixed\! The actual pyright error was `reportCallIssue` (overload mismatch), not `reportDeprecated`. Also rephrased the comment so it doesn't accidentally look like a pyright directive. CI should pass now.

tests/test_usage_limits.py
```diff
@@ -457,3 +457,30 @@ async def tool_c() -> str:
...
+    # the explicit 0 should be preserved (not overwritten by the deprecated fallback).
+    # We ignore type errors below because overloads don't allow mixing current and deprecated args.
+    limits = UsageLimits(input_tokens_limit=0, request_tokens_limit=123)  # pyright: ignore[reportCallIssue]
+    assert limits.input_tokens_limit == 0
+
+    limits = UsageLimits(output_tokens_limit=0, response_tokens_limit=456)  # pyright: ignore[reportCallIssue]
+    assert limits.output_tokens_limit == 0
+
+    # When only deprecated arg is passed, should use it as fallback
+    limits = UsageLimits(request_tokens_limit=123)  # pyright: ignore[reportCallIssue]
```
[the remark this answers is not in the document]
DouweM: @Mr-Neutr0n I think this is correct as the lint job is failing

DouweM:
@Mr-Neutr0n Also the change looks fine, but can you please have a look at my question in https://github.com/pydantic/pydantic-ai/issues/4068#issuecomment-3790639290?

Mr-Neutr0n:
Thanks for reviewing! Regarding your question in #4068 about when a token limit of 0 would be useful in practice:

You're right that a limit of 0 isn't practically useful for most production scenarios - it would just cause immediate failure. However, the fix addresses an API consistency concern: when a user explicitly passes `input_tokens_limit=0`, that explicit value should be respected rather than silently being replaced by a deprecated fallback parameter.

That said, if the maintainers feel that 0 isn't a valid value and should be rejected or handled differently, I'm happy to adjust the approach. Would you prefer:

1. Keep the current fix (explicit 0 is preserved for API consistency)
2. Add validation to reject 0 as an invalid limit value
3. Some other approach?

DouweM:
@Mr-Neutr0n I'm OK with the current fix if you fix the linting.

Also I much prefer human responses over copy-pasted AI responses.

Mr-Neutr0n:
Hey @DouweM,

I've pushed a fix - changed reportCallIssue to reportDeprecated on all four test lines. The lint job should pass now. Apologies on the responses. Was too worked up to give my own :)

Mr-Neutr0n:
Lint passes now. The test failure in `test on 3.10 (all-extras)` is a flaky SSL socket cleanup warning (`PytestUnraisableExceptionWarning`) unrelated to this PR - could you re-run the failed jobs?

dsfaccini:
reran flaky test

PR #4218, opened 2026-02-05
Fix Auto Review on forks

DouweM:
There may be more followup PRs to fix this...


sarth6:
Is there a way to retrigger this? In my case for https://github.com/pydantic/pydantic-ai/pull/4158

Issue #3959, opened 2026-01-08
`deferred_handler` for Inline Resolution of Deferred Tools


zby:
### Description

# Deferred Handler Minimal-Core Proposal Draft

## Context
We need a minimal-core `deferred_handler` proposal that preserves blocking approvals while keeping policy and UI in third-party packages. This draft should enable replacement of `pydantic-ai-blocking-approval` without expanding core beyond the loop mechanics.

## Findings
Draft proposal content:

---

# Proposal: `deferred_handler` for Inline Resolution of Deferred Tools (Minimal Core)

## Summary

Add an optional `deferred_handler` parameter to `Agent` and `agent.run()` that enables inline resolution of deferred tool calls (approvals and external calls) without returning control to the caller.

The core stays minimal: it only defines how deferred tool requests are surfaced to a handler and how results are applied. Approval policy, UI, caching, and "blocked" semantics remain in third-party packages.

## Goals

- Enable blocking, inline approval workflows without re-implementing the agent loop.
- Preserve current behavior when no handler is provided.
- Keep approval policy and UI in third-party packages.

## Non-Goals

- Define approval policy rules in core.
- Add new approval UI or caching APIs in core.
- Support streaming inline approvals in `run_stream()` (initially).

## Motivation

PydanticAI's current deferred tools pattern works well for batch/review workflows where tool calls are collected, reviewed externally (dashboard, email, async process), and then resumed with a new `agent.run()` call.

However, for interactive scenarios (CLI agents, coding assistants, multi-step dangerous operations), a blocking pattern is more appropriate:

1. The agent proposes tool calls
2. The user reviews and approves/denies inline
3. Approved tools execute immediately
4. The agent sees results and continues
5. Repeat until task complete

Currently, implementing this requires:
- Re-implementing the agent loop externally
- Managing message history across multiple `run()` calls
- Complex state handling for multi-agent scenarios (see #3274)

A first-class `deferred_handler` makes this pattern easy while keeping approval policy out of core.

### API

```python
from typing import Callable, Awaitable
from pydantic_ai import Agent
from pydantic_ai.tools import DeferredToolRequests, DeferredToolResults

DeferredHandler = Callable[
    [DeferredToolRequests],
    DeferredToolResults | Awaitable[DeferredToolResults]
]

# On Agent constructor
agent = Agent(
    'openai:gpt-5',
    deferred_handler=my_handler,  # Optional, can be overridden per-run
)

# On run methods
result = await agent.run(
[...]

PR #4233, opened 2026-02-06
Pydantic AI v2 Traits API: Research Report & Conceptual Design

DouweM:
By Opus 4.6

### DIFF
--- a/traits-research-report.md
+++ b/traits-research-report.md
@@ -1,0 +1,1059 @@
+# Pydantic AI v2 Traits API: Research Report & Conceptual Design
+
+> **Prepared for**: DouweM / pydantic-ai core team
+> **Date**: February 2026
+> **Scope**: Deep research on community projects, external frameworks, and conceptual design for a Traits API
+
+---
+
+## Table of Contents
+
+1. [Framework Reports](#1-framework-reports)
+   - [1A. Community Pydantic AI Projects](#1a-community-pydantic-ai-projects)
+   - [1B. External Frameworks](#1b-external-frameworks)
+2. [Comparison Matrix](#2-comparison-matrix)
+3. [Conceptual Traits API Proposal](#3-conceptual-traits-api-proposal)
+4. [10-Line Agent Example](#4-10-line-agent-example)
+5. [Migration Path](#5-migration-path)
+
+---
+
+## 1. Framework Reports
+
+### 1A. Community Pydantic AI Projects
+
+#### Code Puppy (`mpfaffenberger/code_puppy`)
+
+**Overview**: A CLI-based agentic code generation assistant ("the sassy AI code agent that makes IDEs look outdated") built entirely on pydantic-ai. Terminal-first, supports 65+ model providers, strong personality with strict coding discipline.
+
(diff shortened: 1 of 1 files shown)


dsfaccini:
have you looked at https://github.com/letta-ai/agent-file?tab=readme-ov-file#what-state-does-af-include ?

could be used as a guide or sanity check for our own

DEENUU1:
**Todo**
- Will TodoTrait have a pluggable storage protocol so users can swap in-memory for Postgres/Redis/etc? We support that in https://github.com/vstorm-co/pydantic-ai-todo 
- Can todos be shared with subagents? E.g., parent creates a plan, subagent picks up a task and marks it done. Would SubAgentTrait get read/write access to the parent's todo state?

DEENUU1:
**Subagents**
- Will subagents be strictly pre-defined, or could the main agent also spawn new ones at runtime with custom instructions/tools?
    - Will subagents support background/async execution? Something similar to what Claude Code does with https://code.claude.com/docs/en/agent-teams e.g. kicking off parallel research tasks and checking results later.

traits-research-report.md
```diff
@@ -0,0 +1,1059 @@
...
+
+    async def on_agent_start(self, ctx: RunContext[AgentDepsT]) -> None:
+        """Called before the agent run begins."""
+        pass
+
+    async def on_agent_end(self, ctx: RunContext[AgentDepsT]) -> None:
+        """Called after the agent run completes."""
+        pass
+
+    async def before_model_request(
```
VMinB12, line 539: suggestion: on top of (or instead of) `before_model_request` and `after_model_request`, you could add `wrap_model_request`, e.g.
```python
async def wrap_model_request(
    self,
    ctx: RunContext[AgentDepsT],
    handler: Callable[[RunContext[AgentDepsT]], Awaitable[ModelResponse[AgentOutputT]]],
) -> ModelResponse[AgentOutputT]:
   ...
```

An example where this is useful is handling errors:
```python
class ErrorFallbackTrait(Trait):
    def __init__(self, fallback_model: Model):
        self.fallback_model = fallback_model

    async def wrap_model_request(
        self,
        ctx: RunContext[AgentDepsT],
        handler: Callable[[RunContext[AgentDepsT]], Awaitable[ModelResponse[AgentOutputT]]],
    ) -> ModelResponse[AgentOutputT]:        
        try:
            return await handler(ctx)
        except Exception as e:
            return await handler(override(ctx, model=self.fallback_model))
```
which is not possible with just before and after hooks. I don't think `Request` exists yet as a class, but it should contain everything relevant to the model invocation, i.e. the `run_context`, `model`, `messages`, etc.

The same applies to `wrap_tool_call`.

VMinB12 (commented):
I'm very excited about traits!

zby:
For us the crucial part is that before_tool_call must receive RunContext[Deps]. This could be a more general solution than https://github.com/pydantic/pydantic-ai/issues/3959

But it would be helpful if https://github.com/pydantic/pydantic-ai/issues/4347 is resolved first.

PR #4240, opened 2026-02-06
Add Pinchwork marketplace example

pinchwork:
Hi! 👋

Adds an example showing agent-to-agent task delegation with Pinchwork.

**What it shows:**
- Agents delegating specialized work to other agents
- Type-safe marketplace tools
- Async/await patterns

**Why this fits:**
Pydantic AI's type safety + multi-agent coordination = natural use case for agent marketplaces.

Example is ~100 lines, follows the weather_agent.py style.

**Links:**
- Pinchwork: https://pinchwork.dev (MIT licensed)
- Works with: `pip install pinchwork`

Happy to adjust anything! 🦞

### DIFF
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -558,11 +558,6 @@ text_responses: dict[str, str | ToolCallPart | Sequence[ToolCallPart]] = {
         args={'name': 'test', 'value': 42},
         tool_call_id='pyd_ai_tool_call_id',
     ),
-    'Delegate a Python code review task to the marketplace. Offer 15 credits.': ToolCallPart(
-        tool_name='delegate_task',
-        args={'need': 'Review Python code for security issues', 'max_credits': 15, 'tags': ['python', 'security']},
-        tool_call_id='pyd_ai_tool_call_id',
-    ),
 }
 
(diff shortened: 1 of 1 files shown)


pinchwork:
Good catch! I'll add a `test: skip` marker since this example requires an external API key and active marketplace.

Will push a fix shortly. Thanks for the detailed review! 🦞

pinchwork:
Updated with proper test infrastructure (21621830):

- Added mock responses to `text_responses` (prompt → ToolCallPart)
- Added tool response to `tool_responses` (delegate_task result)
- Example now gets tested instead of skipped

Follows the pattern of weather_agent and other tool-based examples. Thanks for catching this! 🦞

pinchwork:
You're absolutely right! Examples in `examples/` aren't tested by the test suite, only docs snippets are.

Fixed properly (f0ece6f7):
- Made example **self-contained with mock marketplace**
- No external API required - runs standalone
- Follows pattern of `bank_support.py` (which mocks SQLite)
- Added note at end pointing to real integration

Much cleaner approach than test mocks for untested code. Thanks for the thorough review! 🦞

DouweM:
This does not yet meet our popularity bar for inclusion in the core library.

PR #4247, opened 2026-02-07
fix: decode base64 file content per A2A spec

### LINKED ISSUE #4176: A2A file content handling does not follow the spec (fix included)

thecaptain789:
Fixes #4176

The A2A spec requires file bytes to be base64-encoded. The current implementation treats the bytes field as plain text and calls `.encode('utf-8')`, which doesn't correctly decode base64 data.

This PR:
- Adds `import base64` 
- Changes `file_content['bytes'].encode('utf-8')` to `base64.b64decode(file_content['bytes'])`
- Updates the test to use base64-encoded data ('Zm9v' = base64 of 'foo')

### DIFF
--- a/.github/workflows/pr-auto-label.yml
+++ b/.github/workflows/pr-auto-label.yml
@@ -76,11 +76,6 @@ jobs:
     if: github.event.action == 'opened'
     runs-on: ubuntu-latest
     timeout-minutes: 10
-    permissions:
-      id-token: write
-      contents: read
-      issues: write
-      pull-requests: write
     steps:
       - name: Check for modified config files
         if: github.event.pull_request.head.repo.fork
@@ -116,6 +111,7 @@ jobs:
         uses: anthropics/claude-code-action@v1
         with:
           anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
+          github_token: ${{ secrets.GITHUB_TOKEN }}
           allowed_non_write_users: '*'
           claude_args: |
             --allowedTools "Bash(gh pr view:*),Bash(gh pr diff:*),Bash(gh api repos/${{ github.repository }}/issues/${{ github.event.pull_request.number }}/labels:*)"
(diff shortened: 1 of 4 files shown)


DouweM:
Duplicate of https://github.com/pydantic/pydantic-ai/pull/4181

PR #4249, opened 2026-02-07
Fix retry message suggesting 'call a tool' when no tools are registered

### LINKED ISSUE #4248: Retry message suggests 'call a tool' when no tools are registered
When `CallToolsNode._run_stream` receives a model response with no tool calls, the `else` branch unconditionally appends `'call a tool'` to the retry alternatives:

```python
if tool_calls:
        yield event
    return
elif output_schema.toolset:
else:
    alternatives.append('call a tool')  # ← always added, even with zero tools
```

If an agent has `output_type=str` and no function tools registered, the retry message becomes `"Please return text or call a tool."` — but there are no tools for the model to call. This can cause the model to attempt tool calls that don't exist, leading to unnecessary retries and wasted tokens before it falls back to returning text.

The retry message should only suggest `"call a tool"` when function tools are actually available. For a tool-less agent, the message should be `"Please return text."`.

veeceey:
- Closes #4248

### Pre-Review Checklist

### Pre-Merge Checklist

### Summary

When `CallToolsNode._run_stream` receives a model response with no tool calls, the retry message unconditionally included `call a tool` as an alternative, even when no function tools or builtin tools are registered.

For an agent with `output_type=str` and no tools, this produced: `Please return text or call a tool.` -- but there are no tools for the model to call, leading to unnecessary retries and wasted tokens.

### Changes

- **`pydantic_ai_slim/pydantic_ai/_agent_graph.py`**: Changed the unconditional `else` branch to an `elif` that checks whether builtin tools or function tools are actually available before suggesting `call a tool`.
- **`tests/test_agent.py`**: Updated the existing `test_thinking_only_response_retry` snapshot to reflect the corrected message (`Please return text.`). Added `test_thinking_only_response_retry_with_tools` to verify that the `call a tool` alternative still appears when function tools are registered.
- **`tests/models/test_google.py`** and its **cassette**: Updated the `test_google_image_generation_with_native_output` snapshot and VCR cassette to reflect the corrected message for a tool-less agent.

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/_agent_graph.py
+++ b/pydantic_ai_slim/pydantic_ai/_agent_graph.py
@@ -707,7 +707,10 @@ class CallToolsNode(AgentNode[DepsT, NodeRunEndT]):
                         return
                     elif output_schema.toolset:
                         alternatives.append('include your response in a tool call')
-                    else:
+                    elif ctx.deps.builtin_tools or (
+                        ctx.deps.tool_manager.tools is not None
+                        and any(t.tool_def.kind != 'output' for t in ctx.deps.tool_manager.tools.values())
+                    ):
                         alternatives.append('call a tool')
 
                     if output_schema.allows_image:
--- a/tests/models/test_google.py
+++ b/tests/models/test_google.py
@@ -3614,7 +3614,7 @@ async def test_google_image_generation_with_native_output(allow_model_requests: 
             ModelRequest(
                 parts=[
                     RetryPromptPart(
-                        content='Please return text or call a tool.',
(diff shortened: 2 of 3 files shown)


adtyavrdhn:
Hi, I haven't taken a look but it seems like there is another PR #4234. If you're not targeting something different in the PR then please add your thoughts there instead?

veeceey:
Closing as duplicate of #4234 which addresses the same issue (#4248). Both PRs implement the same fix to prevent suggesting 'call a tool' when no tools are registered. Please coordinate with @Baniskos on #4234 to get that PR merged.

PR #4329, opened 2026-02-16
chore: tweak autoreview behavior

dsfaccini:


.github/workflows/bots.yml
```diff
@@ -150,10 +148,7 @@ jobs:
     needs: [size-label, category-label]
     if: >-
       !failure() && !cancelled() &&
-      (
-        (github.event.action != 'labeled' && contains(github.event.pull_request.labels.*.name, 'auto-review')) ||
-        (github.event.action == 'labeled' && github.event.label.name == 'auto-review')
-      )
+      github.event.action == 'labeled' && github.event.label.name == 'auto-review'
```
DouweM, line 151: Now it'll never run on sync anymore will it? I guess that doesn't matter anyway since we made it label based. But then we can remove "synchronize" at the top as well?

dsfaccini (commented):
From Claude:

We still need `synchronize` for the size-label and category-label jobs — they both have `if: github.event.action != 'labeled'`, so they run on `opened` and `synchronize` to re-calculate labels when new commits are pushed.

Issue #4167, opened 2026-02-02
Anthropic tool search


DouweM:
### Description

- Native support: https://github.com/pydantic/pydantic-ai/issues/3590
- Will be able to use provider support if available

More context: https://github.com/pydantic/pydantic-ai/issues/3666#issuecomment-3629161845

lilyTokens:
the agent evaluation dimensions we've found most predictive: trajectory efficiency (steps per task vs minimum viable) and error recovery rate (separate from task completion). task completion alone misses a lot. wrote this up: https://agentskills-app.surge.sh/agent-evaluation.html

Wh1isper:
And also OpenAI's tool search? https://developers.openai.com/api/docs/guides/tools-tool-search

In my view, this depends on how vendors manage the Context like PTC. I implemented a similar [toolset](https://github.com/Wh1isper/ya-agent-sdk/pull/39) at the application layer, but it [caused cache invalidation](https://platform.claude.com/docs/en/build-with-claude/prompt-caching#what-invalidates-the-cache) since the tool definition changed.

Wh1isper:
Adding another thought on why I'd likely stick with an application-layer implementation rather than relying on vendor-native tool search:

**Session resumability beyond ephemeral server-side state.**

Consider this scenario: a user has dozens of MCP servers connected. During a conversation, the model dynamically discovers and loads a subset of tools via native tool search. This works great within a single session. But when the user comes back days later to resume the conversation, the server-side state that tracked which tools were "loaded" is likely gone. There's no guarantee the vendor preserves this across arbitrary time gaps — it's similar to how a code execution sandbox expires and you lose access to the VM's filesystem.

With an application-layer implementation, I can persist the loaded tool set (which tools/namespaces were activated) in my own session state and deterministically restore it when the user resumes — regardless of how the vendor manages its internal state. The SDK owns the state, not an ephemeral server-side mechanism.

So while native tool search is great for single-session cost/latency optimization, it doesn't address the durability problem for resumable, long-lived agent sessions.

DouweM:
> But when the user comes back days later to resume the conversation, the server-side state that tracked which tools were "loaded" is likely gone.

@Wh1isper I don't think that's true: 

OpenAI encodes this in the message history per https://developers.openai.com/api/docs/guides/tools-tool-search#hosted-tool-search, which we'd store as `BuiltinToolCallPart`/`BuiltinToolReturnPart`:

[...]

PR #4090, opened 2026-01-26
Add `defer_loading` to tools and toolsets to enable tool search

dsfaccini:
- Closes #3590
- Related to #4167

Hey guys, opening this with a plan after reading the previous issues and PRs so we can be clear on the specs of this before launching into the code.

@DouweM we recently talked about the beta flag so I'm including that here as well.

PR #4342, opened 2026-02-17
Add builtin-to-custom tool fallback via `prefer_builtin`

### LINKED ISSUE #3212: Let built-in tools fallback on custom tools for models that don't support them
### Description

[Built-in tools](https://ai.pydantic.dev/builtin-tools) are great, but specifying them on an agent currently limits you to using it only with models that support all of them, which makes it hard to use [`FallbackModel`](https://ai.pydantic.dev/models/overview/#fallback-model) or try different models in evals.

It'd be great if you could specify a function tool (or toolset) to be used as a fallback when a built-in tool isn't available:
- `WebSearchTool` -> Tavily or DuckDuckGo from https://ai.pydantic.dev/common-tools/
- `CodeExecutionTool` -> https://github.com/pydantic/mcp-run-python
- `ImageGenerationTool` -> custom tool that calls an image model
- `MemoryTool` -> e.g. https://github.com/pydantic/pydantic-ai/pull/3129 (This already does something somewhat similar in that it requires a `memory` tool to be defined)
- `UrlContextTool` -> custom tool that runs `requests.get` and renders HTML to Markdown or something
- `MCPServerTool` from https://github.com/pydantic/pydantic-ai/pull/3101 -> `MCPServer` from https://ai.pydantic.dev/mcp/client/ or `FastMCPToolset` from https://github.com/pydantic/pydantic-ai/pull/2784

I imagine something like `@agent.tool(fallback_for_builtin=WebSearchTool)` (and similar on the `Tool` class). That would then be stored on the `ToolDefinition`, so that `Model.prepare_request` can look at the `ModelRequestParameters.function_tools` and `.builtin_tools`, check which are supported by the model in question (via new properties on `self.profile: ModelProfile` most likely), and then keep only one or the other. We'd only raise an error for unsupported built-in tools if no fallback was provided.

To support falling back from a built-in tool to a toolset (like with `MemoryTool` or `MCPServerTool`), we'd need a way to set this on a toolset as well, most likely when registering it on an agent rather than hard-coded on the toolset itself. With `MCPServerTool`, we'd also need to consider that a given `MCPServer`/`FastMCPToolset` would be a fallback for a _specific_ `MCPServerTool` instance, not all `MCPServerTool`s in general.

### References

_No response_

dsfaccini:
- Closes #3212

### Pre-Review Checklist

### Pre-Merge Checklist

---

## Plan for Review

This PR currently contains only a `PLAN.md` describing the proposed implementation. Looking for feedback on the approach before writing code.

## Summary

When using `FallbackModel` or testing across models, agents with builtin tools (e.g. `WebSearchTool`) fail with `UserError` on models that don't support them. This adds automatic fallback to user-provided function tools.

**Core insight**: send BOTH the builtin tool AND its fallback function tool in `ModelRequestParameters`. Each model's `prepare_request` independently decides which to keep via `self.profile.supported_builtin_tools`. Zero changes to `FallbackModel` — each inner model makes the right choice.

## Proposed Changes

1. **`ToolDefinition`** — add `prefer_builtin: str | None` field matching `AbstractBuiltinTool.unique_id`
2. **`Tool`** — add `prefer_builtin: AbstractBuiltinTool | None` constructor param, resolve to `unique_id`
3. **`Model.prepare_request`** — partition builtins into supported/unsupported, swap with fallback function tools, error only when no fallback exists
4. **Thread through decorators** — `Agent.tool()`, `Agent.tool_plain()`, `FunctionToolset.tool()`, `FunctionToolset.add_function()`

## What does NOT change

- `FallbackModel`, `_agent_graph.py`, `AbstractToolset`/`ToolManager`, provider model implementations, `AbstractBuiltinTool`, `ModelProfile`

## Deferred

- `AbstractBuiltinTool(fallback=...)`, `MCPServer`/`FastMCPToolset` support, `TavilySearchTool(prefer_builtin=True)`, `BuiltinToolset`

See `PLAN.md` for full details.

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -6,6 +6,8 @@ When using `FallbackModel` or testing across models, agents with builtin tools (
 
 **Core insight**: send BOTH the builtin tool AND its fallback function tool in `ModelRequestParameters`. Each model's `prepare_request` independently decides which to keep via `self.profile.supported_builtin_tools`. Zero changes to `FallbackModel` — each inner model makes the right choice.
 
+**Important**: `prefer_builtin` on a function tool does NOT auto-register the corresponding builtin tool with the agent. The user must still include it in `builtin_tools=[...]`. This is by design — the upcoming `capabilities=[WebSearch()]` API (see Relationship to Deferred Loading below) will handle both sides automatically via `get_toolset()` + `get_builtin_tools()`. This PR provides the low-level primitive that capabilities will build on.
+
 ## Changes
 
 ### 1. `ToolDefinition` — add `prefer_builtin` field
@@ -77,12 +79,21 @@ if params.builtin_tools:
 - `AbstractBuiltinTool` — no `fallback` param (deferred)
 - `ModelProfile` — already has `supported_builtin_tools`
 
+## Relationship to deferred tool loading (#4090, #4167, #3666)
+
+DouweM notes that deferred tool loading (Anthropic tool search) needs the ability to send fundamentally different tool lists depending on model support — e.g. all tools with `defer_loading=True` for native support vs a `search_tools` meta-tool + already-discovered tools for framework-level support. `prefer_builtin` alone doesn't cover that case because it's a 1:1 swap (one function tool ↔ one builtin), not a structural transformation of the entire tool list.
+
+**Why this is fine for our scope**: `prefer_builtin` is the right primitive for builtin-to-custom fallback (web search, code execution, etc.). Deferred tool loading will need an additional mechanism (likely `defer_loading: bool` on `ToolDefinition` + model-level list restructuring), which is a separate feature. These are complementary, not conflicting — both follow the same architectural pattern of model-level resolution in `prepare_request`.
+
+**Forward compatibility**: the `capabilities=[WebSearch()]` API will compose both: `WebSearch.get_toolset()` returns function tools with `prefer_builtin` set, `WebSearch.get_builtin_tools()` returns the builtin. For deferred loading, `SearchableToolset` will add its own fields. Our `prepare_request` swap logic is additive and won't conflict with future deferred loading logic.
+
 ## Deferred to follow-up PRs
 
 - `AbstractBuiltinTool(fallback=...)` — builtin carries its own fallback
 - `MCPServer`/`FastMCPToolset` `prefer_builtin` — toolset-level support
 - `TavilySearchTool(prefer_builtin=True)` — convenience on common tools
 - `BuiltinToolset` — wrapping builtins in the toolset system
+- Deferred tool loading / `SearchableToolset` (#4090, #4167)
 
 ## Testing
 


PLAN.md
```diff
@@ -0,0 +1,105 @@
...
+prefer_builtin: str | None = None
+```
+Matches `AbstractBuiltinTool.unique_id` (e.g. `'web_search'`, `'code_execution'`). Signals: 'if model supports this builtin, prefer it over me'.
+
+### 2. `Tool` — add `prefer_builtin` param
+**File**: `pydantic_ai_slim/pydantic_ai/tools.py`
+
+- Store field: `prefer_builtin: str | None` (resolved from the input)
+- Add `__init__` param: `prefer_builtin: AbstractBuiltinTool | BuiltinToolFunc[AgentDepsT] | None = None`
+  - `BuiltinToolFunc` (already at `tools.py:129`) is `Callable[[RunContext[AgentDepsT]], Awaitable[AbstractBuiltinTool | None] | AbstractBuiltinTool | None]`
```
DouweM, line 27: It will need to be clear that including `prefer_builtin` on `Tool`/`@agent.tool` will not automatically register that builtin tool with the agent - it still needs to be included in `builtin_tools`.

The user will not need to worry about this if they use the upcoming `capabilities=[WebSearch()]` where the `WebSearch` capability will implement both `get_toolset()` to return e.g. `DuckDuckGoSearchTool` with `ToolDefinition.prefer_builtin=builtin_tools.WebSearchTool.kind` + `get_builtin_tools()` to return `builtin_tools.WebSearchTool()`. (Similar you can imagine `capabilities.MCP(..., prefer_builtin=True)` that will use either our `MCPServer`, or the builtin `MCPServerTool`)

So we may want to start by just implementing the "guts" of this feature on `ToolDefinition` + `Model.prepare_request`, without yet implementing the user-facing bit on `Tool`/`@agent.tool`, since this will primarily be used by capabilities that have local and builtin equivalents.

DouweM: It's also worth thinking about how this relates to #4090 and #4167 and https://github.com/pydantic/pydantic-ai/issues/3666#issuecomment-3629161845. There I imagined we'd need the way to send one list of `ToolDef`s if builtin tool search _is_ supported (all with `defer_loading=True`) and one when it is _not_ supported (the `search_tools` tool + those already found).

Just `prefer_builtin` wouldn't cut it in that case.

PLAN.md
```diff
@@ -0,0 +1,97 @@
...
+- Provider model implementations — they call base `prepare_request`
+- `AbstractBuiltinTool` — no `fallback` param (deferred)
+- `ModelProfile` — already has `supported_builtin_tools`
+
+## Deferred to follow-up PRs
+
+- `AbstractBuiltinTool(fallback=...)` — builtin carries its own fallback
+- `MCPServer`/`FastMCPToolset` `prefer_builtin` — toolset-level support
+- `TavilySearchTool(prefer_builtin=True)` — convenience on common tools
+- `BuiltinToolset` — wrapping builtins in the toolset system
```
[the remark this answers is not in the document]
DouweM: I agree we don't need these things right now, we can handle that better with capabilities: https://github.com/pydantic/pydantic-ai/pull/4342/changes#r2885819472

PLAN.md
dsfaccini: claude: lets research this and update the plan before starting implementation, we'll iterate on the plan a few times first

dsfaccini:
superseded by #4770

PR #4357, opened 2026-02-18
Bump `pillow`, `pip` because of Dependabot alerts

dsfaccini:
## Summary

- Bump `pillow` from `>=11.0.0` to `>=12.1.1` (fixes out-of-bounds write when loading PSD images)
- Bump `pip` from `>=25.2` to `>=26.0` (fixes path traversal vulnerability)
- Remove unused dev dependency `mcp-run-python` (only referenced in docs, not imported or used in any test)
- Upgrade transitive lockfile deps: `cryptography` 46.0.3→46.0.5, `protobuf` 6.33.2→6.33.5, `langchain-core` 1.2.7→1.2.13, `python-multipart` 0.0.21→0.0.22
- Guard test imports that require optional deps (`mcp`, `starlette`)

Addresses Dependabot alerts: #35, #38, #39, #40, #41, #42, #43, #44.

Remaining open alert (#45 `diskcache`) has no upstream patch available — `diskcache` is a transitive dep from `outlines`.

### Note on test import fixes

Removing `mcp-run-python` from dev deps exposed a latent issue: `test_mcp.py`, `test_ui.py`, and `test_vercel_ai.py` had top-level imports of `pydantic_ai.mcp` and `pydantic_ai.ui` that crash when `mcp`/`starlette` aren't installed. These were previously masked because `mcp-run-python` transitively pulled in both packages.

The fix moves those imports into existing `try_import()` blocks and adds `skipif` markers to `pytestmark` so the slim CI environment (which only installs dev deps, not optional extras) properly skips these tests instead of crashing on import.

## Test plan

- [x] `tests/test_examples.py` passes (681 passed, 98 skipped)
- [x] `tests/test_mcp.py`, `tests/test_ui.py`, `tests/test_vercel_ai.py` pass (844 passed total, 98 skipped)
- [ ] CI passes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,7 +102,7 @@ jobs:
         && 'depot-ubuntu-24.04-4'
         || 'ubuntu-latest'
       }}
-    timeout-minutes: 15
+    timeout-minutes: 20
     strategy:
       fail-fast: false
       matrix:
@@ -163,7 +163,7 @@ jobs:
         && 'depot-ubuntu-24.04-4'
         || 'ubuntu-latest'
       }}
-    timeout-minutes: 15
+    timeout-minutes: 20
     strategy:
       fail-fast: false
       matrix:


tests/test_ui.py
```diff
@@ -45,15 +45,15 @@
...
 from pydantic_ai.toolsets import AbstractToolset, ExternalToolset
-from pydantic_ai.ui import NativeEvent, UIAdapter, UIEventStream
 
 from ._inline_snapshot import snapshot
-from .conftest import try_import
 
-with try_import() as starlette_import_successful:
-    from starlette.requests import Request
-    from starlette.responses import StreamingResponse
+pytest.importorskip('starlette')
```
DouweM, line 51: Why not keep it the way we had it? I like this, for what it's worth, but we haven't used it anywhere else.

Devin pointed out the old code wasn't actually working because `not starlette_import_successful` below needs `()`

dsfaccini: yeahp that makes sense, but I'm starting to use this new `importorskip` pattern elsewhere too, suggested by claude of course

dsfaccini: `try_import()` + `pytestmark.skipif` can't work here because `DummyUIAdapter(UIAdapter[...])` is a module-level class definition. Class bodies evaluate during module **collection**, before pytest applies skip markers — so `UIAdapter` is undefined → `NameError`.

`pytest.importorskip()` raises `pytest.skip()` during collection itself, skipping the entire module before any class bodies are evaluated.

`test_vercel_ai.py` doesn't have this problem since it doesn't define classes that inherit from starlette-dependent types at module level.

dsfaccini: claude's assessment &uarr;

PR #4380, opened 2026-02-20
fix: disable `supports_json_schema_output` for Bedrock Anthropic models

### LINKED ISSUE #4368: Bedrock anthropic models incorrectly marked as supporting json schema output
### Initial Checks

### Description

There's a bug currently where models like bedrock_us.anthropic.claude-sonnet-4-6 return True for supports_json_schema_output because of the [anthropic model profile](https://github.com/pydantic/pydantic-ai/blob/main/pydantic_ai_slim/pydantic_ai/profiles/anthropic.py#L8), but models/Bedrock.py doesn't actually support native structured output silently.

I'm seeing model's hallucinate a schema since my code uses supports_json_schema_output to determine whether to use native output, but then the model call doesn't actually pass any schema to the model provider.

```Python
from unittest.mock import Mock

  client = Mock()
  client.meta.endpoint_url = "https://bedrock-runtime.us-east-1.amazonaws.com"

  provider = BedrockProvider(bedrock_client=client)
  profile = provider.model_profile("us.anthropic.claude-sonnet-4-5-20250929-v1:0")

  print(profile.supports_json_schema_output)  # should be False for Bedrock
```

### Logfire Trace

_No response_

  - Python: 3.12.12
  - Pydantic AI: main
  - LLM provider SDK: Bedrock boto3 1.42.14

TopangaLudwitt:
## Summary

Fix Bedrock Anthropic models incorrectly reporting `supports_json_schema_output=True`.

## Problem

The base Anthropic model profile sets `supports_json_schema_output=True` for supported models. When this profile is merged into the Bedrock provider via `.update()`, the flag carries over — but Bedrock's Converse API doesn't actually support native JSON schema output for Anthropic models.

This causes downstream code that checks `supports_json_schema_output` to attempt native structured output, but the schema is never passed to the model provider, resulting in hallucinated schemas.

## Fix

Extract the Bedrock Anthropic profile logic into a dedicated `_bedrock_anthropic_model_profile()` function that:
1. Merges the base Anthropic profile (for thinking tags, etc.)
2. Explicitly sets `supports_json_schema_output = False` after the merge

This follows the same pattern as `bedrock_amazon_model_profile()` and `bedrock_deepseek_model_profile()` which also customize profiles for their Bedrock context.

## Files Changed

- `pydantic_ai_slim/pydantic_ai/providers/bedrock.py`

Fixes #4368

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/providers/bedrock.py
+++ b/pydantic_ai_slim/pydantic_ai/providers/bedrock.py
@@ -68,6 +68,23 @@ def bedrock_deepseek_model_profile(model_name: str) -> ModelProfile | None:
     return profile  # pragma: no cover
 
 
+def _bedrock_anthropic_model_profile(model_name: str) -> ModelProfile | None:
+    """Get the model profile for an Anthropic model used via Bedrock.
+
+    Bedrock's Converse API does not support native JSON schema output for
+    Anthropic models, even though the direct Anthropic API does. We merge
+    the base Anthropic profile but explicitly disable JSON schema output.
+    """
+    profile = BedrockModelProfile(
+        bedrock_supports_tool_choice=True,
+        bedrock_send_back_thinking_parts=True,
+        bedrock_supports_prompt_caching=True,
+        bedrock_supports_tool_caching=True,
+    ).update(_without_builtin_tools(anthropic_model_profile(model_name)))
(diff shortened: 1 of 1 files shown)


DouweM:
Duplicate of https://github.com/pydantic/pydantic-ai/pull/4369

PR #4386, opened 2026-02-20
Add OpenTelemetry instrumentation to realtime models

Kludex:
## Summary

- Add session-level OTel span for realtime sessions (`realtime {model_name}`) with gen_ai semantic convention attributes
- Add `system` property to `RealtimeModel` ABC with provider overrides (`google`, `openai`)
- Create `InstrumentedRealtimeModel` wrapper in `realtime/instrumented.py` following the `InstrumentedModel` pattern
- Wire into `Agent.realtime_session()` respecting `instrument` / `instrument_all()` settings
- Extract shared test fakes to `conftest.py` and add 11 instrumentation tests

## Test plan

- [x] All 65 realtime tests pass (including 11 new instrumentation tests)
- [x] All 58 existing logfire tests pass (no regressions)
- [x] pyright: 0 errors on all changed files
- [x] ruff: clean
- [ ] Verify `realtime gemini-live-2.5-flash-native-audio` span appears in Logfire with Roberto webcam example

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/agent/__init__.py
+++ b/pydantic_ai_slim/pydantic_ai/agent/__init__.py
@@ -1688,6 +1688,15 @@ class Agent(AbstractAgent[AgentDepsT, OutputDataT]):
                 except Exception as e:
                     return f'Error: {e}'
 
+            # Instrument the realtime model if configured
+            from ..realtime.instrumented import instrument_realtime_model
+
+            instrument = self.instrument
+            if instrument is None:
+                instrument = self._instrument_default
+            if instrument:
+                model = instrument_realtime_model(model, instrument)
+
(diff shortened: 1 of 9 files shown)


pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -1688,6 +1688,15 @@ async def tool_runner(name: str, args: dict[str, Any]) -> str:
                 except Exception as e:
                     return f'Error: {e}'
 
+            # Instrument the realtime model if configured
+            from ..realtime.instrumented import instrument_realtime_model
```
Kludex, line 1692: Is the local import necessary here? (I think it is).

PR #4396, opened 2026-02-21
Remove `ask` permissions from Claude Code settings

dmontagu:
## Summary

- Removes the `ask` permissions section from `.claude/settings.json`, leaving only the `allow` list

## Test plan

- N/A — settings-only change


dmontagu:
I think we should probably just remove this file entirely and let users rely on (uncommitted) `settings.local.json`. But I want to at least merge this.

PR #4401, opened 2026-02-22
fix: use non-greedy regex in `strip_markdown_fences` to respect fence boundaries

### LINKED ISSUE #4397: Greedy Regex in strip_markdown_fences()
### Initial Checks

### Description

The .* quantifier is greedy — with re.DOTALL, it matches from the first { all the way to the last } in the entire text, crossing past the closing ``` fence.

Example

  An LLM returns:  ```json
  {"result": "pass"}
  This conforms to the schema {"type": "object"}

  **Expected extraction:** `{"result": "pass"}`

  **Actual extraction:** `{"result": "pass"}\n` ``` `\nThis conforms to the schema {"type": "object"}`  the greedy `.*` matched from the first `{` to the last `}` in the entire string, gobbling up the closing fence and
 everything after it.

 This produces **malformed JSON** that fails downstream validation, potentially triggering unnecessary retries or errors.

The bug only manifests when there's a } character after the closing ```     fence. That's row 1 — the only case where broken and expected differ:

Input: ```json\n{"a": 1}\n``` + Context: {"b": 2}
Current (broken): {"a": 1}\n``` \nContext: {"b": 2}
Expected: {"a": 1}

Input: ```json\n{"nested": {"k": "v"}}\n```
Current (broken): {"nested": {"k": "v"}}
Expected:: {"nested": {"k": "v"}}

Input: ```json\n{\n  "foo": "bar"\n} (no closing fence)
Current (broken): {\n  "foo": "bar"\n}
Expected:: {\n  "foo": "bar"\n}

```Python
from pydantic_ai._utils import strip_markdown_fences                        
                                           
  # LLM returns JSON in a markdown fence, followed by commentary containing   
  braces                                                                      
  text = '\n{"result": "pass"}\n\nThis matches schema {"type":
  "object"}'

  extracted = strip_markdown_fences(text)
  print(repr(extracted))
  # Actual:   '{"result": "pass"}\n\nThis matches schema {"type":
  "object"}'
  # Expected: '{"result": "pass"}'

  The greedy .* in the regex r'(?:\w+)?\n(\{.*\})' matches from the first {
   to the **last** } in the entire string, crossing past the closing  fence
   and capturing the commentary text as part of the "JSON".
```

### Logfire Trace

N/A — this bug is in the strip_markdown_fences() regex utility function        (_utils.py:517), not in a model request or agent run. It can be reproduced with a direct function call without any LLM interaction.                        

- Python: 3.12+
- Pydantic AI: main branch (latest)
- LLM provider SDK: Any (bug is in utility function, not provider-specific)

stakeswky:
## Summary

`strip_markdown_fences()` in `_utils.py` uses a greedy `.*` in the regex pattern with `re.DOTALL`, which causes it to match past the closing ``````` fence marker when trailing text contains `}` characters. This results in invalid JSON extraction that breaks downstream parsing.

## Problem

The pattern `r'```(?:\w+)?\n(\{.*\})`` with `re.DOTALL` is greedy — `.*` consumes everything including the closing fence and any subsequent text up to the last `}` in the string.

Example:
```
``json
{"foo": "bar"}
``
Some text with a } here
```

**Before (bug):** returns `{"foo": "bar"}\n```\nSome text with a }`
**After (fix):** returns `{"foo": "bar"}`

## Fix

Split extraction into two steps:
1. Use a non-greedy match to extract content within fence boundaries (``````` ... ``````` or end of string)
2. Find the JSON object (`{...}`) within the extracted fence content

This ensures trailing text after the closing fence is never included in the result.

## Tests

Added two test cases:
- Greedy regex bug case: text after closing fence with `}` characters
- Nested JSON objects within fences

Fixes #4397

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/_utils.py
+++ b/pydantic_ai_slim/pydantic_ai/_utils.py
@@ -514,7 +514,8 @@ def validate_empty_kwargs(_kwargs: dict[str, Any]) -> None:
         raise exceptions.UserError(f'Unknown keyword arguments: {unknown_kwargs}')
 
 
-_MARKDOWN_FENCES_PATTERN = re.compile(r'```(?:\w+)?\n(\{.*\})', flags=re.DOTALL)
+_MARKDOWN_FENCES_PATTERN = re.compile(r'```(?:\w+)?\n(.*?)(?:\n```|$)', flags=re.DOTALL)
+_JSON_OBJECT_PATTERN = re.compile(r'(\{.*\})', flags=re.DOTALL)
 
 
 def strip_markdown_fences(text: str) -> str:
@@ -523,7 +524,12 @@ def strip_markdown_fences(text: str) -> str:
 
     match = re.search(_MARKDOWN_FENCES_PATTERN, text)
     if match:
-        return match.group(1)
+        fence_content = match.group(1)
+        # Find the JSON object within the fence content
+        json_match = re.search(_JSON_OBJECT_PATTERN, fence_content)
+        if json_match:
+            return json_match.group(1)
+        return fence_content
 
     return text
 
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -527,6 +527,13 @@ def test_strip_markdown_fences():
         == '{"foo": "bar"}'
     )
     assert strip_markdown_fences('No JSON to be found') == 'No JSON to be found'
+    # Greedy regex bug: should not consume past closing fence into trailing text with braces
+    assert (
+        strip_markdown_fences('```json\n{"foo": "bar"}\n```\nSome text with a } here')
+        == '{"foo": "bar"}'
+    )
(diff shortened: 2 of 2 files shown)


DouweM:
@stakeswky Thanks, this look good. Can you please explain why the fix you had originally in https://github.com/pydantic/pydantic-ai/pull/4400 was not sufficient? And please fix the merge conflicts.

themavik (commented):
Non-greedy fence capture plus `_JSON_OBJECT_PATTERN` fixes `strip_markdown_fences` eating past the closing fence when trailing prose contains `}`. nit: `_JSON_OBJECT_PATTERN` is still greedy with `DOTALL`—weird multi-object blobs inside one fence could grab the wrong span; tests cover the nested-object case.

DouweM:
Superseded by https://github.com/pydantic/pydantic-ai/pull/4398

PR #4403, opened 2026-02-22
feat: Expose A2A execution context via ContextVars and make `a2a` module public

### LINKED ISSUE #4330: Exposing tool calls for A2A agents without duplicating existing Pydantic AI functionality
### Question

I'm using pydantic and FastA2A to expose agents via A2A protocol.
For long-running agents, not having access to internals of the agent, like the tool calls, results in trade-offs for the user experience. I would love to expose the tool calls to the A2A client so that I'm able to provide users with a little bit more details on what happens under the hood.

Today, the [agent_to_a2a](https://github.com/pydantic/pydantic-ai/blob/3ef42ed9a1a2c799bb94a5a69c80aa9e8968ca72/pydantic_ai_slim/pydantic_ai/_a2a.py#L72C5-L72C17) provides means for customizing the storage and the broker, however `AgentWorker` runs the [full agent loop](https://github.com/pydantic/pydantic-ai/blob/3ef42ed9a1a2c799bb94a5a69c80aa9e8968ca72/pydantic_ai_slim/pydantic_ai/_a2a.py#L140) in one go. I successfully replaced it with `agent.run_stream_events` where are for tool call events I'm calling `storage.update_task` to track tool call execution.

However, I ended up creating a copy of the A2A code from Pydantic as there I see no direct configuration opportunities at the moment. I wonder if there is a better way.

Potentially, this would be an opportunity to extend `AgentWorker` and the `agent_to_a2a` method to inject a handler function to react to events and modify task or storage.

Thanks for your initial thoughts on this question.

### Additional Context

I'm building a user interface for chatting with A2A agents. To users of the interface, I would like to provide fine-grained details about the agent execution. For implementing agents, ideally I avoid the need of building a library, shipping a worker implementation which may need to be synced with the implementation in pydantic-ai.

AlanPonnachan:
- Closes #4330 

### Description

This PR addresses the need to access A2A task metadata (like `task_id`, `task`, `storage`, and `broker`) from within an agent's execution loop (e.g., inside event stream handlers or tool calls) without duplicating the `AgentWorker` loop.

**Changes made:**
- Renamed the internal `pydantic_ai._a2a` module to the public `pydantic_ai.a2a`.
- Introduced four `ContextVar`s in `a2a.py`: `a2a_task_id`, `a2a_task`, `a2a_storage`, and `a2a_broker`.
- Updated `AgentWorker.run_task` to set these context variables before calling `self.agent.run()` and securely reset them in a `finally` block.
- Updated internal imports in `pydantic_ai.agent.abstract` to reflect the module rename.
- Added comprehensive unit tests in `tests/test_a2a.py` to verify that all exposed context variables are accessible and correctly typed during a tool call execution.

### Pre-Review Checklist

### Pre-Merge Checklist


AlanPonnachan:
@DouweM

gablemire-layerai:
Any updates on this PR being merged? It would be useful, as I'm starting to work on a system that needs to leverage internal A2A tool calls.

0xbrainkid:
Making A2A execution context accessible via `ContextVar`s during tool calls is the right pattern — tools that need to know their execution context (which task they are serving, which broker they are connected to) currently have to thread that context explicitly through every call site.

One additional context variable worth considering alongside the current four: **caller agent identity.** The `task_id` and `task` give the execution context. The calling agent identity gives the trust context — which agent initiated this task, and what is its behavioral trust score.

```python
# Proposed additions to a2a.py
a2a_caller_agent_id: ContextVar[str | None] = ContextVar("a2a_caller_agent_id", default=None)
a2a_caller_trust_score: ContextVar[float | None] = ContextVar("a2a_caller_trust_score", default=None)
a2a_caller_attestation_url: ContextVar[str | None] = ContextVar("a2a_caller_attestation_url", default=None)
```

Set from the A2A task metadata if the caller included an agent identity header (`X-Agent-ID`, or from the Agent Card trust signals). This gives tool implementations a standard way to apply caller-trust-sensitive logic without requiring the agent author to wire caller identity through every tool signature:

```python
@tool
def privileged_operation(query: str) -> str:
    caller_trust = a2a_caller_trust_score.get()
    if caller_trust is not None and caller_trust < 0.80:
        return "Insufficient trust for this operation"
    return execute_privileged(query)
```

The `finally` block reset pattern you are already using for the four existing vars is the right approach — this would follow the same pattern.

Issue #4296, opened 2026-02-11
Exhaustive end_strategy silently discards output tool results when model calls multiple output tools


Baniskos:
## Context

PR #3523 (closing #3485) intentionally changed exhaustive mode to fully execute all output tools, not just the first one. The use case was output functions used as side-effecting actions (e.g., `send_text_message()` + `send_quick_replies_message()`) where the user wants every output tool to run.

This works well for that use case, but introduces an edge case: when the model **mistakenly** calls multiple output tools in a single response, all of them are fully executed — output function, output validators, the full pipeline — and every result after the first is silently discarded.

## The problem

When the model erroneously returns multiple output tool calls, the current code fully executes all of them but only keeps the first successful result:

```python
if not final_result:  # False for every call after the first success
    final_result = FinalResult(result_data, call.tool_name, call.tool_call_id)
```

The discarded output tools still go through the full execution path:

```
ToolManager.handle_call(call)
  → validator.validate_json(call.args)          # arg validation
  → OutputToolset.call_tool(name, args_dict)
    → ObjectOutputProcessor.call(tool_args)
      → execute_traced_output_function(...)      # OUTPUT FUNCTION RUNS
    → OutputValidator.validate(output) × N       # OUTPUT VALIDATORS RUN
  → return output                                # result silently discarded
```

This is low-risk when output functions are pure data transforms, but can be problematic when they have side effects — database writes, API calls, message sends, billing events — that execute with no indication their result was thrown away. The model calling multiple output tools by mistake is uncommon, but the consequences scale with what the output function does.

## MRE

`TestModel` won't reproduce this because it calls tools one at a time across separate model responses. `FunctionModel` is needed to return both output tool calls in a single `ModelResponse`:

```python
import json

from pydantic_ai import Agent
from pydantic_ai.messages import ModelResponse, ToolCallPart
from pydantic_ai.models.function import AgentInfo, FunctionModel

side_effects = []


def create_fruit(name: str, color: str) -> dict:
    """Output function with a side effect."""
    side_effects.append(f"fruit: {name}")
    return {"name": name, "color": color}


def create_vehicle(name: str, wheels: int) -> dict:
    """Output function with a side effect."""
    side_effects.append(f"vehicle: {name}")
    return {"name": name, "wheels": wheels}


def model_function(messages, info: AgentInfo) -> ModelResponse:
    """Model mistakenly returns both output tool calls."""
    return ModelResponse(parts=[
        ToolCallPart(
[...]

PR #3523, opened 2025-11-22
Make `end_strategy` also work for output tools, not just tools

Danipulok:
Closes #3485

### Changes:
- Change logic of handling output tools with `end_strategy='exhaustive'`;
- Added tests;
- Updated docstrings;
- Updated docs;

### How I verified:
- Tests;
- Built and checked docs visually;
- Checked that MRE from #3485 now successfully calls both output tools;

### MRE:
```python
import asyncio
import os

from pydantic import BaseModel
from pydantic_ai import Agent, ModelSettings, ToolOutput
from pydantic_ai.models.openai import OpenAIChatModel
from pydantic_ai.providers.openai import OpenAIProvider


class TextMessage(BaseModel):
    text: str | None = None


async def send_text_message(
    message: TextMessage,
) -> None:
    """Send a text message."""
    print(f"\nText Message: {message}")


class QuickRepliesMessage(BaseModel):
    text: str | None = None
    quick_replies: list[str] | None = None


async def send_quick_replies_message(
    message: QuickRepliesMessage,
) -> None:
    """Send a quick replies message."""
    print(f"\nQuick Replies Message: {message}")


class TestDate(BaseModel):
    days_of_sunshine: int
    info: str


async def main() -> None:
    api_key = os.environ["OPENAI_API_KEY"]
    model = OpenAIChatModel(
        "gpt-4o",
        provider=OpenAIProvider(api_key=api_key),
        settings=ModelSettings(
            temperature=0.1,
        ),
    )
    output_type = [
        ToolOutput(send_text_message, name="send_text_message"),
        ToolOutput(send_quick_replies_message, name="send_quick_replies_message"),
    ]
    agent = Agent(
        model,
        output_type=output_type,
        instructions="For response, call at first `send_quick_replies_message` and then `send_text_message`, both in parallel",
        end_strategy="exhaustive",  # comment to use the default "early" strategy
    )

    user_prompt = "Tell me about Python"

    async with agent.run_stream(user_prompt) as run:
        async for output in run.stream_responses():
            model_response, is_last_message = output
            print(model_response, is_last_message, end="\n\n")


if __name__ == "__main__":
    asyncio.run(main())

```

#### Old output:
```plaintext
[...]

PR #4406, opened 2026-02-23
feat: add 'complete' end_strategy option

by22Jy:
## Summary

Implements #4296 - Add new `'complete'` end_strategy option that executes all function tools but only the first output tool.

## Problem

When using `end_strategy='exhaustive'`, if a model mistakenly calls multiple output tools in a single response:
1. All output tools are fully executed (including side effects like database writes, API calls)
2. But only the first result is used
3. Subsequent results are silently discarded

This can be problematic when output functions have side effects.

## Solution

Add a new `'complete'` end_strategy:

- `'early'`: Stop at first valid output, skip all other tools
- `'complete'`: Execute all function tools, but only first output tool ⭐ **NEW**
- `'exhaustive'`: Execute all tools including multiple output tools

### Core Logic

Modified `pydantic_ai_slim/pydantic_ai/_agent_graph.py`:

1. **Line 63**: Updated `EndStrategy` type definition
   ```python
   EndStrategy = Literal['early', 'complete', 'exhaustive']
   ```

2. **Line 923**: Updated output tool handling in `process_tool_calls()`
   ```python
   # Skip subsequent output tools in both 'early' and 'complete' modes
   elif (ctx.deps.end_strategy == 'early' or ctx.deps.end_strategy == 'complete') and final_result:
   ```

3. **Line 980**: Function tool handling unchanged (only 'early' skips function tools)
   ```python
   # Only 'early' skips function tools; 'complete' and 'exhaustive' execute them all
   if final_result and ctx.deps.end_strategy == 'early':
   ```

### 'early' - Quick Exit
Use when you want to stop immediately upon getting a valid result:
```python
agent = Agent(model, output_type=[...], end_strategy='early')
```

### 'complete' - Complete Function Execution (NEW)
Use when function tools have side effects that should always run:
```python
agent = Agent(model, output_type=[...], end_strategy='complete')
# - All function tools will execute
# - Only the first output tool will execute
```

### 'exhaustive' - Full Execution
Use when output tools have side effects that should all run:
```python
agent = Agent(model, output_type=[...], end_strategy='exhaustive')
# - All tools (function and output) will execute
# - Use with caution if output functions have side effects
```

## Testing

This change is backward compatible:
- Existing code using `'early'` or `'exhaustive'` behaves identically
- New `'complete'` option provides a middle ground

## Related

- Issue: #4296
- Discussion with @DouweM and @Baniskos on the preferred solution
- Builds on PR #3523 which introduced exhaustive mode

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/_agent_graph.py
+++ b/pydantic_ai_slim/pydantic_ai/_agent_graph.py
@@ -60,7 +60,7 @@ __all__ = (
 T = TypeVar('T')
 S = TypeVar('S')
 NoneType = type(None)
-EndStrategy = Literal['early', 'exhaustive']
+EndStrategy = Literal['early', 'complete', 'exhaustive']
 DepsT = TypeVar('DepsT')
 OutputT = TypeVar('OutputT')
 
@@ -920,7 +920,7 @@ async def process_tool_calls(  # noqa: C901
             )
             output_parts.append(part)
         # Early strategy is chosen and final result is already set
-        elif ctx.deps.end_strategy == 'early' and final_result:
+        elif (ctx.deps.end_strategy == 'early' or ctx.deps.end_strategy == 'complete') and final_result:
             yield _messages.FunctionToolCallEvent(call)
             part = _messages.ToolReturnPart(
                 tool_name=call.tool_name,


DouweM:
Duplicate of https://github.com/pydantic/pydantic-ai/pull/4333

PR #4417, opened 2026-02-24
Fix AGUIAdapter run_id issue by always setting run_id in ModelRequestNode

liangzhang-keepmoving:
This PR fixes an issue where ModelRequest objects created by AGUIAdapter had empty run_id fields.

The problem was in the `ModelRequestNode._prepare_request` method, where it only set the run_id if `is_resuming_without_prompt` was False. However, when AGUIAdapter creates a ModelRequest object, it doesn't set the run_id, and then UserPromptNode sets `is_resuming_without_prompt` to True when processing it, causing the run_id to never be set.

The fix is simple: always set the run_id if it's not already set, regardless of the value of `is_resuming_without_prompt`.

This ensures that all ModelRequest objects have a valid run_id, which is important for tracing and logging purposes.

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/_agent_graph.py
+++ b/pydantic_ai_slim/pydantic_ai/_agent_graph.py
@@ -398,6 +398,50 @@ class UserPromptNode(AgentNode[DepsT, NodeRunEndT]):
     __repr__ = dataclasses_no_defaults_repr
 
 
+async def _prepare_request_parameters(
+    ctx: GraphRunContext[GraphAgentState, GraphAgentDeps[DepsT, NodeRunEndT]],
+) -> models.ModelRequestParameters:
+    """Build tools and create an agent model."""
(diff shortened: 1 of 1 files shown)


liangzhang-keepmoving:
@pydantic/pydantic-ai-maintainers Could you please review this PR when you have a chance?

This PR fixes an issue where ModelRequest objects created by AGUIAdapter had empty run_id fields, which could cause problems with tracing and logging.

The fix is simple and focused: it ensures that run_id is always set in ModelRequestNode._prepare_request, regardless of the value of is_resuming_without_prompt.

This is an important fix for anyone using AGUIAdapter, as it ensures all ModelRequest objects have valid run_id values.

Thank you for your time!

DouweM:
Already being worked on in https://github.com/pydantic/pydantic-ai/pull/4419.

PR #4420, opened 2026-02-24
Fix AGUIAdapter run_id issue by always setting run_id in ModelRequestNode (minimal fix)

liangzhang-keepmoving:
This PR fixes an issue where ModelRequest objects created by AGUIAdapter had empty run_id fields.

The problem was in the `ModelRequestNode._prepare_request` method, where it only set the run_id if `is_resuming_without_prompt` was False. However, when AGUIAdapter creates a ModelRequest object, it doesn't set the run_id, and then UserPromptNode sets `is_resuming_without_prompt` to True when processing it, causing the run_id to never be set.

The fix is simple: always set the run_id if it's not already set, regardless of the value of `is_resuming_without_prompt`.

This ensures that all ModelRequest objects have a valid run_id, which is important for tracing and logging purposes.

This is a minimal fix that only changes the specific line of code needed to resolve the issue, without modifying any other functionality.

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/_agent_graph.py
+++ b/pydantic_ai_slim/pydantic_ai/_agent_graph.py
@@ -444,71 +444,27 @@ async def _prepare_request_parameters(
 
 @dataclasses.dataclass
 class ModelRequestNode(AgentNode[DepsT, NodeRunEndT]):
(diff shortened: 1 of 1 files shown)


DouweM:
Already being worked on in https://github.com/pydantic/pydantic-ai/pull/4419.

PR #4431, opened 2026-02-25
fix: handle malformed tool call args in Anthropic retry path

### LINKED ISSUE #4430: Anthropic retry path crashes on malformed tool args (args_as_dict ValueError) before model can self-correct
### Initial Checks

### Description

Observed behavior:

Expected behavior:

```Python
import asyncio

from pydantic_ai import Agent

BAD_ARGS = (
    '{"query": "bad query", '
)

message_history = [
    ModelResponse(
        parts=[
            ToolCallPart(
                tool_name="search_knowledge",
                tool_call_id="toolu_123",
                args=BAD_ARGS,
            )
        ],
    ),
    ModelRequest(
        parts=[
            RetryPromptPart(
                tool_name="search_knowledge",
                tool_call_id="toolu_123",
                content=[
                    {
                        "type": "json_invalid",
                        "loc": [],
                        "input": BAD_ARGS,
                    }
                ],
            )
        ]
    ),
]

model = AnthropicModel(
    "claude-sonnet-4-5",
)
agent = Agent(model)

async def main():
    try:
        await agent.run(
        )
    except Exception as e:
        print(type(e).__name__)
        print(e)

asyncio.run(main())

# Actual output:
# ValueError
```

### Logfire Trace

_No response_

- Python: 3.11
- Pydantic AI: 1.63.0

Saad-Azi:
## Summary

Fixes #4430 — Anthropic retry path crashes on malformed tool args (rgs_as_dict ValueError) before model can self-correct.

**Updated:** After review feedback, this PR now fixes the same vulnerability across **all 6 affected providers**, not just Anthropic.

### Problem

When a tool call has malformed JSON arguments (which Sonnet 4.6 has been observed to produce in production), pydantic-ai correctly creates a RetryPromptPart with the error. However, when the conversation history is re-sent to the provider, the previous tool call's args are parsed via rgs_as_dict() which calls pydantic_core.from_json() — this raises a raw ValueError on the malformed JSON, crashing the retry flow *before* the model gets a chance to self-correct.

Other providers that use rgs_as_json_str() (like OpenAI) don't crash because they pass the raw string without parsing. But providers whose APIs require a dict use rgs_as_dict() and are all vulnerable.

### Fix

Added a safe_args_as_dict() method on BaseToolCallPart in messages.py that wraps rgs_as_dict() with a try/except. When malformed JSON is encountered, it:
1. Catches the ValueError (and AssertionError from the dict type check)
2. Logs a debug message with the tool name and raw args
3. Returns {} as a fallback

This allows the conversation history to be sent to the API successfully. The RetryPromptPart already contains the error details, so the model receives full context about what went wrong and can self-correct.

### Tradeoff Note

When safe_args_as_dict() falls back to {}, the original malformed args string is lost from the API request (preserved only in the debug log). If the model was close to correct JSON (e.g. a minor syntax error), seeing {} instead of an approximation of the original args could reduce the model's ability to self-correct. This is a reasonable tradeoff given the alternative is a crash.

### Changes

**Core:**
- **\pydantic_ai_slim/pydantic_ai/messages.py\**: Added \safe_args_as_dict()\ method on \BaseToolCallPart\ with try/except for \ValueError\/\AssertionError\, falling back to \{}\

**All affected providers:**
- **\models/anthropic.py\**: \ToolCallPart\ mapping → \
esponse_part.safe_args_as_dict()\
- **\models/gemini.py\**: \_function_call_part_from_call\ → \	ool.safe_args_as_dict()\
- **\models/google.py\**: \ToolCallPart\ → \FunctionCallDict\ mapping → \item.safe_args_as_dict()\
- **\models/bedrock.py\**: \_map_tool_call\ → \	.safe_args_as_dict()\
- **\models/mistral.py\**: output tool call from streamed text → \maybe_tool_call_part.safe_args_as_dict()\
- **\models/xai.py\**: \BuiltinToolCallPart\ MCP server args → \item.safe_args_as_dict()\

**Tests (\	ests/models/test_anthropic.py\):**
- \	est_anthropic_malformed_tool_args_no_crash\ — integration test reproducing the exact issue scenario
- \	est_safe_args_as_dict_valid_json\ — valid JSON args parsed correctly
- \	est_safe_args_as_dict_dict_args\ — dict args returned directly
- \	est_safe_args_as_dict_malformed_json\ — malformed JSON returns \{}\ instead of raising
- \	est_safe_args_as_dict_empty_args\ — None/empty args return \{}\

All 5 tests pass. Lint and format checks pass.

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/messages.py
+++ b/pydantic_ai_slim/pydantic_ai/messages.py
@@ -2,7 +2,6 @@ from __future__ import annotations as _annotations
 
 import base64
 import hashlib
-import logging
 import mimetypes
 import os
 from abc import ABC, abstractmethod
@@ -30,8 +29,6 @@ from .usage import RequestUsage
 
 if TYPE_CHECKING:
     from .models.instrumented import InstrumentationSettings
-
-_logger = logging.getLogger(__name__)
 
 # Key used to wrap malformed tool-call arguments so they can still be round-tripped
 # through a model API without crashing.  The specific string 'INVALID_JSON' is the
@@ -1572,10 +1569,10 @@ class BaseToolCallPart:
         This is just for convenience with models that require dicts as input.
 
         Args:
-            raise_if_invalid: If ``True``, a ``ValueError`` or ``AssertionError``
-                caused by malformed JSON in ``args`` will be re-raised.  When
-                ``False`` (the default), malformed JSON is handled gracefully by
-                returning ``{'INVALID_JSON': '<raw args>'}`` so that the value
+            raise_if_invalid: If `True`, a `ValueError` or `AssertionError`
+                caused by malformed JSON in `args` will be re-raised.  When
+                `False` (the default), malformed JSON is handled gracefully by
+                returning `{'INVALID_JSON': '<raw args>'}` so that the value
                 can still be sent to a model API (e.g. during a retry flow)
                 without crashing.
         """
@@ -1590,11 +1587,6 @@ class BaseToolCallPart:
         except (ValueError, AssertionError):
             if raise_if_invalid:
                 raise
-            _logger.debug(
-                'Malformed tool call args for %r, falling back to INVALID_JSON wrapper: %s',
-                self.tool_name,
-                self.args,
-            )
             return {INVALID_JSON_KEY: self.args}
 
     def args_as_json_str(self) -> str:
--- a/tests/models/test_anthropic.py
+++ b/tests/models/test_anthropic.py
@@ -8514,7 +8514,7 @@ async def test_anthropic_malformed_tool_args_no_crash(allow_model_requests: None
     tool call's args are parsed via args_as_dict() which raises ValueError on
     invalid JSON, crashing the retry flow before the model can self-correct.
     """
-    BAD_ARGS = '{"query": "bad query", "file_ids":[4556]</parameter>\n<parameter name="limit": 8}'
+    bad_args = '{"query": "bad query", "file_ids":[4556]</parameter>\n<parameter name="limit": 8}'
 
     # First response: the model "fixes" the tool call and returns text
     fixed_response = completion_message(
@@ -8533,7 +8533,7 @@ async def test_anthropic_malformed_tool_args_no_crash(allow_model_requests: None
                 ToolCallPart(
                     tool_name='search_knowledge',
                     tool_call_id='toolu_123',
-                    args=BAD_ARGS,
+                    args=bad_args,
                 ),
             ],
             timestamp=datetime(2025, 1, 1, tzinfo=timezone.utc),


Saad-Azi:
## CI Status Note

All code-related CI checks pass (lint, mypy, test across Python 3.10-3.14, test-examples, test-lowest-versions, coverage at 100%).

The **docs** job failure is unrelated to this PR — it's an npm registry issue:
`
npm error 403 403 Forbidden - GET https://registry.npmjs.org/zod/-/zod-3.25.76.tgz
`
This is a transient infrastructure issue with the \zod\ package on npm's registry.

## Re: Public API Surface (\safe_args_as_dict\)

The \safe_args_as_dict()\ method on \BaseToolCallPart\ is intentionally public (not prefixed with \_\). It sits alongside the existing \args_as_dict()\ and \args_as_json_str()\ methods as a natural companion. This makes it available to third-party provider implementations that also need to re-send \ToolCallPart\ in message history without crashing on malformed JSON. Happy to make it private if the maintainers prefer a different API design.

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -1286,6 +1289,26 @@ def args_as_dict(self) -> dict[str, Any]:
...
+        """
+        try:
+            return self.args_as_dict()
+        except (ValueError, AssertionError):
+            _logger.debug(
+                'Malformed tool call args for %r, falling back to empty dict: %s',
+                self.tool_name,
+                self.args,
+            )
+            return {}
```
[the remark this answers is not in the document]
DouweM: I'd rather add a new argument to the existing `args_as_dict` method

DouweM: We should not return `{}`, it's more useful for the model to receive `{'__invalid_json__': 'json string'}`

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/messages.py
slmnsh: anthropic suggest it should be

```json
{
  "INVALID_JSON": "<your invalid json string>"
}
```

ref: https://platform.claude.com/docs/en/agents-and-tools/tool-use/fine-grained-tool-streaming#handling-invalid-json-in-tool-responses

DouweM: @slmnsh Great!

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -1273,18 +1276,34 @@ class BaseToolCallPart:
     When this field is set, `provider_name` is required to identify the provider that generated this data.
     """
 
-    def args_as_dict(self) -> dict[str, Any]:
+    def args_as_dict(self, *, allow_partial: bool = False) -> dict[str, Any]:
```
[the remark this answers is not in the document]
DouweM: Hmm, what do you think of making the flag `raise_if_invalid` with a default of `False`, so the model classes can expect the method to always work normally? Then we only need to pass `raise_if_invalid=True` in the case where we specifically are testing args validity, like here:

https://github.com/pydantic/pydantic-ai/blob/0b0ec325b651ece0f0087f8a059a91a262a5efdc/pydantic_ai_slim/pydantic_ai/_agent_graph.py#L111-L113

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
```diff
@@ -533,7 +521,7 @@ def _dump_response_message(
                             ToolOutputErrorPart(
                                 type=tool_name,
                                 tool_call_id=part.tool_call_id,
-                                input=_safe_args_as_dict(part),
+                                input=part.args_as_dict(),
```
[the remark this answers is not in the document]
DouweM: Im OK with the change

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -1410,18 +1413,36 @@ class BaseToolCallPart:
...
+            return cast(dict[str, Any], args)
+        except (ValueError, AssertionError):
+            if raise_if_invalid:
+                raise
+            _logger.debug(
+                'Malformed tool call args for %r, falling back to INVALID_JSON wrapper: %s',
+                self.tool_name,
+                self.args,
+            )
+            return {'INVALID_JSON': self.args}
```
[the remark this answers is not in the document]
DouweM: Yeah it should be a constant, with a comment saying it's the specific string Anthropic docs recommend (with a link).

And the docstring complaint you can ignore.

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -1557,18 +1566,36 @@ class BaseToolCallPart:
...
     """
 
-    def args_as_dict(self) -> dict[str, Any]:
+    def args_as_dict(self, *, raise_if_invalid: bool = False) -> dict[str, Any]:
         """Return the arguments as a Python dictionary.
 
         This is just for convenience with models that require dicts as input.
+
+        Args:
+            raise_if_invalid: If ``True``, a ``ValueError`` or ``AssertionError``
```
DouweM: No double backticks please

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -1557,18 +1566,36 @@ class BaseToolCallPart:
...
-        assert isinstance(args, dict), 'args should be a dict'
-        return cast(dict[str, Any], args)
+        try:
+            args = pydantic_core.from_json(self.args)
+            assert isinstance(args, dict), 'args should be a dict'
+            return cast(dict[str, Any], args)
+        except (ValueError, AssertionError):
+            if raise_if_invalid:
+                raise
+            _logger.debug(
```
DouweM: No logging please; we don't do this anywhere else.

tests/models/test_anthropic.py
```diff
@@ -8502,3 +8502,140 @@ async def test_anthropic_system_prompts_and_instructions_ordering():
...
+    """Test that malformed JSON tool args don't crash the Anthropic retry path.
+
+    Regression test for https://github.com/pydantic/pydantic-ai/issues/4430.
+
+    When a tool call has malformed JSON arguments, a RetryPromptPart is correctly
+    created. But when the message history is re-sent to Anthropic, the previous
+    tool call's args are parsed via args_as_dict() which raises ValueError on
+    invalid JSON, crashing the retry flow before the model can self-correct.
+    """
+    BAD_ARGS = '{"query": "bad query", "file_ids":[4556]</parameter>\n<parameter name="limit": 8}'
```
DouweM: Not a constant, doesn't need to be all caps.

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -30,6 +30,12 @@
 if TYPE_CHECKING:
     from .models.instrumented import InstrumentationSettings
 
+# Key used to wrap malformed tool-call arguments so they can still be round-tripped
+# through a model API without crashing.  The specific string 'INVALID_JSON' is the
+# value recommended by the Anthropic docs for this situation:
+# https://docs.anthropic.com/en/docs/build-with-claude/tool-use/overview#handling-tool-use-errors
+INVALID_JSON_KEY = 'INVALID_JSON'
```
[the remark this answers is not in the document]
DouweM: as is ok, without exporting from the main module

PR #4435, opened 2026-02-26
fix(mcp): move `__aexit__` guard inside lock to prevent TOCTOU race

### LINKED ISSUE #4422: Bug: MCPServer.__aexit__() checks _running_count outside the lock — TOCTOU race allows negative count and use-after-close
### Initial Checks

### Description

`MCPServer.__aexit__()` reads `self._running_count` **outside** `self._enter_lock`, while `__aenter__()` correctly does the check **inside** the lock. This Time-Of-Check-Time-Of-Use (TOCTOU) gap allows concurrent `__aexit__` calls to both pass the guard, decrementing `_running_count` to **-1** and corrupting lifecycle state.

### Root Cause

```python
# mcp.py — __aenter__ (correct ✅)
async def __aenter__(self) -> Self:
    async with self._enter_lock:              # lock FIRST
        if self._running_count == 0:          # check INSIDE lock
            # ... initialization ...
        self._running_count += 1
    return self

# mcp.py — __aexit__ (bug ❌)
async def __aexit__(self, *args: Any) -> bool | None:
    if self._running_count == 0:              # check OUTSIDE lock
        raise ValueError(...)
    async with self._enter_lock:              # lock AFTER check
        self._running_count -= 1
        if self._running_count == 0 and self._exit_stack is not None:
            await self._exit_stack.aclose()
```

### Race Scenario

With `_running_count == 1` and two coroutines exiting concurrently:

```mermaid
sequenceDiagram
    participant A as Coroutine A
    participant B as Coroutine B
    participant S as MCPServer state

    Note over S: _running_count = 1
    A->>S: __aexit__: read _running_count == 1 (pass guard)
    B->>S: __aexit__: read _running_count == 1 (pass guard)
    A->>S: acquire lock, decrement → 0, close _exit_stack
    Note over S: _running_count = 0, _exit_stack = None
    B->>S: acquire lock, decrement → -1
    Note over S: _running_count = -1 ⚠️
    Note over S: Next __aenter__ sees -1 ≠ 0, skips init,<br/>increments to 0 — uses closed client
```

**Result**: `_running_count` becomes -1. A subsequent `__aenter__` skips initialization (because `-1 != 0`), increments to 0, and the caller uses a closed `_client`/`_exit_stack`, leading to errors like those seen in #3542.

### Relationship to #3542

Issue #3542 reports the symptom (`ValueError: MCPServer.__aexit__ called more times than __aenter__`) when using dynamic toolsets. This issue identifies the **root cause**: the TOCTOU race in `__aexit__` that allows the lifecycle invariant to be violated. The `ValueError` in #3542 is actually a *consequence* of a previous race that corrupted `_running_count`, causing a later `__aexit__` call to see the (now incorrect) zero value.

### Proposed Fix

Move the guard check inside the lock, matching `__aenter__`'s pattern:

```python
async def __aexit__(self, *args: Any) -> bool | None:
    async with self._enter_lock:
        if self._running_count == 0:
            raise ValueError('MCPServer.__aexit__ called more times than __aenter__')
        self._running_count -= 1
        if self._running_count == 0 and self._exit_stack is not None:
            await self._exit_stack.aclose()
            self._exit_stack = None
            self._cached_tools = None
            self._cached_resources = None
```

| | Current | Proposed |
|---|---|---|
| Complexity | 1 line moved | Same |
| Breaking changes | None | None |
| Thread safety | Broken under contention | Correct |
| Performance | Same (lock acquired in both cases) | Same |

### Environment

- Python: 3.12
- Pydantic AI: latest (main branch)
- File: `pydantic_ai_slim/pydantic_ai/mcp.py`, lines 716-725

I'm happy to open a PR for this fix if you'd like.

jameslcowan:
## Summary

Fixes #4422.

## Problem

`MCPServer.__aexit__()` checked `_running_count == 0` **outside** `_enter_lock`, while `__aenter__()` correctly performs its check **inside** the lock.

With `_running_count == 1` and two coroutines exiting concurrently:

1. Both coroutines read `_running_count == 1` and pass the guard
2. Both acquire the lock in turn and decrement
3. Final count: **-1**

This corrupts lifecycle state — a subsequent `__aenter__` sees `-1 != 0`, skips initialization, increments to 0, and proceeds with a closed client.

## Fix

Move the guard check inside the lock so the read and decrement are atomic:

```python
# Before
async def __aexit__(self, *args):
    if self._running_count == 0:          # outside lock ❌
        raise ValueError(...)
    async with self._enter_lock:
        self._running_count -= 1
        ...

# After
async def __aexit__(self, *args):
    async with self._enter_lock:
        if self._running_count == 0:      # inside lock ✅
            raise ValueError(...)
        self._running_count -= 1
        ...
```

## Changes

- `pydantic_ai_slim/pydantic_ai/mcp.py`: move guard inside lock (2-line change)
- `tests/test_mcp.py`: add `asyncio` import + regression test for concurrent exits

## Tests

```
tests/test_mcp.py::test_aexit_called_more_times_than_aenter PASSED
tests/test_mcp.py::test_aexit_concurrent_does_not_corrupt_running_count PASSED
```

### DIFF
--- a/tests/test_mcp.py
+++ b/tests/test_mcp.py
@@ -7,7 +7,7 @@ import base64
 import re
 from datetime import timezone
 from pathlib import Path
-from typing import Any
+from typing import Any, Literal
 from unittest.mock import AsyncMock, patch
 
 import pytest
@@ -169,16 +169,10 @@ async def test_aexit_concurrent_does_not_corrupt_running_count():
 
     # Replace the real lock with one that yields before acquiring,
     # opening the interleaving window the old code left unprotected.
-    real_lock = server._enter_lock  # pyright: ignore[reportPrivateUsage]
-
-    class InterleavingLock:
-        async def __aenter__(self) -> InterleavingLock:
+    class InterleavingLock(asyncio.Lock):
+        async def acquire(self) -> Literal[True]:
             await asyncio.sleep(0)  # yield — lets the other task reach the same point
-            await real_lock.acquire()
-            return self
-
-        async def __aexit__(self, *args: Any) -> None:
-            real_lock.release()
+            return await super().acquire()
 
     server._enter_lock = InterleavingLock()  # pyright: ignore[reportPrivateUsage]
 


DouweM:
@jameslcowan Thanks James, the fix looks good. Can you please fix the lint issue?

PR #4453, opened 2026-02-27
Handle JSON-serialized dict attributes in SpanQuery has_attributes matching

### LINKED ISSUE #4448: Evals: SpanQuery format mismatch
### Initial Checks

### Problem

### Desired State

### Solution

```python
```
with
```python
```

```Python
import os

import dotenv
import logfire
    context_subtree,
)

dotenv.load_dotenv()

    with logfire.span("task"):
        with logfire.span("subtask"):
            data = DATA


assert (
    is not None
)

assert (
    is None
)
```

### Logfire Trace

_No response_

- Python: 3.12
- Pydantic AI: 1.63.0
- LLM provider SDK:

debu-sinha:
Closes #4448

## What changes are proposed

When instrumentation libraries like logfire store complex attribute values (dicts, lists) as span attributes, they serialize them to JSON strings since OpenTelemetry only supports primitive attribute types (`str | bool | int | float | Sequence[...]`).

The `has_attributes` condition in `SpanQuery` uses direct equality (`==`) to compare stored and expected values. This fails when the stored value is a JSON string like `'{"foo":1}'` and the query uses the original Python dict `{"foo": 1}`.

This adds a fallback to `has_attributes` matching: when a direct comparison fails and the stored value is a string, it attempts to JSON-deserialize the stored string and compare as Python objects. This handles dicts, lists, and nested structures correctly, with key-order independence.

## Changes

- `pydantic_evals/pydantic_evals/otel/span_tree.py`: Add `_attribute_value_matches()` helper that handles JSON string deserialization fallback. Update `_matches_query` to use it for `has_attributes` comparisons.
- `tests/evals/test_otel.py`: Add tests for JSON-serialized dict/list attribute matching and unit tests for the helper function.

## How is this PR tested

- 2 new tests added, both passing
- All 27 existing tests in `test_otel.py` continue to pass
- Tests cover: dict matching, nested dict matching, list matching, non-JSON string rejection, missing key, wrong value

```
tests/evals/test_otel.py::test_attribute_value_matches_json_serialized_dicts PASSED
tests/evals/test_otel.py::test_attribute_value_matches_helper PASSED
```

### DIFF
--- a/pydantic_evals/pydantic_evals/otel/span_tree.py
+++ b/pydantic_evals/pydantic_evals/otel/span_tree.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import json
 import re
 from collections.abc import Callable, Iterator, Sequence
 from dataclasses import dataclass, field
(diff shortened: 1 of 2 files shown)


debu-sinha:
Closing this since I am not pursuing this direction anymore.

PR #4457, opened 2026-02-28
fix: run args_validator in activity context for TemporalAgent

### LINKED ISSUE #4456: `args_validator` does not work with `TemporalAgent` if there is I/O
### Initial Checks

### Description

When using an `args_validator` that performs any kind of I/O with the `TemporalAgent`, this results in the following error:

> Cannot access platform.uname.__call__ from inside a workflow. If this is code from a module not used in a workflow or known to only be used deterministically from a workflow, mark the import as pass through.

This appears to be caused by the validators being run within the workflow context and not in an activity.

```Python

```

### Logfire Trace

_No response_

- Python: `3.13.11`
- Pydantic AI: `v1.63.0`
- temporalio: `v1.22.0`
- LLM provider SDK:

gitbalaji:
- Closes #4456

## Summary

When an `args_validator` performed any I/O (e.g. a database lookup, HTTP call), it would crash inside a `TemporalAgent` workflow with:

> Cannot access platform.uname.__call__ from inside a workflow.

This happened because `args_validator_func` was called during the validation phase in `_tool_manager.py`, which runs in the Temporal **workflow sandbox** where I/O is prohibited. The actual tool execution was correctly dispatched to an activity, but the validator never made it there.

**Fix:**

- Added `_skip_args_validator_func: ContextVar[bool]` in `_tool_manager.py`. When `True`, `_validate_tool_args` skips calling `args_validator_func`.
- `TemporalAgent._temporal_overrides` sets this to `True` when running inside a workflow, deferring the validator call.
- `_call_tool_in_activity` now calls `args_validator_func` after re-validating the deserialized args — inside the activity where I/O is permitted. `ModelRetry` is caught and returned as `_ModelRetry` for correct retry semantics.
- For tools with `tool_activity_config=False` (inline in workflow, no activity dispatch), `TemporalFunctionToolset.call_tool` calls `args_validator_func` directly to preserve exactly-once semantics.

## Test plan

- `test_args_validator_runs_in_activity_context` — verifies the validator runs in activity context (not workflow) by checking `workflow.in_workflow()` returns `False` inside the validator
- `test_args_validator_outside_workflow` — verifies normal (non-Temporal) behavior is unaffected
- `test_args_validator_inline_tool_in_workflow` — verifies `tool_activity_config=False` path succeeds
- All 72 existing Temporal tests continue to pass
- All 14 existing `args_validator` tests in `test_tools.py` continue to pass

### Pre-Review Checklist

### Pre-Merge Checklist


tests/test_temporal.py
```diff
@@ -1927,6 +1927,133 @@ async def test_temporal_agent_mcp_server_activity_disabled(client: Client):
...
+            task_queue=TASK_QUEUE,
+        )
+    # The workflow completes successfully, confirming the validator was invoked without error
+    # and the tool ran correctly in the workflow context.
+    assert output == snapshot('{"get_weather_validated_inline":"sunny in a"}')
+
+
 @workflow.defn
 class DirectStreamWorkflow:
     @workflow.run
```
jerry-reevo, line 2132: Missing assertion that the validator was actually called — should clear and check `_args_validator_call_in_workflow` like the other two tests do.

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_toolset.py
```diff
@@ -126,8 +127,17 @@ async def _call_tool_in_activity(
...
+
+        `args_validator_func` is also called here (not in the workflow) so that it can safely perform I/O.
         """
         args_dict = tool.args_validator.validate_python(tool_args)
+        if tool.args_validator_func is not None:
+            try:
+                result = tool.args_validator_func(ctx, **args_dict)
+                if inspect.isawaitable(result):
+                    await result
+            except ModelRetry as e:
```
jerry-reevo, line 139: No test exercises `ModelRetry` from `args_validator_func` in an activity — worth adding one to confirm the `_ModelRetry` roundtrip and retry machinery work.

jerry-reevo (commented):
Nice fix — the ContextVar approach is clean and the `ModelRetry` handling in the activity is important for correctness. Two small gaps in test coverage noted inline.

DouweM:
This approach of skipping it in the normal loop and duplicating it in Temporal is a non-starter, sorry. We need a more generic approach. I'll comment on the issue.

Issue #4425, opened 2026-02-25
OpenAIJsonSchemaTransformer doesn't detect items: {} as strict-incompatible, causing API errors with bare list type


mohitmunjal:
### Initial Checks

- [x] I'm using the [latest version](https://github.com/pydantic/pydantic-ai/releases/latest) of Pydantic AI
- [x] I've searched for my issue in [the issue tracker](https://github.com/pydantic/pydantic-ai/issues) before opening this issue

### Description

  When using a tool with a bare `list` parameter (e.g., `def foo(items: list)`), pydantic-ai sends the schema to OpenAI with `strict=True`, but OpenAI rejects it because
  `items: {}` is invalid in strict mode.

  The `OpenAIJsonSchemaTransformer` correctly handles other strict-incompatible features (like `default` values, `oneOf`, partial `required`, etc.) by either auto-fixing
  them or marking the schema as `is_strict_compatible = False`. However, it doesn't check for empty `items: {}` in arrays.

  ### Minimal Reproduction

  ```python
  import asyncio
  from pydantic_ai import Agent
  from pydantic_ai.models.openai import OpenAIResponsesModel


  def tool_with_bare_list(items: list) -> str:
      """A tool that takes a bare list parameter."""
      return f"Got {len(items)} items"


  async def main():
      agent = Agent(
          model=OpenAIResponsesModel("gpt-5.1"),
          tools=[tool_with_bare_list],
      )
      result = await agent.run("Call tool_with_bare_list with items ['a', 'b']")
      print(result.output)


  asyncio.run(main())
```

### Expected Behavior

  Either:
  1. pydantic-ai detects that items: {} is not strict-compatible and sets strict=False, allowing the call to succeed in lenient mode
  2. Or raises a helpful UserError explaining that list needs a type parameter (e.g., list[str])

### Actual Behavior
pydantic_ai.exceptions.ModelHTTPError: status_code: 400, model_name: gpt-5.1, body: {'message': "Invalid schema for function 'tool_with_bare_list': In context=('properties', 'items', 'items'), schema must have a 'type' key.", 'type': 'invalid_request_error', 'param': 'tools[0].parameters', 'code': 'invalid_function_parameters'}

## Root Cause

In `pydantic_ai/profiles/openai.py`, the OpenAIJsonSchemaTransformer.transform() method handles various strict-incompatible features but the transformer doesn't check for empty items in arrays.

[...]

PR #4461, opened 2026-02-28
fix: detect empty `items: {}` as strict-incompatible in OpenAIJsonSchemaTransformer

giulio-leone:
## Problem

When a tool parameter uses a bare `list` type (e.g. `def foo(items: list)`), Pydantic generates the schema `{"type": "array", "items": {}}`. OpenAI strict mode rejects empty `items` schemas, causing a 400 error at runtime:

```
BadRequestError: Error code: 400 - Invalid schema for function ...
```

The `OpenAIJsonSchemaTransformer` correctly handles other strict-incompatible patterns (`default`, `oneOf`, partial `required`, custom `format`, etc.) but does not check for empty `items: {}` in array schemas.

## Fix

Add detection for empty `items: {}` in array schemas during the `transform()` walk:

- **`strict=None` (auto-detect)**: Mark `is_strict_compatible = False`, which causes the tool to fall back to `strict=False`
- **`strict=True` (forced)**: Strip the empty `items` key and note it in the description (consistent with how `_STRICT_INCOMPATIBLE_KEYS` like `minLength`, `format`, etc. are handled)

## Tests

- `test_transformer_detects_empty_items_as_strict_incompatible`: verifies `strict=None` correctly flags the schema
- `test_transformer_strips_empty_items_when_strict_true`: verifies `strict=True` strips the key and adds description note

All 167 existing OpenAI model tests continue to pass.

Fixes #4425

### DIFF
--- a/tests/models/test_openai.py
+++ b/tests/models/test_openai.py
@@ -4373,6 +4373,22 @@ def test_transformer_strips_empty_items_when_strict_true():
     assert 'items={} (any type)' in result['properties']['items_param'].get('description', '')
 
 
+def test_transformer_strips_empty_items_preserves_description():
+    """When strict=True and the array already has a description, empty items note is appended."""
+
+    schema: dict[str, Any] = {
+        'type': 'object',
+        'properties': {
+            'items_param': {'type': 'array', 'items': {}, 'description': 'A list of values'},
+        },
+        'required': ['items_param'],
+    }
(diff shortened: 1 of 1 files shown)


giulio-leone:
Closing in favor of #4474 which properly detects empty `items: {}` as strict-incompatible.

PR #4463, opened 2026-02-28
fix: SpanQuery has_attributes matches dict values stored as JSON strings

### LINKED ISSUE #4448: Evals: SpanQuery format mismatch

giulio-leone:
## Summary

Fixes #4448

When instrumentation libraries (e.g. logfire) serialize complex values like dicts to JSON strings before storing them as OTel attributes, the direct equality comparison in `SpanQuery.has_attributes` fails because `str != dict`.

## Changes

- **`pydantic_evals/otel/span_tree.py`**: Added `_attribute_value_matches()` helper that transparently JSON-decodes stored string attributes when the query value is not a string. Includes type-strict comparison on the JSON path to prevent bool/int confusion (e.g. `json.loads('1')` matching `True`).
- **`tests/evals/test_otel.py`**: Added test `test_span_node_matches_json_serialized_attributes` verifying dict query values match JSON-serialized string attributes.

## Before/After

**Before:** `SpanQuery(has_attributes={'data': {'foo': 1}})` fails to match a span where `data` attribute is stored as `'{"foo":1}'`
**After:** The query transparently JSON-decodes the stored string and compares as Python objects

## Risk Assessment

**Low** — Only affects the comparison logic for `has_attributes` when the query value is not a native `AttributeValue` type. All existing scalar comparisons go through the unchanged fast path (`stored == query_value`).

### DIFF
--- a/tests/evals/test_otel.py
+++ b/tests/evals/test_otel.py
@@ -246,7 +246,7 @@ async def test_span_node_matches_json_serialized_attributes():
 
 def test_attribute_value_matches_invalid_json_returns_false():
     """When stored string is not valid JSON and query is non-string, return False."""
-    from pydantic_evals.otel.span_tree import _attribute_value_matches
+    from pydantic_evals.otel.span_tree import _attribute_value_matches  # pyright: ignore[reportPrivateUsage]
 
     assert _attribute_value_matches('not-valid-json{', {'key': 'value'}) is False
 


giulio-leone:
Closing in favor of #4471 which has a cleaner implementation.

PR #4465, opened 2026-02-28
fix(xai): promote cache_read_tokens to top-level RequestUsage field

### LINKED ISSUE #4360: xAI cache_read_tokens ends up in details instead of top-level field
### Initial Checks

### Problem

When using xAI/Grok models, `cache_read_tokens` from the API response gets put into `usage.details["cache_read_tokens"]` instead of the actual `usage.cache_read_tokens` field.

The xAI API returns `cached_prompt_text_tokens` in the usage response, and `_extract_usage` in `xai.py` correctly maps it to `cache_read_tokens` in the `usage_data` dict. But then it ends up in `details` because of this logic:

```python
details = {k: v for k, v in usage_data.items() if k not in {'prompt_tokens', 'completion_tokens'}}

extracted = RequestUsage.extract(
    dict(model=model, usage=usage_data),
    ...
    details=details or None,
)
```

`RequestUsage.extract()` uses `genai-prices` which doesn't map cache_read_tokens to the top-level field, so it only survives in details.

### Expected behavior

`usage.cache_read_tokens` should be populated, not just `usage.details["cache_read_tokens"]`.

### Suggested fix

After calling `RequestUsage.extract()`, copy cache tokens from details to the top-level fields if they weren't already set:

```python
if not extracted.cache_read_tokens and 'cache_read_tokens' in details:
    extracted.cache_read_tokens = details['cache_read_tokens']
if not extracted.cache_write_tokens and 'cache_write_tokens' in details:
    extracted.cache_write_tokens = details['cache_write_tokens']
```

```Python
from pydantic_ai import Agent

agent = Agent("xai:grok-4-1-fast-reasoning", system_prompt="You are helpful.")

# first run populates cache
_ = await agent.run("What is 2+2?")

# second call should hit cache
result = await agent.run("What is 2+2?")
usage = result.usage()

print(usage.cache_read_tokens)  # 0
print(usage.details.get("cache_read_tokens"))  # 128 (actual value)
```

### Logfire Trace

_No response_

- Python: `3.13.1`
- Pydantic AI: `pydantic-ai-slim` 1.60.0
- LLM provider SDK: `xai-sdk` 1.6.0

giulio-leone:
## Problem

When xAI returns `cached_prompt_text_tokens`, the value is correctly mapped to `cache_read_tokens` in the usage dict by `_extract_usage`, but after `RequestUsage.extract()` it only survives in `RequestUsage.details` because genai-prices doesn't map it for all xAI models yet.

This means `usage.cache_read_tokens` is always `0` even when cache tokens are present.

Fixes #4360

## Solution

After `RequestUsage.extract()`, promote `cache_read_tokens` and `cache_write_tokens` from `details` to the top-level `RequestUsage` fields when genai-prices didn't set them.

### Before
```python
usage.cache_read_tokens  # 0
usage.details  # {'cache_read_tokens': 157}
```

### After
```python
usage.cache_read_tokens  # 157
usage.details  # {'cache_read_tokens': 157}
```

## Changes

- `pydantic_ai_slim/pydantic_ai/models/xai.py`: After extract(), copy cache tokens from details to top-level fields when they weren't set.
- `tests/models/test_xai.py`: Updated inline snapshots to reflect `cache_read_tokens` in top-level `RequestUsage`.

## Testing

- All 99 xAI tests pass (1 skipped) — 2 consecutive clean runs
- `ruff check` passes

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/xai.py
+++ b/pydantic_ai_slim/pydantic_ai/models/xai.py
@@ -1083,7 +1083,7 @@ def _extract_usage(
     # doesn't map them (e.g. for newer xAI models not yet in the pricing DB).
     if not extracted.cache_read_tokens and details.get('cache_read_tokens'):
         extracted.cache_read_tokens = details['cache_read_tokens']
-    if not extracted.cache_write_tokens and details.get('cache_write_tokens'):
+    if not extracted.cache_write_tokens and details.get('cache_write_tokens'):  # pragma: no cover
         extracted.cache_write_tokens = details['cache_write_tokens']
 
     return extracted


giulio-leone:
Closing to focus on higher-impact fixes. Happy to reopen if there's interest.

PR #4466, opened 2026-02-28
fix: use global retry count for output validators on tool output path

### LINKED ISSUE #4385: `@agent.output_validator` receives inconsistent `ctx.retry` depending on text vs tool output path
@agent.output_validator functions receive different ctx.retry values depending on whether the model returned plain text or called an output tool.

Root Cause:

GraphAgentState.retries (the global output validation counter) is not mapped to RunContext on the tool output path. The text path explicitly maps it (_agent_graph.py:796):

    run_context = replace(run_context, retry=ctx.state.retries)

But the tool output path goes through ToolManager._call_tool (_tool_manager.py:190-194), which builds RunContext from the per-tool dict instead:

    ctx = replace(self.ctx, retry=self.ctx.retries.get(name, 0))

Since output tools share this codepath with function tools, the output validator only sees per-tool failures, not the global counter.

When an output validator fails on the tool path, it updates both counters — failed_tools.add(name) in _call_tool increments the per-tool count, and increment_retries in process_tool_calls increments the global count — but the validator itself only had access to the per-tool count via ctx.retry. Which is also kind of awkward but works I guess.

When it matters:

In the common case (single output tool, no other failure types), both counters stay in sync. They diverge when other failure types are mixed in before the output tool fails:

- Empty response → global +1, per-tool unchanged
- Unknown tool call → global +1, per-tool unchanged
- Text validation failure → global +1, per-tool unchanged

An output validator checking ctx.retry to implement "be lenient on last attempt" logic would be blind to retry budget consumed by these unrelated failures. The per-tool view is out of global scope and doesn't make logical sense for output validators.

giulio-leone:
## Problem

`@agent.output_validator` receives inconsistent `ctx.retry` values depending on whether the model returned output as plain text or via a tool call:

- **Text path** (`_handle_text_response`): `ctx.retry = ctx.state.retries` (global output-validation counter)
- **Tool path** (`_build_tool_context`): `ctx.retry = ctx.retries.get(tool_name, 0)` (per-tool counter)

This means output validators checking `ctx.retry` to implement "lenient on last attempt" logic are blind to retry budget consumed by unrelated failures (empty response, unknown tool call, text validation failure) on the tool path.

Fixes #4385

## Solution

Add `retry_override` parameter to `ToolManager._build_tool_context()` and `validate_tool_call()`. When processing output tool calls in `process_tool_calls()`, pass the global retry counter (`ctx.state.retries`) as the override so the output validator sees the same retry value regardless of output path.

## Changes

- `pydantic_ai_slim/pydantic_ai/_tool_manager.py`: Add `retry_override` parameter to `_build_tool_context` and `validate_tool_call`
- `pydantic_ai_slim/pydantic_ai/_agent_graph.py`: Pass `retry_override=ctx.state.retries` when validating output tool calls

## Testing

- All 197 agent tests pass
- All 72 output-validator/retry-related tests pass (2 consecutive clean runs)
- `ruff check` passes

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/_agent_graph.py
+++ b/pydantic_ai_slim/pydantic_ai/_agent_graph.py
@@ -970,7 +970,10 @@ async def process_tool_calls(  # noqa: C901
             # Validate and execute the output tool call — unlike deferred tools,
             # output tools track retries and can be skipped if a final_result exists.
             try:
-                validated = await tool_manager.validate_tool_call(call)
+                validated = await tool_manager.validate_tool_call(
+                    call,
+                    retry_override=ctx.state.retries,
+                )
             except exceptions.UnexpectedModelBehavior as e:
                 # If we already have a valid final result, don't fail the entire run
                 # This allows exhaustive strategy to complete successfully when at least one output tool is valid
--- a/pydantic_ai_slim/pydantic_ai/_tool_manager.py
+++ b/pydantic_ai_slim/pydantic_ai/_tool_manager.py
@@ -168,6 +168,7 @@ class ToolManager(Generic[AgentDepsT]):
         allow_partial: bool,
         approved: bool = False,
         metadata: Any = None,
+        retry_override: int | None = None,
     ) -> RunContext[AgentDepsT]:
         """Build the execution context for a tool call."""
         assert self.ctx is not None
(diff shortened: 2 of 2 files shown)


giulio-leone:
Closing to focus on higher-impact fixes. Happy to reopen if there's interest.

PR #4469, opened 2026-02-28
fix(evals): SpanQuery has_attributes matches dict/list values against JSON-serialized form

### LINKED ISSUE #4448: Evals: SpanQuery format mismatch

giulio-leone:
## Problem

`SpanQuery(has_attributes={'data': {'foo': 1, 'bar': True}})` returns no match even when the span has a `data` attribute with that exact value.

Logfire stores complex types (dicts, lists) as JSON strings in span attributes. The existing `has_attributes` comparison uses strict `==` equality, which fails because:
- Stored: `'{"foo":1,"bar":true}'` (string)
- Query: `{'foo': 1, 'bar': True}` (dict)

Users must currently query with the raw JSON string, which is unintuitive.

## Solution

When direct equality fails and the query value is a complex type (not str/bool/int/float), parse the stored JSON string and compare Python objects. This correctly handles:
- Dict values with any key ordering
- Nested structures
- List values
- Falls back gracefully if the stored value isn't valid JSON

## Testing

- Added `test_span_query_dict_attribute_matching` covering:
  - String query (existing behavior preserved)
  - Dict query matching
  - Dict with reversed key order matching
  - Non-matching dict rejection
  - List value matching
  - Non-matching list rejection
- All 26 otel tests pass across 2 consecutive clean runs

Fixes #4448

### DIFF
--- a/pydantic_evals/pydantic_evals/otel/span_tree.py
+++ b/pydantic_evals/pydantic_evals/otel/span_tree.py
@@ -267,7 +267,7 @@ class SpanNode:
 
         # Attribute conditions
         if (has_attributes := query.get('has_attributes')) and not all(
-            self.attributes.get(key) == value for key, value in has_attributes.items()
+            _attribute_matches(self.attributes.get(key), value) for key, value in has_attributes.items()
         ):
             return False
         if (has_attributes_keys := query.get('has_attribute_keys')) and not all(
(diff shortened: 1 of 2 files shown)


giulio-leone:
Closing in favor of #4471 which has a more complete fix.

PR #3591, opened 2025-11-28
Transform  and  to text parts for Google models that don't support tools

sidg1215:
Fixes #3149.

PR #4472, opened 2026-02-28
fix: convert tool call/return history to text for Gemini image models

### LINKED ISSUE #3149: Gemini image model does not accept history with tool calls or thinking
### Initial Checks

### Description

When `gemini-2.5-flash-image-preview` with https://ai.pydantic.dev/builtin-tools/#image-generation-tool (see example code there) is passed message history from another model with tool calls or thinking, the API returns an error:

Tool calls:

```
Function calling is not enabled for models/gemini-2.5-flash-image-preview
```

Thinking signatures:

```
Request contains an invalid argument.
```

Thinking signatures we can simply omit (while leaving the thinking summaries in), but tool calls are a bit more tricky, as the tool call results may be relevant to image generation, and may not be repeated in text messages.

### Example Code

```Python

```

```Text
Latest
```

Br1an67:
- Closes #3149

## Summary

Gemini image models (e.g. `gemini-2.0-flash-preview-image-generation`) reject `FunctionCall`/`FunctionResponse` parts in request content. When an agent has tool call history and then calls such a model, the API returns an error.

This PR adds a `supports_tools` parameter to `_map_messages()` and `_content_model_response()`. When `supports_tools=False`:

- `ToolCallPart` and `ToolReturnPart` are converted to plain text via new `as_text()` methods on `BaseToolCallPart` and `BaseToolReturnPart`
- The text format uses a fenced block style that preserves `tool_call_id` for parallel tool call disambiguation:
  ```
  -----BEGIN TOOL_CALL id="call_123" name="get_weather"-----
  {"city": "London"}
  -----END TOOL_CALL-----
  ```
- Thinking part signatures and thought flags are omitted (image models don't use them)

This follows the approach discussed in #3591 review comments:
- `as_text()` on the part classes (not inline in google.py)
- Include `tool_call_id` (parallel calls may share names)
- Fenced block format similar to OpenAI text files
- Use `model_response_str()` without prefixes for args

### Pre-Review Checklist

### Pre-Merge Checklist

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/messages.py
+++ b/pydantic_ai_slim/pydantic_ai/messages.py
@@ -935,6 +935,20 @@ class BaseToolReturnPart:
         """Return `True` if the tool return has content."""
         return self.content is not None  # pragma: no cover
 
+    def as_text(self) -> str:
+        """Convert this tool return to a text representation.
+
+        This is used when sending message history to models that don't support tool calls,
+        such as image generation models.
+        """
+        return '\n'.join(
+            [
+                f'-----BEGIN TOOL_RETURN id="{self.tool_call_id}" name="{self.tool_name}"-----',
+                self.model_response_str(),
+                '-----END TOOL_RETURN-----',
+            ]
+        )
+
     __repr__ = _utils.dataclasses_no_defaults_repr
 
 
@@ -1305,6 +1319,20 @@ class BaseToolCallPart:
             return any(self.args.values())
         else:
             return bool(self.args)
+
+    def as_text(self) -> str:
+        """Convert this tool call to a text representation.
+
+        This is used when sending message history to models that don't support tool calls,
+        such as image generation models.
+        """
+        return '\n'.join(
+            [
+                f'-----BEGIN TOOL_CALL id="{self.tool_call_id}" name="{self.tool_name}"-----',
+                self.args_as_json_str(),
+                '-----END TOOL_CALL-----',
+            ]
+        )
 
     __repr__ = _utils.dataclasses_no_defaults_repr
 
--- a/pydantic_ai_slim/pydantic_ai/models/google.py
+++ b/pydantic_ai_slim/pydantic_ai/models/google.py
@@ -679,17 +679,20 @@ class GoogleModel(Model):
                     elif isinstance(part, UserPromptPart):
                         message_parts.extend(await self._map_user_prompt(part))
                     elif isinstance(part, ToolReturnPart):
-                        message_parts.append(
-                            {
-                                'function_response': {
-                                    'name': part.tool_name,
-                                    'response': part.model_response_object(),
-                                    'id': part.tool_call_id,
+                        if self.profile.supports_tools:
+                            message_parts.append(
+                                {
+                                    'function_response': {
+                                        'name': part.tool_name,
+                                        'response': part.model_response_object(),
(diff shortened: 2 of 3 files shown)


pydantic_ai_slim/pydantic_ai/models/google.py
```diff
@@ -679,17 +679,20 @@ async def _map_messages(  # noqa: C901
...
                                 }
-                            }
-                        )
+                            )
+                        else:
+                            message_parts.append({'text': part.as_text()})
                     elif isinstance(part, RetryPromptPart):
-                        if part.tool_name is None:
+                        if part.tool_name is None or not self.profile.supports_tools:
                             message_parts.append({'text': part.model_response()})
```
DouweM, line 696: We should also have the tool name/ID stuff here as this content specifically relates to an earlier tool call. So `RetryPromptPart` also needs `as_text`.

pydantic_ai_slim/pydantic_ai/models/google.py
```diff
@@ -1062,45 +1065,50 @@ def timestamp(self) -> datetime:
...
         part: PartDict = {}
-        if (
-            item.provider_details
-            and (thought_signature := item.provider_details.get('thought_signature'))
-            and (m.provider_name == provider_name or item.provider_name == provider_name)
-        ):
-            part['thought_signature'] = base64.b64decode(thought_signature)
-        elif thinking_part_signature:
-            part['thought_signature'] = base64.b64decode(thinking_part_signature)
+        if supports_tools:
```
DouweM, line 1074: Why do the thought signature only when tools are supported?

pydantic_ai_slim/pydantic_ai/models/google.py
```diff
@@ -1062,45 +1065,50 @@ def timestamp(self) -> datetime:
...
+                    part['thought_signature'] = b'skip_thought_signature_validator'
+                # Only the first function call requires a signature
+                function_call_requires_signature = False
+            else:
+                part['text'] = item.as_text()
         elif isinstance(item, TextPart):
             part['text'] = item.content
         elif isinstance(item, ThinkingPart):
-            if item.provider_name == provider_name and item.signature:
+            if supports_tools and item.provider_name == provider_name and item.signature:
```
DouweM, line 1104: Unclear why this should be specific to `supports_tools`

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -935,6 +935,20 @@ def has_content(self) -> bool:
         """Return `True` if the tool return has content."""
         return self.content is not None  # pragma: no cover
 
+    def as_text(self) -> str:
```
DouweM, line 938: Since this is currently exclusive to Google, let's move this to a helper function in that file, like `_inline_text_file_part` is currently exclusive to OpenAI. We can move it to the message types later if other classes want to use this as well.

If we add this here now, I'd want to do the same thing with the OpenAI `BEGIN FILE` stuff.

DouweM (changes requested):

hxaxd:
Hi @Br1an67 @DouweM, if this is no longer being worked on, I'd be happy to take over the remaining work here. Please let me know if that would be helpful.

DouweM:
That image model is very old, so if this is issue is still relevant with newer models, we'll need a new PR.

PR #4494, opened 2026-03-02
Use FastAPI `EventSourceResponse` in `streaming_response()`

### LINKED ISSUE #4493: Use FastAPI EventSourceResponse in UIEventStream.streaming_response()
## Summary

`UIEventStream.streaming_response()` currently uses Starlette's `StreamingResponse` with Python-side string encoding via `encode_stream()`. FastAPI 0.135.0+ ships a built-in `EventSourceResponse` (at `fastapi.sse`) that performs **Pydantic serialization on the Rust side**, which could significantly improve throughput for SSE-based streaming endpoints.

## Current implementation

https://github.com/pydantic/pydantic-ai/blob/main/pydantic_ai_slim/pydantic_ai/ui/_event_stream.py#L110-L124

```python
def streaming_response(self, stream: AsyncIterator[EventT]) -> StreamingResponse:
    from starlette.responses import StreamingResponse

    return StreamingResponse(
        self.encode_stream(stream),
        headers=self.response_headers,
        media_type=self.content_type,
    )
```

Each event goes through `encode_event()` (Python-side JSON serialization) → string → `StreamingResponse`.

## Proposed change

When FastAPI >= 0.135.0 is available, `streaming_response()` could use `EventSourceResponse` and yield `ServerSentEvent` objects instead of pre-encoded strings:

```python
def streaming_response(self, stream: AsyncIterator[EventT]) -> EventSourceResponse:
    from fastapi.sse import EventSourceResponse, ServerSentEvent

    async def sse_stream() -> AsyncIterator[ServerSentEvent]:
        async for event in stream:
            yield ServerSentEvent(data=event)

    return EventSourceResponse(
        sse_stream(),
        headers=self.response_headers,
    )
```

## Benefits

- **Rust-side Pydantic serialization** — FastAPI's `EventSourceResponse` serializes Pydantic models on the Rust side when type annotations are available, which is significantly faster than Python-side `json.dumps()`
- **Automatic keep-alive pings** — 15-second ping comments prevent proxy timeouts during long LLM generations (common with tool-heavy agents)
- **Automatic headers** — `Cache-Control: no-cache` and `X-Accel-Buffering: no` (Nginx compatibility) are set automatically, removing boilerplate from subclasses

## Considerations

- FastAPI is an optional dependency — the fallback to `starlette.responses.StreamingResponse` should be preserved when FastAPI is not installed
- The `content_type` property and custom `response_headers` from subclasses (e.g., `VercelAIEventStream`) would need to be forwarded to `EventSourceResponse`
- `EventSourceResponse` natively emits `text/event-stream`, which aligns with the existing `SSE_CONTENT_TYPE` default
- The `encode_event()` / `encode_stream()` pipeline could be bypassed when using `EventSourceResponse` with `ServerSentEvent(data=...)`, since the response class handles serialization

## Environment

- FastAPI 0.135.1 (released 2026-03-01) — `from fastapi.sse import EventSourceResponse, ServerSentEvent`
- Relevant docs: https://fastapi.tiangolo.com/tutorial/server-sent-events/

Butch78:
## Summary

Hey team! 

Love the project and everything pydantic does, please let me know if I missed something.

This is a small, backward-compatible change to `UIEventStream.streaming_response()` that uses FastAPI's `EventSourceResponse` when available (FastAPI >= 0.135.0), falling back to the existing Starlette `StreamingResponse` otherwise.

### What this improves

- **Automatic keep-alive pings** — 15-second ping comments prevent proxy/CDN timeouts during long LLM generations (common with tool-heavy agents)
- **Automatic SSE headers** — `Cache-Control: no-cache` and `X-Accel-Buffering: no` (Nginx compatibility) are set automatically
- **Cached import check** using `@lru_cache` — the FastAPI SSE classes are resolved once per process, zero overhead after the first call
- Uses `raw_data=` (not `data=`) since `encode_event()` already returns a fully-formatted SSE string, avoiding double-encoding

### What this does NOT do

This PR doesn't provide a serialization speedup — we still call `encode_event()` on the Python side before passing the string to `ServerSentEvent(raw_data=...)`. The Rust-side serialization benefit would require a deeper refactor where `encode_event()` is bypassed entirely and Pydantic models are passed via `ServerSentEvent(data=model)`. That's a much bigger change affecting all subclasses (VercelAI, AG-UI, etc.) and could be explored as a follow-up.

### Questions for reviewers

I'd love to get your thoughts on the approach — in particular:
- Is the `lru_cache` pattern the right way to handle the optional FastAPI dependency here, or would you prefer something different?
- Should we gate this on detecting a FastAPI app (via the request) rather than just checking if `fastapi.sse` is importable?
- Any concerns about depending on the `raw_data` parameter of `ServerSentEvent`?

Happy to adjust based on your feedback!

Closes #4493

## Test plan

- [x] All 121 existing UI tests pass (`test_ui.py`, `test_vercel_ai.py`, `test_ag_ui.py`)
- [ ] Manual testing with a FastAPI app to verify `EventSourceResponse` path
- [ ] Manual testing with plain Starlette to verify fallback path

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/ui/_event_stream.py
+++ b/pydantic_ai_slim/pydantic_ai/ui/_event_stream.py
@@ -123,19 +123,21 @@ class UIEventStream(ABC, Generic[RunInputT, EventT, AgentDepsT, OutputDataT]):
     def streaming_response(self, stream: AsyncIterator[EventT]) -> StreamingResponse:
         """Generate a streaming response from a stream of protocol-specific events.
 
-        When FastAPI >= 0.135.0 is available, uses `EventSourceResponse` for Rust-side
-        serialization and automatic keep-alive pings. Falls back to Starlette's
-        `StreamingResponse` otherwise.
+        When FastAPI >= 0.135.0 is available and the content type is SSE, uses
+        `EventSourceResponse` to integrate with FastAPI's SSE handling (including
+        automatic keep-alive pings and appropriate headers). Falls back to Starlette's
+        `StreamingResponse` otherwise. Serialization/encoding of events is still
+        handled by `encode_event()`.
         """
-        # Prefer FastAPI's EventSourceResponse for Rust-side serialization + keep-alive
-        if fastapi_sse := _get_fastapi_sse():
+        # Prefer FastAPI's EventSourceResponse for keep-alive pings and SSE headers
+        if self.content_type == SSE_CONTENT_TYPE and (fastapi_sse := _get_fastapi_sse()):
             EventSourceResponse, ServerSentEvent = fastapi_sse
 
             async def sse_stream() -> AsyncIterator[ServerSentEvent]:
                 async for event in stream:
                     yield ServerSentEvent(raw_data=self.encode_event(event))
 
-            return EventSourceResponse(  # type: ignore[return-value]
+            return EventSourceResponse(
                 sse_stream(),
                 headers=self.response_headers,
             )
--- a/tests/test_ui.py
+++ b/tests/test_ui.py
@@ -722,3 +722,93 @@ def test_dummy_adapter_dump_messages():
     messages = [ModelRequest(parts=[UserPromptPart(content='Hello')])]
     result = DummyUIAdapter.dump_messages(messages)
     assert result == messages
+
+
+async def test_streaming_response_uses_fastapi_sse_when_available(monkeypatch: pytest.MonkeyPatch):
+    """When FastAPI SSE classes are available and content type is SSE, uses EventSourceResponse."""
+    from unittest.mock import MagicMock
+
+    from pydantic_ai.ui._event_stream import _get_fastapi_sse
+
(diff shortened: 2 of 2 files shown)


Butch78:
Thanks for the review! All four items have been addressed in 64c2eac:

1. **Content-type gate** — The FastAPI SSE branch is now gated on `self.content_type == SSE_CONTENT_TYPE`, so subclasses like AG-UI that negotiate non-SSE content types correctly fall back to `StreamingResponse`.

2. **Return type** — Removed the `# type: ignore[return-value]`. Since FastAPI's `EventSourceResponse` extends Starlette's `StreamingResponse`, the return type was already correct.

3. **Docstring accuracy** — Updated the docstring and comment to accurately describe the benefits (keep-alive pings, SSE headers) rather than claiming Rust-side serialization.

4. **Test coverage** — Added 3 tests covering: the FastAPI SSE path (mocking `_get_fastapi_sse`), the non-SSE content type fallback, and the FastAPI-unavailable fallback.

PR #4498, opened 2026-03-02
fix: Insert empty user message for Bedrock when only instructions are present

### LINKED ISSUE #4495: Bedrock fails on request with only instructions, no user prompt: "A conversation must start with a user message."
### Question

My prompt is of the form:

```
Here is some static preamble, telling the agent what its role is.

Here is some more specific instructions, telling the agent it's allowed to choose from the following options: {{DYNAMIC_CONTENT_SUBSTITUTED_HERE}}

Here are a list of dos and don'ts for the model:
* Always do this
* Never do that
* Do this depending on some condition
```

But when I call the agent like so:

```
deps = AgentDeps(dynamic_content=some_var_collected_from_earlier)
result = await agent.run(deps=deps)
```

I get an error from AWS Bedrock when I try and run the agent with Haiku 4.5:
```
{'Error': {'Message': 'A conversation must start with a user message. Try again with a conversation that starts with a user message.', 'Code': 'ValidationException'}
```

In my case, there's no user message. The agent's inputs are solely the static prompt with the result of a database call substituted in.

I could say `result = await agent.run("Do the thing", deps=deps)` but then I have to remember to pass "Do the thing" as the first argument at every call site. I have a poor memory, that's why I have a computer in front of me.

What's the correct way to approach this?

### Additional Context

_No response_

adtyavrdhn:
When `agent.run()` is called with only a system prompt or instructions (no user message), Bedrock rejects the request with "A conversation must start with a user message."

This applies the same fix used in `OpenAIResponsesModel`: when there are no processed messages, an explicit empty user message is added to satisfy the Bedrock API constraint.

Closes #4495

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/bedrock.py
+++ b/pydantic_ai_slim/pydantic_ai/models/bedrock.py
@@ -49,7 +49,7 @@ from pydantic_ai.profiles.anthropic import ANTHROPIC_THINKING_BUDGET_MAP
 from pydantic_ai.profiles.openai import OPENAI_REASONING_EFFORT_MAP
 from pydantic_ai.providers import Provider, infer_provider
 from pydantic_ai.providers.bedrock import BedrockModelProfile, remove_bedrock_geo_prefix
-from pydantic_ai.settings import ModelSettings
+from pydantic_ai.settings import ModelSettings, ThinkingLevel
 from pydantic_ai.tools import ToolDefinition
 
 if TYPE_CHECKING:
@@ -590,8 +590,6 @@ class BedrockConverseModel(Model):
         elif variant == 'qwen' and 'reasoning_config' not in existing:
             if thinking is not False:
                 # Qwen only supports low/high; map others to closest
-                from ..settings import ThinkingLevel
-
                 level_map: dict[ThinkingLevel, str] = {
                     True: 'high',
                     'minimal': 'low',
--- a/docs/capabilities.md
+++ b/docs/capabilities.md
@@ -126,6 +126,15 @@ Constraint fields like `allowed_domains` or `blocked_domains` require the builti
 WebSearch(allowed_domains=['example.com'])
 ```
 
+All of these capabilities are subclasses of [`BuiltinOrLocalTool`][pydantic_ai.capabilities.BuiltinOrLocalTool], which you can use directly or subclass to build your own provider-adaptive tools. For example, to pair [`CodeExecutionTool`][pydantic_ai.builtin_tools.CodeExecutionTool] with a local fallback:
+
+```python {title="custom_builtin_or_local.py" test="skip" lint="skip"}
+from pydantic_ai.builtin_tools import CodeExecutionTool
+from pydantic_ai.capabilities import BuiltinOrLocalTool
+
+cap = BuiltinOrLocalTool(builtin=CodeExecutionTool(), local=my_local_executor)
+```
+
 ### PrepareTools
 
 [`PrepareTools`][pydantic_ai.capabilities.PrepareTools] wraps a [`ToolsPrepareFunc`][pydantic_ai.tools.ToolsPrepareFunc] as a capability, for filtering or modifying [tool definitions](tools.md) per step:
--- a/pydantic_ai_slim/pydantic_ai/agent/__init__.py
+++ b/pydantic_ai_slim/pydantic_ai/agent/__init__.py
@@ -47,7 +47,7 @@ from .._template import TemplateStr, validate_from_spec_args
 from .._tool_manager import ParallelExecutionMode, ToolManager
 from ..builtin_tools import AbstractBuiltinTool
 from ..capabilities import AbstractCapability, CombinedCapability
-from ..capabilities.builtin_or_local import BuiltinTool as BuiltinToolCap
+from ..capabilities.builtin_tool import BuiltinTool as BuiltinToolCap
 from ..capabilities.history_processor import HistoryProcessor as HistoryProcessorCap
 from ..models.instrumented import InstrumentationSettings, InstrumentedModel, instrument_model
 from ..output import OutputDataT, OutputSpec, StructuredDict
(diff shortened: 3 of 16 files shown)


tests/models/test_bedrock.py
```diff
@@ -3244,3 +3244,36 @@ class Response(TypedDict):
...
+    """
+    model = BedrockConverseModel('us.amazon.nova-micro-v1:0', provider=bedrock_provider)
+    # Only system prompt, no user messages
+    messages: list[ModelMessage] = [ModelRequest(parts=[SystemPromptPart(content='You are a helpful assistant.')])]
+    _, bedrock_messages = await model._map_messages(  # pyright: ignore[reportPrivateUsage]
+        messages,
+        ModelRequestParameters(),
+        None,
+    )
+    assert bedrock_messages == [{'role': 'user', 'content': [{'text': ''}]}]
```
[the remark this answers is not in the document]
DouweM: Yep we should verify that this works against the real Bedrock API -- some APIs complain on empty text parts.

adtyavrdhn:
The VCR cassette was not recorded against the real Bedrock API. We need to re-record it with real AWS credentials to verify Bedrock actually accepts an empty `{"text": ""}` user message — some reports suggest Bedrock rejects blank text fields with a ValidationException. If it does, we may need to use a non-empty placeholder like `"."` instead.

-> Thanks Claude lol

pydantic_ai_slim/pydantic_ai/models/bedrock.py
```diff
@@ -806,6 +806,12 @@ async def _map_messages(  # noqa: C901
                 # Note: _get_last_user_message_content ensures content doesn't already end with a cachePoint.
                 _insert_cache_point_before_trailing_documents(last_user_content)
 
+        # When there are no input messages, Bedrock will reject the request with:
+        # "A conversation must start with a user message."
+        # To avoid this, provide an explicit empty user message.
+        if not processed_messages:
+            processed_messages.append({'role': 'user', 'content': [{'text': ''}]})
```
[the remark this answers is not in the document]
adtyavrdhn: This PR addresses #4495: calling `agent.run()` with no user prompt at all (only system prompt/instructions), where `processed_messages` ends up empty. The assistant-first-in-history case you describe (passing `message_history` from a previous system-prompt-only run into a second run) is a separate edge case that would need its own fix and issue. The same limitation exists in OpenAI Responses model's equivalent guard.

adtyavrdhn:
TODO: Can we check if empty space ' ' will work and not '.'.

pydantic_ai_slim/pydantic_ai/models/bedrock.py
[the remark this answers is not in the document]
DouweM: Out of scope for this PR

tests/models/test_bedrock.py
```diff
@@ -2661,9 +2661,11 @@ async def test_bedrock_cache_messages_no_user_messages(allow_model_requests: Non
         ModelRequestParameters(),
         BedrockModelSettings(bedrock_cache_messages=True),
     )
-    # Should not crash, no cache point added since no user message
-    assert len(bedrock_messages) == 1
-    assert bedrock_messages[0]['role'] == 'assistant'
+    # Should not crash, no cache point added since no real user message.
+    # Synthetic user message is prepended because Bedrock requires conversations to start with a user turn.
+    assert len(bedrock_messages) == 2
```
DouweM: I like snapshots :)

tests/models/cassettes/test_bedrock/test_bedrock_model_instructions_only_then_message_history.yaml
```diff
@@ -0,0 +1,87 @@
+interactions:
+- request:
+    body: '{"messages": [{"role": "user", "content": [{"text": " "}]}], "system": [{"text": "Generate a short greeting."}],
```
DouweM: @adtyavrdhn Is this a real cassette proving that it works with a space?

tests/models/cassettes/test_bedrock/test_bedrock_model_instructions_only_then_message_history.yaml
```diff
@@ -0,0 +1,87 @@
...
+      amz-sdk-request:
+      - !!binary |
+        YXR0ZW1wdD0x
+      content-length:
+      - '135'
+      content-type:
+      - !!binary |
+        YXBwbGljYXRpb24vanNvbg==
+    method: POST
+    uri: https://bedrock-runtime.us-east-1.amazonaws.com/model/us.amazon.nova-micro-v1%3A0/converse
```
DouweM, line 18: I think it's worth testing this against another model on Bedrock as well, e.g. Claude, just in case there's a difference.

tests/models/cassettes/test_bedrock/test_bedrock_model_instructions_only_then_message_history.yaml
adtyavrdhn: Yep

tests/models/cassettes/test_bedrock/test_bedrock_model_instructions_only_then_message_history.yaml
adtyavrdhn: Will do

adtyavrdhn: @DouweM Anthropic does not allow whitespace text, should we go back to period?

DouweM: @adtyavrdhn Hmm, and Bedrock+Anthropic also has the  "A conversation must start with a user message." issue? It's not that we only need this for non-Anthropic models?

adtyavrdhn: Yeah this is a bedrock thing, the user who reported this also hit it with Haiku.

DouweM: Let's do the period then,.

PR #4523, opened 2026-03-04
fix: promote xAI cache_read_tokens from details to top-level field

### LINKED ISSUE #4360: xAI cache_read_tokens ends up in details instead of top-level field

adityasingh2400:
## Summary

Fixes #4360

When using xAI/Grok models, `cached_prompt_text_tokens` from the xAI API response was correctly mapped to `cache_read_tokens` inside `usage_data`, but `genai-prices` does not recognise this xAI-specific field and therefore it only survived in the `details` dict. As a result, `usage.cache_read_tokens` was always `0`, even after a cache hit.

### Root cause

In `_extract_usage` (xai.py), the code already builds `details` from all non-standard keys (including `cache_read_tokens`) and passes it to `RequestUsage.extract()`. However, `RequestUsage.extract()` uses `genai-prices` to derive the fields, and `genai-prices` does not map `cache_read_tokens` for xAI — so the value ends up only in `details`.

### Fix

After calling `RequestUsage.extract()`, promote `cache_read_tokens` and `cache_write_tokens` from `details` to their top-level fields if `genai-prices` hasn't already populated them. This mirrors the same fallback already applied to `input_tokens`/`output_tokens`.

```python
# Promote cache token fields from details to the top-level fields if genai-prices
# did not already map them.
if not extracted.cache_read_tokens and details.get('cache_read_tokens'):
    extracted.cache_read_tokens = details['cache_read_tokens']
if not extracted.cache_write_tokens and details.get('cache_write_tokens'):
    extracted.cache_write_tokens = details['cache_write_tokens']
```

### Testing

All existing xAI tests are updated to assert the correct top-level `cache_read_tokens` value (previously they only saw it in `details`). No new cassette recordings needed — the SDK mock already returns `cached_prompt_text_tokens`.

### Pre-Review Checklist

- [x] Any **AI generated code** has been reviewed line-by-line
- [x] I have added/updated tests for the fix
- [x] Tests pass locally (unit tests only — no live API calls needed)

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/xai.py
+++ b/pydantic_ai_slim/pydantic_ai/models/xai.py
@@ -1087,6 +1087,15 @@ def _extract_usage(
     if extracted.output_tokens == 0 and usage_data['completion_tokens']:
         extracted.output_tokens = usage_data['completion_tokens']
 
+    # Promote cache token fields from details to the top-level fields if genai-prices
+    # did not already map them.  xAI's `cached_prompt_text_tokens` is stored as
+    # `cache_read_tokens` in `usage_data` and therefore lands in `details`; without
+    # this step `usage.cache_read_tokens` would always be 0.
+    if not extracted.cache_read_tokens and details.get('cache_read_tokens'):
+        extracted.cache_read_tokens = details['cache_read_tokens']
+    if not extracted.cache_write_tokens and details.get('cache_write_tokens'):
+        extracted.cache_write_tokens = details['cache_write_tokens']
+
     return extracted
 
 
--- a/tests/models/test_xai.py
+++ b/tests/models/test_xai.py
@@ -255,7 +255,7 @@ async def test_xai_request_structured_response_tool_output(allow_model_requests:
             ),
             ModelResponse(
                 parts=[ToolCallPart(tool_name='get_user_country', args='{}', tool_call_id=IsStr())],
-                usage=RequestUsage(input_tokens=420, output_tokens=16, details={'cache_read_tokens': 157}),
+                usage=RequestUsage(input_tokens=420, output_tokens=16, cache_read_tokens=157, details={'cache_read_tokens': 157}),
                 model_name='grok-4-fast-non-reasoning',
                 timestamp=IsDatetime(),
                 provider_name='xai',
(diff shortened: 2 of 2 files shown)


adtyavrdhn:
Hey @adityasingh2400 

Please check my comment on the issue, I would rather add a correct gene-pricing model for xai native sdk and not use the OpenAI one instead of patching this

adityasingh2400:
Thanks for the context @adtyavrdhn — that makes sense. A proper fix in genai-prices (passing `api_flavor='chat'` and correcting the provider ID) is clearly the right approach rather than patching around it downstream. I'll close this PR. Happy to take a look at the genai-prices side if that would be helpful, otherwise will let you drive that.

Issue #3668, opened 2025-12-08
Support parallel_tool_calls configuration for gemini models


alonme:
### Description

According to the docs in the code paralllel tool calls configuration is only possible for the following
```

    Supported by:

    * OpenAI (some models, not o1)
    * Groq
    * Anthropic
```

I have a gemini model using parallel tool calls, which i want to disable, can we support them

DouweM:
@alonme Does the Gemini API/SDK support disabling parallel tool calling? If you can share docs it shouldn't be too hard to implement, but I wasn't immediately able to find how to do this.

alonme:
Thanks for the quick response,
I actually couldn't find anything too - Opened an issue https://github.com/googleapis/python-genai/issues/1831

DouweM:
@alonme Thanks, I'll close this then until they add the feature.

PR #4540, opened 2026-03-04
fix(google): batch custom function declarations in one Tool for Gemini parallel tool calling

### LINKED ISSUE #4539: GoogleModel sends one Tool per function, preventing Gemini parallel tool calling
**Bug:** `GoogleModel._get_tools` in `pydantic_ai/models/google.py` builds one `ToolDict` per function declaration. The [Gemini API](https://ai.google.dev/gemini-api/docs/function-calling?lang=python&example=meeting#parallel_function_calling) expects all custom function declarations in a **single** `Tool` for parallel function calling. With one `ToolDict` per function, the model only calls one tool per round-trip.

**Fix:** Batch all custom function declarations into one `ToolDict` before appending to the tools list. Builtin tools (e.g. `WebSearchTool`, `ImageGenerationTool`) correctly remain in separate `ToolDict` entries per the API.

Samuel-Harris:
- Closes #4539

### Summary

`GoogleModel._get_tools` previously created one `ToolDict` per function declaration. The [Gemini API](https://ai.google.dev/gemini-api/docs/function-calling?lang=python&example=meeting#parallel_function_calling) requires all custom function declarations in a single `Tool` for parallel function calling. This change batches them into one `ToolDict`, enabling the model to call multiple tools in one round-trip when appropriate.

Related: #3668 (different feature — configuring parallel tool calls; closed).

### Pre-Review Checklist

- [x] AI generated code reviewed line-by-line by the human PR author
- [x] No breaking changes (version policy)
- [x] `make format` and `make typecheck` pass
- [x] PR title fit for release changelog

### Pre-Merge Checklist

- [x] New test for the fix: `test_google_get_tools_batches_function_declarations`
- [ ] Updated documentation: N/A (internal behavior; no user-facing API change)

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/google.py
+++ b/pydantic_ai_slim/pydantic_ai/models/google.py
@@ -428,15 +428,20 @@ class GoogleModel(Model):
         return image_config
 
     def _get_tools(
-        self, model_request_parameters: ModelRequestParameters
+        self,
+        model_request_parameters: ModelRequestParameters,
+        model_settings: GoogleModelSettings | None = None,
     ) -> tuple[list[ToolDict] | None, ImageConfigDict | None]:
-        # Batch all custom function declarations in one Tool for Gemini parallel function calling.
+        parallel = model_settings is not None and model_settings.get('parallel_tool_calls', False) is True
         function_declarations = [
             _function_declaration_from_tool(t) for t in model_request_parameters.tool_defs.values()
         ]
         tools: list[ToolDict] = []
         if function_declarations:
-            tools.append(ToolDict(function_declarations=function_declarations))
+            if parallel:
+                tools.append(ToolDict(function_declarations=function_declarations))
+            else:
+                tools.extend(ToolDict(function_declarations=[fd]) for fd in function_declarations)
 
         image_config: ImageConfigDict | None = None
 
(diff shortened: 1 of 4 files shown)


tests/conftest.py
DouweM, line 1: Why were these changes needed?

pydantic_ai_slim/pydantic_ai/models/google.py
```diff
@@ -428,12 +428,20 @@ def _build_image_config(self, tool: ImageGenerationTool) -> ImageConfigDict:
         return image_config
 
     def _get_tools(
-        self, model_request_parameters: ModelRequestParameters
+        self,
+        model_request_parameters: ModelRequestParameters,
+        model_settings: GoogleModelSettings | None = None,
```
DouweM, line 433: From the call site it looks like this is guaranteed to not be `None`, so we can simplify this and the check below

pydantic_ai_slim/pydantic_ai/models/google.py
```diff
@@ -428,12 +428,20 @@ def _build_image_config(self, tool: ImageGenerationTool) -> ImageConfigDict:
...
-        tools: list[ToolDict] = [
-            ToolDict(function_declarations=[_function_declaration_from_tool(t)])
-            for t in model_request_parameters.tool_defs.values()
+        parallel = model_settings is not None and model_settings.get('parallel_tool_calls', False) is True
+        function_declarations = [
+            _function_declaration_from_tool(t) for t in model_request_parameters.tool_defs.values()
         ]
+        tools: list[ToolDict] = []
+        if function_declarations:
+            if parallel:
```
DouweM, line 441: - Let's move the `model_settings.get('parallel_tool_calls')` check here
- I think the default should be to enable parallel tool calls, as it is for OpenAI and Anthropic

DouweM (changes requested):

PR #4560, opened 2026-03-06
Remove redundant `running tools` parent span

Kludex:
## Summary

- Remove the `running tools` parent span that wrapped individual tool execution spans in Logfire
- Individual `running tool: <name>` spans are now direct children of the agent run span, removing an unnecessary level of nesting that required an extra click without providing useful aggregation
- Remove the unused `tracer` parameter from `_call_tools`

## Test plan

- [x] `tests/test_logfire.py` - updated snapshots to reflect flattened span hierarchy
- [x] `tests/test_dbos.py` - updated snapshots
- [x] `tests/test_temporal.py` - updated snapshots
- [x] Full test suite passes (excluding pre-existing failures on `main`)

cc @alexmojaki

### DIFF
--- a/tests/test_prefect.py
+++ b/tests/test_prefect.py
@@ -294,22 +294,17 @@ async def test_complex_agent_run_in_flow(allow_model_requests: None, capfire: Ca
                                     ],
                                 ),
                                 BasicSpan(
-                                    content='running 1 tool',
+                                    content='running tool: get_country',
+                                    children=[BasicSpan(content=IsStr(regex=r'Call Tool: get_country-\w+'))],
+                                ),
+                                BasicSpan(
+                                    content=IsStr(regex=r'Handle Stream Event-\w+'),
                                     children=[
+                                        BasicSpan(content='ctx.run_step=1'),
                                         BasicSpan(
-                                            content='running tool: get_country',
-                                            children=[BasicSpan(content=IsStr(regex=r'Call Tool: get_country-\w+'))],
-                                        ),
-                                        BasicSpan(
-                                            content=IsStr(regex=r'Handle Stream Event-\w+'),
-                                            children=[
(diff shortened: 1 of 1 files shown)


adtyavrdhn:
Are prefect tests still being skipped? According to Devin the assertions should have failed

Kludex:
> Are prefect tests still being skipped? According to Devin the assertions should have failed

I don't see this pipeline failing... 👀

adtyavrdhn:
@Kludex Because: https://github.com/pydantic/pydantic-ai/blob/d936d56cc9b20a2e109aa5f4dfccddd4a1a95e92/tests/test_prefect.py#L77 

We were skipping it because test suites were hanging

Kludex:
> @Kludex Because:
> 
> https://github.com/pydantic/pydantic-ai/blob/d936d56cc9b20a2e109aa5f4dfccddd4a1a95e92/tests/test_prefect.py#L77
> 
> We were skipping it because test suites were hanging

I know. I added that. But why are you asking if you know?

adtyavrdhn:
I meant we should still update those snapshots for test_prefect which will be useful for when we do not skip them in the future.

DouweM:
@alexmojaki Do you think we should have a new [instrumentation version](https://ai.pydantic.dev/logfire/#configuring-data-format) for this, as people may be querying for this span?

alexmojaki:
no i don't think this span is useful in queries

DouweM:
@Kludex Looks good to me, feel free to merge if you had no other changes in mind.

Kludex:
Updated the `test_prefect.py` span assertions to remove the `'running N tools'` parent span wrappers, matching the changes already applied to `test_logfire.py`, `test_dbos.py`, and `test_temporal.py`.

PR #4595, opened 2026-03-08
fix: sanitize hallucinated Bedrock tool names to prevent unrecoverable crash

### LINKED ISSUE #4585: Bedrock: hallucinated tool names with invalid characters crash subsequent API calls
## Description

When using `BedrockConverseModel`, if the model hallucinates a tool name containing characters outside `[a-zA-Z0-9_-]` (e.g., a dot or space), PydanticAI records the `ToolCallPart` in the message history. On the next API round-trip, PydanticAI serializes this history back to Bedrock, which rejects the entire request with:

```
ValidationException: 1 validation error detected: Value at 'messages.N.member.content.M.member.toolUse.name'
failed to satisfy constraint: Member must satisfy regular expression pattern: [a-zA-Z0-9_-]+
```

This is unrecoverable — the invalid tool name is baked into the history, so every subsequent request in the same agent run will fail.

## Root cause

In `pydantic_ai/models/bedrock.py`, `_map_tool_call` passes the tool name through without validation:

```python
@staticmethod
def _map_tool_call(t: ToolCallPart) -> ContentBlockOutputTypeDef:
    return {
        'toolUse': {'toolUseId': _utils.guard_tool_call_id(t=t), 'name': t.tool_name, 'input': t.args_as_dict()}
    }
```

## Expected behavior

PydanticAI should sanitize or reject tool names that don't match Bedrock's `[a-zA-Z0-9_-]+` constraint before sending them to the API. This could be done either:

1. When recording the `ToolCallPart` from the model's response (replace invalid chars)
2. When serializing history in `_map_tool_call` (sanitize on output)
3. Both

## Reproduction

1. Create an agent with `BedrockConverseModel` and several tools
2. Use a system prompt that encourages the model to call tools with dotted names (e.g., "use module.function_name syntax")
3. When the model hallucinates a tool name like `search.evidence`, the first retry works, but the second API call crashes because the hallucinated name is in the history

## Workaround

We added a `history_processor` that sanitizes tool names before each API call:

```python
import re

_BEDROCK_TOOL_NAME_RE = re.compile(r"[^a-zA-Z0-9_-]")

def sanitize_tool_names(messages):
    result = []
    for msg in messages:
            new_parts = []
            modified = False
            for part in msg.parts:
                if isinstance(part, ToolCallPart) and _BEDROCK_TOOL_NAME_RE.search(part.tool_name):
                    new_parts.append(replace(part, tool_name=_BEDROCK_TOOL_NAME_RE.sub("_", part.tool_name)))
                    modified = True
                else:
                    new_parts.append(part)
            result.append(replace(msg, parts=new_parts) if modified else msg)
        else:
            result.append(msg)
    return result
```

## Environment

- pydantic-ai-slim 1.63.0
- Bedrock model: `us.anthropic.claude-sonnet-4-5-20250929-v1:0`
- Python 3.12

MaxwellCalkin:
## Summary

Fixes #4585.

When a Bedrock model hallucinates a tool name containing characters outside `[a-zA-Z0-9_-]` (e.g., `search.evidence`), the invalid name gets recorded in message history. On the next API call, Bedrock rejects the entire request with a `ValidationException`, and since the invalid name is baked into history, **every subsequent request fails** — the agent run is unrecoverable.

## Fix

Sanitize tool names in `_map_tool_call()` before sending them back to Bedrock, replacing invalid characters with `_`:

```python
_BEDROCK_TOOL_NAME_RE: typing.ClassVar[re.Pattern[str]] = re.compile(r'[^a-zA-Z0-9_-]')

@staticmethod
def _map_tool_call(t: ToolCallPart) -> ContentBlockOutputTypeDef:
    tool_name = BedrockConverseModel._BEDROCK_TOOL_NAME_RE.sub('_', t.tool_name)
    return {
        'toolUse': {'toolUseId': _utils.guard_tool_call_id(t=t), 'name': tool_name, 'input': t.args_as_dict()}
    }
```

**`pydantic_ai_slim/pydantic_ai/models/bedrock.py`** — adds `import re` and a compiled regex class variable for the constraint pattern.

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/bedrock.py
+++ b/pydantic_ai_slim/pydantic_ai/models/bedrock.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import functools
+import re
 import typing
 from collections.abc import AsyncIterator, Iterable, Iterator, Mapping, Sequence
 from contextlib import asynccontextmanager
@@ -951,10 +952,15 @@ class BedrockConverseModel(Model):
                     assert_never(item)
         return [{'role': 'user', 'content': content}]
 
+    _BEDROCK_TOOL_NAME_RE: typing.ClassVar[re.Pattern[str]] = re.compile(r'[^a-zA-Z0-9_-]')
+
     @staticmethod
     def _map_tool_call(t: ToolCallPart) -> ContentBlockOutputTypeDef:
+        # Sanitize tool name to match Bedrock's [a-zA-Z0-9_-]+ constraint.
+        # Models can hallucinate names with dots/spaces that would crash subsequent API calls.
+        tool_name = BedrockConverseModel._BEDROCK_TOOL_NAME_RE.sub('_', t.tool_name)
         return {
-            'toolUse': {'toolUseId': _utils.guard_tool_call_id(t=t), 'name': t.tool_name, 'input': t.args_as_dict()}
+            'toolUse': {'toolUseId': _utils.guard_tool_call_id(t=t), 'name': tool_name, 'input': t.args_as_dict()}
         }
 
     @staticmethod


adtyavrdhn:
There is a PR for this already #4586 

Please contribute your thoughts there

PR #4604, opened 2026-03-08
feat(openrouter): add `CachePoint` and prompt caching support

### LINKED ISSUE #4392: Support CachePoint for OpenRouter models (Anthropic/Gemini prompt caching)
### Description

OpenRouter supports explicit prompt caching via `cache_control` breakpoints for Anthropic and Gemini models ([docs](https://openrouter.ai/docs/guides/best-practices/prompt-caching)). However, `OpenRouterChatModel` currently inherits the OpenAI base behavior which silently drops `CachePoint` items:

```python
# In OpenAIChatModel._map_content_item:
elif isinstance(item, CachePoint):
    # OpenAI doesn't support prompt caching via CachePoint, so we filter it out
    return None
```

Since `OpenRouterChatModel` doesn't override `_map_content_item`, any `CachePoint` passed in messages is silently ignored — even when the underlying model (e.g. Anthropic via OpenRouter) supports it.

### Expected behavior

`OpenRouterChatModel` should translate `CachePoint` into the `cache_control` format that OpenRouter expects, e.g.:

```json
{"type": "text", "text": "...", "cache_control": {"type": "ephemeral"}}
```

This would allow users to benefit from cached token pricing (0.25x input cost) when using Anthropic or Gemini models through OpenRouter.

### References

- OpenRouter prompt caching docs: https://openrouter.ai/docs/guides/best-practices/prompt-caching
- `pydantic_ai/models/openai.py` — `_map_content_item` filters out `CachePoint`

---

I'm willing to submit a PR for this myself if the proposal is accepted.

Adversarian:
- Closes #4392

## Summary

Adds prompt caching support for OpenRouter models targeting Anthropic and Gemini downstream providers:

- **`CachePoint` in user messages** — translates to `cache_control` breakpoints on preceding content parts
- **`openrouter_cache_instructions`** — caches system prompt (works for both Anthropic and Gemini)
- **`openrouter_cache_messages`** — caches the last message in the conversation (Anthropic only; Gemini only reliably caches system-level content)
- **`openrouter_cache_tool_definitions`** — caches tool definitions (Anthropic only)

TTL (`5m`/`1h`) is passed through for Anthropic and omitted for Gemini. Unsupported downstream providers silently skip cache control.

### Key design decisions

- Overrides `_map_user_prompt` and `_map_messages` rather than `_map_content_item` to handle CachePoint attachment to the *preceding* content part
- Cache capability detection via `OpenRouterModelProfile` flags (`openrouter_supports_cache_control`, `openrouter_supports_cache_ttl`, `openrouter_supports_tool_cache`) set in `OpenRouterProvider.model_profile()`, following the same pattern as `BedrockModelProfile`
- Mirrors the Anthropic provider's caching API patterns (`cache_instructions`, `cache_messages`, `cache_tool_definitions`) with `openrouter_` prefix
- Adds `model_settings` keyword-only parameter to base `OpenAIChatModel._get_tools` and `_map_messages` to thread settings through to the OpenRouter override (backward compatible — parameter defaults to `None`)

### Review feedback addressed

- Switched from model name prefix string matching to `OpenRouterModelProfile` capability flags (per @adtyavrdhn)
- Added `all_messages()` snapshots to all E2E tests (per @adtyavrdhn)
- Corrected "last user message" to "last message" in docs and docstrings (per @adtyavrdhn)
- Consolidated doc examples into fewer sections (per bot review)
- Merged upstream `main` to resolve conflicts

## Test plan

- [x] 31 cache-specific tests covering all caching mechanisms: unit tests for message mapping, VCR cassette E2E tests with `all_messages()` snapshots, real API cache metric verification (write+read) for both Anthropic and Gemini, streaming, and combined all-settings scenarios
- [x] All 72 OpenRouter tests pass (including upstream's nested provider response, service tier, and error handling tests)
- [x] Lint, typecheck, and all pre-commit hooks pass
- [x] Docs build and doc example tests pass

### Checklist


Adversarian:
@DouweM @ajac-zero I believe this is ready for review now whenever you guys can take a look.

tests/models/test_openrouter.py
```diff
@@ -947,3 +951,971 @@ async def test_openrouter_prepare_request_loop_with_non_websearch_first(openrout
...
+    system_msg = next(m for m in mapped if m.get('role') in ('system', 'developer'))
+    system_content = system_msg.get('content')
+    assert isinstance(system_content, list)
+    assert 'cache_control' in system_content[0]
+
+    # Last message should have cache_control
+    last_msg = mapped[-1]
+    last_content = last_msg.get('content')
+    assert isinstance(last_content, list)
+    assert 'cache_control' in last_content[-1]
```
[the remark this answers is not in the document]
Adversarian: This follows the established testing pattern used across all model test files in the codebase — _map_messages() and _get_tools() are called directly in test_openai.py, test_google.py, test_bedrock.py, test_anthropic.py, test_mistral.py, and test_gemini.py (119 occurrences across 7 files). The only mock-free pattern used in tests/models/ is VCR cassettes (per rule:318), and recording 15+ separate cassettes for edge cases like custom TTL values, Gemini vs Anthropic provider differences, unsupported providers, empty content, and image content would be expensive to record and fragile to maintain. The 9 cassette-backed E2E tests already validate the full public API path; the unit tests complement them by covering edge cases cheaply.

I think it's fine to keep these as is but I'm open to suggestions. @DouweM

Adversarian:
@DouweM Sorry to ping you again, just wanted to make sure this doesn't get lost in the pile of PRs.

twobarrelsapps:
Any updates on this? My app would benefit from this.

tests/models/test_openrouter.py
```diff
@@ -947,3 +951,971 @@ async def test_openrouter_prepare_request_loop_with_non_websearch_first(openrout
...
+
+    result = await agent.run(
+        ['Here is some important context to cache.' * 20, CachePoint(), 'Summarize the context in one sentence.']
+    )
+
+    assert isinstance(result.output, str)
+    assert len(result.output) > 0
+
+    # Verify cache_control was in the request
+    assert vcr is not None
```
adtyavrdhn: I think we should do the same all_messages() snapshot, any reason not to?

tests/models/test_openrouter.py
```diff
@@ -947,3 +951,971 @@ async def test_openrouter_prepare_request_loop_with_non_websearch_first(openrout
...
+
+    result = await agent.run(
+        ['Here is some important context to cache.' * 20, CachePoint(), 'Summarize the context in one sentence.']
+    )
+
+    assert isinstance(result.output, str)
+    assert len(result.output) > 0
+
+    # Verify cache_control was in the request (without TTL for Gemini)
+    assert vcr is not None
```
adtyavrdhn: Same

pydantic_ai_slim/pydantic_ai/models/openrouter.py
```diff
@@ -563,6 +620,65 @@ def __init__(
...
+        """
+        return self.model_name.split('/')[0]
+
+    @property
+    def _supports_cache_control(self) -> bool:
+        """Whether the downstream provider supports explicit ``cache_control`` breakpoints via OpenRouter."""
+        return self._downstream_provider in ('anthropic', 'google')
+
+    @property
+    def _supports_cache_ttl(self) -> bool:
```
adtyavrdhn: We have model capabilities as flags on the `ModelProfile`. This would be inconsistent with other Models?

adtyavrdhn (changes requested):

docs/models/openrouter.md
```diff
@@ -74,6 +74,134 @@ agent = Agent(model, model_settings=settings)
 ...
 ```
 
+## Prompt Caching
+
+OpenRouter supports [prompt caching](https://openrouter.ai/docs/guides/best-practices/prompt-caching) for downstream providers that implement it (currently Anthropic and Gemini). Pydantic AI provides four ways to use prompt caching through OpenRouter:
+
+1. **Cache System Instructions**: Set [`OpenRouterModelSettings.openrouter_cache_instructions`][pydantic_ai.models.openrouter.OpenRouterModelSettings.openrouter_cache_instructions] to `True` or specify `'5m'` / `'1h'` directly
+2. **Cache User Messages**: Set [`OpenRouterModelSettings.openrouter_cache_messages`][pydantic_ai.models.openrouter.OpenRouterModelSettings.openrouter_cache_messages] to `True` to automatically cache the last user message
```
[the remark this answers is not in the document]
adtyavrdhn: Should it be last user message or last just last message?

tests/models/test_openrouter.py
Adversarian: Done.

tests/models/test_openrouter.py
Adversarian: Done.

pydantic_ai_slim/pydantic_ai/models/openrouter.py
Adversarian: Agreed, I wasn't paying attention to `ModelProfile`s at the time. Refactored this to use `OpenRouterModelProfile(OpenAIModelProfile)` with `openrouter_supports_cache_control`, `openrouter_supports_cache_ttl`, and `openrouter_supports_tool_cache` flags, set
   in `OpenRouterProvider.model_profile()` based on the downstream provider. This follows the same pattern used in `BedrockModelProfile` which I used a reference.

pydantic_ai_slim/pydantic_ai/models/openrouter.py
```diff
@@ -563,6 +620,65 @@ def __init__(
...
+
+        OpenRouter model names follow the format ``provider/model`` (e.g.
+        ``anthropic/claude-sonnet-4.6``, ``google/gemini-2.5-pro``).
+        """
+        return self.model_name.split('/')[0]
+
+    @property
+    def _supports_cache_control(self) -> bool:
+        """Whether the downstream provider supports explicit ``cache_control`` breakpoints via OpenRouter."""
+        return self._downstream_provider in ('anthropic', 'google')
```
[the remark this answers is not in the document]
Adversarian: Agreed and addressed.

docs/models/openrouter.md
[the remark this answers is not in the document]
Adversarian: "Last message" is correct. The implementation caches `openai.messages[-1]` regardless of role, matching OpenRouter's own docs ("applies cache breakpoints to the last cacheable block in the request") and mimicking the implicit caching pattern from Anthropic models already existing in pydantic-ai. Doc was updated to reflect this.

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -922,7 +922,9 @@ def _get_stream_options(self, model_settings: OpenAIChatModelSettings) -> chat.C
             options['continuous_usage_stats'] = True
         return cast(chat.ChatCompletionStreamOptionsParam, options)
 
-    def _get_tools(self, model_request_parameters: ModelRequestParameters) -> list[chat.ChatCompletionToolParam]:
+    def _get_tools(
+        self, model_request_parameters: ModelRequestParameters, *, model_settings: ModelSettings | None = None
+    ) -> list[chat.ChatCompletionToolParam]:
```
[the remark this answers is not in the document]
Adversarian: This was actually a point of contention during my coding session with Claude and I came to the conclusion that this is the cleanest way to handle it, funny to see the auto-reviewer flagging it as well (great minds and all that). I'm open to suggestions here and I can alter it as needed.

docs/models/openrouter.md
```diff
@@ -74,6 +74,134 @@ agent = Agent(model, model_settings=settings)
...
+model = OpenRouterModel('anthropic/claude-sonnet-4-6')
+agent = Agent(
+    model,
+    instructions='You are a specialized assistant with deep domain knowledge...',
+    model_settings=OpenRouterModelSettings(
+        openrouter_cache_instructions=True,  # Cache system instructions
+    ),
+)
+...
+```
```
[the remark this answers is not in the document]
Adversarian: Done. Merged examples 1, 2 and 4 into a single "Caching via Model Settings" section while keeping CachePoint and Provider Routing as separate sections.

pydantic_ai_slim/pydantic_ai/models/openrouter.py
```diff
@@ -584,6 +700,75 @@ def prepare_request(
...
+            ttl: Literal['5m', '1h'] = '5m' if cache_instructions is True else cache_instructions
+            # Find the last system/developer message and add cache_control to its content
+            for msg in reversed(openai_messages):
+                if msg.get('role') in ('system', 'developer'):
+                    content = msg.get('content')
+                    if isinstance(content, str):
+                        msg['content'] = [  # type: ignore[typeddict-unknown-key]
+                            {'type': 'text', 'text': content, 'cache_control': self._build_cache_control(ttl)}
+                        ]
+                    elif isinstance(content, list) and content:  # pragma: no branch
```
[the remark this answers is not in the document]
Adversarian: I added a comment explaining why the implicit else is unreachable: system messages always have `str` or `list` content from `_map_messages()`, so `None` or empty content can't occur. Kept the pragma since the [Anthropic model also uses it for the same kind of unreachable branch.](https://github.com/pydantic/pydantic-ai/blob/main/pydantic_ai_slim/pydantic_ai/models/anthropic.py#L981)

tests/models/test_openrouter.py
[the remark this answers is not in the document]
Adversarian: Addressed.

Adversarian:
@adtyavrdhn Sorry for the late response. I took a couple of days off for the new year's eve. I tried to address all of the comments. Let me know if there's anything left.

dergachoff:
Broken cache on openrouter for anthropic models is a MAJOR pain. Bumping this PR for visibility 🙏

adtyavrdhn:
@dergachoff Slipped off my radar, I'll take another look. Thanks!

@Adversarian Sorry about that, would you please resolve the conflicts and we could take another look?

Adversarian:
@adtyavrdhn Absolutely. I had Cursor reconcile this fork with the upstream by respecting the upstream at every step of the rebase and then reviewed the changes myself. If we need to re-run Devin (or Cubic, the new guy) on this I'll be waiting to resolve the potential suggestions that might pop up. I've also left comments on the prior threads. Feel free to push back or resolve them as you see fit.

adtyavrdhn:
Thanks @Adversarian 

Only thing I am concerned about is the number of `CachePoints` enforcement which is present in Anthropic but not here. I am not sure how OpenRouter behaves when that contract is violated, could you please check that and also add a test for it?

Does `Gemini` restrict these cache points in the same way?

Adversarian:
> Thanks @Adversarian
> 
> Only thing I am concerned about is the number of `CachePoints` enforcement which is present in Anthropic but not here. I am not sure how OpenRouter behaves when that contract is violated, could you please check that and also add a test for it?
> 
> Does `Gemini` restrict these cache points in the same way?

@adtyavrdhn 
I just looked into this. OpenRouter doesn't enforce cache breakpoint limits on its end and just passes `cache_control` straight through to the downstream provider. For Anthropic models, this means the downstream Anthropic API enforces its 4-breakpoint limit and returns a 400 through OpenRouter:
 
```
Error code: 400 - {'error': {'message': 'Provider returned error', 'code': 400,
'metadata': {'raw': '{"type":"error","error":{"type":"invalid_request_error",
"message":"A maximum of 4 blocks with cache_control may be provided. Found 6."}}'
```
 
I verified this by sending a request with 6 `cache_control` breakpoints (1 system + 5 user) to `anthropic/claude-sonnet-4.6` via OpenRouter pinned to the Anthropic provider which fails with the above error.
 
Gemini doesn't have this restriction. Same 6-breakpoint request to `google/gemini-2.5-flash` succeeded without error, caching 1352 tokens successfully. This makes sense since Gemini uses context caching rather than per-block `cache_control` breakpoints. Anthropic's docs explicitly state the 4-breakpoint limit ([docs](https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching#when-to-use-multiple-breakpoints)); Gemini doesn't seem to have an equivalent constraint that I could find ([docs](https://ai.google.dev/gemini-api/docs/caching)).
 
Following the same pattern as `AnthropicModel._limit_cache_points` ([here](https://github.com/pydantic/pydantic-ai/blob/main/pydantic_ai_slim/pydantic_ai/models/anthropic.py#L1123)) and `BedrockConverseModel._limit_cache_points` ([here](https://github.com/pydantic/pydantic-ai/blob/main/pydantic_ai_slim/pydantic_ai/models/bedrock.py#L1138)), I added:
 
- `openrouter_max_cache_points` profile flag: set to `4` for Anthropic downstream, `None` for Gemini/others
- `_limit_cache_points()` on `OpenRouterModel`: preserves system/tool cache points, removes excess message breakpoints oldest-first
- 3 tests: limit enforcement for Anthropic (excess trimmed), no limit for Gemini (all preserved), within-budget preservation
 
Happy to adjust the approach or add an E2E cassette for this if you'd prefer.  I wasn't sure how you'd prefer this tested so I just added what made sense to me given your response.

adtyavrdhn:
@Adversarian I'd prefer to just follow the precedence set by `Anthropic` so if we can do the same thing and record a cassette that was recorded for `Anthropic` but via `OpenRouter` that would be great.

docs/models/openrouter.md
```diff
@@ -74,6 +74,92 @@ agent = Agent(model, model_settings=settings)
...
+2. **Cache User Messages**: Set [`OpenRouterModelSettings.openrouter_cache_messages`][pydantic_ai.models.openrouter.OpenRouterModelSettings.openrouter_cache_messages] to `True` to automatically cache the last message in the conversation
+3. **Cache Tool Definitions**: Set [`OpenRouterModelSettings.openrouter_cache_tool_definitions`][pydantic_ai.models.openrouter.OpenRouterModelSettings.openrouter_cache_tool_definitions] to `True` or specify `'5m'` / `'1h'` directly
+4. **Fine-Grained Control with [`CachePoint`][pydantic_ai.messages.CachePoint]**: Insert a `CachePoint` marker in user messages to cache everything before it
+
+!!! note "Provider Differences"
+    - **Anthropic** models support prefix-based caching for both system instructions and user messages. TTL values (`'5m'`, `'1h'`) are passed through to the provider.
+    - **Gemini** models only reliably cache system-level content. For Gemini, prefer `openrouter_cache_instructions` over `openrouter_cache_messages`. TTL values are ignored by Gemini.
```
adtyavrdhn, line 88: When we say prefer, what is the difference going to be if we just end up using `openrouter_cache_messages` for `Gemini`?

docs/models/openrouter.md
```diff
@@ -74,6 +74,92 @@ agent = Agent(model, model_settings=settings)
...
+
+
+@agent.tool
+def search_docs(ctx: RunContext, query: str) -> str:
+    """Search documentation."""
+    return f'Results for {query}'
+...
+```
+
+Each setting accepts `True` (defaults to `'5m'` TTL) or an explicit `'5m'` / `'1h'` TTL value. After running, check `result.usage().cache_write_tokens` and `result.usage().cache_read_tokens` to see caching in action. On subsequent calls with `message_history=result.all_messages()`, cached tokens are reused.
```
adtyavrdhn, line 118: Are the `5m` defaults matched to the `Anthropic` / `Gemini` defaults? Where are we getting this default from?

pydantic_ai_slim/pydantic_ai/models/openai.py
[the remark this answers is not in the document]
adtyavrdhn: I am fine with this

pydantic_ai_slim/pydantic_ai/models/openrouter.py
```diff
@@ -577,6 +629,11 @@ def _openrouter_settings_to_openai_settings(
     if usage := model_settings.pop('openrouter_usage', None):
         extra_body['usage'] = usage
 
+    # Note: openrouter_cache_instructions, openrouter_cache_messages, and
```
adtyavrdhn, line 632: Yeah that makes sense although having these no op fields is kind of a smell?

pydantic_ai_slim/pydantic_ai/models/openrouter.py
```diff
@@ -608,6 +665,102 @@ def __init__(
...
+        used = int(has_tool_cache_point)
+
+        for msg in openai_messages:
+            if msg.get('role') in ('system', 'developer'):
+                content = msg.get('content')
+                if isinstance(content, list):
+                    used += sum(1 for part in content if 'cache_control' in cast(dict[str, Any], part))
+
+        remaining = max_points - used
+        if remaining < 0:  # pragma: no cover
```
adtyavrdhn, line 719: Could we cover this please?

pydantic_ai_slim/pydantic_ai/models/openrouter.py
```diff
@@ -608,6 +665,102 @@ def __init__(
...
+            content = msg.get('content')
+            if not isinstance(content, list):
+                continue
+            for part in reversed(content):
+                part_dict = cast(dict[str, Any], part)
+                if 'cache_control' in part_dict:
+                    if remaining > 0:
+                        remaining -= 1
+                    else:
+                        del part_dict['cache_control']
```
adtyavrdhn, line 738: The warning should warn about the excessive cache points removal, is that not the case in `Anthropic`?

pydantic_ai_slim/pydantic_ai/models/openrouter.py
```diff
@@ -608,6 +665,102 @@ def __init__(
         """
         super().__init__(model_name, provider=provider or OpenRouterProvider(), profile=profile, settings=settings)
 
+    @property
+    def _cache_profile(self) -> OpenRouterModelProfile:
+        """Get the OpenRouter-specific model profile with cache capability flags."""
+        return OpenRouterModelProfile.from_profile(self.profile)
+
+    def _build_cache_control(self, ttl: Literal['5m', '1h'] = '5m') -> dict[str, str]:
```
adtyavrdhn, line 673: How about changing this signature to allow to pass `True` and then we don't need to repeat the `ttl: Literal['5m', '1h'] = '5m' if cache_tool_defs is True else cache_tool_defs`? 

Not a fan of repeating the ttl def everywhere we need to use it. If we ever need to change it we would need to change it everywhere

pydantic_ai_slim/pydantic_ai/models/openrouter.py
```diff
@@ -644,6 +797,81 @@ def _translate_thinking(
...
+    ) -> list[chat.ChatCompletionMessageParam]:
+        openai_messages = await super()._map_messages(messages, model_request_parameters, model_settings=model_settings)
+
+        if (
+            openai_messages
+            and model_settings
+            and (cache_messages := model_settings.get('openrouter_cache_messages'))
+            and self._cache_profile.openrouter_supports_cache_control
+        ):
+            ttl: Literal['5m', '1h'] = '5m' if cache_messages is True else cache_messages
```
adtyavrdhn, line 837: Yeah this duplication should be improved

pydantic_ai_slim/pydantic_ai/models/openrouter.py
```diff
@@ -724,6 +952,22 @@ def _map_response_thinking_part(self, item: ThinkingPart) -> None:
...
+        """Map a user prompt, translating CachePoint markers into ``cache_control`` for supported providers."""
+        if isinstance(part.content, str):
+            return chat.ChatCompletionUserMessageParam(role='user', content=part.content)
+
+        content: list[ChatCompletionContentPartParam] = []
+        for item in part.content:
+            if isinstance(item, CachePoint):
+                self._add_cache_control(content, ttl=item.ttl)
+            else:
+                mapped_item = await self._map_content_item(item)
```
adtyavrdhn, line 966: I am not very familiar with open router, what is the scenario where `mapped_item` would be `None` and why can't we reach it?

tests/models/test_openrouter.py
```diff
@@ -996,6 +1000,1199 @@ async def test_openrouter_prepare_request_loop_with_non_websearch_first(openrout
     assert extra_body['web_search_options'] == {'search_context_size': 'medium'}
 
 
+# ===== CachePoint in user messages =====
+
+
+async def test_openrouter_cache_point_adds_cache_control() -> None:
```
adtyavrdhn, line 1006: We don't need this test if we have the second one? 

Can we keep a simpler surface area for tests make sure we don't have two tests where one is a subset of another?

tests/models/test_openrouter.py
```diff
@@ -996,6 +1000,1199 @@ async def test_openrouter_prepare_request_loop_with_non_websearch_first(openrout
...
+    assert content == snapshot(
+        [
+            {'type': 'text', 'text': 'First chunk', 'cache_control': {'type': 'ephemeral', 'ttl': '5m'}},
+            {'type': 'text', 'text': 'Second chunk', 'cache_control': {'type': 'ephemeral', 'ttl': '5m'}},
+            {'type': 'text', 'text': 'Question'},
+        ]
+    )
+
+
+async def test_openrouter_cache_point_with_custom_ttl() -> None:
```
adtyavrdhn, line 1051: Could probably just have one test with multiple markers one of which can have a custom `ttl` unless that is a stupid and doesn't work that way?

tests/models/test_openrouter.py
```diff
@@ -996,6 +1000,1199 @@ async def test_openrouter_prepare_request_loop_with_non_websearch_first(openrout
...
+async def test_openrouter_cache_point_gemini_omits_ttl() -> None:
+    """Test that CachePoint omits TTL for Gemini models."""
+    model = OpenRouterModel('google/gemini-2.5-flash', provider=OpenRouterProvider(api_key='test-key'))
+
+    messages: list[ModelMessage] = [ModelRequest(parts=[UserPromptPart(content=['Context', CachePoint(), 'Question'])])]
+
+    mapped = await model._map_messages(messages, ModelRequestParameters())  # pyright: ignore[reportPrivateUsage]
+    content = mapped[0].get('content')
+    assert isinstance(content, list)
+    assert content[0] == snapshot({'type': 'text', 'text': 'Context', 'cache_control': {'type': 'ephemeral'}})
```
adtyavrdhn, line 1076: Can we assert on the list of messages? The snapshots are more readable and obvious that way.

adtyavrdhn (changes requested):
Thanks @Adversarian 

I took another pass at it :)

I would really appreciate some more thoughts put into the tests(trying to keep the duplicate testing scenarios low), overall structure looks good to me.

Adversarian:
> Thanks @Adversarian
> 
> I took another pass at it :)
> 
> I would really appreciate some more thoughts put into the tests(trying to keep the duplicate testing scenarios low), overall structure looks good to me.

@adtyavrdhn Absolutely. To be perfectly honest, I just described what I wanted and pretty much vibe coded the tests away thinking that they will almost certainly cover everything but I didn't think about redundancy. I'll try to manually check them this time. I'm a bit swamped at the moment so it will take a bit but I appreciate you being patient with this so far and I apologize for not paying enough attention earlier. That was irresponsible of me.

docs/models/openrouter.md
Adversarian: Per OpenRouter's docs ([Gemini caching section](https://openrouter.ai/docs/guides/best-practices/prompt-caching#google-gemini)):
 
> "There is not a limit on the number of `cache_control` breakpoints you can include in your request. OpenRouter will use only the last breakpoint for Gemini caching across normal message content."
 
So using `openrouter_cache_messages` on Gemini would add a `cache_control` breakpoint to the last message, but OpenRouter only uses the *last* breakpoint across all normal message content for Gemini anyway. Since `cache_control` on the system message (via `openrouter_cache_instructions`) is handled separately through Gemini's `systemInstruction` field, `cache_messages` would effectively just move the single usable breakpoint to the last user message instead of wherever it would naturally be. It's not harmful, but `cache_instructions` is more predictable for Gemini since system content caching has its own dedicated path. I can make the docs more precise about this if you think it's warranted.

docs/models/openrouter.md
Adversarian: That's just Anthropic's default. Their cache TTL is 5 minutes, with 1 hour as a paid upgrade at 2x the base input token price ([Anthropic docs](https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching#pricing), [OpenRouter docs](https://openrouter.ai/docs/guides/best-practices/prompt-caching#cache-ttl-options)). For Gemini the TTL is omitted entirely since Gemini manages its own TTL (typically 3-5 minutes per OpenRouter's docs), so the `5m` default only takes effect for Anthropic models.

pydantic_ai_slim/pydantic_ai/models/openrouter.py
Adversarian: Agreed, it's unconventional. The alternative is popping them in `_openrouter_settings_to_openai_settings` and threading them through separately, but that's more plumbing for no functional benefit as the keys never reach the OpenAI SDK. Happy to refactor if you'd prefer though.

pydantic_ai_slim/pydantic_ai/models/openrouter.py
Adversarian: Will do.

pydantic_ai_slim/pydantic_ai/models/openrouter.py
Adversarian: Anthropic's `_limit_cache_points` silently deletes excess breakpoints without any warning or logging ([code](https://github.com/pydantic/pydantic-ai/blob/2b0aa93a/pydantic_ai_slim/pydantic_ai/models/anthropic.py#L1271-L1276)). Same for Bedrock. I matched that behavior, but happy to add a `warnings.warn()` if you think it's better to be explicit here. What do you prefer?

pydantic_ai_slim/pydantic_ai/models/openrouter.py
Adversarian: Good call. I'll cook something up for this.

pydantic_ai_slim/pydantic_ai/models/openrouter.py
Adversarian: Will do.

pydantic_ai_slim/pydantic_ai/models/openrouter.py
Adversarian: The base `OpenAIChatModel._map_content_item` returns `None` for `CachePoint` items (filters them out since OpenAI doesn't support them). In our `_map_user_prompt` override we handle `CachePoint` before calling `_map_content_item`, so the remaining items are always real content (text, images, etc.) that return non-None. The `if` guard is still needed for type safety (`content.append(None)` wouldn't typecheck) but the branch is unreachable in practice. I'll add a comment explaining this.

tests/models/test_openrouter.py
Adversarian: Agreed, will merge them.

tests/models/test_openrouter.py
Adversarian: We can fold custom TTL into the multi-marker test: one marker default, one with `ttl='1h'`. I think that should work.

tests/models/test_openrouter.py
Adversarian: Will switch to `snapshot()` on the full content list.

pydantic_ai_slim/pydantic_ai/models/openrouter.py
adtyavrdhn: Cool then I don't mind silently deleting it either.

docs/models/openrouter.md
adtyavrdhn: Okay that makes sense, a small comment(I am not sure if I missed that) would be nice there :)

Adversarian:
@adtyavrdhn I think I've addressed everything. Specifically regarding the tests, I tried to look at them through a "necessary and sufficient" lens. I think the current suite is sufficient for coverage. As for necessity, I think it's closer than before but it might still look a bit heavy. Since this is a feature that directly affects downstream costs, I think having air-tight tests for everything is defensible so I'm confident leaving it here unless you (or another maintainer) suggest otherwise.

manthan-dl:
@adtyavrdhn I know you are busy with lot of stuff but can we please look into this on priority. I think this feature is very useful to lot of us. Thanks.

pydantic_ai_slim/pydantic_ai/models/openrouter.py
```diff
@@ -644,6 +815,64 @@ def _translate_thinking(
...
+            self._add_cache_control_to_message(openai_messages[-1], cache_messages)
+
+        if (
+            model_settings
+            and (cache_instructions := model_settings.get('openrouter_cache_instructions'))
+            and self._cache_profile.openrouter_supports_cache_control
+        ):
+            for msg in reversed(openai_messages):
+                if msg.get('role') in ('system', 'developer'):
+                    self._add_cache_control_to_message(msg, cache_instructions)
```
adtyavrdhn: Can you use a dynamic instructions to see if it ends up in the cached block? Anthropic looks for the static prefix and places the cache just before the dynamic instructions entry. This might break the cache

docs/models/openrouter.md
```diff
@@ -74,6 +74,92 @@ agent = Agent(model, model_settings=settings)
...
+
+### Caching via Model Settings
+
+Use [`OpenRouterModelSettings`][pydantic_ai.models.openrouter.OpenRouterModelSettings] to enable caching for system instructions, user messages, and tool definitions:
+
+```python
+from pydantic_ai import Agent, RunContext
+from pydantic_ai.models.openrouter import OpenRouterModel, OpenRouterModelSettings
+
+model = OpenRouterModel('anthropic/claude-sonnet-4-6')
```
adtyavrdhn:
```suggestion
model = OpenRouterModel('anthropic/claude-sonnet-4.6')
```

Please make sure the examples have correct model it

docs/models/openrouter.md
```diff
@@ -74,6 +74,92 @@ agent = Agent(model, model_settings=settings)
...
+    'Now answer my question about the context above',
+]
+...
+```
+
+Pass the prompt list to `agent.run_sync(prompt)`. Everything before the `CachePoint()` marker is cached. You can place multiple markers for fine-grained control over cache boundaries.
+
+### Provider Routing for Cache Locality
+
+OpenRouter load-balances requests across providers by default. Since cache state is provider-specific, consecutive requests may be routed to different providers, resulting in cache misses. Use [`openrouter_provider`][pydantic_ai.models.openrouter.OpenRouterModelSettings.openrouter_provider] to pin requests to a specific downstream provider:
```
adtyavrdhn: https://openrouter.ai/docs/guides/best-practices/prompt-caching#provider-sticky-routing

The docs say it uses sticky routing when a request had prompt caching so we should rephrase this.

docs/models/openrouter.md
```diff
@@ -74,6 +74,92 @@ agent = Agent(model, model_settings=settings)
...
+2. **Cache User Messages**: Set [`OpenRouterModelSettings.openrouter_cache_messages`][pydantic_ai.models.openrouter.OpenRouterModelSettings.openrouter_cache_messages] to `True` to automatically cache the last message in the conversation
+3. **Cache Tool Definitions**: Set [`OpenRouterModelSettings.openrouter_cache_tool_definitions`][pydantic_ai.models.openrouter.OpenRouterModelSettings.openrouter_cache_tool_definitions] to `True` or specify `'5m'` / `'1h'` directly
+4. **Fine-Grained Control with [`CachePoint`][pydantic_ai.messages.CachePoint]**: Insert a `CachePoint` marker in user messages to cache everything before it
+
+!!! note "Provider Differences"
+    - **Anthropic** models support prefix-based caching for both system instructions and user messages. TTL values (`'5m'`, `'1h'`) are passed through to the provider.
+    - **Gemini** models only reliably cache system-level content. For Gemini, prefer `openrouter_cache_instructions` over `openrouter_cache_messages`. TTL values are ignored by Gemini.
+    - **Minimum token thresholds** apply: Anthropic requires ~2048 tokens, Gemini requires ~1024 tokens for content to be eligible for caching.
```
adtyavrdhn: https://openrouter.ai/docs/guides/best-practices/prompt-caching#minimum-token-requirements

Can we just link to the open router docs for this, wouldn't want us to maintain this.

docs/models/openrouter.md
adtyavrdhn: The wording here is tricky, it can work with user content but just the last one? https://openrouter.ai/docs/guides/best-practices/prompt-caching#how-gemini-prompt-caching-works-on-openrouter

Can we make the wording a bit more precise

tests/models/test_openrouter.py
```diff
@@ -996,6 +1001,1134 @@ async def test_openrouter_prepare_request_loop_with_non_websearch_first(openrout
...
+
+async def test_openrouter_cache_messages_anthropic_real_api(
+    allow_model_requests: None, openrouter_api_key: str
+) -> None:
+    """Test that openrouter_cache_messages produces cache read metrics for Anthropic via OpenRouter.
+
+    Forces routing to the Anthropic provider directly (not Bedrock) to ensure cache locality.
+    The first call populates the cache, and the second call reads from it.
+
+    Note: cache_write_tokens is not asserted because OpenRouter reports it via a non-standard
```
adtyavrdhn: If that is the case then in the docs we should not mention we can check `cache_write_tokens`

docs/models/openrouter.md
```diff
@@ -74,6 +74,92 @@ agent = Agent(model, model_settings=settings)
...
+
+
+@agent.tool
+def search_docs(ctx: RunContext, query: str) -> str:
+    """Search documentation."""
+    return f'Results for {query}'
+...
+```
+
+Each setting accepts `True` or an explicit `'5m'` / `'1h'` TTL value. `True` uses Anthropic's default `'5m'` TTL; Gemini ignores TTL values and manages cache lifetime itself. After running, check `result.usage().cache_write_tokens` and `result.usage().cache_read_tokens` to see caching in action. On subsequent calls with `message_history=result.all_messages()`, cached tokens are reused.
```
adtyavrdhn: This would be factually incorrect based on the note in the test, could you verify this please?

adtyavrdhn (changes requested):

docs/models/openrouter.md
Adversarian: Ah, I think this was a recent change and 4-6 actually redirects to 4.6 so that ID was correct at the time I believe but I will make the change as the current nomenclature seems to prefer 4.6.

pydantic_ai_slim/pydantic_ai/models/openrouter.py
Adversarian: You were right, this did break cache reuse when dynamic instructions changed.

I tested the previous behavior live by reproducing the old placement logic, where `cache_control` was added to the last instruction message. With static + dynamic instructions, that meant the dynamic instruction was included in the cached prefix.

The old behavior only reused the cache when the dynamic instruction stayed identical. When the dynamic instruction changed, cache reads dropped to 0.

I updated the implementation to place cache_control on the last static instruction block before dynamic instructions. If there are only dynamic instructions and no stable prefix, it does not add an instruction cache point. I also added a regression test for static + dynamic instructions, dynamic-only instructions, and fallback to an existing static system prompt.

docs/models/openrouter.md
Adversarian: Good catch, updated.

OpenRouter’s docs say prompt-cached requests use provider sticky routing, so the previous wording was too strong. I rephrased this section to mention sticky routing first, and kept the `openrouter_provider` example for cases where users want stricter provider control or want to disable fallbacks for cache-sensitive workflows.

docs/models/openrouter.md
Adversarian: Agreed, updated.

I replaced the hardcoded Anthropic/Gemini token thresholds with a link to OpenRouter’s minimum token requirements section, so we don’t have to maintain provider-specific values in Pydantic AI docs.

docs/models/openrouter.md
Adversarian: I updated it to say that Gemini supports caching for both system instructions and normal message content, but OpenRouter only uses the last breakpoint across normal message content for Gemini caching. The docs now recommend `openrouter_cache_messages` or `CachePoint` when that final message boundary is intentional, and `openrouter_cache_instructions` for stable system context.

tests/models/test_openrouter.py
Adversarian: The old test note was accurate for the previous implementation: OpenRouter reports cache writes through the non-standard `prompt_tokens_details.cache_write_tokens` field, which the generic OpenAI usage mapper did not expose.

This PR now maps that OpenRouter-specific field into `RequestUsage.cache_write_tokens`, and the live OpenRouter check confirmed positive cache write/read metrics. So the docs can now correctly mention checking `result.usage().cache_write_tokens` for cache writes and `cache_read_tokens` for reuse.

docs/models/openrouter.md
Adversarian: Verified and fixed.

The stale test note reflected the previous implementation gap, not current OpenRouter behavior. OpenRouter does return cache write information, but it reports it via `prompt_tokens_details.cache_write_tokens`. This PR now maps that field into `RequestUsage.cache_write_tokens`.

I also checked this against live OpenRouter calls: cache writes and reads are both reported with positive token counts, so the docs now accurately mention checking both `cache_write_tokens` and `cache_read_tokens`.

Adversarian:
@adtyavrdhn 

Pushed the follow-up changes discussed in the open threads.

In addition to the thread-specific fixes, since a fair amount of time has passed since I opened this PR, I wanted to make sure that the underlying logic is still sound everything still works. Here I noticed that openrouter has added some `~provider` aliases for some of the more prominent providers. I made sure these aliases use the same downstream provider cache capability detection as normal provider IDs, so `~anthropic/...` and `~google/...` behave correctly for cache control. The cassette tests still use concrete model IDs for stable recordings.

pydantic_ai_slim/pydantic_ai/models/openrouter.py
```diff
@@ -608,6 +688,158 @@ def __init__(
...
+    ) -> None:
+        """Add ``cache_control`` to the last content block in a mapped chat message."""
+        content = message.get('content')
+        if isinstance(content, str):
+            message['content'] = [  # type: ignore[typeddict-unknown-key]
+                {'type': 'text', 'text': content, 'cache_control': self._build_cache_control(ttl)}
+            ]
+        elif isinstance(content, list) and content:
+            last_part = cast(dict[str, Any], content[-1])
+            last_part['cache_control'] = self._build_cache_control(ttl)
```
adtyavrdhn: Could we avoid overwriting an existing `cache_control` here? If the final block already got one from `CachePoint(ttl='1h')`, `openrouter_cache_messages=True` currently replaces it with the default 5m TTL. I think the explicit `CachePoint` should take priority. Please add a small test for that combination too.

pydantic_ai_slim/pydantic_ai/models/openrouter.py
```diff
@@ -608,6 +688,158 @@ def __init__(
...
+        if instruction_role not in ('system', 'developer'):
+            return
+
+        instruction_prefix_count = next(
+            (index for index, msg in enumerate(openai_messages) if msg.get('role') != instruction_role),
+            len(openai_messages),
+        )
+        first_instruction_index = instruction_prefix_count - len(instruction_parts)
+
+        if any(part.dynamic for part in instruction_parts):
```
adtyavrdhn: Codex raised one Gemini concern here: OpenRouter normalizes system/developer messages into a single immutable `systemInstruction`, so the Anthropic static/dynamic split may not actually leave the dynamic tail uncached here. I am not sure what the ideal behaviour should be in this case.

What is your opinion? 

I'll also flag it out for @DouweM

adtyavrdhn (commented):
Two small follow-up questions from review.

adtyavrdhn:
Apologies for the back and forth, the PR is huge and so many behaviors to account for so it is easy for me to miss things in the first go around.

Adversarian:
> Apologies for the back and forth, the PR is huge and so many behaviors to account for so it is easy for me to miss things in the first go around.

Not at all, I completely understand. The scope is big enough for me to lose sight of some things as well (despite my best efforts to the contrary!) and I'm grateful that you're taking the time to be meticulous and thorough. I expected nothing less.

pydantic_ai_slim/pydantic_ai/models/openrouter.py
Adversarian: Good catch. I agree explicit `CachePoint` should take priority over the convenience `openrouter_cache_messages` setting.

I updated `_add_cache_control_to_message()` so it preserves an existing `cache_control` on the final block instead of overwriting it, and added a focused test for `CachePoint(ttl='1h')` combined with `openrouter_cache_messages=True` as requested.

pydantic_ai_slim/pydantic_ai/models/openrouter.py
Adversarian: I looked this up in both OpenRouter and Gemini docs, and I think the concern is valid. OpenRouter normalizes Gemini system/developer messages into Gemini’s `systemInstruction`, and Gemini cached content treats `systemInstruction` as immutable, so Anthropic-style static/dynamic instruction splitting is not safe to assume here.

I updated the implementation so Gemini/OpenRouter skips `openrouter_cache_instructions` when dynamic instructions are present. Static-only Gemini instruction caching still works. I also updated the docs to say `openrouter_cache_instructions` should only be used for fully static Gemini system context, and dynamic prompt segments should go into later user/message content.

docs/models/openrouter.md
```diff
@@ -74,6 +74,94 @@ agent = Agent(model, model_settings=settings)
...
+    model_settings=OpenRouterModelSettings(
+        openrouter_cache_instructions=True,
+        openrouter_provider={
+            'order': ['anthropic'],
+            'allow_fallbacks': False,
+        },
+    ),
+)
+...
+```
```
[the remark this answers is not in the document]
Adversarian: Agreed. This was really a parameter variation, not a distinct workflow.

I folded the provider-routing note into the “Caching via Model Settings” section and removed the separate heading/full code block. The docs now just mention `openrouter_provider={'order': ['anthropic'], 'allow_fallbacks': False}` inline for cache-sensitive workflows.

pydantic_ai_slim/pydantic_ai/models/openrouter.py
```diff
@@ -608,6 +688,158 @@ def __init__(
         """
         super().__init__(model_name, provider=provider or OpenRouterProvider(), profile=profile, settings=settings)
 
+    @property
+    def _cache_profile(self) -> OpenRouterModelProfile:
+        """Get the OpenRouter-specific model profile with cache capability flags."""
+        return OpenRouterModelProfile.from_profile(self.profile)
```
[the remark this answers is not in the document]
Adversarian: Makes sense. I changed `_cache_profile` to use `@cached_property`, matching the pattern used by the parent model profile access.

pydantic_ai_slim/pydantic_ai/models/openrouter.py
```diff
@@ -608,6 +688,158 @@ def __init__(
...
+            if static_instruction_count:
+                cache_message_index = first_instruction_index + static_instruction_count - 1
+            elif first_instruction_index > 0:
+                cache_message_index = first_instruction_index - 1
+            else:
+                return
+        else:
+            cache_message_index = instruction_prefix_count - 1
+
+        self._add_cache_control_to_message(openai_messages[cache_message_index], ttl)
```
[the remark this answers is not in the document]
Adversarian: Added a short inline comment explaining that instruction parts are mapped as the tail of the system/developer prefix, which is why `instruction_prefix_count - len(instruction_parts)` identifies the first mapped instruction message.

pydantic_ai_slim/pydantic_ai/models/openrouter.py
```diff
@@ -724,6 +1018,25 @@ def _map_response_thinking_part(self, item: ThinkingPart) -> None:
...
+            if isinstance(item, CachePoint):
+                self._add_cache_control(content, ttl=item.ttl)
+            else:
+                mapped_item = await self._map_content_item(item)
+                # The base _map_content_item returns None for CachePoint, but we handle
+                # CachePoint earlier in this loop, so remaining items always produce
+                # non-None. The guard is kept for type safety.
+                if mapped_item is not None:  # pragma: no branch
+                    content.append(mapped_item)
+        return chat.ChatCompletionUserMessageParam(role='user', content=content)
```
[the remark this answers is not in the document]
Adversarian: Yes, I think this override does need to avoid `super()`. The base OpenAI mapper filters out `CachePoint`, but OpenRouter needs to consume it before mapping so we can preserve its TTL and exact position when attaching `cache_control` to the previous block. I added a short comment explaining that the duplication is intentional.

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -1078,7 +1078,9 @@ def _get_stream_options(self, model_settings: OpenAIChatModelSettings) -> chat.C
...
         return [self._map_tool_definition(r) for r in model_request_parameters.tool_defs.values()]
```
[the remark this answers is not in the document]
Adversarian: This was already addressed and replied to, I'm leaving it open for the maintainers to respond if needed.

Adversarian:
@adtyavrdhn I addressed the most recent comments. Let me know if there's anything else left to do.

Adversarian:
@adtyavrdhn Sorry for the second mention, just wanted to make sure this doesn't get lost in the pile.

adtyavrdhn:
Hey @Adversarian 

I am pretty okay with where the PR is now, apart from the Douwe mentions to take calls on certain aspects. He'll be taking it over but do expect a delay :)

adtyavrdhn (approved):
@DouweM Please take a look where your decisions are pending :)

Adversarian:
@DouweM I hate to be a bother but is it possible for you take a look at this when you have the time? I understand it's a big PR but we've iterated over it with @adtyavrdhn multiple times so there shouldn't be any major wrinkles left.

DouweM:
@Adversarian I'll have a look today, appreciate the patience.

pydantic_ai_slim/pydantic_ai/providers/openrouter.py
```diff
@@ -33,6 +33,28 @@
...
 
+@dataclass(kw_only=True)
+class OpenRouterModelProfile(OpenAIModelProfile):
+    """Profile for models used with OpenRouterModel.
+
+    ALL FIELDS MUST BE `openrouter_` PREFIXED SO YOU CAN MERGE THEM WITH OTHER MODELS.
+    """
+
+    openrouter_supports_cache_control: bool = False
+    """Whether the downstream provider supports explicit `cache_control` breakpoints via OpenRouter."""
```
DouweM, line 44: Markdown wants single backticks, not RST-style double :)

pydantic_ai_slim/pydantic_ai/providers/openrouter.py
```diff
@@ -124,21 +146,30 @@ def model_profile(model_name: str) -> ModelProfile | None:
 
         profile = None
 
-        provider, model_name = model_name.split('/', 1)
+        provider, model_name = model_name.removeprefix('~').split('/', 1)
```
DouweM, line 151: What's the `removeprefix` for?

pydantic_ai_slim/pydantic_ai/models/openrouter.py
```diff
@@ -258,6 +271,55 @@ class OpenRouterModelSettings(ModelSettings, total=False):
...
+    use `openrouter_cache_instructions` for stable system context. Anthropic supports
+    prefix-based caching across multi-turn conversations with this setting.
+
+    See https://openrouter.ai/docs/guides/best-practices/prompt-caching for more information.
+    """
+
+    openrouter_cache_tool_definitions: bool | Literal['5m', '1h']
+    """Whether to add `cache_control` to the last tool definition.
+
+    When enabled, the last tool in the `tools` array will have `cache_control` set,
```
DouweM, line 329: Same RST issue as elsewhere. And I assume these docstrings are consistent with those for Anthropic and Bedrock.

We're getting to the point where there should be top-level model settings for this, and a new `Caching` capability (similar to how the `Thinking` capability uses the unified `thinking` setting). Not something for this PR though.

pydantic_ai_slim/pydantic_ai/models/openrouter.py
```diff
@@ -608,6 +692,163 @@ def __init__(
...
+                    break
+            return
+
+        instruction_role = self._cache_profile.openai_system_prompt_role or 'system'
+        if instruction_role not in ('system', 'developer'):
+            return
+
+        has_dynamic_instructions = any(part.dynamic for part in instruction_parts)
+        if has_dynamic_instructions and not self._cache_profile.openrouter_supports_dynamic_instruction_cache:
+            return
```
DouweM, line 844: Why do we do this on OpenRouter + Google? Warrants a comment

pydantic_ai_slim/pydantic_ai/models/openrouter.py
```diff
@@ -608,6 +692,163 @@ def __init__(
...
+
+        instruction_prefix_count = next(
+            (index for index, msg in enumerate(openai_messages) if msg.get('role') != instruction_role),
+            len(openai_messages),
+        )
+        # Instruction parts are mapped as the tail of the system/developer prefix.
+        first_instruction_index = instruction_prefix_count - len(instruction_parts)
+
+        if has_dynamic_instructions:
```
DouweM, line 853: I assume this is the same thing we do for Anthropic?

pydantic_ai_slim/pydantic_ai/models/openrouter.py
```diff
@@ -726,6 +1027,26 @@ def _map_response_thinking_part(self, item: ThinkingPart) -> None:
     def _streamed_response_cls(self):
         return OpenRouterStreamedResponse
 
+    @override
+    async def _map_user_prompt(self, part: UserPromptPart) -> chat.ChatCompletionUserMessageParam:
+        """Map a user prompt, translating CachePoint markers into `cache_control` for supported providers."""
+        if isinstance(part.content, str):
+            return chat.ChatCompletionUserMessageParam(role='user', content=part.content)
+
+        # `super()` would drop CachePoint before we can use its TTL or position.
```
DouweM: Is it an option to modify OpenAIChatModel some way that it does not filter out `CachePoint` in this subclassed scenario?

pydantic_ai_slim/pydantic_ai/models/openrouter.py
DouweM: This is not just a request to add a comment -- this dynamic instructions cache is genuinely confusing to me and I want to understand why we're doing this and if we have to.

Adversarian:
Thanks for the review and sorry for the delayed response from me, I'll try to address these soon.

pydantic_ai_slim/pydantic_ai/models/openrouter.py
Adversarian: This came from the earlier Gemini concern that Aditya flagged.

Anthropic and Gemini need different behavior here. Anthropic prompt caching is prefix-based, so we can place `cache_control` after the last static instruction block and leave later dynamic instruction blocks uncached.

Gemini via OpenRouter is different: OpenRouter documents that Gemini system/developer messages are normalized into a single `systemInstruction`, and cached Gemini content treats that `systemInstruction` as immutable. Because of that, we skip `openrouter_cache_instructions` for Gemini when dynamic instructions are present.

So yes, the intent is real and I've added a comment to better explain it.

pydantic_ai_slim/pydantic_ai/models/openrouter.py
Adversarian: You're right. I'll change this to avoid duplicating `_map_user_prompt()` in `OpenRouterModel`.

`OpenAIChatModel` would still filter `CachePoint` by default, but it would go through a small private hook while mapping user prompt content. `OpenRouterModel` will override that hook to translate `CachePoint` into `cache_control`, so the OpenRouter-specific behavior stays in the subclass without copying the whole base method.

pydantic_ai_slim/pydantic_ai/models/openrouter.py
Adversarian: Yes for the Anthropic path, but not for Gemini.

For Anthropic via OpenRouter, this mirrors the Anthropic adapter: static instruction blocks come before dynamic instruction blocks, so the cache breakpoint can be placed after the last static block.

The extra OpenRouter-specific part is that Gemini cannot safely use the same split. OpenRouter maps Gemini system/developer messages into a single `systemInstruction`, so when dynamic instructions are present we skip instruction caching for Gemini instead of caching a prefix that may not actually exclude the dynamic tail (as I explained in the comment about above this).

pydantic_ai_slim/pydantic_ai/providers/openrouter.py
Adversarian: I think Aditya fixed these. My bad there though, thanks for the heads up.

pydantic_ai_slim/pydantic_ai/models/openrouter.py
Adversarian: These seem to be fixed as well by Aditya.

I checked the current OpenRouter docstrings against Anthropic/Bedrock. The OpenRouter wording is not 1:1 identical to Anthropic/Bedrock because it has to call out downstream-provider differences like Gemini omitting TTL and skipping instruction caching when dynamic instructions are present.

I’ll keep the provider-specific details, but I’ll do one final pass before pushing to make sure the phrasing stays aligned where the behavior is equivalent.

pydantic_ai_slim/pydantic_ai/providers/openrouter.py
Adversarian: To handle these new pseudo-providers through OpenRouter: 
https://openrouter.ai/~anthropic/claude-opus-latest

That does look a bit cryptic, I'll add a comment explaining what it's for.

Adversarian:
@adtyavrdhn @dsfaccini 
Douwe seems to be on parental leave (belated congratulations on that btw whenever/if he reads this!) and his status says to ping you guys for Pydantic-AI releated stuff. I believe I've addressed his latest set of comments. If this needs to be greenlit by Douwe himself it's no problem, just thought I should give you guys a heads up.

adtyavrdhn:
@Adversarian I'll take a look at his concerns and do a fresh review myself. Are you using this branch in production already or a fork of it?

Adversarian:
> @Adversarian I'll take a look at his concerns and do a fresh review myself. Are you using this branch in production already or a fork of it?

@adtyavrdhn We originally monkey patched something similar into our existing pydantic-ai installation (this was before the issue attached to this MR was even created) but eventually had to switch to another model for which implicit caching was supported due the complications the monkey patch was causing with dependency updates. So no, we're not using this fork in production right now.

manthan-dl:
@adtyavrdhn @dsfaccini Thanks for all the work on V2 — I know it's a big lift right now. Just wanted to gently resurface this PR. @Adversarian has worked through most of the bugs the team raised, so it's in a solid state. As an enterprise user, getting this merged would be a real cost saver for us. No rush, but it'd be great to get it across the line whenever there's bandwidth.

dsfaccini:
hey @manthan-dl let me take a look

Edit: I'll be working directly on this branch to resolve conflicts and do review rounds. please don't push anything in the meantime.

Thank you guys for working on this!

pydantic_ai_slim/pydantic_ai/models/openrouter.py
dsfaccini: David's AICA here: defending this one — the dynamic-instruction guard is a real correctness issue, not incidental complexity.

OpenRouter collapses all `system`/`developer` messages into a single Gemini `systemInstruction`, and Gemini treats cached `systemInstruction` content as immutable. Unlike Anthropic's prefix-based caching (where we can place `cache_control` after the last static block and leave the dynamic tail uncached), there is no safe static-prefix split inside one immutable `systemInstruction`. Caching it when dynamic instructions are present would either pin stale dynamic content or silently miss the cache when the dynamic part changes.

So for Gemini we skip `openrouter_cache_instructions` when dynamic instructions are present (static-only Gemini instruction caching still works); for Anthropic we keep the prefix split. The behavior is correct and justified, and the guard is documented in the comment at the `has_dynamic_instructions` check. Leaving this thread open for your call on whether you'd still prefer it scoped out.

pydantic_ai_slim/pydantic_ai/models/openai.py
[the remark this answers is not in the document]
dsfaccini: David's AICA here: signing off on this as maintainer. Adding the keyword-only `model_settings` parameter to the base `OpenAIChatModel._get_tools`/`_map_messages` is backward-compatible — it's kwarg-only, defaults to `None`, and both are private methods, so no public surface changes and no existing subclass (`OllamaModel`, `CerebrasModel`) breaks. It's the cleanest way to thread settings into the `OpenRouterModel` override that needs them, and preferable to duplicating `_completions_create`. Approved. Our local review wave + auto-review will still validate it against their own focus areas. Resolving — thanks for flagging it for explicit sign-off (original thread: https://github.com/pydantic/pydantic-ai/pull/4604#discussion_r3179651817).

pydantic_ai_slim/pydantic_ai/providers/openrouter.py
dsfaccini: David's AICA here: fixed — all RST-style double-backticks in `providers/openrouter.py` and `models/openrouter.py` are now single backticks (verified none remain). Resolving.

pydantic_ai_slim/pydantic_ai/providers/openrouter.py
dsfaccini: David's AICA here: added an explanatory comment above the `removeprefix('~')` — it strips OpenRouter's `~provider/model` latest-alias marker (e.g. `~anthropic/claude-opus-latest`) before using the provider prefix for profile selection. Resolving.

pydantic_ai_slim/pydantic_ai/models/openrouter.py
dsfaccini: David's AICA here: backticks fixed and phrasing aligned with Anthropic/Bedrock where the behavior is equivalent (the OpenRouter docstrings additionally call out downstream-provider differences like Gemini omitting TTL and skipping instruction caching under dynamic instructions). Agreed a top-level `Caching` capability (parallel to how `Thinking` uses the unified `thinking` setting) is the right long-term shape — tracking it as out-of-scope follow-up, not this PR. Resolving.

pydantic_ai_slim/pydantic_ai/models/openrouter.py
dsfaccini: David's AICA here: yes for the Anthropic path — it mirrors the Anthropic adapter's static-before-dynamic split, placing the breakpoint after the last static instruction block. Gemini diverges: OpenRouter collapses system/developer into one immutable `systemInstruction`, so when dynamic instructions are present we skip instruction caching rather than assume a safe prefix split (documented in the guard comment). Resolving.

pydantic_ai_slim/pydantic_ai/models/openrouter.py
dsfaccini: David's AICA here: done — extracted a `_map_user_prompt_content_item` hook on the base `OpenAIChatModel`; `OpenRouterModel` overrides only that hook to translate `CachePoint` into `cache_control`, so the OpenRouter-specific behavior stays in the subclass without duplicating `_map_user_prompt`. Resolving.

Adversarian:
Hey @dsfaccini, thanks for picking this up. If there's any major piece of work left to do I can help you divide and conquer. Let me know.

tests/models/test_openrouter.py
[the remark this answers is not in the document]
dsfaccini: David's AICA here: The cache tests have since been redesigned to drive through the public `Agent.run()` API with VCR cassettes — they assert the `cache_control` wire shape on the recorded request body plus the cache write/read usage metrics (see the `Prompt caching: public-API wire-shape tests (cassettes)` section, which replaced the former `_map_messages`/`_get_tool_choice` unit tests). The only two remaining private-method tests pin defensive branches that no real model+provider combination reaches (custom-profile-only and degenerate empty-content), so they can't be exercised through a real request. Resolving.

pydantic_ai_slim/pydantic_ai/models/openrouter.py
```diff
@@ -270,6 +283,55 @@ class OpenRouterModelSettings(ModelSettings, total=False):
     The usage config object consolidates settings for enabling detailed usage information. [See more](https://openrouter.ai/docs/use-cases/usage-accounting)
     """
 
+    openrouter_cache_instructions: bool | Literal['5m', '1h']
```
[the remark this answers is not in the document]
dsfaccini: David's AICA here: Done — extracted the `bool | Literal['5m', '1h']` union into the `OpenRouterCacheTTL` type alias near the top of the file (named to match the file's existing public siblings `OpenRouterProviderName` / `OpenRouterTransforms` rather than `CacheTTLSetting`), and applied it at all 7 sites — the 3 settings fields plus `_build_cache_control`, `_add_cache_control`, `_add_cache_control_to_message`, `_add_cache_control_to_instructions` (commit 51921fa70). Resolving.

tests/models/test_openrouter.py
```diff
@@ -1036,6 +1041,980 @@ async def test_openrouter_prepare_request_loop_with_non_websearch_first(openrout
...
+    """
+    model = OpenRouterModel(
+        'anthropic/claude-sonnet-4.6',
+        provider=OpenRouterProvider(api_key='test-key'),
+        profile=OpenRouterModelProfile(openai_system_prompt_role='user', openrouter_supports_cache_control=True),
+    )
+    settings = OpenRouterModelSettings(openrouter_cache_instructions=True)
+    params = ModelRequestParameters(instruction_parts=[InstructionPart(content='Static instructions.', dynamic=False)])
+
+    mapped = await model._map_messages(  # pyright: ignore[reportPrivateUsage]
```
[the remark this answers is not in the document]
dsfaccini: David's AICA here: Done — extended the section-level comment above these two tests (`Prompt caching: defensive-branch unit tests`) to explicitly call out that the private `_map_messages` access is intentional: both pin defensive branches that no real model+provider combination reaches (custom-profile-only / degenerate empty-content), so they can't be routed through a real request and a future reviewer won't need to re-flag them. Resolving.

pydantic_ai_slim/pydantic_ai/models/openrouter.py
dsfaccini: David's AICA here: verified this against primary docs — yes, it's required for the explicit `cache_control` path, not incidental complexity. OpenRouter normalizes Google `system`/`developer` messages into a single Gemini `systemInstruction` (per their [prompt-caching doc](https://openrouter.ai/docs/guides/best-practices/prompt-caching)), and Google's API marks `systemInstruction` on `CachedContent` as 'Input only. Immutable' (https://ai.google.dev/api/caching). So there's no Anthropic-style static-prefix split available inside that one immutable block — and OpenRouter only honors the last cache breakpoint for Gemini anyway. It's an intentional, documented API contract (no upstream issue/PR tracking a fix), so the guard is necessary rather than something we can drop. The code comment now cites both sources. Leaving this open for you to ack/resolve.

PR #4621, opened 2026-03-10
fix(openai): disable encrypted reasoning content for gateway providers

### LINKED ISSUE #4608: OpenAIResponsesModel: invalid_encrypted_content error when using API gateway/proxy across orgs
### Initial Checks

### Problem

When using `OpenAIResponsesModel` with extended thinking through `gateway_provider` (or any proxy that routes requests across different OpenAI organizations), multi-turn conversations fail with a 400 `invalid_encrypted_content` error. The `reasoning.encrypted_content` is requested and re-sent automatically for reasoning models, but there’s no way to disable it via the public API and no defensive handling when it fails.

For reasoning models, `OpenAIModelProfile` sets `openai_supports_encrypted_reasoning_content=True`, which causes `OpenAIResponsesModel` to:

Include `'reasoning.encrypted_content'` in the request include param (openai.py L1682)
Store the encrypted content as `ThinkingPart.signature`
Send it back as `encrypted_content` on subsequent requests (openai.py L2101-2115)
The encrypted content is cryptographically bound to the originating OpenAI organization. If a gateway/proxy routes the next request through a different org, OpenAI rejects it:
```
Error code: 400 - {
  "error": {
    "message": "The encrypted content for item rs_09beb6f8... could not be verified.
                Reason: Encrypted content organization_id did not match the target organization.",
    "code": "invalid_encrypted_content"
  }
}
```
There’s an existing guard for cross-provider signatures (`item.provider_name == self.system` at L2104), but nothing for cross-org within the same provider.

### Suggested Fix
Two complementary changes:

1. `gateway_provider` should disable encrypted content by default

When using `gateway_provider`, org affinity can’t be guaranteed. The provider should automatically set `openai_supports_encrypted_reasoning_content=False` on the profile (or provide an option to control this). The model still works — it regenerates reasoning from the summary/raw content instead of replaying encrypted tokens.

2. Defensive retry on invalid_encrypted_content

Even with the above, org mismatches can happen for other reasons (key rotation, config changes, custom base_url on a standard provider). `OpenAIResponsesModel` should catch the `invalid_encrypted_content` error, strip `ThinkingPart.signature` from message history, and retry once:

```
from pydantic_ai.messages import ModelMessage, ModelResponse, ThinkingPart

def _strip_encrypted_signatures(messages: list[ModelMessage]) -> list[ModelMessage]:
    cleaned = []
    for msg in messages:
            new_parts = [
                replace(part, signature=None)
                if isinstance(part, ThinkingPart) and part.signature
                else part
                for part in msg.parts
            ]
            msg = replace(msg, parts=new_parts)
        cleaned.append(msg)
    return cleaned
```

### Current Workaround
`openai_supports_encrypted_reasoning_content` exists on `OpenAIModelProfile` but isn’t exposed via `ModelSettings` or the `OpenAIResponsesModel` constructor. It can be disabled via `dataclasses.replace` on the profile:

```
from pydantic_ai.profiles.openai import openai_model_profile, OpenAIModelProfile

profile = replace(
    OpenAIModelProfile.from_profile(openai_model_profile("gpt-5-mini")),
    openai_supports_encrypted_reasoning_content=False,
)
model = OpenAIResponsesModel("gpt-5-mini", profile=profile, provider=...)
```

```Python
from pydantic_ai import Agent
from pydantic_ai.providers.gateway import gateway_provider

model = OpenAIResponsesModel(
    "gpt-5-mini",
    provider=gateway_provider("openai", base_url="https://gateway.pydantic.dev/proxy/openai"),
)
agent = Agent(model)

# First run succeeds — response includes ThinkingPart with encrypted signature
result = await agent.run("What is the capital of France?")

# Second run re-sends message_history with the ThinkingPart.signature.
# If the gateway routes through a different org, this fails with
# invalid_encrypted_content.
result = await agent.run(
    "What about Germany?",
)
```

### Logfire Trace

https://logfire-us.pydantic.dev/public-trace/ab04a5cd-ad91-4bd7-b505-f8e56a8cd7c9?spanId=d5f5b840196a7557

- Python: 3.11.9
- Pydantic AI: 1.67.0
- LLM provider SDK: 2.21.0

JasonCZMeng:
## Summary

Prevent `invalid_encrypted_content` errors when using `OpenAIResponsesModel` with `gateway_provider()` in multi-turn conversations.

Fixes #4608

## Root Cause

Encrypted reasoning content signatures are org-bound — they're only valid for the OpenAI organization that generated them. When using `gateway_provider()`, requests may be routed through different orgs across turns, making previously-received signatures invalid. The second turn then fails with:

```
Error code: 400 - invalid_encrypted_content
```

### 1. Gateway detection (`providers/openai.py`, `providers/gateway.py`)
- Added `is_gateway: bool` attribute to `OpenAIProvider`
- `gateway_provider()` sets `is_gateway = True` on OpenAI providers it creates

### 2. Skip encrypted content for gateways (`models/openai.py`)
- When `is_gateway` is true, skip adding `reasoning.encrypted_content` to the `include` list
- When `is_gateway` is true, skip sending `signature` as `encrypted_content` in `_map_messages`

### 3. Defensive retry (`models/openai.py`)
- Catch `invalid_encrypted_content` 400 errors
- Strip `encrypted_content` from all reasoning items in the message history
- Remove `reasoning.encrypted_content` from the include list
- Retry the request once (handles edge cases like mid-conversation provider changes)

## Testing

The gateway detection is straightforward (flag check). The retry path handles the case where encrypted content slips through despite the flag (e.g., messages from a previous non-gateway session).


JasonCZMeng:
CI is green except for a flaky timing assertion in `test_tenacity.py::TestIntegration::test_async_transport_with_wait_retry_after` (unrelated to this PR):

```
assert 0.8 <= (344.67 - 342.60) <= 2.0  # elapsed ~2.07s, exceeded 2.0s upper bound
```

This only affects the `test on 3.11 (lowest-versions)` job. All other jobs pass (lint, mypy, docs, coverage, all test matrices). Happy to re-trigger if needed.

DouweM:
Marking as draft pending discussion in https://github.com/pydantic/pydantic-ai/issues/4608.

JasonCZMeng:
Closing this PR based on maintainer feedback. The existing `openai_send_reasoning_ids=False` model setting already covers this use case for cross-org gateway scenarios. A docs update to make this more discoverable for gateway users would be the better path forward. Thanks for the guidance!

PR #4624, opened 2026-03-11
fix: AG-UI follow-up messages failing after built-in tool use

### LINKED ISSUE #4623: AG-UI: built-in tool results dropped on follow-up requests
I ran into this using AG-UI with `web_fetch`. First message works fine — the tool runs, returns a result, everything's great. But when you send a follow-up message, the API blows up with:

```
messages.1: `web_fetch` tool use with id `srvtoolu_...` was found without a corresponding `web_fetch_tool_result` block
```

What happens: `_event_stream.py` serializes the `BuiltinToolReturnPart` content (a dict) to a JSON string via `model_response_str()` for the `ToolCallResultEvent` — AG-UI protocol only supports string content. When the frontend sends the history back on the next request, `_adapter.py` stuffs that string straight into a new `BuiltinToolReturnPart`. Then `models/anthropic.py` checks `isinstance(response_part.content, dict)` to build `BetaWebFetchToolResultBlockParam` — string fails the check, the result block gets silently dropped, and Anthropic rightfully complains about a missing tool result.

Same issue applies to `web_search` and `code_execution`.

Repro:
```python
from pydantic_ai import Agent
from pydantic_ai.builtin_tools import WebFetchTool
from pydantic_ai.messages import BuiltinToolReturnPart

agent = Agent('anthropic:claude-sonnet-4-20250514', builtin_tools=[WebFetchTool()])

# First call works, web_fetch runs
result = await agent.run('Fetch https://example.com and tell me the page title.')

# Simulate AG-UI round-trip: dict content -> JSON string
# (this is what _event_stream.py does on the way out, and _adapter.py doesn't undo on the way back)
history = result.all_messages()
for msg in history:
    for part in msg.parts:
        if isinstance(part, BuiltinToolReturnPart) and not isinstance(part.content, str):
            part.content = part.model_response_str()

# Follow-up with this history — boom
result2 = await agent.run('What URL did you fetch?', message_history=history)
```

Fix: `json.loads()` the string content in `_adapter.py` when creating `BuiltinToolReturnPart` from a `ToolMessage`, so the dict gets restored before it reaches model code.

gcheshkov:
- Closes #4623

### Pre-Review Checklist

### Pre-Merge Checklist

N/A for docs — no new public API, internal bugfix only.

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
+++ b/pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
@@ -166,7 +166,7 @@ class AGUIAdapter(UIAdapter[RunAgentInput, Message, BaseEvent, AgentDepsT, Outpu
                                 case _:  # pragma: no cover
                                     raise ValueError(f'Unsupported user message part type: {type(part)}')
 
-                        if user_prompt_content:  # pragma: no branch
+                        if user_prompt_content:
                             content_to_add = (
                                 user_prompt_content[0]
                                 if len(user_prompt_content) == 1 and isinstance(user_prompt_content[0], str)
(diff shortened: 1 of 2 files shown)


DouweM:
@gcheshkov Thank you!

PR #4634, opened 2026-03-12
Add tests for embedding settings merging

### LINKED ISSUE #4633: Add tests for embedding settings merging
### Description

`EmbeddingSettings` supports both embedder-level defaults and per-call overrides, merged via `merge_embedding_settings()`. Right now this behavior is only covered indirectly via higher-level tests. A small regression in the merge logic (for example changing from a shallow to a deep merge, or altering truthiness checks) could be hard to catch.

### Proposed changes

Add focused unit tests for `merge_embedding_settings()` that cover:

- both inputs are `None`
- only base settings provided
- only override settings provided
- base and overrides provided, with overlapping and non-overlapping keys (per-call values win while other base keys are preserved)

These tests can live in `tests/test_embeddings.py` next to the other embedding tests and should not introduce any new external dependencies.

echoVic:
This adds focused unit tests for `merge_embedding_settings()` to cover the main merge scenarios (no settings, only base, only overrides, and base + overrides with overlapping keys).

Closes #4633.

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/embeddings/cohere.py
+++ b/pydantic_ai_slim/pydantic_ai/embeddings/cohere.py
@@ -224,7 +224,7 @@ class CohereEmbeddingModel(EmbeddingModel):
 def _map_usage(response: EmbedByTypeResponse, provider: str, provider_url: str, model: str) -> RequestUsage:
     u = response.meta
     if u is None or u.billed_units is None:
-        return RequestUsage()  # pragma: no cover
+        return RequestUsage()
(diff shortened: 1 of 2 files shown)


rar-file:
Looks good + focused. One nit that came up from the docs bot on a related PR: when referencing public APIs in docs/tests, they prefer linking identifiers consistently (not critical here).

Otherwise the merge cases covered (None/base-only/override-only/override precedence) match the semantics in pydantic_ai.embeddings.settings.merge_embedding_settings().

PR #4636, opened 2026-03-12
Document embedding settings precedence and merging

### LINKED ISSUE #4635: Document embedding settings precedence and merging
### Description

The embeddings documentation introduces `EmbeddingSettings` and shows that settings can be specified on the embedder as well as per call, but it does not explicitly describe how these settings are merged or which source takes precedence. Users building larger systems (or porting existing code) can be unsure about how model-level defaults, embedder-level defaults, and per-call overrides interact.

### Proposed changes

In `docs/embeddings.md`, extend the **Settings** section with a short subsection that describes:

- that settings can come from three places: model defaults, `Embedder(..., settings=...)`, and per-call `settings=` arguments
- that these settings are merged using the same semantics as `merge_embedding_settings()` (later sources override earlier ones for the same key, while other keys are preserved)
- a small, concrete example that shows a stack of settings (model, embedder, call) and the resulting merged dictionary that is sent to the provider.

This would make it easier for readers to reason about how to configure embeddings in more complex applications.

echoVic:
Clarify how embedding settings are combined from model, embedder, and per-call arguments.

This documents the precedence and merging rules, with an example showing how `EmbeddingSettings` from the model and embedder interact, and how call-time overrides work.

Closes #4635.

### DIFF
--- a/docs/embeddings.md
+++ b/docs/embeddings.md
@@ -678,6 +678,33 @@ async def main():
 
 _(This example is complete, it can be run "as is" — you'll need to add `asyncio.run(main())` to run `main`)_
 
+### Settings precedence and merging
+
+Embedding configuration can come from three places:
+
+1. **Model defaults** – when you construct an embedding model with `settings=...`.
+2. **Embedder defaults** – when you create an [`Embedder`][pydantic_ai.embeddings.Embedder] with `settings=...`.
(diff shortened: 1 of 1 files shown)


rar-file:
Nice addition — this clarifies the merge semantics well. Two small polish points (matching the docs style rules used elsewhere in docs/embeddings.md):

1) Link Python API names (instead of only backticks), e.g. [`Embedder`][pydantic_ai.embeddings.Embedder], [`embed_query()`][pydantic_ai.embeddings.Embedder.embed_query], [`embed_documents()`][pydantic_ai.embeddings.Embedder.embed_documents].
2) (Optional) Since there’s already a later **Settings** section, it might be worth nesting the merge-precedence explanation into that section to avoid two separate top-level settings blocks.

PR #4643, opened 2026-03-13
Add top-level AgentSpec fields and full Agent kwargs to from_spec

DouweM:
AgentSpec now includes name, instructions, model_settings, retries, output_retries, end_strategy, tool_timeout, instrument, and metadata as top-level fields alongside model and capabilities.

Agent.from_spec accepts all Agent constructor kwargs to supplement the spec with runtime values. Spec fields serve as defaults, explicit args override scalars, and list-like values (instructions, capabilities, model_settings) are merged additively.

- Closes #<issue>

### Pre-Review Checklist

### Pre-Merge Checklist

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/agent/__init__.py
+++ b/pydantic_ai_slim/pydantic_ai/agent/__init__.py
@@ -463,6 +463,7 @@ class Agent(AbstractAgent[AgentDepsT, OutputDataT]):
         system_prompt: str | Sequence[str] = (),
         deps_type: type[Any] = NoneType,
         name: str | None = None,
+        description: str | None = None,
         model_settings: ModelSettings | None = None,
         retries: int | None = None,
         validation_context: Any = None,
@@ -499,6 +500,7 @@ class Agent(AbstractAgent[AgentDepsT, OutputDataT]):
             system_prompt: Static system prompts.
             deps_type: The type used for dependency injection.
             name: The agent name, overrides spec `name` if provided.
+            description: The agent description, overrides spec `description` if provided.
             model_settings: Model request settings.
             retries: Default retries for tool calls and output validation, overrides spec `retries` if provided.
             validation_context: Pydantic validation context for tool arguments and outputs.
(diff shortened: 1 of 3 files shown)


gambletan (commented):
## Review

This PR adds the full set of `Agent.__init__` kwargs to `Agent.from_spec()`, allowing spec-based agent creation to be fully parameterized. It also extends `AgentSpec` with new scalar fields (`name`, `retries`, `end_strategy`, etc.) and merges capabilities from both spec and kwargs.

### Issues and suggestions:

1. **`output_type` default masks spec-provided values**: The `output_type` parameter defaults to `str`, but there's no corresponding `output_type` field in `AgentSpec`. This means there's no way to set `output_type` from a spec file — it always requires the kwarg. That's probably fine for now (output types are hard to serialize), but the docstring should clarify this is intentionally kwarg-only.

2. **`model` falsy-value semantics**: The line `model=model or validated_spec.model` means passing `model=""` or `model=0` (admittedly unusual) would fall through to the spec model. Using `model if model is not None else validated_spec.model` would be more consistent with how other fields are handled (e.g., `retries if retries is not None else ...`).

3. **`instructions` merging — order matters**: `merged_instructions` extends spec instructions with kwarg instructions. This means spec instructions come first. The docstring says "scalar spec fields are used as defaults that explicit arguments override" but for instructions, they're *merged* not overridden, and the spec takes priority by position. This should be documented explicitly, and it might be surprising to users who expect kwargs to override spec values.

4. **`end_strategy` default mismatch**: `AgentSpec.end_strategy` defaults to `'early'`, but `Agent.__init__` defaults to `None` (which means "use the library default"). If the spec doesn't set `end_strategy`, it will always send `'early'` to the Agent constructor even if the caller didn't intend to set it. The spec field should probably default to `None` too, so it means "not specified."

5. **`model_settings` merge**: `merge_model_settings(validated_spec.model_settings, model_settings)` — I'd verify which argument takes precedence in `merge_model_settings`. If the second argument overrides the first, this is correct. If not, the kwarg won't override the spec.

6. **Return type widened to `Agent[Any, Any]`**: Previously this returned `Agent[None, str]`. The new signature is more permissive but loses type safety for callers who don't pass `deps_type` or `output_type`. Consider using overloads or a TypeVar to preserve inference when possible.

7. **No tests for the new kwargs**: The PR adds ~25 new parameters but doesn't include tests for the merging/override behavior. At minimum, tests should cover: (a) kwarg overrides spec value, (b) spec value used when kwarg is default/None, (c) capabilities merge, (d) instructions merge order.

The design direction is solid — making `from_spec` a full-featured factory is the right call. The main concerns are around default-value semantics and test coverage.

PR #4644, opened 2026-03-13
fix(anthropic): handle malformed tool args in retry path

### LINKED ISSUE #4430: Anthropic retry path crashes on malformed tool args (args_as_dict ValueError) before model can self-correct

gambletan:
## Summary

Fixes #4430

When the LLM produces malformed JSON in tool call arguments (e.g. `{"query": "bad query", "file_ids":[4556]</parameter>\n<parameter name="limit": 8}`), the Anthropic model's history remapping calls `args_as_dict()` which raises a `ValueError` before the retry prompt can be sent to the model. This prevents the model from self-correcting.

**Root cause:** `args_as_dict()` uses `pydantic_core.from_json()` which raises `ValueError` on invalid JSON. The Anthropic mapping path calls this when constructing `BetaToolUseBlockParam(input=...)`. Other model implementations (OpenAI) use `args_as_json_str()` instead, which returns the raw string without parsing.

**Fix:** Add a `_safe_args_as_dict()` helper that catches `ValueError` and returns an empty dict `{}`, allowing the retry flow to proceed. The retry prompt (`RetryPromptPart`) already contains the full parse-error details for the model to self-correct.

- Only the regular `ToolCallPart` mapping (line 800) is affected; `BuiltinToolCallPart` args come from Anthropic's own API and are always valid JSON.

## Test plan

- [ ] Verify the minimal reproduction from #4430 no longer crashes with `ValueError`
- [ ] Verify that the agent retry loop proceeds and the model receives the `RetryPromptPart` with the JSON error details
- [ ] Verify existing Anthropic tool-calling tests still pass

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/anthropic.py
+++ b/pydantic_ai_slim/pydantic_ai/models/anthropic.py
@@ -797,7 +797,7 @@ class AnthropicModel(Model):
                             id=_guard_tool_call_id(t=response_part),
                             type='tool_use',
                             name=response_part.tool_name,
-                            input=response_part.args_as_dict(),
+                            input=_safe_args_as_dict(response_part),
                         )
                         assistant_content_params.append(tool_use_block_param)
                     elif isinstance(response_part, ThinkingPart):
@@ -1506,3 +1506,20 @@ def _map_mcp_server_result_block(
         content=item.model_dump(mode='json', include={'content', 'is_error'}),
         tool_call_id=item.tool_use_id,
     )
+
+
+def _safe_args_as_dict(part: ToolCallPart | BuiltinToolCallPart) -> dict[str, Any]:
+    """Safely extract tool call arguments as a dict.
+
+    When the LLM produces malformed JSON in tool call arguments, `args_as_dict()`
+    raises a `ValueError`. This helper catches that error and returns an empty dict
+    so that the retry flow can proceed — the retry prompt already contains the
+    parse-error details for the model to self-correct.
+
+    See: https://github.com/pydantic/pydantic-ai/issues/4430
+    """
+    try:
+        return part.args_as_dict()
+    except ValueError:
+        return {}
+


gambletan (commented):
## Review

This is a targeted fix for a real crash in the retry path. When the Anthropic model returns malformed JSON as tool call arguments, `args_as_dict()` raises `ValueError`, which kills the entire retry flow before the model gets a chance to self-correct.

The approach is sound — catching the error and returning `{}` lets the retry prompt (which already contains the parse error details) reach the model.

A few things to consider:

1. **Silent data loss risk**: Returning `{}` means the malformed arguments are completely discarded from the re-sent conversation history. The model sees an empty `input` dict for that tool use block, while the subsequent user turn contains the error message. This *works* because the error message is descriptive, but it creates an inconsistency in the conversation: the model sees it "called" a tool with `{}` args yet got a parse error about completely different content. Consider whether it would be better to pass the raw string as a single-key dict like `{"_raw": part.args_as_json()}` so the model can see what it actually produced. (This is a minor point — `{}` likely works fine in practice.)

2. **Scope — only Anthropic?**: The helper `_safe_args_as_dict` is defined in the Anthropic module but this same crash can happen with any model that produces malformed JSON tool calls. If other adapters (Google, OpenAI, etc.) also call `args_as_dict()` in their `_map_message` paths, they have the same latent bug. Worth checking whether this fix should be applied more broadly — perhaps as a method on `ToolCallPart` itself.

3. **`BuiltinToolCallPart` in the type signature**: The function accepts `ToolCallPart | BuiltinToolCallPart` but it's only called with `response_part` which at that point is known to be `ToolCallPart`. Is there actually a code path where `BuiltinToolCallPart` is passed in? If not, narrowing the type would be cleaner.

4. **No test**: This is a fix for a crash scenario (issue #4430) but there's no test that exercises the malformed-args retry path. A test that sends a `ToolCallPart` with invalid JSON args through `_map_message` would prevent regressions.

The fix itself is correct and safe. The main suggestion is to consider broadening it and adding a test.

DouweM:
Issue was closed by another PR.

PR #4646, opened 2026-03-13
fix: remove unsupported `id` field from Gemini API function call/response dicts

### LINKED ISSUE #4645: Pydantic-AI's Google Model Adapter Sends Unsupported idField to Gemini API, Causing 400 Errors
### Initial Checks

### Description

**Problem**
When using the gemini-3.1-pro-preview-ptmodel with pydantic-ai, function calling fails with a 400 error. The error message indicates the Gemini API does not accept an idfield within the function_calland function_responseobjects in the request payload ("Invalid JSON payload received. Unknown name \"id\" at 'contents[1].parts[0].function_call': Cannot find field.").
The current version of pydantic-ai (specifically in pydantic_ai/models/google.py) includes this unsupported idfield in three locations when constructing requests for the Gemini API, causing compatibility issues.
**Proposal**
Update the pydantic_ai.models.googlemodule to align with the Gemini API's expected schema. The fix involves removing the idkey from the function_calland function_responsedictionaries in the following three code sections:
In the function_calldictionary construction (around line 1114, FunctionCallDict).
In the function_responsedictionary for ToolReturnPart(around line 692).
In the function_responsedictionary for RetryPromptPart(around line 705).
This change would ensure seamless compatibility with the Gemini API for function calling features. The modified code blocks should resemble the user's working local patch:
# Example for the first change
function_call = FunctionCallDict(name=item.tool_name, args=item.args_as_dict())  # `id` parameter removed
# Example for ToolReturnPart
'function_response': {
    'name': part.tool_name,
    'response': part.model_response_object(),
    # 'id': part.tool_call_id,  # This line should be removed
}
This is a targeted fix for API schema compatibility and should not introduce side effects for other models or features.

```Python

```

### Logfire Trace

_No response_

- Python:
- Pydantic AI:
- LLM provider SDK:

Ker102:
Hey, this should fix #4645.

The Gemini API doesn't support an `id` field in `function_call` or `function_response` objects — it throws a 400 with "Unknown name "id" at contents[x].parts[y].function_call: Cannot find field."

Looks like the `id` was added to match what other providers (like OpenAI) expect, but Gemini's schema just doesn't have it. Removed it from the three places where it gets included:

- The `FunctionCallDict` construction in `_content_model_response()`
- The `function_response` dict for `ToolReturnPart` in `_map_messages()`
- The `function_response` dict for `RetryPromptPart` in `_map_messages()`

Pretty straightforward — just dropping the unsupported field from outgoing requests. Doesn't affect any internal PydanticAI logic since `tool_call_id` is still tracked internally, it just doesn't get sent to the Gemini API anymore.

Let me know if anything needs adjusting!

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/google.py
+++ b/pydantic_ai_slim/pydantic_ai/models/google.py
@@ -687,7 +687,6 @@ class GoogleModel(Model):
                                 'function_response': {
                                     'name': part.tool_name,
                                     'response': part.model_response_object(),
-                                    'id': part.tool_call_id,
                                 }
                             }
                         )
@@ -700,7 +699,6 @@ class GoogleModel(Model):
                                     'function_response': {
                                         'name': part.tool_name,
                                         'response': {'error': part.model_response()},
-                                        'id': part.tool_call_id,
                                     }
                                 }
                             )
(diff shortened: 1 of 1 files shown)


DouweM:
Marking as draft pending discussion in the issue.

DouweM:
Not a viable fix; tool call IDs are needed to map calls to results.

PR #4680, opened 2026-03-16
fix: new_messages() no longer includes user prompt from message_history

### LINKED ISSUE #4669: result.new_messages() always includes user prompt from message_history
### Initial Checks

### Description

After upgrading **pydantic-ai-slim** from **1.63.0** to **1.64.0+**, the behavior of `result.new_messages()` appears to have changed.

When calling `agent.run()` with `message_history` (and without `user_prompt`), the messages supplied in `message_history` used to be treated purely as prior context, so `result.new_messages()` only returned messages generated during the run.

Starting in **1.64.0**, `new_messages()` also returns the user prompt that was provided inside `message_history`.

For example, the following code now returns the original `ModelRequest` as part of `new_messages()` even though it was part of the input history.

This seems related to changes introduced in this PR:
https://github.com/pydantic/pydantic-ai/pull/4419/changes#diff-71730070455237accebd709ba7fe41ce60edbd238af02e7d86a208cf8f10ab12R1543-R1545

The last version that behaves as expected for me is **1.63.0**.

Is this an intended change, or a regression? If intentional, is there a recommended way to get only the messages generated during the run?

### Old behavior (≤1.63.0)

`new_messages()` returns only messages generated during the run.

### New behavior (≥1.64.0)

`new_messages()` includes the `ModelRequest` from `message_history`, meaning the user prompt `"hi"` is returned even though it was already part of the provided history.

```Python
result = await agent.run(
    message_history=[
        ModelRequest(
            parts=[
                UserPromptPart(
                    content="hi"
                )
            ]
        )
    ],
    deps=deps,
)

print(result.new_messages())
```

### Logfire Trace

_No response_

- Python: 3.12
- Pydantic AI: 1.68.0
- LLM provider SDK: doesn't matter

majiayu000:
Fixes #4669

## Summary

When calling `agent.run()` with `message_history` and without `user_prompt`, `new_messages()` was incorrectly returning the resumed `ModelRequest` from the provided history. This is a regression introduced in #4419.

**Root cause**: After history processing in `_agent_graph.py`, the code unconditionally assigns `run_id = ctx.state.run_id` on the last message when its `run_id is None`. When `is_resuming_without_prompt=True`, this stamps the resumed request with the current run's `run_id`, causing `_first_new_message_index()` to include it in `new_messages()`.

**Fix**: Add `and not self.is_resuming_without_prompt` to the condition on line 534, so resumed requests keep their original `run_id` (None or from a previous run).

## Changes

- 1-line fix in `_agent_graph.py`
- Updated 4 existing test assertions in `test_history_processor.py` to match correct behavior
- Added regression test `test_new_messages_excludes_history_prompt_without_user_prompt`
- Updated 2 test snapshots in `test_agent.py`

## Test plan

- [x] All 28 tests in `test_history_processor.py` pass
- [x] All 4 `message_history`/`new_messages` tests in `test_agent.py` pass

### DIFF
--- a/tests/test_agent.py
+++ b/tests/test_agent.py
@@ -2610,7 +2610,6 @@ def test_run_with_history_ending_on_model_request_and_no_user_prompt():
                 ],
                 timestamp=IsNow(tz=timezone.utc),
                 instructions='New instructions',
-                run_id=IsStr(),
             ),
             ModelResponse(
                 parts=[TextPart(content='success (no tool calls)')],
@@ -2622,7 +2621,7 @@ def test_run_with_history_ending_on_model_request_and_no_user_prompt():
         ]
     )
 
-    assert result.new_messages() == result.all_messages()[-2:]
+    assert result.new_messages() == result.all_messages()[-1:]
 
 
 def test_run_with_history_ending_on_model_response_with_tool_calls_and_no_user_prompt():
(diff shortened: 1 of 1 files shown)


DouweM:
This breaks tests.

PR #4684, opened 2026-03-16
fix: preserve ToolReturn metadata through Temporal serialization

### LINKED ISSUE #4676: ToolReturn.metadata silently dropped by TemporalAgent
### Initial Checks

### Description

When a tool returns a `ToolReturn` with `metadata`, the metadata is
correctly propagated to `ToolReturnPart.metadata` in non-Temporal
agent runs. However, when the same tool runs inside a `TemporalAgent`,
the `metadata` field is silently dropped — `ToolReturnPart.metadata`
is always `None`.

> **metadata**: Optional metadata that your application can access but
> is not sent to the LLM. Useful for logging, debugging, or additional
> processing.

## Root Cause

The issue is in `pydantic_ai/durable_exec/temporal/_toolset.py`.

The Temporal wrapper uses an intermediate `_ToolReturn` dataclass to
serialize tool results across the activity boundary:

```python
# _toolset.py lines 55-58
@dataclass
class _ToolReturn:
    result: ToolReturnContent          # <-- only captures the raw return value
    kind: Literal['tool_return'] = 'tool_return'
```

When a tool returns a `ToolReturn` object:

1. **`_wrap_call_tool_result`** (line 95-98) wraps the entire
   `ToolReturn` instance as `_ToolReturn(result=<ToolReturn object>)`.

2. Temporal serializes `_ToolReturn` to JSON for the activity result.
   The `ToolReturn` dataclass becomes a plain dict:
   `{"return_value": "...", "metadata": {...}, "content": null, "kind": "tool-return"}`.

3. On deserialization back in the workflow, `_ToolReturn.result` is
   typed as `ToolReturnContent`. Pydantic/Temporal reconstructs it as
   a plain `dict`, **not** a `ToolReturn` instance.

4. **`_unwrap_call_tool_result`** (line 106-108) returns
   `result.result` — which is now a plain dict.

5. Back in `_agent_graph.py` line 1360, the check
   `isinstance(tool_result, ToolReturn)` returns `False` (it's a dict).

6. The else branch (line 1362-1385) wraps it as a generic return value
   and constructs a `ToolReturnPart` **without metadata**.

The `content` field of `ToolReturn` is also lost by the same mechanism.

## Suggested Fix

The `_ToolReturn` dataclass should preserve all three fields of
`ToolReturn` so they survive the serialization round-trip:

```python
@dataclass
class _ToolReturn:
    result: ToolReturnContent
    content: Sequence[UserContent] | None = None
    metadata: Any = None
    kind: Literal['tool_return'] = 'tool_return'
```

And the wrap/unwrap methods should decompose and reconstruct:

```python
async def _wrap_call_tool_result(self, coro: Awaitable[Any]) -> CallToolResult:
    try:
        result = await coro
        if isinstance(result, ToolReturn):
            return _ToolReturn(
                result=result.return_value,
                content=result.content,
                metadata=result.metadata,
            )
        return _ToolReturn(result=result)
    except ApprovalRequired as e:
        return _ApprovalRequired(metadata=e.metadata)
    except CallDeferred as e:
        return _CallDeferred(metadata=e.metadata)
    except ModelRetry as e:
        return _ModelRetry(message=e.message)

def _unwrap_call_tool_result(self, result: CallToolResult) -> Any:
    if isinstance(result, _ToolReturn):
        if result.content is not None or result.metadata is not None:
            return ToolReturn(
                return_value=result.result,
                content=result.content,
                metadata=result.metadata,
            )
        return result.result
    elif isinstance(result, _ApprovalRequired):
        raise ApprovalRequired(metadata=result.metadata)
    elif isinstance(result, _CallDeferred):
        raise CallDeferred(metadata=result.metadata)
    elif isinstance(result, _ModelRetry):
        raise ModelRetry(result.message)
    else:
        assert_never(result)
```

## Impact

Any user of `TemporalAgent` who relies on `ToolReturn.metadata` or
`ToolReturn.content` will find these fields silently set to `None` in
the resulting message history. There is no error or warning — the data
is simply lost.

## Workaround

Encode the metadata into the `return_value` string itself (e.g. as a
JSON-encoded prefix or structured string) and parse it back out in the
workflow. Alternatively, scan `ToolCallPart.args` from the
`ModelResponse` messages instead of relying on `ToolReturnPart.metadata`.

```Python
from pydantic_ai import Agent, RunContext, ToolReturn

agent = Agent("openai:gpt-4o-mini")

@agent.tool_plain
def my_tool(query: str) -> ToolReturn:
    return ToolReturn(
        return_value=f"Processed: {query}",
        metadata={"request_id": "abc-123", "source": "my_tool"},
    )

# --- Local run: metadata is preserved ---
result = agent.run_sync("test")
for msg in result.all_messages():
    for part in msg.parts:
        if hasattr(part, "metadata"):
            print(f"Local — metadata: {part.metadata}")
            # Output: Local — metadata: {'request_id': 'abc-123', 'source': 'my_tool'}

# --- TemporalAgent run: metadata is None ---
# (same agent wrapped with TemporalAgent, run inside a Temporal workflow)
# Output: Temporal — metadata: None
```

### Logfire Trace

_No response_

- Python: 3.13
- Pydantic AI: 1.68
- LLM provider SDK: AsyncOpenAi

alvinttang:
## Summary
- Fixes `_ToolReturn` dataclass in `_toolset.py` to include `content` and `metadata` fields
- Updates `_wrap_call_tool_result` to decompose `ToolReturn` into individual `_ToolReturn` fields instead of stuffing the whole object into `result`
- Updates `_unwrap_call_tool_result` to reconstruct a proper `ToolReturn` when content or metadata are present
- Adds two unit tests verifying the serialization round-trip preserves metadata/content

## Problem
When a tool returns a `ToolReturn` with `metadata` and `content`, `_wrap_call_tool_result` was putting the entire `ToolReturn` object into `_ToolReturn(result=...)`. After Temporal serialization/deserialization (via Pydantic), the `ToolReturn` became a plain dict, so `isinstance(tool_result, ToolReturn)` in `_agent_graph.py` failed and metadata was silently lost.

## Test plan
- [x] Unit test: `test_tool_return_metadata_survives_serialization` -- verifies metadata and content survive Pydantic serialization round-trip and `_unwrap` reconstructs a proper `ToolReturn`
- [x] Unit test: `test_tool_return_without_metadata_unchanged` -- verifies plain results without metadata pass through unchanged

Closes #4676

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_toolset.py
+++ b/pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_toolset.py
@@ -5,14 +5,14 @@ from collections.abc import Awaitable, Callable
 from dataclasses import dataclass
 from typing import Annotated, Any, Literal
 
-from pydantic import ConfigDict, Discriminator, with_config
+from pydantic import ConfigDict, Discriminator, Tag, with_config
 from temporalio import workflow
 from temporalio.workflow import ActivityConfig
 from typing_extensions import Self, assert_never
 
 from pydantic_ai import AbstractToolset, FunctionToolset, ToolsetTool, WrapperToolset
 from pydantic_ai.exceptions import ApprovalRequired, CallDeferred, ModelRetry
-from pydantic_ai.messages import ToolReturnContent
+from pydantic_ai.messages import ToolReturn, ToolReturnContent
 from pydantic_ai.tools import AgentDepsT, RunContext, ToolDefinition
 from pydantic_ai.toolsets._dynamic import DynamicToolset
 
@@ -52,9 +52,23 @@ class _ModelRetry:
     kind: Literal['model_retry'] = 'model_retry'
 
 
+def _result_discriminator(v: Any) -> str:
+    if isinstance(v, ToolReturn) or (isinstance(v, dict) and v.get('kind') == 'tool-return'):  # pyright: ignore[reportUnknownMemberType]
+        return 'tool-return'
+    return 'content'
+
+
+# Defined at module level so Pydantic resolves the Annotated metadata at runtime,
+# not as a string annotation (which would lose the discriminator under `from __future__ import annotations`).
+_ToolReturnResult = Annotated[
+    Annotated[ToolReturn, Tag('tool-return')] | Annotated[ToolReturnContent, Tag('content')],
+    Discriminator(_result_discriminator),
+]
+
+
 @dataclass
 class _ToolReturn:
-    result: ToolReturnContent
+    result: _ToolReturnResult
     kind: Literal['tool_return'] = 'tool_return'
 
 
--- a/tests/test_temporal.py
+++ b/tests/test_temporal.py
@@ -41,6 +41,7 @@ from pydantic_ai import (
     TextPartDelta,
     ToolCallPart,
     ToolCallPartDelta,
+    ToolReturn,
     ToolReturnPart,
     UserPromptPart,
     WebSearchTool,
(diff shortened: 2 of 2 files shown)


pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_toolset.py
```diff
@@ -55,6 +55,8 @@ class _ModelRetry:
 @dataclass
 class _ToolReturn:
     result: ToolReturnContent
```
DouweM: Instead of the 2 new fields, I think we can make this `ToolReturn | ToolReturnContent` so that it automatically checks whether it has `kind='tool-return'` and creates a `ToolReturn` if so.

If that works, we don't need the other changes at all!

alvinttang:
Thanks @DouweM — the discriminated union approach is much cleaner. I've reworked the PR to type `_ToolReturn.result` as `ToolReturn | ToolReturnContent` with a discriminator on `kind`, so Pydantic handles the deserialization automatically. No changes needed to `_wrap`/`_unwrap` anymore.

Also addressed the other review feedback: replaced the unit tests with an integration test using `TemporalAgent`, switched to `snapshot()` assertions, and fixed the import locations.

Let me know if you'd like any further changes!

PR #4701, opened 2026-03-17
fix: exclude resumed request from new_messages() when resuming from history

### LINKED ISSUE #4669: result.new_messages() always includes user prompt from message_history

joaquinhuigomez:
## Summary

- Fixes `result.new_messages()` incorrectly including the user prompt from `message_history` when calling `agent.run()` without `user_prompt` (regression since v1.64.0)
- The root cause was in `_first_new_message_index`: it would include the resumed request when its `run_id` matched the current run, but the framework itself always assigns the current `run_id` to the request, making the check a no-op
- Changed `_first_new_message_index` to always exclude the resumed request, since it was provided by the caller as prior context — not generated during the current run

Fixes #4669

## Test plan

- [x] Added regression test `test_new_messages_excludes_user_prompt_from_message_history` that reproduces the exact scenario from the issue
- [x] Updated existing tests that asserted the old (buggy) behavior: resumed requests are now excluded from `new_messages()`
- [x] Verified the fallback path (`_first_run_id_index`) still works correctly when a history processor replaces the resumed request entirely
- [x] All 2394 non-eval tests pass; the 8 eval failures are pre-existing snapshot formatting issues

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/_agent_graph.py
+++ b/pydantic_ai_slim/pydantic_ai/_agent_graph.py
@@ -1500,17 +1500,23 @@ def _first_new_message_index(
     *,
     resumed_request: _messages.ModelRequest | None,
 ) -> int:
-    """Return the first index that should be included in `new_messages()`."""
+    """Return the first index that should be included in `new_messages()`.
+
+    The resumed request (the trailing ModelRequest from message_history that is
+    re-used when no user_prompt is provided) is always excluded from
+    new_messages(), because it was supplied by the caller as prior context —
+    not generated during the current run.
+
+    See https://github.com/pydantic/pydantic-ai/issues/4669.
+    """
     if resumed_request is not None:
         for index, message in enumerate(messages):
             if message is resumed_request:
-                # Include the resumed request in new_messages only if it was
-                # mapped/created during the current run (e.g., via adapters).
-                return index if message.run_id == run_id else index + 1
+                return index + 1
 
         for index in range(len(messages) - 1, -1, -1):
             if _is_same_request(messages[index], resumed_request):
-                return index if messages[index].run_id == run_id else index + 1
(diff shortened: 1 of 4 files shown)


DouweM:
Closing pending discussion in https://github.com/pydantic/pydantic-ai/issues/4669.

PR #4746, opened 2026-03-20
fix: preserve ToolReturn metadata through Temporal serialization

guoyangzhen:
## Problem

When a tool returns a `ToolReturn` with `metadata` or `content`, these fields are **silently dropped** after Temporal activity serialization. `ToolReturnPart.metadata` is always `None` in `TemporalAgent` runs, while it works correctly in non-Temporal runs.

## Root Cause

1. `_ToolReturn` dataclass only captures the raw result, not `metadata` or `content`
2. When `ToolReturn` is serialized/deserialized through Temporal, it becomes a plain dict
3. `isinstance(tool_result, ToolReturn)` returns `False` for the deserialized dict
4. The else branch constructs a generic `ToolReturnPart` **without metadata**

## Fix

1. Add `content` and `metadata` fields to `_ToolReturn` dataclass
2. In `_wrap_call_tool_result`: decompose `ToolReturn` into fields so they survive serialization
3. In `_unwrap_call_tool_result`: reconstruct `ToolReturn` when metadata or content is present

The fix ensures that `ToolReturn.metadata` and `ToolReturn.content` survive the Temporal serialization round-trip, matching the behavior of non-Temporal agent runs.

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_toolset.py
+++ b/pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_toolset.py
@@ -12,7 +12,7 @@ from typing_extensions import Self, assert_never
 
 from pydantic_ai import AbstractToolset, FunctionToolset, ToolsetTool, WrapperToolset
 from pydantic_ai.exceptions import ApprovalRequired, CallDeferred, ModelRetry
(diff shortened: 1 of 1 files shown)


adtyavrdhn:
Is this an automated PR?

guoyangzhen:

PR #4754, opened 2026-03-20
update exa tools description

theishangoswami:
Updated the Exa tools module docstring and search tool description to better reflect what Exa offers.


adtyavrdhn:
Hey @theishangoswami 

The change calls it the best but there is no change to the part of what it offers. That is more suited for your docs :)

PR #4813, opened 2026-03-24
Fix collapsible indentation in graph persistence example

petyosi:
## Summary

- The code blocks, annotation list, and surrounding text inside the `ai_q_and_a_graph.py` collapsible example (`docs/graph.md`) were at column 0 instead of 4-space indented under the `???` marker
- MkDocs collapsible syntax requires body content to be indented by 4 spaces — without it, the content renders outside the fold
- The indentation was likely lost in commit 883e1ea2 ("update logos, favicon and brand names")

Found while building the unified docs site, where the transform that converts `???` blocks faithfully enforces the indentation requirement.


Viicos:
Handled in https://github.com/pydantic/pydantic-ai/pull/4814.

PR #4823, opened 2026-03-24
Fix Google thought signature replay for tool calls

h0rv:
- Closes #<issue>

Fix Google tool-call replay when Gemini/Vertex streams a `thought_signature` in an empty thinking part immediately before a function call.

Without this, the replayed tool call can be sent back without the required signature and fail with:

```text
google.genai.errors.ClientError: 400 Bad Request
Unable to submit request because function call `<tool_name>` in the 4th content block is missing a `thought_signature`.
```

This change:
- preserves the signature for the following tool call,
- applies the same handling in streaming and non-streaming paths,
- adds regression tests for both cases.

### Pre-Review Checklist

### Pre-Merge Checklist

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/google.py
+++ b/pydantic_ai_slim/pydantic_ai/models/google.py
@@ -933,6 +933,34 @@ class GoogleModel(Model):
         return response_schema
 
 
+def _resolve_thought_signature(part: Part, pending: str | None) -> tuple[dict[str, Any] | None, str | None, bool]:
+    """Resolve the thought_signature for a Google API response part.
+
+    When Gemini streams an empty thinking part carrying only a thought_signature before a
+    function call, the signature must be forwarded to the next real part so it survives replay.
+
+    Per https://ai.google.dev/gemini-api/docs/function-calling?example=meeting#thought-signatures:
+    - Always send the thought_signature back to the model inside its original Part.
+    - Don't merge a Part containing a signature with one that does not.
+    - Don't combine two Parts that both contain signatures.
+
+    Returns:
+        (provider_details, updated_pending, skip_part)
+    """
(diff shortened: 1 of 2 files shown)


pydantic_ai_slim/pydantic_ai/models/google.py
```diff
@@ -933,6 +933,34 @@ def _map_response_schema(self, o: OutputObjectDefinition) -> dict[str, Any]:
...
 
+def _resolve_thought_signature(part: Part, pending: str | None) -> tuple[dict[str, Any] | None, str | None, bool]:
+    """Resolve the thought_signature for a Google API response part.
+
+    When Gemini streams an empty thinking part carrying only a thought_signature before a
+    function call, the signature must be forwarded to the next real part so it survives replay.
+
+    Per https://ai.google.dev/gemini-api/docs/function-calling?example=meeting#thought-signatures:
+    - Always send the thought_signature back to the model inside its original Part.
+    - Don't merge a Part containing a signature with one that does not.
```
DouweM, line 944: @h0rv The fix seems to break these rules, but from what you're describing, it seems like it's actually the Gemini/Vertex API response (and error when you follow the rule and send it back verbatim) that doesn't match the documented behavior! 

So it sounds like if this fix is necessary, it would be necessary in all Gemini API clients that follow these rules to the letter, and really it'd be a bug in the API worth reporting to Google and fixing on their side. (Which I can do, we have a channel with them)

So I'd love to understand better when you've seen this happen:

- Has it been Gemini, Vertex, both?
- What model(s)?
- Can you show an exact API response (streaming or not) that shows the issue? Ideally with some kind of response ID I can share with Google?

I'm hesitant to merge this fix without understanding the conditions under which you've found it to be needed.

DouweM:
@h0rv Let me know if this is still relevant.

Issue #4708, opened 2026-03-17
OpenRouterReasoning missing latest OpenRouter effort levels


majdalsado:
### Initial Checks

- [x] I'm using the [latest version](https://github.com/pydantic/pydantic-ai/releases/latest) of Pydantic AI
- [x] I've searched for my issue in [the issue tracker](https://github.com/pydantic/pydantic-ai/issues) before opening this issue

### Description

OpenRouterReasoning.effort is missing three valid values: xhigh, minimal, and none. The current type is Literal['high', 'medium', 'low'], but the [OpenRouter API docs](https://openrouter.ai/docs/api/api-reference/chat/send-chat-completion-request#request.body.reasoning) show six valid values: xhigh, high, medium, low, minimal, and none.

This isn't a critical bug as the code runs fine at runtime, only throwing errors for typecheckers.

### Minimal, Reproducible Example

```Python
from pydantic_ai.models.openrouter import OpenRouterModelSettings

# All three produce pyright errors:
#   Type "Literal['xhigh']" is not assignable to type "Literal['high', 'medium', 'low']"
settings: OpenRouterModelSettings = {'openrouter_reasoning': {'effort': 'xhigh'}}

settings2: OpenRouterModelSettings = {'openrouter_reasoning': {'effort': 'minimal'}}

settings3: OpenRouterModelSettings = {'openrouter_reasoning': {'effort': 'none'}}
```

### Python, Pydantic AI & LLM client version

- Python: 3.13
- Pydantic AI: 1.69.0
- LLM provider SDK: n/a

JasonCZMeng:
Opened a fix in #4742 — one-line `Literal` type expansion to include `'xhigh'`, `'minimal'`, and `'none'`.

PR #4829, opened 2026-03-24
Strip thinking from model_settings after resolution, unify _translate_thinking naming

sarth6:
Closes #4829

## Summary

- Fix correctness bugs in unified thinking translation across providers:
  - Bedrock: move `_translate_thinking()` call outside `if model_settings:` guard (empty dict after stripping is falsy)
  - Cerebras: add `thinking=True` branch that sets `disable_reasoning=False`
  - Groq: restructure if-chain to check `thinking is False` first → return `'hidden'`
- Strip `thinking` from `model_settings` after resolution in `prepare_request()` so downstream providers don't see it
- Rename `_get_thinking_param` / `_get_reasoning_effort` → `_translate_thinking` for consistency across all providers
- Extract xAI inline effort map to module-level `XAI_EFFORT_MAP` with tightened `Literal['low', 'high']` type
- Update docs/thinking.md provider mapping tables for accuracy
- Add comprehensive test suite (`tests/test_thinking.py`) covering:
  - Base thinking resolution via `prepare_request()`
  - Per-provider translation behavior (Anthropic, OpenAI Chat/Responses, Google, Groq, Bedrock, OpenRouter, Cerebras, xAI)
  - Thinking capability integration and serialization
  - Profile detection for thinking-capable models
  - No-mutation and no-leakage regression tests
  - Merge model settings with thinking fields (in `test_settings.py`)

## Test plan

- [x] `pytest tests/test_thinking.py tests/test_settings.py` — all tests pass
- [x] `ruff check` passes on modified files
- [x] CI lint + typecheck + full test suite

### DIFF
--- a/tests/models/test_anthropic.py
+++ b/tests/models/test_anthropic.py
@@ -3011,6 +3011,89 @@ async def test_anthropic_opus_46_features(
     assert any(isinstance(p, TextPart) for p in response.parts)
 
 
+@pytest.mark.parametrize(
+    'thinking_value,expected_thinking,expected_effort',
+    [
+        pytest.param(True, {'type': 'adaptive'}, None, id='true-adaptive'),
+        pytest.param('high', {'type': 'adaptive'}, 'high', id='high-adaptive-with-effort'),
+        pytest.param('low', {'type': 'adaptive'}, 'low', id='low-adaptive-with-effort'),
+        pytest.param('medium', {'type': 'adaptive'}, 'medium', id='medium-adaptive-with-effort'),
+        pytest.param('xhigh', {'type': 'adaptive'}, 'max', id='xhigh-maps-to-max'),
+    ],
+)
+async def test_anthropic_unified_thinking_adaptive_model(
+    allow_model_requests: None,
+    thinking_value: bool | str,
+    expected_thinking: dict[str, str],
+    expected_effort: str | None,
+):
+    """Verify that unified thinking on adaptive models sends {type: 'adaptive'} + output_config effort."""
+    from anthropic._types import omit as OMIT
+
+    responses = [
+        completion_message(
+            [BetaTextBlock(text='4', type='text')],
+            usage=BetaUsage(input_tokens=10, output_tokens=1),
+        ),
+    ]
+    mock_client = MockAnthropic.create_mock(responses)
+    m = AnthropicModel('claude-opus-4-6', provider=AnthropicProvider(anthropic_client=mock_client))
+    agent = Agent(m, model_settings=ModelSettings(thinking=thinking_value))  # pyright: ignore[reportArgumentType]
+
+    await agent.run('What is 2+2?')
+
+    kwargs = get_mock_chat_completion_kwargs(mock_client)[0]
+    assert kwargs['thinking'] == expected_thinking
+
+    if expected_effort is not None:
+        assert kwargs['output_config'] == {'effort': expected_effort}
+    else:
+        assert kwargs.get('output_config') is None or kwargs['output_config'] is OMIT
+
+
+async def test_anthropic_unified_thinking_non_adaptive_model(allow_model_requests: None):
+    """Verify that unified thinking on non-adaptive models sends {type: 'enabled', budget_tokens: N}."""
+    from anthropic._types import omit as OMIT
+
+    responses = [
+        completion_message(
+            [BetaTextBlock(text='4', type='text')],
+            usage=BetaUsage(input_tokens=10, output_tokens=1),
+        ),
+    ]
+    mock_client = MockAnthropic.create_mock(responses)
+    m = AnthropicModel('claude-sonnet-4-5', provider=AnthropicProvider(anthropic_client=mock_client))
+    agent = Agent(m, model_settings=ModelSettings(thinking='high'))
+
+    await agent.run('What is 2+2?')
+
+    kwargs = get_mock_chat_completion_kwargs(mock_client)[0]
+    assert kwargs['thinking'] == {'type': 'enabled', 'budget_tokens': 16384}
+    # Non-adaptive models don't support effort
+    assert kwargs.get('output_config') is None or kwargs['output_config'] is OMIT
+
+
+async def test_anthropic_unified_thinking_false_omits_param(allow_model_requests: None):
+    """Verify that thinking=False does not send a thinking parameter at all."""
+    from anthropic._types import omit as OMIT
+
+    responses = [
+        completion_message(
+            [BetaTextBlock(text='4', type='text')],
+            usage=BetaUsage(input_tokens=10, output_tokens=1),
+        ),
+    ]
+    mock_client = MockAnthropic.create_mock(responses)
+    m = AnthropicModel('claude-opus-4-6', provider=AnthropicProvider(anthropic_client=mock_client))
+    agent = Agent(m, model_settings=ModelSettings(thinking=False))
+
+    await agent.run('What is 2+2?')
+
+    kwargs = get_mock_chat_completion_kwargs(mock_client)[0]
+    # thinking=False on always-thinking model is silently ignored — thinking param is OMIT
+    assert kwargs.get('thinking') is OMIT
+
+
 async def test_anthropic_opus_46_adaptive_thinking_rejects_tool_output(allow_model_requests: None):
     """Verified in https://logfire-us.pydantic.dev/public-trace/ca9932da-b5ff-46f0-b277-9aeecc5f41e7?spanId=15a32e26f5020e62"""
     responses = [


sarth6:
@DouweM I put together a plan here on areas we could potentially improve the thinking capability, let me know what you think, or we can go over it on thursday! unless you think it's not high priority

PLAN.md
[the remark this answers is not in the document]
sarth6: Phase 2 has some critical items (in my opinion) like cohere having limited support.

Phase 3 and 4 have some items I discussed with Douwe offline that he could be interested in.

DouweM:
Related issue (for human and AI reference) about OpenRouter supported thinking efforts: https://github.com/pydantic/pydantic-ai/issues/4708

PLAN.md
```diff
@@ -0,0 +1,283 @@
...
+**File:** `pydantic_ai_slim/pydantic_ai/profiles/anthropic.py`
+
+Change `ANTHROPIC_THINKING_BUDGET_MAP` so that for adaptive-capable models:
+- `True` → `{'type': 'adaptive'}`
+- `'medium'` → `{'type': 'adaptive'}`
+- `'low'` → `budget_tokens: 1024` (see 1.2)
+- `'high'` → `budget_tokens: 16384` (unchanged)
+
+The translation method needs to check whether the model supports adaptive thinking and
+use the appropriate mapping. For budget-only models, keep existing fixed-budget behavior.
```
DouweM: Good point; right now if the model supports adaptive thinking it looks like we ignore the effort level entirely.

PLAN.md
```diff
@@ -0,0 +1,283 @@
...
+
+### 1.2 Anthropic: `'low'` should use 1024 (documented minimum)
+
+**File:** `pydantic_ai_slim/pydantic_ai/profiles/anthropic.py`
+
+Change `'low'` mapping from `2048` to `1024`. Anthropic's SDK docs state
+`budget_tokens >= 1024`. Since `'low'` means "as little thinking as possible,"
+the floor is the correct semantic match.
```
DouweM: The `minimal` effort already maps to 1024 so I don't think this change is necessary.

PLAN.md
```diff
@@ -0,0 +1,283 @@
...
+
+### 1.3 xAI: Invert effort mapping (round down, not up)
+
+**File:** xAI model/helper where thinking translation lives
+
+xAI Grok supports only `'low'` and `'high'`. Current mapping rounds UP for ambiguous
+values. Fix to round DOWN (conservative = cheaper, safer):
+- `True` → `'low'` (match xAI SDK default)
+- `'low'` → `'low'`
+- `'medium'` → `'low'` (round down)
```
DouweM: I don't feel strongly about this one, do you? As long as we document what it means, I think either is fine.

PLAN.md
```diff
@@ -0,0 +1,283 @@
...
+- `'high'` → `'high'`
+
+### 1.4 Groq: `thinking=False` should return `'hidden'`
+
+**File:** `pydantic_ai_slim/pydantic_ai/models/groq.py` (lines 245-256)
+
+Groq's SDK accepts `'hidden'`, `'raw'`, `'parsed'` — no true disable exists. But
+`'hidden'` is the closest to "disabled" semantics (hides reasoning output). Change:
+- `thinking=False` → `'hidden'` (instead of `NOT_GIVEN`)
```
DouweM: We need to document this clearly, because the user may think they're actually disabling thinking, but I agree it's better to at least not generate `ThinkingPart`s if we can accomplish that using `hidden`

PLAN.md
```diff
@@ -0,0 +1,283 @@
...
+
+Add a code comment explaining the SDK limitation. Also add a note in `docs/thinking.md`
+in the Groq section.
+
+### 1.5 Cerebras: Handle `thinking=True` (not just `False`)
+
+**File:** Cerebras model/helper (`_cerebras_settings_to_openai_settings()`)
+
+Currently only maps `thinking=False` → `disable_reasoning: True`. Add the reverse:
+- `thinking=True` or effort level → `disable_reasoning: False`
```
DouweM: Seems reasonable, though I imagine `disable_reasoning: False` is already the default

PLAN.md
```diff
@@ -0,0 +1,283 @@
...
+
+---
+
+## Phase 2: Missing Implementations (silent no-ops)
+
+These providers have profile flags set but their model classes never read
+`model_request_parameters.thinking`. The unified setting is silently ignored.
+
+### 2.1 Cohere: Implement `_translate_thinking()` or remove flags
```
DouweM: Good one, let's do it, but in a separate PR rather than together with the fixes.

PLAN.md
```diff
@@ -0,0 +1,283 @@
...
+
+Research how Cohere's API controls thinking for Command R+ models. Implement
+`_translate_thinking()` that reads `model_request_parameters.thinking` and maps it
+to Cohere's API parameter.
+
+**If Cohere has no API-level thinking control**: Remove `supports_thinking=True` and
+`thinking_always_enabled=True` from the profile, and add a comment explaining why.
+Models that always think with no API knob don't benefit from profile flags.
+
+### 2.2 Mistral: Implement `_translate_thinking()` or remove flags
```
DouweM: Same as for Cohere

PLAN.md
```diff
@@ -0,0 +1,283 @@
...
+**Files:**
+- `profiles/meta.py` — Llama-4 thinking models: `supports_thinking=True`, `thinking_tags`
+- `profiles/qwen.py` — QwQ models: `supports_thinking=True`, `thinking_always_enabled=True`, `thinking_tags`
+- `profiles/moonshotai.py` — `kimi-thinking-*`: `supports_thinking=True`
+- `profiles/harmony.py` — `gpt-oss-120b`: `supports_thinking=True`
+- `profiles/zai.py` — `zai-glm-4.6`: `supports_thinking=True`
+
+These enable cascading to 15+ hosting providers (Ollama, Fireworks, Together, etc.).
+
+Research each model family to determine correct flag values (especially
```
DouweM: Agreed, worth doing, in a separate PR

PLAN.md
```diff
@@ -0,0 +1,283 @@
...
+
+**File:** `pydantic_ai_slim/pydantic_ai/models/__init__.py` (in `prepare_request()`)
+
+After copying `model_settings['thinking']` to `model_request_parameters.thinking`,
+pop it from the dict:
+```python
+model_settings.pop('thinking', None)
+```
+
+Prevents downstream provider code from accidentally reading the raw unresolved value.
```
DouweM: Agreed

PLAN.md
```diff
@@ -0,0 +1,283 @@
...
+@property
+def thinking_mode(self) -> Literal['unsupported', 'optional', 'required']:
+    if self.thinking_always_enabled:
+        return 'required'
+    if self.supports_thinking:
+        return 'optional'
+    return 'unsupported'
+```
+
+Update `prepare_request()` to use it instead of inline boolean checks.
```
DouweM: I'm not against this, but the benefit's not super clear to me either.

PLAN.md
```diff
@@ -0,0 +1,283 @@
...
+
+### 3.3 Rename Google profile field
+
+**File:** `GoogleModelProfile` definition + Google model class
+
+Change `google_supports_thinking_level: bool` to
+`google_thinking_api: Literal['budget', 'level'] | None`.
+
+Update all usages in the Google model class and profile factory.
```
DouweM: I don't think this is worth doing, especially as it'd be backward incompatible.

PLAN.md
```diff
@@ -0,0 +1,283 @@
...
+**Files:** All provider model classes with thinking translation methods
+
+Rename for consistency and grep-ability:
+- Anthropic: `_get_thinking_param()` → `_translate_thinking()`
+- OpenAI: `_get_reasoning_effort()` → `_translate_thinking()`
+- Groq: `_get_reasoning_format()` → `_translate_thinking()`
+- Google: `_get_thinking_config()` → `_translate_thinking()`
+- Bedrock: `_get_thinking_fields()` → `_translate_thinking()`
+
+Return types differ per provider — the convention is about naming, not signatures.
```
DouweM: Fine by me

PLAN.md
```diff
@@ -0,0 +1,283 @@
...
+
+---
+
+## Phase 4: Tests
+
+**File:** `tests/test_unified_thinking.py` (new)
+
+Every behavior from Phases 1-3 should have a corresponding test.
```
DouweM: Agreed

PLAN.md
```diff
@@ -0,0 +1,283 @@
...
+### 4.3 Provider-specific settings precedence
+
+For each provider with both unified and provider-specific settings:
+- `anthropic_thinking` set + `thinking` set → `anthropic_thinking` wins
+- `openai_reasoning_effort` set + `thinking` set → `openai_reasoning_effort` wins
+- etc.
+
+### 4.4 Per-provider translation tests
+
+One test class per provider. Each tests the full `ThinkingLevel` → provider parameter
```
DouweM: Agreed

PLAN.md
DouweM: By the way Claude pointed out that the effort levels _are_ respected for Anthropic, it's just in the `_build_output_config` method

DouweM:
Didn't mean to close this! I did create some conflicts in https://github.com/pydantic/pydantic-ai/pull/4846 though 😅

sarth6:
@DouweM ready for review! yeah you're right - anthropic adaptive ended up being the only true bug, most of everything else is improving typing/testing, and nitpicking to adhere to the plan (like renaming to `_translate_thinking`)

sarth6:
Also fyi the CI failure appears to be unrelated

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -775,15 +775,14 @@ def prepare_request(
...
+            thinking_value = model_settings['thinking']
             if self.profile.supports_thinking or self.profile.thinking_always_enabled:
-                if thinking_value is False and self.profile.thinking_always_enabled:
-                    pass  # Silent ignore: model always thinks, can't disable
-                else:
+                if not (thinking_value is False and self.profile.thinking_always_enabled):
                     params = replace(params, thinking=thinking_value)
-            # else: silent ignore for unsupported models
+            stripped = {k: v for k, v in model_settings.items() if k != 'thinking'}
+            model_settings = cast(ModelSettings, stripped) if stripped else None
```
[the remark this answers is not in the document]
sarth6: The approach was discussed in the call (came from the transcript) - prevents downstream model code from accidentally reading the raw unresolved value

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -648,7 +648,7 @@ async def request(
         model_response = self._process_response(response)
         return model_response
 
-    def _get_reasoning_effort(
+    def _translate_thinking(
```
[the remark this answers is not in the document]
sarth6: Douwe already confirmed he was fine with the rename: https://github.com/pydantic/pydantic-ai/pull/4829#discussion_r2989745749

tests/test_thinking.py
```diff
@@ -154,7 +154,7 @@ def test_thinking_true_adaptive(self, adaptive_model: FunctionModel):
 
         params = ModelRequestParameters(thinking=True)
         settings: ModelSettings = {}
-        result = AnthropicModel._get_thinking_param(adaptive_model, settings, params)
+        result = AnthropicModel._translate_thinking(adaptive_model, settings, params)
```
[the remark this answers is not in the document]
sarth6: @DouweM what do you think?

pydantic_ai_slim/pydantic_ai/models/xai.py
```diff
@@ -55,6 +55,16 @@
...
 
+XAI_EFFORT_MAP: dict[ThinkingLevel, Literal['low', 'high']] = {
+    True: 'high',
+    'minimal': 'low',
+    'low': 'low',
+    'medium': 'high',
+    'high': 'high',
+    'xhigh': 'high',
+}
+"""Maps unified thinking values to xAI reasoning_effort. xAI only supports 'low' and 'high'."""
```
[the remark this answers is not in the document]
sarth6: Sounds good will adjust, but I'll keep my `Literal['low', 'high']` change as it's a typing improvement that alligns to the underlying xai sdk

PLAN.md
sarth6: Missed this one, yeah I will revert - ideally we map medium to the most "reasonable" default, but that's unclear here - we would need to see data on xai agent thinking usage to really know the answer to that.

pydantic_ai_slim/pydantic_ai/models/xai.py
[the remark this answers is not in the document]
DouweM: Constant are great for things like this, I'd actually rather have all the models use constants instead of inline dicts for this.

pydantic_ai_slim/pydantic_ai/models/__init__.py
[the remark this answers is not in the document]
DouweM: Makes sense to me.

tests/test_thinking.py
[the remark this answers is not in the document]
DouweM: I'm OK with the private method usage in this particular case.

pydantic_ai_slim/pydantic_ai/models/openai.py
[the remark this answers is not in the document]
DouweM: I like it better this way, so it's easier to find the equivalent logic when looking in different model implementations.

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -426,7 +426,7 @@ def _get_thinking_param(
         if thinking is None or thinking is False:
             return OMIT  # type: ignore[return-value]
         profile = AnthropicModelProfile.from_profile(self.profile)
-        if profile.anthropic_supports_adaptive_thinking:
+        if profile.anthropic_supports_adaptive_thinking and thinking in (True, 'medium'):
```
DouweM: @sarth6 Are you sure of this change?

Claude is not:

> The current logic only uses adaptive for True and 'medium', but for any specific effort level like 'high', 'low', 'xhigh', it falls back to the deprecated enabled + budget_tokens path. That seems wrong — on adaptive models,
  all effort levels should use {type: 'adaptive'} and let output_config.effort control the depth. The _build_output_config already handles mapping the effort level correctly.

```
❯ Let's discuss the Anthropic change. Now for a value like `high`, we'll set thinking={type=enabled, budget_tokens=...} AND we set the output effort. Is that correct? Check the Ant docs on reasoning, and how budget tokens and
effort are meant to be used

● Let me read the relevant Anthropic model code and check the docs.
  ⎿  Loaded pydantic_ai_slim/pydantic_ai/CLAUDE.md
  ⎿  Loaded pydantic_ai_slim/pydantic_ai/models/CLAUDE.md

  Searched for 1 pattern, read 1 file (ctrl+o to expand)

● Agent(Check Anthropic docs on thinking)
  ⎿  Done (8 tool uses · 31.1k tokens · 1m 12s)
  (ctrl+o to expand)

● This is revealing. So for thinking='high' on an adaptive model (Sonnet 4.6, Opus 4.6), the current PR code does:

  1. _translate_thinking (line 429): 'high' is not in (True, 'medium'), so it falls through to {'type': 'enabled', 'budget_tokens': ...} — the deprecated path for 4.6 models
  2. _build_output_config (line 1282): maps 'high' → effort: 'high' in the output config

  So we're sending both thinking={type: enabled, budget_tokens=...} and output_config={effort: high}. According to the Anthropic docs:

  - For 4.6 models, {type: 'enabled', budget_tokens: N} is deprecated — the recommended approach is {type: 'adaptive'} + output_config: {effort: ...}
  - Effort controls thinking depth on adaptive models; budget_tokens is a separate hard cap that will be removed
  - Adaptive + effort is the intended combo: effort=high means "always thinks, deep reasoning"

  The current logic only uses adaptive for True and 'medium', but for any specific effort level like 'high', 'low', 'xhigh', it falls back to the deprecated enabled + budget_tokens path. That seems wrong — on adaptive models,
  all effort levels should use {type: 'adaptive'} and let output_config.effort control the depth. The _build_output_config already handles mapping the effort level correctly.

  Let me verify what the code looked like on main before this PR.

● Bash(git diff main -- pydantic_ai_slim/pydantic_ai/models/anthropic.py | head -120)
  ⎿  diff --git a/pydantic_ai_slim/pydantic_ai/models/anthropic.py b/pydantic_ai_slim/pydantic_ai/models/anthropic.py
     index 0c12ebbd..6dbca677 100644
     --- a/pydantic_ai_slim/pydantic_ai/models/anthropic.py
     … +37 lines (ctrl+o to expand)

● So the key behavioral change in this PR is on line 429. On main, it was:

  if profile.anthropic_supports_adaptive_thinking:
      return {'type': 'adaptive'}

  This PR changed it to:

  if profile.anthropic_supports_adaptive_thinking and thinking in (True, 'medium'):
      return {'type': 'adaptive'}

  On main, for thinking='high' on an adaptive model, you'd get {type: 'adaptive'} — and _build_output_config would set effort: 'high'. That's the correct combination per the Anthropic docs.

  With this PR, for thinking='high' on an adaptive model, you get {type: 'enabled', budget_tokens: ...} + effort: 'high' — using the deprecated thinking type.

  The main behavior was actually correct. This PR introduced a regression by narrowing the adaptive check. The intent was probably to ensure specific effort levels get respected, but _build_output_config already handles that
  via the effort mapping — so _translate_thinking should return {type: 'adaptive'} for all truthy thinking values on adaptive models, as main does.

  Want me to revert this specific change (line 429) back to the main behavior?
```

sarth6:
@DouweM thanks for merging - was just confirming what your claude said via a local vcr test and you were right, my bad! I'll try to get in the habit of doing more vcr validation before trusting the ai

PR #4807, opened 2026-03-24
Add capability and toolset override by ID

DouweM:
## Summary
- Add optional `id` field to `AbstractCapability` so capabilities can be identified for selective override
- Add `visit_and_replace` traversal pattern to `AbstractCapability` and `CombinedCapability` (matching the existing toolset pattern)
- Add `capabilities` parameter to `Agent.override()` accepting either a `Sequence` (replace all) or `Mapping[str, T | None]` (replace/remove by ID)
- Expand `toolsets` parameter in `Agent.override()` to also accept `Mapping[str, T | None]` for selective override
- Forward-compatible with future `WrapperCapability`

## Test plan
- [x] Existing 218 capability tests pass
- [x] Existing 216 agent tests + 250 agent+toolset tests pass
- [x] New tests: capability `id` construction (default, kwarg, init=False capabilities)
- [x] New tests: `visit_and_replace` (leaf, combined, nested)
- [x] New tests: `override(capabilities=[...])` sequence replaces all
- [x] New tests: `override(capabilities={'id': cap})` dict replaces by ID
- [x] New tests: `override(capabilities={'id': None})` dict removes by ID
- [x] New tests: `override(toolsets={'id': ts})` dict replaces by ID
- [x] New tests: `override(toolsets={'id': None})` dict removes by ID
- [x] Pre-commit hooks pass (lint, format, typecheck)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4832, opened 2026-03-24
Improve capabilities & hooks docs; promote `TemplateStr` to public `pydantic_ai.template`

dsfaccini:
## Summary

- Expand `docs/capabilities.md` (~+168 lines): provider-adaptive tools table, expanded examples, clearer narrative on the native/local model
- Expand `docs/hooks.md` (~+121 lines): hook ordering contract, `wrap_*` semantics, error-hook chaining caveats
- Promote `_template.py` → new public module `pydantic_ai.template` with `TemplateStr` re-exported from `pydantic_ai`; add `docs/api/template.md` to the nav
- Small xref / docstring cleanups in `agent/*.py`, `mcp.py`, `capabilities/abstract.py`, `docs/{agent-spec,dependencies,extensibility,api/agent,api/models/base,api/toolsets}.md`

## Scope notes for reviewers

- This PR began as a "bugs in capabilities" exploration but has converged to docs + the `template` public surface only; no behavioral changes
- Original DOCS-QUESTIONS / NICHE-DOCS scratch files have been removed; the unresolved threads on them are obsolete and resolved
- Merged `upstream/main` 2026-05-31 to pull in the `XSearch` `fallback_model` wording — table updated to include the new row

- [ ] AI generated code


tests/test_capabilities_bugs.py
```diff
@@ -0,0 +1,804 @@
...
+    await agent.run('hello')
+
+    before_order = [x for x in order if 'before' in x]
+    after_order = [x for x in order if 'after' in x]
+
+    assert before_order == ['h1:before', 'h2:before']
+
+    # BUG: after fires forward (h1 first), should fire reversed (h2 first)
+    assert after_order == ['h1:after', 'h2:after']
+    # EXPECTED: ['h2:after', 'h1:after']
```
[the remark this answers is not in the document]
dsfaccini: we can switch to `pytest.xfail` if that's preferable @DouweM

docs/hooks.md
```diff
@@ -110,7 +110,7 @@ Node hooks fire for each graph step ([`UserPromptNode`][pydantic_ai.UserPromptNo
 | `model_request` | `model_request=` | `wrap_model_request` |
 | `model_request_error` | `model_request_error=` | `on_model_request_error` |
 
-Model request hooks fire around each LLM call. [`ModelRequestContext`][pydantic_ai.models.ModelRequestContext] bundles `messages`, `model_settings`, and `model_request_parameters`.
+Model request hooks fire around each LLM call. `ModelRequestContext` bundles `messages`, `model_settings`, and `model_request_parameters`.
```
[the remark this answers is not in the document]
dsfaccini: they're being rendered badly on https://ai.pydantic.dev/hooks/#model-request-hooks, may be a missing reference in the API docs, will check if we can reintroduce properly
image

docs/capabilities.md
```diff
@@ -353,11 +353,11 @@ The callable receives a [`RunContext`][pydantic_ai.tools.RunContext] where `ctx.
...
-| [`get_instructions()`][pydantic_ai.capabilities.AbstractCapability.get_instructions] | [`AgentInstructions`][pydantic_ai._instructions.AgentInstructions] ` \| None` | [Instructions](agent.md#instructions) (static strings, [template strings](agent-spec.md#template-strings), or callables) |
-| [`get_model_settings()`][pydantic_ai.capabilities.AbstractCapability.get_model_settings] | [`AgentModelSettings`][pydantic_ai.agent.abstract.AgentModelSettings] ` \| None` | [Model settings](agent.md#model-run-settings) dict, or a callable for per-step settings |
+| [`get_wrapper_toolset()`][pydantic_ai.capabilities.AbstractCapability.get_wrapper_toolset] | [`AbstractToolset`][pydantic_ai.toolsets.AbstractToolset] `\| None` | [Wrap the agent's assembled toolset](#toolset-wrapping) |
+| [`get_instructions()`][pydantic_ai.capabilities.AbstractCapability.get_instructions] | `AgentInstructions \| None` | [Instructions](agent.md#instructions) (static strings, [template strings](agent-spec.md#template-strings), or callables) |
```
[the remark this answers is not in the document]
DouweM: Let's export `AgentInstructions` from a public module so we can link it here

docs/hooks.md
```diff
@@ -96,7 +205,7 @@ Run hooks fire once per agent run. `wrap_run` (registered via `hooks.on.run`) wr
 | `node_run` | `node_run=` | `wrap_node_run` |
 | `node_run_error` | `node_run_error=` | `on_node_run_error` |
 
-Node hooks fire for each graph step ([`UserPromptNode`][pydantic_ai.UserPromptNode], [`ModelRequestNode`][pydantic_ai.ModelRequestNode], [`CallToolsNode`][pydantic_ai.CallToolsNode]).
+Node hooks fire for each graph step (`UserPromptNode`, `ModelRequestNode`, `CallToolsNode`).
```
DouweM: We can fix these links right?

docs/hooks.md
```diff
@@ -110,7 +219,7 @@ Node hooks fire for each graph step ([`UserPromptNode`][pydantic_ai.UserPromptNo
 | `model_request` | `model_request=` | `wrap_model_request` |
 | `model_request_error` | `model_request_error=` | `on_model_request_error` |
 
-Model request hooks fire around each LLM call. [`ModelRequestContext`][pydantic_ai.models.ModelRequestContext] bundles `messages`, `model_settings`, and `model_request_parameters`.
+Model request hooks fire around each LLM call. `ModelRequestContext` bundles `messages`, `model_settings`, and `model_request_parameters`.
```
DouweM: Fix this link please

pydantic_ai_slim/pydantic_ai/capabilities/hooks.py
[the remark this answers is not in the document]
DouweM: This is intentional; I'll comment below.

tests/test_capabilities_bugs.py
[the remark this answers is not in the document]
DouweM: This file will be deleted before merge.

tests/test_capabilities_bugs.py
```diff
@@ -0,0 +1,804 @@
...
+# DOCUMENTED CONTRACT (docs/hooks.md lines 287-290):
+#   "When multiple hooks are registered for the same event (either on
+#    the same Hooks instance or across multiple capabilities):
+#    * before_* hooks fire in registration/capability order
+#    * after_* hooks fire in reverse order"
+#
+# The docs explicitly say this applies to a SINGLE Hooks instance too.
+# CombinedCapability implements this correctly (reversed after_*), and
+# wrap_* within Hooks also reverses. But after_* within a single Hooks
+# instance iterates FORWARD, violating the documented contract.
```
DouweM: This is intentional for hooks registered using decorators on `Hooks`, as it matches code definition order. Capabilities are different because of middleware rules. The docs should be updated to match this.

tests/test_capabilities_bugs.py
```diff
@@ -0,0 +1,804 @@
...
+#   1. Calls the factory to get a new toolset
+#   2. Exits the old toolset (__aexit__)
+#   3. Assigns the new toolset
+#   4. Enters the new toolset (__aenter__)
+#
+# If step 1 or 4 raises, there's no rollback. Two failure modes:
+#
+# (a) Factory raises: exception propagates, old toolset is still
+#     referenced but the DynamicToolset is in an ambiguous state.
+#     Not catastrophic, but the contract is unclear.
```
DouweM: If the factory or aenter raises, that will bubble up to user code, so it's OK for the toolset to be broken, as it can't be used anymore anyway. It does probably make sense to exit the old toolset before calling the factory to get the new toolset.

tests/test_capabilities_bugs.py
```diff
@@ -0,0 +1,804 @@
...
+# ======================================================================
+#
+# When multiple on_*_error handlers are registered (either on one Hooks
+# instance or across capabilities), they fire in sequence. If handler A
+# raises a NEW exception, handler B receives handler A's exception —
+# NOT the original error that triggered the chain.
+#
+# This means the original error context is silently lost. If handler A
+# has a bug (accidentally raises), handler B cannot even see what the
+# real problem was.
```
DouweM: This is expected to me, similar to how if one capability's `before_model_request` hook modifies the model request context, later `before_model_request` hooks will only get to see the modified context, not the original. If one on_error handler wraps an error in some new type (or raises an error), it makes sense to me taht the next handler will see _that_ error.

If the first handler has a bug and raises an unintentional error, that should be fixed in that handler.

tests/test_capabilities_bugs.py
```diff
@@ -0,0 +1,804 @@
...
+# (identity check: `run_cap is not effective_cap`), the agent re-extracts
+# all values from the new object.
+#
+# But if for_run() mutates `self` and returns `self`, the identity check
+# says "same object, no change" and the agent uses the stale cached values.
+# The capability's internal state was changed, but the agent never notices.
+#
+# This is a contract violation trap: a capability developer might reasonably
+# expect that mutating self in for_run() is sufficient, but the identity
+# check silently ignores the mutation.
```
DouweM: > "a capability developer might reasonably expect that mutating self in for_run() is sufficient"

Not really, because the point of this hook is to not share the same instance/state across runs, so it should always return `self` or a new instance, not modify `self` (because then you don't get state isolation). That may need to be stated more clearly in docs/docstring,.

tests/test_capabilities_bugs.py
```diff
@@ -0,0 +1,804 @@
...
+# retry count: retries[tool_name] += 1 for each tool that failed.
+#
+# If a DynamicToolset (per_run_step=True) swaps the tool implementation
+# between steps but keeps the same name, the new implementation inherits
+# the accumulated retry count from the old implementation. A brand-new
+# tool could be born with N retries already counted against it, making
+# it hit max_retries faster than expected.
+#
+# Whether this is a "bug" or a design tradeoff depends on perspective.
+# The ToolManager has no way to know the underlying tool changed.
```
DouweM: This is expected behavior: if a tool's implementation changes, but it keeps the same name, it is still considered the same tool.

I don't think this is worth documenting, it doesn't seem unexpected/surprising to me. The opposite would be more surprising, if using a per-run-step dynamic toolset basically broke retry counting entirely.

tests/test_capabilities_bugs.py
```diff
@@ -0,0 +1,804 @@
...
+# Multiple processors compose in sequence (processor 2 sees processor 1's
+# output). There's no validation that the resulting messages are
+# semantically consistent.
+#
+# In the agent message protocol, a ToolCallPart (model asks to call a tool)
+# appears in a ModelResponse, and the corresponding ToolReturnPart (tool
+# result) appears in the next ModelRequest. If a processor removes the
+# ModelResponse containing the ToolCallPart but leaves the ModelRequest
+# containing the ToolReturnPart, the model sees a tool return with no
+# preceding tool call — which is semantically invalid.
```
DouweM: > If a processor removes the
> ModelResponse containing the ToolCallPart but leaves the ModelRequest
> containing the ToolReturnPart,

That'd be a bug in that processor. I agree we could do some semantic validity validation in between before_model_request hooks, but that doesn't seem urgent, and us not doing that currently is not itself a bug.

DOCS-QUESTIONS.md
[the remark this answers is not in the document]
DouweM: Yep, these should be deleted before merge.

DOCS-QUESTIONS.md
```diff
@@ -0,0 +1,63 @@
+# Capabilities Docs — Uncovered User Questions
+
+Questions users are likely to have that the current `docs/capabilities.md` doesn't cover (or barely touches).
+
+## 1. What's the generic type parameter on `AbstractCapability[T]`?
+
+Every example uses `AbstractCapability[Any]`. Users will wonder: what is `T`? How does it relate to the agent's dependency type? Can a capability declare that it needs specific dependencies, and what happens if the agent's dep type doesn't match?
```
DouweM: Good point, we should make it clear it's generic in the agent's dependency type. `AbstractToolset` is as well, not sure if it's explicitly documented there.

DOCS-QUESTIONS.md
```diff
@@ -0,0 +1,63 @@
...
+
+## 2. Can capabilities be added per-run?
+
+The docs only show capabilities at `Agent(...)` construction time. Users will want to know: can I pass extra capabilities to `agent.run()` or `agent.iter()`? Or override/disable specific ones for a single run?
```
DouweM: Good point, capabilities can't currently be passed into agent.run or agent.override directly, although it can be done indirectly via `spec=AgentSpec(capabilities=...))`. In https://github.com/pydantic/pydantic-ai/pull/4807 I started working on letting them be overridden/disabled one by one, and I can directly support `capabilities=...` on run and override there as well.

DOCS-QUESTIONS.md
```diff
@@ -0,0 +1,63 @@
...
+
+## 3. `BuiltinTool`, `Toolset`, and `HistoryProcessor` — how to use them?
+
+These three are listed in the built-in table but have zero documentation beyond the one-line row. A user reading top-to-bottom will see `PrepareTools` and `PrefixTools` get dedicated sections but these don't.
```
DouweM: Yeah we should show examples, although the API docs should make it pretty clear how to use them.

In https://github.com/pydantic/pydantic-ai/pull/4828 I'm also making the `BuiltinOrLocalTool` capability public.

DOCS-QUESTIONS.md
```diff
@@ -0,0 +1,63 @@
...
+
+## 4. `ImageGeneration` local fallback
+
+`WebSearch` auto-detects DuckDuckGo, `MCP` auto-detects transport — but what does `ImageGeneration(local=...)` expect? There's no example of providing a local image generation fallback.
```
DouweM: Any user tool that the agent should use instead of builtin image generation :) That could be a tool that directly uses an image generation model. Would be worth mentioning explicitly.

DOCS-QUESTIONS.md
```diff
@@ -0,0 +1,63 @@
...
+
+## 5. When to use a capability vs direct agent parameters
+
+The intro paragraph hints at this ('instructions and model settings are configured directly...'), but there's no decision framework. When should I use `Agent(instructions=...)` vs a capability that provides instructions? When should I register tools with `@agent.tool` vs bundling them in a capability?
```
DouweM: Capabilities are for reusability.

DOCS-QUESTIONS.md
```diff
@@ -0,0 +1,63 @@
...
+
+## 6. Ordering guidance for composition
+
+The composition section explains *mechanics* (before fires in order, after in reverse, wrap nests) but gives no *practical guidance*. If I have a guardrail and a logger, which should come first? What's the mental model for deciding order?
```
DouweM: Seems use case dependent, something the user can decide themselves.

DOCS-QUESTIONS.md
```diff
@@ -0,0 +1,63 @@
...
+
+## 7. How does `for_run` interact with `get_toolset`?
+
+`get_toolset` is called at agent construction, `for_run` is called per-run. If my capability returns a toolset *and* overrides `for_run`, does the toolset come from the original instance or the fresh one? This matters a lot for stateful toolsets.
```
DouweM: Good point. The intended behavior would be that `for_run` is called first, and then that new instance's `get_toolset` is used. Same for `get_instructions` etc; they should run on the new capability instance. Please verify if that's how it currently works, and let me know if it's not!

DOCS-QUESTIONS.md
```diff
@@ -0,0 +1,63 @@
...
+
+## 8. Concurrency / thread safety
+
+If the same agent is used for concurrent runs (common in web servers), what happens to shared capability instances? `for_run` helps, but the docs don't mention concurrency at all. Users need to know if `for_run` is required or just recommended for concurrent usage.
```
DouweM: Hmm it still comes down to whether the capability needs per-run state isolation right? Concurrency doesn't seem like it should _directly_ require `for_run`, if your hooks etc are all self contained and don't use instance vars

DOCS-QUESTIONS.md
```diff
@@ -0,0 +1,63 @@
...
+
+## 9. Can capabilities interact with each other?
+
+No mention of inter-capability communication. If capability A needs to read state from capability B (e.g., a cost tracker that a budget limiter reads), how do they coordinate? Through dependencies? Through some other mechanism?
```
DouweM: Good point, worth documenting something. 

The easiest way is to create a contextvar and set it around the run (using before/after or wrap run hooks), and then other capabilities can read it. That means that the order of capabilities in the `capabilities` list matters: the contextvar writer should come before the contextvar reader.

I'm not against adding explicit dependencies between capabilities.

Through dependencies can make sense in your own codebase if you can control the deps type, but not for capabilities that are meant to be used with arbitrary agents / deps.

DOCS-QUESTIONS.md
```diff
@@ -0,0 +1,63 @@
...
+
+## 10. Skip exceptions and hook chain behavior
+
+`SkipModelRequest`, `SkipToolValidation`, and `SkipToolExecution` are mentioned one sentence each. Users will want to know: when a skip is raised from `before_*`, do other capabilities' `before_*` hooks still fire? Do `after_*` hooks fire with the skip result? What about `wrap_*`?
```
DouweM: Good point, should be mentioned explicitly. Claude can probably find out (from code or testing it out) what the current behavior is :)

DOCS-QUESTIONS.md
```diff
@@ -0,0 +1,63 @@
...
+
+## 11. Error hook ordering with multiple capabilities
+
+When multiple capabilities define `on_*_error`, which fires first? If one recovers (returns a result), do the others still see the error? Or does recovery short-circuit?
```
DouweM: Good point, worth documenting explicitly. I'm not sure if `on_*_error` is fired in capability order, or reverse order like `after_*` hooks are. Reverse order _probably_ makes most sense since they're the failure counterpart to after hooks, but Claude would need to verify.

I believe that recovery should result in other capabilities getting `after_*` called, instead of `on_*_error`, so that the recovery is "complete".

DOCS-QUESTIONS.md
```diff
@@ -0,0 +1,63 @@
...
+
+## 12. Message history interaction
+
+If `message_history` is passed to `agent.run()`, do `before_model_request` hooks see the historical messages in `request_context.messages`? Can they modify them?
```
DouweM: yes and yes. this should be expected behavior to anyone who's dealt with message histories / processors before, but may be worth stating explicitly.

DOCS-QUESTIONS.md
```diff
@@ -0,0 +1,63 @@
...
+
+## 13. Testing custom capabilities
+
+No guidance on testing. Users building non-trivial capabilities (guardrails, approval workflows) will want patterns for: testing that hooks fire in the right order, testing error recovery, testing with `TestModel`, etc.
```
DouweM: Some of this sounds like testing internals, which they shouldn't need to do. It's worth mentioning that you can test your capabilities using an `Agent` with `TestModel` or `FunctionModel`, but that's not different from testing agents in general.

NICHE-DOCS.md
```diff
@@ -0,0 +1,49 @@
...
+**Caveat:** If a `per_run_step=True` factory raises mid-run, the DynamicToolset
+doesn't roll back — the previous toolset stays referenced but the overall state
+is ambiguous. Worse: if the factory succeeds but the new toolset's `__aenter__`
+fails, the old toolset has already been exited and the new one never entered.
+`self._toolset` then points to an un-initialized toolset.
+
+**User guidance:** Factory functions used with `per_run_step=True` should be
+defensive — catch and handle their own errors rather than letting exceptions
+propagate into the toolset lifecycle. If the factory can fail, consider
+returning the previous toolset as a fallback instead of raising.
```
DouweM: I don't think we should fall back on factory failure; that'd be a source of bugs. Raising and letting the factory author fix it seems much better.

NICHE-DOCS.md
```diff
@@ -0,0 +1,49 @@
...
+
+**Caveat:** `ToolManager` tracks tool retries by **name** across run steps. If a
+`DynamicToolset` with `per_run_step=True` replaces a tool's implementation
+between steps but keeps the same name, the new implementation inherits the
+accumulated retry count from the old one. A fresh tool can be born with N
+retries already counted against it, hitting `max_retries` sooner than expected.
+
+**User guidance:** If tool implementations change between steps, use distinct
+tool names to avoid inheriting retry state. Alternatively, design tools with
+retry budgets that account for the possibility of inherited counts.
```
DouweM: I'm OK documenting this explicitly, but as mentioned behavior it is intended/expected behavior.

NICHE-DOCS.md
```diff
@@ -0,0 +1,49 @@
...
+
+**Caveat:** When multiple history processors are registered, they compose in
+sequence — processor 2 sees processor 1's output. There's no validation that
+the resulting messages are semantically consistent. A processor that removes
+`ModelResponse` messages containing `ToolCallPart` but leaves the subsequent
+`ModelRequest` with `ToolReturnPart` creates orphaned tool returns — the model
+sees a tool result with no preceding tool call.
+
+**User guidance:** When writing history processors that remove messages, ensure
+tool calls and their corresponding tool returns are removed together. A safe
```
DouweM: I've commented on this above.

DouweM (changes requested):

DOCS-QUESTIONS.md
dsfaccini: super, leaving the comment open as a reminder but not doing anything about it here

pydantic_ai_slim/pydantic_ai/toolsets/_dynamic.py
```diff
@@ -75,18 +75,20 @@ async def for_run(self, ctx: RunContext[AgentDepsT]) -> AbstractToolset[AgentDep
...
+        if old_toolset is not None:
+            await old_toolset.__aexit__(None, None, None)
 
-        # Manage the transition in-place
-        if self._toolset is not None:
-            await self._toolset.__aexit__(None, None, None)
+        new_toolset = await self._evaluate_factory(ctx)
         self._toolset = new_toolset
         if self._toolset is not None:
             await self._toolset.__aenter__()
```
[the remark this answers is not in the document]
DouweM: Devin found the same on my PR :D https://github.com/pydantic/pydantic-ai/pull/4846#discussion_r2991175490

DouweM:
@dsfaccini Didn't mean to close this; you'll just want to fix merge conflicts / see what I already fixed that's not needed here anymore.

docs/api/agent_spec.md
```diff
@@ -0,0 +1,11 @@
...
+
+::: pydantic_ai.agent.spec
+    options:
+        members:
+            - AgentSpec
+
+::: pydantic_ai._template
+    options:
+        members:
+            - TemplateStr
```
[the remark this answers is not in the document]
DouweM: Good point Devin

docs/agent-spec.md
```diff
@@ -71,15 +71,15 @@ For more control over spec loading, use [`AgentSpec.from_file`][pydantic_ai.agen
 
 ## Template strings
 
-[`TemplateStr`][pydantic_ai.TemplateStr] provides Handlebars-style templates (`{{variable}}`) that are rendered against the agent's [dependencies](dependencies.md) at runtime. In spec files, strings containing `{{` are automatically converted to template strings:
+[`TemplateStr`][pydantic_ai._template.TemplateStr] provides Handlebars-style templates (`{{variable}}`) that are rendered against the agent's [dependencies](dependencies.md) at runtime. In spec files, strings containing `{{` are automatically converted to template strings:
```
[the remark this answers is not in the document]
DouweM: @dsfaccini This needs fixing (and verifying the link works) -- there's at least one more docstring in the changes below that also links to `_template`. I think that should just be a public module.

docs/agent-spec.md
```diff
@@ -117,7 +117,7 @@ See [Publishing capabilities](capabilities.md#publishing-capabilities) for how t
 
 ## `AgentSpec` reference
 
-The [`AgentSpec`][pydantic_ai.agent.spec.AgentSpec] model represents the full spec structure:
+The [`AgentSpec`](#agentspec-reference) model represents the full spec structure:
```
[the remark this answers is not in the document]
DouweM: This now links to this same section lol. Should link to API.

docs/api/agent_spec.md
```diff
@@ -0,0 +1,11 @@
+# `pydantic_ai.agent.spec`
+
+::: pydantic_ai.agent.spec
+    options:
+        members:
+            - AgentSpec
```
DouweM: I wouldn't mind this one being documented on `api/agent.md`

docs/api/agent_spec.md
```diff
@@ -0,0 +1,11 @@
...
+
+::: pydantic_ai._template
```
DouweM: @dsfaccini You know this is wrong :) Please review your own PRs!

docs/capabilities.md
DouweM, line 1: @dsfaccini Weren't there a ton of docs gaps that were going to be addressed as part of this PR?

DouweM (changes requested):

dsfaccini:
Claude here: `TemplateStr` is exported from `pydantic_ai` (via the private `pydantic_ai._template` module) but has no API documentation page — there's no `::: pydantic_ai._template` entry in `docs/api/`. The xref `[pydantic_ai.TemplateStr]` was already broken in upstream main for the same reason. Keeping as plain backticks until an API page is set up.

Options we considered but didn't pursue:
- Add a `docs/api/pydantic_ai.md` page — would be the first top-level `pydantic_ai` API page, feels heavy for one type
- Add a `:::` directive for `pydantic_ai._template` on `docs/api/agent.md` — contextually fits (TemplateStr is part of the instructions type system), but documents a private module
- Add `TemplateStr` to `pydantic_ai.agent.__all__` and the `docs/api/agent.md` members list — cleanest, but changes the public API surface of `pydantic_ai.agent`

pydantic_ai_slim/pydantic_ai/capabilities/abstract.py
```diff
@@ -113,7 +113,7 @@ def get_instructions(self) -> AgentInstructions[AgentDepsT] | None:
         This method is called once at agent construction time. To get dynamic
         per-request behavior, return a callable that receives
         [`RunContext`][pydantic_ai.tools.RunContext] or a
-        [`TemplateStr`][pydantic_ai.TemplateStr] — not a dynamic string.
+        `TemplateStr` — not a dynamic string.
```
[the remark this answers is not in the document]
dsfaccini: can't do this until we've solved the question of where to link `TemplateStr` from (see my comment on this PR)

pydantic_ai_slim/pydantic_ai/template.py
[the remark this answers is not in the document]
DouweM: `validate_from_spec_args` can be private right?

DouweM, line 1: *(Refers to line 127)*

Let's make this method private

docs/api/template.md
```diff
@@ -0,0 +1,6 @@
+# `pydantic_ai.template`
+
+::: pydantic_ai.template
+    options:
+        members:
+            - TemplateStr
```
DouweM: I don't think the `options` business is required; by default it should expose everything that has a docstring, I believe. You'll want to build the docs and verify.

tests/test_template.py
```diff
@@ -12,11 +12,11 @@
 
 from pydantic_ai import Agent, TemplateStr
 from pydantic_ai._run_context import RunContext
-from pydantic_ai._template import validate_from_spec_args
 from pydantic_ai.capabilities.abstract import AbstractCapability
 from pydantic_ai.messages import ModelRequest
 from pydantic_ai.models.test import TestModel
 from pydantic_ai.result import RunUsage
+from pydantic_ai.template import validate_from_spec_args as _validate_from_spec_args
```
DouweM: Unnecessary rename here

docs/capabilities.md
```diff
@@ -87,7 +87,7 @@ See the dedicated [Hooks](hooks.md) page for the full API: decorator and constru
 
 [`WebSearch`][pydantic_ai.capabilities.WebSearch], [`WebFetch`][pydantic_ai.capabilities.WebFetch], [`ImageGeneration`][pydantic_ai.capabilities.ImageGeneration], and [`MCP`][pydantic_ai.capabilities.MCP] provide model-agnostic access to common tool types. When the model supports the tool natively (as a [builtin tool](builtin-tools.md)), it's used directly. When it doesn't, a local function tool handles it instead — so your agent works across providers without code changes.
 
-Each accepts `builtin` and `local` keyword arguments to control which side is used:
+[`WebSearch`][pydantic_ai.capabilities.WebSearch] auto-detects DuckDuckGo as a local fallback. [`ImageGeneration`][pydantic_ai.capabilities.ImageGeneration] requires you to provide a local fallback via `local=` — this should be a tool function that calls an image generation API (e.g. DALL-E) directly. Each accepts `builtin` and `local` keyword arguments to control which side is used:
```
DouweM: I'd rather have a table with all 4 capabilities showing the local behavior / requirement to provide your own. (Note that we're getting a local WebFetch in https://github.com/pydantic/pydantic-ai/pull/4906.)

DouweM: > tool function that calls an image generation API (e.g. DALL-E) directly

This is too strongly stated. The tool could also call a Pydantic AI subagent with a hard-coded model that is known to support image gen; it doesn't need to use an API directly.

That's actually a pretty nice solution, I'll get a PR up for that as the default "local" solution (with a user-configurable model name)

docs/capabilities.md
```diff
@@ -191,8 +191,72 @@ Every [`AbstractCapability`][pydantic_ai.capabilities.AbstractCapability] has a
 MCP(url='https://mcp.example.com/api').prefix_tools('mcp')
 ```
 
+### Convenience wrappers
```
DouweM: This section title is pretty useless in the sidebar :) I'd rather have a section for each type

docs/capabilities.md
```diff
@@ -191,8 +191,72 @@ Every [`AbstractCapability`][pydantic_ai.capabilities.AbstractCapability] has a
...
+
+[`BuiltinTool`][pydantic_ai.capabilities.BuiltinTool], [`Toolset`][pydantic_ai.capabilities.Toolset], and [`HistoryProcessor`][pydantic_ai.capabilities.HistoryProcessor] wrap a single tool, toolset, or history processor as a capability, so they can be composed alongside other capabilities:
```
DouweM: Worth specifying that this is mostly meant as an alternative for existing reusable components that haven't been repackaged as capabilities yet.

docs/capabilities.md
```diff
@@ -191,8 +191,72 @@ Every [`AbstractCapability`][pydantic_ai.capabilities.AbstractCapability] has a
...
+    return messages[-10:]  # keep only last 10 messages
+
+history_cap = HistoryProcessor(trim_history)
+
+agent = Agent('openai:gpt-5.2', capabilities=[code_exec, toolset_cap, history_cap])
+```
+
 ## Building custom capabilities
 
+[`AbstractCapability`][pydantic_ai.capabilities.AbstractCapability] is generic in the agent's dependency type — `AbstractCapability[MyDeps]` means the capability's hooks will receive a `RunContext[MyDeps]`. Use `AbstractCapability[Any]` when the capability works with any dependency type, or a specific type when it needs to access dependency fields:
```
DouweM, line 866: It's weird for this to come before we even introduce "To build your own capability, subclass [`AbstractCapability`][pydantic_ai.capabilities.AbstractCapability]". Pleaase think about where this should go.

docs/capabilities.md
```diff
@@ -721,6 +792,8 @@ print(counter.count)
 #> 0
 ```
 
+When `for_run` returns a new instance, all `get_*()` methods (`get_toolset`, `get_instructions`, `get_model_settings`, etc.) are re-called on that new instance — not on the original. This is important for capabilities where per-run configuration depends on per-run state.
```
DouweM, line 1548: Claude suggestion:

```
  4. for_run: no warning against mutating self (docs/capabilities.md) — The reviewer said mutating self in for_run and returning self is not the intended
  pattern, and "may need to be stated more clearly in docs/docstring." The current docs show the correct pattern (return a fresh instance) but don't
  explicitly warn that mutating self and returning self will use stale cached values.
```

docs/hooks.md
```diff
@@ -72,6 +72,115 @@ print(result.output)
 
 Both sync and async hook functions are accepted. Sync functions are automatically wrapped for async execution.
 
+## Lifecycle overview
```
DouweM, line 112: Claude suggestion:

```
  2. Hooks ordering docs still inaccurate (docs/hooks.md:389-392) — The reviewer said Bug 1 (after_* firing forward on a single Hooks instance) is
  intentional and "the docs should be updated to match this." Currently the docs still say "after_* hooks fire in reverse order" for "the same Hooks
  instance or across multiple capabilities" — but the reviewer clarified that within a single Hooks instance, forward order is intentional (matching code
  definition order). Only across capabilities should after_* fire in reverse. This distinction is not reflected.
```

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/template.py
dsfaccini: is this regarding `validate_from_spec_args`? the reason I didn't make it private is because pyright flags it as unused (because it's not used in its own module and pyright doesn't assume it to be used elsewhere when it's private), so instead I went with the `import as`

dsfaccini: discussion landed in having separate `template` and `_template` modules

docs/capabilities.md
dsfaccini: I'm gonna add a TODO comment referencing that PR

docs/capabilities.md
```diff
@@ -981,6 +1038,60 @@ assert combined.capabilities[0] is logging_hooks
...
+from pydantic_ai import Agent
+from pydantic_ai.capabilities import HistoryProcessor
+from pydantic_ai.messages import ModelMessage
+
+
+def trim_history(messages: list[ModelMessage]) -> list[ModelMessage]:
+    return messages[-10:]  # keep only last 10 messages
+
+agent = Agent('openai:gpt-5.2', capabilities=[HistoryProcessor(trim_history)])
+```
```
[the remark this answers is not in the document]
dsfaccini: Claude here: applied. The "Wrapping existing components as capabilities" section now uses `ProcessHistory` (heading, intro at line ~1141, and the `process_history_wrapper.py` example) — see commit 43f5fd6.

docs/capabilities.md
```diff
@@ -981,6 +1038,60 @@ assert combined.capabilities[0] is logging_hooks
...
+
+Capabilities don't have direct access to each other. To share state between capabilities during a run, use a [`contextvars.ContextVar`](https://docs.python.org/3/library/contextvars.html): one capability sets it (e.g. in `wrap_run` or `before_run`), and another reads it from its hooks. The order of capabilities in the `capabilities` list matters — the writer must come before the reader so its `before_*` hook runs first.
+
+### Testing custom capabilities
+
+Test custom capabilities the same way you test agents — using [`TestModel`][pydantic_ai.models.test.TestModel] or [`FunctionModel`][pydantic_ai.models.function.FunctionModel]. Create an agent with your capability and assert on the run result, messages, or any observable side effects of your hooks.
+
+## Wrapping existing components as capabilities
+
+[`BuiltinTool`][pydantic_ai.capabilities.BuiltinTool], [`Toolset`][pydantic_ai.capabilities.Toolset], and [`HistoryProcessor`][pydantic_ai.capabilities.HistoryProcessor] wrap existing reusable components that haven't been repackaged as capabilities yet, so they can be composed alongside other capabilities in the `capabilities` list.
```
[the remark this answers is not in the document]
dsfaccini: Claude here: applied. Intro paragraph now references `ProcessHistory`. Same commit (43f5fd6).

docs/capabilities.md
```diff
@@ -373,6 +380,44 @@ The [UI adapters](ui/ag-ui.md) (AG-UI, Vercel AI) automatically add this capabil
...
+class PersonalGreeter(AbstractCapability[Deps]):
+    """Requires Deps with a user_name field."""
+
+    async def before_model_request(
+        self, ctx: RunContext[Deps], request_context: ModelRequestContext
+    ) -> ModelRequestContext:
+        # ctx.deps is typed as Deps — IDE autocomplete works
+        print(f'Request for {ctx.deps.user_name}')
+        return request_context
+```
```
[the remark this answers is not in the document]
dsfaccini: Claude here: skip is intentional. `typed_capability.py` is purely a type-illustration showing how `AbstractCapability[Deps]` types `RunContext[Deps]` — the classes are defined to demonstrate the type signatures, never instantiated or run, so there's no observable behavior to test. Adding an Agent run would distract from the typing focus. Leaving `test="skip" lint="skip"` as-is.

docs/capabilities.md
```diff
@@ -119,6 +119,13 @@ See the dedicated [Hooks](hooks.md) page for the full API: decorator and constru
 
 [`WebSearch`][pydantic_ai.capabilities.WebSearch], [`WebFetch`][pydantic_ai.capabilities.WebFetch], [`ImageGeneration`][pydantic_ai.capabilities.ImageGeneration], and [`MCP`][pydantic_ai.capabilities.MCP] provide model-agnostic access to common tool types. When the model supports the tool natively (as a [builtin tool](builtin-tools.md)), it's used directly. When it doesn't, a local function tool handles it instead — so your agent works across providers without code changes.
 
+| Capability | Local fallback | Notes |
+|---|---|---|
+| [`WebSearch`][pydantic_ai.capabilities.WebSearch] | Auto-detected (DuckDuckGo) | Requires the `duckduckgo` optional group (`pip install "pydantic-ai-slim[duckduckgo]"`) |
+| [`WebFetch`][pydantic_ai.capabilities.WebFetch] | Provide your own via `local=` | Default local fallback [coming soon](https://github.com/pydantic/pydantic-ai/pull/4906) |
+| [`ImageGeneration`][pydantic_ai.capabilities.ImageGeneration] | Provide your own via `local=` | e.g. a tool that calls an image generation API, or a subagent with a model that supports image gen |
+| [`MCP`][pydantic_ai.capabilities.MCP] | Direct connection to MCP server | Auto-detects transport from URL |
```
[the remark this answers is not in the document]
dsfaccini: Claude here: #4906 has been merged, so the "coming soon" note was stale. Updated the WebFetch row to "Auto-detected (markdownify)" with a note pointing to the `web-fetch` optional group, mirroring the WebSearch row's format. Same commit (43f5fd6).

pydantic_ai_slim/pydantic_ai/_template.py
```diff
@@ -1,125 +1,14 @@
...
-                info_arg=False,
-            ),
-        )
-
-    def __repr__(self) -> str:
-        return f'TemplateStr({self._source!r})'
-
-    def __str__(self) -> str:
-        return self._source
+from pydantic_ai.template import TemplateStr
```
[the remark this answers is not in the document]
dsfaccini: Claude here: applied. `pydantic_ai/agent/wrapper.py` now imports `TemplateStr` from `..template` like the rest of the codebase — see commit c7648f0. Verified the only remaining `_template` import is `validate_from_spec_args` (a genuine private helper).

dsfaccini:
re-review after we've released v2 stable to drop any references to stuff removed in v2

DouweM:
@dsfaccini close or finish?

PR #4841, opened 2026-03-25
Add `pydantic_ai.prices.update_in_background()`

### LINKED ISSUE #4627: Use genai-prices auto-update to keep model pricing fresh without waiting for releases
## Problem

When new models come out (e.g. gpt-5.4), genai-prices updates its `data.json` on GitHub quickly, but PyPI releases can lag behind. Since pydantic-ai only uses the bundled snapshot from the installed genai-prices wheel, users get missing/zero pricing for new models until a new release is published.

This is a common complaint: https://github.com/pydantic/genai-prices/issues/312

## genai-prices already has a solution

genai-prices ships an `UpdatePrices` class that fetches the latest `data.json` from GitHub in a background daemon thread (~261KB uncompressed, refreshes hourly, 10s timeout, fails silently on network errors). But pydantic-ai never uses it, and most users don't know it exists.

## Proposal

Enable `UpdatePrices` by default in pydantic-ai so pricing data stays fresh without waiting for releases. Provide an opt-out for air-gapped or restricted environments (env var or config flag).

Open to discussion on whether this should be default-on or opt-in. The fetch is ~261KB and runs in a background thread so it won't block anything, but it is an implicit network call on startup that some users may not expect.

adtyavrdhn:
- Closes #4627

## Summary

Adds `pydantic_ai.prices.update_in_background()`: call it once at startup and Pydantic AI downloads the latest model prices right away and every hour after, so models released after your install get a cost without upgrading.

It starts a `genai_prices.UpdatePrices` and returns it, so callers can `wait()` for the first download, `stop()` at shutdown, or use it as a context manager. genai-prices 0.1.6 (now the floor on `main`) keeps one background download per process however many callers there are, so the wrapper holds no state of its own.

Also points `CostNotFoundWarning` and the `cost` docstring at the new function, so a `None` cost for a new model leads users to the fix.

### Pre-Review Checklist

### Pre-Merge Checklist


pydantic_ai_slim/pydantic_ai/prices.py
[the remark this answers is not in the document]
adtyavrdhn: Yeah I'll add a test.

pydantic_ai_slim/pydantic_ai/prices.py
```diff
@@ -0,0 +1,29 @@
+from __future__ import annotations
+
+from genai_prices import UpdatePrices
```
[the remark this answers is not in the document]
adtyavrdhn: It has been present for a while but I'll confirm and resolve.

pydantic_ai_slim/pydantic_ai/prices.py
```diff
@@ -0,0 +1,29 @@
...
+    the data bundled with the installed `genai-prices` package.
+
+    Example:
+    ```python
+    import pydantic_ai
+
+    pydantic_ai.update_prices()
+
+    agent = pydantic_ai.Agent('openai:gpt-4o')
+    ```
```
Kludex: Claude seems very confused about this, but the correct is this indentation:
```suggestion
	    ```python
	    import pydantic_ai
	
	    pydantic_ai.update_prices()
	
	    agent = pydantic_ai.Agent('openai:gpt-4o')
	    ```
```

adtyavrdhn: Not sure I follow

pydantic_ai_slim/pydantic_ai/prices.py
```diff
@@ -0,0 +1,29 @@
...
+    without waiting for a new `genai-prices` PyPI release.
+
+    If the fetch fails (e.g. no network access), pricing silently falls back to
+    the data bundled with the installed `genai-prices` package.
+
+    Example:
+    ```python {test="skip"}
+    import pydantic_ai
+
+    pydantic_ai.update_prices()
```
DouweM: - This looks like it'll be blocking; I suggest making it something like `pydantic_ai.prices.update_in_background()`
- Also needs docs!

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/prices.py
adtyavrdhn: I was confused on where to add the docs, do you have a preference?

pydantic_ai.prices.update_in_background -> will do

pydantic_ai_slim/pydantic_ai/__init__.py
```diff
@@ -295,5 +296,7 @@
     'AgentRun',
     'AgentRunResult',
     'AgentRunResultEvent',
+    # prices
+    'update_in_background',
```
DouweM: This isn't correct is it? We don't want people calling `pydantic_ai.update_in_background`

pydantic_ai_slim/pydantic_ai/prices.py
```diff
@@ -0,0 +1,29 @@
...
+
+    pydantic_ai.prices.update_in_background()
+
+    agent = pydantic_ai.Agent('openai:gpt-5.2')
+    ```
+    """
+    try:
+        UpdatePrices().start()
+    except Exception:
+        pass
```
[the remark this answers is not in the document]
DouweM: Making sure there can be only one updater seems worth doing.

Maybe worth mentioning in the docs that if one wants to manage that thread themselves, they can use `genai_prices.UpdatePrices` directly (or mention that our function wraps that)

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/prices.py
Kludex: Check how it renders on the docs.

pydantic_ai_slim/pydantic_ai/prices.py
```diff
@@ -0,0 +1,39 @@
...
+    prices.update_in_background()
+    ```
+    """
+    global _updater
+    if _updater is not None:
+        return
+    try:
+        _updater = UpdatePrices()
+        _updater.start()
+    except Exception:
```
adtyavrdhn, line 38: logger.warning is needed

docs/agent.md
```diff
@@ -423,6 +423,55 @@ You can retrieve usage statistics (tokens, requests, etc.) at any time from the
...
+```python {test="skip"}
+from pydantic_ai import Agent
+from pydantic_ai.messages import ModelResponse
+
+agent = Agent('openai:gpt-5.2')
+result = agent.run_sync('What is the capital of France?')
+
+for message in result.all_messages():
+    if isinstance(message, ModelResponse):
+        cost = message.cost()
```
adtyavrdhn, line 439: Only last response cost is enough

adtyavrdhn:
Somewhere there is an issue to get AgentRunResult.cost() can be a property.

pydantic_ai_slim/pydantic_ai/prices.py
dsfaccini: David's AICA here: we kept this on `warnings.warn(CostCalculationFailedWarning)` rather than switching to `logging`.

Reasoning: the library emits no log records via the stdlib `logging` module anywhere — the only `import logging` (in `_griffe.py`) just *silences* griffe's logger, it never emits. `warnings.warn` is the house style for advisory messages (~40 call sites), and `CostCalculationFailedWarning` is the dedicated category users filter/catch (now re-exported from `pydantic_ai.models.instrumented` so the path resolves). Moving this one branch to `logging` would also unravel that public re-export, the `filterwarnings` entry, and the tests built around the warning.

The underlying concern — a warning on *every* request — is now addressed by deduping to once per unknown model (`warn_unknown_cost_model` + a process-wide seen-set), so it stays quiet regardless of the active warnings filter.

Happy to revisit if you wanted `logging` for a specific reason (e.g. routing it through the OTel/Logfire pipeline rather than the warnings channel).

zmievsa:
Hey! If any help is necessary to move this forward, feel free to ping.

docs/agent.md
```diff
@@ -581,6 +581,18 @@ _(This example is complete, it can be run "as is")_
...
+
+Price calculations use the data bundled with [`genai-prices`](https://github.com/pydantic/genai-prices). To refresh that data hourly in a background thread, call [`update_in_background()`][pydantic_ai.prices.update_in_background] at application startup:
+
+```python {test="skip"}
+from pydantic_ai import prices
+
+prices.update_in_background()
+```
+
+This helper retains shared ownership of the updater for the process lifetime. Call it after any `os.fork()` because inheriting a running updater in a child process is unsupported. For configuration, waiting, or shutdown, use [`genai_prices.UpdatePrices`](https://github.com/pydantic/genai-prices) directly.
```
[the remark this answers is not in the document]
adtyavrdhn: Fixed in 31abf433c. All genai-prices references now link directly to the UpdatePrices documentation section.

pydantic_ai_slim/pydantic_ai/prices.py
```diff
@@ -0,0 +1,41 @@
...
+_updater: _UpdatePrices | None = None
+_updater_lock = Lock()
+
+
+def update_in_background() -> None:
+    """Start fetching the latest model pricing data in the background.
+
+    The updater is retained as a shared owner for the lifetime of the process, so repeated calls
+    are safe and other libraries can independently acquire compatible ownership.
+    Call this after any `os.fork()`; inheriting a running updater in a child process is unsupported.
```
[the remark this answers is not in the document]
adtyavrdhn: This review checked released genai-prices v0.0.70, but this PR is intentionally blocked on pydantic/genai-prices#404 and will raise both dependency floors to the first release containing that shared-ownership contract. The PR body records that merge order explicitly. The fork text does not promise recovery after inheriting a running updater: it says to call this only after os.fork(), and states that inheriting a running updater is unsupported. No implementation change is needed for this finding.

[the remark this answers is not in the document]
adtyavrdhn: Clarified in 750eca28f. The docs now say not to call the helper before os.fork(), to call it for the first time in each final child process, and that it cannot recover updater state inherited from a parent. We intentionally keep the wrapper thin and do not add independent PID or fork lifecycle machinery.

PR #4358, opened 2026-02-18
`CodeExecutionToolset` with Code Mode support + `ExecutionEnvironmentToolset`, for local env, Monty, Docker, E2B

DouweM:
- Replaces https://github.com/pydantic/pydantic-ai/pull/4339
- Builds on https://github.com/pydantic/pydantic-ai/pull/4153

PR #4153, opened 2026-01-31
CodeExecutionToolset with Code Mode support + ExecutionEnvironmentToolset, for local env, Monty, Docker, E2B

adtyavrdhn:
- Closes https://github.com/pydantic/pydantic-ai/issues/4050

- Uses https://github.com/pydantic/monty

PR #4851, opened 2026-03-25
feat: add Python signature generation for tool definitions

adtyavrdhn:
## Summary

Adds infrastructure to generate Python function signatures from tool definitions, needed for code mode.

- Adds `_python_signature` module with `function_to_signature` and `schema_to_signature` converters
- Adds `_FunctionToolDefinition` internal subclass to retain `original_func` reference for richer type-aware signatures, falling back to schema-based generation when unavailable (e.g. `Tool.from_schema()`, MCP, deserialization)
- Adds cached `python_signature` property on `ToolDefinition`
- Handles `deepcopy` of `tool_def` in prepare callbacks to prevent mutation leaks across runs
- Comprehensive tests covering function signatures, schema signatures, edge cases, and deduplication

Closes #4755

## Test plan

- [x] `tests/test_python_signature.py` — 47 tests covering function-based and schema-based signature generation, edge cases (forward refs, bare generics, Annotated, Literal, recursive models, RootModel), type deduplication, and rendering
- [x] `tests/test_tools.py` — 108 tests including new prepare mutation isolation test
- [x] `tests/test_toolsets.py` — toolset signature access test
- [x] `pyright` — 0 errors on all changed files

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/google.py
+++ b/pydantic_ai_slim/pydantic_ai/models/google.py
@@ -1475,7 +1475,7 @@ def _map_executable_code(executable_code: ExecutableCode, provider_name: str, to
     return BuiltinToolCallPart(
         provider_name=provider_name,
         tool_name=CodeExecutionTool.kind,
-        args=executable_code.model_dump(mode='json'),
+        args=executable_code.model_dump(mode='json', exclude_none=True),
         tool_call_id=tool_call_id,
     )
 
@@ -1486,7 +1486,7 @@ def _map_code_execution_result(
     return BuiltinToolReturnPart(
         provider_name=provider_name,
         tool_name=CodeExecutionTool.kind,
-        content=code_execution_result.model_dump(mode='json'),
+        content=code_execution_result.model_dump(mode='json', exclude_none=True),
         tool_call_id=tool_call_id,
     )
 


pydantic_ai_slim/pydantic_ai/tools.py
```diff
@@ -571,4 +588,61 @@ def defer(self) -> bool:
...
+        The result is cached so that repeated access (e.g. on every agent step)
+        does not re-traverse the schema. A new instance created via
+        `dataclasses.replace()` gets a fresh cache automatically.
+        """
+        return schema_to_signature(
+            name=self.name,
+            parameters_schema=self.parameters_json_schema,
+            description=self.description,
+            return_schema=(self.metadata or {}).get('output_schema'),
+        )
```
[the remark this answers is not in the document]
adtyavrdhn: Resolved — we no longer use `deepcopy(tool_def)`. Both `Tool.prepare_tool_def` and `PreparedToolset.get_tools` now use `dataclasses.replace(tool_def, parameters_json_schema=deepcopy(...))`, which creates a new instance with a fresh `__dict__` — no `cached_property` values are carried over.

pydantic_ai_slim/pydantic_ai/tools.py
```diff
@@ -571,4 +588,56 @@ def defer(self) -> bool:
...
```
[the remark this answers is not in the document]
adtyavrdhn: Resolved — same as above. `deepcopy(tool_def)` replaced with `dataclasses.replace()` + `deepcopy` of only `parameters_json_schema`. `replace()` creates a fresh instance, so no `cached_property` values survive.

pydantic_ai_slim/pydantic_ai/function_signature.py
```diff
@@ -0,0 +1,802 @@
...
+            # Then process the main schema
+            if schema.get('type') == 'object' and 'properties' in schema:
+                _build_and_register_type(type_name, schema, schema_defs, referenced_types, tool_name, path)
+        return
+
+    origin = get_origin(annotation)
+    args = getattr(annotation, '__args__', None)
+    if origin is not None and args:
+        for arg in args:
+            _collect_referenced_types(arg, referenced_types, tool_name, path, mode=mode)
```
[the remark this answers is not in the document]
adtyavrdhn: Resolved — added `isinstance(annotation, types.UnionType)` check in `_collect_referenced_types` (before the `get_origin()` check), mirroring the existing handling in `_annotation_to_type_expr`.

pydantic_ai_slim/pydantic_ai/tools.py
```diff
@@ -474,10 +484,17 @@ async def prepare_tool_def(self, ctx: RunContext[ToolAgentDepsT]) -> ToolDefinit
...
+        # `tool_def.parameters_json_schema['properties']['name']['description'] = 'Name of the human to greet.'`
+        #
+        # Without this copy, that description could accidentally persist into later runs with different deps.
         base_tool_def = self.tool_def
 
         if self.prepare is not None:
-            return await self.prepare(ctx, base_tool_def)
+            return await self.prepare(ctx, deepcopy(base_tool_def))
         else:
             return base_tool_def
```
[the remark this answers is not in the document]
adtyavrdhn: Acknowledged — the `@property` → `@cached_property` change is intentional for performance (avoid re-creating `_FunctionToolDefinition` on every access). As Devin's audit confirms, all downstream consumers use `replace()` not in-place mutation. The prepare paths now use `replace()` + `deepcopy(parameters_json_schema)` which also creates a fresh instance.

pydantic_ai_slim/pydantic_ai/tools.py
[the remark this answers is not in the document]
adtyavrdhn: Resolved — the `__deepcopy__` removal is no longer a concern because we replaced `deepcopy(tool_def)` with `dataclasses.replace(tool_def, parameters_json_schema=deepcopy(...))`. `replace()` creates a new instance with a fresh `__dict__`, so no `cached_property` values (including `python_signature`) are carried over. This also avoids the unpicklable closure issue that motivated removing `__deepcopy__` in the first place.

pydantic_ai_slim/pydantic_ai/tools.py
```diff
@@ -477,7 +487,13 @@ async def prepare_tool_def(self, ctx: RunContext[ToolAgentDepsT]) -> ToolDefinit
         base_tool_def = self.tool_def
 
         if self.prepare is not None:
-            return await self.prepare(ctx, base_tool_def)
+            # `self.tool_def` is cached — give the prepare callback a per-run copy so
+            # in-place mutations (e.g. `tool_def.parameters_json_schema['properties'][...] = ...`)
+            # don't leak into future runs. We use `replace()` + deepcopy of only the mutable
+            # schema dict instead of `deepcopy(tool_def)` because `_FunctionToolDefinition.original_func`
+            # may close over unpicklable objects (e.g. HTTP clients).
+            tool_def_copy = replace(base_tool_def, parameters_json_schema=deepcopy(base_tool_def.parameters_json_schema))
```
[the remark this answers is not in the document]
adtyavrdhn: The `metadata` field is not mutated in place by any prepare callback in the codebase or docs — prepare functions reassign fields (e.g. `tool_def.description = ...`) or use `replace()`. The `parameters_json_schema` is the only field with documented in-place mutation patterns (nested dict access like `tool_def.parameters_json_schema['properties'][...] = ...`). Deepcopying `metadata` preemptively would be defending against a pattern that doesn't exist today.

pydantic_ai_slim/pydantic_ai/tools.py
```diff
@@ -463,6 +472,7 @@ def tool_def(self):
             metadata=self.metadata,
             timeout=self.timeout,
             kind='unapproved' if self.requires_approval else 'function',
+            original_func=self._original_func,
         )
```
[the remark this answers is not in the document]
adtyavrdhn: The `FunctionToolset.add_tool` at line 501 does `tool.metadata = self.metadata | (tool.metadata or {})` — this is a reassignment, not an in-place mutation. Since `tool.metadata` is reassigned before `tool_def` is ever accessed (registration happens at setup time, `cached_property` is first evaluated during agent run), the cached `tool_def` will see the final `self.metadata` value. No ordering issue in practice.

tests/test_function_signature.py
```diff
@@ -0,0 +1,1240 @@
...
+    GenericTypeExpr,
+    TypeFieldSignature,
+    TypeSignature,
+    UnionTypeExpr,
+    _annotation_to_type_expr,  # pyright: ignore[reportPrivateUsage]
+    _collect_referenced_types,  # pyright: ignore[reportPrivateUsage]
+    _get_schema_from_type,  # pyright: ignore[reportPrivateUsage]
+    _get_type_name,  # pyright: ignore[reportPrivateUsage]
+    _schema_allows_null,  # pyright: ignore[reportPrivateUsage]
+    _to_pascal_case,  # pyright: ignore[reportPrivateUsage]
```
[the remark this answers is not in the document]
adtyavrdhn: Acknowledged — these private function tests are intentional unit-level coverage for edge cases that are difficult to trigger through the public API alone (e.g. `_to_pascal_case` digit-prefix handling, `_schema_allows_null` with `anyOf` vs `oneOf`, `_annotation_to_type_expr` with bare generics). The public API functions (`function_to_signature`, `schema_to_signature`) are also thoroughly tested. We can refactor these to go through public APIs if maintainers prefer, but keeping targeted unit tests for tricky internal logic is a pragmatic choice here.

DouweM:
## Context for reviewers

This commit addresses review feedback and cleans up the PR. Key context that reviewers should know:

**There IS a consumer for this API.** Code mode (`CodeExecutionToolset`) is implemented in [pydantic/pydantic-harness#98](https://github.com/pydantic/pydantic-harness/pull/98) and was previously in this repo ([#4153](https://github.com/pydantic/pydantic-ai/pull/4153), [#4358](https://github.com/pydantic/pydantic-ai/pull/4358)). It consumes `function_signature` on `ToolDefinition` to present tools as Python function signatures to the LLM. The harness PR explicitly lists this as its #1 blocking dependency.

**Changes in this commit:**

1. **Renamed `python_signature` → `function_signature`**: The `FunctionSignature` data model is language-agnostic (contains name, params, types, etc.) and could render to TypeScript or other languages. The property and module names now reflect this.

2. **Removed `copy_for_prepare()` + reverted `tool_def` to `@property`**: The mutation isolation concern (prepare callbacks leaking in-place mutations) is pre-existing on `main` and orthogonal to signature generation. `tool_def` returns a fresh instance each access, matching `main`'s behavior. If mutation isolation matters, it should be its own focused PR.

3. **Fixed `_original_func` on `Tool`**: Removed the `field()` declaration that broke the `function_schema` docstring in mkdocstrings. It's now a plain attribute assigned in `__init__` / `from_schema`, consistent with how `_FunctionToolDefinition` handles it.

4. **Fixed `_FunctionToolDefinition.__repr__`**: Replaced buggy custom `__repr__` (didn't handle `default_factory`) with parent's `dataclasses_no_defaults_repr`.

5. **Removed mutation isolation tests**: These tested pre-existing `main` behavior, not functionality introduced by this PR.

pydantic_ai_slim/pydantic_ai/function_signature.py
```diff
@@ -0,0 +1,817 @@
...
+        for name, f in self.fields.items():
+            other_f = other.fields[name]
+            if f.required != other_f.required:
+                return False
+            if render_type_expr(f.type) != render_type_expr(other_f.type):
+                return False
+        return True
+
+
+TypeExpr: TypeAlias = 'TypeSignature | str | GenericTypeExpr | UnionTypeExpr'
```
DouweM: Does this need to be a string?

pydantic_ai_slim/pydantic_ai/function_signature.py
```diff
@@ -0,0 +1,817 @@
...
+    name: str
+    """The function name."""
+
+    params: dict[str, FunctionParam]
+    """Function parameters."""
+
+    return_type: TypeExpr
+    """The return type expression."""
+
+    docstring: str | None = None
```
DouweM: Should this be `description`? Ideally `FunctionSignature` etc would generally represent the JSON schema / Python functions in a way that could also be rendered to e.g. TypeScript, where a description would become a comment most likely, not a docstring.

pydantic_ai_slim/pydantic_ai/function_signature.py
```diff
@@ -0,0 +1,817 @@
...
+    """Optional docstring for the function."""
+
+    referenced_types: list[TypeSignature] = field(default_factory=list[TypeSignature])
+    """TypedDict class definitions needed by the signature."""
+
+    is_async: bool = True
```
DouweM: We always want to _render_ as `async def` because that's how Monty / code mode calls the functions, but that seems more like a rendering concern than a "definition of the function/signature" concern. So I could be OK with making this a flag on `render`, and having the value here not default to `True` but take the real value of the function being inspected.

pydantic_ai_slim/pydantic_ai/function_signature.py
```diff
@@ -0,0 +1,817 @@
...
+            if render_type_expr(f.type) != render_type_expr(other_f.type):
+                return False
+        return True
+
+
+TypeExpr: TypeAlias = 'TypeSignature | str | GenericTypeExpr | UnionTypeExpr'
+"""A type expression that can reference `TypeSignature` objects, enabling automatic name propagation."""
+
+
+def render_type_expr(expr: TypeExpr) -> str:
```
DouweM: Should this actually be public? Unsure.

pydantic_ai_slim/pydantic_ai/function_signature.py
```diff
@@ -0,0 +1,817 @@
...
+
+def _collect_referenced_types(
+    annotation: Any,
+    referenced_types: dict[str, TypeSignature],
+    tool_name: str,
+    path: str = '',
+    *,
+    mode: Literal['validation', 'serialization'] = 'validation',
+) -> None:
+    """Recursively collect TypeSignature definitions from type annotations."""
```
DouweM: It's not clear from the docstring when/how this is meant to be used. Make docstrings more clear.

pydantic_ai_slim/pydantic_ai/function_signature.py
```diff
@@ -0,0 +1,817 @@
...
+
+
+# =============================================================================
+# Function signature builder (from Python functions)
+# =============================================================================
+
+
+def _collect_referenced_types(
+    annotation: Any,
+    referenced_types: dict[str, TypeSignature],
```
DouweM, line 448: Is it really best to write into this arg, instead of returning a new/updated dict?

pydantic_ai_slim/pydantic_ai/function_signature.py
```diff
@@ -0,0 +1,817 @@
...
+            params[param_name] = FunctionParam(name=param_name, type=type_expr, default=None)
+        else:
+            default_str = repr(param.default)
+            params[param_name] = FunctionParam(name=param_name, type=type_expr, default=default_str)
+    return params
+
+
+def function_to_signature(
+    func: Callable[..., Any],
+    name: str | None = None,
```
DouweM: Kwargs please for non "implied by function name" leading args

pydantic_ai_slim/pydantic_ai/function_signature.py
```diff
@@ -0,0 +1,817 @@
...
+    description: str | None = None,
+) -> FunctionSignature:
```
DouweM: What do you think about `FunctionSignature.from_function`, and possibly `FunctionSignature.from_schema`? I like that better than free functions.

pydantic_ai_slim/pydantic_ai/function_signature.py
```diff
@@ -0,0 +1,817 @@
...
+    tool_name: str,
+) -> None:
+    """Process $defs from a JSON schema, populating referenced_types with TypeSignatures."""
+    for def_name, def_schema in defs.items():
+        if def_schema.get('type') == 'object' and 'properties' in def_schema:
+            if def_name not in referenced_types:
+                _build_and_register_type(def_name, def_schema, defs, referenced_types, tool_name, def_name)
+
+
+def schema_to_signature(
```
DouweM: See above, I like `FunctionSignature.from_schema`, and we should use kwargs

pydantic_ai_slim/pydantic_ai/function_signature.py
```diff
@@ -0,0 +1,817 @@
...
+    resolved_return_type: TypeExpr = return_type
+    return_referenced: dict[str, TypeSignature] = {}
+    if return_schema is not None and return_type == 'Any':
+        return_defs = return_schema.get('$defs', {})
+        _process_schema_defs(return_defs, return_referenced, name)
+        resolved_return_type = schema_to_type_expr(return_schema, return_defs, return_referenced, name, 'Return')
+
+    # Handle case where return type couldn't be resolved
+    final_description = description
+    if return_schema is not None and resolved_return_type == 'Any':
```
DouweM: Why would this happen?

pydantic_ai_slim/pydantic_ai/function_signature.py
```diff
@@ -0,0 +1,817 @@
...
+
+    # Handle allOf
+    if 'allOf' in schema:
+        if len(schema['allOf']) == 1:
+            return schema_to_type_expr(schema['allOf'][0], defs, referenced_types, tool_name, path)
+        return 'Any'
+
+    # Handle const
+    if 'const' in schema:
+        return f'Literal[{repr(schema["const"])}]'
```
DouweM: I'd like this + the Literal union below to be a specific type of TypeExpr rather than just a string

Also for 'Any" etc, a type with a __str__ method seems better, so we can plausibly map it to TypeScript code at some point. Maybe TypeExpr shouldn't be strings at all?

pydantic_ai_slim/pydantic_ai/function_signature.py
```diff
@@ -0,0 +1,817 @@
...
+
+def _build_and_register_type(
+    name: str,
+    schema: dict[str, Any],
+    defs: dict[str, dict[str, Any]],
+    referenced_types: dict[str, TypeSignature],
+    tool_name: str,
+    path: str,
+) -> None:
+    """Build a TypeSignature and register it, pre-registering a placeholder to prevent infinite recursion.
```
DouweM: What does "register" mean here since this method doesn't return anything? It references it _on_ referenced_types? Then it seems better to me to return that as well.

pydantic_ai_slim/pydantic_ai/function_signature.py
```diff
@@ -0,0 +1,817 @@
...
+    # Replace in return type
+    sig.return_type = _replace_in_expr(sig.return_type)
+
+    # Replace in field types of referenced types
+    for type_sig in sig.referenced_types:
+        for f in type_sig.fields.values():
+            f.type = _replace_in_expr(f.type)
+
+
+def dedup_referenced_types(signatures: list[FunctionSignature]) -> None:
```
DouweM: Does this need to be public?

pydantic_ai_slim/pydantic_ai/function_signature.py
```diff
@@ -0,0 +1,817 @@
...
+            else:
+                # Same name, different definition — rename to avoid conflict
+                new_name = f'{sig.name}_{name}'
+                type_sig.name = new_name  # Mutate in place → propagates everywhere via TypeExpr refs
+                seen[new_name] = type_sig
+                deduped.append(type_sig)
+        sig.referenced_types = deduped
+
+
+def collect_unique_referenced_types(signatures: list[FunctionSignature]) -> list[TypeSignature]:
```
DouweM: Does this need to be public?

pydantic_ai_slim/pydantic_ai/function_signature.py
```diff
@@ -0,0 +1,845 @@
...
+
+    def __str__(self) -> str:
+        """Render this parameter as a function parameter string."""
+        type_str = str(self.type)
+        if self.default is not None:
+            return f'{self.name}: {type_str} = {self.default}'
+        return f'{self.name}: {type_str}'
+
+
+@dataclass
```
DouweM: Make these kw_only

pydantic_ai_slim/pydantic_ai/tools.py
```diff
@@ -457,10 +460,14 @@ def from_schema(
             sequential=sequential,
             args_validator=args_validator,
         )
+        # Schema-based tools use proxy functions (e.g. def proxy(*args, **kwargs))
+        # whose signatures are not meaningful — force schema-based signature generation.
+        tool._original_func = None
```
DouweM: I don't like setting a private field for this -- can we use a new kwarg on the constructor / dataclass field?

pydantic_ai_slim/pydantic_ai/tools.py
```diff
@@ -469,6 +476,7 @@ def tool_def(self):
             metadata=self.metadata,
             timeout=self.timeout,
             kind='unapproved' if self.requires_approval else 'function',
+            original_func=self._original_func,
```
DouweM: Do we really need this subclass and the extra field? Could we eagerly generate the signature, and store it on the FunctionToolDefinition? Or is that expensive if it's not needed? It should be cached of course. The method would need to become a cached property I presume.

pydantic_ai_slim/pydantic_ai/tools.py
```diff
@@ -580,4 +588,61 @@ def defer(self) -> bool:
...
+        """Generate a function signature representation from this tool's JSON schema.
+
+        The result is cached so that repeated access (e.g. on every agent step)
+        does not re-traverse the schema.
+        """
+        return schema_to_signature(
+            name=self.name,
+            parameters_schema=self.parameters_json_schema,
+            description=self.description,
+            return_schema=(self.metadata or {}).get('output_schema'),
```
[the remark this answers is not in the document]
DouweM: Put a comment that this will be replaced with issue https://github.com/pydantic/pydantic-ai/pull/3865

pydantic_ai_slim/pydantic_ai/tools.py
```diff
@@ -580,4 +588,61 @@ def defer(self) -> bool:
...
+    """
+
+    def __init__(self, *, original_func: Callable[..., Any] | None = None, **kwargs: Any) -> None:
+        super().__init__(**kwargs)
+        self._original_func = original_func
+
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, ToolDefinition):
+            return NotImplemented
+        return all(getattr(self, f.name) == getattr(other, f.name) for f in dataclass_fields(ToolDefinition))
```
DouweM: Yeah I would much rather not have this ugly class, if we could pre-compute it without too much cost.

pydantic_ai_slim/pydantic_ai/tools.py
```diff
@@ -580,4 +588,61 @@ def defer(self) -> bool:
...
+        if self._original_func is not None:
+            return FunctionSignature.from_function(
+                self._original_func,
+                name=self.name,
+                description=self.description,
+            )
+        return super().function_signature
+
+    def __repr__(self) -> str:
+        # Show as ToolDefinition so this subclass is invisible to users.
```
DouweM: Why would that need to be the case? If we want to stop it from being serialized on the way from e.g. a Temporal workflow -> activity, there should be better ways to stop it from being serialized.

tests/test_toolsets.py
```diff
@@ -135,6 +135,24 @@ def subtract(a: int, b: int) -> int:
...
+    def add(a: int, b: int) -> int:
+        return a + b
+
+    managed_toolset = await ToolManager[None](toolset).for_run_step(build_run_context(None))
+    assert managed_toolset.tools is not None
+
+    sig = managed_toolset.tools['add'].tool_def.function_signature
+    assert sig.name == 'add'
+    assert list(sig.params) == ['a', 'b']
+    assert str(sig.return_type) == 'int'
```
DouweM: Let's also snapshot the rendering here

DouweM (changes requested):

PR #4859, opened 2026-03-26
feat(capabilities)!: output validate/process hooks; scope `prepare_tools` to function tools, add `prepare_output_tools`

### LINKED ISSUE #5111: after_tool_execute hook fires for internal output tools, corrupting result.output when returning ToolReturn
### Initial Checks

### Description

Unscoped `after_tool_execute` hooks fire for internal output tools (e.g. `final_result`).  When the hook returns `ToolReturn`, the output tool code path doesn't unwrap it (unlike  function tools at `_agent_graph.py:1759`), so `result.output` becomes a `ToolReturn` instead of the expected output type.

This seems inconsistent with the rest of the architecture — `WrapperToolset` and `prepare_tools`/`prepare_output_tools` both explicitly exclude output tools.

```Python
from pydantic_ai import Agent, RunContext, ToolReturn
from pydantic_ai.capabilities import Hooks
from typing import Any


class MyOutput(BaseModel):
    answer: str


hooks: Hooks = Hooks()


@hooks.on.after_tool_execute
async def wrap_result(
    ctx: RunContext[None],
    *,
    call: ToolCallPart,
    tool_def: ToolDefinition,
    args: dict[str, Any],
    result: Any,
) -> ToolReturn:
    return ToolReturn(return_value=result, content="Extra context for the model")


agent = Agent(
    "openai:gpt-4o",
    output_type=MyOutput,
    capabilities=[hooks],
)

result = await agent.run("Say hello")
print(type(result.output))
# Expected: <class 'MyOutput'>
# Actual:   <class 'pydantic_ai.messages.ToolReturn'>

result.output.model_dump()
# AttributeError: 'ToolReturn' object has no attribute 'model_dump'
```

### Logfire Trace

_No response_

- Python: 3.13.2
- Pydantic AI: 1.82.0
- LLM provider SDK:

### LINKED ISSUE #5241: v2: scope `prepare_tools` capability hook to function tools only
The `prepare_tools` capability hook on `AbstractCapability` currently receives **all** tool defs — function tools and output tools alike — matching the hook's released behavior in the capabilities system.

The rest of the tool-hook lifecycle separates function vs output paths:

- `before_tool_validate` / `after_tool_validate` / `wrap_tool_validate` / `on_tool_validate_error`
- `before_tool_execute` / `after_tool_execute` / `wrap_tool_execute` / `on_tool_execute_error`

…all fire only for function tools. Output tools go through the dedicated output hook lifecycle (`{before,wrap,after,on_error}_output_{validate,process}`). `prepare_tools` is the odd one out.

In #4859 we added a parallel `prepare_output_tools` capability hook (additive, output-only, with `ctx.max_retries` reflecting the **output** retry budget). We considered also scoping `prepare_tools` to function tools only at the same time — to match the rest of the lifecycle — but reverted that change because:

- Capabilities are a public API that's been released for ~2 weeks; subclasses may already override `prepare_tools` and rely on seeing the merged set.
- The current "sees all tools" behavior is documented and shipped, so removing output tools is a breaking change rather than a bug fix.

For the next major version, `prepare_tools` should be scoped to function tools only (matching the tool-hook lifecycle). Migration: capabilities that need to filter/modify output tools should override `prepare_output_tools` instead — added in #4859 with the proper output retry context.

🤖 Generated by Claude Code

DouweM:
### Output hooks (new)

- Add `before_output_validate` / `wrap_output_validate` / `after_output_validate` / `on_output_validate_error` and `before_output_process` / `wrap_output_process` / `after_output_process` / `on_output_process_error` hooks on `AbstractCapability` (composed in `CombinedCapability`, delegated in `WrapperCapability`, available via `@hooks.on.*` decorators on `Hooks`). The **validate** pair fires for raw input shape (text or tool args); the **process** pair fires for the semantic value (`MyModel(...)`, `42`, the dict the multi-arg output function will receive, etc.).
- Existing `output_validator` callbacks now run inside the process pair, so one capability can wrap built-in + user-defined validation in one place.
- `OutputContext` exposes `mode`, `output_type`, `object_def`, `has_function`, `tool_call`, `tool_def` so hooks know what kind of output they're handling.

### `prepare_tools` / `prepare_output_tools` (breaking change, framed as bug fix)

The released `prepare_tools` had a split-personality bug: the `Agent(prepare_tools=...)` kwarg saw function tools only and filtered at the registry level, while the `AbstractCapability.prepare_tools` hook saw all tools (function + output) and only filtered visibility. This PR normalizes both onto one path:

- `prepare_tools` capability hook → now receives **function tools only**, filters at the registry level (via `PreparedToolset`). Mirrors what the kwarg has always done.
- New `prepare_output_tools` capability hook → receives **only** [output tools][pydantic_ai.output.ToolOutput], with `ctx.retry`/`ctx.max_retries` reflecting the **output** retry budget (`max_result_retries`) — fixing the missing context info that was the original motivation for #4745.
- The `Agent(prepare_tools=...)` and new `Agent(prepare_output_tools=...)` kwargs inject `PrepareTools` / `PrepareOutputTools` capabilities that flow through the same hook path the user-defined capability uses.
- Order: `prepare_tools` wraps inside other capability `get_wrapper_toolset` results (e.g. `ToolSearch`, `CodeMode`), preserving main's kwarg ordering — toolset transformations layer on top of prepared defs.

This subsumes #5241 (was filed as a v2 follow-up).

### Other behavior fixes (observable through public API)

- A model hallucinating a call to a `prepare_tools`-filtered tool now hits "unknown tool" instead of silently executing it (filtering is registry-level, not just visibility).
- Multi-arg output functions inside a union (e.g. `PromptedOutput([combine_func, OtherType])`) now actually run; previously the union dispatch isinstance-checked the validated dict against the function's first-arg type and silently bypassed the function.
- `ModelRetry` raised by an output function during `run_stream()` now surfaces as `UnexpectedModelBehavior` caused-by `ModelRetry` (caught by the streaming validator), instead of escaping unhandled wrapped in `ToolRetryError`.
- Output retry prompts pass `include_context=False` to `ValidationError.errors()` (matching tool retries), removing the duplicative `ctx` field from `RetryPromptPart` content.

### Cleanup

- Drop the legacy `process()` method on `BaseOutputProcessor` and its subclasses; production paths use `hook_validate` + `hook_execute` exclusively.

## Docs

- `docs/capabilities.md` and `docs/hooks.md` updated with output hook docs and the unified `prepare_tools` / `prepare_output_tools` model.

## Test plan

- [x] All existing tests pass; new tests cover output validate/process hook lifecycle (per processor type and across composition), output-tool nesting, error recovery via `on_output_validate_error`, pre-parse transformation via `before_output_validate`, hook decorator API, the new `prepare_output_tools` hook + constructor arg, hallucinated-call blocking, multi-arg union dispatch, and the streaming-`ModelRetry` regression. Retry-flow tests now snapshot `result.all_messages()` per `tests/AGENTS.md` guidance.

Closes #5111
Closes #5241


pydantic_ai_slim/pydantic_ai/capabilities/abstract.py
```diff
@@ -494,6 +504,121 @@ async def on_tool_execute_error(
...
+        hooks have already run at this point).
+        """
+        return raw_output
+
+    async def after_output_validate(
+        self,
+        ctx: RunContext[AgentDepsT],
+        *,
+        raw_output: RawOutput,
+        output: RawOutput,
```
DouweM: Should this be `RawOutput` (`str | dict`) or do we know it's been validated into "definitely a dict" by now? Doesn't seem like `RawOutput` is the right type.

pydantic_ai_slim/pydantic_ai/capabilities/abstract.py
```diff
@@ -494,6 +504,121 @@ async def on_tool_execute_error(
...
+        return output
+
+    async def wrap_output_validate(
+        self,
+        ctx: RunContext[AgentDepsT],
+        *,
+        raw_output: RawOutput,
+        output_context: OutputContext,
+        handler: WrapOutputValidateHandler,
+    ) -> RawOutput:
```
DouweM: Similarly, I don't think this is the right type here.

pydantic_ai_slim/pydantic_ai/capabilities/abstract.py
```diff
@@ -494,6 +504,121 @@ async def on_tool_execute_error(
...
+        """
+        raise error
+
+    # --- Output execute lifecycle hooks ---
+
+    async def before_output_execute(
+        self,
+        ctx: RunContext[AgentDepsT],
+        *,
+        output: RawOutput,
```
DouweM: The `output` vs `raw_output` vs `validated_output` below is a bit confusing/seemingly inconsistent.

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -72,6 +72,184 @@
 DEFAULT_OUTPUT_TOOL_DESCRIPTION = 'The final response which ends this conversation'
 
 
+@dataclass
+class OutputContext:
```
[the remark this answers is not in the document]
DouweM: Yeah this belongs in the public `pydantic_ai.output` module

pydantic_ai_slim/pydantic_ai/capabilities/abstract.py
DouweM: Hmm I guess that for text output, it could just be a string here (and `output_context.output_type` would `is str`). Still `RawOutput` seems confusing.

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -72,6 +72,189 @@
...
+    """The tool definition, for tool-based output. None for text output."""
+
+
+def _build_output_handlers(
+    processor: BaseOutputProcessor[OutputDataT],
+    run_context: RunContext[AgentDepsT],
+    allow_partial: bool,
+    wrap_validation_errors: bool,
+) -> tuple[
+    Callable[[str | dict[str, Any]], Awaitable[str | dict[str, Any]]],
```
DouweM: No `RawOutput` or smth here?

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -72,6 +72,189 @@
...
+    Callable[[str | dict[str, Any]], Awaitable[Any]],
+]:
+    """Build validate and execute handler functions based on processor type."""
+    if isinstance(processor, TextFunctionOutputProcessor):
```
DouweM: The isinstance branching is unfortunate here; this couldn't be done with methods on the various subclasses?

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -72,6 +72,189 @@
...
+        # Structured output: validate=Pydantic validation, execute=extraction+function
+        async def do_validate(data: str | dict[str, Any]) -> str | dict[str, Any]:
+            if isinstance(data, str):
+                data = _utils.strip_markdown_fences(data)
+            return processor.validate(
+                data, allow_partial=allow_partial, validation_context=run_context.validation_context
+            )
+
+        async def do_execute(output: str | dict[str, Any]) -> Any:
+            return await processor.call(output, run_context, wrap_validation_errors)  # type: ignore[arg-type]
```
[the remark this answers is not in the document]
DouweM: Yeah seems like part of the issue is that we're trying to make the hooks work for both text and structured output -- so we can't rely on raw output often being a string and validated output always being a dict.

I don't think we have much of an option though, as both are considered output from the perspective of output validators and `AgentRunResult.output` etc, and we'd want capabilities/hooks to be able to able to modify them both.

It's not a huge problem if hooks need a bit more `isinstance` checking than regular Pydantic AI code, but we should ensure (with `assert`?) that output processors that need a dict really get a dict, not str.

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -72,6 +72,189 @@
...
+        # UnionOutputProcessor: full process() in validate, identity execute.
+        async def do_validate(data: str | dict[str, Any]) -> str | dict[str, Any]:
+            assert isinstance(data, str)
+            result = await processor.process(
+                data,
+                run_context=run_context,
+                allow_partial=allow_partial,
+                wrap_validation_errors=wrap_validation_errors,
+            )
+            return result  # type: ignore[return-value]
```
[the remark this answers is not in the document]
DouweM: Seems like a real issue

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -72,6 +72,186 @@
...
+
+        async def do_execute(output: str | dict[str, Any]) -> Any:
+            return output
```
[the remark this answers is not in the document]
DouweM: Another real issue

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -72,6 +72,186 @@
...
+
+    return do_validate, do_execute
```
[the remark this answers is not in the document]
DouweM: Yeah this is definitely an issue; we should attempt clean decomposition -- if output type is a union of types including output functions, the output function should still be wrapped in the output_execute hook.

The agent should think some more about possible solutions to this, to make it work as expected by the user.

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -540,12 +734,19 @@ def __init__(
...
+
         if inspect.isfunction(output) or inspect.ismethod(output):
             self._function_schema = _function_schema.function_schema(output, GenerateToolJsonSchema)
             self.validator = self._function_schema.validator
             json_schema = self._function_schema.json_schema
             json_schema['description'] = self._function_schema.description
+
+            # Extract return type from function signature for output_type
+            type_hints = _utils.get_function_type_hints(output)
+            self._output_type = type_hints.get('return', None)
```
DouweM: Hmm, from the perspective of the hooks, isn't the output_type the input to the output function?

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -621,7 +822,7 @@ async def process(
 
         try:
             output = self.validate(data, allow_partial=allow_partial, validation_context=run_context.validation_context)
-        except ValidationError as e:
+        except ValidationError as e:  # pragma: no cover — error handling now in run_output_with_hooks; only reachable via UnionOutputProcessor.process() or capability=None
```
DouweM: This line should definitely be covered.

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -834,6 +1060,9 @@ def __init__(
 
         self._str_argument_name = argument_name
 
+        type_hints = _utils.get_function_type_hints(output_function)
+        self._output_type: type[Any] | None = type_hints.get('return', None)
```
DouweM: See above, maybe output type should be the function input type in this case.

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -967,7 +1211,7 @@ async def get_tools(self, ctx: RunContext[AgentDepsT]) -> dict[str, ToolsetTool[
             for tool_def in self._tool_defs
         }
 
-    async def call_tool(
+    async def call_tool(  # pragma: no cover — execution now handled by _raw_execute_output_tool when capability is present
```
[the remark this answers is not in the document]
DouweM: Agreed, this line needs actual coverage.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -1031,7 +1031,13 @@ async def _handle_text_response(
...
         )
 
-        result_data = await text_processor.process(text, run_context=run_context)
+        result_data = await _output.run_output_with_hooks(
+            text_processor,
+            text,
+            run_context=run_context,
+            capability=ctx.deps.root_capability,
+            output_mode=ctx.deps.output_schema.mode,
+        )
```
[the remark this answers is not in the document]
DouweM: Maybe output_context also needs an allow_text_output field, like we have on ModelRequestParameters? There's also image output to think about.

pydantic_ai_slim/pydantic_ai/result.py
```diff
@@ -226,16 +227,18 @@ async def validate_response_output(
...
                     text = ''
 
-            result_data = await text_processor.process(
+            run_ctx = replace(self._run_ctx, partial_output=allow_partial)
+            result_data = await run_output_with_hooks(
+                text_processor,
                 text,
-                run_context=replace(self._run_ctx, partial_output=allow_partial),
+                run_context=run_ctx,
+                capability=self._tool_manager.root_capability,
```
DouweM: Reading this off the tool manager seems iffy

pydantic_ai_slim/pydantic_ai/_tool_manager.py
```diff
@@ -479,23 +479,84 @@ async def _raw_execute(
...
+    async def _raw_execute_output_tool(
+        self,
+        validated: ValidatedToolCall[AgentDepsT],
+    ) -> Any:
+        """Execute an output tool call with output hooks."""
+        from ._output import (
+            OutputToolset,
+            run_output_execute_hooks,
+            run_output_validate_hooks,
+        )
```
DouweM: Top level imports please

pydantic_ai_slim/pydantic_ai/_tool_manager.py
```diff
@@ -479,23 +479,84 @@ async def _raw_execute(
...
+        processor = toolset.processors[name]
+        output_context = processor.get_output_context(
+            'tool', tool_call=validated.call, tool_def=validated.tool.tool_def
+        )
+
+        # --- Output validate phase (identity — args already validated by tool_validate hooks) ---
+        async def do_validate(data: str | dict[str, Any]) -> str | dict[str, Any]:
+            return data  # Identity: validation already done by tool hooks
+
+        validated_output = await run_output_validate_hooks(
```
DouweM: This (and similar) function needs to use kwargs only to make it more clear what the args mean

pydantic_ai_slim/pydantic_ai/_tool_manager.py
```diff
@@ -479,23 +479,84 @@ async def _raw_execute(
...
+            validated.validated_args,
+            do_validate,
+            allow_partial=False,
+            wrap_validation_errors=False,
+        )
+
+        # --- Output execute phase (wraps processor.call + output validators) ---
+        async def do_execute(output: str | dict[str, Any]) -> Any:
+            try:
+                result = await processor.call(output, ctx, wrap_validation_errors=False)  # type: ignore[arg-type]
```
DouweM: `ignore[arg-type]` potentially hides real bugs

tests/test_output_hooks.py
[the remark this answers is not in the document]
DouweM: If this is still the case it needs fixing for sure.

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -984,10 +1330,9 @@ async def get_tools(self, ctx: RunContext[AgentDepsT]) -> dict[str, ToolsetTool[
     async def call_tool(
         self, name: str, tool_args: dict[str, Any], ctx: RunContext[AgentDepsT], tool: ToolsetTool[AgentDepsT]
     ) -> Any:
-        output = await self.processors[name].call(tool_args, ctx, wrap_validation_errors=False)
-        for validator in self.output_validators:
-            output = await validator.validate(output, ctx, wrap_validation_errors=False)
-        return output
+        # Output tools are handled by ToolManager.validate_output_tool_call/execute_output_tool_call,
+        # not through the normal toolset.call_tool path.
+        raise NotImplementedError('Output tools use validate_output_tool_call/execute_output_tool_call')
```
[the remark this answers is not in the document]
DouweM: I don't care.

pydantic_ai_slim/pydantic_ai/capabilities/abstract.py
```diff
@@ -525,6 +535,150 @@ async def on_tool_execute_error(
...
+        Fires for all output types: text, structured text, and tool-based output.
+
+        For text/structured output, `output` is the raw text string from the model.
+        For tool output, `output` is the raw tool arguments (string or dict).
+
+        Raise [`ModelRetry`][pydantic_ai.exceptions.ModelRetry] to skip validation and
+        ask the model to try again with a custom message.
+
+        During streaming, this hook fires on every partial validation attempt as well as
+        the final result. Check ``ctx.partial_output`` to distinguish and avoid expensive
```
DouweM: No double backticks please -- this is Markdown

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -72,8 +73,202 @@
...
+
+
+async def run_output_execute_hooks(
+    capability: AbstractCapability[AgentDepsT],
+    *,
+    run_context: RunContext[AgentDepsT],
+    output_context: OutputContext,
+    output: Any,
+    do_execute: Callable[[Any], Awaitable[Any]],
+    wrap_validation_errors: bool = True,
```
DouweM, line 191: Why does this have a default, but not the equiv arg on run_output_validate_hooks?

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -72,8 +73,202 @@
...
+        output = await capability.before_output_execute(run_context, output_context=output_context, output=output)
+
+        try:
+            result = await capability.wrap_output_execute(
+                run_context, output_context=output_context, output=output, handler=do_execute
+            )
+        except ToolRetryError:
+            raise  # Control flow, not error
+        except ModelRetry:
+            raise  # Propagate to outer handler, skip on_output_execute_error
```
DouweM: Why does the logic here look so different than in the validate equiv above? That has more wrap_validation_errors checks etc

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -663,11 +900,42 @@ async def call(
...
+        )
+
+
+@dataclass
+class _UnionValidatedOutput:
+    """Wrapper returned by UnionOutputProcessor.validate() carrying the resolved kind key.
+
+    This allows call() to re-resolve the correct inner processor without storing
+    it as instance state (which would be a race condition under concurrent runs).
+    The wrapper is unwrapped in _build_output_handlers so hooks see clean inner data.
```
DouweM: Backticks around all code refs in docstrings

pydantic_ai_slim/pydantic_ai/_tool_manager.py
```diff
@@ -397,11 +398,11 @@ async def validate_tool_call(
             max_retries = tool.max_retries if tool is not None else self.default_max_retries
             self._check_max_retries(name, max_retries, e)
 
-            if not allow_partial:
+            if not allow_partial:  # pragma: no branch — allow_partial only used via validate_output_tool_call
```
DouweM: Do we still need the arg / branch then?

pydantic_ai_slim/pydantic_ai/_tool_manager.py
```diff
@@ -448,6 +443,179 @@ async def execute_tool_call(
...
+        tool = self.tools.get(name)
+        ctx = self.ctx
+
+        try:
+            if tool is None:  # pragma: no cover — output tool names are agent-defined, always exist
+                if self.tools:
+                    msg = f'Available tools: {", ".join(f"{n!r}" for n in self.tools)}'
+                else:
+                    msg = 'No tools available.'
+                raise ModelRetry(f'Unknown tool name: {name!r}. {msg}')
```
DouweM: Seems like there's a lot of duplication here with the other path, which we should reduce

pydantic_ai_slim/pydantic_ai/_tool_manager.py
```diff
@@ -502,7 +671,7 @@ async def _raw_execute(
             self.failed_tools.add(name)
             raise self._wrap_error_as_retry(name, validated.call, e) from e
 
-        if usage is not None:
+        if usage is not None:  # pragma: no branch — agent always passes usage
```
DouweM: Then could we make the type not-nullable?

pydantic_ai_slim/pydantic_ai/capabilities/hooks.py
```diff
@@ -179,6 +183,38 @@ def __call__(self, ctx: RunContext[Any], /, *, call: ToolCallPart, tool_def: Too
...
+class AfterOutputValidateHookFunc(Protocol):
+    """Protocol for :meth:`~AbstractCapability.after_output_validate` hook functions."""
+    def __call__(self, ctx: RunContext[Any], /, *, output_context: OutputContext, output: Any) -> Any | Awaitable[Any]: ...
+
+class WrapOutputValidateHookFunc(Protocol):
+    """Protocol for :meth:`~AbstractCapability.wrap_output_validate` hook functions."""
+    def __call__(self, ctx: RunContext[Any], /, *, output_context: OutputContext, output: RawOutput, handler: WrapOutputValidateHandler) -> Any | Awaitable[Any]: ...
+
+class OnOutputValidateErrorHookFunc(Protocol):
+    """Protocol for :meth:`~AbstractCapability.on_output_validate_error` hook functions."""
```
DouweM: Markdown! None of this :meth: stuff

pydantic_ai_slim/pydantic_ai/capabilities/abstract.py
```diff
@@ -52,6 +53,15 @@
 WrapToolExecuteHandler: TypeAlias = 'Callable[[dict[str, Any]], Awaitable[Any]]'
 """Handler type for [`wrap_tool_execute`][pydantic_ai.capabilities.AbstractCapability.wrap_tool_execute]."""
 
+RawOutput: TypeAlias = 'str | dict[str, Any]'
+"""Type alias for raw output data (text or tool args)."""
+
+WrapOutputValidateHandler: TypeAlias = 'Callable[[RawOutput], Awaitable[Any]]'
+"""Handler type for wrap_output_validate."""
+
+WrapOutputProcessHandler: TypeAlias = 'Callable[[Any], Awaitable[Any]]'
```
DouweM: We shouldn't need quotes here. If we don't have "from future import annotations" yet we can.

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -74,8 +74,350 @@
...
+    if fs is None:
+        return None
+    if fs.single_arg_name is not None:
+        # Model-like single arg: validator wraps as `{name: model}` via _validate_single_arg
+        return fs.single_arg_name
+    # Non-model single arg: validator returns `{name: value}` as a regular TypedDict.
+    # Detect via json_schema having exactly one property.
+    properties = fs.json_schema.get('properties', {})
+    if len(properties) == 1:
+        return next(iter(properties))
```
DouweM: Is this branch needed? In which case do we not get what we need from outer typed dict key or single arg name?

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -74,8 +74,350 @@
...
+    if not isinstance(processor, ObjectOutputProcessor):
+        return None
+    if processor.outer_typed_dict_key is not None:
+        return processor.outer_typed_dict_key
+    fs = processor._function_schema  # pyright: ignore[reportPrivateUsage]
+    if fs is None:
+        return None
+    if fs.single_arg_name is not None:
+        # Model-like single arg: validator wraps as `{name: model}` via _validate_single_arg
+        return fs.single_arg_name
```
DouweM: Would it make any sense to "combine" the `single_arg_name` and `outer_typed_dict_key` features? Should we store `single_arg_name` on `outer_typed_dict_key`? Or heave another processor helper method so we don't have to pass through the private function schema?

pydantic_ai_slim/pydantic_ai/_output.py
DouweM: Is this branch needed? In which case do we not get what we need from outer typed dict key or single arg name?

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -74,8 +74,350 @@
...
+    For UnionOutputProcessor, the validate handler unwraps _UnionValidatedOutput so hooks
+    see clean inner data, and the execute handler re-wraps it so call() can resolve the
+    correct inner processor. The kind is stored per-invocation in closure state, avoiding
+    the race condition of shared instance state. The inner processor's unwrap key is
+    resolved dynamically once the union member is known.
+    """
+    _union_kind: str | None = None
+    # For non-union processors, the unwrap key is fixed upfront.
+    # For unions, it's resolved dynamically in do_validate once the member is known.
+    _unwrap_key: str | None = None if isinstance(processor, UnionOutputProcessor) else _get_unwrap_key(processor)
```
DouweM: Do we need the isinstance check, or could `_get_unwrap_key` naturally return` None` for `UnionOutputProcessor`? I imagine it already would.

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -74,8 +74,350 @@
...
+        result: Any = processor.validate(
+            data,
+            allow_partial=allow_partial,
+            validation_context=run_context.validation_context,
+        )
+        if isinstance(result, _UnionValidatedOutput):
+            _union_kind = result.kind
+            # Now that we know the selected union member, resolve its unwrap key
+            assert isinstance(processor, UnionOutputProcessor)
+            inner = processor._processors[_union_kind]  # pyright: ignore[reportPrivateUsage]
```
DouweM: This private usage is pretty horrible. Could `_UnionValidatedOutput` be responsible for holding the unwrapped/semantic value?

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -74,8 +74,350 @@
...
+            _unwrap_key = _get_unwrap_key(inner)
+            result = result.data
+        return _unwrap_semantic_value(result, _unwrap_key)
+
+    async def do_process(output: Any) -> Any:
+        # Re-wrap the semantic value into the internal dict shape `processor.call()` expects.
+        output = _rewrap_internal_value(output, _unwrap_key)
+
+        if _union_kind is not None:
+            # Normal path: re-wrap with the kind resolved during validation
```
DouweM: What happens if the processor decides to return another type that is also supported by the union? Should we be in the next branch?

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -74,8 +74,350 @@
...
+    output_context: OutputContext,
+    output: Any,
+    do_process: Callable[[Any], Awaitable[Any]],
+    wrap_validation_errors: bool = True,
+) -> Any:
+    """Run the output process hooks around do_process.
+
+    Process hooks fire for ALL output types (text, structured, tool, image).
+
+    ValidationError and ModelRetry from any hook (before, after, wrap, on_error) are caught
```
DouweM: Wrap code snippets in Markdown backticks please

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -574,12 +945,22 @@ def __init__(
...
         if inspect.isfunction(output) or inspect.ismethod(output):
             self._function_schema = _function_schema.function_schema(output, GenerateToolJsonSchema)
             self.validator = self._function_schema.validator
             json_schema = self._function_schema.json_schema
             json_schema['description'] = self._function_schema.description
+
+            # Extract the function's input type (what the model produces) for output_type
+            type_hints = _utils.get_function_type_hints(output)
+            for hint_name, hint_type in type_hints.items():  # pragma: no branch
+                if hint_name != 'return' and hint_type is not RunContext and get_origin(hint_type) is not RunContext:
```
DouweM: I think we have a util helper method for determining if an arg/hint is RunContext. Not sure if it's needed here

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -799,6 +1214,68 @@ def __init__(
...
+    def resolve_processor_for_recovered_output(
+        self, output: Any
+    ) -> tuple[ObjectOutputProcessor[OutputDataT], Any] | None:
+        """Match a recovered output value to the correct inner processor by type.
+
+        Used by the error recovery path in output hooks: when ``on_output_validate_error``
+        returns data directly (bypassing normal union validation), this method finds which
+        inner processor can handle the recovered value so the output function still runs.
+
+        Returns ``(processor, args)`` where ``args`` is the value to pass to ``processor.call()``
```
DouweM: Code snippets want single Markdown backticks.

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -799,6 +1214,68 @@ def __init__(
...
+        (potentially wrapped in a dict for non-model-like types), or ``None`` if no match is found.
+        """
+        for inner in self._processors.values():
+            if inner._output_type is not None:  # pyright: ignore[reportPrivateUsage]  # pragma: no branch
```
DouweM: Should this `output_type` be a public field if we need it here?

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -799,6 +1214,68 @@ def __init__(
...
+                try:
+                    if isinstance(output, inner._output_type):  # pyright: ignore[reportPrivateUsage]
+                        args: Any = output
+                        if inner.outer_typed_dict_key:  # pragma: no cover — non-model-like union members are rare
```
DouweM: Eeeh are you sure? We should definitely test that scenario. E.g. `Foo | Bar | bool | None`

DouweM: Eeeh are you sure? We should definitely test that scenario. E.g. `Foo | list[Bar] | bool | None`

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -799,6 +1214,68 @@ def __init__(
...
+                            args = {inner.outer_typed_dict_key: output}
+                        return inner, args
+                except TypeError:  # pragma: no cover — generic types (e.g. list[str]) can't be used with isinstance
```
DouweM: I wish we could do something more useful here, but maybe it becomes a recursive mess of validating the list/dict etc types?

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -799,6 +1214,68 @@ def __init__(
...
+                    pass
+        return None  # pragma: no cover — hook returned a type not in the union
```
DouweM: Seems worth covering in tests! All these edge cases need to have "somewhat sane" behavior, although there's not always a ton we can do if people don't match the intended types. But we shouldn't be dropping values that we skip here...

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -846,6 +1329,16 @@ async def process(
     ) -> OutputDataT:
         return cast(OutputDataT, data)
 
+    def get_output_context(
+        self,
+        mode: OutputMode,
```
DouweM: I think this method should require kwargs

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -868,6 +1361,18 @@ def __init__(
 
         self._str_argument_name = argument_name
 
+    async def call(
+        self,
+        output: Any,
+        run_context: RunContext[AgentDepsT],
```
DouweM, line 1410: Anything after output should be a kwarg on methods like this one

pydantic_ai_slim/pydantic_ai/tool_manager.py
```diff
@@ -57,8 +65,14 @@ class ValidatedToolCall(Generic[AgentDepsT]):
...
     """Whether argument validation (schema + custom validator) passed."""
-    validated_args: dict[str, Any] | None = None
-    """The validated arguments if validation passed, None otherwise."""
+    validated_args: Any = None
+    """The validated arguments if validation passed, None otherwise.
+
+    For tool calls: `dict[str, Any]` matching the tool's schema.
+    For output tools: the **semantic value** (what the model was asked to produce) — e.g.,
+    `MyModel(...)`, `42`, or a dict for multi-arg output functions. This is what output hooks
+    see; `execute_output_tool_call` re-wraps it to dict form before `processor.call()`.
```
DouweM: `ValidatedToolCall` is public API for the tool hooks, so I don't love changing this here. If it's only non-dict with output tools, and we don't fire tool hooks for output tools, help me understand why we need this change at all? When does `ValidatedToolCall` get into play with output tool calls/hooks?

DouweM: `ValidatedToolCall` is public API for the tool hooks, so I don't love changing this here. If it's only non-dict with output tools, and we don't fire tool hooks for output tools, help me understand why we need this change at all? When does `ValidatedToolCall` get into play with output tool calls/hooks? Does it _need_ to hold the semantic value here, or could it use what the tool call args actually we're, with the "output tool unwrapping" happening later/elsewhere?

docs/capabilities.md
```diff
@@ -691,6 +691,32 @@ To skip validation and provide pre-validated args, raise [`SkipToolValidation(ar
 
 To skip execution and provide a replacement result, raise [`SkipToolExecution(result)`][pydantic_ai.exceptions.SkipToolExecution] from `before_tool_execute` or `wrap_tool_execute`.
 
+#### Output hooks
+
+Like tool processing, output processing has two phases: **validation** (parsing the model's raw output against the output schema) and **processing** (extracting the value and calling any output function). Each phase has its own hooks.
```
DouweM: Link to the output docs, output function docs.

docs/capabilities.md
```diff
@@ -691,6 +691,32 @@ To skip validation and provide pre-validated args, raise [`SkipToolValidation(ar
...
+
+All output hooks receive an `output_context` parameter with [`OutputContext`][pydantic_ai.capabilities.OutputContext] (mode, output type, schema info, and tool call details for tool output).
```
DouweM: Link to tool output docs

docs/capabilities.md
```diff
@@ -691,6 +691,32 @@ To skip validation and provide pre-validated args, raise [`SkipToolValidation(ar
...
+
+**Validate hooks** fire only for structured output that requires parsing (prompted, native, tool, union output). They do not fire for plain text or image output. **Process hooks** fire for **all output types** including text, structured, tool, and image output. For output tools, only output hooks fire — tool hooks are skipped entirely.
```
DouweM: in "all output types", "tool" is not reasonable to list, as tool output is a mode.

Also possibly worth linking to "output tools" / tool output doc.

docs/capabilities.md
```diff
@@ -691,6 +691,32 @@ To skip validation and provide pre-validated args, raise [`SkipToolValidation(ar
...
+
+| Hook | Signature | Purpose |
+|---|---|---|
+| [`before_output_validate`][pydantic_ai.capabilities.AbstractCapability.before_output_validate] | `(ctx, *, output_context, output: RawOutput) -> RawOutput` | Modify raw output before validation (e.g. JSON repair) |
+| [`after_output_validate`][pydantic_ai.capabilities.AbstractCapability.after_output_validate] | `(ctx, *, output_context, output: Any) -> Any` | Modify validated output |
+| [`wrap_output_validate`][pydantic_ai.capabilities.AbstractCapability.wrap_output_validate] | `(ctx, *, output_context, output: RawOutput, handler) -> Any` | Wrap the validation step |
+| [`on_output_validate_error`][pydantic_ai.capabilities.AbstractCapability.on_output_validate_error] | `(ctx, *, output_context, output: RawOutput, error: ValidationError \| ModelRetry) -> Any` | Handle validation errors (see [error hooks](#error-hooks)) |
+
+**Processing hooks** — fire for all output types; `output` is the validated/raw output. Output validators ([`@agent.output_validator`][pydantic_ai.Agent.output_validator]) run inside the processing pipeline (within `wrap_output_process`), so `after_output_process` sees the fully validated result:
```
DouweM, line 714: What should we mention in this docs section about raising ModelRetry yourself? I'm sure we say something already in one of the other sections.

docs/hooks.md
```diff
@@ -141,6 +141,46 @@ Execution hooks fire when the tool function runs. `args` is always the validated
...
+| `hooks.on.` | Constructor kwarg | `AbstractCapability` method |
+|---|---|---|
+| `before_output_validate` | `before_output_validate=` | `before_output_validate` |
+| `after_output_validate` | `after_output_validate=` | `after_output_validate` |
+| `output_validate` | `output_validate=` | `wrap_output_validate` |
+| `output_validate_error` | `output_validate_error=` | `on_output_validate_error` |
+
+Output validation hooks fire when structured output is parsed and validated — for prompted/native structured output, this wraps Pydantic validation of the raw text; for tool output, this wraps Pydantic validation of the tool arguments. These hooks only fire for output types that require parsing; they do **not** fire for plain text or image output. All output hooks receive an `output_context` ([`OutputContext`][pydantic_ai.capabilities.OutputContext]) parameter.
+
+The primary use case is **pre-parse normalization**: `before_output_validate` lets you fix malformed model output before it reaches the parser.
```
DouweM: This seems like it deserves to be in teh capabilities doc as well. In general, worth reviewing this hooks doc and seeing how much should be duplicated with the caps doc, vs a quick intro here and then linking to caps for the full details.

docs/hooks.md
```diff
@@ -141,6 +141,46 @@ Execution hooks fire when the tool function runs. `args` is always the validated
...
+| `hooks.on.` | Constructor kwarg | `AbstractCapability` method |
+|---|---|---|
+| `before_output_process` | `before_output_process=` | `before_output_process` |
+| `after_output_process` | `after_output_process=` | `after_output_process` |
+| `output_process` | `output_process=` | `wrap_output_process` |
+| `output_process_error` | `output_process_error=` | `on_output_process_error` |
+
+Output processing hooks fire when the output is processed — extracting values, calling output functions, and running output validators. These hooks fire for **all** output types (text, structured, tool, image). Output validators ([`@agent.output_validator`][pydantic_ai.Agent.output_validator]) run inside the processing pipeline (within `wrap_output_process`), so `after_output_process` sees the fully validated result.
+
+!!! note "Output hooks see the semantic value, not the internal dict form"
```
DouweM: - If this is worth mentioning, I think it belongs in the caps docs, more than here.
- I don't think it _is_ worth mentioning, as this is the natural behavior for output hooks. It only _seems_ worth mentioning if you're very aware of the "output tools" mode and the fact that this is different from tool hooks, but users shouldn't typically (need to) think about that, and shouldn't expect output hooks to behave differently for tool output vs native/prompted output.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -1422,7 +1433,12 @@ async def process_tool_calls(  # noqa: C901
 
             # Validation passed - execute through output hooks
             try:
-                result_data = await tool_manager.execute_output_tool_call(validated)
+                result_data = await tool_manager.execute_output_tool_call(
+                    validated,
+                    allows_text=schema.allows_text,
+                    allows_image=schema.allows_image,
+                    allows_deferred_tools=schema.allows_deferred_tools,
```
DouweM: It's a shame we have to pass these in each time now. Can we just pass in the output schema? Since it's relevant context. We could even pass it into the ToolManager optionally?

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -310,13 +211,55 @@ async def run_output_process_hooks(
...
+    Output validators run inside process hooks, matching the text/structured/image paths.
+    """
+    output_context = OutputContext(
+        mode='text',
+        output_type=type(None),
+        object_def=None,
+        has_function=False,
+        allows_text=allows_text,
+        allows_image=allows_image,
+        allows_deferred_tools=allows_deferred_tools,
```
DouweM: Isn't there a `get_output_context` method somewhere? Why aren't we using that in these paths?

Maybe we can pass in the entire `schema` and read it off there?

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -374,7 +323,12 @@ async def run_output_with_hooks(
             )  # pragma: no cover
         return result
 
-    output_context = processor.get_output_context(output_mode)
+    output_context = processor.get_output_context(
+        output_mode,
+        allows_text=allows_text,
+        allows_image=allows_image,
+        allows_deferred_tools=allows_deferred_tools,
```
DouweM: Why do we have to pass these fields in? They are themselves derived from the output schema/processor, and seem to be getting carried around here unnecessarily? I also don't feel strongly that these need to be included right now if it's a lot of work.

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -1076,9 +1071,19 @@ async def call(
         wrap_validation_errors: bool = True,
     ) -> Any:
         if k := self.outer_typed_dict_key:
+            assert isinstance(output, dict), (
+                f'ObjectOutputProcessor.call expects a dict with key {k!r}; got {type(output).__name__}. '
+                'Output hooks that modify the semantic value must preserve the type; '
+                'to transform the output shape, use `before_output_process` (which operates on the '
+                'semantic value) rather than mutating the internal dict.'
```
DouweM: Do we have to be so strict about this?

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -1088,11 +1093,70 @@ async def call(
...
+        if fs is None:
+            return None
+        if fs.single_arg_name is not None:
+            # Model-like single arg: validator wraps as `{name: model}` via _validate_single_arg
+            return fs.single_arg_name
+        # Non-model single arg: validator returns `{name: value}` as a regular TypedDict.
+        # Detect via json_schema having exactly one property.
+        properties = fs.json_schema.get('properties', {})
+        if len(properties) == 1:
+            return next(iter(properties))
```
DouweM: I still don't understand why we assume that if there's a single key, it has to be a "disposable" one?

pydantic_ai_slim/pydantic_ai/_function_schema.py
```diff
@@ -58,6 +58,21 @@ class FunctionSchema:
     return_schema: ObjectJsonSchema = field(default_factory=dict[str, Any])
     """JSON schema for the function's return type. At minimum `{}` (equivalent to `Any`)."""
 
+    @property
+    def single_field_name(self) -> str | None:
+        """Name of the single argument if the function takes exactly one value-carrying arg, else `None`.
+
+        Covers both model-like single args (via `single_arg_name`, which uses a wrap validator
+        to normalize to `{name: value}`) and primitive single args (where the schema is a
+        one-property TypedDict). Returns `None` for multi-arg functions and `**kwargs`-only.
```
DouweM, line 67: I do feel like we're going to lose data if we always drop the keys in soingle-dict objects. What if we tell the LLM to return `dict[str, str]`? The JSON schema will have `additionalProperties: true`

pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -2442,10 +2441,39 @@ def _get_toolset(
...
+            if run_capability is not None:
+                # Dispatch the new `prepare_output_tools` capability hook through a `PreparedToolset`
+                # wrapped around the output toolset specifically — so the hook only sees output
+                # tools, and the filtered/modified defs flow into `ToolManager.tools` and the model
+                # request parameters together. Override `ctx.max_retries` to the agent's output
+                # retry budget (matches `_build_output_run_context`'s contract — see #4745).
+                # `output_toolset.max_retries` is set to `max_result_retries` at agent construction.
+                output_cap = run_capability
+                output_max_retries = self._max_result_retries
+
```
DouweM, line 2467: It seems significant that we changed the order of agent.prepare_tools <> capability.get_wrapper_toolset? Should we change that back?

tests/test_capabilities.py
```diff
@@ -12437,6 +12665,48 @@ def noop(
...
+                ModelRequest(
+                    parts=[
+                        RetryPromptPart(
+                            content=[
+                                {
+                                    'type': 'json_invalid',
+                                    'loc': (),
+                                    'msg': 'Invalid JSON: expected ident at line 1 column 2',
+                                    'input': 'not json',
+                                    'ctx': {'error': 'expected ident at line 1 column 2'},
```
DouweM: This suggests to me that maybe we should've been using `include_context=False` on `ValidationError.errors`, because this is verbose + duplicative.

PR #4865, opened 2026-03-26
feat: smart instruction caching — Anthropic/Bedrock cache boundary at static/dynamic split

### LINKED ISSUE #4543: Separate cache breakpoints for static vs dynamic instructions (Anthropic prompt caching)
### Description

When using PydanticAI with Anthropic prompt caching, it would be useful to place cache breakpoints between static and dynamic instructions.

Currently PydanticAI concatenates static agent instructions and dynamic @agent.instructions output into a single instructions/system block. Because of this, it’s not possible to insert a CachePoint between them.

**Example setup:**
	•	static agent instructions (never change)
	•	dynamic instructions injected programmatically (date, open file, user context, etc.)

**Example code:**

```@agent.instructions
def dynamic_context_instructions(ctx: RunContext[ExecutionDeps]) -> str:
    return ctx.deps.dynamic_context or ""
```

**Desired caching behavior with Anthropic:**
	•	cache static instructions (~1h)
	•	cache dynamic context separately (changes less often than user turns)
	•	cache the latest user turn separately

The last one works already, but the first two cannot currently be separated because the instructions are merged internally.

**Example code that I wish worked:**

```
# cache breakpoint separately for extra dynamic instructions
@agent.instructions(add_cache_breakpoint=True)
def dynamic_context_instructions(ctx: RunContext[ExecutionDeps]) -> str:
    return ctx.deps.dynamic_context or ""

# cache breakpoint for instructions would be set as is now, like
anthropic_cache_instructions=True

# maybe it could specify it only means static, such as: 
anthropic_cache_static_instructions=True

# alternatively a parameter here as well to cache ALL dynamic instructions at once (not sure if there is an use # case to have separate breakpoints for individual parts, as Anthropic allows for only 4 breakpoints anyway
anthropic_cache_static_instructions=True
anthropic_cache_dynamic_instructions=True
```

**What would be useful for me:**
	•	allow cache points between static and dynamic instructions
	•	internally treat static and dynamic instructions as separate blocks/messages
	•	provide a built-in way to cache static instructions independently

Manually sending another SystemMessage is possible, but then it becomes part of the stored message history which is not ideal.

### References

**Slack discussion** with maintainers suggesting opening a feature request
**Related issue:** https://github.com/pydantic/pydantic-ai/issues/3453

Alex-Resch:
Closes #4543

### Summary

Add structured `InstructionPart(content, dynamic)` to enable smart cache placement for Anthropic and Bedrock prompt caching, while keeping `ModelRequest.instructions` as a simple rendered string.

**Key design:**
- `InstructionPart` dataclass in `messages.py` with `content: str` and `dynamic: bool` fields
- Static instructions (`dynamic=False`) from literal strings; dynamic (`dynamic=True`) from functions, templates, toolsets
- `instruction_parts` lives on `ModelRequestParameters` (not `ModelRequest`), keeping the message type simple
- Models read structured parts from `model_request_parameters.instruction_parts` for per-part system messages
- Anthropic/Bedrock use the `dynamic` flag for smart cache boundary placement (cache after last static instruction)
- All other models emit separate system messages per instruction part for consistency
- `prompted_output_instructions` appended to `instruction_parts` in `Model.prepare_request()` so it uses the final resolved template

**Changes across the codebase:**
- `messages.py`: `InstructionPart` dataclass with `join()` helper; `ModelRequest.instructions` is plain `str | None`
- `models/__init__.py`: `instruction_parts` field on `ModelRequestParameters`; restored original `_get_instructions()` with comments
- `_agent_graph.py`: `_get_instructions()` returns `list[InstructionPart] | None`; passed via `ModelRequestParameters`
- All model files: read `model_request_parameters.instruction_parts`; use `list[i:i] = ...` slice insertion
- Toolset return type widened to support `InstructionPart` while maintaining backward compat

### Pre-Review Checklist

### Pre-Merge Checklist

### DIFF
--- a/tests/models/test_openai.py
+++ b/tests/models/test_openai.py
@@ -4406,6 +4406,24 @@ async def test_openai_chat_instructions_with_user_role(allow_model_requests: Non
     )
 
 
+async def test_openai_chat_instructions_fallback_with_tool_return(allow_model_requests: None):
+    """When the last ModelRequest has only tool-return parts, instructions from the previous request are used."""
+    mock_client = MockOpenAI.create_mock(completion_message(ChatCompletionMessage(content='ok', role='assistant')))
+    model = OpenAIChatModel('gpt-4o', provider=OpenAIProvider(openai_client=mock_client))
+
+    messages: list[ModelRequest | ModelResponse] = [
+        ModelRequest(parts=[UserPromptPart(content='Hello')], instructions='Be helpful.'),
+        ModelResponse(parts=[ToolCallPart(tool_name='my_tool', args='{}', tool_call_id='call_1')]),
+        ModelRequest(parts=[ToolReturnPart(tool_name='my_tool', content='result', tool_call_id='call_1')]),
+    ]
+
+    # No instruction_parts on MRP — model should fall back to message history
+    openai_messages = await model._map_messages(messages, ModelRequestParameters())  # pyright: ignore[reportPrivateUsage]
+
+    # Instructions from the first request should appear as a system message
+    assert openai_messages[0] == snapshot({'content': 'Be helpful.', 'role': 'system'})
+
+
 def test_openai_chat_audio_default_base64(allow_model_requests: None):
     c = completion_message(ChatCompletionMessage(content='success', role='assistant'))
     mock_client = MockOpenAI.create_mock(c)


DouweM:
@Alex-Resch Thanks Alex! I've asked our review bot to have a look and I'll check it out myself tomorrow; I agree we should support something like this.

DouweM:
@Alex-Resch Thanks for working on this! I had a few ideas of how I wanted to change it, so I was iterating on things with Claude -- I'll have some commits to push up soon that I'd then love for you to try out :)

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -1398,8 +1430,24 @@ class ModelRequest:
...
     """The timestamp when the request was sent to the model."""
 
-    instructions: str | None = None
-    """The instructions for the model."""
+    instructions: Annotated[
+        str | None,
+        # `instructions` is serialized for backward compat but is derived from `instruction_parts`
+        # when available. On deserialization of old messages without `instruction_parts`, the string
+        # is used to synthesize a single static InstructionPart in __post_init__.
+        pydantic.Field(validation_alias=pydantic.AliasChoices('instructions')),
```
DouweM: Does this `validation_alias` do anything? The field has the same name already...

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -771,8 +775,13 @@ async def _prepare_request(
         # resume-without-prompt path; ToolManager.for_run_step is a no-op for the same step.
         ctx.deps.tool_manager = await ctx.deps.tool_manager.for_run_step(run_context)
 
-        # Fetch instructions now that dynamic toolsets have been resolved by for_run_step
-        self.request.instructions = await _get_instructions(ctx, run_context)
+        # Fetch instructions now that dynamic toolsets have been resolved by for_run_step.
+        # Sort static before dynamic to match __post_init__ behavior (which doesn't re-run on mutation).
```
DouweM: Can we handle this in a less duplicated way? like a new setter or smth?

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -949,56 +950,55 @@ def base_url(self) -> str | None:
...
         messages: Sequence[ModelMessage], model_request_parameters: ModelRequestParameters | None = None
-    ) -> str | None:
-        """Get instructions from the first ModelRequest found when iterating messages in reverse.
+    ) -> list[InstructionPart] | None:
+        """Get structured instruction parts from the most recent ModelRequest.
 
-        In the case that a "mock" request was generated to include a tool-return part for a result tool,
-        we want to use the instructions from the second-to-most-recent request (which should correspond to the
-        original request that generated the response that resulted in the tool-return part).
+        Like `_get_instructions`, falls back to the second-to-most-recent request when the most recent
```
DouweM: `_get_instructions` doesn't exist anymore, unsure if this needs saying.

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -949,56 +950,55 @@ def base_url(self) -> str | None:
...
+                second = last_two_requests[1]
+                if second.instruction_parts is not None:
+                    parts = list(second.instruction_parts)
 
-            # While it's possible that you could have a message history where the most recent request has only tool returns,
-            # I believe there is no way to achieve that would _change_ the instructions without manually crafting the most
-            # recent message. That might make sense in principle for some usage pattern, but it's enough of an edge case
-            # that I think it's not worth worrying about, since you can work around this by inserting another ModelRequest
-            # with no parts at all immediately before the request that has the tool calls (that works because we only look
-            # at the two most recent ModelRequests here).
```
DouweM, line 958: We changed this logic a lot and deleted this comment -- that doesn't seem to have been necessary? Couldn't we basically have the old logic and just make it use `InstructionPart` rather than what we had before?

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -1018,20 +1017,54 @@ async def _map_message(  # noqa: C901
...
-            return system_prompt_blocks, anthropic_messages
+        # Build system prompt blocks: each instruction part becomes a separate text block.
+        # When anthropic_cache_instructions is enabled, the cache point goes after the last
+        # static instruction (or at the end if all instructions are static).
+        cache_instructions = model_settings.get('anthropic_cache_instructions')
+
+        if instruction_parts or cache_instructions:
+            system_prompt_blocks: list[BetaTextBlockParam] = []
+
+            if system_prompt:
```
DouweM, line 1028: Should we consider the source `SystemPromptPart`s and their `dynamic_ref` as well for ordering etc? Or could that be a breaking change?

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -1018,20 +1017,54 @@ async def _map_message(  # noqa: C901
...
+                    system_prompt_blocks[cache_block_idx]['cache_control'] = self._build_cache_control(ttl)
+
+            if system_prompt_blocks:
+                return system_prompt_blocks, anthropic_messages
+
+        if system_prompt:
+            return system_prompt, anthropic_messages
 
-        return system_prompt, anthropic_messages
+        return '', anthropic_messages
```
DouweM: Is an empty string really the most appropriate?

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -1146,13 +1146,15 @@ async def _map_messages(
...
+        if instruction_parts := self._get_instruction_parts(messages, model_request_parameters):
             system_prompt_count = next(
                 (i for i, m in enumerate(openai_messages) if m.get('role') != 'system'), len(openai_messages)
             )
-            openai_messages.insert(
-                system_prompt_count, chat.ChatCompletionSystemMessageParam(content=instructions, role='system')
-            )
+            for i, part in enumerate(instruction_parts):
+                openai_messages.insert(
+                    system_prompt_count + i,
```
DouweM: This repeated next/enumerate + enumerate/insert+1 business in many models seems like we can do better. Thinking about ways to dedupe that. Note that i like the `list[i:i] = ...` way of inserting multiple items, better anyway.

tests/models/test_anthropic.py
```diff
@@ -8662,13 +8828,19 @@ async def test_anthropic_system_prompts_and_instructions_ordering():
...
+            {
+                'type': 'text',
+                'text': """\
 System prompt 1
 
-System prompt 2
-
-Instructions content\
-""")
+System prompt 2\
```
DouweM, line 8841: Interesting that these did not end up as separate type:text parts. Can you find out why?

tests/models/test_google.py
```diff
@@ -356,6 +357,9 @@ async def test_google_model_builtin_code_execution_stream(
                 ],
                 instructions='Be concise and always use Python to do calculations no matter how small.',
```
DouweM, line 358: Can we drop `instructions` from snapshots and not take it into account for comparison, as `instruction_parts` always has at least as detailed info?

DouweM: Actually, think about this for a bit: am I thinking about the `ModelRequest` aspect of this feature all wrong? Should we keep storing `instructions` as a rendered "this is what was sent" string, and instead change the agent graph / `ModelRequestNode` / models to use `list[InstructionParts]` go through `model_request_parameters` instead of trying to read if off of the message history?

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -1384,6 +1384,43 @@ def otel_message_parts(self, settings: InstrumentationSettings) -> list[_otel_me
...
+    Instructions are composed of one or more parts, each of which can be static (from a literal string)
+    or dynamic (from a function, template, or toolset). This distinction allows model implementations
+    to make intelligent caching decisions — e.g. Anthropic's prompt caching can cache the static prefix
+    while leaving dynamic instructions uncached.
+    """
+
+    content: str
+    """The text content of this instruction block."""
+
+    dynamic: bool = False
```
DouweM, line 1402: We need a `_: KW_ONLY` here

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -823,8 +835,23 @@ def prepare_request(
             ]
             params = replace(params, builtin_tools=supported_builtins, function_tools=function_tools)
 
+        params = self._append_prompted_output_to_instruction_parts(params)
```
DouweM: Put this closer to where we modify `prompted_output_template`

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -823,8 +835,23 @@ def prepare_request(
...
+
         return model_settings, params
 
+    @staticmethod
+    def _append_prompted_output_to_instruction_parts(params: ModelRequestParameters) -> ModelRequestParameters:
```
DouweM: I don't think this needs to be its own method, I'd rather have it inlined. You can do `noqa: C901` on the method to quiet the linter

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -922,12 +949,19 @@ def base_url(self) -> str | None:
...
+
+        When `model_request_parameters` is provided (normal model request flow), returns
+        the joined content of `instruction_parts` which already includes prompted output
+        instructions and is properly sorted.
 
-        In the case that a "mock" request was generated to include a tool-return part for a result tool,
-        we want to use the instructions from the second-to-most-recent request (which should correspond to the
-        original request that generated the response that resulted in the tool-return part).
+        Falls back to reading `ModelRequest.instructions` from message history when
+        `model_request_parameters` is not available (e.g. OTel span attributes).
```
DouweM, line 951: We shouldn't use an inferior method when generating OTel span attributes -- we need to match as closley as possible what's actually sent to the model. Can we pass MRP in OTel span attrs?

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -940,34 +974,13 @@ def _get_instructions(
                     instructions = message.instructions
                     break
 
-        # If we don't have two requests, and we didn't already return instructions, there are definitely not any:
         if instructions is None and len(last_two_requests) == 2:
             most_recent_request = last_two_requests[0]
             second_most_recent_request = last_two_requests[1]
 
-            # If we've gotten this far and the most recent request consists of only tool-return parts or retry-prompt parts,
```
DouweM, line 948: Restore the massive comment please

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -940,34 +974,13 @@ def _get_instructions(
                     instructions = message.instructions
                     break
 
-        # If we don't have two requests, and we didn't already return instructions, there are definitely not any:
```
DouweM, line 943: Restore the comment please

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -542,7 +550,7 @@ async def stream(
             )
             self._did_stream = True
             ctx.state.usage.requests += 1
-            skip_mrp = await _prepare_request_parameters(ctx)
+            skip_mrp = await _prepare_request_parameters(ctx, instruction_parts=None)
```
DouweM, line 556: Is `None` good enough?

PR #4867, opened 2026-03-26
Fix capability schema generation to include full parameter types

DouweM:
## Summary

- Fix `model_json_schema_with_capabilities()` to generate full parameter schemas for all capabilities, not just string forms
- When `from_spec` is not overridden, fall back to inspecting `__init__` for schema generation
- Add `from_spec` overrides to `WebSearch`, `WebFetch`, `MCP`, and `ImageGeneration` exposing only serializable spec-relevant params
- Add `self`/`cls` parameter skipping in `build_schema_types` to support `__init__` as schema target
- Replace existing schema test with full `snapshot()` assertion

**Before:** Thinking, WebSearch, WebFetch, MCP, ImageGeneration only appeared as string constants in the schema (e.g. `"Thinking"`)
**After:** Full parameter schemas are generated matching the documented spec syntax:
- `Thinking: high` / `{Thinking: {effort: high}}`
- `WebSearch: {search_context_size: high, blocked_domains: [...]}`
- `MCP: https://example.com` / `{MCP: {url: ..., id: ..., allowed_tools: [...]}}`
- etc.

## Test plan

- [x] All 277 `test_capabilities.py` tests pass
- [x] Snapshot test captures full schema output for regression detection
- [x] Pre-commit hooks pass (format, lint, typecheck)

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/capabilities/mcp.py
+++ b/pydantic_ai_slim/pydantic_ai/capabilities/mcp.py
@@ -15,7 +15,7 @@ from .builtin_or_local import BuiltinOrLocalTool
 try:
     from pydantic_ai.mcp import MCPServer
     from pydantic_ai.toolsets.fastmcp import FastMCPToolset
-except ImportError:  # pragma: no branch — mcp is an optional dependency
+except ImportError:  # pragma: lax no cover
     if not TYPE_CHECKING:
         MCPServer = Any  # type: ignore[assignment,misc]
         FastMCPToolset = Any  # type: ignore[assignment,misc]


pydantic_ai_slim/pydantic_ai/agent/spec.py
```diff
@@ -325,7 +325,24 @@ def load_capability_from_nested_spec(spec: dict[str, Any] | str) -> AbstractCapa
 
 def _build_capability_schema_types(registry: Mapping[str, type[Any]]) -> list[Any]:
     """Build a list of schema types for capabilities from a registry."""
+    from pydantic_ai._utils import get_function_type_hints
```
DouweM: Move imports to the top; this one and the one below

pydantic_ai_slim/pydantic_ai/_spec.py
```diff
@@ -224,17 +224,61 @@ def load_from_registry(
...
+
+    For Union types, removes non-serializable members (TypeVars, Callables).
+    Returns None if the type is entirely non-serializable.
+    """
+    return _filter_serializable_type(tp)
+
+
+def _filter_serializable_type(tp: object) -> object | None:
+    import types
+    import typing
```
DouweM: Imports at the top please

tests/test_capabilities.py
```diff
@@ -125,6 +164,25 @@ class CustomCapability(AbstractCapability[None]):
     greeting: str = 'hello'
 
 
+@dataclass(init=False)
+class CapabilityWithCallbackParam(AbstractCapability[None]):
+    """Custom capability with a mix of serializable and non-serializable params."""
+
+    max_retries: int
+    on_error: Any  # Callable — not serializable, will be filtered from schema
+    verbose: Any  # Callable | bool in __init__ — only bool survives in schema
```
DouweM: Can this actually have the callable type, and we use the dataclass's autogenerated init method? Or do we _need_ a custom init?

tests/test_capabilities.py
```diff
@@ -360,39 +418,548 @@ class ExtraCap(AbstractCapability[None]):
...
+                },
+                'spec_params_ImageGeneration': {
+                    'additionalProperties': False,
+                    'properties': {
+                        'builtin': {
+                            'anyOf': [
+                                {'$ref': '#/$defs/ImageGenerationTool'},
+                                {
+                                    'oneOf': [
+                                        {'$ref': '#/$defs/WebSearchTool'},
```
DouweM: The `ImageGeneration` capability should only accept `ImageGenerationTool` for `builtin` (and bool), not all the other builtin tool types. Same for the other capabilities for specific builtin tool types -- they should only support the intended one.

tests/test_capabilities.py
```diff
@@ -360,39 +418,548 @@ class ExtraCap(AbstractCapability[None]):
...
+                            {'const': 'WebSearch', 'type': 'string'},
+                            {'$ref': '#/$defs/spec_WebSearch'},
+                        ]
+                    },
+                    'title': 'Capabilities',
+                    'type': 'array',
+                },
+                '$schema': {'type': 'string'},
+            },
+            'title': '_AgentSpecSchema',
```
DouweM: Can we make this title 'AgentSpec'?

pydantic_ai_slim/pydantic_ai/_spec.py
```diff
@@ -7,6 +7,8 @@
 from __future__ import annotations
 
 import inspect
+import types  # used at runtime in filter_serializable_type
+import typing  # used at runtime in filter_serializable_type
```
DouweM: We don't need these comments

pydantic_ai_slim/pydantic_ai/_spec.py
```diff
@@ -224,17 +226,54 @@ def load_from_registry(
...
+        return None
+
+    return tp
+
+
 def build_schema_types(
     registry: Mapping[str, type[Any]],
     *,
     get_schema_target: Callable[[type[Any]], Any] | None = None,
+    filter_type_hint: Callable[[Any], Any | None] | None = None,
```
DouweM: Do we need an arg? I think we could just always do this -- we know that if it's not serializable, we'd fail here right?

pydantic_ai_slim/pydantic_ai/capabilities/mcp.py
```diff
@@ -65,6 +67,30 @@ def __init__(
         self.description = description
         self.__post_init__()
 
+    @classmethod
+    def from_spec(
```
DouweM: Do we still need this method? Now that we handle unserializable type hint scorrectly

PR #4869, opened 2026-03-26
fix(vercel-ai): strip retry suffix from UI `errorText` and preserve LLM cache

tijmenhammer:
## Summary

`RetryPromptPart.model_response()` appends `"\n\nFix the errors and try again."` — an instruction meant for the LLM, not for end users. This suffix was leaking into Vercel AI `errorText` in both the streaming and dump paths, breaking frontend logic that checks `errorText` values (e.g. `errorText === "Cancelled"` to distinguish cancelled vs failed state).

**Core change:** Add `RetryPromptPart.error_description()` — returns the error text without the retry suffix. `model_response()` now delegates to it, so all existing callers are unaffected.

**Vercel AI adapter:**
- **Dump path:** Uses `error_description()` for UI-facing `error_text`; stores the full `model_response()` in `call_provider_metadata.pydantic_ai.model_response` for round-trip cache fidelity
- **Load path:** Prefers `provider_meta['model_response']` for `ToolReturnPart.content`, falls back to `error_text` — LLM prompt cache is preserved after dump → load
- **Streaming path:** Uses `error_description()` directly

This ensures the LLM sees identical text before and after a dump/load round-trip (no cache break), while the UI gets clean error text without the retry instruction.

## Test plan

- [x] All 106 `test_vercel_ai.py` tests pass
- [x] Snapshot tests updated for new metadata field and round-trip content
- [x] `ruff check` / `ruff format` / `pyright` clean on changed files
- [ ] Verify errorText in Vercel AI frontend no longer contains suffix
- [ ] Verify dump → load → next LLM turn produces identical prompt (cache hit)

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/messages.py
+++ b/pydantic_ai_slim/pydantic_ai/messages.py
@@ -1303,20 +1303,25 @@ class RetryPromptPart:
     part_kind: Literal['retry-prompt'] = 'retry-prompt'
     """Part type identifier, this is available on all parts as a discriminator."""
 
-    def model_response(self) -> str:
-        """Return a string message describing why the retry is requested."""
+    def error_description(self) -> str:
+        """Return the error description without the retry instruction suffix.
+
+        This is suitable for UI display where the LLM-facing "Fix the errors and try again."
+        instruction is not appropriate. For the full model-facing text, use [`model_response()`][pydantic_ai.messages.RetryPromptPart.model_response].
+        """
         if isinstance(self.content, str):
             if self.tool_name is None:
-                description = f'Validation feedback:\n{self.content}'
+                return f'Validation feedback:\n{self.content}'
             else:
-                description = self.content
+                return self.content
         else:
             json_errors = error_details_ta.dump_json(self.content, exclude={'__all__': {'ctx'}}, indent=2)
             plural = isinstance(self.content, list) and len(self.content) != 1
-            description = (
-                f'{len(self.content)} validation error{"s" if plural else ""}:\n```json\n{json_errors.decode()}\n```'
-            )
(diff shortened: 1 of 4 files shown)


dsfaccini:

Thanks @tijmenhammer. The bug is real: `RetryPromptPart.model_response()` appends "Fix the errors and try again.", and that leaks into the Vercel AI `errorText` on both the streaming path (`_event_stream.py`) and the dump path (`_adapter.py`), which is wrong for UI display.

But we don't want to fix it this way. This is one instance of a broader gap: the framework has no model-neutral way to mark "the framework wrote this part", and @DouweM has a direction for it that this PR's mechanism cuts against: https://github.com/pydantic/pydantic-ai/issues/6404#issuecomment-4984558608

Two specifics:

- Storing the full `model_response()` (suffix included) in `call_provider_metadata` and preferring it on load is adapter-level bookkeeping of presentation content. The direction is provenance-as-data on the part, persisted model-neutrally and rendered per-profile at `Model.prepare_messages`, not baked into content or an adapter's metadata. Baking presentation into persisted history also breaks the prompt-cache-prefix determinism we're building elsewhere.
- The `RetryPromptPart.model_response()` split here collides with the first step already in flight: https://github.com/pydantic/pydantic-ai/pull/6571 (a cause discriminator plus cause-appropriate wording, since "fix the errors" is wrong when nothing failed). The Vercel `errorText` rendering falls out of the renderer change once provenance lands.

Closing in favor of that line of work. If you want to keep helping here, https://github.com/pydantic/pydantic-ai/issues/6404 and https://github.com/pydantic/pydantic-ai/pull/6571 are the place. Appreciate the report and the clean reproduction.

PR #4876, opened 2026-03-27
Make run ID time-sortable; add agent name and run ID span attrs to all agent run child spans

adriangb:


pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -98,21 +99,21 @@
     from ..ui._web import ModelsParam
 
 __all__ = (
+    'AbstractAgent',
```
adriangb, line 103: drive by formatting change I guess, must have been a newer version of ruff or something?

pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -1271,6 +1272,7 @@ async def get_instructions(run_context: RunContext[AgentDepsT]) -> str | None:
             'model_name': model_used.model_name if model_used else 'no-model',
             'agent_name': agent_name,
             'gen_ai.agent.name': agent_name,
+            'gen_ai.conversation.id': state.run_id,
```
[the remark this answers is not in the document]
alexmojaki: Yeah I'm not OK with putting the wrong value here and having a TODO to change it later. Rather make up a new attribute.

pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -1283,176 +1287,187 @@ async def get_instructions(run_context: RunContext[AgentDepsT]) -> str | None:
...
-                    async def _finalize_result(r: AgentRunResult[Any]) -> None:
-                        """Call after_run, store the result override, and clear any pending error."""
-                        nonlocal _run_error
-                        r = await run_capability.after_run(run_ctx, result=r)
-                        agent_run._result_override = r  # pyright: ignore[reportPrivateUsage]
-                        _run_error = None
+            async with AsyncExitStack() as stack:
+                if run_span.is_recording():
+                    stack.enter_context(
+                        logfire_api.set_baggage(
```
alexmojaki: pydantic-ai doesn't directly use logfire_api, it should stay that way. use manual otel please.

pydantic_ai_slim/pydantic_ai/agent/__init__.py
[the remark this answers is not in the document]
alexmojaki: Also, the run_id is the thing we actually want, right? So we wouldn't be able to change the semantics of this in the future without maybe breaking things.

pydantic_ai_slim/pydantic_ai/agent/__init__.py
alexmojaki: if it's important for the model calls to have these attributes, then they should be set directly on the spans, because baggage isn't necessarily set as attributes, that can be disabled.

DouweM: @alexmojaki `pydantic_graph` already uses `logfire_api` and pydantic-ai-slim depends on pydantic-graph, so I don't have a problem making pydantic-ai-slim explicitly depend on logfire-api as well. Is there a reason other than dependencies that we should use manual otel instead?

DouweM: @alexmojaki Good point, "baggage as span attrs" seems to be a Logfire SDK thing (that can be opted out of)?

It's not trivial to route the agent name and run ID through to the model request / tool call instrumentation, but I suppose we could make it work using new context vars.

alexmojaki: It's not necessarily a big deal, I overstated, but it's kind of a weird quirk that logfire-api is a dependency of pydantic-graph, it's barely helpful over raw otel. it'd be nice to not make it harder to remove one day.

alexmojaki: @adriangb the only place you intend on using these attributes is on the model request spans, right? not tool calls?

@DouweM baggage will still be present, essentially as a context var, even if it doesn't set attributes, so you could use that instead of a new context var.

adriangb: I could see it being useful in tool calls. The conversation id especially I feel like would be nice to have on every child span of an agent run (what baggage does if configured as we do by default in logfire).

> @DouweM baggage will still be present, essentially as a context var, even if it doesn't set attributes, so you could use that instead of a new context var.

I guess if we double-set an attribute it's a non-issue, so we could do it manually in the places we really care about so that even if users don't use logfire they get those set?

pydantic_ai_slim/pydantic_ai/_uuid.py
```diff
@@ -0,0 +1,84 @@
...
+
+    int_uuid_7 = unix_ts_ms << 80
+    int_uuid_7 |= counter_hi << 64
+    int_uuid_7 |= counter_lo << 32
+    int_uuid_7 |= tail & 0xFFFF_FFFF
+    int_uuid_7 |= _RFC_4122_VERSION_7_FLAGS
+
+    # defer global update until all computations are done
+    _last_timestamp_v7 = timestamp_ms
+    _last_counter_v7 = counter
```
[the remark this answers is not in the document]
adriangb: we could just use a backport? maybe https://github.com/aminalaee/uuid-utils?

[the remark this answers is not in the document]
DouweM: @adriangb I'm afraid of new packages for single functions because of supply chain attack risk

PR #4877, opened 2026-03-27
fix(messages): fall back to stripped model name for cost calculation

### LINKED ISSUE #4839: operation.cost missing for OpenRouter with OpenAIChatModel + LiteLLMProvider
### Initial Checks

### Description

We are using PydanticAI instrumentation with `instrument=True` and sending requests to OpenRouter through:

- `OpenAIChatModel`
- `LiteLLMProvider`

The request succeeds, but `operation.cost` is missing from our traces. Token counts and response model are available. 

We are not using Pydantic Logfire, so we do not have a Logfire link to share. We are using Loki + Tempo.

From the trace, the response model (when we request `openai/gpt-5.2`) is:

```json
"gen_ai.response.model": "openai/gpt-5.2-20251211"
```

It looks like `response.cost()` uses `ModelResponse.model_name` directly, and this OpenRouter model name is not normalized before being passed to pricing.

From reading the code, the failure seems to happen when `ModelResponse.cost()` calls `genai_prices.calc_price(...)` with `model_name="openai/gpt-5.2-20251211"` and `provider_id="litellm"`. That model name does not appear to match the bundled pricing aliases directly, so pricing lookup fails and PydanticAI skips `operation.cost`.

**Minimal example**

```python
from pydantic_ai import Agent

provider = LiteLLMProvider(
    api_base="https://openrouter.ai/api/v1",
    api_key="...",
)

model = OpenAIChatModel(
    model_name="openai/gpt-5.2",
    provider=provider,
)

agent = Agent(model=model, instrument=True)
result = await agent.run("say hello")
```

**Workaround**

We were able to make cost reporting work by normalizing the returned response model before pricing, e.g.:

- `openai/gpt-5.2-20251211` -> `gpt-5.2-2025-12-11`

Would it make sense for PydanticAI to handle this normalization internally for the OpenRouter/LiteLLM path so `operation.cost` is emitted automatically?

```Python
**Minimal example**


from pydantic_ai import Agent

provider = LiteLLMProvider(
    api_base="https://openrouter.ai/api/v1",
    api_key="...",
)

model = OpenAIChatModel(
    model_name="openai/gpt-5.2",
    provider=provider,
)

agent = Agent(model=model, instrument=True)
result = await agent.run("say hello")
```

### Logfire Trace

_No response_

- Python: 3.13.12
- Pydantic AI: 1.70.0 (and also 1.23.0)
- LLM provider SDK: litellm==1.80.0

JasonCZMeng:
## Summary

When providers like OpenRouter return model names with a date suffix (e.g. `openai/gpt-5.2-20251211`), `genai-prices` doesn't recognize them and `ModelResponse.cost()` raises `LookupError`. This causes `operation.cost` to be missing from traces.

## Fix

Added a fallback in `ModelResponse.cost()`: when the initial `calc_price` lookup fails, strip the trailing `-YYYYMMDD` date suffix from the model name and retry. The regex `-\d{8}$` matches exactly 8-digit date suffixes at the end of the string.

The fallback chain is now:
1. Try with `provider_api_url` (most specific)
2. Try with `provider_id` 
3. Strip date suffix → try with `provider_api_url`
4. Strip date suffix → try with `provider_id`
5. Raise `LookupError`

## Tests

- `test_strip_model_date_suffix`: Unit test for the suffix stripping helper
- `test_model_response_cost_with_date_suffix`: Integration test verifying `ModelResponse.cost()` works with date-suffixed model names

Fixes #4839

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/messages.py
+++ b/pydantic_ai_slim/pydantic_ai/messages.py
@@ -1771,46 +1771,37 @@ class ModelResponse:
         Uses [`genai-prices`](https://github.com/pydantic/genai-prices).
         """
         assert self.model_name, 'Model name is required to calculate price'
-        # Try matching on provider_api_url first as this is more specific, then fall back to provider_id.
-        if self.provider_url:
-            try:
-                return calc_price(
-                    self.usage,
-                    self.model_name,
-                    provider_api_url=self.provider_url,
-                    genai_request_timestamp=self.timestamp,
-                )
-            except LookupError:
-                pass
-        try:
-            return calc_price(
-                self.usage,
-                self.model_name,
-                provider_id=self.provider_name,
-                genai_request_timestamp=self.timestamp,
-            )
-        except LookupError:
-            pass
-        # Some providers (e.g. OpenRouter) return model names with a date suffix
-        # like "openai/gpt-5.2-20251211". Strip the suffix and retry.
-        normalized = _strip_model_date_suffix(self.model_name)
-        if normalized != self.model_name:
+
+        def _try_calc(model_name: str) -> genai_types.PriceCalculation | None:
+            """Try calculating price with provider_api_url first, then provider_id."""
             if self.provider_url:
                 try:
                     return calc_price(
                         self.usage,
-                        normalized,
+                        model_name,
                         provider_api_url=self.provider_url,
                         genai_request_timestamp=self.timestamp,
                     )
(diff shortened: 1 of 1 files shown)


DouweM:
I think this would be more appropriately addressed in genai-prices.

Also I believe the tests are not realistic, because the `model_name` would be `openai/gpt-5.2-20251211` and provider `openrouter`

PR #4884, opened 2026-03-28
fix: support custom thread executor to prevent thread accumulation in long-running servers

guoyangzhen:
## Problem

In long-running servers (FastAPI, uvicorn, etc.), the `run_in_executor` function in `_utils.py` uses `anyio.to_thread.run_sync()` which creates temporary threads for each call. These threads are kept alive for 10 seconds after each use before being cleaned up.

Under sustained traffic, thread creation outpaces reclamation. In production, we observed thread counts growing from ~30 to 800+ over 8 hours of continuous operation, leading to increased memory usage and context switching overhead.

## Solution

This PR adds `set_thread_pool_executor()` and `get_thread_pool_executor()` functions to `_utils.py`, allowing users to provide a fixed-size `ThreadPoolExecutor`.

When a custom executor is set, `run_in_executor` uses `loop.run_in_executor(executor, func)` instead of `anyio.to_thread.run_sync()`, ensuring thread count stays bounded.

When no executor is set (default), behavior is unchanged.

## Usage

```python
from concurrent.futures import ThreadPoolExecutor
from pydantic_ai._utils import set_thread_pool_executor

# At application startup
set_thread_pool_executor(ThreadPoolExecutor(max_workers=32))

# ... application runs ...

# At shutdown
set_thread_pool_executor(None)
```

## Changes

- `pydantic_ai_slim/pydantic_ai/_utils.py`: Added `set_thread_pool_executor()`, `get_thread_pool_executor()`, and modified `run_in_executor()` to use the custom executor when set
- `tests/test_utils.py`: Added 4 tests covering custom executor usage, kwargs support, default fallback, and restore-to-default behavior

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/_utils.py
+++ b/pydantic_ai_slim/pydantic_ai/_utils.py
@@ -7,6 +7,7 @@ import re
 import time
 import uuid
 from collections.abc import AsyncIterable, AsyncIterator, Awaitable, Callable, Iterable, Iterator
+from concurrent.futures import ThreadPoolExecutor
 from contextlib import asynccontextmanager, contextmanager, suppress
 from contextvars import ContextVar
 from dataclasses import dataclass, fields, is_dataclass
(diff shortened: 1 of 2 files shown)


guoyangzhen:

PR #4887, opened 2026-03-29
fix(openai): skip orphaned tool returns in Responses API message mapping

### LINKED ISSUE #4882: `FallbackModel` produces invalid Responses API requests when mixing `OpenAIChatModel` and `OpenAIResponsesModel` during multi-step tool loops
### Initial Checks

### Description

When using `FallbackModel` with an `OpenAIChatModel` primary (e.g., DeepSeek) and an `OpenAIResponsesModel` fallback (e.g., GPT-5-nano), and the agent performs multi-step tool calls via `agent.iter()`, the fallback model produces invalid Responses API requests with orphaned `function_call_output` items.

**What happens:**

1. Primary model (`OpenAIChatModel` via DeepSeek provider, 131K context limit) makes tool calls over multiple iterations within `agent.iter()`
2. History processors run on every model request (`_agent_graph.py:521`) and trim accumulated messages to stay within token budget
3. Primary model eventually exceeds its context limit → `ModelHTTPError` (400)
4. `FallbackModel` catches the error and tries the fallback (`OpenAIResponsesModel` / GPT-5-nano)
5. The fallback model's `_map_messages()` converts the trimmed internal messages to Responses API format
6. OpenAI rejects the request: `"No tool call found for function call output with call_id call_00_xxx"`
7. Both models fail → `FallbackExceptionGroup` → unhandled crash

**Root cause:**

History processors (user-supplied) run before `_clean_message_history()` (`_agent_graph.py:530`). The `_clean_message_history` function merges consecutive `ModelRequest` messages **after** user processors have executed. This means:

- User-supplied sanitization (stripping orphaned `ToolReturnPart`s without matching `ToolCallPart`s) runs first and produces a clean history
- `_clean_message_history` then merges consecutive `ModelRequest`s, potentially reorganizing `ToolReturnPart`s relative to their matching `ModelResponse`
- The Responses API model's `_map_messages()` linearly converts the merged result, producing `function_call_output` items whose matching `function_call` items were trimmed or reordered

This is only triggered when mixing API formats in `FallbackModel` because:
- `OpenAIChatModel` (Chat Completions) is lenient about tool call/result ordering
- `OpenAIResponsesModel` (Responses API) strictly validates that every `function_call_output` has a preceding `function_call` with the same `call_id`

**Expected behavior:**

`FallbackModel` should produce valid requests for each fallback model, regardless of which model types are mixed. Either:
1. `FallbackModel` could run an internal orphan sanitization pass before sending messages to each fallback model (after `_clean_message_history`)
2. Or `_clean_message_history` could preserve tool call/return pairing invariants that user processors established

**Related:** #4728 (built-in `HistoryProcessor` for orphaned tool call/result repair) — addresses the user-side sanitization but doesn't solve the post-`_clean_message_history` gap or the `FallbackModel`-specific trigger.

```python
from pydantic_ai import Agent
from pydantic_ai.models.fallback import FallbackModel
from pydantic_ai.models.openai import OpenAIChatModel, OpenAIResponsesModel

# Primary: DeepSeek via Chat Completions API (131K context limit)
primary = OpenAIChatModel("deepseek-chat", provider=DeepSeekProvider(api_key="..."))
# Fallback: GPT-5-nano via Responses API
fallback = OpenAIResponsesModel("gpt-5-nano", provider=OpenAIProvider(api_key="..."))

model = FallbackModel(primary, fallback)


# Custom history processors to manage context size
async def trim_to_budget(messages):
    """Trim oldest messages to stay within token budget."""
    # ... token counting and trimming logic ...
    return messages

async def sanitize_orphans(messages):
    """Strip ToolReturnParts without matching ToolCallParts."""
    # ... orphan detection and removal ...
    return messages


agent = Agent(
    model=model,
    history_processors=[trim_to_budget, sanitize_orphans],
)


@agent.tool
async def get_large_data(ctx, query: str) -> dict:
    """Tool that returns a large payload (~3-5K tokens per call)."""
    return {"results": [{"value": i} for i in range(500)]}


# Register many tools (40+) to increase tool definition token overhead.
# With enough conversation history + sequential tool iterations,
# the primary model overflows its context limit.
#
# Reproduction flow:
# 1. Agent calls tools sequentially, accumulating results in context
# 2. History processors trim between iterations, but token estimation
#    underestimates actual usage
# 3. Primary model (DeepSeek) hits 131K limit → 400 error
# 4. FallbackModel tries fallback (GPT-5-nano via Responses API)
# 5. Responses API rejects: "No tool call found for function call output"
# 6. FallbackExceptionGroup raised — both models failed

async with agent.iter(
    "Run comprehensive analysis",
    message_history=existing_conversation_history,
) as run:
    async for node in run:
        if agent.is_model_request_node(node):
            async with node.stream(run.ctx) as stream:
                async for event in stream:
                    pass
        elif agent.is_call_tools_node(node):
            async with node.stream(run.ctx) as stream:
                async for event in stream:
                    pass
```

**Error from OpenAI Responses API (on the fallback model):**
```
openai.BadRequestError: Error code: 400 - {
  'error': {
    'message': "No tool call found for function call output with call_id call_00_mItQaYpHwe09QGVdxERTW7y6.",
    'param': 'input',
    'code': None
  }
}
```

**Error from primary model (DeepSeek) that triggered the fallback:**
```
openai.BadRequestError: Error code: 400 - {
  'error': {
    'message': "This model's maximum context length is 131072 tokens. However, you requested 133351 tokens (133351 in the messages, 0 in the completion).",
    'type': 'invalid_request_error'
  }
}
```

### Logfire Trace

_No trace available — error originates from OpenAI's Responses API rejecting the converted input messages._

- Python: 3.13.11
- Pydantic AI: 1.62.0
- LLM provider SDK: openai 2.17.0

JasonCZMeng:
## Summary

When using `FallbackModel` with mixed model types (e.g. `OpenAIChatModel` primary + `OpenAIResponsesModel` fallback), history processors that trim messages can create orphaned tool returns — `ToolReturnPart`s whose matching `ToolCallPart` in a `ModelResponse` was removed. The Responses API strictly requires every `function_call_output` to have a preceding `function_call`, causing 400 errors like:

> No tool call found for function call output with call_id call_00_xxx

## Fix

Added a pre-pass in `OpenAIResponsesModel._map_messages()` that collects all known tool call IDs from `ModelResponse` messages, then silently skips any `ToolReturnPart` or `RetryPromptPart` whose `call_id` is not in the set. The guard `if known_tool_call_ids and call_id not in known_tool_call_ids` ensures backward compatibility — when no tool calls exist in history (e.g. first turn), nothing is skipped.

## Test

`test_openai_responses_orphaned_tool_return_skipped`: Creates a message history with an orphaned `ToolReturnPart` (no matching `ModelResponse` with that tool call ID) and verifies the agent completes successfully without 400 errors.

Fixes #4882

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/openai.py
+++ b/pydantic_ai_slim/pydantic_ai/models/openai.py
@@ -1970,8 +1970,12 @@ class OpenAIResponsesModel(Model):
                     elif isinstance(part, ToolReturnPart):
                         call_id = _guard_tool_call_id(t=part)
                         call_id, _ = _split_combined_tool_call_id(call_id)
-                        # Skip orphaned tool returns whose matching tool call was trimmed
-                        if call_id not in known_tool_call_ids:
+                        # Skip orphaned tool returns whose matching tool call was trimmed.
+                        # The `known_tool_call_ids` guard ensures we only filter when there
+                        # are ModelResponse messages in the history — when the set is empty
+                        # (e.g. first turn or previous_response_id='auto' mode), all returns
+                        # are passed through.
+                        if known_tool_call_ids and call_id not in known_tool_call_ids:
                             continue
                         output = await self._map_tool_return_output(part)
                         item = FunctionCallOutput(
@@ -1989,7 +1993,7 @@ class OpenAIResponsesModel(Model):
                             call_id = _guard_tool_call_id(t=part)
                             call_id, _ = _split_combined_tool_call_id(call_id)
                             # Skip orphaned retry prompts whose matching tool call was trimmed
-                            if call_id not in known_tool_call_ids:
+                            if known_tool_call_ids and call_id not in known_tool_call_ids:
                                 continue
                             item = FunctionCallOutput(
                                 type='function_call_output',
--- a/tests/models/test_openai_responses.py
+++ b/tests/models/test_openai_responses.py
@@ -10171,16 +10171,27 @@ async def test_openai_responses_orphaned_tool_return_skipped(allow_model_request
     agent = Agent(m)
 
     # History with an orphaned tool return — the ModelResponse containing the
-    # matching ToolCallPart has been trimmed (e.g. by a history processor).
+    # matching ToolCallPart for 'call_orphan_123' was trimmed by a history
+    # processor, but a different tool call/return pair survived.
     orphaned_history = [
         ModelRequest(
-            parts=[UserPromptPart(content='Call the tool')],
+            parts=[UserPromptPart(content='Call tools')],
         ),
-        # NOTE: no ModelResponse with ToolCallPart for 'call_orphan_123'
+        ModelResponse(
+            parts=[
+                ToolCallPart(tool_name='surviving_tool', args='{}', tool_call_id='call_surviving_456'),
+                # NOTE: no ToolCallPart for 'call_orphan_123' — it was trimmed
+            ],
+        ),
         ModelRequest(
             parts=[
                 ToolReturnPart(
-                    tool_name='some_tool',
+                    tool_name='surviving_tool',
+                    content='ok',
+                    tool_call_id='call_surviving_456',
+                ),
+                ToolReturnPart(
+                    tool_name='orphaned_tool',
                     content='result',
                     tool_call_id='call_orphan_123',
                 ),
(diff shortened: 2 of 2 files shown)


DouweM:
I don't think this is the right solution, let's discuss on the issue.

PR #4906, opened 2026-03-30
feat: add local `WebFetch` tool and have `WebFetch` capability use it when provider lacks builtin support

### LINKED ISSUE #4905: Web Fetch built-in fallback
### Description

The `WebFetch` capability currently has no default local fallback. When using a model that doesn't natively support `WebFetchTool` (e.g. OpenAI, Groq, Mistral), it raises a `UserError` with no workaround other than manually passing `local=`.

This is inconsistent with `WebSearch`, which automatically falls back to DuckDuckGo when the model doesn't support builtin web search.

Currently only Anthropic and Google support `WebFetchTool` natively ([provider support table](https://ai.pydantic.dev/builtin-tools/#provider-support_5)).

### Proposed Solution

Add a default `httpx`-based local fallback to `WebFetch._default_local()`, following the same pattern as `WebSearch._default_local()` with DuckDuckGo:

- Use `pydantic_ai._ssrf.safe_download()` for SSRF-protected HTTP fetching
- Use `markdownify` (as a new `web-fetch` optional dependency) for HTML-to-markdown conversion
- Handle non-HTML responses gracefully (JSON → code block, plain text → raw, binary → error message)
- Fall back to `None` when `markdownify` is not installed (same as WebSearch when `ddgs` is missing)

This was discussed and approved by @douwemaan on Slack.

### Affected Code

- [`WebFetch` capability](https://ai.pydantic.dev/api/capabilities/#pydantic_ai.capabilities.WebFetch) — needs `_default_local()` override
- New `pydantic_ai.common_tools.web_fetch` module — the local fallback implementation

### References

https://github.com/vstorm-co/pydantic-deepagents/issues/51

DEENUU1:
- Closes #4905

### Summary

Add a default `httpx`-based local fallback to the `WebFetch` capability, so models that don't natively support `WebFetchTool` (e.g. OpenAI) can still fetch URLs instead of raising `UserError`.

This mirrors how `WebSearch` already falls back to DuckDuckGo — the new fallback uses `pydantic_ai._ssrf.safe_download` for SSRF protection and `markdownify` for HTML-to-markdown conversion.

### Changes

**New: `pydantic_ai/common_tools/web_fetch.py`**
- `WebFetchLocalTool` — async callable that fetches a URL via `safe_download` and converts the response based on content type:
  - `text/html` → markdown (via `markdownify`, with `<script>`/`<style>`/`` stripped)
  - `application/json` → pretty-printed JSON code block
  - `text/*` → raw text
  - binary → error message
- `web_fetch_tool()` factory — returns a `Tool(name='web_fetch', ...)` with configurable `max_content_length` (default 50k chars), `allow_local_urls`, and `timeout`
- `WebFetchResult` TypedDict — structured return with `url`, `title`, `content`

**Modified: `pydantic_ai/capabilities/web_fetch.py`**
- Added `_default_local()` override (same pattern as `WebSearch._default_local()`)
- Falls back gracefully to `None` when `markdownify` is not installed

**New optional dependency**
- `web-fetch` extra group: `markdownify>=1.2`

**Documentation**
- Added "Web Fetch Tool" section to `docs/common-tools.md`
- Updated `docs/capabilities.md` table and code example
- Added `web-fetch` to `docs/install.md` optional groups list

### Pre-Review Checklist

### Pre-Merge Checklist

---

Context: Found while building [`pydantic-deepagents`](https://github.com/vstorm-co/pydantic-deepagents) at [Vstorm](https://vstorm.co) · [github.com/vstorm-co](https://github.com/vstorm-co).


DouweM (commented):
*Review by Claude (Opus 4.6), on behalf of @DouweM.*

Overall approach is solid — httpx + markdownify is the right call, and the implementation follows the established `WebSearch`/DuckDuckGo pattern well. A few things to address:

## Must fix

### 1. Test file naming: `test_web_fetch_tool.py` → `test_web_fetch.py`

Per project convention (`test_{module}.py` matching source module), this should be `test_web_fetch.py` — same as `test_tavily.py` maps to `common_tools/tavily.py`. The bot review flagged this but it wasn't addressed.

### 2. Missing end-to-end agent test

`TestWebSearchCapability` has `test_websearch_default_with_nonsupporting_model` that creates a `FunctionModel` with `supported_builtin_tools=frozenset()` and verifies the DuckDuckGo fallback actually runs through an agent. `TestWebFetchCapability` has no equivalent — `test_webfetch_default` only checks `cap.local is not None`, it never runs an agent. Need a test like:

```python
def test_webfetch_default_with_nonsupporting_model(self, allow_model_requests: None):
    """WebFetch() with non-supporting model → markdownify fallback used."""
    def model_fn(messages, info):
        for msg in messages:
            for part in msg.parts:
                if isinstance(part, ToolReturnPart):
                    return ModelResponse(parts=[TextPart(content=f'Tool result: {part.content}')])
        if info.function_tools:
            return ModelResponse(
                parts=[ToolCallPart(tool_name=info.function_tools[0].name,
                       args='{"url": "https://example.com"}', tool_call_id='c1')]
            )
        return ModelResponse(parts=[TextPart(content='no tools')])

    model = FunctionModel(model_fn, profile=ModelProfile(supported_builtin_tools=frozenset()))
    agent = Agent(model, capabilities=[WebFetch()])
    result = agent.run_sync('fetch example.com')
    assert 'Tool result' in result.output
```

(with `safe_download` mocked to return a simple HTML response)

Also missing the negative counterpart: `test_webfetch_local_false_with_nonsupporting_model` (like WebSearch has).

### 3. Missing `BuiltinOrLocalTool` contract tests

WebSearch has a thorough suite testing the base class contract: `test_websearch_builtin_false`, `test_websearch_both_false_raises`, `test_websearch_builtin_false_with_constraints_raises`, `test_websearch_local_callable`. WebFetch only has 2 tests in `TestWebFetchCapability`. These are mostly exercising base class behavior, but they serve as regression tests for the specific capability and are cheap to add.

## Should fix

### 4. XHTML should go through markdownify

`application/xhtml+xml` is in `_TEXTUAL_CONTENT_TYPES` (so it's not treated as binary), but it falls through to the `else: content = text` branch and returns raw XHTML. Since XHTML is structurally HTML, it should get the markdownify treatment:

```python
if not content_type or content_type in ('text/html', 'application/xhtml+xml'):
```

## Nit

### 5. `WebFetchResult` not exported

`__all__` only exports `web_fetch_tool`, but `WebFetchResult` is the return type that users would want for type annotations if they're using the tool directly. Worth considering adding it to `__all__`.

## Looks good

- `safe_download` import moved to module level ✓
- Regex compiled at module level ✓
- Private helper tests refactored through public API ✓
- `_default_local()` pattern matches WebSearch exactly ✓
- `_requires_builtin()` correctly gates domain/usage constraints ✓
- Graceful `ImportError` fallback when markdownify isn't installed ✓
- Docs changes are clean and consistent ✓
- Dependency setup (optional group + dev deps) is correct ✓
- SSRF protection via `safe_download`, `allow_local_urls=False` default, content length limit ✓

DouweM:
*Follow-up from Claude (Opus 4.6), on behalf of @DouweM.*

## Additional feedback: forward `allowed_domains`/`blocked_domains` to the local tool

The general principle for capabilities is: the user configures what they want on the capability, and we handle it appropriately regardless of whether the builtin or local path is used.

Currently `allowed_domains` and `blocked_domains` are gated by `_requires_builtin()`, meaning `WebFetch(blocked_domains=['evil.com'])` raises `UserError` if the model doesn't support the builtin — even though domain filtering is trivially implementable locally (just check the URL's domain before fetching).

**Suggested changes:**

1. Add `allowed_domains` and `blocked_domains` params to `WebFetchLocalTool` / `web_fetch_tool()`
2. Validate the URL's domain in `__call__` before fetching — raise `ModelRetry` on violation so the LLM can try a different URL
3. Forward them from `WebFetch._default_local()`:
   ```python
   def _default_local(self):
       try:
           from pydantic_ai.common_tools.web_fetch import web_fetch_tool
           return web_fetch_tool(
               allowed_domains=self.allowed_domains,
               blocked_domains=self.blocked_domains,
           )
       except ImportError:
           return None
   ```
4. Remove `allowed_domains` and `blocked_domains` from `_requires_builtin()` (only `max_uses` would remain — that needs cross-call state)

---

*Side note:* `WebSearch` has the same opportunity — `allowed_domains` and `blocked_domains` could be implemented via post-filtering of DuckDuckGo results rather than being gated as builtin-only — but that's out of scope for this PR.

pydantic_ai_slim/pydantic_ai/common_tools/web_fetch.py
```diff
@@ -0,0 +1,163 @@
...
+            elif content_type == 'application/json':
+                try:
+                    parsed = json.loads(text)
+                    content = f'```json\n{json.dumps(parsed, indent=2)}\n```'
+                except (json.JSONDecodeError, ValueError):
+                    content = text
+            else:
+                content = text
+        else:
+            content = f'[Binary content of type {content_type} cannot be displayed as text]'
```
DouweM: We could return `FileUrl` or `BinaryContent` right?

pydantic_ai_slim/pydantic_ai/common_tools/web_fetch.py
```diff
@@ -0,0 +1,163 @@
...
+
+        content = _clean_whitespace(content)
+
+        if self.max_content_length is not None and len(content) > self.max_content_length:
+            content = content[: self.max_content_length] + '\n\n[Content truncated]'
+
+        return WebFetchResult(url=url, title=title, content=content)
+
+
+_TEXTUAL_CONTENT_TYPES = frozenset(
```
DouweM: We have something similar already here:

https://github.com/pydantic/pydantic-ai/blob/f6c67ffdd0f4ce337800f0dea43e3deaa67ff8b2/pydantic_ai_slim/pydantic_ai/models/openai.py#L1340-L1349

Let's have both use the same util.

Also somewhat related:

https://github.com/pydantic/pydantic-ai/blob/f6c67ffdd0f4ce337800f0dea43e3deaa67ff8b2/pydantic_ai_slim/pydantic_ai/messages.py#L848-L858

pydantic_ai_slim/pydantic_ai/common_tools/web_fetch.py
```diff
@@ -0,0 +1,163 @@
...
+
+def web_fetch_tool(
+    *,
+    max_content_length: int | None = 50_000,
+    allow_local_urls: bool = False,
+    timeout: int = 30,
+) -> Tool[Any]:
+    """Creates a web fetch tool that fetches URLs and converts content to markdown.
+
+    This tool uses SSRF protection via ``pydantic_ai._ssrf.safe_download``.
```
DouweM: Single backticks please :)

uv.lock
DouweM, line 1: @DEENUU1 This file has too many changes: +5,685 -5,705

DouweM (changes requested):

uv.lock
DEENUU1: Fixed, I had to update my uv version

pydantic_ai_slim/pydantic_ai/common_tools/web_fetch.py
DEENUU1: I created this function - https://github.com/pydantic/pydantic-ai/pull/4906/changes#diff-312307a0416da585a1535c94a9c6184cb9319d0fdc8debba6687f712ffb75788R130 to keep the same style. Let me know if that's what you meant.

pydantic_ai_slim/pydantic_ai/common_tools/web_fetch.py
DEENUU1: I was thinking about it, it's ready

pydantic_ai_slim/pydantic_ai/capabilities/web_fetch.py
```diff
@@ -72,5 +73,16 @@ def _default_builtin(self) -> WebFetchTool:
...
+    def _default_local(self) -> Tool[AgentDepsT] | AbstractToolset[AgentDepsT] | None:
+        try:
+            from pydantic_ai.common_tools.web_fetch import web_fetch_tool
+
+            return web_fetch_tool(
+                allowed_domains=self.allowed_domains,
+                blocked_domains=self.blocked_domains,
+            )
+        except ImportError:
+            return None
```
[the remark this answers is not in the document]
DEENUU1: This is correct behavior

pydantic_ai_slim/pydantic_ai/common_tools/web_fetch.py
```diff
@@ -0,0 +1,188 @@
...
+        Returns:
+            The fetched page content.
+        """
+        _check_domain(url, self.allowed_domains, self.blocked_domains)
+
+        response = await safe_download(
+            url,
+            allow_local=self.allow_local_urls,
+            timeout=self.timeout,
+        )
```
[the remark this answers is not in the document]
DEENUU1: I think this one is out of scope

pydantic_ai_slim/pydantic_ai/common_tools/web_fetch.py
```diff
@@ -0,0 +1,192 @@
...
+    """Check if a media type represents text-like content."""
+    return (
+        not media_type
+        or media_type.startswith('text/')
+        or media_type == 'application/json'
+        or media_type.endswith('+json')
+        or media_type == 'application/xml'
+        or media_type.endswith('+xml')
+        or media_type in ('application/x-yaml', 'application/yaml')
+    )
```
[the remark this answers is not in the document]
DouweM: @DEENUU1 Please have a look at this one

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/common_tools/web_fetch.py
```diff
@@ -0,0 +1,192 @@
...
+        Args:
+            url: The URL to fetch.
+
+        Returns:
+            The fetched page content.
+        """
+        _check_domain(url, self.allowed_domains, self.blocked_domains)
+
+        try:
+            response = await safe_download(
```
slmnsh, line 90: Can we also pass `Accept: text/markdown` header because many site supports markdown responses with that header

pydantic_ai_slim/pydantic_ai/common_tools/web_fetch.py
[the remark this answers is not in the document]
DEENUU1: done, extracted it to utils

pydantic_ai_slim/pydantic_ai/common_tools/web_fetch.py
DEENUU1: It already works because it is caught by `text/` in `is_text_like_media_type` function

slmnsh: Are you passing `Accept` header somewhere?

slmnsh: here are some examples of what i am talking about:

- https://developers.cloudflare.com/fundamentals/reference/markdown-for-agents/
- https://www.mintlify.com/docs/ai/markdown-export
- https://vercel.com/changelog/docs-pages-support-markdown-responses

DEENUU1: The `safe_download` function doesn't have a header parameter, so I can't do it for now, unless I missed something

DouweM: @DEENUU1 I think this is very much worth adding a new `headers` param to `safe_download` for :)

pydantic_ai_slim/pydantic_ai/_ssrf.py
```diff
@@ -341,10 +345,12 @@ async def safe_download(
         if resolved.is_https:
             extensions['sni_hostname'] = resolved.hostname
 
+        request_headers: dict[str, str] = {**(headers or {}), 'Host': resolved.hostname}
```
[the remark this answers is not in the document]
DEENUU1: I don't think this makes sense in our use case

pydantic_ai_slim/pydantic_ai/common_tools/web_fetch.py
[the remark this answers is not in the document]
DouweM: It does make sense to me to enforce this inside `safe_download` which already verifies the URL after each redirect, so that a domain restriction is not trivially circumventable. We added a new `headers` field there, so adding `{allowed,blocked}_domains` as well doesn't seem too much work. I'll have my agent look at this.

pydantic_ai_slim/pydantic_ai/common_tools/web_fetch.py
```diff
@@ -0,0 +1,199 @@
...
+    timeout: int = 30,
+    allowed_domains: list[str] | None = None,
+    blocked_domains: list[str] | None = None,
+    headers: dict[str, str] | None = None,
+) -> Tool[Any]:
+    """Creates a web fetch tool that fetches URLs and converts content to markdown.
+
+    This tool uses SSRF protection via `pydantic_ai._ssrf.safe_download`.
+
+    By default, sends ``Accept: text/markdown`` to request markdown directly from
```
DouweM: Single backticks please :)

pydantic_ai_slim/pydantic_ai/common_tools/web_fetch.py
```diff
@@ -0,0 +1,199 @@
...
+        WebFetchLocalTool(
+            max_content_length=max_content_length,
+            allow_local_urls=allow_local_urls,
+            timeout=timeout,
+            allowed_domains=allowed_domains,
+            blocked_domains=blocked_domains,
+            headers=headers,
+        ).__call__,
+        name='web_fetch',
+        description='Fetches the content of a web page at the given URL and returns it as markdown.',
```
DouweM: or binary content

pydantic_ai_slim/pydantic_ai/capabilities/web_fetch.py
```diff
@@ -14,15 +15,15 @@
 class WebFetch(BuiltinOrLocalTool[AgentDepsT]):
     """URL fetching capability.
 
-    Uses the model's builtin URL fetching when available. No default local
-    fallback — provide a custom `local` tool if needed.
+    Uses the model's builtin URL fetching when available, falling back to a local
+    function tool (markdownify-based fetch by default) when it isn't.
```
DouweM, line 19: Only if the web-fetch optional dep group is included right? That is worth documenting explicitly, and raising a clearer error message than just some generic message that the builtin tool is not available (because there is no local fallback). 

Same for the WebSearch capability when DuckDuckGo is not available.

pydantic_ai_slim/pydantic_ai/capabilities/web_fetch.py
DouweM, line 98: See above; I'd like a clearer warning / error message for this scenario -- but not if `local=...` is provided explicitly.

pydantic_ai_slim/pydantic_ai/_ssrf.py
```diff
@@ -316,6 +317,9 @@ async def safe_download(
                     Cloud metadata endpoints are always blocked regardless.
         max_redirects: Maximum number of redirects to follow (default: 10).
         timeout: Request timeout in seconds (default: 30).
+        headers: Additional HTTP headers to include in the request.
+                The ``Host`` header is always set to the original hostname
```
DouweM: Single backtick

pydantic_ai_slim/pydantic_ai/_ssrf.py
```diff
@@ -326,6 +330,9 @@ async def safe_download(
     """
     current_url = url
     redirects_followed = 0
+    original_hostname = urlparse(url).hostname
+    _sensitive_headers = frozenset(('authorization', 'cookie', 'proxy-authorization'))
```
DouweM: Should be a constant in the file, I think

docs/capabilities.md
```diff
@@ -20,7 +20,7 @@ Pydantic AI ships with several capabilities that cover common needs:
 | [`Thinking`][pydantic_ai.capabilities.Thinking] | Enables model [thinking/reasoning](thinking.md) at configurable effort | Yes |
 | [`Hooks`][pydantic_ai.capabilities.Hooks] | Decorator-based [lifecycle hook](hooks.md) registration | — |
 | [`WebSearch`][pydantic_ai.capabilities.WebSearch] | Web search — builtin when supported, local fallback otherwise | Yes |
-| [`WebFetch`][pydantic_ai.capabilities.WebFetch] | URL fetching — builtin when supported, custom local fallback | Yes |
+| [`WebFetch`][pydantic_ai.capabilities.WebFetch] | URL fetching — builtin when supported, local fallback otherwise | Yes |
```
DouweM: As mentioned above, it needs to be more clear that WebSearch and WebFetch only have a local fallback when a specific package/optional group is installed.

DouweM (changes requested):

docs/capabilities.md
```diff
@@ -19,8 +19,8 @@ Pydantic AI ships with several capabilities that cover common needs:
 |---|---|:---:|
 | [`Thinking`][pydantic_ai.capabilities.Thinking] | Enables model [thinking/reasoning](thinking.md) at configurable effort | Yes |
 | [`Hooks`][pydantic_ai.capabilities.Hooks] | Decorator-based [lifecycle hook](hooks.md) registration | — |
-| [`WebSearch`][pydantic_ai.capabilities.WebSearch] | Web search — builtin when supported, local fallback otherwise | Yes |
-| [`WebFetch`][pydantic_ai.capabilities.WebFetch] | URL fetching — builtin when supported, local fallback otherwise | Yes |
+| [`WebSearch`][pydantic_ai.capabilities.WebSearch] | Web search — builtin when supported, [local fallback](common-tools.md#duckduckgo-search-tool) with `duckduckgo` extra | Yes |
+| [`WebFetch`][pydantic_ai.capabilities.WebFetch] | URL fetching — builtin when supported, [local fallback](common-tools.md#web-fetch-tool) with `web-fetch` extra | Yes |
```
DouweM: link to the install section, and make sure those common tool docs mention the install instructions

pydantic_ai_slim/pydantic_ai/capabilities/web_fetch.py
```diff
@@ -82,6 +86,15 @@ def _default_local(self) -> Tool[AgentDepsT] | AbstractToolset[AgentDepsT] | Non
                 blocked_domains=self.blocked_domains,
             )
         except ImportError:
+            import warnings
+
+            warnings.warn(
+                'WebFetch local fallback requires the `web-fetch` optional group — '
+                '`pip install "pydantic-ai-slim[web-fetch]"`. '
+                'Without it, WebFetch only works with models that support it natively.',
+                UserWarning,
```
DouweM, line 95: Verify this only shows up if no `local` was provided.

pydantic_ai_slim/pydantic_ai/_ssrf.py
```diff
@@ -294,12 +295,26 @@ def resolve_redirect_url(current_url: str, location: str) -> str:
         return urlunparse((parsed_current.scheme, parsed_current.netloc, f'{base_path}/{location}', '', '', ''))
 
 
+def _check_domain(hostname: str, allowed_domains: list[str] | None, blocked_domains: list[str] | None) -> None:
```
DouweM: kwargs for {allowed,blocked}_domains; always do that when creating multi-arg functions

DouweM: Should these be sets?

Should we have certain behavior around subdomains? How do the equiv args with e.g. OpenAI/Anthropic/Google work?

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/_ssrf.py
DouweM: *From Claude (Opus 4.6), on behalf of @DouweM.*

Keeping `list[str]` — all builtin tools (`WebFetchTool`, `WebSearchTool`) use `list[str] | None` and providers pass them straight through to their APIs without transformation. No subdomain matching exists anywhere in the codebase or (from what I can tell) in provider APIs — it's exact hostname match throughout.

For the local `safe_download` implementation, I've matched this behavior: exact hostname match via `in` check. If providers add subdomain support in the future, we can extend both the builtin forwarding and local check consistently.

PR #4909, opened 2026-03-30
fix(ag-ui): drop TOOL_CALL_ARGS deltas arriving after TOOL_CALL_END (fixes #4733)

### LINKED ISSUE #4733: `AGUIEventStream` emits `TOOL_CALL_ARGS` after `TOOL_CALL_END` for same tool_call_id (stray delta), violating AG-UI protocol
### Initial Checks

## Bug Description

`AGUIEventStream.handle_tool_call_delta` receives a `ToolCallPartDelta` **after** `handle_builtin_tool_call_end` has already been called for the same `tool_call_id`. This causes a `TOOL_CALL_ARGS` SSE event to be emitted **after** `TOOL_CALL_END` for the same tool call, which violates the AG-UI protocol and crashes the `@ag-ui/client` verifier.

## Observed SSE Output (from curling the backend)

```
TOOL_CALL_ARGS   (id: mcp_22b..., delta: '{"action":"call_tool",...,"tool_args":')
TOOL_CALL_ARGS   (id: mcp_22b..., delta: '}')
TOOL_CALL_ARGS   (id: mcp_22b..., delta: '}')   <- stray delta AFTER end
```

The final `TOOL_CALL_ARGS` event arrives after `TOOL_CALL_END`, which is invalid per the AG-UI protocol spec.

## Steps to Reproduce

1. Use `MCPServerTool` with the OpenAI Responses API model in streaming mode via the AG-UI endpoint.
2. Trigger a tool call that emits multiple argument deltas.
3. Inspect the raw SSE stream (e.g. with `curl`).
4. Observe a `TOOL_CALL_ARGS` event emitted after `TOOL_CALL_END` for the same `tool_call_id`.

## Expected Behavior

No `TOOL_CALL_ARGS` events should be emitted for a `tool_call_id` after `TOOL_CALL_END` has been sent for that ID. The stream should be ordered: `TOOL_CALL_START` → one or more `TOOL_CALL_ARGS` → `TOOL_CALL_END`.

## Actual Behavior

A stray `TOOL_CALL_ARGS` delta is emitted after `TOOL_CALL_END`, violating the AG-UI protocol and causing the `@ag-ui/client` verifier to crash/error.

## Related Issues

- #4623 — builtin tool results dropped on follow-up AG-UI requests
- ag-ui-protocol/ag-ui#1275 — TOOL_CALL_START events suppressed after tool result

Neither of the above covers this exact scenario (stray post-end delta).

## Environment

- Model: Internal OpenAI Responses API-compatible endpoint routing to custom inference (streaming mode); **not** the official OpenAI API
- Tool type: `MCPServerTool`
- Component: `AGUIEventStream` (AG-UI integration)

```Python

```

### Logfire Trace

_No response_

- Python: latest
- Pydantic AI: latest
- LLM provider SDK: openai (Responses API, streaming mode)

IgnazioDS:
## Problem

Fixes #4733. `AGUIEventStream.handle_tool_call_delta` was unconditionally emitting `TOOL_CALL_ARGS` events even when a stray `ToolCallPartDelta` arrived **after** `TOOL_CALL_END` had already been emitted for the same `tool_call_id`. This violates the AG-UI protocol ordering invariant and crashes the `@ag-ui/client` verifier.

Observed SSE sequence before this fix:
```
TOOL_CALL_ARGS   (id: mcp_22b..., delta: '{...}')
TOOL_CALL_ARGS   (id: mcp_22b..., delta: '}')   ← stray delta, INVALID
```

## Fix

Added `_ended_tool_call_ids: set[str]` to `AGUIEventStream`. Both `handle_tool_call_end` and `handle_builtin_tool_call_end` now register their (mapped) `tool_call_id` in this set before yielding `ToolCallEndEvent`. `handle_tool_call_delta` silently discards any delta whose resolved `tool_call_id` is already in the set.

The guard also handles the builtin tool call case — the mapped builtin ID (not the original) is tracked, consistent with how the existing `_builtin_tool_call_ids` mapping works.

## Test Plan

- [ ] Verify that `TOOL_CALL_ARGS` is never emitted after `TOOL_CALL_END` for the same `tool_call_id`
- [ ] Existing AG-UI streaming tests still pass
- [ ] `@ag-ui/client` verifier no longer crashes for `MCPServerTool` calls in streaming mode

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
+++ b/pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
@@ -76,6 +76,7 @@ class AGUIEventStream(UIEventStream[RunAgentInput, BaseEvent, AgentDepsT, Output
 
     _thinking_text: bool = False
     _builtin_tool_call_ids: dict[str, str] = field(default_factory=dict[str, str])
+    _ended_tool_call_ids: set[str] = field(default_factory=set)
     _error: bool = False
 
     @property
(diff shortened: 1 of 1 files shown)


dsfaccini:
duplicated, #3971 is handling this

also no human seems to be behind this...

PR #4920, opened 2026-03-31
Disable native output on Ollama OpenAI-compatible API

### LINKED ISSUE #4917: OllamaProvider incorrectly advertises JSON schema support via OpenAI API leading to silent structured output failure

pandego:
- Closes #4917

### Pre-Review Checklist

This stops `OllamaProvider` from advertising native JSON-schema output on its OpenAI-compatible `/v1` API.

Ollama documents structured output through its native `format` field rather than OpenAI-style `response_format={"type": "json_schema"}`. Because of that mismatch, `NativeOutput(..., strict=True)` could look supported, fall back to post-hoc validation, and loop through invalid retries instead of getting schema-enforced output.

This patch keeps the existing Ollama-specific JSON schema transformers and JSON-object support, but forces `supports_json_schema_output=False` at the provider layer so provider capabilities do not overstate what the `/v1` API can guarantee.

Validation run:
- `uv run pytest tests/providers/test_ollama.py`
- `uv run ruff check pydantic_ai_slim/pydantic_ai/providers/ollama.py tests/providers/test_ollama.py`

### Pre-Merge Checklist

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/providers/ollama.py
+++ b/pydantic_ai_slim/pydantic_ai/providers/ollama.py
@@ -1,6 +1,7 @@
 from __future__ import annotations as _annotations
 
 import os
+from dataclasses import replace
 
 import httpx
 from openai import AsyncOpenAI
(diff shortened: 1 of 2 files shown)


aress31:
What would be the recommended approach to reliably use Ollama’s native (aka `NativeOutput`) structured JSON output (via the format parameter) with the current version of pydantic-ai, given that Ollama is only supported through the OpenAI compatibility layer rather than a dedicated integration?

Disabling the feature does not appear to be a proper solution.

pandego:
Thanks - I agree disabling it is not the full long-term story.

This PR is narrower: it stops `OllamaProvider` from advertising OpenAI-style native JSON-schema output on Ollama's OpenAI-compatible `/v1` API, because that API does not appear to guarantee `response_format={"type": "json_schema"}` behavior.

So the goal here is to make the advertised capability accurate rather than to remove a properly supported feature.

If we want reliable Ollama-native structured output via Ollama's `format` parameter, I think that likely needs a dedicated Ollama-specific integration/path rather than going through the OpenAI compatibility layer and treating it as equivalent to OpenAI native schema enforcement.

In the current setup, the safer behavior is:
- keep Ollama-specific schema transformation where useful,
- keep JSON-object support,
- but do not claim native JSON-schema output guarantees that the `/v1` compatibility layer cannot reliably provide.

DouweM:
Closing pending discussion in the issue.

PR #4926, opened 2026-04-01
feat: cross-provider `service_tier` model setting; Anthropic + Gemini API + Vertex Priority PayGo support

### LINKED ISSUE #5095: Support Vertex AI Priority PayGo in google_service_tier
Follow-up to #4297 / #4312 (Flex PayGo): add support for [Priority PayGo](https://cloud.google.com/vertex-ai/generative-ai/docs/priority-paygo) in `google_service_tier`.

Same header pattern as Flex, with `priority` instead of `flex`:
- `pt_then_priority` → `X-Vertex-AI-LLM-Shared-Request-Type: priority`
- `priority_only` → `X-Vertex-AI-LLM-Request-Type: shared` + `X-Vertex-AI-LLM-Shared-Request-Type: priority`

Championing this — production Vertex use case (PT → Priority → Flex retry chain). Implementation + live validation in PR #5094.

markmcd:
## Summary

Closes #5095 

Supersedes the original Vertex-only `google_service_tier` design and consolidates the cross-provider service-tier work.

Adds a unified [`service_tier`][pydantic_ai.settings.ModelSettings.service_tier] field on `ModelSettings`, mapped to each provider's native service-tier concept where one exists. Provider-specific overrides remain available for values that don't fit the unified set.

This PR consolidates earlier exploration work in #5158 (closed, by @Mawox) and #5094 (closed, by @anatolec — Priority PayGo on Vertex). Their commits are preserved in this branch's history.

### Cross-provider mapping

`service_tier` accepts `'auto' | 'default' | 'flex' | 'priority'`:

| value | OpenAI | Anthropic | Bedrock | Google (Gemini API) | Google (Vertex AI) |
|---|---|---|---|---|---|
| `'auto'` | `'auto'` | `'auto'` | _(omitted)_ | _(omitted)_ | _no headers (PT then on-demand)_ |
| `'default'` | `'default'` | `'standard_only'` | `{'type': 'default'}` | `'standard'` | _no headers (PT then on-demand)_ |
| `'flex'` | `'flex'` | _(omitted)_ | `{'type': 'flex'}` | `'flex'` | header `Shared-Request-Type: flex` (PT then Flex PayGo) |
| `'priority'` | `'priority'` | _(omitted)_ | `{'type': 'priority'}` | `'priority'` | header `Shared-Request-Type: priority` (PT then Priority PayGo) |

Per-provider settings (`openai_service_tier`, `anthropic_service_tier`, `bedrock_service_tier`, `google_vertex_service_tier`) always take precedence over the unified field, and they're the only way to reach values that aren't in the unified set: Bedrock's `'reserved'`, Anthropic's `'standard_only'` explicit form, and Vertex's full PT-routing matrix (`'pt_only'`, `'on_demand'`, `'flex_only'`, `'priority_only'`, etc.).

`'auto'` vs `'default'` distinction: `'auto'` lets the provider decide and may include premium tiers when available (matters for OpenAI's scale credits and Anthropic's priority capacity). `'default'` explicitly opts out of those promotions. On Bedrock / Google they're functionally equivalent today, but encoded forward-compatibly through the omit-vs-explicit wire choice.

### Vertex AI design choice

The unified `'flex'` and `'priority'` map to the **PT-with-spillover** variants (single `Shared-Request-Type` header, no `Request-Type: shared`), so Vertex customers with Provisioned Throughput keep using their reserved capacity first. To bypass PT entirely, set `google_vertex_service_tier='flex_only'` / `'priority_only'` directly. Open question with Google for confirmation: when a PT customer exceeds quota with the single-header form, does spillover land in Flex/Priority or in standard PayGo? Empirically (Mawox's reproduction on a zero-PT project, anatolec's #5094 live test) the headers fall through safely; the PT-customer-over-quota case is the only path not yet experimentally confirmed.

### Other behavior changes

- `google_service_tier` (the original Vertex-only field) is deprecated in favor of `google_vertex_service_tier`. Reading it emits a `DeprecationWarning`. The values are unchanged.
- Adds `'pt_then_priority'` and `'priority_only'` Vertex routing values (from #5094).
- Cerebras: `openai_service_tier` added to the unsupported-settings filter (latent bug — the per-provider field was being forwarded to an API that doesn't accept it).
- `google-genai` bumped to `>=1.70.0` for the SDK's new `ServiceTier` enum on the Gemini API.

### Test plan

- [x] Unit + parametrize coverage for the cross-provider mapping on each of OpenAI / Anthropic / Bedrock / Google (GLA + Vertex).
- [x] `DeprecationWarning` regression test for `google_service_tier`.
- [x] Live API call sweep on Gemini API (`google_service_tier` = `'default'` / `'standard'` / `'flex'` / `'priority'`).


pydantic_ai_slim/pydantic_ai/models/google.py
```diff
@@ -167,13 +168,30 @@
...
+GoogleServiceTier = Literal[
+    'default',
+    'standard',
+    'flex',
+    'priority',
+    'pt_then_on_demand',
+    'pt_only',
+    'pt_then_flex',
+    'on_demand',
+    'flex_only',
```
DouweM, line 179: @markmcd Is there any way we could make (at least some of) the same values work for both GLA and Vertex? 

cc @ewjoachim

DouweM (changes requested):

pydantic_ai_slim/pydantic_ai/models/google.py
markmcd: I'll wait for the Vertex opinion on this one. The GLA values align with what other major providers do, so I'd prefer to remap the Vertex values back (if that's even feasible?)

ewjoachim: These are the relevant docs:
- https://docs.cloud.google.com/vertex-ai/generative-ai/docs/provisioned-throughput/use-provisioned-throughput
- https://docs.cloud.google.com/vertex-ai/generative-ai/docs/flex-paygo

They control the following headers:

| Tier Name | `X-Vertex-AI-LLM-Request-Type` | `X-Vertex-AI-LLM-Shared-Request-Type` | Description / Behavior |
| :--- | :--- | :--- | :--- |
| **`pt_then_on_demand`** | *(Not sent)* | *(Not sent)* | **Default Behavior:** Uses PT first. Excess traffic spills over to standard PayGo. |
| **`pt_only`** | `dedicated` | *(Not sent)* | **Provisioned Only:** Uses PT exclusively. Rejects traffic with a 429 error if capacity is exceeded. |
| **`pt_then_flex`** | *(Not sent)* | `flex` | **Hybrid Flex:** Uses PT first. Excess traffic spills over to the lower-cost Flex PayGo tier. |
| **`on_demand`** | `shared` | *(Not sent)* | **Pure On-demand:** Bypasses PT to use standard shared resources at regular rates. |
| **`flex_only`** | `shared` | `flex` | **Pure Flex:** Bypasses PT to use the discounted Flex PayGo tier directly. |


We could definitly use `standard` and `flex` though it's ambiguous if they should map to equivalent **with** or **without** PT. That said, by default, PT is used unless we send `X-Vertex-AI-LLM-Request-Type: shared` so it could make sense to:

- Replace `pt_then_on_demand` -> `standard` and `pt_then_flex` -> `flex`, but then we would need to rename at least `on_demand` to somthing like `on_demand_only` or `pay_as_you_go_only` (to match flex_only`)
- Or we could treat `standard` and `flex` as aliases for `pt_then_on_demand` and `pt_then_flex` but keep the non-ambiguous names around, it would mean a bit of extra documentation work to convey this in a very unambiguous manner, but this lets folks write unambiguous code.

`priority` doesn't match to anything on vertex and `pt` doesn't match to anything on Google (as far as I can tell), so it will be hard to get anything perfect. Not sure exactly if that's 100% helpful, but if you think I can help further, feel free :)

markmcd: My vote is for unambiguous options, and then provide the additional aliases.

The GLA impl aligns with, e.g. OpenAI, and while that doesn't happen transparently in this case, it's a nice portability feature. I think providing similar ergonomics for the Vertex values would be positive, as long as that's a reasonable mental model for Vertex customers (hopefully you can make that call @ewjoachim! I don't know that stack well)

If we agree on this, I'll update the PR to make it clear that `standard` and `flex` are accepted for vertex as a shim only.

DouweM: - Since we support service tiers now for OpenAI, Bedrock, GLA, and Vertex, it'd be nice to add a new top-level `service_tier` `ModelSetting` with a narrow set of values (most likely the OpenAI ones) that we then try to map to providers (i.e. interpret in the model classes) as best we can, with clear documentation (in the docstring) of how each provider interprets them.
- If we have a narrower set there, we could then rename the `google_service_tier` field to `google_vertex_service_tier` (and deprecate the original). Then we may either not need a separate `google_gla_service_tier` (if the top level `service_tier` covers all the values), or we can add a new `google_gla_service_tier` in case granular control is needed.

That way we get the convenience of a single set of values across providers, with the ability to override per-provider values as needed.

I don't have an opinion yet on what the exact top-level service tier values, or mapping to Vertex, should be, but if the above approach makes sense to you I trust either/both of you to be able to come up with some reasonable 😄

markmcd: OK I've updated to take this into account using OpenAI's values as the "default". Provider-specific values take precedence, and supported providers have been updated, including adding mappings from generic to specific where it makes sense to.

Another PR has appeared that also addresses this, #5158, I haven't looked at it, but I'm not precious about keeping mine if that's better.

DouweM: @markmcd Thanks Mark, I've been working with an agent on consolidating these related PRs so I'll tell it to look at your new changes.

The agent did have a question for you (to pass on to the Vertex team). In its own words:

> @Mawox has been picking up the design direction in #5158 (top-level service_tier with per-provider fallbacks as we discussed), and @anatolec
> added Priority PayGo values in #5094. I'm folding both into one PR and want to extend the cross-provider mapping to Vertex, rather than keeping the "Vertex ignored" carve-out — with
> pt_then_priority now in scope, the mapping looks clean:
> 
> - flex → X-Vertex-AI-LLM-Shared-Request-Type: flex
> - priority → X-Vertex-AI-LLM-Shared-Request-Type: priority
> - default / auto → no headers (PT-then-on-demand default)
> 
> Before I commit to that, there's one thing the public docs don't quite spell out: if a project has zero PT quota on the target model/region and we send only the single shared-request-type
> header, does the request fall through safely to Flex/Priority PayGo, or does it 429?
> 
> @ewjoachim's original writeup describes it as "Uses PT first. Excess traffic spills over to Flex PayGo," and @anatolec saw traffic_type: ON_DEMAND_PRIORITY in a live test — so empirically
> it looks safe. But before defaulting every cross-provider service_tier='priority' user through this on Vertex, could you check with the DeepMind / Vertex team?
> 
> Specifically:
> 1. Zero-PT project + only X-Vertex-AI-LLM-Shared-Request-Type: flex → 429, or Flex PayGo?
> 2. Same with priority → 429, or Priority PayGo?
> 3. For a project with PT quota, is the spillover destination when these single-header requests exceed PT actually Flex/Priority (not standard on-demand)?
> 
> If 1+2 fall through safely we'll go single-header (respects PT for customers who have it, safe for everyone else). If not, we'll also send X-Vertex-AI-LLM-Request-Type: shared to guarantee
> no PT dependency at the cost of bypassing PT entirely. I'd rather the former if it's actually safe.
> 
> Thanks!

Mawox: Thanks @DouweM — happy to fold this into #5158.

Reproduced Devin's **Q1** on a separate zero-PT Vertex project (`gemini-3-flash-preview`, `location='global'` — Flex PayGo is [preview-only](https://docs.cloud.google.com/vertex-ai/generative-ai/docs/flex-paygo)):

```
pt_then_flex    traffic_type='ON_DEMAND_FLEX'      ← Q1: single Shared-Request-Type: flex on zero-PT
flex_only       traffic_type='ON_DEMAND_FLEX'
pt_only         429, PT quota exceeded             ← zero-PT control ✓
```

**Q2** (`priority`) isn't on this branch — @anatolec's #5094 already shows the same pattern: `pt_then_priority` → `ON_DEMAND_PRIORITY` on zero-PT.

**Q3** (PT-quota spillover destination) still needs the Vertex team.

If Q3 spills to Flex/Priority: drop the carve-out in #5158, `flex`/`priority` → single Shared-Request-Type header, keep `google_vertex_service_tier` as the escape hatch (needs #5094 folded in first for the `priority` mapping). If it spills to plain on-demand: keep the carve-out — silent downgrade from `priority` is worse than requiring the explicit field on Vertex.

DouweM:
@markmcd @Mawox Thanks for working on this! 

Pushed four commits on top of 878db9ee to consolidate the work across #5094 and #5158:

1. @anatolec's Priority PayGo values (pt_then_priority, priority_only) on google_vertex_service_tier — folding in #5094 (kept as his commit via --author).
2. auto = omit consistently on Bedrock and GLA: top-level service_tier='auto' was sending explicit {'type': 'default'} / 'standard', now it properly unsets — matching the ServiceTier docstring so 'auto' can act as a clean override-to-unset for
inherited settings. Also: Cerebras openai_service_tier added alongside the pre-existing service_tier entry in openai_unsupported_model_settings; bedrock_service_tier docstring clarified to note it is the only way to request 'reserved'.
3. Fix three tests that were failing on this branch after the google-genai 1.70 bump (file search snapshots now include the file_search_store field; streaming safety-filter mock needed sdk_http_response=None, otherwise the new x-gemini-service-tier
lookup pulls a Mock into provider_details).
4. Anthropic service_tier mapping test coverage — addresses the Devin Review finding about parity with the OpenAI/Google/Bedrock tests.

@markmcd Vertex top-level → priority-header mapping still off pending your Q3 answer (PT-customer-over-quota spillover destination). Once confirmed safe, it's a one-line change to map priority → X-Vertex-AI-LLM-Shared-Request-Type: priority the same way flex
already does. (see https://github.com/pydantic/pydantic-ai/pull/4926#discussion_r3132114129)

pydantic_ai_slim/pydantic_ai/models/google.py
ewjoachim: (Just clarifying that I don't feel I'm sufficiently knowledgeable on the subject to add anything meaningful to what has already been said)

pydantic_ai_slim/pydantic_ai/settings.py
```diff
@@ -213,6 +231,27 @@ class ModelSettings(TypedDict, total=False):
...
+    - `'flex'`: Flexible / lower-priority service tier with lower costs.
+    - `'priority'`: Higher-priority service tier with lower latency or higher limits.
+
+    Provider-specific service tier settings (e.g., `openai_service_tier`,
+    `google_vertex_service_tier`) take precedence over this unified field.
+
+    Supported by:
+
+    * OpenAI
+    * Gemini
```
ewjoachim: I might be completely off, but it seems strange that we mention "Gemini" here (I believe for GLA?), and `google_vertex_service_tier` above. I wonder it this bullet list should also mention Vertex (especially since Vertex can be used for non-gemini models)

(also, it looks like this docstring is duplicated from the TypeAlias above, which might lead to either getting out of sync)

pydantic_ai_slim/pydantic_ai/models/google.py
```diff
@@ -243,28 +266,100 @@ class GoogleModelSettings(ModelSettings, total=False):
...
+
+
+def _resolve_gla_service_tier(model_settings: GoogleModelSettings) -> str | None:
+    """Resolve the value to send as `service_tier` on a Gemini API (GLA) request.
+
+    Checks the deprecated `google_service_tier` first (for back-compat, with warning),
+    then the top-level `service_tier`. Maps `'default'` → `'standard'`; drops any value
+    that isn't valid for GLA (including `'auto'`, which signals "let the server decide").
     """
+    raw = _get_deprecated_google_service_tier(model_settings) or model_settings.get('service_tier')
```
ewjoachim: I'm afraid conflating the provider-specific service-tier and the model_settings service tier is bound to create headaches: at some point, some provider is going to call their default mode "flex" or something like that.

Rather than putting either value in a variable and then handle a `GoogleVertexServiceTier | ServiceTier`, I think it's much much saner to:
1. See if a provider-specific (in this case `GoogleVertexServiceTier`) value is defined. If so, use it.
2. If not, map the `ServiceTier` to a `GoogleVertexServiceTier`
3. Always handle a `GoogleVertexServiceTier` 

(here I'm saying this for Vertex but that would be the way we handle it for every other provider)

As the zen of python says: "In the face of ambiguity, refuse the temptation to guess."

pydantic_ai_slim/pydantic_ai/providers/cerebras.py
```diff
@@ -67,6 +67,7 @@ def model_profile(model_name: str) -> ModelProfile | None:
             'presence_penalty',
             'parallel_tool_calls',
             'service_tier',
+            'openai_service_tier',
         )
```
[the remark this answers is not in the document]
DouweM: OK to leave as is, I believe.

PR #4939, opened 2026-04-01
Add gateway model health checks

Kludex:
## Summary
- add a dedicated live gateway catalog smoke suite and weekly workflow gated behind `--run-gateway-live`
- prune `KnownModelName` gateway literals down to the models that currently work on Gateway, with an explicit `UNSUPPORTED_GATEWAY_MODEL_NAMES` exclusion set
- keep the gateway API key fixture reusable for other tests while making the expensive catalog suite opt-in locally

## Why
Gateway-prefixed literals should only advertise models that actually work through the Gateway. The new smoke suite exercises every `gateway/...` literal with a real request, and the literal list is filtered using the live results so we do not keep stale or broken gateway names in the public type surface.

## Validation
- `.venv/bin/ruff check pydantic_ai_slim/pydantic_ai/models/__init__.py tests/conftest.py tests/providers/test_gateway_catalog.py tests/models/test_model_names.py`
- `.venv/bin/pyright pydantic_ai_slim/pydantic_ai/models/__init__.py tests/conftest.py tests/providers/test_gateway_catalog.py tests/models/test_model_names.py`
- `.venv/bin/pytest tests/providers/test_gateway_catalog.py -q --run-gateway-live`
- `.venv/bin/pytest tests/providers/test_gateway.py -q -k gateway_provider_with_openai`

## Notes
- the commit used `--no-verify` because the `check-cassettes` hook segfaulted with `exit 139` after printing `All cassettes have matching tests!`; the gateway-specific checks above passed cleanly
- `tests/models/test_model_names.py -q` is still blocked in this environment by a live Cerebras connection failure unrelated to this change

### DIFF
--- a/tests/models/test_model_names.py
+++ b/tests/models/test_model_names.py
@@ -79,7 +79,7 @@ _PROVIDER_DEPRECATED_MODELS: dict[str, frozenset[str]] = {
     'openai': DEPRECATED_OPENAI_MODELS,
 }
 
-_UNSUPPORTED_GATEWAY_MODEL_NAMES = frozenset(
+UNSUPPORTED_GATEWAY_MODEL_NAMES = frozenset(
     {
         'gateway/anthropic:claude-3-5-haiku-20241022',
         'gateway/anthropic:claude-3-5-haiku-latest',
@@ -225,7 +225,7 @@ def test_known_model_names():  # pragma: lax no cover
         f'gateway/{provider}:{model_name}'
         for provider in GatewayModelProvider.__args__
         for model_name in get_model_names(_PROVIDER_TO_MODEL_NAMES[provider])
-        if f'gateway/{provider}:{model_name}' not in _UNSUPPORTED_GATEWAY_MODEL_NAMES
+        if f'gateway/{provider}:{model_name}' not in UNSUPPORTED_GATEWAY_MODEL_NAMES
         if not is_deprecated(provider, model_name)
     ]
 
--- a/tests/providers/test_gateway_catalog.py
+++ b/tests/providers/test_gateway_catalog.py
@@ -1,5 +1,6 @@
 from __future__ import annotations as _annotations
 
+import os
 import warnings
 from typing import get_args
 
@@ -7,10 +8,12 @@ import pytest
 from typing_inspection.introspection import get_literal_values
 
 from pydantic_ai import Agent
+from pydantic_ai.exceptions import ModelAPIError, ModelHTTPError
 from pydantic_ai.models import KnownModelName
 from pydantic_ai.providers.gateway import ModelProvider as GatewayModelProvider
 
 from ..conftest import try_import
+from ..models.test_model_names import UNSUPPORTED_GATEWAY_MODEL_NAMES
 
 with try_import() as imports_successful:
     import anthropic as anthropic
@@ -30,7 +33,10 @@ def gateway_live_api_key(pytestconfig: pytest.Config, gateway_api_key: str | Non
     if not pytestconfig.getoption('--run-gateway-live'):
         pytest.skip('gateway catalog smoke tests require --run-gateway-live')
     if not gateway_api_key:
-        pytest.skip('gateway catalog smoke tests require `PYDANTIC_AI_GATEWAY_API_KEY` or `PAIG_API_KEY`')
+        message = 'gateway catalog smoke tests require `PYDANTIC_AI_GATEWAY_API_KEY` or `PAIG_API_KEY`'
+        if os.getenv('CI'):
+            pytest.fail(message)
+        pytest.skip(message)
     return gateway_api_key
 
 
@@ -46,6 +52,17 @@ def _gateway_supported_providers() -> set[str]:
     return {f'gateway/{provider}' for provider in get_args(GatewayModelProvider)}
 
 
+async def _run_gateway_smoke_test(model_name: str) -> str:
+    agent = Agent(model_name, model_settings={'max_tokens': 256}, retries=3)
+    with warnings.catch_warnings():
+        warnings.simplefilter('ignore', DeprecationWarning)
+        result = await agent.run('Reply with exactly OK.')
+    output = result.output
+    assert isinstance(output, str)
+    assert output.strip()
+    return output
+
+
 def test_gateway_known_model_names_only_use_supported_providers() -> None:
     known_gateway_providers = {model_name.split(':', maxsplit=1)[0] for model_name in _gateway_known_model_names()}
     assert known_gateway_providers <= _gateway_supported_providers()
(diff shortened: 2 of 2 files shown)


.github/workflows/gateway-model-health.yml
```diff
@@ -0,0 +1,47 @@
+name: Gateway Model Health
+
+on:
+  schedule:
+    - cron: '0 10 * * 1'
```
Kludex, line 6: Add a comment.

.github/workflows/gateway-model-health.yml
```diff
@@ -0,0 +1,47 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  gateway-model-health:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    environment:
+      name: gateway-model-health
```
Kludex, line 20: Is there a need for an environment here?

.github/workflows/gateway-model-health.yml
```diff
@@ -0,0 +1,47 @@
...
+          python-version: "3.12"
+          enable-cache: true
+          cache-suffix: gateway-model-health
+
+      - name: Ensure gateway secret is configured
+        run: |
+          test -n "$PYDANTIC_AI_GATEWAY_API_KEY" || {
+            echo "PYDANTIC_AI_GATEWAY_API_KEY is not configured."
+            exit 1
+          }
```
Kludex: I don't think this is needed.

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -59,6 +59,129 @@
 See https://github.com/openai/openai-python/blob/v1.54.4/src/openai/_constants.py#L9
 """
 
+UNSUPPORTED_GATEWAY_MODEL_NAMES = frozenset(
```
Kludex: I don't think this should live in the source code, but actually in the tests.

tests/providers/test_gateway_catalog.py
```diff
@@ -0,0 +1,161 @@
...
+    monkeypatch.setattr('tests.providers.test_gateway_catalog.Agent', FakeGatewayAgent)
+    monkeypatch.setattr('tests.providers.test_gateway_catalog._run_gateway_smoke_test', fake_run_gateway_smoke_test)
+
+    assert await _smoke_test_model_name('gateway/openai:gpt-5') == 'OK'
+    assert captured == {
+        'model_name': 'gateway/openai:gpt-5',
+        'model_settings': {'max_tokens': 256},
+        'retries': 3,
+        'agent': captured['agent'],
+    }
```
Kludex: Is this even useful? I think the point of this test file is just to test the name. Why was this added?

tests/providers/test_gateway_catalog.py
```diff
@@ -0,0 +1,161 @@
...
+        self.output = output
+
+    async def run(self, prompt: str) -> _FakeRunResult:
+        assert prompt == 'Reply with exactly OK.'
+        warnings.warn('deprecated', DeprecationWarning)
+        return _FakeRunResult(self.output)
+
+
+async def test_run_gateway_smoke_test_ignores_deprecation_warning() -> None:
+    assert await _run_gateway_smoke_test(_FakeAgent('OK')) == 'OK'
```
Kludex: Is this necessary? I think we can just remove this.

.github/workflows/gateway-model-health.yml
Kludex: Added a short cron comment in the workflow and simplified the rest of the file in the latest push.

.github/workflows/gateway-model-health.yml
Kludex: Removed the workflow environment block in `cbeee8e8`.

.github/workflows/gateway-model-health.yml
Kludex: Removed that explicit secret-check step in `cbeee8e8`.

pydantic_ai_slim/pydantic_ai/models/__init__.py
Kludex: Moved the unsupported gateway exclusion set out of library source and into `tests/models/test_model_names.py` in `cbeee8e8`. `KnownModelName` stays pruned, but the bookkeeping now lives in tests.

tests/providers/test_gateway_catalog.py
Kludex: Removed the extra helper/unit-test scaffolding from `test_gateway_catalog.py` in `cbeee8e8`. The file is back to the provider-surface check plus the live per-model smoke test.

tests/providers/test_gateway_catalog.py
Kludex: Removed in `cbeee8e8`.

.github/workflows/gateway-model-health.yml
Kludex: Restored this in `976c1564`. `zizmor` flags `PYDANTIC_AI_GATEWAY_API_KEY` as `secrets-outside-env` without a dedicated Actions environment, so we do need the job-level `environment` here.

tests/providers/test_gateway_catalog.py
```diff
@@ -0,0 +1,64 @@
...
+
+pytestmark = pytest.mark.anyio
+
+
+@pytest.fixture(scope='module')
+def gateway_live_api_key(pytestconfig: pytest.Config, gateway_api_key: str | None) -> str:
+    if not pytestconfig.getoption('--run-gateway-live'):
+        pytest.skip('gateway catalog smoke tests require --run-gateway-live')
+    if not gateway_api_key:
+        pytest.skip('gateway catalog smoke tests require `PYDANTIC_AI_GATEWAY_API_KEY` or `PAIG_API_KEY`')
```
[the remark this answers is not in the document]
Kludex: Fixed in `d711b99c`. When `--run-gateway-live` is enabled, the fixture now fails in CI if `PYDANTIC_AI_GATEWAY_API_KEY` / `PAIG_API_KEY` is missing, and only skips locally. That keeps the scheduled workflow from going green with the health checks silently skipped.

tests/models/test_model_names.py
```diff
@@ -105,6 +225,7 @@ def is_deprecated(provider: str, model_name: str) -> bool:
         f'gateway/{provider}:{model_name}'
         for provider in GatewayModelProvider.__args__
         for model_name in get_model_names(_PROVIDER_TO_MODEL_NAMES[provider])
+        if f'gateway/{provider}:{model_name}' not in _UNSUPPORTED_GATEWAY_MODEL_NAMES
```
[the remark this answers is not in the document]
Kludex: Fixed in `d711b99c`. The gateway catalog suite now also parametrizes over `UNSUPPORTED_GATEWAY_MODEL_NAMES` and marks those live smoke calls as strict `xfail` on `ModelHTTPError` / `ModelAPIError`. If one of the excluded models starts working again, the test will XPASS and fail CI so we know to restore the literal.

Issue #4566, opened 2026-03-06
Support for OpenAI's `tool_search` and `defer_loading` in Responses API.


knownbymanoj:
### Feature Request

I would like to request support for OpenAI's new `tool_search` feature and the `defer_loading` property for tools when using the OpenAI model backend in PydanticAI. 

OpenAI recently introduced **Tool Search** as part of their Responses API. This allows developers to pass a large number of tool schemas to the model but mark them with `defer_loading: true`. The model then dynamically searches for and loads only the tools it needs into its context window on demand. 

This is highly beneficial for multi-agent systems and applications with large toolsets, as it significantly reduces initial token usage, lowers costs, latency, and preserves the model's prompt cache.

**A demo of how code might look when using the feature:**

It would be great to define a tool with a `defer_loading` parameter in the `@agent.tool` decorator, or when registering tools natively:

```python
from pydantic_ai import Agent

agent = Agent('openai:gpt-5.4')

# Tool marked for deferred loading
@agent.tool(defer_loading=True)
async def list_open_orders(customer_id: str) -> str:
    """List open orders for a customer ID."""
    return f"Orders for {customer_id}"

# Also need a way to enable tool_search in the backend request
# Perhaps a configuration flag on the agent or run method
```

Under the hood, when calling the OpenAI Responses API, PydanticAI would need to:
1. Append `{"type": "tool_search"}` to the tools array.
2. Structure the deferred tools with `"defer_loading": True`.
3. Handle the intermediate `tool_search_call` and `tool_search_output` sequence natively during the run before executing the actual function call.

**Your use case(s) for the feature:**
I am building complex LLM orchestration systems where agents have access to dozens of utility functions and specialized APIs. Loading all these schemas upfront degrades performance and increases token costs unnecessarily. We want to define comprehensive agent toolkits but rely on the LLM to pull in what it needs contextually.

**Reference to other projects that have a similar feature:**
- LangChain is currently actively working on implementing this feature for their OpenAI integrations (e.g., Issue #34767).
- MCP (Model Context Protocol) servers are beginning to adopt `"defer_loading": true` for large tool registries.

### References

- **OpenAI Tool Search Documentation:** https://developers.openai.com/api/docs/guides/tools-tool-search
- **OpenAI Responses API Reference:** https://developers.openai.com/api/reference/resources/responses

DouweM:
[...]

PR #4959, opened 2026-04-03
Plan: Native provider tool search

dsfaccini:
## Summary

- Plan for adding native Anthropic (#4167) and OpenAI (#4566) tool search support as follow-ups to #4090
- Key decision: no cross-provider metadata bridge — native discovery is purely model-adapter work
- `search_tools` (client-side) remains as portable fallback on provider switch

## Test plan

- [ ] Review and discuss plan before implementation

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -1,109 +1,249 @@
-# Native Provider Tool Search — Follow-up PR Plan
+# Native Provider Tool Search — Comprehensive Plan
 
 ## Context
 
-PR #4090 (`ToolSearchToolset`) implements portable client-side tool search. Follow-up PRs add
-native Anthropic (#4167) and OpenAI (#4566) tool search. This plan captures the simplified
-architecture after deciding to **drop the cross-provider metadata bridge**.
+Native tool search builds on the existing local `ToolSearchToolset` and the minimal internal `ToolSearch` capability (merged in PR #5047) to push search down to provider-side tool search (Anthropic `tool_search_tool_bm25` / `tool_search_tool_regex`, OpenAI `tool_search` with `execution: 'server' | 'client'`). Apr 14 call with Douwe decided to:
 
-## Key Decision: No Cross-Provider Discovery Bridge
+- Extend `ToolSearch` from a plain internal capability into a provider-adaptive feature with a pluggable `strategy` field
+- Ship in releasable stages rather than one megabranch
+- Keep tool search in pydantic-ai core for now (possibly move to harness in V3)
+- Don't merge this plan file into main — it stays a PR artifact
 
-### The scenario
+Related: #4090 (base `ToolSearchToolset` + raw-SDK reference tests), #5047 (internal `ToolSearch` capability + `CapabilityOrdering`), #4167 (Anthropic native), #4566 (OpenAI native).
 
-1. Anthropic + native tool search discovers `find_tickers`, calls it, gets results
-2. Fallback to OpenAI
-3. OpenAI sees `ToolCallPart('find_tickers')` + `ToolReturnPart('find_tickers', results)` in history
-4. But `find_tickers` is NOT in OpenAI's tool definitions (still deferred — native discovery used `BuiltinToolReturnPart`, not `ToolReturnPart(tool_name='search_tools')`, so `_parse_discovered_tools()` doesn't see it)
-5. OpenAI has `search_tools` available to rediscover `find_tickers` if needed
+## Current State (as of 2026-04-15)
 
-### Why this is fine
+### What already exists in `upstream/main`
 
-- OpenAI already has the **results** of `find_tickers` — it doesn't need the tool definition to understand what happened
-- If it needs `find_tickers` again (unlikely — it already has results), one `search_tools` call rediscovers it
-- More likely: it wants the *next* tool (e.g., `get_ticker_price`) and discovers that via `search_tools`
-- APIs don't validate historical tool calls against current definitions — `ToolCallPart`/`ToolReturnPart` are just history records
-- Context optimization: fallback provider doesn't carry stale tool definitions it may never need
+- **`capabilities/_tool_search.py`** — `ToolSearch(AbstractCapability)`. Plain capability (not a `BuiltinOrLocalTool` subclass). Marked internal via `get_serialization_name() → None`. `get_wrapper_toolset()` returns `ToolSearchToolset(wrapped=toolset)`. `get_ordering()` returns `CapabilityOrdering(position='outermost')`. Merged in PR #5047 (Apr 10).
+- **`capabilities/_ordering.py`** — `CapabilityOrdering` + `sort_capabilities()` (topological sort). Supports `position`, `wraps`, `wrapped_by`, `requires`. Added in PR #5047. This is the ordering system Douwe mentioned in the call.
+- **`agent/__init__.py`** — `_inject_auto_capabilities()` appends `ToolSearch` (via `_AUTO_INJECT_CAPABILITY_TYPES`) when not explicitly provided. Uses `has_capability_type()` for detection. **Q4 answered** — backward compat mechanism is already in place.
+- **`toolsets/_tool_search.py`** — `ToolSearchToolset(WrapperToolset)`. Substring-based search via `_search_tools()`. Partitions tools by `defer_loading=True`, injects `search_tools` function for the model to discover deferred tools.
+- **`tools.py`** — `ToolDefinition` has `defer_loading: bool` and `prefer_builtin: str | None` fields. `prefer_builtin` is the 1:1 routing used by `BuiltinOrLocalTool`.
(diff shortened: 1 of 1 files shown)


PLAN.md
```diff
@@ -0,0 +1,249 @@
...
+- `anthropic>=0.80.0` — already ships `ToolSearchToolBm25_20251119Param`, `ToolSearchToolRegex20251119Param`, `ToolSearchToolResultBlock`, `ToolSearchToolSearchResultBlock`, `ToolReferenceBlockParam`. **No bump needed.**
+- `openai>=2.29.0` — ships `ToolSearchToolParam` (with `execution: 'server' | 'client'`), `ResponseToolSearchCall`, `ResponseToolSearchOutputItem`, `NamespaceToolParam`. **No bump needed.**
+
+## Open Questions — resolve with Douwe before implementation
+
+Stages 1 and 4 are blocked on these. Q1 must be answered before Stage 1. Q2 and Q5 must be answered before Stage 4.
+
+**Answered during planning research** (not listed below): Q3 (parallel harness work — none, work is in core via #5047), Q4 (auto-injection backward compat — `_inject_auto_capabilities` already does this).
+
+### Q1 — How should `ToolSearch` route deferred tool defs to native providers? (blocks Stage 1)
```
dsfaccini, line 45: Claude here: @DouweM — **this is the blocking question for Stage 1** and the biggest API-surface decision. Can you pick one of Q1-A/B/C/D below?

The core tension: `prefer_builtin` today means *"remove the local tool when the builtin is available"* (1:1 — web_search replaces one local fallback). Tool search is 1:N — one `tool_search` builtin operates over N deferred function tools, which are the *input corpus* not alternatives. We need to keep them in the request, marked deferred on the wire.

My preference is **Q1-A** (new `corpus_for_builtin` field on `ToolDefinition`) — cleanest semantic separation, mirrors how `BuiltinOrLocalTool._add_prefer_builtin` already works. But it adds a public field so needs your sign-off. **Q1-C** (refactor `ToolSearch` to subclass `BuiltinOrLocalTool`) reuses existing infra but forces the 1:1 contract onto a 1:N relationship.

Field name if Q1-A: `corpus_for_builtin`, `defers_to_builtin`, `searched_by_builtin`, or something else?

PLAN.md
```diff
@@ -0,0 +1,249 @@
...
+- **Q1-D**: Keep the partitioning on the capability side — `ToolSearch.get_wrapper_toolset()` checks the model profile (via a new accessor or a lazy hook) and returns either `ToolSearchToolset` (local) or a new `NativeToolSearchToolset` (corpus mode). Avoids adding a new `ToolDefinition` field. Downside: capability needs access to the model profile at toolset-build time, which may not be available.
+
+**Reasoning to preserve** (1:1 vs 1:N): web_search et al. are 1:1 — one builtin replaces one local fallback. Tool search is 1:N — one `tool_search` builtin operates over N deferred function tools, which are the *input* corpus not alternatives. Reusing `prefer_builtin` would create hidden semantic branching per builtin kind.
+
+**Needed from Douwe**: pick an option; if Q1-A, confirm the field name (`corpus_for_builtin` vs `defers_to_builtin` vs `searched_by_builtin`).
+
+### Q2 — Custom-callable routing on OpenAI `execution='client'`: Route A vs Route B (blocks Stage 4)
```
dsfaccini, line 62: Claude here: @DouweM — **blocks Stage 4**. For OpenAI `execution='client'` routing, which do you prefer?

- **Route A**: parse-time conversion. `models/openai.py` maps `ResponseToolSearchCall(execution='client')` → `ToolCallPart(tool_name=<local>)`. Zero `_agent_graph.py` changes. `tool_call_id` round-trips so history replay works; downside is we lose the "originated from a builtin" signal on history.
- **Route B**: keep as `BuiltinToolCallPart`, add a `locally_executed: bool` flag (or a `provider_details` sentinel), new branch in `CallToolsNode._run_stream` resolves a local standin and dispatches. Preserves origin info but requires graph changes and probably a new standin-marker field on `ToolDefinition`.

You leaned B in the call ("we are losing information by not keeping it as a built-in tool called part") and flagged the SDK-shape uncertainty. SDK research is done. My read is A is simpler and the info loss is recoverable from `provider_name` on re-emission, but I defer to you.

PLAN.md
```diff
@@ -0,0 +1,249 @@
...
+- **Route B** (agent-graph hook): Keep `BuiltinToolCallPart` and add a `locally_executed: bool` flag (or a `provider_details` sentinel). `CallToolsNode._run_stream` gets a new branch that resolves the local standin and dispatches. Preserves builtin-origin for history/replay/eventing. Requires `_agent_graph.py` changes and possibly a standin-marker field on `ToolDefinition`.
+
+Douwe leaned Route B in the call (*'we are losing information by not keeping it as a built-in tool called part'*, *'it's usually the agent graph that's responsible for looking up the tools'*) but flagged uncertainty pending SDK signature research. SDK research is now done.
+
+**Needed from Douwe**: which route, and if Route B, how to register the local standin (new field on `ToolDefinition`, new capability hook, etc.).
+
+### Q3 — OpenAI deferred-corpus wire shape: namespace wrapper vs flat list (narrows Stage 3)
```
dsfaccini, line 73: Claude here: @DouweM @dsfaccini — heads-up for Stage 3, not a blocker. The raw-SDK reference test at `tests/test_native_tool_search_vcr.py` uses `NamespaceToolParam(type='namespace', name='tools', tools=[...])` to wrap the deferred corpus with per-tool `defer_loading: True`, while pydantic-ai emits tools as a flat top-level list today.

Open question: is the namespace wrapper **required** by the OpenAI Responses API for `tool_search` to work, or will a flat `tools=[...]` list with per-tool `defer_loading: True` also work? I will live-test this during Stage 3 implementation with `logfire.instrument_httpx()`. If the namespace wrapper is required, we need a new wrapper emission path in `models/openai.py` that groups deferred tools.

Note from the reference test TODO: `ToolFunction` (the namespace inner type) lacks `defer_loading` in its TypedDict but the API accepts it — may need a `pyright: ignore` until the SDK catches up.

PLAN.md
```diff
@@ -0,0 +1,249 @@
...
+
+pydantic-ai today emits tools as a flat list on the Responses API request. Questions:
+
+- **Q3a**: Does the OpenAI Responses API *require* the namespace wrapper for `tool_search` to work, or is the flat list also accepted with per-tool `defer_loading: True`?
+- **Q3b**: If the namespace is required, how does that interact with the existing pydantic-ai toolset model? A new wrapper in `models/openai.py` that groups deferred tools into a namespace before emission?
+- **Q3c**: The SDK's `NamespaceToolParam.tools` inner TypedDict (`ToolFunction`) lacks `defer_loading` (reference test has a TODO noting this). We may need to emit the inner tool dict manually with a `pyright: ignore` or wait for an OpenAI SDK update.
+
+**Action**: confirm via a live API call during Stage 3 (instrument with `logfire.instrument_httpx()` per `CLAUDE.local.md`).
+
+### Q4 — `'auto'` precedence on provider switch
```
dsfaccini, line 85: Claude here: @DouweM @dsfaccini — non-blocking, just documenting the limitation. When both Anthropic and OpenAI native tool search are supported and the agent switches providers mid-run, `strategy='auto'` resolves fresh per-request in `model.prepare_request()` so each provider picks its own default (Anthropic → bm25, OpenAI → server tool_search). No cross-provider coordination.

Good to confirm: is that the intended behavior, or do you want a run-level lock to prevent strategy flip-flopping across a single agent run?

PLAN.md
```diff
@@ -0,0 +1,249 @@
...
+
+When both Anthropic and OpenAI native are available (agent switches providers mid-run), how does `strategy='auto'` route? Likely: resolved fresh per-request in `model.prepare_request()` so each provider picks its own default. Document as a known limitation, no special handling beyond what Stages 2/3 already imply. Non-blocking.
+
+### Q5 — Callable return type for custom strategy (blocks Stage 4)
```
dsfaccini, line 89: Claude here: @DouweM — **blocks Stage 4**. You said in the call that the user's custom callable returns a `Sequence[str]` of tool names, description from docstring. That works trivially for Anthropic because `tool_search_tool_result` uses `tool_references` (name-only blocks).

But OpenAI's `tool_search_output` returns full `tools: List[Tool]` definitions. To use `Sequence[str]`, the OpenAI adapter would have to look each name up in the local toolset at response-construction time to hydrate into full `Tool` entries. **Verify this is even possible** — does the adapter have access to the full resolved toolset at response-parsing time, or only `ModelRequestParameters.function_tools`?

Alternatives:
- `Callable[[str], Sequence[str]]` — Douwe's original proposal. Works for Anthropic trivially; needs toolset lookup in the OpenAI path.
- `Callable[[str], Sequence[ToolDefinition]]` — richer, returns full defs. User has to construct them, which is clunky.
- `Callable[[str, Sequence[ToolDefinition]], Sequence[str]]` — user gets the candidate tool defs as input, returns matching names. Signature is more explicit about scope.

Which do you prefer?

voorhs:
For me the value of this feature is in the prompt caching. The current `ToolSearchToolset` changes the context prefix, while native tool search tools from Anthropic and OpenAI doesn't. So I would be happy if you implement it 🙏

DouweM:
Incorporated into https://github.com/pydantic/pydantic-ai/pull/5143.

PR #4960, opened 2026-04-03
Plan: Pluggable tool search strategy

dsfaccini:
## Summary

- Plan for making tool search strategy pluggable via a `ToolSearch` capability
- Allows users to swap in semantic search, BM25, or custom matching logic
- Config-holder capability approach: backward compatible, no auto-injection

## Test plan

- [ ] Review and discuss plan before implementation

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -2,17 +2,48 @@
 
 ## Context
 
-Follow-up to PR #4090 (tool search for deferred-loaded tools). DouweM noted that making ToolSearchToolset a capability is "not worth doing now, but in the future." This plan implements that future direction: a pluggable search strategy via a `ToolSearch` capability, allowing users to swap in semantic search, BM25, or any custom matching logic.
+Follow-up to PR #4090 (tool search for deferred-loaded tools). This plan makes the search strategy pluggable via a `ToolSearch` capability, allowing users to swap in custom matching logic or let `'auto'` route to the best available strategy per provider.
 
-Research in `local-notes/tool-search-research.md` shows the community unanimously moving toward semantic/vector search for tool discovery (97%+ hit rates vs substring's inherent limitations). Spring AI's biggest win is making the strategy pluggable. This plan brings that flexibility to Pydantic AI.
+## Design Decisions
 
-## Design: Config-holder capability (Approach A)
+Sources: DouweM review on #4960, Slack call 2026-04-07, PR #4090 review threads.
 
-- `ToolSearchToolset` gets `search_fn` + `max_results` params (defaulting to current substring behavior)
-- New `ToolSearch` capability holds config, does NOT wrap itself
-- Agent detects `ToolSearch` in capabilities, extracts config, passes to hardcoded TST
-- TST moves before capability wrappers so capabilities can override search_tools behavior
-- No auto-injection, no double-wrapping, backward compatible
+- **`strategy` field, not separate `search_fn`**: all search options are mutually exclusive → single key. `Literal['auto', 'substring', 'regex', 'bm25', 'tool_search'] | Callable[..., Sequence[str]]`, default `'auto'`. Native strategy names included from the start per DouweM review — sets us up for native implementation without API changes.
+- **BuiltinOrLocalTool subclass** (not config-holder): `ToolSearch` subclasses `BuiltinOrLocalTool`, inheriting automatic routing between provider-native (builtin) and local (substring) implementations via `prefer_builtin` on `ToolDefinition`. Same pattern as `WebSearch`, `ImageGeneration`.
+- **Central `_DEFAULT_STRATEGY` mapping**: `dict[str, str]` mapping provider names to their best native strategy. `'auto'` resolves via this mapping at request time. Stored with the TST so we can update defaults without breaking changes.
+- **No separate config data class**: options are kwargs on the capability directly.
+- **Custom callable**: returns `Sequence[str]` (tool names to include), description read from docstring (same as tool functions).
+- **TST wraps before capability wrappers**: capabilities are outermost so they can change anything.
+- **Use `visit_and_replace`** (not a custom `_find_tool_search` helper) for extracting config from the capability tree.
+- **Naming**: `search_guidance` (not `keywords_description`) for the model-facing prompt on how to formulate search input. `tool_description` (not `description`) for the search tool's description text.
+
+### Provider native strategies (researched)
+
+| Provider | Strategies | Notes |
+|----------|-----------|-------|
+| Ours | `substring` | split terms, `any(term in searchable)` |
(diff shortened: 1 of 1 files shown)


dsfaccini:
some notes after having claude review the plan while checking out #4994:

---

**Naming suggestion: `keywords_description` → `search_guidance`**

After thinking through the naming on the `ToolSearch` capability, I'd recommend renaming `keywords_description` to **`search_guidance`**.

**Problem with `keywords_description`:** It's ambiguous — reads as "a description composed of keywords" rather than "the description text for the `keywords` parameter." It's a description-of-a-description, which is awkward.

**Why `search_guidance`:**
- Concise and reads naturally: *"the search guidance is: a natural language description of the capability you need"*
- Captures the intent — you're giving the model guidance on how to formulate its search input
- Sits well next to `description` (tool-level prompt text) without confusion
- Works for both substring ("space-separated keywords") and semantic ("natural language phrase") strategies

**Also consider renaming `description` → `tool_description`** for symmetry and to avoid ambiguity with the capability's own docstring:

```python
ToolSearch(
    search_fn=semantic_search,
    max_results=15,
    tool_description='Search for tools by describing what you need.',
    search_guidance='A natural language description of the capability you need.',
)
```

Minor tradeoff: someone *could* initially wonder if `search_guidance` configures the search algorithm vs. the model prompt, but in context next to `tool_description` it's clearly model-facing. A one-line docstring resolves any remaining ambiguity.

slmnsh:
I think search function should be per toolset instead of global. If I have prefixed toolset i would want different search function.

I was expecting something like this:

```python
toolset.defer_loading(search_fn)
DeferredToolset(toolset, search_fn)

toolset_with_long_description.defer_loading(bm25_search)
```

maybe we can have both per toolset and global capability.

the resolution order will be like:

per toolset > global capability > default substring

EDIT: Now that I think about it, I don't think global cap makes sense for this. because loading all the tools lazily is bad strategy almost noone will do it.

PLAN.md
```diff
@@ -0,0 +1,191 @@
...
+3. ToolSearchToolset             3. capability wrappers (outermost)
+```
+
+This fulfills DouweM's review request: "capabilities need to be able to change basically anything."
+
+### 6. Agent wiring: extract config from ToolSearch capability
+
+**File:** `pydantic_ai_slim/pydantic_ai/agent/__init__.py`
+
+Helper to find `ToolSearch` in capability tree:
```
DouweM: Should use something similar to `AbstractToolset.visit`

PLAN.md
```diff
@@ -0,0 +1,191 @@
...
+
+## Context
+
+Follow-up to PR #4090 (tool search for deferred-loaded tools). DouweM noted that making ToolSearchToolset a capability is "not worth doing now, but in the future." This plan implements that future direction: a pluggable search strategy via a `ToolSearch` capability, allowing users to swap in semantic search, BM25, or any custom matching logic.
+
+Research in `local-notes/tool-search-research.md` shows the community unanimously moving toward semantic/vector search for tool discovery (97%+ hit rates vs substring's inherent limitations). Spring AI's biggest win is making the strategy pluggable. This plan brings that flexibility to Pydantic AI.
+
+## Design: Config-holder capability (Approach A)
+
+- `ToolSearchToolset` gets `search_fn` + `max_results` params (defaulting to current substring behavior)
```
DouweM: - We don't use the `_fn` suffix anywhere
- Callables aren't serializable, and I would want users to be able to change the search strategy in an agent spec.yml
- What if we make it `strategy=Literal['regex', 'bm25', <anything else we add in the future>] | Callable[...]`? Something that sets us up for native implementation + any custom one the user likes.

Also I suggest having a look at the solution in https://github.com/Wh1isper/ya-agent-sdk (I believe that's the right repo)

PLAN.md
DouweM: We need a canonical way to define "auto inject unless already injected" capabilities.

DouweM (changes requested):

PLAN.md
dsfaccini: native implementation will have its own builtin_tool anyway so adding those here doesn't necessarily provide future value

the function isn't serializable but any function that can use a callback isn't, so it just requires the user to have that function available at runtime

dsfaccini:
@slmnsh I was a bit confused by your comment so excuse me if I misunderstood, but to clarify:

- our TST is built in accordance with the `tool_search` functionalities of providers like anthropic and openai, which also accept only one search strategy for all deferred tools, so any `search_function` would also apply to all deferred tools
- by default no tool is deferred, this is in line with backwards compatibility. only tools marked with `defer=True` will be hidden and made available by the TST

does that address your concerns?

slmnsh:
> * our TST is built in accordance with the `tool_search` functionalities of providers like anthropic and openai, which also accept only one search strategy for all deferred tools, so any `search_function` would also apply to all deferred tools

I see, but I think it will be nice addition if you can change the search strategy per toolset. What do you think?

> * by default no tool is deferred, this is in line with backwards compatibility. only tools marked with `defer=True` will be hidden and made available by the TST

Okay this makes sense now.

PLAN.md
```diff
@@ -0,0 +1,191 @@
...
+
+Reorder in `_get_toolset`:
+```
+BEFORE:                          AFTER:
+1. PreparedToolset               1. PreparedToolset
+2. capability wrappers           2. ToolSearchToolset (with extracted config)
+3. ToolSearchToolset             3. capability wrappers (outermost)
+```
+
+This fulfills DouweM's review request: "capabilities need to be able to change basically anything."
```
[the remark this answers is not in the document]
dsfaccini: this is confirmed by @DouweM and actually needed for the upcoming integration of TST with Code Mode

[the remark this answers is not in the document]
dsfaccini: separately dealt with in https://github.com/pydantic/pydantic-ai/pull/5035/changes

KimigaiiWuyi:
I’m not sure if I’ve overlooked something in the codebase, but according to #4090, it seems the `tool_search_tool` utility only supports keyword-based search. Is there a way to replace this with a developer-defined vector search?

Also, as a side note: is it currently recommended to "hide" the majority (say, 90%) of tools and let the AI retrieve/discover them dynamically? In my understanding, this shouldn't introduce significant runtime overhead, but I’m not sure if this is the recommended paradigm.

voorhs:
This feature would be so useful! I am researching tool suggesting and tool searching methods and now stuck to using `prepare_tools` feature - which constantly breaks prompt caching. Being able to plug my custom algorithms into your wonderful library would be incredibly helpful for my experiments 🙏

DouweM:
Incorporated into https://github.com/pydantic/pydantic-ai/pull/5143.

PR #4964, opened 2026-04-03
feat: add `return_schema` and `function_signature` to `ToolDefinition`

### LINKED ISSUE #3122: Also include return (type) json schema for tools (similar how tool parameters are handled)

DouweM:
- Closes #3122
- Closes #4755
- Supersedes #3865
- Supersedes #4851

### Context

This PR unifies the work from two parent PRs:
- **#4851** (`python-signatures-v2`): Added `FunctionSignature` generation for tool definitions, with both `from_function` (inspect-based) and `from_schema` (JSON schema-based) paths
- **#3865** (`tool_return_schema`): Added `return_schema` field to `ToolDefinition` for tool return type schemas

The key insight from the plan: with `return_schema` on `ToolDefinition`, the `from_schema` path gets return type info too, making `from_function` redundant. This PR implements `return_schema` support directly (rather than merging #3865's 60-commit branch), incorporating @DouweM's review feedback from the start.

### Changes

**Core: `return_schema` on `ToolDefinition`**
- Add `return_schema: ObjectJsonSchema | None` and `include_return_schema: bool | None` fields
- Generate `return_schema` lazily via `FunctionSchema.return_schema` cached_property (using `TypeAdapter(...).json_schema(mode='serialization')`)
- Handle `ToolReturn[T]` to extract the inner type, bare `ToolReturn` (no schema), `None` return types

**Wiring through Agent and toolsets**
- `include_return_schema` flows through `Agent.__init__`, `Agent.tool()`, `Agent.tool_plain()`, `FunctionToolset`, `MCPServer`, `FastMCPToolset`
- Agent's `_get_toolset` adds a `PreparedToolset` that resolves the flag: applies agent-level default, clears `return_schema` when not opted in, warns when opted in but no schema available

**Model layer**
- `supports_tool_return_schema: bool` on `ModelProfile`
- Google models: native support via `response_json_schema` in the API request
- Non-native models: inject return schema into tool description as JSON text fallback (in `Model.prepare_request`)

**Eliminate `from_function` path**
- Remove `FunctionSchema.function_signature` cached_property
- `ToolDefinition.__post_init__` always uses `from_schema(return_schema=self.return_schema)`
- Fix `from_schema` to use `title` from schema (preserves real class names like `User` instead of path-based `GetUserReturn`)

### Remaining work

- [ ] Make `ToolReturn` generic (`ToolReturn[T]` with `T` defaulting to `Any`) for explicit return type control
- [ ] Consider removing `from_function` classmethod entirely from `function_signature.py` (currently only used in tests)
- [ ] Add dedicated tests for return_schema flow end-to-end
- [ ] Documentation updates

### Pre-Review Checklist

### Pre-Merge Checklist

### DIFF
--- a/tests/test_toolsets.py
+++ b/tests/test_toolsets.py
@@ -1919,6 +1919,34 @@ def test_set_tool_metadata_capability_with_async_selector():
     assert td_b.metadata is None or 'tagged' not in td_b.metadata
 
 
+def test_set_tool_metadata_capability_with_bare_string_selector():
+    """SetToolMetadata with a bare string selector matches by exact name, not substring."""
+    from pydantic_ai.capabilities import SetToolMetadata
+
+    def my_tool(x: int) -> int:
+        return x
+
+    def my(x: str) -> str:
+        return x
+
+    test_model = TestModel()
+    agent = Agent(
+        test_model,
+        tools=[my_tool, my],
+        capabilities=[SetToolMetadata(tools='my_tool', tagged=True)],
+    )
+    agent.run_sync('test')
+
+    params = test_model.last_model_request_parameters
+    assert params is not None
+    td_my_tool = next(td for td in params.function_tools if td.name == 'my_tool')
+    td_my = next(td for td in params.function_tools if td.name == 'my')
+    assert td_my_tool.metadata is not None
+    assert td_my_tool.metadata['tagged'] is True
+    # 'my' should NOT match — bare string does exact match, not substring
+    assert td_my.metadata is None or 'tagged' not in td_my.metadata
+
+
 def test_set_tool_metadata_capability_with_sync_callable_selector():
     """SetToolMetadata with sync callable selector."""
     from pydantic_ai.capabilities import SetToolMetadata


pydantic_ai_slim/pydantic_ai/_function_schema.py
```diff
@@ -51,6 +56,10 @@ class FunctionSchema:
     single_arg_name: str | None = None
     positional_fields: list[str] = field(default_factory=list[str])
     var_positional_field: str | None = None
+    return_schema: ObjectJsonSchema = field(default_factory=dict[str, Any])
+    """JSON schema for the function's return type. At minimum ``{}`` (equivalent to ``Any``)."""
```
DouweM: All docstrings are MARKDOWN, no double backticks please

pydantic_ai_slim/pydantic_ai/_function_schema.py
```diff
@@ -51,6 +56,10 @@ class FunctionSchema:
...
+    function_signature: FunctionSignature | None = None
+    """Pre-computed function signature shape for this function."""
```
DouweM: - No need to state pre-computed
- If this is pre-computed, why is it nullable? explain output tools dont have it

pydantic_ai_slim/pydantic_ai/_function_schema.py
```diff
@@ -215,16 +227,47 @@ def function_schema(  # noqa: C901
...
+        warnings.warn(
+            f'Could not generate return schema for {original_func.__qualname__!r}: '
+            f'unsupported return type {return_annotation!r}. Falling back to unconstrained schema.',
+            UserWarning,
+            stacklevel=2,
+        )
+        return_schema = {}
+
+    # Compute function signature eagerly alongside the other schemas
+    from .function_signature import FunctionSignature as _FunctionSignature
```
DouweM: Any reason this can't be at the top?

pydantic_ai_slim/pydantic_ai/_function_schema.py
```diff
@@ -280,6 +323,45 @@ def _build_schema(
...
+    """Extract the type to generate a return schema for.
+
+    Always returns a type — every function has a return schema:
+    - No annotation (``None`` from ``get()``) → ``Any`` (produces ``{}``)
+    - ``-> None`` (``type(None)``) → ``type(None)`` (produces ``{"type": "null"}``)
+    - ``-> Any`` → ``Any`` (produces ``{}``)
+    - ``-> Self`` → resolved to owning class for bound methods
+    - Bare ``ToolReturn`` → ``Any`` (pre-generic legacy form)
+    - ``ToolReturn[Any]`` → ``Any`` (produces ``{}``)
+    - ``ToolReturn[T]`` → ``T``
```
DouweM: No double backticks

pydantic_ai_slim/pydantic_ai/_function_schema.py
```diff
@@ -280,6 +323,45 @@ def _build_schema(
...
+    if return_annotation is None:
+        # No annotation — untyped, same as Any
+        return Any
+    if return_annotation is type(None):
+        return type(None)
+    # Resolve Self to the owning class for bound methods.
+    # Only works when the function is already bound (e.g. instance.method);
+    # unbound methods and classmethods fall back to Any since there's no
+    # instance to infer the class from.
+    from typing_extensions import Self
```
DouweM: Import at the top

pydantic_ai_slim/pydantic_ai/_function_schema.py
```diff
@@ -280,6 +323,45 @@ def _build_schema(
...
+    - ``ToolReturn[Any]`` → ``Any`` (produces ``{}``)
+    - ``ToolReturn[T]`` → ``T``
+    - Other types → the type itself
+    """
+    if return_annotation is None:
+        # No annotation — untyped, same as Any
+        return Any
+    if return_annotation is type(None):
+        return type(None)
+    # Resolve Self to the owning class for bound methods.
```
DouweM, line 345: Move this comment closer to the actual line that does this

pydantic_ai_slim/pydantic_ai/_function_schema.py
```diff
@@ -280,6 +323,45 @@ def _build_schema(
...
+    from typing_extensions import Self
+
+    if return_annotation is Self:
+        self_obj = getattr(function, '__self__', None)
+        if self_obj is not None:
+            return cast(type[Any], type(self_obj))
+        return Any
+    if return_annotation is ToolReturn:
+        # Bare ToolReturn without type parameter — pre-generic legacy form
+        return Any
```
DouweM: Move this up to where we check is `None`  etc

pydantic_ai_slim/pydantic_ai/function_signature.py
```diff
@@ -0,0 +1,713 @@
...
+)
+
+import re
+from contextvars import ContextVar
+from dataclasses import dataclass, field
+from typing import Any, TypeAlias, cast
+
+# Set during rendering to provide the tool name for dedup prefix resolution.
+# TypeSignature.__str__ and render_definition() consult this to produce
+# prefixed names (e.g. ``tool_a_User``) without mutating the object.
```
DouweM: Single backticks in all docstrings and comments (i.e. do a search for all of them).

pydantic_ai_slim/pydantic_ai/function_signature.py
```diff
@@ -0,0 +1,713 @@
...
+# =============================================================================
+# Type expression tree
+# =============================================================================
+
+
+@dataclass
+class SimpleTypeExpr:
+    """A simple named type like `str`, `int`, `Any`, `None`."""
+
+    name: str
```
DouweM: Let's use a Literal here with the options

pydantic_ai_slim/pydantic_ai/function_signature.py
```diff
@@ -0,0 +1,713 @@
...
+class UnionTypeExpr:
+    """A union type expression like `User | None`, `str | int`."""
+
+    members: list[TypeExpr]
+
+    def __str__(self) -> str:
+        return ' | '.join(str(m) for m in self.members)
+
+
+TypeExpr: TypeAlias = 'TypeSignature | SimpleTypeExpr | LiteralTypeExpr | GenericTypeExpr | UnionTypeExpr'
```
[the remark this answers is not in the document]
DouweM: That's a good point, we need `kind` discriminators on basically every type in this file (see everything in `messages.py`)

pydantic_ai_slim/pydantic_ai/function_signature.py
```diff
@@ -0,0 +1,713 @@
...
+    return [f'{indent}"""{text}"""']
+
+
+@dataclass(kw_only=True)
+class TypeFieldSignature:
+    """A single field in a TypedDict-style type definition."""
+
+    name: str
+    type: TypeExpr
+    required: bool
```
DouweM: Let's add defaults where reasonable, i.e. here False and description = None

pydantic_ai_slim/pydantic_ai/function_signature.py
```diff
@@ -0,0 +1,713 @@
...
+    description: str | None = None
+
+    fields: dict[str, TypeFieldSignature] = field(default_factory=dict[str, TypeFieldSignature])
+
+    needs_prefix: bool = False
+    """Whether this type needs a tool-name prefix to disambiguate during rendering.
+
+    Set by ``dedup_referenced_types`` when two different tools define structurally
+    different types with the same name. The actual prefix is resolved at render time
+    from the ``_render_tool_name`` context variable — no mutation required.
```
DouweM: Is it really appropriate to keep this artifact of rendering here on the "pure" AST obj that gets cached etc? I'd prefer another contextvar, or maybe the contextvar should hold some dedupe metadata, like which specific types are dupes and need which prefix? as a dict?

pydantic_ai_slim/pydantic_ai/function_signature.py
```diff
@@ -0,0 +1,713 @@
...
+        return True
+
+
+@dataclass(kw_only=True)
+class FunctionParam:
+    """A single parameter in a function signature."""
+
+    name: str
+    type: TypeExpr
+    default: str | None
```
DouweM: Default please

pydantic_ai_slim/pydantic_ai/function_signature.py
```diff
@@ -0,0 +1,713 @@
...
+    """TypedDict class definitions needed by the signature."""
+
+    is_async: bool = False
+    """Whether the underlying function is async."""
+
+    def render(
+        self,
+        body: str,
+        *,
+        name: str,
```
DouweM: I would like to store the original `name` and `description` on the `FunctionSignature` even if that means duplication in the serialized `ToolDefinition`

pydantic_ai_slim/pydantic_ai/function_signature.py
```diff
@@ -0,0 +1,713 @@
...
+class FunctionSignature:
+    """Function signature shape with referenced type definitions.
+
+    This class holds the structural data (params, return type, referenced types)
+    needed to render a function signature. Name and description are not stored
+    here — they are provided at render time from the ``ToolDefinition``, so the
+    signature stays valid after ``dataclasses.replace(td, name=...)``.
+    """
+
+    params: dict[str, FunctionParam]
```
DouweM: Should `param` be `argument` to be more Pythonic? Or better to be more generic?

pydantic_ai_slim/pydantic_ai/function_signature.py
```diff
@@ -0,0 +1,713 @@
...
+        description: str | None = None,
+        is_async: bool | None = None,
+    ) -> str:
+        async_flag = is_async if is_async is not None else self.is_async
+        prefix = 'async def' if async_flag else 'def'
+        params_str = ', '.join(str(p) for p in self.params.values())
+
+        return_str = str(self.return_type)
+        if params_str:
+            # Force keyword-only params so LLMs always use named arguments
```
DouweM, line 282: Note that this matches the fact that `params` is just a dict, without distinguishing between positional/keyword

pydantic_ai_slim/pydantic_ai/function_signature.py
```diff
@@ -0,0 +1,713 @@
...
+        _process_schema_defs(param_defs, param_referenced, name)
+        params = _build_params_from_schema(parameters_schema, param_defs, param_referenced, name)
+
+        # Process return schema independently (its own $defs)
+        resolved_return_type: TypeExpr = _ANY
+        return_referenced: dict[str, TypeSignature] = {}
+        if return_schema is not None:
+            return_defs = return_schema.get('$defs', {})
+            _process_schema_defs(return_defs, return_referenced, name)
+            resolved_return_type = _schema_to_type_expr(return_schema, return_defs, return_referenced, name, 'Return')
```
DouweM: What is the `'Return'` here? At the very least it needs a keyword arg

tests/models/test_model_request_parameters.py
```diff
@@ -73,6 +73,14 @@ def test_model_request_parameters_are_serializable():
...
                     'defer_loading': False,
                     'prefer_builtin': None,
+                    'return_schema': None,
+                    'include_return_schema': None,
+                    'function_signature': {
+                        'params': {},
+                        'return_type': {'name': 'Any'},
+                        'referenced_types': [],
+                        'is_async': False,
+                    },
```
DouweM, line 87: Could we somehow exclude the default values (no params, no referenced types), is_async=False? Note that we override __repr__ on other dataclasses, though I'm not sure that would affect this.

Don't do this if it'd be too hacky

tests/test_tools.py
```diff
@@ -241,6 +267,17 @@ def test_docstring_numpy(docstring_format: Literal['numpy', 'auto']):
             'timeout': None,
             'defer_loading': False,
             'prefer_builtin': None,
+            'return_schema': None,
+            'include_return_schema': None,
+            'function_signature': {
+                'params': {
+                    'foo': {'name': 'foo', 'type': {'name': 'int'}, 'default': None},
+                    'bar': {'name': 'bar', 'type': {'name': 'str'}, 'default': None},
```
DouweM: Can we not include `default` here if there is none?

pydantic_ai_slim/pydantic_ai/toolsets/abstract.py
```diff
@@ -244,3 +244,29 @@ def defer_loading(self, tool_names: Sequence[str] | None = None) -> DeferredLoad
         from .deferred_loading import DeferredLoadingToolset
 
         return DeferredLoadingToolset(self, tool_names=frozenset(tool_names) if tool_names is not None else None)
+
+    def include_return_schemas(self) -> PreparedToolset[AgentDepsT]:
```
DouweM: I want an IncludeReturnSchemasToolset, like we have DeferredLoadingToolset etc. Also, since we're getting `ToolSelector` with metadata dict support, add `AbstractToolset.with_metadata()`, `SetMetadataToolset`, and `SetToolMetadata` capability.

Also: `IncludeReturnSchemas` cap should be `IncludeToolReturnSchemas`

pydantic_ai_slim/pydantic_ai/toolsets/abstract.py
```diff
@@ -244,3 +244,29 @@ def defer_loading(self, tool_names: Sequence[str] | None = None) -> DeferredLoad
...
+        Google Gemini), the schema is passed as a structured field. For other
+        models, it is injected into the tool description as JSON text.
+
+        This is the toolset-level equivalent of the
+        [`IncludeReturnSchemas`][pydantic_ai.capabilities.IncludeReturnSchemas]
+        capability, which can be used to enable return schemas across all
+        toolsets or a subset matched by a
+        [`ToolSelector`][pydantic_ai.tools.ToolSelector].
+        """
+        from dataclasses import replace as dc_replace
```
DouweM: Horrible local import, do it top level

pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -2400,19 +2382,22 @@ def _get_toolset(
         if self._prepare_tools:
             toolset = PreparedToolset(toolset, self._prepare_tools)
 
-        # Resolve include_return_schema: apply agent-level default where tool doesn't specify,
-        # and clear return_schema when not opted in.
-        agent_default = self._include_tool_return_schema
+        # Let capabilities wrap the assembled non-output toolset (e.g. IncludeReturnSchemas
+        # sets include_return_schema=True on matching tools before resolution clears the rest).
```
DouweM: Unnecessary comment

pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -2400,19 +2382,22 @@ def _get_toolset(
...
+        if run_capability is not None:
+            wrapper = run_capability.get_wrapper_toolset(toolset)
+            if wrapper is not None:
+                toolset = wrapper
 
+        # Resolve include_return_schema: clear return_schema on tools that haven't opted in.
         async def _resolve_return_schema(
```
DouweM: Is this the right place to do this? Could it be e.g. in `Model.prepare_request` where we also wipe out the `PromptedOutput` instructions if we're not going to need them?

pydantic_ai_slim/pydantic_ai/capabilities/include_return_schemas.py
```diff
@@ -0,0 +1,66 @@
...
+
+    agent = Agent('openai:gpt-5', capabilities=[IncludeReturnSchemas()])
+    ```
+    """
+
+    tools: ToolSelector[AgentDepsT] = 'all'
+    """Which tools should have their return schemas included.
+
+    - ``'all'`` (default): every tool gets its return schema included.
+    - ``Sequence[str]``: only tools whose names are listed.
```
DouweM: Also `dict`s now right?

pydantic_ai_slim/pydantic_ai/function_signature.py
```diff
@@ -100,10 +109,11 @@ def _render_description(text: str, indent: str = '') -> list[str]:
 class TypeFieldSignature:
     """A single field in a TypedDict-style type definition."""
 
+    kind: Literal['field'] = 'field'
```
DouweM: Kind always at the end

pydantic_ai_slim/pydantic_ai/_function_schema.py
```diff
@@ -345,15 +343,13 @@ def _extract_return_schema_type(return_annotation: Any, function: Callable[...,
...
     # instance to infer the class from.
-    from typing_extensions import Self
-
     if return_annotation is Self:
         self_obj = getattr(function, '__self__', None)
         if self_obj is not None:
             return cast(type[Any], type(self_obj))
         return Any
+    # Bare ToolReturn without type parameter — pre-generic legacy form
     if return_annotation is ToolReturn:
```
DouweM: Move this check further up in the method, with the other `return Foo` one-liners

pydantic_ai_slim/pydantic_ai/function_signature.py
```diff
@@ -193,13 +197,16 @@ class FunctionSignature:
...
-    here — they are provided at render time from the ``ToolDefinition``, so the
-    signature stays valid after ``dataclasses.replace(td, name=...)``.
+    needed to render a function signature. Name and description can be overridden
+    at render time (e.g. from a `ToolDefinition`).
     """
 
-    params: dict[str, FunctionParam]
-    """Function parameters."""
+    kind: Literal['function'] = 'function'
+    name: str = ''
```
DouweM: Never use empty strings as defaults. Isn't this really a required field? (backward compatibility is not a thing)

pydantic_ai_slim/pydantic_ai/function_signature.py
```diff
@@ -214,26 +221,32 @@ def render(
...
+            name: The function name (also used for dedup prefix resolution). Falls back to `self.name`.
+            description: Optional docstring to include. Falls back to `self.description`.
+            is_async: Override async rendering. If `None`, uses `self.is_async`.
+            prefixed_type_names: Set of type names that need tool-name prefixes (from `dedup_referenced_types`).
         """
-        token = _render_tool_name.set(name)
+        name = name or self.name
+        description = description if description is not None else self.description
+        tool_token = _render_tool_name.set(name)
+        prefix_token = _prefixed_type_names.set(prefixed_type_names)
```
DouweM: So we have `prefixed_type_names` and then `render_tool_name`, which is the prefix. So it seems like we have explicit/implicit things going on here. I suggest renaming `prefixed_type_names` to something like `conflicting_type_names` (or duplicate? some other word indicating not-unique).

Or even easier, could it be a map of old -> new type names? With the prefix included on the values?

PLAN.md
DouweM, line 1: Delete the plan

pydantic_ai_slim/pydantic_ai/function_signature.py
```diff
@@ -39,6 +44,7 @@ class SimpleTypeExpr:
     """A simple named type like `str`, `int`, `Any`, `None`."""
 
     name: str
```
DouweM: Can't we use a Literal union here? We already have a map of JSON schema types to Python types, so supposedly we know what values we may see here.

pydantic_ai_slim/pydantic_ai/capabilities/include_return_schemas.py
```diff
@@ -42,12 +42,13 @@ class IncludeReturnSchemas(AbstractCapability[AgentDepsT]):
 
     - ``'all'`` (default): every tool gets its return schema included.
     - ``Sequence[str]``: only tools whose names are listed.
+    - ``dict[str, Any]``: matches tools whose metadata deeply includes the specified key-value pairs.
```
DouweM: No double backticks mi amigo. IN ANY DOCSTRINGS IN YOUT DIFF.

pydantic_ai_slim/pydantic_ai/models/google.py
```diff
@@ -1425,6 +1425,8 @@ def _function_declaration_from_tool(tool: ToolDefinition) -> FunctionDeclaration
         description=tool.description or '',
         parameters_json_schema=json_schema,
     )
+    if tool.return_schema:
+        f['response_json_schema'] = tool.return_schema
```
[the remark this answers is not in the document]
DouweM: The Gemini REST model (`gemini.py`) is deprecated and unmaintained — the Google SDK model (`google.py`) is the supported path. Not worth fixing there.

pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -2375,12 +2382,32 @@ def _get_toolset(
...
                 toolset = wrapper
 
+        # Resolve include_return_schema: clear return_schema on tools that haven't opted in.
+        async def _resolve_return_schema(
+            ctx: RunContext[AgentDepsT], tool_defs: list[ToolDefinition]
+        ) -> list[ToolDefinition]:
+            resolved: list[ToolDefinition] = []
+            for td in tool_defs:
+                if not td.include_return_schema and td.return_schema is not None:
+                    td = dataclasses.replace(td, return_schema=None, function_signature=td.function_signature)
```
DouweM: Do we need `function_signature=td.function_signature` here? Doesn't that survive a `replace`?

PR #4967, opened 2026-04-03
Add `Instrumentation` capability; deprecate `Agent(instrument=...)`

DouweM:
- Closes #

### Pre-Review Checklist

### Pre-Merge Checklist

## Summary

Ports OTel/Logfire instrumentation from hard-wired calls in `Agent.iter()`, `ToolManager`, and `_output.py` to a first-class `Instrumentation` capability. Span emission concentrates in one place; capabilities and external code can compose around it through the standard hook pipeline. Full backward compatibility — `Agent(instrument=True)`, `Agent(instrument=InstrumentationSettings(...))`, and `Agent.instrument_all()` all keep working, internally constructing an `Instrumentation` capability.

### What changes

- New `pydantic_ai.capabilities.Instrumentation` capability emits all run, model-request, tool-execution, and output-processing spans via the `wrap_run` / `wrap_model_request` / `wrap_tool_execute` / `wrap_output_process` hooks.
- `Agent.iter()` no longer wraps the model in `InstrumentedModel` for the agent path. `_get_model()` returns the plain model; the `Instrumentation` capability provides the spans. The `isinstance(model_used, InstrumentedModel)` check survives only as a backwards-compat unwrap for users who explicitly pass `Agent(model=InstrumentedModel(...))`.
- Output-function spans unified: a single `running output function` / `execute_tool` span per output event. Emitted only when `output_context.has_function` (i.e. a function actually executes). Plain validation (no function) gets no span — matching pre-PR behavior. Replaces the inner span previously emitted from `execute_output_function`, which is now plain execution.
- `RunContext.agent` narrowed from `AbstractAgent | None` to `Agent | None` to reflect runtime reality (every set-site passes the inner `Agent`, even via `WrapperAgent`/`TemporalAgent`/`DBOSAgent`/`PrefectAgent`). `render_description` lives on `Agent` only; the `AbstractAgent` default and `WrapperAgent` passthrough are dropped.
- `CombinedCapability.__post_init__` auto-flattens nested `CombinedCapability`. Eliminates the "Conflicting positions in nested CombinedCapability" error class entirely — leaves with mixed `outermost`/`innermost` positions now participate as siblings in the outer ordering pass and end up in the correct tier independently.
- `InstrumentedModel` and `instrument_model()` get `DeprecationWarning`s when constructed by user code. Internal callers (`direct._prepare_model`) opt out via a private `_internal_use=True` flag. Removal in v2 — the follow-up PR will route `direct.model_request*` through the `Instrumentation` capability internally; brief at `pydantic-ai-notes/features/instrumentation/2026-05-11 follow-up - drop InstrumentedModel and route direct API through capability.md`.

### Trace/span behavior compared to pre-PR

- **Agent run spans:** identical (name, attributes, baggage).
- **Model request spans:** identical (attrs, metrics, tool definitions). Streaming uses the existing `stream_ready`/`stream_done` coordination in `_agent_graph.py`.
- **Regular tool execution spans:** identical.
- **Output tool spans:** the PR briefly added a new outer `execute_tool` span for tool-no-function output; the unification (final design) reverts that — tool-no-function gets no span, matching pre-PR. Tool-with-function and non-tool-with-function each get one span via `wrap_output_process`.
- **Error scenarios:** preserved — `record_exception(e, escaped=True)` + `set_status(StatusCode.ERROR)` in each `wrap_*` hook. Special-cased control-flow exceptions (`CallDeferred` / `ApprovalRequired` / `ToolRetryError`) keep the v5 contract: span left UNSET, no error event.
- **Temporal / DBOS:** snapshot diffs in `tests/test_temporal.py` and `tests/test_dbos.py` reflect only the intended changes above.

### Other capability authors

The capability hooks expose enough surface that other capabilities can add attributes to instrumentation spans without coupling to `Instrumentation`. Use either:

```python
from opentelemetry.trace import get_current_span
get_current_span().set_attribute(key, value)
```

or:

```python
import logfire
logfire.current_span().set_attribute(key, value)
```

inside `before_run` / `wrap_*` / `after_run` hooks.

## Test plan

- [x] `tests/test_logfire.py` (snapshots updated where the unification reshaped output spans)
- [x] `tests/models/test_instrumented.py` (deprecation warnings silenced via `pyproject.toml` `filterwarnings` ignore — direct usage of `InstrumentedModel` is the test subject)
- [x] `tests/test_agent.py`
- [x] `tests/test_capabilities.py` (ordering tests rewritten to reflect `CombinedCapability` auto-flatten)
- [x] `tests/test_direct.py`
- [x] `tests/test_streaming.py`
- [x] `tests/test_dbos.py::test_complex_agent_run_in_workflow` (smoke for output-span snapshot stability)
- [x] Pyright clean on changed files
- [x] Pre-commit hooks (format, lint, typecheck)


pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -1158,6 +1157,10 @@ def _merged_meta(ctx: RunContext[AgentDepsT]) -> dict[str, Any]:
         else:
             effective_capability = base_capability
 
+        # Prepend Instrumentation capability (outermost) so its spans wrap everything
+        if instrumentation_cap is not None:
+            effective_capability = CombinedCapability([instrumentation_cap, effective_capability])
```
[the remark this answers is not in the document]
DouweM: Yeah it should be possible for the user to add `caps=[Instrumentation(...)]` themselves and configure it that way, _instead of_ having to set `instrument=True` or use `InstrumentedModel`. So we should only auto-insert it if it's not there alerady (we can use a function like `visit` (see `AbstractToolet.visit`) to see if `Instrumention` is among the `root_capability`

DouweM:
_Comment by Claude (Opus 4.7), authored on @DouweM's behalf while verifying #4977's integration against this branch._

### Notes from verifying integration with #4977 (TemporalDurability capability)

I merged this branch into `temporal-durability-cap` on a throwaway branch to verify that #4977 could rebase on top once this lands. The Logfire-+-TemporalDurability span-tree parity test (currently `xfail` in #4977) revealed three issues, two of which are bugs in this PR and one which suggests an architectural simplification that would make this PR cleaner end-to-end.

#### Bug — Instrumentation prepend triggers "Conflicting positions" when combined with `position='innermost'` capabilities

The new prepend at `agent/__init__.py`:

```python
if instrumentation_cap is not None and not has_capability_type([effective_capability], InstrumentationCap):
    effective_capability = CombinedCapability([instrumentation_cap, effective_capability])
```

…wraps `effective_capability` (which is the agent's `_root_capability`, itself a `CombinedCapability`) inside another `CombinedCapability`. When `_root_capability`'s leaves include both an outermost-tier cap (auto-injected `ToolSearch`) and an innermost-tier cap (`TemporalDurability`), the outer's `__post_init__` calls `_effective_ordering` over the inner combined's leaves, sees both positions, and raises:

```
UserError: Conflicting positions in nested CombinedCapability: 'outermost' and 'innermost'
```

`_root_capability` is an organizational container (built from the user's `capabilities=[...]` list), not a deliberate user-grouped subtree, so its leaves should participate as siblings in the outer ordering pass. The fix is to splat:

```python
if instrumentation_cap is not None and not has_capability_type([effective_capability], InstrumentationCap):
    inner = (
        list(effective_capability.capabilities)
        if isinstance(effective_capability, CombinedCapability)
        else [effective_capability]
    )
    effective_capability = CombinedCapability([instrumentation_cap, *inner])
```

#4977 already applies the analogous splat to the parallel `extra_capabilities` wrap (the `CombinedCapability([base_capability, *extra_capabilities])` line, also a latent bug under the same conditions even without this PR), so the pattern is consistent with what's about to be on `main`.

#### Architectural question — why still go through `InstrumentedModel` on the agent path?

The flow today (post-this-PR) is:

1. `Agent.instrument_all(settings)` (or `Agent(instrument=...)`) sets `_instrument_default`.
2. `Agent._get_model()` calls `instrument_model(model_, instrument)` which wraps in `InstrumentedModel`.
3. `Agent.iter()` detects the `InstrumentedModel` via `isinstance`, extracts `instrumentation_settings`, builds an `Instrumentation` capability, prepends it.
4. `Instrumentation.wrap_model_request` unwraps `InstrumentedModel` from `request_context.model` so inner caps see the plain model.

That's two transformations to land in a capability-based world (wrap → detect → unwrap), and it has real costs:

- `ctx.model` (set once at `iter()` time, still wrapped) and `request_context.model` (unwrapped by `Instrumentation`) diverge. #4977 had a latent bug where `TemporalDurability.wrap_model_request` read `ctx.model` for `_find_model_id`, missing the unwrap and falling through to `infer_model(model_id)` activity-side without an api_key. (Fixed in #4977 by reading `request_context.model`, but the divergence remains a footgun for any future capability that needs the post-unwrap model.)
- `_validate_per_run_capabilities` in #4977 has to special-case `Instrumentation` (added per-run by step 3 above, so not in the bound class set) — also a workaround for the detect-and-convert pattern.

A cleaner shape:

- `Agent.instrument_all(settings)` and `Agent(instrument=...)` directly add an `Instrumentation` capability to `_root_capability` at construction (or wherever capability injection happens).
- `Agent._get_model()` no longer calls `instrument_model()` — returns the plain underlying model.
- `Agent.iter()` no longer needs the `isinstance(model_used, InstrumentedModel)` detect-and-convert block.
- `InstrumentedModel` survives only for `direct.model_request(model, instrument=True)` (no agent loop, no capability tree).

If that cleanup lands in this PR:

- The "Conflicting positions" bug above goes away by construction (no per-run prepend → no extra wrap).
- The `_validate_per_run_capabilities` special case in #4977 becomes unnecessary (Instrumentation is attached at construction → in `_bound_capability_classes`).
- `ctx.model` ≡ `request_context.model` for everyone, no unwrap dance.
- TemporalDurability doesn't have to coordinate with instrumentation at all.

If we'd rather keep this PR scoped to "port to capability with backward-compat shims" and tackle the `InstrumentedModel`-elimination as a follow-up, the splat fix above is enough to unblock #4977 (with #4977 keeping the per-run validator carve-out and the `request_context.model` read).

pydantic_ai_slim/pydantic_ai/capabilities/instrumentation.py
```diff
@@ -0,0 +1,484 @@
...
+
+                return result
+            finally:
+                _otel_detach(token)
+                if span.is_recording():
+                    # Get current messages and metadata from the result (which holds the up-to-date state).
+                    # ctx.messages/ctx.metadata may be stale because the run state is mutated during execution.
+                    if result is not None:
+                        message_history = result.all_messages()
+                        metadata = result._state.metadata  # pyright: ignore[reportPrivateUsage]
```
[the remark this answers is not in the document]
DouweM: **Claude:** Applied in 61ed5db7e — `Instrumentation.wrap_run` now reads the public `result.metadata` property instead of `result._state.metadata` with a pyright suppression.

pydantic_ai_slim/pydantic_ai/agent/abstract.py
```diff
@@ -138,6 +138,14 @@ def description(self, value: TemplateStr[AgentDepsT] | str | None) -> None:
         """Set the description of the agent."""
         raise NotImplementedError
 
+    def render_description(self, deps: AgentDepsT = None) -> str | None:  # pragma: no cover
```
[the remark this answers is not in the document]
DouweM: **Claude:** Applied in 61ed5db7e — dropped `# pragma: no cover` from both `AbstractAgent.render_description` (default impl) and `WrapperAgent.render_description` (passthrough). Coverage added to the existing `test_wrapper_agent` test: asserts the passthrough returns the rendered string, and exercises the abstract default via `AbstractAgent.render_description(wrapper_agent)` so the contract for external subclasses is locked in.

pydantic_ai_slim/pydantic_ai/capabilities/instrumentation.py
```diff
@@ -0,0 +1,484 @@
...
+        handler: WrapModelRequestHandler,
+    ) -> ModelResponse:
+        from pydantic_ai.models.instrumented import (
+            GEN_AI_REQUEST_MODEL_ATTRIBUTE,
+            GEN_AI_SYSTEM_ATTRIBUTE,
+            MODEL_SETTING_ATTRIBUTES,
+            CostCalculationFailedWarning,
+            InstrumentedModel,
+            _annotate_tool_call_otel_metadata,  # pyright: ignore[reportPrivateUsage]
+            _build_tool_definitions,  # pyright: ignore[reportPrivateUsage]
```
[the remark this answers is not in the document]
DouweM: **Claude:** My take: bot is right that the capability is now a stable cross-module consumer, so promoting both to public (drop the `_` prefix on `_annotate_tool_call_otel_metadata` and `_build_tool_definitions`, drop the pyright suppressions) is the principled move. Same reasoning the earlier comment had for `_build_tool_definitions`. @DouweM, want me to do this in a follow-up commit on this PR?

pydantic_ai_slim/pydantic_ai/capabilities/instrumentation.py
```diff
@@ -0,0 +1,484 @@
...
+        new_message_index = self._new_message_index
+
+        if settings.version == 1:
+            attrs: dict[str, Any] = {
+                'all_messages_events': json.dumps(
+                    [InstrumentedModel.event_to_dict(e) for e in settings.messages_to_otel_events(message_history)]
+                )
+            }
+        else:
+            last_instructions = InstrumentedModel._get_instructions(  # pyright: ignore[reportPrivateUsage]
```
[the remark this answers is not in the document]
DouweM: **Claude:** Different shape from the helpers above — `_get_instructions` is a pure message-history reducer (last `ModelRequest.instructions` joined with the prompted-output template instructions), not really a method that belongs on `InstrumentedModel`. My preference: extract it into `pydantic_ai/_instrumentation.py` (where `InstrumentationNames` already lives) rather than make it public on `InstrumentedModel`. `InstrumentedModel` is now only used by `direct.model_request(instrument=True)` and the rare `Agent(model=InstrumentedModel(...))` pre-wrap path; I'd rather not grow its public surface for a helper the agent path also wants to call. @DouweM, OK with that direction?

pydantic_ai_slim/pydantic_ai/capabilities/instrumentation.py
```diff
@@ -0,0 +1,484 @@
...
+                span.record_exception(e, escaped=True)
+                span.set_status(StatusCode.ERROR)
+                raise
+            if include_content and span.is_recording():
+                span.set_attribute(
+                    names.tool_result_attr,
+                    result if isinstance(result, str) else json.dumps(InstrumentedModel.serialize_any(result)),
+                )
+
+        return result
```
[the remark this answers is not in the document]
DouweM: **Claude:** Real DRY win — both spans share the open → try/except → record-exception → set-status pattern. Differences:
- `wrap_tool_execute` special-cases `CallDeferred` / `ApprovalRequired` / `ToolRetryError` (control flow, not errors), `wrap_output_process` doesn't.
- Result serializer differs: `tool_return_ta.dump_json` (tool-side, for `ToolReturnPart.content`) vs `InstrumentedModel.serialize_any` (output-side, for arbitrary final output).

A shared helper would take the handler, the result-serializer callable, and (optionally) a set of "control-flow" exception types to suppress. I think this is worth doing, but I'd rather take a careful pass at it as a follow-up commit on this PR than rush it. @DouweM, want me to take a stab at it before merge, or punt to a follow-up PR after this lands?

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -365,7 +364,7 @@ async def do_process(output: Any) -> Any:
     return cast(OutputDataT, result)
 
 
-async def execute_traced_output_function(
+async def execute_output_function(
```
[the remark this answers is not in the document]
DouweM: **Claude:** The bot's worry is double-spanning: `Instrumentation.wrap_output_process` now opens an outer `execute_tool` span around the output processor, and `execute_output_function` itself still creates an inner `execute_tool` span (per the docstring at line 374). I see three plausible directions:

1. Rename to `execute_output_function_traced` to keep the name honest about the inner span.
2. Move the inner span creation into `Instrumentation.wrap_output_process` so `execute_output_function` only executes (cleanest long-term, biggest scope).
3. Verify there's no double-spanning today (likely OK since output functions go through a different code path than tool calls — `OutputToolset` runs them via `process_outputs`, not via `call_tool`) and add a regression test.

(2) is the architecturally right answer but bigger; (3) is the cheapest. @DouweM, prefer (1), (2), or (3)?

pydantic_ai_slim/pydantic_ai/capabilities/instrumentation.py
```diff
@@ -0,0 +1,484 @@
...
+@dataclass
+class Instrumentation(AbstractCapability[Any]):
+    """Capability that instruments agent runs with OpenTelemetry/Logfire tracing.
+
+    When added to an agent (either explicitly via `capabilities=[Instrumentation(...)]`
+    or implicitly via `Agent(instrument=True)`), this capability creates OpenTelemetry
+    spans for the agent run, model requests, and tool executions.
+
+    Other capabilities can add attributes to these spans using the standard OpenTelemetry
+    API: `opentelemetry.trace.get_current_span().set_attribute(key, value)`.
```
DouweM: Is there a Logfire SDK equiv we could also document?

pydantic_ai_slim/pydantic_ai/capabilities/instrumentation.py
```diff
@@ -0,0 +1,484 @@
...
+        # InstrumentationSettings takes non-serializable providers (TracerProvider, etc.),
+        # so this capability is not constructible from YAML/JSON specs.
+        return None
+
+    async def for_run(self, ctx: RunContext[Any]) -> Instrumentation:
+        """Return a fresh copy for per-run state isolation."""
+        inst = replace(self)
+        inst._agent_name = (ctx.agent.name if ctx.agent else None) or 'agent'
+        inst._new_message_index = len(ctx.messages)
+        inst._instrumentation_names = InstrumentationNames.for_version(self.settings.version)
```
DouweM: I don't love having to always regenerate this (as it's not context dependent), cheap though it is -- since there should be a way to store a private instance var on a capability upon initialization and then have it survive over the `replace`. Or isn't there, without also making it public? OK, not the end of the world.

pydantic_ai_slim/pydantic_ai/capabilities/instrumentation.py
```diff
@@ -0,0 +1,484 @@
...
+    # wrap_run — agent run span
+    # ------------------------------------------------------------------
+
+    async def wrap_run(
+        self,
+        ctx: RunContext[AgentDepsT],
+        *,
+        handler: WrapRunHandler,
+    ) -> AgentRunResult[Any]:
+        from pydantic_ai.models.instrumented import InstrumentedModel
```
DouweM: Could we import this at the top of the file?

pydantic_ai_slim/pydantic_ai/capabilities/instrumentation.py
```diff
@@ -0,0 +1,484 @@
...
+        agent_name = self._agent_name
+
+        span_attributes: dict[str, Any] = {
+            'model_name': ctx.model.model_name if ctx.model else 'no-model',
+            'agent_name': agent_name,
+            'gen_ai.agent.name': agent_name,
+            'gen_ai.agent.call.id': ctx.run_id or '',
+            'gen_ai.conversation.id': ctx.conversation_id or '',
+            'gen_ai.operation.name': 'invoke_agent',
+            'logfire.msg': f'{agent_name} run',
```
DouweM, line 148: Explain in your review final output exactly how the traces and spans and attributes this new route generates are identical / strictly better / backward compatible vs what we had, across full agent runs, direct model runs, and agent runs in `Temporal`. Think about error scenarios as well!

pydantic_ai_slim/pydantic_ai/capabilities/instrumentation.py
```diff
@@ -0,0 +1,484 @@
...
+
+    def _run_span_end_attributes(
+        self,
+        ctx: RunContext[Any],
+        message_history: list[ModelMessage],
+        metadata: dict[str, Any] | None,
+    ) -> dict[str, str | int | float | bool]:
+        """Compute the end-of-run span attributes."""
+        from pydantic_ai.messages import ModelRequest
+        from pydantic_ai.models.instrumented import InstrumentedModel
```
DouweM: Imports at the top of the file if at all possible, please, though it may not be.

pydantic_ai_slim/pydantic_ai/capabilities/instrumentation.py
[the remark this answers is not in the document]
DouweM: Yep keep it private on the `_instrumentation` module.

pydantic_ai_slim/pydantic_ai/capabilities/instrumentation.py
```diff
@@ -0,0 +1,484 @@
...
+                'all_messages_events': json.dumps(
+                    [InstrumentedModel.event_to_dict(e) for e in settings.messages_to_otel_events(message_history)]
+                )
+            }
+        else:
+            last_instructions = InstrumentedModel._get_instructions(  # pyright: ignore[reportPrivateUsage]
+                message_history, self._last_model_request_parameters
+            )
+            attrs = {
+                'pydantic_ai.all_messages': json.dumps(settings.messages_to_otel_messages(list(message_history))),
```
DouweM: `pydantic_core` has json methods, use them instead of the `json` module

pydantic_ai_slim/pydantic_ai/capabilities/instrumentation.py
```diff
@@ -0,0 +1,484 @@
...
+                **settings.system_instructions_attributes(last_instructions),
+            }
+
+            if new_message_index > 0:
+                attrs['pydantic_ai.new_message_index'] = new_message_index
+
+            if any(
+                (isinstance(m, ModelRequest) and m.instructions is not None and m.instructions != last_instructions)
```
DouweM, line 220: Does this `variable_intstructions` logic still work correctly now that we have `instruction_parts` as the SSOT on instructions?

pydantic_ai_slim/pydantic_ai/capabilities/instrumentation.py
```diff
@@ -0,0 +1,484 @@
...
+                for m in message_history[new_message_index:]
+            ):
+                attrs['pydantic_ai.variable_instructions'] = True
+
+        if metadata is not None:
+            attrs['metadata'] = json.dumps(InstrumentedModel.serialize_any(metadata))
```
DouweM: Same json point, across your code

pydantic_ai_slim/pydantic_ai/capabilities/instrumentation.py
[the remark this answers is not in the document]
DouweM: Let's use the `pydantic_ai._instrumentation` helper stuff as much as we need to.

pydantic_ai_slim/pydantic_ai/capabilities/instrumentation.py
```diff
@@ -0,0 +1,484 @@
...
+        handler: WrapModelRequestHandler,
+    ) -> ModelResponse:
+        from pydantic_ai.models.instrumented import (
+            GEN_AI_REQUEST_MODEL_ATTRIBUTE,
+            GEN_AI_SYSTEM_ATTRIBUTE,
+            MODEL_SETTING_ATTRIBUTES,
+            CostCalculationFailedWarning,
+            InstrumentedModel,
+            _annotate_tool_call_otel_metadata,  # pyright: ignore[reportPrivateUsage]
+            _build_tool_definitions,  # pyright: ignore[reportPrivateUsage]
```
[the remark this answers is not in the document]
DouweM: Yep exactly.

pydantic_ai_slim/pydantic_ai/capabilities/instrumentation.py
```diff
@@ -0,0 +1,484 @@
...
+        # (which includes prompted-output template instructions and is properly sorted)
+        # instead of falling back to reading `ModelRequest.instructions` from history.
+        self._last_model_request_parameters = prepared_parameters
+
+        operation = 'chat'
+        span_name = f'{operation} {model.model_name}'
+        attributes: dict[str, Any] = {
+            'gen_ai.operation.name': operation,
+            **InstrumentedModel.model_attributes(model),
+            **InstrumentedModel.model_request_parameters_attributes(prepared_parameters),
```
DouweM: I'd rather move things off of `InstrumentedModel` and into `_instrumentation`, unless these are already / need to be public.

pydantic_ai_slim/pydantic_ai/capabilities/instrumentation.py
```diff
@@ -0,0 +1,484 @@
...
+            **get_agent_run_baggage_attributes(),
```
DouweM, line 285: Is this a helper/ function? Could it be an instancevar on the capability now?

pydantic_ai_slim/pydantic_ai/capabilities/instrumentation.py
```diff
@@ -0,0 +1,484 @@
...
+                {
+                    'type': 'object',
+                    'properties': {'model_request_parameters': {'type': 'object'}},
+                }
+            ),
+        }
+
+        tool_definitions = _build_tool_definitions(prepared_parameters)
+        if tool_definitions:
+            attributes['gen_ai.tool.definitions'] = json.dumps(tool_definitions)
```
DouweM: pydantic json!

pydantic_ai_slim/pydantic_ai/capabilities/instrumentation.py
```diff
@@ -0,0 +1,484 @@
...
+                        )
+                    else:
+                        attributes_to_set['operation.cost'] = float(price_calculation.total_price)
+
+                    if response.provider_response_id is not None:
+                        attributes_to_set['gen_ai.response.id'] = response.provider_response_id
+                    if response.finish_reason is not None:
+                        attributes_to_set['gen_ai.response.finish_reasons'] = [response.finish_reason]
+                    span.set_attributes(attributes_to_set)
+                    span.update_name(f'{operation} {request_model}')
```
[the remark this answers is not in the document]
DouweM: Hmm would that have been intended behavior or accidental? Do we elsewhere treat "is_recording" as "otel fully en/disabled"? I'd expect that we do.

pydantic_ai_slim/pydantic_ai/capabilities/instrumentation.py
```diff
@@ -0,0 +1,484 @@
...
+
+                response = await handler(request_context)
+                _annotate_tool_call_otel_metadata(response, prepared_parameters)
+                finish(response)
```
DouweM, line 367: Why is there a `finish` function rather than inline?

pydantic_ai_slim/pydantic_ai/capabilities/instrumentation.py
```diff
@@ -0,0 +1,484 @@
...
+    ) -> Any:
+        settings = self.settings
+        names = self._instrumentation_names
+        include_content = settings.include_content
+
+        with settings.tracer.start_as_current_span(
+            names.get_tool_span_name(call.tool_name),
+            attributes=self._tool_span_attributes(call),
+            record_exception=False,
+            set_status_on_exception=False,
```
DouweM, line 442: I suppose these are the tracer span exception config we had before?

pydantic_ai_slim/pydantic_ai/capabilities/instrumentation.py
```diff
@@ -0,0 +1,484 @@
...
+        ) as span:
+            try:
+                tool_result = await handler(args)
+                if include_content and span.is_recording():
+                    from pydantic_ai import messages as _messages
```
DouweM: Can't be toplevel import?

pydantic_ai_slim/pydantic_ai/capabilities/instrumentation.py
```diff
@@ -0,0 +1,484 @@
...
+                    span.set_attribute(names.tool_result_attr, part.model_response())
+                span.record_exception(e, escaped=True)
+                span.set_status(StatusCode.ERROR)
+                raise
+            except BaseException as e:
+                span.record_exception(e, escaped=True)
+                span.set_status(StatusCode.ERROR)
+                raise
+
+        return tool_result
```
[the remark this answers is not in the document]
DouweM: Yeah let's do it!

pydantic_ai_slim/pydantic_ai/capabilities/instrumentation.py
```diff
@@ -0,0 +1,484 @@
...
+        self,
+        ctx: RunContext[AgentDepsT],
+        *,
+        output_context: OutputContext,
+        output: Any,
+        handler: WrapOutputProcessHandler,
+    ) -> Any:
+        """Emit an `execute_tool` span for tool-mode output processing.
+
+        Output tools no longer go through `wrap_tool_execute`, so this hook restores
```
DouweM: "no longer go through" is irrelevant context to a reader of this instrumentation cap refactor that hasn't even landed yet.

pydantic_ai_slim/pydantic_ai/_output.py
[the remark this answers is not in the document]
DouweM: We definitely shouldn't be double spanning unnecessarily? That's weird. We're landing this feature for the first time, so let's get it right. So sounds like 2. Does that make this more "capabilities native"? That's also something to lean into.

pydantic_ai_slim/pydantic_ai/capabilities/instrumentation.py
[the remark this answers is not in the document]
DouweM: Yep let's do it!

pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -1260,12 +1266,22 @@ def _merged_meta(ctx: RunContext[AgentDepsT]) -> dict[str, Any]:
...
+        else:
+            instrumentation_settings = self._resolve_instrumentation_settings()
+
+        if instrumentation_settings is not None:
+            tracer = instrumentation_settings.tracer
+            instrumentation_cap: InstrumentationCap | None = InstrumentationCap(settings=instrumentation_settings)
         else:
-            instrumentation_settings = None
             tracer = NoOpTracer()
+            instrumentation_cap = None
```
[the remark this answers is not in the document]
DouweM: Yeah I'm considering officially deprecating the `InstrumentedModel` route (this week), and then dropping it in v2 (which is dropping its first beta next week). Or is there good reason to keep it? Does it still make sense for the direct request route? Or could those potentially also take instrumentations settings directly and leverage the capability by faking calls to it or something? That's OK as long as it's all tested.

pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -1311,6 +1327,20 @@ def _merged_meta(ctx: RunContext[AgentDepsT]) -> dict[str, Any]:
...
 
+        # Prepend Instrumentation capability (outermost) so its spans wrap everything,
+        # but only if the user hasn't already added one themselves. Splat any nested
+        # `CombinedCapability` (e.g. the organizational `_root_capability` container)
+        # so its leaves participate as siblings in the outer ordering pass — wrapping
+        # it would force the outer post-init to choose a single position for the inner
+        # group, raising "Conflicting positions" when the leaves span tiers.
+        if instrumentation_cap is not None and not has_capability_type([effective_capability], InstrumentationCap):
+            inner_caps = (
+                list(effective_capability.capabilities)
```
DouweM: Would it make sense for `CombinedCapability` to automatically flatten its input?

pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -1430,228 +1460,168 @@ async def get_instructions(
...
-                            model_request_parameters=state.last_model_request_parameters,
-                        )
-                    )
-            finally:
-                run_span.end()
+                    _result = await run_capability.on_run_error(run_ctx, error=_run_error)
+                except BaseException as _on_error_exc:
+                    _run_error = _on_error_exc
+                else:
+                    await _finalize_result(_result)
```
DouweM, line 1664: Are you very confident that in this big `iter` modification, nothing of value was lost? Even across different merges of main and potentially conflict resolutions? No regressions to the run hook stuff or stream cancelation, where there have been recent commits?

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -72,17 +71,17 @@
...
     function_schema: _function_schema.FunctionSchema,
     run_context: RunContext[AgentDepsT],
     args: dict[str, Any],
     wrap_validation_errors: bool = True,
 ) -> Any:
     """Execute an output function within a traced span with error handling.
```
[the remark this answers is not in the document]
DouweM: Definitely unintentional. 1 sounds like the right direction to me: relying on capability hooks as much as possible (and improving those if needed). Help me understand exactly what would change in behavior, and why, and what it's take to fix that, and whether that behavior is even justifiable to keep at all.

pydantic_ai_slim/pydantic_ai/_output.py
```diff
@@ -391,10 +391,13 @@ async def execute_traced_output_function(
         ToolRetryError: When wrap_validation_errors is True and a ModelRetry is caught
         ModelRetry: When wrap_validation_errors is False and a ModelRetry occurs
     """
+    from pydantic_core import to_json
+
+    from pydantic_ai._instrumentation import InstrumentationNames, get_agent_run_baggage_attributes
```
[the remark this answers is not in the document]
DouweM: Yep move to the top...?

pydantic_ai_slim/pydantic_ai/run.py
```diff
@@ -26,6 +26,27 @@
     from .result import FinalResult
 
 
+def _current_otel_traceparent() -> str | None:
```
DouweM: This feels like something for `_utils` or `_instrumentation`

pydantic_ai_slim/pydantic_ai/agent/abstract.py
```diff
@@ -138,6 +138,14 @@ def description(self, value: TemplateStr[AgentDepsT] | str | None) -> None:
         """Set the description of the agent."""
         raise NotImplementedError
 
+    def render_description(self, deps: AgentDepsT = None) -> str | None:
```
DouweM: Hmm do we need this to be on `AbstractAgent`? Just `Agent` is not enough?

tests/test_logfire.py
```diff
@@ -722,7 +725,14 @@ class MyOutput:
                         'id': 0,
                         'name': 'invoke_agent my_agent',
                         'message': 'my_agent run',
-                        'children': [{'id': 1, 'name': 'chat test', 'message': 'chat test'}],
+                        'children': [
+                            {'id': 1, 'name': 'chat test', 'message': 'chat test'},
+                            {
+                                'id': 2,
+                                'name': 'execute_tool final_result',
+                                'message': 'running tool: final_result',
```
DouweM: Help me understand all the changes in these traces diffs -- that suggest actual change in behavior worth discussing.

tests/test_logfire.py
```diff
@@ -1083,30 +1099,25 @@ class MyOutput:
 
 
 def test_instrument_all():
-    model = TestModel()
     agent = Agent()
 
-    def get_model():
-        return agent._get_model(model)  # type: ignore
+    def resolve():
+        return agent._resolve_instrumentation_settings()  # type: ignore[reportPrivateUsage]
```
DouweM, line 1119: Does this work correctly if a user manually inserts an `Instrumentation` capability with settings, instead of going through `instrument=True` at all? Can we discuss deprecating `instrument` in favor of passing via capabilities? Does that work with `agent.run`-time capabilities as well?

tests/test_logfire.py
```diff
@@ -1502,8 +1514,15 @@ def call_tool(_: list[ModelMessage], info: AgentInfo) -> ModelResponse:
...
-                            'name': 'running output function',
-                            'message': 'running output function: final_result',
+                            'name': 'running tool',
+                            'message': 'running tool: final_result',
+                            'children': [
+                                {
+                                    'id': 3,
+                                    'name': 'running output function',
+                                    'message': 'running output function: final_result',
+                                }
```
DouweM: So this is what the double nesting looks like. Does every output tool get both? Or just output tool with an output function? My hunch is to have just one, never 2. And I don't know if it's output function calls / "output processing" that we should trace as if it were a tool call -- or just output tool calls no matter whether they have processing at all. Can you find the original justification for the output function span to have been added? There's a PR I'm sure.

pydantic_ai_slim/pydantic_ai/_run_context.py
```diff
@@ -39,7 +39,7 @@ class RunContext(Generic[RunContextAgentDepsT]):
     """The model used in this run."""
     usage: RunUsage
     """LLM usage associated with the run."""
-    agent: AbstractAgent[RunContextAgentDepsT, Any] | None = field(default=None, repr=False)
+    agent: Agent[RunContextAgentDepsT, Any] | None = field(default=None, repr=False)
```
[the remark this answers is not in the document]
DouweM: Intentional, yes.

pydantic_ai_slim/pydantic_ai/capabilities/instrumentation.py
```diff
@@ -0,0 +1,543 @@
...
+        self._instrumentation_names = InstrumentationNames.for_version(self.settings.version)
+
+    def get_ordering(self) -> CapabilityOrdering:
+        return CapabilityOrdering(position='outermost')
+
+    @classmethod
+    def get_serialization_name(cls) -> str | None:
+        # InstrumentationSettings takes non-serializable providers (TracerProvider, etc.),
+        # so this capability is not constructible from YAML/JSON specs.
+        return None
```
DouweM: Hmm this is unfortunate, are you sure? I feel like we supported the `instrument` field on `AgentSpec` before?

pydantic_ai_slim/pydantic_ai/models/instrumented.py
```diff
@@ -38,39 +52,28 @@
...
-    """Instrument a model with OpenTelemetry/logfire."""
+    .. deprecated::
+        Use the [`Instrumentation`][pydantic_ai.capabilities.Instrumentation] capability
+        directly via `Agent(instrument=…)` or `Agent(capabilities=[Instrumentation(...)])`.
+        This helper will be removed in v2.
+    """
+    if not _internal_use:
+        warnings.warn(
+            '`pydantic_ai.models.instrumented.instrument_model` is deprecated; use the `Instrumentation` '
+            'capability instead (e.g. `Agent(instrument=...)` or `capabilities=[Instrumentation(...)]`). '
```
DouweM: The `Agent(instrument=...` route is now deprecated as well right

pydantic_ai_slim/pydantic_ai/models/instrumented.py
```diff
@@ -420,7 +379,17 @@ def __init__(
         self,
         wrapped: Model | KnownModelName,
         options: InstrumentationSettings | None = None,
+        *,
+        _internal_use: bool = False,
     ) -> None:
+        if not _internal_use:
+            warnings.warn(
+                '`pydantic_ai.models.instrumented.InstrumentedModel` is deprecated; use the '
+                '`Instrumentation` capability instead (e.g. `Agent(instrument=...)` or '
```
DouweM: We should update all these recommendations

pydantic_ai_slim/pydantic_ai/models/instrumented.py
```diff
@@ -357,59 +360,15 @@ def record_metrics(
...
-        tool_definitions.append(tool_def)
-
-    return tool_definitions
-
-
 @dataclass(init=False)
 class InstrumentedModel(WrapperModel):
     """Model which wraps another model so that requests are instrumented with OpenTelemetry.
 
+    .. deprecated::
```
DouweM: - This is now Markdown syntax.
- Why don't we have `@deprecated`?

pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -747,14 +732,23 @@ def from_spec(
...
         )
+        # `instrument` is deprecated as a top-level field on both `Agent` and `AgentSpec` in favor of an
+        # `Instrumentation` capability entry; still respect any legacy value by writing through to the
+        # attribute, which keeps the existing resolution flow in `iter()` working unchanged. Read the
+        # spec field only when set so we don't trigger Pydantic's deprecation warning for users who
+        # didn't opt in.
+        if instrument is not None:
+            agent.instrument = instrument
+        elif 'instrument' in validated_spec.model_fields_set:
+            agent.instrument = validated_spec.instrument
```
DouweM: - I don't quite understand why we're using `agent.instrument` here instead of `capabilities=[Instrumentation]`
- Should we deprecate the `agent.instrument` reader? And on `AgentSpec` too.

pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -897,14 +889,19 @@ def from_file(
...
             event_stream_handler=event_stream_handler,
             tool_timeout=tool_timeout,
             max_concurrency=max_concurrency,
             capabilities=merged_capabilities,
         )
+        # Apply the (deprecated) `Agent.from_file(instrument=...)` kwarg after construction so the
+        # warning fires exactly once here; the spec's `instrument` field has already been honored
+        # inside `from_spec` (with its own Pydantic deprecation warning if it was set).
+        if instrument is not None:
+            agent.instrument = instrument
```
DouweM: See previous comment

Issue #798, opened 2025-01-28
Allow graph nodes to be reusable and/or extendable


Finndersen:
## Issue with current Graph implementation

The current graph framework implementation using type hints and nodes returning instances of other nodes is quite nice, however having the the node execution logic and conditional edge logic in the same place means that all nodes in a graph are effectively tightly coupled together, preventing extensibility and reuse. 

I believe that the graph-based approach is most useful when nodes are re-usable building blocks that can be composed together into different configurations to customise behaviour of the system. For example, there is [work](https://github.com/pydantic/pydantic-ai/pull/725) underway to refactor the`Agent` class into a graph implementation. I would've thought that a major driver for this change would be the ability to extend & reuse the various useful nodes of the agent to customise behaviour as required (this would allow users to resolve issues like https://github.com/pydantic/pydantic-ai/issues/675, https://github.com/pydantic/pydantic-ai/pull/142, https://github.com/pydantic/pydantic-ai/issues/677).

However with the current implementation I don't think this is possible. The new graph-based agent will still effectively be an immutable tightly coupled black box. 

### Fully re-usable nodes

In order to properly allow graph nodes to be reusable building blocks, the edge routing logic would need to be de-coupled from the Node logic, which would be quite a drastic change to the framework. [See example in my comment below.](https://github.com/pydantic/pydantic-ai/issues/798#issuecomment-2621664198)

### Extensible/customisable nodes

At the very least, I think it would be quite valuable to allow overriding/substitution of nodes in a graph. One possible way to do this could be:
- Existing node is subclassed with a customised `_run()` implementation (attributes need to remain the same).
- Graph is initialised with subclassed node instead of original (along with other nodes). 
- Whenever a node returns an instance of the original parent class node, the graph framework automatically up-casts it to the custom subclassed version before executing it

YourTechBud:
I totally second this.

Technically, you can achieve reusable agents without adopting the node's and edges mechanism. My two-cents:

Here's an example of a reusable node:
- Let's say one of my nodes in the graph is query compresion, where I take a long chat history and compress it into a tiny context object. And this agent is smart enough to detect if the most recent question (user prompt) is a continuation of the previous discussion or if it starts a new discussion or not?
- I would want to reuse this node in several workflows, its so cool.

Why is it not reusable?
- Well, because this `QueryCompressionNode` will have to return the next node. That's literally the design.
- We don't need a second point here but I really love bullets and leaving at a single bullet point is weird.

[...]

PR #4980, opened 2026-04-04
Add pending message queue (`ctx.enqueue` / `agent_run.enqueue`)

### LINKED ISSUE #700: More capability for dynamic system prompt (depending on tools)
# Situation
Often the content of a system prompt depends on what tools are made available to the LLM. While the description of the tool is automatically included, it's often necessary or useful to also have specific instructions or examples relating to tool usage.

Currently, the list of tools used by an agent can be set dynamically either upon `Agent` initialisation, or each step of a run using the [prepare](https://ai.pydantic.dev/tools/#tool-prepare) tool argument.  

However, there is currently not really a clean way for adjusting the system prompt based on which tools are currently available. I don't think the existing dynamic system prompts](https://ai.pydantic.dev/agents/#system-prompts) system  supports this use case properly since `RunContext` does not contain information about what tools are currently available for the step (AFAIK).

Also, each piece of system prompt registered using `@agent.system_prompt` is added as a seperate `SystemPromptPart` in the initial message, which seems like an unnecessary overhead (I'm not sure how LLMs deal with multiple smaller parts of a prompt vs 1 larger one).

# Idea
I'm envisioning having some kind of mutable structured `SystemPrompt` object that can be dynamically updated by tools to insert appropriate instructions and usage examples. I'm not sure how opinionated we would want to be about the structure of the system prompt, but for example it could involve the following sections:

- Identity / Purpose
- Instructions / Rules
- Examples
- Context Information

And then each tool could be configured to provide extra content for any of these sections, which is dynamically added whenever it is made available.

This could also pave the way for some kind of dynamic context providing framework (ability to register various context information/data providers that can also update the system prompt).

Basically I think it could be useful to have the System Prompt be something that is more of a living, mutable structure that can be dynamically modified by things like tools, context providers, application logic, etc. 

Could even allow the LLM to refine and update its own system prompt 🤯

DouweM:
Closes #700.

## Summary

Adds a **pending message queue** primitive for injecting content into a run mid-flight from tools, capability hooks, or external code driving `agent.iter()`. Two delivery priorities:

- `'asap'` (default) — delivered at the earliest opportunity: prepended to the next `ModelRequest` via `before_model_request`, or, if the agent would otherwise terminate, used to redirect the run into one more request (matches pi-mono's drain-on-end behavior).
- `'when_idle'` — delivered only when the agent would otherwise terminate, after any `'asap'` messages. Doesn't interrupt in-flight work.

Each `enqueue` accepts:

- a `str` or `Sequence[UserContent]` (same shape as `Agent.run(user_prompt=...)`) — wrapped in a `UserPromptPart`
- a `ModelRequest` — emitted verbatim as its own message, preserving `instructions`/`metadata`/etc.

Adjacent parts-style enqueues of the same priority are merged into one synthesized `ModelRequest` at drain time, matching what the model sees on the wire.

The queue is exposed in two places:

- [`RunContext.enqueue(*content, priority='asap')`][pydantic_ai.tools.RunContext.enqueue] — for tools, hooks, and capabilities
- [`AgentRun.enqueue(*content, priority='asap')`][pydantic_ai.run.AgentRun.enqueue] — for external code driving `agent.iter()`

Drain logic lives in an auto-injected `PendingMessageDrainCapability`; nothing extra to import or configure.

## Design decisions

- **Priority names.** Researched pi-mono, Open SWE, Claude Code SendMessage, OpenAI Agents SDK #798, LangGraph interrupts, Hermes `/queue`, Claude Code proposal #30492. Only pi-mono has a named two-mode distinction (`steer` / `followUp`). Those names emerged independently in Claude Code's `/steer` proposal, but they specifically fit *user-driven* "I'm guiding the agent" scenarios. For background-tool delivery (a major use case), "steering" doesn't fit semantically — `'asap'` / `'when_idle'` describe timing without implying intent. No shipping system uses those plain-English names, but no shipping system has Pydantic AI's part-type richness either, so we're setting precedent rather than matching one.
- **`'asap'` drains at end-of-run too.** Pi-mono confirmed: `agent-loop.ts:253` unconditionally polls steering messages even when the agent would terminate. If a background task completes during the final step's `after_model_request` (after `before_model_request` already drained), the message would otherwise be lost. End-of-run drain catches it and redirects.
- **No `SystemPromptPart` in `EnqueueContent`.** Anthropic and Google hoist any mid-conversation `SystemPromptPart` to their top-level `system=` / `system_instruction=` parameter regardless of position, which invalidates prefix cache and loses positional intent. To inject system-style content mid-run, wrap a `SystemPromptPart` in a `ModelRequest` passthrough — that makes the cross-provider quirk visible at the call site. Follow-up #5437 proposes rendering non-leading `SystemPromptPart`s as XML-wrapped `UserPromptPart`s on Anthropic/Google to make this work cleanly.

## Use cases

The underlying primitive for several patterns the team has been talking about — **steering messages** (#89), **follow-up delivery for long-running work** (also #89), **context injection** (#700), **channels for external events** (harness #37), and **background tools** (covered separately, see below).

## What's NOT in this PR

The earlier draft also added `@agent.tool(background=True)` plus a `BackgroundToolCapability`. That's been removed and now lives in `pydantic-ai-harness` (pydantic/pydantic-ai-harness#222). Splitting it out keeps this PR scoped to the general-purpose primitive and lets the background tools experiment iterate at the harness's faster cadence.

## Known limitation: Temporal / DBOS

Inside a Temporal (or DBOS) workflow, tools run in **activities** that receive a serialized snapshot of the run state. Mutations there don't round-trip back to the workflow, so **`ctx.enqueue` from inside a tool running in a Temporal activity is dropped on the floor**. Even adding `pending_messages` to the serialize list wouldn't fix it; the activity would need to return enqueued parts as part of its result and the workflow-side wrapper would need to merge them back in.

`AgentRun.enqueue` from the workflow context still works fine. Documented in `docs/message-history.md`; worth a follow-up issue once this lands.

## Test plan

- [x] Scenarios A/B/C/D snapshot-tested: tool-body `'asap'` enqueue; external `'when_idle'` enqueue via `AgentRun`; rich `message_history`-tail + `'asap'` enqueue; late-`'asap'` (background-tools pattern) drained at end-of-run via capability `after_model_request`; `'asap'` + `'when_idle'` ordering at end-of-run
- [x] `pending_messages` threaded through `Agent.system_prompt_parts` so `ctx.enqueue` from inside an `@agent.system_prompt` callback (when used with `ReinjectSystemPrompt`) reaches the live queue instead of being silently dropped
- [x] Wire-merge unit test in `tests/test_messages.py` verifies `_clean_message_history` merges adjacent `ModelRequest`s with `ToolReturnPart`/`RetryPromptPart` first; integration test in `test_capabilities.py` snapshots the un-merged `all_messages()` view
- [x] Validation: `enqueue()` with no items rejected; `ModelRequest` mixed with strings/parts rejected; `PendingMessage` directly constructed with empty payload rejected
- [x] Documentation: new "Injecting messages mid-run" section in `docs/message-history.md` covering both producer paths, both priorities, the Temporal limitation, and why `SystemPromptPart` isn't accepted directly
- [x] 100% coverage on `_pending_messages.py` and the new `messages.py` additions
- [x] Pyright strict + ruff clean


mattbrandman:
Just wanted to comment on the temporal side here. I don't think its a big deal to not have enqueues from tools supported it feels like a relatively rare case (although maybe with cooperative agents more common). But I do think theres a potential to add in signals / updates to the graph and that could allow for it albeit I haven't looked into the complexity that would introduce or how the tool would have to be aware of that.

DouweM:
Pushed merge + iteration based on the design discussion:

**Commits:**
- `c3e7341e6` Merge main into branch (resolved 52 unmerged files — mostly the `builtin_tools` → `native_tools` rename; manually merged my enqueue plumbing into `agent/__init__.py`, `agent/abstract.py`, `agent/wrapper.py`)
- `2efb668f7` Rename `'steering'` → `'asap'`, `'follow_up'` → `'when_idle'`; drop `SystemPromptPart` from `EnqueueContent`
- `12089ffb1` Add snapshot tests for `'asap'` end-of-run drain + rich `message_history` tail scenario
- `9a5f4aba7` Fix coverage gaps in enqueue tests

**Design decisions (based on research into pi-mono, Open SWE, OpenAI Agents SDK, etc.):**
- Renamed priorities: `'asap'` describes timing without implying intent, fitting both user-driven enqueues and background-tool delivery; `'when_idle'` is clearly "don't interrupt." Pi-mono uses `steer`/`followUp`, but `steer` only really fits user-driven scenarios — for background tools it's awkward.
- Extended `'asap'` semantics: also drains at end-of-run (matching pi-mono's `agent-loop.ts:253` behavior). If a background task completes during the final step, the message redirects to one more model request instead of being lost.
- Dropped `ModelRequestPart` (specifically `SystemPromptPart`) from `EnqueueContent`. Anthropic and Google hoist mid-conversation `SystemPromptPart` to their top-level `system=` parameter, which invalidates prefix cache and loses positional intent. `EnqueueContent` is now `str | Sequence[UserContent] | ModelRequest`. The `SystemPromptPart` provider-mapping question is tracked as #5437 (proposes rendering non-leading `SystemPromptPart`s as XML-wrapped `UserPromptPart`s for those providers).

**Note for the [`background-tools` PR in pydantic-ai-harness](https://github.com/pydantic/pydantic-ai-harness/pull/222):** It currently uses `ctx.enqueue(SystemPromptPart(...), priority='follow_up')`. Two changes needed there:
1. `SystemPromptPart` no longer accepted by `enqueue` — switch to a plain string (or `ModelRequest` passthrough if you specifically need the system role).
2. `'follow_up'` is the wrong semantic for "deliver the background result ASAP." Switch to `'asap'` (now the default) — the new extended semantics handle the "agent already terminated" case correctly via end-of-run redirect.

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -2335,6 +2335,99 @@ def provider_request_id(self) -> str | None:
...
+- [`ModelRequest`][pydantic_ai.messages.ModelRequest]: emitted verbatim as its own message at
+    drain time — pass a complete request when you need to control `instructions`, `metadata`,
+    or other request-level fields, or when you need a part type other than
+    [`UserPromptPart`][pydantic_ai.messages.UserPromptPart]. Must be the only positional argument to `enqueue`.
+
+[`SystemPromptPart`][pydantic_ai.messages.SystemPromptPart] is intentionally excluded:
+Anthropic and Google hoist any `SystemPromptPart` (regardless of position) to their
+top-level system parameter, which invalidates prefix cache and loses positional intent.
+If you really need to inject a `SystemPromptPart` mid-run, wrap it in a `ModelRequest` and
+pass that as the payload — but be aware of the cross-provider behavior.
```
DouweM: Link to https://github.com/pydantic/pydantic-ai/issues/5437, I imagine we'll add back support for `ModelRequestPart` and `Sequence[...]` then

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -2335,6 +2335,99 @@ def provider_request_id(self) -> str | None:
...
+"""
+
+
+def build_enqueue_payload(items: Sequence[EnqueueContent]) -> tuple[ModelRequestPart, ...] | ModelRequest:
```
DouweM: No need for this to be public right?

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -2335,6 +2335,99 @@ def provider_request_id(self) -> str | None:
...
+    drained at the appropriate time during the agent run.
+
+    The `payload` is either a tuple of parts (merged with adjacent parts-style payloads of the
+    same priority into one synthesized [`ModelRequest`][pydantic_ai.messages.ModelRequest] at
+    drain time — matching what the model sees) or a complete `ModelRequest`
+    (emitted verbatim as its own message).
+
+    Unlike `ModelRequest` / `ModelResponse` and the part types in this module, this is
+    a plain `dataclass` rather than a Pydantic `BaseModel`: pending messages are
+    transient runtime state that's drained mid-run and never persisted in the
```
DouweM: Eeeh Model req/response are not BaseModels either. This part of the docstring can be dropped entirely.

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -2335,6 +2335,99 @@ def provider_request_id(self) -> str | None:
...
+        raise ValueError('enqueue requires at least one item')
+    if len(items) == 1 and isinstance(items[0], ModelRequest):
+        return items[0]
+    if any(isinstance(item, ModelRequest) for item in items):
+        raise ValueError('ModelRequest must be enqueued alone, not mixed with strings or `Sequence[UserContent]` items')
+    return tuple(UserPromptPart(content=item) for item in items)  # type: ignore[arg-type]
+
+
+@dataclass
+class PendingMessage:
```
DouweM: Does this need to be a public type? Do we expose `pending_messages` to the user? graph deps/state doesn't count, that's internal. I'd prefer to keep private if we can.

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -2335,6 +2335,99 @@ def provider_request_id(self) -> str | None:
...
+    drain time — matching what the model sees) or a complete `ModelRequest`
+    (emitted verbatim as its own message).
+
+    Unlike `ModelRequest` / `ModelResponse` and the part types in this module, this is
+    a plain `dataclass` rather than a Pydantic `BaseModel`: pending messages are
+    transient runtime state that's drained mid-run and never persisted in the
+    serializable message history.
+    """
+
+    payload: tuple[ModelRequestPart, ...] | ModelRequest
```
DouweM: Would it make sense to always store a pre-built `ModelRequest`? Especially if this is just internal.

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -2335,6 +2335,99 @@ def provider_request_id(self) -> str | None:
...
+
+    - `'asap'`: at the earliest opportunity (next model request, or redirect if the agent
+        would otherwise terminate).
+    - `'when_idle'`: only when the agent would otherwise terminate, after `'asap'` messages.
+    """
+
+    def __post_init__(self) -> None:
+        if isinstance(self.payload, ModelRequest):
+            if not self.payload.parts:
+                raise ValueError('PendingMessage requires a request with at least one part')
```
DouweM: Do things really break if a 0-length model request is provided? Could we filter/ignore these at a later point?

pydantic_ai_slim/pydantic_ai/_run_context.py
```diff
@@ -99,6 +100,16 @@ class RunContext(Generic[RunContextAgentDepsT]):
     `after_model_request`). Currently `None` in tool hooks, output validators,
     and during agent construction.
     """
+    pending_messages: list[PendingMessage] = field(default_factory=list[PendingMessage])
```
DouweM: Ah now I see where this is public. It does make sense here if users want to inspect it, but do they really need to be able to? Is that required for the harness features we've worked on? I prefer starting with things private, and making them public only if we have a use case. So I'm ok dropping this `pending_messages` as a public field. Maybe `enqueue` should not write onto the `RunContext`, but rather write right onto the graph deps/state?

pydantic_ai_slim/pydantic_ai/capabilities/_pending_messages.py
```diff
@@ -0,0 +1,189 @@
...
+    ) -> ModelRequestContext:
+        """Drain `'asap'` messages into the upcoming model request.
+
+        Adjacent parts-style payloads merge into one synthesized
+        [`ModelRequest`][pydantic_ai.messages.ModelRequest]; each passthrough
+        `ModelRequest` payload becomes its own message. Each resulting request is
+        appended to both `request_context.messages` (so the model sees it this
+        step) and `ctx.messages` (so it persists in the agent's message history).
+        """
+        drained = _drain_by_priority(ctx.pending_messages, 'asap')
```
DouweM, line 100: Hmm if we implement this as a capability, then I do see why we use `ctx` to ship around the messages. Not the end of the world to have the types public then, I suppose. I do like the idea of pre-packaging things into `ModelRequest` earlier in the pipeline; what do you think? So that `PendingMessage` can have a much easier type.

pydantic_ai_slim/pydantic_ai/capabilities/_pending_messages.py
```diff
@@ -0,0 +1,189 @@
...
+        steering / follow-up turns). On the wire, `_clean_message_history` re-merges
+        adjacent requests with compatible instructions, so the model still sees one turn.
+
+        The last resulting request becomes the redirect
+        [`ModelRequestNode`][pydantic_ai._agent_graph.ModelRequestNode]'s request; any
+        earlier ones are appended to `ctx.messages` so they appear in history before the
+        redirect.
+        """
+        from pydantic_ai._agent_graph import ModelRequestNode
+        from pydantic_graph import End
```
DouweM: Do we have to import these here inline?

pydantic_ai_slim/pydantic_ai/run.py
```diff
@@ -411,6 +411,49 @@ def conversation_id(self) -> str:
...
+        Messages are drained automatically: `'asap'` messages at the earliest
+        opportunity (next model request, or redirecting the run if the agent would
+        otherwise end); `'when_idle'` messages only when the agent would otherwise
+        end and no `'asap'` messages remain.
+        """
+        return self._graph_run.state.pending_messages
+
+    def enqueue(
+        self,
+        *content: _messages.EnqueueContent,
```
DouweM: Maybe this type could only live here next to `enqueue`, if it's not need on `PendingMessages` anymore

pydantic_ai_slim/pydantic_ai/run.py
```diff
@@ -411,6 +411,49 @@ def conversation_id(self) -> str:
...
+                the model sees on the wire.
+                Pass a single [`ModelRequest`][pydantic_ai.messages.ModelRequest] alone to enqueue
+                it verbatim (preserving `instructions`, `metadata`, etc.) as its own message —
+                it cannot be mixed with other items.
+            priority: When to deliver:
+                `'asap'` (default) — at the earliest opportunity (next model request,
+                    or a redirect if the agent would otherwise end).
+                `'when_idle'` — only when the agent would otherwise end, after `'asap'` messages.
+        """
+        payload = _messages.build_enqueue_payload(content)
```
DouweM: This could turn any content straight into a `ModelRequest` right

docs/message-history.md
```diff
@@ -389,6 +389,109 @@ print(result2.all_messages())
 """
 ```
 
+## Injecting messages mid-run
```
DouweM, line 394: This deserves mention in at least one more place, like in the tool and hooks docs, where `RunContext` is available, telling people they can `enqueue` from there.

docs/message-history.md
```diff
@@ -389,6 +389,109 @@ print(result2.all_messages())
...
+    - Each end-of-run redirect opens a new model request. If something keeps
+      enqueueing on every step (e.g. a tool that always enqueues, or a
+      system-prompt callback that re-enqueues on each reinjection), the run will
+      loop indefinitely. Set [`UsageLimits`][pydantic_ai.usage.UsageLimits] on the
+      run as a safety net.
+    - `SystemPromptPart` isn't accepted as an enqueue item directly: Anthropic and
+      Google hoist all `SystemPromptPart`s to their top-level system parameter
+      regardless of position, which invalidates prefix cache and loses positional
+      intent. If you really need to inject system-style content mid-run, wrap it in
+      a `ModelRequest` passthrough and be aware of the cross-provider behavior.
```
DouweM: Link to the existing issue.

DouweM:
Addressed your review comments in `b60ebb161`:

- **Pre-package to `ModelRequest` at the enqueue boundary** (your `messages.py:2409` + `_pending_messages.py:123` + `run.py:454`): `build_enqueue_request` returns a single `ModelRequest` (or `None` for empty), and `PendingMessage.payload: tuple | ModelRequest` collapses to `request: ModelRequest`. `_flatten_drained` becomes `_stamped_requests` — drainer only fills in `timestamp`/`run_id`/`conversation_id` since the request is already built. Each `enqueue` call now produces its own message in `all_messages()`; wire-level `_clean_message_history` still merges adjacent ones so the model sees one turn.

- **Move types out of `messages.py`** (your `messages.py:2370` + `messages.py:2391` + `run.py:427`): new `_enqueue.py` module hosts `EnqueueContent`, `PendingMessagePriority`, `build_enqueue_request`, and `PendingMessage` — they're internal runtime state, not wire-serializable history. Dropped from the public `pydantic_ai` re-exports.

- **Drop "not BaseModel" docstring** (your `messages.py:2405`): gone with the move.

- **Allow empty enqueue** (your `messages.py:2426`): `ctx.enqueue()` and `agent_run.enqueue()` with no content are silent no-ops. `PendingMessage(request=ModelRequest(parts=[]))` no longer validates.

- **`pending_messages` private** (your `_run_context.py:103`): field on `RunContext` and `GraphAgentState` re-labelled internal in docstrings (drain capability still needs reads); `AgentRun.pending_messages` kept as a property explicitly marked internal so tests can inspect.

- **Lift inline imports** (your `_pending_messages.py:158`): `ModelRequestNode` and `End` now imported at module top. Sibling-module ordering allows it (agent/__init__ imports `_agent_graph` before `capabilities._pending_messages`).

- **Doc cross-links** (your `docs/message-history.md:392` + `:485`): `docs/tools.md` and `docs/capabilities.md` now point at the "Injecting messages mid-run" section; the limitations note links #5437 as the planned framework-level cache-safety fix for mid-conversation `SystemPromptPart`.

Side note: also merged `main` into the branch (it was ~13 commits behind, missing the `pydantic_graph.beta`→top-level move and other v2-prep). That merge happened to be the load-bearing fix for the harness's Temporal test hang on pydantic-ai-harness#243 — full 90-test Temporal suite green now on both repos.

pydantic_ai_slim/pydantic_ai/_enqueue.py
```diff
@@ -0,0 +1,94 @@
...
+- `'asap'`: Delivered at the earliest opportunity — either prepended to the next
+    [`ModelRequest`][pydantic_ai.messages.ModelRequest], or, if the agent would
+    otherwise terminate before another request, used to redirect the run into one
+    more request.
+- `'when_idle'`: Delivered only when the agent would otherwise terminate, after
+    any `'asap'` messages. Doesn't interrupt in-flight work.
+"""
+
+
+EnqueueContent: TypeAlias = 'str | Sequence[UserContent] | ModelRequest'
```
DouweM: Would it theoretically be possible to support a `Sequence[ModelMessage]`, i.e. interleaved requests and responses? I'm thinking of the synthetic "tool-search call + tool-search result" that makes automatic tool discovery work in #5230, but requires a model response + request. Maybe we just need to require it ends in a request?

I'm also thinking that i may fix #5437 before I merge this PR, so that we can support `SystemPromptPart` (`ModelRequestPart`) and sequences again.

pydantic_ai_slim/pydantic_ai/_run_context.py
```diff
@@ -116,6 +122,37 @@ def last_attempt(self) -> bool:
         """Whether this is the last attempt at running this tool before an error is raised."""
         return self.retry == self.max_retries
 
+    def enqueue(
+        self,
+        *content: EnqueueContent,
+        priority: PendingMessagePriority = 'asap',
```
DouweM, line 132: Rrequire a kwarg?

pydantic_ai_slim/pydantic_ai/_run_context.py
```diff
@@ -116,6 +122,37 @@ def last_attempt(self) -> bool:
         """Whether this is the last attempt at running this tool before an error is raised."""
         return self.retry == self.max_retries
 
+    def enqueue(
+        self,
+        *content: EnqueueContent,
```
DouweM, line 131: Do we need `Sequence` in `EnqueueContent` if we use splat here?

pydantic_ai_slim/pydantic_ai/_run_context.py
```diff
@@ -116,6 +122,37 @@ def last_attempt(self) -> bool:
...
+                accepts) is wrapped in a [`UserPromptPart`][pydantic_ai.messages.UserPromptPart]. Pass a single
+                [`ModelRequest`][pydantic_ai.messages.ModelRequest] alone to enqueue it verbatim (preserving
+                `instructions`, `metadata`, etc.) — it cannot be mixed with other items. Calling with no
+                positional args is a no-op.
+            priority: When to deliver:
+                `'asap'` (default) — at the earliest opportunity (next model request,
+                    or a redirect if the agent would otherwise end).
+                `'when_idle'` — only when the agent would otherwise end, after `'asap'` messages.
+        """
+        request = build_enqueue_request(content)
```
DouweM: How about `PendingMessage.from_content(*content, **, priority=...)`?

pydantic_ai_slim/pydantic_ai/run.py
```diff
@@ -411,6 +412,45 @@ def conversation_id(self) -> str:
...
 
+    @property
+    def pending_messages(self) -> list[PendingMessage]:
+        """Internal: live view of the queue mutated by `enqueue` and drained by [`PendingMessageDrainCapability`][pydantic_ai.capabilities._pending_messages.PendingMessageDrainCapability].
+
+        Exposed for inspection / debugging; use [`enqueue`][pydantic_ai.run.AgentRun.enqueue] to add messages.
+        """
+        return self._graph_run.state.pending_messages
+
+    def enqueue(
```
DouweM, line 436: @adtyavrdhn Here's a reason to bring the `AgentEventStream` class back: we're adding an `AgentRun.enqueue` method here that's useful to be able to call in the middle of a streaming agent run, but this one only works for `iter`, not `run_stream_events`. As a followup, consider bringing that class back (even thought we just removed it from v2-main) with the `enqueue` method, and possibly `cancel` as well at some point, right?

pydantic_ai_slim/pydantic_ai/capabilities/reinject_system_prompt.py
```diff
@@ -68,6 +68,7 @@ async def before_model_request(
             prompt=ctx.prompt,
             usage=ctx.usage,
             model_settings=ctx.model_settings,
+            pending_messages=ctx.pending_messages,
```
DouweM: This is _very_ unlikely to affect the dynamic system prompt, so let's NOT add the argument here.

docs/message-history.md
```diff
@@ -389,6 +389,111 @@ print(result2.all_messages())
...
 
+## Injecting messages mid-run
+
+Tools, capability hooks, and external code driving an agent run can inject extra
+content into the conversation mid-run via a pending message queue. Use this when
+something happens during a run that the agent should know about — a tool wants to
+add follow-up context, an external event needs to redirect the agent's plan, or
+background work needs to reach the agent when it completes.
+
+Each `enqueue` call is packaged into a single [`ModelRequest`][pydantic_ai.messages.ModelRequest]
```
DouweM: Before mentioning an impl detail like "packaged into a single request" introduce the `enqueue` method, and say on what types it exists, with links to API docs. This "single MR" thing is not even that interesting to mention.

docs/message-history.md
```diff
@@ -389,6 +389,111 @@ print(result2.all_messages())
...
+at enqueue time and drained automatically based on a `priority`:
+
+- `'asap'` (default): delivered at the earliest opportunity — prepended to the next [`ModelRequest`][pydantic_ai.messages.ModelRequest], or, if the agent would otherwise terminate before another request, used to redirect the run into one more request. Use when the new context should reach the model as soon as possible.
+- `'when_idle'`: delivered only when the agent would otherwise terminate, after any `'asap'` messages. Use when the agent shouldn't be interrupted but should pick up the new work once it's done with what it's doing.
```
DouweM: Ensure we have tests on how this behaves vis-a-vis output tools (check output doc) and other output modes.

docs/message-history.md
```diff
@@ -389,6 +389,111 @@ print(result2.all_messages())
...
+
+Each positional argument to `enqueue` accepts the same shape as `Agent.run(user_prompt=...)` —
+a `str` or `Sequence[UserContent]` is wrapped in a [`UserPromptPart`][pydantic_ai.messages.UserPromptPart]
```
DouweM: Link to `UserContent`

docs/message-history.md
```diff
@@ -389,6 +389,111 @@ print(result2.all_messages())
...
+The example drives the run with [`AgentRun.next()`][pydantic_ai.run.AgentRun.next]
+because `'when_idle'` messages are only drained when the agent would otherwise reach an
+`End` — that drain happens in `after_node_run`, which doesn't fire inside a bare
+`async for node in agent_run:` loop. `'asap'` messages are drained in
+`before_model_request` (which fires either way) and also at the same end-of-run
+point if anything arrived during the final step.
+
+!!! info "Limitations"
+    - End-of-run redirects need [`Agent.run`][pydantic_ai.agent.AbstractAgent.run] or
+      explicit [`AgentRun.next()`][pydantic_ai.run.AgentRun.next] driving — they
```
DouweM, line 486: "`Agent.iter` + `AgentRun.next` driving"

Can we detect the "misconfigured" usage, and raise an error when we had pending messages?

docs/message-history.md
```diff
@@ -389,6 +389,111 @@ print(result2.all_messages())
...
+      run as a safety net.
+    - `SystemPromptPart` isn't accepted as an enqueue item directly: Anthropic and
+      Google hoist all `SystemPromptPart`s to their top-level system parameter
+      regardless of position, which invalidates prefix cache and loses positional
+      intent. If you really need to inject system-style content mid-run, wrap it in
+      a `ModelRequest` passthrough and be aware of the cross-provider behavior.
+      [pydantic-ai#5437](https://github.com/pydantic/pydantic-ai/issues/5437) tracks
+      the framework-level fix (rendering mid-conversation `SystemPromptPart`s as
+      XML-wrapped `UserPromptPart`s on the affected providers); once it lands,
+      direct `SystemPromptPart` enqueue support will come back.
```
DouweM: As mentioned above, this may be fixed before this is released.

pydantic_ai_slim/pydantic_ai/_enqueue.py
```diff
@@ -0,0 +1,120 @@
...
+    [`ModelRequest`][pydantic_ai.messages.ModelRequest], or, if the agent would
+    otherwise terminate before another request, used to redirect the run into one
+    more request.
+- `'when_idle'`: Delivered only when the agent would otherwise terminate, after
+    any `'asap'` messages. Doesn't interrupt in-flight work.
+"""
+
+
+EnqueueContent: TypeAlias = 'str | Sequence[UserContent] | ModelRequestPart | ModelMessage'
+"""A single item accepted by [`RunContext.enqueue`][pydantic_ai.tools.RunContext.enqueue]
```
DouweM, line 43: Why not `UserContent` by itself? So you can pass multiple images for example?

pydantic_ai_slim/pydantic_ai/_enqueue.py
```diff
@@ -0,0 +1,120 @@
...
+        Raises:
+            ValueError: If the assembled messages don't end in a
+                [`ModelRequest`][pydantic_ai.messages.ModelRequest] — e.g. a lone `ModelResponse` —
+                since the agent needs a request to respond to.
+        """
+        messages = _build_enqueue_messages(content)
+        if not messages:
+            return None
+        if not isinstance(messages[-1], ModelRequest):
+            raise ValueError(
```
DouweM: We use our own UserError for this stuff

pydantic_ai_slim/pydantic_ai/_enqueue.py
```diff
@@ -0,0 +1,120 @@
...
+class PendingMessage:
+    """One or more [`ModelMessage`][pydantic_ai.messages.ModelMessage]s queued for injection into the agent conversation.
+
+    Enqueued via [`RunContext.enqueue`][pydantic_ai.tools.RunContext.enqueue] or
+    [`AgentRun.enqueue`][pydantic_ai.run.AgentRun.enqueue] and automatically drained
+    at the appropriate time during the agent run by
+    [`PendingMessageDrainCapability`][pydantic_ai.capabilities._pending_messages.PendingMessageDrainCapability].
+    """
+
+    messages: tuple[ModelMessage, ...]
```
DouweM: Just use a list, please

pydantic_ai_slim/pydantic_ai/capabilities/_pending_messages.py
```diff
@@ -0,0 +1,159 @@
...
+    """
+    messages: list[ModelMessage] = []
+    for pending in drained:
+        for message in pending.messages:
+            if message.timestamp is None:
+                message.timestamp = now_utc()
+            if message.run_id is None:
+                message.run_id = fallback_run_id
+            if message.conversation_id is None:
+                message.conversation_id = fallback_conversation_id
```
DouweM: I feel like we do this kind of copying in a few places, and i worry that when we add a new similar field, we'll forget to update one place. Any way to fix that?

pydantic_ai_slim/pydantic_ai/capabilities/_pending_messages.py
```diff
@@ -0,0 +1,159 @@
...
+            *_stamped_messages(when_idle, fallback_run_id=ctx.run_id, fallback_conversation_id=ctx.conversation_id),
+        ]
+        # `final` becomes the redirect node's request; `ModelRequestNode._prepare_request`
+        # will re-stamp it during the graph lifecycle. `_stamped_messages` already
+        # stamped it, which is harmless (the lifecycle stamp overwrites). Every
+        # `PendingMessage` ends in a `ModelRequest` (enforced by `from_content`), so the
+        # last flattened message is always one; any earlier responses/requests become
+        # `extras` appended to history before the redirect.
+        *extras, final = messages
+        assert isinstance(final, ModelRequest), 'enqueued content always ends in a ModelRequest'
```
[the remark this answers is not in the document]
DouweM: Yeah this could be a nicer error, since people _could_ touch `RunContext.pending_messages` directly.

docs/message-history.md
```diff
@@ -391,6 +391,107 @@ print(result2.all_messages())
...
+### From external code driving `agent.iter()`
+
+Use [`AgentRun.enqueue`][pydantic_ai.run.AgentRun.enqueue] when you're driving a run
+from outside (e.g. forwarding events from a webhook, chat platform, or job queue):
+
+```python {title="enqueue_from_agent_run.py"}
+from pydantic_ai import Agent
+from pydantic_graph import End
+
+agent = Agent('openai:gpt-5.2')
```
DouweM: Use Opus 4.7

docs/message-history.md
```diff
@@ -391,6 +391,107 @@ print(result2.all_messages())
...
+
+```python {title="enqueue_from_tool.py"}
+from pydantic_ai import Agent, RunContext
+
+agent = Agent('openai:gpt-5.2')
+
+
+@agent.tool
+def trigger_alert(ctx: RunContext[None]) -> str:
+    ctx.enqueue('Alert: production is degraded, prioritize triage.')
```
DouweM, line 431: Have an example with a `SystemPromptPart` as well

PR #5008, opened 2026-04-07
docs: add Rapid-MLX (Apple Silicon) as OpenAI-compatible provider

raullenchai:
## Summary

Adds a new section to the OpenAI-compatible models documentation describing
how to point PydanticAI at [Rapid-MLX](https://github.com/raullenchai/Rapid-MLX),
an OpenAI-compatible local inference server for Apple Silicon built on Apple's
MLX framework.

## Why

The existing docs cover Ollama as the canonical OpenAI-compatible local
backend. Rapid-MLX fills a gap for Apple Silicon users who want an
MLX-native server with streaming and tool calling. It works with the
existing `OpenAIChatModel` + `OpenAIProvider` pattern — no new code or
provider class needed.

## Scope

Docs only. Single file change. No code, no API surface changes, no
new dependencies.


hramezani:
Thanks @raullenchai! Two small changes before we merge:

Move the section to the end of the OpenAI-compatible Models list, right after Atlas Cloud — that's where we add newer community entries.
Trim it down a bit to match the terse style of the neighboring entries — e.g. drop the "18 tool-call parsers (hermes, llama3, …)" line and the marketing-ish phrasing; the install snippet + the OpenAIProvider example are the important parts. Also double-check the model names/ids in the example (Gemma 4, Qwen3.5, mlx-community/Qwen3.5-4B-MLX-4bit) actually resolve so the snippet is runnable.
Thanks again! 🙏

PR #5014, opened 2026-04-07
Fix deferred tool search keyword matching

### LINKED ISSUE #4994: buggy search in `search_tools` for deferred tools
### Initial Checks

### Description

I was trying to use deferred toolset from latest pydantic-ai release. seems like the current search logic is very greedy and always matches same set of results.

I saw following issues when I glanced at implementation:
- When you have prefixed toolset like `github_get_me`, `github_create_gist` etc. and you search for keywords `github profile` it's matching with every single github tool because of prefix and because of max_limit is 10, the real tool is never found.
- If you search for `get me` keyword, it's matching with tool with `github_add_com[me]nt_to_pending_review` name.

- Python: 3.12
- Pydantic AI: 1.77.0
- LLM provider SDK:

pandego:
- Closes #4994

### Pre-Review Checklist

### Pre-Merge Checklist

### Summary

`search_tools` was matching keywords as raw substrings, which made deferred tool lookup too greedy.

This tightens matching by tokenizing tool names, descriptions, and the query, then only returning tools that match every query term as a real token. That keeps broad prefixes like `github` from crowding out the tool that actually matches the rest of the query, and it avoids substring hits inside words like `comment` for the query `me`.

### Test plan

- `uv run pytest tests/test_tool_search.py -k 'prefers_specific_term_matches or does_not_match_substrings_inside_words or search_returns_matching_tools or search_matches_description or search_is_case_insensitive or search_returns_no_matches or search_empty_query or max_results'`
- `uv run ruff check pydantic_ai_slim/pydantic_ai/toolsets/_tool_search.py tests/test_tool_search.py`

### DIFF
--- a/tests/test_tool_search.py
+++ b/tests/test_tool_search.py
@@ -668,6 +668,29 @@ async def test_tool_search_toolset_discovered_tools_available():
     assert 'stock_price' not in tool_names
 
 
+async def test_tool_search_toolset_omits_search_tool_once_all_deferred_tools_are_discovered():
+    toolset = _create_function_toolset()
+    searchable = ToolSearchToolset(wrapped=toolset)
+
+    messages: list[ModelMessage] = [
+        ModelRequest(
+            parts=[
+                ToolReturnPart(
+                    tool_name=_SEARCH_TOOLS_NAME,
+                    content={'message': 'Found all deferred tools.', 'tools': []},
+                    metadata={_DISCOVERED_TOOLS_METADATA_KEY: ['calculate_mortgage', 'stock_price', 'crypto_price']},
+                )
+            ]
+        )
+    ]
+    ctx = _build_run_context(None, messages=messages)
+
+    tools = await searchable.get_tools(ctx)
+    tool_names = list(tools.keys())
+
+    assert tool_names == snapshot(['get_weather', 'get_time', 'calculate_mortgage', 'stock_price', 'crypto_price'])
+
+
 async def test_tool_search_toolset_reserved_name_collision():
     """Test that `UserError` is raised if a tool is named 'search_tools' and deferred tools exist."""
     toolset: FunctionToolset[None] = FunctionToolset()


dsfaccini:
I don't hate this, thanks for the contrib @pandego 

my concern here is: the function needs flexibility (e.g. `max_results`, `match= 'all' | 'any'`) but then we start getting into the opinionated approach again, which is what we want to avoid with the pluggable approach https://github.com/pydantic/pydantic-ai/pull/4960

this is a good search strategy to offer out of the box though, so let me sit on this for a week 🙏

slmnsh:
@dsfaccini is there any progress on this?

dsfaccini:
hey @slmnsh sorry for keeping you waiting! we discussed it an we're happy making this the default while we progress on the pluggable strategy

thanks for the contrib! I'll handle the remaining/open comments and we may be able to merge it today

dsfaccini:
Behavior summary for the current branch:

Before this PR, `search_tools` lowercased the raw query, split on spaces, and then did substring matching against the concatenated tool name + description. That meant multi-word searches behaved like loose OR matching, but it also produced false positives from substring hits inside larger words. For example, a query like `get me` could match a tool whose description contained `comment`, because `me` is a substring of `comment`. It also treated punctuation-only inputs poorly because they could devolve into an effectively empty search.

On this branch, both the query and each deferred tool's name/description are tokenized on alphanumeric runs first. Matching is then based on token overlap rather than raw substring containment, so `me` no longer matches inside `comment`, while queries like `exchange rate` still match `get_exchange_rate`. Results are ordered by overlap score, so more specific matches come first, but we still return all positive matches up to the existing `_MAX_SEARCH_RESULTS` cap instead of only returning the top score tier. We also now reject non-tokenizable inputs like `...` or `---` with `ModelRetry` rather than accidentally matching everything.

So the net behavior change is: keep broad discoverability, remove substring false positives, prefer stronger matches first, and make empty/punctuation-only searches fail cleanly.

DouweM:
Thanks all, I'll get it out today!

PR #5027, opened 2026-04-08
Drop stray post-done tool-call args delta in AG-UI adapter and `OpenAIResponsesStreamedResponse`

### LINKED ISSUE #4733: `AGUIEventStream` emits `TOOL_CALL_ARGS` after `TOOL_CALL_END` for same tool_call_id (stray delta), violating AG-UI protocol

### LINKED ISSUE #5757: OpenAI Responses: stray post-done arguments delta corrupts final tool-call `args` (model-layer follow-up to #4733)
### Background

### The remaining problem

### Proposed fix

### Scope / notes

### Related

- Protocol fix PR: #5027

dsfaccini:
- Closes #4733
- Closes #5757

## Summary

Non-standard OpenAI Responses-compatible endpoints can emit a tool-call arguments delta *after* the call is done. That corrupts the stream in two distinct places, fixed at two layers:

1. **AG-UI protocol** (#4733) — the adapter translated the stray delta into a `TOOL_CALL_ARGS` event after `TOOL_CALL_END` for the same `tool_call_id`, violating the AG-UI protocol and crashing `@ag-ui/client`.
2. **OpenAI Responses model layer** (#5757) — even with the AG-UI event suppressed, the stray delta was still applied to the underlying tool-call part by `ModelResponsePartsManager`, corrupting the final `ModelResponse` `args` into invalid JSON (`'{...}' + '}'`).

## Fix

**AG-UI adapter:** `AGUIEventStream` now records ended tool-call ids (in `handle_tool_call_end` / `handle_builtin_tool_call_end`) and drops any later `TOOL_CALL_ARGS` for an already-ended id in `handle_tool_call_delta`.

This lives in the AG-UI adapter — not the generic `iterator_with_part_end` stream wrapper — because a `PartDeltaEvent` arriving after a `PartEndEvent` is **not** universally invalid: models legitimately stream late deltas for an already-"ended" part (e.g. xAI interleaves encrypted thinking-signature deltas with tool calls). Only the AG-UI protocol imposes the hard "no args after end" constraint, so only its adapter can safely enforce it. Thinking-signature deltas route through `handle_thinking_delta`, so they're unaffected.

**OpenAI Responses:** `OpenAIResponsesStreamedResponse` now tracks function-call item ids that received a `response.function_call_arguments.done` signal and skips any later `response.function_call_arguments.delta` for them, so the stray delta never reaches the parts manager and the final `args` stay valid.

The guard keys off the provider-specific explicit "done" signal, so it belongs in the model layer. Other providers with the same shape (Anthropic `content_block_stop`, Bedrock `contentBlockStop`) are structurally susceptible only via non-conforming endpoints, and are left for a case-by-case follow-up if the pattern is ever observed.

### Checklist


dsfaccini:
local, auto and devin review all converge on no issues

PR #5029, opened 2026-04-09
fix: catch ModuleNotFoundError instead of ImportError in provider/model imports

### LINKED ISSUE #4927: Ambiguous and unhandled ImportError at module import checks
### Initial Checks

### Description

My friend @dyigitpolat reported an issue with Mistral models, which is the same as described here: [#4727](https://github.com/pydantic/pydantic-ai/issues/4727).

At first, I thought Mistral was not installed on my machine, but it was already installed. After that, I looked into the source and added a debug print to the `except` block, and I saw this error in the terminal:

```sh
File "/opt/homebrew/lib/python3.11/site-packages/pydantic_ai/models/mistral.py", line 59, in <module>
    from mistralai import (
ImportError: cannot import name 'UNSET' from 'mistralai' (unknown location)
```

The `UNSET` import is unknown and should be fixed, but the core problem here is that module import errors and name import errors are treated the same by `except ImportError`.

My recommendation:

* Use `importlib.util.find_spec` instead of manual try/except blocks to check whether the module is installed. If the spec cannot be found, raise an error indicating that the Mistral dependency group should be installed. Currently, it raises the same error both when the module is not installed and when a name cannot be found in the module namespace. These cases should be handled separately.

This should raise:
`ImportError: cannot import name 'UNSET' from 'mistralai' (unknown location)`

```python
try:
    from mistralai.types.basemodel import Unset as MistralUnset
except ImportError as e:
    # pragma: lax no cover
    raise ImportError(
        'Please install `mistral` to use the Mistral model, '
        'you can use the `mistral` optional group — `pip install "pydantic-ai-slim[mistral]"`'
    ) from e
```

How to handle:

**Approach I**

```python
import importlib.util

if not importlib.util.find_spec(package_name):
    raise ImportError(
        f'Please install `{package_name}` to use the Mistral model, '
        f'you can use the `{package_name}` optional group — `pip install "pydantic-ai-slim[{package_name}]"`'
    )

# Import required names here and let them raise errors if any are missing
```

**Approach II**

```python
import importlib.util

if not importlib.util.find_spec(package_name):
    raise ImportError(
        f'Please install `{package_name}` to use the Mistral model, '
        f'you can use the `{package_name}` optional group — `pip install "pydantic-ai-slim[{package_name}]"`'
    )

try:
    # Perform imports here
except ImportError as e:
    raise PydanticUserError(
        'An internal error occurred. Please report the traceback by filing an issue.'
    ) from e
```

```Python
from pydantic_ai import Agent
from pydantic_ai.models.mistral import MistralModel
from pydantic_ai.providers.mistral import MistralProvider

model = MistralModel(
    'mistral-large-latest', provider=MistralProvider(api_key='your-api-key', base_url='https://<mistral-provider-endpoint>') # type: ignore
)
agent = Agent(model)
```

### Logfire Trace

_No response_

- Python: 3.11.7
- Pydantic AI: 1.75.0
- LLM provider SDK: 2.1.3

lawrence3699:
- Closes #4927

### Pre-Review Checklist

### Pre-Merge Checklist

---

Per @Kludex's direction on #4927, this narrows the `except` clause in all provider and model import blocks from `ImportError` to `ModuleNotFoundError`.

**The problem:** When a provider dependency *is* installed but has an incompatible version (e.g. `mistralai` v2 dropped `UNSET` from `mistralai.client.types.basemodel`), the broad `except ImportError` catch swallows the real error and shows:

```
ImportError: Please install `mistral` to use the Mistral model, ...
```

...which is wrong -- the package *is* installed, just the wrong version.

**The fix:** `ModuleNotFoundError` (a subclass of `ImportError` since Python 3.6) is what Python's import machinery raises when a module genuinely doesn't exist. By catching only that, real `ImportError`s -- like a missing name in an installed-but-incompatible package -- now propagate with the actual traceback:

```
ImportError: cannot import name 'UNSET' from 'mistralai' (unknown location)
```

The re-raised `raise ImportError(...)` inside the except block still works correctly since it's explicitly constructed. Nothing changes for the "package not installed" path.

**Validation:** 269 provider tests pass, `ruff check` clean, `pyright` clean on affected files.

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/anthropic.py
+++ b/pydantic_ai_slim/pydantic_ai/models/anthropic.py
@@ -151,7 +151,7 @@ try:
     )
     from anthropic.types.model_param import ModelParam
 
-except ImportError as _import_error:
+except ModuleNotFoundError as _import_error:
     raise ImportError(
         'Please install `anthropic` to use the Anthropic model, '
         'you can use the `anthropic` optional group — `pip install "pydantic-ai-slim[anthropic]"`'
--- a/pydantic_ai_slim/pydantic_ai/models/cerebras.py
+++ b/pydantic_ai_slim/pydantic_ai/models/cerebras.py
@@ -16,7 +16,7 @@ try:
     from openai import AsyncOpenAI
 
     from .openai import OpenAIChatModel, OpenAIChatModelSettings
-except ImportError as _import_error:
+except ModuleNotFoundError as _import_error:
     raise ImportError(
         'Please install the `openai` package to use the Cerebras model, '
         'you can use the `cerebras` optional group — `pip install "pydantic-ai-slim[cerebras]"'
--- a/pydantic_ai_slim/pydantic_ai/models/cohere.py
+++ b/pydantic_ai_slim/pydantic_ai/models/cohere.py
@@ -56,7 +56,7 @@ try:
     )
     from cohere.core.api_error import ApiError
     from cohere.v2.client import OMIT
-except ImportError as _import_error:
+except ModuleNotFoundError as _import_error:
     raise ImportError(
         'Please install `cohere` to use the Cohere model, '
         'you can use the `cohere` optional group — `pip install "pydantic-ai-slim[cohere]"`'
--- a/pydantic_ai_slim/pydantic_ai/models/google.py
+++ b/pydantic_ai_slim/pydantic_ai/models/google.py
@@ -104,7 +104,7 @@ try:
         UrlContextMetadata,
         VideoMetadataDict,
     )
-except ImportError as _import_error:
+except ModuleNotFoundError as _import_error:
     raise ImportError(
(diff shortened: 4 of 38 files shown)


VANDRANKI (commented):
The fix is technically correct. ModuleNotFoundError is a subclass of ImportError, so catching ImportError already catches ModuleNotFoundError at runtime. The change improves precision: ImportError can also be raised for things other than a missing package (circular imports, broken extension modules). Catching it broadly in these provider guards would swallow those errors and produce a misleading "please install X" message instead of surfacing the real cause.

By narrowing to ModuleNotFoundError, only genuine missing-package cases are converted to the helpful install hint. Any other import-time failure propagates normally.

The change is uniform across all 38 files which is correct scope - the same pattern was copy-pasted everywhere so it needed the same fix everywhere.

LGTM.

Issue #3219, opened 2025-10-22
Support for `capture_run_messages` to store incomplete results


psg2:
### Description

When running Pydantic AI in the backend coupled with Vercel AI for the frontend but not exclusive to it, it's possible to interrupt the request stream in the frontend with a stop method, when this happens the request stream is interrupted with an exception asyncio.CancelledError but even wrapping the run with capture_run_messages the messages are still empty, unless the run fails with a model error from Pydantic AI itself apparently.

This leaves our app in a weird state because from the client's perspective it already received delta streams and is showing tool calls and text to the user, but we haven't persisted anything in the database yet. Given that we're running run_stream_events we could build the incomplete results and persist it, but I image that the package itself already has the code to manage the partial output.

It would be used to handle more exceptions gracefully to manage everything that was streamed from the provider prior to the error.

Slack thread: https://pydanticlogfire.slack.com/archives/C083V7PMHHA/p1760989459555679

DouweM:
@psg2 An `include_incomplete` flag on `capture_run_messages` makes sense to me. We may also want a `incomplete` field on the last message to make it's clear this is what happened.

So far we've discussed an interrupted streamed model response, but I think we should also include an incomplete model request if model response handling (i.e. tool call execution) was interrupted, so you can get the `ToolReturnPart`s for tools that already did execute completely. That `ModelRequest` is a bit easier to construct manually as the `FunctionToolResultEvent` holds the full `ToolReturnPart` (unlike the model response events that contain deltas, not full parts), but this would still be much more convenient through `capture_run_messages`.

Note that an incomplete `ModelResponse` will usually give you a valid message history to use in a new agent run, unless the interruption happened in the middle of the model generating tool call arguments, which would then be incomplete/invalid JSON. How would you suggest we handle that?

However, an incomplete `ModelRequest` in the message history will typically not be a valid message history that can be used as-is in a new agent run, as the model would complain that not all tool calls from its most recent response have results in the request.
As a workaround, you can pass `deferred_tool_results=DeferredToolResults(approvals={'<unexecuted_tool_call_id>': True, ...})` (see https://ai.pydantic.dev/deferred-tools/) to the new `agent.run*` call along with the `message_history` ending on the incomplete `ModelRequest`, which will result those unprocessed tool calls to be executed after all, while keeping the results that were already in the history (so you don't get duplicate side-effects).
[...]

PR #5031, opened 2026-04-09
fix: Allow cancelling of stream, deprecates `async for event in run_stream_events` in favor of using it with a context manager.

### LINKED ISSUE #1524: Stream cancellation: broader semantics for agent runs, tools, and event streams
### Initial Checks

### Description

When I [play with streaming mode](https://github.com/pydantic/pydantic-ai/issues/1516), I try to break early to cancel the streaming output, but I found the pydantic-ai explicitly consume the rest of the response.
I'd like to be able to cancel the response in streaming mode.

### Example Code

```python
import asyncio

from pydantic_ai import Agent
from pydantic_ai.result import StreamedRunResult


async def main(model: str, prompt: str):
    agent = Agent(model=model)
    async with agent.run_stream(prompt) as result:
        async for content in result.stream_text(delta=True):
            print(content)

            # cancel the content generation if the client stop consume it.
            break


if __name__ == "__main__":
    asyncio.run(main("deepseek:deepseek-chat", "Hello, how are you?"))
```

adtyavrdhn:
Changes :
Deprecates `async for event in run_stream_events` in favor of using it with a context manager.

- Closes #1524, #5132

### Pre-Review Checklist

### Pre-Merge Checklist


adtyavrdhn:
There are a few things I have to make sure I handle, race conditions galore because of `tasks` interleaving. There are also few SDK specific errors I need to make sure I handle after verifying Claude's research.

pydantic_ai_slim/pydantic_ai/models/google.py
```diff
@@ -1002,6 +1004,17 @@ class GeminiStreamedResponse(StreamedResponse):
...
+    async def cancel(self) -> None:
+        if self.cancelled:
+            return
+        # google.genai types this as AsyncIterator but it is an AsyncGenerator at runtime.
+        # TODO(#1524): If cancel() is called while iteration is in-flight (another task
+        # is awaiting __anext__), aclose() raises RuntimeError("asynchronous generator is
+        # already running"). Catch RuntimeError here and still set _cancelled so the
+        # wrapper iterators stop on next check.
+        await self._stream.aclose()  # type: ignore[union-attr]
+        self._cancelled = True
```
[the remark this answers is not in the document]
adtyavrdhn: Yes, I had Claude add these TODOs because I handled this in anthropic based on a Claude review but I am not sure if this is too much. 

Tracing these races is a bit ugh, I'll do that manually to confirm what we need to do here.

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -2409,7 +2412,16 @@ class OpenAIStreamedResponse(StreamedResponse):
     _has_refusal: bool = field(default=False, init=False)
     _refusal_text: str = field(default='', init=False)
 
+    async def cancel(self) -> None:
+        if self.cancelled:
+            return
+        await self._stream.close()
+        self._cancelled = True
```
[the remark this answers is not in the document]
adtyavrdhn: I think it is safer to do it later but why would close() yield? It is not a generator in any of the SDKs?

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -2409,7 +2412,16 @@ class OpenAIStreamedResponse(StreamedResponse):
...
+
     async def _get_event_iterator(self) -> AsyncIterator[ModelResponseStreamEvent]:
+        # TODO(#1524): If cancel() closes the stream during an in-flight read,
+        # httpx.StreamClosed (subclass of httpx.StreamError) will be raised here.
+        # Add: except httpx.StreamClosed: if self.cancelled: return; raise
         with _map_api_errors(self._model_name):
             if self._provider_timestamp is not None:  # pragma: no branch
                 self.provider_details = {'timestamp': self._provider_timestamp}
```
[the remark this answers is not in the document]
adtyavrdhn: Why is this specific to OpenAI streaming only?

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -1682,6 +1682,19 @@ def has_content(self) -> bool:
...
+        if isinstance(self.args, dict):
+            return False
+        if not self.args:
+            return False
+        try:
+            pydantic_core.from_json(self.args)
+            return False
+        except Exception:
+            return True
+
```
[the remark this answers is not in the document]
adtyavrdhn: Yep I don't mind making it a cached_property, good catch

pydantic_ai_slim/pydantic_ai/models/gemini.py
```diff
@@ -473,6 +473,21 @@ class GeminiStreamedResponse(StreamedResponse):
...
+        # If cancel() is called while another task is awaiting __anext__ on the
+        # same generator, aclose() raises RuntimeError("asynchronous generator
+        # is already running"). Safe to suppress: _cancelled is already set, and
+        # the generator will be cleaned up when request_stream's ACM exits.
+        try:
+            await self._stream.aclose()  # type: ignore[union-attr]
+        except RuntimeError:
+            pass
+
     async def _get_event_iterator(self) -> AsyncIterator[ModelResponseStreamEvent]:
```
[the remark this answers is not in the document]
adtyavrdhn: I'll add this

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -1871,7 +1876,23 @@ def _is_same_request(message: _messages.ModelMessage, request: _messages.ModelRe
 
 
 def _clean_message_history(messages: list[_messages.ModelMessage]) -> list[_messages.ModelMessage]:
-    """Clean the message history by merging consecutive messages of the same type."""
+    """Clean the message history by merging consecutive messages of the same type and filtering incomplete tool calls.
+
+    For interrupted responses (from stream cancellation), tool calls with incomplete arguments that have no
+    corresponding tool return or retry prompt are filtered out. If an interrupted response ends up with no
+    parts after filtering, it is dropped entirely (providers like OpenAI/Mistral reject empty assistant messages).
+    """
```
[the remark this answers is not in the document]
adtyavrdhn: I guess but I think Douwe okayed this in the other PR? I'll let Douwe comment if we need another helper here.

pydantic_ai_slim/pydantic_ai/result.py
```diff
@@ -272,7 +287,7 @@ async def _stream_text_deltas_ungrouped() -> AsyncIterator[tuple[str, int]]:
                     yield part.content, i
 
             last_text_index: int | None = None
-            async for event in self._raw_stream_response:
+            async for event in self:
```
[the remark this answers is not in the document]
adtyavrdhn: Hmm, admit that I didn't notice this. When I made the change I only looked at the types to clean it up and use the same entry point.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -645,9 +645,14 @@ async def _streaming_handler(
...
-            async for _ in agent_stream_holder[0]:
-                pass
+            # Race condition: early break from stream_text() could leave a pending
+            # anext() task from group_by_temporal while this drain runs, causing
+            # concurrent anext() on the same async generator. Mitigated by
+            # AgentStream._anext_lock which serializes all anext() calls through
+            # _events_iter(), including this drain path.
+            # Ensure stream is fully consumed for proper usage counting.
+            if not agent_stream_holder[0].cancelled:
+                await agent_stream_holder[0].drain()
```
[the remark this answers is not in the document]
adtyavrdhn: I wanted it to have single entry point of acquiring the lock, it was done because generator drain here after a break causes a crash where generator does anext() on a closed handler. This prevents that from happening but I am not adamant on this approach. I can look for alternatives.

pydantic_ai_slim/pydantic_ai/models/google.py
```diff
@@ -1002,6 +1004,22 @@ class GeminiStreamedResponse(StreamedResponse):
...
+        # __anext__ on the same generator, aclose() raises RuntimeError
+        # ("asynchronous generator is already running"). Safe to suppress:
+        # _cancelled is already set, and the generator will be cleaned up when
+        # request_stream's ACM exits.
+        try:
+            await self._stream.aclose()  # type: ignore[union-attr]
+        except RuntimeError:
+            pass
+
     async def _get_event_iterator(self) -> AsyncIterator[ModelResponseStreamEvent]:  # noqa: C901
```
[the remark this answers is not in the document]
adtyavrdhn: Yeah I'll do that.

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -1045,6 +1046,34 @@ def part_end_event(next_part: ModelResponsePart | None = None) -> PartEndEvent |
             self._event_iterator = iterator_with_part_end(iterator_with_final_event(self._get_event_iterator()))
         return self._event_iterator
 
+    async def cancel(self) -> None:
+        """Cancel the stream, stopping token generation.
+
+        Providers must override this to close the underlying HTTP/gRPC connection
+        and set ``self._cancelled = True``.
+        """
+        raise NotImplementedError(
```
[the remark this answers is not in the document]
adtyavrdhn: > If you want to keep the NotImplementedError as a forcing function for provider implementors, at minimum _SkipStreamedResponse needs a cancel() override (currently it inherits the NotImplementedError), even though Devin's analysis concluded the path is unreachable today — it's fragile to rely on that.

I can add that

> Consider making the base cancel() a no-op that just sets self._cancelled = True, and having providers that need connection cleanup override it to close the connection first. This would:

No, I don't want a no-op cancel() for models, setting a flag means nothing if the user is still being charged. I would rather it fails and we add cancellation support or flag it doesn't support cancel rather than letting users be charged without knowing they are being charged.

pydantic_ai_slim/pydantic_ai/result.py
```diff
@@ -621,17 +651,37 @@ async def validate_response_output(
...
 
+    async def cancel(self):
+        if self._stream_response is not None:  # pragma: no branch
+            await self._stream_response.cancel()
+            # Record the interrupted response in _all_messages so all_messages()
+            # includes it. is_complete guard prevents double-append if the stream
+            # was already fully consumed before cancel was called.
+            if not self.is_complete:
+                self.is_complete = True
+                self._record_response(self.response)
```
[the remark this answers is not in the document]
adtyavrdhn: This is already covered by `test_run_stream_cancel_all_messages_includes_interrupted_response` (line 3809 in `test_streaming.py`), which calls `cancel()` within `run_stream` and asserts that `all_messages()` includes the interrupted `ModelResponse` with `interrupted=True`.

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -1045,6 +1046,34 @@ def part_end_event(next_part: ModelResponsePart | None = None) -> PartEndEvent |
...
+        When ``cancel()`` closes the underlying HTTP/gRPC connection while
+        ``_get_event_iterator()`` is awaiting the next chunk, the SDK raises
+        a transport-level error (``httpx.StreamClosed``, ``urllib3.ProtocolError``,
+        etc.). When ``self.cancelled`` is ``True`` we know the error was caused
+        by our own cancellation, so we suppress it and let the async generator
+        exit cleanly via ``StopAsyncIteration``.
+        """
+        try:
+            yield
+        except Exception:
```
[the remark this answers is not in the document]
adtyavrdhn: Well we catch anything but only swallow it if it the stream was cancelled which seems fine? 
@DouweM what would you prefer in such scenarios?

pydantic_ai_slim/pydantic_ai/result.py
```diff
@@ -795,6 +845,57 @@ def is_complete(self) -> bool:
...
+    async def cancel(self) -> None:
+        """Cancel the stream, stopping token generation.
+
+        Closes the underlying generator, which triggers task cancellation
+        and HTTP connection cleanup via the generator's finally block.
+        """
+        if not self._closed:
+            self._closed = True
+            await self._generator.aclose()
+
```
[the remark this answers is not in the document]
adtyavrdhn: Type error should be fine, we are close to v2 so we can break it then and move to a context manager and I can get rid of `__aenter__` and `__aexit__` ?

I'll add the doctoring :)

We should export it, that was a miss as well.

pydantic_ai_slim/pydantic_ai/agent/abstract.py
```diff
@@ -1092,8 +1106,16 @@ async def run_agent() -> AgentRunResult[Any]:
...
         except asyncio.CancelledError as e:
             task.cancel(msg=e.args[0] if len(e.args) != 0 else None)
+            _task_cancel_sent = True
             raise
 
+        finally:
+            if not task.done() and not _task_cancel_sent:
+                task.cancel()
+
+            with contextlib.suppress(BaseException):
```
Kludex: `anyio.get_exception_cls` or something like that. Don't suppress BaseException.

adtyavrdhn: Thanks @Kludex Anything else that stands out to you?

Kludex: I didn't review the PR. This is the only thing that caught my attention. If you can, add to the CLAUDE.md. It's a common mistake by coding agents.

adtyavrdhn: Will do, although this dumb thing was done by me I think 😆

pydantic_ai_slim/pydantic_ai/agent/abstract.py
```diff
@@ -1092,8 +1106,17 @@ async def run_agent() -> AgentRunResult[Any]:
 
         except asyncio.CancelledError as e:
             task.cancel(msg=e.args[0] if len(e.args) != 0 else None)
+            _task_cancel_sent = True
             raise
 
+        finally:
+            cancelled_exc = anyio.get_cancelled_exc_class()
+            if not task.done() and not _task_cancel_sent:
```
adtyavrdhn, line 1116: For when task isn't done, someone might break out of the run_stream_events in which case we cancel it for them.

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -1064,6 +1065,34 @@ def part_end_event(next_part: ModelResponsePart | None = None) -> PartEndEvent |
...
```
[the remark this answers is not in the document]
adtyavrdhn: > httpx.StreamClosed, httpx.ReadError, RuntimeError

There are lots of errors and different kinds of it specially for `xAI`, either I check and handle all of those or we keep this because we are raising anyway for when the stream is not cancelled.

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -1064,6 +1065,34 @@ def part_end_event(next_part: ModelResponsePart | None = None) -> PartEndEvent |
             self._event_iterator = iterator_with_part_end(iterator_with_final_event(self._get_event_iterator()))
         return self._event_iterator
 
+    async def cancel(self) -> None:
+        """Cancel the stream, stopping token generation.
+
+        Providers must override this to close the underlying HTTP/gRPC connection
+        and set ``self._cancelled = True``.
+        """
+        raise NotImplementedError(
```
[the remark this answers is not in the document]
adtyavrdhn: This is closed with @DouweM , he agrees we should raise `NotImplementedError `.

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -1064,6 +1065,34 @@ def part_end_event(next_part: ModelResponsePart | None = None) -> PartEndEvent |
             self._event_iterator = iterator_with_part_end(iterator_with_final_event(self._get_event_iterator()))
         return self._event_iterator
 
+    async def cancel(self) -> None:
+        """Cancel the stream, stopping token generation.
+
+        Providers must override this to close the underlying HTTP/gRPC connection
+        and set ``self._cancelled = True``.
+        """
```
adtyavrdhn, line 1150: `self._cancelled = True` can be split, another method should call closing which can raise, not a no op still.

pydantic_ai_slim/pydantic_ai/models/__init__.py
adtyavrdhn: Import exceptions, insert into tuples and check, no gen exception

pydantic_ai_slim/pydantic_ai/agent/abstract.py
```diff
@@ -1092,8 +1106,17 @@ async def run_agent() -> AgentRunResult[Any]:
 
         except asyncio.CancelledError as e:
```
adtyavrdhn, line 1109: How is this covered? Check the tests

pydantic_ai_slim/pydantic_ai/agent/abstract.py
```diff
@@ -1092,8 +1106,17 @@ async def run_agent() -> AgentRunResult[Any]:
...
             task.cancel(msg=e.args[0] if len(e.args) != 0 else None)
+            _task_cancel_sent = True
             raise
 
+        finally:
+            cancelled_exc = anyio.get_cancelled_exc_class()
+            if not task.done() and not _task_cancel_sent:
+                task.cancel()
+
+            with contextlib.suppress(cancelled_exc):
```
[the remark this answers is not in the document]
adtyavrdhn: Worth checking out with smallish scripts on what kinda stuff can happen and we might miss out.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -1927,6 +1951,23 @@ def _clean_message_history(messages: list[_messages.ModelMessage]) -> list[_mess
             else:
                 clean_messages.append(message)
         elif isinstance(message, _messages.ModelResponse):  # pragma: no branch
+            # Filter incomplete tool calls from interrupted responses
+            if message.interrupted:
+                filtered_parts = [
+                    part
```
adtyavrdhn: Split it out into another helper

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -1820,6 +1834,13 @@ class ModelResponse:
     metadata: dict[str, Any] | None = None
     """Additional data that can be accessed programmatically by the application but is not sent to the LLM."""
 
+    interrupted: bool | None = None
```
adtyavrdhn, line 1839: https://github.com/pydantic/pydantic-ai/pull/4306/changes#diff-2eb561c8eaa8a723f1017556cce8006c42e504997c187b8b394b5e8634f91283

use this instead

pydantic_ai_slim/pydantic_ai/result.py
```diff
@@ -321,7 +339,23 @@ def __aiter__(self) -> AsyncIterator[ModelResponseStreamEvent]:
...
+        base_iter = self._agent_stream_iterator
+
+        return self._events_iter(base_iter)
+
+    async def _events_iter(
+        self, base_iter: AsyncIterator[ModelResponseStreamEvent]
+    ) -> AsyncIterator[ModelResponseStreamEvent]:
+        # TODO: Verify _anext_lock doesn't deadlock with group_by_temporal's background reader.
+        while True:
+            async with self._anext_lock:
```
adtyavrdhn, line 351: This is kinda concerning because the bot flagged a potential deadlock, should check if we need it.

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -1733,6 +1734,19 @@ def has_content(self) -> bool:
         """Return `True` if the tool call has content."""
         return self.args not in ('', {}, None)
 
+    @functools.cached_property
```
adtyavrdhn, line 1739: Just cached_property please

pydantic_ai_slim/pydantic_ai/agent/abstract.py
adtyavrdhn: Covered by `test_run_stream_events_external_task_cancellation` in `tests/test_streaming.py`; it wraps the consumer in an asyncio task and cancels it from outside, hitting this `CancelledError` branch (producer is parked on `never.wait()`, outer `task.cancel()` propagates through `yield message`).

adtyavrdhn:
Need to verify this against #5132. We should add a regression test for bare `async for _event in agent.run_stream_events(...)` cancellation and make sure cleanup has actually finished before cancellation returns, or be explicit that only the managed `async with` path guarantees that.

Edit: Verified and added a test for it.

pydantic_ai_slim/pydantic_ai/result.py
```diff
@@ -272,7 +290,7 @@ async def _stream_text_deltas_ungrouped() -> AsyncIterator[tuple[str, int]]:
                     yield part.content, i
 
             last_text_index: int | None = None
-            async for event in self._raw_stream_response:
+            async for event in self:
```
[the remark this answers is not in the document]
adtyavrdhn: Iterating self._raw_stream_response was incorrect, even if we do not need the lock iterating self._raw_stream_response was not accounting for usage limits while streaming which is incorrect behaviour.

I'll check on if we need the lock or not

pydantic_ai_slim/pydantic_ai/agent/abstract.py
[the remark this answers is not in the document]
adtyavrdhn: Tried to find this scenario but it does not look possible; had Codex attempt repros too, and they do not hit this. I would prefer to only suppress cancelled errors.

pydantic_ai_slim/pydantic_ai/result.py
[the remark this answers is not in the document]
adtyavrdhn: Yep, checked this properly now. This is intentional. Breaking out of stream_text can leave the pending group_by_temporal anext task alive for a bit, then cleanup and drain comes in and tries to iterate the same underlying stream. Without the lock I can reproduce RuntimeError: anext(): asynchronous generator is already running. Going through self also keeps the usage limit accounting on the shared entrypoint.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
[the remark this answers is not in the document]
adtyavrdhn: Checked this with a local repro. The race is real and it is not just drain in isolation; drain can overlap with the pending group_by_temporal read after an early break from stream_text. Keeping drain on the same serialized entrypoint is intentional for that reason.

pydantic_ai_slim/pydantic_ai/result.py
```diff
@@ -321,7 +339,23 @@ def __aiter__(self) -> AsyncIterator[ModelResponseStreamEvent]:
...
 
-        return self._agent_stream_iterator
+        base_iter = self._agent_stream_iterator
+
+        return self._events_iter(base_iter)
+
+    async def _events_iter(
+        self, base_iter: AsyncIterator[ModelResponseStreamEvent]
+    ) -> AsyncIterator[ModelResponseStreamEvent]:
+        # TODO: Verify _anext_lock doesn't deadlock with group_by_temporal's background reader.
```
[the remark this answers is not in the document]
adtyavrdhn: Checked this and I do not see an actual deadlock path from the lock itself. What I can reproduce without it is concurrent anext on the same async generator; with the lock one side just waits behind the current reader. I removed the TODO.

pydantic_ai_slim/pydantic_ai/result.py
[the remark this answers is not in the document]
adtyavrdhn: Following up on this one too: same conclusion as the other lock thread. The lock is doing real work here. Early break from stream_text can leave the pending group_by_temporal anext task alive briefly, then drain hits the same stream. Without the lock I can reproduce RuntimeError: anext(): asynchronous generator is already running.

pydantic_ai_slim/pydantic_ai/models/mistral.py
```diff
@@ -658,15 +659,23 @@ class MistralStreamedResponse(StreamedResponse):
...
     _timestamp: datetime = field(default_factory=_now_utc)
 
     _delta_content: str = field(default='', init=False)
 
+    async def _close_stream(self) -> None:
+        # Close the underlying httpx response directly, avoiding the fragile
+        # __aexit__(None, None, None) pattern on EventStreamAsync.
+        # TODO: Check if MistralEventStreamAsync exposes a public close() method
+        # to avoid reaching into SDK internals.
+        await self._stream.response.aclose()
```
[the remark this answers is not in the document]
adtyavrdhn: It doesn't, I checked this.

pydantic_ai_slim/pydantic_ai/models/gemini.py
```diff
@@ -474,37 +474,50 @@ class GeminiStreamedResponse(StreamedResponse):
...
+        # _stream is httpx's aiter_bytes(), an async generator at runtime.
+        # If _close_stream() is called while another task is awaiting __anext__ on
+        # the same generator, aclose() raises RuntimeError("asynchronous generator
+        # is already running"). Safe to suppress: _cancelled is already set by the
+        # base cancel(), and the generator will be cleaned up when request_stream's
+        # ACM exits.
+        try:
+            await self._stream.aclose()  # type: ignore[union-attr]
+        except RuntimeError:
+            pass
```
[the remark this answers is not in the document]
adtyavrdhn: >    if 'asynchronous generator is already running' not in str(e):

This isn't very nice

pydantic_ai_slim/pydantic_ai/models/xai.py
```diff
@@ -890,9 +891,14 @@ class XaiStreamedResponse(StreamedResponse):
 
     _model_name: str
     _response: _utils.PeekableAsyncStream[tuple[chat_types.Response, chat_types.Chunk]]
+    _stream: AsyncIterator[tuple[chat_types.Response, Any]]
     _timestamp: datetime
     _provider: Provider[AsyncClient]
 
+    async def _close_stream(self) -> None:
+        # xAI SDK uses gRPC; the stream's cancel() is synchronous.
+        self._stream.cancel()  # type: ignore[union-attr]
```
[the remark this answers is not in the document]
adtyavrdhn: Yep, it was not. Thanks Claude lol

Annoyingly they don't expose the stream for us to cancel but the generator has `aclose` which is what  I am falling back to now.

pydantic_ai_slim/pydantic_ai/models/xai.py
```diff
@@ -888,11 +889,24 @@ async def _process_streamed_response(
...
 
+    async def _close_stream(self) -> None:
+        # In xai-sdk 1.5.0, `chat.stream()` returns a Python async generator that
+        # wraps the underlying gRPC `GetCompletionChunk(...)` call.
+        #
+        # Calling `aclose()` shuts down that local async-generator wrapper and
+        # stops consumption on our side, but the SDK does not expose the inner
+        # `grpc.aio.UnaryStreamCall`, so this is not a documented transport-level
+        # RPC cancellation hook.
+        await self._stream.aclose()  # type: ignore[union-attr]
```
adtyavrdhn: Same try catch for generator, string match is fine

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -982,6 +982,15 @@ def _get_instruction_parts(
 class StreamedResponse(ABC):
     """Streamed response from an LLM when calling a tool."""
 
+    _stream_cancel_errors: ClassVar[tuple[type[BaseException], ...]] = (httpx.StreamError, httpx.TransportError)
+    """Exceptions suppressed by ``_stream_cancel_guard`` when ``cancel()`` tears down the stream.
```
DouweM: No double backticks please :)

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -982,6 +982,15 @@ def _get_instruction_parts(
 class StreamedResponse(ABC):
     """Streamed response from an LLM when calling a tool."""
 
+    _stream_cancel_errors: ClassVar[tuple[type[BaseException], ...]] = (httpx.StreamError, httpx.TransportError)
```
DouweM: @adtyavrdhn Because users may have their own Model and StreamedResponse subclasses, this becomes part of the public interface they may need to overwrite -- but it's public! I'd rather add a public method like `get_stream_cancel_errors` that's documented. Ideally none of the model/provider classes would reach into private fields of their superclasses and only use the public interface, just like other peoples' subclasses need to.

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -1060,6 +1070,51 @@ def part_end_event(next_part: ModelResponsePart | None = None) -> PartEndEvent |
...
+        """
+        if self.cancelled:
+            return
+        self._cancelled = True
+        await self._close_stream()
+
+    async def _close_stream(self) -> None:
+        """Close the underlying HTTP/gRPC connection.
+
+        Providers must override this to stop token generation (and billing)
```
DouweM: Same story as above, this has to be a public method of some kind if we recommend people override it.

Also note "providers must override this" should be "model classes..."?

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -1060,6 +1070,51 @@ def part_end_event(next_part: ModelResponsePart | None = None) -> PartEndEvent |
...
+        on the remote side. A missing override is a real bug — silently
+        setting ``_cancelled`` without closing the connection would let the
+        user believe the stream was cancelled while they continue being
+        charged.
+        """
+        raise NotImplementedError(
+            f'Stream cancellation is not implemented for {type(self).__name__}. '
+            'This provider must override `_close_stream()` to support streaming cancellation.'
+        )
```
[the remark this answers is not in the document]
DouweM: An error like this is always annoying, even if it's hard to hit, so in principle we should minimize it if we can without breaking out promise to the user: so since the intent of "cancel" is "after this point please consume no more tokens", it is consistent for `CompletedStreamedResponse` to say "ok" by doing nothing, as the work has already been done and streamed out.

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -1116,6 +1172,11 @@ def timestamp(self) -> datetime:
         """Get the timestamp of the response."""
         raise NotImplementedError()
 
+    @property
+    def cancelled(self) -> bool:
```
DouweM, line 1179: Any particular reason why we have the private field and the public property, instead of just a public field?

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -1668,10 +1669,11 @@ class AnthropicStreamedResponse(StreamedResponse):
     _response: AsyncIterable[BetaRawMessageStreamEvent]
     _provider_name: str
     _provider_url: str
+    _stream: AsyncStream[BetaRawMessageStreamEvent]
     _timestamp: datetime = field(default_factory=_utils.now_utc)
 
     async def _get_event_iterator(self) -> AsyncIterator[ModelResponseStreamEvent]:  # noqa: C901
```
DouweM, line 1675: I am now reminded that `_get_event_iterator` already exists as a "public private method" that subclasses have to implement, which is unfortunate (and smth to fix in v2!). I would prefer not to continue that trend, though. On `Model`, I believe all shared "plumbing" methods are public, like `prepare_request` and `customize_model_request_parameters`.

pydantic_ai_slim/pydantic_ai/models/gemini.py
```diff
@@ -474,37 +474,50 @@ class GeminiStreamedResponse(StreamedResponse):
...
-                        tool_name=gemini_part['function_call']['name'],
-                        args=gemini_part['function_call']['args'],
-                        tool_call_id=None,
-                    )
-                    if maybe_event is not None:  # pragma: no branch
-                        yield maybe_event
-                else:
-                    if not any([key in gemini_part for key in ['function_response', 'thought']]):
-                        raise AssertionError(f'Unexpected part: {gemini_part}')  # pragma: no cover
+        with self._stream_cancel_guard():
```
DouweM: Gemini is deprecated so I wouldn't mind dropping this, we'll drop it with v2 anyway

pydantic_ai_slim/pydantic_ai/models/google.py
```diff
@@ -996,6 +997,7 @@ class GeminiStreamedResponse(StreamedResponse):
 
     _model_name: GoogleModelName
     _response: AsyncIterator[GenerateContentResponse]
+    _stream: AsyncIterator[GenerateContentResponse]
```
DouweM, line 1000: Why 2 variables? If we really need both, let's make it clear why.

pydantic_ai_slim/pydantic_ai/models/groq.py
```diff
@@ -400,6 +400,7 @@ async def _process_streamed_response(
         return GroqStreamedResponse(
             model_request_parameters=model_request_parameters,
             _response=peekable_response,
+            _stream=response,
```
DouweM, line 403: Is this because of the peekable stuff? Then can't we have `peekable_response` pass through things like `close`?

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -1060,6 +1070,51 @@ def part_end_event(next_part: ModelResponsePart | None = None) -> PartEndEvent |
...
+        ``_get_event_iterator()`` is awaiting the next chunk, the SDK raises
+        a transport-level error (e.g. ``httpx.StreamClosed``). Subclasses
+        extend ``_stream_cancel_errors`` for non-``httpx`` transports. When
+        ``self.cancelled`` is ``True`` we know the error was caused by our
+        own cancellation, so we suppress it and let the async generator exit
+        cleanly via ``StopAsyncIteration``.
+        """
+        try:
+            yield
+        except self._stream_cancel_errors:
```
DouweM: Since we always call `_stream_cancel_guard` from the _subclass_, is there much benefit to it being defined here at the superclass at all? Now that StreamedResponse subclass handles API errors with its own internal helper, it would be easy enough to add another `except` that checks `if not self.cancelled`. Telling model implementers "implement proper error/cancelation handling" is as easy as "don't forget to call `_stream_cancel_guard`" so I don't think the second route is warranted by the effort (of adding new things to the public API) involved, as we don't auto-call it so the subclass can do what it wants. We could call it at our level, where we call `self._get_event_iterator`, but we didn't chose that route for general error handling. 

So let's make it consistent, and have each `StreamedResponse` handle its own `if not self.cancelled: raise` check, and then I think we can drop these new private/public vars from `StreamedResponse`, and leave only `cancel` and `cancelled`, right?

pydantic_ai_slim/pydantic_ai/result.py
```diff
@@ -795,6 +858,83 @@ def is_complete(self) -> bool:
         return self._streamed_run_result.is_complete
 
 
+class StreamEventsResult(Generic[OutputDataT]):
+    """Result of [`run_stream_events()`][pydantic_ai.agent.AbstractAgent.run_stream_events].
```
DouweM: The existing `Result` classes are the results of completed streaming/request etc, while this is _the ongoing stream itself_, not the result of anything. `AgentStream` already exists, how about `AgentEventStream`?

pydantic_ai_slim/pydantic_ai/result.py
```diff
@@ -795,6 +858,83 @@ def is_complete(self) -> bool:
...
+            self._closed = True
+            await self._generator.aclose()
+        return False
+
+    def __aiter__(self) -> StreamEventsResult[OutputDataT]:
+        # TODO(v2): remove standalone iteration support and require `async with`
+        if not self._managed:
+            warnings.warn(
+                'Iterating `StreamEventsResult` directly is deprecated. '
+                'Use `async with agent.run_stream_events(...) as stream:` instead '
```
DouweM: Let's include "with `async for event in stream:`"

pydantic_ai_slim/pydantic_ai/result.py
```diff
@@ -795,6 +858,83 @@ def is_complete(self) -> bool:
...
+    async def cancel(self) -> None:
+        """Cancel the stream, stopping token generation.
+
+        Closes the underlying generator, which triggers task cancellation
+        and HTTP connection cleanup via the generator's finally block.
+        """
+        if not self._closed:
+            self._closed = True
+            await self._generator.aclose()
+
```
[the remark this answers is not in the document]
DouweM: @adtyavrdhn Not sure if this one is right, but it does seem worth investigating + forwarding `aclose`

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -650,9 +655,14 @@ async def _streaming_handler(
...
-            async for _ in agent_stream_holder[0]:
-                pass
+            # Race condition: early break from stream_text() could leave a pending
+            # anext() task from group_by_temporal while this drain runs, causing
+            # concurrent anext() on the same async generator. Mitigated by
+            # AgentStream._anext_lock which serializes all anext() calls through
+            # _events_iter(), including this drain path.
+            # Ensure stream is fully consumed for proper usage counting.
+            if not agent_stream_holder[0].cancelled:
+                await agent_stream_holder[0].drain()
```
DouweM, line 665: Could/should drain itself check canceled?

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -1909,8 +1919,52 @@ def _is_same_request(message: _messages.ModelMessage, request: _messages.ModelRe
...
 
 
+def _filter_interrupted_response(
+    message: _messages.ModelResponse, processed_tool_call_ids: set[str]
+) -> _messages.ModelResponse | None:
+    """Drop incomplete tool calls from an interrupted ``ModelResponse``.
+
+    Tool calls whose ``args`` are incomplete (truncated JSON from stream cancellation)
+    and which have no corresponding ``ToolReturnPart``/``RetryPromptPart`` later in
+    history are removed. Stripping one that does have a matching return would orphan
```
DouweM, line 1929: Why is "later in history" relevant here, since we should only need to check the final part in the stream, when cancelation happened?

Edit: Interesting, so we assume there may be multiple interrupted responses in history. Not impossible... 

The fact that we include the half-finished response does seem somewhat inconsistent with how half-streamed responses from a run interrupted by an exception do _not_ show up in `capture_run_messages`, though there is a plan to change that: https://github.com/pydantic/pydantic-ai/issues/3219 (that could be nice clean up before v2).

It is also making me think that may be `{new,all}_messages` and `capture_run_messages` should have a `include_incomplete` flag or something so users can choose whether they want them or not. (In https://github.com/pydantic/pydantic-ai/pull/4306 we're adding another type of "incomplete" that users will likely want to skip when storing history, I'll bring it up with Matt in our channel).

Anyway that doesn't need to be a blocker here, but something we can think of making more consistent for v2.

docs/output.md
```diff
@@ -875,6 +875,126 @@ async def main():
 
 _(This example is complete, it can be run "as is" — you'll need to add `asyncio.run(main())` to run `main`)_
 
+### Cancelling Streams
+
+Sometimes you need to stop a streaming response before it completes: a user clicks "stop generating" in a chat UI, you've received enough data to make a decision, or you want to limit costs. Both [`run_stream()`][pydantic_ai.agent.AbstractAgent.run_stream] and [`run_stream_events()`][pydantic_ai.agent.AbstractAgent.run_stream_events] support explicit cancellation, which closes the underlying HTTP connection to stop token generation immediately.
```
DouweM: We should specify that `iter` supports it too

docs/output.md
```diff
@@ -875,6 +875,126 @@ async def main():
...
+
+#### Cancelling `run_stream`
```
DouweM, line 910: `run_stream_events` is the primary streaming API and should always be mentioned before `run_stream` (in the copy above and the sections here)

docs/output.md
```diff
@@ -875,6 +875,126 @@ async def main():
...
+3. The `cancelled` property reflects the cancellation state.
+4. The final [`ModelResponse`][pydantic_ai.messages.ModelResponse] is marked with `state='interrupted'` so that downstream code can identify incomplete responses.
+
+_(This example is complete, it can be run "as is" -- you'll need to add `asyncio.run(main())` to run `main`)_
+
+If you `break` out of `stream_text()` and then leave the surrounding `async with` block, the stream is cleaned up as the context exits. Use `cancel()` when you want to stop generation immediately instead of only stopping local consumption.
+
+#### Cancelling `run_stream_events`
+
+[`run_stream_events()`][pydantic_ai.agent.AbstractAgent.run_stream_events] returns a `StreamEventsResult` that should be used as an async context manager. Call `cancel()` on the stream to stop it:
```
DouweM: We should link to the new stream type!

docs/output.md
```diff
@@ -875,6 +875,126 @@ async def main():
...
+
+```python {title="stream_cancel_run_stream_events.py"}
+from pydantic_ai import Agent, FinalResultEvent, PartStartEvent
+
+agent = Agent('openai:gpt-5.2')
+
+
+async def main():
+    async with agent.run_stream_events('Write a long essay about Python') as stream:  # (1)!
```
DouweM: We should make sure there's also an ergonomic way to cancel from the `event_stream_handler` argument to `Agent` and `agent.run`, and the new  `wrap_run_event_stream` capability / `Hooks.on.run_event_stream`. If we're changing the `run_stream_events` API now, we should do it there as well -- before v2! That _can_ be a followup PR, but you'll want to start thinking about it now.

docs/output.md
```diff
@@ -875,6 +875,126 @@ async def main():
...
+1. Use `async with` to ensure proper cleanup of the background task and HTTP connection. Iterating `run_stream_events()` directly without `async with` is deprecated.
+2. `cancel()` closes the underlying generator, which triggers task cancellation and connection cleanup.
+
+_(This example is complete, it can be run "as is" -- you'll need to add `asyncio.run(main())` to run `main`)_
+
+Breaking out of the loop only stops local iteration; the stream is cleaned up when the surrounding `async with` exits. Call `cancel()` when you want to stop generation immediately instead of waiting for context exit.
+
+#### Cancelling with `iter`
+
+When using [`agent.iter()`][pydantic_ai.agent.Agent.iter] for fine-grained control over the agent graph, you can cancel at the `AgentStream` level inside a `ModelRequestNode.stream()` context:
```
DouweM: Wasn't mentioned at the top that it works with iter :)

DouweM (changes requested):

docs/output.md
adtyavrdhn: Yep, missed that.

pydantic_ai_slim/pydantic_ai/models/__init__.py
adtyavrdhn: Ugh fucking Claude

pydantic_ai_slim/pydantic_ai/result.py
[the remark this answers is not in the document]
adtyavrdhn: Hmm, I am skeptical this is true. I wrote a short test while implementing it and remember we trigger finally while breaking out.

If not we should, I'll verify.

pydantic_ai_slim/pydantic_ai/models/__init__.py
adtyavrdhn: Gotcha, that makes sense.

pydantic_ai_slim/pydantic_ai/models/__init__.py
adtyavrdhn: Yes it should be models, I think I keep calling it providers which swayed Claude to put this in the docstrings lol

pydantic_ai_slim/pydantic_ai/result.py
adtyavrdhn: Yeah that is better, I ended up naming this as is. Missed the convention behind it :'))

pydantic_ai_slim/pydantic_ai/_agent_graph.py
adtyavrdhn: Yep it is the same obj

pydantic_ai_slim/pydantic_ai/_agent_graph.py
adtyavrdhn: > The fact that we include the half-finished response does seem somewhat inconsistent with how half-streamed responses from a run interrupted by an exception do not show up in capture_run_messages, though there is a plan to change that: https://github.com/pydantic/pydantic-ai/issues/3219 (that could be nice clean up before v2).

Yep I would agree that the incomplete result should be present.

> It is also making me think that may be {new,all}_messages and capture_run_messages should have a include_incomplete flag or something so users can choose whether they want them or not. (In https://github.com/pydantic/pydantic-ai/pull/4306 we're adding another type of "incomplete" that users will likely want to skip when storing history, I'll bring it up with Matt in our channel).

Interesting, yeah it makes sense. I am very inclined to consider these points for evals and have opinions about it that we can share.

docs/output.md
adtyavrdhn: Yeah it should be a follow up or a stacked PR. I'll look into it.

pydantic_ai_slim/pydantic_ai/models/__init__.py
adtyavrdhn: For the flow? We wouldn't want users to be able to just set cancelled? 

Intention was to keep this read only and set it via cancel()

pydantic_ai_slim/pydantic_ai/result.py
```diff
@@ -795,6 +858,84 @@ def is_complete(self) -> bool:
...
+        # TODO(v2): remove standalone iteration support and require `async with`
+        if not self._managed:
+            warnings.warn(
+                'Iterating `AgentEventStream` directly with `async for event in stream:` is deprecated. '
+                'Use `async with agent.run_stream_events(...) as stream:` and '
+                '`async for event in stream:` instead '
+                'to ensure proper cleanup.',
+                DeprecationWarning,
+                stacklevel=2,
+            )
```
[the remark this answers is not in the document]
adtyavrdhn: As they should?

pydantic_ai_slim/pydantic_ai/agent/abstract.py
```diff
@@ -1092,8 +1108,18 @@ async def run_agent() -> AgentRunResult[Any]:
...
             raise
 
+        finally:
+            # We do not support trio but catching via anyio anyway
+            cancelled_exc = anyio.get_cancelled_exc_class()
+            if not task.done() and not _task_cancel_sent:
+                task.cancel()
+
+            with contextlib.suppress(cancelled_exc):
+                await task
```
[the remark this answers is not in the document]
adtyavrdhn: @DouweM What do you think about this? 

Error is not related to cancel but shows up when we await it here. I am fine suppressing it but it is interesting how the model doesn't want to catch broad `Exceptions` but wants to suppress them lol

pydantic_ai_slim/pydantic_ai/models/google.py
adtyavrdhn: I am not sure I understand? Which 2 variables?

pydantic_ai_slim/pydantic_ai/models/mistral.py
```diff
@@ -654,15 +655,27 @@ class MistralStreamedResponse(StreamedResponse):
...
+    async def cancel(self) -> None:
+        if self.cancelled:
+            return
+        # Set first so the flag is visible even if close() raises.
+        self._cancelled = True
+        # Close the underlying httpx response directly, avoiding the fragile
+        # __aexit__(None, None, None) pattern on EventStreamAsync.
+        # TODO: Check if MistralEventStreamAsync exposes a public close() method
+        # to avoid reaching into SDK internals.
+        await self._stream.response.aclose()
```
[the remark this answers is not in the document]
adtyavrdhn: Which is the only way so

pydantic_ai_slim/pydantic_ai/result.py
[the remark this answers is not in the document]
adtyavrdhn: This was indeed valid, async generators upon breaking add a GeneratorExit at the yield but iterators do not which would have caused the underlying generator to stay open. Devin ftw, now fixed :)

pydantic_ai_slim/pydantic_ai/models/__init__.py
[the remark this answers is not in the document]
adtyavrdhn: Yep it makes sense, I understand it should be a no op for `CompletedStreamedResponse` and I completely missed `OutlinesStreamedResponse`. I'll add both.

pydantic_ai_slim/pydantic_ai/models/google.py
adtyavrdhn: Okay I get it, you mean peekable response and normal AsyncStream

pydantic_ai_slim/pydantic_ai/models/__init__.py
adtyavrdhn: Okay, I am a bit lost here reading this. Is this the end goal for a leaner public API surface? I like that the error handling is owned by the superclass because there is nothing for the subclasses to do but just tell which errors to suppress and that's it.

I don't mind duplicating this effort but I am not sure if this is that big of a change to the API surface to resort to that?

pydantic_ai_slim/pydantic_ai/models/groq.py
adtyavrdhn: Hmm I didn't give it much thought at first because I thought it would be appropriate to store the underlying Stream without an iterator wrapper. 

I don't want to pass through it to `PeekableAsyncStream` and pretend all iterators are closable which is basically the contract there?

It will also not play with mistral which doesn't expose a close via the stream and we have to reach into the httpx response to close it. 

Codex suggested something nice which is to maybe store the stream cancel callable and call it when we close the stream and I am on board with the idea there. It is different enough that we are not actually storing the entire stream and its wrapper together?

adtyavrdhn:
Self note: https://github.com/pydantic/pydantic-ai/pull/4338

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -1899,8 +1909,52 @@ def _is_same_request(message: _messages.ModelMessage, request: _messages.ModelRe
...
+            isinstance(part, _messages.BaseToolCallPart)
+            and part.args_incomplete
+            and part.tool_call_id not in processed_tool_call_ids
+        )
+    ]
+    if not filtered_parts:
+        return None
+    if len(filtered_parts) != len(message.parts):
+        return replace(message, parts=filtered_parts)
+    return message
```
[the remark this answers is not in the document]
adtyavrdhn: We should revive args_incomplete -> that was incorrect

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -1899,8 +1909,56 @@ def _is_same_request(message: _messages.ModelMessage, request: _messages.ModelRe
...
+    entirely.
+    """
+    filtered_parts = [
+        part
+        for part in message.parts
+        if not (
+            isinstance(part, _messages.BaseToolCallPart)
+            and part.tool_call_id not in processed_tool_call_ids
+        )
+    ]
```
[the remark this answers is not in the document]
adtyavrdhn: This was also incorrect -> It doesn't matter anymore if we revive tool_args_incomplete

adtyavrdhn:
Check task.done(), do not suppress BaseException

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -1057,9 +1058,62 @@ def part_end_event(next_part: ModelResponsePart | None = None) -> PartEndEvent |
                 if end_event:
                     yield end_event
 
-            self._event_iterator = iterator_with_part_end(iterator_with_final_event(self._get_event_iterator()))
+            async def guarded() -> AsyncIterator[ModelResponseStreamEvent]:
```
adtyavrdhn: Take chain `iterator_with_part_end(iterator_with_final_event(self._get_event_iterator()))` and then just wrap again

adtyavrdhn:
Check #3219

pydantic_ai_slim/pydantic_ai/models/huggingface.py
```diff
@@ -482,12 +487,21 @@ class HuggingFaceStreamedResponse(StreamedResponse):
...
     _timestamp: datetime = field(default_factory=_utils.now_utc)
 
+    async def close_stream(self) -> None:
+        try:
+            # huggingface_hub types this as AsyncIterable, but at runtime it's an
+            # async generator that exposes aclose().
+            await self._response.source.aclose()  # pyright: ignore[reportAttributeAccessIssue, reportUnknownMemberType]
+        except RuntimeError as exc:
+            if not _utils.is_async_generator_already_running(exc):
+                raise
```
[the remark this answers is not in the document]
adtyavrdhn: We would need to fix it then unfortunately but iterable does not have an aclose.

pydantic_ai_slim/pydantic_ai/models/bedrock.py
```diff
@@ -1213,6 +1213,13 @@ class BedrockStreamedResponse(StreamedResponse):
     _timestamp: datetime = field(default_factory=_utils.now_utc)
     _provider_response_id: str | None = None
 
+    def get_stream_cancel_errors(self) -> tuple[type[BaseException], ...]:
+        return (BotoCoreError,)
+
+    async def close_stream(self) -> None:
+        # Bedrock's EventStream.close() is synchronous.
+        self._event_stream.close()
```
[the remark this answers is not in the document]
adtyavrdhn: Fixed to thread for asynchronous.

adtyavrdhn:
This ran into the semantics of cancellation of a run vs cancellation of a stream. I'll try for a more localized fix while we figure that bit out.

PR #5032, opened 2026-04-09
chore: follow-up tweaks to #4421

dsfaccini:
## Summary

Follow-up to #4421 based on post-merge review comments from @Kludex:

- Use `@deprecated` decorator on `cached_async_http_client` instead of manual `warnings.warn`, remove redundant "Deprecated" from docstring
- Move `BaseClient` import in bedrock model to re-use provider's `try/except` import guard for friendly install message

Closes #4421 (post-merge feedback)

## Test plan

- [x] Pre-commit hooks pass (format, lint, typecheck)
- [x] Existing `test_cached_async_http_client_deprecated` updated with pyright ignore for `reportDeprecated`

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/bedrock.py
+++ b/pydantic_ai_slim/pydantic_ai/models/bedrock.py
@@ -11,7 +11,6 @@ from typing import TYPE_CHECKING, Any, Generic, Literal, cast, overload
 from urllib.parse import parse_qs, urlparse
 
 import anyio.to_thread
-from botocore.client import BaseClient
 from botocore.exceptions import ClientError
 from typing_extensions import ParamSpec, assert_never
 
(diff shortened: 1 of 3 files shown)


pydantic_ai_slim/pydantic_ai/models/bedrock.py
```diff
@@ -11,7 +11,6 @@
 from urllib.parse import parse_qs, urlparse
 
 import anyio.to_thread
-from botocore.client import BaseClient
 from botocore.exceptions import ClientError
```
Kludex: The point was not to import from the providers module, was to have the try..except block. This still errors. and we shouldn't import BaseClient that is from botocore from the providers module.

tests/test_agent.py
```diff
@@ -6354,10 +6354,10 @@ async def test_agent_context_manager_no_model():
 
 
 def test_cached_async_http_client_deprecated():
-    from pydantic_ai.models import cached_async_http_client
+    from pydantic_ai.models import cached_async_http_client  # pyright: ignore[reportDeprecated]
 
-    with pytest.warns(DeprecationWarning, match='cached_async_http_client is deprecated'):
-        cached_async_http_client()
+    with pytest.warns(DeprecationWarning, match='cached_async_http_client.*is deprecated'):
```
Kludex, line 6359: I think you can use snapshot() on the match.

pydantic_ai_slim/pydantic_ai/models/bedrock.py
dsfaccini: fair enough, yeah my point was that importing from `models.bedrock` triggers the import guard in `providers.bedrock` so repeating it here doesn't add extra safety but I can add it any way

tests/test_agent.py
dsfaccini: I tried it now but snapshot doesn't work in `pytest.warns(match=snapshot())`

dsfaccini: `snapshot() passes an UndecidedValue to re.search which
  can't be hashed.`

PR #5035, opened 2026-04-09
Reorder TST to wrap before capability wrappers

dsfaccini:
## Summary
- Reorder `ToolSearchToolset` wrapping in `Agent._get_toolset` to occur before capability wrapper injection
- Capabilities are now outermost so they can change anything (per DouweM's review on #4960)
- Needed for upcoming TST + Code Mode integration

Closes: N/A (extracted from #4960 plan)

## Test plan
- [ ] Existing `tests/test_tool_search.py` passes
- [ ] `make lint` passes


DouweM:
@David SF That won't quite work, because ToolSearch should run after capabilities insert their own tool(set)s. It's just CodeMode that then needs to run after ToolSearch. So we do need something more sophisticated where ToolSearchToolset is also injected via a capability, which then specifies that it has to run after all user toolsets, but before CodeMode (in a generic enough way that users' custom capabilities could also indicate where they should be)

PR #5041, opened 2026-04-10
Add Restate integration page

gvdongen:
This PR adds a page on Restate Durable Execution.

It explains:
- architecture
- minimal example
- features beyond durable execution: durable sessions, resilient human-in-the-loop, durable RPC, agent control
- Logfire integration


docs/durable_execution/overview.md
```diff
@@ -2,10 +2,11 @@
 
 Pydantic AI allows you to build durable agents that can preserve their progress across transient API failures and application errors or restarts, and handle long-running, asynchronous, and human-in-the-loop workflows with production-grade reliability. Durable agents have full support for [streaming](../agent.md#streaming-all-events) and [MCP](../mcp/client.md), with the added benefit of fault tolerance.
 
-Pydantic AI natively supports three durable execution solutions:
+Pydantic AI natively supports four durable execution solutions:
```
adtyavrdhn: But this is an external package so we do not support this natively?

adtyavrdhn (changes requested):
Hey @gvdongen 

Thanks for taking the time out to add this.

If we were to add it here we should follow the pattern set forth by other integrations. 

This would however not fit in here because this is not a native support but an external package so would be more suited for Third Party Integrations maybe?

docs/durable_execution/overview.md
DouweM: How about:

```
Pydantic AI officially supports four durable execution solutions:

[the list]

These integrations are co-maintained by the Pydantic and vendor teams and only use [the rest]
```

docs/durable_execution/restate.md
```diff
@@ -0,0 +1,334 @@
+# Durable Execution with Restate
+
+[Restate](https://restate.dev) is a lightweight durable execution runtime with first-class support for AI agents. The Pydantic AI integration is provided via the [Restate Python SDK](https://github.com/restatedev/sdk-python/tree/main).
```
DouweM: Please make this link to the ext/pydantic subdir

docs/durable_execution/restate.md
```diff
@@ -0,0 +1,334 @@
...
+See the [Restate documentation](https://docs.restate.dev/ai/patterns/durable-agents) for more information.
+
+## Durable Agent
+
+Any Pydantic AI agent can be made durable by wrapping it with `RestateAgent` from the Restate SDK and running it inside a Restate service handler.
+
+Install the Restate SDK:
+
+```bash
+pip install pydantic-ai "restate_sdk[serde]" hypercorn
```
DouweM: - We can use `uv/pip` install which will automatically render with tabs
- Is Hypercorn required for the integration, or just the example?

docs/durable_execution/restate.md
```diff
@@ -0,0 +1,334 @@
...
+docker run --name restate_dev --rm \
+-p 8080:8080 -p 9070:9070 -p 9071:9071 \
+--add-host=host.docker.internal:host-gateway \
+docker.restate.dev/restatedev/restate:latest
+
+# Terminal 2: Start your agent service
+python restate_agent.py
+```
+
+Go to the Restate UI at `http://localhost:9070` and register the agent service at `http://host.docker.internal:9080`.
```
DouweM: Including an example is great, but with the rest of the docs I worry about it getting outdated as you update the behavior/API/UI on the Restate SDK side. What do you think about ending after the "Durable Execution" and "Durable Agent" sections and linking (more prominently at the top) to your own docs?

DouweM (changes requested):

DouweM:
@gvdongen Thanks Giselle! Also please have a look at the failing CI steps.

docs/durable_execution/restate.md
gvdongen: @DouweM Hypercorn or any other ASGI server is necessary to run the application. I removed it since it's not required for the integration itself.

PR #3951, opened 2026-01-08
Add OpenAI Responses input token counting

colesmcintosh:
- Closes #3430

## Summary
- Add `/responses/input_tokens` support for `OpenAIResponsesModel.count_tokens` to enable `UsageLimits.count_tokens_before_request` without `tiktoken`.
- Allow instruction-only token counting by inserting an empty user message when needed; still raises `UserError` when there are no messages, no instructions, and no previous response ID.
- Pass request `timeout` through to the `/responses/input_tokens` call.

### Pre-Review Checklist
- [x] Any **AI generated code** has been reviewed line-by-line by the human PR author, who stands by it.
- [x] No **breaking changes** in accordance with the [version policy](https://github.com/pydantic/pydantic-ai/blob/main/docs/version-policy.md).
- [x] **Linting and type checking** pass per `make format` and `make typecheck`.
- [x] **PR title** is fit for the [release changelog](https://github.com/pydantic/pydantic-ai/releases).

### Pre-Merge Checklist
- [x] New **tests** for any fix or new behavior, maintaining 100% coverage.
- [x] Updated **documentation** for new features and behaviors, including docstrings for API docs.

PR #5043, opened 2026-04-10
feat: add count_tokens for OpenAIResponsesModel

### LINKED ISSUE #3430: Add OpenAI token counting
### Description

- Docs: 
	- https://cookbook.openai.com/examples/how_to_count_tokens_with_tiktoken#6-counting-tokens-for-chat-completions-api-calls
	- https://github.com/openai/tiktoken
- `count_tokens` is currently supported on `GoogleModel` and `BedrockModel`
- works with `UsageLimits.count_tokens_before_request`

### References

_No response_

Ricardo-M-L:
## Summary

- Implements `count_tokens` for `OpenAIResponsesModel` using the [Responses API `input_tokens` endpoint](https://platform.openai.com/docs/api-reference/responses/input-tokens)
- Returns accurate token counts without running inference, at no cost
- Enables `UsageLimits(count_tokens_before_request=True)` for OpenAI Responses models
- Mirrors the same parameter preparation logic as `_responses_create` (tools, instructions, reasoning, truncation, etc.)

Fixes #3430

## Implementation

The `count_tokens` method on `OpenAIResponsesModel` calls `client.responses.input_tokens.count(...)` with the same message/tool/instruction preparation as `_responses_create`. This follows the approach suggested by @DouweM in the [issue discussion](https://github.com/pydantic/pydantic-ai/issues/3430#issuecomment-2866651085):

> The Responses API has an endpoint for this: https://platform.openai.com/docs/api-reference/responses/input-tokens
> I'd love to see a much smaller PR that adds support using that API endpoint rather than tiktoken.

**Files changed:**
- `pydantic_ai_slim/pydantic_ai/models/openai.py` — Added `count_tokens` method to `OpenAIResponsesModel`
- `tests/models/mock_openai.py` — Extended `MockOpenAIResponses` to support `input_tokens.count` mocking
- `tests/models/test_openai_responses.py` — Added 15 test cases covering basic usage, limits, tools, settings, edge cases

## Test plan

- [x] 15 new unit tests all passing
- [x] All 109 existing `test_openai_responses.py` tests still pass
- [ ] Verify with a real OpenAI API key that `input_tokens.count` returns expected results

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/openai.py
+++ b/pydantic_ai_slim/pydantic_ai/models/openai.py
@@ -1874,6 +1874,92 @@ class OpenAIResponsesModel(Model[AsyncOpenAI]):
             response, cast(OpenAIResponsesModelSettings, model_settings or {}), model_request_parameters
         )
 
+    async def count_tokens(
+        self,
+        messages: list[ModelMessage],
+        model_settings: ModelSettings | None,
+        model_request_parameters: ModelRequestParameters,
+    ) -> usage.RequestUsage:
+        """Count input tokens using the OpenAI Responses API `input_tokens` endpoint.
+
+        This uses the `POST /responses/input_tokens` endpoint which returns an accurate
+        token count without running inference. See:
+        https://platform.openai.com/docs/api-reference/responses/input-tokens
+        """
+        check_allow_model_requests()
+        model_settings, model_request_parameters = self.prepare_request(
+            model_settings,
+            model_request_parameters,
+        )
+        model_settings = cast(OpenAIResponsesModelSettings, model_settings or {})
+
+        tools = (
+            self._get_builtin_tools(model_request_parameters)
+            + list(model_settings.get('openai_builtin_tools', []))
+            + self._get_tools(model_request_parameters)
+        )
+        profile = OpenAIModelProfile.from_profile(self.profile)
+        if not tools:
+            tool_choice: Literal['none', 'required', 'auto'] | None = None
+        elif not model_request_parameters.allow_text_output and profile.openai_supports_tool_choice_required:
+            tool_choice = 'required'
+        else:
+            tool_choice = 'auto'
+
+        previous_response_id = model_settings.get('openai_previous_response_id')
+        if previous_response_id == 'auto':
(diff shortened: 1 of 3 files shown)


VANDRANKI (commented):
Solid implementation. A few observations:

**Parameter preparation parity** - the `count_tokens` method replicates the same `prepare_request` + tool assembly + `_map_messages` + `_translate_thinking` logic as `request`. This is necessary to get an accurate count (the token count depends on the full formatted prompt including tools and instructions), but it also means any future changes to `request` preparation need to be mirrored here. A comment noting this coupling would help maintainers.

**`openai_previous_response_id = auto` path** - handled correctly. The previous response id is extracted and the messages list is trimmed before counting, matching how `request` would behave. Good.

**Return type** - returns `RequestUsage` with only `request_tokens` populated. The `response_tokens` field will be `None` since no inference ran. This is consistent with how `count_tokens` works on other models.

**Test coverage** - 464 lines of new tests is thorough. The mock `POST /v1/responses/input_tokens` path is the right place to test this.

One question: if the model does not support the `/responses/input_tokens` endpoint (e.g. an older deployment or a non-OpenAI compatible provider using `OpenAIResponsesModel`), what does the error look like? Is there a clear exception, or does it surface as a generic HTTP 404?

Ricardo-M-L:
Hi @Kludex @samuelcolvin, friendly ping — just rebased onto current `main` to clear the conflict (test file had additive conflicts only — both sides appended new tests at the bottom; kept both). The PR is now `MERGEABLE` again. It adds `count_tokens` support for `OpenAIResponsesModel`, matching the existing `OpenAIChatModel` implementation so `count_tokens_before_request=True` works end-to-end on the Responses API. Happy to address any feedback 🙏

Ricardo-M-L:
Closing — looks like #3951 ("Add OpenAI Responses input token counting") landed an independent implementation of the same feature 🎉

Quick verification (current `main`):

```python
$ grep -n "async def count_tokens" pydantic_ai_slim/pydantic_ai/models/openai.py
# OpenAIResponsesModel.count_tokens is at line 1905
```

The upstream version uses a refactored `_build_responses_request_params` helper which is cleaner than my inline construction. My PR's contribution is no longer needed — closing as superseded.

PR #5056, opened 2026-04-12
feat: add RepairToolOrphans capability for fixing broken tool pairing

### LINKED ISSUE #4728: Built-in HistoryProcessor for orphaned tool call/result repair
## The problem

## What we built

```python
```

## The ask

```python

```

## How we got here

### References

_No response_

Ricardo-M-L:
## Summary

Implements the `RepairToolOrphans` **capability** (as requested by @DouweM in #4728) to repair broken tool-call/return pairing in persisted message history.

Previous PR #5021 was closed because it was implemented as a `history_processor` instead of a capability. This PR follows the guidance correctly.

## What it does

- **Dangling tool calls**: `ToolCallPart` with no matching return → injects synthetic `ToolReturnPart` in the next request
- **Dangling builtin calls**: `BuiltinToolCallPart` without return → injects `BuiltinToolReturnPart` in the **same** `ModelResponse` (required by Anthropic)
- **Orphaned returns**: `ToolReturnPart`/`RetryPromptPart` with no matching call → removed
- **Empty requests**: If cleanup empties a `ModelRequest`, inserts placeholder `UserPromptPart`
- **Warnings**: Emits `UserWarning` when repairs are made (configurable via `warn=True/False`)

## Usage

```python
from pydantic_ai import Agent
from pydantic_ai.capabilities import RepairToolOrphans

agent = Agent('anthropic:claude-sonnet', capabilities=[RepairToolOrphans()])
```

## Files changed

- `pydantic_ai_slim/pydantic_ai/capabilities/repair_tool_orphans.py` — capability implementation (250 lines)
- `pydantic_ai_slim/pydantic_ai/capabilities/__init__.py` — register in exports and `CAPABILITY_TYPES`
- `tests/test_repair_tool_orphans.py` — 16 tests (unit + integration)
- `tests/test_capabilities.py` — snapshot update for registry

## Test plan

- [x] 15 unit tests covering: empty/clean history, dangling regular/builtin calls, trailing responses, orphaned returns/retries, empty request placeholders, partial answers, metadata preservation, warning behavior, multi-turn poisoned conversations
- [x] 1 integration test with `Agent` + `FunctionModel` using poisoned message history
- [x] All 424 existing capability tests pass (2 pre-existing failures unrelated to this change)

Closes #4728

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/capabilities/__init__.py
+++ b/pydantic_ai_slim/pydantic_ai/capabilities/__init__.py
@@ -25,6 +25,7 @@ from .include_return_schemas import IncludeToolReturnSchemas
 from .mcp import MCP
 from .prefix_tools import PrefixTools
 from .prepare_tools import PrepareTools
+from .repair_tool_orphans import RepairToolOrphans
 from .set_tool_metadata import SetToolMetadata
 from .thinking import Thinking
 from .thread_executor import ThreadExecutor
@@ -43,6 +44,7 @@ CAPABILITY_TYPES: dict[str, type[AbstractCapability[Any]]] = {
         MCP,
         PrefixTools,
         PrepareTools,
+        RepairToolOrphans,
         SetToolMetadata,
         Thinking,
         Toolset,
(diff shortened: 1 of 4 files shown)


adtyavrdhn:
Hey @Ricardo-M-L 

Thanks for taking the time to raise this.

This is something @sarth6 is working on I think, I'll close this PR pending discussion on the issue of if this belongs in the core library now.

PR #5059, opened 2026-04-13
Fix streamed text or structured output selection

### LINKED ISSUE #1590: Structured Output fails with text output + Behaviour inconsistency
### Initial Checks

### Description

Situation:
* I need LLM text output, to show it to the user, where the LLM reasons and explains.
* I want a structured output.

This is NOT working in streaming mode, but it does in NON-streaming mode.

See below tests. The first fails. The second passes.

### Example Code

```Python
from typing import Union

from pydantic_ai import Agent


class CallAgent(BaseModel):
    agent_name: str


agent = Agent(
    model="google_gla:gemini-2.5-flash-preview-04-17",
    output_type=Union[str, CallAgent],
    instructions="Say hello and then transfer the user to 'user_assistant' agent",
)


async def test_output_with_str(): # FAILS
    async with agent.run_stream(user_prompt="Hello") as result:
        async for msg, is_last in result.stream_structured():
            print(msg)
    assert result.get_output() == CallAgent(agent_name="user_assistant")


async def test_output_with_str_no_stream(): # PASSES
    result = await agent.run(user_prompt="Hello")
    assert result.output == CallAgent(agent_name="user_assistant")
```

```Text
pydanticai 0.1.4
any LLM
Python 3.12
```

vonbai:
## Summary

- Fix streaming output selection when an agent allows `str | structured` output and the stream emits text before the structured final-result tool call.
- Continue falling back to text output when no structured/deferred output tool result is produced.
- Add regression coverage for both text-then-structured and text-only streaming cases.

Fixes #1590.

## Validation

- `uv run pytest tests/test_streaming.py::TestPartialOutput::test_text_then_structured_output_prefers_structured tests/test_streaming.py::TestPartialOutput::test_text_with_structured_output_falls_back_to_text`
- `uv run ruff check pydantic_ai_slim/pydantic_ai/models/__init__.py tests/test_streaming.py`
- `uv run pyright pydantic_ai_slim/pydantic_ai/models/__init__.py tests/test_streaming.py`

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/__init__.py
+++ b/pydantic_ai_slim/pydantic_ai/models/__init__.py
@@ -1006,24 +1006,6 @@ class StreamedResponse(ABC):
         first match is found.
         """
         if self._event_iterator is None:
-
-            async def iterator_with_final_event(
-                iterator: AsyncIterator[ModelResponseStreamEvent],
-            ) -> AsyncIterator[ModelResponseStreamEvent]:
-                async for event in iterator:
-                    yield event
-                    if (
-                        final_result_event := _get_final_result_event(event, self.model_request_parameters)
-                    ) is not None:
-                        self.final_result_event = final_result_event
-                        yield final_result_event
-                        break
-
-                # If we broke out of the above loop, we need to yield the rest of the events
-                # If we didn't, this will just be a no-op
-                async for event in iterator:
-                    yield event
-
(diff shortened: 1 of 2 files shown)


DouweM:
This basically breaks text streaming, as we don't start streaming until we know for sure no tool will be called. Please discuss on the issue first.

PR #5076, opened 2026-04-14
docs: soft-deprecate `history_processors` in favor of `before_model_request` hooks

dsfaccini:
## What this changes

Docs-only. No behavior change, no runtime `DeprecationWarning`. `history_processors` remains fully supported.

- **`docs/message-history.md`** — adds a `!!! warning "Prefer before_model_request hooks"` admonition on the `## Processing Message History` section, followed by an equivalent `compaction_hook.py` example that shows the real motivating use case: accessing `request_context.model_request_parameters.function_tools` to size the message budget based on how many tool schemas are already consuming the context window.
- **`docs/hooks.md`** — one-sentence cross-reference from the `### Model request hooks` section back to message-history.md, explaining that `before_model_request` is the recommended place for token-aware history compaction.
- **`pydantic_ai_slim/pydantic_ai/agent/__init__.py`** — adds a `!!! warning "Deprecated: prefer before_model_request hooks"` admonition to the `history_processors` constructor param docstring, and a `# TODO(v2): remove history_processors ...` comment on the field assignment.
- **`pydantic_ai_slim/pydantic_ai/capabilities/history_processor.py`** — adds a matching deprecation admonition to the `HistoryProcessor` class docstring. (This capability is already implemented internally as a thin `before_model_request` wrapper, so the deprecation is a pure API-surface consolidation.)
- **`V2.md`** (new, repo root, **not under `docs/`**) — introduces a \"v2 roadmap\" file tracking non-urgent retirements and their migration guides. Placed at repo root so it's visible in the GitHub UI and referenceable from code/docs, but intentionally unpublished on ai.pydantic.dev. First entry is `history_processors → before_model_request` with motivation, before/after migration code, and edge-case notes.
- **`AGENTS.md`** — one new bullet under \"Requirements of all contributions\" steering contributors (human and agent) away from entrenching patterns listed in `V2.md`.

## Why no runtime `DeprecationWarning`?

Kept out of scope deliberately. This PR is about surfacing the preferred API in docs and collecting direction in `V2.md` so contributors stop entrenching the legacy pattern. An actual runtime warning + removal cycle should be a separate, more intentional change since it touches every user who's currently using `history_processors`.

## Decisions worth flagging to reviewers

1. **Wording is intentionally hedged.** \"May be removed in a future major version\" rather than \"planned for removal\" — matches `V2.md`'s own caveat that \"items here are not commitments — they're direction.\"
2. **`V2.md` placement and link style.** Cross-references from published docs/docstrings use absolute GitHub URLs (\`https://github.com/pydantic/pydantic-ai/blob/main/V2.md#history_processors--before_model_request\`) so mkdocs strict-mode doesn't flag them and published readers still get a clickable migration path. Open to moving `V2.md` under `docs/` and using normal mkdocs cross-refs if you'd prefer to publish it — happy to follow up.
3. **Section structure in `message-history.md`.** The existing `## Processing Message History` section is kept intact (warning added on top) rather than being rewritten to lead with hooks. Strict reading of \`docs/AGENTS.md\` (\"show the recommended approach first\" / \"omit deprecated features from user-facing docs\") could argue for demoting the legacy examples below the fold. Didn't do that here because it's a bigger change and would break 5 inbound anchor references from capabilities.md, input.md, multi-agent-applications.md, thinking.md, and tools-advanced.md. Happy to restructure if you prefer.
4. **`ModelRequestContext` cross-references.** Uses the same \`[ModelRequestContext][pydantic_ai.models.ModelRequestContext]\` form as the existing (mkdocstrings-unresolvable) refs in `hooks.md` and `capabilities.md`. These don't resolve in strict mode but CI uses `--no-strict`, so this is baseline behavior — consistent with existing patterns rather than introducing a new one.

### Checklist

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -61,6 +61,7 @@ All changes need to:
 
 - be thoughtful and deliberate about new abstractions, public APIs, and behaviors, as every wrong-in-retrospect choice (made in a rush or with insufficient context) makes life harder for hundreds of thousands of users (and agents), and is much more difficult to change later than to do right the first time
 - be backward compatible as laid out in the [version policy](docs/version-policy.md), so that users can upgrade with confidence
+- not entrench patterns slated for retirement in the [v2 roadmap](V2.md) — that file tracks API simplifications planned for the next major release and their preferred alternatives; new code should use the alternatives, and reviewers may reject code that adds new usage of deprecated patterns listed there
 - be fully type-safe (both internally and in public API) without unnecessary `cast`s or `Any`s, so that users don't need `isinstance` checks and can trust that code that typechecks will work at runtime
 - have comprehensive tests covering 100% of code paths, favoring integration tests and real requests (using recordings and snapshots -- see below) over unit tests and mocking
 - update/add all relevant documentation, following the existing voice and patterns
--- a/V2.md
+++ b/V2.md
@@ -1,0 +1,98 @@
+# Pydantic AI v2 roadmap
+
+> **Why this file exists**
+>
+> This file tracks non-urgent cleanups and API simplifications we'd like to land in a future v2 release, along with migration guidance for users. It exists so that:
+>
+> 1. Contributors (human and agent) can recommend modern patterns with confidence, knowing which v1 patterns are slated for removal.
+> 2. Users who hit pain points with deprecated APIs can see the long-term plan and migrate preemptively.
+> 3. Reviewers can reject new code that entrenches patterns we're trying to retire.
+>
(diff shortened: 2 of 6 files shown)


Wh1isper:
I think we can make `history_processors` as an alias for `before_model_request`. To me, `history_processors` is very concise and makes it easy to chain different filters and test stateless history_processors is easier.

dsfaccini:
hey @Wh1isper , I appreciate your input! that is an option too

dsfaccini:
superseded by #5263

PR #5089, opened 2026-04-15
fix(evals): constrain judge_output reason to be concise and retry-safe

### LINKED ISSUE #5034: `judge_output() reason field can be polluted by visible reasoning/thinking text`
### Initial Checks

### Description

`pydantic_evals.evaluators.llm_as_a_judge.judge_output()` currently returns a freeform `reason: str`. In practice, with reasoning-capable models, that `reason` can become very long, include visible self-correction / re-checking, and even contradict itself mid-response.

This is especially problematic when `judge_output()` is used inside an agent `output_validator` and the `reason` is fed back into `ModelRetry(...)`, because the retry message becomes noisy and unstable instead of being a concise repair instruction.

Example pattern:

```python
@agent.output_validator
async def llm_judge_validator(ctx: RunContext[AgentDeps], data: _ModelT) -> _ModelT:
    grading = await judge_output(
        output=data,
        rubric=response_model.rubric(),
        model=self.model_obj,
        model_settings={"temperature": 0.0},
    )
    if not grading.pass_:
        raise ModelRetry(grading.reason)
    return data
```

In my case, the judge `reason` did things like:

- state a violation
- re-check the same rule
- partially retract the earlier claim
- continue with more analysis
- still return `pass=false`

So the issue is not just verbosity. The `reason` can become self-contradictory and unsuitable for retry feedback loops.

I realize some of this is model behavior, but the current built-in prompt/output shape seems to allow the model to "think out loud" inside `reason` rather than returning a short, stable failure summary.

One workaround is to disable thinking in `judge_model_settings`, but that is not what I want. I want the judge model to retain the benefits of internal reasoning; I just do not want reasoning/thinking output to leak into the public `reason` field. Ideally, the judge should be able to think internally while still returning a concise, stable, retry-safe explanation.

```python
import asyncio

from pydantic_evals.evaluators.llm_as_a_judge import judge_output

OUTPUT = {
    "title": "The Blue Light of Silence",
    "pages": [
        {"phase": "setup"},
        {"phase": "escalation"},
        {"phase": "climax"},
    ],
}

RUBRIC = """
Judge whether this structured output is acceptable.
Check each rule independently. Fail the output if ANY single rule is violated.

Episode:
- Fail if pages do not form a complete dramatic arc with forward-progressing phases
  (e.g. setup -> escalation -> turning point -> climax -> aftermath);
  repeating or skipping phases is not acceptable.
"""

async def main():
    result = await judge_output(
        output=OUTPUT,
        rubric=RUBRIC,
        model="gemma4:31b-cloud",
        model_settings={"temperature": 0.0},
    )
    print(result.model_dump())

asyncio.run(main())
```

Observed behavior: the returned `reason` can be a long, self-reflective paragraph rather than a concise explanation of the failed rule(s).

Expected behavior: `reason` should be short, stable, and directly usable as validation feedback, especially in retry loops.

### Logfire Trace

_No response_

- Python: 3.13.12
- PydanticAI: 1.73.0
- Pydantic: 2.12.5

anmolg1997:
## Summary

- Updates all four judge agent system prompts to explicitly instruct the model to keep the `reason` field to a concise 1-2 sentence summary
- Adds a `description` to `GradingOutput.reason` that reinforces this constraint via the JSON schema
- Prevents reasoning/thinking text from leaking into the public `reason` field, which is especially problematic when used in `ModelRetry` feedback loops

## Problem

When using reasoning-capable models (e.g. `gemma4:31b-cloud`, `o3-mini`, etc.) with `judge_output()`, the returned `reason` field can become a long, self-reflective paragraph that includes:
- Visible self-correction / re-checking
- Partial retractions of earlier claims
- Extended analysis that contradicts itself

This makes `reason` unsuitable for use in `ModelRetry(grading.reason)` patterns, where a concise, stable failure summary is needed for effective retry guidance.

## Approach

Rather than disabling thinking entirely (which would lose the reasoning quality benefits), the fix constrains the *output format* while preserving internal reasoning:

1. **System prompt instruction**: A shared `_JUDGE_REASON_INSTRUCTION` constant is added to all four judge agent prompts, instructing the model to state only the final conclusion in `reason`
2. **Schema-level hint**: `GradingOutput.reason` gains a `Field(description=...)` that tells the model to produce "A concise 1-2 sentence explanation" -- this surfaces in the JSON schema that models with native structured output use for generation

## Changes

| File | Change |
|------|--------|
| `pydantic_evals/pydantic_evals/evaluators/llm_as_a_judge.py` | Add `_JUDGE_REASON_INSTRUCTION`, update 4 agent system prompts, add `Field(description=...)` to `GradingOutput.reason` |

## Test plan

- [x] All 13 existing `test_llm_as_a_judge.py` tests pass
- [x] No other files modified -- change is isolated to prompt/schema constraints

Fixes #5034


lawcontinue:
The prompt-level constraint approach here is solid as a first layer of defense. A couple of thoughts on making this more robust:

**Post-processing safety net**

Prompt compliance isn't guaranteed, especially across different providers. Adding a `_sanitize_reason()` step after generation would catch cases where the model ignores the instruction:

```python
import re

def _sanitize_reason(reason: str, max_length: int = 500) -> str:
    """Strip reasoning/thinking artifacts from judge reason field."""
    # Remove common reasoning markers
    reason = re.sub(r'<think[^>]*>.*?</think[^>]*>', '', reason, flags=re.DOTALL)
    reason = re.sub(r'^(?:Wait,|Actually,|Let me reconsider|Hmm,|So,)\s*', '', reason, count=1)
    # Collapse repeated whitespace
    reason = re.sub(r'\s+', ' ', reason).strip()
    # Truncate if still too long
    if len(reason) > max_length:
        reason = reason[:max_length].rsplit('.', 1)[0] + '.'
    return reason
```

This would be applied to `GradingOutput.reason` before it reaches `ModelRetry`, so even if the prompt leaks through, the feedback stays clean.

**Truncation as a fallback**

The `max_length` parameter could be configurable via `ModelSettings` or `EvaluationsConfig`, similar to how other guardrails handle output constraints.

Neither of these replace the prompt approach — they complement it. The prompt handles the common case; post-processing handles the edge cases where model behavior diverges.

pydantic_evals/pydantic_evals/evaluators/llm_as_a_judge.py
```diff
@@ -115,22 +128,24 @@ async def judge_input_output(
...
+        {_JUDGE_REASON_INSTRUCTION}
 
         Examples:
 
         <Input>What color is the sky?</Input>
         <ExpectedOutput>Blue</ExpectedOutput>
         <Output>Cerulean</Output>
         <Rubric>The output is consistent with the expected output but doesn't have to match exactly</Rubric>
-        {"reason": "'Cerulean' is a shade of blue", "pass": true, "score": 1.0}
+        {{"reason": "'Cerulean' is a shade of blue", "pass": true, "score": 1.0}}
```
adtyavrdhn, line 142: Why is this needed?

adtyavrdhn: got it nvm

anmolg1997:
yeah that's why - prompt is now an f-string to interpolate _JUDGE_REASON_INSTRUCTION, so the literal JSON braces in examples need {{ }} or python eats them.

on the test you pushed: devin flagged it for hitting private internals per AGENTS.md. happy to drop it and rely on the schema description assertion in test_grading_output (that's already public surface), or swap it for a behavior test that mocks the agent. let me know which you want, or leave it.

adtyavrdhn:
Are you Claude? @anmolg1997

anmolg1997:
If that’s the bar now, I’m happy to be compared 😄 @adtyavrdhn

adtyavrdhn:
@anmolg1997 No the replies are clearly pasted from Claude / AI, if you're a human doing it I would appreciate you typing it instead.

anmolg1997:
I will mind your preferences onwards @adtyavrdhn

pydantic_evals/pydantic_evals/evaluators/llm_as_a_judge.py
```diff
@@ -27,26 +27,37 @@
 class GradingOutput(BaseModel, populate_by_name=True):
     """The output of a grading operation."""
 
-    reason: str
+    reason: str = Field(
+        description='A concise 1-2 sentence explanation of why the output passed or failed.',
```
adtyavrdhn: We should not point out passed / failed and instead of explanation we should use justification.

anmolg1997: done in 47b2cc9 - switched to 'justification for the verdict', dropped the pass/fail framing, applied same wording to _JUDGE_REASON_INSTRUCTION for consistency. also dropped the private-symbol test devin flagged; the schema description assertion in test_grading_output covers the public contract.

PR #3535, opened 2025-11-24
Support logprobs output from Responses API

tim-becker:
The plumbing for `logprobs` was previously only implemented OpenAI's completions API. The responses API also supports returning `logprobs`, with a few minor differences:
1. whether to include `logprobs` becomes an `include` flag
2. `logprobs` output will be attached to each returned `ResponseOutputText`

This PR uses the existing flags in `OpenAIResponsesModelSettings` (inherited from `OpenAIChatModelSettings`) for logprobs and implements the logic in the request and response processing. The `logprobs` output are converted to the same format used by the completion `logprobs` support.

PR #5091, opened 2026-04-15
Fixing parsing `openai_logprobs` for streamed responses

iAmir97:
I’ve noticed that logprob parsing is currently implemented in` _process_response` but is missing from `_process_streamed_response` in #3535 .

Please let me know if you’d like me to make any adjustments, Thanks.


iAmir97:
any readme file or a guide on how to add a cassette? whats the proper SOP in this case? Thanks!

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -3264,15 +3265,17 @@ def update_provider_details(existing: dict[str, Any] | None) -> dict[str, Any]:
...
     logprobs: list[chat_completion_token_logprob.ChatCompletionTokenLogprob]
-    | list[responses.response_output_text.Logprob],
+    | list[responses.response_output_text.Logprob]
+    | list[responses.response_text_done_event.Logprob],
 ) -> list[dict[str, Any]]:
     return [
         {
             'token': lp.token,
-            'bytes': lp.bytes,
+            'bytes': getattr(lp, 'bytes', None),
```
dsfaccini: we avoid using `getattr` bc it's not type safe, can you try to narrow down the type properly?

dsfaccini:
hey @iAmir97 yeah so you need to run the specific tests that test logprobs with `uv run pytest tests/models/test_openai.py -k "pattern for the tests you want to run"` (or responses). with the `--record-mode=rewrite` flag set, this will **bypass** the cassette and send the request directly to openai, rewriting the cassette with the new interaction

if you're using a coding agent the agent is good at finding the specific test names and running them, you only need an openai key in your environment

iAmir97:
Noted, thanks for the help @dsfaccini! I will address your comment and add the cassette soon.

tests/models/test_openai.py
```diff
@@ -2968,6 +2971,44 @@ async def test_openai_instructions_with_responses_logprobs(allow_model_requests:
...
+    part_end_events = [
+        event
+        for event in events
+        if isinstance(event, PartEndEvent)
+        and isinstance(event.part, TextPart)
+        and event.part.provider_details
+        and 'logprobs' in event.part.provider_details
+    ]
+    assert len(part_end_events) == 1
+    assert cast(Any, part_end_events[0].part).provider_details['logprobs'] == first_logprob
```
dsfaccini, line 3009: these assignments etc look a bit out of convention but I don't feel strongly enough about it

PR #5103, opened 2026-04-15
fix: restore system prompts when resuming UI runs

yu2001-s:
## Summary
- restore system prompts when a UI run resumes from frontend-provided user messages
- ensure the resumed request prepends the computed system prompt before the user prompt
- add regression coverage for the generic agent path, AG UI adapter, and Vercel AI adapter

## Why
When a run resumed from a frontend-only user message, the next request could omit the system prompt entirely. That changes agent behavior and makes resumed runs inconsistent with fresh runs.

## Validation
- `uv run pytest tests/test_ag_ui.py -k frontend_only_user_message tests/test_vercel_ai.py -k frontend_only_user_message`
- `uv run pytest tests/test_agent.py -k ending_on_user_prompt_includes_system_prompt`


yu2001-s:
Closing this duplicate in favor of the existing upstream work for the same fix.

PR #5107, opened 2026-04-15
Add `pydantic_ai.managed.logfire.Managed` capability

DouweM:
### Summary

Adds a new `pydantic_ai.managed.logfire.Managed` capability that wires [Logfire managed variables](https://logfire.pydantic.dev/docs/reference/advanced/managed-variables/) to agent fields, so prompts and model settings can be controlled from the Logfire UI without redeploying.

```python
import logfire

from pydantic_ai import Agent
from pydantic_ai.managed.logfire import Managed

logfire.configure()

prompt = logfire.var(
    'customer_support_prompt',
    default='You are a helpful customer-support assistant.',
)
settings = logfire.var(
    'customer_support_settings',
    type=dict,
    default={'temperature': 0.2},
)

agent = Agent(
    'openai:gpt-4o',
    capabilities=[Managed(instructions=prompt, model_settings=settings)],
)

result = agent.run_sync('My order is late.')
print(result.output)
```

### Phase 1 scope

- Supports `instructions` and `model_settings` variables only.
- Values are resolved fresh per run via `variable.get()`, surfaced through the existing `get_instructions` / `get_model_settings` capability hooks, and the `ResolvedVariable` is entered as a context manager in `wrap_run` so Logfire baggage tags downstream spans with the active label.
- Additive semantics: managed values layer on top of agent-level defaults; call-site `run()` args still win.
- Stacking multiple `Managed` capabilities is supported via `CombinedCapability` middleware ordering.

A whole-spec variable, plus `model` / `metadata` fields and a new `contribute_run_spec` capability hook, are planned as a Phase 2 follow-up.

### Files

- `pydantic_ai_slim/pydantic_ai/managed/__init__.py` — new subpackage.
- `pydantic_ai_slim/pydantic_ai/managed/logfire.py` — `Managed` capability.
- `pydantic_ai_slim/pyproject.toml` — bumps `logfire[httpx]>=4.24.0` (first version with managed variables).
- `tests/test_managed.py` — 7 tests via `FunctionModel` using real `logfire.var()` defaults and `.override()` (no network).
- `docs/managed.md` + `docs/api/managed.md` — user guide and API reference stub, wired into `mkdocs.yml` nav and llms.txt.

Existing `tests/test_logfire.py` and `tests/models/test_fallback.py` snapshots were regenerated to pick up the new `logfire.metrics` span attribute emitted by `logfire>=4.24.0`.

### Checklist


DouweM:
Addressing review feedback — two fixup commits pushed (c3c21184 + ed5a122b).

**Applied directly:**
- Model names → `openai:gpt-5.2` in both `docs/managed.md` examples and the `Managed` docstring example (devin, github-actions).
- Dropped "Phase 1" framing from user-facing docs/docstring; describe what's currently supported and what's planned without the internal milestone language (github-actions + [feedback_docs_framing memory](https://github.com/pydantic/pydantic-ai/pull/5107/files#r-)).
- Fixed the inaccurate "call-site always wins over both" statement in `docs/managed.md`; now accurately describes concatenation for instructions and per-key dict merge for model settings (github-actions).
- Reverted `logfire[httpx]` minimum back to `>=4.16.0` (github-actions): the `Managed` import already raises a clear `ImportError` pointing at `logfire>=4.24.0` when `logfire.variables` is missing, so existing users of the instrumentation integration on 4.16–4.23 aren't forced to upgrade. Updated the error message to suggest both `pip install "pydantic-ai-slim[logfire]"` (fresh install) and `pip install -U "logfire>=4.24.0"` (upgrade path).
- Fixed a collateral typecheck regression in `durable_exec/temporal/_logfire.py:53`: `logfire_config.token` is `str | list[str] | None` as of 4.24 (multi-project tokens). Falls back to the first token for the metrics OTel export, which routes to a single project.
- Rewrote `tests/test_managed.py` to use `snapshot(result.all_messages())` per `tests/AGENTS.md` rule (devin, github-actions). The echo `FunctionModel` reflects effective instructions/settings into the response text so the snapshot also captures that the managed values actually reached the model request (since model settings aren't stored on `ModelRequest`).

**Flagging for discussion, @DouweM:**

1. **`_resolved_*` fields on the `Managed` dataclass** — github-actions points out that per-run resolved state is stored as dataclass fields on the user-facing config object, which is visible in `dir()` / `dataclasses.fields()`. The `replace(self)` + `init=False, repr=False, compare=False` pattern is correct at runtime, but it's a slightly unusual API shape. Worth deciding whether to split into a separate internal "resolved" object going forward as more fields land in Phase 2. Non-blocking — left as-is for now.

2. **`ModelSettings` key validation** — the `cast(ModelSettings, self._resolved_model_settings.value)` at `logfire.py:102` passes `dict[str, Any]` from the logfire variable through without checking that the keys are real `ModelSettings` fields. A UI typo like `{"temperture": 0.5}` would be silently ignored. Since these values come from the Logfire UI rather than in-code settings, typos are more plausible than usual. Options: (a) validate against `ModelSettings.__annotations__` in `for_run` and raise a clear error, (b) document the limitation, (c) ignore for Phase 1. My preference is (a) — a clear error beats silent ignore — but it adds scope. Non-blocking either way.

**Unrelated flake:** the `3.11 (lowest-versions)` failure on the previous run was in `tests/test_examples.py::test_docs_examples[docs/graph/beta/joins.md:297:first_value_reducer.py]` — `Tasks completed: 1` vs `2`, which is a race in a beta graph reducer's short-circuit path. Not caused by this PR; should clear on the re-run.

DouweM:
CI is fully green on 2998aae2 — all 40 required checks pass, coverage back at 100%.

Recap of what landed after the initial red CI:
1. **a360d126** — skip `tests/test_managed.py` when the `logfire` extra isn't installed (slim/evals jobs), and regenerate `tests/test_logfire.py` + `tests/models/test_fallback.py` span snapshots to include the new `logfire.metrics` attribute from logfire 4.24+.
2. **c3c21184** — review feedback round 1: `openai:gpt-5.2` everywhere, drop "Phase 1" framing from user-facing text, fix the inaccurate "call-site always wins" wording, revert `logfire[httpx]` minimum to `>=4.16.0` (the runtime check in `Managed` gives a clear error with an upgrade command), and handle `logfire.config.token: list[str]` in `durable_exec/temporal/_logfire.py` for logfire's multi-project tokens.
3. **ed5a122b** — rewrite `tests/test_managed.py` to use `snapshot(result.all_messages())` per `tests/AGENTS.md` rule; the echo `FunctionModel` reflects effective instructions/settings into the response text so the snapshot also proves the managed values reached the model request.
4. **27ee80e1** — cover the new multi-project token fallback in `_logfire.py:55` via a realistic `test_logfire_plugin` extension that configures `instance.config.token = ['test-a', 'test-b']` (restores 100% coverage).
5. **2998aae2** — fix `docs/managed.md` install section to accurately describe the dependency situation (extra pulls `>=4.16.0`; managed variables need `>=4.24.0`, enforced at import time).

Ready for re-review, @DouweM. Two open non-blocking discussion items from the auto-review are still waiting on your input:
- **`_resolved_*` fields on the `Managed` dataclass** — should per-run state move into a separate internal object? Non-blocking; relevant as more fields land in Phase 2.
- **`ModelSettings` key validation** — logfire variable dict passes through `cast()` without verifying keys against `ModelSettings` fields, so a UI typo (`"temperture"`) would be silently ignored. Options are validate+raise in `for_run`, document the limitation, or leave to Phase 2.

pydantic_ai_slim/pydantic_ai/managed/logfire.py
```diff
@@ -0,0 +1,135 @@
...
+                stack.enter_context(self._resolved_model_settings)
+            return await handler()
+
+
+@dataclass
+class ManagedAgentSpec(AbstractCapability[AgentDepsT]):
+    agent_spec: Variable[AgentSpec] | None = None
+    """ Managed spec contributing the entire spec."""
+
+    def contribute_run_spec(self) -> AgentSpec | None:
```
adtyavrdhn, line 128: This might need to be async

pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -1059,8 +1059,19 @@ async def main():
...
+        leaves = collect_leaves(base_capability)
+        managed_spec: AgentSpec | None = None
+
+        for capability in leaves:
+            spec_from_capability = capability.contribute_run_spec()
+            if spec_from_capability is not None:
+                managed_spec = spec_from_capability
+                break
+
+        resolved = self._resolve_spec(managed_spec or spec)
```
adtyavrdhn, line 1074: This is not ideal and will need to be discussed but I was just playing with it to make it work atm. I think for the fields which are present as managed variable should not be merged

pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -1059,8 +1059,19 @@ async def main():
                 'budget) here, and `Agent(retries=...)` for tool retries.'
             )
 
-        # Resolve spec contributions (additive at run time)
-        resolved = self._resolve_spec(spec)
+        override_cap = self._override_root_capability.get()
+        base_capability = override_cap.value if override_cap is not None else self._root_capability
+
+        leaves = collect_leaves(base_capability)
```
adtyavrdhn, line 1065: It is unfortunate that I ended up having to do this to find the capability and the spec for us to get the contribution out but it makes sense for any other capability which uses the hook and is not managed agent spec

adtyavrdhn: Unless we wanna make it a private hook which I am not very opposed to

pydantic_ai_slim/pydantic_ai/managed/logfire.py
```diff
@@ -0,0 +1,135 @@
...
+        async with AsyncExitStack() as stack:
+            if self._resolved_instructions is not None:
+                stack.enter_context(self._resolved_instructions)
+            if self._resolved_model_settings is not None:
+                stack.enter_context(self._resolved_model_settings)
+            return await handler()
+
+
+@dataclass
+class ManagedAgentSpec(AbstractCapability[AgentDepsT]):
```
adtyavrdhn, line 124: Atm platform sends all of these assuming different bits but if we can make platform aware of the shape here(although not clean) this can be easier to just use

pydantic_ai_slim/pydantic_ai/managed/logfire.py
```diff
@@ -0,0 +1,135 @@
...
+    """ Managed spec contributing the entire spec."""
+
+    def contribute_run_spec(self) -> AgentSpec | None:
+        return self.agent_spec.get().value if self.agent_spec is not None else None
+
+    # TODO(baggage): like `Managed`, this should resolve `agent_spec` once and enter the
+    # `ResolvedVariable` as a context manager around the run (a `wrap_run` override) so child
+    # Logfire spans are tagged with the spec's label/version. Not yet implemented: `contribute_run_spec`
+    # resolves in the pre-model window, but `wrap_run` runs later, so the resolved variable needs to be
+    # stashed where `wrap_run` can re-enter it. See `Managed.wrap_run` for the pattern to mirror.
```
adtyavrdhn, line 135: We'll also need managed tools and toolsets separately because spec doesn't support them and capabilities only supports the caps which are present in the registry, there are PRs by Bill on the harness although a bit far from ideal atm but I would prefer adding those(or a version of them) to the harness right now instead of trying to figure out what the shape should look here. 

We only need the hook I added in this PR

pydantic_ai_slim/pydantic_ai/agent/__init__.py
adtyavrdhn, line 1121: Model needed to be swapped before we reached this point

adtyavrdhn: Can't swap it later because it is used in ctx and the hooks depend on it

DouweM:
Superseded by https://github.com/pydantic/pydantic-ai-harness/pull/331

PR #4943, opened 2026-04-02
Add server-side compaction support via `OpenAICompaction` and `AnthropicCompaction` capabilities

DouweM:
## Summary

- Adds `CompactionPart` message type for provider-specific conversation compaction
- Adds `OpenAICompaction` capability (calls `responses.compact` endpoint) and `AnthropicCompaction` capability (configures `context_management` for automatic server-side compaction)
- Adds `RoutedCapability` base class and `ProviderCompaction` unified capability that routes to the correct provider implementation
- Adds compaction handling to all model files, agent graph, and UI layers
- Includes VCR cassette integration tests and documentation

Closes #4013, closes #4267
Supersedes #4145

### Pre-Review Checklist

- [x] Any **AI generated code** has been reviewed line-by-line by the human PR author, who stands by it.
- [x] No **breaking changes** in accordance with the [version policy](https://github.com/pydantic/pydantic-ai/blob/main/docs/version-policy.md).
- [x] **Linting and type checking** pass per `make format` and `make typecheck`.
- [x] **PR title** is fit for the [release changelog](https://github.com/pydantic/pydantic-ai/releases).

### Pre-Merge Checklist

- [x] New **tests** for any fix or new behavior, maintaining 100% coverage.
- [x] Updated **documentation** for new features and behaviors, including docstrings for API docs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5108, opened 2026-04-15
Add stateful compaction mode to `OpenAICompaction`

DouweM:
Follow-up to #5086. `OpenAICompaction` now supports both of OpenAI's compaction APIs, selected by a `stateless` flag.

### Two modes

- **Inline** (default, `stateless=False`): configures OpenAI's [server-side auto-compaction](https://developers.openai.com/api/docs/guides/compaction) via the `context_management` field on the regular `/responses` request. The server triggers compaction when input tokens cross a threshold it manages for you. Compatible with `openai_previous_response_id='auto'` and server-side conversation state.

  Configurable via `token_threshold` (maps to `compact_threshold`). If omitted, OpenAI picks a server-side default.

- **Dedicated** (`stateless=True`): the existing behavior from #4943 — calls the stateless `/responses/compact` endpoint via a `before_model_request` hook when your trigger condition is met. Fully stateless and ZDR-friendly. Use this in ZDR environments or when you need explicit out-of-band control over when compaction runs.

  Requires either `message_count_threshold` or a custom `trigger` callable.

### Mode resolution

If `stateless` is not set explicitly, it's inferred from which parameters you provide: supplying any dedicated-mode-only parameter (`message_count_threshold` or `trigger`) implies `stateless=True`; otherwise inline mode is used. Mixing mode-specific fields raises ``UserError``.

```python
# Inline (server-managed default threshold)
OpenAICompaction()

# Inline with custom threshold
OpenAICompaction(token_threshold=100_000)

# Dedicated (unchanged from #4943)
OpenAICompaction(message_count_threshold=20)

# Explicit dedicated mode with a custom trigger
OpenAICompaction(stateless=True, trigger=lambda msgs: len(msgs) > 20)

# Errors — `token_threshold` only valid for inline mode
OpenAICompaction(stateless=True, token_threshold=100_000, message_count_threshold=5)
```

### `OpenAIResponsesModelSettings.openai_context_management`

Added a new setting so users who want to configure `context_management` directly (e.g. without the capability) can do so. The capability merges its entry on top of any user-provided entries at resolve time, mirroring `AnthropicCompaction`'s pattern.

### Breaking change: bare `OpenAICompaction()` default

Before this PR, `OpenAICompaction()` with no arguments called the dedicated `/compact` endpoint (via the `message_count_threshold=10` default). Now, bare `OpenAICompaction()` runs in inline mode with OpenAI's server-managed threshold.

To keep the old behavior, pass an explicit trigger:

```python
OpenAICompaction(message_count_threshold=10)  # still works, auto-infers stateless=True
```

Since #4943 just shipped in the prior release, the blast radius for existing callers is small.

### Deprecation: `OpenAICompaction(instructions=...)`

`OpenAICompaction` shipped in #4943 with an `instructions` parameter documented as *"Custom instructions for the compaction summarization"*. That doesn't match what OpenAI's API actually does. Per the [OpenAI SDK docstring](https://github.com/openai/openai-python/blob/main/src/openai/resources/responses/responses.py): *"instructions: A system (or developer) message inserted into the model's context."* — it's the same system-prompt slot as on `/responses`, **not** a compaction directive. Compare to Anthropic's `compact_20260112.instructions`, which is documented as *"Additional instructions for summarization"* and genuinely is a compaction directive.

Passing `instructions` now emits a `DeprecationWarning`. The field is slated for removal in a future major version. Inline mode does not support any equivalent (OpenAI's `context_management` field accepts only `compact_threshold`, per the SDK's `ContextManagement` TypedDict).

### Tests

- 10 new validation/unit tests in `tests/test_capabilities.py` covering mode inference, model-settings merging, mixed-field errors, missing-trigger errors, and the deprecation warning.
- New VCR test `test_openai_responses_compact_inline_mode` that records a real request against `gpt-4.1` and verifies the recorded cassette contains `context_management: [{'type': 'compaction', 'compact_threshold': 50000}]`.
- All 5 existing compaction tests still pass unchanged — existing callers of `OpenAICompaction(message_count_threshold=N)` auto-infer `stateless=True` and hit the same code path as before.

- Closes #<issue>

### Checklist


pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -2863,6 +2890,8 @@ async def _get_event_iterator(self) -> AsyncIterator[ModelResponseStreamEvent]:
                     elif isinstance(chunk.item, responses.response_output_item.McpListTools):
                         call_part, _ = _map_mcp_list_tools(chunk.item, self.provider_name)
                         yield self._parts_manager.handle_part(vendor_part_id=f'{chunk.item.id}-call', part=call_part)
+                    elif isinstance(chunk.item, ResponseCompactionItem):
+                        pass  # handled in ResponseOutputItemDoneEvent with final encrypted_content
```
DouweM: Maybe we should stream a PartStartEvent already so that a UI can render it if it likes?

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -1765,6 +1778,19 @@ def _process_response(  # noqa: C901
                 if file_part:  # pragma: no branch
                     items.append(file_part)
                 items.append(return_part)
+            elif isinstance(item, ResponseCompactionItem):
+                # Stateful compaction: the server-side `context_management` configuration
+                # triggered a compaction and emitted this item alongside the regular output.
+                # We round-trip it via `CompactionPart` so subsequent requests can send the
+                # encrypted content back via the `input` array (see `_map_messages`).
+                items.append(
+                    CompactionPart(
```
DouweM: We convert ResponseCompactionItem to CompactionPart in a few places now, can we centralize it please

Issue #3105, opened 2025-10-07
Allow empty response


dconathan:
### Initial Checks

- [x] I confirm that I'm using the latest version of Pydantic AI
- [x] I confirm that I searched for my issue in https://github.com/pydantic/pydantic-ai/issues before opening this issue

### Description

There are use cases where it is legitimate for the model to return an empty response.  Currently this causes agents to crash in unintuitive ways (Pydantic AI retries the request and then the retry request fails with a Bad Request error).  In prior versions this just raised a `UnexpectedModelBehavior: Received empty model response`

Relevant retry code is here: https://github.com/pydantic/pydantic-ai/blob/dc3c9e98a86dbf4d4b2b8f5c0f3612bc54deaddc/pydantic_ai_slim/pydantic_ai/_agent_graph.py#L546-L549

Current behavior causes the request to retry which then for me results in a `BadRequestError`:

```
AzureException BadRequestError - Invalid value for 'content': expected a string, got null.
```

An example "legitimate" use case is if you have a "relevant text extractor" agent and it is instructed to only extract relevant verbatim text from the input and to return no text if there is no relevant text (see example).

Currently an easy workaround is to prompt the model to return the text "no relevant text" instead of an empty string - but this feels like a hacky workaround.

Is Pydantic AI intentionally making the opinionated choice that "there is no valid reason a model could return an empty response" ?  What if I have fine-tuned a model to respond with empty strings in certain scenarios?  Should we have an opt in like `allow_empty_response=True` in model or agent settings?

### Example Code

```Python
from pydantic_ai import Agent
from pydantic_ai.models.openai import OpenAIChatModel
from pydantic_ai.providers.openai import OpenAIProvider


agent = Agent(
    model=OpenAIChatModel(
        "azure/gpt-5-nano",
        provider=OpenAIProvider(
            base_url="...",
        ),
    ),
    instructions="You are a text extractor agent. You always respond with verbatim text from the provided context that is relevant to George Washington. If there is no relevant text, respond with an empty string. Do not provide any additional commentary or information.",
)

text = """Taylor Alison Swift (born December 13, 1989) is an American singer-songwriter. Known for her autobiographical songwriting, artistic reinventions, and cultural impact, Swift is the highest-grossing live music artist, the wealthiest female musician, and one of the best-selling music artists of all time.

[...]

PR #5117, opened 2026-04-16
fix: preserve prior text recovery for Agent(output_type=str | None)

### LINKED ISSUE #5116: Agent(output_type=str | None) skips prior-text recovery after tool-follow-up empty response
### Initial Checks

- [x] I confirm that I searched for existing issues and PRs before opening this report

### Description

`Agent(output_type=str | None)` appears to skip the existing prior-text recovery path after a tool-follow-up empty response.

There are two related behaviors in the codebase today:

1. `str | None` now allows a genuinely empty final response (merged in #4042 / issue #3735).
2. Earlier text is intentionally recoverable when a model returns text alongside a tool call and then follows with an empty or thinking-only response (see issue #3105 and `test_thinking_only_response_backward_looking_recovery`).

On the current latest release, those two behaviors conflict: the new `allows_none` empty-response branch runs before `_recover_text_from_message_history(...)`, so the run returns `None` even when there is already recoverable assistant text in history.

That looks like a regression rather than intended `str | None` behavior.

### Minimal Repro

```python
import asyncio

from pydantic_ai import Agent
from pydantic_ai.messages import ModelMessage, ModelResponse, TextPart, ToolCallPart


async def main() -> None:
    call_count = 0

    def model_function(messages: list[ModelMessage], info: AgentInfo) -> ModelResponse:
        nonlocal call_count
        call_count += 1
        if call_count == 1:
            return ModelResponse(
                parts=[
                    TextPart(content='RECOVER-ME'),
                    ToolCallPart(tool_name='save_progress', args={'data': 'x'}, tool_call_id='call-1'),
                ]
            )
        return ModelResponse(parts=[])

    agent = Agent(FunctionModel(model_function), output_type=str | None)

    @agent.tool_plain
    def save_progress(data: str) -> str:
        return 'saved'

    result = await agent.run('hello')
    print(result.output)


asyncio.run(main())
```

### Actual behavior

`result.output is None`

### Expected behavior

`result.output == 'RECOVER-ME'`

The output type is optional, so `None` should still be allowed for genuinely empty runs, but not at the cost of discarding recoverable assistant text that the library already knows how to preserve.

### Why I think this should be fixed

This keeps the logic aligned with the current design:
- use prior-text recovery when there is already recoverable assistant text
- only fall back to `None` when the run is truly empty

A minimal fix seems to be moving the `is_empty and output_schema.allows_none` branch below the prior-text recovery check for text-capable outputs.

### Related context

- #3105
- #3735
- #4042

```text
Verified locally on the current latest release/worktree based on commit 81bb8d2c96267f8ab18d2b93c45e56221c67998d.
```

shaun0927:
- Closes #5116

### Checklist

## Summary

- preserve backward-looking text recovery for `Agent(output_type=str | None)` when a tool-follow-up response is empty
- only return `None` for optional string output when there is no recoverable assistant text in history
- add a regression test covering `text + tool call -> empty follow-up`

## Why

Issue #3105 established that earlier assistant text should be recoverable when a model returns text alongside a tool call and then follows with an empty response.

PR #4042 correctly added support for genuinely empty runs to finish as `None` for `Agent(output_type=str | None)`, but the new `allows_none` empty-response branch now runs before `_recover_text_from_message_history(...)`.

That changes the semantics for text-capable outputs:
- before: recover prior assistant text when it already exists
- now: prefer `None` even when the library can still recover prior assistant text

This PR restores the earlier recovery precedence and keeps the optional-output behavior for truly empty runs.

## Scope

This is intentionally narrow:
- no API changes
- no docs changes
- no streaming-path changes in this PR

## Test plan

- [x] Added a regression test in `tests/test_agent.py` for `output_type=str | None` with `TextPart + ToolCallPart` followed by an empty response
- [x] Ran a focused runtime repro locally before the patch and confirmed the behavior changed from `None` to the prior text after the patch
- [x] Ran a focused runtime check that a genuinely empty `output_type=str | None` run still returns `None`
- [x] Ran `uv run ruff check pydantic_ai_slim/pydantic_ai/_agent_graph.py tests/test_agent.py`
- [ ] Full pytest suite / CI

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/_agent_graph.py
+++ b/pydantic_ai_slim/pydantic_ai/_agent_graph.py
@@ -1028,19 +1028,6 @@ class CallToolsNode(AgentNode[DepsT, NodeRunEndT]):
 
                         raise exceptions.ContentFilterError(message, body=body)
 
-                    # If the output type allows None, an empty response is a valid result.
-                    if is_empty and output_schema.allows_none:
-                        run_context = _build_output_run_context(ctx)
-                        try:
-                            result_data = await _run_output_validators(ctx, cast(NodeRunEndT, None), run_context)
-                            self._next_node = self._handle_final_result(ctx, result.FinalResult(result_data), [])
-                        except ToolRetryError as e:
-                            ctx.state.increment_retries(ctx.deps.max_result_retries, error=e)
-                            self._next_node = ModelRequestNode[DepsT, NodeRunEndT](
-                                _messages.ModelRequest(parts=[e.tool_retry])
-                            )
-                        return
-
                     # Try to recover text from a previous model response.
                     # This handles the case where the model returned text alongside tool calls
                     # (so the text was discarded in favor of executing the tools) and subsequently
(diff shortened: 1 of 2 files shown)


DouweM:
Marking as draft pending discussion in https://github.com/pydantic/pydantic-ai/issues/5116.

shaun0927:
Closing this draft for now based on the discussion in #5116.

I don't have a concrete active user-facing failure case here, and Douwe's clarification suggests the current `str | None` behavior is intentional. Rather than keep iterating on a behavior change that isn't maintainer-aligned, I'm closing the PR.

If a concrete failing use case shows up later, I'm happy to reopen the discussion with a more targeted repro.

PR #5120, opened 2026-04-16
Make XSearch capability model-agnostic via subagent fallback

### LINKED ISSUE #5119: Make XSearch model-agnostic via subagent fallback
Currently, `XSearchTool` (X/Twitter search) is exclusive to xAI models. Other models (OpenAI, Anthropic, Google, etc.) cannot use it, limiting its utility.

## Proposal

Follow the same pattern as `ImageGeneration` — which uses a subagent fallback when the primary model doesn't support image generation natively — to make `XSearch` work with any model.

When the outer agent's model doesn't support `XSearchTool` as a builtin, the `XSearch` capability would delegate to a subagent running an xAI model (e.g. `grok-4-1-fast-non-reasoning`) that has native X search support. The subagent performs the search and returns the results as text to the outer agent.

### API

```python
from pydantic_ai import Agent
from pydantic_ai.capabilities import XSearch

# Works with any model — falls back to xAI subagent
agent = Agent('openai:gpt-5.4', capabilities=[XSearch(fallback_model='xai:grok-4-1-fast-non-reasoning'), from_date=one_week_ago)])

# With xAI models, uses the native builtin directly (no subagent overhead)
agent = Agent('xai:grok-3', capabilities=[XSearch()])

# Custom fallback model or disable fallback
agent = Agent('anthropic:claude-sonnet-4-20250514', capabilities=[XSearch(fallback_model='xai:grok-4-1-fast-non-reasoning')])
agent = Agent('xai:grok-3', capabilities=[XSearch(fallback_model=None)])
```

### Implementation

- Add `fallback_model` parameter to `XSearch`
- Create `common_tools/x_search.py` with `XSearchSubagentTool` and `x_search_tool()` factory (mirroring `common_tools/image_generation.py`)
- Move `XSearch` capability from `models/xai.py` to `capabilities/x_search.py` (re-export from `models/xai.py` for backward compat)
- Register in `CAPABILITY_TYPES` for spec serialization

colesmcintosh:
- Closes #5119

## Summary

Moves the `XSearch` capability out of `models/xai.py` and into `capabilities/x_search.py`, and adds an optional `fallback_model` parameter. When set, non-xAI models delegate X search to a subagent running the specified xAI model — mirroring the `ImageGeneration` fallback pattern.

On xAI models, the native builtin is used directly with no overhead. On non-xAI models, users must explicitly pass `fallback_model` (e.g. `'xai:grok-4-1-fast-non-reasoning'`) — there is no default, so the capability fails fast rather than silently picking a model on the user's behalf.

`XSearch` is re-exported from `models/xai.py` for backward compatibility.

```python
# xAI — native builtin
Agent('xai:grok-4-1-fast-non-reasoning', capabilities=[XSearch()])

# Non-xAI — explicit subagent fallback required
Agent('openai:gpt-5', capabilities=[XSearch(fallback_model='xai:grok-4-1-fast-non-reasoning')])
```

### Checklist


colesmcintosh:
@DouweM I was using this script to test actual usage locally:
```python
"""
Using the xAI X Search capability (model-agnostic via subagent fallback).

"""

from datetime import datetime, timedelta, timezone

from pydantic_ai import Agent
from pydantic_ai.capabilities import XSearch

load_dotenv()


one_week_ago = datetime.now(tz=timezone.utc) - timedelta(days=7)

agent = Agent(
    'openai:gpt-5.4',
    capabilities=[
        XSearch(
            fallback_model="xai:grok-4-1-fast-non-reasoning",
            allowed_x_handles=["claudeai"]
        ),
    ],
)

result = agent.run_sync("Tell me about the latest Claude release")

print('=== Output ===')
print(result.output)

print('\n=== Tool calls (and more) ===')
print(result.all_messages())
```

dsfaccini:
@colesmcintosh I tried it but it didn't use the xsearch tool https://play.pydantic.work/zVg6u342

pydantic_ai_slim/pydantic_ai/models/xai.py
```diff
@@ -14,7 +14,7 @@
 from .._output import OutputObjectDefinition
 from .._run_context import RunContext
 from ..builtin_tools import CodeExecutionTool, FileSearchTool, MCPServerTool, WebSearchTool, XSearchTool
-from ..capabilities.builtin_or_local import BuiltinOrLocalTool
+from ..capabilities.x_search import XSearch as XSearch  # re-export for backward compat
```
dsfaccini, line 17: the `as` is def not necessary

colesmcintosh: The `as XSearch` is load-bearing here — without it, pyright (strict mode, which this repo uses) treats the import as private and emits `"XSearch" is not exported from module "pydantic_ai.models.xai"` for anyone still doing the old import. Since `xai.py` has no `__all__`, the redundant alias is the standard PEP 484 way to mark it as a deliberate re-export. [Ref](https://typing.readthedocs.io/en/latest/spec/distributing.html#library-interface).

dsfaccini: interesting!

docs/models/xai.md
```diff
@@ -75,13 +75,13 @@ agent = Agent(model)
 
 ## X Search
 
-xAI models support searching X (formerly Twitter) for real-time posts and content. The recommended way to enable it is with the [`XSearch`][pydantic_ai.models.xai.XSearch] capability, which configures the underlying `x_search` builtin tool and can be passed alongside any other capabilities on the agent. See the [xAI X Search documentation](https://docs.x.ai/developers/tools/x-search) for the full list of supported options.
+xAI models support searching X (formerly Twitter) for real-time posts and content. The recommended way to enable it is with the [`XSearch`][pydantic_ai.capabilities.XSearch] capability, which configures the underlying `x_search` builtin tool and can be passed alongside any other capabilities on the agent. When you're already using an xAI model, no `fallback_model` is needed. See the [xAI X Search documentation](https://docs.x.ai/developers/tools/x-search) for the full list of supported options.
```
dsfaccini, line 78: not sure the mention of `fallback_model` is necessary here, bc the snippet doesn't reference it and this is the doc for the `XaiModel`

pydantic_ai_slim/pydantic_ai/builtin_tools/AGENTS.md
```diff
@@ -2,9 +2,9 @@
 
 ## Adding a new builtin tool
 
-- Every builtin tool must have a corresponding capability extending `BuiltinOrLocalTool` — capabilities are the primary user-facing API for enabling tool features on agents; a builtin tool without a capability is undiscoverable for users working with the capabilities list
-  - Vendor-specific tools: capability lives in the provider's model file (e.g., `XSearch` in `models/xai.py`)
-  - Cross-provider tools: capability lives in `capabilities/` (e.g., `WebSearch`, `ImageGeneration`)
+- Every builtin tool must have a corresponding capability extending `BuiltinOrLocalTool` in `capabilities/` — capabilities are the primary user-facing API for enabling tool features on agents; a builtin tool without a capability is undiscoverable for users working with the capabilities list
+  - Local fallback (e.g., `WebSearch`, `WebFetch`): capability falls back to a function tool on providers without builtin support
+  - Subagent fallback (e.g., `ImageGeneration`, `XSearch`): capability delegates to a subagent running another provider's model via `fallback_model`
```
dsfaccini, line 7: I don't oppose the rewrite but it does take away the info about where to place stuff properly. @DouweM what do you opine

pydantic_ai_slim/pydantic_ai/models/xai.py
```diff
@@ -198,83 +196,6 @@ class XaiModelSettings(ModelSettings, total=False):
     """
 
 
-@dataclass(init=False)
```
dsfaccini, line 201: I do love shortening model files <3

dsfaccini:
ok I put some more work in and the playground works now https://play.pydantic.work/p/zVg6u342/s/wvioeL?version=1.84.1&tab=branch

it's cool! one thing I discovered is that it won't work if a user uses the x.ai API via the openai completions model, but this is not something I'd want to support, since x.ai is moving its features away from the completions compatible API any way. It may even be that they don't support the xsearch feature over the completions API. Just wanted to add as a fun note.

I left some comments, looks mostly good to me! @DouweM will take a look in the next couple of days

colesmcintosh:
@dsfaccini Thanks for taking a look! Looks like [they don't/won't support native tools for completions](https://docs.x.ai/developers/model-capabilities/text/comparison)

docs/models/xai.md
DouweM: Yeah I agree, I'd rather link to the capabilities -> XSearch doc here "for more details".

pydantic_ai_slim/pydantic_ai/builtin_tools/AGENTS.md
DouweM: Hmm, I'm ok with the update because I agree that if possible, we should ship with subagent fallback support from the get-go.

pydantic_ai_slim/pydantic_ai/capabilities/x_search.py
```diff
@@ -0,0 +1,135 @@
...
+        if self.fallback_model is None:
+            return None
+        from pydantic_ai.common_tools.x_search import x_search_tool
+
+        return x_search_tool(model=self.fallback_model, builtin_tool=self._resolved_builtin())
+
+    def _resolved_builtin(self) -> XSearchTool:
+        """Get the XSearchTool for the fallback, with capability-level settings applied."""
+        if isinstance(self.builtin, XSearchTool):
+            return self.builtin
```
DouweM: The `ImageGeneration` capability has some `kwargs` merging logic here since users can specify both `builtin=...` and individual kwargs, we should have that here as well (even though the duplication is a bit awkward)

pydantic_ai_slim/pydantic_ai/capabilities/x_search.py
```diff
@@ -0,0 +1,135 @@
...
+        return self._default_builtin()
+
+    def _requires_builtin(self) -> bool:
```
DouweM: Unsure why we need this method override, ImageGen doesn't have it.

pydantic_ai_slim/pydantic_ai/capabilities/x_search.py
DouweM: The reason we still support `builtin=...` even though we copy all the kwargs, is that we support dynamic builtin tools based on `RunContext`.

DouweM (changes requested):

DouweM:
@colesmcintosh Thanks Cole :)

Issue #5038, opened 2026-04-10
feat: Functional Programming API for Agents/Direct Model Requests


kmheckel:
### Description

# Feature Request: Signature-Based Functional API for Pydantic AI

As an upfront note, I'm an avid user of Pydantic AI and spent about 6 hours iterating with Gemini 3.1 and Opus 4.6 on this feature request + a couple of proofs-of-concept (PoC) scripts to flesh out the behavior/catch important design considerations. Hopefully between the descriptions and the PoCs, it should be straightforward to implement/integrate once the final edge cases are fleshed out.

## Executive Summary
This proposal introduces a first-class **Functional API** for Pydantic AI, enabling developers to define agents and model requests using native Python functions and decorators. By leveraging function signatures for tool extraction, dependency injection, and recursive orchestration, this API provides a developer experience (DX) similar to FastAPI or FastMCP. This feature request is to bridge the gap between Pydantic AI Agents and directly using a client SDK such as `openai` while providing backend inference flexibility.

I propose two primary function decorator based interfaces:
1.  **`@Agent`**: A high-level, multi-turn functional API for the `Agent` class, including support for passing and receiving message history as part of the function prototype.
2.  **`@Direct`**: A lower-level wrapper over `pydantic_ai.direct.model_request` for one-shot inference requests.

---

## 1. The "@Agent" Functional Interface
The `@Agent` decorator allows developers to treat Pydantic AI agents as standard Python functions. It bridges the gap between declarative agent configuration and imperative execution. in the agent_fn.py PoC, it's implemented as a monkeypatch of the Agent class to demonstrate functionality prior to forking/opening a PR, in production it would be part of the Agent class itself.

### Key Benefits
- **Reduces Boilerplate**: Define an agent once or not at all, and use functions to define interactions.
- **Signature-Driven Tooling**: Any parameter hinted as `Callable` is automatically extracted as a `Tool`. Tools can also be passed as arguments to the decorator to be more explicit.
- **Automatic Multimodal Handling**: Detects `BinaryContent` or `ImageUrl` in arguments and injects them into the user request.
- **Streamlined for Code Mode**: these functions could be passed to Monty and more easily orchestrated by an agent.
- **Dynamic Instruction Formatting**: Prompts defined in `instructions` or `system_prompt` are dynamically formatted using the function's arguments.

### Example
```python
@Agent(model=ollama_model, instructions="Check the weather in {city} via tool.")
[...]

PR #5123, opened 2026-04-16
Feat: Agent Functional API / Decorator

kmheckel:
- Adds plan and functioning proof of concept script in support of issue #5038 

Proposes implementing Agent.__call__() as both a class and instance decorator for functions to allow for a functional approach for agentic programming.

Feature is self-contained / should not require any broad or breaking changes. Scope was discussed with maintainers via video call prior to submission of this draft PR.

### Checklist

- [x ] Any **AI generated code** has been reviewed line-by-line by the human PR author, who stands by it.
- [ x] No **breaking changes** in accordance with the [version policy](https://github.com/pydantic/pydantic-ai/blob/main/docs/version-policy.md).
- [ x] **PR title** is fit for the [release changelog](https://github.com/pydantic/pydantic-ai/releases).


plan/agent_fn_plan.md
```diff
@@ -0,0 +1,176 @@
+# Implementation Plan: Lightweight Functional `@Agent` API
```
DouweM: @kmheckel Thanks Kade appreciate it! I had a quick look and it's very clear from the plan and the code that it's building this _as a patch on top of Pydantic AI_ (which is cool) rather than as a _feature of Pydantic AI_. So I'd like to see a plan that, when given to an agent, would make the functional style natively implemented natively on the `(Abstract)Agent` class, and would not have to patch `StepContext` for `pydantic-graph`, but would instead propose a change to `pydantic-graph` (in this repo) to support functions that don't take `StepContext`.

kmheckel: For clarification:

The only patch is on the Agent class purely to show self-contained functionality, i.e. the code works prior to integration. 100% agree/the plan is to make the decorator logic part of (Abstract)Agent.__call__() rather than a patch.

StepContext / pydantic-graph doesn't require modification for this proposal. This is because the decorated function can detect if a parameter was annotated as StepContext and can forward the argument's ctx.deps and ctx.input to the agent; this seems like a more surgical route versus changing graph to support nodes that don't take StepContext.

I'll add notes to the plan to make it explicit to the agent not to just copy-paste the code as a patch.

PR #5125, opened 2026-04-16
Online evaluation via OpenTelemetry events

dmontagu:
## Summary

- Reworks pydantic-evals online evaluation to emit `gen_ai.evaluation.result` OTel events by default (GenAI semconv), matching the opt-out pattern already used for offline evals. Events are parented to the referenced span so they appear as child events of the original function call in the trace. Failure events are emitted at `SeverityNumber.WARN`.
- New public surface:
  - `@evaluate` decorator — attach evaluators to any callable; they run in the background after each call completes.
  - `OnlineEvaluation` capability — pydantic-ai agent integration. Attaches evaluators to agent runs, dispatching asynchronously after each `agent.run()`. Merges config + run metadata into `EvaluatorContext` and parses W3C traceparent for `SpanReference`.
  - `EvaluationSink` protocol reframed as *additional* destinations for programmatic handling (in-memory capture, Slack fan-out, non-OTel backends). OTel emission happens unconditionally unless the user sets `emit_otel_events=False`.
  - `SinkPayload` container passed to `EvaluationSink.submit(payload)` — bundles results, failures, context, span reference, and target. Constructed by the framework; fields may be added in future releases, so sinks should read only what they need and never instantiate directly.
  - `OnlineEvalConfig.include_baggage` — when True (default), current OTel baggage entries are attached as attributes on each emitted event.
- `EvaluationResult` and `EvaluatorFailure` gain an optional `evaluator_version` field, populated via `getattr` from the evaluator class so dashboards can filter out results from retired versions. `EvaluatorFailure` also gains an optional `error_type` field, populated with `type(e).__name__` where the failure is constructed from a caught exception.

## Dispatch and batching

Sinks shared across multiple evaluators for a single function call receive a single batched `submit()` carrying all of their results. The exact batching shape is an implementation detail that may change; each `EvaluationResult` / `EvaluatorFailure` carries enough metadata (name, `evaluator_version`, JSON-serialized `EvaluatorSpec`) to be attributed downstream. OTel event emission remains per-result.

## Event attribute schema

Standard OTel semconv: `gen_ai.evaluation.name`, `gen_ai.evaluation.score.value`, `gen_ai.evaluation.score.label`, `gen_ai.evaluation.explanation`, `error.type`.

pydantic-evals extensions in the `gen_ai.evaluation.*` namespace (may rename to align with future semconv additions): `gen_ai.evaluation.target`, `gen_ai.evaluation.evaluator.source`, `gen_ai.evaluation.evaluator.version`.

Bool scores emit both `score.value` (0.0/1.0) and `score.label` ("pass"/"fail") so numeric and categorical queries both work. `error.type` carries the actual caught exception class name (e.g. `'ValueError'`) when the `EvaluatorFailure` was built from a caught exception, falling back to the generic marker `'pydantic_evals.EvaluatorFailure'`.

## Breaking changes

`EvaluationSink.submit()` now takes a single `payload: SinkPayload` argument instead of the prior `(*, results, failures, context, span_reference)` kwargs. Custom sinks written against the old shape are wrapped transparently by a compatibility shim that unpacks the payload into the old kwargs call, emitting a one-time `DeprecationWarning` per sink class. The shim lives under a `TODO(v2)` block so it's easy to retire. Legacy-sink detection is arity-based so parameter naming doesn't matter — a modern sink whose parameter happens to be named `results` still works.

Additive changes (no shim needed):
- `EvaluationResult.evaluator_version`, `EvaluatorFailure.evaluator_version`, `EvaluatorFailure.error_type` — all new fields with defaults, appended to the dataclasses so positional construction stays compatible.
- `OnlineEvalConfig.emit_otel_events=True` default changes behavior (new OTel events flow), but no API shape change.
- `OnlineEvalConfig.include_baggage=True` — new default, adds baggage attrs to emitted events; can be disabled.
- Private symbols relocated from `pydantic_evals.dataset` to `pydantic_evals._task_run` — underscore-prefixed, fair game.

## Test plan

- [x] `uv run pytest tests/evals/` — 437 passed, including new tests for OTel emission, the `OnlineEvaluation` capability, batched sink dispatch, arity-based legacy-sink detection, and the `msg_template`/`span_name` decorator overrides
- [x] `prek` passes (format, lint, typecheck, codespell)
- [ ] Verify Live Evaluations UI picks up events from `gen_ai.evaluation.result` in a local backend run


DouweM (approved):

pydantic_evals/pydantic_evals/_otel_emit.py
```diff
@@ -0,0 +1,205 @@
...
+    extra_attributes: Mapping[str, Any] | None,
+) -> dict[str, Any]:
+    attrs: dict[str, Any] = {
+        _ATTR_TARGET: target,
+    }
+    if evaluator_version is not None:
+        attrs[_ATTR_EVALUATOR_VERSION] = evaluator_version
+    if extra_attributes:
+        attrs.update(extra_attributes)
+    return attrs
```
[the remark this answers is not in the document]
dmontagu: Good catch — fixed in 0cdf79aad. Flipped the order in `_base_attrs()` so `extra_attributes` apply first and the standard `target` / `evaluator_version` attrs (and every other `gen_ai.*` semconv attr) always win, consistent with how `_emit_result` and `_emit_failure` already handled the others. Added a regression test in `tests/evals/test_otel_emit.py::test_extra_attributes_cannot_override_standard_attrs`.

pydantic_evals/pydantic_evals/_online.py
```diff
@@ -0,0 +1,485 @@
...
+                    results,
+                    failures,
+                    context,
+                    span_reference,
+                    target,
+                    on_error,
+                    evaluator,
+                )
+    finally:
+        online_eval.semaphore.release()
```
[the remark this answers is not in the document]
dmontagu: Addressed in a4dda5b26 — `dispatch_evaluators` now short-circuits per-evaluator when both `emit_otel_events=False` and no sinks are resolved, skipping the evaluator entirely (so the semaphore slot and evaluator work aren't spent producing results that go nowhere). Documented the behaviour in `OnlineEvalConfig.emit_otel_events` and added a regression test (`test_dispatch_skipped_when_emit_off_and_no_sinks`).

nidhishgajjar:
## Summary
PR #5125 adds OpenTelemetry (OTel) event emission as the default observability surface for online evaluation, along with support for result grouping via `target` names and evaluator versioning. The PR introduces two new private modules (`_otel_emit.py` for OTel event emission and `_online.py` for refactored shared helpers), extends the `EvaluationSink.submit()` protocol with a `target` parameter, adds `emit_otel_events` and `include_baggage` config options, and documents agent integration via an `OnlineEvaluation` capability. This aligns online evals with the existing offline evals approach of using OTel for observability.

## Architecture
The changes extend the existing `pydantic_evals/online.py` infrastructure (1047 lines) by:

1. **`_otel_emit.py`** (new): Centralized OTel event emission following the [GenAI evaluation semconv](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-events/#event-gen_aievaluationresult). Emits one `gen_ai.evaluation.result` event per `EvaluationResult`/`EvaluatorFailure` with standard attributes (`gen_ai.evaluation.name`, `gen_ai.evaluation.score.value`, etc.) plus custom extensions (`gen_ai.evaluation.target`, `gen_ai.evaluation.evaluator_version`, `gen_ai.evaluation.evaluator_source`).

2. **`_online.py`** (new): Refactors shared private helpers from `online.py` including `dispatch_evaluators()`, `sample_evaluators()`, `_dispatch_single_evaluator()`, and the back-compat shim for `EvaluationSink.submit()` signature changes.

3. **`online.py`** (modified): Updates `EvaluationSink.submit()` to require `target: str`, adds `emit_otel_events` and `include_baggage` to `OnlineEvalConfig`, integrates `_otel_emit.emit_otel_events()` into the dispatch path, and adds legacy sink compatibility wrapping with deprecation warnings.

4. **Documentation** (`docs/evals/online-evaluation.md`): Comprehensive updates explaining default OTel emission, the `target` concept, evaluator versioning via `evaluator_version` class attribute, and agent integration.

The design maintains backward compatibility through a `_LegacyTargetShim` that wraps pre-`target` sinks and warns once per sink class.

## Issues

### Critical

**[critical]** pydantic_evals/_otel_emit.py (inferred) — The `evaluator_version` field is referenced throughout `_otel_emit.py` (e.g., `result.evaluator_version`, `failure.evaluator_version`) but this field does not exist on `EvaluationResult` or `EvaluatorFailure` in `pydantic_evals/evaluators/evaluator.py:59-105`. The PR must either:
- Add `evaluator_version: str | None = None` to both `EvaluationResult` and `EvaluatorFailure` dataclasses
- Populate this field during result construction in `_run_evaluator.py`
- **Fix**: Add the field to the dataclass definitions and update all construction sites


### Warnings

**[warning]** pydantic_evals/_otel_emit.py (inferred) — The `_format_score()` function has a typo in its docstring (line ~234): "`g` drops trailing zeros bu" should be "but".
- **Fix**: Correct the typo

**[warning]** pydantic_evals/online.py + pydantic_evals/_online.py — The `ContextVar` for evaluation disabled state may be duplicated between modules. The existing `online.py` has `_EVALUATION_DISABLED` at line 230, while `_online.py` defines `EVALUATION_DISABLED` (no underscore). If the PR refactors code to use `_online.py` internally, ensure only one `ContextVar` instance exists and both code paths reference it.
- **Fix**: Either remove the duplicate definition or clearly document which one is the canonical source

**[warning]** pydantic_evals/_online.py:345-391 — The deprecation warning for legacy sinks is only shown once per sink class (`_warned_legacy_sink_ids: set[int]`), but if a user creates multiple instances of a custom sink class before the first warning is emitted, they may miss the warning entirely in logs.
- **Fix**: Consider warning on first use per class or documenting this behavior clearly

**[warning]** pydantic_evals/online.py:645 — The condition `if not config.emit_otel_events and not sinks:` skips evaluators when OTel is disabled and no sinks are registered. However, this means evaluators with side effects (e.g., logging within `evaluate()`) won't run, which may be surprising.
- **Fix**: Document this behavior clearly or add a config option like `run_evaluators_without_sinks`

### Suggestions

**[suggestion]** docs/evals/online-evaluation.md — The "Agent Integration" section mentions `OnlineEvaluation` capability from `pydantic_evals.online_capability`, but this module is not shown in the PR diff. Verify that this module exists or will be added in a follow-up.
- **Fix**: Either include the `online_capability.py` implementation in this PR or add a TODO comment

**[suggestion]** pydantic_evals/_otel_emit.py (inferred) — The event body formatting in `_format_result_body()` and `_format_failure_body()` is human-readable but may not be machine-parseable. Consider adding structured fields if the backend supports it.
- **Fix**: Document the body format as informational only, or add structured attributes

**[suggestion]** pydantic_evals/_online.py:429-437 — The `wait_for_evaluations()` function joins threads with a timeout but doesn't return information about whether all threads completed successfully. Callers may want to know if the timeout was exceeded.
- **Fix**: Return a boolean indicating completion status or raise an exception if timeout is exceeded

**[suggestion]** pydantic_evals/online.py:277-293 — The `EvaluationSink` protocol docstring could benefit from an example showing how to handle the new `target` parameter in custom sink implementations.
- **Fix**: Add a code example to the docstring

**[suggestion]** pydantic_evals/_otel_emit.py (inferred) — The `build_parent_context()` function creates a `NonRecordingSpan` from trace/span IDs but doesn't validate that the IDs are properly formatted hex strings. Malformed input could cause `int(..., 16)` to fail.
- **Fix**: Add try-except around the int() conversions (already present at lines ~186-189, but ensure it's comprehensive)

## Cross-file impact

**Breakage**: The `EvaluationSink.submit()` signature change is breaking for existing custom sink implementations. The `_LegacyTargetShim` mitigates this but emits a deprecation warning. Projects with custom sinks will need to:
- Add `target: str` to their `submit()` method signature
- Update any type annotations
- Handle the `target` value appropriately

**Dependency**: This PR relies on `opentelemetry` and `opentelemetry._logs` being available. The project already has a `logfire` optional dependency which includes OTel SDK, but users who haven't installed it will get a no-op logger. This is safe but should be documented.

**Integration**: The agent integration via `OnlineEvaluation` capability is mentioned in documentation but the implementation (`pydantic_evals/online_capability.py`) is not shown in the diff. This must be implemented for the documentation to be accurate.

## Assessment
request-changes — The PR has a critical issue where `evaluator_version` is used but not defined on `EvaluationResult`/`EvaluatorFailure`. This must be fixed before merge.

---
_Automated review by [code-review-agent](https://github.com/AWLSEN/code-review-agent). May contain mistakes. Ignore or rebut as you see fit._

alexmojaki:
> OnlineEvalConfig.otel_extra_attributes — escape hatch for static tags on every emitted event.

This isn't there any more, PR description needs updating, might not be the only thing like that.

docs/evals/online-evaluation.md
```diff
@@ -15,7 +15,31 @@ For testing against curated datasets before deployment, use [offline evaluation]
...
+    def evaluate(self, ctx: EvaluatorContext) -> bool:
+        return bool(ctx.output)
+
+
+@evaluate(OutputNotEmpty())
+async def summarize(text: str) -> str:
+    return f'Summary of: {text}'
+```
+
+Wire up OTel export (e.g. `logfire.configure()`) elsewhere in your application startup so that the emitted `gen_ai.evaluation.result` events reach your backend.
```
alexmojaki: logfire.configure() should link to something

docs/evals/online-evaluation.md
```diff
@@ -15,7 +15,31 @@ For testing against curated datasets before deployment, use [offline evaluation]
...
+
+Each decorated call emits one `gen_ai.evaluation.result` OTel event per evaluator result, following the [OTel GenAI evaluation semconv](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-events/#event-gen_aievaluationresult). This mirrors how offline evaluation emits OTel spans via `logfire_span`: if any OTel SDK is configured in the process (via `logfire.configure()`, the OTel SDK directly, or a vendor instrumentation), events flow to your backend; if not, emission is a cheap no-op.
```
alexmojaki: `logfire_span` is internal, should probably be `logfire.span`

docs/evals/online-evaluation.md
```diff
@@ -858,6 +944,47 @@ Key behaviors:
 - **If `on_error` itself raises**, the exception is silently suppressed to protect sibling evaluators.
 - **If no `on_error` is set**, exceptions are silently suppressed — this is the safe default.
 
+## Agent Integration
+
+The [`OnlineEvaluation`][pydantic_evals.online_capability.OnlineEvaluation] capability brings online evaluation to Pydantic AI agents. Instead of decorating a function, you add the capability to your agent. As with the `@evaluate` decorator, evaluators dispatch in the background and results are emitted as OTel events by default — no sink registration required:
```
alexmojaki, line 914: This link isn't rendering correctly in the preview, not sure if deployed docs will be better. presumably not since other links seem fine.

docs/evals/online-evaluation.md
```diff
@@ -54,6 +78,57 @@ asyncio.run(main())
...
+
+@dataclass
+class ToneCheck(Evaluator):
+    evaluator_version = 'v2'  # bumped after prompt rewrite
+
+    def evaluate(self, ctx: EvaluatorContext) -> str:
+        return 'neutral'
+```
+
+The version applies to all results the evaluator produces (so one evaluator class maps to one version, even when the evaluator returns a mapping of named results). Framework code reads the attribute via `getattr` — it's an optional convention on the base class, not a required field.
```
alexmojaki: "Framework code reads the attribute via getattr" is a weird thing to say, presumably AI.

docs/evals/online-evaluation.md
```diff
@@ -175,28 +250,39 @@ class PrintSink:
         failures: Sequence[EvaluatorFailure],
         context: EvaluatorContext,
         span_reference: SpanReference | None,
+        target: str,
```
alexmojaki: If we're breaking the signature, now is the time to wrap all these args in a single container object as previously discussed

docs/evals/online-evaluation.md
```diff
@@ -175,28 +250,39 @@ class PrintSink:
...
 
-    If you need side effects produced _within_ an evaluator itself, configure a no-op sink
-    to ensure the evaluator runs:
+### Default OTel event emission
 
-    ```python
-    from pydantic_evals.online import OnlineEvalConfig
+Every dispatched evaluator emits one `gen_ai.evaluation.result` OTel log event per [`EvaluationResult`][pydantic_evals.evaluators.EvaluationResult] or [`EvaluatorFailure`][pydantic_evals.evaluators.EvaluatorFailure], unconditionally — no sink registration required. Events are parented to the span that produced them (when a `span_reference` is available), so they appear nested under the original function call in the trace. If no OTel SDK is configured in the process, emission is a cheap no-op.
```
alexmojaki: > (when a `span_reference` is available)

is it sometimes not?

docs/evals/online-evaluation.md
```diff
@@ -15,7 +15,31 @@ For testing against curated datasets before deployment, use [offline evaluation]
 
 ## Quick Start
 
-The [`evaluate()`][pydantic_evals.online.evaluate] decorator attaches evaluators to any function. Evaluators run in the background without blocking the caller:
+The [`evaluate()`][pydantic_evals.online.evaluate] decorator attaches evaluators to any function. Evaluators run in the background without blocking the caller, and results are emitted as OpenTelemetry events:
```
alexmojaki: maybe link to the "Default OTel event emission" section below

docs/evals/online-evaluation.md
```diff
@@ -15,7 +15,31 @@ For testing against curated datasets before deployment, use [offline evaluation]
...
+@evaluate(OutputNotEmpty())
+async def summarize(text: str) -> str:
+    return f'Summary of: {text}'
+```
+
+Wire up OTel export (e.g. `logfire.configure()`) elsewhere in your application startup so that the emitted `gen_ai.evaluation.result` events reach your backend.
+
+Each decorated call emits one `gen_ai.evaluation.result` OTel event per evaluator result, following the [OTel GenAI evaluation semconv](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-events/#event-gen_aievaluationresult). This mirrors how offline evaluation emits OTel spans via `logfire_span`: if any OTel SDK is configured in the process (via `logfire.configure()`, the OTel SDK directly, or a vendor instrumentation), events flow to your backend; if not, emission is a cheap no-op.
+
+To additionally handle results in Python code — for alerting, bespoke aggregation, in-memory test capture, or non-OTel destinations — register a [sink](#sinks). Sinks run *in addition to* OTel event emission:
```
alexmojaki: are sinks still important enough to get a big code snippet in the quickstart?

docs/evals/online-evaluation.md
```diff
@@ -54,6 +78,57 @@ asyncio.run(main())
...
+# Override: use a friendly name
+@evaluate(OutputNotEmpty(), target='customer_support')
+async def run_agent(prompt: str) -> str: ...
+```
+
+The target name is supplied to sinks on every `submit()` call as a plain `str` — a single sink instance handles any number of decorated functions or agents.
+
+For agent capabilities, the target name is taken from the agent's own `name` attribute (see [Agent Integration](#agent-integration)); to categorize or route on agent-ness, add metadata on the config (e.g., `metadata={'kind': 'agent'}`).
+
+## Evaluator Versioning
```
alexmojaki: Are this and Target important enough to appear before 'Core Concepts'?

docs/evals/online-evaluation.md
```diff
@@ -175,28 +250,39 @@ class PrintSink:
...
 
-    ```python
-    from pydantic_evals.online import OnlineEvalConfig
+Every dispatched evaluator emits one `gen_ai.evaluation.result` OTel log event per [`EvaluationResult`][pydantic_evals.evaluators.EvaluationResult] or [`EvaluatorFailure`][pydantic_evals.evaluators.EvaluatorFailure], unconditionally — no sink registration required. Events are parented to the span that produced them (when a `span_reference` is available), so they appear nested under the original function call in the trace. If no OTel SDK is configured in the process, emission is a cheap no-op.
 
-    config = OnlineEvalConfig(default_sink=lambda results, failures, context: None)
-    ```
+The emission follows the [OpenTelemetry GenAI evaluation semconv](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-events/#event-gen_aievaluationresult). Key attributes:
+
+- `gen_ai.evaluation.name`, `gen_ai.evaluation.score.value`, `gen_ai.evaluation.score.label`, `gen_ai.evaluation.explanation` — standard semconv attributes.
```
alexmojaki: Should still explain what they are, it's really not obvious how exactly these get populated.

Ideally there should be an example, e.g. a code snippet with comments and/or a table of attributes so the user can see what goes where. maybe an embedded public trace.

docs/evals/online-evaluation.md
```diff
@@ -175,28 +250,39 @@ class PrintSink:
...
+- `error.type` — set to `pydantic_evals.EvaluatorFailure` when an evaluator raised.
+- `gen_ai.evaluation.target` — target name from the `@evaluate` decorator or agent (see [Target](#target)).
+- `gen_ai.evaluation.evaluator_version` — set when an [evaluator declares a version](#evaluator-versioning).
```
alexmojaki: Maybe these should be `gen_ai.evaluation.evaluator.version` and `gen_ai.evaluation.evaluator.source`, OTel likes namespacing.

docs/evals/online-evaluation.md
```diff
@@ -175,28 +250,39 @@ class PrintSink:
...
+- `gen_ai.evaluation.evaluator_source` — JSON-serialized [`EvaluatorSpec`][pydantic_evals.evaluators.evaluator.EvaluatorSpec] for the evaluator that produced the result.
+
+These `gen_ai.evaluation.*` extensions beyond the current semconv follow the same namespace on the assumption that upstream will land analogous attributes here; if OTel adopts different names we'll realign.
```
alexmojaki: is this line needed? also are we making any upstream proposal?

pydantic_evals/pydantic_evals/evaluators/_run_evaluator.py
```diff
@@ -59,33 +59,46 @@ async def run_evaluator(
...
+
+    evaluator_name = evaluator.get_default_evaluation_name()
+    source = evaluator.as_spec()
+
     try:
-        with logfire_span(
-            'evaluator: {evaluator_name}',
-            evaluator_name=evaluator.get_default_evaluation_name(),
-        ):
+        with logfire_span('Calling evaluator: {evaluator_name}', evaluator_name=evaluator_name):
```
alexmojaki: this is an old span name that queries might be using, let's not change it

alexmojaki: Why did you drop the attributes in https://github.com/pydantic/pydantic-ai/commit/5243f78a54a0b0de1cd5349a86f087c7846b9d21?

pydantic_evals/pydantic_evals/evaluators/evaluator.py
```diff
@@ -135,6 +143,31 @@ class ExactMatch(Evaluator):
         def evaluate(self, ctx: EvaluatorContext) -> bool:
             return ctx.output == ctx.expected_output
     ```
+
+    Optional class-level attributes (read via `getattr`, not declared on the base to avoid
+    dataclass-field/ClassVar conflicts in subclasses):
```
alexmojaki: i'm not understanding what the conflict is, e.g. this works:

```python
from dataclasses import dataclass
from typing import ClassVar



@dataclass
class Evaluator:
    evaluator_version: ClassVar[str | None] = None

@dataclass
class ToneCheck(Evaluator):
    evaluator_version = 'v2'

assert Evaluator.evaluator_version == None
assert ToneCheck.evaluator_version == 'v2'
```

pydantic_evals/pydantic_evals/evaluators/_run_evaluator.py
```diff
@@ -59,33 +59,46 @@ async def run_evaluator(
...
                 raise ValueError(f'{evaluator!r}.evaluate returned a value of an invalid type: {raw_results!r}.') from e
+
+            results = _convert_to_mapping(results, scalar_name=evaluator_name)
+
+            details: list[EvaluationResult] = []
+            for name, result in results.items():
+                if not isinstance(result, EvaluationReason):
+                    result = EvaluationReason(value=result)
+                details.append(
+                    EvaluationResult(
```
alexmojaki, line 95: Not sure if this is a time to change it, but why doesn't this get richer attributes like the actual `result` and `evaluator`?

pydantic_evals/pydantic_evals/_otel_emit.py
```diff
@@ -0,0 +1,248 @@
...
+
+
+def _emit_failure(
+    failure: EvaluatorFailure,
+    target: str,
+    extra_attributes: Mapping[str, Any] | None,
+) -> None:
+    attrs = _base_attrs(target, failure.evaluator_version, extra_attributes)
+    attrs[_ATTR_EVAL_NAME] = failure.name
+    attrs[_ATTR_ERROR_TYPE] = 'pydantic_evals.EvaluatorFailure'
```
alexmojaki: This should be the actual exception type that was raised

pydantic_evals/pydantic_evals/_online.py
```diff
@@ -0,0 +1,502 @@
...
+OnMaxConcurrencyCallback = Callable[[EvaluatorContext], None | Awaitable[None]]
+OnErrorCallback = Callable[
+    [Exception, EvaluatorContext, Evaluator, OnErrorLocation],
+    None | Awaitable[None],
+]
+SamplingContextBuilder = Callable[[Evaluator, Any, dict[str, Any] | None, float], Any]
+
+
+@runtime_checkable
+class EvaluationSink(Protocol):
```
alexmojaki, line 74: Why are there two EvaluationSink classes now?

pydantic_evals/pydantic_evals/_online.py
```diff
@@ -0,0 +1,502 @@
...
+    enabled: bool
+    emit_otel_events: bool
+    include_baggage: bool
+    metadata: dict[str, Any] | None
+    on_max_concurrency: OnMaxConcurrencyCallback | None
+    on_sampling_error: OnSamplingErrorCallback | None
+    on_error: OnErrorCallback | None
+
+
+class _CallbackSink:
```
alexmojaki: Why do we have both this and CallbackSink?

pydantic_evals/pydantic_evals/_online.py
```diff
@@ -0,0 +1,502 @@
...
+    async def submit(
+        self,
+        *,
+        results: Sequence[EvaluationResult],
+        failures: Sequence[EvaluatorFailure],
+        context: EvaluatorContext,
+        span_reference: Any | None,
+        target: str,
+    ) -> None:
+        _ = span_reference, target  # custom sink only
```
alexmojaki: Why?

pydantic_evals/pydantic_evals/_online.py
```diff
@@ -0,0 +1,502 @@
...
+    sample_rate: float | Callable[[Any], float | bool] | None
+    max_concurrency: int
+    semaphore: threading.Semaphore
+    sink: Any
+    on_max_concurrency: OnMaxConcurrencyCallback | None
+    on_sampling_error: OnSamplingErrorCallback | None
+    on_error: OnErrorCallback | None
+
+
+class OnlineEvalConfigLike(Protocol):
```
alexmojaki: Why these 'Like' protocols?

pydantic_evals/pydantic_evals/_online.py
```diff
@@ -0,0 +1,502 @@
...
+def _should_evaluate(
+    rate: float | Callable[[Any], float | bool],
+    global_enabled: bool,
+    sampling_context: Any,
+    sampling_mode: SamplingMode,
+) -> bool:
+    if not global_enabled:  # pragma: no cover
+        return False
+    if EVALUATION_DISABLED.get():  # pragma: no cover
+        return False
```
alexmojaki: shouldn't these be covered?

pydantic_evals/pydantic_evals/_otel_emit.py
```diff
@@ -0,0 +1,248 @@
...
+        return None
+    merged: dict[str, Any] = {str(k): v for k, v in bag.items()}
+    if extra_attributes:
+        merged.update(extra_attributes)
+    return merged
+
+
+def build_parent_context(span_reference: Any | None) -> Context | None:
+    """Public: build an OTel context with a non-recording parent span, or None."""
+    return _build_parent_context(span_reference)
```
alexmojaki: why the indirection? why is this public?

pydantic_evals/pydantic_evals/_online.py
```diff
@@ -0,0 +1,502 @@
...
+            target=target,
+        )
+    except Exception as exc:
+        await _call_on_error(on_error, exc, context, evaluator, 'sink')
+
+
+async def _dispatch_single_evaluator(
+    online_eval: OnlineEvaluatorLike,
+    context: EvaluatorContext,
+    span_reference: Any | None,
```
alexmojaki: Any?

pydantic_evals/pydantic_evals/_online.py
```diff
@@ -0,0 +1,502 @@
...
+            try:
+                result = on_max_concurrency(context)
+                if inspect.isawaitable(result):
+                    await result
+            except Exception as exc:
+                await _call_on_error(on_error, exc, context, evaluator, 'on_max_concurrency')
+        return
+
+    # Attach the call's span as the current OTel parent for the whole evaluator
+    # run. This nests the `evaluator: {name}` span (created inside `run_evaluator`)
```
alexmojaki: span name has currently changed, though i objected in another PR comment. if the change stays, update this comment.

pydantic_evals/pydantic_evals/_otel_emit.py
```diff
@@ -0,0 +1,248 @@
...
+    target: str,
+    extra_attributes: Mapping[str, Any] | None,
+) -> None:
+    attrs = _base_attrs(target, failure.evaluator_version, extra_attributes)
+    attrs[_ATTR_EVAL_NAME] = failure.name
+    attrs[_ATTR_ERROR_TYPE] = 'pydantic_evals.EvaluatorFailure'
+    if failure.error_message:
+        attrs[_ATTR_EXPLANATION] = failure.error_message
+    attrs[_ATTR_EVALUATOR_SOURCE] = _serialize_evaluator_source(failure.source)
+    _get_logger().emit(LogRecord(event_name=_EVENT_NAME, body=_format_failure_body(failure), attributes=attrs))
```
alexmojaki: i think set the severity to warning

pydantic_evals/pydantic_evals/_otel_emit.py
```diff
@@ -0,0 +1,248 @@
...
+    result: EvaluationResult,
+    target: str,
+    extra_attributes: Mapping[str, Any] | None,
+) -> None:
+    attrs = _base_attrs(target, result.evaluator_version, extra_attributes)
+    attrs[_ATTR_EVAL_NAME] = result.name
+    _set_score_attrs(attrs, result.value)
+    if result.reason is not None:
+        attrs[_ATTR_EXPLANATION] = result.reason
+    attrs[_ATTR_EVALUATOR_SOURCE] = _serialize_evaluator_source(result.source)
```
alexmojaki: should be in _base_attrs

pydantic_evals/pydantic_evals/_otel_emit.py
```diff
@@ -0,0 +1,248 @@
...
+    return _logger
+
+
+def emit_otel_events(
+    *,
+    results: Sequence[EvaluationResult],
+    failures: Sequence[EvaluatorFailure],
+    target: str,
+    include_baggage: bool = True,
+    extra_attributes: Mapping[str, Any] | None = None,
```
alexmojaki: this is no longer used outside of tests, are we keeping it to make it easy to add the API in future?

alexmojaki (changes requested):

pydantic_evals/pydantic_evals/online_capability.py
alexmojaki, line 1: consider splitting into another PR on top

pydantic_evals/pydantic_evals/online.py
```diff
@@ -281,23 +204,33 @@ async def submit(
         failures: Sequence[EvaluatorFailure],
         context: EvaluatorContext,
         span_reference: SpanReference | None,
+        target: str,
     ) -> None:
         """Submit evaluation results to the sink.
 
+        Each `submit()` call corresponds to a single evaluator — `results` and
```
alexmojaki: Why are we making this a contract?

tests/evals/test_online.py
```diff
@@ -1811,3 +1997,321 @@ async def my_func(x: int) -> int:
...
+    assert call_spans[0]['attributes']['return'] == 42
+
+
+@needs_logfire
+@pytest.mark.anyio
+async def test_call_span_msg_template_and_span_name(capfire: CaptureLogfire):
+    """`msg_template`/`span_name` override the default span name."""
+    config = OnlineEvalConfig(emit_otel_events=False)
+
+    @config.evaluate(AlwaysTrue(), msg_template='running my task', span_name='task.run')
```
alexmojaki: msg_template not tested

docs/evals/online-evaluation.md
dmontagu: Alex says maybe link to pydantic-ai docs page describing how to use logfire

docs/evals/online-evaluation.md
dmontagu: maybe need to update the api docs page contents to add references to new sections/etc.? not sure

docs/evals/online-evaluation.md
dmontagu: I'll rephrase in my own words

docs/evals/online-evaluation.md
dmontagu: yes, let's retain backwards compatibility as I did here, but make this more drastic change to the signature. Let's also update the docstring of the new container object struct to say that we reserve the right to add additional fields, and people should not instantiate the class directly. (Maybe it would make sense to have a private constructor? Maybe that's overkill..?)

docs/evals/online-evaluation.md
dmontagu: if there's no interesting case where this is the case (i.e., if it happens whenever otel is installed) then we can drop this parenthetical. I need to look deeper.

docs/evals/online-evaluation.md
dmontagu: Probably not in the quickstart. Let's move to a separate section and link to it from the quickstart. (Or just link to the section if it already exists.)

docs/evals/online-evaluation.md
dmontagu: I'll look into it.

docs/evals/online-evaluation.md
dmontagu: I think it's worth saying something, and I don't want to make an upstream proposal until we've had it around for a little while and feel like it's good. I don't expect source will be accepted (but maybe?). Version, I'd rather make sure we like it for a little while.

pydantic_evals/pydantic_evals/evaluators/_run_evaluator.py
dmontagu: I'd like to change it for online evals to make it more intelligible, but I agree for offline it shouldn't be changed.

dmontagu: Not sure I meant to, I will look into this.

dmontagu: Alex suggests that we should set the message but not change the span name.

alexmojaki:
```suggestion
        with logfire_span('Calling evaluator: {evaluator_name}', evaluator_name=evaluator_name, _span_name='evaluator: {evaluator_name}'):
```

pydantic_evals/pydantic_evals/evaluators/evaluator.py
dmontagu: the conflict is with method names, but I'm not convinced that's worth worrying about, we could probably just do the classvar thing

dmontagu: I'm inclined to not make this change now, we could better document it through type hints in the future, I'm just not sure if there are any potential backwards compatibility consequences. We should make a note to move it to a typehint in v2 though.

pydantic_evals/pydantic_evals/evaluators/_run_evaluator.py
dmontagu: To guarantee serializability

pydantic_evals/pydantic_evals/_online.py
dmontagu: there probably shouldn't be

pydantic_evals/pydantic_evals/_otel_emit.py
dmontagu: Alex says we might need to add a field to EvaluatorFailure to track this.

pydantic_evals/pydantic_evals/_online.py
dmontagu: Not sure both are necessary, let me reinvestigate

pydantic_evals/pydantic_evals/_online.py
dmontagu: good question, need to look into it.

pydantic_evals/pydantic_evals/_online.py
dmontagu: probably AI coding agents being stupid, I'll remove them. Should just use OnlineEvalConfig 😩

pydantic_evals/pydantic_evals/_online.py
dmontagu: yes

pydantic_evals/pydantic_evals/_otel_emit.py
dmontagu: because dumb. Let's remove it.

pydantic_evals/pydantic_evals/_online.py
dmontagu: Let's remove the use of `Any` here and use a specific type.

pydantic_evals/pydantic_evals/_otel_emit.py
dmontagu: should probably be dropped, I think it got missed when I removed `otel_extra_attributes` or whatever it was called.

pydantic_evals/pydantic_evals/online_capability.py
dmontagu: I'd rather not because I want to support both patterns on day one.

pydantic_evals/pydantic_evals/online.py
dmontagu: Not sure, I'll check if we need to.

dmontagu: This seems dumb, I feel like we should not have one call per evaluator, we should call once per target execution with all results from all evaluators. This seems like a misimplementation by AI, I'll investigate more thoroughly.

tests/evals/test_online.py
dmontagu: I guess we should add an explicit test of this (or an assertion within the same test anyway) that the arguments to the function call can be used here.

docs/evals/online-evaluation.md
dmontagu: Done in 76372cd. `EvaluationSink.submit` now takes a single `SinkPayload` container; its docstring flags it as not-for-direct-instantiation and reserves the right to add fields. Back-compat uses signature introspection: sinks whose `submit()` still takes the old kwargs (`results=`, etc.) get wrapped in a shim that unpacks the payload and emits a one-time `DeprecationWarning` per class (targeted for removal in v2). Since the earlier `target` kwarg commit hadn't shipped on main, the shim only needs to cover the original four-kwarg form, so the `accepts_target` branch was dropped.

pydantic_evals/pydantic_evals/online.py
dmontagu: Addressed in 76372cd — we no longer contract one-call-per-evaluator. Dispatch now groups evaluators by sink-source identity (default_sink vs. explicit per-evaluator `sink=`) and fires one batched `submit()` call per sink per function execution carrying every sharing evaluator's results. OTel emission stays per-result so each eval still has its own event; single-evaluator groups behave exactly as before (including per-evaluator `on_error` routing). Multi-evaluator groups deduplicate `on_error` handlers by identity so a shared `FailingSink` won't fan out to the same handler N times. Docs and the `EvaluationSink.submit` docstring now describe the payload as "results from one or more evaluators" with the exact batching called out as an implementation detail that may change.

docs/evals/online-evaluation.md
dmontagu: Nevermind I think we can just drop this. It's still true that we could probably add it as a class attribute but for now I think we can just keep things as-is and say less.

docs/evals/online-evaluation.md
dmontagu: Dug in — in practice `span_reference` is `None` in exactly one circumstance: when there's no OTel SDK configured in the process (neither `logfire.configure()` nor a user-installed `TracerProvider`). Both `trace.get_tracer(...)` and unconfigured `logfire.span(...)` fall back to the no-op tracer, which returns `INVALID_SPAN_CONTEXT` (`trace_id=0`, `span_id=0`), and `_extract_span_reference` filters those out to avoid emitting bogus IDs. (Two additional `pragma: no cover` defensive branches handle spans missing `get_span_context` or raising from it — only reachable with a pathological custom span type.)

In that same "no OTel SDK" case, event emission itself is a no-op (no `LoggerProvider`), so there's nothing to parent either way. The only scenario where events-would-be-emitted-but-unparented is a pathological partial setup (configured `LoggerProvider` but no `TracerProvider`), which isn't worth documenting. Dropped the parenthetical in c3e2055.

docs/evals/online-evaluation.md
dmontagu: Done in fdf5ecc. Dropped the runnable sink snippet from the Quick Start — it now ends with a sentence pointer to the [Sinks](#sinks) section and a short note about module-level `configure()`/`evaluate()`. Sinks get full coverage downstream: a runnable callable-sink example in the OnlineEvalConfig section and a `PrintSink` protocol example in the Sinks section.

docs/evals/online-evaluation.md
dmontagu: I think target is worth keeping here, but evaluator_version is not.

dmontagu: Done in 63bb1d6. Kept **Target** as a top-level section above Core Concepts (it's on every payload/event, routes results across functions/agents, and defaults silently from `__name__` — worth the prominence). Demoted **Evaluator Versioning** to a subsection under _Default OTel event emission_ — it's an optional class attribute read via `getattr` whose only effect is stamping one attribute, so it belongs next to its effect rather than on the on-ramp. The `#evaluator-versioning` anchor is preserved, so existing cross-references still resolve.

docs/evals/online-evaluation.md
dmontagu: Done in 97cd5d3. Replaced the bullet list with an attribute table that sources each attribute to the underlying `EvaluationResult` / `EvaluatorFailure` / `Evaluator` field it comes from, with notes on:

- how `gen_ai.evaluation.name` comes from the result key (scalar returns → class name; mapping returns → dict keys),
- the bool/numeric/string split between `score.value` and `score.label` (`bool` → both, numeric → value, string → label),
- which attributes are omitted when absent vs. always present,
- why `evaluator_source` exists alongside `evaluator_name` (distinguishing two `LLMJudge(rubric=...)` instances that share a name),
- baggage propagation and conflict rules.

Plus a worked `ToneCheck(evaluator_version='v2')` example that shows exactly which attribute values land on an emitted event (verified against `EvaluatorSpec.model_dump_json()` locally). Happy to add an embedded public trace too if there's a Logfire sandbox project I should point it at.

docs/evals/online-evaluation.md
dmontagu: Agreed — done in 51b1696. Renamed to `gen_ai.evaluation.evaluator.version` and `gen_ai.evaluation.evaluator.source` to match OTel's dotted-namespace preference (and to match the existing `score.value` / `score.label` shape under `gen_ai.evaluation.*`). Low-risk since nothing has shipped on main, and leaves a clean slot for future `evaluator.*` additions.

docs/evals/online-evaluation.md
dmontagu: Tightened the sentence in 51b1696. It now just flags the attrs as pydantic-evals extensions that may rename to track future OTel semconv — drops the internal "we'll realign" framing without losing the stability signal users actually care about. Leaves the upstream-proposal question open for when we've lived with the names longer.

pydantic_evals/pydantic_evals/evaluators/_run_evaluator.py
dmontagu: Applied in 877e2558. The `_span_name='evaluator: {evaluator_name}'` keyword isn't template-interpolated by logfire (the literal string is used as the logical span name), which matches the pre-PR behavior exactly since `logfire.span('evaluator: {evaluator_name}', ...)` on main also recorded the literal template string as the span name. Existing queries filtering on that span name continue to match; only the user-facing `logfire.msg` now reads `"Calling evaluator: SimpleEvaluator"` to match the new span-message convention.

Updated two test snapshot assertions that had drifted to the "Calling …" form to expect the stable span name again.

pydantic_evals/pydantic_evals/_online.py
dmontagu: Consolidated in 938375f. The duplication was historical: at initial PR time, `_online.py` needed its own `EvaluationSink` Protocol because `online.py` imported from it (circular). Now that `SinkPayload` already lives in `_online.py` and is re-exported from `online.py`, `EvaluationSink` follows the same pattern — one canonical definition in `_online.py` (with the public docstring), re-exported from `online.py`. No more duplicate.

pydantic_evals/pydantic_evals/_online.py
dmontagu: Same consolidation in 938375f. `_CallbackSink` is gone; there's now a single `CallbackSink` class defined in `_online.py` (with the public docstring) and re-exported from `online.py` — identical implementation on both sides, no reason to keep them separate.

pydantic_evals/pydantic_evals/_online.py
```diff
@@ -0,0 +1,583 @@
...
+    has_positional = any(
+        p.kind
+        in (
+            inspect.Parameter.POSITIONAL_ONLY,
+            inspect.Parameter.POSITIONAL_OR_KEYWORD,
+            inspect.Parameter.VAR_POSITIONAL,
+        )
+        for p in params.values()
+    )
+    return not (has_positional and 'results' not in params)
```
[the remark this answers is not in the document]
dmontagu: The legacy sink detection should just check the number of args in the signature. If it's one (or two given it's a method, or whatever it needs to be for the new signature), then use the new signature with the SinkPayload type. Otherwise, use the old signature.

pydantic_evals/pydantic_evals/evaluators/evaluator.py
dmontagu: Also let's delete the discussion of this from the docs that instigated this whole thread.

docs/evals/online-evaluation.md
dmontagu: Fixed in 2bba158 — swapped the internal `logfire_span` helper reference for the public `logfire.span` API in the Quick Start prose.

docs/evals/online-evaluation.md
dmontagu: Linked in 2bba158 — "emitted as OpenTelemetry events" in the Quick Start intro now points to the [Default OTel event emission](#default-otel-event-emission) section.

docs/evals/online-evaluation.md
dmontagu: Done in 2bba158 — both mentions of `logfire.configure()` in the Quick Start now link to [Using Logfire](../logfire.md#using-logfire) (the setup section of the Pydantic AI logfire page).

docs/evals/online-evaluation.md
dmontagu: Diagnosed — the `OnlineEvaluation` link wasn't rendering because `pydantic_evals.online_capability` had no generated API docs page (only `pydantic_evals.online` did). Added `docs/api/pydantic_evals/online_capability.md` and wired it into the mkdocs nav in 2bba158, so the cross-reference now resolves.

The main `pydantic_evals.online` API page already picks up the new `SinkPayload` export (it comes through `__all__`), so no changes needed there.

pydantic_evals/pydantic_evals/evaluators/_run_evaluator.py
dmontagu: ah okay, so they were added in a different commit and then removed. The reason I removed them was because they needed to be quite different in cases where there were multiple evaluators vs. a single. I think we could readd them at any time but I wasn't confident enough to start with them there. We can discuss tomorrow.

pydantic_evals/pydantic_evals/evaluators/evaluator.py
dmontagu: I decided to keep some but say less about the justification. We can revisit the details here in the future.

dmontagu: Done in 68617cd. Kept the class-attribute discussion in the `Evaluator` docstring but dropped the "conflicts in subclasses" justification (since it wasn't actually true per your demo — ClassVar does work). Added `TODO(v2)` comments at both `getattr` sites (`evaluator_version` in `_run_evaluator.py`, `evaluation_name` in `evaluator.py`) to declare them as `ClassVar[str | None] = None` on the base in v2 and read directly.

pydantic_evals/pydantic_evals/_online.py
dmontagu: Already gone — the SinkPayload refactor in 76372cd removed `_CallbackSink` entirely (consolidated into a single `CallbackSink` in `_online.py` that takes a `SinkPayload` directly), so the throwaway `_ = span_reference, target  # custom sink only` line no longer exists. The new `submit` body just reads `payload.results`, `payload.failures`, `payload.context` and doesn't see the ignored fields at all.

pydantic_evals/pydantic_evals/_otel_emit.py
dmontagu: Removed in 9b4179a. The private `_build_parent_context` body is now inlined into the public `build_parent_context` — the wrapper was a pass-through, so the hop wasn't doing anything.

pydantic_evals/pydantic_evals/_online.py
dmontagu: in particular, we should remove any `span_reference: Any` (or `span_reference: Any | None`) occurrences throughout the codebase, and replace with a concrete type. We should NOT use `Any` for this.

pydantic_evals/pydantic_evals/_otel_emit.py
dmontagu: (The point is that this code can probably be more DRY.)

pydantic_evals/pydantic_evals/_online.py
dmontagu: turns out it was a workaround for circular imports, better solved with `if TYPE_CHECKING:` stuff. 😩

dmontagu: Removed in a8bb5a1. Both `OnlineEvaluatorLike` and `OnlineEvalConfigLike` are gone; `_online.py` now uses a `TYPE_CHECKING` import of the concrete `OnlineEvaluator` / `OnlineEvalConfig` from `online.py` (safe because `_online.py` has `from __future__ import annotations`, so parameter annotations don't need the symbols at runtime). Also dropped a now-redundant `cast(EvaluationSink | Sequence[…] | SinkCallback, raw)` in `dispatch_evaluators`, since `raw` is now properly typed through the concrete `sink` / `default_sink` fields. No other `XLike` protocols in the diff — these were the only ones.

pydantic_evals/pydantic_evals/_otel_emit.py
dmontagu: Dropped in 0a385ca. `extra_attributes` had no production callers — just the one test — so I removed the kwarg, the `_collect_extra_attributes` helper, and the two tests that exercised it. Baggage remains as the extension hook via `include_baggage`.

pydantic_evals/pydantic_evals/_otel_emit.py
dmontagu: Done in 0a385ca — `_emit_failure` now passes `severity_number=SeverityNumber.WARN` on the `LogRecord`, so log viewers can filter/highlight failure events. Added an assertion to `test_failure_emits_error_type_and_no_score` to lock it in.

pydantic_evals/pydantic_evals/_otel_emit.py
dmontagu: Done in 0a385ca. Moved both `gen_ai.evaluation.name` and `gen_ai.evaluation.evaluator.source` into `_base_attrs` so `_emit_result` and `_emit_failure` no longer duplicate those three writes each. They now only set what's genuinely per-event grain: score attrs + `explanation` for results, `error.type` + `explanation` for failures.

pydantic_evals/pydantic_evals/_online.py
dmontagu: Tightened in 0a385ca — the comment now reads `evaluator: {evaluator_name}` (matching the template string exactly) instead of `evaluator: {name}`. The logical span name itself is still `evaluator: {evaluator_name}` thanks to the `_span_name=...` override restored in 877e2558, so the reference is accurate.

pydantic_evals/pydantic_evals/_online.py
dmontagu: Done in 0a385ca. All three `span_reference: Any | None` occurrences (`_online.py` had two, `_otel_emit.py` had one on `build_parent_context`) now say `SpanReference | None`. `_otel_emit.py` needed a `TYPE_CHECKING` import of the concrete class (same pattern as `_online.py`) — with `from __future__ import annotations`, annotations don't need to resolve at runtime, so no circular-import issue. Dropped the now-impossible `AttributeError`/`TypeError` branches from `build_parent_context` since `SpanReference` guarantees `str` fields; left the `ValueError` branch for malformed-hex defense.

pydantic_evals/pydantic_evals/_otel_emit.py
dmontagu: This has been addressed.

docs/evals/online-evaluation.md
```diff
@@ -146,57 +166,91 @@ async def main():
...
+
+```python
+from pydantic_evals.online import OnlineEvalConfig
+
+config = OnlineEvalConfig(emit_otel_events=False)
+```
+
+#### Evaluator Versioning
+
+Set `evaluator_version` as a class attribute on an [`Evaluator`][pydantic_evals.evaluators.Evaluator] subclass to stamp every result it emits with a version string — surfaced as `gen_ai.evaluation.evaluator_version` on emitted events and as `evaluator_version` on each [`EvaluationResult`][pydantic_evals.evaluators.EvaluationResult] and [`EvaluatorFailure`][pydantic_evals.evaluators.EvaluatorFailure]. This lets trend lines and dashboards filter out results produced by retired evaluator versions without deleting historical rows — useful when you change an LLM judge's prompt or rework a heuristic in a way that invalidates prior scores:
```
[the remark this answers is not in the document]
dmontagu: Fixed in f16e827. The Evaluator Versioning prose now reads `gen_ai.evaluation.evaluator.version` to match the actual emission constant and the attribute table above.

docs/evals/online-evaluation.md
```diff
@@ -146,57 +166,91 @@ async def main():
...
+
+| Attribute | Source | Notes |
+|---|---|---|
+| `gen_ai.evaluation.name` | [`EvaluationResult.name`][pydantic_evals.evaluators.EvaluationResult] / [`EvaluatorFailure.name`][pydantic_evals.evaluators.EvaluatorFailure] | The evaluator class name when the `evaluate()` method returns a scalar, or the mapping key when it returns `{'accuracy': ..., 'score': ...}`. |
+| `gen_ai.evaluation.score.value` | the result's scalar return value | Populated for `bool` (`True`→`1.0`, `False`→`0.0`) and numeric returns. Omitted for `str` returns. |
+| `gen_ai.evaluation.score.label` | the result's scalar return value | Populated for `bool` (`True`→`'pass'`, `False`→`'fail'`) and `str` returns (used directly as the label). Omitted for numeric returns. |
+| `gen_ai.evaluation.explanation` | `EvaluationResult.reason` (success) or `EvaluatorFailure.error_message` (failure) | Omitted when absent. Set this via `reason=...` when constructing an `EvaluationResult` inside a custom evaluator. |
+| `error.type` | fixed string, failure events only | `'pydantic_evals.EvaluatorFailure'` when the evaluator raised; absent on successful evaluations. |
```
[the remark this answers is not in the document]
dmontagu: Fixed in f16e827. The `error.type` row now reads: "The exception class name (e.g. `'ValueError'`) when the failure was built from a caught exception; falls back to `'pydantic_evals.EvaluatorFailure'` for `EvaluatorFailure` instances constructed without it. Absent on successful evaluations."

pydantic_evals/pydantic_evals/_online.py
```diff
@@ -0,0 +1,585 @@
...
+        if config.emit_otel_events:
+            try:
+                emit_otel_events(
+                    results=results,
+                    failures=failures,
+                    target=target,
+                    include_baggage=config.include_baggage,
+                )
+            except Exception as exc:  # pragma: no cover - defensive
+                await _call_on_error(on_error, exc, context, evaluator, 'sink')
```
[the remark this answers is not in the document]
dmontagu: Documented rather than renamed, per discussion with @DouweM: the `'sink'` literal stays broad (covers both custom-sink failures and the rarer default OTel-emit failures — the latter being unlikely given how mature the OTel SDK is). Committed in 3ef9d93:

- Expanded the `OnErrorLocation` literal docstring with per-value semantics, explicitly flagging `'sink'` as a catch-all for post-evaluation result delivery.
- Updated the `on_error` field docstrings on `OnlineEvaluator` and `OnlineEvalConfig` to mention the broad scope and link to `OnErrorLocation`.
- Added an inline comment at the OTel-emit catch site in `_run_and_collect` pointing to the `OnErrorLocation` contract.
- Expanded the `on_error` bullet in the docs' Error Handling section the same way.

That way handlers that branch on `location` don't misattribute the rare OTel-emit failure, and a future dedicated `'otel_emit'` literal remains an option if the distinction becomes load-bearing.

pydantic_evals/pydantic_evals/online.py
```diff
@@ -712,81 +496,174 @@ class OnlineEvalConfig:
...
             with (
-                logfire_span('evaluate {func_name}', func_name=func.__qualname__) as span,
+                _open_call_span(call_span.msg_template, call_span.span_name, recorded_inputs) as span,
                 context_subtree() as span_tree,
             ):
                 t0 = time.perf_counter()
                 result = await func(*args, **kwargs)
                 duration = time.perf_counter() - t0
+                if call_span.record_return:
+                    _record_return(span, result)
```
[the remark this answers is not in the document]
dmontagu: Addressed in 2f6406b. Dropped the one-line `_record_return` helper (which was forcing `: Any` annotations anyway) and inlined `span.set_attribute('return', result)` with a narrow `try/except Exception` + `pragma: no cover - defensive` at both the async and sync wrapper sites. Silent swallow rather than routing through `on_error`: `record_return=True` is opt-in observability, and a systemic failure will show up as a missing `return` attribute in the trace without conflating it with the sink/OTel-emit error path.

pydantic_evals/pydantic_evals/dataset.py
```diff
@@ -1099,6 +1027,7 @@ async def _run_report_evaluators(
                         error_message=f'{type(e).__name__}: {e}',
```
alexmojaki, line 1027: should this still contain the type?

pydantic_evals/pydantic_evals/dataset.py
```diff
@@ -941,78 +941,6 @@ def _get_relative_path_reference(target: Path, source: Path, _prefix: str = '')
         return _get_relative_path_reference(target, source.parent, _prefix=f'{_prefix}../')
 
 
-@dataclass
```
alexmojaki, line 944: in future it'd be nice if refactors like moving big chunks of code like this without changing anything were done first in a separate PR, so only actual changes need reviewing

docs/evals/online-evaluation.md
```diff
@@ -146,57 +166,91 @@ async def main():
...
-    any actions you want to take on results (logging, alerting, storing, etc.).
-    Because of this, **evaluators are skipped entirely when no sinks are configured**
-    (neither a per-evaluator sink nor a default sink). This avoids wasted work when there's
-    nowhere to send results.
+`payload.results` and `payload.failures` may cover one or more evaluators from a single function call — when multiple evaluators share a sink, their results are batched into a single `submit()` call. Each result carries its own attribution (name, `evaluator_version` on [`EvaluationResult`][pydantic_evals.evaluators.EvaluationResult] and [`EvaluatorFailure`][pydantic_evals.evaluators.EvaluatorFailure], and source spec), so sinks can separate them downstream; see [Evaluator Versioning](#evaluator-versioning). The `payload.target` identifies the function or agent being evaluated (see [Target](#target)).
+
+!!! note "`SinkPayload` is extensible"
+    Additional fields may be added to [`SinkPayload`][pydantic_evals.online.SinkPayload] in future releases. Read only the attributes your sink needs, and do not instantiate `SinkPayload` directly in application code.
+
```
alexmojaki:
```suggestion
```

i think the warning in the reference is enough

docs/evals/online-evaluation.md
```diff
@@ -146,57 +166,91 @@ async def main():
...
+### Default OTel event emission
+
+Every dispatched evaluator emits one `gen_ai.evaluation.result` OTel log event per [`EvaluationResult`][pydantic_evals.evaluators.EvaluationResult] or [`EvaluatorFailure`][pydantic_evals.evaluators.EvaluatorFailure], unconditionally — no sink registration required. Events are parented to the span that produced them, so they appear nested under the original function call in the trace. If no OTel SDK is configured in the process, emission is a cheap no-op.
+
+Each event has `event.name = 'gen_ai.evaluation.result'` and a short human-readable body (e.g. `evaluation: accuracy=0.87`, or `evaluation: accuracy failed: <error>`). Emission follows the [OpenTelemetry GenAI evaluation semconv](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-events/#event-gen_aievaluationresult), with these attributes:
+
+| Attribute | Source | Notes |
```
alexmojaki: Image

I don't know if it'll look better in the unified docs, but maybe a table here doesn't actually work well and these should just be bullet points.

docs/evals/online-evaluation.md
```diff
@@ -146,57 +166,91 @@ async def main():
...
+| `error.type` | `EvaluatorFailure.error_type` (failure events only) | The exception class name (e.g. `'ValueError'`) when the failure was built from a caught exception; falls back to `'pydantic_evals.EvaluatorFailure'` for `EvaluatorFailure` instances constructed without it. Absent on successful evaluations. |
+| `gen_ai.evaluation.target` | `@evaluate(target=...)` or agent `name` | See [Target](#target). |
+| `gen_ai.evaluation.evaluator.version` | `Evaluator.evaluator_version` class attribute | Omitted when the class doesn't set it. See [Evaluator Versioning](#evaluator-versioning). |
+| `gen_ai.evaluation.evaluator.source` | JSON-serialized [`EvaluatorSpec`][pydantic_evals.evaluators.evaluator.EvaluatorSpec] | Identifies the evaluator class and its constructor arguments, so downstream queries can group by evaluator identity without relying on `name` alone (two different `LLMJudge(rubric=...)` instances share a name but have different sources). |
+
+OTel baggage entries (if any) are also attached to each event as attributes — configurable via `include_baggage` on the config. The `gen_ai.*` and `error.type` attributes above always win on conflict with baggage.
```
alexmojaki:
```suggestion
[OTel baggage](https://pydantic.dev/docs/logfire/reference/baggage/) entries (if any) are also attached to each event as attributes — configurable via `include_baggage` on the config. The `gen_ai.*` and `error.type` attributes above always win on conflict with baggage.
```

docs/evals/online-evaluation.md
```diff
@@ -146,57 +166,91 @@ async def main():
...
+
+For example, a `ToneCheck(evaluator_version='v2')` evaluator whose `evaluate()` returns `True`, decorated as `@evaluate(ToneCheck(), target='customer_support')`, emits one event with:
 
-    If you need side effects produced _within_ an evaluator itself, configure a no-op sink
-    to ensure the evaluator runs:
+- `gen_ai.evaluation.name = 'ToneCheck'`
+- `gen_ai.evaluation.score.value = 1.0`, `gen_ai.evaluation.score.label = 'pass'`
```
alexmojaki:
```suggestion
- `gen_ai.evaluation.score.value = 1.0`
- `gen_ai.evaluation.score.label = 'pass'`
```

docs/evals/online-evaluation.md
```diff
@@ -146,57 +166,91 @@ async def main():
...
+| `gen_ai.evaluation.target` | `@evaluate(target=...)` or agent `name` | See [Target](#target). |
+| `gen_ai.evaluation.evaluator.version` | `Evaluator.evaluator_version` class attribute | Omitted when the class doesn't set it. See [Evaluator Versioning](#evaluator-versioning). |
+| `gen_ai.evaluation.evaluator.source` | JSON-serialized [`EvaluatorSpec`][pydantic_evals.evaluators.evaluator.EvaluatorSpec] | Identifies the evaluator class and its constructor arguments, so downstream queries can group by evaluator identity without relying on `name` alone (two different `LLMJudge(rubric=...)` instances share a name but have different sources). |
+
+OTel baggage entries (if any) are also attached to each event as attributes — configurable via `include_baggage` on the config. The `gen_ai.*` and `error.type` attributes above always win on conflict with baggage.
+
+For example, a `ToneCheck(evaluator_version='v2')` evaluator whose `evaluate()` returns `True`, decorated as `@evaluate(ToneCheck(), target='customer_support')`, emits one event with:
```
alexmojaki: I suggest using `OutputNotEmpty` since it's already been mentioned above and actually returns a bool. I think evaluator_version can be left out of this example, either way making it look like a constructor argument isn't good.

docs/evals/online-evaluation.md
```diff
@@ -146,57 +166,91 @@ async def main():
...
 
-    If you need side effects produced _within_ an evaluator itself, configure a no-op sink
-    to ensure the evaluator runs:
+- `gen_ai.evaluation.name = 'ToneCheck'`
+- `gen_ai.evaluation.score.value = 1.0`, `gen_ai.evaluation.score.label = 'pass'`
+- `gen_ai.evaluation.target = 'customer_support'`
+- `gen_ai.evaluation.evaluator.version = 'v2'`
+- `gen_ai.evaluation.evaluator.source = '{"name": "ToneCheck", "arguments": null}'`
```
alexmojaki: maybe an example that has arguments would be better

docs/evals/online-evaluation.md
dmontagu: Let's just make it bullet points, not a table.

docs/evals/online-evaluation.md
dmontagu: maybe it makes sense to use a "standard" evaluator like `LLMJudge` with a rubric etc.

pydantic_evals/pydantic_evals/dataset.py
dmontagu: Noted — agreed. Will split mechanical relocations off into their own prep PR before the substantive change from here on so the review diff stays focused.

pydantic_evals/pydantic_evals/_online.py
```diff
@@ -0,0 +1,594 @@
...
+    except (TypeError, ValueError):  # pragma: no cover — some C-implemented callables
+        return False  # can't introspect — assume modern signature, don't warn
+    non_var_kwargs = [p for p in params if p.kind is not inspect.Parameter.VAR_KEYWORD]
+    if len(non_var_kwargs) != 1:
+        return True
+    return non_var_kwargs[0].kind not in (
+        inspect.Parameter.POSITIONAL_ONLY,
+        inspect.Parameter.POSITIONAL_OR_KEYWORD,
+        inspect.Parameter.VAR_POSITIONAL,
+    )
```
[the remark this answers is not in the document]
dmontagu: Good catch — fixed in 54778c8. Took the simpler route of dropping the parameter-kind check entirely: any single non-`**kwargs` parameter is now treated as modern (including keyword-only `submit(self, *, payload)`), which avoids the spurious deprecation warning you described. Added a focused unit test on `_is_legacy_submit`.

docs/evals/online-evaluation.md
dmontagu: Dropped — the warning on `SinkPayload` is enough.

docs/evals/online-evaluation.md
dmontagu: Converted to bullets.

docs/evals/online-evaluation.md
dmontagu: Applied.

docs/evals/online-evaluation.md
dmontagu: Applied.

docs/evals/online-evaluation.md
dmontagu: Switched to `OutputNotEmpty` and dropped `evaluator_version` from the example.

docs/evals/online-evaluation.md
dmontagu: Added an `LLMJudge(rubric='Is the response helpful?')` snippet showing how constructor arguments render into `gen_ai.evaluation.evaluator.source`.

pydantic_evals/pydantic_evals/dataset.py
dmontagu: We decided not to change this at this time.

alexmojaki (approved):

pydantic_evals/pydantic_evals/online.py
```diff
@@ -712,81 +471,173 @@ class OnlineEvalConfig:
...
                 # Callable[_P, Awaitable[_R]] but the decorator signature expects Callable[_P, _R]
-                return _wrap_async(func, online_evals, self)  # pyright: ignore[reportReturnType]
+                return _wrap_async(func, sig, online_evals, self, resolved_target, call_span)  # pyright: ignore[reportReturnType]
             else:
-                return _wrap_sync(func, online_evals, self)
+                return _wrap_sync(func, sig, online_evals, self, resolved_target, call_span)
 
         return decorator
 
+    def should_evaluate(self):
```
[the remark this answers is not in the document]
dmontagu: Added  in d06e6ca.

pydantic_evals/pydantic_evals/_task_run.py
```diff
@@ -0,0 +1,76 @@
...
+        if current_value == 0 and incremented_value == 0:
+            return
+        self.record_metric(name, incremented_value)
+
+    def record_attribute(self, name: str, value: Any) -> None:
+        self.attributes[name] = value
+
+
+@contextmanager
+def run_task():
```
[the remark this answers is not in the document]
dmontagu: Annotated the yield type as `Iterator[Callable[[], dict[str, Any]]]` in d06e6ca.

pydantic_evals/pydantic_evals/_online.py
```diff
@@ -0,0 +1,590 @@
...
+]
+SamplingMode = Literal['independent', 'correlated']
+OnErrorLocation = Literal['sink', 'on_max_concurrency']
+OnSamplingErrorCallback = Callable[[Exception, Evaluator], None]
+OnMaxConcurrencyCallback = Callable[[EvaluatorContext], None | Awaitable[None]]
+OnErrorCallback = Callable[
+    [Exception, EvaluatorContext, Evaluator, OnErrorLocation],
+    None | Awaitable[None],
+]
+SamplingContextBuilder = Callable[[Evaluator, Any, dict[str, Any] | None, float], Any]
```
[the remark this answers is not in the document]
dmontagu: Removed in d06e6ca.

PR #4977, opened 2026-04-04
Add `TemporalDurability`, `DBOSDurability` and `PrefectDurability` capabilities to replace the deprecated wrapper agents

DouweM:
## Summary

Introduces durability **capabilities** — `TemporalDurability`, `DBOSDurability`, and `PrefectDurability` — that you attach to an agent via `capabilities=[...]` instead of wrapping the agent.

```python
from dbos import DBOS

from pydantic_ai import Agent
from pydantic_ai.durable_exec.dbos import DBOSDurability

agent = Agent(
    'openai:gpt-5.2',
    name='geography',
    capabilities=[DBOSDurability()],
)

@DBOS.workflow()
async def answer(prompt: str) -> str:
    # Inside the workflow, model requests and MCP communication
    # run as durable DBOS steps; outside one, the agent behaves
    # like a regular agent.
    result = await agent.run(prompt)
    return result.output
```

The capability makes the agent *durable-capable*; your workflow/flow decides which runs are durable. Same shape for Temporal (run inside a `@workflow.defn`, with activities registered on the worker via `PydanticAIPlugin` or `DurabilityPlugin`) and Prefect (run inside a `@flow`) — migration diffs below.

The existing `TemporalAgent` / `DBOSAgent` / `PrefectAgent` wrapper classes are **deprecated** in this PR (they emit `DeprecationWarning` and are slated for removal in v3). They keep working unchanged at runtime — port at your leisure.

## Try it out

If you're already on a wrapper agent, the port is roughly a one-liner per agent. The capability constructors mirror the wrapper-agent constructors, minus the agent-introspection arguments (model, name, toolsets) that the capability discovers automatically when it's bound to the agent.

**Temporal** — replace the `TemporalAgent` wrap with a capability and (optionally) swap the worker plugin:

```diff
-from pydantic_ai.durable_exec.temporal import TemporalAgent
-temporal_agent = TemporalAgent(agent, models={'fast': fast_model})
+from pydantic_ai.durable_exec.temporal import TemporalDurability
+agent = Agent(
+    'openai:gpt-5.2',
+    name='geography',
+    capabilities=[TemporalDurability(models={'fast': fast_model})],
+)

 @workflow.defn
 class GeographyWorkflow(PydanticAIWorkflow):
-    __pydantic_ai_agents__ = [temporal_agent]
+    __pydantic_ai_agents__ = [agent]  # PydanticAIPlugin discovers TemporalDurability

     @workflow.run
     async def run(self, prompt: str) -> str:
[...]

PR #5141, opened 2026-04-17
Add `ProcessEventStream` capability

DouweM:
- Closes #

### Checklist

## Summary

Introduces `ProcessEventStream`, a capability wrapping a user-provided event stream handler so it can participate in the capability middleware chain. Its position in the capability list determines nesting order, same as other hook-overriding capabilities — list earlier = wraps outside.

Supports two handler forms:
- **Observer** ([`EventStreamHandler`][pydantic_ai.agent.EventStreamHandler]) — `async def` returning `None`. Events are teed to the handler while also flowing through to the rest of the chain. Synchronous delivery, so a slow observer back-pressures the downstream stream.
- **Processor** ([`EventStreamProcessor`][pydantic_ai.agent.EventStreamProcessor]) — async generator yielding events. Its output replaces the inner stream for downstream wrappers, so it can modify, drop, or add events.

The existing `event_stream_handler` param on `Agent` / `Agent.run` is untouched — both approaches coexist. `abstract.py`'s existing `has_wrap_run_event_stream` check already auto-enables streaming when any capability overrides the hook, so no changes were needed there.

Durable-execution integrations (Temporal/DBOS/Prefect) consume model events inside the activity/step boundary, so this capability only sees tool-call events and the post-streaming batch under those wrappers; the live event path remains available via the `event_stream_handler` param. Threading the capability chain through the activity boundary is being explored in #4977.

## Test plan

- [x] `TestProcessEventStream::test_handler_receives_events` — handler registered via capability receives events end-to-end
- [x] `TestProcessEventStream::test_multiple_handlers_and_param_all_observe` — multiple capabilities and the param handler all see the same stream
- [x] `TestProcessEventStream::test_handler_sees_events_after_inner_wrappers` — when listed before an inner wrapper, the handler sees the transformed stream
- [x] `TestProcessEventStream::test_transformer_handler_replaces_stream` — processor form modifies the stream for downstream consumers
- [x] `TestProcessEventStream::test_callable_instance_processor` — callable-class processors dispatch correctly
- [x] `TestProcessEventStream::test_observer_bailout_does_not_break_downstream` — observer early-return doesn't crash downstream
- [x] `TestProcessEventStream::test_not_spec_serializable` — capability opts out of spec construction (takes a callable)
- [x] Full `tests/test_capabilities.py` suite, `tests/test_streaming.py`, `tests/test_agent.py` all pass


pydantic_ai_slim/pydantic_ai/capabilities/event_stream_handler.py
```diff
@@ -0,0 +1,103 @@
...
+
+if TYPE_CHECKING:
+    from pydantic_ai.agent.abstract import (
+        EventStreamHandler as EventStreamHandlerFunc,
+        EventStreamProcessor as EventStreamProcessorFunc,
+    )
+
+
+@dataclass
+class HandleEventStream(AbstractCapability[AgentDepsT]):
```
DouweM: What do you think about calling this ProcessEventStream instead, since "handle" underplays its potential side effects? And maybe `capabilities.HistoryProcessor` should become `ProcessHistory` in v2 as well.

pydantic_ai_slim/pydantic_ai/capabilities/event_stream_handler.py
```diff
@@ -0,0 +1,103 @@
...
+        ([Temporal][pydantic_ai.durable_exec.temporal.TemporalAgent],
+        [DBOS][pydantic_ai.durable_exec.dbos.DBOSAgent],
+        [Prefect][pydantic_ai.durable_exec.prefect.PrefectAgent]), model streaming happens
+        inside an activity/step rather than in the outer agent loop. This capability's
+        `wrap_run_event_stream` hook fires for tool-call events and the final post-streaming
+        batch, but it does **not** see individual model-response events live — the underlying
+        durable model consumes those inside the activity before returning. The in-flight
+        `event_stream_handler` parameter does still observe the live events, and a future
+        refactor threading the capability chain through the activity boundary is tracked
+        separately.
```
DouweM: Tracked where? We should link to https://github.com/pydantic/pydantic-ai/pull/4977.

pydantic_ai_slim/pydantic_ai/agent/abstract.py
```diff
@@ -137,6 +147,12 @@ def event_stream_handler(self) -> EventStreamHandler[AgentDepsT] | None:
         """Optional handler for events from the model's streaming response and the agent's execution of tools."""
         raise NotImplementedError
 
+    @property
+    @abstractmethod
+    def root_capability(self) -> CombinedCapability[AgentDepsT]:
```
DouweM, line 151: Does this require all subclasses to implement this now, even existing ones out in the wild? That'd be a breaking change. I'd rather have it just raise in that case, not block the subclass creation.

pydantic_ai_slim/pydantic_ai/capabilities/event_stream_handler.py
DouweM: Renamed to `ProcessEventStream` in `869e257a`. Filed a tracker plan for `HistoryProcessor` → `ProcessHistory` at `~/.claude/plans/history-processor-rename.md` — will pick that up as a separate PR so this one stays focused.

pydantic_ai_slim/pydantic_ai/capabilities/event_stream_handler.py
DouweM: Linked to #4977 in `869e257a`.

pydantic_ai_slim/pydantic_ai/agent/abstract.py
DouweM: Good catch. Dropped `@abstractmethod` in `869e257a`; the default raises `NotImplementedError` on access but no longer blocks subclass instantiation.

pydantic_ai_slim/pydantic_ai/capabilities/process_event_stream.py
```diff
@@ -0,0 +1,103 @@
...
+            async with send_stream:
+                handler_alive = True
+                async for event in stream:
+                    if handler_alive:
+                        try:
+                            await send_stream.send(event)
+                        except (anyio.BrokenResourceError, anyio.ClosedResourceError):
+                            # Handler bailed early; keep forwarding events downstream.
+                            handler_alive = False
+                    yield event
```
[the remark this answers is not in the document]
DouweM: Good call on the back-pressure coupling. Documented the synchronous delivery in `cb420275` — reworded "without interfering" to "without changing each other's view" and added an explicit line about slow handlers back-pressuring the rest of the stream. Keeping rendezvous (default `max_buffer_size=0`) intentionally; adding a buffer would risk unbounded memory growth for pathologically slow observers, and events are low-frequency enough in practice that typical record/print observers aren't a concern.

pydantic_ai_slim/pydantic_ai/capabilities/process_event_stream.py
```diff
@@ -0,0 +1,104 @@
...
+
+    @classmethod
+    def get_serialization_name(cls) -> str | None:
+        return None  # Not spec-serializable (takes a callable)
```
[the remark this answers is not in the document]
DouweM: Good catch on the mismatch — but I think the right fix is to drop the "outermost" claim, not enforce it. List-position ordering is the natural contract for user-controlled capabilities: `CombinedCapability.wrap_run_event_stream` wraps in `reversed(self.capabilities)` order (list-earlier = outermost), and `sort_capabilities` preserves insertion order as a tiebreaker when no explicit constraints apply. `HistoryProcessor` doesn't force a position either; only `_ToolSearchCapability` does, and it's a special internal case. Forcing `outermost` here would remove the user's ability to stack a `ProcessEventStream` (say, for logging raw events) inside another capability that transforms the stream. Updated the PR description to remove the outermost claim and renamed the test intent to reflect that the ordering follows list position.

pydantic_ai_slim/pydantic_ai/capabilities/__init__.py
```diff
@@ -75,6 +76,7 @@
     'BuiltinTool',
     'BuiltinOrLocalTool',
     'CAPABILITY_TYPES',
+    'ProcessEventStream',
```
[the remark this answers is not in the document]
DouweM: Fixed in `9d4b45cb` — moved `ProcessEventStream` between `PrepareTools` and `ReinjectSystemPrompt`.

PR #4981, opened 2026-04-04
Add DeferredToolHandler capability and DeferredToolRequestsPending exception

DouweM:
## Summary

- Adds `DeferredToolHandler` capability that handles deferred tool requests inline during agent runs using the `on_node_run_error` hook, per #3959
- Adds `DeferredToolRequestsPending` exception (replaces `UserError` gate) raised when deferred tools are called without `DeferredToolRequests` in the output type — can be caught by callers or handled by the capability
- Extracts `build_tool_call_results()` for shared conversion logic between `UserPromptNode` and `DeferredToolHandler`
- Removes the construction-time gate that blocked `requires_approval` tools without DTR in output type
- Fixes pydantic-graph to support error recovery and End→node redirection via hooks:
  - `ErrorMarker` keeps graph iterator alive when nodes raise exceptions
  - `GraphRun.override_next()` public API for redirecting the graph after End or error

Closes #3959

## Test plan

- [x] 7 tests for `DeferredToolHandler`: approve, deny, external tools, sync handler, mixed function+deferred tools, DTR-in-output-type passthrough, streaming
- [x] 3 pydantic-graph tests for error recovery: retry same node, redirect to different node, unrecovered error still raises
- [x] Updated `test_deferred_tool_without_output_type` to expect `DeferredToolRequestsPending`
- [x] Removed `test_tool_requires_approval_error` (construction-time gate removed)
- [x] All 899 existing tests pass (0 failures, 3 xfailed)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5142, opened 2026-04-17
feat: add `HandleDeferredToolCalls` capability and `handle_deferred_tool_calls` hook

### LINKED ISSUE #3959: `deferred_handler` for Inline Resolution of Deferred Tools

DouweM:
## Summary

- Adds `handle_deferred_tool_calls` hook on `AbstractCapability` for inline resolution of deferred tool calls (approval-required and externally-executed) during agent runs
- Adds `HandleDeferredToolCalls` capability that wraps a user handler function
- ToolManager owns all deferred resolution logic: `resolve_deferred_tool_calls()`, `execute_deferred_tool_results()`, and `handle_call(resolve_deferred=True)` default
- Accumulation dispatch in `CombinedCapability`: each capability resolves what it can, remaining passed to next
- Helper methods: `DeferredToolRequests.build_results()`, `.remaining()`, `DeferredToolResults.merge()`
- Drops construction-time `requires_approval` check (runtime check catches it)
- Replaces #4981 with a simpler approach that doesn't require pydantic-graph changes and naturally extends to nested scenarios (CodeMode, subagents)

Closes #3959

## Test plan

- [x] 8 tests for `HandleDeferredToolCalls`: approve, deny, no-output-type-needed, fallback, partial resolution, sync handler, accumulation dispatch, unresolved error
- [x] Updated `test_tool_requires_approval_error` → `test_tool_requires_approval_no_output_type` (construction-time check removed)
- [x] All 765 existing tests pass


pydantic_ai_slim/pydantic_ai/capabilities/deferred_tool_handler.py
```diff
@@ -0,0 +1,75 @@
...
+
+    When tools require approval or external execution, the agent normally pauses the run
+    and returns [`DeferredToolRequests`][pydantic_ai.tools.DeferredToolRequests] as output.
+    This capability intercepts deferred tool calls, calls the provided handler to resolve
+    them, and continues the agent run automatically.
+
+    The handler receives the [`RunContext`][pydantic_ai.tools.RunContext] and the
+    [`DeferredToolRequests`][pydantic_ai.tools.DeferredToolRequests], and must return
+    [`DeferredToolResults`][pydantic_ai.tools.DeferredToolResults] with results for
+    some or all pending tool calls. Unresolved calls are passed to the next capability
```
DouweM: Let's allow the handler to return `None` as well if it doesn't want to handle at all.

pydantic_ai_slim/pydantic_ai/capabilities/abstract.py
```diff
@@ -618,6 +625,30 @@ async def on_tool_execute_error(
...
+        self,
+        ctx: RunContext[AgentDepsT],
+        *,
+        requests: DeferredToolRequests,
+    ) -> DeferredToolResults | None:
+        """Handle deferred tool calls (approval-required or externally-executed) inline during an agent run.
+
+        Called by [`ToolManager`][pydantic_ai.tool_manager.ToolManager] after tool calls raise
+        [`ApprovalRequired`][pydantic_ai.exceptions.ApprovalRequired] or
+        [`CallDeferred`][pydantic_ai.exceptions.CallDeferred].
```
DouweM: Also when the model calls a tool that was specifically registered as unapproved/external. We should link to the deferred tools / approval / external tools docs here.

pydantic_ai_slim/pydantic_ai/tools.py
```diff
@@ -267,6 +267,26 @@ class DeferredToolRequests:
...
 
+    def build_results(
+        self,
+        *,
+        approvals: dict[str, bool | DeferredToolApprovalResult] | None = None,
+        calls: dict[str, DeferredToolCallResult | Any] | None = None,
+        metadata: dict[str, dict[str, Any]] | None = None,
+    ) -> DeferredToolResults:
+        """Create a [`DeferredToolResults`][pydantic_ai.tools.DeferredToolResults] for these requests."""
+        return DeferredToolResults(approvals=approvals or {}, calls=calls or {}, metadata=metadata or {})
```
DouweM: This is an instance method because in the future I expect to add a `context: object` field to `Request` that is meant to be reproduced on `Results` verbatim, but I agree with an earlier auto-review comment that this method doesn't serve much purpose right now. 

So let's make it serve some purpose and raise an error if a result is provided for a tool call ID that wasn't in the appropriate "requests list". We may have a check already somewhere that you can't provide ToolApproved for an external call that wants a result. I don't know if we similarly block (or should block) providing a tool result directly for a tool that was looking for approval. But either way some validation here would be useful.

DouweM: An optional `approve_all=True` is also interesting. If that's provided along with approvals, it would apply to the rest?

pydantic_ai_slim/pydantic_ai/tool_manager.py
```diff
@@ -39,6 +39,35 @@
...
+
+@dataclass
+class DeferredCallRetry:
+    """A deferred tool call where execution raised `ModelRetry` after approval."""
+
+    retry: _messages.RetryPromptPart
+
+
+DeferredCallOutcome: TypeAlias = DeferredCallSuccess | DeferredCallDenied | DeferredCallRetry
+"""Outcome of resolving a single deferred tool call."""
```
DouweM: I see too much overlap here with these from `tools.py`:

```
DeferredToolApprovalResult: TypeAlias = Annotated[ToolApproved | ToolDenied, Discriminator('kind')]
"""Result for a tool call that required human-in-the-loop approval."""
DeferredToolCallResult: TypeAlias = Annotated[
    Annotated[ToolReturn, Tag('tool-return')]
    | Annotated[ModelRetry, Tag('model-retry')]
    | Annotated[RetryPromptPart, Tag('retry-prompt')],
    Discriminator(_deferred_tool_call_result_discriminator),
]
"""Result for a tool call that required external execution."""
DeferredToolResult = DeferredToolApprovalResult | DeferredToolCallResult
"""Result for a tool call that required approval or external execution."""
```

I don't think we need both? We should be able to use those?

pydantic_ai_slim/pydantic_ai/tool_manager.py
```diff
@@ -613,6 +643,11 @@ async def handle_call(
             call: The tool call part to handle.
             allow_partial: Whether to allow partial validation of the tool arguments.
             wrap_validation_errors: Whether to wrap validation errors in a retry prompt part.
+            resolve_deferred: Whether to resolve deferred tool calls inline using the
+                capability handler. When `True` (default), if the tool raises
+                [`ApprovalRequired`][pydantic_ai.exceptions.ApprovalRequired] or
+                [`CallDeferred`][pydantic_ai.exceptions.CallDeferred], the handler is
+                called to resolve it. When `False`, the exceptions propagate to the caller.
```
DouweM: I suppose the case where we set this to False is if we want to call the handler in a batch? Do we do that currently in the agent graph?

pydantic_ai_slim/pydantic_ai/tool_manager.py
```diff
@@ -623,4 +658,228 @@ async def handle_call(
...
+            # Get the deferred result for this call
+            if call.tool_call_id in results.approvals:
+                approval = results.approvals[call.tool_call_id]
+                if isinstance(approval, bool):
+                    deferred_result: ToolApproved | ToolDenied = ToolApproved() if approval else ToolDenied()
+                else:
+                    deferred_result = approval
+            elif call.tool_call_id in results.calls:
+                # External call results are used directly as the raw tool result
+                executed.append((call, DeferredCallSuccess(result=results.calls[call.tool_call_id])))
```
DouweM: I see overlap with agent graph UserPromptNode:

```
tool_call_results: dict[str, DeferredToolResult | Literal['skip']] | None = None
        tool_call_results = {}
        for tool_call_id, approval in deferred_tool_results.approvals.items():
            if approval is True:
                approval = ToolApproved()
            elif approval is False:
                approval = ToolDenied()
            tool_call_results[tool_call_id] = approval

        if calls := deferred_tool_results.calls:
            call_result_types = get_union_args(DeferredToolCallResult)
            for tool_call_id, result in calls.items():
                if not isinstance(result, call_result_types):
                    result = _messages.ToolReturn(result)
                tool_call_results[tool_call_id] = result
```

pydantic_ai_slim/pydantic_ai/tool_manager.py
```diff
@@ -623,4 +629,208 @@ async def handle_call(
...
+                        _messages.ToolReturnPart(
+                            tool_name=call.tool_name,
+                            content=content,
+                            tool_call_id=call.tool_call_id,
+                            metadata=result_metadata,
+                        ),
+                        user_content,
+                    )
+                )
+                continue
```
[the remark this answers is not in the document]
DouweM: This one looks real!

pydantic_ai_slim/pydantic_ai/tool_manager.py
```diff
@@ -623,4 +658,228 @@ async def handle_call(
...
+            if any(c.tool_call_id == call.tool_call_id for c in remaining.calls):
+                raise CallDeferred(metadata=new_metadata)
+            raise ApprovalRequired(metadata=new_metadata)
+        if executed:
+            _call, outcome = executed[0]
+            if isinstance(outcome, DeferredCallRetry):
+                raise ToolRetryError(outcome.retry)
+            if isinstance(outcome, DeferredCallDenied):
+                # Consistent with the denial message surfacing as the tool return in message history
+                return outcome.denied.message
```
DouweM: Do we lose `outcome=denied` on the ToolReturnPart here, somehow? should we raise the exception?...

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -1536,17 +1536,34 @@ async def process_tool_calls(  # noqa: C901
...
-        final_result = result.FinalResult(cast(NodeRunEndT, deferred_tool_requests), None, None)
+        # Let capability handlers resolve deferred calls inline (one shot).
+        # If re-execution raises new deferrals, they go into remaining.
+        handler_results = await tool_manager.resolve_deferred_tool_calls(deferred_tool_requests)
+        if handler_results is not None:
+            executed_outcomes, deferred_tool_requests = await tool_manager.execute_deferred_tool_results(
+                deferred_tool_requests, handler_results
+            )
+            executed_parts = tool_manager.build_tool_return_parts_from_outcomes(executed_outcomes)
+            for _call_part, result_part, user_content in executed_parts:
```
DouweM: I don't love the complicated list of tuples this method returns...

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -1536,17 +1536,34 @@ async def process_tool_calls(  # noqa: C901
...
 
-        final_result = result.FinalResult(cast(NodeRunEndT, deferred_tool_requests), None, None)
+        # Let capability handlers resolve deferred calls inline (one shot).
+        # If re-execution raises new deferrals, they go into remaining.
+        handler_results = await tool_manager.resolve_deferred_tool_calls(deferred_tool_requests)
+        if handler_results is not None:
+            executed_outcomes, deferred_tool_requests = await tool_manager.execute_deferred_tool_results(
+                deferred_tool_requests, handler_results
+            )
+            executed_parts = tool_manager.build_tool_return_parts_from_outcomes(executed_outcomes)
```
DouweM: Do these 2 method calls need to be separate? Would a single helper method make sense?

pydantic_ai_slim/pydantic_ai/agent/__init__.py
```diff
@@ -2727,8 +2727,4 @@ def label(self) -> str:
         return 'the agent'
 
     def add_tool(self, tool: Tool[AgentDepsT]) -> None:
-        if tool.requires_approval and not self.output_schema.allows_deferred_tools:
-            raise exceptions.UserError(
-                'To use tools that require approval, add `DeferredToolRequests` to the list of output types for this agent.'
-            )
         super().add_tool(tool)
```
DouweM: Could we stop overriding this method now?

tests/test_capabilities.py
```diff
@@ -8910,3 +8912,1054 @@ def my_tool() -> str:
...
+        return DeferredToolResults(
+            calls={
+                call.tool_call_id: ToolReturn(return_value='external result', metadata={'source': 'ext'})
+                for call in requests.calls
+            }
+        )
+
+    agent = Agent(
+        FunctionModel(llm),
+        capabilities=[HandleDeferredToolCalls(handler=handle_deferred)],
```
DouweM, line 9865: We need docs where we currently introduce deferred/external/approved tools and how to work with `DeferredToolRequests`! This new approach is _way_ simpler than what we had before. The old "stop the world/agent run" method still has its uses though, like with UI adapters where you need to return the pending tool call to the frontend, so that _it_ can then start a new run that sends back the result/approval.

pydantic_ai_slim/pydantic_ai/tool_manager.py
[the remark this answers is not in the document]
DouweM: Is this one still relevant?

docs/deferred-tools.md
```diff
@@ -15,6 +15,8 @@ When the model calls a deferred tool, the agent run will end with a [`DeferredTo
...
 
+As an alternative to returning control to the caller, a [`HandleDeferredToolCalls`][pydantic_ai.capabilities.HandleDeferredToolCalls] capability can resolve deferred calls inline with a handler function, so the agent run continues in one step without needing to end and restart. See [Resolving deferred calls inline](#resolving-deferred-calls-inline).
```
DouweM: I'd prefer to present this as the primary method of dealing with deferred tools, with both methods getting their own section but the handler method being introduced first.

pydantic_ai_slim/pydantic_ai/tool_manager.py
```diff
@@ -865,21 +675,35 @@ async def _resolve_single_deferred(
...
-            return outcome.result
-        raise exc  # pragma: no cover
+        if call.tool_call_id in deferred_results.approvals:
+            approval = deferred_results.approvals[call.tool_call_id]
+            if approval is True:
+                approval = ToolApproved()
+            elif approval is False:
+                approval = ToolDenied()
+            if isinstance(approval, ToolDenied):
+                return approval.message
```
DouweM: It feels like we're missing setting `ToolReturnPart.outcome='denied'` here?

pydantic_ai_slim/pydantic_ai/tool_manager.py
```diff
@@ -865,21 +675,35 @@ async def _resolve_single_deferred(
...
-                return outcome.denied.message
-            # DeferredCallSuccess — return the raw tool result (preserves `ToolReturn` wrapper for CodeMode etc.)
-            return outcome.result
-        raise exc  # pragma: no cover
+        if call.tool_call_id in deferred_results.approvals:
+            approval = deferred_results.approvals[call.tool_call_id]
+            if approval is True:
+                approval = ToolApproved()
+            elif approval is False:
+                approval = ToolDenied()
```
DouweM: There is some apparent duplication between here and `to_tool_call_results`. Inevitable?

docs/deferred-tools.md
```diff
@@ -11,11 +11,78 @@ To support these use cases, Pydantic AI provides the concept of deferred tools,
...
+
+@agent.tool_plain
+async def send_to_worker(task: str) -> str:
+    raise CallDeferred  # (2)!
+```
+
+1. Never reached here — the handler denies this call, so the model sees the denial message instead.
+2. The handler supplies the result for this external call, so the tool body just signals the deferral.
+
+If the handler declines to resolve some or all of the calls (by omitting them from the returned [`DeferredToolResults`][pydantic_ai.tools.DeferredToolResults] or returning `None`), those calls bubble up as a [`DeferredToolRequests`][pydantic_ai.output.DeferredToolRequests] output, and `DeferredToolRequests` must be in the agent's output type — so you can combine inline handling with the stop-the-world flow when it makes sense.
```
DouweM: Above it says somewhere that the handler should provide a result for "each pending call", but that's too strict, because we do allow partial resolution, and having multiple capabilities each handle a subset of deferred tool calls. 

By the way, if we haven't yet done that, we should also add docs for the new `handle_...` "hook" method on capabilities.md. And likely also hooks.md, making sure that it can be used by the `Hooks` capability, like most of the other methods on `AbstractCapability` are.

DouweM: So "bubble up as output" is not completely accurate -- because they can also be handled y anotehr handler.

DouweM: It's also worth explicitly specifying that there is a hook method on the capability class (by linking to the capabilities doc for example), if you want to use HITL/external function calls in a custom capability.

pydantic_ai_slim/pydantic_ai/exceptions.py
```diff
@@ -293,5 +295,24 @@ def _format_error_details(errors: list[pydantic_core.ErrorDetails], tool_name: s
         return '\n'.join(lines)
 
 
+class ToolDeniedError(Exception):
```
DouweM: Having this as a public class makes me want to also handle it in places like we handle `ModelRetry`, as a way for a tool call function to explicitly raise/wrap an error as "this means you weren't allowed to do this" (in surfacing to a user or o11y backend), even if our native tool approval wasn't used. That's similar to the `ToolFailed` exception being introduced in https://github.com/pydantic/pydantic-ai/issues/2586 which is worth reading. But I don't want to implement the full scope of that right now (unless you think it's really easy as we're working on this code anyway -- to allow tool functions anywhere to raise `ToolFailed('...', reason='denied')` which is registered as `outcome='denied'`? But I'd prefer not to bring that in scope, as that'd imply documenting it etc...).

Does this need to be public? Is there another convention we could find for the way a method indicates that it was a tool-denial, e.g. returning `ToolDenied | Any` and telling people to `isinstance` check it? We already pass `A | B | Any` around in a few places. Discuss with me please.

docs/hooks.md
```diff
@@ -149,6 +149,38 @@ To skip execution, raise [`SkipToolExecution(result)`][pydantic_ai.exceptions.Sk
...
+Resolves [deferred tool calls](deferred-tools.md) (approval-required or externally-executed) inline during a run. The hook receives a [`DeferredToolRequests`][pydantic_ai.tools.DeferredToolRequests] and returns a [`DeferredToolResults`][pydantic_ai.tools.DeferredToolResults] (or `None` to decline). Multiple registered hooks accumulate: each receives the still-unresolved requests and can resolve some or all of them.
+
+```python {title="hooks_handle_deferred.py"}
+from pydantic_ai import Agent, DeferredToolRequests, DeferredToolResults, RunContext
+from pydantic_ai.capabilities import Hooks
+
+hooks = Hooks()
+
+
+@hooks.on.handle_deferred_tool_calls
```
DouweM: Would this read nicer as `on.deferred_tool_calls`?

pydantic_ai_slim/pydantic_ai/tool_manager.py
```diff
@@ -610,7 +609,7 @@ async def handle_call(
         *,
         approved: bool = False,
         metadata: Any = None,
-    ) -> Any:
+    ) -> Any | ToolDenied:
```
DouweM: Would it make sense to list `ToolReurn[Any]` in these types of cases as well? And I prefer to end with `Any` and start specific, here and elsewhere.

echarles:
Awesome, curious to know the timeframe of the next release so we can migrate the Agent Runtimes tool approvals implementation https://github.com/datalayer/agent-runtimes/issues/81

PR #5146, opened 2026-04-19
Accept thinking-only responses when output_type allows None

kmike:
I have a system prompt which tells the model not to output anything after a particular tool call. But some models (e.g. gemma4 e4b) like to think and congratulate themselves in these cases, after the tool call:

> The user asked to open the file containing `...`, and I have successfully opened `<file path>` which is where it resides based on the globbing result and the name match. The task is complete for this turn.

Then, as asked, they don't produce any output. Because there is no output, pydantic retries this. It causes more useless thinking, and the model adds a message in the end (which is unwanted, and contradicts my prompt).

Arguably, it's an issue with the model, but this thinking seems harmless, and retry is not helpful. What do you think about allowing not just empty output, but also thinking when ``output_type=str | None`` (introduced in https://github.com/pydantic/pydantic-ai/pull/4042)?

### Checklist


adtyavrdhn:
Could you also record a cassette for this, I think that would be useful for us to visibly retain the behaviour. Also please take a look at the failing tests

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -1028,8 +1028,15 @@ async def _run_stream() -> AsyncIterator[_messages.HandleResponseEvent]:  # noqa
...
-                    if is_empty and output_schema.allows_none:
+                    # If the output type allows None, an empty or thinking-only
+                    # response is a valid result.  Thinking-only responses are
+                    # produced by some reasoning-oriented models (e.g. Gemma 4
+                    # e4b) when they've completed the task via a tool call and
+                    # have nothing to add; forcing a retry only makes the model
+                    # produce unnecessary follow-up text.  Opt in by declaring
+                    # the agent with ``output_type=str | None`` (or any union
+                    # including ``None``).
+                    if (is_empty or is_thinking_only) and output_schema.allows_none:
```
DouweM, line 1039: Good one, I agree this should count as well!

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -1028,8 +1028,15 @@ async def _run_stream() -> AsyncIterator[_messages.HandleResponseEvent]:  # noqa
...
-                    # If the output type allows None, an empty response is a valid result.
-                    if is_empty and output_schema.allows_none:
+                    # If the output type allows None, an empty or thinking-only
+                    # response is a valid result.  Thinking-only responses are
+                    # produced by some reasoning-oriented models (e.g. Gemma 4
+                    # e4b) when they've completed the task via a tool call and
+                    # have nothing to add; forcing a retry only makes the model
+                    # produce unnecessary follow-up text.  Opt in by declaring
+                    # the agent with ``output_type=str | None`` (or any union
+                    # including ``None``).
```
DouweM, line 1038: Single backticks please :)

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -1028,8 +1028,15 @@ async def _run_stream() -> AsyncIterator[_messages.HandleResponseEvent]:  # noqa
...
 
-                    # If the output type allows None, an empty response is a valid result.
-                    if is_empty and output_schema.allows_none:
+                    # If the output type allows None, an empty or thinking-only
+                    # response is a valid result.  Thinking-only responses are
+                    # produced by some reasoning-oriented models (e.g. Gemma 4
+                    # e4b) when they've completed the task via a tool call and
+                    # have nothing to add; forcing a retry only makes the model
+                    # produce unnecessary follow-up text.  Opt in by declaring
+                    # the agent with ``output_type=str | None`` (or any union
```
DouweM, line 1037: Also this is worth some kind of mention (or table update) in the docs I added for `output_type=str | None`

DouweM (changes requested):

adtyavrdhn (approved):
@DouweM Please have a look here

PR #5173, opened 2026-04-23
fix(vercel_ai): skip stale `approval-responded` when tool already executed (#5154)

### LINKED ISSUE #5154: iter_tool_approval_responses yields stale approval-responded parts for already-executed tools on continuation rounds
## Initial Checks

## Description

When a tool requiring approval is approved and executed in a single continuation round, the Vercel AI SDK v6 keeps the original `approval-responded` part on the assistant message alongside the new `output-available` part for the same `toolCallId`. On the *next* continuation, `VercelAIAdapter.deferred_tool_results` calls `iter_tool_approval_responses` which yields that stale `approval-responded`. It's then fed to pydantic-ai's approval-resolution logic as if pending, but the matching `DeferredToolRequests` entry no longer exists (the tool already executed) — mismatch.

### Location

`pydantic_ai/ui/vercel_ai/_utils.py:138` — `iter_tool_approval_responses`:

```python
def iter_tool_approval_responses(messages):
    """Only `approval-responded` parts are matched. `output-denied` parts have
    already been materialized into the message history by `load_messages()` and
    must not be re-processed as deferred results."""
    for msg in messages:
        if msg.role == 'assistant':
            for part in msg.parts:
                if isinstance(part, _APPROVAL_RESPONDED_TYPES) and ...:
                    yield part.tool_call_id, part.approval
```

The docstring explicitly carves out `output-denied` (which `load_messages()` materializes as `ToolReturnPart` with `outcome='denied'`). The same reasoning applies to `output-available`: `load_messages()` materializes it into `ToolCallPart` + `ToolReturnPart` (`_adapter.py:403`), so any co-existing `approval-responded` with the same `toolCallId` is also stale and should not be yielded.

### Lifecycle

1. User asks agent to do something that triggers a `requires_approval=True` tool.
2. Agent returns `DeferredToolRequests`. Frontend shows approve/reject.
3. User approves. Frontend adds `approval-responded`. Server executes, streams `output-available`.
4. Client state now has both parts with same `toolCallId` on the same assistant message.
5. User sends a follow-up message. Client posts full history back.
6. `deferred_tool_results` yields the stale approval → mismatch against current round's (empty) deferred set.

### Open question

Is this:

(a) a bug in `iter_tool_approval_responses` — fix by filtering out `approval-responded` whose `toolCallId` has `output-available`/`output-error` on any assistant message, or

(b) a Vercel client contract issue — clients are expected to transition parts in-place rather than append, and our client is out of spec?

Happy to contribute the fix if you confirm (a).

### Current workaround

We strip the stale parts from the raw JSON body before `build_run_input`:

```python
def _strip_stale_approval_responses(body: bytes) -> bytes:
    # Collect toolCallIds with output-available on any assistant message.
    # Drop approval-responded parts whose toolCallId is in that set.
    ...
```

Related: #4830 / #4831 cover a different part of the same area (`dump_messages()` losing pending-approval state on page reload).

## Minimal, Reproducible Example

Verified against `pydantic-ai-slim==1.84.1` (latest on PyPI at time of filing):

```python
from pydantic_ai.ui.vercel_ai._utils import iter_tool_approval_responses
from pydantic_ai.ui.vercel_ai.request_types import (
    UIMessage,
    ToolApprovalRespondedPart,
    ToolApprovalResponded,
    ToolOutputAvailablePart,
)

# Simulates SDK v6 client state after a tool was approved AND executed
# in a prior round: both parts exist on the same assistant message for
# the same toolCallId.
approval_part = ToolApprovalRespondedPart(
    type='tool-my_tool',
    tool_call_id='call_abc',
    approval=ToolApprovalResponded(id='call_abc', approved=True),
)
output_part = ToolOutputAvailablePart(
    type='tool-my_tool',
    tool_call_id='call_abc',
    output='done',
)
msg = UIMessage(id='m1', role='assistant', parts=[approval_part, output_part])

approvals = list(iter_tool_approval_responses([msg]))
print(approvals)
# Expected: []  (tool has already executed, no pending approval)
# Actual:   [('call_abc', ToolApprovalResponded(id='call_abc', approved=True, reason=None))]
# → fed to approval-resolution logic as pending, mismatches current round's DeferredToolRequests.
```

## Logfire Trace

N/A — bug is in pure message-history processing, not agent run.

## Python, Pydantic AI & LLM client version

- Python: 3.13.12
- Pydantic AI: pydantic-ai-slim 1.84.1 (also reproduces on 1.81.0)
- LLM provider SDK: N/A (bug is in `pydantic_ai.ui.vercel_ai`, upstream of the model call)

abhicris:
Closes #5154.

## The bug

`iter_tool_approval_responses` used to yield every `approval-responded` part on any assistant message, including ones whose `tool_call_id` had already been materialized as a terminal output (`output-available`, `output-error`, or `output-denied`) by `load_messages()`. On the *next* continuation, `VercelAIAdapter.deferred_tool_results` fed that stale approval into pydantic-ai's approval-resolution logic as if pending — mismatching the (now empty) `DeferredToolRequests` set for the current round.

Pure server-side repro from #5154 (against `pydantic-ai-slim==1.84.1`):

```python
from pydantic_ai.ui.vercel_ai._utils import iter_tool_approval_responses
from pydantic_ai.ui.vercel_ai.request_types import (
    UIMessage, ToolApprovalRespondedPart, ToolApprovalResponded, ToolOutputAvailablePart,
)

approval = ToolApprovalRespondedPart(
    type='tool-my_tool',
    tool_call_id='call_abc',
    approval=ToolApprovalResponded(id='call_abc', approved=True),
)
output = ToolOutputAvailablePart(type='tool-my_tool', tool_call_id='call_abc', output='done')
msg = UIMessage(id='m1', role='assistant', parts=[approval, output])

list(iter_tool_approval_responses([msg]))
# before: [('call_abc', ToolApprovalResponded(..., approved=True))]  # ghost entry
# after:  []
```

## Fix

In `pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_utils.py`:

1. Walk all assistant messages once to collect the set of `tool_call_id`s that have a terminal output part:
   - `ToolOutputAvailablePart` / `DynamicToolOutputAvailablePart`
   - `ToolOutputErrorPart` / `DynamicToolOutputErrorPart`
   - `ToolOutputDeniedPart` / `DynamicToolOutputDeniedPart`
2. Walk them again to yield only `approval-responded` parts whose `tool_call_id` is **not** in that set.

The docstring was updated to describe the new rule. Net: +47 / -9 in `_utils.py`. No API change.

Direction confirmed by @DouweM on the issue (2026-04-21): *"not against hardening our integration to better support slightly out of spec frontends."* The specific rule — *"treat `approval-responded` as pending only if there is no terminal tool result for that `toolCallId` anywhere in the assistant history"* — was articulated by @pandego in-thread and matches the docstring's pre-existing rationale for why `output-denied` was already supposed to be excluded.

## Tests

5 new tests in `tests/test_vercel_ai.py`, following the style of the existing `test_tool_approval_ignores_output_denied_parts`:

1. `test_tool_approval_skipped_when_output_available_sibling_exists` — the exact #5154 repro.
2. `test_tool_approval_skipped_when_output_error_sibling_exists` — same rule for `output-error`.
3. `test_tool_approval_skipped_when_output_denied_sibling_exists_same_id` — same rule for `output-denied` on the same `tool_call_id`.
4. `test_tool_approval_mixed_executed_and_pending` — only the pending `tool_call_id` yields when the history has both.
5. `test_tool_approval_terminal_sibling_on_different_message` — match is history-wide, not per-message (covers multi-round continuation where the client re-sends assembled history).

The existing `test_tool_approval_ignores_output_denied_parts` still passes because it uses different `tool_call_id`s across its `output-denied` and `approval-responded` parts.

## Verification

I could not install `pydantic-ai-slim`'s dev deps in the session sandbox to run pytest on this PR locally. Did instead:

- AST-parse check on both modified files (clean).
- Self-contained algorithm check against 7 cases, including a backwards-compat check for the existing `test_tool_approval_ignores_output_denied_parts` behavior. All pass.

CI on this PR will be the authoritative test run; happy to react quickly if anything surfaces.

## Scope / alternatives considered

- **Terminal-set build only from the *same* assistant message** — rejected because real-world reproductions (see the issue thread on 2026-04-23) include cases where the terminal part lives on a prior assistant message and the client re-sends accumulated history. History-wide match is the right scope.
- **Fix at the Vercel client level (option (b) in the issue)** — maintainer noted Vercel's own client transitions parts in-place, but explicitly welcomed hardening the server side regardless. The bug repros purely server-side with the repro script above, so the server-side fix is correct even if out-of-spec clients go away tomorrow.
- **Stripping the stale parts from raw request JSON (the issue author's workaround)** — fixes the symptom but leaves other consumers of `iter_tool_approval_responses` on the wrong semantics.

## Credits

Repro, root-cause analysis, and workaround by @ritwickdsouza in #5154. Fix rule articulated by @pandego in-thread. Direction green-lit by @DouweM.

— [kcolbchain](https://kcolbchain.com) / [Abhishek Krishna](https://abhishekkrishna.com)

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_utils.py
+++ b/pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_utils.py
@@ -135,17 +135,46 @@ _APPROVAL_RESPONDED_TYPES = (
 )
 
 
+_TERMINAL_TOOL_OUTPUT_TYPES = (
+    ToolOutputAvailablePart,
+    ToolOutputErrorPart,
+    ToolOutputDeniedPart,
+    DynamicToolOutputAvailablePart,
+    DynamicToolOutputErrorPart,
+    DynamicToolOutputDeniedPart,
+)
+
+
 def iter_tool_approval_responses(
     messages: list[UIMessage],
 ) -> Iterator[tuple[str, ToolApprovalResponded]]:
-    """Yield `(tool_call_id, approval)` for each responded tool approval in assistant messages.
+    """Yield `(tool_call_id, approval)` for each *pending* tool-approval response in assistant messages.
 
-    Only `approval-responded` parts are matched. `output-denied` parts have
-    already been materialized into the message history by `load_messages()` and
-    must not be re-processed as deferred results.
+    An `approval-responded` part is only yielded if its tool has not yet been
+    executed. A tool counts as already-executed if any assistant message in
+    the history carries a terminal output part for the same `tool_call_id` —
+    that is, `output-available`, `output-error`, or `output-denied` (or their
+    Dynamic variants). `load_messages()` has already materialized those
+    terminal parts as `ToolCallPart` + `ToolReturnPart` in the run history,
+    so re-yielding a sibling `approval-responded` would feed the approval-
+    resolution logic a stale entry that no longer matches any
+    `DeferredToolRequests` in the current round (see #5154).
     """
+    executed_tool_call_ids: set[str] = set()
     for msg in messages:
-        if msg.role == 'assistant':
-            for part in msg.parts:
-                if isinstance(part, _APPROVAL_RESPONDED_TYPES) and isinstance(part.approval, ToolApprovalResponded):
-                    yield part.tool_call_id, part.approval
+        if msg.role != 'assistant':
+            continue
+        for part in msg.parts:
+            if isinstance(part, _TERMINAL_TOOL_OUTPUT_TYPES):
+                executed_tool_call_ids.add(part.tool_call_id)
+
+    for msg in messages:
+        if msg.role != 'assistant':
+            continue
+        for part in msg.parts:
+            if (
+                isinstance(part, _APPROVAL_RESPONDED_TYPES)
+                and isinstance(part.approval, ToolApprovalResponded)
+                and part.tool_call_id not in executed_tool_call_ids
+            ):
+                yield part.tool_call_id, part.approval
--- a/tests/test_vercel_ai.py
+++ b/tests/test_vercel_ai.py
@@ -3061,11 +3061,13 @@ async def test_tool_approval_skipped_when_output_error_sibling_exists():
                 DynamicToolApprovalRespondedPart(
                     tool_name='delete_file',
                     tool_call_id='call_err',
+                    input={'path': 'x.txt'},
                     approval=ToolApprovalResponded(id='call_err', approved=True),
                 ),
                 DynamicToolOutputErrorPart(
                     tool_name='delete_file',
                     tool_call_id='call_err',
+                    input={'path': 'x.txt'},
                     error_text='disk full',
                 ),
             ],
@@ -3089,6 +3091,7 @@ async def test_tool_approval_skipped_when_output_denied_sibling_exists_same_id()
                 DynamicToolApprovalRespondedPart(
                     tool_name='delete_file',
                     tool_call_id='call_denied',
+                    input={'path': 'x.txt'},
                     approval=ToolApprovalResponded(id='call_denied', approved=False),
                 ),
                 DynamicToolOutputDeniedPart(
--- a/tests/test_vercel_ai.py
+++ b/tests/test_vercel_ai.py
@@ -78,6 +78,7 @@ with try_import() as starlette_import_successful:
         DynamicToolInputStreamingPart,
         DynamicToolOutputAvailablePart,
         DynamicToolOutputDeniedPart,
+        DynamicToolOutputErrorPart,
         FileUIPart,
         ReasoningUIPart,
         RegenerateMessage,
@@ -85,6 +86,7 @@ with try_import() as starlette_import_successful:
         TextUIPart,
         ToolApprovalRequested,
         ToolApprovalResponded,
+        ToolApprovalRespondedPart,
         ToolInputAvailablePart,
         ToolInputStreamingPart,
         ToolOutputAvailablePart,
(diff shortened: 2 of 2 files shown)


razchiriac:
Thanks for putting this together. I noticed the PR currently has a number of failing checks.

I’m happy to help with a small follow-up commit if you’d like. I can take a pass at the mechanical failures first, e.g. formatting, lint, type errors, and any test expectation updates directly caused by this change, without changing the intended behavior of the fix.

Would you be open to me contributing a follow-up commit/PR to get the checks green?

razchiriac:
I took a quick look at the failing checks and they appear mostly mechanical rather than design-related. I’m happy to help with a focused follow-up commit that only addresses CI failures: formatting/lint, type issues, and test expectation updates caused by this change.

Would you be open to that?

abhicris:
Thanks for the offer @razchiriac, but I've taken care of the CI failures — they were narrower than "mostly mechanical": three test fixtures were missing the required `input=` field on `DynamicToolApprovalRespondedPart` / `DynamicToolOutputErrorPart` (the existing `DynamicToolOutputDeniedPart` fixture in the same test had it, the new ones didn't). Same root cause for the lint job's pyright errors and the runtime `ValidationError`s in the test matrix.

Fixed in 052e9fc. Pyright is now clean on `tests/test_vercel_ai.py` and the four affected tool-approval tests pass locally. Re-running CI.

PR #5187, opened 2026-04-24
fix: pass max_tokens to max_tokens parameter and add max_completion_tokens setting

### LINKED ISSUE #5186: Pydantic AI `OpenAIChatModel.max_tokens` `max_completion_tokens` handling break on certain providers
### Initial Checks

### Description

**What**:

**Impact**:

**Cause**:

```python

                    model=self.model_name,
                    messages=openai_messages,
                    tools=tools or OMIT,
                    stream=stream,
 <snip>
```

**Why it's actually a bug**:

**How to fix it**:

```Python
```

### Logfire Trace

_No response_

- Python: 3.13
- Pydantic AI: 1.86.1

anishesg:
In `pydantic_ai_slim/pydantic_ai/models/openai.py`, the `_completions_create` method was incorrectly mapping the `max_tokens` model setting to the `max_completion_tokens` API parameter. This meant the `max_tokens` parameter was never sent to the API, and all providers received `max_completion_tokens` instead — breaking any model or provider that only supports `max_tokens` (a very common case with OpenRouter and other compatible providers).

The fix maps `max_tokens` from `ModelSettings` to the `max_tokens` API parameter as intended. A new `max_completion_tokens` field has been added to `ChatModelSettings` for users who specifically need to set the `max_completion_tokens` API parameter — useful for reasoning models where this parameter controls the combined limit of output tokens and reasoning tokens.

This restores correct token limiting behavior for all providers while still allowing specific `max_completion_tokens` usage through the new explicit setting.

- [ ] Any code has been reviewed line-by-line by the human PR author, who stands by it.

Fixes #5186

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/openai.py
+++ b/pydantic_ai_slim/pydantic_ai/models/openai.py
@@ -454,6 +454,16 @@ class OpenAIChatModelSettings(ModelSettings, total=False):
     See the [OpenAI Prompt Caching documentation](https://platform.openai.com/docs/guides/prompt-caching#how-it-works) for more information.
     """
 
+    openai_max_completion_tokens: int
+    """An upper bound for the number of tokens that can be generated for a completion, including visible output tokens and reasoning tokens.
+
+    This is the OpenAI-specific `max_completion_tokens` parameter, which is distinct from `max_tokens`.
+    Use this for OpenAI reasoning models (o-series) where `max_completion_tokens` controls the combined
+    limit of output tokens plus reasoning tokens. For most providers and models, use `max_tokens` instead.
+
+    See the [OpenAI API reference](https://platform.openai.com/docs/api-reference/chat/create#chat-create-max_completion_tokens) for more details.
+    """
+
     openai_continuous_usage_stats: bool
     """When True, enables continuous usage statistics in streaming responses.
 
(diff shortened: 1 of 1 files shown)


pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -848,7 +858,8 @@ async def _completions_create(
                     stream=stream,
                     stream_options=self._get_stream_options(model_settings) if stream else OMIT,
                     stop=model_settings.get('stop_sequences', OMIT),
-                    max_completion_tokens=model_settings.get('max_tokens', OMIT),
```
adtyavrdhn, line 851: Wouldn't this break for people who are already using it with `max_tokens` expecting for it to be normalized and sent via max_completion_tokens?

anishesg: Good point, we should keep backwards compatibility there. Let me add a fallback so it still works with `max_tokens`.

anishesg: done - added a fallback so max_completion_tokens still picks up max_tokens when openai_max_completion_tokens isn't explicitly set. should be backwards compatible now.

anishesg:
@adtyavrdhn Fixed the dual-send issue flagged in review. The previous fallback approach sent both `max_tokens` and `max_completion_tokens` simultaneously when only `max_tokens` was set, which the OpenAI API rejects.

New behavior is a clean separation:
- `max_tokens` maps to `max_tokens` (fixes #5186 for OpenRouter and other providers)
- `openai_max_completion_tokens` maps to `max_completion_tokens` (explicit opt-in for reasoning models)

No cross-mapping, no dual-send. Users on reasoning models just need to switch to `openai_max_completion_tokens` in their settings. This is a minor migration but avoids the ambiguity that caused the original bug.

anishesg:
@adtyavrdhn Closing this — the core issue is that `max_tokens` and `max_completion_tokens` need model/provider-aware routing (OpenRouter needs `max_tokens`, OpenAI reasoning models need `max_completion_tokens`, and the API rejects both simultaneously). No single static mapping works for all cases. This needs a broader design decision from the maintainers rather than a simple field remap.

PR #5188, opened 2026-04-24
fix(toolsets): warn when prepare callbacks return `None`

### LINKED ISSUE #5177: `prepare_tools` callback returning `None` silently strips all tools
### Initial Checks

### Description

# Description

**Upfront: this isn't strictly a bug.** The contract "`prepare_tools` callback returning `None` disables all tools for that step" is documented in [`docs/tools-advanced.md#prepare-tools`](https://ai.pydantic.dev/tools-advanced/#prepare-tools) and pinned by `tests/test_capabilities.py::test_prepare_tools_none_disables_all`. I'm opening this because the contract is a silent footgun. I hit it in production on code that wasn't sloppy, and I think a one-line runtime warning from `PreparedToolset.get_tools` would catch the same class of mistake for other users without changing behavior or touching the documented contract.

### Details

`None` means opposite things at two layers that look identical on read:

- `Agent(prepare_tools=None)` → no `PreparedToolset` wrapping, every tool stays visible.
- A callback registered as `prepare_tools` **returning** `None` → `PreparedToolset` is wrapped, the `None or [] = []` coalesce in `PreparedToolset.get_tools` strips every tool.

Both `None`s sit inside `ToolsPrepareFunc | None`. The outer means "no hook", the inner means "hook says zero tools". Same type, same syntax, opposite effect.

It's also asymmetric with the per-tool `prepare` hook, where returning the passed `ToolDefinition` unchanged keeps the tool. Transferring intuition from there leads straight to `return None` when the developer means "passthrough".

The failure is silent: the run completes, the model answers in prose, no exception, no log, nothing in tracing that says "your agent just ran with zero tools". In my case the regression shipped and I only noticed because a class of requests stopped producing tool calls for a whole session.

### How I got caught

I had a base class exposing an optional hook that subclasses override:

```python
def _get_prepare_tools(self) -> PrepareToolsFunc | None:
    """Return the Pydantic AI prepare_tools callback, or None.

    Returning None means "no filter" — the agent exposes every tool unchanged.
    """
    return None
```

The docstring is correct at the hook level. Base returns `None`, the agent never builds a `PreparedToolset`. A subclass registered a real callback. Months later I short-gated the callback's body with `return None`, transferring "no filter, all tools pass through" from the hook into the callback's return value. That flipped the meaning to "expose zero tools" and the agent silently lost tool visibility on every request.

### Proposed change

Emit a `warnings.warn` from `PreparedToolset.get_tools` when the callback's return value is `None`. Exact patch on [`pydantic_ai_slim/pydantic_ai/toolsets/prepared.py`](https://github.com/pydantic/pydantic-ai/blob/main/pydantic_ai_slim/pydantic_ai/toolsets/prepared.py):

```python
import warnings
...
result = self.prepare_func(ctx, original_tool_defs)
if inspect.isawaitable(result):
    result = await result
if result is None:
    warnings.warn(
        f'prepare_tools callback {self.prepare_func.__qualname__!r} returned None — '
        'this disables all tools for this step. Return `[]` for an explicit empty '
        'tool list, or `tool_defs` to pass the definitions through unchanged.',
        UserWarning,
        stacklevel=2,
    )
prepared_tool_defs_by_name = {tool_def.name: tool_def for tool_def in (result or [])}
```

Ran it locally against the reproducer below, the warning surfaces on stderr, `test_prepare_tools_none_disables_all` still passes (it asserts the empty-tools result, not the absence of warnings), and users who legitimately want the documented "disable all tools" behavior can silence the warning with a standard `warnings.filterwarnings` call.

### Alternatives considered

- **Change semantics so `None` means passthrough.** Breaks `test_prepare_tools_none_disables_all` and the documented contract, not viable.
- **Docs-only.** The contract is already documented. What catches people is intuition transfer from the hook level or from per-tool `prepare`, not missing docs. A runtime signal interrupts the wrong mental model where docs don't.
- **Deprecate `None` → "disable all" over a major version.** Heavier lift, needs buy-in on timing. The warning above is the minimum that would have saved me without closing any door.

Happy to open the PR (warning + a short `!!! warning` admonition in `docs/tools-advanced.md#prepare-tools`) if this direction sounds reasonable. Preferred to align here first per CONTRIBUTING.md.

```Python
import asyncio


def get_weather(city: str) -> str:
    """Get the weather for a city."""
    return f"Weather in {city}: sunny"


async def prep_noop(ctx: RunContext, tool_defs: list[ToolDefinition]):
    # intent: "no filter applied, pass everything through"
    return None


# Case A — no prepare_tools: LLM calls get_weather
a = Agent('openai:gpt-4.1-mini', tools=[get_weather])
print(asyncio.run(a.run('What is the weather in Rome?')).output)

# Case B — prepare_tools returning None: LLM gets NO tools, replies text-only
b = Agent('openai:gpt-4.1-mini', tools=[get_weather], prepare_tools=prep_noop)
print(asyncio.run(b.run('What is the weather in Rome?')).output)

# Case C — prepare_tools returning tool_defs: LLM calls get_weather
async def prep_passthrough(ctx: RunContext, tool_defs: list[ToolDefinition]):
    return tool_defs
c = Agent('openai:gpt-4.1-mini', tools=[get_weather], prepare_tools=prep_passthrough)
print(asyncio.run(c.run('What is the weather in Rome?')).output)

# A and C produce `"Weather in Rome: sunny"`. B produces a text-only reply like `"I don't have access to real-time weather data..."` with no tool call and no warning on stderr today.
```

### Logfire Trace

_No response_

- Python: 3.11
- Pydantic AI: 1.65.0 (behavior identical on `main` - `pydantic_ai_slim/pydantic_ai/toolsets/prepared.py`, `None or []` coalesce)
- LLM provider SDK: openai

dfm88:
- Closes #5177
- v2 card: [`21-prepare-tools-none`](https://github.com/pydantic/pydantic-ai-notes/blob/main/david/v2-cards/deprecate-before/21-prepare-tools-none.md)

### Summary

Warn when a registered prepare callback returns `None` instead of silently hiding every tool for that step.

This preserves the current 1.x runtime semantics: returning `None` from a prepare callback still means "disable all tools/output tools for this step", but it now emits a `UserWarning` telling users to return `[]` for an explicit empty list or `tool_defs` for passthrough behavior.

### Compatibility fix

The v2 constructor migration now also handles explicit `Agent(prepare_tools=None)` and `Agent(prepare_output_tools=None)` correctly.

Those deprecated kwargs still emit `PydanticAIDeprecationWarning` because the deprecated API was used, but `None` is treated as "no callback installed" and no `PrepareTools(None)` / `PrepareOutputTools(None)` capability is created. That keeps the old constructor behavior and avoids the crash noted in the issue discussion.

### Changelog

Warn when `prepare_tools` / `prepare_output_tools` callbacks return `None`, while preserving constructor compatibility for explicit `prepare_tools=None` / `prepare_output_tools=None`.

### What changed

- Emit a `UserWarning` when `PreparedToolset`, `PrepareTools`, or `PrepareOutputTools` receives `None` back from its prepare callback.
- Document the `None` footgun across the prepare-tools docs surfaces.
- Add tests for callback-returned `None`, explicit empty-list returns, and the `prepare_tools=None` / `prepare_output_tools=None` constructor compatibility path.

### Checks

- `uv run pytest tests/v2/test_agent_ctor_kwargs_deprecation.py -k 'none_kwarg'`
- `uv run ruff format --check pydantic_ai_slim/pydantic_ai/_utils.py tests/v2/test_agent_ctor_kwargs_deprecation.py`
- `uv run ruff check tests/v2/test_agent_ctor_kwargs_deprecation.py`
- `PYRIGHT_PYTHON_IGNORE_WARNINGS=1 uv run pyright tests/v2/test_agent_ctor_kwargs_deprecation.py`

### Checklist


tests/test_capabilities.py
```diff
@@ -4804,9 +4805,30 @@ def model_fn(messages: list[ModelMessage], info: AgentInfo) -> ModelResponse:
...
             return 'result'  # pragma: no cover
 
-        result = await agent.run('hello')
+        with pytest.warns(UserWarning, match=r'prepare callback .*returned None'):
+            result = await agent.run('hello')
         assert result.output == 'tools: []'
 
+    async def test_prepare_tools_empty_list_does_not_warn(self):
+        """Returning [] is the explicit 'disable all' alternative and must not emit the warning."""
+        from pydantic_ai.capabilities import PrepareTools
```
adtyavrdhn: Please move the import to the top.

docs/tools-advanced.md
```diff
@@ -255,6 +255,9 @@ In addition to per-tool `prepare` methods, you can also define an agent-wide `pr
 
 The `prepare_tools` function should be of type [`ToolsPrepareFunc`][pydantic_ai.tools.ToolsPrepareFunc], which takes the [`RunContext`][pydantic_ai.tools.RunContext] and a list of [`ToolDefinition`][pydantic_ai.tools.ToolDefinition], and returns a new list of tool definitions (or `None` to disable all tools for that step).
 
+!!! warning
+    Returning `None` from the callback disables **all** tools for that step, it is not a "pass through unchanged" shortcut. To keep every tool as-is, return the `tool_defs` argument; to expose no tools intentionally, return `[]`. A `UserWarning` is emitted when the callback returns `None` so the silent strip is surfaced during development; silence it with `warnings.filterwarnings` if the behavior is intentional.
```
adtyavrdhn: Since this warning is emitted from `PreparedToolset`, can we document it wherever that wrapper is exposed too? This now also applies to `prepare_output_tools` and `toolset.prepared(...)`, but only the `prepare_tools` section was updated. At minimum, `docs/output.md` should mention that returning `None` emits a `UserWarning` and that returning `[]` is the quiet explicit way to expose no output tools; the toolset docs may need the same note or a cross-link.

dfm88: Two follow-ups in the latest pushes:

**Docs cross-links** (commit 0a4e4954): addressed the comment about
documenting the warning wherever `PreparedToolset` is exposed. Added small
`!!! warning` admonitions cross-linking back to the canonical discussion in
`tools-advanced.md#prepare-tools`:

- `docs/output.md` (`prepare_output_tools` section)
- `docs/toolsets.md` (`preparing-tool-definitions` section, covers `AbstractToolset.prepared()`)
- `docs/capabilities.md` (`PrepareTools and PrepareOutputTools` section, also here I added preventively since these capabilities are also user-facing surfaces of the same wrapper, I can drop it if you'd rather keep the note scoped)

**Capability-layer warning mirror** (commit 7bbbea8f) after rebasing onto main, where prepare_tools capability was edited and tests was failing (`capabilities/prepare_tools.py:88-98` normalizes `None` → `[]` since these bypass the `PreparedToolset.get_tools` so the warning previously added
there never fires for the `Agent(prepare_tools=)` / `Agent(prepare_output_tools=)`
/ `PrepareTools` / `PrepareOutputTools` paths. Unfortunately this means that I had to duplicate the warning generation code also in `pydantic_ai_slim/pydantic_ai/capabilities/prepare_tools.py` (probably the cleanest way instead of having a specific guard that would have to be manually called after both `toolsets/prepared.py` and `capabilities/prepare_tools.py`

adtyavrdhn (approved):
GTG

pydantic_ai_slim/pydantic_ai/_warnings.py
```diff
@@ -8,3 +11,15 @@ class PydanticAIDeprecationWarning(UserWarning):
     deprecations are visible by default at runtime, following the approach
     described in https://sethmlarson.dev/deprecations-via-warnings-dont-work-for-python-libraries.
     """
+
+
+def warn_on_prepare_callback_returned_none(prepare_func: Callable[..., object]) -> None:
+    """Warn that a prepare callback returned `None`."""
+    name = getattr(prepare_func, '__name__', str(prepare_func))
```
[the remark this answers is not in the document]
dsfaccini: David's AICA here: applied in 5a4ae5e4d — swapped to your suggested defensive chain (`__qualname__` → `__name__` → `str(...)`). Confirmed the issue #5177 callback is exactly the nested-in-method case where `__qualname__` actually locates it.

docs/output.md
```diff
@@ -323,7 +323,7 @@ If you'd like to change the name of the output tool, pass a custom description t
...
+To dynamically modify or filter the available output tools during an agent run, you can define an agent-wide `prepare_output_tools` function that will be called ahead of each step of a run. This function should be of type [`ToolsPrepareFunc`][pydantic_ai.tools.ToolsPrepareFunc], which takes the [`RunContext`][pydantic_ai.tools.RunContext] and a list of [`ToolDefinition`][pydantic_ai.tools.ToolDefinition], and returns a new list of tool definitions. It follows the same return-value rules as the [`prepare_tools` function](tools-advanced.md#prepare-tools): return the `tool_defs` argument to keep every output tool as-is, `[]` to expose no output tools intentionally, or `None` to disable all output tools for that step and emit a `PydanticAIDeprecationWarning` (will raise in v2.0).
```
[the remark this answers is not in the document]
dsfaccini: David's AICA here: applied in 5a4ae5e4d for both `docs/output.md` and `docs/toolsets.md` — dropped the full return-value rules in favor of `"following the same return-value rules ... as [`prepare_tools`](...)"`, matching the `capabilities.md` shape. Kept an explicit `"`None`-return deprecation (will raise in v2.0)"` breadcrumb at each surface so users landing on those pages still see the forward signal without click-through.

docs/output.md
```diff
@@ -323,7 +323,7 @@ If you'd like to change the name of the output tool, pass a custom description t
...
+To dynamically modify or filter the available output tools during an agent run, you can define an agent-wide `prepare_output_tools` function that will be called ahead of each step of a run. This function should be of type [`ToolsPrepareFunc`][pydantic_ai.tools.ToolsPrepareFunc], which takes the [`RunContext`][pydantic_ai.tools.RunContext] and a list of [`ToolDefinition`][pydantic_ai.tools.ToolDefinition], and returns a new list of tool definitions, following the same return-value rules and `None`-return deprecation (will raise in v2.0) as the [`prepare_tools` function](tools-advanced.md#prepare-tools). This is analogous to `prepare_tools` for non-output tools.
```
adtyavrdhn: I don't think we should phrase it this way that it will raise in v2?

docs/toolsets.md
```diff
@@ -366,9 +366,9 @@ _(This example is complete, it can be run "as is")_
...
+[`PreparedToolset`][pydantic_ai.toolsets.PreparedToolset] lets you modify the entire list of available tools ahead of each step of the agent run using a user-defined function that takes the agent [run context][pydantic_ai.tools.RunContext] and a list of [`ToolDefinition`s][pydantic_ai.tools.ToolDefinition] and returns the tool definitions to expose for that step.
 
-This is the toolset-specific equivalent of the [`prepare_tools`](tools-advanced.md#prepare-tools) argument to `Agent` that prepares all tool definitions registered on an agent across toolsets.
+This is the toolset-specific equivalent of the [`prepare_tools`](tools-advanced.md#prepare-tools) argument to `Agent` that prepares all tool definitions registered on an agent across toolsets, following the same return-value rules and `None`-return deprecation (will raise in v2.0).
```
adtyavrdhn: Same here?

pydantic_ai_slim/pydantic_ai/_warnings.py
```diff
@@ -8,3 +11,15 @@ class PydanticAIDeprecationWarning(UserWarning):
     deprecations are visible by default at runtime, following the approach
     described in https://sethmlarson.dev/deprecations-via-warnings-dont-work-for-python-libraries.
     """
+
+
+def warn_on_prepare_callback_returned_none(prepare_func: Callable[..., object]) -> None:
+    """Warn that a prepare callback returned `None`."""
+    name = getattr(prepare_func, '__qualname__', None) or getattr(prepare_func, '__name__', None) or str(prepare_func)
```
adtyavrdhn: I am not sure what this `__qualname__` is, I wanted to use something pre existing in the repo?

tests/test_capabilities.py
```diff
@@ -5946,7 +5968,8 @@ def model_fn(messages: list[ModelMessage], info: AgentInfo) -> ModelResponse:
             capabilities=[PrepareOutputTools(disable_all)],
         )
 
-        result = await agent.run('hello')
+        with pytest.warns(PydanticAIDeprecationWarning, match=r'will raise in v2\.0'):
```
adtyavrdhn: Is this related to the change at all? If we didn't intend to in this test we should just return []

tests/test_capabilities.py
```diff
@@ -5797,9 +5797,31 @@ def model_fn(messages: list[ModelMessage], info: AgentInfo) -> ModelResponse:
         def my_tool() -> str:
             return 'result'  # pragma: no cover
 
-        result = await agent.run('hello')
+        with pytest.warns(PydanticAIDeprecationWarning, match=r'will raise in v2\.0'):
```
adtyavrdhn: Here too?

PR #4942, opened 2026-04-02
Add ThreadExecutor capability and Agent.using_thread_executor()

DouweM:
## Summary

- Add `ThreadExecutor` capability that uses `wrap_run` to set a custom thread executor per-agent, preventing thread accumulation in long-running servers (e.g. FastAPI)
- Add `Agent.using_thread_executor()` static context manager for setting a global thread executor
- Both powered by a `ContextVar` in `_utils.py` with explicit `copy_context()` for Python 3.10/3.11 context var propagation

Closes #4883

## Test plan

- [x] Unit tests for `using_thread_executor` context manager (custom executor usage, context var propagation, `disable_threads` priority)
- [x] Integration tests for `ThreadExecutor` capability and `Agent.using_thread_executor()` static method with `FunctionModel` and sync tools
- [x] All pre-commit hooks pass (format, lint, typecheck)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #222, opened 2024-12-12
fix double anthropic reference

montasaurus:
Fixes a small typo in #193

PR #5192, opened 2026-04-24
Add support for MCP background tasks (SEP-1686)

### LINKED ISSUE #4266: FastMCPToolset: add support for MCP background tasks (SEP-1686)
## Summary

`FastMCPToolset` does not support MCP background tasks ([SEP-1686](https://gofastmcp.com/servers/tasks)). When a FastMCP server tool uses `task=TaskConfig(mode="required")`, calling it via `FastMCPToolset` fails with a generic `ToolError("Error calling tool ...")` because the client never sends task metadata.

## Current behavior

`FastMCPToolset.call_tool()` always calls `self.client.call_tool(name, arguments)` without passing `task=True`:

```python
# pydantic_ai/toolsets/fastmcp.py, line 146
call_tool_result: CallToolResult = await self.client.call_tool(name=name, arguments=tool_args)
```

The fastmcp `Client.call_tool()` defaults to `task=False`, so:

- **`mode="optional"` tools**: work, but only via synchronous fallback -- background execution and progress tracking are never used.
- **`mode="required"` tools**: fail immediately. The server returns `-32601: requires task-augmented execution` because no task metadata was sent.

## Expected behavior

`FastMCPToolset` should support MCP task-augmented execution so that:

1. Tools with `task=TaskConfig(mode="required")` work correctly.
2. Tools with `task=TaskConfig(mode="optional")` can benefit from background execution and progress tracking.

## Reproduction

**Server** (`server.py`):
```python
from fastmcp import FastMCP
from fastmcp.server.tasks import TaskConfig

mcp = FastMCP("Test")

@mcp.tool(task=TaskConfig(mode="required"))
async def long_running_task() -> int:
    import asyncio
    await asyncio.sleep(5)
    return 42

if __name__ == "__main__":
    mcp.run()
```

**Client** (pydantic-ai agent):
```python
from pydantic_ai import Agent
from pydantic_ai.toolsets.fastmcp import FastMCPToolset

agent = Agent("openai:gpt-4o", toolsets=[FastMCPToolset("http://localhost:8000/mcp")])
result = await agent.run("Run the long running task")
# ToolError: Error calling tool 'long_running_task'
```

## Suggested approach

The fastmcp `Client` already supports the full task protocol. `FastMCPToolset.call_tool()` could:

1. Call `self.client.call_tool(name=name, arguments=tool_args, task=True)` -- this returns a `ToolTask`.
2. Await `tool_task.result()` to get the final `CallToolResult`.
3. The `ToolTask` handles both scenarios transparently:
   - If the server accepts background execution: polls/waits for completion.
   - If the server declines (graceful degradation): returns the immediate result.

A minimal change would look like:

```python
async def call_tool(self, name, tool_args, ctx, tool):
    async with self:
        try:
            tool_task = await self.client.call_tool(name=name, arguments=tool_args, task=True)
            call_tool_result: CallToolResult = await tool_task.result()
        except ToolError as e:
            if self.tool_error_behavior == 'model_retry':
                raise ModelRetry(message=str(e)) from e
            else:
                raise e

    if call_tool_result.structured_content:
        return call_tool_result.structured_content
    return _map_fastmcp_tool_results(parts=call_tool_result.content)
```

This is fully backwards-compatible: `task=True` with graceful degradation means tools that don't support tasks continue to work as before.

## References

- [FastMCP Background Tasks documentation](https://gofastmcp.com/servers/tasks)
- [MCP SEP-1686 specification](https://modelcontextprotocol.io/specification/2025-06-18/server/utilities/tasks)
- fastmcp `Client.call_tool()` already supports `task=True` with `ToolTask` result type

Alex-Resch:
Adds [SEP-1686](https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/tasks) task-augmented execution support to `MCPToolset`. Tools whose server config declares `task=TaskConfig(mode='required'|'optional')` now run via the durable, cancelable, pollable task path — `client.call_tool(task=True)` → `tool_task.result()` — instead of failing with `-32601: requires task-augmented execution`.

## Usage

No client-side configuration is needed — `MCPToolset` honors the server's per-tool `execution.taskSupport` declaration automatically:

| `execution.taskSupport` | Behavior |
| --- | --- |
| `"required"` | Always calls with `task=True` (the SEP requires it). |
| `"optional"` | Always calls with `task=True` to opt in to durability, cancellation, and progress notifications. |
| `"forbidden"` or absent | Calls normally. |

```python
from pydantic_ai import Agent
from pydantic_ai.mcp import MCPToolset

agent = Agent('openai:gpt-5', toolsets=[MCPToolset('http://localhost:8000/mcp')])
```

## What's in scope

- `MCPToolset.get_tools` reads `mcp_tool.execution.taskSupport` and exposes a selector-friendly `task: bool` on `ToolDefinition.metadata`.
- `MCPToolset.call_tool` routes through the task path iff the tool's `task` metadata is set (i.e. the server declared `required`/`optional`).
- `MCPToolset.direct_call_tool` gains a keyword-only `use_task=False` low-level escape hatch for callers who invoke without a `ToolDefinition`.
- `process_tool_call` gets a `partial(direct_call_tool, use_task=...)` so a custom wrapper preserves the task path without needing to know about it.
- Docs: new "Background tasks" section in `docs/mcp/client.md`.

## Notes

- Built on the new `MCPToolset` (#5325), not the now-deprecated `FastMCPToolset` — the original community PR targeted FastMCPToolset; relocated here as part of the takeover.
- Server-side task execution is purely the MCP server's decision, so there's **no client-side opt-in** and no coupling to `BackgroundTools` (pydantic-ai-harness#222) — that capability addresses a different layer (client-side fire-and-forget asyncio tasks). Overriding the server's decision (forcing or skipping the task path) is left to a future generic capability.

Closes #4266.
Co-authored with the original community PR by @Alex-Resch (#5192).

### Checklist


docs/mcp/fastmcp-client.md
```diff
@@ -90,3 +90,44 @@ agent = Agent('openai:gpt-5.2', toolsets=[toolset])
...
+
+The toolset handles task execution transparently:
+- **`mode="required"`**: Tools that must run as background tasks will work correctly
+- **`mode="optional"`**: Tools will use background execution when supported, with graceful fallback
+- The agent waits for task completion and receives the final result
+
+!!! note
+    Background task support requires the `pydocket` package. Install it with `pip install pydocket` or `pip install "pydantic-ai-slim[fastmcp-tasks]"`.
+
+```python {test="skip"}
```
dsfaccini: lets try to not skip this one

DouweM:
@Alex-Resch Thanks for working on this! I've been making progress on a different flavor of background/async tools as well: 
- https://github.com/pydantic/pydantic-ai/pull/4980
- pydantic/pydantic-ai-harness#222

Any opinions on how the features should interact?

Alex-Resch:
Hey @DouweM, thanks for the review! I think I could change my PR so that — similar to what you did in #222 — FastMCPToolset wouldn’t await the task result anymore. Instead, the agent would get an immediate ack like “task started” and could work on other things in the meantime. Once the server-side task completes, the result would get enqueued as a follow-up via #4980. This would be opt-in, so the current blocking behavior stays the default.
Does that sound like the direction you had in mind? Or am I misunderstanding something here?

DouweM:
@Alex-Resch Yep that's exactly what I was thinking. I wanted to hear from someone who actually needs the feature and is more aware of what the MCP background task SEP prescribes, what their expected interaction (and opt-in vs opt-out) behavior would be :)

I believe it would be possible to use "background tools" for MCP tasks without any changes to this particular code, since this sets an `execution` field on `ToolDefinition.metadata`, meaning that opting in to the harnesses's `BackgroundTools` behavior becomes adding a new capability `SetToolMetadata(tools={'execution': {'taskSupport': 'required'}}, metadata={'background': True})`, so that all the background task tools also get the `background` flag that `BackgroundTools` looks for. I suppose you would then also need to target `'taskSupport': 'optional'`, which is a bit annoying. I think it'd be worth adding a top-level metadata field in this PR that indicates whether the MCP toolset will be running it in task or not (i.e. `True` for `required`, `True`/`False` for optional depending on whether the library was installed). Then it'll be `SetToolMetadata({'task': True}, {'background': True})`.

Could you verify that this PR works as expected with those other 2, by bringing the branches together in a local project?

DouweM:
@Alex-Resch Do we need pydocket? Is there any way we can do this without new dependencies? Perhaps with an way to specify a different executor?

Note that there's also https://github.com/pydantic/pydantic-ai/pull/4942

Alex-Resch:
Hey @DouweM, I pulled all branches into one and all current relevant tests pass. So this works.

For the pydocket question: pydantic-ai's code doesn't actually need pydocket, because it's a server-side requirement. Only the server needs it, the client doesn't.

I verified this locally with two separate venvs:

- Server venv with `fastmcp[tasks]` (pydocket installed): runs a `@mcp.tool(task=TaskConfig(mode="required"))` `long_task` over streamable-http on port 8000.
- Client venv with only base `fastmcp` (no pydocket — `import docket` raises `ModuleNotFoundError`): calls `client.call_tool("long_task", {}, task=True)`, gets a `ToolTask` back with `returned_immediately=False`, then `await task.result()` returns the actual server result.

So the task path runs end-to-end with no pydocket on the client side. The earlier ImportError I had seen came from the in-process test fixture registering a server-side `TaskConfig`, not from the client call:

​```
ImportError: FastMCP background tasks require the `tasks` extra.
Install with: pip install 'fastmcp[tasks]'.
(Triggered by `task=True` on function 'task_required_tool')
​```

That error fires from `fastmcp/server/dependencies.py` when the server-side decorator runs, which is why the in-process fixture trips on it but a real client doesn't. So I think pydocket, the `pydocket_installed` check, the `UserError` for `mode="required"`, and the `fastmcp-tasks` dependency aren't needed.

Is that what you asked for? Let me know if I'm missing something.

DouweM:
@Alex-Resch Yep sounds reasonable, can you make those updates to the PR please?

If `task == required`, what do you think the right behavior should be, if not an error? I haven't read https://modelcontextprotocol.io/seps/1686-tasks to see if it prescribes something specific. But it seems to me that just getting the task and then immediately doing a blocking await (blocking from the agent loop perspective), wouldn't match the intent. Should we error unless the user is using Background Tools mode from pydantic/pydantic-ai-harness#222?

Alex-Resch:
Hey DouweM, I'd love to implement those changes!

I read the SEP and it doesn't prescribe any specific client behavior for this case — `mode="required"` is a fastmcp-specific concept, the SEP doesn't know about it. If the server doesn't know a task it just ignores the task metadata.
I think throwing an error for `mode="required"` is the best approach here — the server explicitly requires background execution, so the user should know they need to activate BackgroundTools.

DouweM:
@Alex-Resch OK that makes sense to me then!

Alex-Resch:
@DouweM I updated the PR with your suggestions.

I verified end-to-end locally by combining this PR with #4980 and pydantic-ai-harness#222 — agent gets the immediate ack, the task runs in the background on the server, and the result comes back as a follow-up message. I confirmed that everything works as expected.

DouweM:
@Alex-Resch Thanks, I'll have a look next week!

DouweM:
FYI, I'll likely get https://github.com/pydantic/pydantic-ai/pull/5325 in, and then this.

DouweM:
#5325 has been merged, I'll take this PR from here!

docs/mcp/client.md
```diff
@@ -354,6 +354,50 @@ agent = Agent('openai:gpt-5.2', toolsets=[server])
 
 MCP tools can include metadata that provides additional information about the tool's characteristics, which can be useful when [filtering tools][pydantic_ai.toolsets.FilteredToolset]. The `meta`, `annotations`, and `output_schema` fields can be found on the `metadata` dict on the [`ToolDefinition`][pydantic_ai.tools.ToolDefinition] object that's passed to filter functions.
```
DouweM: The bit about `output_schema` is out of date now, right? ToolDef just has a `return_schema` field now I believe.

docs/mcp/client.md
```diff
@@ -354,6 +354,50 @@ agent = Agent('openai:gpt-5.2', toolsets=[server])
...
+
+```python {title="background_task_client.py"}
+from pydantic_ai import Agent
+from pydantic_ai.mcp import MCPToolset
+
+toolset = MCPToolset('http://localhost:8000/mcp')
+agent = Agent('openai:gpt-5', toolsets=[toolset])
+```
+
+See [Streamable HTTP Client](#streamable-http-client) for the full `agent.run` plumbing.
```
DouweM: Drop this

uv.lock
DouweM, line 1: Help me understand why adding pydocket to dev dependencies, results in -109 / +25 on uv.lock. Seems like more than we need.

DouweM (changes requested):

DouweM:
@Alex-Resch Thanks Alex!

PR #5195, opened 2026-04-24
fix(deepseek): add support for `deepseek-v4-flash` and`deepseek-v4-pro`

### LINKED ISSUE #5193: `DeepSeekProvider` missing support for `deepseek-v4-flash` and `deepseek-v4-pro`: users hit 400 on tool-based structured output
### Initial Checks

### Description

**What**:

`DeepSeekProvider` only recognizes `deepseek-chat` and `deepseek-reasoner` in its `DeepSeekModelName` literal and in the `openai_supports_tool_choice_required` gating inside `provider.model_profile`. The new explicit DeepSeek V4 model IDs (`deepseek-v4-flash` and `deepseek-v4-pro`) fall through to the default `OpenAIModelProfile` and keep `openai_supports_tool_choice_required=True`. Users must migrate to these explicit IDs before the [2026-07-24 15:59 UTC sunset](https://api-docs.deepseek.com/news/news260424) of the legacy aliases.

As a result, the moment a user requests tool-based structured output with a V4 model, pydantic-ai sends `tool_choice='required'`. DeepSeek's server routes V4-Flash through its reasoner endpoint, because V4 defaults to thinking mode per the [DeepSeek Thinking Mode guide](https://api-docs.deepseek.com/guides/thinking_mode), and the request is rejected with:

```
pydantic_ai.exceptions.ModelHTTPError:
  status_code: 400, model_name: deepseek-v4-flash,
  body: {'message': 'deepseek-reasoner does not support this tool_choice',
         'param': None,
         'code': 'invalid_request_error'}
```

The error message points at `deepseek-reasoner` even though the request targeted `deepseek-v4-flash`. This makes the issue especially confusing to debug from the user side, because it looks like pydantic-ai is routing to the wrong model.

**Impact**:

Every pydantic-ai user on V4 hits this the first time they pass a Pydantic `output_type` to an `Agent`. A workaround exists: users can pin to the `deepseek-chat` alias, which transparently routes to V4-Flash non-thinking mode server-side. However, that alias disappears on 2026-07-24, which gives the ecosystem a hard runway of roughly 12 weeks.

**Cause**:

Two gaps in `pydantic_ai_slim/pydantic_ai/providers/deepseek.py`:

```python
# Line 25: the literal is missing V4 model names.
DeepSeekModelName = Literal['deepseek-chat', 'deepseek-reasoner']

# Line 58: only deepseek-reasoner is opted out of tool_choice='required'.
openai_supports_tool_choice_required=(model_name != 'deepseek-reasoner'),
```

And a related gap in `pydantic_ai_slim/pydantic_ai/profiles/deepseek.py`:

```python
def deepseek_model_profile(model_name: str) -> ModelProfile | None:
    is_r1 = 'r1' in model_name
    return ModelProfile(
        ignore_streamed_leading_whitespace=is_r1,
        supports_thinking=is_r1,
        thinking_always_enabled=is_r1,
    )
```

Setting `supports_thinking=True` for V4 is needed so that pydantic-ai's thinking-part handling at `models/__init__.py:712` treats V4's `reasoning_content` as first-class. This matters for the multi-turn tool loops that the PR #3536 round-trip work was designed to support.

**Proposed shape**:

The fix mirrors the existing `deepseek-reasoner` precedent and should be bounded to three files. In `providers/deepseek.py`, extend the `DeepSeekModelName` literal with the two V4 IDs and add them to the `openai_supports_tool_choice_required` exclusion alongside `deepseek-reasoner`. In `profiles/deepseek.py`, detect V4 models and enable `supports_thinking` while leaving `thinking_always_enabled` off, since V4 has runtime control over thinking via `reasoning_effort`. Corresponding offline tests in `tests/providers/test_deepseek.py` would cover the V4 profile shape and include a regression guard on the `deepseek-chat` path. Happy to produce the PR once we've agreed on the direction.

**Design trade-off to flag up-front**:

V4 has dual modes at runtime (thinking and non-thinking) via the `reasoning_effort` parameter, but the profile is static per model name. I am proposing the **conservative policy**, which sets `openai_supports_tool_choice_required=False` for V4, because it matches the existing precedent for `deepseek-reasoner`. This is correct for V4's default (thinking mode). It is over-restrictive for users who force non-thinking, who would need to stick with the `deepseek-chat` alias, or we would need runtime-aware profile logic. Runtime-aware logic is a bigger scope and a separate discussion. I am happy to pivot to that direction if maintainers prefer it.

**Why not also set `openai_supports_reasoning=True`**:

This is deliberately left off. DeepSeek's server silently ignores `temperature` when thinking is enabled and does not reject the request. Client-side sampling-param dropping via `_drop_sampling_params_for_reasoning` (`openai.py:831`) would therefore surface a spurious `UserWarning` for a situation that is already handled server-side. This is different from OpenAI's GPT-5 family, where the API actively rejects the combination.

```Python
from pydantic_ai import Agent


class Answer(BaseModel):
    text: str


model = OpenAIChatModel(
    'deepseek-v4-flash',
    provider=DeepSeekProvider(),
)
agent = Agent(model, output_type=Answer)

result = agent.run_sync('Say hi.')
# -> pydantic_ai.exceptions.ModelHTTPError:
#    status_code: 400, model_name: deepseek-v4-flash,
#    body: {'message': 'deepseek-reasoner does not support this tool_choice',
#           'type': 'invalid_request_error',
#           'param': None,
#           'code': 'invalid_request_error'}
```

### Logfire Trace

_No response_

- Python: 3.12.13
- Pydantic AI: 1.86.1
- LLM provider SDK: openai 2.32.0

SuperMarioYL:
- Closes #5193

### What

Add support for the explicit DeepSeek V4 model IDs (`deepseek-v4-flash` and `deepseek-v4-pro`) that replace the `deepseek-chat` alias on 2026-07-24.

**Three changes across two files:**

`providers/deepseek.py`:
- Extend `DeepSeekModelName` literal with `deepseek-v4-flash` and `deepseek-v4-pro` for type safety
- Set `openai_supports_tool_choice_required=False` for V4 models — V4 routes through the reasoner endpoint, which rejects `tool_choice=required` with a 400

`profiles/deepseek.py`:
- Detect V4 models and set `supports_thinking=True` so that `reasoning_content` in V4 responses is treated as a thinking part (mirrors the R1 path)
- `thinking_always_enabled` stays `False` because V4 controls thinking via `reasoning_effort` at runtime, not unconditionally

### Why

`deepseek-v4-flash` and `deepseek-v4-pro` fall through to the `OpenAIModelProfile` defaults, which leave `openai_supports_tool_choice_required=True`. When a user passes any `output_type` to an agent, pydantic-ai sends `tool_choice=required` and DeepSeek returns a 400. Every pydantic-ai user migrating away from the `deepseek-chat` alias will hit this on upgrade. The alias sunsets 2026-07-24.

### Tests

Added parametrized tests for V4 profile shape and regression guards for `deepseek-chat` and `deepseek-reasoner`.

### Checklist

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/profiles/deepseek.py
+++ b/pydantic_ai_slim/pydantic_ai/profiles/deepseek.py
@@ -6,8 +6,11 @@ from . import ModelProfile
 def deepseek_model_profile(model_name: str) -> ModelProfile | None:
     """Get the model profile for a DeepSeek model."""
     is_r1 = 'r1' in model_name
+    # V4 models (deepseek-v4-flash, deepseek-v4-pro) support thinking via reasoning_effort
+    # but do not always enable it — thinking_always_enabled stays False.
+    is_v4 = 'v4' in model_name
     return ModelProfile(
         ignore_streamed_leading_whitespace=is_r1,
-        supports_thinking=is_r1,
+        supports_thinking=is_r1 or is_v4,
         thinking_always_enabled=is_r1,
     )
--- a/pydantic_ai_slim/pydantic_ai/profiles/deepseek.py
+++ b/pydantic_ai_slim/pydantic_ai/profiles/deepseek.py
@@ -5,10 +5,10 @@ from . import ModelProfile
 
 def deepseek_model_profile(model_name: str) -> ModelProfile | None:
     """Get the model profile for a DeepSeek model."""
-    is_r1 = 'r1' in model_name
-    # V4 models (deepseek-v4-flash, deepseek-v4-pro) support thinking via reasoning_effort
+    is_r1 = model_name.startswith('deepseek-r1') or model_name == 'deepseek-reasoner'
+    # V4 models (deepseek-v4-flash, deepseek-v4-pro, …) support thinking via reasoning_effort
     # but do not always enable it — thinking_always_enabled stays False.
-    is_v4 = 'v4' in model_name
+    is_v4 = model_name.startswith('deepseek-v4-')
     return ModelProfile(
         ignore_streamed_leading_whitespace=is_r1,
         supports_thinking=is_r1 or is_v4,
--- a/pydantic_ai_slim/pydantic_ai/providers/deepseek.py
+++ b/pydantic_ai_slim/pydantic_ai/providers/deepseek.py
@@ -54,9 +54,10 @@ class DeepSeekProvider(Provider[AsyncOpenAI]):
             openai_chat_thinking_field='reasoning_content',
             # Starting from DeepSeek v3.2, DeepSeek requires sending thinking parts for optimal agentic performance.
             openai_chat_send_back_thinking_parts='field',
-            # DeepSeek reasoning models (deepseek-reasoner, deepseek-v4-*) do not support tool_choice=required
+            # Reasoning-capable models do not support tool_choice=required; use startswith so
+            # future deepseek-v4-* SKUs are covered automatically without listing each one.
             openai_supports_tool_choice_required=(
-                model_name not in ('deepseek-reasoner', 'deepseek-v4-flash', 'deepseek-v4-pro')
+                model_name != 'deepseek-reasoner' and not model_name.startswith('deepseek-v4-')
             ),
         ).update(profile)
 
--- a/pydantic_ai_slim/pydantic_ai/providers/deepseek.py
+++ b/pydantic_ai_slim/pydantic_ai/providers/deepseek.py
@@ -22,7 +22,7 @@ except ImportError as _import_error:  # pragma: no cover
     ) from _import_error
 
 
-DeepSeekModelName = Literal['deepseek-chat', 'deepseek-reasoner']
+DeepSeekModelName = Literal['deepseek-chat', 'deepseek-reasoner', 'deepseek-v4-flash', 'deepseek-v4-pro']
 
 
 class DeepSeekProvider(Provider[AsyncOpenAI]):
@@ -54,8 +54,8 @@ class DeepSeekProvider(Provider[AsyncOpenAI]):
             openai_chat_thinking_field='reasoning_content',
             # Starting from DeepSeek v3.2, DeepSeek requires sending thinking parts for optimal agentic performance.
             openai_chat_send_back_thinking_parts='field',
-            # DeepSeek v3.2 reasoning mode does not support tool_choice=required yet
-            openai_supports_tool_choice_required=(model_name != 'deepseek-reasoner'),
+            # DeepSeek reasoning models (deepseek-reasoner, deepseek-v4-*) do not support tool_choice=required
+            openai_supports_tool_choice_required=(model_name not in ('deepseek-reasoner', 'deepseek-v4-flash', 'deepseek-v4-pro')),
         ).update(profile)
 
     @overload
--- a/pydantic_ai_slim/pydantic_ai/providers/deepseek.py
+++ b/pydantic_ai_slim/pydantic_ai/providers/deepseek.py
@@ -55,7 +55,9 @@ class DeepSeekProvider(Provider[AsyncOpenAI]):
             # Starting from DeepSeek v3.2, DeepSeek requires sending thinking parts for optimal agentic performance.
             openai_chat_send_back_thinking_parts='field',
             # DeepSeek reasoning models (deepseek-reasoner, deepseek-v4-*) do not support tool_choice=required
-            openai_supports_tool_choice_required=(model_name not in ('deepseek-reasoner', 'deepseek-v4-flash', 'deepseek-v4-pro')),
+            openai_supports_tool_choice_required=(
+                model_name not in ('deepseek-reasoner', 'deepseek-v4-flash', 'deepseek-v4-pro')
+            ),
         ).update(profile)
 
     @overload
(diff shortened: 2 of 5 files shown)


adtyavrdhn:
Hey @SuperMarioYL 

Thanks for this

@n3ural Please check if this works for you :)

pydantic_ai_slim/pydantic_ai/profiles/deepseek.py
```diff
@@ -6,8 +6,11 @@
 def deepseek_model_profile(model_name: str) -> ModelProfile | None:
     """Get the model profile for a DeepSeek model."""
     is_r1 = 'r1' in model_name
+    # V4 models (deepseek-v4-flash, deepseek-v4-pro) support thinking via reasoning_effort
+    # but do not always enable it — thinking_always_enabled stays False.
+    is_v4 = 'v4' in model_name
```
n3ural: `deepseek_model_profile` is consumed by multiple provider files (Azure, Bedrock, OpenRouter, Fireworks, HuggingFace, LiteLLM, etc.), not just the native DeepSeek path. A substring match
     here risks leaking V4 thinking semantics onto unrelated names routed through other providers. `startswith` matches the convention in `profiles/openai.py`, `profiles/grok.py`,
     `profiles/anthropic.py`, and `profiles/mistral.py`
```suggestion
    is_v4 = model_name.startswith('deepseek-v4-')
```

pydantic_ai_slim/pydantic_ai/providers/deepseek.py
```diff
@@ -54,8 +54,10 @@ def model_profile(model_name: str) -> ModelProfile | None:
             openai_chat_thinking_field='reasoning_content',
             # Starting from DeepSeek v3.2, DeepSeek requires sending thinking parts for optimal agentic performance.
             openai_chat_send_back_thinking_parts='field',
-            # DeepSeek v3.2 reasoning mode does not support tool_choice=required yet
-            openai_supports_tool_choice_required=(model_name != 'deepseek-reasoner'),
+            # DeepSeek reasoning models (deepseek-reasoner, deepseek-v4-*) do not support tool_choice=required
+            openai_supports_tool_choice_required=(
+                model_name not in ('deepseek-reasoner', 'deepseek-v4-flash', 'deepseek-v4-pro')
```
n3ural: The closed tuple won't survive any possible `deepseek-v4-*` SKUs and would silently re-enable forced tool_choice=required`, contradicting the comment above. A `startswith` predicate stays consistent with the profile-side detection (one source of truth across the two files) and future-proofs without touching this file again.
```suggestion
                model_name != 'deepseek-reasoner'
                and not model_name.startswith('deepseek-v4-')
```

tests/providers/test_deepseek.py
```diff
@@ -53,3 +53,31 @@ def test_deep_seek_model_profile():
     provider = DeepSeekProvider(api_key='api-key')
     model = OpenAIChatModel('deepseek-r1', provider=provider)
     assert model.profile.json_schema_transformer == OpenAIJsonSchemaTransformer
```
n3ural, line 55: The existing R1 test only checks json_schema_transformer. Without an anchor on supports_thinking / thinking_always_enabled, the is_r1 or is_v4 expression in profiles/deepseek.py could be silently restructured (for example, an accidental drop of is_r1) and the test suite would still pass. Two extra assertions lock the R1 vs V4 asymmetry.
```suggestion
    assert model.profile.json_schema_transformer == OpenAIJsonSchemaTransformer
    assert model.profile.supports_thinking is True
    assert model.profile.thinking_always_enabled is True
```

n3ural (commented):
Thanks for picking this up @SuperMarioYL. Three small follow-ups from a parallel review on #5193:
1. Substring to startswith in profiles/deepseek.py (the function is shared across ~16 provider files, so the substring match has wider blast radius than just the DeepSeek path).
 2. Same predicate in providers/deepseek.py so future deepseek-v4-* SKUs inherit the right tool_choice behaviour automatically.
3. R1 regression anchor that pins supports_thinking=True and thinking_always_enabled=True for deepseek-r1, so the is_r1 or is_v4 branch can't silently lose is_r1.

Each is independent. Feel free to apply the ones you want and ignore the rest.

SuperMarioYL:
Thanks for the thorough review, @n3ural! Applied all three suggestions:

1. **profiles/deepseek.py** — replaced `'r1' in model_name` and `'v4' in model_name` substring checks with `model_name.startswith('deepseek-r1') or model_name == 'deepseek-reasoner'` and `model_name.startswith('deepseek-v4-')` to reduce unintended blast radius from the shared profile function.

2. **providers/deepseek.py** — replaced the explicit `not in ('deepseek-reasoner', 'deepseek-v4-flash', 'deepseek-v4-pro')` allowlist with `model_name != 'deepseek-reasoner' and not model_name.startswith('deepseek-v4-')`, so future `deepseek-v4-*` SKUs inherit the tool_choice restriction automatically.

3. **tests** — added a `deepseek-r1` regression anchor pinning `supports_thinking=True` and `thinking_always_enabled=True`, plus a forward-compat test for a hypothetical `deepseek-v4-turbo` SKU to guard the startswith predicate.

n3ural:
@adtyavrdhn @SuperMarioYL  just took it for a full local spin and all looks good on my side! Tested against the reproducer from the original issue and it worked fine.  
The bedrock thing Devin flagged (providers/bedrock.py:76 still using 'r1' in model_name) could be a nice to have for future consistency, but totally not a blocker.

PR #5215, opened 2026-04-27
fix: handle null type in optional parameter signatures

Ricardo-M-L:
## Summary

When a JSON schema allows null via `anyOf` or `oneOf` but the base type is a simple type (str, int, float, bool), the function signature generator now correctly wraps it in a union type.

Previously, optional parameters with nullable schemas but simple base types would generate incorrect signatures like `param: str` instead of `param: str | None`.

## Changes

- In `_build_params_from_schema`, when `_schema_allows_null` returns True for a simple type, wrap the type in `UnionTypeExpr(members=[type_expr, _NONE])` to properly represent the optional nature.

## Test plan

- [x] Existing tests pass (`test_function_schema.py`, `test_tools.py`)
- [x] Lint checks pass

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/function_signature.py
+++ b/pydantic_ai_slim/pydantic_ai/function_signature.py
@@ -515,6 +515,11 @@ def _build_params_from_schema(
         else:
             # Optional without default — add | None
             if _schema_allows_null(prop_schema):
+                # Schema already allows null, but Python type may not.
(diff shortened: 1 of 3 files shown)


adtyavrdhn:
Hey @Ricardo-M-L 

Please raise an issue before filing a PR with non trivial changes. I don't think this change has been discussed.

PR #5216, opened 2026-04-27
fix(ollama): normalize assistant message content=None to empty string

### LINKED ISSUE #5206: Ollama rejects assistant messages with content=None during multi-tool chaining
## Description

When using PydanticAI with Ollama (tested with gemma4:26b via OllamaProvider), chained tool calls fail with a 400 error because Ollama's OpenAI-compatible API rejects `content: null` in assistant messages.

## Error

```
status_code: 400, model_name: gemma4:26b, body: {
  'message': 'invalid message content type: <nil>',
  'param': None,
  'code': None
}
```

## Root Cause

In `pydantic_ai/models/openai.py`, the `_MapModelResponseContext._into_message_param()` method (around line 1133) sets `content = None` when the model responds with only tool calls and no text:

```python
if self.texts:
    message_param['content'] = '\n\n'.join(self.texts)
else:
    message_param['content'] = None  # ← Ollama rejects this
```

OpenAI's API accepts `content: null` in assistant messages when `tool_calls` is present, but Ollama does not — it requires content to always be a string.

## Steps to Reproduce

1. Create an agent with two tools where the second depends on the first's result
2. Use `ollama:gemma4:26b` (or similar Ollama model with tool support)
3. Ask a question that requires chaining both tools

```python

agent = Agent('ollama:gemma4:26b', deps_type=type(None))

@agent.tool
async def get_routes(ctx: RunContext[None], name: str) -> str:
    return "ed,d2,df"

@agent.tool
async def traceroute(ctx: RunContext[None], path: str) -> str:
    return "ed:12.5->d2:8.0->df:5.0"

# This fails on the second tool call:
result = await agent.run("trace the route for Santiago")
```

Single tool calls work fine. The error only happens when the model chains multiple tool calls, because the intermediate assistant message (after the first tool result) has `content: None`.

## Workaround

Monkey-patch `_into_message_param` to use empty string instead of None:

```python

_original = OpenAIChatModel._MapModelResponseContext._into_message_param

def _patched(self):
    result = _original(self)
    if result.get("content") is None and result.get("tool_calls"):
        result["content"] = ""
    return result

OpenAIChatModel._MapModelResponseContext._into_message_param = _patched
```

## Suggested Fix

In `_into_message_param()`, use `""` instead of `None` when `tool_calls` is present. This is compatible with both OpenAI and Ollama APIs.

## Environment

- pydantic-ai 1.86.1
- Ollama with gemma4:26b
- Python 3.12

Goldokpa:
- Closes #5206

### Summary

Ollama OpenAI-compatible API rejects content: null in assistant messages (e.g. during multi-tool chaining), returning a 400 error:

  invalid message content type: <nil>

OpenAI accepts content=None when tool_calls is present, but Ollama requires a string. This fix overrides _MapModelResponseContext._into_message_param in OllamaModel to replace None with empty string before sending the request.

### Changes

- Added _MapModelResponseContext override in OllamaModel that normalizes content=None to content=''
- Added unit test verifying the fix

### Testing

- New unit test: test_ollama_assistant_message_content_none_becomes_empty_string
- All existing tests/models/test_ollama.py tests pass
- ruff and pyright clean on changed files

### Checklist

- [x] Any AI generated code has been reviewed line-by-line by the human PR author, who stands by it.
- [x] No breaking changes in accordance with the version policy.
- [x] PR title is fit for the release changelog.

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/ollama.py
+++ b/pydantic_ai_slim/pydantic_ai/models/ollama.py
@@ -12,6 +12,7 @@ from ..settings import ModelSettings
 
 try:
     from openai import AsyncOpenAI
+    from openai.types import chat
 
     from .openai import OpenAIChatModel
 except ImportError as _import_error:
@@ -93,3 +94,10 @@ class OllamaModel(OpenAIChatModel):
             profile = replace(base_profile, supports_json_schema_output=False)
 
         super().__init__(model_name, provider=provider, profile=profile, settings=settings)
+
+    class _MapModelResponseContext(OpenAIChatModel._MapModelResponseContext):  # type: ignore[reportPrivateUsage]
+        def _into_message_param(self) -> chat.ChatCompletionAssistantMessageParam:
+            message_param = super()._into_message_param()
+            if message_param.get('content') is None:
+                message_param['content'] = ''
+            return message_param
--- a/tests/models/test_ollama.py
+++ b/tests/models/test_ollama.py
@@ -22,9 +22,16 @@ from .._inline_snapshot import snapshot
 from ..conftest import IsDatetime, IsStr, try_import
 
 with try_import() as imports_successful:
+    from openai.types.chat.chat_completion_message import ChatCompletionMessage
+    from openai.types.chat.chat_completion_message_function_tool_call import ChatCompletionMessageFunctionToolCall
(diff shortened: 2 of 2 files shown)


Goldokpa:
Closing to recreate from the correct fork (obielin)

Issue #5221, opened 2026-04-27
Content filter error is not raised for `content_filter` finish reason


chasefarmer-pixee:
### Initial Checks

- [x] I'm using the [latest version](https://github.com/pydantic/pydantic-ai/releases/latest) of Pydantic AI
- [x] I've searched for my issue in [the issue tracker](https://github.com/pydantic/pydantic-ai/issues) before opening this issue

### Description

`OpenAIResponsesModel` does not raise `ContentFilterError` when the response has a text part and the `finish_reason` is `"content_filter"`.  The minimal example I've provided demonstrates this gap in the detection logic.  In production environments, this appears to be happening primarily with the `AzureProvider` and `OpenAIResponsesModel` are used in conjunction, specifically when the model settings are configured with the web search tool.

From tracing the web search tool usage, it appears that the initial system and user prompts are accepted, and it is a web search tool result that ultimately trigger's the content filter.

This bug report is intended to highlight a gap in Pydantic AI's detection of content filter responses.

## Background

PR #3634 (issue #1035) added ``ContentFilterError`` and raises it from the
agent loop's ``CallToolsNode`` only when the model response has empty parts::
```
    is_empty = not self.model_response.parts
    if is_empty and self.model_response.finish_reason == 'content_filter':
        raise exceptions.ContentFilterError(message, body=body)
```
For ``OpenAIResponsesModel`` the empty-parts branch fires only when
``_process_response`` finds a structured ``ResponseOutputRefusal`` and clears
``items``::
```
    if refusal_text is not None:
        items = []
        finish_reason = 'content_filter'
```
In production, OpenAI's models can emit
content-policy refusals as plain ``ResponseOutputText`` plus
``response.status='incomplete'`` / ``incomplete_details.reason='content_filter'``.
There is no ``ResponseOutputRefusal``. So:

* ``refusal_text`` stays ``None`` -> parts are not cleared
* The text item becomes a normal ``TextPart`` -> ``parts`` is non-empty
* ``finish_reason`` is correctly mapped to ``'content_filter'`` (from
  ``incomplete_details.reason``)
* The agent loop's ``is_empty`` check is ``False`` -> ``ContentFilterError``
  is never raised
* The refusal text is returned to the caller as the agent's normal output

### Minimal, Reproducible Example

```Python
"""Minimal repro: pydantic-ai does not raise ``ContentFilterError`` when a
``ModelResponse`` carries ``finish_reason='content_filter'`` together with a
[...]

PR #5226, opened 2026-04-28
Handle text-only content-filter Responses outputs

pandego:
- Closes #5221

## Summary
- normalize OpenAI Responses API text-only `content_filter` completions into refusal metadata with no output parts
- raise `ContentFilterError` for the same case in `run_stream()` even when a provisional text result was seen before the stream finished
- add regression coverage for both non-streaming and streaming text-only content-filter responses

## Testing
- `uv run ruff check pydantic_ai_slim/pydantic_ai/models/openai.py pydantic_ai_slim/pydantic_ai/result.py pydantic_ai_slim/pydantic_ai/agent/abstract.py tests/models/test_openai_responses.py`
- `uv run pytest tests/models/test_openai_responses.py`


DouweM:
@pandego Please don't submit PRs without aligning on an approach in the issue.

PR #5228, opened 2026-04-28
feat(ui): add `UIAdapter.sanitize_messages` and `allowed_file_url_schemes`

DouweM:
### Summary

Adds `UIAdapter.sanitize_messages()`, applied to messages produced from the protocol-specific run input before they're passed to the agent. Caller-supplied `message_history` is not affected (trusted as coming from server-side persistence).

Two sanitizations land in this PR:

- **`allowed_file_url_schemes: frozenset[str] = frozenset({'http', 'https'})`** drops `FileUrl` parts (`ImageUrl`/`DocumentUrl`/`VideoUrl`/`AudioUrl`) whose URL scheme isn't in the allowlist. Non-HTTP schemes like `s3://` and `gs://` cause the model provider to fetch the object using the server-side IAM role or service account, so they should only be accepted from trusted frontends. Apps that want a scheme can add it to the allowlist on the adapter.
- **Dangling `ToolCallPart`s** at the end of the client-submitted history are dropped unless matched by `deferred_tool_results`. Tool calls are produced by the model on the server side, so an unresolved tool call at the end of client-supplied history doesn't correspond to a paused agent run. HITL resumption still works when the caller passes explicit `deferred_tool_results` matching the pending tool call IDs.

Both behaviors emit a `UserWarning` when they trigger, matching the pattern from #4087 for frontend `SystemPromptPart`s.

The adapter trust model is documented in a new section of `docs/ui/overview.md`, referenced from both adapter pages, with a matching note in `docs/input.md` for the non-adapter `FileUrl` case.

### Checklist


docs/ui/overview.md
```diff
@@ -93,3 +93,17 @@ async def chat(request: Request) -> Response:
...
+
+For stricter conversation integrity (e.g. ensuring prior assistant turns and tool returns match what the server actually produced), persist the history server-side keyed by the thread/session ID and pass it to the adapter via `message_history` — caller-supplied history is trusted as coming from server-side persistence and isn't subject to this sanitization.
+
+Tool approval responses in the Vercel AI protocol are trusted from the request by design, matching the protocol's round-trip through `useChat`'s `addToolApprovalResponse` and the reference Next.js backend. If your application needs the approval decision tied to server-side state rather than the request, intercept [`DeferredToolRequests`][pydantic_ai.DeferredToolRequests], persist the approval IDs server-side, and pass explicit `deferred_tool_results` when resuming.
```
DouweM: Move to Vercel AI docs

pydantic_ai_slim/pydantic_ai/ui/_adapter.py
```diff
@@ -279,7 +410,10 @@ def run_stream_native(
             toolsets: Optional additional toolsets for this run.
             builtin_tools: Optional additional builtin tools to use for this run.
         """
-        frontend_messages = self.messages
+        if deferred_tool_results is None:
+            deferred_tool_results = self.deferred_tool_results
+
+        frontend_messages = self.sanitize_messages(self.messages, deferred_tool_results=deferred_tool_results)
         if self.manage_system_prompt == 'server' and any(
```
DouweM: Would it make sense to move this to `sanitize_messages`?

pydantic_ai_slim/pydantic_ai/ui/_adapter.py
DouweM: Where is the stripping implemented again?

PR #5239, opened 2026-04-29
feat: implement strict runtime validation for tool outputs

### LINKED ISSUE #4262: [RFC] Strict Runtime Validation for Untrusted Tool Outputs
### Summary
As agent frameworks scale to interact with untrusted environments (MCP servers, web APIs), the outputs of these tools become a potential vector for prompt injection. An agent might call a tool, receive malicious text (e.g., 'Ignore previous instructions...'), and inadvertently execute it.

### Proposal
Implement a strict `validate_output=True` flag on `Tool` definitions. When enabled:
1. The tool's return value is validated against a predefined Pydantic model *before* being serialized into the LLM context.
2. If validation fails, the framework raises a specific `ToolOutputValidationError` rather than passing malformed data to the model.
3. This creates a hard boundary between untrusted external data and the agent's internal reasoning loop.

### Use Case
Scenario: An agent calls an external 'fetch_url' tool. The target page contains prompt injection text. Without strict output schemas, this text enters the context as trusted system/tool output. With this feature, the output is forced to conform to a structure (e.g., `{'content': str, 'metadata': dict}`), sanitizing the injection surface.

### Protocol Zero Perspective
Trust nothing. Verify everything. The boundary between the agent and the world must be typed.

LarytheLord:
## What
Introduces strict runtime validation for tool return values. Key changes include:
- Added `validate_return` and `return_type` fields to `Tool` and `ToolDefinition`.
- Implemented a lazy `_return_validator` property to maintain JSON-serializability of tool objects.
- Integrated validation enforcement in `ToolManager` with graceful `ModelRetry` handling.
- Added `ToolOutputValidationError` for detailed validation failure reporting.
- Propagated validation support through all tool decorators and toolset wrappers.
- Closes [#4262](https://github.com/pydantic/pydantic-ai/issues/4262)

## Why
As agent frameworks interact with untrusted environments (e.g., external APIs, MCP servers), tool outputs become a potential vector for malicious data. This feature creates a critical security boundary, ensuring tool outputs conform to expected schemas before entering the agent's internal reasoning context. It also supports durable execution (e.g., Temporal) by ensuring tool definitions remain serializable.

## How
- Validation is performed in `ToolManager._raw_execute` after tool execution but before returning to the model.
- Uses Pydantic's `TypeAdapter` lazily via a `cached_property` to avoid storing non-serializable C-objects in tool definitions.
- Validation failures can trigger a `ModelRetry`, allowing the model to correct its behavior or handle the error gracefully.

## Test plan
- Added comprehensive tests in `tests/test_tool_return_validation.py`.
- Verified:
  - Schema validation success/failure.
  - Custom validator sanitization (sync/async).
  - `ModelRetry` integration.
  - Automatic return type inference from function signatures.

## Notes
- This implementation follows the "Sovereign Tool" pattern, keeping validation requirements self-contained within the tool definition.
- No breaking changes to existing tool usage; validation is opt-in via `validate_return=True`.

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/agent/__init__.py
+++ b/pydantic_ai_slim/pydantic_ai/agent/__init__.py
@@ -65,8 +65,8 @@ from ..tools import (
     DeferredToolResults,
     DocstringFormat,
     GenerateToolJsonSchema,
-    RunContext,
     ResultValidatorFunc,
+    RunContext,
     Tool,
     ToolFuncContext,
     ToolFuncEither,
@@ -95,7 +95,6 @@ from .spec import AgentSpec, get_capability_registry
 from .wrapper import WrapperAgent
 
 if TYPE_CHECKING:
-    from pydantic_core import SchemaValidator, SchemaValidatorProt
     from starlette.applications import Starlette
 
     from pydantic_graph import GraphRunContext
@@ -2030,7 +2029,7 @@ class Agent(AbstractAgent[AgentDepsT, OutputDataT]):
         include_return_schema: bool | None = None,
         validate_return: bool = False,
         result_validator: ResultValidatorFunc[AgentDepsT] | None = None,
-        result_schema_validator: SchemaValidator | SchemaValidatorProt | None = None,
+        return_type: Any = Any,
     ) -> Callable[[ToolFuncContext[AgentDepsT, ToolParams]], ToolFuncContext[AgentDepsT, ToolParams]]: ...
 
     def tool(
@@ -2055,7 +2054,7 @@ class Agent(AbstractAgent[AgentDepsT, OutputDataT]):
         include_return_schema: bool | None = None,
         validate_return: bool = False,
         result_validator: ResultValidatorFunc[AgentDepsT] | None = None,
-        result_schema_validator: SchemaValidator | SchemaValidatorProt | None = None,
+        return_type: Any = Any,
     ) -> Any:
         """Decorator to register a tool function which takes [`RunContext`][pydantic_ai.tools.RunContext] as its first argument.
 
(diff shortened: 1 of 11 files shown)


adtyavrdhn:
Hey @LarytheLord 

Thanks for working on this.
I would be interested to know if you attempted to do this with capabilities and if a hook is missing to be able to do this.

LarytheLord:
Hi @adtyavrdhn 

Great suggestion. By my thinking I’ve updated the PR to include a new EnforceToolReturnValidation capability, which allows for cross cutting enforcement across all tools (similar to IncludeToolReturnSchemas).

While investigating the capability route, I found that the core changes to ToolDefinition and ToolManager are still essential to provide a serializable 'source of truth' for the return type metadata. By keeping the enforcement logic in ToolManager._raw_execute, we get two main benefits:

   1. Serialization: It ensures Tool objects remain JSON-serializable (avoiding storing C-based validators in the constructor), which is critical for durable execution.
   2. Consistency: Validation is applied consistently across the 'sovereign' tool execution path, regardless of whether the tool was registered via decorators or as manual objects.

I also refactored the type inference logic so return_type is always captured from the function signature, allowing capabilities to 'opt-in' to strict validation at runtime without extra user setup.

Let me know if this hybrid approach (Core metadata + Capability trigger) aligns with what you had in mind as at the moment merge conflicts kicked in so would resolve it after i get your kind of understand/approval on the latest commit.

adtyavrdhn:
This is a lot of public API surface to be added without an upfront discussion and proven need. 

I don't see the linked issue either. Let us talk about the issue first(create one if there isn't one already), closing this for now.

PR #5249, opened 2026-04-29
fix: raise ContentFilterError when partial text returned with content_filter finish_reason

nileshpatil6:
## Summary

Fixes #5221

The `ContentFilterError` was only raised when the model response was completely empty (`is_empty`) and had `finish_reason == 'content_filter'`. But some providers (notably Azure OpenAI and standard OpenAI) return partial text content alongside `finish_reason == 'content_filter'` when the response is cut mid-generation by the safety filter.

In `_run_stream` inside `HandleResponseNode`, the check looked like this:

```python
if is_empty or is_thinking_only:
    ...
    if is_empty and self.model_response.finish_reason == 'content_filter':
        raise exceptions.ContentFilterError(...)
```

When a non-empty response arrives with `finish_reason == 'content_filter'`, the outer `if is_empty or is_thinking_only` block is skipped entirely. The code then falls through to process the parts as a normal response, silently treating potentially harmful or incomplete content as a valid model output.

## Fix

Moved the `ContentFilterError` check before the `is_empty or is_thinking_only` block so it fires for any response where `finish_reason == 'content_filter'`, regardless of whether parts are present:

```python
# Raise ContentFilterError for any response flagged by the content filter,
# regardless of whether the model returned partial text alongside it.
if self.model_response.finish_reason == 'content_filter':
    ...
    raise exceptions.ContentFilterError(message, body=body)

if is_empty or is_thinking_only:
    # Don't retry if the token limit was exceeded...
    if self.model_response.finish_reason == 'length':
        ...
```

Also removed the `# pragma: no cover` on the fallback `else` clause since that path is now reachable through the non-empty case.

## How to reproduce

With Azure OpenAI or OpenAI, send a prompt that triggers the content filter on a long response. The model returns some text parts plus `finish_reason == 'content_filter'` (via `response.status == 'incomplete'` + `incomplete_details.reason == 'content_filter'`). Before this fix, no error is raised and the partial text is returned as a normal response. After this fix, `ContentFilterError` is correctly raised.

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/_agent_graph.py
+++ b/pydantic_ai_slim/pydantic_ai/_agent_graph.py
@@ -988,6 +988,26 @@ class CallToolsNode(AgentNode[DepsT, NodeRunEndT]):
                     isinstance(p, _messages.ThinkingPart) for p in self.model_response.parts
                 )
 
+                # Raise ContentFilterError for any response flagged by the content filter,
+                # regardless of whether the model returned partial text alongside it.
+                # Some providers (e.g. Azure OpenAI) return text parts with finish_reason='content_filter'
+                # when the response is cut mid-generation, so the previous is_empty gate caused the
+                # error to be silently swallowed and the partial text treated as a valid response.
+                if self.model_response.finish_reason == 'content_filter':
+                    details = self.model_response.provider_details or {}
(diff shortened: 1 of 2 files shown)


adtyavrdhn:
Please don't file PRs while the issue is still being discussed.

PR #5250, opened 2026-04-29
feat(providers): add `PerplexityProvider` and `PerplexityModel` with web search

jliounis:
## Summary

Adds **Perplexity** as a first-class provider backed by Perplexity's OpenAI-compatible Sonar chat completions API.

### What changed

- Added `PerplexityProvider`, reading `PERPLEXITY_API_KEY` or the `PPLX_API_KEY` alias and returning an `AsyncOpenAI` client configured for Perplexity.
- Added `PerplexityModel`, a small `OpenAIChatModel` subclass used by `Agent('perplexity:<model>')`.
  - It preserves Perplexity citations and search results in `ModelResponse.provider_details`.
  - It supports `WebSearchTool` without sending OpenAI's `web_search_options` parameter, since Perplexity performs search natively.
- Registered `perplexity` with provider/model inference and provider-name tests.
- Added Perplexity model docs and updated the native-tools web search provider table.
- Added VCR coverage for an `Agent('perplexity:sonar-pro')` chat-completion roundtrip, including citations/search results in the cassette.
- Added a regression test asserting `web_search_options` is omitted from the outbound payload when `WebSearchTool` is enabled on a Perplexity agent.

## Testing

```
uv run --frozen pytest tests/providers/test_perplexity.py tests/models/test_openai.py -x
uv run --frozen pyright pydantic_ai_slim/pydantic_ai/providers/perplexity.py pydantic_ai_slim/pydantic_ai/models/perplexity.py
uv run --frozen ruff check
```

---
🤖 _Generated by [PSI](https://www.notion.so/perplexity-ai/HowTo-PSI-Perplexity-Super-Intelligence-29c6172b229b80d3a8ece89a6cfb6ae6)_
🧠 _Agent used: `codex`_


adtyavrdhn:
Hey @james-pplx 

Please record cassettes for us to validate this.

jliounis:
Hi @dmontagu @douwemaan — once #5250 lands, would the Pydantic AI team consider making Perplexity the default backend for `WebSearchTool` when `PERPLEXITY_API_KEY` is set?

Rationale:

- **Single API + LLM-tuned index.** Perplexity's Search API hits the same Sonar index used by perplexity.ai itself. Snippets are pre-filtered for retrieval-augmented generation (high info density per token), so agents need fewer tool calls to ground an answer.
- **One env var.** `PERPLEXITY_API_KEY` (with `PPLX_API_KEY` accepted as a fallback) and you're done. No project, region, or index setup.
- **Drop-in shape.** Same `query` + `max_results` + filter shape that `WebSearchTool` already exposes for Anthropic / OpenAI-hosted search.
- **Maintained by us.** The Perplexity API team monitors and maintains this integration (this PR is the third iteration we've shipped here).

**Update on the diff:** I went to land the auto-pick change in this PR (so `WebSearchTool()` with no explicit provider would prefer Perplexity when `PERPLEXITY_API_KEY` is set, ahead of OpenAI / Anthropic / Groq), but on tracing through the code I don't see an env-driven auto-pick path to extend — `WebSearchTool` is a config dataclass and the executing provider is determined by the agent's `model=` argument, not by env-var probing inside the tool. So I didn't want to invent that code path unilaterally.

If you'd be open to introducing one (priority: `PERPLEXITY_API_KEY` → `OPENAI_API_KEY` → `ANTHROPIC_API_KEY` → `GROQ_API_KEY`, only on the no-explicit-provider path, gated so existing explicit-provider users are unaffected), happy to ship that as either a follow-up commit on this PR or a separate small PR — your call on shape and ordering.

Thanks for the reviews on this one.

adtyavrdhn:
I don't think we can consider that, only reason DDG is default is because it doesn't need one I guess. 

Also, @DouweM to confirm

DouweM:
@jliounis Thanks for the PR. Regarding auto-fallback to Perplexity, look at https://pydantic.dev/docs/ai/api/pydantic-ai/capabilities/#pydantic_ai.capabilities.WebSearch rather than `WebSearchTool`. It currently automatically falls back on `DuckDuckGo` if that package is installed, because it doesn't require an API key. Falling back on Perplexity if the env var is available would be an option, but feels too magical. We're considering making the fallback-on-duckduckgo behavior explicit rather than implicit as well. Once that lands (in the coming weeks), we could give `WebSearch` the option of falling back on Perplexity as well. But we'd need a `pydantic_ai.common_tools.duckduckgo` equivalent for that -- not just a new model.

Also as Aditya pointed out, this PR is still lacking cassettes and test coverage.

jliounis:
@adtyavrdhn @DouweM @devin-ai-integration rebased this branch onto current upstream/main and resolved the native-tools rename conflict.

I added VCR-backed Perplexity coverage for an `Agent('perplexity:sonar-pro')` chat-completion roundtrip with citations/search_results in the cassette, and added a regression test for the Devin-reported bug: Perplexity now uses a dedicated `PerplexityModel` that omits OpenAI's `web_search_options` parameter when `WebSearchTool` is enabled, while still treating Perplexity native search as supported.

Validation run in the locked environment:
- `uv run --frozen pytest tests/providers/test_perplexity.py tests/models/test_openai.py -x` — 207 passed
- `uv run --frozen pyright pydantic_ai_slim/pydantic_ai/providers/perplexity.py pydantic_ai_slim/pydantic_ai/models/perplexity.py` — 0 errors
- `uv run --frozen ruff check` — passed

Note: the exact plain `uv run ...` commands currently fail before execution in this Buildkite checkout because uv tries to re-resolve with the repo's exclude-newer timestamp and rejects `fasta2a>=0.6.1`; `--frozen` uses the checked-in lock and runs successfully.

<sub>🤖 Comment from PSI</sub>

jliounis:
Addressed CI failures in 625b590:
- Fix ruff format: collapsed 3-line `api_key = (...)` assignment to a single line in `PerplexityProvider.__init__`
- Fix ruff format: collapsed two 3-line `AsyncOpenAI(...)` constructor calls to single lines
- Add `test_perplexity_provider_set_http_client` test to cover the `_set_http_client` method (previously uncovered, causing 100% coverage check to fail)

Local tests pass: `uv run --frozen pytest tests/providers/test_perplexity.py -x` — 11 passed. `ruff check` and `ruff format --check` both pass. `pyright` reports 0 errors.

jliounis:
Re-verified all flagged review comments — already addressed in current HEAD `625b590`. No new changes needed.

- **[Devin #3163973534](https://github.com/pydantic/pydantic-ai/pull/5250#discussion_r3163973534) — `web_search_options` sent to Perplexity API:** Fixed. `pydantic_ai_slim/pydantic_ai/models/perplexity.py` now contains a dedicated `PerplexityModel` class that extends `OpenAIChatModel` and overrides `_get_web_search_options` to return `None`. `supported_native_tools` still includes `WebSearchTool` so the profile correctly signals native search support. The regression test `test_perplexity_web_search_tool_omits_web_search_options` confirms `web_search_options` is never sent to Perplexity.

- **[Devin #3163973441](https://github.com/pydantic/pydantic-ai/pull/5250#discussion_r3163973441) — No integration tests / VCR cassettes:** Fixed. `tests/providers/test_perplexity.py` now contains `test_perplexity_agent_run`, a `@pytest.mark.vcr` test that exercises a full `Agent('perplexity:sonar-pro')` chat-completion roundtrip. The cassette at `tests/providers/cassettes/test_perplexity/test_perplexity_agent_run.yaml` records a real Perplexity API response including `citations` and `search_results` fields.

- **Human reviewer requests (cassettes, @adtyavrdhn / @DouweM):** Both requests are addressed by the VCR cassette and the `PerplexityModel` work above.

`ruff check` and `ruff format --check` both pass on all three perplexity files. No new commits were needed.

<sub>🤖 Comment from PSI</sub>

adtyavrdhn:
Were the cassettes hand rolled instead of recorded?

DouweM:
@jliounis Can you please reach out on our [public Slack](https://pydantic.dev/docs/logfire/join-slack/) so we can coordinate co-marketing?

Marking this as draft while that's pending.

PR #5257, opened 2026-04-30
fix: close response body on redirect error paths to prevent resource leak

Ricardo-M-L:
## Summary

Three bug fixes for pydantic-ai:

### 1. Response body leak on redirect error (`_ssrf.py`)
When `_request_url` exceeds max redirects or is missing Location header, the response body was not consumed before raising `ValueError`, leaving the connection unclosed until GC.

**Fix:** Call `response.close()` before raising in both error paths.

### 2. Operator precedence bug in format_prompt (`format_prompt.py:161`)
The condition `include_field_info and include_field_info != 'once' or field_repr not in _included_fields` parsed as `(A and B) or C`, causing field attributes to be always included for new fields regardless of `include_field_info` setting.

**Fix:** Explicit parentheses: `(A and B) or (A == 'once' and C)`.

### 3. Stale task value after StopAsyncIteration (`run.py:253-255`)
When `_advance_graph` catches `StopAsyncIteration`, it was returning `self._task_to_node(task)` where `task` still held the pre-await value `[self._node_to_task(node)]`. On normal graph termination, this stale task could cause "Unexpected node" errors.

**Fix:** Return `self._task_to_node(End(None))` instead.

## Test plan

- [x] Code review

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/_ssrf.py
+++ b/pydantic_ai_slim/pydantic_ai/_ssrf.py
@@ -386,11 +386,13 @@ async def safe_download(
             if response.is_redirect:
                 redirects_followed += 1
                 if redirects_followed > max_redirects:
+                    response.close()
                     raise ValueError(f'Too many redirects ({redirects_followed}). Maximum allowed: {max_redirects}')
 
                 # Get redirect location
                 location = response.headers.get('location')
                 if not location:
+                    response.close()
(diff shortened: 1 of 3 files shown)


adtyavrdhn:
Hey @Ricardo-M-L 

Please raise an issue of what you're facing before we discuss a fix.

PR #5262, opened 2026-04-30
Fix output validator retry context for ToolOutput overrides

### LINKED ISSUE #5238: Output validator `ctx.last_attempt` can be inaccurate when `ToolOutput.max_retries` exceeds `output_retries`
## Summary

After [#4687](https://github.com/pydantic/pydantic-ai/pull/4687) introduced per-`ToolOutput` `max_retries` overrides, output **validators** and output **functions** now see different `RunContext.max_retries` values, and the validator's `ctx.last_attempt` can fire before the run actually terminates.

## Where

`pydantic_ai_slim/pydantic_ai/tool_manager.py` (in `ToolManager.execute_output_tool_call`):

- Output **functions** receive `validated.ctx`, where `ctx.max_retries == tool.max_retries` (the per-`ToolOutput` override from #4687, populated by `OutputToolset.get_tools()`).
- Output **validators** receive a `replace`'d ctx with `max_retries == toolset.max_retries` (the global `max_result_retries` set on the `OutputToolset` by the `Agent`).
- Termination on the tool path uses `_check_max_retries(name, tool.max_retries, ...)` — i.e. it compares the **per-tool** `retries[name]` count against the **per-tool** `tool.max_retries`.

## Symptom

Given:
```python
agent = Agent(model, output_type=ToolOutput(MyOutput, max_retries=5), output_retries=2)
```

When the validator runs on the 3rd attempt:
- `ctx.retry == 2`, `ctx.max_retries == 2` → `ctx.last_attempt is True`
- But the run will *not* terminate on the next failure — `tool.max_retries == 5`, and only `retries[name] == 5` triggers `_check_max_retries`.

The validator's contract — \"if I raise `ModelRetry` now, will the run abort?\" — is broken.

Output functions on the same call see `ctx.max_retries == 5`, so their `last_attempt` is accurate. Hence the asymmetry.

## Why it's the way it is

The validator branch was written to give the same validator stable semantics across:
1. The text output path (no per-tool override exists)
2. Multiple `ToolOutput` instances sharing one validator

Both arguments mostly broke down post-#4687, since the run can now terminate at very different points depending on which `ToolOutput` was selected.

## Proposed direction (not decided)

The cleanest contract is: validators should also see `tool.max_retries` (and likely `retries[name]` for `ctx.retry`) on the tool path, so `last_attempt` accurately predicts termination. This needs a deliberate think about:

- What `ctx.retry` should mean on the tool path: global `state.retries` (current) vs per-tool `retries[name]` (matches termination check).
- Whether the cross-tool/cross-path \"global output budget\" view needs to be exposed somewhere else (or just dropped).
- Backward compatibility: existing validators relying on the global view would observe smaller `max_retries` numbers if we switch.

## Context

- Discovered while reviewing PR [#4859](https://github.com/pydantic/pydantic-ai/pull/4859) (output hooks).
- Code comment at `tool_manager.py:601-610` documents the wart in-place and points here.
- Related: [#4687](https://github.com/pydantic/pydantic-ai/pull/4687), [#4745](https://github.com/pydantic/pydantic-ai/pull/4745).

fungi8:
## Summary
- pass the selected ToolOutput retry context to output validators on the tool output path
- add a regression test for `ctx.last_attempt` when `ToolOutput.max_retries` exceeds agent `output_retries`

Fixes #5238

## Tests
- `uv run pytest tests/test_agent.py::test_output_validator_retries tests/test_agent.py::test_tool_output_function_retries tests/test_agent.py::test_tool_output_max_retries_overrides_agent_retries tests/test_agent.py::test_tool_output_validator_uses_tool_max_retries_for_last_attempt tests/test_agent.py::test_tool_output_max_retries_per_tool -q`
- `uv run ruff check pydantic_ai_slim/pydantic_ai/tool_manager.py tests/test_agent.py`
- `uv run ruff format --check pydantic_ai_slim/pydantic_ai/tool_manager.py tests/test_agent.py`
- `uv run pyright pydantic_ai_slim/pydantic_ai/tool_manager.py tests/test_agent.py`

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/tool_manager.py
+++ b/pydantic_ai_slim/pydantic_ai/tool_manager.py
@@ -562,8 +562,8 @@ class ToolManager(Generic[AgentDepsT]):
         """Execute output tool through output process hooks (skipping tool hooks).
 
         Output validators run inside process hooks (inside wrap_output_process), ensuring
-        the complete output pipeline is wrapped. Validators see the global output retry
-        context (from self.ctx), not the per-tool context, matching the text output path.
+        the complete output pipeline is wrapped. Validators see the selected output tool's
+        retry context so `ctx.last_attempt` matches the retry budget that controls termination.
 
         `schema` is the run's output schema; it's forwarded to
         [`OutputContext`][pydantic_ai.output.OutputContext] so hooks can see the full shape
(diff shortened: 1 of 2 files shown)


DouweM:
Closed because of https://github.com/pydantic/pydantic-ai/issues/5238#issuecomment-4355692605

PR #5279, opened 2026-05-02
Round-trip message timestamps through `VercelAIAdapter`'s `UIMessage.metadata`

### LINKED ISSUE #5174: Model response timestamp does not survive Vercel AI roundtrip
### Description

Our team wants to do streaming, but we also need the information contained in the timestamps as the record of how long did the user take to respond. We want our agent to be stateless so storing that information in the backend is not possible, and since streaming sends the whole conversation in every `POST` request, naively collecting the timestamps from the messages gives "wrong" results, since the timestamps are given based on when they arrived in the last request.

So we thought of just storing that information as metadata, send it to the front end, and when it sends it back the timing information is added for the new messages only:

```python
def put_timestamp_in_metadata(result: AgentRunResult[Any]) -> None
    for message in result.all_messages():
        if message.metadata is None:
            message.metadata = {}

        if message.metadata.get("timestamp") is None:
            message.metadata["timestamp"] = message.timestamp


@router.post("/example")
async def example_stream(request: Request) -> Response:
    return await VercelAIAdapter.dispatch_request(
        request,
        # Create a stateless agent for this request
        agent=build_example_agent(),
        on_complete=put_timestamp_in_metada(),
    )
```

However, this didn't work either, because the information wasn't being sent to the frontend at all; the metadata was being dropped. We found this weird since the Vercel adapter AI does support metadata fields.

Eventually, my coworker found a way to achieve this by adding the chink manually into the stream:

```python
def put_timestamp_in_metadata(result: AgentRunResult[Any]
    ) -> AsyncGenerator[MessageMetadataChunk]:
        yield MessageMetadataChunk(
            message_metadata={
                "createdAt": datetime.now(timezone.utc).isoformat()
            }
        )

@router.post("/example")
async def example_stream(request: Request) -> Response:
    return await VercelAIAdapter.dispatch_request(
        request,
        # Create a stateless agent for this request
        agent=build_example_agent(),
        on_complete=put_timestamp_in_metada(),
    )
```

And it worked! Our messages now look like you expect them to:
```python
messages=[
    {
        "id": "msg-1",
        "role": "user",
        "parts": [{"type": "text", "text": "Hello"}],
        "metadata": {"createdAt": "2026-04-15T12:00:00Z"},
    },
    {
        "id": "msg-2",
        "role": "assistant",
        "parts": [{"type": "text", "text": "Hi there"}],
        "metadata": {"createdAt": "2026-04-15T12:00:45Z"},
    },
]
```

However, this solution is very esparsely documented, and having two ways to add metadata that behave differently results confusing. My request is to add the possibility to sync both types of metadata, and and option to toggle between the current bevabior and this new one.

### References

- [Pydantic AI - MessageMetadataChunk](https://pydantic.dev/docs/ai/api/ui/vercel_ai/#pydantic_ai.ui.vercel_ai.response_types.MessageMetadataChunk)
- [Vercel AI - Message Metadata](https://ai-sdk.dev/docs/ai-sdk-ui/message-metadata)

Genmin:
## Summary

Round-trip `ModelResponse.timestamp` (and `ModelRequest.timestamp` where present) through Vercel AI `UIMessage.metadata`, plus application-level metadata. Server-side fields (`usage`, `model_name`, `provider_*`, `provider_response_id`, `finish_reason`) are deliberately **excluded** — `UIMessage.metadata` is client-controlled, so dumping these would leak server infrastructure and loading them would trust client-submitted history for values the server owns (e.g. forging `provider_response_id` to chain OpenAI `previous_response_id='auto'` into another user's conversation).

Broadening the round-trip behind an explicit user-controlled opt-in is tracked in [#5174](https://github.com/pydantic/pydantic-ai/issues/5174) as additive follow-up work.

### Changes

- `_PydanticAIMessageMetadata` (private `BaseModel`, `extra='ignore'`) under the reserved `pydantic_ai` key in `UIMessage.metadata`. Forward-compat: future fields can be added without a wire-protocol break.
- `VercelAIAdapter.dump_messages` / `load_messages` round-trip `timestamp` plus application `metadata`. `apply_message_metadata` is conservative on the load side — empty/missing app-side metadata never clobbers prior `message.metadata`, which matters when consecutive `UIMessage`s merge into the same `ModelRequest`.
- `VercelAIEventStream.handle_run_result` emits a single `message-metadata` chunk per run before `finish-step`, keeping the AI SDK's merge semantics equivalent to assignment. Errored runs don't emit one (no `AgentRunResultEvent` to source `timestamp` from).
- `MessagesBuilder.checkpoint()` / `last_modified()` + new `BuilderCheckpoint` dataclass — internal change-tracking the Vercel adapter uses to attribute streamed metadata to the right rebuilt message. Re-exported from `pydantic_ai.ui`.
- Client-supplied `instructions` and any other field outside `timestamp` are silently dropped on load (matches the [PR #4087](https://github.com/pydantic/pydantic-ai/pull/4087) trust-filter precedent for `SystemPromptPart`s).
- Streaming dump-path asymmetry documented: streamed `message-metadata` chunks fire only for assistant responses, whereas `dump_messages` populates both request and response sides.

Fixes #5174.

## Tests

- `tests/test_vercel_ai.py` — full dump-load round-trip (`test_adapter_dump_load_roundtrip_with_message_metadata`), `MessageMetadataChunk` emission (`test_event_stream_emits_message_metadata`), single-chunk-per-run invariant in multi-step runs (`test_event_stream_emits_single_message_metadata_per_run`), instruction-injection rejection (`test_adapter_load_rejects_client_supplied_instructions`), malformed-payload rejection (`test_adapter_load_ignores_malformed_pydantic_metadata`), merged-`ModelRequest` app-metadata preservation (`test_adapter_load_preserves_application_metadata_across_merged_messages`), and ~5 more around edge cases (empty `pydantic_ai` block, no target message, app-only metadata, etc.). ~42 existing snapshots updated to use `IsStr()` matchers for runtime timestamps.


pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
```diff
@@ -234,19 +246,25 @@ def deferred_tool_results(self) -> DeferredToolResults | None:
...
+        cls, messages: Sequence[UIMessage], *, preserve_message_metadata: bool = False
+    ) -> list[ModelMessage]:
         """Transform Vercel AI messages into Pydantic AI messages."""
         builder = MessagesBuilder()
 
         for msg in messages:
+            previous_message_count = len(builder.messages)
+            previous_last_message = builder.messages[-1] if builder.messages else None
+            previous_last_parts_count = len(previous_last_message.parts) if previous_last_message else 0
+
```
dsfaccini: do we need to declare these here if they are first used on line 481? do we need to declare these at all, or can we inline them?

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
```diff
@@ -460,8 +478,37 @@ def load_messages(cls, messages: Sequence[UIMessage]) -> list[ModelMessage]:  #
...
+    @staticmethod
+    def _apply_ui_message_metadata(
+        messages: list[ModelMessage],
+        *,
+        role: Literal['system', 'user', 'assistant'],
+        metadata: Any,
+        previous_last_message: ModelMessage | None,
+        previous_last_parts_count: int,
+        previous_message_count: int,
+    ) -> None:
```
dsfaccini: related to the last comment: if these `previous_` variables are all derived from `messages` why do we need to declare them outside and pass as arguments?

tests/test_vercel_ai.py
```diff
@@ -4643,6 +4646,172 @@ async def test_adapter_dump_load_roundtrip_without_timestamps():
...
+    assert request.metadata is None
+    assert request.instructions is None
+    assert response.metadata is None
+    assert not response.usage.has_values()
+    assert response.model_name is None
+    assert response.provider_name is None
+    assert response.provider_url is None
+    assert response.provider_details is None
+    assert response.provider_response_id is None
+    assert response.finish_reason is None
```
dsfaccini: not sure what this proves, if anything a snapshot would be better

tests/test_vercel_ai.py
```diff
@@ -4643,6 +4646,172 @@ async def test_adapter_dump_load_roundtrip_without_timestamps():
...
+            finish_reason='stop',
+            run_id='run-123',
+            conversation_id='conversation-456',
+            metadata={'createdAt': '2026-04-15T12:00:45Z'},
+        ),
+    ]
+
+    ui_messages = VercelAIAdapter.dump_messages(original_messages, preserve_message_metadata=True)
+
+    assert ui_messages[0].metadata == {
```
dsfaccini: snapshots would be better on both of these

dsfaccini:
hey @Genmin I started reviewing but need to reset my system, will have to finish later

dsfaccini:
Claude here: I pushed a follow-up commit (3cba8ba) addressing review feedback. Summary:

- **Dropped the `preserve_message_metadata` flag.** Metadata now round-trips by default to match the existing part-level pattern (`load_provider_metadata`/`dump_provider_metadata`) and DouweM's direction in #5174 ("approach should be similar to how we deal with metadata on parts"). `from_request`, `dispatch_request`, `dump_messages`, `load_messages`, and `VercelAIEventStream` no longer take the flag.

- **Introduced `_PydanticAIMessageMetadata` as the wire schema** for the reserved `pydantic_ai` key. Replaces the hand-rolled `_metadata_str` / `_metadata_dict` / `_metadata_finish_reason` helpers and the bare `try/except Exception` in `_validate_with`. Validation is centralized; `model_dump(mode='json', exclude_defaults=True)` produces clean output and `Z`-suffix timestamps. Field changes need a deprecation cycle.

- **Stopped restoring `ModelRequest.instructions` from client-controlled metadata.** This was a new injection path: `instructions` is behavior-shaping but isn't covered by the `manage_system_prompt` filter (#4087), which strips `SystemPromptPart`s. `_agent_graph.py:822` overwrites `instructions` per-request anyway, so loading the field from history can't change what the model sees but does pollute the merge logic at `_agent_graph.py:2046-2058`. Mirrors the existing trust filter. Dump still includes it for frontend display. New regression test: `test_adapter_load_rejects_client_supplied_instructions`.

- **Stopped emitting duplicate `pydantic_ai` metadata** when a `ModelRequest` splits into separate system + user `UIMessage`s. Metadata now goes only on the trailing one, so reload doesn't re-apply it twice to the merged `ModelRequest`.

- **Lifted the load-side targeting onto `MessagesBuilder`.** New `checkpoint()` + `last_modified(checkpoint, of_type=...)` methods (with public `BuilderCheckpoint`) replace the `previous_message_count` / `previous_last_message` / `previous_last_parts_count` plumbing through `_apply_ui_message_metadata`. Encapsulates the "did the tail grow vs. new message appended" invariants in the class that owns them — keeps Vercel from reaching into builder internals. Additive API; AG-UI is unaffected.

- **Tests**: dropped imports of private `apply_message_metadata` / `dump_message_metadata` (per `tests/CLAUDE.md` "Test through public APIs"); switched line-by-line asserts to `assert == snapshot([...])`; added `test_adapter_load_rejects_client_supplied_instructions` and `test_adapter_load_ignores_malformed_pydantic_metadata`; ran `--inline-snapshot=fix` then swapped runtime-generated `timestamp` / `run_id` / `conversation_id` strings for `IsStr()` matchers across ~42 existing snapshots that gained a `pydantic_ai` block now that metadata is always emitted.

- **Docs**: rewrote the "Message metadata" section in `docs/ui/vercel-ai.md` — drops opt-in language, documents the reserved `pydantic_ai` key, calls out the streaming-only-at-end limitation (chunk fires once per run from `AgentRunResultEvent`, so it only carries fields known at run completion), and explains why `instructions` is intentionally not restored.

dsfaccini:
David's AICA here:

Addressed the latest auto-review round in 9678421d2:

**`instructions` removed from `UIMessage.metadata` entirely** (devin's flag + the type comment)
Previously `instructions` was dumped-but-not-restored. The dump was justified only as "for frontend display" — speculative, no actual consumer. Since the load path already (correctly) refuses to restore it, dumping it served no round-trip purpose and exposed server-side instructions to the client. It's now neither dumped nor restored, and the field is gone from `_PydanticAIMessageMetadata`. This is strictly more aligned with the #4087 trust model. `test_adapter_load_rejects_client_supplied_instructions` stays as the regression guard — a crafted `{'pydantic_ai': {'instructions': ...}}` is dropped by the schema's `extra='ignore'`.

**`BuilderCheckpoint` kept out of the public API**
Removed from `pydantic_ai/ui/__init__.py`. It's only used internally by the Vercel adapter via `builder.checkpoint()` / `last_modified()`; no external consumer needs the type. Can be exported later if a third-party adapter does.

**`dump_message_metadata` dump branch** — converted the `if/else` to explicit `if / elif isinstance(_, ModelResponse) / else: assert_never` per the type-system guideline.

@DouweM these are "trim speculative public/wire surface" calls — happy to revisit if you'd rather keep `instructions` in the dump or `BuilderCheckpoint` exported.

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_utils.py
```diff
@@ -80,6 +117,89 @@ def dump_provider_metadata(
...
+            run_id=message.run_id,
+            conversation_id=message.conversation_id,
+            usage=message.usage if message.usage.has_values() else None,
+            model_name=message.model_name,
+            provider_name=message.provider_name,
+            provider_url=message.provider_url,
+            provider_details=message.provider_details,
+            provider_response_id=message.provider_response_id,
+            finish_reason=message.finish_reason,
+        )
```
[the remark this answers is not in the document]
dsfaccini: David's AICA here: Done in 9678421d2 — converted to `if / elif isinstance(message, ModelResponse) / else: assert_never`.

pydantic_ai_slim/pydantic_ai/ui/__init__.py
```diff
@@ -18,6 +18,7 @@
     'NativeEvent',
     'OnCompleteFunc',
     'MessagesBuilder',
+    'BuilderCheckpoint',
```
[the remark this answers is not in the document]
dsfaccini: David's AICA here: Done in 9678421d2 — `BuilderCheckpoint` removed from `pydantic_ai/ui/__init__.py` (import + `__all__`); it stays an internal class in `_messages_builder.py`. Rationale in https://github.com/pydantic/pydantic-ai/pull/5279#issuecomment-4502806785 — can be exported later if a third-party adapter needs it.

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_utils.py
```diff
@@ -80,6 +117,89 @@ def dump_provider_metadata(
...
+    """
+    metadata = dict(message.metadata) if message.metadata else {}
+
+    if isinstance(message, ModelRequest):
+        pydantic_metadata = _PydanticAIMessageMetadata(
+            timestamp=message.timestamp,
+            run_id=message.run_id,
+            conversation_id=message.conversation_id,
+            instructions=message.instructions,
+        )
```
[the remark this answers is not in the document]
dsfaccini: David's AICA here: Addressed in 9678421d2 — `instructions` is no longer dumped into `UIMessage.metadata` at all (the field is removed from `_PydanticAIMessageMetadata`), so it's neither exposed to nor sourced from client-controlled history. Details in https://github.com/pydantic/pydantic-ai/pull/5279#issuecomment-4502806785.

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_utils.py
```diff
@@ -80,6 +117,89 @@ def dump_provider_metadata(
         return filtered if filtered else None
 
 
+def dump_message_metadata(message: ModelMessage) -> dict[str, Any]:
```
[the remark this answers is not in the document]
dsfaccini: David's AICA here: Keeping the `dict[str, Any]` return — this was a deliberate earlier call on this PR. Returning `dict | None` with the empty→None conversion inside the function reintroduces a type-unreachable branch in the streaming caller: for a `ModelResponse` the dump is never empty (`ModelResponse.timestamp` is non-optional), so an `if metadata := ...` guard there can't take its False branch and tripped CI's `fail_under=100` coverage gate (https://github.com/pydantic/pydantic-ai/actions/runs/25617139825/job/75197582334). With the always-dict return, the single `dump_messages` caller normalizes empty→None at the assignment site and the streaming path drops the guard entirely.

tests/test_vercel_ai.py
```diff
@@ -4808,6 +5181,70 @@ async def event_generator():
...
+        provider_response_id='resp-123',
+        finish_reason='stop',
+        metadata={'createdAt': '2026-04-15T12:00:45Z'},
+    )
+
+    async def event_generator():
+        yield PartStartEvent(index=0, part=response.parts[0])
+        yield PartEndEvent(index=0, part=response.parts[0])
+        result = AgentRunResult(output='Hello')
+        result._state.message_history = [response]  # pyright: ignore[reportPrivateUsage]
```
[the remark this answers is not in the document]
dsfaccini: David's AICA here: Looked into both suggested paths and neither is achievable as stated:

- **"Use a public method to set up the fixture"** — `AgentRunResult` is a dataclass whose only public field is `output`; `message_history` lives in the private `_state` with no public setter. There is no public API to construct an `AgentRunResult` carrying a given message history, so `_state` is the only path.
- **"Restructure to a full end-to-end streaming run"** — `VercelAIAdapter.run_stream()` drives the model's streaming path, and a streamed response (e.g. `FunctionModel` with a `stream_function`) is assembled from text/tool-call deltas that cannot carry response-level fields (`provider_details`, `finish_reason`, `provider_response_id`, application `metadata`, a fixed `timestamp`). A real streamed run therefore can't produce the rich `message-metadata` payload this test exists to verify — it would collapse into a near-duplicate of `test_run_stream_text_and_thinking`.

This test deliberately unit-tests `VercelAIEventStream` (a public, exported class) fed with public event types; the single `result._state.message_history = [...]` line is the narrowly-scoped exception needed to hand a synthetic `ModelResponse` to the stream, and it's already marked with `# pyright: ignore[reportPrivateUsage]`. The rich-field round-trip is separately covered through the public API in `test_adapter_dump_load_roundtrip_with_message_metadata`.

A VCR test against a real provider would be the stronger long-term version (real provider metadata, no `_state`); leaving that as a possible follow-up. Resolving for now.

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
dsfaccini: David's AICA here: Addressed by the `BuilderCheckpoint` refactor — the `previous_*` variables are gone; change-tracking now lives on `MessagesBuilder.checkpoint()` / `last_modified()`. Resolving.

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
dsfaccini: David's AICA here: Addressed by the `BuilderCheckpoint` refactor — the `previous_*` variables are gone; change-tracking now lives on `MessagesBuilder.checkpoint()` / `last_modified()`. Resolving.

tests/test_vercel_ai.py
dsfaccini: David's AICA here: Addressed — these assertions are now snapshot-based. Resolving.

tests/test_vercel_ai.py
dsfaccini: David's AICA here: Addressed — these assertions are now snapshot-based. Resolving.

pydantic_ai_slim/pydantic_ai/ui/__init__.py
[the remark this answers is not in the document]
dsfaccini: David's AICA here: Reopening this — on reflection the private split is internally inconsistent. `MessagesBuilder` is public (it's in `pydantic_ai.ui.__all__`), and so are its `checkpoint()` and `last_modified()` methods. `checkpoint()` *returns* a `BuilderCheckpoint` and `last_modified()` *consumes* one — but the type itself isn't importable, so the public API has a hole in the middle.

A third-party adapter (the exact "if a third-party UI adapter might need it" case) can call the methods, but the moment it factors checkpoint handling into a typed helper it can't spell the parameter type:

```python
from typing import Any

from pydantic_ai.messages import ModelResponse
from pydantic_ai.ui import MessagesBuilder


def attach_response_metadata(builder: MessagesBuilder, checkpoint, metadata: dict[str, Any]) -> None:
    # `checkpoint` is a BuilderCheckpoint — but there is no public name to annotate it with.
    message = builder.last_modified(checkpoint, of_type=ModelResponse)
    if message is not None:
        message.metadata = metadata
```

The options for that `checkpoint` parameter are all bad: leave it unannotated (pyright in strict mode flags it, and the caller loses type safety), or `from pydantic_ai.ui._messages_builder import BuilderCheckpoint` — reaching into a private module. A public method whose parameter/return type is unspeakable is a worse compatibility commitment than just exporting the type.

So I've put it back in `__all__` (importable as `from pydantic_ai.ui import BuilderCheckpoint`). It's a frozen dataclass with a small, stable shape, so the backwards-compat surface is minimal. Happy to revert if you'd still rather keep it internal — flagging the inconsistency so you can weigh in.

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_utils.py
```diff
@@ -29,12 +42,39 @@
...
+    # re-resolved by the agent on every request, not message state to round-trip.
+    # Carrying it here would expose confidential prompt guidance to the client.
+    timestamp: datetime | None = None
+    run_id: str | None = None
+    conversation_id: str | None = None
+
+    usage: RequestUsage | None = None
+    model_name: str | None = None
+    provider_name: str | None = None
+    provider_url: str | None = None
```
DouweM: This will potentially contain sensitive information, so we should drop it.

pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_utils.py
```diff
@@ -29,12 +42,39 @@
...
+    provider_details: dict[str, Any] | None = None
+    provider_response_id: str | None = None
```
DouweM: There's too much potentially sensitive data being sent to the frontend here without explicit opt-in. `timestamp` is harmless, but some others of these are not obviously-safe to either expose to or accept from the client. E.g. `provider_response_id` is used by `OpenAIResponsesModel`'s `openai_previous_response_id='auto'` model setting, so if a user can inject it, they could end up gaining access to another user's conversation (if they had the response ID somehow).

`run_id` and `conversation_id` are also not necessarily safe to accept from the user depending on how the application code handles them.

So we need to limit this PR to only safe fields (like `timestamp`) and make the others opt-in, somehow, in a new PR.

DouweM (changes requested):

dsfaccini:
David's AICA here: addressed in 0b5c1d215 — scope cut to `timestamp` only.

`_PydanticAIMessageMetadata` (and the `dump`/`apply` paths) now carry just `timestamp`. `run_id`, `conversation_id`, `usage`, `model_name`, `finish_reason`, and all four `provider_*` fields are removed from both the outbound dump and the inbound load. The `pydantic_ai` envelope and the `_PydanticAIMessageMetadata` schema with `extra='ignore'` stay so adding fields later is non-breaking.

Tests narrowed accordingly — `test_adapter_dump_load_roundtrip_with_message_metadata` now asserts what's preserved (`timestamp` + application metadata) and what's *not* (server/provider fields). Added `test_adapter_load_pydantic_metadata_without_timestamp` for branch coverage of the valid-block-no-timestamp path. Docs section trimmed to match.

The broader user-controlled opt-in (e.g. `include_metadata: bool | UIIncludeMetadata`) is deferred to a follow-up PR — additive (new optional kwarg), non-breaking against the current state.

CI is running now.

Issue #748, opened 2025-01-22
Gemini causes 'Event loop is closed'  when running inside an async context


asaf:
Hi,

This is an attempt to use `google-gla:gemini-2.0-flash-exp` via streamit in async,

Code below causes `RuntimeError: Event loop is closed`, same code works just fine for `openai:gpt-4o`

```python
import asyncio

import streamlit as st
from pydantic_ai import Agent
from pydantic_ai.messages import ModelRequest, UserPromptPart


async def main():
    st.title("Simple chat")

    # Initialize chat history
    if "messages" not in st.session_state:
        st.session_state.messages = []

    # Display chat messages from history on app rerun
    for message in st.session_state.messages:
        part1 = message.parts[0]
        role = ""
        if isinstance(part1, UserPromptPart):
            role = "user"
        with st.chat_message("human" if role == "user" else "ai"):
            st.markdown(message.parts[0].content)

    # Accept user input
    if prompt := st.chat_input("Say something..."):
        # Display user message in chat message container
        st.chat_message("user").markdown(prompt)

        # Add user message to chat history
        st.session_state.messages.append(
            ModelRequest(parts=[UserPromptPart(content=prompt)])
        )

        # Display assistant response in chat message container
        with st.chat_message("assistant"):
            response_container = st.empty()  # Placeholder for streaming response
            agent = Agent(
                "google-gla:gemini-2.0-flash-exp",
                # "openai:gpt-4o", # works!
                # https://ai.pydantic.dev/results/#structured-result-validation
                result_type=str,  # type: ignore
                system_prompt="you are a nice bot",
            )
            result = await agent.run(prompt)
            response_container.markdown(result.data)


if __name__ == "__main__":
    asyncio.run(main())
```

izzyacademy:
@asaf, it looks like the event loop in GeminiAgentModel, which is responsible for managing and running asynchronous tasks, has been shut down, and the program is trying to reuse it. Once closed, the event loop cannot execute any more tasks.

Please could you share the **full stack trace of the error** so that we can narrow down which lines are the culprit of this problem.

Thanks in advance!

dmontagu:
Unfortunately, I'm not familiar enough with streamlit to be able to reproduce the issue on my own, I tried and just get this:
```
2025-01-22 13:43:09.553 WARNING streamlit.runtime.scriptrunner_utils.script_run_context: Thread 'MainThread': missing ScriptRunContext! This warning can be ignored when running in bare mode.
[...]

PR #5288, opened 2026-05-04
fix(pydantic-graph): replace deprecated asyncio.get_event_loop() with get_running_loop()

### LINKED ISSUE #1196: `pydantic_graph`: `asyncio.get_event_loop()` is deprecated since python 3.12
### Initial Checks

### Description

asyncio.get_event_loop() has been deprecated since version 3.12

 

The following code at [line 15 in pydantic_graph/_utils.py](https://github.com/pydantic/pydantic-ai/blob/main/pydantic_graph/pydantic_graph/_utils.py#L15) emits a deprecation warning
    event_loop = asyncio.get_event_loop()

Python docs recommend using asyncio.get_running_loop() instead.
https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop

### Example Code

```Python
from pydantic_ai import Agent
agent = Agent(TestModel())
agent.run_sync('hi')"


  /home/user/.local/lib/python3.13/site-packages/pydantic_graph/_utils.py:15: DeprecationWarning: There is no current event loop
    event_loop = asyncio.get_event_loop()
```

```Text
python=3.13
pydantic=2.10.6
pydantic-ai-slim=0.0.42
```

Ghraven:
Fixes #1196

## Problem

`asyncio.get_event_loop()` is deprecated since Python 3.12 and emits a `DeprecationWarning` when called outside of a running event loop (e.g. in tests, fresh threads, or synchronous entry points):

```
DeprecationWarning: There is no current event loop
```

This warning is triggered by `get_event_loop()` in `pydantic_graph/_utils.py`.

## Fix

Replace `asyncio.get_event_loop()` with `asyncio.get_running_loop()`, which is the recommended replacement in Python 3.10+:

- When called from within a coroutine or task, `get_running_loop()` returns the running loop immediately — no deprecation warning, no overhead.
- When called outside a running loop (synchronous context, fresh thread), it raises `RuntimeError`, which is caught to create and set a new event loop — the same fallback behaviour as before, without the warning.

A docstring is added explaining the rationale.

## Change

```python
# Before
def get_event_loop():
    try:
        event_loop = asyncio.get_event_loop()
    except RuntimeError:
        event_loop = asyncio.new_event_loop()
        asyncio.set_event_loop(event_loop)
    return event_loop

# After
def get_event_loop():
    try:
        return asyncio.get_running_loop()
    except RuntimeError:
        event_loop = asyncio.new_event_loop()
        asyncio.set_event_loop(event_loop)
        return event_loop
```

No behaviour change for callers — the function still returns a usable event loop in all cases.

### DIFF
--- a/pydantic_graph/pydantic_graph/_utils.py
+++ b/pydantic_graph/pydantic_graph/_utils.py
@@ -51,12 +51,23 @@ except ImportError:  # pragma: no cover
 
 
 def get_event_loop():
+    """Return the running event loop, or create and set a new one if none exists.
+
+    Uses ``asyncio.get_running_loop()`` when called from within a coroutine or
+    callback, falling back to ``asyncio.new_event_loop()`` when there is no
+    running loop (e.g. in a synchronous context or a fresh thread).
+
+    Note:
+        ``asyncio.get_event_loop()`` is deprecated since Python 3.12 and emits
+        a ``DeprecationWarning`` when there is no current event loop. This
+        implementation avoids that warning while preserving the same behaviour.
+    """
     try:
-        event_loop = asyncio.get_event_loop()
+        return asyncio.get_running_loop()
     except RuntimeError:
         event_loop = asyncio.new_event_loop()
         asyncio.set_event_loop(event_loop)
-    return event_loop
+        return event_loop
 
 
 def get_union_args(tp: Any) -> tuple[Any, ...]:


adtyavrdhn:
Is it really a drop in replacement? I am skeptical while dealing with asyncio things.

Ghraven:
Hey @adtyavrdhn, fair question — here's the reasoning:

`asyncio.get_running_loop()` and `asyncio.get_event_loop()` both return the **same running loop** when called from within a coroutine or callback. The difference only matters in a **synchronous / no-loop context**:

- `get_event_loop()` on Python 3.12+: emits `DeprecationWarning` if no current loop is set, then creates one internally anyway
- `get_running_loop()`: raises `RuntimeError` immediately and cleanly when there's no running loop — which we then catch to create a new one explicitly

The end result is identical — a usable event loop is returned in both cases — but the `RuntimeError` path is cleaner and intentional rather than relying on deprecated internal fallback behaviour. The Python docs explicitly recommend this pattern as the replacement. Happy to add a note to the docstring making this more explicit if that would help.

adtyavrdhn:
@Kludex You reverted it(#1204) last time so it isn't straightforward I guess. Any insights here?

Ghraven:
Thanks @adtyavrdhn for flagging PR #1252 — I wasn't aware of that history.

Looking at the revert, the original issue was that `get_running_loop()` breaks when called in a genuinely sync context (no running loop at all), which the previous fix also tried to handle but got wrong in some edge cases.

Our implementation catches `RuntimeError` from `get_running_loop()` and falls back to creating a new loop — so it should be behaviourally equivalent to the old `get_event_loop()` path. But if @Kludex has specific concerns about why that pattern is still incorrect (e.g. thread safety, nested loops, specific Python version edge cases), I'd genuinely like to understand before pushing further.

@Kludex — happy to close this if the consensus is that a simple swap isn't the right fix here, or to rework it based on your guidance. What was the actual failure mode in the reverted PR?

Ghraven:
Gentle ping! This replaces the deprecated `asyncio.get_event_loop()` call which raises a `DeprecationWarning` in Python 3.12+. Happy to adjust the approach if needed. Thanks!

adtyavrdhn:
Hey @Ghraven 

I am going to have to look deeper into this to make sure we don't break anything which we might not have a test for, it isn't something I consider to be swappable as is when Marcelo already tried it once and it didn't go super well.

Kludex:
I mean, look at the commits that added and removed it. What was the reason I did that? I'm pretty sure there are issues linked that point the reason.

Ghraven:
Thanks @Kludex, you're right — I went and read the history:

- #1204 made exactly the swap I'm proposing (`get_event_loop` → `get_running_loop` with `RuntimeError` fallback).
- #1252 reverted it because of #748 — the "Event loop is closed" failures with Gemini + Streamlit (and similar sync-entry-point contexts), which appear to be tied to httpx connection pooling reusing a connection across event loops.

So the simple swap isn't actually behaviour-preserving in those edge cases — my fallback of `new_event_loop()` + `set_event_loop()` can land users in the same broken state that #1204 did, just with an extra log line.

Given that, I think the right move is to close this PR. The deprecation warning is real but the safe fix probably needs a different shape (e.g. detecting the streaming-from-sync case and not creating a new loop at all, or scoping the loop change more narrowly). Happy to take another look in a separate PR if you'd want that — but only if you think it's worth pursuing. Closing now to keep the queue clean. Thanks for the pointer!

Issue #5132, opened 2026-04-17
Cancellation cleanup in run_stream_events returns before internal tasks finish


Wh1isper:
### Initial Checks

- [x] I confirm that I'm using the latest version of Pydantic AI for the original integration issue
- [x] I confirm that I searched for existing issues before opening this report

### Note on authorship

I am filing this as an AI coding assistant working with a user, and I want to be explicit that this report is bot-authored. I have tried to keep it factual and scoped to what I could reproduce directly.

### Summary

I found a stable cancellation-cleanup problem in `run_stream_events()`.

When the consumer task is cancelled while events are still streaming:

- the consumer task finishes immediately
- internal Pydantic AI tasks are still pending for a short time afterwards
- cleanup of the underlying `request_stream(...).__aexit__` happens later
- `run_agent()` can finish with an unhandled `anyio.BrokenResourceError`

This looks related to the broader cancellation work discussed in #1524.

I also saw a deeper failure in a real TUI integration involving:

- `Task was destroyed but it is pending!`
- `ValueError: <Token var=<ContextVar name='pydantic_ai.current_run_context' ...> was created in a different Context`

I do not yet have a standalone reproducer for that exact traceback, so this issue focuses on the stable reproduction below.

### Environment

- `pydantic-ai`: `1.81.0`
- Python: `3.13.7`
- OS: Linux

### Stable reproduction

```python
import asyncio
from contextlib import asynccontextmanager
from datetime import datetime

from pydantic_ai import Agent
from pydantic_ai.messages import ModelMessage
from pydantic_ai.models import Model, StreamedResponse
from pydantic_ai.tools import RunContext
from pydantic_ai.usage import RequestUsage

cleanup_finished = asyncio.Event()


class FinalTextStream(StreamedResponse):
    async def _get_event_iterator(self):
        self._usage = RequestUsage()
        for event in self._parts_manager.handle_text_delta(vendor_part_id=0, content='hello '):
            yield event
        await asyncio.sleep(0.2)
        for event in self._parts_manager.handle_text_delta(vendor_part_id=0, content='world'):
            yield event

    @property
    def model_name(self):
        return 'slow'

    @property
    def provider_name(self):
        return 'test'

    @property
    def provider_url(self):
        return 'https://test'

    @property
    def timestamp(self):
        return datetime.now()


class SlowCleanupModel(Model):
    @property
    def system(self):
        return 'test'

    @property
    def model_name(self):
        return 'slow'

    @property
[...]

PR #5291, opened 2026-05-04
Stop draining streamed responses after early exit

### LINKED ISSUE #1524: Stream cancellation: broader semantics for agent runs, tools, and event streams

adtyavrdhn:
- Closes #1524
- Related #5132

### Summary

This is a narrower replacement for #5031. The previous PR tried to define explicit cancellation behavior for streams, event streams, and runs. That quickly runs into unresolved semantics around whether cancellation means "stop this model response stream", "stop the whole agent graph", or "interrupt currently running tool calls".

This PR only fixes the smaller bug from #1524: if a caller exits `run_stream()` early, Pydantic AI should not keep consuming the provider stream in the background.

The main changes are:

- `ModelRequestNode.stream()` no longer drains `AgentStream` to EOF after yielding it to the caller.
- Exiting the stream context now lets the underlying `model.request_stream(...)` context manager unwind instead of continuing to pull chunks for usage accounting.
- `PeekableAsyncStream` now supports `aclose()`, and `FunctionModel`, Google, HuggingFace, and xAI streaming request paths close their async iterators in `finally` blocks.
- `run_stream_events()` now returns a small async context manager wrapper, so callers can use `async with agent.run_stream_events(...) as events:` for deterministic producer cleanup. This does not add a `cancel()` method or define run-level cancellation semantics.

### Why this is different from #5031

This PR does not add:

- `StreamedRunResult.cancel()`
- `AgentEventStream.cancel()`
- `RunCancelled`
- `cancel(end_run=...)`
- cancellation semantics for tool execution
- synthetic tool-return messages for interrupted tool calls
- an incomplete/interrupted response field in serialized message history

Those are still worth discussing, but they are larger API and behavior questions. This PR is intentionally only about not continuing to read a provider stream after the user stops consuming it.

### What happens to the socket/underlying provider stream?

This PR does not expose a new public cancellation hook, but it still closes the stream by unwinding the existing provider context.

The flow is:

1. The user exits `async with agent.run_stream(...)` or breaks from the streamed output and leaves the context.
2. `ModelRequestNode.stream()` no longer drains the `AgentStream` to EOF.
3. `stream_done` is set.
4. The internal `_streaming_handler` resumes and exits `async with req_ctx.model.request_stream(...) as sr:`.
5. The provider's `request_stream()` cleanup runs.

For providers that already wrap SDK streams in an async context manager, such as OpenAI, Anthropic, Groq, Mistral, and the custom Gemini path, that means the SDK/http response context exits. For Google GenAI, HuggingFace, and xAI, this PR adds best-effort `aclose()` cleanup around the runtime async iterator. xAI in particular is still best-effort: the SDK stream object exposes async-generator cleanup locally, but this PR does not claim a documented transport-level RPC cancellation hook.

The important behavior is that Pydantic AI itself stops pulling chunks. If a provider SDK properly maps stream close/finalization to connection close, token generation should stop there. If a provider SDK does not expose real transport cancellation, this PR still avoids our previous behavior of actively reading the rest of the response.

### Tests

- `uv run pytest tests/test_streaming.py`
- `uv run pytest tests/test_messages.py tests/test_streaming.py`
- `uv run pytest tests/test_direct.py -k "model_request_stream"`
- `uv run ruff check pydantic_ai_slim/pydantic_ai/_agent_graph.py pydantic_ai_slim/pydantic_ai/_utils.py pydantic_ai_slim/pydantic_ai/agent/abstract.py pydantic_ai_slim/pydantic_ai/models/function.py pydantic_ai_slim/pydantic_ai/models/google.py pydantic_ai_slim/pydantic_ai/models/huggingface.py pydantic_ai_slim/pydantic_ai/models/xai.py pydantic_ai_slim/pydantic_ai/result.py pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py pydantic_ai_slim/pydantic_ai/durable_exec/prefect/_agent.py pydantic_ai_slim/pydantic_ai/durable_exec/dbos/_agent.py tests/test_streaming.py`
- `uv run pyright pydantic_ai_slim/pydantic_ai/_agent_graph.py pydantic_ai_slim/pydantic_ai/_utils.py pydantic_ai_slim/pydantic_ai/agent/abstract.py pydantic_ai_slim/pydantic_ai/messages.py pydantic_ai_slim/pydantic_ai/models/__init__.py pydantic_ai_slim/pydantic_ai/models/function.py pydantic_ai_slim/pydantic_ai/models/google.py pydantic_ai_slim/pydantic_ai/models/huggingface.py pydantic_ai_slim/pydantic_ai/models/xai.py pydantic_ai_slim/pydantic_ai/result.py tests/test_streaming.py`

The local commit hook's full-repo pyright target was not usable in this fresh worktree because optional example/durable dependencies such as FastAPI, Prefect, DBOS, pandas, and others were not installed. The targeted pyright checks above passed.

### Checklist

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/_agent_graph.py
+++ b/pydantic_ai_slim/pydantic_ai/_agent_graph.py
@@ -666,11 +666,9 @@ class ModelRequestNode(AgentNode[DepsT, NodeRunEndT]):
 
         # Normal path: handler was called, stream is ready
         stream_error: BaseException | None = None
+        agent_stream = agent_stream_holder[0]
         try:
-            yield agent_stream_holder[0]
-            # Ensure stream is fully consumed for proper usage counting
-            async for _ in agent_stream_holder[0]:
-                pass
+            yield agent_stream
         except BaseException as exc:
             stream_error = exc
             raise
--- a/pydantic_ai_slim/pydantic_ai/agent/abstract.py
+++ b/pydantic_ai_slim/pydantic_ai/agent/abstract.py
@@ -1060,8 +1060,9 @@ class AbstractAgent(Generic[AgentDepsT, OutputDataT], ABC):
 
         async def main():
             events: list[AgentStreamEvent | AgentRunResultEvent] = []
-            async for event in agent.run_stream_events('What is the capital of France?'):
-                events.append(event)
+            async with agent.run_stream_events('What is the capital of France?') as stream:
+                async for event in stream:
+                    events.append(event)
             print(events)
             '''
             [
@@ -1079,7 +1080,7 @@ class AbstractAgent(Generic[AgentDepsT, OutputDataT], ABC):
         ```
 
         Arguments are the same as for [`self.run`][pydantic_ai.agent.AbstractAgent.run],
-        except that `event_stream_handler` is now allowed.
+        except that `event_stream_handler` is managed internally by this method.
 
         Args:
             user_prompt: User input to start/continue the conversation.
@@ -1105,8 +1106,8 @@ class AbstractAgent(Generic[AgentDepsT, OutputDataT], ABC):
             spec: Optional agent spec to apply for this run. At run time, spec values are additive.
 
         Returns:
-            An async iterable of stream events `AgentStreamEvent` and finally a `AgentRunResultEvent` with the final
-            run result.
+            An [`AgentEventStream`][pydantic_ai.result.AgentEventStream], which is both an async iterator of stream
+            events and an async context manager for deterministic cleanup.
         """
         if infer_name and self.name is None:
             self._infer_name(inspect.currentframe())
--- a/pydantic_ai_slim/pydantic_ai/agent/abstract.py
+++ b/pydantic_ai_slim/pydantic_ai/agent/abstract.py
@@ -1203,8 +1203,7 @@ class AbstractAgent(Generic[AgentDepsT, OutputDataT], ABC):
             raise
 
         except BaseException:
-            if not task.done():
-                task.cancel()
+            task.cancel()
 
             # The consumer side is already exiting. Await the producer only to
             # retrieve its exception and finish cleanup; it must not replace the
--- a/pydantic_ai_slim/pydantic_ai/agent/abstract.py
+++ b/pydantic_ai_slim/pydantic_ai/agent/abstract.py
@@ -5,7 +5,7 @@ import inspect
 from abc import ABC, abstractmethod
 from collections.abc import AsyncIterable, AsyncIterator, Awaitable, Callable, Iterator, Mapping, Sequence
 from concurrent.futures import Executor
-from contextlib import AbstractAsyncContextManager, asynccontextmanager, contextmanager
+from contextlib import AbstractAsyncContextManager, asynccontextmanager, contextmanager, suppress
 from types import FrameType
 from typing import TYPE_CHECKING, Any, Generic, TypeAlias, cast, overload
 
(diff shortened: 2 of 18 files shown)


pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -666,11 +666,9 @@ async def _streaming_handler(
 
         # Normal path: handler was called, stream is ready
         stream_error: BaseException | None = None
+        agent_stream = agent_stream_holder[0]
         try:
-            yield agent_stream_holder[0]
-            # Ensure stream is fully consumed for proper usage counting
-            async for _ in agent_stream_holder[0]:
```
adtyavrdhn, line 672: This is the biggest change we need to talk about, not draining the stream.

tests/test_usage_limits.py
```diff
@@ -143,6 +143,8 @@ async def ret_a(x: str) -> str:
                     tool_calls=1,
                 )
             )
+            async for _ in result.stream_text(debounce_by=None):
```
adtyavrdhn, line 146: This was needed to keep the test behaviour same but the point needs to be discussed. If we do not drain the usage will be different but draining after the user has exited the stream context is incorrect behaviour.

pydantic_ai_slim/pydantic_ai/models/google.py
```diff
@@ -538,7 +538,10 @@ async def request_stream(
         )
         model_settings = cast(GoogleModelSettings, model_settings or {})
         response = await self._generate_content(messages, True, model_settings, model_request_parameters)
-        yield await self._process_streamed_response(response, model_request_parameters)  # type: ignore
+        try:
+            yield await self._process_streamed_response(response, model_request_parameters)  # type: ignore
```
adtyavrdhn, line 542: We need this because when the response is in `async with response` we can rely on the provider to close and run the exits when the user leaves this context. Same cannot be true for generators who lack this, in this case we do a best effort basis of closing the generator from our end.

tests/test_usage_limits.py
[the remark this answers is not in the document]
adtyavrdhn: Yes this is a change that we have to talk about here.

pydantic_ai_slim/pydantic_ai/agent/abstract.py
```diff
@@ -1197,7 +1200,14 @@ async def run_agent() -> AgentRunResult[Any]:
 
         except asyncio.CancelledError as e:
             task.cancel(msg=e.args[0] if len(e.args) != 0 else None)
+            task_cancel_sent = True
             raise
+        finally:
+            if not task.done() and not task_cancel_sent:
+                task.cancel()
+
+            with suppress(asyncio.CancelledError, anyio.BrokenResourceError, anyio.ClosedResourceError):
```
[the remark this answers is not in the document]
adtyavrdhn: So much text :// 

Okay so await task in finally overwrites the producer's error if it was a non cancelled error, sneaky thing.  

If there is an exception when we are in finally then we should retain the original one otherwise raise as is because the exception was from the producer task which we have no business catching.

pydantic_ai_slim/pydantic_ai/models/google.py
adtyavrdhn: Yeah so this wouldn't always work if we just rely on CPython to clean up generators when no one references it.

PR #5293, opened 2026-05-04
fix(fastmcp): respect is_error tool results

### LINKED ISSUE #5217: MCP isError:true is indiscriminately converted to ModelRetry — non-retryable tool errors should reach the LLM as informational results
### Initial Checks

### Description

When an MCP tool returns `isError: true`, `direct_call_tool` unconditionally raises `ModelRetry`:

```python
# pydantic_ai/mcp.py — direct_call_tool
if result.isError:
    text = "\n".join(str(part) for part in content)
    raise ModelRetry(text)
```

This treats **all** tool errors as retryable — the LLM is told "try again" regardless of the error's nature. The `isError: true` flag is set for every non-2xx HTTP response from an MCP backend (via FastMCP's `raise_for_status()` → exception → `CallToolResult(isError=true)`), so all 4xx and 5xx responses become `ModelRetry`.

**What goes wrong:**

The error is delivered to the LLM as a retry prompt rather than an informational tool result. This makes it harder for the LLM to understand the nature of the failure, and encourages it to retry the same call instead of informing the user or choosing a different approach. After `max_retries` is exhausted, the run crashes with `UnexpectedModelBehavior` and the error context is lost entirely.

**The distinction that's missing** — MCP tool errors should be split into retryable and non-retryable:

| Category | Examples | What the LLM should do |
|----------|---------|----------------------|
| **Retryable** | Connection timeout, 429 Too Many Requests, 502/503 | Retry (current behavior is correct) |
| **Non-retryable** | 400, 404, 422 "Transcript too short", "Unsupported language" | Inform the user, choose a different approach |

The MCP spec only provides the `isError` boolean — it doesn't prescribe retry semantics. But pydantic-ai's blanket conversion to `ModelRetry` assumes all errors are retryable, which is incorrect for 4xx client errors and application-level failures.

Anthropic's own [Certified Architect guide](https://github.com/paullarionov/claude-certified-architect/blob/main/guide_en.MD) explicitly recommends structured error responses with `errorCategory` (transient/validation/permission) and `isRetryable` fields — but pydantic-ai doesn't inspect any of this.

**Proposed solution:**

Change the default: return `isError` content as a tool result with `outcome='failed'` instead of raising `ModelRetry`. The `outcome` field on `ToolReturnPart` already exists (#4363). The LLM sees the error message and can decide what to do — inform the user, try a different approach, or call the tool again on its own. Protocol-level `McpError` exceptions (connection failures, JSON-RPC errors) would still raise as they do today.

As an optional extension: add a configurable hook on MCP server constructors for consumers who need custom retryability logic (e.g. "retry 502s but surface 422s").

**Related issues:**

- #4114 (`ToolFailed` exception) — would have allowed tools to fail without killing the run; went stale
- #4363 (`ToolReturnPart` `outcome` field) — added `outcome` but there's no way to produce `outcome='failed'` from tool code today
- #1989 — the opposite problem (McpError crashing the run); fix added `ModelRetry`, which this issue argues went too far
- #2586 — original request for tool failure handling

```python
from fastapi import FastAPI
from fastapi.responses import JSONResponse
from fastmcp import FastMCP
from pydantic_ai import Agent
from pydantic_ai.mcp import MCPServerHTTP

# Backend: FastAPI app that returns 422 for a known condition
app = FastAPI()

@app.post("/check")
async def check(input: dict):
    if input.get("value") == "bad":
        return JSONResponse(status_code=422, content={"error": "InvalidInput", "detail": "Value is not acceptable"})
    return {"result": "ok"}

mcp = FastMCP.from_fastapi(app=app)

# Agent side: MCP tool call hits 422
# Expected: LLM sees the error detail and can inform the user
# Actual: ModelRetry is raised, LLM retries futilely, run crashes after max_retries
agent = Agent("openai:gpt-4o", mcp_servers=[MCPServerHTTP(url="http://localhost:8000/mcp")])
result = agent.run_sync("Check the value 'bad'")
# → UnexpectedModelBehavior: Exceeded maximum retries
# The LLM never saw "Value is not acceptable"
```

- Python: 3.12
- Pydantic AI: 1.80.0
- LLM provider SDK: openai 1.x / anthropic 0.x

fengfeng-zi:
## Summary
- handle FastMCP tool results with `is_error=True` in `direct_call_tool()`
- map the error result through `tool_error_behavior` so non-retryable tool errors do not silently bypass policy
- add focused regression tests for both `model_retry` and `error` behaviors

Closes #5217

## Testing
- `uv run pytest tests/test_fastmcp.py -k "is_error_result or error_behavior"`

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/toolsets/fastmcp.py
+++ b/pydantic_ai_slim/pydantic_ai/toolsets/fastmcp.py
@@ -255,6 +255,12 @@ class FastMCPToolset(AbstractToolset[AgentDepsT]):
                 else:
                     raise e
 
+        if call_tool_result.is_error:
+            message = '\n'.join(part.text for part in call_tool_result.content if isinstance(part, TextContent))
+            if self.tool_error_behavior == 'model_retry':
+                raise ModelRetry(message=message or 'MCP tool call failed')
+            raise ToolError(message or 'MCP tool call failed')
+
         # Prefer structured content if there are only text parts, which per the docs would contain the JSON-encoded structured content for backward compatibility.
         # See https://github.com/modelcontextprotocol/python-sdk#structured-output
         if (structured := call_tool_result.structured_content) and all(
--- a/tests/test_fastmcp.py
+++ b/tests/test_fastmcp.py
@@ -6,6 +6,7 @@ import base64
 from pathlib import Path
 from tempfile import TemporaryDirectory
 from typing import Any
+from unittest.mock import AsyncMock
 
 import pytest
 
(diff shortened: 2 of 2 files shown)


DouweM:
This does not match the goal of the issue; we should send back `ToolReturnPart` with `outcome='error'` (and possibly communicate that failure to the model somehow), NOT to raise a `ToolError` that may either be rewrapped by `ModelRetry`/`RetryPromptPart` (incorrectly, per that issue) or be bubbled up to the user.

PR #5292, opened 2026-05-04
fix(vercel_ai): emit `tool-input-(available|error)` on `FunctionToolCallEvent`

sadra-barikbin:
# fix(vercel_ai): emit `tool-input-(available|error)` on `FunctionToolCallEvent`

## Context

In the Vercel AI adapter, `ToolInputAvailableChunk` was emitted from
`handle_tool_call_end`, which fires on `PartEndEvent` for a `ToolCallPart` —
i.e. as soon as the model finishes streaming the call, *before* validation.
That has two consequences:

1. For deferred / client-side tools, where the chunk effectively acts as a
   "go execute this" signal, the frontend could be handed args the server
   would have rejected.
2. There was no way to surface a validation *failure* as a distinct chunk —
   the input chunk fired regardless.

This change moves the emission to `handle_function_tool_call`, which fires
post-validation in `_agent_graph.py`, and tightens a few related loose ends.

### Vercel AI adapter — `pydantic_ai/ui/vercel_ai/_event_stream.py`

- Replace `handle_tool_call_end` emission with `handle_function_tool_call`.
- On `args_valid=True`: emit `ToolInputAvailableChunk`. Use
  `event.validated_args` for `input` when populated, falling back to
  `part.args_as_dict()` (so the frontend sees the *coerced* value).
- On `args_valid=False`: emit `ToolInputErrorChunk` with `errorText` from
  `validation_error.tool_retry.model_response()`.
- Skip emission entirely when `args_valid is None` *and* no
  `validation_error` — covers resume of non-`ToolApproved` deferred results
  (`ToolDenied`, `ModelRetry`, direct return) and the output-tool
  end-strategy-skipped path. Re-announcing "input available" for those would
  be misleading. The `args_valid=False, validation_error=None` case
  (`UnexpectedModelBehavior` raised during validation) is also skipped — the
  outer error path closes the pending tool call with `tool-output-error`.

### Event surface — `pydantic_ai/messages.py`

- Add `validated_args: dict[str, Any] | None` and
  `validation_error: ToolRetryError | None` to `FunctionToolCallEvent`,
  mirroring `ValidatedToolCall`.
- `validation_error` is `compare=False`: the same information is carried by
  the following `FunctionToolResultEvent.RetryPromptPart`, and including it
  in equality breaks snapshot tooling (since `ToolRetryError` doesn't roundtrip
  cleanly through `copy.deepcopy` + repr).
- `validated_args` carries the documented "mild lie" — it's typed as a dict
[...]

PR #5320, opened 2026-05-06
Yield `OutputToolCallEvent`/`OutputToolResultEvent` for output tool calls; deprecate function-tool events for failing output tool calls

### LINKED ISSUE #2640: Failed to yield ToolReturnPart streaming with structured output type via tool call
### Initial Checks

### Description

When building an agent that uses tool calls to handle structured outputs, the final result tool return is never streamed back, it is only returned when the complete list of messages is returned. Specifically this part https://github.com/pydantic/pydantic-ai/blob/f5f7b8075a93f2a0e62f754f75d07562e60bf552/pydantic_ai_slim/pydantic_ai/_agent_graph.py#L636. The tool return is created and added to the outputs but not yielded.

This may have been on purpose, and I'd like to understand why. In most use cases, structured outputs are one-shot runs where there is no chat back with the model. So not including the tool return is fine. But I wanted to chat with the model and assumed the streamed responses would give me an equivalent history for chat applications as calling `all_messages()`.

Without yielding the tool return, the streamed responses cause a dangling tool call which causes issues if you try to feed that back as message history. 

Happy to submit a PR to add the yielding if you folks want. Let me know what you think!

### Example Code

```Python
# Run in a notebook, hence nest_asyncio
# Modified from https://ai.pydantic.dev/agents/#streaming-all-events-and-output
import asyncio
from pydantic_ai import Agent, StructuredDict
from pydantic_ai.messages import FinalResultEvent, FunctionToolCallEvent, FunctionToolResultEvent
import nest_asyncio

nest_asyncio.apply()

async def run_agent(agent, prompt):
    output_messages = []
    # Begin a node-by-node, streaming iteration
    async with agent.iter(prompt) as run:
        async for node in run:
            if Agent.is_user_prompt_node(node):
                # A user prompt node => The user has provided input
                output_messages.append(f'=== UserPromptNode: {node.user_prompt} ===')
            elif Agent.is_model_request_node(node):
                # A model request node => We can stream tokens from the model's request
                async with node.stream(run.ctx) as request_stream:
                    async for event in request_stream:
                        pass
                    output_messages.append(request_stream.get())
            elif Agent.is_call_tools_node(node):
                # A handle-response node => The model returned some data, potentially calls a tool
                async with node.stream(run.ctx) as handle_stream:
                    async for event in handle_stream:
                        if isinstance(event, FunctionToolCallEvent):
                            output_messages.append(event)
                        elif isinstance(event, FunctionToolResultEvent):
                            output_messages.append(event.result)
            elif Agent.is_end_node(node):
                # Once an End node is reached, the agent run is complete
                assert run.result is not None
                assert run.result.output == node.data.output
                output_messages.append(run.result.output)
    return output_messages


class TestSchema(BaseModel):
    user_name: str
    user_description: str


json_schema = TestSchema.model_json_schema()
agent = Agent(
    "openai:gpt-4o",
    system_prompt="Either ask for clarification or respond with the final result tool.",
    output_type=StructuredDict(json_schema),
)
# Full run sync to get all the messages
result = agent.run_sync("Hi")

for msg in result.all_messages():
    # The first message with be ModelRequest for system promptand user prompt
    # Second is the tool call
    # Third is the tool return
    print(msg)
    print()

print("==============")
msgs = asyncio.run(run_agent(agent, "Hi"))
for msg in msgs:
    # First is user prompt
    # Second is tool call
    # Third is just output - no tool return
    print(msg)
    print()

"""RESULTS
ModelRequest(parts=[SystemPromptPart(content='Either ask for clarification or respond with the final result tool.', timestamp=datetime.datetime(2025, 8, 22, 5, 28, 23, 285240, tzinfo=datetime.timezone.utc)), UserPromptPart(content='Hi', timestamp=datetime.datetime(2025, 8, 22, 5, 28, 23, 285246, tzinfo=datetime.timezone.utc))])

ModelResponse(parts=[ToolCallPart(tool_name='final_result', args='{"user_name":"Hello, how can I assist you today?","user_description":"User greeted."}', tool_call_id='call_xxFXtp52ubL9XPdC0XK21EbD')], usage=RequestUsage(input_tokens=64, output_tokens=30, details={'accepted_prediction_tokens': 0, 'audio_tokens': 0, 'reasoning_tokens': 0, 'rejected_prediction_tokens': 0}), model_name='gpt-4o-2024-08-06', timestamp=datetime.datetime(2025, 8, 22, 5, 28, 23, tzinfo=TzInfo(UTC)), provider_request_id='chatcmpl-C7Ec3CbnlDC0oZGYKS7xxQQdFtaCB')

ModelRequest(parts=[ToolReturnPart(tool_name='final_result', content='Final result processed.', tool_call_id='call_xxFXtp52ubL9XPdC0XK21EbD', timestamp=datetime.datetime(2025, 8, 22, 5, 28, 24, 19144, tzinfo=datetime.timezone.utc))])

==============
=== UserPromptNode: Hi ===

ModelResponse(parts=[ToolCallPart(tool_name='final_result', args='', tool_call_id='call_Dt32fw0NnEoOKhPLdtpSeFsi')], usage=RequestUsage(details={'accepted_prediction_tokens': 0, 'audio_tokens': 0, 'reasoning_tokens': 0, 'rejected_prediction_tokens': 0}), model_name='gpt-4o', timestamp=datetime.datetime(2025, 8, 22, 5, 28, 24, tzinfo=TzInfo(UTC)))

{'user_name': 'Hi', 'user_description': 'An initial greeting message.'}
"""
# I am expecting a toolreturnpart object in the second response just like in the first
```

```Text
Python version: 3.13.5 (main, Jul 11 2025, 22:26:07) [Clang 20.1.4 ]
Pydantic AI version: 0.7.4
LLM Client OpenAI version: 1.100.2
```

DouweM:
- Closes #2640
- Supersedes #5052

Output tool calls (the model's "submit final answer" call) didn't fire any call/result events on the success path, so consumers building agent runs from the event stream saw a dangling tool call without a result. On failure paths (validation/execution retry, skipped due to existing final result) they reused `FunctionToolCallEvent` / `FunctionToolResultEvent`, which conflated user-defined function tools with the conceptually-distinct output tool.

This adds two new event classes:

- [`OutputToolCallEvent`][pydantic_ai.messages.OutputToolCallEvent] — yielded when an output tool is about to be invoked
- [`OutputToolResultEvent`][pydantic_ai.messages.OutputToolResultEvent] — yielded once the output tool has completed

Plus shared base classes [`ToolCallEvent`][pydantic_ai.messages.ToolCallEvent] / [`ToolResultEvent`][pydantic_ai.messages.ToolResultEvent] so handlers that want unified treatment can match both function and output tool events in a single `case` (useful for logging, telemetry, or rebuilding message history from events alone). Match against the specific subclass when you need to discriminate.

`process_tool_calls` yields the new pair on every output-tool branch (success, validation/execution failure, early-strategy skip), so message history rebuilt from events alone no longer has dangling tool calls.

UI adapters (AG-UI, Vercel AI, base [`UIEventStream`][pydantic_ai.ui.UIEventStream]) route output-tool results through new `handle_output_tool_call` / `handle_output_tool_result` hooks. AG-UI and Vercel AI handle them the same as function tool results today; subclasses can override to discriminate.

[`FunctionToolResultEvent.result`][pydantic_ai.messages.FunctionToolResultEvent] is renamed to `part` (parallel to `ToolCallEvent.part`); `result` stays as a `@deprecated` read-side alias.

#### Backward compatibility during the v2 transition

Output-tool failure paths (skipped because a final result already exists, or validation/execution failure triggering a retry) still emit the legacy `FunctionToolCallEvent` / `FunctionToolResultEvent` alongside the new events, so existing `isinstance(event, FunctionToolCallEvent)` consumers keep seeing them. The legacy emission for output tool calls will stop in v2 — switch to `OutputToolCallEvent` / `OutputToolResultEvent` (or the shared `ToolCallEvent` / `ToolResultEvent` bases) before then. The UI adapter base class dedupes the dual emission internally so downstream protocol streams (AG-UI, Vercel AI) only see one result per tool call.

#### Why separate classes (not a shared `ToolCallEvent` with a `kind` flag)

- Aligns with the existing `event_kind` discriminator pattern — there's no precedent for a within-class sub-discriminator, and a single class with two discriminator fields would be awkward under \`pydantic.Discriminator\`.
- Output tools are conceptually distinct (the model's "submit final answer" call vs. a user-defined function call). For streaming UIs in particular, presenting an output-tool call as a "tool call" misleads end users.
- The shared base classes preserve the unified-handling option for consumers that want it.

### Checklist


pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -1475,15 +1506,23 @@ async def process_tool_calls(  # noqa: C901
...
                 assert validated.validation_error is not None
                 if final_result:
-                    for event in _emit_skipped_output_tool(
-                        call, 'Output tool not used - output failed validation.', output_parts, args_valid=False
+                    for event in _emit_output_tool_events(
+                        call,
+                        'Output tool not used - output failed validation.',
+                        output_parts,
+                        args_valid=False,
+                        legacy=True,
```
DouweM: Do we need the legacy param?

docs/capabilities.md
```diff
@@ -806,15 +806,20 @@ class StreamAuditor(AbstractCapability[Any]):
         stream: AsyncIterable[AgentStreamEvent],
     ) -> AsyncIterable[AgentStreamEvent]:
         async for event in stream:
-            if isinstance(event, FunctionToolCallEvent):
+            if isinstance(event, BaseToolCallEvent):
```
DouweM: This make it seem like these mirror to `BaseToolCallPart`. But really it mirrors to `ToolCallPart`, there is no event for `BuiltinToolCallPart` (for good reason). So maybe call it `ToolCallEvent`?

docs/capabilities.md
```diff
@@ -806,15 +806,20 @@ class StreamAuditor(AbstractCapability[Any]):
...
                 print(f'Tool called: {event.part.tool_name}')
-            elif isinstance(event, FunctionToolResultEvent):
-                print(f'Tool result: {event.tool_return.content!r}')
+            elif isinstance(event, BaseToolResultEvent):
+                print(f'Tool result: {event.result.content!r}')
```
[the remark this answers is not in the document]
DouweM: Good reason to drop `test=skip`

tests/test_ui.py
```diff
@@ -613,14 +622,46 @@ def bad_output(value: str) -> str:
...
+    agent = Agent(model=FunctionModel(stream_function=stream_function), output_type=OutputType)
+
+    request = DummyUIRunInput(messages=[ModelRequest.user_text_prompt('Hello')])
+    adapter = DummyUIAdapter(agent, request)
+    events = [event async for event in adapter.run_stream()]
+
+    # The first (failed) tool call should produce ONE `<output-tool-result>`, not two — the legacy
+    # `FunctionToolResultEvent` for `out_1` is dedupped at the UI layer.
+    assert sum(1 for e in events if "<output-tool-result name='final_result'>" in e) == 2
+    assert sum(1 for e in events if "<function-tool-result name='final_result'>" in e) == 0
```
DouweM: I wanna se the snapshot of the entire history as well

tests/models/test_bedrock.py
```diff
@@ -3853,6 +3855,20 @@ class Response(TypedDict):
                     provider_details={'status': 'success'},
                 )
             ),
+            OutputToolCallEvent(
+                part=ToolCallPart(
+                    tool_name='final_result', args='{"result":7006652.0}', tool_call_id='tooluse_ptgCcZ0uQu-UUMz0abqoWw'
+                ),
+                args_valid=True,
```
DouweM, line 3862: This does kind of ask for a `validated_args` field like what https://github.com/pydantic/pydantic-ai/pull/5292 is also trying to add (read that with a subagent, it's related to us). Not really scope I wanna pull in necessarily, though, but would be useful on the API surface...?

tests/models/test_bedrock.py
```diff
@@ -3853,6 +3855,20 @@ class Response(TypedDict):
...
+            ),
+            OutputToolResultEvent(
+                result=ToolReturnPart(
```
DouweM: I don't think `result` is the best name here. If we're adding new types (and going to do v2 anyway), can we name it something better? If there is a `result` key on here, I'd expect it it to take the final value of the output that was returned from the agent -- as in post output functions etc. If we're adding these events, let's get the naming right. We can change it on the base event as well and then deprecate the old name on the `FunctionToolResult` event. I don't think we _necessarily_ need to store the final processed `output` here as well, but if it's easy, let's. Then we have `part` (symmetrical with `ToolCallEvent`) and `output`. `FunctionToolResultEvent` already makes it easily available through `.part.content`, but Output tool event doesn't. (Leaving it off is OK too, as there are other ways to get the agent's final output, and it's not sent to the model.)

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -2505,29 +2509,58 @@ class FunctionToolCallEvent:
...
         return self.part.tool_call_id  # pragma: no cover
 
-    __repr__ = _utils.dataclasses_no_defaults_repr
+
+@dataclass(repr=False)
+class OutputToolCallEvent(BaseToolCallEvent):
+    """An event indicating the start of a call to an output tool (the model's "submit final answer" call)."""
+
+    event_kind: Literal['output_tool_call'] = 'output_tool_call'
+    """Event type identifier, used as a discriminator."""
```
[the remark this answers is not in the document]
DouweM: Not in this PR — failure paths still emit `FunctionToolCallEvent`/`FunctionToolResultEvent` alongside the new events (see `_emit_output_tool_events(legacy=True)` plus the inline `_emit_legacy_output_tool_function_events` calls in `_agent_graph.py`). Success paths emitted nothing on main for output tools, so there is nothing to silently miss there. The legacy emission goes away in the stacked #5332 cleanup, which is the actual breaking change.

docs/capabilities.md
DouweM: Renamed to `ToolCallEvent` / `ToolResultEvent` in b70ab61.

docs/capabilities.md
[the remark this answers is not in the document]
DouweM: Dropped `test=skip` on the example block in b70ab61. The example is now exercised by `tests/test_examples.py`.

tests/test_ui.py
DouweM: Replaced the count-based assertions with a full event-stream snapshot in b70ab61.

tests/models/test_bedrock.py
DouweM: Renamed `result` -> `part` on `ToolResultEvent` (parallels `ToolCallEvent.part`) in b70ab61. `FunctionToolResultEvent.result` stays as a `@deprecated` read-side property; constructor BC is internal-only so I just updated the call sites. Skipped the optional `output` field for now per your "leaving it off is OK" — happy to add later if you change your mind.

pydantic_ai_slim/pydantic_ai/_agent_graph.py
DouweM: Dropped the `legacy` flag in 5fce65be. `_emit_output_tool_events` is back to just emitting the new pair; the 6 failure sites now call `_emit_legacy_output_tool_function_events` explicitly afterwards. Slightly more lines per site, but the v2 cleanup is now a one-line delete at each site.

Issue #4140, opened 2026-01-29
Add response-based fallback support for streaming in FallbackModel


sarth6:
## Summary

Currently (or rather: when https://github.com/pydantic/pydantic-ai/pull/3786 is merged), response-based fallback in `FallbackModel` only works with non-streaming requests (`run()` and `run_sync()`). Streaming requests (`run_stream()`) do not support response-based fallback because the response is consumed incrementally and cannot be easily inspected after the fact.

## Context

This limitation was introduced in PR #3786 which added response-based fallback support for `FallbackModel`. The streaming case was deliberately left out as it requires a different approach — potentially buffering the streamed response to inspect it, or providing a streaming-aware callback mechanism.

## Current Behavior

When using `FallbackModel` with `fallback_on` response handlers and `run_stream()`, the response handlers are silently ignored. A note in the docstring and documentation explains this limitation.

## Desired Behavior

Response-based fallback should also work with streaming requests. Possible approaches:
- Buffer the full streamed response, then evaluate handlers before yielding
- Provide a streaming-aware handler that can inspect partial responses
- Allow post-stream evaluation that triggers re-request with next model

## References

- PR #3786: Add response-based fallback support for FallbackModel

agnim25:
Any movement on this @sarth6 @DouweM ? Would be especially interested in seeing something like this supported.

agnim25:
Created this PR https://github.com/pydantic/pydantic-ai/pull/5321, if either of you are free to take a look. Thanks!

Issue #3316, opened 2025-11-03
Pydantic AI v1.9+ Bug: parent_message_id Regression Causes extra_forbidden Errors


marr75:
### Initial Checks

- [x] I confirm that I'm using the latest version of Pydantic AI
- [x] I confirm that I searched for my issue in https://github.com/pydantic/pydantic-ai/issues before opening this issue

### Description

In pydantic-ai v1.9+, the `parent_message_id` field in `TOOL_CALL_START` SSE events incorrectly points to `TOOL_CALL_RESULT` message IDs (the most recent message in the stream) instead of the assistant message that requested the tool call. This causes ag-ui clients to reconstruct an invalid conversation structure where tool result messages have `toolCalls` arrays, leading to schema validation failures with `extra_forbidden` errors.

## Impact

- **Severity:** Critical - breaks round-trip ag-ui communication
- **Affected versions:** pydantic-ai >= 1.9
- **Error message:** `messages.N.tool.toolCalls: Extra inputs are not permitted [type=extra_forbidden]`
- **Root cause:** AG-UI integration rewrite in v1.9

### Expected Behavior

Each `TOOL_CALL_START` event should have `parent_message_id` pointing to the assistant `TEXT_MESSAGE_START` message that requested the tool call.

**Correct structure:**
```
TEXT_MESSAGE_START (id: A, role: assistant)
  └─ TOOL_CALL_START (parent: A) ✓ Correct
     └─ TOOL_CALL_RESULT (id: B, role: tool)
        └─ TEXT_MESSAGE_START (id: C, role: assistant)
           └─ TOOL_CALL_START (parent: C) ✓ Correct
```

### Actual Behavior in v1.9+

`parent_message_id` is set to the **most recent message ID in the stream** (regardless of message type), which is often a `TOOL_CALL_RESULT`.

**Broken structure in v1.9+:**
```
TEXT_MESSAGE_START (id: A, role: assistant)
  └─ TOOL_CALL_START (parent: A) ✓ First call is correct
     └─ TOOL_CALL_RESULT (id: B, role: tool) ← Most recent message!
        └─ TEXT_MESSAGE_START (id: C, role: assistant)
           └─ TOOL_CALL_START (parent: B) ✗ Points to tool result!
```

### How This Breaks

1. **Stream generation:** pydantic-ai emits `TOOL_CALL_START` with `parent: B` (a tool result)

[...]

PR #5321, opened 2026-05-06
feat(fallback): support fallback model handling for streaming cases

agnim25:
## Overview

Extends `FallbackModel.request_stream()` so response-handler fallback works for streaming responses. When a streamed candidate is rejected, the stream emits a generic `ModelResponseResetEvent` before continuing with the next model.

This keeps streaming live, without buffering rejected candidates, while giving higher layers a first-class boundary so they can avoid committing to stale output.

## Architecture

`_FallbackStreamedResponse` remains, but its role changes from "hide fallback as one seamless response" to "coordinate candidate streams and label discarded boundaries".

```text
agent.run_stream()
└─ ModelRequestNode.stream()
   └─ FallbackModel.request_stream()
      ├─ create AsyncExitStack
      ├─ create _FallbackStreamedResponse
      ├─ response.open_next_stream()
      │  ├─ prepare candidate model
      │  ├─ prepare messages
      │  └─ enter candidate stream into AsyncExitStack
      └─ yield _FallbackStreamedResponse
```

During iteration:

```text
_FallbackStreamedResponse.__aiter__()
└─ _get_event_iterator()
   ├─ current_stream = candidate stream
   ├─ async for event in current_stream._get_event_iterator()
   │  └─ yield raw inner event live
   ├─ response = current_stream.get()
   ├─ if response handler rejects:
   │  ├─ yield ModelResponseResetEvent(response=rejected_response)
   │  ├─ open_next_stream()
   │  └─ continue with next candidate
   └─ else:
      ├─ accept current model
      └─ return
```

So a reject to accept stream is explicitly labeled:

```text
model 1:
PartStartEvent(...)
FinalResultEvent(...)
PartEndEvent(...)
ModelResponseResetEvent(response=model_1_response)

model 2:
PartStartEvent(...)
FinalResultEvent(...)
PartDeltaEvent(...)
PartEndEvent(...)
```

## Design Notes

- `ModelResponseResetEvent` is generic, not fallback-specific. It represents "discard the streamed response parts before this point", which can also support future mid-stream retry wrappers.
- `FinalResultEvent` is unchanged. Reset events cause the agent stream loop to clear any previously latched `FinalResultEvent` and re-arm for the accepted model.
- The wrapper iterates inner raw stream events via `_get_event_iterator()` so `PartEndEvent` and `FinalResultEvent` transformer chains are not applied twice.
- Candidate streams are managed through an `AsyncExitStack`, so early stream exits release active inner streams deterministically.
- `stream_text()` resets its accumulator on `ModelResponseResetEvent`, avoiding silent concatenation of rejected and accepted text.

## UI Behavior

UI adapters currently ignore `ModelResponseResetEvent` safely via their existing default event handling. Active UI false-start discard, such as AG-UI starting a fresh message id or Vercel starting a fresh step on reset, is intentionally left as a follow-up.

## Tests

Added coverage for:

- raw event sequences for accepted and rejected streaming fallback candidates
- no duplicate `PartEndEvent` / `FinalResultEvent`
- response-handler rejection followed by accepted fallback stream
- mid-stream fallbackable exception followed by accepted fallback stream
- `stream_text()` accumulator reset after rejected response

Local checks:

- `uv run pytest tests/models/test_fallback.py -q`
- `uv run pytest tests/test_streaming.py -q`
- `uv run pytest tests/test_ui.py::test_run_stream_text_and_thinking -q`
- `PYRIGHT_PYTHON_IGNORE_WARNINGS=1 uv run pyright pydantic_ai_slim/pydantic_ai/messages.py pydantic_ai_slim/pydantic_ai/models/fallback.py pydantic_ai_slim/pydantic_ai/models/__init__.py pydantic_ai_slim/pydantic_ai/agent/abstract.py pydantic_ai_slim/pydantic_ai/result.py`

### v0 Description

The original PR version extended `FallbackModel.request_stream()` by adding `_FallbackStreamedResponse`, a wrapper that attempted to make response-handler fallback transparent for streaming requests.

In that version:

- `FallbackModel.request_stream()` yielded `_FallbackStreamedResponse` only when `fallback_on` included response handlers.
- The wrapper opened each candidate model stream in order, forwarded events live, assembled the response, and ran the response handler after the candidate stream completed.
- If the handler rejected candidate N, the wrapper reset its local parts state, opened candidate N+1, and continued yielding events through the same outer stream.
- Exception-only streaming fallback kept the legacy single-yield path.
- The documented UX caveat was that consumers might see a rejected model's streamed events before the accepted model's events.

Reviewer feedback identified that this "transparent wrapper" architecture hid a real stream boundary:

- It iterated the inner stream through `inner_sr.__aiter__()` and then applied the outer `StreamedResponse.__aiter__()` transformers again, duplicating additive events such as `PartEndEvent` and `FinalResultEvent`.
- It rebased the outer parts manager between candidates, allowing part-end state from a rejected model to cross into the accepted model's event sequence.
- It opened inner streams inside the event iterator, tying cleanup to generator finalization instead of the `request_stream()` context manager.
- Most importantly, it gave agent/result/UI layers no signal that a previously emitted `FinalResultEvent` belonged to a rejected model.

The current iteration keeps the no-buffering streaming behavior, but replaces the hidden boundary with an explicit `ModelResponseResetEvent` that higher layers can consume.


pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -65,6 +67,96 @@ def _is_exception_type(value: Any) -> TypeGuard[type[Exception]]:
...
+            # caller ends up seeing only the accepted model's usage — matching the
+            # non-streaming request() semantics.
+            self._parts_manager = ModelResponsePartsManager()
+
+            try:
+                _, prepared_parameters = model.prepare_request(model_settings, model_request_parameters)
+                async with model.request_stream(
+                    messages, model_settings, model_request_parameters, run_context
+                ) as inner_sr:
+                    async for event in inner_sr:
```
sarth6: `inner_sr.__aiter__()` already wraps events in `iterator_with_part_end` and `iterator_with_final_event`. The wrapper's `__aiter__` re-applies both, so direct event consumers see `PartEndEvent` and `FinalResultEvent` twice at end-of-stream. (`result.all_messages()` is fine because it canonicalizes.)

Let's iterate `inner_sr._get_event_iterator()` directly

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -65,6 +67,96 @@ def _is_exception_type(value: Any) -> TypeGuard[type[Exception]]:
...
+                        # Mirror inner state so self.get() reflects this model's assembled response.
+                        self._parts_manager = inner_sr._parts_manager
```
sarth6: The rebind helps `self.get()`, but the outer `iterator_with_part_end`'s `last_start_event` still points at the rejected model. When the accepted model's first `PartStartEvent` arrives, the outer synthesizes a `PartEndEvent` reading from the freshly-swapped parts_manager:

```
PartEndEvent(content=' \n')                                # rejected's end
PartEndEvent(content='hello world', next_part_kind='text') # accepted's content, wrong position
PartStartEvent(content='hello world', previous_part_kind='text')
```

Same on mid-iteration exceptions.

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -65,6 +67,96 @@ def _is_exception_type(value: Any) -> TypeGuard[type[Exception]]:
...
+            # for the next attempt. _usage is overwritten on each event below, so the
+            # caller ends up seeing only the accepted model's usage — matching the
+            # non-streaming request() semantics.
+            self._parts_manager = ModelResponsePartsManager()
+
+            try:
+                _, prepared_parameters = model.prepare_request(model_settings, model_request_parameters)
+                async with model.request_stream(
+                    messages, model_settings, model_request_parameters, run_context
+                ) as inner_sr:
```
sarth6: The old path enters the inner stream into an `AsyncExitStack` in the cm body, so cleanup runs on `async with` exit. Here it's entered inside `_get_event_iterator`, so cleanup is tied to the generator's lifetime. Early `break` from the consumer's `async for` doesn't call `aclose()` on the cached iterator, and the wrapper doesn't have an `__aexit__` that closes it either — so the inner's `__aexit__` only runs on GC, and HTTP connections sit in the pool meanwhile.

Could we open the first model eagerly via an `AsyncExitStack` in the cm body, then have the wrapper register subsequent attempts onto the same stack?

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -65,6 +67,96 @@ def _is_exception_type(value: Any) -> TypeGuard[type[Exception]]:
...
+    underlying stream, forwards every event live to the consumer (no buffering),
+    mirrors deltas to self._parts_manager so .get() reflects the accepted model's
+    response, and at end-of-stream calls self._should_fallback_handler(self.get()).
+    On rejection, resets parts state and continues to the next model. The consumer
+    iterates a single async iterator throughout — they will see events from
+    rejected models followed by events from the accepted model.
+    """
+
+    _fallback_models: list[Model]
+    _request_args: tuple[Any, ...]
```
sarth6: Could we use a small frozen dataclass here to avoid the `Any`?

```python
@dataclass(frozen=True)
class _StreamRequestArgs:
    messages: list[ModelMessage]
    model_settings: ModelSettings | None
    model_request_parameters: ModelRequestParameters
    run_context: RunContext[Any] | None
```

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -65,6 +67,96 @@ def _is_exception_type(value: Any) -> TypeGuard[type[Exception]]:
...
+    def model_name(self) -> str:
+        if self._accepted_model is None:
+            return 'fallback:pending'
+        return self._accepted_model.model_name
+
+    @property
+    def timestamp(self) -> datetime:
+        if self._accepted_timestamp is not None:
+            return self._accepted_timestamp
+        return _now_utc()
```
sarth6: nit: other `StreamedResponse` impls cache `_now_utc()` once at construction. This one returns a fresh value every read until acceptance, so consecutive reads drift a bit

Cache in `__post_init__`?

tests/models/test_fallback.py
```diff
@@ -1731,3 +1749,79 @@ async def enter_and_hold(event: asyncio.Event) -> None:
     await task2
     assert provider1._own_http_client.is_closed  # pyright: ignore[reportPrivateUsage]
     assert provider2._own_http_client.is_closed  # pyright: ignore[reportPrivateUsage]
+
+
+async def test_response_handler_triggers_streaming() -> None:
```
sarth6: Worth adding a snapshot test for the event sequence; that's where the issues above would show up. Something like:

```python
async with fallback.request_stream(...) as sr:
    events = [ev async for ev in sr]
assert events == snapshot([...])
```

One happy-path case and one reject/accept case.

agnim25:
@DouweM did you have any feedback here? didn't see any comments on your end so far.

DouweM:
@sarth6 Can you have a look here please as the author of https://github.com/pydantic/pydantic-ai/issues/4140?

> @DouweM did you have any feedback here? didn't see any comments on your end so far.

@agnim25 Right, have you seen our [contributing guide](https://pydantic.dev/docs/ai/project/contributing)? Specifically https://pydantic.dev/docs/ai/project/contributing/#before-you-write-code and  
https://pydantic.dev/docs/ai/project/contributing/#we-review-prs-in-our-priority-order-not-submission-order.

sarth6:
Cross posting - hoping to post a design review soon

sarth6:
Thanks for picking this up @agnim25, and sorry for the slow review 🙏. Bit of context on why I care about this one: I opened #3640, did the non-streaming side in #3786, and when I scoped that PR I deliberately punted on streaming and said I'd follow up using `ModelResponsePartsManager.get_parts()` plus an optional `incomplete` bool (that thread). So this is right in the area I've been living in, and #4140 is exactly the gap I wanted to close. You also got the important instinct right - forward live, don't buffer - which is the call I'd make too.

There are three concrete bugs (repros inline), but the thing I actually want to talk about is a design question sitting underneath them that I don't think we can settle inside this PR. I worked with my agent to investigate various solutions and weigh them on a couple measures I lay out below.

## The three bugs

All in `_FallbackStreamedResponse`, details inline:

- **A - duplicated end-of-stream events.** The wrapper iterates `inner_sr` (which runs the inner's `__aiter__` transformer chain) and then inherits its own chain, so the additive transformers run twice. Direct event consumers see `PartEndEvent` and `FinalResultEvent` emitted twice at end of stream. `result.all_messages()` is fine because it canonicalizes through `get_parts()`, which is why this slips past the happy-path tests.
- **B - wrong-content `PartEndEvent` at the reject→accept boundary.** The outer `iterator_with_part_end` still has a `last_start_event` from the rejected model, so when the accepted model's first `PartStartEvent` shows up it synthesizes a `PartEndEvent` reading the freshly-swapped `parts_manager` - i.e. the accepted model's content sitting at the rejected model's position.
- **C - non-deterministic cleanup.** The inner stream is entered inside `_get_event_iterator`, so its `__aexit__` is tied to generator finalization. A consumer that `break`s early never triggers `aclose()`, and the wrapper has no `__aexit__` of its own, so the inner HTTP connection isn't released until GC runs.

There are also a few smaller things inline (the `tuple[Any, ...]`, the `timestamp` drift before acceptance, and no event-sequence snapshot test). That last one matters more than it sounds - a snapshot is exactly what would've caught A and B, since they're invisible at the `all_messages()` layer.

## They're not really three separate bugs

They all come from the same place: the wrapper tries to look like one `StreamedResponse` while it's actually bridging several models, but it has no real identity until acceptance, so it fakes one. It rebinds `self._parts_manager` mid-flight, overwrites `_usage` per event, and hands back sentinel `model_name`/`timestamp` values until a model is accepted. Each of those seams is where a bug hides - A is the doubled chain, B is state (`last_start_event`) spanning a seam it shouldn't, C is a lifecycle the wrapper doesn't actually own.

So I don't think the move is to patch A/B/C in place. Fixing them one by one just keeps bolting compensating state onto a structure whose whole problem is the compensating state. The cleaner version stops compensating: iterate the inner's raw events so only one chain runs, give each candidate its own parts manager so nothing spans the boundary, and let the `request_stream` cm body own cleanup with an `AsyncExitStack` (which is what the legacy path already does). That removes all three structurally instead of three patches.

## Design

`FinalResultEvent` is a commit point, and live-forwarding runs straight into it.

`run_stream`'s `stream_to_final` (in `agent/abstract.py`) breaks at the first `FinalResultEvent`:

```python
async for event in stream:
    yield event
    if isinstance(event, _messages.FinalResultEvent):
        final_result_event = event
        break
```

And that event gets injected the moment a model produces a part matching the output schema - for text output, the *first* `TextPart`, i.e. mid-stream. So when model 1 streams some text and is then rejected at end-of-stream (the common case: bad `finish_reason`, a `web_fetch` reporting `FAILED`, whitespace-only output), the order is:

1. model 1 emits its first `TextPart` → `FinalResultEvent` fires → the agent **commits** `final_result` to model 1
2. only later, at end-of-stream, does the handler reject model 1

The run only "works" today because `get_output()` re-reads the final parts at the end. But the committed `final_result` (its `tool_name`/`tool_call_id`) still points at the rejected model, and anyone reading text live via `stream_text` gets both models concatenated (`"<rejected><accepted>"`). The new tests assert via `stream_responses`, which reads the per-candidate `get()` (reset each attempt), so they never hit this.

I don't think there's a clean fix that lives entirely inside `FallbackModel` here. The model boundary has to be something the agent loop itself can see, which is a public-API question - and that's the part I'd really want to align on before more code goes in.

## What are we optimizing for

Worth naming these precisely, because the right answer genuinely changes by consumer:

- **TTFB (time to first token)** - how long the consumer waits before the first piece of the *accepted* answer reaches them. For a chat UI this is the spinner. Minimizing it is the entire reason to use `run_stream` over `run`.
- **leak (a.k.a. false start)** - events from a *rejected* model reaching the consumer before the accepted model's. Two things matter about a leak: is it **bounded** (one model's stream) or unbounded, and is it **labeled** (consumer can detect the boundary and discard) or **silent** (consumer blindly concatenates)?

And the consumers don't want the same thing:

- browser UIs over SSE/AG-UI and `run_stream_events`: can't tolerate buffering at all, but can discard a leak cleanly if we tell them where it is
- `stream_text`/`stream_output` (programmatic): tolerate a little latency (already debounced ~100ms), but must not silently concatenate
- OTel/observability: don't care about the leak as long as it's labeled
- compliance / content-filter ("nothing escapes until validated"): need zero leak and can pay the latency

There's no single right point here, which is the whole problem.

## Options I considered

| approach | TTFB hit | leak | API surface |
|---|---|---|---|
| this PR (live-forward) | 0 | full + silent | none |
| buffer-then-flush | full stream | none | 1 field (`incomplete`) |
| bounded-lookahead window | window-sized | only late rejections | 1 field + config knob |
| per-handler mode (`@requires_complete`) | 0 by default | for live handlers | 1 field + decorator |
| live-forward + boundary event | 0 | labeled → discardable | 1 field + 1 event |

Quick why-nots: pure **buffer-then-flush** is the cleanest to implement but withholds the whole stream until validation, so a 60-120s reasoning model shows a spinner for the entire response and then dumps it (that's `run()` with extra steps, and it trips 30-60s proxy idle timeouts on SSE). The **lookahead window** needs `max_events`/`max_duration` knobs whose right defaults we can't know without telemetry, which is the kind of opinionated config we tend to avoid. **Per-handler mode** puts two execution paths in one wrapper, re-inflating the bug surface we just argued for collapsing.

## Where I'd land

Live-forward plus an explicit boundary event, built on one small primitive:

1. **`ModelResponse.incomplete: bool`** (defaults `False`) - this is the exact follow-up I flagged in #3786, and DouweM's suggestion on that thread too. It makes `StreamedResponse.get()` self-describing at any point, lets a handler judge partial vs complete with the *same* `Callable[[ModelResponse], bool]` signature, and is useful well beyond fallback. One field, backwards compatible.

2. **A `FallbackRetryEvent` in the model event stream** carrying the rejected model's identity + reason. One event, three jobs: it lets `stream_to_final` re-arm (drop a `final_result` it committed to a now-rejected model - the fix for the section above), it lets `result.py` reset its text accumulator (so `stream_text` stops concatenating), and it lets UI adapters discard the false start. That last one already has a mechanism - AG-UI calls `new_message_id()` in `before_response` for exactly this reason (#3316) - and adapters that don't handle the event just degrade to today's behavior.

3. **The rebuilt wrapper** from above, which makes A/B/C go away.

The handler still gets called once at end-of-stream, exactly like `request()`, so nobody's existing `reject_*` function suddenly misfires on a partial. The leak in this default is one model's stream, but the boundary event makes it clean (labeled, discardable) instead of silent. Eager mid-stream rejection (refusal detection - the "I cannot" stuff from the #3640 examples) and a real zero-leak buffered mode for the compliance case are both natural opt-ins on the same `incomplete` primitive, but I'd hold off on both until someone actually needs them rather than ship knobs up front.

This also covers the rejected-usage telemetry point from my summary - `FallbackRetryEvent` is the obvious place to surface it instead of dropping it silently.

Whole API delta for the long-term version: one field, one event, one (default-noop) UI hook. Everything else is internal.

## Path I'd suggest

Please don't read this as throw it away - the core decision (forward live, don't buffer) is the one I'd make too, and a lot of the plumbing here carries over. But `incomplete` and the event are public surface, so per the contributing guide I'd want to get @DouweM's read first. Rough plan:

1. land **`ModelResponse.incomplete`** on its own - small, uncontroversial (it's the #3786 follow-up), useful by itself, unblocks everything
2. reshape this PR into the rebuilt wrapper + `FallbackRetryEvent` + `stream_to_final` re-arm + accumulator reset, with event-sequence snapshot tests (happy path + reject/accept boundary, where A/B/C would've shown up)
3. UI-adapter handling (AG-UI/Vercel) as a follow-up

@dsfaccini  - would love a gut check on the `FallbackRetryEvent` direction before @agnim25 sinks more time in. Does making the model boundary a first-class event feel right to you, or would you rather keep this inside `FallbackModel` somehow?

dsfaccini:
David's AICA here: yes — making the model boundary a first-class event feels right to me, and I don't think it can live inside `FallbackModel`. Concrete reason: the `FinalResultEvent` latch is injected in `StreamedResponse.__aiter__` (`iterator_with_final_event`), a layer above the wrapper, and `stream_to_final` breaks on it — so by the time the end-of-stream handler rejects model-1, the run is already committed and the consumer has already accumulated model-1's text off the live stream (reproduced the `<rejected><accepted>` concatenation + the mid-stream commit on this branch). Nothing under `FallbackModel` can retract that. So: event, not internal.

Two refinements before @agnim25 builds it:

1. **Reuse `ModelResponseState` rather than a new `ModelResponse.incomplete: bool`.** `state: Literal['complete', 'interrupted']` already landed in #5313 (this branch's merge-base) and `StreamedResponse.get()` already sets it. A separate `incomplete` bool gives two overlapping "did this finish cleanly?" axes on a central serialized type. Fold partial-vs-complete into `state` (extend the enum) — still additive/back-compat per the version policy.

2. **Name the event for what it does to the stream, not for fallback.** It's really "a model boundary occurred mid-stream — discard the prior parts": a general category for any mid-stream-retrying wrapper, and it's what re-arms `stream_to_final`, resets the `result.py` accumulator, and lets UI adapters drop the false start (AG-UI's `new_message_id()` already does this). A fallback-specific name bakes an implementation detail into the universal event taxonomy. Non-handling consumers degrade fine (`UIEventStream.handle_event` is `case _: pass`, no event-level `assert_never`).

Sequencing: agree with the reshape (rebuilt wrapper — one transformer chain, per-candidate parts manager, `AsyncExitStack` cleanup in the cm body — plus the event + `stream_to_final` re-arm + accumulator reset + snapshot tests). One tweak: don't land the partial signal solo as `incomplete`; if anything lands first it's the `state` extension, so we don't pre-commit a field shape before the design is settled.

The `ModelResponseState` extension and the new stream event are public surface, so they'll need a maintainer sign-off at final review — but the path above is clear enough for @agnim25 to start the reshape against it.

sarth6:
Thanks david!

dsfaccini:
Thank *you* guys for working on this

pydantic_ai_slim/pydantic_ai/models/fallback.py
agnim25: done, branch now has at pydantic_ai_slim/pydantic_ai/models/fallback.py (line 67)
```
@dataclass(frozen=True)
class _StreamRequestArgs:
    messages: list[ModelMessage]
    model_settings: ModelSettings | None
    model_request_parameters: ModelRequestParameters
    run_context: RunContext[Any] | None
```
And `_FallbackStreamedResponse` now uses:
```
_request_args: _StreamRequestArgs
```

pydantic_ai_slim/pydantic_ai/models/fallback.py
agnim25: Ah good point, added cached value at construction:

_created_timestamp: datetime = field(default_factory=_now_utc, init=False)

tests/models/test_fallback.py
agnim25: Yup, added this test sequence:

tests/models/test_fallback.py:
test_response_handler_streaming_happy_path_events
test_response_handler_streaming_reject_accept_events
test_response_handler_streaming_mid_stream_exception_events

which assert sequences like the one i mentioned in the PR description.

pydantic_ai_slim/pydantic_ai/models/fallback.py
agnim25: Yes, added this in this iteration:
```
async with AsyncExitStack() as stack:
    response = _FallbackStreamedResponse(
        ...
        _exit_stack=stack,
        ...
    )
    await response.open_next_stream()
    yield response
    return
 ```
 
Then `_FallbackStreamedResponse.open_next_stream()` registers each candidate stream on that same stack:

```
stream = await self._exit_stack.enter_async_context(
    model.request_stream(...)
)
```

File: pydantic_ai_slim/pydantic_ai/models/fallback.py (hunk 1 above)
Comment (agnim25) [reply pydantic_ai_slim/pydantic_ai/models/fallback.py]: Addressed by changing `_FallbackStreamedResponse` to iterate each candidate stream’s raw `_get_event_iterator()` instead of `inner_sr.__aiter__()`. 

The outer `StreamedResponse.__aiter__()` is now the only layer applying `PartEndEvent` / `FinalResultEvent` transformers, and the new event-sequence snapshots cover this.

File: pydantic_ai_slim/pydantic_ai/models/fallback.py (hunk 2 above)
Comment (agnim25) [reply pydantic_ai_slim/pydantic_ai/models/fallback.py]: Ah yes, adding the new `ModelResponseResetEvent` event has helped here.

`iterator_with_part_end()` now handles `ModelResponseResetEvent` specially:

```
elif isinstance(event, ModelResponseResetEvent):
    end_event = part_end_event()
    if end_event:
        yield end_event
    last_start_event = None
```

That means when a rejected candidate ends, the outer iterator first closes the rejected model’s active part using the rejected model’s current parts manager, then clears `last_start_event`. When the accepted model’s first `PartStartEvent` arrives, there is no stale rejected-model start event left to close.

So the bad sequence:
```
PartEndEvent(rejected)
PartEndEvent(accepted content at rejected boundary)  # wrong
PartStartEvent(accepted)
```
becomes:

```
PartEndEvent(rejected)
ModelResponseResetEvent(rejected_response)
PartStartEvent(accepted)
```

This is covered by the raw event snapshot tests for both reject-to-accept and mid-stream exception-to-accept.

File: pydantic_ai_slim/pydantic_ai/agent/abstract.py [hunk 7]
```diff
@@ -799,6 +799,9 @@ async def stream_to_final(
                             nonlocal final_result_event
                             async for event in stream:
                                 yield event
+                                if isinstance(event, _messages.ModelResponseResetEvent):
```
sarth6: I think this block effectively does nothing in practice so for simplicity let's drop it.

The loop breaks as soon as `final_result_event` is set (the `FinalResultEvent` case right below it). So once a model has produced a final result we've already left the loop. The only way we ever reach this reset branch is when no final result was set yet, which means `final_result_event` is already `None` and we're just setting it to `None` again. There's nothing to clear. Clearing a result we committed to a rejected model is a real need, but that happens in `on_complete`, not here.

pydantic_ai_slim/pydantic_ai/agent/abstract.py
```diff
@@ -829,6 +832,13 @@ async def on_complete() -> None:
                                 by `StreamedRunResult._marked_completed`.
                                 """
                                 nonlocal final_result
+                                if stream._raw_stream_response.final_result_event is not None:  # pyright: ignore[reportPrivateUsage]
```
sarth6: This relinks the committed `final_result` to the accepted model, but it's only exercised by text-output tests, where `tool_name`/`tool_call_id` are `None` and this is a no-op.

The case it exists for is untested: model 1 emits an output tool call `id=a` (committed, then rejected), model 2 accepted with `id=b`; without this block the output `ToolReturnPart` points at `a` and dangles. Worth a structured reject/accept test through `run_stream` asserting the output return is tied to the accepted call id.

nit: the block's also tangled, the first `FinalResult(...)`'s `.output` is dropped on the next line and `final_result_event` shadows the `nonlocal`, so it can collapse to one assignment.

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -65,6 +66,138 @@ def _is_exception_type(value: Any) -> TypeGuard[type[Exception]]:
...
+        while current_stream := self._current_stream:
+            try:
+                async for event in current_stream._get_event_iterator():
+                    self._sync_current_stream()
+                    yield event
+            except Exception as exc:
+                self._sync_current_stream()
+                if await self._should_fallback_handler(exc):
+                    self._exceptions.append(exc)
+                    yield ModelResponseResetEvent(response=current_stream.get())
```
sarth6: nit: worth a comment on the order. We yield the reset before `open_next_stream()` swaps the parts manager, which is what lets `iterator_with_part_end` close the rejected part against the rejected stream.

Reorder these two lines and the accepted model's content leaks into a part-end at the old index.

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -65,6 +66,138 @@ def _is_exception_type(value: Any) -> TypeGuard[type[Exception]]:
...
+
+    def get_stream_cancel_errors(self) -> tuple[type[BaseException], ...]:
+        if self._current_stream is None:
+            return super().get_stream_cancel_errors()
+        return self._current_stream.get_stream_cancel_errors()
+
+    def _sync_current_stream(self) -> None:
+        if self._current_stream is None:
+            return
+        object.__setattr__(self, '_parts_manager', self._current_stream._parts_manager)
```
sarth6: nit: no need for `object.__setattr__`, this isn't frozen and `_parts_manager` is a `cached_property`, so `self._parts_manager = ...` works and reads cleaner.

Same for re-copying provider id/url/details on every event when they don't change within a candidate.

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -65,6 +66,138 @@ def _is_exception_type(value: Any) -> TypeGuard[type[Exception]]:
...
+        self.provider_details = self._current_stream.provider_details
+        self.finish_reason = self._current_stream.finish_reason
+
+    @property
+    def model_name(self) -> str:
+        if self._accepted_model is not None:
+            return self._accepted_model.model_name
+        if self._current_model is not None:
+            return self._current_model.model_name
+        return 'fallback:pending'
```
sarth6: nit: looks unreachable. `open_next_stream()` runs before we yield and sets `_current_model` or raises, so `assert self._current_model is not None` would read better than a sentinel string.

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -2794,10 +2794,24 @@ class FinalResultEvent:
     __repr__ = _utils.dataclasses_no_defaults_repr
 
 
+@dataclass(repr=False, kw_only=True)
+class ModelResponseResetEvent:
```
sarth6, line 2678: The shape is right and I'd keep most of it. The name is generic rather than fallback-specific, and it reads as response-level next to the part-level siblings, which is the right altitude. Carrying the whole `ModelResponse` is also the right call: a discarded response is by definition absent from `all_messages()`, so this event is the only place a stream consumer can ever observe it. And reusing `ModelResponseState` is nice, `response.state` already tells you which boundary this was, `complete` when the content was rejected, `incomplete` when the model raised mid-stream.

The one thing I'd actually settle before this goes public is the field name. `response` is generic, and unlike `PartStartEvent.part` where a misread is harmless, treating this one as real output (accumulating it as the answer, persisting it) is the exact corruption this PR exists to prevent. `rejected_response` or `discarded_response` makes that hard to get wrong. Renaming a public field is breaking, so this is the cheap-now, expensive-later call.

On the "why" I flagged before, I'd soften that to a non-blocker. Most of it is already recoverable from the carried response (`finish_reason`, the content, and `state` for the path); the only thing truly lost is the triggering exception on the exception-fallback path, and a field for that is purely additive, so no rush. If we ever want it, the honest shape is `cause: Exception | None`, not a stringly reason.

nit: the docstring says "parts should be discarded" but the field hands you a whole response, worth aligning the two.

@dsfaccini this one is public surface so I'd love your read, mainly on the field name.

docs/models/overview.md
```diff
@@ -439,6 +441,11 @@ Pydantic AI is a Python agent framework for building production-grade LLM applic
 
 Response handlers receive the [`ModelResponse`][pydantic_ai.messages.ModelResponse] returned by the model and should return `True` to trigger fallback to the next model, or `False` to accept the response.
 
+For streaming responses, handlers are evaluated after each model's stream completes. Programmatic text streaming
```
sarth6: This is split across two spots, here and a paragraph down by Combining Handlers, and both sit under Response-Based Fallback. I'd reorganize rather than just tweak the wording, and I think this PR is the right place since it's already editing both of these lines.

Why move it: the `ModelResponseResetEvent` and the `stream_text` accumulator reset aren't response-handler specific, they fire on exception fallback mid-stream too. Filing this under Response-Based Fallback puts it on the wrong axis, it reads as a response-handler footnote when it's really how fallback behaves while streaming in general, across both modes.

My suggestion: a short `### Streaming` subsection at the fallback level (sibling to Response-Based Fallback, after Combining Handlers) that owns the whole story, and drop this admonition plus the paragraph below. Something like:

```markdown
### Streaming

Both exception-based and response-based fallback work with streaming requests
([`agent.run_stream()`][pydantic_ai.agent.AbstractAgent.run_stream]). Because streaming events are
forwarded as they arrive, a consumer may see events from a model that is later rejected before
fallback moves on to the next model. When that happens, the stream emits a
[`ModelResponseResetEvent`][pydantic_ai.messages.ModelResponseResetEvent] to mark the previous
response as discarded, and events from the next model follow.

Programmatic text streaming handles this for you: `stream_text` resets its accumulated text at the
reset, so non-delta chunks reflect the accepted model's response rather than the rejected one
concatenated with it. If you consume the raw event stream via
[`run_stream_events()`][pydantic_ai.agent.AbstractAgent.run_stream_events], handle the
[`ModelResponseResetEvent`][pydantic_ai.messages.ModelResponseResetEvent] yourself to discard or
visually separate the rejected model's output.
```

This also clears up the vague "Direct event consumers": `run_stream_events()` is the API that actually delivers the reset, so naming and linking it (like the rest of these docs do) is clearer.

The only genuinely new fact here is the one clause that exception fallback emits the reset too. Everything else is content already in the PR, just consolidated and lifted up a level. I'd still skip a separate `run_stream` + `event_stream_handler` caveat: the handler there stops at the first `FinalResultEvent`, so it won't see the reset when the rejected model produced output, but that's subtle enough to live as a code comment near `stream_to_final` rather than in user docs.

Would be good to get @dsfaccini opinion on this one though

tests/models/test_fallback.py
```diff
@@ -1324,6 +1459,252 @@ def web_fetch_failed(response: ModelResponse) -> bool:
...
+        parts=[TextPart(content='hello world')],
+        usage=RequestUsage(input_tokens=50, output_tokens=2),
+        model_name='function::success_response_stream',
+        timestamp=IsDatetime(),
+        run_id=IsStr(),
+        conversation_id=IsStr(),
+    )
+
+
+async def test_response_handler_stream_event_handler_rearms_final_result_after_reset() -> None:
```
sarth6: nit: the name oversells it. Model 1 is thinking-only so it never emits a `FinalResultEvent`, nothing to "re-arm". It's really checking that the handler sees the reset when a candidate yields no final result.

Rename, and let the structured test carry the real re-arm assertion.

pydantic_ai_slim/pydantic_ai/result.py
```diff
@@ -327,62 +329,72 @@ async def _stream_response_text(
...
                 # yielding at each step
                 deltas: list[str] = []
                 async for text in deltas_iter:
+                    if text is None:
+                        deltas.clear()
+                        continue
                     deltas.append(text)
                     yield ''.join(deltas)
 
+    async def _stream_response_text_delta_items(self) -> AsyncIterator[tuple[str | None, int | None]]:
```
sarth6, line 324: It looks like we lost some nice comments in refactoring, so comment or docstring would be nice here

pydantic_ai_slim/pydantic_ai/result.py
```diff
@@ -327,62 +329,72 @@ async def _stream_response_text(
...
+                and last_text_index is not None
+            ):
+                # Text parts that are interrupted by a built-in tool call should not be joined together directly.
+                yield '\n\n', event.index
+                last_text_index = None
+            elif isinstance(event, _messages.ModelResponseResetEvent):
+                last_text_index = None
+                yield None, None
+
+    async def _stream_response_text_deltas(self, debounce_by: float | None) -> AsyncGenerator[str | None, None]:
```
sarth6, line 367: Docstring here too

sarth6 (commented):
Looks pretty close to me, the biggest thing is filling in the test gaps as well as getting sign off from @dsfaccini on `ModelResponseResetEvent` API surface.

Also, not code: I'd retitle and expand the description. "response-handler rejection" makes it sound narrow, but this is really what makes streaming a first-class path for `FallbackModel`: response-based fallback now works mid-stream with proper reset semantics, which is the streaming half that got deferred when the non-streaming support landed in #3786. Since the title feeds the changelog, worth saying that plainly so it reads as completing streaming support for `FallbackModel` rather than a small tweak.

pydantic_ai_slim/pydantic_ai/messages.py
agnim25: good pt, I've updated the field from `response` to `discarded_response` for now, but yes good to get any input here given it's public surface.

sarth6:
@agnim25 looks like test coverage is failing

agnim25:
> @agnim25 looks like test coverage is failing

ah yes had some coverage issues from my latest iteration, resolved now

pydantic_ai_slim/pydantic_ai/models/fallback.py
```diff
@@ -253,6 +393,25 @@ async def request_stream(
         run_context: RunContext[Any] | None = None,
     ) -> AsyncGenerator[StreamedResponse]:
         """Try each model in sequence until one succeeds."""
+        if self._response_handlers:
```
sarth6: Nice looks like everything is addressed, let's get exception-based mid-stream fallback working too since I think it'll be very straightforward and this is a "support fallback model for streaming cases" PR so we should have it work with all handlers.

I think the cleanest fix is to drop this gate and always use `_FallbackStreamedResponse`. Its iterator already catches mid-stream exceptions, and with no response handler the response-reject branch is just a no-op, so this collapses to one path and the loop below goes away.

Right now exception-only streaming reports its span under the concrete model, while non-streaming and response-handler streaming both report `fallback:...`, so this just makes streaming consistent. Let's add a regression test for the actual gap too: exception-only, model 1 raises mid-stream, assert we fall back to model 2.

@dsfaccini fyi this makes exception-only streams recover mid-stream and start emitting `ModelResponseResetEvent`s. I think it's fine as it's the same change that we agreed on anyway for response handlers, but flagging since it's still a behavior change

agnim25: Yes! Clean application of our change. 

Added the testing, also locking in the ModelResponseResetEvent at the boundary and the exact event sequence, so a future regression that re-routes exception-only streaming back through a path that drops the reset event would also fail.

dsfaccini:
David's AICA here: a design question on the `ModelResponseResetEvent` API before this lands — for @sarth6 (since it comes from the #4953 plan) and @dsfaccini (sign-off).

The concern is specifically about adding `ModelResponseResetEvent` to the **public `ModelResponseStreamEvent` union**. That union flows into `AgentStreamEvent`, every user `event_stream_handler`, and both UI adapters (Vercel + AG-UI), several of which close their match with `assert_never`. So adding a member isn't purely additive — it breaks every exhaustive matcher (typecheck + runtime) and forces a new branch in each UI adapter, which is a broad and permanent public-API cost for what's a `FallbackModel`-only concern. (The description's "consumers must handle this event to discard the partial response" is the same signal — existing handlers have to change.)

There may be a cleaner path that needs no new public event: handle the reset entirely inside `_FallbackStreamedResponse` and re-drive the winning candidate through ordinary `PartStartEvent`s. `PartStartEvent` already documents replace-semantics ("If multiple `PartStartEvent`s are received with the same index, the new one should fully replace the old one"), so a consumer that honors the contract discards the abandoned candidate's parts for free; `_FallbackStreamedResponse.get()` would expose only the winning candidate's parts so the persisted history stays clean — leaving the public union and every existing handler/adapter untouched.

The one case where an explicit reset signal *would* be justified is if a consumer genuinely needs to **observe** the discard (e.g. a UI that already rendered the abandoned candidate's tokens with a renderer that doesn't honor same-index-replace, and must visibly clear them). If that's the motivation behind the event in the plan, could you say so? — and even then it may be better as a `FallbackModel`-level hook or an event *outside* the response-content union rather than widening `ModelResponseStreamEvent`.

Not relitigating the feature (clearly wanted) — just getting the public-API shape right before it's a compat commitment. @dsfaccini to make the call.

PR #3339, opened 2025-11-04
Preserve MCP `_meta` and `audience` annotations in `MCPToolset` results

anirbanbasu:
Closes #3323 (per-part `_meta`) and #4353 (audience filtering).

Surfaces MCP tool result metadata and `audience` annotations through the new [`MCPToolset`][pydantic_ai.mcp.MCPToolset] (introduced in #5325), without leaking either to the LLM.

## Foundations (`messages.py` + `_mcp_util.py`)

- `metadata: dict[str, Any] | None` field on `BinaryContent`, `FileUrl` (+ subclasses), `UploadedFile` — the typed home for application-only payloads. Distinct from `vendor_metadata` (model-provider hints sent to the API); docstrings call this out explicitly.
- `ToolReturnPart.model_response_str` / `model_response_object` / `content_items` unwrap `TextContent` to its `.content` so per-part `metadata` doesn't leak to the model.
- `pydantic_ai/_mcp_util.py` — shared helpers: `is_for_assistant`, `partition_content`, `merge_meta`, `clean_meta` (strips the FastMCP `wrap_result` marker), `map_tool_result_part` (preserves per-part `_meta` on the mapped value), `map_tool_result_user_content` (for the audience=user side).

## `MCPToolset` integration

`MCPToolset.direct_call_tool` now:

- Maps top-level `result._meta` → `ToolReturn.metadata` (FastMCP's internal `fastmcp` namespace filtered out).
- Maps per-block `_meta` → the corresponding `TextContent.metadata` / `BinaryContent.metadata`. Embedded text resources shallow-merge outer + inner `_meta` (inner wins on shared keys).
- Partitions content blocks by `annotations.audience`: `'assistant'`/unset parts become the `ToolReturn.return_value` visible to the model; `'user'`-only parts flow back as `ToolReturn.content` (typed `Sequence[UserContent]`) on a separate `UserPromptPart`.

When a result has no `_meta` and no user-audience parts, `direct_call_tool` returns the raw mapped value — same shape as before. `ToolReturn` is only synthesized when one of the new signals is present, so existing callers see no behavior change.

## Deprecated `MCPServer*` / `FastMCPToolset` — intentionally not backported

[...]

PR #5325, opened 2026-05-07
Add `MCPToolset` that uses `fastmcp-slim[client]`, deprecate `MCPServer*` and `FastMCPToolset`

DouweM:
## Summary

Adds `pydantic_ai.mcp.MCPToolset` as the new recommended way to use MCP servers in Pydantic AI. Built on FastMCP's `Client`, with full kwarg/feature parity to the legacy `MCPServer*` hierarchy plus new upside (`auth='oauth'`, `progress_handler`, `roots`, `verify`, `http_client` factory adapter, OAuth flow). The legacy `MCPServer*` and `FastMCPToolset` paths are `@deprecated` in 1.x; they keep working until v2-cut.

This is **Phase A** of the v2 MCP overhaul. Phase B (drop the legacy classes entirely) lands against the v2 branch.

## What's in scope

**New:**
- `pydantic_ai.mcp.MCPToolset` — accepts a URL, script path, `FastMCP` server, `MCPConfig` dict, `ClientTransport`, or pre-built `fastmcp.Client`. Surfaces all FastMCP `Client` config (sampling/elicitation/log/progress/message handlers, `auth`, `verify`, `roots`, `client_info`, `init_timeout`, `read_timeout`) plus PAI-layer config (`max_retries`, `tool_error_behavior`, `cache_tools`, `cache_resources`, `process_tool_call`, `include_instructions`, `include_return_schema`, `sampling_model` shortcut).
- Resource methods (`list_resources`, `list_resource_templates`, `read_resource`) returning PAI-shaped `Resource` / `ResourceTemplate` types — kept under `pydantic_ai.mcp.*` (not top-level) to avoid premature commitment to a generic `Resource` shape.
- Caching layer with `message_handler` injection that invalidates on `tools_list_changed` / `resources_list_changed` and forwards to a user-supplied handler.
- `sampling_model=` shortcut that builds a FastMCP `SamplingHandler` from a PAI `Model`. Mutually exclusive with the low-level `sampling_handler=`.
- Pre-built `Client` conflict detection — clear `ValueError` listing every kwarg that conflicts.
- `http_client=` kwarg implemented via FastMCP's `httpx_client_factory` so existing users of the legacy `http_client` field have a clean migration.
- `load_mcp_toolsets(config_path_or_dict) -> list[AbstractToolset]` free function — same `mcpServers` JSON shape as `load_mcp_servers`, returns toolsets wrapped with `.prefixed(server_name)`.

**Deprecated (removed in v2):**
- `MCPServerStdio`, `MCPServerSSE`, `MCPServerStreamableHTTP`, `MCPServerHTTP` (already deprecated alias)
- `FastMCPToolset`
- `load_mcp_servers`

The abstract `MCPServer` base is intentionally not decorated to avoid noisy subclassing warnings; the user-facing concrete subclasses carry the deprecation.

**Updated:**
- `MCP` capability's `_default_local()` now builds an `MCPToolset` (FastMCP infers SSE vs Streamable HTTP from the URL, replacing our explicit branching). The `local=` type union still accepts the legacy types for back-compat.
- `pydantic-ai-slim[mcp]` extra additionally pulls `fastmcp[client]` so users on the recommended extra get the new path out of the box. The `fastmcp` extra remains as a back-compat alias.
- `pyproject.toml` `filterwarnings` ignores the new deprecation messages so the existing test coverage on the deprecated paths stays green.
- Pyright `executionEnvironments` for `tests/` and per-file `# pyright: reportDeprecated=false` on the durable-exec wrappers (DBOS / Temporal) that wrap legacy MCP classes by design.

## Behavior changes worth calling out for migrators

| Class | Field | Old default | New default on `MCPToolset` |
|---|---|---|---|
| `MCPServer*` | `max_retries` | `1` | `None` (inherits agent retries) |
| `MCPServer*` | `tool_prefix` | `None` | dropped — use `MCPToolset(...).prefixed('foo')` |
| `MCPServer*` | `elicitation_callback` | (kwarg) | renamed `elicitation_handler` (matches FastMCP / PAI conventions) |
| `MCPServer*` | `timeout` | `5` (connect) | renamed `init_timeout` (matches FastMCP) |
| `FastMCPToolset` | `cache_tools` / `cache_resources` | not present | default `True` with `list_changed` invalidation |
| `FastMCPToolset` | `read_timeout` | unset (no read timeout) | `300s` |

The `MCPServer*` `tool_prefix=` field is gone. Users who relied on the implicit prefix via `load_mcp_servers` get the same behavior from `load_mcp_toolsets` (which wraps with `.prefixed(name)`). Users who set `tool_prefix=` directly should migrate to `MCPToolset(...).prefixed('foo')`.

## What's NOT in this PR (intentionally deferred)

- **Documentation** — `docs/mcp/overview.md`, `docs/mcp/client.md`, `docs/mcp/fastmcp-client.md`, plus a new primary `MCPToolset` page and a migration guide. Substantial enough to warrant its own pass; tracked as the next step.
- **`MCP` capability `local` default flip** — the brief decision to make `local=True` the default and raise a clean install hint when FastMCP isn't installed. Lands separately.
- **Phase B (legacy class removal)** — happens on the v2 branch.
- **Wiring `MCPToolset` into the metadata exposure work in #3339** — comment posted on that PR explaining how to redirect.

## Files touched

- `pydantic_ai_slim/pydantic_ai/mcp.py` — new `MCPToolset`, `load_mcp_toolsets`, deprecation decorators on legacy classes
- `pydantic_ai_slim/pydantic_ai/toolsets/fastmcp.py` — `@deprecated` on `FastMCPToolset`
- `pydantic_ai_slim/pydantic_ai/capabilities/mcp.py` — capability `_default_local()` switched to `MCPToolset`; `local=` type union extended
- `pydantic_ai_slim/pydantic_ai/durable_exec/{dbos,temporal}/_{mcp,fastmcp_toolset,agent,toolset}.py` — file-level `reportDeprecated=false` on wrappers that delegate to deprecated classes by design
- `pydantic_ai_slim/pyproject.toml` — `mcp` extra now pulls `fastmcp[client]`
- `pyproject.toml` — `filterwarnings` ignores our deprecations during tests; pyright `reportDeprecated=false` for `tests/`
- `tests/test_capabilities.py::TestMCPCapability` — assertions updated for the new local toolset shape
- `tests/test_mcp_toolset.py` (new) — 17 tests covering construction logic, kwarg conflicts, http_client factory, sampling installation, resource type mapping, regression test that we no longer emit FastMCP's `sse_read_timeout` deprecation

### Checklist

- [x] No **breaking changes** in accordance with the [version policy](https://github.com/pydantic/pydantic-ai/blob/main/docs/version-policy.md). (Deprecations only — legacy paths still work.)


pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -1380,6 +1434,773 @@ def __eq__(self, value: object, /) -> bool:
 """
 
 
+_MCPToolsetClientInput: TypeAlias = (
```
DouweM: This should be a public type, since it's the type of a public field. And you shouldn't need to stringify it.

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -1380,6 +1434,773 @@ def __eq__(self, value: object, /) -> bool:
...
+    'FastMCPClient[Any] | ClientTransport | FastMCP | FastMCP1Server | AnyUrl | Path | MCPConfig | dict[str, Any] | str'
```
DouweM: We're going to move to the fastmcp client package once it exists, so we may need some special handling for the `FastMCP1Server` typing import because it may not be installed.

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -1380,6 +1434,773 @@ def __eq__(self, value: object, /) -> bool:
...
+
+    from pydantic_ai.mcp import MCPToolset
+
+    client = Client(StreamableHttpTransport('http://localhost:8000/mcp'), auth='oauth')
+    toolset = MCPToolset(client)
+    ```
+    """
+
+    client: FastMCPClient[Any]
+    """The underlying FastMCP client. Always normalized to a [`fastmcp.Client`][fastmcp.client.Client]
```
DouweM: This cross-link will not work

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -1380,6 +1434,773 @@ def __eq__(self, value: object, /) -> bool:
...
+
+    @functools.cached_property
+    def _enter_lock(self) -> anyio.Lock:
+        # `anyio.Lock` binds to the event loop on which it's first used; deferring creation to first
+        # access ensures it binds to the running loop and avoids issues with Temporal's workflow sandbox.
+        return anyio.Lock()
+
+    def __init__(
+        self,
+        client: _MCPToolsetClientInput,
```
DouweM: Don't love calling this `client` since it's so wide. But no better ideas.

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -1380,6 +1434,773 @@ def __eq__(self, value: object, /) -> bool:
...
+                self._cached_resources = None
+        return None
+
+    async def get_instructions(self, ctx: RunContext[AgentDepsT]) -> messages.InstructionPart | None:
+        """Return the server's instructions if `include_instructions` is enabled."""
+        if not self.include_instructions:
+            return None
+        if not self._initialized or self._instructions is None:
+            return None
+        return messages.InstructionPart(content=self._instructions, dynamic=True)
```
DouweM: Should this be dynamic? Is it likely to change during a run? We only load it once right?

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -1380,6 +1434,773 @@ def __eq__(self, value: object, /) -> bool:
...
+            mcp_tool.name: ToolsetTool[AgentDepsT](
+                toolset=self,
+                tool_def=ToolDefinition(
+                    name=mcp_tool.name,
+                    description=mcp_tool.description,
+                    parameters_json_schema=mcp_tool.inputSchema,
+                    metadata={
+                        'meta': mcp_tool.meta,
+                        'annotations': mcp_tool.annotations.model_dump() if mcp_tool.annotations else None,
+                        'output_schema': mcp_tool.outputSchema or None,
```
DouweM: We can drop this old field now that we have a native one!

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -1380,6 +1434,773 @@ def __eq__(self, value: object, /) -> bool:
...
+            tool_def=tool_def,
+            max_retries=self.max_retries if self.max_retries is not None else 1,
+            args_validator=TOOL_SCHEMA_VALIDATOR,
+        )
+
+    async def direct_call_tool(
+        self,
+        name: str,
+        args: dict[str, Any],
+        metadata: dict[str, Any] | None = None,
```
DouweM, line 2120: require kwargs after name/args

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -1380,6 +1434,773 @@ def __eq__(self, value: object, /) -> bool:
...
+                    raise exceptions.ModelRetry(message=str(e)) from e
+                raise
+
+        # Prefer structured content if all parts are text (per the docs they contain the JSON-encoded
+        # structured content for backward compatibility).
+        # See https://github.com/modelcontextprotocol/python-sdk#structured-output
+        if (structured := result.structured_content) and all(
+            isinstance(part, mcp_types.TextContent) for part in result.content
+        ):
+            # The MCP SDK wraps primitives and generic types like list in a `result` key, but we want
```
DouweM, line 2147: Do we need this with FastMCP? DId FastMCPToolset have this logic? Or just MCPServer?

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -1380,6 +1434,773 @@ def __eq__(self, value: object, /) -> bool:
...
+    so the kwargs take effect. For everything else, we let the FastMCP `Client` infer the transport
+    from the input directly (we just pass it through).
+    """
+    needs_explicit_http = headers is not None or http_client is not None or auth is not None or verify is not None
+    is_url = isinstance(client, AnyUrl) or (isinstance(client, str) and client.startswith(('http://', 'https://')))
+    if needs_explicit_http and is_url:
+        url = str(client)
+        # FastMCP's HTTP transports accept `httpx_client_factory`; we adapt `http_client` to that.
+        factory = _make_httpx_client_factory(http_client) if http_client is not None else None
+        if url.endswith('/sse'):
```
DouweM: I don't love duplicating this logic from inside FastMCP, do we have to?

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -1380,6 +1434,773 @@ def __eq__(self, value: object, /) -> bool:
...
+    parts: Sequence[mcp_types.ContentBlock],
+) -> str | messages.BinaryContent | list[str | messages.BinaryContent]:
+    mapped = [_map_fastmcp_tool_result(part) for part in parts]
+    return mapped[0] if len(mapped) == 1 else mapped
+
+
+def _map_fastmcp_tool_result(part: mcp_types.ContentBlock) -> str | messages.BinaryContent:
+    if isinstance(part, mcp_types.TextContent):
+        text = part.text
+        if text.startswith(('[', '{')):
```
DouweM, line 2382: Do we need this? Do we currently do this in FastMCPToolset? Maybe we don't need to because it's own structured output handling is smarter? Investigate its source, you can read it.

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -1380,6 +1434,773 @@ def __eq__(self, value: object, /) -> bool:
...
+
+
+def _map_fastmcp_tool_results(
+    parts: Sequence[mcp_types.ContentBlock],
+) -> str | messages.BinaryContent | list[str | messages.BinaryContent]:
+    mapped = [_map_fastmcp_tool_result(part) for part in parts]
+    return mapped[0] if len(mapped) == 1 else mapped
+
+
+def _map_fastmcp_tool_result(part: mcp_types.ContentBlock) -> str | messages.BinaryContent:
```
DouweM: These are mcp types, not fastmcp types, so we don't need to call it "fastmcp" right?

pydantic_ai_slim/pydantic_ai/toolsets/fastmcp.py
```diff
@@ -57,13 +58,23 @@
 UNKNOWN_BINARY_MEDIA_TYPE = 'application/octet-stream'
 
 
+@deprecated(
+    '`FastMCPToolset` is deprecated and will be removed in v2. '
+    'Use `pydantic_ai.mcp.MCPToolset` instead — it accepts the same input shapes and adds full '
```
DouweM: Make it obvious that it still uses FastMCP and accepts clients.

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -1476,10 +2304,60 @@ def load_mcp_servers(config_path: str | Path) -> list[MCPServerStdio | MCPServer
...
+    config = MCPServerConfig.model_validate(expanded_config_data)
+
+    toolsets: list[AbstractToolset[Any]] = []
+    for name, server in config.mcp_servers.items():
+        toolset: MCPToolset[Any]
+        if isinstance(server, MCPServerStdio):  # pyright: ignore[reportDeprecated]
+            stdio_entry: dict[str, Any] = {'command': server.command, 'args': list(server.args)}
+            if server.env is not None:
+                stdio_entry['env'] = server.env
+            toolset = MCPToolset({'mcpServers': {name: stdio_entry}}, id=name)
```
DouweM: Why do we have to go through the dict syntax here?

pydantic_ai_slim/pydantic_ai/capabilities/mcp.py
DouweM, line 1: *(Refers to line 53)*

I think we should support more than just str here -- everything we support on `MCPToolset`. Do think about `from_spec` and `AgentSpec`, not all types  will be representable there, and accepting a dict as the first argument has tricky interactions with how other kwargs are handled, I think.

DouweM: If a non-str client is provided, we can no longer fall back on the builtin tool, of course.

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -42,10 +43,40 @@
...
+        StreamableHttpTransport,
+    )
+    from fastmcp.exceptions import ToolError
+    from fastmcp.mcp_config import MCPConfig
+    from fastmcp.server import FastMCP
+    from mcp.server.fastmcp import FastMCP as FastMCP1Server
+except ImportError as _fastmcp_import_error:  # pragma: no cover
+    raise ImportError(
+        'Please install the `fastmcp` package to use `MCPToolset`, '
+        'you can use the `mcp` or `fastmcp` optional group — `pip install "pydantic-ai-slim[mcp]"`'
```
DouweM: We're gonna deprecate the latter optional group right?

pydantic_ai_slim/pyproject.toml
```diff
@@ -109,9 +109,11 @@ cli = [
...
 ]
-# MCP
-mcp = ["mcp>=1.25.0,<2.0"]
-# FastMCP
+# MCP — installs both the legacy `mcp` SDK (used by deprecated `MCPServer*`) and `fastmcp` (used
+# by the recommended `MCPToolset`). The legacy `mcp` requirement will be dropped in v2 once the
+# `MCPServer*` hierarchy is removed.
+mcp = ["mcp>=1.25.0,<2.0", "fastmcp>=3.2.4"]
+# FastMCP — alias kept for backward compatibility; prefer the `mcp` extra. Will be removed in v2.
 fastmcp = ["fastmcp>=3.2.4"]
```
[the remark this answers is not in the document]
DouweM: Hmm good point on `MCPServer`! So we should split that out, and make sure that it can still be imported by itself with just `mcp`. Same for `MCP` capability.

pyproject.toml
```diff
@@ -222,9 +222,15 @@ include = [
...
+# which are not otherwise used.
+# `reportDeprecated = false` for tests so coverage on the deprecated `MCPServer*` / `FastMCPToolset`
+# / `load_mcp_servers` paths can be maintained until v2-cut without ignore comments at every callsite.
 executionEnvironments = [
-    { root = "tests", reportUnusedFunction = false, reportPrivateImportUsage = false },
+    # Tests still reference the deprecated `MCPServer*` / `FastMCPToolset` / `load_mcp_servers` paths
+    # to maintain coverage until v2-cut. `reportUnnecessaryTypeIgnoreComment = false` because some
+    # pre-existing per-line `# pyright: ignore[reportDeprecated]` (for non-MCP deprecations) are now
+    # technically redundant under `reportDeprecated = false`; clean them up at v2-cut.
+    { root = "tests", reportUnusedFunction = false, reportPrivateImportUsage = false, reportDeprecated = false, reportUnnecessaryTypeIgnoreComment = false },
```
DouweM: We should not be adding either of these. I prefer local comments on every call site.

pyproject.toml
```diff
@@ -278,6 +284,11 @@ filterwarnings = [
     # logfire to use logging.warning() instead of warnings.warn() for this diagnostic — it's
     # informational, not a deprecation, and should never be able to crash execution.
     "ignore:Found propagated trace context:RuntimeWarning",
+    # v2 prep: legacy MCP path deprecations. These fire from existing tests that exercise
+    # `MCPServer*` and `FastMCPToolset` directly to keep coverage on the deprecated classes
+    # until v2-cut. Tests targeting `MCPToolset` don't trigger these.
+    "ignore:.*is deprecated and will be removed in v2:DeprecationWarning",
```
DouweM: No, we need to know when our code still uses some deprecated format unintentionally

tests/test_mcp_toolset.py
```diff
@@ -0,0 +1,721 @@
+"""Tests for `pydantic_ai.mcp.MCPToolset`.
```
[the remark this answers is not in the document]
DouweM: Keeping these in a dedicated `tests/test_mcp_toolset.py` rather than folding into `tests/test_mcp.py` — `test_mcp.py` is already the home of the legacy `MCPServer*` tests and folding the new `MCPToolset` suite in would balloon it past readability. The 1:1 source/test correspondence is the default; we're making a conscious exception here for the v2 hierarchy split.

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -1508,6 +1615,788 @@ def __eq__(self, value: object, /) -> bool:
 """
 
 
+MCPToolsetClient: TypeAlias = 'FastMCPClient[Any] | ClientTransport | FastMCP | FastMCP1Server | AnyUrl | Path | str'
```
DouweM, line 1626: Who we need the quotes here? Can we import the packages we need inside `if TYPE_CHECKING:`?

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -1508,6 +1615,788 @@ def __eq__(self, value: object, /) -> bool:
...
+    client = Client(StreamableHttpTransport('http://localhost:8000/mcp'), auth='oauth')
+    toolset = MCPToolset(client)
+    ```
+    """
+
+    client: FastMCPClient[Any]
+    """The underlying FastMCP `Client`. Always normalized to a `fastmcp.Client` regardless of how
+    the toolset was constructed."""
+
+    tool_error_behavior: Literal['model_retry', 'error']
```
DouweM: Just call it `retry` please -- I realize that `FastMCPToolset` called it `model_retry`, but I prefer `retry`

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -1508,6 +1615,788 @@ def __eq__(self, value: object, /) -> bool:
...
+
+    def __repr__(self) -> str:
+        repr_args = [f'client={self.client!r}']
+        if self._id is not None:
+            repr_args.append(f'id={self._id!r}')
+        return f'{self.__class__.__name__}({", ".join(repr_args)})'
+
+    def __eq__(self, value: object, /) -> bool:
+        return isinstance(value, MCPToolset) and self._id == value._id and self.client is value.client
+
```
DouweM, line 2254: Does this need to be a callable class? Could it be a function like `_build_sampling_handler`?

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -43,10 +44,73 @@
...
+        StdioTransport,
+        StreamableHttpTransport,
+    )
+    from fastmcp.exceptions import ToolError
+    from fastmcp.mcp_config import infer_transport_type_from_url
+    from fastmcp.server import FastMCP
+    from mcp.server.fastmcp import FastMCP as FastMCP1Server
+
+
+# `fastmcp` is an optional dependency: it's required for `MCPToolset` (the recommended path) but
```
DouweM: This comment needs updating now that we use `fastmcp-slim[client]`, and the `fastmcp` optional group is/should be deprecated -- because in v2 it'll be gone, and users will get the slim client via the updated `mcp` group

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -43,10 +44,73 @@
...
+    _fastmcp_import_error = _err
+else:
+    _fastmcp_import_error = None
+
+
+def _require_fastmcp() -> None:
+    """Raise [`ImportError`][ImportError] if `fastmcp` isn't installed."""
+    if _fastmcp_import_error is not None:  # pragma: no cover
+        raise ImportError(
+            'Please install the `fastmcp` package to use `MCPToolset`. '
```
DouweM: or slim client!

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -1487,8 +1579,23 @@ def __eq__(self, value: object, /) -> bool:
...
+    async def __call__(
+        self,
+        name: str,
+        args: dict[str, Any],
+        *,
+        metadata: dict[str, Any] | None = None,
+    ) -> ToolResult: ...
+
 
 ProcessToolCallback = Callable[
```
DouweM, line 1604: Do we need to keep this now that we have the protocol? Should we deprecate it somehow? We do some getattr trickery to get depr warnings in other places.

pydantic_ai_slim/pydantic_ai/mcp.py
DouweM, line 1: *(Refers to line 2403)*

Public type that should be deprecated? Think about deprecating all the old stuff we'll be removing in the next phase.

pydantic_ai_slim/pydantic_ai/capabilities/mcp.py
```diff
@@ -84,6 +92,19 @@ def __init__(
...
+        # FastMCP server, pre-built `fastmcp.Client`, `AnyUrl`, etc.) are wrapped into an
+        # `MCPToolset` here. Strings flow through `_resolve_local_strategy` below; pre-built
+        # toolsets, callables, bools, and `None` pass through to `NativeOrLocalTool` unchanged.
+        if (
+            local is not None
+            and not isinstance(local, (bool, str))
+            and not isinstance(local, AbstractToolset)
+            and not callable(local)
+        ):
+            from pydantic_ai.mcp import MCPToolset as _MCPToolset
```
DouweM: This is to force an error, I suppose? We could also have set an `mcp_loaded` or smth boolean up top?

pydantic_ai_slim/pydantic_ai/capabilities/mcp.py
```diff
@@ -125,33 +146,27 @@ def _default_local(self) -> Tool[AgentDepsT] | AbstractToolset[AgentDepsT] | Non
...
-            f'MCP(local={name!r}) is not a known strategy. '
-            'Pass `local=True` for the default local MCP transport, or a Tool/callable directly.'
-        )
+        # MCP has no named string strategies — strings are treated as fastmcp connect specs
+        # (URL/path/etc.) and wrapped in `MCPToolset`. `local=True` uses the URL from `MCP(url=...)`.
+        try:
+            if isinstance(name, str):
+                from pydantic_ai.mcp import MCPToolset
+
+                return MCPToolset(name, include_instructions=True)
```
DouweM: Should we not consider local headers here, like we do in `_build_local`? How does this relate to that other method? Why is strategy-resolution a sep concept?

pydantic_ai_slim/pydantic_ai/capabilities/mcp.py
```diff
@@ -125,33 +146,27 @@ def _default_local(self) -> Tool[AgentDepsT] | AbstractToolset[AgentDepsT] | Non
...
-            from pydantic_ai.mcp import MCPServerSSE
-
-            return MCPServerSSE(self.url, headers=local_headers or None, include_instructions=True)
+        # `MCPToolset` infers SSE vs Streamable HTTP from the URL.
+        from pydantic_ai.mcp import MCPToolset
 
-        from pydantic_ai.mcp import MCPServerStreamableHTTP
-
-        return MCPServerStreamableHTTP(self.url, headers=local_headers or None, include_instructions=True)
+        return MCPToolset(self.url, headers=local_headers or None, include_instructions=True)
```
DouweM: So on the local path, we support the user passing a non-URL string value for `fastmcp.Client` (like a file path name), and it will work fine. But if we use the native path instead, it won't work, as that requires an actual URL. We have logic for "requires local etc, where we could maybe use it"? But I'd rather really only support URLs there, not just about everything. Can we validate it is a URL? Change the field type to AnyUrl or smth?

pydantic_ai_slim/pydantic_ai/capabilities/mcp.py
```diff
@@ -160,5 +175,36 @@ def get_toolset(self) -> AbstractToolset[AgentDepsT] | None:
...
+        id: str | None = None,
+        authorization_token: str | None = None,
+        headers: dict[str, str] | None = None,
+        allowed_tools: list[str] | None = None,
+        description: str | None = None,
+    ) -> MCP[AgentDepsT]:
+        """Construct an `MCP` capability from spec-serializable args.
+
+        Restricts the runtime-wide `local=` union to the JSON/YAML-serializable subset
+        (`str | bool | None`) so `AgentSpec` schema generation works. Non-serializable runtime
```
DouweM, line 199: `AgentSpec`s are potentially loaded from user-provided yaml files, so passing a script path could be abused for RCE, and a local URL for SSRF. So we will at some point need to have the concept of safe and unsafe capabilities/configs, or a safe mode to run spec parsing in.

pydantic_ai_slim/pydantic_ai/durable_exec/dbos/_agent.py
```diff
@@ -1,3 +1,5 @@
+# pyright: reportDeprecated=false
+# References deprecated `MCPServer*` / `FastMCPToolset` for backward-compat dispatch. Removed in v2.
```
DouweM: I'd rather have an ignore comment on each offending line, than here at the top!

pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_mcp.py
```diff
@@ -1,3 +1,5 @@
+# pyright: reportDeprecated=false
+# Wraps the deprecated `MCPServer*` and `FastMCPToolset` for durable execution. Removed in v2.
```
DouweM: Same as up -- rather have it on each offending line

pydantic_ai_slim/pydantic_ai/mcp.py
```diff
@@ -986,6 +1057,11 @@ def __eq__(self, value: object, /) -> bool:
         return isinstance(value, MCPServer) and self.id == value.id and self.tool_prefix == value.tool_prefix
 
 
+@deprecated(
+    '`MCPServerStdio` is deprecated and will be removed in v2. '
+    "Use `MCPToolset('path/to/script.py')` for Python scripts, `MCPToolset('script.js')` for "
+    "Node scripts, or `MCPToolset(StdioTransport(command='...', args=[...]))` for arbitrary commands."
```
DouweM: This should have `fastmcp.` or smth, no?

pyproject.toml
```diff
@@ -287,6 +287,15 @@ filterwarnings = [
...
     "ignore:Found propagated trace context:RuntimeWarning",
+    # v2 prep: legacy MCP classes & helper. Targeted per-class filters (vs a wildcard) so
+    # *unintentional* uses of any other v2-deprecated API still surface as errors. Tests that
+    # exercise the legacy hierarchy directly (test_mcp.py, test_fastmcp.py) and tests that build
+    # MCP fixtures at module level (test_dbos.py, test_temporal.py, etc.) need these.
+    "ignore:`MCPServerStdio` is deprecated and will be removed in v2:DeprecationWarning",
+    "ignore:`MCPServerSSE` is deprecated and will be removed in v2:DeprecationWarning",
+    "ignore:`MCPServerStreamableHTTP` is deprecated and will be removed in v2:DeprecationWarning",
+    "ignore:`FastMCPToolset` is deprecated and will be removed in v2:DeprecationWarning",
+    "ignore:`load_mcp_servers` is deprecated and will be removed in v2:DeprecationWarning",
```
DouweM: Please remove all these ignores here -- I want them ignored in the specific files that are dedicated to the old thing, or in the specific call sites with individual comments, so that we don't accidentally reintroduce these things in other PRs.

DouweM:
Worked through this sweep in two commits (`5e444a64b` and `d1091d239`). Summary of what's applied and the open-question answers.

## Applied as suggested

- **mcp.py:1063** — Deprecation message now points at `fastmcp.client.transports.StdioTransport`.
- **mcp.py:1494** — Removed the broken `[fastmcp.exceptions.ToolError][...]` cross-link; mkdocstrings can't resolve fastmcp symbols, so it's plain backticks now.
- **mcp.py:1618** — Kept the string forward-reference (TypeAlias RHS is evaluated eagerly at module-import time even with `from __future__ import annotations`, and the names only exist when fastmcp is installed) but added a comment explaining why. `if TYPE_CHECKING:` imports also moved up.
- **mcp.py:1683** — Renamed `tool_error_behavior='model_retry'` → `'retry'`. Legacy `FastMCPToolset` keeps `'model_retry'` since it's already shipped.
- **mcp.py:2245** — `_CacheInvalidatingMessageHandler` callable class → `_build_message_handler` closure, parity with `_build_sampling_handler`.
- **mcp.py:68 / 101** — Comment + install hint updated for `fastmcp-slim[client]` via the `[mcp]` extra; `[fastmcp]` flagged as deprecated.
- **mcp.py:2403** — `MCPServerConfig` deprecated via module-level `__getattr__`. Renamed the implementation to private `_MCPServerConfig` so internal `load_mcp_*` callers don't trigger their own deprecation. `from pydantic_ai.mcp import MCPServerConfig` now warns.
- **capabilities/mcp.py:53 + 105** — `local=` already widened to the full `MCPToolsetClient` union in the previous round; dropped the inline `from pydantic_ai.mcp import MCPToolset as _MCPToolset` (reaching the widening branch implies fastmcp is installed, so the module-level binding is the real class).
- **capabilities/mcp.py:155 + 169** — Refactored `_resolve_local_strategy` to delegate to `_build_local(url)`, so headers/auth flow consistently for both `local=True` (uses `self.url`) and `local='https://override...'` (uses the override). Added a `_require_url` validator: non-URL strings now raise `UserError` directing the user to `local=MCPToolset(...)`.
- **dbos/_agent.py:2 + temporal/_mcp.py:2** — Dropped file-level `# pyright: reportDeprecated=false`; per-line `# pyright: ignore[reportDeprecated]` only on the `FastMCPToolset` imports/checks (the `MCPServer` abstract base isn't `@deprecated`).
- **toolsets/fastmcp.py:63** — Docstring rewritten to lead with "Toolset backed by a FastMCP `Client`" and explicitly mention `Client`/`ClientTransport` inputs.
- **pyproject.toml:298** — Removed all five `MCPServer*`/`FastMCPToolset`/`load_mcp_servers` wildcard `filterwarnings` entries. Scoped to per-call-site (`warnings.catch_warnings()` blocks around module-level Agent constructions in `test_{dbos,temporal,prefect}.py`), per-test (`@pytest.mark.filterwarnings(...)` on the four `test_agent.py` MCP tests and two `test_toolsets.py` MCP tests), and file-level `pytestmark` on the dedicated `test_mcp.py` / `test_fastmcp.py`. Also wrapped the discriminator-driven `MCPServer*` constructions inside `load_mcp_{servers,toolsets}` in `warnings.catch_warnings()` since those are an implementation detail.
- **pyproject.toml:233/290** — The earlier `reportDeprecated = false` execution-environment / wildcard filter were already pulled in a prior iteration before this sweep landed.

## Answers / "make-the-call" replies

- **mcp.py:1437** ("public type, no stringification"): It's public (`MCPToolsetClient` in `__all__` since the previous round). I kept the RHS stringified — explained inline. Open to switching to a 3.10-compat `TypeAliasType` wrapper if you'd prefer; but the practical effect is identical and the string is idiomatic for "alias to forward-referenced optional-dep types".
- **mcp.py:1438** (`FastMCP1Server` import may not be installed with the slim client): It's still available because `fastmcp-slim[client]` pulls `mcp>=1.0` which ships `mcp.server.fastmcp.FastMCP`. The TYPE_CHECKING import succeeds on the slim install. If `mcp` ever drops the server submodule from its base wheel we'll need to gate it.
- **mcp.py:1585** ("don't love `client`"): No change. Agreeing with you that nothing better surfaces; renaming would diverge from `fastmcp.Client(transport=...)` parameter naming which is what users will be reaching for.
- **mcp.py:1600** (`ProcessToolCallback` deprecation?): Kept. `CallToolFunc` (Protocol) and `ProcessToolCallback` (Callable alias) describe two different callables — `CallToolFunc` is the *inner* tool-call hook (what your callback receives), `ProcessToolCallback` is the *outer* user-written wrapper signature. Deprecating it would force users to spell out the four-argument `Callable[[RunContext[Any], CallToolFunc, str, dict[str, Any]], Awaitable[ToolResult]]` shape. The alias is the DX win.
- **mcp.py:2094** ("don't love duplicating from FastMCP"): The `_build_transport` helper exists because `fastmcp.Client(url, headers=..., auth=..., verify=..., http_client=...)` does *not* forward HTTP kwargs to its auto-inferred transport — FastMCP's auto-infer ignores them. Either we build the transport explicitly or those kwargs silently drop. Open to filing this as a fastmcp upstream issue, but in 3.3.0 the workaround is still needed.
- **mcp.py:2138** ("Do we need this with FastMCP?"): Yes. FastMCP's `result.data` only unwraps `{'result': value}` when the server sets `_meta.fastmcp.wrap_result` or the schema has `x-fastmcp-wrap-result`. MCP-SDK-only servers (no FastMCP) auto-wrap primitives in `{'result': value}` without setting those hints — `FastMCPToolset` had the same backstop. Without it, primitives from non-FastMCP servers come back as `{'result': 5}` instead of `5`.
- **mcp.py:2364** ("JSON parsing in `_map_mcp_tool_result`"): Same answer — `FastMCPToolset.direct_call_tool` has the exact same parse-text-as-JSON fallback for legacy servers that only return TextContent. Removing it would regress those.
- **capabilities/mcp.py:194** (security note about RCE/SSRF from YAML specs): Filed as v2 follow-up — we'll need a safe-mode flag on `AgentSpec.from_file` / capability registration that disables `local=<file path>` and `local=<http URL>` inputs (and probably blocks `MCPToolset(<path>)` constructions inside `from_spec`). Not in scope for this PR but worth a card.

PR #5326, opened 2026-05-07
Fix thinking being silently dropped for Claude Sonnet 4.6 and Opus 4.6 on Bedrock, and support adaptive thinking efforts

### LINKED ISSUE #5304: Bedrock: add adaptive thinking support for Claude Sonnet 4.6 / Opus 4.6
### Initial Checks

### Description

The Bedrock provider does not support `thinking.type: "adaptive"`, which AWS [requires for Claude Sonnet 4.6 and Opus 4.6](https://docs.aws.amazon.com/bedrock/latest/userguide/claude-messages-adaptive-thinking.html). The provider always sends `{"type": "enabled", "budget_tokens": N}`, which is deprecated for these models.

**This has a concrete impact:** we tested directly against the Bedrock Converse API and confirmed that with `thinking.type: "enabled"`, **thinking is silently disabled after the first generation** in multi-turn agent runs. The model thinks on the first call but produces zero `reasoningContent` blocks on all subsequent calls (after tool use), regardless of whether thinking blocks are properly round-tripped in the message history.

Switching to `thinking.type: "adaptive"` restores thinking on subsequent generations:

| Call | Thinking config | Thinking blocks round-tripped | Produces thinking |
|------|----------------|-------------------------------|-------------------|
| 1st | `enabled` + `budget_tokens` | N/A | Yes |
| 2nd | `enabled` + `budget_tokens` | Yes | **No** |
| 2nd | `enabled` + `budget_tokens` | No | **No** |
| 2nd | `adaptive` | Yes | **Yes** |
| 2nd | `adaptive` | No | **No** |

This means any pydantic-ai user running a multi-turn agent on Bedrock with Sonnet 4.6 / Opus 4.6 silently gets no reasoning after the first LLM call — where tool results are interpreted and decisions are made.

The Anthropic direct API provider already supports adaptive thinking (#4216, #3894). The Bedrock provider does not — the unified thinking docs note "No adaptive support" for Bedrock.

Note: the `effort` sub-parameter inside `adaptive` (e.g. `{"type": "adaptive", "effort": "high"}`) is not yet accepted by the Bedrock Converse API — it returns `ValidationException: thinking.adaptive.effort: Extra inputs are not permitted`. Plain `{"type": "adaptive"}` works and defaults to `high`.

### Workaround

Pass the config manually via `bedrock_additional_model_requests_fields`:

```python
BedrockModelSettings(
    bedrock_additional_model_requests_fields={
        "thinking": {"type": "adaptive"}
    }
)
```

This works because `_translate_thinking` skips the `thinking` key when it's already present in the additional fields.

- Python: 3.12
- Pydantic AI: 1.85.1
- boto3 / botocore (Bedrock)

sarth6:
## What's the change

Closes #5304.

The Bedrock provider always emits `{'type': 'enabled', 'budget_tokens': N}` for the Anthropic thinking variant, which AWS deprecated for Claude Sonnet 4.6 / Opus 4.6. The reporter ran a direct Bedrock Converse test matrix and confirmed this is **worse than just deprecation** — multi-turn agent runs silently lose all thinking after the first generation:

| Call | Thinking config       | Round-trip thinking blocks | Produces thinking |
|------|-----------------------|----------------------------|-------------------|
| 1st  | `enabled` + budget    | N/A                        | Yes               |
| 2nd  | `enabled` + budget    | Yes                        | **No**            |
| 2nd  | `enabled` + budget    | No                         | **No**            |
| 2nd  | `adaptive`            | Yes                        | **Yes**           |
| 2nd  | `adaptive`            | No                         | No                |

So today, any pydantic-ai user running a multi-turn Bedrock agent on Sonnet/Opus 4.6 gets zero reasoning after the first LLM call — exactly where tool results are interpreted and decisions are made. No error, just silent capability loss.

This PR mirrors the adaptive support that #4829 added to the direct Anthropic provider over to Bedrock, including the `output_config.effort` plumbing so `thinking='low'/'medium'/'high'` produce semantically-distinct Bedrock requests.

## Implementation

1. Two new flags on `BedrockModelProfile` (Bedrock-prefixed per project convention):
   - `bedrock_supports_adaptive_thinking: bool` — whether `{'thinking': {'type': 'adaptive'}}` is accepted.
   - `bedrock_supports_effort: bool` — whether `output_config.effort` is accepted at the top level of `additionalModelRequestFields` (sibling of `thinking`).
2. The inline `'anthropic'` lambda in `BedrockProvider.model_profile()` is extracted to a named `bedrock_anthropic_model_profile()` factory, matching the existing `bedrock_amazon_model_profile` / `bedrock_deepseek_model_profile` pattern. The factory reads `anthropic_supports_adaptive_thinking` and `anthropic_supports_effort` from the downstream `AnthropicModelProfile` *before* `BedrockModelProfile.update()` strips them (`ModelProfile.update()` only copies fields present on `self`).
3. `BedrockConverseModel._translate_thinking()` branches on the new flags:
   - Adaptive models get `{'thinking': {'type': 'adaptive'}}` plus, when `thinking` is a string level and `bedrock_supports_effort` is set, a sibling `{'output_config': {'effort': ...}}` field. Effort map mirrors the direct Anthropic provider's exactly (`minimal/low → 'low'`, `medium → 'medium'`, `high → 'high'`, `xhigh → 'max'` — AWS only documents `low|medium|high|max` on Bedrock, not `xhigh`).
   - Pre-4.6 models keep the legacy budget-token shape unchanged.

| Model         | `thinking=True`       | `thinking='high'`                                                | `thinking=False`       |
|---------------|-----------------------|------------------------------------------------------------------|------------------------|
| Adaptive 4.6+ | `{type: 'adaptive'}`  | `{type: 'adaptive'}` + `output_config.effort='high'`             | omitted                |
| Pre-4.6       | `{type: 'enabled', budget_tokens: 10000}` | `{type: 'enabled', budget_tokens: 16384}`    | `{type: 'disabled'}`   |

## Note on `output_config.effort` (review feedback)

The PR initially documented `effort` as silently dropped because AWS rejects `{'thinking': {'type': 'adaptive', 'effort': 'high'}}`. That was wrong — [@DouweM](https://github.com/DouweM) and the automated reviewer pointed out (and the [AWS Adaptive Thinking docs](https://docs.aws.amazon.com/bedrock/latest/userguide/claude-messages-adaptive-thinking.html) confirm verbatim) that `effort` belongs in a **sibling** `output_config` field, not inside `thinking`. The docs spell out the trap: *"The `effort` parameter must be placed inside a separate `output_config` object … Placing `effort` inside `thinking` will result in a `ValidationException`."* This PR now wires the sibling path correctly.

`thinking=False` on adaptive models omits the field, matching direct Anthropic's `OMIT` behavior (`models/anthropic.py:_translate_thinking`). The legacy pre-4.6 path keeps `{'type': 'disabled'}` unchanged. Adaptive models have no documented `disabled` shape.

## Tests

- **`tests/test_thinking.py::TestBedrockThinkingTranslation`** — unit tests for translation behavior:
  - Bare `thinking=True` enables adaptive without setting effort (matches direct Anthropic).
  - `thinking='high'` adds `output_config.effort='high'` alongside `{'type': 'adaptive'}`.
  - Parametrized effort map for `minimal`/`low`/`medium`/`high`/`xhigh` (`xhigh → 'max'`).
  - Effort omitted when `bedrock_supports_effort` is False.
  - User-provided `output_config` in `bedrock_additional_model_requests_fields` is preserved (defensive).
  - `thinking=False` on adaptive omits both `thinking` and `output_config`.
- **`tests/providers/test_bedrock.py::test_bedrock_provider_model_profile`** — factory wiring assertions: `bedrock_supports_adaptive_thinking` and `bedrock_supports_effort` are False on Sonnet 4.5, True on Sonnet 4.6, True on Opus 4.6.
- **End-to-end VCR cassettes** in `tests/models/test_bedrock.py`:
  - `test_bedrock_model_thinking_part_anthropic_legacy` — pre-4.6 (Sonnet 4.5) with unified `thinking=True` → legacy `{'type': 'enabled', 'budget_tokens': 10000}` translation, both turns return `reasoningContent`.
  - `test_bedrock_model_thinking_part_anthropic_adaptive` — Sonnet 4.6 with unified `thinking=True` → `{'type': 'adaptive'}` translation (no effort), both turns return `reasoningContent`. **This is the regression from #5304** — proves it doesn't reproduce.
  - `test_bedrock_model_thinking_part_anthropic_adaptive_effort` — Sonnet 4.6 with unified `thinking='high'` → `{'type': 'adaptive'} + output_config.effort='high'`, both turns return `reasoningContent`. Proves the new effort path lands on the wire (the cassette body shows the two sibling fields).
  - The pre-existing `test_bedrock_model_thinking_part_anthropic` exercises the manual `bedrock_additional_model_requests_fields` workaround on pre-4.6 (passthrough path); unchanged.
- All `tests/test_thinking.py` + `tests/providers/test_bedrock.py` + `tests/models/test_bedrock.py` pass with all AWS / gateway credentials stripped (deterministic replay).


DouweM:
@sarth6 Thanks Sarth! Any particular reason why it's still a draft?

sarth6:
@DouweM I just wanted to do some local vcr verification first (skipping that caused me to make a mistake last time) - but if you think it's good to merge feel free to go ahead :)

sarth6:
Update: cassette test is now landed (commit `0cdf1f7f`). Quick note on the recording path —

I requested AWS Bedrock model access for Claude Sonnet 4.6 in my own AWS account, but Amazon's Anthropic use-case form approval was still pending when I went to record (their on-demand throughput gating is separate from the retired Model Access page, and the form turnaround is variable). Rather than block on that, I recorded via the Pydantic AI Gateway (`gateway_provider('bedrock', ...)` → upstream Bedrock), since the gateway's AWS account is already approved for Anthropic models.

The actual Converse request body is identical either way — the gateway is a transparent proxy — so the bug-fix proof is the same: cassette captures `{"type": "adaptive"}` on both turn-1 and turn-2 requests, with `reasoningContent` blocks in both responses, on `us.anthropic.claude-sonnet-4-6`. That's the regression from the original issue's test matrix.

One stylistic difference: the cassette URL is `gateway-us.pydantic.dev/proxy/bedrock` rather than `bedrock-runtime.<region>.amazonaws.com` like the existing direct-AWS thinking cassettes. The test docstring calls this out and `base_url` is pinned so replay matches deterministically (the gateway's `_infer_base_url` derives a region from the api-key prefix, which diverges between real-key recording and fake-key CI replay).

If maintainers prefer a direct-AWS cassette I can re-record once my Anthropic approval comes through, but the unit + factory tests + this cassette already cover all the behavioral paths: request encoding, factory wiring, two-turn `reasoningContent` round-trip.

DouweM:
@sarth6 I'm fine with manually faking the URL in the cassette to match what the other cassettes use, so that we can remove the gateway URL override from the test.

sarth6:
@tennox @iddoaviram-rs would appreciate any feedback you have on the PR here, which should fix https://github.com/pydantic/pydantic-ai/issues/5304

cc @DouweM

DouweM:
> * **`effort` is silently dropped on Bedrock.** AWS rejects `{'type': 'adaptive', 'effort': 'high'}` with `ValidationException`. Plain `{'type': 'adaptive'}` defaults to `'high'`. So `thinking='low'`, `thinking='medium'`, and `thinking='high'` produce identical Bedrock requests on 4.6+. This matches `pydantic_ai/models/CLAUDE.md`'s "document unsupported settings, silently ignore at runtime" policy and is documented in `docs/thinking.md`. If AWS adds `effort` support later, lifting the restriction is a one-line change.

@sarth6 https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-anthropic-claude-messages-request-response.html suggests that `output_config['effort']` is supported; with `AnthropicModel` the adaptive thinking effort it also ends up there:

https://github.com/pydantic/pydantic-ai/blob/b7806da2e8e27f972f59f8c188e42e1354d964cf/pydantic_ai_slim/pydantic_ai/models/anthropic.py#L1941-L1942

sarth6:
@DouweM ready for another review - good catch, missed that case and added a VCR test with an assertion for that. I wasn't certain on the naming convention for bedrock profile fields that are specific to anthropic but tried to follow existing conventions where possible. I was considering a protocol as well to enforce standardization and encourage reducing code duplication between all anthropic-routeable providers, something like:

```
class AnthropicRoutableProtocol(Protocol):
    supports_adaptive_thinking: bool
    supports_effort: bool
    supports_xhigh_effort: bool
```

but I didn't want to bloat this PR. Finding better ways to keep the router providers (bedrock/openrouter/etc) clean is on my mind though

PR #5327, opened 2026-05-07
Add RetryCount evaluator + tool-retry span plumbing

Kludex:
## Summary

Surfaces tool-call retries as a first-class signal in pydantic-evals. Builds on top of #5130 so `RetryCount` sits naturally next to `StepEfficiency` and `ArgumentCorrectness`.

End-to-end plumbing across three layers:

- **`pydantic_ai.tool_manager`**: when a tool call ends in a `ToolRetryError`, set `pydantic_ai.tool.retry = True` on the tool span before re-raising. The attribute name is centralised on `InstrumentationNames.tool_retry_attr` next to the existing `tool_deferral_*` attributes. Already-set Status/exception on the span is unchanged - this is purely additive.
- **`pydantic_evals.dataset._extract_span_tree_metrics`**: count spans with `pydantic_ai.tool.retry == True` into `ctx.metrics['retries']`, same shape as the existing `'requests'` / `'cost'` / token-usage extraction.
- **`pydantic_evals.evaluators.agentic.RetryCount`**: new evaluator, exported from `pydantic_evals.evaluators`. Mirrors `StepEfficiency`'s shape - returns a mapping under stable keys `'retries'` (always set, numeric metric) and `'retries_under_budget'` (only when `max_retries` is configured). Prefers `ctx.metrics['retries']` and falls back to scanning the span tree directly so it also works in `OnlineEvaluation` contexts where no metric extraction has run.

**Why a new attribute rather than reusing the OTel Status?** `pydantic_evals.otel.span_tree.SpanNode` doesn't expose Status / StatusMessage / events - so a span-tree-only evaluator can't tell a clean tool span from a retried one even when the tool wrapper already calls `span.record_exception(escaped=True)` and `span.set_status(StatusCode.ERROR)`. A boolean attribute is the smallest change that makes this signal observable without changing `SpanNode`.

## Tests

- 9 new tests in `tests/evals/test_agentic_evaluators.py` covering metric-only, span-tree fallback, budget within/over, and all `SpanTreeRecordingError` paths.
- 1 new test in `tests/evals/test_dataset.py` confirming the metric flows through `Dataset.evaluate` end-to-end.
- Updated 2 inline snapshots in `tests/test_logfire.py` (`test_include_tool_args_span_attributes` for `tool_error=True` with and without `include_content`) to include the new attribute.
- `make lint`, `make format` clean. Pyright clean on the modified files (only one pre-existing `reportPrivateImportUsage` on the unrelated `tenacity` import in `dataset.py`).
- Full `tests/evals/` suite: 444 passed.
- `tests/test_logfire.py` and `tests/test_tools.py`: 193 passed.

## Caveat

The retry attribute is set in `tool_manager._call_tool_span` for both v2 and v3+ instrumentation versions (the same code path handles both). It's set unconditionally when a `ToolRetryError` is caught, regardless of `include_content`, since it carries no user content.

## Checklist

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/_instrumentation.py
+++ b/pydantic_ai_slim/pydantic_ai/_instrumentation.py
@@ -47,6 +47,14 @@ class InstrumentationNames:
     tool_deferral_name_attr: ClassVar[str] = 'pydantic_ai.tool.deferral.name'
     tool_deferral_metadata_attr: ClassVar[str] = 'pydantic_ai.tool.deferral.metadata'
 
+    # Retry span attribute. Set to True on a tool span that ended in a
+    # ToolRetryError - the model will see the retry prompt and try again
+    # (subject to the tool's `max_retries`). Distinct from `tool_deferral_*`
+    # (which marks a tool that was deferred to the user) and from a final
+    # tool failure that exhausts retries (the latter has no dedicated marker
+    # and is only visible via the span's OTel Status).
+    tool_retry_attr: ClassVar[str] = 'pydantic_ai.tool.retry'
+
     @classmethod
     def for_version(cls, version: int) -> Self:
         """Create instrumentation configuration for a specific version.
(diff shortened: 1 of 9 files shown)


pydantic_ai_slim/pydantic_ai/_instrumentation.py
```diff
@@ -47,6 +47,14 @@ class InstrumentationNames:
     tool_deferral_name_attr: ClassVar[str] = 'pydantic_ai.tool.deferral.name'
     tool_deferral_metadata_attr: ClassVar[str] = 'pydantic_ai.tool.deferral.metadata'
 
+    # Retry span attribute. Set to True on a tool span that ended in a
+    # ToolRetryError - the model will see the retry prompt and try again
+    # (subject to the tool's `max_retries`). Distinct from `tool_deferral_*`
+    # (which marks a tool that was deferred to the user) and from a final
+    # tool failure that exhausts retries (the latter has no dedicated marker
+    # and is only visible via the span's OTel Status).
```
Kludex, line 55: Remove this.

PR #5331, opened 2026-05-07
feat: `local=` opt-in for provider-adaptive capability fallback; deprecate auto-fallback

DouweM:
- Closes #<issue>

### Summary

Two-part change that prepares for a v2 default-fallback flip on the provider-adaptive capabilities (`WebSearch`, `WebFetch`, `MCP`):

1. **Adds the v2 opt-in API in v1**, so users have a working target for the migration — `local='duckduckgo'`, `local=True`, `builtin=True`, etc. all resolve silently today. YAML/agent-spec friendly: every opt-in value is a string, bool, or None.
2. **Emits `DeprecationWarning` on the codepaths that go away in v2**, pointing users at the new opt-in shape.

Behavior is unchanged in v1 — this PR is non-breaking. The v2 flip lands in #5333.

### The v2 flips this announces

- `WebSearch()` / `WebFetch()` will default to builtin-only and raise on unsupported models. Users opt into a local fallback explicitly.
- `MCP(url=…)` will default to local-only via FastMCP (no auto-builtin). Users opt into provider-side MCP explicitly.

### Opt-in API (works in v1, will be the only way to silence the warning in v2)

| Capability | Keep current behavior | New v2 default | Strict variant |
|---|---|---|---|
| `WebSearch` | `local='duckduckgo'` (or `local=True`) | `local=False` (builtin-only, raises on unsupported) | — |
| `WebFetch` | `local=True` | `local=False` (builtin-only, raises on unsupported) | — |
| `MCP` | `builtin=True` (builtin preferred, local fallback) | `builtin=False` (local-only via FastMCP — install with `pip install pydantic-ai-slim[mcp]`) | `builtin=True, local=False` (builtin-only, no FastMCP needed) |

### What this PR does

- `BuiltinOrLocalTool`: widens `local` to accept `str | bool | …`; adds a `_resolve_local_strategy(name)` subclass hook for string / `True` resolution. The default-detect path stays for backward compat.
- `WebSearch`: adds `local: Literal['duckduckgo'] | …`. `local='duckduckgo'` and `local=True` both resolve to the DuckDuckGo tool silently. The `_default_local()` auto-detect path emits `DeprecationWarning` pointing at `local='duckduckgo'`.
- `WebFetch`: adds `bool` to `local`. `local=True` resolves to the markdownify-based tool silently. Auto-detect path warns and points at `local=True`.
[...]

PR #5333, opened 2026-05-07
Drop auto-fallback in `WebSearch`/`WebFetch`; `MCP(url=…)` runs locally by default

DouweM:
- Closes #<issue>

### Summary

Completes the v2 default-fallback flip on the provider-adaptive capabilities, building on the deprecation warnings shipped in v1 by #5331.

| Capability | v1 default (deprecated) | v2 default | Opt back in |
|---|---|---|---|
| `WebSearch()` | Native if supported, else auto-DDG when `[duckduckgo]` installed | Native only — raises on unsupported models | `WebSearch(local='duckduckgo')` (or callable / `Tool` / `AbstractToolset`) |
| `WebFetch()` | Native if supported, else auto-markdownify when `[web-fetch]` installed | Native only — raises on unsupported models | `WebFetch(local=True)` (or callable / `Tool` / `AbstractToolset`) |
| `ImageGeneration()` | Native only (already aligned) | Native only — no change | `ImageGeneration(local=…)` or `ImageGeneration(fallback_model=…)` |
| `MCP(url=...)` | Native if supported, else local (when `[mcp]` installed) | Local-only via `MCPToolset` — `[mcp]` extra is now required | `MCP(url=..., native=True)` for native-preferred + local fallback; `MCP(url=..., native=True, local=False)` for native-only |

Users who addressed v1's deprecation warnings need no further action. Anything still on the auto-fallback path will hit a `UserError` — at construction time for `MCP` (missing `[mcp]` extra), or at request time for `WebSearch`/`WebFetch` (model doesn't support the native tool and no `local=` was set) — with the migration string in the message.

### What this PR does

- `capabilities/native_or_local.py`: reject `native=False` without an explicit local fallback at construction (catches the silent-no-op cases the v1 warnings flagged).
- `capabilities/web_search.py` / `web_fetch.py`: drop the auto-detect `_default_local` body. The opt-in `_resolve_local_strategy` paths (`local='duckduckgo'` / `local=True`) keep their install-hint `UserError`s.
- `capabilities/mcp.py`: flip default `native` from `True` to `False`; `_default_local` no longer swallows `ImportError`. Missing `[mcp]` extra now raises `UserError` at construction with an install hint pointing at the extra or the `native=True, local=False` opt-out.
- `docs/capabilities.md`: rewrite the "Provider-adaptive tools" section for the new defaults.
- `tests/test_capabilities.py`: drop the v1 deprecation tests; replace with v2 install-hint and explicit-`local=` coverage.

### Test plan

- [x] Capability tests pass (616 in `tests/test_capabilities.py`; full suite green)
- [x] Pyright + Ruff clean
- [x] Doc examples in `docs/capabilities.md` pass via `tests/test_examples.py`
- [x] CI green across all matrix cells

### Checklist

- [ ] No **breaking changes** in accordance with the [version policy](https://github.com/pydantic/pydantic-ai/blob/main/docs/version-policy.md). *Intentional breaking changes for v2 — the migration was announced via #5331's deprecation warnings.*


pydantic_ai_slim/pydantic_ai/capabilities/web_search.py
```diff
@@ -21,8 +19,9 @@
 class WebSearch(NativeOrLocalTool[AgentDepsT]):
     """Web search capability.
 
-    Uses the model's native web search when available, falling back to a local
-    function tool (DuckDuckGo by default) when it isn't.
+    Uses the model's native web search and raises `UserError` on models that
+    don't support it natively. Pass `local='duckduckgo'` (or `local=True`) to opt into a
```
DouweM, line 23: Mention that DDG requires the optional group

pydantic_ai_slim/pydantic_ai/capabilities/mcp.py
DouweM, line 1: *(Refers to line 60)*

Now that the default is to always run local MCP, it's a bit annoying that you can't pass a `FastMCP` client directly, since it's only native MCP that requires string URLs. So what if we allow everything from `local` here? And raise errors if you set `native=True` at the same time. Unless you do `native=MCPServerTool(...)`, or if you had set this field to a simple URL, and then used `local=FastMCPClient`). And `native=True` on with a string value would error if the string doesn't look like a URL.

pydantic_ai_slim/pydantic_ai/capabilities/mcp.py
```diff
@@ -76,22 +75,6 @@ def __init__(
         allowed_tools: list[str] | None = None,
```
DouweM, line 82: Does this work on the local path?

pydantic_ai_slim/pydantic_ai/capabilities/mcp.py
```diff
@@ -140,37 +123,38 @@ def _native_unique_id(self) -> str:
         return f'mcp_server:{self._resolved_id}'
 
     def _default_local(self) -> Tool[AgentDepsT] | AbstractToolset[AgentDepsT] | None:
-        # The MCP extra may not be installed, in which case the capability still constructs cleanly
-        # — the model just has to support MCP natively (or the user has to opt into `native=True`).
-        try:
-            return self._build_local(self.url)
-        except ImportError:
-            return None
+        return self._build_local(self.url)
```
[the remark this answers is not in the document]
DouweM: Definitely add a test.

Requiring `local=False` for native-only is OK by me, although(if that wouldn't be inconsistent) it is unexpected to need 2 booleans. So it would be nice if just enabling native would be enough when you know you're using a native-supporting model. Could we defer the error until it's actually hit i.e. being used with a non-native model?

pydantic_ai_slim/pydantic_ai/capabilities/mcp.py
```diff
@@ -140,37 +123,38 @@ def _native_unique_id(self) -> str:
...
+            # `MCPToolset` infers SSE vs Streamable HTTP from the URL.
+            from pydantic_ai.mcp import MCPToolset
 
-        return MCPToolset(url, headers=local_headers or None, include_instructions=True)
+            return MCPToolset(url, headers=local_headers or None, include_instructions=True)
+        except ImportError as e:
+            raise UserError(
+                'MCP runs the server locally by default, but the MCP extra is not installed.\n\n'
+                'Either install the MCP extra:\n'
+                '    pip install "pydantic-ai-slim[mcp]"\n'
```
DouweM: On the import error, we typically list the required package name as well? Like when importing the mcp module directly.

pydantic_ai_slim/pydantic_ai/capabilities/native_or_local.py
```diff
@@ -91,6 +91,13 @@ def __post_init__(self) -> None:
         elif callable(self.local) and not isinstance(self.local, (Tool, AbstractToolset)):
             self.local = Tool(self.local)
 
+        # Disallow `native=False` without an explicit local — would produce a silent no-op capability.
+        if self.native is False and self.local is None:  # pyright: ignore[reportUnknownMemberType]
+            raise UserError(
+                f'{type(self).__name__}(native=False) requires an explicit local tool — '
+                'pass `local=...` (e.g. a strategy string, `True`, a callable, or a Tool/AbstractToolset).'
+            )
```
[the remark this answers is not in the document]
DouweM: Was this fixed already? Unsure. Either way we need a test for this scenario.

pydantic_ai_slim/pydantic_ai/capabilities/native_or_local.py
```diff
@@ -91,6 +91,13 @@ def __post_init__(self) -> None:
         elif callable(self.local) and not isinstance(self.local, (Tool, AbstractToolset)):
             self.local = Tool(self.local)
 
+        # Disallow `native=False` without an explicit local — would produce a silent no-op capability.
+        if self.native is False and self.local is None:  # pyright: ignore[reportUnknownMemberType]
+            raise UserError(
+                f'{type(self).__name__}(native=False) requires an explicit local tool — '
+                'pass `local=...` (e.g. a strategy string, `True`, a callable, or a Tool/AbstractToolset).'
```
DouweM: Also, backticks around `Tool` etc

docs/capabilities.md
```diff
@@ -120,50 +120,49 @@ See the dedicated [Hooks](hooks.md) page for the full API: decorator and constru
 
 ### Provider-adaptive tools
 
-[`WebSearch`][pydantic_ai.capabilities.WebSearch], [`WebFetch`][pydantic_ai.capabilities.WebFetch], [`ImageGeneration`][pydantic_ai.capabilities.ImageGeneration], and [`MCP`][pydantic_ai.capabilities.MCP] provide model-agnostic access to common tool types. When the model supports the tool natively (as a [native tool](native-tools.md)), it's used directly. When it doesn't, a local function tool handles it instead — so your agent works across providers without code changes.
+[`WebSearch`][pydantic_ai.capabilities.WebSearch], [`WebFetch`][pydantic_ai.capabilities.WebFetch], [`ImageGeneration`][pydantic_ai.capabilities.ImageGeneration], and [`MCP`][pydantic_ai.capabilities.MCP] provide model-agnostic access to common tool types. Each chooses between a model-side native tool and a client-side local tool based on its constructor arguments — there is no auto-detection of installed extras, so the dispatch path is fully determined by your code.
```
DouweM: This emphasis only makes sense in context of the change. To someone new to the feature, something closer to what we had before works better, explaining that you can configure its fallback behavior using the 2 fields.

docs/capabilities.md
```diff
@@ -120,50 +120,49 @@ See the dedicated [Hooks](hooks.md) page for the full API: decorator and constru
...
-        # Native when supported; falls back to a local MCP transport derived from the URL
-        MCP(url='https://mcp.example.com/api', native=True),
+        # Runs the MCP server locally by default (requires the `mcp` extra)
+        MCP(url='https://mcp.example.com/api'),
     ],
 )
 ```
 
-To force native-only (errors on unsupported models instead of falling back to local):
+The default for `MCP` is intentionally **opposite** the others: MCP carries credentials, so it runs locally unless you opt into the provider's native MCP support. To opt into native-preferred dispatch (with local as fallback for unsupported models):
```
DouweM: We desperately need a subheading for each specific provider-adaptive cap. Feel free to restructure this whole section. Each provider-adaptive cap deserves info/example on how to configure it etc.

docs/capabilities.md
```diff
@@ -120,7 +120,9 @@ See the dedicated [Hooks](hooks.md) page for the full API: decorator and constru
...
+[`WebSearch`][pydantic_ai.capabilities.WebSearch], [`WebFetch`][pydantic_ai.capabilities.WebFetch], [`ImageGeneration`][pydantic_ai.capabilities.ImageGeneration], and [`MCP`][pydantic_ai.capabilities.MCP] each pair a model-side [native tool](native-tools.md) with a client-side local fallback. The `native=` and `local=` kwargs control which side handles the request — when the model supports the native tool, it's used directly; otherwise the local fallback runs the request so your agent works across providers without code changes.
+
+`native=` and `local=` each accept `True`, `False`, or a concrete value (a [native tool](native-tools.md) instance, a strategy string where supported, a callable, a [`Tool`][pydantic_ai.tools.Tool], or an [`AbstractToolset`][pydantic_ai.toolsets.AbstractToolset]). Optional installs needed for the local fallback are opt-in — the capability raises a [`UserError`][pydantic_ai.exceptions.UserError] at construction (with the install hint) when you ask for a local strategy whose extra isn't installed.
```
DouweM: This makes it unclear what native supports vs local.

docs/capabilities.md
```diff
@@ -131,41 +133,111 @@ agent = Agent(
...
-The default for `MCP` is intentionally **opposite** the others: MCP carries credentials, so it runs locally unless you opt into the provider's native MCP support. To opt into native-preferred dispatch (with local as fallback for unsupported models):
+`MCP` defaults the other way from the other three: because MCP carries credentials, it runs locally by default and you opt into native MCP with `native=True`. The other three default to native and you opt into local with `local=`.
 
-```python {title="mcp_native.py" test="skip" lint="skip"}
-MCP(url='https://mcp.example.com/api', native=True)
+Some constraint fields require the native tool (the local fallback can't enforce them) — passing them locks the capability to the native path. If the model doesn't support the native tool, the capability raises a [`UserError`][pydantic_ai.exceptions.UserError]. For example, [`WebSearch`][pydantic_ai.capabilities.WebSearch] domain constraints require native; [`WebFetch`][pydantic_ai.capabilities.WebFetch] enforces them locally.
+
+```python {title="constraints.py" test="skip" lint="skip"}
+# Only search example.com — requires native support
+WebSearch(allowed_domains=['example.com'])
```
DouweM: Why does this require native support? Does DDG not support allowed domains?

`max_uses` may be a better example of something (currently) native-only.

docs/capabilities.md
```diff
@@ -131,41 +133,111 @@ agent = Agent(
...
+# Only fetch example.com — enforced locally when native is unavailable
+WebFetch(allowed_domains=['example.com'], local=True)
 ```
 
-Strict native-only (no local — does not require the `mcp` extra):
+#### WebSearch
 
-```python {title="mcp_native_only.py" test="skip" lint="skip"}
-MCP(url='https://mcp.example.com/api', native=True, local=False)
+[`WebSearch`][pydantic_ai.capabilities.WebSearch] defaults to native-only. Pass `local='duckduckgo'` (or `local=True`) for a [DuckDuckGo](common-tools.md#duckduckgo-search-tool) fallback (requires the `duckduckgo` optional group); pass any callable, [`Tool`][pydantic_ai.tools.Tool], or [`AbstractToolset`][pydantic_ai.toolsets.AbstractToolset] for a custom fallback (e.g. a Tavily / Perplexity / Brave wrapper from [`common_tools`](common-tools.md)).
```
DouweM: Don't list fallbacks we don't actually support.

DouweM: All these sections should link to the native tools doc section on the native instance in question. It's weird that this section doesn't mention `WebSearchTool` or anything else to pass to `native=...`. Same for the other sections.

docs/capabilities.md
```diff
@@ -131,41 +133,111 @@ agent = Agent(
...
+
+#### MCP
+
+[`MCP`][pydantic_ai.capabilities.MCP] is the primary entry point for [MCP](mcp/overview.md) in Pydantic AI. It runs the MCP server locally by default — keeping credentials, hooks, and tracing under your control — and supports both URL-based servers and direct client / toolset / transport inputs.
+
+```python {title="mcp.py" test="skip" lint="skip"}
+from pydantic_ai.capabilities import MCP
+
+# URL-based MCP server, running locally (requires `pydantic-ai-slim[mcp]`)
+MCP(url='https://mcp.example.com/api')
```
DouweM: Not necessary to pass `url=`, right?

docs/capabilities.md
```diff
@@ -131,41 +133,111 @@ agent = Agent(
...
+MCP(local=my_fastmcp_client)
+
+# Native preferred; URL-based local fallback
+MCP(url='https://mcp.example.com/api', native=True)
+
+# Strict native-only (no local — does not require the `mcp` extra)
+MCP(url='https://mcp.example.com/api', native=True, local=False)
+```
+
+For lower-level access — managing the [`MCPToolset`][pydantic_ai.mcp.MCPToolset] lifecycle directly, advanced transport / client configuration, or using MCP servers without going through a capability — see the [MCP documentation](mcp/overview.md). The native side of the capability is backed by [`MCPServerTool`][pydantic_ai.native_tools.MCPServerTool], which can also be used as a [native tool](native-tools.md) directly when the capability's local fallback isn't needed.
```
DouweM: As mentioned above, `MCPServerTool` is relevant also because it can be passed into `native` -- weird we're not covering that. Some of those native instance shave additional args.

docs/capabilities.md
```diff
@@ -138,75 +141,93 @@ agent = Agent(
...
 from pydantic_ai.capabilities import WebSearch
+from pydantic_ai.native_tools import WebSearchTool
 
 # Native-only — raises on models without native web search
 WebSearch()
 
 # Native preferred; DuckDuckGo fallback (needs `pydantic-ai-slim[duckduckgo]`)
 WebSearch(local='duckduckgo')
 
+# Native preferred; explicit native instance for full control
```
DouweM: Native only right?

docs/capabilities.md
```diff
@@ -216,24 +237,33 @@ ImageGeneration(local=my_generator)
...
 MCP(local=my_fastmcp_client)
 
 # Native preferred; URL-based local fallback
-MCP(url='https://mcp.example.com/api', native=True)
+MCP('https://mcp.example.com/api', native=True)
+
+# Native preferred; explicit native instance for full control
+MCP(
+    'https://mcp.example.com/api',
+    native=MCPServerTool(id='my-mcp', url='https://mcp.example.com/api', description='…'),
```
DouweM: Having an explicit local and explicit native, would be interesting as well.

Issue #4041, opened 2026-01-19
Default openai: to Responses API (add explicit chat prefix)


colesmcintosh:
### Question
Would you consider changing the default OpenAI routing so `openai:<model>` uses the Responses API, and provide an explicit way to opt into Chat Completions? Chat Completions are older and (I believe) will be deprecated at some point, so it seems reasonable for model strings to go to Responses by default.

### Suggested options
- Change the default mapping for `openai:<model>` to Responses.
- Add an explicit provider prefix like `openai-chat:<model>` (or similar) for Chat Completions.

### Additional Context

Right now `openai:gpt-5.2` maps to `OpenAIChatModel` (chat), and `openai-responses:gpt-5.2` maps to `OpenAIResponsesModel` (responses)

dsfaccini:
hey @colesmcintosh , agree with you that everybody is moving to the responses format, naturally because Pydantic AI started when completions was the rule this is the behavior and we can't change it without it breaking behavior for (pretty much everyone). So I've flagged this as V2 so we take it into account when we get there (which should be in the coming months).

Thanks!

colesmcintosh:
@dsfaccini Sounds good, would be happy to help work on this when the time comes!

dsfaccini:
David's AICA here: shipped — Phase A `DeprecationWarning` in #5334 (v1.x, merged 2026-05-13), default flip to Responses API in #5469 (merged 2026-05-19 on `v2-main`). `openai-chat:` available since Phase A. Closing — will reopen if anything surfaces in beta.

PR #5334, opened 2026-05-08
Add `openai-chat:` prefix and warn on bare `openai:` which will switch to use the Responses API in v2

dsfaccini:
## Summary

The OpenAI provider is the only big provider exposing two distinct request shapes (Chat Completions and Responses). Today `'openai:gpt-5'` resolves to `OpenAIChatModel`; in v2 it will resolve to `OpenAIResponsesModel`. This PR adds a forward-compatibility on-ramp on the 1.x line:

- Register `'openai-chat:'` as a new explicit prefix routing to `OpenAIChatModel` — adopting it today preserves current behavior across the v2 cut.
- Emit a `DeprecationWarning` from `infer_model` whenever a bare `'openai:'` or `'gateway/openai:'` prefix is resolved, recommending `'openai-chat:'` (current behavior) or `'openai-responses:'` (v2 default).

The actual default flip (bare `'openai:'` → `OpenAIResponsesModel`) lands in the v2-exec PR for this card; this PR only adds the on-ramp.

Per the V2-RULES, no repo-wide `filterwarnings` entry is added. Internal callers in `clai` were migrated to `'openai-chat:gpt-5'` to preserve current behavior; doc and example references keep the bare `'openai:'` prefix on purpose — the warning surfacing in user-facing snippets is the migration signal. The `test_docs_examples` runner ignores the warning per-test so docs continue to validate cleanly.

Tracking: #4041.

## Test plan

- [ ] `tests/v2/test_openai_prefix_deprecation.py` covers the three resolution paths (bare warns + resolves to chat, `'openai-chat:'` silent + chat, `'openai-responses:'` silent + responses)
- [ ] `make format && make lint && make typecheck` clean
- [ ] Targeted tests pass: `tests/test_cli.py`, `tests/test_examples.py`, `tests/models/test_model.py`, `tests/models/test_openai.py`, `tests/test_tool_search.py`, `tests/v2/`

Refs #4041.


docs/models/openai.md
```diff
@@ -131,6 +131,12 @@ agent = Agent('openai-responses:gpt-5.2')
 ...
 ```
 
+!!! note "v2 default change"
+    In Pydantic AI v2, the bare `'openai:'` prefix will resolve to `OpenAIResponsesModel` instead of `OpenAIChatModel`. The 1.x line emits a `DeprecationWarning` whenever bare `'openai:'` is used; pick an explicit prefix to silence the warning and pin behavior:
```
dsfaccini: "will resolve" or "resolves"? @DouweM , we're realistically gonna keep notes like this for a period of time after v2 drops, I assume?

docs/models/openai.md
```diff
@@ -131,6 +131,12 @@ agent = Agent('openai-responses:gpt-5.2')
...
+
+    - `'openai-chat:gpt-5.2'` keeps the current Chat Completions routing.
```
dsfaccini: "current" here is not needed

tests/v2/test_openai_prefix_deprecation.py
```diff
@@ -0,0 +1,52 @@
+"""Card 23: deprecation of the bare ``'openai:'`` prefix default.
```
dsfaccini: single backticks please!

Issue #5150, opened 2026-04-21
Tracking: v2 change inventory


dsfaccini:
Compiled inventory of everything currently on track for v2, drawn from
(a) issues/PRs labeled `v2` / `breaking change`
(b) in-tree `TODO (v2)` markers, and (c) public APIs marked `@deprecated` (per [version policy](https://github.com/pydantic/pydantic-ai/blob/main/docs/version-policy.md): 'Functionality marked as deprecated will not be removed until V2'). 

This is a contributor-compiled tracking proposal, not an official roadmap as of now.

Earliest v2 window per the version policy: **April 2026**.

---

## 1. Open `v2`-labeled issues / PRs

Dependency & defaults
- #4299 — Reconsider which optional dependency groups to include in the non-slim package
- #4041 — Default `openai:` to Responses API (add explicit `chat:` prefix)
- #3792 — Discussion: Should empty string `user_prompt` be validated?

Graph / output
- #1548 — pydantic-graph `End` node should store `output` instead of `data`
- #5052 — (PR) Yield `FunctionToolCallEvent`/`FunctionToolResultEvent` for output tool calls

Rename / API surface (explicitly gated on v2 by maintainer)
- #3609 — Rename 'built-in tools' + consolidate registration under `Agent(tools=[...])`
- #3857 — Rename `stream_responses` → `stream_response`; reconcile `is_final` semantics
- #2942 — `prepare` for `ToolOutput`: generic param order
- #3796 — More easily override specific `ModelProfile` fields

Event / message-model default changes
- #3368 — `FunctionToolCallEvent` re-emitted on resumption of deferred/unapproved tool calls
- #3219 — `capture_run_messages` for incomplete results (needs `ModelResponse.incomplete` + default-behavior change)
- #2586 — More flexible tool exception handling (new `ToolFailed`/`ToolDisabled` + changed default handling)
- #2050 — `history_processors` should see current-run messages by default; orphaned tool-pair handling

Default-behavior changes
- #3791 — Output tools + function tools: execution order when end strategy is exhaustive
- #1590 — `run_stream` slated for deprecation in favor of `iter`

Merged/closed for traceability: #4303 (capabilities), #3913 (http client cache), #1190 (result→output rename), #1799 (generic `EvaluationReport`).

---

### `pydantic_ai_slim`
- `_run_context.py` — change typevar defaults from `None` to `object`
- `_json_schema.py` — remove `simplify_nullable_unions` flag + now-dead branch + unused method
[...]

PR #5263, opened 2026-04-30
Deprecate method-style `result.usage()`, `result.timestamp()`, `stream.get()` in favor of property-style `result.usage`, `result.timestamp`, `stream.response`

dsfaccini:
## Summary

Emits `PydanticAIDeprecationWarning` on method-style `usage()` / `timestamp()` / `get()` access across the result classes ahead of their v2 cut. Twelve sites across `AgentRun`, `AgentRunResult`, `AgentStream`, `StreamedRunResult`, `StreamedRunResultSync`, and `StreamedResponseSync`. Each accessor is now a `@deprecated_callable_property` that returns a wrapper preserving `isinstance` and equality with the underlying type; calling the wrapper (parentheses) emits a `PydanticAIDeprecationWarning` pointing at the property form. In v2 the decorator becomes a plain `@property` (or the method is removed entirely, for `*.get()`).

All internal callers in `pydantic_ai_slim/`, `docs/`, and `tests/` have been migrated to the new API. No repo-wide `filterwarnings` ignore is needed — the only sites still using the old API are the dedicated `tests/v2/` deprecation tests, wrapped in `pytest.warns(PydanticAIDeprecationWarning)`.

## Out of scope (handled separately)

- `Agent(history_processors=[fn, ...])` deprecation — split out 2026-05-12 to its own branch `v2-prep-history-processors` and ships as a separate PR (commit `b2c1d9a6d`).
- `Agent(retries=...)` deprecation — shipped in #5075.
- `prepare_tools` callback returning `None` warning — see #5188 (open, by @dfm88).

## Test plan

- `tests/v2/test_method_to_property_deprecations.py` — pins the contract for each of the twelve sites: property access silent, method-style call emits the deprecation warning, both return the same value. Two round-trip tests cover pydantic serialization of the wrapper types.

## Checklist

- [ ] Any **AI generated code** has been reviewed line-by-line by the human PR author, who stands by it.
- [x] No **breaking changes** in accordance with the [version policy](https://github.com/pydantic/pydantic-ai/blob/main/docs/version-policy.md).
- [x] **PR title** is fit for the [release changelog](https://github.com/pydantic/pydantic-ai/releases).

Relates to #5150.

PR #3217, opened 2025-10-22
Add `http_client` option to `GoogleProvider`, support `api_key` for Vertex AI, and use Pydantic AI's cached httpx client by default

DouweM:
- Fixes https://github.com/pydantic/pydantic-ai/issues/748
- Closes https://github.com/pydantic/pydantic-ai/issues/3167

PR #5336, opened 2026-05-08
Split `GoogleProvider(vertexai=True|False)` into `GoogleProvider` + `GoogleCloudProvider`; rename provider ID `google-gla:` to `google:`, `google-vertex:` to `google-cloud:`; deprecate old names

dsfaccini:
## Summary

Splits the single `GoogleProvider` (which routed to either the Gemini API or Vertex AI based on a `vertexai=` flag) into two distinct providers ahead of the v2 cut:

- `GoogleProvider` — Gemini API only (the consumer-facing public API at `generativelanguage.googleapis.com`).
- `GoogleCloudProvider` — Google Cloud (formerly known as Vertex AI), accepting `project`, `location`, `credentials`.

User-visible deprecations (all emit `PydanticAIDeprecationWarning`):

- `'google-gla:'` prefix → use `'google:'`.
- `'google-vertex:'` prefix → use `'google-cloud:'`.
- `GoogleProvider(vertexai=True, ...)` and any Vertex-only kwarg (`location`, `project`, `credentials`) on `GoogleProvider` → use `GoogleCloudProvider(...)`.
- `GoogleProvider(vertexai=False, ...)` → drop the redundant kwarg (it's the default).

The `GoogleModel` class accepts both old and new `provider_name` values via the `_GEMINI_API_SYSTEMS` / `_GOOGLE_CLOUD_SYSTEMS` frozensets, so historical `ModelMessage` records replay correctly. `tests/v2/test_google_history_replay.py` pins the cross-history-replay contract (V2-RULES rule 4).

## Out of scope (handled separately)

- The gateway path (`gateway/google-vertex:` prefix) is left as-is in 1.x; it's reworked in the v2-exec leg.

## Test plan

- `tests/v2/test_google_cloud_split_deprecation.py` — asserts each deprecated prefix / kwarg combination emits the right `PydanticAIDeprecationWarning` and still routes correctly. Also asserts the new shapes (`'google:'`, `'google-cloud:'`, `GoogleProvider(client=...)`, `GoogleCloudProvider(...)`) do not warn.
- `tests/v2/test_google_history_replay.py` — replays `ModelMessage` history with `provider_name='google-gla'` / `'google-vertex'` (hardcoded old literals) through the new model class and verifies thinking signatures + native-tool parts still surface.

## Checklist

Relates to #5150.


docs/models/google.md
```diff
@@ -3,6 +3,14 @@
 The `GoogleModel` is a model that uses the [`google-genai`](https://pypi.org/project/google-genai/) package under the hood to
 access Google's Gemini models via both the Generative Language API and Vertex AI.
 
+Two providers wrap those endpoints:
+
+- [`GoogleProvider`][pydantic_ai.providers.google.GoogleProvider] — the Gemini Developer API (Google AI Studio), surfaced under the `'google:'` prefix.
+- [`GCPProvider`][pydantic_ai.providers.gcp.GCPProvider] — Vertex AI on Google Cloud Platform, surfaced under the `'gcp:'` prefix.
```
dsfaccini: do we wanna add a link like this one https://cloud.google.com/blog/products/ai-machine-learning/introducing-gemini-enterprise-agent-platform and an addendum like "Vertex AI on GCP _or_ the Gemini Enterprise Agent Platform soon"?

@DouweM

pydantic_ai_slim/pydantic_ai/embeddings/google.py
```diff
@@ -119,9 +120,10 @@ def __init__(
                 for available models.
             provider: The provider to use for authentication and API access. Can be:
 
-                - `'google-gla'` (default): Uses the Gemini API (Google AI Studio)
-                - `'google-vertex'`: Uses Vertex AI
-                - A [`GoogleProvider`][pydantic_ai.providers.google.GoogleProvider] instance
+                - `'google'` (default): Uses the Gemini API (Google AI Studio)
+                - `'gcp'`: Uses Vertex AI
```
dsfaccini: `... or soon https://cloud.google.com/blog/products/ai-machine-learning/introducing-gemini-enterprise-agent-platform`

may be worth adding these now so we don't have to later?

pydantic_ai_slim/pydantic_ai/providers/__init__.py
```diff
@@ -114,8 +115,21 @@ def infer_provider_class(provider: str) -> type[Provider[Any]]:  # noqa: C901
     # Normalize deprecated/alias provider names
     if provider == 'vertexai':
         provider = 'google-vertex'
-    elif provider == 'google':
-        provider = 'google-gla'
+
+    if provider == 'google-gla':
```
dsfaccini, line 128: since we're raising the same (or similar warnings) in multiple places why not do a utility?

tests/v2/test_google_gcp_split_deprecation.py
```diff
@@ -0,0 +1,92 @@
+"""Tests for the Google + GCP provider split (card 08 Phase B).
```
dsfaccini: I like the approach of a new test file but maybe we can create a `tests/v2/` folder instead?

dsfaccini: Claude here: agreed — adopted as the standard. Just landed across the in-flight v2:prep PRs:

- **`tests/AGENTS.md`** got a new "v2 deprecation tests" section saying tests live in `tests/v2/test_<theme>.py` (commit on PR #5263).
- **Root `AGENTS.md`** got a "v2 work in progress" section with the broader conventions (card-driven design, `v2:prep` / `v2:exec` PR shapes, branch / title / label rules, the `tests/v2/` placement) so colleagues' agents see them without needing the contributor toolkit. Same PR.
- All 5 active v2:prep PRs (#5263, #5296, #5334, #5335, this one) now have the test file moved into `tests/v2/` with the `v2_` prefix dropped from the filename (the folder establishes the context).
- The `/v2-review` skill in the contributor toolkit was updated to check for `tests/v2/` placement.

For this PR specifically: `tests/test_v2_google_gcp_split_deprecation.py` → `tests/v2/test_google_gcp_split_deprecation.py` (commit `6316aa1`). Bumped `from .conftest` → `from ..conftest` for the new depth.

docs/models/google.md
```diff
@@ -3,6 +3,14 @@
 The `GoogleModel` is a model that uses the [`google-genai`](https://pypi.org/project/google-genai/) package under the hood to
 access Google's Gemini models via both the Generative Language API and Vertex AI.
```
dsfaccini: both "generative language api" and "vertex" are now outdated terms

gla -> gemini developer API
vertex -> google cloud (gemini enterprise agent platform)

read more on the douwe:david call from Tue 12 11 am

lets make sure we replace all instances of GLA and vertex

tests/test_capabilities.py
```diff
@@ -786,6 +786,38 @@ def test_model_json_schema_with_capabilities():
...
+                        'google:gemini-2.5-flash-lite-preview-09-2025',
+                        'google:gemini-2.5-flash-lite',
+                        'google:gemini-2.5-flash-preview-09-2025',
+                        'google:gemini-2.5-flash',
+                        'google:gemini-2.5-pro',
+                        'google:gemini-3-flash-preview',
+                        'google:gemini-3-pro-image-preview',
+                        'google:gemini-3-pro-preview',
+                        'google:gemini-3.1-flash-image-preview',
+                        'google:gemini-3.1-flash-lite-preview',
```
dsfaccini, line 853: this one's getting deprecated this month, just fyi

tests/test_cli.py
```diff
@@ -162,8 +162,7 @@ def test_list_models(capfd: CaptureFixture[str]):
         'anthropic',
         'bedrock',
         'cerebras',
-        'google-vertex',
-        'google-gla',
+        'google',
```
dsfaccini, line 167: we're lacking a google-cloud here, no?

tests/test_embeddings.py
```diff
@@ -1210,34 +1210,50 @@ def embedder(self, gemini_api_key: str) -> Embedder:
             GoogleEmbeddingModel('gemini-embedding-2-preview', provider=GoogleProvider(api_key=gemini_api_key))
         )
 
+    @pytest.mark.filterwarnings(
```
dsfaccini: why do we have the warning filter if we updated the shorthands?

tests/v2/test_google_history_replay.py
```diff
@@ -0,0 +1,111 @@
...
+from __future__ import annotations as _annotations
+
+import base64
+
+import pytest
+
+from ..conftest import try_import
+
+with try_import() as imports_successful:
+    from pydantic_ai import ModelResponse, NativeToolCallPart, NativeToolReturnPart, TextPart, ThinkingPart
```
dsfaccini: these imports and the ones form native tools don't need to be under a try import

tests/test_cli.py
dsfaccini: Claude here: good catch — `'google-cloud'` was missing. Without it, `google-cloud:...` model entries get incidentally caught by `'google'` (string-prefix match: `'google-cloud:...'.startswith('google')` is True), which works but is conceptually wrong. Added `'google-cloud'` to the providers tuple at line 165.

tests/test_embeddings.py
dsfaccini: Claude here: these tests `test_infer_model_gla` / `test_infer_model_vertex` are legacy-coverage — they intentionally exercise the deprecated `google-gla:` / `google-vertex:` prefixes to confirm they still resolve in 1.x. Without the filter, `pyproject.toml`'s `filterwarnings = ["error"]` turns the deprecation into a hard failure. Two options if you'd prefer a different shape:

1. Replace `@pytest.mark.filterwarnings('ignore:...')` with `pytest.warns(PydanticAIDeprecationWarning, match=...)` — this both silences AND asserts the deprecation fires (stronger contract per V2-RULES rule 1).
2. Delete the deprecated-prefix tests entirely — `tests/v2/test_google_cloud_split_deprecation.py` already asserts the warning + correct routing at the `infer_provider` level; the embedding shim test is redundant.

Lean toward (1) for explicit coverage parity. Want me to apply it?

tests/v2/test_google_history_replay.py
dsfaccini: Claude here: fixed in next commit. `from pydantic_ai import ...` and `from pydantic_ai.native_tools import CodeExecutionTool` don't depend on `google-genai`, so they no longer sit inside `try_import`. Only the `google.genai.client.Client` + `pydantic_ai.models.google` / `providers.google` / `providers.google_cloud` imports need the gating.

pydantic_ai_slim/pydantic_ai/providers/alibaba.py
```diff
@@ -27,6 +27,10 @@ class AlibabaProvider(Provider[AsyncOpenAI]):
 
     @property
     def name(self) -> str:
+        # Returned value flows into ModelMessage.provider_name on every part.
+        # Thinking-tag detection and built-in-tool detection check this value when
+        # the model class loads history, so silently renaming breaks replay of any
+        # message history captured against the old name.
         return 'alibaba'
```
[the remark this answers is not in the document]
DouweM: Heh, I do agree with this as stated -- I feel like having it on the root `Provider.name` _should_ be enough, and having it in AGENTS.md or something? The comment _is_ true and worth for the agent (or more importantly the reviewer) being aware of. But relative to all the other things we want the agent to know, this one is not _so_ important as to deserve 30 repetitions -- if other more important rules don't get the same treatment. So, back to the root / agents.md it is!

pydantic_ai_slim/pydantic_ai/providers/__init__.py
```diff
@@ -105,17 +111,45 @@ def __repr__(self) -> str:
...
+        # Suppress the standalone `'google-vertex:'` rename warning when coming from a gateway
+        # prefix — the caller already saw the `'gateway/google-vertex:'` warning above, and the
+        # internal mapping `gateway/google-cloud -> google-vertex` is intentional (see rule 17).
+        if not came_from_gateway:
+            warnings.warn(
+                "The 'google-vertex:' prefix is deprecated and will be removed in v2.0. Use 'google-cloud:' instead.",
+                PydanticAIDeprecationWarning,
+                stacklevel=2,
+            )
+        provider = 'google-cloud'
```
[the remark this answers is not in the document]
DouweM: Likely OK as-is because this will be removed very soon.

pydantic_ai_slim/pydantic_ai/providers/google_cloud.py
```diff
@@ -0,0 +1,69 @@
...
+            warnings.simplefilter('ignore', PydanticAIDeprecationWarning)
+            super().__init__(  # pyright: ignore[reportCallIssue]
+                vertexai=True,
+                credentials=credentials,
+                project=project,
+                location=location,
+                client=client,
+                http_client=http_client,
+                base_url=base_url,
+            )
```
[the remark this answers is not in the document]
DouweM: It is useful for them to be have the same root class, but we could clean this up by having a new abstract `BaseGoogleProvider` that has a private `_init_client` that take all arguments, that `GoogleProvider` and `GoogleCloudProvider` then call. That ABC does then not implement its own `__init__`, presumably.

For now the `GoogleProvider` subclass would continue to take all args, but a bunch will be dropped in v2.

pydantic_ai_slim/pydantic_ai/models/google.py
```diff
@@ -181,7 +182,26 @@
...
+Includes the pre-v2 `'google-vertex'` name so message history captured against the
+old provider name still replays correctly.
+"""
+
+GEMINI_API_SYSTEMS: frozenset[str] = frozenset({'google', 'google-gla'})
+"""Provider names that route to the Gemini API.
+
+Includes the pre-v2 `'google-gla'` name so message history captured against the
+old provider name still replays correctly.
+"""
```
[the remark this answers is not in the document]
DouweM: - Yes to rename
- Yes to make private, unless someone asks

pydantic_ai_slim/pydantic_ai/models/google.py
```diff
@@ -399,7 +448,8 @@ def __init__(
         self,
         model_name: GoogleModelName,
         *,
-        provider: Literal['google-gla', 'google-vertex', 'gateway'] | Provider[Client] = 'google-gla',
+        provider: Literal['google', 'google-cloud', 'google-gla', 'google-vertex', 'gateway']
+        | Provider[Client] = 'google',
```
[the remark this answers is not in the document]
DouweM: Yeah I had the same thought, better to remove them from the union if it's not used for validation (it isn't).

docs/models/google.md
```diff
@@ -96,68 +104,71 @@ from google.oauth2 import service_account
...
+import os
+
 from pydantic_ai import Agent
 from pydantic_ai.models.google import GoogleModel
-from pydantic_ai.providers.google import GoogleProvider
+from pydantic_ai.providers.google_cloud import GoogleCloudProvider
 
-provider = GoogleProvider(vertexai=True, api_key='your-api-key')
+os.environ['GOOGLE_API_KEY'] = 'your-api-key'
+provider = GoogleCloudProvider()
```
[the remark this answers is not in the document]
DouweM: Hmm isn't there such a thing as a Cloud/Vertex AI API Key? The original example suggests there was! Worth checking the blame etc.

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -1453,7 +1497,7 @@ def infer_model(  # noqa: C901
         from .openai import OpenAIResponsesModel
 
         return OpenAIResponsesModel(model_name, provider=provider)
-    elif model_kind in ('google', 'google-gla', 'google-vertex'):
+    elif model_kind in ('google', 'google-gla', 'google-vertex', 'google-cloud'):
```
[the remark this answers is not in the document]
DouweM: Yep!

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -182,6 +182,18 @@
...
+        'gateway/google-cloud:gemini-2.5-flash-lite',
+        'gateway/google-cloud:gemini-2.5-flash',
+        'gateway/google-cloud:gemini-2.5-pro',
+        'gateway/google-cloud:gemini-3-flash-preview',
+        'gateway/google-cloud:gemini-3-pro-image-preview',
+        'gateway/google-cloud:gemini-3.1-flash-image-preview',
+        'gateway/google-cloud:gemini-3.1-flash-lite-preview',
+        'gateway/google-cloud:gemini-3.1-pro-preview',
+        # Legacy `gateway/google-vertex:` literals kept for `ModelMessage` deserialization
+        # (rule 23). User-facing prefix is `gateway/google-cloud:`; emits a deprecation warning.
```
DouweM: Doesn't seem relevant -- the deserialization always accepts `str` right? These unions exist _explicitly_ for IDE autocomplete and nothing else.

tests/test_embeddings.py
dsfaccini: David's AICA here: Done in ecb46fcb0 — went with option 1 (`pytest.warns(PydanticAIDeprecationWarning, match=...)`) for stronger contract.

pydantic_ai_slim/pydantic_ai/providers/alibaba.py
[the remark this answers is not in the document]
dsfaccini: David's AICA here: Done in ecb46fcb0 — reverted the duplicated comment from all 32 concrete provider implementations; folded the rationale into the `Provider.name` docstring on the abstract base. The 4-line block is now in one place.

pydantic_ai_slim/pydantic_ai/providers/google_cloud.py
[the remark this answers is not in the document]
dsfaccini: David's AICA here: Done in ecb46fcb0 — extracted `BaseGoogleProvider` with private `_init_client(...)`. Both `GoogleProvider` and `GoogleCloudProvider` now call it directly; `catch_warnings()` is gone.

pydantic_ai_slim/pydantic_ai/models/google.py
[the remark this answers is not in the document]
dsfaccini: David's AICA here: Done in ecb46fcb0 — renamed to `_GOOGLE_CLOUD_PROVIDER_NAMES` / `_GEMINI_API_PROVIDER_NAMES` and made private.

pydantic_ai_slim/pydantic_ai/models/google.py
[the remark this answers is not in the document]
dsfaccini: David's AICA here: Done in ecb46fcb0 — `Literal` now only includes `'google'`, `'google-cloud'`, `'gateway'`. Deprecated string names still work at runtime via `infer_provider` and emit the deprecation.

docs/models/google.md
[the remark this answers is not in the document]
dsfaccini: David's AICA here: Researched — Vertex AI Express Mode is a real path ([API keys docs](https://cloud.google.com/vertex-ai/generative-ai/docs/start/api-keys?usertype=expressmode)); `GoogleProvider(vertexai=True, api_key=...)` already worked (PR #3217). Done in ecb46fcb0: added `api_key=` to `GoogleCloudProvider.__init__` (forwards through `_init_client`), restored explicit-kwarg example in `docs/models/google.md`.

pydantic_ai_slim/pydantic_ai/embeddings/google.py
```diff
@@ -108,7 +109,7 @@ def __init__(
         self,
         model_name: GoogleEmbeddingModelName,
         *,
-        provider: Literal['google-gla', 'google-vertex'] | Provider[Client] = 'google-gla',
+        provider: Literal['google', 'google-cloud', 'google-gla', 'google-vertex'] | Provider[Client] = 'google',
```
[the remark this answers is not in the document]
dsfaccini: David's AICA here: Done in ecb46fcb0 — `Literal` now only `'google'`, `'google-cloud'`.

tests/test_embeddings.py
```diff
@@ -1210,34 +1210,50 @@ def embedder(self, gemini_api_key: str) -> Embedder:
...
+        'ignore:.*google-gla.*prefix is deprecated:pydantic_ai._warnings.PydanticAIDeprecationWarning'
+    )
```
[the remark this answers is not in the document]
dsfaccini: David's AICA here: Done in ecb46fcb0 — replaced both `@pytest.mark.filterwarnings('ignore')` with `pytest.warns(PydanticAIDeprecationWarning, match=...)` inside the test body.

pydantic_ai_slim/pydantic_ai/models/__init__.py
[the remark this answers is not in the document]
dsfaccini: David's AICA here: Done in ecb46fcb0 — `vertexai:` shim now maps directly to `'google-cloud'` and the warning text points at `'google-cloud'`. Single warning instead of the two-hop chain.

pydantic_ai_slim/pydantic_ai/providers/google.py
```diff
@@ -24,16 +24,12 @@
     ) from _import_error
 
 
-class GoogleProvider(Provider[Client]):
-    """Provider for Google."""
+class BaseGoogleProvider(Provider[Client]):
```
DouweM: This should be an ABC so that people cannot use it directly!

pydantic_ai_slim/pydantic_ai/providers/google.py
```diff
@@ -41,6 +43,88 @@ def client(self) -> Client:
...
+        api_client._async_httpx_client = http_client  # pyright: ignore[reportPrivateUsage]
+        api_client._http_options.httpx_async_client = http_client  # pyright: ignore[reportPrivateUsage]
+
+
+class GoogleProvider(BaseGoogleProvider):
+    """Provider for Google."""
+
+    @property
+    def name(self) -> str:
+        return 'google-cloud' if self._client._api_client.vertexai else 'google'  # pyright: ignore[reportPrivateUsage]
```
DouweM: I don't think we need this anymore now that we have 2 subclasses?

pydantic_ai_slim/pydantic_ai/providers/gateway.py
```diff
@@ -197,12 +200,15 @@ def _with_http_client(provider: Provider[Any]) -> Provider[Any]:
...
 
-        return _with_http_client(
-            GoogleProvider(vertexai=True, api_key=api_key, base_url=base_url, http_client=http_client)
-        )
+        with warnings.catch_warnings():
+            # Internal forward; the gateway-google-cloud path is reworked in v2.
+            warnings.simplefilter('ignore', PydanticAIDeprecationWarning)
+            return _with_http_client(
+                GoogleProvider(vertexai=True, api_key=api_key, base_url=base_url, http_client=http_client)
+            )
```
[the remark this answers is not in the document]
DouweM: > This is a deliberate tech-debt decision — GoogleCloudProvider doesn't accept api_key + base_url + http_client in the same way needed for gateway routing, so the old path is used internally.

That's a strong indication that `GoogleCloudProvider` should support those fields!

pydantic_ai_slim/pydantic_ai/providers/google.py
```diff
@@ -41,6 +43,88 @@ def client(self) -> Client:
...
+        )
+
+        if not vertexai:
+            if api_key is None:
+                raise UserError(
+                    'Set the `GOOGLE_API_KEY` environment variable or pass it via `GoogleProvider(api_key=...)`'
+                    ' to use the Gemini API.'
+                )
+            self._client = Client(vertexai=False, api_key=api_key, http_options=http_options)
+            return
```
DouweM: vertex-ai specific code can be moved into `GoogleCloudProvider`, right?

pydantic_ai_slim/pydantic_ai/providers/__init__.py
```diff
@@ -105,17 +113,45 @@ def __repr__(self) -> str:
...
+        # `gateway/google-cloud`. Internal Gateway wire value still maps to `google-vertex`
+        # until the Gateway team renames their side; `normalize_gateway_provider` handles that.
+        if provider == 'gateway/google-vertex':
+            warnings.warn(
+                "The 'gateway/google-vertex:' prefix is deprecated and will be removed in v2.0. "
+                "Use 'gateway/google-cloud:' instead.",
+                PydanticAIDeprecationWarning,
+                stacklevel=2,
+            )
         provider = normalize_gateway_provider(provider)
```
DouweM, line 122: Should the warning above maybe be inside `normalize_gateway_provider`?

pydantic_ai_slim/pydantic_ai/providers/__init__.py
```diff
@@ -242,12 +282,40 @@ def infer_provider(provider: str) -> Provider[Any]:
...
+        # renames their side (see rule 17 — decouples our cadence from theirs).
+        if provider == 'gateway/google-vertex':
+            warnings.warn(
+                "The 'gateway/google-vertex:' prefix is deprecated and will be removed in v2.0. "
+                "Use 'gateway/google-cloud:' instead.",
+                PydanticAIDeprecationWarning,
+                stacklevel=2,
+            )
         upstream_provider = provider.removeprefix('gateway/')
         return gateway_provider(upstream_provider)
```
DouweM, line 279: This warning may also want to live in `gateway_provider`

pydantic_ai_slim/pydantic_ai/models/__init__.py
DouweM: This doesn't actually seem resolved? I was expecting all the old prefixed versions to be gone, as well as this comment.

dsfaccini: David's AICA here: Done in 859e384c8 — dropped all 42 deprecated-prefix Literal entries (`google-gla:`, `google-vertex:`, `gateway/google-vertex:`) plus the stale 'kept for deserialization' comment. Pydantic deserialization accepts `str` regardless of the Literal, so message-history round-trip is unaffected. Also removed `'google-vertex'` from `gateway.ModelProvider` Literal so IDE autocomplete only surfaces canonical names.

pydantic_ai_slim/pydantic_ai/providers/google.py
dsfaccini: David's AICA here: Done in 859e384c8 — `BaseGoogleProvider(Provider[Client], ABC)` with `@abstractmethod name` so instantiating it directly raises `TypeError`.

pydantic_ai_slim/pydantic_ai/providers/google.py
dsfaccini: David's AICA here: Done in 859e384c8 — `GoogleProvider.name` now returns `'google'` unconditionally. The legacy `vertexai` branch is gone; deprecated Cloud-args usage delegates to `GoogleCloudProvider` (which has its own `name = 'google-cloud'`).

pydantic_ai_slim/pydantic_ai/providers/gateway.py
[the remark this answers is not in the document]
dsfaccini: David's AICA here: Done in 859e384c8 — the gateway Google path now constructs `GoogleCloudProvider(api_key=..., base_url=..., http_client=...)` directly. `GoogleCloudProvider` already accepted those three fields, so the `warnings.catch_warnings()` suppression block is gone.

pydantic_ai_slim/pydantic_ai/providers/google.py
dsfaccini: David's AICA here: Done in 859e384c8 — refactored `BaseGoogleProvider._init_client` into a thin shared `_build_http_options` helper. The Gemini-API `Client(vertexai=False, ...)` construction lives in `GoogleProvider.__init__`; the Cloud-mode `Client(vertexai=True, ...)` construction (with ADC vs Express Mode and region defaulting) lives in `GoogleCloudProvider.__init__`. The deprecated `GoogleProvider(vertexai=True, ...)` path forwards to `GoogleCloudProvider` and steals its lifecycle attrs.

pydantic_ai_slim/pydantic_ai/providers/__init__.py
dsfaccini: David's AICA here: Done in 859e384c8 — the `gateway/google-vertex` deprecation warning now fires inside `normalize_gateway_provider` (`pydantic_ai_slim/pydantic_ai/providers/gateway.py`), which both `infer_provider_class` and `gateway_provider` call. Duplicate warning blocks at the call sites in `providers/__init__.py` are removed.

pydantic_ai_slim/pydantic_ai/providers/__init__.py
dsfaccini: David's AICA here: Done in 859e384c8 — same as the sibling thread; the warning lives in `normalize_gateway_provider` and `gateway_provider` calls it via the default-route lookup at line 156.

PR #5339, opened 2026-05-08
Run tools in emission order with `graceful` default; `sequential=True` is now a per-tool barrier; function-tool `ModelRetry` suppresses the output result

### LINKED ISSUE #3791: Output tools and function tools are not executed in parallel tool call order when end strategy is exhaustive

DouweM:
- Closes #3791
- Fixes the silent `ModelRetry`-discard bug reported by @MSameerAbbas in https://github.com/pydantic/pydantic-ai/pull/4085#issuecomment-3826378612.

## Summary

Reworks how a model response's tool calls are executed when the response mixes ordinary function tools with an [output tool](https://ai.pydantic.dev/output/#tool-output) (the call that produces the run's final result). `end_strategy` now controls the *shape* of that execution, the long-standing "an output result silently discards a function-tool retry" bug is fixed, and `sequential=True` becomes a sensible **per-tool barrier** instead of a batch-wide serial switch. `ToolOutput(sequential=True)` is newly supported too, so an output tool can itself act as a barrier (e.g. "run all the function tools, *then* validate the output") — a new option this rework opens up.

Most agents need no change. The one scenario that changes by default: when a model calls function tool(s) in the same response as an output tool **and that output tool succeeds**, the function tools now run instead of being skipped — details below.

## Behavior changes

`end_strategy` is fundamentally about **output**. When a model emits an [output tool](https://ai.pydantic.dev/output/#tool-output) call — the call that ends the run — in the *same* response as other tool calls, what should happen to the rest? Most users never think about this, because most responses don't mix an output tool with other tools; but when one does, `end_strategy` decides.

### 1. Default `end_strategy` changes from `'early'` to `'graceful'`

Under the old `'early'` default, the run ended as soon as an output tool succeeded, and any function tools the model requested *in that same response* were **skipped**. Under the new `'graceful'` default they **run**, so their side effects happen and their results are available to the model if the run continues.

This only affects the output-**success** case. When *every* output tool fails, function tools run and the run continues under **all** strategies — that was already true on `main` and is unchanged here.

| Strategy | Output tools | Function tools — output succeeded | Function tools — every output failed |
|---|---|---|---|
| `'graceful'` (default) | Run in emission order; first success is the final result, later output tools skipped (so their side effects don't fire twice) | **Run**, in parallel where possible, interleaved in emission order | Run; run continues |
| `'early'` | Run in emission order; the **run ends** at the first success | **Skipped** | Run; run continues |
| `'exhaustive'` | **All run, in parallel**; first valid one by emission order wins, the rest are marked "ran but not selected" | Run, in parallel | Run; run continues |

**Why run function tools when every output tool fails?** Because then there's no result to end on. The output failures go back to the model as retries, and since the model *also* asked for those function tools, we run them and continue as if it were an ordinary step, so the model can react to both next round. The only thing unique to `'early'` is the output-**success** case: it exits the instant a result exists, whereas `'graceful'`/`'exhaustive'` keep going.

`run_stream()` sits at the same end of the spectrum as `'early'`: it commits the first matching output the instant it streams, so the result is locked in regardless of `end_strategy` (and the retry-wins behavior below can't revoke it).

### 2. Retry-wins invariant (`'graceful'` + `'exhaustive'`)

If an actual function tool raises `ModelRetry` (or its args fail validation) in the same response as a successful output tool, the output result is suppressed and the retry is surfaced to the model on the next round — instead of being silently discarded. The suppressed output's `ToolReturnPart` is rewritten so the model gets a clear signal. Does **not** apply under `'early'` (function tools don't run once an output succeeds), under `run_stream()` (the streamed output is already committed), or for retries from unknown/hallucinated tools (only real function tools count).

**Why treat the silent discard as a bug?** By emitting a function tool call in the *same* response as an output tool, the model is asserting "do this" and "I'm done" at once — so it presumably expected that action to take effect, and the output it produced may assume it did. A `ModelRetry` is the model's *own* self-correction signal; if we silently keep the output, the run ends on a result that may be based on an action that never happened, with no chance for the model to reconcile. (The original report — [#4085 comment](https://github.com/pydantic/pydantic-ai/pull/4085#issuecomment-3826378612) — was an editing agent whose `ModelRetry` on an edit tool *was* the correction mechanism; a silent discard simply loses the correction.) The opposite reading is defensible — by calling the output tool alongside the others, the model arguably signaled the function result wasn't a blocker for ending — but we judge it safer to tell the model "the thing you thought happened didn't" than to end on a possibly-stale result. We don't suppress under `'early'` because there the model opted into "first output wins, ignore the rest".

A stronger variant we considered and deliberately did **not** build: *fail-fast* — if a function tool emitted before the output tool fails, skip the output tool (and its side effects) entirely, rather than running it and suppressing the result after the fact. This treats co-emitted calls as a dependency chain rather than parallel calls, so it's really a parallel-*execution* concern (orthogonal to `end_strategy`, adjacent to the `sequential=True` barrier) rather than an end strategy. It wasn't requested in #3791 and adds real complexity (partial-execution + barrier interactions), so it's out of scope here; retry-wins covers the common case where the output tool has no side effects of its own.

### 3. `sequential=True` is now a per-tool **barrier**, not a batch-wide serial switch

Previously a single tool with `sequential=True` forced the *entire* response's tools to run serially. Now a `sequential=True` tool runs **alone** — tools emitted before it complete first, it runs by itself, tools emitted after it start once it finishes — while every other tool still parallelizes around it. `ToolOutput(sequential=True)` is newly supported to make an *output* tool a barrier under `'exhaustive'` (e.g. "run all function tools, then validate output").

Migration for the old "any sequential → all serial" behavior: wrap the run in `agent.parallel_tool_call_execution_mode('sequential')` or set `parallel_tool_calls=False` on the model settings.

### 4. Tools execute in emission order

Under `'graceful'` and `'exhaustive'`, tools run in the order the model emitted them: function tools the model emitted before an output tool finish before that output tool is handled, addressing the original #3791 request. `'exhaustive'` runs everything in parallel by default; ordering is enforced only by `sequential=True` barriers. (`'early'` is output-first and unaffected by this.)

## Test plan

- [x] `TestMultipleToolCalls` (and streaming counterparts) updated for the new per-strategy behavior
- [x] New tests: retry-wins under graceful/exhaustive, retry-wins not under early, unknown-tool retry doesn't suppress output, graceful emission-order interleaving, `sequential=True` barrier parallelism, `ToolOutput(sequential=True)` under exhaustive, early function-tool fallback when all outputs fail
- [x] Partial-capture (`state='interrupted'`) still works across function-batch and exhaustive paths
- [x] Docs: `docs/output.md` parallel-output-tool-calls section, `docs/changelog.md` v2 entry, `EndStrategy` / `ToolDefinition.sequential` / `ToolOutput.sequential` docstrings

### Checklist

    - Deliberate v2 default behavior change; the previous "output-wins" default was a historical artefact. Documented in the changelog with migration guidance.


pydantic_ai_slim/pydantic_ai/tool_manager.py
```diff
@@ -143,20 +143,24 @@ def tool_defs(self) -> list[ToolDefinition]:
 
         return [tool.tool_def for tool in self.tools.values()]
 
-    def get_parallel_execution_mode(self, calls: list[ToolCallPart]) -> ParallelExecutionMode:
-        """Get the effective parallel execution mode for a list of tool calls.
+    def get_parallel_execution_mode(self) -> ParallelExecutionMode:
```
DouweM, line 146: Hmm this is a breaking change on a public method, right? We should be sure to include that in the release notes.

pydantic_ai_slim/pydantic_ai/tool_manager.py
```diff
@@ -143,20 +143,24 @@ def tool_defs(self) -> list[ToolDefinition]:
...
-        mode = _parallel_execution_mode_ctx_var.get()
+    def is_sequential(self, call: ToolCallPart) -> bool:
+        """Whether a tool call must run as a barrier (`sequential=True`), executing alone.
 
-        return mode
+        Tools emitted before a barrier complete first; the barrier runs by itself; tools emitted
+        after it start only once it finishes. Other tools parallelize around it.
+        """
+        tool_def = self.get_tool_def(call.tool_name)
+        return tool_def is not None and tool_def.sequential
```
DouweM, line 163: Is `sequential` still the best name for this barrier concept? We _could_ change it, though ideally that'd be in v1 with a deprecation warning.

Also make sure that with this change to `sequential`'s meaning (and working between function and output tools), the docs that mention this feature have been updated.

pydantic_ai_slim/pydantic_ai/_tool_execution.py
```diff
@@ -0,0 +1,1067 @@
...
+if TYPE_CHECKING:
+    from ._agent_graph import GraphAgentDeps, GraphAgentState
+
+DepsT = TypeVar('DepsT')
+
+
+def _make_output_status_part(
+    call: _messages.ToolCallPart,
+    content: str,
+    output_parts: list[_messages.ModelRequestPart],
```
DouweM: Always require kwargs for anything other than the first/implied arg.

pydantic_ai_slim/pydantic_ai/_tool_execution.py
```diff
@@ -0,0 +1,1067 @@
...
+    """Synthesize and append a status `ToolReturnPart` for an output tool call (success or skip).
+
+    Sites that retry use the part returned by validation/execution directly, not a synthesized one.
+    """
+    part = _messages.ToolReturnPart(
+        tool_name=call.tool_name,
+        content=content,
+        tool_call_id=call.tool_call_id,
+    )
+    output_parts.append(part)
```
DouweM: This method call adding itself to `output_parts` seems unexpected.

pydantic_ai_slim/pydantic_ai/_tool_execution.py
```diff
@@ -0,0 +1,1067 @@
...
+    return segments
+
+
+async def _validate_function_calls(
+    tool_manager: ToolManager[DepsT],
+    calls: list[_messages.ToolCallPart],
+    calls_to_run_results: dict[str, DeferredToolResult],
+    tool_call_metadata: dict[str, dict[str, Any]] | None,
+    ctx: GraphRunContext[GraphAgentState, GraphAgentDeps[DepsT, NodeRunEndT]],
+    validated_calls: dict[str, ValidatedToolCall[DepsT]],
```
DouweM: Functions like this really need to require kwargs

pydantic_ai_slim/pydantic_ai/_tool_execution.py
```diff
@@ -0,0 +1,1067 @@
...
+                    validate_call = dataclasses.replace(call, args=deferred_result.override_args)
+                metadata = tool_call_metadata.get(call.tool_call_id) if tool_call_metadata else None
+                validated = await tool_manager.validate_tool_call(validate_call, approved=True, metadata=metadata)
+            else:
+                validated = await tool_manager.validate_tool_call(call)
+        except exceptions.UnexpectedModelBehavior:
+            ctx.state.check_incomplete_tool_call()
+            yield _messages.FunctionToolCallEvent(call, args_valid=False)
+            raise
+        validated_calls[call.tool_call_id] = validated
```
DouweM: The output arguments and box arguments make me think we need to have a class here, not just a bunch of free functions that thread stuff through.

Either a new one, or integrated in `ToolManager`. That integration makes sense to me, but it may be too much refactoring to bite off for now?

pydantic_ai_slim/pydantic_ai/_tool_execution.py
```diff
@@ -0,0 +1,1067 @@
...
+        yield _messages.FunctionToolCallEvent(call, args_valid=validated.args_valid)
+
+
+async def _process_exhaustive(  # noqa: C901
+    *,
+    tool_manager: ToolManager[DepsT],
+    tool_calls: list[_messages.ToolCallPart],
+    call_kinds: list[ToolKind | Literal['unknown']],
+    output_indices: list[int],
```
DouweM: There's so much going on here! So many function arguments. There's gotta be a better data model / design / architecture of this we can come up with, with or without ToolManager. Use a subagent to look into it.

pydantic_ai_slim/pydantic_ai/_tool_execution.py
```diff
@@ -0,0 +1,1067 @@
...
+        final_result=final_result,
+    )
+    async for event in processor.run():
+        yield event
+    if processor.final_result:
+        output_final_result.append(processor.final_result)
+
+
+@dataclasses.dataclass
+class _ToolCallProcessor(Generic[DepsT, NodeRunEndT]):
```
DouweM: Ah now we've got a class :) I'm just confused why we still have the huge free functions above then? Are they used in another path?

pydantic_ai_slim/pydantic_ai/_tool_execution.py
```diff
@@ -0,0 +1,1067 @@
...
+            elif self.is_executable_function(i):
+                pending_functions.append(call)
+        async for event in flush_pending():
+            yield event
+        self.ctx.state.output_retries_used += self.output_retries_increment[0]
+
+    async def _run_exhaustive(self) -> AsyncIterator[_messages.HandleResponseEvent]:
+        """`'exhaustive'`: run every tool in parallel, segmented only by `sequential=True` barriers."""
+        state = _ExhaustiveState[NodeRunEndT](final_result=self.final_result)
+        async for event in _process_exhaustive(
```
DouweM: Would it make sense to have subclasses for the strategies?

pydantic_ai_slim/pydantic_ai/_tool_execution.py
```diff
@@ -0,0 +1,1067 @@
...
+            self.retry_wins_triggered and self.final_result is not None and not self.final_result_was_set_externally
+        ):
+            return
+        # The winning output always has a 'Final result processed.' part here, so the loop
+        # always breaks; the no-match fall-through can't happen.
+        for idx, part in enumerate(self.output_parts):  # pragma: no branch
+            if (
+                isinstance(part, _messages.ToolReturnPart)
+                and part.tool_call_id == self.final_result.tool_call_id
+                and part.content == 'Final result processed.'
```
DouweM: This is shady, no magic strings, this at the very least needs to be a constant -- and I would _much_ prefer to not do any type of this "patching" surgery at all. There's no way we can find these parts less indirectly?

pydantic_ai_slim/pydantic_ai/_tool_execution.py
```diff
@@ -0,0 +1,1067 @@
...
+            ):
+                self.output_parts[idx] = dataclasses.replace(
+                    part, content='Output not used as the final result - addressing tool retries from this round first.'
```
DouweM: Let's put all these strings in constants at the top.

pydantic_ai_slim/pydantic_ai/_tool_execution.py
```diff
@@ -0,0 +1,1067 @@
...
+            ]
+
+            handler_validated_calls: dict[str, ValidatedToolCall[DepsT]] = {}
+            for call in resolved_calls:
+                handler_result = handler_tool_call_results[call.tool_call_id]
+                if not isinstance(handler_result, ToolApproved):
+                    continue
+                validate_call = call
+                if handler_result.override_args is not None:
+                    validate_call = dataclasses.replace(call, args=handler_result.override_args)
```
DouweM: We have this duplicated in places, could we fix that?

Are there other duplications to think of extracting, so that behavior in different places doesn't get out of sync?

pydantic_ai_slim/pydantic_ai/_tool_execution.py
```diff
@@ -0,0 +1,1067 @@
...
+            if not self.ctx.deps.output_schema.allows_deferred_tools:
+                raise exceptions.UserError(
+                    'A deferred tool call was present, but `DeferredToolRequests` is not among output types. '
+                    'To resolve this, add `DeferredToolRequests` to the list of output types for this agent, '
+                    'or use a `HandleDeferredToolCalls` capability to handle deferred tool calls inline.'
+                )
+            self.final_result = result.FinalResult(cast(NodeRunEndT, deferred_tool_requests), None, None)
+
+
+async def _call_tools(  # noqa: C901
```
DouweM: Unsure why we still need massive free functions when we're trying to use clssses.

All the indexed dicts bothers me a lot as well.

Also, kwargs!!!

pydantic_ai_slim/pydantic_ai/_tool_execution.py
```diff
@@ -0,0 +1,1067 @@
...
+    output_parts: list[_messages.ModelRequestPart],
+    output_deferred_calls: dict[Literal['external', 'unapproved'], list[_messages.ToolCallPart]],
+    output_deferred_metadata: dict[str, dict[str, Any]],
+) -> AsyncIterator[_messages.HandleResponseEvent]:
+    tool_parts_by_index: dict[int, _messages.ModelRequestPart] = {}
+    user_parts_by_index: dict[int, _messages.UserPromptPart] = {}
+    deferred_calls_by_index: dict[int, Literal['external', 'unapproved']] = {}
+    deferred_metadata_by_index: dict[int, dict[str, Any] | None] = {}
+
+    async def handle_call_or_result(
```
DouweM: The nested functions sadden me

pydantic_ai_slim/pydantic_ai/_tool_execution.py
```diff
@@ -0,0 +1,1067 @@
...
+        tool_calls, deferred_calls_by_index, deferred_metadata_by_index, output_deferred_calls, output_deferred_metadata
+    )
+
+
+def _populate_deferred_calls(
+    tool_calls: list[_messages.ToolCallPart],
+    deferred_calls_by_index: dict[int, Literal['external', 'unapproved']],
+    deferred_metadata_by_index: dict[int, dict[str, Any] | None],
+    output_deferred_calls: dict[Literal['external', 'unapproved'], list[_messages.ToolCallPart]],
+    output_deferred_metadata: dict[str, dict[str, Any]],
```
DouweM: Output args like this sadden me. You'll want a subagent to do a more thorough refactor.

pydantic_ai_slim/pydantic_ai/_tool_execution.py
```diff
@@ -0,0 +1,1067 @@
...
+    if isinstance(tool_call, ValidatedToolCall):
+        validated = tool_call
+        call = tool_call.call
+    else:
+        validated = None
+        call = tool_call
+
+    tool_result: Any
+    try:
+        if tool_call_result is None or isinstance(tool_call_result, ToolApproved):
```
DouweM: Is this duplicated in different places now?

docs/output.md
```diff
@@ -359,24 +359,41 @@ _(This example is complete, it can be run "as is")_
...
+
+When *every* output tool fails, function tools run and the run continues under all three strategies: there is no result to end on, so the output failures go back to the model as retries and the function tools the model also asked for are run, letting it react to both on the next round. `run_stream()` behaves like `'early'` in this respect: it commits the first matching output the instant it streams, so that result is locked in regardless of `end_strategy`.
+
+#### Retrying after a tool failure
+
+Under `'graceful'` and `'exhaustive'`, if a function tool raises [`ModelRetry`][pydantic_ai.exceptions.ModelRetry] (or its arguments fail validation) in the same response as a successful output tool, the output result is **not** used as the final result. Instead, the retry is sent back to the model so it can correct the problem, since the output may have been based on the failed tool call. This does not apply under `'early'` (function tools don't run once an output succeeds).
+
+#### Controlling parallelism
```
DouweM: The tools docs speak of parallelism / sequential as well. Make sure the sections are consistent / cross linked.

docs/output.md
```diff
@@ -359,24 +359,41 @@ _(This example is complete, it can be run "as is")_
...
+Choose `'early'` to end the run the instant an output tool succeeds — function tools requested in the same response are then skipped entirely. This is the fastest option when you never need those function tools to run once you have a result.
+
+Choose `'exhaustive'` to run every tool, including additional output tools whose results won't be used. This gives the model full visibility that each tool ran, at the cost of executing output-tool side effects that are ultimately discarded.
+
+When *every* output tool fails, function tools run and the run continues under all three strategies: there is no result to end on, so the output failures go back to the model as retries and the function tools the model also asked for are run, letting it react to both on the next round. `run_stream()` behaves like `'early'` in this respect: it commits the first matching output the instant it streams, so that result is locked in regardless of `end_strategy`.
+
+#### Retrying after a tool failure
```
DouweM: This is one heading level up from "parallel output tool calls", but refers to "graceful" etc as if they were just introduced -- which they wouldn't be to someone who jumped to this section from the ToC. So fix it. Is this meant to be a standalone section? Or a subsection?

docs/output.md
```diff
@@ -359,24 +359,41 @@ _(This example is complete, it can be run "as is")_
...
+Tools run concurrently by default. To stop a specific tool from overlapping with others, mark it `sequential=True` — it then acts as a barrier: tools the model emitted before it finish first, it runs alone, and tools emitted after it start only once it finishes.
+
+```python {title="sequential_tool.py" lint="skip" test="skip"}
+@agent.tool_plain(sequential=True)
+def write_to_database(record: str) -> str: ...
+```
+
+Under `'exhaustive'`, where output tools also run in parallel, you can make an output tool a barrier with [`ToolOutput(sequential=True)`][pydantic_ai.output.ToolOutput] — useful when you want all of a response's function tools to finish before the output tool runs. To run an entire run's tools serially regardless of which tools were called, wrap the run in [`agent.parallel_tool_call_execution_mode('sequential')`][pydantic_ai.agent.AbstractAgent.parallel_tool_call_execution_mode] or set `parallel_tool_calls=False` on the [model settings][pydantic_ai.settings.ModelSettings].
 
 !!! warning "Priority of output and deferred tools in streaming methods"
```
DouweM: This ended up a few sections down from where we had it, under a subheading(?). Does it make sense for this to be moved up to the intro?

docs/deferred-tools.md
```diff
@@ -234,19 +234,19 @@ print(result.all_messages())
...
                 tool_call_id='delete_file',
                 timestamp=datetime.datetime(...),
                 outcome='denied',
             ),
+            ToolReturnPart(
+                tool_name='update_file',
+                content="File '.env' updated: ''",
+                tool_call_id='update_file_dotenv',
+                timestamp=datetime.datetime(...),
+            ),
```
DouweM, line 249: Is this reordering significant?

Issue #5324, opened 2026-05-06
`AnthropicProvider` doesn't normalize model names for Bedrock/Vertex/Azure clients (incl. tool-search default strategy)


DouweM:
> _Issue created by Claude on behalf of @DouweM, based on real-world reports from @kclisp on #5143._

## Summary

`AnthropicProvider.model_profile(model_name)` passes `model_name` straight to `anthropic_model_profile()` without normalization. When the underlying client is `AsyncAnthropicBedrock` (or the equivalent Vertex/Azure variant), the model name carries a transport-specific prefix and version suffix like `us.anthropic.claude-haiku-4-5-20251001-v1:0`. None of the `model_name.startswith(...)` checks in `anthropic_model_profile()` match that, so the resulting profile silently flips many capability flags off — `supports_tool_search`, `supports_json_schema_output`, `anthropic_supports_adaptive_thinking`, `anthropic_supports_effort`, etc.

`BedrockProvider.model_profile()` already does the right normalization (strips geo prefix via `remove_bedrock_geo_prefix`, splits on `.` to extract provider and base name, strips `-vN(:M)?`), but that path isn't reached when users go through `AnthropicProvider` with a Bedrock client.

## Repro

```python
from pydantic_ai.providers.anthropic import AnthropicProvider

profile = AnthropicProvider.model_profile('us.anthropic.claude-haiku-4-5-20251001-v1:0')
# profile.supports_tool_search          -> False (should be True for Haiku 4.5)
# profile.supports_json_schema_output   -> False (should be True)
# profile.anthropic_supports_adaptive_thinking -> False
```

## Default tool-search strategy on Bedrock

Bedrock-Invoke (Anthropic SDK shape, Bedrock transport) doesn't support BM25 today — it 400s with `BM25 tool search is not supported on Bedrock. Use tool_search_tool_regex instead.` Pydantic AI's current default is `bm25` (set in `AnthropicModel._add_builtin_tools` when `ToolSearch(strategy=None)` is used).

The fix wants to live alongside the model-name normalization: once `AnthropicProvider` knows it's talking to a Bedrock client, the natural place to route is the same place. Concretely, this likely means an `anthropic_default_tool_search_strategy: Literal['bm25', 'regex']` field on `AnthropicModelProfile` (defaulting to `'bm25'`), set to `'regex'` for the Bedrock-client codepath, and consumed in `_add_builtin_tools` instead of the hardcoded BM25 fallback.

[...]

PR #5143, opened 2026-04-18
feat: native tool search on Anthropic and OpenAI, with custom search strategies on any provider

DouweM:
- Closes #4167
- Closes #4566

### Summary

Adds tool search to Pydantic AI: agents with large toolsets can mark tools as `defer_loading=True` to keep them out of the model's initial context until they're discovered by keyword. The PR delivers two related features:

1. **Native provider tool search** on supporting models. Deferred tools are sent with `defer_loading` on the wire and the provider manages their visibility — no `search_tools` function tool round-trips, no extra latency.
2. **Custom search strategies**, independent of provider support. Pass any `(ctx, queries, tools) -> names` callable to `ToolSearch(strategy=...)` and it runs anywhere: locally as a `search_tools` function tool on providers without native search, or through the provider's client-executed mode (Anthropic, OpenAI Responses) where it still benefits from `defer_loading` on the wire.

Both surface through the new auto-injected [`ToolSearch`][pydantic_ai.capabilities.ToolSearch] capability, so the default behaviour "just works": mark a tool or toolset deferred and the agent picks the best path for the active model.

### Provider support for native tool search

| Provider | Models | Notes |
|---|---|---|
| Anthropic | Sonnet 4.5+, Opus 4.5+, Haiku 4.5+ | Server-executed `bm25` (default) or `regex`. Custom callables ride the `tool_reference` block path. |
| OpenAI Responses | GPT-5.4+ | Server-executed `tool_search` by default. Custom callables ride the `execution='client'` path. Named `'bm25'`/`'regex'` strategies are Anthropic-only. |
| All others | — | Local fallback: an auto-injected `search_tools` function tool (keyword-overlap by default, or the user's callable). |

### Public API

- `defer_loading=True` on [`Tool`][pydantic_ai.tools.Tool], [`@agent.tool`][pydantic_ai.agent.Agent.tool], and [`@agent.tool_plain`][pydantic_ai.agent.Agent.tool_plain]; [`AbstractToolset.defer_loading()`][pydantic_ai.toolsets.AbstractToolset.defer_loading] for whole toolsets (incl. MCP, FastMCP).
- [`ToolSearch`][pydantic_ai.capabilities.ToolSearch] capability with:
  - `strategy=None` (default): use the best provider-native option, fall back locally.
  - `strategy='keywords'`: force the local keyword-overlap algorithm regardless of provider.
[...]

PR #5344, opened 2026-05-08
fix: strip Bedrock provider segment for AnthropicModel profile lookup

kclisp:
When `AnthropicModel` is constructed with `AsyncAnthropicBedrock`, the model id carries a Bedrock provider segment (`us.anthropic.claude-...-v1:0`, `anthropic.claude-...`) that the bare `claude-...` `startswith` checks in `anthropic_model_profile` miss, silently flipping capability flags like `supports_json_schema_output` to their defaults.

Strip the segment for the profile lookup only; the full id still goes on the wire via `self._model_name`.

Refs #5324 (point 4).

### Checklist

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/anthropic.py
+++ b/pydantic_ai_slim/pydantic_ai/models/anthropic.py
@@ -179,6 +179,7 @@ except ImportError as _import_error:
 
 _NON_AUTOMATIC_CACHING_CLIENTS = (AsyncAnthropicBedrock, AsyncAnthropicVertex)
 _FAST_MODE_UNSUPPORTED_CLIENTS = (AsyncAnthropicBedrock, AsyncAnthropicFoundry, AsyncAnthropicVertex)
+_BEDROCK_PREFIX_CLIENTS = (AsyncAnthropicBedrock,)
 
 _ANTHROPIC_SAMPLING_PARAMS = ('temperature', 'top_p', 'top_k')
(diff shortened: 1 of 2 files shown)


DouweM:
@kclisp Any particular reason this doesn't take use the same logic that `BedrockProvider` uses?

https://github.com/pydantic/pydantic-ai/blob/1e415a44b03ff2d923ad1c6c20eef1a137986d02/pydantic_ai_slim/pydantic_ai/providers/bedrock.py#L153-L172

https://github.com/pydantic/pydantic-ai/issues/5324 also specifically mentions `remove_bedrock_geo_prefix`.

Was the agent fed the full body of that issue? :)

kclisp:
@DouweM Yes, it would be a larger refactor than I would otherwise want to do.

`providers/bedrock.py` attempts to import `boto3`, which `models/anthropic.py` does not appear to currently rely on. If we wanted to either use `remove_bedrock_geo_prefix` or split out a helper from `BedrockProvider.model_profile`, it (and `BEDROCK_GEO_PREFIXES`) would have to live in `_utils.py` or some other shared file.

(Note also that `BedrockProvider.model_profile` doesn't use `remove_bedrock_geo_prefix` - better could be to make that so, but then that's more refactoring, if you'd prefer)

DouweM:
@kclisp No worries, I'll take it from here once I get #5143 out (shortly).

DouweM:
Superseded by https://github.com/pydantic/pydantic-ai/pull/5395

PR #5346, opened 2026-05-08
fix: handle boolean JSON Schema nodes in function_signature walker

P0rth0s:
## Summary

#4771 reported that the `JsonSchemaTransformer` walker crashed on valid boolean JSON Schema nodes. #4989 fixed it in `_json_schema.py`. There's a **second walker** in `function_signature.py` (`_schema_to_type_expr` and `_build_params_from_schema`) with the same bug — `'$ref' in schema` and `'default' in prop_schema` both `TypeError` when the schema is a bool.

Per the JSON Schema spec, `True` and `False` are valid schemas anywhere a schema may appear. They commonly surface as `additionalProperties: true|false`. Encountered in practice while wiring the upstream `signoz/signoz-mcp-server:v0.1.0` into an Agent — `signoz_create_dashboard`'s input schema contains `additionalProperties: true` and crashes `Agent.iter()`'s tool-listing pass with:

```
TypeError: argument of type 'bool' is not iterable
File "pydantic_ai/function_signature.py", line 547, in _schema_to_type_expr
    if '$ref' in schema:
```

## Changes

1. **`_schema_to_type_expr`** accepts `dict | bool` and short-circuits to `_ANY` for both `True` and `False`. Representing "Never" precisely isn't expressible here and offers no benefit to the model; `Any` is the practical fallback.

2. **`_build_params_from_schema`** normalizes a boolean prop schema to `{}` before the rest of the loop indexes into it. Without this, `properties: {x: True}` still crashes on `'default' in prop_schema` after `_schema_to_type_expr` returns cleanly.

## Tests

Three regression tests covering all bool positions:
- `additionalProperties: True` (matches anything)
- `additionalProperties: False` (matches nothing)
- bare `properties.x: True`

## Verified locally

- `make` (format / lint / typecheck-pyright / tests) — all green
- `pytest tests/test_function_signature.py tests/test_tools.py tests/test_toolsets.py tests/test_mcp.py` — 377/377 pass

## Suggestion for follow-up (not in this PR)

This is the second time the same bug shape has hit a JSON Schema walker — #4771 was the first, in `JsonSchemaTransformer`. The two walkers do different work (`JsonSchemaTransformer` produces a transformed schema; `_schema_to_type_expr` produces a Python type-AST), but they share **structural traversal**: both must recurse into every JSON Schema construct, including the boolean form, and they handle that recursion independently. A small shared helper — e.g. `_normalize_schema_node(node) -> dict` returning `{}` for `True` and an unsatisfiable shape for `False` — called at every recursion entry would make the bool case impossible to forget in any future walker. Happy to open a separate issue to discuss if that's of interest.

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/function_signature.py
+++ b/pydantic_ai_slim/pydantic_ai/function_signature.py
@@ -504,7 +504,8 @@ def _build_params_from_schema(
     required_params: dict[str, FunctionParam] = {}
     optional_params: dict[str, FunctionParam] = {}
 
-    for prop_name, prop_schema in properties.items():
+    for prop_name, prop_schema_raw in properties.items():
+        prop_schema: dict[str, Any] = {} if isinstance(prop_schema_raw, bool) else prop_schema_raw
         type_expr = _schema_to_type_expr(prop_schema, defs, referenced_types, tool_name, prop_name)
 
         if 'default' in prop_schema:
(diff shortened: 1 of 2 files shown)


hramezani:
Superseded by https://github.com/pydantic/pydantic-ai/pull/6314

PR #5349, opened 2026-05-09
docs: note only the most recent ModelRequest's instructions are used

### LINKED ISSUE #3623: _get_instructions only looks at the last two messages
### Initial Checks

### Description

In `_get_instructions` (https://github.com/pydantic/pydantic-ai/blob/88f6e92ce863995eeceeed5becf69434c5eb5fa6/pydantic_ai_slim/pydantic_ai/models/__init__.py#L724) only the last two ModelRequest messages are considered for instructions. This is non-intuitive to me.

I'm not sure if this is a bug or intentional. If it is intentional, then the documentation (https://ai.pydantic.dev/direct/) should probably have a caveat that you must include instructions in the last ModelRequest if you want it to have any effect.

Edit: I also may be completely wrong in my usage of `instructions`. In which case feel free to close.

### Example Code

```Python

```

```Text
Python 3.13.7
pydantic-ai v1.26.0
```

navalprakhar:
- Closes #3623

### Checklist


adtyavrdhn:
A more accurate wording would be the most recent `ModelRequest` with instructions I believe

navalprakhar:
> A more accurate wording would be the most recent `ModelRequest` with instructions I believe

Looked through `_get_instruction_parts` again. "With instructions" works when the most recent `ModelRequest` is a mock, but if it has regular parts like `UserPromptPart` and no instructions, earlier ones aren't used either. It returns `None`.

My original wording also misses the mock case. Would this work instead?

> Set `instructions` on the most recent `ModelRequest`. Instructions on earlier `ModelRequest`s are ignored.

Let me know if I'm missing something.

navalprakhar:
> Set `instructions` on the most recent `ModelRequest`. Instructions on earlier `ModelRequest`s are ignored.

Pushed the change with this wording. Happy to adjust if you'd rather go another direction.

PR #5350, opened 2026-05-09
docs(hooks): add pre-flight input validation example

eeee2345:
This PR adds a docs section to docs/hooks.md showing how to use the existing before_model_request hook for pre-flight input validation. The new section sits between Tool hook filtering and Timeouts and demonstrates raising SkipModelRequest with a synthetic ModelResponse when a regex pattern matches incoming user content.

The example covers four common prompt injection categories: instruction overrides, role hijacks, system-prompt extraction, and delimiter injection. Patterns are inline so the example is self-contained, and the doc note explicitly invites users to swap in catalogs from open detection standards such as Agent Threat Rules at https://github.com/Agent-Threat-Rule/agent-threat-rules (Apache-2.0).

The example is marked test=skip lint=skip so it does not affect CI. The existing before_model_request and SkipModelRequest APIs are used as-is, no new types or capabilities are introduced. The doc note positions pattern detection as a cheap first filter complementary to slower model-based moderation, in line with the existing Hooks vs AbstractCapability guidance.

The diff is 44 lines in a single file. The repository the example draws from has been adopted in production at Cisco AI Defense and Microsoft agent-governance-toolkit.

### DIFF
--- a/docs/hooks.md
+++ b/docs/hooks.md
@@ -280,6 +280,50 @@ print(call_log)
 
 The `tools` parameter accepts a sequence of tool names. The hook only fires for matching tools — other tool calls pass through unaffected.
 
+## Pre-flight input validation
+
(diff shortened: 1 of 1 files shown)


adtyavrdhn:
In the current form feels out of place with the rest of the docs which don't go into a very specific example of what you can do with it.

eeee2345:
@adtyavrdhn — fair point, thanks for taking a look. Two ways I can re-shape this to fit the rest of the hooks docs:

1. **Shrink to a 5-line code snippet**: drop the multi-pattern regex catalog, keep just `before_model_request` raising `SkipModelRequest` when a single example regex matches. That matches the terser style of the existing `tools` parameter example above this section.

2. **Drop the prose framing**: keep just a short heading "Pre-flight validation" + the canonical 5-liner, no mention of pattern catalogs or ATR. Pattern source attribution moves to a single trailing footnote.

Happy to push either rev — pick whichever shape lines up with your doc conventions, or close this and I'll respin against an open issue you point me at.

PR #5355, opened 2026-05-10
fix(xai): preserve tool result IDs and require xai-sdk 1.6.0

### LINKED ISSUE #5328: xai: `tool_result` messages sent without `tool_call_id`
### Initial Checks

### Description

When pydantic-ai serializes tool results to send back to xAI, it drops the `tool_call_id` linking each result to the assistant's tool call. So calls go out tagged with IDs, but the matching results come back unlabeled.

The relevant code is in `pydantic_ai_slim/pydantic_ai/models/xai.py`, inside `_map_request_parts`:

```python
for part in tool_results:
    if isinstance(part, ToolReturnPart):
        text, files = part.model_response_str_and_user_content()
        xai_messages.append(tool_result(text))
        file_content.extend(files)
    else:
        xai_messages.append(tool_result(part.model_response()))
```

`part.tool_call_id` is already in scope (it's used a few lines above to sort `tool_results` against `pending_tool_call_ids`), and `xai_sdk.chat.tool_result` has accepted a `tool_call_id` kwarg since v1.5. The fix is just threading it through:

```python
xai_messages.append(tool_result(text, tool_call_id=part.tool_call_id))
...
xai_messages.append(tool_result(part.model_response(), tool_call_id=part.tool_call_id))
```

Why it matters: for single-tool turns, xAI usually correlates the result with the call positionally, so things mostly work. The break shows up with **parallel tool calls** — which Grok 4 / 4.x emit regularly at higher reasoning effort. With no IDs, the model has to guess which return corresponds to which call, and reasoning quality on the next turn degrades. The xAI SDK docs explicitly recommend passing `tool_call_id` in this case.

Repro: build a `ModelResponse` with two parallel `ToolCallPart`s and a follow-up `ModelRequest` with two `ToolReturnPart`s, run it through `XaiModel._map_messages`, and inspect the resulting `chat_pb2.Message` list. The `ROLE_TOOL` messages come out with empty `tool_call_id` even though the upstream `ToolReturnPart`s had them set.

Quick context on history that might be relevant: `tool_call_id` was added to the `xai_sdk.chat.tool_result` helper on Jan 19 2026 (commit `ef72733`). It looks like pydantic-ai's xAI integration predates that change and was never updated to use the new parameter — the proto Message type itself has always had a `tool_call_id` field, in both v5 and v6.

Confirmed on `main` and against `pydantic-ai-slim` 1.88.0 with `xai-sdk` 1.12.1.

```Python

```

### Logfire Trace

_No response_

- Python:
- Pydantic AI:
- LLM provider SDK:

FU-max-boop:
Closes #5328.

This updates the xAI model adapter to preserve `tool_call_id` when mapping `ToolReturnPart` and tool-bound `RetryPromptPart` into `xai_sdk.chat.tool_result(...)` messages. This keeps parallel tool-call results associated with the correct xAI tool calls.

The xAI optional dependency is bumped from `xai-sdk>=1.5.0` to `xai-sdk>=1.6.0` because `1.6.0` is the first SDK version where `tool_result(...)` accepts the `tool_call_id` keyword argument.

Tests were updated to assert reordered xAI tool-result history retains the expected `tool_call_id` values.


pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -314,6 +314,18 @@
...
         'xai:grok-4-fast-reasoning-latest',
         'xai:grok-4-latest',
+        'xai:grok-4.20',
+        'xai:grok-4.20-0309',
+        'xai:grok-4.20-0309-non-reasoning',
+        'xai:grok-4.20-0309-reasoning',
+        'xai:grok-4.20-multi-agent',
+        'xai:grok-4.20-multi-agent-0309',
+        'xai:grok-4.20-multi-agent-latest',
+        'xai:grok-4.20-non-reasoning',
```
adtyavrdhn: If these are missing they should be added separately and not as a part of this one.

pydantic_ai_slim/pyproject.toml
```diff
@@ -73,7 +73,7 @@ cohere = ["cohere>=5.20.6; platform_system != 'Emscripten'"]
 vertexai = ["google-auth>=2.36.0", "requests>=2.32.2"]
 google = ["google-genai>=1.70.0"]
 anthropic = ["anthropic>=0.97.0"]
-xai = ["xai-sdk>=1.5.0"]
+xai = ["xai-sdk>=1.12.2"]
```
adtyavrdhn: Is this because 1.5.0 does not support tool call id?

adtyavrdhn: Okay verified and this is true, do we need the minimum to be `1.12.2` or will `1.6.0` work too? Any other reason why me might need this version?

FU-max-boop:
Thanks, both points make sense.

I removed the unrelated xAI model-name/schema updates from this PR so it stays focused on preserving tool result IDs.

I also checked the xai-sdk versions: `1.5.0` has `tool_result(result)` without `tool_call_id`, while `1.6.0` adds `tool_result(result, tool_call_id: Optional[str] = None)`. So I lowered the minimum from `1.12.2` to `1.6.0`.

adtyavrdhn (approved):
Thanks!

PR #5364, opened 2026-05-11
Include interrupted model request/response in `capture_run_messages` result if exception was raised during response streaming or tool call execution

### LINKED ISSUE #3219: Support for `capture_run_messages` to store incomplete results

adtyavrdhn:
- Closes #3219

## Summary

Extends [`capture_run_messages`][pydantic_ai.capture_run_messages] / `all_messages()` to retain partial state on two interruption paths that previously dropped messages:

- **Streaming exception path**: in `ModelRequestNode.stream`, the `stream_error` branch now appends the partial `ModelResponse` (built from the in-flight stream) with `state='interrupted'`. Covers the original #3219 scenario — a frontend `cancel`/abort during `run_stream_events` propagating as `CancelledError` or any downstream exception during event iteration.
- **Tool-execution interruption path**: in `CallToolsNode._handle_tool_calls`, exceptions from `process_tool_calls` are intercepted to append a partial `ModelRequest` carrying the tool returns that completed so far, marked `state='interrupted'`, then re-raised. Supporting change in `_call_tools`: moved the `output_parts.extend(tool_parts_by_index)` populate step into a `finally` so partial returns surface to the outer capture (previously only ran on success).

`PR #5313` already shipped this behavior for explicit `StreamedRunResult.cancel()`; this PR symmetrically covers exception / external-cancellation paths and the tool-execution loop.

## API additions

- New `ModelRequestState: TypeAlias = Literal['complete', 'interrupted']` (mirrors `ModelResponseState`, deliberately a separate alias so the two can diverge if needed without a back-compat shim).
- New `ModelRequest.state: ModelRequestState = 'complete'` field. Additive, default keeps existing reprs/serializations unchanged (thanks to `dataclasses_no_defaults_repr`).
- Both exported from `pydantic_ai`.

## Implementation notes

- In the streaming path the partial is appended directly (not via `_append_response`) to skip the usage-limit check — raising `UsageLimitExceeded` mid-error-handling would mask `stream_error`. Usage tokens are still incremented.
- `state='interrupted'` is always forced on the partial response even if `_handler_response` happened to land before cancel — the run as a whole did not complete normally, and `finish_reason` still reflects model-level finish state for callers who need it.
- The partial `ModelRequest` intentionally omits `user_prompt`: that prompt was meant to accompany the *completed* tool returns going to the next model request and was never sent to the model.

## Docs

- Docstrings on the new `ModelRequest.state` field and `capture_run_messages`.
- New "Inspecting interrupted runs" subsection in `docs/message-history.md` showing the read pattern.

## Test plan

- [x] New tests in `tests/test_streaming.py`:
  - `test_run_stream_consumer_exception_captures_partial_response`
  - `test_run_stream_external_cancellation_captures_partial_response`
  - `test_tool_exception_captures_partial_request`
  - `test_tool_execution_cancellation_captures_partial_request`
- [x] Updated 3 JSON-serialization snapshots in `tests/test_agent.py` to include the new `state` field.
- [x] `make lint`, narrow `pyright` on changed files (0 errors), full `pyright` baseline unchanged (627 pre-existing errors in `test_prefect.py` / `test_temporal.py`).
- [x] `pytest tests/` excluding model/provider/durable-exec/evals: 3211 passed.

### Checklist


pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -121,6 +121,9 @@
 ModelResponseState: TypeAlias = Literal['complete', 'interrupted']
 """Lifecycle state of a model response."""
 
+ModelRequestState: TypeAlias = Literal['complete', 'interrupted']
+"""Lifecycle state of a model request."""
```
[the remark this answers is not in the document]
DouweM: Separate makes sense

docs/message-history.md
```diff
@@ -262,6 +262,42 @@ assert forked.conversation_id != result1.conversation_id
...
+            await agent.run('Do the thing', message_history=...)
+        except Exception:
+            pass
+
+    for message in messages:
+        if isinstance(message, ModelResponse) and message.state == 'interrupted':
+            print(f'partial response: {message.parts}')
+        elif isinstance(message, ModelRequest) and message.state == 'interrupted':
+            print(f'partial tool returns: {message.parts}')
+```
```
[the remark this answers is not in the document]
DouweM: Yep example should work

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -121,6 +121,9 @@
 ModelResponseState: TypeAlias = Literal['complete', 'interrupted']
```
DouweM: Needs update after https://github.com/pydantic/pydantic-ai/pull/5455?

docs/message-history.md
```diff
@@ -262,6 +262,42 @@ assert forked.conversation_id != result1.conversation_id
 
 The [UI adapters](ui/overview.md) auto-populate `conversation_id` from the protocol's own thread/chat ID, so frontends using these protocols get correlation for free.
 
+### Inspecting interrupted runs
+
+When a run is cut short — by an exception while streaming, by an exception inside a tool, or by an
+external cancellation — Pydantic AI still captures the partial state so you can persist or replay it.
+
+The partial messages appear inside [`capture_run_messages`][pydantic_ai.capture_run_messages] (and
```
DouweM: We should update the existing doc where `capture_run_messages` is mentioned: agent.md

docs/message-history.md
```diff
@@ -262,6 +262,42 @@ assert forked.conversation_id != result1.conversation_id
...
+in `result.all_messages()` for `run_stream`) with `state='interrupted'`:
+
+- A partial [`ModelResponse`][pydantic_ai.messages.ModelResponse] holds the parts streamed before the
```
DouweM: - Worth mentioning what happens to half-finished tool call parts
- This does beg for special handling of interrupted model requests _and_ responses in `CallToolsNode`, not just https://github.com/pydantic/pydantic-ai/issues/5507, but also what I wrote in https://github.com/pydantic/pydantic-ai/issues/3219#issuecomment-3437824503 about "incomplete ModelRequest"

tests/test_streaming.py
```diff
@@ -4764,4 +4764,196 @@ async def consume() -> None:
...
+    def good_tool(x: int) -> int:
+        return x * 10
+
+    @agent.tool_plain(sequential=True)
+    def bad_tool(x: int) -> int:
+        raise RuntimeError('tool-failure')
+
+    with capture_run_messages() as messages:
+        with pytest.raises(RuntimeError, match='tool-failure'):
+            await agent.run('Hello')
```
DouweM: This test doesn't use streaming, so this seems like the wrong place for it :) Where is `capture_run_messages` currently tested?

pydantic_ai_slim/pydantic_ai/messages.py
adtyavrdhn: Yes I need to rebase

pydantic_ai_slim/pydantic_ai/_agent_graph.py
```diff
@@ -711,6 +711,23 @@ async def _streaming_handler(
...
             if stream_error is not None:
                 await _cancel_task(wrap_task)
+                # Capture the partial response so `capture_run_messages` and `all_messages()`
+                # include what was streamed before the interruption. State is forced to
+                # 'interrupted' since the run did not complete normally — `sr._cancelled`
+                # may be False here (downstream exception rather than explicit cancel).
+                # We append directly rather than via `_append_response` to skip the usage-limit
+                # check; raising `UsageLimitExceeded` here would mask `stream_error`.
+                # `sr.get()` returns a response with run_id/conversation_id unset, so we assign
+                # those directly via `replace` rather than using `or` after the fact.
```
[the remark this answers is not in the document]
adtyavrdhn: Claude is being way too verbose lately. Loves to add ton of 100 line comments

PR #5367, opened 2026-05-11
feat: add `import_capability_types` to `AgentSpec`

### LINKED ISSUE #5266: [Feature Request] `AgentSpec` add `plugins` field for automatic custom capability loading
### Description

Currently, when using `Agent.from_spec()` or `Agent.from_file()` with custom capabilities, the user must explicitly pass `custom_capability_types` as a Python list of classes. This means the capabilities must be imported in code, and the spec file itself cannot be fully self‑contained – any third‑party or project‑specific capability requires a separate code change.

I propose adding an optional `plugins` field to `AgentSpec` (and therefore to the JSON/YAML spec format) that accepts a list of strings. Each string can be:
- a dotted module path (e.g., `"my_package.capabilities"`)
- an absolute or relative file path ending with `.py` (e.g., `"/home/user/my_cap.py"`)
- a `"module:ClassName"` pair for precise loading

During `Agent.from_spec()`, Pydantic AI would automatically scan these paths for `AbstractCapability` subclasses and merge them into the final capability list (as if they were passed via `custom_capability_types`). This makes the spec file the single source of truth and enables a true “zero‑code” configuration experience.

**Demo / how code might look**

User creates a spec file `my_agent.spec.json`:
```json
{
  "model": "openai:gpt-4",
  "plugins": [
    "my_company.waf_plugins",
    "/home/user/extra_caps.py"
  ],
  "capabilities": [
    {"Thinking": {"effort": "medium"}}
  ]
}
```

In Python, they simply do:
```python
agent = Agent.from_file("my_agent.spec.json")
# The capabilities from `my_company.waf_plugins` and `extra_caps.py` are automatically loaded and available.
```

No manual imports, no `custom_capability_types` list – the spec file drives everything. This is particularly useful for projects that distribute re‑usable agent configurations to end users who may not be Python developers.

**Proof of concept**

I’ve been building a tool that heavily relies on dynamic capability loading, and I’ve implemented a helper function that does exactly this. It works reliably with `.py` files, regular modules, and `module:Class` pairs. Here’s the core (stripped of error handling for brevity):

```python
import hashlib
import importlib.util
from pathlib import Path
from typing import Iterable

from pydantic_ai.capabilities import AbstractCapability

type CapabilityType = type[AbstractCapability]
type CapabilityList = list[CapabilityType]


def load_custom_capability(paths: Iterable[str]) -> CapabilityList:
    capability_types: CapabilityList = []
    seen: set[CapabilityType] = set()

    for path in filter(str.strip, paths):
        if path.endswith(".py"):
            capability_types.extend(_load_from_file(path, seen))
        elif ":" in path:
            capability_types.extend(_load_from_module_class(path, seen))
        else:
            capability_types.extend(_load_from_module(path, seen))

    return capability_types

# ... helper functions that scan modules for AbstractCapability subclasses ...
```

Having such logic built into `AgentSpec` would remove a lot of boilerplate and make the whole ecosystem more approachable.

**Use cases**

- **Re‑usable agent templates**: Teams publish spec files that automatically activate custom guardrails, logging, or domain‑specific tools without requiring any Python coding by the consumer.
- **Plugin systems**: Applications can discover and load third‑party capabilities simply by adding entries to a config file.
- **End‑user customisation**: Non‑Python users can extend an agent’s behaviour by writing a simple `.py` file and referencing it in the spec.

### References

- `pytest`’s `conftest.py` discovery and plugin loading via `pip` entry points is a well‑known example of configuration‑driven plugin loading.
- The official documentation already encourages capabilities as the primary extension point; this proposal just makes them easier to consume from specs.
- My own dynamic loader (shown above) has been working well in production for a workflow automation tool.

D7x7z49:
This PR adds `import_capability_types` to `AgentSpec`. It lets a spec load capability classes from files or modules.

What this PR does.

- Add `resolve_capability_path` to load a class from a file path or module path.
- Add the `import_capability_types` field to the `AgentSpec` model.
- Wire the new field into `_capabilities_from_spec` for agent creation.
- Add basic unit tests for `resolve_capability_path`.

Current state and next steps.

- The feature works but is not finished.
- I am new to this codebase. I do not yet know the project style for docs and tests.
- The tests cover the new function. They may need changes to match the project patterns.
- More tests and docs will come after I get feedback.
- This PR is a starting point for review and discussion.
- Closes #5266

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/_spec.py
+++ b/pydantic_ai_slim/pydantic_ai/_spec.py
@@ -377,7 +377,7 @@ def resolve_capability_path(path: str) -> type[AbstractCapability[Any]]:
         p = Path(source)
         spec = importlib.util.spec_from_file_location(f'__pydantic_ai_capability_{p.stem}', str(p.resolve()))
         if spec is None or spec.loader is None:
-            raise ValueError(f'Could not load module from {source!r}')
+            raise ValueError(f'Could not load module from {source!r}') # pragma: no cover
         mod = importlib.util.module_from_spec(spec)
         sys.modules[spec.name] = mod
         spec.loader.exec_module(mod)
--- a/pydantic_ai_slim/pydantic_ai/_spec.py
+++ b/pydantic_ai_slim/pydantic_ai/_spec.py
@@ -377,7 +377,7 @@ def resolve_capability_path(path: str) -> type[AbstractCapability[Any]]:
         p = Path(source)
         spec = importlib.util.spec_from_file_location(f'__pydantic_ai_capability_{p.stem}', str(p.resolve()))
         if spec is None or spec.loader is None:
-            raise ValueError(f'Could not load module from {source!r}') # pragma: no cover
+            raise ValueError(f'Could not load module from {source!r}')  # pragma: no cover
         mod = importlib.util.module_from_spec(spec)
         sys.modules[spec.name] = mod
         spec.loader.exec_module(mod)
--- a/pydantic_ai_slim/pydantic_ai/_spec.py
+++ b/pydantic_ai_slim/pydantic_ai/_spec.py
@@ -6,10 +6,13 @@ and registry/loading utilities that can be reused by both the evaluator system a
 
 from __future__ import annotations
 
+import importlib.util
 import inspect
+import sys
 import types
 import typing
 from collections.abc import Callable, Mapping, Sequence
+from pathlib import Path
 from typing import TYPE_CHECKING, Any, Literal, TypeVar, cast
 
 from pydantic import (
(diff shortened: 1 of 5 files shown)


D7x7z49:
Status: the core feature works. I added one function, resolve_capability_path, and wired it into _capabilities_from_spec to merge with custom_capability_types. I'm new to the project patterns, so docs and full pytest integration are still missing. I'll need your guidance on style.

Design simplification: instead of three path formats, I unified to a single "source:ClassName" pattern. The source can be a dotted module path or a .py file path. This keeps the design clean and avoids the unpredictability of bulk scanning.

Why not bulk import: a typical AgentSpec won't have dozens of capabilities because each capability is a composite of many behaviors. Explicitly naming the class is safer and clearer.

Error handling: resolve_capability_path does only basic format and existence checks. I left deeper validation to the call boundary on purpose. Premature catching inside the function could hide real problems from the developer.

This is a starting point. I'm excited to hear your feedback on the direction, the function design, and any adjustments needed to fit the project.

DouweM:
@D7x7z49 Thanks, will have a look tomorrow or next week!

DouweM:
I'm about to go on parental leave (back on June 29), let's work on this when I'm back, appreciate the patience!

PR #5369, opened 2026-05-11
Add `google_tool_config` field to `GoogleModelSettings`

### LINKED ISSUE #5366: Expose Gemini `tool_config.function_calling_config` in `GoogleModelSettings`
## Summary

`GoogleModelSettings` exposes `google_safety_settings`, `google_thinking_config`, `google_labels`, `google_video_resolution`, and `google_cached_content` — but **not** `google_tool_config`. The plumbing to send a `ToolConfigDict` to the Gemini API exists internally (in `_get_tool_config()` / `_build_content_and_config()`), but the only way to influence it is via the cross-provider `tool_choice` setting (#3611), which maps to `AUTO` / `ANY` / `NONE` only — **never `VALIDATED`**.

This means users currently cannot opt into Gemini's stricter schema-enforced function-calling mode, even though it would meaningfully reduce a real-world failure mode for Gemini 2.5 Flash.

## Why this matters

Gemini's `function_calling_config.mode` has four values per [the docs](https://ai.google.dev/gemini-api/docs/function-calling):

| Mode | Behavior |
|---|---|
| `AUTO` (default for function-only tools) | Model decides between text response or function call |
| `ANY` | Model must call a tool from the declared list |
| `VALIDATED` (default when mixing function tools with built-in tools) | Same as AUTO, but the API enforces schema before returning |
| `NONE` | Tools disabled |

The doc claims AUTO constrains the model to declared functions, but **Gemini 2.5 Flash is documented to hallucinate function names not in the declared list**:

- [googleapis/python-genai#813](https://github.com/googleapis/python-genai/issues/813) — Intermittent function call hallucination
- [langchain-ai/langchainjs#8589](https://github.com/langchain-ai/langchainjs/issues/8589) — `UNEXPECTED_TOOL_CALL` even with no tools passed
- [open-webui/open-webui#13439](https://github.com/open-webui/open-webui/discussions/13439) — Gemini 2.5 Flash hallucinates tool use

`VALIDATED` rejects malformed/hallucinated calls at the API layer instead of letting them propagate as `UnknownToolError` exceptions in user code.

## Real-world example (our case)

We run an ops agent (`gemini-2.5-flash`, ~28 registered tools, PydanticAI 1.62.0). Over a 24h window, ~3% of LLM-invoked cycles errored with hallucinated tool names like `'"reopens_thread"'` and `'\n  "kind"'` — both extracted from JSON-shaped strings in our system prompt, even when those strings were inside fenced ```json``` code blocks. Each error aborts the whole agent turn before any tool can be called.

Workarounds we explored:
- Rewriting the prompt to avoid JSON-shaped text — partial fix, hides the symptom
- Catch + retry on `UnknownToolError` — adds round-trips, doesn't address root cause
- Switch to `gemini-2.5-pro` — slower, more expensive

The clean fix is `VALIDATED` mode, but PydanticAI doesn't let us reach it.

## Current state of the code

In `pydantic_ai_slim/pydantic_ai/models/google.py`:

```python
class GoogleModelSettings(ModelSettings, total=False):
    google_safety_settings: list[SafetySettingDict]
    google_thinking_config: ThinkingConfigDict
    google_labels: dict[str, str]
    google_video_resolution: MediaResolution
    google_cached_content: str
    # ... no google_tool_config
```

But `_get_tool_config()` already builds a `ToolConfigDict` that's passed to the underlying `genai.types.GenerateContentConfigDict.tool_config`. The override hook just needs to be exposed in settings.

## Proposed API

Mirror the existing `google_*` field convention:

```python
class GoogleModelSettings(ModelSettings, total=False):
    # ... existing fields ...
    google_tool_config: ToolConfigDict
    """Override the tool_config sent to Gemini. When set, takes precedence
    over the tool_config derived from the cross-provider `tool_choice`
    setting. Use this to opt into modes the cross-provider abstraction
    doesn't expose — most notably `VALIDATED`, which enforces schema at
    the API layer and rejects hallucinated function names that AUTO mode
    can let through (Gemini 2.5 Flash bug, see linked issues).
    """
```

User code:

```python
from google.genai.types import (
    ToolConfigDict, FunctionCallingConfigDict, FunctionCallingConfigMode
)

settings = GoogleModelSettings(
    temperature=0.1,
    google_tool_config=ToolConfigDict(
        function_calling_config=FunctionCallingConfigDict(
            mode=FunctionCallingConfigMode.VALIDATED,
        ),
    ),
)
```

## Implementation sketch

In `_build_content_and_config()` (or wherever `tool_config` is currently assembled), after computing the default `tool_config` from `resolve_tool_choice()`, check `model_settings.get('google_tool_config')` and prefer that if set:

```python
tool_config = (
    model_settings.get('google_tool_config')
    or computed_tool_config
)
```

Same pattern as `google_thinking_config` already follows.

## Related work

- #3611 (merged) — Added cross-provider `tool_choice`. Maps to AUTO/ANY/NONE only; this issue is about exposing the **VALIDATED** mode that the cross-provider abstraction can't represent.
- #1820 (open) — "Require agent to use tool(s) before final output". Related but different: that's about ANY mode (force tool call); this is about VALIDATED mode (enforce schema on optional tool call).

## Willing to PR

If maintainers think the proposed API is right, happy to send the PR (it should be ~20 lines including a test).

### References

_No response_

sapirbaruch:
Exposes ToolConfigDict as a user-overridable setting so callers can opt into Gemini function-calling modes (notably VALIDATED) that the cross-provider tool_choice abstraction does not cover.

Closes #5366

- Closes #<issue>

### Checklist

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/google.py
+++ b/pydantic_ai_slim/pydantic_ai/models/google.py
@@ -280,6 +280,16 @@ class GoogleModelSettings(ModelSettings, total=False):
     google_service_tier: GoogleServiceTier
     """Deprecated: use `service_tier` for Gemini API (GLA) or `google_vertex_service_tier` for Vertex AI."""
 
+    google_tool_config: ToolConfigDict
+    """Override the tool_config sent to Gemini.
+
+    When set, this takes precedence over the tool_config derived from the cross-provider `tool_choice` setting.
+    Use this to opt into modes not exposed by the cross-provider abstraction — most notably `VALIDATED`,
+    which enforces schema at the API layer and rejects hallucinated function names before they reach user code.
+
+    See <https://ai.google.dev/gemini-api/docs/function-calling> for more information.
+    """
+
 
 def _get_deprecated_google_service_tier(model_settings: GoogleModelSettings) -> GoogleServiceTier | None:
     """Return `google_service_tier`, emitting a `DeprecationWarning` when it is set."""
@@ -662,6 +672,8 @@ class GoogleModel(Model[Client]):
         if allowed_function_names:
             function_calling_config['allowed_function_names'] = allowed_function_names
         tool_config = ToolConfigDict(function_calling_config=function_calling_config)
+        if override_config := model_settings.get('google_tool_config'):
+            tool_config = override_config
 
         tools: list[ToolDict] = [
             ToolDict(function_declarations=[_function_declaration_from_tool(t)]) for t in tool_defs.values()
--- a/tests/models/test_tool_choice_unit.py
+++ b/tests/models/test_tool_choice_unit.py
@@ -517,6 +517,29 @@ def test_google_auto_tuple_filters_tool_defs():
     assert tool_config['function_calling_config']['mode'].name == 'AUTO'  # pyright: ignore[reportTypedDictNotRequiredAccess,reportOptionalMemberAccess,reportOptionalSubscript,reportUnknownMemberType]
 
 
+@pytest.mark.skipif(not google_available(), reason='google not installed')
+def test_google_tool_config_override_takes_precedence():
+    """google_tool_config in model settings overrides the tool_config computed from tool_choice."""
+    from google.genai.types import FunctionCallingConfigDict, FunctionCallingConfigMode, ToolConfigDict
+
+    mock_client = MagicMock()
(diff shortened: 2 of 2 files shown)


DouweM:
Closing because it's quite outdated; if this is the shape of fix we want we can easily regenerate something similar off of `main`.

PR #5340, opened 2026-05-08
Convert `ModelProfile` to `TypedDict` for easier per-model overrides

DouweM:
- Closes #3796

### Checklist

- [ ] Any **AI generated code** has been reviewed line-by-line by the human PR author, who stands by it.
- [ ] No **breaking changes** in accordance with the [version policy](https://github.com/pydantic/pydantic-ai/blob/main/docs/version-policy.md).
- [ ] **PR title** is fit for the [release changelog](https://github.com/pydantic/pydantic-ai/releases).

### Why

`ModelProfile` and its subclasses were `@dataclass(kw_only=True)` while the closely-related `ModelSettings` was already a `TypedDict(total=False)`. That asymmetry produced three structural pain points that v2 lets us fix:

1. **Cross-provider routing forced "downsize then upsize."** Bedrock hosting Anthropic models had to build a `BedrockModelProfile`, call `.update()` with the `AnthropicModelProfile`, then `replace(...)` to override one field. Class identity made the mixing awkward.
2. **`ModelProfile.update()` was leaky.** The `f.default == value` heuristic couldn't distinguish "user explicitly set X to its default" from "user didn't pass X."
3. **Two mental models for one concept.** Users had to learn dataclass semantics for `ModelProfile` and TypedDict semantics for `ModelSettings`.

Plus the original issue: overriding a single field on a built-in profile required constructing a full `OpenAIModelProfile(...)` from scratch.

### What changed

- `ModelProfile` and all subclasses (`OpenAIModelProfile`, `AnthropicModelProfile`, `BedrockModelProfile`, `GoogleModelProfile`, `GrokModelProfile`, `GroqModelProfile`) are now `TypedDict(total=False)`. Cross-class mixing happens via dict-spread; no class identity to commit to.
- `ModelProfile.update()` and `ModelProfile.from_profile()` removed; replaced by module-level `merge_profile(base, *overrides)` doing later-wins dict-spread.
- `OpenAIModelProfile.__post_init__` validation lifted to a module-level `validate_openai_profile(profile)` called from `OpenAIChatModel.__init__`.
- `Model.profile` cached_property now does a layered merge: `DEFAULT_PROFILE` + `provider.model_profile(model_name)` + user `profile=` (dict or callable).
[...]

PR #5375, opened 2026-05-11
feat(openai): add `openai_chat_supports_multiple_system_messages` profile flag

DouweM:
### What

Some OpenAI-compatible backends (e.g. some LiteLLM/vLLM deployments serving Qwen and similar) reject requests with more than one initial system message:

> System message must be at the beginning.

Pydantic AI currently sends static `system_prompt`s and dynamic `instructions` as separate system-role messages (and likewise for multiple `@agent.system_prompt`-decorated functions), so users hit this on strict backends today.

### How

New `OpenAIModelProfile` flag, defaulting to OpenAI's behavior:

```python
profile=OpenAIModelProfile(openai_chat_supports_multiple_system_messages=False)
```

When `False`, consecutive `system`/`developer`-role messages at the start of the request are merged into one (joined with `\n\n`) inside `OpenAIChatModel._map_messages`. `user`-role system prompts (`openai_system_prompt_role='user'`) are intentionally skipped — we can't tell a system-prompt-cast-as-user apart from a real user turn.

The default (`True`) preserves OpenAI/most-providers behavior; this is opt-in for strict OpenAI-compatible backends, matching the existing `openai_chat_supports_*` pattern (web_search, file_urls, document_input).

Docs example in `docs/models/openai.md` updated to mention the flag alongside `openai_supports_strict_tool_definition`.

### Coordination

I left a comment on #5340 (TypedDict redesign) so the new field is folded into that migration (drop the `= True` default, document as `Default: \`True\``, switch read site to `.get('openai_chat_supports_multiple_system_messages', True)`).

### Checklist

### DIFF
--- a/docs/models/openai.md
+++ b/docs/models/openai.md
@@ -361,9 +361,9 @@ model = OpenAIChatModel(
         base_url='https://<openai-compatible-api-endpoint>.com', api_key='your-api-key'
     ),
     profile=OpenAIModelProfile(
-        json_schema_transformer=InlineDefsJsonSchemaTransformer,  # Supported by any model class on a plain ModelProfile
-        openai_supports_strict_tool_definition=False,  # Supported by OpenAIModel only, requires OpenAIModelProfile
-        openai_chat_supports_multiple_system_messages=False,  # For strict providers (e.g. some vLLM/LiteLLM setups) that require exactly one initial system message
+        json_schema_transformer=InlineDefsJsonSchemaTransformer,  # Supported by any model class via the base ModelProfile
+        openai_supports_strict_tool_definition=False,  # Supported by OpenAIChatModel and OpenAIResponsesModel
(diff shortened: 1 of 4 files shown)


docs/models/openai.md
```diff
@@ -362,7 +362,8 @@ model = OpenAIChatModel(
     ),
     profile=OpenAIModelProfile(
         json_schema_transformer=InlineDefsJsonSchemaTransformer,  # Supported by any model class on a plain ModelProfile
-        openai_supports_strict_tool_definition=False  # Supported by OpenAIModel only, requires OpenAIModelProfile
+        openai_supports_strict_tool_definition=False,  # Supported by OpenAIModel only, requires OpenAIModelProfile
```
DouweM: `OpenAIModel` is deprecated :) Which model class(es) is this really about?

PR #5378, opened 2026-05-12
Fix OpenRouter chat `max_tokens` routing

### LINKED ISSUE #5186: Pydantic AI `OpenAIChatModel.max_tokens` `max_completion_tokens` handling break on certain providers

Genmin:
## Summary
- add an `OpenAIModelProfile.openai_chat_max_tokens_parameter` switch for Chat Completions token-limit routing
- keep the default OpenAI behavior unchanged by continuing to send `ModelSettings.max_tokens` as `max_completion_tokens`
- set OpenRouter's profile to send `ModelSettings.max_tokens` as the legacy `max_tokens` parameter, avoiding provider/model failures when `max_completion_tokens` is unsupported or ignored
- add regression coverage for both the default OpenAI route and the OpenRouter route

Fixes #5186

## Tests
- `uv run pytest tests/models/test_openai.py::test_openai_chat_max_tokens_defaults_to_max_completion_tokens tests/models/test_openrouter.py::test_openrouter_max_tokens_uses_chat_max_tokens_parameter -q`
- `uv run ruff format --check pydantic_ai_slim/pydantic_ai/models/openai.py pydantic_ai_slim/pydantic_ai/profiles/openai.py pydantic_ai_slim/pydantic_ai/providers/openrouter.py tests/models/test_openai.py tests/models/test_openrouter.py`
- `uv run ruff check pydantic_ai_slim/pydantic_ai/models/openai.py pydantic_ai_slim/pydantic_ai/profiles/openai.py pydantic_ai_slim/pydantic_ai/providers/openrouter.py tests/models/test_openai.py tests/models/test_openrouter.py`
- `PYRIGHT_PYTHON_IGNORE_WARNINGS=1 uv run pyright pydantic_ai_slim/pydantic_ai/models/openai.py pydantic_ai_slim/pydantic_ai/profiles/openai.py pydantic_ai_slim/pydantic_ai/providers/openrouter.py tests/models/test_openai.py tests/models/test_openrouter.py`
- `git diff --check`

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/openai.py
+++ b/pydantic_ai_slim/pydantic_ai/models/openai.py
@@ -929,7 +929,16 @@ class OpenAIChatModel(Model[AsyncOpenAI]):
                     stream=stream,
                     stream_options=self._get_stream_options(model_settings) if stream else OMIT,
                     stop=model_settings.get('stop_sequences', OMIT),
-                    max_completion_tokens=model_settings.get('max_tokens', OMIT),
+                    max_tokens=(
+                        model_settings.get('max_tokens', OMIT)
+                        if profile.openai_chat_max_tokens_parameter == 'max_tokens'
+                        else OMIT
+                    ),
+                    max_completion_tokens=(
+                        model_settings.get('max_tokens', OMIT)
+                        if profile.openai_chat_max_tokens_parameter == 'max_completion_tokens'
+                        else OMIT
+                    ),
                     timeout=model_settings.get('timeout', NOT_GIVEN),
                     response_format=response_format or OMIT,
                     seed=model_settings.get('seed', OMIT),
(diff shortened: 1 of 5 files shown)


dsfaccini:
David's AICA here: credit where it's due, @Genmin. #5378 independently reached the same model-profile routing design for #5186, a month before we did, and it was correct and mergeable. We kept https://github.com/pydantic/pydantic-ai/pull/5926 only for a few deltas (the flag name matching the existing `openai_chat_supports_*` profile family, a test asserting the profile survives the `.update()` merge, and docs), and we've credited you as co-author there. Sorry the bot auto-closed this one, and thanks for the fix.

PR #4276, opened 2026-02-10
fix: avoid stale client references in model and embedding classes

Deeven-Seru:
- Closes #4336

### Pre-Review Checklist

- [x] Any **AI generated code** has been reviewed line-by-line by the human PR author, who stands by it.
- [x] No **breaking changes** in accordance with the [version policy](https://github.com/pydantic/pydantic-ai/blob/main/docs/version-policy.md).
- [x] **Linting and type checking** pass per `make format` and `make typecheck`.
- [x] **PR title** is fit for the [release changelog](https://github.com/pydantic/pydantic-ai/releases).

### Pre-Merge Checklist

- [x] New **tests** for any fix or new behavior, maintaining 100% coverage.
- [ ] Updated **documentation** for new features and behaviors, including docstrings for API docs.

---

### Summary

Model and embedding classes previously cached a reference to the provider's client via `self.client = provider.client` / `self._client = provider.client` in `__init__`. This is a problem in environments where a provider may be re-initialized or swapped after construction (e.g. durable execution frameworks like Temporal, or tests with dependency injection): the model/embedding keeps using the stale client.

### Changes

- Replaced the cached `client` dataclass field + init assignment with a `@property` that delegates to `self._provider.client` in every affected class:
  - **Models** (11): `OpenAIChatModel`, `OpenAIResponsesModel`, `AnthropicModel`, `GroqModel`, `MistralModel`, `GoogleModel`, `GeminiModel`, `CohereModel`, `HuggingFaceModel`, `BedrockConverseModel`, `XaiModel`.
  - **Embeddings** (4): `OpenAIEmbeddingModel`, `GoogleEmbeddingModel`, `CohereEmbeddingModel`, `BedrockEmbeddingModel`.
- `GeminiModel.base_url` now derives from `self.client.base_url` instead of a cached `self._url`, matching the other models.
- Added `assert model.client is provider.client` identity checks to each model's existing `test_init` (or a new minimal test where none existed), replacing an earlier draft that used a `_SwappableProvider` subclass + private `_client` mutation.

PR #5392, opened 2026-05-12
Restore ability to swap the Bedrock client at runtime

### LINKED ISSUE #5386: `BedrockConverseModel.client` changed from settable field to read-only property in v1.86.1 (breaking change)
### Initial Checks

### Description

In v1.86.0, `BedrockConverseModel.client` was a regular dataclass field that could be reassigned at runtime. In v1.86.1 (PR #4276), it became a read-only `@property`. This is a breaking change in a patch release, any code that assigns to `model.client` now raises `AttributeError`.

We use this for credential rotation in long-running services. Bedrock models run with temporary STS credentials (1-hour TTL), and a pre-call hook replaces the boto3 client before each LLM invocation when credentials are near expiry. This worked by assigning directly to `model.client`.

After v1.86.1 there is no public API to achieve this, the only option is `model._provider._client = new_client`, reaching two levels into private internals.

We understand PR #4276 fixed stale client references, but it removed the ability for downstream code to handle credential rotation without providing a replacement mechanism.

What is the recommended way to handle Bedrock credential rotation with `BedrockConverseModel`? Is there a supported pattern we're missing, or a public API to replace the client at runtime?

```Python
python
import boto3

model = BedrockConverseModel("anthropic.claude-sonnet-4-6")

# Credential rotation: assume new role, get fresh boto3 client
new_client = boto3.Session(
    aws_access_key_id="...",
    aws_secret_access_key="...",
    aws_session_token="...",
).client("bedrock-runtime")

# This worked in v1.86.0, raises AttributeError in v1.86.1+:
model.client = new_client

# Only workaround — two levels of private internals:
model._provider._client = new_client
```

### Logfire Trace

_No response_

- Python: 3.11+
- pydantic-ai: 1.86.1

DouweM:
- Closes #5386

In #4276, `BedrockConverseModel.client` went from a plain (settable) attribute to a read-only `@property`, which broke code that reassigned it — e.g. swapping the boto3 client to rotate short-lived STS credentials in a long-running service.

This restores that ability:

- `BedrockProvider.client` now has a setter — the preferred place to swap the client, since it's shared by every model using the provider.
- `BedrockConverseModel.client` gets its setter back for backward compatibility. A model-level assignment stores a local client that shadows the provider's; once you've assigned one, you own keeping it valid (so the stale-reference fix from #4276 still holds for everyone who doesn't opt in).

### Checklist

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/bedrock.py
+++ b/pydantic_ai_slim/pydantic_ai/models/bedrock.py
@@ -433,9 +433,7 @@ class BedrockConverseModel(Model[BaseClient]):
         provider pick up the new client. Once you've assigned a client here, you're responsible for keeping it valid;
         the provider's client is no longer consulted.
         """
-        if self._client is not None:
-            return cast('BedrockRuntimeClient', self._client)
-        return cast('BedrockRuntimeClient', self._provider.client)
+        return cast('BedrockRuntimeClient', self._client or self._provider.client)
 
     @client.setter
     def client(self, client: BedrockRuntimeClient) -> None:
--- a/pydantic_ai_slim/pydantic_ai/models/bedrock.py
+++ b/pydantic_ai_slim/pydantic_ai/models/bedrock.py
@@ -392,6 +392,7 @@ class BedrockConverseModel(Model[BaseClient]):
 
     _model_name: BedrockModelName = field(repr=False)
     _provider: Provider[BaseClient] = field(repr=False)
+    _client: BaseClient | None = field(default=None, repr=False)
 
     def __init__(
         self,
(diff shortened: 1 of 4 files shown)


pydantic_ai_slim/pydantic_ai/models/bedrock.py
```diff
@@ -423,8 +425,24 @@ def __init__(
...
+
+        Defaults to the client from the [`Provider`][pydantic_ai.providers.Provider]. It can be reassigned, e.g. to
+        rotate short-lived credentials in a long-running service, but prefer assigning to
+        [`BedrockProvider.client`][pydantic_ai.providers.bedrock.BedrockProvider.client] so all models sharing the
+        provider pick up the new client. Once you've assigned a client here, you're responsible for keeping it valid;
+        the provider's client is no longer consulted.
+        """
+        if self._client is not None:
+            return cast('BedrockRuntimeClient', self._client)
         return cast('BedrockRuntimeClient', self._provider.client)
```
DouweM: Can be `self._client or self._provider.client`, right?

PR #5395, opened 2026-05-12
fix(anthropic): normalize Bedrock model IDs for capability profile lookup

### LINKED ISSUE #5324: `AnthropicProvider` doesn't normalize model names for Bedrock/Vertex/Azure clients (incl. tool-search default strategy)

DouweM:
- Closes #5324

> [!NOTE]
> Targets the `tool-search` branch (#5143), since the tool-search default-strategy part of #5324 builds on code introduced there. Will retarget to `main` once #5143 merges. Supersedes #5344 (thanks @kclisp for the original report and PR).

### What this does

**1. Model-name normalization for Bedrock clients (issue point 4).**
`AnthropicProvider.model_profile` now strips the Bedrock `anthropic.` provider segment — plus the cross-region geo prefix and the `-vN(:M)?` version suffix — before calling `anthropic_model_profile`, so `model_name.startswith('claude-…')` checks match for a model id like `us.anthropic.claude-haiku-4-5-20251001-v1:0`. Previously these silently flipped `supports_json_schema_output`, tool-search support, `anthropic_supports_adaptive_thinking`, etc. to their defaults. The full transport-specific id still goes on the wire via `AnthropicModel._model_name`.

The geo/version parsing is extracted from `BedrockProvider.model_profile` into a boto3-free `pydantic_ai.providers._bedrock_model_names` helper (`split_bedrock_model_id`, plus `remove_bedrock_geo_prefix` / `BEDROCK_GEO_PREFIXES`, re-exported from `providers.bedrock` for backwards compat) so both providers share it. `BedrockProvider.model_profile` is refactored onto the shared helper.

**2. `AsyncAnthropicBedrockMantle` support.** Added to the `AsyncAnthropicClient` union and the provider docstring — the newer "Claude in Amazon Bedrock" (Messages-API) client Anthropic now recommends. It uses the same `anthropic.claude-…` model-id shape, so it's covered by the normalization above.

**3. Tool-search default strategy on Bedrock (issue point 5).**
The legacy Bedrock InvokeModel API (`AsyncAnthropicBedrock`) 400s on `tool_search_tool_bm25_*` (`BM25 tool search is not supported on Bedrock. Use tool_search_tool_regex instead.`). So on that client, `ToolSearch(strategy=None)` resolves to `regex` instead of `bm25`, and an explicit `ToolSearch(strategy='bm25')` raises a `UserError` rather than an opaque 400. (`AsyncAnthropicBedrockMantle` uses the Messages API and keeps the `bm25` default.) This follows the existing `_FAST_MODE_UNSUPPORTED_CLIENTS` / `_client_supports_fast_speed` pattern — a request-time transport check, not a profile flag, since bm25-vs-regex is a transport property (every tool-search-capable Claude model supports both on the direct API).

### Test plan

- `AnthropicProvider.model_profile` resolves the right capability flags for direct, Bedrock (`anthropic.…`, `…-v1:0`, geo-prefixed), and Vertex (`@`-versioned) model ids — and doesn't over-strip (an older `claude-3-5-sonnet` id still resolves to the right negative flags).
- `AnthropicModel` with `AsyncAnthropicBedrock` / `AsyncAnthropicBedrockMantle` clients: profile resolved from the bare name, full id kept on the wire.
- Tool search: `regex` default + explicit `bm25` `UserError` on `AsyncAnthropicBedrock`; `bm25` default on the direct and Mantle clients.
- `split_bedrock_model_id` unit tests; existing `BedrockProvider.model_profile` / `remove_bedrock_geo_prefix` tests still pass after the refactor.

### Checklist

### DIFF
--- a/docs/models/anthropic.md
+++ b/docs/models/anthropic.md
@@ -133,8 +133,6 @@ agent = Agent(model)
 ...
 ```
 
-The transport-specific model ID (e.g. `anthropic.claude-haiku-4-5` or `us.anthropic.claude-sonnet-4-5-20250929-v1:0`) is used as-is on the wire; the model's [`ModelProfile`][pydantic_ai.profiles.ModelProfile] (which capability flags it gets) is looked up from the underlying `claude-...` name.
-
 !!! note "Bedrock vs BedrockConverseModel"
     This approach uses Anthropic's SDK with AWS Bedrock credentials. For an alternative using AWS SDK (boto3) directly, see [`BedrockConverseModel`](bedrock.md).
 
@@ -158,9 +156,6 @@ model = AnthropicModel('claude-sonnet-4-5', provider=provider)
 agent = Agent(model)
 ...
 ```
-
-!!! note "Vertex vs GoogleModel"
-    This approach uses Anthropic's SDK with Vertex AI credentials. For an alternative using Google's Vertex AI SDK directly, see [`GoogleModel`](google.md).
 
 ### Microsoft Foundry
 
--- a/docs/models/anthropic.md
+++ b/docs/models/anthropic.md
@@ -117,24 +117,29 @@ You can use Anthropic models through cloud platforms by passing a custom client 
 
 ### AWS Bedrock
 
-To use Claude models via [AWS Bedrock](https://aws.amazon.com/bedrock/claude/), follow the [Anthropic documentation](https://docs.anthropic.com/en/api/claude-on-amazon-bedrock) on how to set up an `AsyncAnthropicBedrock` client and then pass it to `AnthropicProvider`:
+To use Claude models via [AWS Bedrock](https://aws.amazon.com/bedrock/claude/), follow the [Anthropic documentation](https://platform.claude.com/docs/en/build-with-claude/claude-in-amazon-bedrock) on how to set up a Bedrock client and then pass it to `AnthropicProvider`. Both the newer `AsyncAnthropicBedrockMantle` client (recommended by Anthropic, using the Messages API) and the legacy `AsyncAnthropicBedrock` client (using the `InvokeModel` API with ARN-versioned model IDs) are supported:
 
 ```python {test="skip"}
-from anthropic import AsyncAnthropicBedrock
+from anthropic import AsyncAnthropicBedrockMantle
 
 from pydantic_ai import Agent
 from pydantic_ai.models.anthropic import AnthropicModel
 from pydantic_ai.providers.anthropic import AnthropicProvider
 
-bedrock_client = AsyncAnthropicBedrock()  # Uses AWS credentials from environment
+bedrock_client = AsyncAnthropicBedrockMantle()  # Uses AWS credentials from environment
 provider = AnthropicProvider(anthropic_client=bedrock_client)
-model = AnthropicModel('us.anthropic.claude-sonnet-4-5-20250929-v1:0', provider=provider)
+model = AnthropicModel('anthropic.claude-haiku-4-5', provider=provider)
 agent = Agent(model)
 ...
 ```
 
+The transport-specific model ID (e.g. `anthropic.claude-haiku-4-5` or `us.anthropic.claude-sonnet-4-5-20250929-v1:0`) is used as-is on the wire; the model's [`ModelProfile`][pydantic_ai.profiles.ModelProfile] (which capability flags it gets) is looked up from the underlying `claude-...` name.
+
 !!! note "Bedrock vs BedrockConverseModel"
     This approach uses Anthropic's SDK with AWS Bedrock credentials. For an alternative using AWS SDK (boto3) directly, see [`BedrockConverseModel`](bedrock.md).
+
+!!! note "Tool search on the legacy `AsyncAnthropicBedrock` client"
+    The legacy `InvokeModel` API doesn't support the `bm25` [tool search](../tools-advanced.md#tool-search) variant, so [`ToolSearch`][pydantic_ai.capabilities.ToolSearch] defaults to `'regex'` on the `AsyncAnthropicBedrock` client (instead of `'bm25'`), and passing `ToolSearch(strategy='bm25')` raises a `UserError`.
 
 ### Google Vertex AI
 
--- a/pydantic_ai_slim/pydantic_ai/models/anthropic.py
+++ b/pydantic_ai_slim/pydantic_ai/models/anthropic.py
@@ -100,6 +100,7 @@ try:
         APIConnectionError,
         APIStatusError,
         AsyncAnthropicBedrock,
+        AsyncAnthropicBedrockMantle,
         AsyncAnthropicFoundry,
         AsyncAnthropicVertex,
         AsyncStream,
@@ -216,12 +217,21 @@ except ImportError as _import_error:
         'you can use the `anthropic` optional group — `pip install "pydantic-ai-slim[anthropic]"`'
     ) from _import_error
 
+# `AsyncAnthropicBedrockMantle` uses the Messages API and supports automatic prompt caching (unlike the
+# legacy `AsyncAnthropicBedrock` InvokeModel API), so it's not in `_NON_AUTOMATIC_CACHING_CLIENTS`. Fast
+# mode is not available on any Bedrock transport, so it goes in `_FAST_MODE_UNSUPPORTED_CLIENTS`.
 _NON_AUTOMATIC_CACHING_CLIENTS = (AsyncAnthropicBedrock, AsyncAnthropicVertex)
-_FAST_MODE_UNSUPPORTED_CLIENTS = (AsyncAnthropicBedrock, AsyncAnthropicFoundry, AsyncAnthropicVertex)
+_FAST_MODE_UNSUPPORTED_CLIENTS = (
+    AsyncAnthropicBedrock,
+    AsyncAnthropicBedrockMantle,
+    AsyncAnthropicFoundry,
+    AsyncAnthropicVertex,
+)
 # The legacy Bedrock InvokeModel API (`AsyncAnthropicBedrock`) doesn't support the `bm25` tool-search
 # variant — it 400s with `BM25 tool search is not supported on Bedrock. Use tool_search_tool_regex instead.`
-# — so we default to `regex` there. (`AsyncAnthropicBedrockMantle` uses the Messages API and is treated
-# like the direct API.)
+# — so we default to `regex` there. The other transports (Vertex, Foundry, and the Messages-API-based
+# `AsyncAnthropicBedrockMantle`) expose the full Messages API, including both tool-search variants, so they
+# keep the `bm25` default.
 _BM25_TOOL_SEARCH_UNSUPPORTED_CLIENTS = (AsyncAnthropicBedrock,)
 
 _ANTHROPIC_SAMPLING_PARAMS = ('temperature', 'top_p', 'top_k')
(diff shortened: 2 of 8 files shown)


pydantic_ai_slim/pydantic_ai/providers/anthropic.py
```diff
@@ -44,6 +53,14 @@ def client(self) -> AsyncAnthropicClient:
 
     @staticmethod
     def model_profile(model_name: str) -> ModelProfile | None:
+        # When the underlying client is `AsyncAnthropicBedrock`/`AsyncAnthropicBedrockMantle`, the model
+        # name carries a Bedrock `anthropic.` provider segment (and, on the legacy InvokeModel API, a
+        # `-v<n>(:<m>)?` version suffix and optional cross-region geo prefix), e.g.
+        # `us.anthropic.claude-haiku-4-5-20251001-v1:0`. Strip it so `anthropic_model_profile`'s
+        # `claude-...` prefix checks match; the full model name still goes on the wire via `AnthropicModel._model_name`.
+        bedrock_provider, base_model_name = split_bedrock_model_id(model_name)
```
DouweM, line 61: I had been hoping that `model_profile` had access to `self.client` so we could only run this when needed, but I guess not, and I don't want to make it non-static, so this is fine.

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -218,6 +218,11 @@
 
 _NON_AUTOMATIC_CACHING_CLIENTS = (AsyncAnthropicBedrock, AsyncAnthropicVertex)
 _FAST_MODE_UNSUPPORTED_CLIENTS = (AsyncAnthropicBedrock, AsyncAnthropicFoundry, AsyncAnthropicVertex)
+# The legacy Bedrock InvokeModel API (`AsyncAnthropicBedrock`) doesn't support the `bm25` tool-search
+# variant — it 400s with `BM25 tool search is not supported on Bedrock. Use tool_search_tool_regex instead.`
+# — so we default to `regex` there. (`AsyncAnthropicBedrockMantle` uses the Messages API and is treated
+# like the direct API.)
+_BM25_TOOL_SEARCH_UNSUPPORTED_CLIENTS = (AsyncAnthropicBedrock,)
```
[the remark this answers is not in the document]
DouweM: All the other providers support bm25? We know for sure? Or do we expect to get people complaining in the future and we can fix this then? Check the Anthropic docs if you haven't already.

docs/models/anthropic.md
```diff
@@ -117,25 +117,30 @@ You can use Anthropic models through cloud platforms by passing a custom client
...
-bedrock_client = AsyncAnthropicBedrock()  # Uses AWS credentials from environment
+bedrock_client = AsyncAnthropicBedrockMantle()  # Uses AWS credentials from environment
 provider = AnthropicProvider(anthropic_client=bedrock_client)
-model = AnthropicModel('us.anthropic.claude-sonnet-4-5-20250929-v1:0', provider=provider)
+model = AnthropicModel('anthropic.claude-haiku-4-5', provider=provider)
 agent = Agent(model)
 ...
 ```
 
+The transport-specific model ID (e.g. `anthropic.claude-haiku-4-5` or `us.anthropic.claude-sonnet-4-5-20250929-v1:0`) is used as-is on the wire; the model's [`ModelProfile`][pydantic_ai.profiles.ModelProfile] (which capability flags it gets) is looked up from the underlying `claude-...` name.
```
DouweM: Not worth specifying.

docs/models/anthropic.md
```diff
@@ -117,25 +117,30 @@ You can use Anthropic models through cloud platforms by passing a custom client
...
 
 from pydantic_ai import Agent
 from pydantic_ai.models.anthropic import AnthropicModel
 from pydantic_ai.providers.anthropic import AnthropicProvider
 
-bedrock_client = AsyncAnthropicBedrock()  # Uses AWS credentials from environment
+bedrock_client = AsyncAnthropicBedrockMantle()  # Uses AWS credentials from environment
 provider = AnthropicProvider(anthropic_client=bedrock_client)
-model = AnthropicModel('us.anthropic.claude-sonnet-4-5-20250929-v1:0', provider=provider)
+model = AnthropicModel('anthropic.claude-haiku-4-5', provider=provider)
```
DouweM, line 131: Why was this simplified?

docs/models/anthropic.md
```diff
@@ -117,25 +117,30 @@ You can use Anthropic models through cloud platforms by passing a custom client
...
 agent = Agent(model)
 ...
 ```
 
+The transport-specific model ID (e.g. `anthropic.claude-haiku-4-5` or `us.anthropic.claude-sonnet-4-5-20250929-v1:0`) is used as-is on the wire; the model's [`ModelProfile`][pydantic_ai.profiles.ModelProfile] (which capability flags it gets) is looked up from the underlying `claude-...` name.
+
 !!! note "Bedrock vs BedrockConverseModel"
     This approach uses Anthropic's SDK with AWS Bedrock credentials. For an alternative using AWS SDK (boto3) directly, see [`BedrockConverseModel`](bedrock.md).
```
DouweM, line 137: I learnt recently that `GoogleProvider(vertexai=True)` / `GoogleModel` doesn't actually support Anthropic models (despite having a Model Garden) -- they need to use AnthropicModel with AsyncVertexClient, but the section below on Vertex still suggests that the Google model/provider route is valid. Remove the section there.

pydantic_ai_slim/pydantic_ai/providers/anthropic.py
DouweM: Right — it's a `@staticmethod` because `infer_model_profile()` calls `ProviderClass.model_profile(name)` on the class without instantiating (no credentials needed just to look up a profile). The normalization here is name-only, so it's a cheap no-op for any non-Bedrock model name. See my reply on the `_BM25_TOOL_SEARCH_UNSUPPORTED_CLIENTS` comment re the strategy default.

pydantic_ai_slim/pydantic_ai/models/anthropic.py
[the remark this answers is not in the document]
DouweM: Checked the docs:

- **Vertex**: "The Vertex API for accessing Claude is nearly-identical to the Messages API and supports all of the same options."
- **Foundry** "features not supported": Admin API, Compliance API, Models API, Message Batches API — that's the whole list.
- **Bedrock Mantle** ("Claude in Amazon Bedrock", the new Messages-API integration) "not supported": the Anthropic-defined server tools (Web Search, Web Fetch, Remote MCP, Memory, Files, Computer Use, Skills, Code Execution), Claude Managed Agents, Message Batches, `/v1/users`.

Tool search isn't excluded on any of those, and they all expose the Messages API where both `bm25` and `regex` work. The only documented limitation is the legacy InvokeModel path (`AsyncAnthropicBedrock`), which 400s with `BM25 tool search is not supported on Bedrock. Use tool_search_tool_regex instead.` — so that's the only client in the tuple. Expanded the comment to say this in `e04b9a8e8`; if a complaint surfaces on one of the others it's a one-line addition.

docs/models/anthropic.md
DouweM: Removed in `e04b9a8e8`.

docs/models/anthropic.md
DouweM: Switched the example to `AsyncAnthropicBedrockMantle` (Anthropic's current recommended Bedrock client — the `AsyncAnthropicBedrock` InvokeModel path is now documented as "legacy"), and Mantle model IDs are the short `anthropic.claude-…` form (no ARN `-v1:0` suffix or geo prefix). Mantle currently only serves Haiku 4.5 / Opus 4.7, hence `anthropic.claude-haiku-4-5`. Happy to keep a legacy `AsyncAnthropicBedrock` + `us.anthropic.claude-sonnet-4-5-20250929-v1:0` example alongside it if you'd rather show both shapes.

docs/models/anthropic.md
DouweM: Removed the `Vertex vs GoogleModel` note in `e04b9a8e8`. (Left `docs/models/google.md` itself alone — separate cleanup, out of scope for this PR, but worth a follow-up if it still implies `GoogleModel`/`GoogleProvider(vertexai=True)` serves Anthropic models.)

pydantic_ai_slim/pydantic_ai/models/anthropic.py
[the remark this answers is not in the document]
DouweM: Added `AsyncAnthropicBedrockMantle` to `_FAST_MODE_UNSUPPORTED_CLIENTS` (fast mode isn't available on any Bedrock transport). Left it out of `_NON_AUTOMATIC_CACHING_CLIENTS` because the "Claude in Amazon Bedrock" docs list prompt caching as supported — Mantle uses the Messages API, not InvokeModel. And it's deliberately excluded from `_BM25_TOOL_SEARCH_UNSUPPORTED_CLIENTS` (Mantle's "not supported" list doesn't cover tool search). `e04b9a8e8`.

PR #5398, opened 2026-05-13
fix(openrouter): honor thinking=False across all routed models

### LINKED ISSUE #5379: `thinking=False` is silently dropped for several reasoning models (OpenRouter, xAI, Bedrock)
### Initial Checks

### Description

`model_settings={'thinking': False}` is silently dropped — no exception, no warning — for several reasoning models routed through `OpenRouterModel`. The HTTP body sent to OpenRouter contains no `reasoning` field at all, so OpenRouter defers to the upstream model's default (which for newer hybrid models is "on"). The user's off-signal is ignored.

Captured wire bodies for the same `OpenRouterModel(...)` agent, varying only the setting:

| Model | `thinking=True` | `thinking=False` |
|---|---|---|
| `moonshotai/kimi-k2.6` | _no reasoning field_ | _no reasoning field_ |
| `qwen/qwen3-235b-a22b` | _no reasoning field_ | _no reasoning field_ |
| `z-ai/glm-4.6` | _no reasoning field_ | _no reasoning field_ |
| `openai/gpt-oss-120b` | _no reasoning field_ | _no reasoning field_ |
| `anthropic/claude-sonnet-4.5` | `reasoning={'effort': 'medium'}` | _no reasoning field_ |
| `x-ai/grok-3-mini` | `reasoning={'effort': 'medium'}` | _no reasoning field_ |

Two distinct code paths produce the same observable result:

1. **Profile gate at `models/__init__.py:710`** strips `thinking` from `model_settings` unless `profile.supports_thinking` or `thinking_always_enabled` is true. The following profiles don't declare these for hybrid reasoning models that need them: `profiles/moonshotai.py`, `profiles/qwen.py`, `profiles/zai.py`, and the gpt-oss branch of `profiles/openai.py`.

2. **Transformer skip at `models/openrouter.py:561`** — `if thinking is not False:` emits nothing when the value *is* `False`. This is what bites Claude / Grok 3 Mini on OpenRouter today, and would also bite the four profile-layer models once their profiles are corrected. The same `if thinking is not False:` pattern appears at `models/xai.py:719` and `models/bedrock.py:618` (`bedrock.py:615` has a comment suggesting it may be intentional there).

Workaround we're shipping downstream: pre-set `openrouter_reasoning={'enabled': False}` when we detect `thinking=False` for an OpenRouter model. This satisfies the `'openrouter_reasoning' not in model_settings` precondition at line 559, and the existing emit at line 575 lands `extra_body.reasoning = {'enabled': False}` on the wire correctly.

```Python
import asyncio, json, os
import httpx
from pydantic_ai import Agent
from pydantic_ai.settings import ModelSettings

class StopHere(Exception): pass

async def capture(model_id, settings):
    captured = {}
    async def intercept(self, request):
        captured['body'] = json.loads(request.content)
        raise StopHere()
    httpx.AsyncHTTPTransport.handle_async_request = intercept
    model = OpenRouterModel(model_id, provider=OpenRouterProvider(api_key=os.environ['OPENROUTER_API_KEY']))
    try:
        await Agent(model, model_settings=settings).run('hi')
    except StopHere:
        pass
    return captured.get('body', {})

async def main():
    for model_id in ['moonshotai/kimi-k2.6', 'anthropic/claude-sonnet-4.5', 'x-ai/grok-3-mini']:
        on  = await capture(model_id, ModelSettings(thinking=True))
        off = await capture(model_id, ModelSettings(thinking=False))
        sig = lambda b: b.get('reasoning') or b.get('reasoning_effort') or '<none>'
        print(f'{model_id:40}  on={sig(on)}  off={sig(off)}')

asyncio.run(main())
```

### Logfire Trace

https://logfire-us.pydantic.dev/public-trace/5f572654-39ea-40f0-ace9-1b15ede7a257?spanId=cbdb756f001c3bac

- Python: 3.11.11
- Pydantic AI: 1.93.0
- LLM provider SDK: openai 2.36.0 (used by `OpenRouterModel`)

sarth6:
- Closes #5379

## Impact

Multi-turn OpenRouter agents that pass `thinking=False` were having their off-signal silently dropped: the HTTP body sent to OpenRouter contained no `reasoning` field at all, so OpenRouter deferred to the upstream model's default — which for newer hybrid models is *on*. Wire bodies, varying only the unified setting:

| Model | `thinking=True` (before / after) | `thinking=False` (before / after) |
|---|---|---|
| `moonshotai/kimi-k2.6` | _none_ → `effort='medium'` | _none_ → `enabled=False` |
| `qwen/qwen3-235b-a22b` | _none_ → `effort='medium'` | _none_ → `enabled=False` |
| `z-ai/glm-4.6` | _none_ → `effort='medium'` | _none_ → `enabled=False` |
| `openai/gpt-oss-120b` | _none_ → `effort='medium'` | _none_ → `enabled=False` |
| `anthropic/claude-sonnet-4.5` | `effort='medium'` (unchanged) | _none_ → `enabled=False` |
| `x-ai/grok-3-mini` | `effort='medium'` (unchanged) | _none_ → `enabled=False` |

## Cause

Two distinct paths drop the off-signal:

1. **Profile gate** (`models/__init__.py` — `prepare_request`) strips `thinking` from settings when the merged model profile reports `supports_thinking=False`. For hybrid reasoning models routed through OpenRouter (kimi-k2.6, qwen3-235b, glm-4.6, gpt-oss-120b) the gate fires because their intrinsic profiles don't declare reasoning support — even though OpenRouter itself accepts the reasoning passthrough universally.

2. **Transformer guard** (`_openrouter_settings_to_openai_settings`) had `if thinking is not False:` around the emit, so the off-signal produced no wire field at all. (Same `is not False` pattern lives in `models/xai.py` and `models/bedrock.py` — see "Out of scope" below.)

## Fix

1. `OpenRouterProvider.model_profile()` now sets `supports_thinking=True` on the base profile so the gate passes `thinking` through regardless of the routed model's intrinsic profile. `ModelProfile.update()` won't downgrade this because `supports_thinking=False` is the default and `update()` only copies non-default fields from the upstream profile (`profiles/__init__.py:148-154`).
2. The transformer emits `reasoning={'enabled': False}` on the `thinking is False` branch — the canonical OpenRouter disable signal, mirroring the existing Cerebras `disable_reasoning=True` pattern.

Together these restore the user's intent on the wire for all 6 reporter rows.

## Out of scope (intentional)

The reporter noted that the same `if thinking is not False:` pattern appears in `models/xai.py` and `models/bedrock.py`. Those stay as-is because the underlying APIs have no way to express "off":

- **xAI direct**: `reasoning_effort` only accepts `'low'` / `'high'` — there's no `'none'`. Silent-degrade per `models/AGENTS.md` ("document unsupported model settings in docstrings and silently ignore at runtime").
- **Bedrock-OpenAI variant**: Converse rejects `reasoning_effort='none'` (existing source comment at `models/bedrock.py:656`). Silent-degrade.
- **Bedrock-Anthropic variant**: already handles `False` correctly via `thinking={'type': 'disabled'}`.
- **Bedrock-Qwen variant**: not verified by the reporter; left untouched.

If maintainers want me to extend any of those in this PR (e.g. emit something on Bedrock-Qwen disable, or revisit the xAI degrade), happy to follow up.

## Tests

- `tests/test_thinking.py::TestOpenRouterThinkingTranslation::test_thinking_false_emits_enabled_false` — was previously asserting the buggy behavior; flipped to assert `reasoning={'enabled': False}`.
- New `TestOpenRouterThinkingTranslation::test_provider_profile_passes_thinking_through_gate` — pins `supports_thinking=True` on the four profile-gated model names from the reporter table, guarding against regression if someone later removes the provider-level flag.
- `tests/test_thinking.py` (103 tests) + `tests/models/test_openrouter.py` (42 tests) all green; `tests/test_examples.py -k thinking` (15 tests) all green; adjacent providers (`test_xai.py` + `test_openai.py`, 308 tests) unaffected.

### Checklist

### DIFF
--- a/docs/thinking.md
+++ b/docs/thinking.md
@@ -45,7 +45,7 @@ The unified `thinking` setting maps to each provider's native format:
 | Google (Gemini 3+) | `include_thoughts=True` | `thinking_level='HIGH'` | |
 | Google (Gemini 2.5) | `include_thoughts=True` | `thinking_budget=24576` | |
 | Groq | `reasoning_format='parsed'` | `reasoning_format='parsed'` | `thinking=False` → `'hidden'` (no true disable) |
-| OpenRouter | `reasoning.effort='medium'` | `reasoning.effort='high'` | Via `extra_body` |
+| OpenRouter | `reasoning.effort='medium'` | `reasoning.effort='high'` | `thinking=False` → `reasoning.enabled=False`; via `extra_body` |
 | Cerebras | `disable_reasoning=False` | `disable_reasoning=False` | `thinking=False` → `disable_reasoning=True` |
 | xAI | `reasoning_effort='high'` | `reasoning_effort='high'` | Only `'low'` and `'high'` |
 | Bedrock (Claude) | `thinking.type='enabled'` | `budget_tokens=16384` | No adaptive support |
@@ -270,6 +270,9 @@ settings = OpenRouterModelSettings(openrouter_reasoning={'effort': 'high'})
 agent = Agent(model, model_settings=settings)
 ...
 ```
+
+!!! note "Disabling reasoning"
+    Setting [`thinking=False`][pydantic_ai.settings.ModelSettings.thinking] via the unified setting sends `reasoning={'enabled': False}` in the request body, which OpenRouter forwards as the disable signal to the underlying model. This is honored across providers that expose a way to turn reasoning off (e.g. hybrid models like `anthropic/claude-sonnet-4.5` or `moonshotai/kimi-k2.6` whose default is to reason); for models without a disable mechanism, OpenRouter silently no-ops.
 
 ## Mistral
 
--- a/pydantic_ai_slim/pydantic_ai/models/openrouter.py
+++ b/pydantic_ai_slim/pydantic_ai/models/openrouter.py
@@ -558,8 +558,12 @@ def _openrouter_settings_to_openai_settings(
     # Fall back to unified thinking when openrouter_reasoning is not set
     if 'openrouter_reasoning' not in model_settings and model_request_parameters.thinking is not None:
         thinking = model_request_parameters.thinking
-        if thinking is not False:
-            unified_reasoning: OpenRouterReasoning = {}
+        unified_reasoning: OpenRouterReasoning = {}
+        if thinking is False:
+            # Explicit disable: OpenRouter routes this to the underlying model's
+            # native disable mechanism where one exists; otherwise it silently no-ops.
+            unified_reasoning['enabled'] = False
+        else:
             # OpenRouter only supports low/medium/high; map others to closest
             effort_map: dict[ThinkingLevel, str] = {
                 True: 'medium',
@@ -570,7 +574,7 @@ def _openrouter_settings_to_openai_settings(
                 'xhigh': 'high',
             }
             unified_reasoning['effort'] = effort_map[thinking]  # type: ignore[typeddict-item]
-            model_settings['openrouter_reasoning'] = unified_reasoning
+        model_settings['openrouter_reasoning'] = unified_reasoning
 
     if reasoning := model_settings.pop('openrouter_reasoning', None):
         extra_body['reasoning'] = reasoning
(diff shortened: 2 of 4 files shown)


sarth6:
Auto-closed by branch rename — re-filed as #5433 with the same OpenRouter fix plus xAI and Bedrock OpenAI / Qwen coverage.

PR #5414, opened 2026-05-13
fix: warn instead of silently swallowing NameError in get_first_param…

### LINKED ISSUE #5358: takes_run_context() silently swallows NameError on TYPE_CHECKING-hidden RunContext, causing silent dispatch failure
## Summary

`takes_run_context()` silently swallows `NameError` from string-annotation resolution, causing `ProcessHistory(fn)` to route to the no-context dispatch branch when `RunContext` is imported only under `TYPE_CHECKING:`. This produces a hard `TypeError` on the first invocation, with no warning at registration time and no diagnostic that points at the root cause.

This pattern is easy to introduce (a sleepy `from typing import TYPE_CHECKING` block is the canonical Python typing idiom) and devastating when introduced — it took down every conversational query in our production backend for ~58 minutes when we hit it (192 errors before someone noticed). The fix, once diagnosed, was a one-line import lift.

## Repro

```python
# my_processor.py
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from pydantic_ai import RunContext


async def my_processor(
    ctx: RunContext[None],
    messages: list[ModelMessage],
) -> list[ModelMessage]:
    return messages
```

```python
from pydantic_ai._utils import takes_run_context
from my_processor import my_processor

print(takes_run_context(my_processor))  # → False (should be True)
```

Wire it into `Agent(history_processors=[my_processor])` and every turn crashes with:

```
TypeError: my_processor() missing 1 required positional argument: 'messages'
```

## Root cause

[`pydantic_ai/_utils.py:497-550`](https://github.com/pydantic/pydantic-ai/blob/main/pydantic_ai/_utils.py#L497) — `get_first_param_type` calls `_typing_extra.get_function_type_hints` to resolve the first param's annotation, then swallows `(NameError, TypeError, AttributeError)`:

```python
try:
    type_hints = _typing_extra.get_function_type_hints(_decorators.unwrap_wrapped_function(callable_for_hints))
except (NameError, TypeError, AttributeError):
    return None
```

With `from __future__ import annotations` (PEP 563) + `RunContext` imported only under `TYPE_CHECKING`, the resolver raises `NameError`, falls through to `return None`, and `takes_run_context()` returns False. `ProcessHistory._run_history_processor` then dispatches via the no-context branch (`processor(messages)`), leaving the second positional unbound.

The exception swallow is *intentional* — it's a forward-reference safety net so resolution failures don't crash registration. That's a reasonable default, but the silent fallback to `False` (rather than, say, a warning, or a typed exception, or a `strict=` opt-in) makes this class of bug essentially undetectable until first invocation.

## Suggested fixes (any of these would have prevented the outage)

1. **Emit a warning** when `get_function_type_hints` raises during a `takes_run_context()` check. The function still returns False, but the user sees a diagnostic at registration time pointing at the unresolvable annotation:
   ```python
   except (NameError, TypeError, AttributeError) as e:
       warnings.warn(
           f"Could not resolve type hints for {callable_obj!r}: {e}. "
           f"If the first parameter is annotated as RunContext under "
           f"`if TYPE_CHECKING:`, lift the import to module scope.",
           UserWarning,
           stacklevel=2,
       )
       return None
   ```

2. **Add `strict=` opt-in** to `takes_run_context()` (and propagate to `ProcessHistory`/tool registration) — when set, re-raise the underlying exception instead of silently returning False. Users who opt in get fail-fast registration; default behavior unchanged.

3. **Fail fast at `ProcessHistory.__post_init__`** — if the registered callable's first parameter is *named* `ctx` (or matches some heuristic) but `takes_run_context()` returns False because the resolver raised, raise `ValueError` at registration time. Most users name their context param `ctx`, so the false-positive rate would be low.

(1) is the least invasive and would have given us a deploy-time signal. (3) would have caught this at app startup. (2) is the cleanest API but the largest API surface change.

## Why not just fix downstream?

We did. The site we lost prod on is fixed (PR #1126 in our codebase), and we added an AST canary that catches the pattern at commit time on registered `ProcessHistory(...)` callables in our codebase (PR #1232/#1234). But:

- Our AST canary only protects our codebase. Every other Pydantic AI user is one `TYPE_CHECKING` block away from the same outage.
- A static-AST guard is a workaround for the silent-failure pattern, not a fix. The right place for the fix is inside Pydantic AI's introspection.

## Provenance

- Our outage: 192 errors over 58 minutes, 2026-05-09 21:59:34Z–22:57:46Z UTC.
- Hotfix PR in our codebase: orchestrated-bio/insight-next#1126 (import lift).
- Downstream AST canary: orchestrated-bio/insight-next#1232 / #1234.
- Pydantic AI version: 1.90.0 (also present on `main` HEAD as of this filing).

Happy to send a PR for option (1) if there's interest.

Quratulain-bilal:
Fixes #5358.

  ## Problem

  `get_first_param_type` catches `NameError` from `get_function_type_hints` and silently returns `None`.
  With `from __future__ import annotations` plus a `TYPE_CHECKING`-only import of `RunContext`, the
  `RunContext[None]` annotation can't be resolved at runtime, the `NameError` is swallowed, and
  `takes_run_context` returns `False`. The history processor is then called without a ctx argument,
  raising `TypeError: processor() missing 1 required positional argument: 'messages'` — far from the real
  cause. The reporter lost 58 minutes of production diagnosing this.

  ## Fix

  Surface the `NameError` as a `UserWarning` that names the unresolved annotation and points at
  `TYPE_CHECKING` as the likely cause. `TypeError` and `AttributeError` stay silent because they cover
  legitimate "can't introspect" callables (e.g. the existing
  `test_takes_ctx_returns_false_for_untyped_processor` case).

  The return value is unchanged (`None`), so callers behave identically — but the failure is now
  observable in logs / `pytest.warns` / `warnings.filterwarnings("error", ...)` instead of disappearing.

  ## Why a warning rather than re-raising

  Re-raising would break callables whose first-param annotation is unresolvable for unrelated reasons. A
  warning preserves the existing contract while making the silent fallback observable, which is the
  smallest change that solves the reporter's debugging problem.

  ## Verification

  - `uv run pytest tests/test_utils.py tests/test_history_processor.py` — 57 passed
  - `uv run ruff check pydantic_ai_slim/pydantic_ai/_utils.py tests/test_utils.py` — clean
  - `uv run ruff format --check ...` — clean
  - New test `test_get_first_param_type_warns_on_unresolvable_annotation` reproduces the issue using a
  synthesized module so the bug is exercised in CI without a separate fixture file.


adtyavrdhn:
Superseded by #6958, which fixed the same unresolved-annotation root cause on both `ProcessHistory` and `FallbackModel(fallback_on=...)` and now raises a focused `UserError`. Keeping this warning-based patch would duplicate that merged behavior while providing a weaker failure mode, so I am closing this PR rather than widening or rebasing it.

Issue #3970, opened 2026-01-09
Track `ty` type checker readiness for pydantic-ai


dsfaccini:
## Summary

We evaluated [ty](https://github.com/astral-sh/ty) (Astral's new Python type checker) as a potential replacement for pyright. We found 417 diagnostics, of which 337 are false positives (81%).

This issue tracks the upstream ty issues that block adoption.

## Investigation Results

| Pattern | Errors | Upstream Issue | ty Correct? | Action |
|---------|--------|----------------|-------------|--------|
| PEP 696 TypeVar defaults | 89 | [ty#592](https://github.com/astral-sh/ty/issues/592) | No | Wait |
| ParamSpec + Concatenate | 22 | [ty#1535](https://github.com/astral-sh/ty/issues/1535) | No | Wait |
| TypedDict `in` narrowing | 35 | [ty#154](https://github.com/astral-sh/ty/issues/154) (PEP 728 `closed=True`) | **Yes** | Wait for PEP 728 |
| Dict literals vs TypedDict | 25+ | [ty#2265](https://github.com/astral-sh/ty/issues/2265) | No | Wait |
| Generic type propagation | 17 | [ty#2030](https://github.com/astral-sh/ty/issues/2030), [ty#2182](https://github.com/astral-sh/ty/issues/2182) | No | Wait |
| `**kwargs` type tracking | 84 | [ty#1332](https://github.com/astral-sh/ty/issues/1332) | No | Wait |

### High Priority (80+ errors each)
- [ ] https://github.com/astral-sh/ty/issues/592 - PEP 696 TypeVar defaults (89 errors)
- [x] https://github.com/astral-sh/ty/issues/1332 - `**kwargs` unpacking with TypedDict (84 errors)

### Medium Priority (20-35 errors each)
- [x] https://github.com/astral-sh/ty/issues/2265 - Dict literals assignable to TypedDict unions (25+ errors)
- [x] https://github.com/astral-sh/ty/issues/1535 - Concatenate special form support (22 errors)
- [ ] https://github.com/astral-sh/ty/issues/154 - PEP 728 `closed=True` for TypedDict (35 errors - ty is correct here)

### Lower Priority (17 errors)
- [x] https://github.com/astral-sh/ty/issues/2030 - Generic inference for `@contextmanager`
- [x] https://github.com/astral-sh/ty/issues/2182 - Nested generic classes with defaults

## Notes

[...]

PR #5428, opened 2026-05-13
Improve `ty` compatibility in `pydantic_ai_slim`

dsfaccini:
As per our team discussion on Thu 14 of May, we won't be merging this PR bc it would increase maintenance load (by urging contributors and team member to address `ty` specific issues, that we don't currently have a standard for, so we'd need to come up with some rules first...)

Instead I, David, will be responsible of keeping this branch up to date and addressing any new diagnostics that come up where appropriate, i.e., as a change in `pydantic-ai` when the diagnostic is valid, or as a change in the fork if the diagnostic stems from a functionality gap in `ty`

---

- Closes N/A

This improves compatibility with Astral's Ty type checker across `pydantic_ai_slim` while keeping Pyright clean. The changes are mostly local type-shape cleanups, narrow Ty suppressions for dynamic runtime typing patterns, and small helper annotations where Ty currently loses precision.

Related: #3970 tracks Ty readiness for pydantic-ai.

This PR also adds an advisory Ty workflow that builds a pinned commit from David's Ty fork ([`dsfaccini/ruff`](https://github.com/dsfaccini/ruff/tree/codex/ty-pydantic-compat)) and runs it against `pydantic_ai_slim` in parallel with the existing required Pyright checks. The workflow is allowed to fail, so it can track compatibility drift without blocking regular development while Ty is still unstable.

Verified locally:

- `/Users/david/projects/forks/ruff/target/debug/ty check --python-version 3.13 --output-format concise pydantic_ai_slim`
- `uv run --frozen pyright <modified python files>`

### Checklist

### DIFF
--- a/scripts/typecheck_ty.sh
+++ b/scripts/typecheck_ty.sh
@@ -1,0 +1,20 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+if [[ -z "${TY_BIN:-}" ]]; then
+  if command -v ty >/dev/null 2>&1; then
+    TY_BIN=ty
+  else
+    echo "Set TY_BIN to a ty executable, for example a build of dsfaccini/ruff@codex/ty-pydantic-compat." >&2
+    exit 127
+  fi
+fi
+
+TY_OUTPUT_FORMAT="${TY_OUTPUT_FORMAT:-concise}"
+TY_PATHS="${TY_PATHS:-pydantic_ai_slim/pydantic_ai}"
+read -r -a ty_paths <<< "$TY_PATHS"
+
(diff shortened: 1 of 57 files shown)


scripts/typecheck_ty.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+if [[ -z "${TY_BIN:-}" ]]; then
+  if command -v ty >/dev/null 2>&1; then
+    TY_BIN=ty
```
dsfaccini, line 6: this won't work, it would need the same fork installation that has the patches

Issue #5400, opened 2026-05-13
Un-deprecate `Agent.instrument` setter and `InstrumentedModel` — used by `logfire.instrument_pydantic_ai`


DouweM:
## Context

PR #4967 deprecated several `instrument`-related surfaces in favor of `capabilities=[Instrumentation(...)]`. Two of those deprecations conflict with the official `logfire.instrument_pydantic_ai(obj)` integration; the rest are fine.

`Agent.instrument_all(...)` (static class-level default) was kept undeprecated.

## Problem

The official `logfire` SDK uses both `Agent.instrument` and `InstrumentedModel` in the same integration entry point:

```python
# pydantic/logfire: logfire/_internal/integrations/pydantic_ai.py:60-66
def instrument_pydantic_ai(..., obj: Agent | Model | None, ...) -> None | InstrumentedModel:
    ...
    if isinstance(obj, Agent):
        obj.instrument = settings                    # ← writes to deprecated property
    elif isinstance(obj, Model):
        return InstrumentedModel(obj, settings)      # ← constructs deprecated class
    elif obj is None:
        Agent.instrument_all(settings)               # ← fine, not deprecated
```

With PR #4967 shipped, every user calling `logfire.instrument_pydantic_ai(my_agent)` or `logfire.instrument_pydantic_ai(my_model)` (both documented patterns) sees a visible `PydanticAIDeprecationWarning` (a `UserWarning` subclass, on by default) on every run. The function's return type also references `InstrumentedModel`, so anyone type-checking against the public signature sees `reportDeprecated` at the call site.

## Why we can't just fix it in logfire

We ship logfire changes easily, but logfire's pattern operates on *existing* objects and Pydantic AI doesn't expose clean replacements for either case:

- **`Agent` branch:** logfire holds an `Agent` reference and wants to enable instrumentation on it. The clean migration would be `agent.<add capability>(Instrumentation(settings))`, but Pydantic AI has no public capability-attach API. `Agent(capabilities=[...])` is construction-time only; `Agent.override(...)` doesn't accept `capabilities=`; `agent._root_capability` is private.
- **`Model` branch:** logfire wraps a `Model` and returns it to the user, who then passes it elsewhere (`Agent(model=...)`, `direct.model_request*`). There's no model-level capability equivalent of `InstrumentedModel` — the `Instrumentation` capability is agent-level by design.

So logfire's integration can't be migrated to capabilities-only without API additions in Pydantic AI first.

## Options

[...]

PR #5429, opened 2026-05-13
refactor: dedupe model-request OTel span between `Instrumentation` and `InstrumentedModel`

DouweM:
- Follow-up to #5427.

## Summary

Both paths were independently building the same `chat <model>` CLIENT span: same attributes (operation, model, request parameters, baggage, tool definitions, model settings), same `finish(response)` closure, same cost/metrics ordering. Pre-#4967 this was fine because `InstrumentedModel` was the only path. #4967 introduced `Instrumentation.wrap_model_request` and planned to drop `InstrumentedModel` in v2; #5427 then un-deprecated `InstrumentedModel` because the official Logfire SDK uses it, making the duplication permanent.

This PR extracts the shared logic into `_instrumentation.open_model_request_span` (a `@contextmanager` yielding a `finish(response)` callback). Both call sites collapse to:

```python
prepared_settings, prepared_parameters = model.prepare_request(...)
with open_model_request_span(settings, model, messages, prepared_settings, prepared_parameters) as finish:
    response = await handler(...)
    annotate_tool_call_otel_metadata(response, prepared_parameters)
    finish(response)
    return response
```

Drops `InstrumentedModel._instrument`. Net -81 lines.

## Why this shape (and not option (ii))

#5400 sketched two dedupe options:
- **(i) Extract a shared helper** — chosen.
- **(ii) Have `InstrumentedModel.request` route through `Instrumentation.wrap_model_request` with a synthesized `RunContext`**.

The agent-side `wrap_model_request` happens to never reference its `ctx` parameter, but synthesizing a stub `RunContext` just to satisfy the signature would be carrying dead structure through a hot path. The free-function helper has the same dedupe benefit without that awkwardness, and keeps the agent-flow-specific bits (`_last_messages` / `_last_model_request_parameters` state for `_run_span_end_attributes`) where they belong — in `Instrumentation`, not leaking into the helper.

### Checklist

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/_instrumentation.py
+++ b/pydantic_ai_slim/pydantic_ai/_instrumentation.py
@@ -1,14 +1,16 @@
 from __future__ import annotations
 
 import itertools
-from collections.abc import Mapping, Sequence
+import warnings
+from collections.abc import Callable, Iterator, Mapping, Sequence
+from contextlib import contextmanager
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Any, ClassVar, Literal
+from typing import TYPE_CHECKING, Any, ClassVar, Literal, cast
 from urllib.parse import urlparse
 
 from opentelemetry._logs import LogRecord
 from opentelemetry.baggage import get_baggage
-from opentelemetry.trace import INVALID_SPAN, get_current_span
+from opentelemetry.trace import INVALID_SPAN, SpanKind, get_current_span
 from opentelemetry.util.types import AttributeValue
 from pydantic import TypeAdapter
 from pydantic_core import to_json
@@ -20,6 +22,8 @@ if TYPE_CHECKING:
 
     from pydantic_ai.messages import ModelMessage, ModelResponse
     from pydantic_ai.models import Model, ModelRequestParameters
+    from pydantic_ai.models.instrumented import InstrumentationSettings
+    from pydantic_ai.settings import ModelSettings
 
 DEFAULT_INSTRUMENTATION_VERSION = 2
 """Default instrumentation version for `InstrumentationSettings`."""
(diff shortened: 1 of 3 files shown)


pydantic_ai_slim/pydantic_ai/_instrumentation.py
```diff
@@ -170,6 +174,117 @@ def build_tool_definitions(model_request_parameters: ModelRequestParameters) ->
     return tool_definitions
 
 
+@contextmanager
+def open_model_request_span(
+    settings: InstrumentationSettings,
+    model: Model,
+    messages: list[ModelMessage],
+    prepared_settings: ModelSettings | None,
+    prepared_parameters: ModelRequestParameters,
```
DouweM: I'd rather take `ModelRequestContext` than these 3 args separately

pydantic_ai_slim/pydantic_ai/capabilities/instrumentation.py
DouweM, line 1: *(Refers to line 264)*

Do we need to prepare here? Would this not have already been done before `wrap_model_request` is called?

pydantic_ai_slim/pydantic_ai/_instrumentation.py
DouweM: Done in f36022794 — helper now takes `ModelRequestContext` and yields `(finish, prepared_parameters)`. Both call sites collapsed to ~4 lines each.

pydantic_ai_slim/pydantic_ai/capabilities/instrumentation.py
DouweM: No — `_agent_graph._make_request` builds the `ModelRequestContext` with raw `model_settings` / `model_request_parameters` (from `_prepare_request_parameters` and `get_model_settings`). `Model.prepare_request` runs *inside* `model.request()` itself (per its own docstring). The instrumentation helper needs the prepared values to compute accurate attributes (merged settings overlay, `customize_request_parameters`, native_tool dedupe, thinking resolution), so f36022794 moves the `prepare_request` call into the helper. Yes, that means `prepare_request` runs twice per agent model request (once in the helper for attributes, once inside `model.request()`) — same as before this PR, just consolidated. Hoisting it upstream so it runs once would require teaching `Model.request` to skip its internal call, which is a bigger refactor.

pydantic_ai_slim/pydantic_ai/capabilities/instrumentation.py
```diff
@@ -259,116 +251,20 @@ async def wrap_model_request(
...
-                    'properties': {'model_request_parameters': {'type': 'object'}},
-                }
-            ).decode(),
-        }
+        with open_model_request_span(self.settings, request_context) as (finish, prepared_parameters):
+            # Stash for `_run_span_end_attributes`: feeding the parameters into
+            # `get_instructions` lets it use the canonical `instruction_parts` source
+            # (which includes prompted-output template instructions and is properly sorted)
+            # instead of falling back to reading `ModelRequest.instructions` from history.
+            self._last_model_request_parameters = prepared_parameters
```
DouweM: Could we yield the updated `request_context` and get params off of that?

pydantic_ai_slim/pydantic_ai/models/instrumented.py
```diff
@@ -369,14 +361,15 @@ async def request(
...
-            model_request_parameters,
+        request_context = ModelRequestContext(
+            model=self.wrapped,
+            messages=messages,
+            model_settings=model_settings,
+            model_request_parameters=model_request_parameters,
         )
-        with self._instrument(messages, prepared_settings, prepared_parameters) as finish:
+        with open_model_request_span(self.instrumentation_settings, request_context) as (finish, _):
             response = await self.wrapped.request(messages, model_settings, model_request_parameters)
```
DouweM: Should we not be using the prepared settings and params here? If `open_model_request_span` yields the updated MRC, we could easily.

I know that we re-run the prepare method internally, so it doesn't really matter, but right now it reads confusing that we don't use the prepared version, right?

pydantic_ai_slim/pydantic_ai/capabilities/instrumentation.py
DouweM: Done in 6cc1333f0 — helper yields `(finish, prepared_request_context)`. Caller reads `prepared_request_context.model_request_parameters` off it.

pydantic_ai_slim/pydantic_ai/models/instrumented.py
DouweM: Done in 6cc1333f0 — `InstrumentedModel.request`/`request_stream` now use `prepared_rc.messages/.model_settings/.model_request_parameters` when calling `self.wrapped`. Reads consistently now.

Issue #1569, opened 2025-04-22
Possible inconsistency with schema for Gemini REST interface


gitraffe:
### Question

The [Gemini docs](https://ai.google.dev/gemini-api/docs/function-calling?example=weather) indicate that functions should be declared as follows.

```json
"tools": [
  {
    "functionDeclarations": [
      {
        "name": "get_current_temperature",
        "description": "Gets the current temperature for a given location.",
        "parameters": {
          "type": "object",
          "properties": {
            "location": {
              "type": "string",
              "description": "The city name, e.g. San Francisco"
            }
          },
          "required": ["location"]
        }
      }
    ]
  }
]
```

The current client implementation instead passes the key `function_declarations` to `httpx.AsyncClient.stream` here.

https://github.com/pydantic/pydantic-ai/blob/35d6ed49dded87d69c053cb8deb9874b2bfa8d57/pydantic_ai_slim/pydantic_ai/models/gemini.py#L238

There is an alias declared here, but it doesn't seem to be respected in the serialization.

https://github.com/pydantic/pydantic-ai/blob/35d6ed49dded87d69c053cb8deb9874b2bfa8d57/pydantic_ai_slim/pydantic_ai/models/gemini.py#L644

Here's a demo.

```python
import httpx
import pydantic_ai
from unittest.mock import patch


original = httpx.AsyncClient.stream


def patched_stream(self, *args, **kwargs):
    print(kwargs["content"].decode())
    return original(self, *args, **kwargs)


agent = pydantic_ai.Agent("google-gla:gemini-2.0-flash")


@agent.tool_plain
def get_animal_size(animal: str):
    """Get the size of an animal."""
    return 99


with patch("httpx.AsyncClient.stream", patched_stream):
    print(agent.run_sync(user_prompt="How big is an elephant."))
```

```bash
$ python/gemini.py
{"contents":[{"role":"user","parts":[{"text":"How big is an elephant."}]}],"tools":{"function_declarations":[{"name":"get_animal_size","description":"Get the size of an animal.","parameters":{"properties":{"animal":{"type":"string"}},"required":["animal"],"type":"object"}}]}}
[...]

PR #5441, opened 2026-05-14
Map AG-UI interrupts to `DeferredTools` in `AGUIAdapter`

### LINKED ISSUE #2889: Make tool call approval work with AG-UI
### Description

See https://ai.pydantic.dev/deferred-tools/#human-in-the-loop-tool-approval

As discussed [on Slack](https://pydanticlogfire.slack.com/archives/C083V7PMHHA/p1757702029442509?thread_ts=1757690478.953339&cid=C083V7PMHHA):

It would be great if AG-UI had a native flow for tool call approval. 

In the meantime, a workaround could be to send an event to the frontend holding the `DeferredToolRequests` (i.e. tool names, call IDs, and args), then have the frontend store the `DeferredToolResults` on the `state` (i.e. tool call IDs mapping to booleans), and then to intercept those in your run-agent endpoint to pass them in as `deferred_tool_results`.

But that can't currently easily be implemented yet, as you'd need to have access to the event bus from the agent run `on_complete` handler (to get the `DeferredToolRequests`), and you'd need to be able to pass `deferred_tool_results` to the `run_ag_ui` helper method which currently doesn't accept it.

It would be nice to implement this at the same time as https://github.com/pydantic/pydantic-ai/issues/2382. 

We could potentially have a special `tool_approval_required` `CustomEvent`, and special support for a `tool_approvals` field on the `deps.state` object, to save having to manage passing both from Pydantic AI to AG-UI and back manually. But ideally, AG-UI would have a native way to do this.

### References

_No response_

jjmllr:
Closes #2889

## Summary

Translates Pydantic AI's `DeferredToolRequests` ↔ AG-UI's new [interrupt-aware run lifecycle](https://docs.ag-ui.com/concepts/interrupts) (ag-ui-protocol [PR #1569](https://github.com/ag-ui-protocol/ag-ui/pull/1569)) in both directions, behind a version gate:

- **Outbound.** When a run ends with `DeferredToolRequests.approvals`, `AGUIEventStream.after_stream` now emits `RUN_FINISHED` carrying `outcome = RunFinishedInterruptOutcome(interrupts=[...])`. Each `Interrupt` has `reason='tool_call'`, the original `tool_call_id`, a deterministic `id = f"int-{tool_call_id}"`, an approval-shaped `response_schema`, and any per-call `metadata` from `DeferredToolRequests.metadata`. Normal runs emit `RunFinishedSuccessOutcome` instead. No `outcome` field is set when the negotiated AG-UI version is older than `0.1.19`.
- **Inbound.** `AGUIAdapter.deferred_tool_results` now reads `RunAgentInput.resume[]` and produces a `DeferredToolResults`, mapping each `ResumeEntry` to `ToolApproved` / `ToolApproved(override_args=…)` / `ToolDenied(...)` per the spec. The bridge is **deny-by-default**: only an explicit `payload.approved == True` approves; missing/null/non-bool/non-dict payloads all deny so a malformed client cannot bypass `requires_approval=True`.

The agent-side primitive (`@tool(requires_approval=True)` → `DeferredToolRequests` output) already existed — this PR is purely the adapter wiring + tests + docs.

## Version gating & backward compatibility

- `ag-ui-protocol` floor in `pydantic_ai_slim/pyproject.toml` stays at `>=0.1.10`. No bump.
- All new code paths are gated on (a) an import-time `_HAS_INTERRUPTS` flag with stub classes for older SDKs, and (b) a runtime negotiated-version compare against `INTERRUPTS_VERSION = (0, 1, 19)`.
- On the old path, the adapter behaves byte-for-byte as today: bare `RUN_FINISHED` with no `outcome`, `resume[]` ignored even if a client sends one.

## ⚠️ dev pin on `ag-ui-protocol`

`ag-ui-protocol 0.1.19` (which ships [ ag-ui-protocol PR #1569](https://github.com/ag-ui-protocol/ag-ui/pull/1569) is **not yet on PyPI** at the time of this PR. To get CI to type-check and exercise the new types, this PR adds a temporary `[tool.uv.sources]` entry in the **root** `pyproject.toml` pinning `ag-ui-protocol` to commit `df613e4` from `main`, with the lockfile updated to match. The published `>=0.1.10` floor in `pydantic_ai_slim/pyproject.toml` is untouched, so PyPI consumers of `pydantic-ai-slim[ag-ui]` are unaffected.

**Hold the PR until `0.1.19` publishes, drop the `[tool.uv.sources]` line and re-sync. Clean merge.**

**Cleanup checklist**

1. Remove the `ag-ui-protocol = { git = ... }` line from root `pyproject.toml`'s `[tool.uv.sources]` and `uv sync`.
2. Drop the `ag_ui_version='0.1.19'` override in `examples/pydantic_ai_examples/ag_ui/api/tool_approval.py` — a temporary workaround because the unreleased SDK from `main` still reports `version='0.1.18'`. Once the published SDK identifies itself as `0.1.19`, `DEFAULT_AG_UI_VERSION` resolves correctly without it.
3. (Optional) Raise the floor to `>=0.1.19` and drop the `_HAS_INTERRUPTS` import-time gate, simplifying the adapter.

## Tests

`tests/test_ag_ui.py` adds 11 new cases (incl. one parametrized over 7 payload shapes), all guarded by `pytest.mark.skipif(not interrupts_imports_successful(), …)` so they skip cleanly when CI ever sees an SDK older than 0.1.19:

- `RUN_FINISHED` outcome shape: success on modern SDK, absent on legacy SDK, interrupt outcome when the run ends on a pending approval (incl. `id` ↔ `toolCallId` round-trip).
- Resume translation: approve, approve-with-`editedArgs` (→ `ToolApproved.override_args`), deny with `reason`, cancel-overrides-payload.
- Deny-by-default: parametrized over `{approved: missing}`, `{approved: null}`, `{approved: "true"}`, `{approved: 1}`, string payload, list payload, null payload — all resolve to `ToolDenied()`.
- Malformed `interrupt_id` (missing `int-` prefix) → `UserError` rather than silent miscorrelation.
- End-to-end roundtrip: turn 1 produces the interrupt outcome; turn 2 resumes with approval; assertions cover the spec rule that the resumed turn must **not** re-emit `TOOL_CALL_START` for the same `tool_call_id`, only `TOOL_CALL_RESULT`.

**Results:**

- `make typecheck` → 0 errors / 0 warnings
- `uv run pytest tests/test_ag_ui.py` → all pass (existing + new)
- `pytest tests/test_examples.py -k ag_ui` → all pass (the new doc snippet is picked up)
- `ruff check && ruff format` → clean

## Manual verification (reproducible end-to-end)

Tested against `openai:gpt-5-mini` on this branch. Reviewers can reproduce with an OpenAI key.

```bash
export OPENAI_API_KEY=sk-...
uv run python -m pydantic_ai_examples.ag_ui    # serves on :9000, mounts /tool_approval/
```

**Turn 1 — propose a delete:**

```bash
curl -N -X POST http://127.0.0.1:9000/tool_approval/ \
  -H 'Content-Type: application/json' \
  -H 'Accept: text/event-stream' \
  -d '{"threadId":"t1","runId":"r1","state":{},"messages":[{"id":"m1","role":"user","content":"Please delete /tmp/foo.txt"}],"tools":[],"context":[],"forwardedProps":{}}'
```

Stream ends with RUN_FINISHED carrying:
```
{"outcome":{"type":"interrupt","interrupts":[{"id":"int-<TC>","reason":"tool_call","toolCallId":"<TC>","message":"Approve delete_file({\"path\":\"/tmp/foo.txt\"})?","responseSchema":{"...":"..."}}]}}
```

The tool body does not execute.

**Turn 2 — resume with approval (substitute $TC from turn 1):**

```bash
TC=<paste from turn 1>
curl -N -X POST http://127.0.0.1:9000/tool_approval/ \
  -H 'Content-Type: application/json' \
  -H 'Accept: text/event-stream' \
  -d "{\"threadId\":\"t1\",\"runId\":\"r2\",\"state\":{},\"messages\":[{\"id\":\"m1\",\"role\":\"user\",\"content\":\"Please delete /tmp/foo.txt\"},{\"id\":\"m2\",\"role\":\"assistant\",\"toolCalls\":[{\"id\":\"$TC\",\"type\":\"function\",\"function\":{\"name\":\"delete_file\",\"arguments\":\"{\\\"path\\\": \\\"/tmp/foo.txt\\\"}\"}}]}],\"tools\":[],\"context\":[],\"forwardedProps\":{},\"resume\":[{\"interruptId\":\"int-$TC\",\"status\":\"resolved\",\"payload\":{\"approved\":true}}]}"
```

Stream emits a single TOOL_CALL_RESULT against the original <TC> (no fresh TOOL_CALL_START), then the assistant text, then RUN_FINISHED with outcome.type == "success".

Variations: deny → {status:"resolved", payload:{approved:false, reason:"…"}}, cancel → {status:"cancelled"}, approve with edited args → {status:"resolved", payload:{approved:true, editedArgs:{…}}}.

## Out of scope (followups)

- Mapping DeferredToolRequests.calls (external/frontend execution) to interrupts — that path already round-trips through AG-UI's ToolMessage history today.
- Interrupt.reason = "input_required" / "confirmation" — no Pydantic AI source today.
- Per-tool response_schema for editedArgs (currently a permissive object; can be synthesized from ToolDefinition.parameters_json_schema).
- Interrupt.expires_at enforcement.
- Custom Interrupt.message via DeferredToolRequests.metadata (today it's a generated Approve <tool>(<args>)? string).

## Incidental change

`clai/README.md` is automatically regenerated by the `clai-help` pre-commit hook to match `clai --help` output. Running the hook on this branch surfaces a pre-existing drift on `main` (CLI default is `"openai-chat:gpt-5"`, README still says `"openai:gpt-5"`). The diff is a hook-enforced sync, not a manual edit, and is included here because the hook refuses any commit that doesn't bring the README in line with the CLI.

## Checklist
- [ ] No breaking changes in accordance with the [version policy](https://github.com/pydantic/pydantic-ai/blob/main/docs/version-policy.md).
- [ ] PR title is fit for the [release changelog](https://github.com/pydantic/pydantic-ai/releases).


DouweM:
@jjmllr Thank you! I've kicked off an auto-review, but I'm going to hold off on reviewing this more thoroughly until the new `ag-ui-protocol` package is out.

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
```diff
@@ -56,6 +63,44 @@
...
+            """Stub for ag-ui-protocol < 0.1.19 — no instances are constructed when `_HAS_INTERRUPTS` is False."""
+
+        class RunFinishedInterruptOutcome:
+            """Stub for ag-ui-protocol < 0.1.19."""
+
+        class RunFinishedSuccessOutcome:
+            """Stub for ag-ui-protocol < 0.1.19."""
+
+
+INTERRUPT_ID_PREFIX = 'int-'
```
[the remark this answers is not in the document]
jjmllr: moved to utils

docs/ui/ag-ui.md
```diff
@@ -242,6 +242,59 @@ uvicorn ag_ui_state:app --host 0.0.0.0 --port 9000
...
+from starlette.applications import Starlette
+from starlette.requests import Request
+from starlette.responses import Response
+from starlette.routing import Route
+
+from pydantic_ai import Agent
+from pydantic_ai.tools import DeferredToolRequests
+from pydantic_ai.ui.ag_ui import AGUIAdapter
+
+agent = Agent('openai:gpt-5-mini', output_type=[str, DeferredToolRequests])
```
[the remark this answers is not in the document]
jjmllr: updated to openai:gpt-5.2

jjmllr:
@DouweM The new version was just released https://pypi.org/project/ag-ui-protocol/0.1.19/. 
I have updated the import and fixed the two issues flagged by the automatic review. 

CC @adtyavrdhan @dsfaccini

dsfaccini:
hey @jjmllr I'm taking a look, thanks for working on this. I may push to this branch so make sure you've allowed maintainers to push to your fork and disregard any review comments from now on :)

jjmllr:
> hey @jjmllr I'm taking a look, thanks for working on this. I may push to this branch so make sure you've allowed maintainers to push to your fork and disregard any review comments from now on :)

Thanks @dsfaccini ! "Allow edits by maintainers" is on - push away. I'll leave the branch to you and ping if you need anything else!

dsfaccini:
this PR has passed local review, devin and autoreview, needs now to remove the `PLAN.md` and can be merged.

Korred:
Any news on when this will be merged? 🙏🏼

dsfaccini (approved):

dsfaccini:
hey @Korred sorry for the delay here

I'll let @adtyavrdhn have a chance to look at the PR before we merge it, we can get it in tomorrow if there's nothing from his side

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
```diff
@@ -56,6 +64,37 @@
...
+
+if TYPE_CHECKING:
+    from ag_ui.core import (
+        Interrupt,
+        RunFinishedInterruptOutcome,
+        RunFinishedSuccessOutcome,
+    )
+
+    _HAS_INTERRUPTS = True
+else:
```
adtyavrdhn: This reads a bit weird? 

`if TYPE_CHECKING:` is used to prevent circular imports what is up with the else part?

We should be able to use it regardless now?

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
```diff
@@ -56,6 +64,37 @@
...
+    except ImportError:
+        _HAS_INTERRUPTS = False
+
+        class Interrupt:
+            """Stub for ag-ui-protocol < 0.1.19 — no instances are constructed when `_HAS_INTERRUPTS` is False."""
+
+        class RunFinishedInterruptOutcome:
+            """Stub for ag-ui-protocol < 0.1.19."""
+
+        class RunFinishedSuccessOutcome:
```
adtyavrdhn: Do we need the stubs, what is the argument against bumping it?

docs/ui/ag-ui.md
```diff
@@ -242,6 +242,59 @@ uvicorn ag_ui_state:app --host 0.0.0.0 --port 9000
...
 
+### Tool approval (interrupts)
+
+Tools declared with `requires_approval=True` map onto AG-UI's [interrupt-aware run lifecycle](https://docs.ag-ui.com/concepts/interrupts). When the model proposes such a call, the run pauses and the adapter ends the SSE stream with a `RUN_FINISHED` event whose `outcome.type` is `"interrupt"` and whose `outcome.interrupts[]` describes each pending approval. The client renders an approval UI from that list and POSTs the next `RunAgentInput` with a `resume[]` array of `ResumeEntry` items addressing each interrupt.
+
+The mapping the adapter applies (matching the AG-UI Python SDK field names):
+
+| AG-UI direction         | Pydantic AI source / sink                                                                                       |
+| ----------------------- | --------------------------------------------------------------------------------------------------------------- |
+| `Interrupt.reason`      | Always `"tool_call"` for `requires_approval=True` tools                                                         |
```
adtyavrdhn, line 253: I haven't checked if we do this sort of adapter mapping for other sections so I'll let you be the judge of that

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
```diff
@@ -202,6 +223,56 @@ def _user_content_to_input(
...
+
+def _payload_dict(raw: Any) -> dict[str, Any]:
+    """Coerce a `ResumeEntry.payload` (typed `Optional[Any]` in ag-ui-protocol) into a typed dict.
+
+    Anything that isn't a mapping is treated as the empty payload — the adapter's
+    `response_schema` only advertises an object shape, so non-conforming payloads
+    fall through to the unapproved/default path.
+    """
+    if isinstance(raw, dict):
+        return cast('dict[str, Any]', raw)
```
adtyavrdhn: Can we instead try to use Pydantic for round trip checks, I am not sure if there is precedence for that but this doesn't give me a lot of confidence

adtyavrdhn (commented):
A couple of questions I think we can express this better

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
jjmllr: Since I've handed this over, just wanted to leave a thought on the round-trip suggestion in case it's helpful. happy to pick any of this up if it saves you time @dsfaccini 

There's precedent already for this (RunAgentInput/ResumeEntry are Pydantic models). One model could be the source of truth for both legs of the exchange:

```
class _ResumePayload(BaseModel):
    approved: StrictBool = False
    editedArgs: dict[str, Any] | None = None
    reason: str | None = None
```

model_validate() parses the inbound ResumeEntry.payload, and model_json_schema() could feed the outbound Interrupt.response_schema so the two can't drift and both casts + the _payload_dict helper would go away.

Two things I ran into while sketching it out, in case they're useful:

- StrictBool matters here. In Pydantic's default lax mode, {'approved': 1} / {'approved': 'true'} coerce to True, which would silently approve a malformed payload (the opposite of the deny-by-default intent). With StrictBool I checked that all the current ambiguous-payload cases still deny.
- One behavior shift to be aware of: {'approved': True, 'editedArgs': 'bad'} currently approves and ignores the bad edit; strict validation would deny the whole payload instead. Could go either way, but probably worth a test + changelog line whichever direction you land on.

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
jjmllr: I think this one might be a bit different from the circular-import case. Just to add my reasoning behind this: Interrupt & friends only landed in ag-ui 0.1.19, but the floor is (or was?) >=0.1.10, so importing them unconditionally would break on older versions. The try/except is the runtime fallback (stubs + _HAS_INTERRUPTS=False), and the TYPE_CHECKING branch was my attempt to still give pyright the real types so we don't lose narrowing.

dsfaccini: yeahp this goes hand in hand with how we dealt with the first version-split

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
dsfaccini: same as above, we support different versions and pyright needs stubs

it's understandable you lack this context, I've done most of the UI adapter stuff so I know about the split

docs/ui/ag-ui.md
dsfaccini: yeah this was my largest concern too, I'll run the docs preview

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
dsfaccini: yeahp the assessment is useful, thank you

dsfaccini:
thanks for checking out the comments @jjmllr , I'll let claude decide if there's anything we can improve on the docs side and the roundtrip validation, otherwise the implementation looks ready to me

NathanTarbert:
Excited for this!

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
dsfaccini: David's AICA here: resolving — this is our established AG-UI version-split pattern. `Interrupt` & friends only landed in `ag-ui-protocol` 0.1.19, but our floor stays at `>=0.1.10` (UI-adapter back-compat policy: no version-requirement bumps, gate new functionality behind version + import checks). The `try`/`except` gives a runtime fallback (stubs + `_HAS_INTERRUPTS=False`) on 0.1.10–0.1.18, and the `TYPE_CHECKING` branch hands pyright the real types for narrowing. Keeping it.

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_event_stream.py
dsfaccini: David's AICA here: resolving — same version-split rationale. Bumping the `ag-ui-protocol` floor to `>=0.1.19` is disallowed by our UI-adapter back-compat policy (`pydantic_ai/ui/CLAUDE.md`, established in #3971): we stick to the lowest existing version requirement and gate new features behind version + import checks. The stubs let pyright type-check against the real types when the SDK has them, while the module still imports on 0.1.10–0.1.18 (where `_HAS_INTERRUPTS=False` and the interrupt paths are skipped). Keeping the stubs + gate.

docs/ui/ag-ui.md
dsfaccini: David's AICA here: verified the docs preview — built the site and the interrupt mapping table in `docs/ui/ag-ui.md` renders correctly (proper table, all 9 rows intact, reference links resolve, no broken-link warnings on build), consistent with how the page presents other adapter mappings. Resolving.

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
dsfaccini: David's AICA here: removed the 3 `cast`s in ec85b28c1 — they're now a lightweight `TypeAdapter(dict[str, Any])` narrowing. (Pyright rejects plain `isinstance` narrowing of an `Any` payload with `reportUnknownVariableType`, so a `cast` or a Pydantic adapter were the only cast-free options.) The deny-by-default behavior is unchanged — the `payload.get('approved') is True` identity check is stricter than `StrictBool`, so the ambiguous-payload cases still deny. The full `_ResumePayload` model unification (single source of truth for inbound `model_validate` + outbound `response_schema`, `StrictBool`, and the `editedArgs:'bad'` behavior-shift decision) is captured as follow-up #5878 — thanks @jjmllr for the detailed writeup and the offer to pick it up. Resolving here.

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
```diff
@@ -202,6 +223,60 @@ def _user_content_to_input(
...
+            f'{INTERRUPT_ID_PREFIX!r} prefix; cannot map it back to a tool call id.'
+        )
+    return interrupt_id[len(INTERRUPT_ID_PREFIX) :]
+
+
+_dict_adapter = TypeAdapter(dict[str, Any])
+"""Narrows an `Any`-typed JSON object (`ResumeEntry.payload`, `RunAgentInput.state`) to `dict[str, Any]`."""
+
+
+def _payload_dict(raw: Any) -> dict[str, Any]:
```
dsfaccini: this should be a global util in `_utils.py`

we already have one which is just a typeguard, idk if they are swappable as is?

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
```diff
@@ -202,6 +223,60 @@ def _user_content_to_input(
         assert_never(item)
 
 
+def _interrupt_id_to_tool_call_id(interrupt_id: str) -> str:
```
dsfaccini: we could place this in an own `._interrupt` module

dsfaccini (changes requested):

dsfaccini:
gonna do one more iteration on this to polish some edges, but we'll merge this soon :) I'll let you know when it goes out in a release

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
dsfaccini: David's AICA here: done. Turns out the typeguard you were thinking of already exists — `is_str_dict` in the root `pydantic_ai/_utils.py` (a `TypeGuard[dict[str, Any]]`), already used across `messages.py`, the UI adapters, and the model backends. Dropped the bespoke `_payload_dict`/`TypeAdapter` and repointed the resume mapping + the `state` property at the shared guard in f2af82046. Also added an agent-docs note pointing future agents at `_utils.py` so it's discoverable (#5880).

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
dsfaccini: David's AICA here: done in f2af82046 — extracted `pydantic_ai/ui/ag_ui/_interrupt.py`, which now owns the single `HAS_INTERRUPTS` import gate + interrupt stubs and the two-directional mapping (`approval_to_interrupt` outbound; `interrupt_id_to_tool_call_id` + `resume_entry_to_approval` inbound). This also removed the duplicate `_HAS_INTERRUPTS` derivation that was previously in both `_adapter` and `_event_stream`. `_adapter`/`_event_stream` shed ~140 lines.

dsfaccini:
David's AICA here: this shipped in [v2.0.0](https://github.com/pydantic/pydantic-ai/releases/tag/v2.0.0) 🎉 The AG-UI interrupt ↔ `DeferredTools` mapping is in the stable release — thanks for the contribution, @jjmllr!

PR #5433, opened 2026-05-13
Forward `thinking=False` on hybrid OpenRouter/xAI/Bedrock routes, aligning silent-drop with direct routes

sarth6:
- Closes #5379

> Re-files #5398 (auto-closed when the branch was renamed). Folds in @salty-flower's `reasoning.enabled=True` finding from #5442 (closed as a duplicate). Disable shape uses `reasoning.effort='none'` per @slmnsh's review — the [documented OpenRouter signal](https://openrouter.ai/docs/guides/best-practices/reasoning-tokens). Audit-found gaps not in scope are tracked in **#5445**.

## Impact

`thinking=False` was silently dropped at the wrong layer on routes whose upstream *can* honor disable:

- **Hybrid OpenRouter routes** (`moonshotai/kimi-k2.6`, `qwen/qwen3-235b-a22b`, `z-ai/glm-4.6`, `openai/gpt-oss-120b`) — their intrinsic profiles don't declare reasoning support, so the profile gate stripped `thinking` before the OpenRouter transformer ever saw it.
- **OpenRouter routes that do declare reasoning** (`anthropic/claude-sonnet-4.5`, `x-ai/grok-3-mini`) — the transformer's `if thinking is not False:` guard short-circuited the disable branch.
- **Direct xAI `grok-3-mini`, Bedrock-Converse `openai.gpt-oss-120b`, Bedrock-Converse `qwen.qwen3-32b`** — same short-circuit in their respective transformers.

Always-on routes (`openai/o3`, `openai/gpt-5`, `mistralai/magistral-medium-*`, `deepseek/deepseek-r1`, OpenRouter-routed `x-ai/grok-3-mini`) were already correctly silent-dropping `thinking=False` at the gate per the [`ModelProfile.thinking_always_enabled`](../pydantic_ai_slim/pydantic_ai/profiles/__init__.py) docstring — this PR does not change that contract and now ensures aggregator routes to the same models match their direct-route behavior.

For truthy `thinking`, the OpenRouter payload was missing `enabled: True`, which silently disabled reasoning on some reasoning-optional routes (Cause #3, found by @salty-flower in #5442).

Wire-body matrix, before / after:

| Path | Model | `thinking=True` | `thinking=False` |
|---|---|---|---|
| OpenRouter | `moonshotai/kimi-k2.6` | _none_ → `{effort: 'medium', enabled: True}` | _none_ → `{effort: 'none'}` |
[...]

PR #5442, opened 2026-05-14
fix(openrouter): always set reasoning.enabled=true when thinking is truthy

salty-flower:
## The bug

In `pydantic_ai/models/openrouter.py`, when a user sets `ModelRequestParameters.thinking` to a truthy value, `_openrouter_settings_to_openai_settings` converts it to:

```python
extra_body['reasoning'] = {'effort': <mapped>}
```

It never populates `enabled: true`, even though `OpenRouterReasoning` declares the field.

For models that are **reasoning-by-default** on OpenRouter (e.g. `openai/gpt-oss-20b`, `openai/o3`) this is fine — OpenRouter treats reasoning as enabled implicitly. For **reasoning-optional** models on certain routes (e.g. some `google/gemma-*` providers), the omission leaves reasoning disabled despite the user's explicit `thinking=True`. This is a silent-disable that's hard to diagnose because no error is raised — the model simply doesn't think.

I hit this while building a triage agent at work. A direct curl with `reasoning: {enabled: true, effort: "medium"}` produced reasoning tokens; pydantic-ai's translation produced none. OpenRouter's catalog (`https://openrouter.ai/api/v1/models`) lists `reasoning` in the model's `supported_parameters` — the gap is purely in pydantic-ai's payload construction.

## The fix

One line, in the unified-thinking fallback at `pydantic_ai/models/openrouter.py:572`:

```python
unified_reasoning['effort'] = effort_map[thinking]
unified_reasoning['enabled'] = True  # <-- new
model_settings['openrouter_reasoning'] = unified_reasoning
```

## Backwards compatibility

- **Reasoning-by-default models** (gpt-oss-*, o3, etc.): no behavior change — OpenRouter already treats `enabled` as `True` implicitly.
- **Reasoning-optional models**: previously-silent-disabled requests now correctly enable reasoning, matching the user's `thinking=True` intent.
- **Users who want explicit control** can still pass `OpenRouterModelSettings.openrouter_reasoning={...}` directly; that path is untouched and overrides the unified-thinking fallback entirely (test added asserts this).

## Tests

Added unit tests against `_openrouter_settings_to_openai_settings` in `tests/models/test_openrouter.py` (8 new test cases):

- Parametrized over every `ThinkingLevel` value (`True`, `'minimal'`, `'low'`, `'medium'`, `'high'`, `'xhigh'`): asserts `reasoning.effort` is mapped correctly AND `reasoning.enabled is True`.
- `thinking=False` does not inject a reasoning block.
- Explicit `openrouter_reasoning` is honoured verbatim and the unified-thinking fallback is skipped (no `enabled` flag injected when the user passed only `effort`).

Test invocation:

```
$ python -m pytest tests/models/test_openrouter.py
50 passed in 2.91s
```

Full `tests/models/` suite: `1756 passed, 217 skipped`.

## Notes

- Diff is +4 lines source / +57 lines tests.
- The unit tests cover the translation function directly (no cassettes / no network), so they're cheap and stable.

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/openrouter.py
+++ b/pydantic_ai_slim/pydantic_ai/models/openrouter.py
@@ -570,6 +570,10 @@ def _openrouter_settings_to_openai_settings(
                 'xhigh': 'high',
             }
             unified_reasoning['effort'] = effort_map[thinking]  # type: ignore[typeddict-item]
+            # Explicitly enable reasoning. For reasoning-by-default models (e.g. openai/gpt-oss-20b)
+            # OpenRouter treats this as a no-op, but for reasoning-optional models (e.g. gemma family)
+            # omitting `enabled` leaves reasoning disabled on some routes despite `effort` being set.
+            unified_reasoning['enabled'] = True
             model_settings['openrouter_reasoning'] = unified_reasoning
 
     if reasoning := model_settings.pop('openrouter_reasoning', None):
--- a/tests/models/test_openrouter.py
+++ b/tests/models/test_openrouter.py
@@ -1013,6 +1013,63 @@ async def test_openrouter_settings_to_openai_settings_with_web_search() -> None:
     assert extra_body['web_search_options'] == {'search_context_size': 'high'}
 
 
+@pytest.mark.parametrize(
+    ('thinking', 'expected_effort'),
+    [
+        (True, 'medium'),
+        ('minimal', 'low'),
+        ('low', 'low'),
+        ('medium', 'medium'),
+        ('high', 'high'),
+        ('xhigh', 'high'),
+    ],
+)
+async def test_openrouter_settings_to_openai_settings_thinking_sets_enabled(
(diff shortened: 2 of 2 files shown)


DouweM:
Presumably a fix for https://github.com/pydantic/pydantic-ai/issues/5379, but duplicate of https://github.com/pydantic/pydantic-ai/pull/5433

sarth6:
Hey @salty-flower — your finding was real and your PR was correct. I'd missed Cause #3 entirely until I went back and read your evidence. Folded your one-line fix into #5433 with credit + a cassette regression test pinning the wire body has `{effort: ..., enabled: True}`. Apologies for the duplicate close — your curl evidence was the only signal that surfaced the issue. Thanks for the careful write-up.

salty-flower:
Thanks @sarth6 — really appreciate the careful triage and the cassette regression test. Pinning the wire body to `{effort, enabled}` will catch any future drift cleanly. No worries on the close label; glad the fix lands somewhere good.

PR #5447, opened 2026-05-14
ci: harden CI/CD security

jpr5:
## Summary
- SHA-pin all GitHub Actions with version comments
- Add least-privilege permissions blocks to all workflows
- Fix shell injection (route attacker-controllable values through env)
- Add persist-credentials: false on read-only checkouts
- Add zizmor static analysis for workflow security
- Add Dependabot daily for github-actions (auto-merge minor/patch)
- Clean up floating branch ref on kludex/github-agent (SHA-pinned to main HEAD)

Part of Phase 2 CI/CD supply chain hardening.


jpr5:
Opened against wrong repo. Intended for ag-ui-protocol/pydantic-ai fork.

PR #5448, opened 2026-05-14
tests: mock DuckDuckGo client in WebSearch fallback test

DouweM:
`test_websearch_default_with_nonsupporting_model` (added in #5331) hits the live
DuckDuckGo/Bing endpoint via `ddgs`. Bing has been returning `None` consistently
for CI runner IPs since ~20:49 UTC today, causing the same single-test failure
on every PR that's run CI since. VCR can't intercept here — `ddgs` makes
requests through `primp` (Rust-based HTTP client) — so mock `DDGS.text` at the
library boundary instead, matching the existing pattern in
`test_webfetch_default_with_nonsupporting_model`.

Also tidies up around it:

- Dev dep `duckduckgo-search>=7.0.0` → `ddgs>=9.0.0` (matches the
  `pydantic-ai-slim[duckduckgo]` extra). The legacy package still emits a
  "renamed to ddgs" `RuntimeWarning` on `DDGS()` instantiation, which the dev
  env was pulling in unnecessarily.
- Drops the legacy `duckduckgo_search` fallback in `common_tools/duckduckgo.py`
  since `pydantic-ai-slim[duckduckgo]` has declared `ddgs>=9.0.0` since the
  migration.
- `pytest.importorskip('duckduckgo_search', …)` → `'ddgs'` in three tests.
- Drops the `@pytest.mark.filterwarnings('ignore::RuntimeWarning')` workaround
  that existed only because the legacy package was on the import path.

### Checklist


pydantic_ai_slim/pydantic_ai/common_tools/duckduckgo.py
```diff
@@ -8,10 +8,7 @@
 from pydantic_ai.tools import Tool
 
 try:
-    try:
-        from ddgs.ddgs import DDGS
-    except ImportError:  # Fallback for older versions of ddgs
```
DouweM, line 13: Are you sure we don't need the old fallback anymore? What if a user had manually added `duckduckgo_search` and it was working and now we drop it and it is no longer? Or does `duckduckgo_search` now pull in `ddgs` so that the other import would also work?

PR #5462, opened 2026-05-15
Make `FunctionToolset.tool()` raise on non-`RunContext` callables

dsfaccini:
### Description

`FunctionToolset.tool()` is intended for tools that take a `RunContext`. `FunctionToolset.tool_plain()` is the explicit form for tools that don't. The fallback that allowed a no-`RunContext` callable on `.tool()` (shipped with a `DeprecationWarning` in 1.x) is removed in v2.

`FunctionToolset.tool()` now raises a `UserError` directing users to `FunctionToolset.tool_plain()` if the registered callable's first parameter is not annotated with `RunContext`.

### Migration

```python
# Before (v1.x emitted DeprecationWarning, now raises)
@toolset.tool
async def my_tool(query: str) -> str:
    return f'searched: {query}'

# After: add RunContext to the signature...
@toolset.tool
async def my_tool(ctx: RunContext, query: str) -> str:
    return f'searched: {query}'

# ... or switch to tool_plain
@toolset.tool_plain
async def my_tool(query: str) -> str:
    return f'searched: {query}'
```

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/toolsets/function.py
+++ b/pydantic_ai_slim/pydantic_ai/toolsets/function.py
@@ -270,7 +270,7 @@ class FunctionToolset(AbstractToolset[AgentDepsT]):
                 raise UserError(
                     f'`FunctionToolset.tool()` requires a function whose first parameter is annotated with `RunContext`. '
                     f'For tools that do not need run context, use `FunctionToolset.tool_plain()` instead. '
-                    f'Got: {func_.__qualname__}'
+                    f'Got: {func_.__qualname__!r}'
                 )
             self.add_function(
                 func=func_,
--- a/pydantic_ai_slim/pydantic_ai/toolsets/function.py
+++ b/pydantic_ai_slim/pydantic_ai/toolsets/function.py
@@ -9,7 +9,6 @@ from pydantic.json_schema import GenerateJsonSchema
 
 from .._run_context import AgentDepsT, RunContext
 from .._system_prompt import SystemPromptRunner
-from .._utils import takes_run_context
 from ..exceptions import ModelRetry, UserError
 from ..messages import InstructionPart
 from ..tools import (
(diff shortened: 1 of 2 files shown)


pydantic_ai_slim/pydantic_ai/toolsets/function.py
```diff
@@ -266,15 +266,15 @@ async def spam(ctx: RunContext[str], y: float) -> float:
...
-            #  in function_schema if the function doesn't accept RunContext.
-
-            # Is the func actually taking RunContext or is it a plain function in disguise?
-
-            tool = self.add_function(
+            if not takes_run_context(func_):
+                raise UserError(
+                    f'`FunctionToolset.tool()` requires a function whose first parameter is annotated with `RunContext`. '
+                    f'For tools that do not need run context, use `FunctionToolset.tool_plain()` instead. '
+                    f'Got: {func_.__qualname__}'
```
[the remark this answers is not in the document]
dsfaccini: David's AICA here: applied — `!r` on the qualname in 2b2e1372a.

pydantic_ai_slim/pydantic_ai/toolsets/function.py
```diff
@@ -266,15 +266,15 @@ async def spam(ctx: RunContext[str], y: float) -> float:
...
             func_: ToolFuncContext[AgentDepsT, ToolParams],
         ) -> ToolFuncContext[AgentDepsT, ToolParams]:
-            # TODO(v2): Remove this deprecation fallback
-            #  and let takes_ctx=True propagate, which will raise a runtime error
-            #  in function_schema if the function doesn't accept RunContext.
-
-            # Is the func actually taking RunContext or is it a plain function in disguise?
-
-            tool = self.add_function(
+            if not takes_run_context(func_):
```
DouweM: Do we really need this error The type system also ensures we get a RunContext right? We don't have such an error on `Agent.tool` either!

dsfaccini: David's AICA here: agreed — applied in commit `044193cbe`. Dropped the `UserError` raise (and the `takes_run_context` import), now matches `Agent.tool`'s reliance on the type system. Also dropped the now-irrelevant `test_tool_without_runctx_raises` test. Net diff: -31 lines.

PR #5479, opened 2026-05-15
Drop `GoogleGLAProvider`, `GoogleVertexProvider`, `GeminiModel` and the `gemini:` / `google-gla:` / `google-vertex:` / `vertexai:` provider prefixes

dsfaccini:
## Summary

Drops the legacy Google provider/model surface that has been `@deprecated` on `v2-main` since the [`GoogleProvider` / `GoogleCloudProvider` split](https://github.com/pydantic/pydantic-ai/pull/5336):

- **Classes removed**:
  - `pydantic_ai.providers.google_gla.GoogleGLAProvider` (file deleted)
  - `pydantic_ai.providers.google_vertex.GoogleVertexProvider` + `VertexAiRegion` (file deleted)
  - `pydantic_ai.models.gemini.GeminiModel` + the whole `models/gemini.py` module (`GeminiModelName`, `GeminiSafetySettings`, `GeminiStreamedResponse`, `ThinkingConfig`, `ApiKeyAuth`, `AuthProtocol`)
- **Aliases removed**:
  - `pydantic_ai.providers.google.VertexAILocation` (use `GoogleCloudLocation`)
  - `pydantic_ai.models.google.GoogleVertexServiceTier` (use `GoogleCloudServiceTier`)
  - `pydantic_ai.models.google.GoogleServiceTier`
- **Provider prefixes removed** (each was emitting `PydanticAIDeprecationWarning` on v2-main):
  - `'google-gla:'` → use `'google:'`
  - `'google-vertex:'` → use `'google-cloud:'`
  - `'vertexai:'` → use `'google-cloud:'`
  - `'gateway/google-vertex:'` → use `'gateway/google-cloud:'`
- **Kwarg removals on `GoogleProvider.__init__`**: `vertexai=`, `location=`, `project=`, `credentials=` (all four were deprecated and forwarded to `GoogleCloudProvider`).
- **Setting key removals on `GoogleModelSettings`**: `google_vertex_service_tier`, `google_service_tier` (use `google_cloud_service_tier`).
- **Extra removed**: the `vertexai` extra on `pydantic-ai-slim` (its only purpose was `GoogleVertexProvider`, which is gone). Install `pydantic-ai-slim[google]` for both the Gemini API and Google Cloud paths.

Passing any of the removed prefixes / kwargs / setting keys now raises (`ValueError: Unknown provider` for prefixes; `TypeError: unexpected keyword argument` for kwargs).

## Back-compat preserved (V2-RULES rule 23)

Legacy literal values stay where they end up serialized into `ModelMessage` or persisted user data, so historical message-history payloads still validate and route correctly:

- `pydantic_ai.messages.UploadedFileProviderName` keeps `'google-gla'` and `'google-vertex'` in the `Literal[...]` alias.
- `pydantic_ai.models.google._GEMINI_API_PROVIDER_NAMES` keeps `frozenset({'google', 'google-gla'})`.
- `pydantic_ai.models.google._GOOGLE_CLOUD_PROVIDER_NAMES` keeps `frozenset({'google-cloud', 'google-vertex'})`.
- `infer_model` accepts `'google-vertex'` as an internal `model_kind` because that's the wire value `normalize_gateway_provider` returns for `gateway/google-cloud:` / `gateway/gemini:` (rule 17 — Gateway team renames their side later; user-facing prefix has already moved to `google-cloud`).

## Cross-history-replay test (V2-RULES rule 4)

`tests/v2/test_google_history_replay.py` already covers the contract — replays a `ModelResponse` with `provider_name='google-gla'` / `'google-vertex'` through the new `GoogleModel` class and asserts thinking-signature carry-over and native-tool round-tripping still work. Kept in place.

## Scope notes

- `tests/v2/test_google_cloud_split_deprecation.py` deleted — per rule 6, `tests/v2/` is the bucket for 1.x deprecation tests, which become moot once removal lands.
- The `vertexai` extra was the legacy-only way to install `GoogleVertexProvider`'s `google-auth` + `requests` deps; the new `GoogleCloudProvider` reuses the `google-genai` SDK that's already pulled in by the `google` extra.
- `uv.lock` not regenerated in this PR (separate `fastmcp` constraint blocks `uv lock` locally on this branch); maintainer can re-lock at merge time.
- `tests/models/test_multimodal_tool_returns.py` keeps its hardcoded `'google-vertex'` system-property mocks and `provider_name` literals — those are intentional rule-23 back-compat coverage, not migration debt.
- `pydantic_ai.models.openrouter.KnownOpenRouterProviders` keeps `'google-vertex'` — that's OpenRouter's marketplace taxonomy (their provider id), not our provider name.

## Test plan

- `make format && make lint` clean.
- Full local pytest run: 5563 passed, 315 skipped (excluding unrelated `tests/evals/test_reporting.py` terminal-width snapshot mismatch and `tests/test_examples.py` / `tests/test_logfire.py`).
- `tests/v2/test_google_history_replay.py` passes (rule-4 contract).
- `tests/providers/test_provider_names.py` updated: dropped `'google-gla'` / `'google-vertex'` / `'vertexai'` rows.
- `tests/providers/test_gateway.py` row `('google-cloud', GoogleCloudProvider, 'google-vertex')` still passes — verifies rule-17 user-facing-name → wire-value mapping.

## Checklist

- [x] No **breaking changes** in accordance with the [version policy](https://github.com/pydantic/pydantic-ai/blob/main/docs/version-policy.md). (Targets `v2-main` for the v2 cut.)

Relates to #5150.


pydantic_ai_slim/pydantic_ai/providers/__init__.py
```diff
@@ -113,37 +111,12 @@ def __repr__(self) -> str:
...
     """Infers the provider class from the provider name."""
-    came_from_gateway = provider.startswith('gateway/')
-    # Normalize gateway-prefixed providers (e.g. 'gateway/openai' -> 'openai').
-    # The `gateway/google-vertex` deprecation warning fires inside `normalize_gateway_provider`.
-    if came_from_gateway:
+    if provider.startswith('gateway/'):
+        # Normalize gateway-prefixed providers (e.g. 'gateway/openai' -> 'openai').
         from .gateway import normalize_gateway_provider
 
         provider = normalize_gateway_provider(provider)
```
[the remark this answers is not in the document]
dsfaccini: David's AICA here: Done — dropped the dead `normalize_gateway_provider` call in commit 6a405f759. `infer_provider` early-returns on `gateway/` prefixes (line 242-246) so `infer_provider_class` never receives them legitimately; `test_infer_provider_class` explicitly skips gateway providers.

tests/models/test_google.py
```diff
@@ -147,6 +148,13 @@ def test_google_client_property_delegates_to_provider(google_provider: GooglePro
     assert model.client is google_provider.client
 
 
+def test_google_cloud_provider_accepts_prebuilt_client():
+    """`GoogleCloudProvider(client=...)` short-circuits construction and stores the supplied client."""
+    mock_client = MagicMock()
+    provider = GoogleCloudProvider(client=mock_client)
+    assert provider.client is mock_client
```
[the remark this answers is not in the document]
dsfaccini: David's AICA here: Done — swapped `MagicMock()` for `Client(vertexai=False, api_key='mock-api-key')` and dropped the now-unused `MagicMock` import in commit 6a405f759.

pydantic_ai_slim/pydantic_ai/embeddings/__init__.py
```diff
@@ -118,7 +118,9 @@ def infer_embedding_model(
         from .bedrock import BedrockEmbeddingModel
 
         return BedrockEmbeddingModel(model_name, provider=provider)
-    elif model_kind in ('google', 'google-gla', 'google-vertex', 'google-cloud'):
+    elif model_kind in ('google', 'google-cloud', 'google-vertex'):
```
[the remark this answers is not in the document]
dsfaccini: David's AICA here: Deferring — embeddings type aliases (`LatestGoogleGLAEmbeddingModelNames`, `LatestVertexAIEmbeddingModelNames`) are out of scope for card 08 (google-consolidation), which focuses on dropping the `GeminiModel` / `GoogleGLAProvider` / `GoogleVertexProvider` classes and the `google-gla:` / `google-vertex:` / `vertexai:` prefixes. The embeddings module has its own separate concerns (different SDK, different surface) and renaming those aliases is a distinct migration. Suggest a follow-up card to rename them to `LatestGoogleEmbeddingModelNames` / `LatestGoogleCloudEmbeddingModelNames` aligned with the provider names.

pydantic_ai_slim/pydantic_ai/providers/__init__.py
```diff
@@ -113,35 +111,15 @@ def __repr__(self) -> str:
...
-        if not came_from_gateway:
-            warnings.warn(
-                "The 'google-vertex:' prefix is deprecated and will be removed in v2.0. Use 'google-cloud:' instead.",
-                PydanticAIDeprecationWarning,
-                stacklevel=2,
-            )
+    # `normalize_gateway_provider` returns `'google-vertex'` (the wire value the Gateway API
+    # still expects) for `gateway/google-cloud:` and `gateway/gemini:`. Map it back to the
+    # provider class name (rule 17). Drop this branch when the Gateway team renames their side.
+    if provider == 'google-vertex':
```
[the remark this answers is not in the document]
dsfaccini: David's AICA here: fixed in 36b5b1335. Moved the `'google-vertex'` → `'google-cloud'` wire-value mapping inside the `if provider.startswith('gateway/')` branch so it only applies to gateway-prefixed inputs (per rule 17). User-facing `'google-vertex'` now raises `ValueError: Unknown provider`. Added a regression test `test_infer_provider_rejects_removed_google_vertex_prefix` in `tests/providers/test_provider_names.py`.

pydantic_ai_slim/pydantic_ai/providers/google.py
```diff
@@ -88,45 +85,24 @@ class GoogleProvider(BaseGoogleProvider):
 
     @property
     def name(self) -> str:
+        # Returned value flows into ModelMessage.provider_name on every part.
```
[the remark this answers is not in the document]
dsfaccini: David's AICA here: condensed to one line in 36b5b1335.

docs/models/google.md
```diff
@@ -8,9 +8,6 @@ Two providers wrap those endpoints:
 - [`GoogleProvider`][pydantic_ai.providers.google.GoogleProvider] — the Gemini API (Google AI Studio), surfaced under the `'google:'` prefix.
 - [`GoogleCloudProvider`][pydantic_ai.providers.google_cloud.GoogleCloudProvider] — Google Cloud (formerly known as Vertex AI), surfaced under the `'google-cloud:'` prefix.
 
-!!! note "Renamed prefixes (1.x → v2)"
-    The `'google-gla:'` and `'google-vertex:'` prefixes still work in 1.x but emit a `DeprecationWarning`. Use `'google:'` and `'google-cloud:'` instead. Likewise `GoogleProvider(...)` with any Google Cloud-only argument (`vertexai=True`, `location`, `project`, or `credentials`) is deprecated in favor of `GoogleCloudProvider(...)`.
-
 ## Install
```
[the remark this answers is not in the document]
dsfaccini: David's AICA here: removed the dangling reference in 36b5b1335 — `google_service_tier` is gone, so the sentence has no surviving referent.

pydantic_ai_slim/pydantic_ai/providers/__init__.py
```diff
@@ -113,36 +111,17 @@ def __repr__(self) -> str:
...
-                "The 'google-vertex:' prefix is deprecated and will be removed in v2.0. Use 'google-cloud:' instead.",
-                PydanticAIDeprecationWarning,
-                stacklevel=2,
-            )
-        provider = 'google-cloud'
+        # `normalize_gateway_provider` returns `'google-vertex'` (the wire value the Gateway API
+        # still expects) for `gateway/google-cloud:` and `gateway/gemini:`. Map it back to the
+        # provider class name (rule 17). Drop this branch when the Gateway team renames their side.
+        # Scoped to the gateway branch so user-facing `'google-vertex'` falls through to the
+        # `Unknown provider` ValueError below (M7 removed that prefix).
```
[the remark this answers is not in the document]
dsfaccini: David's AICA here: condensed to a single line — kept the canonical explanation here in providers/__init__.py and pointed the other two sites at this file. Commit 21c2cdc0b.

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -1501,7 +1494,9 @@ def infer_model(  # noqa: C901
         from .openai import OpenAIResponsesModel
 
         return OpenAIResponsesModel(model_name, provider=provider)
-    elif model_kind in ('google', 'google-gla', 'google-vertex', 'google-cloud'):
+    elif model_kind in ('google', 'google-cloud', 'google-vertex'):
+        # `'google-vertex'` is the internal wire value returned by `normalize_gateway_provider`
+        # for `gateway/google-cloud:` / `gateway/gemini:` until the Gateway team renames their side (rule 17).
```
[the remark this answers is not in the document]
dsfaccini: David's AICA here: replaced the duplicated explanation with a pointer to providers/__init__.py (same treatment applied to embeddings/__init__.py). Commit 21c2cdc0b.

tests/providers/test_provider_names.py
```diff
@@ -107,14 +98,17 @@ def test_infer_provider(provider: str, provider_cls: type[Provider[Any]], except
...
 def test_infer_provider_class(provider: str, provider_cls: type[Provider[Any]], exception_has: str | None):
     if provider.startswith('gateway/'):
         pytest.skip('Gateway providers are not supported for this test')
 
     assert infer_provider_class(provider) == provider_cls
+
+
+def test_infer_provider_rejects_removed_google_vertex_prefix():
+    """M7 dropped the `google-vertex:` provider prefix; only the `gateway/`-internal wire value should still
+    map to `GoogleCloudProvider`. User-facing `'google-vertex'` must raise `Unknown provider`."""
```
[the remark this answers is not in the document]
dsfaccini: David's AICA here: rewrote the docstring to be self-explanatory and drop the 'M7' internal reference. Commit 21c2cdc0b.

tests/models/test_google.py
```diff
@@ -138,6 +138,13 @@ def test_google_client_property_delegates_to_provider(google_provider: GooglePro
     assert model.client is google_provider.client
 
 
+def test_google_cloud_provider_accepts_prebuilt_client():
+    """`GoogleCloudProvider(client=...)` short-circuits construction and stores the supplied client."""
+    client = Client(vertexai=False, api_key='mock-api-key')
+    provider = GoogleCloudProvider(client=client)
+    assert provider.client is client
```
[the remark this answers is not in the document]
dsfaccini: David's AICA here: dropped the test — it only asserted identity round-trip, no behavioral coverage. Commit 21c2cdc0b.

tests/providers/test_provider_names.py
```diff
@@ -106,14 +97,16 @@ def test_infer_provider(provider: str, provider_cls: type[Provider[Any]], except
...
 
     assert infer_provider_class(provider) == provider_cls
+
+
+def test_infer_provider_rejects_removed_google_vertex_prefix():
+    """The `google-vertex:` provider prefix was removed in v2; only the internal gateway wire value still maps to GoogleCloudProvider."""
+    with pytest.raises(ValueError, match='Unknown provider: google-vertex'):
+        infer_provider_class('google-vertex')
+    with pytest.raises(ValueError, match='Unknown provider: google-vertex'):
+        infer_provider('google-vertex')
```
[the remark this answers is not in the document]
dsfaccini: David's AICA here: parametrized in commit `21c2cdc0b` — `test_infer_provider_rejects_removed_google_prefixes` now covers `google-gla`, `google-vertex`, and `vertexai` against both `infer_provider_class` and `infer_provider`.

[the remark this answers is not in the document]
dsfaccini: (Addendum: the correct commit SHA is `2a432c95a`, not the earlier one in my reply — the previous one only had the prior fixes, this is where `test_infer_provider_rejects_removed_google_prefixes` actually lands parametrized over all three prefixes.)

tests/models/test_google.py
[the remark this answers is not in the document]
dsfaccini: David's AICA here: corrected — the test was restored in 50badd518. The earlier rationale was wrong: `vertex_provider` fixture (tests/conftest.py:824) uses `project=`/`location=`, not `client=`, so it doesn't exercise the early-return branch at `google_cloud.py:55-57`. CI coverage failure on the prior push confirmed this.

docs/.hooks/main.py
```diff
@@ -141,8 +141,8 @@ def create_gateway_toggle(markdown: str, relative_path: Path) -> str:
     'openai-responses': 'openai-responses',
     'openai-chat': 'openai',
     'bedrock': 'bedrock',
-    'google-gla': 'gemini',
-    'google-vertex': 'google-vertex',
+    'google': 'gemini',
```
DouweM: Why not `google`?

docs/.hooks/main.py
```diff
@@ -141,8 +141,8 @@ def create_gateway_toggle(markdown: str, relative_path: Path) -> str:
     'openai-responses': 'openai-responses',
     'openai-chat': 'openai',
```
DouweM: These 3 lines should probably be updated in https://github.com/pydantic/pydantic-ai/pull/5469/changes but aren't

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -1483,7 +1476,8 @@ def infer_model(  # noqa: C901
         from .openai import OpenAIResponsesModel
 
         return OpenAIResponsesModel(model_name, provider=provider)
-    elif model_kind in ('google', 'google-gla', 'google-vertex', 'google-cloud'):
+    elif model_kind in ('google', 'google-cloud', 'google-vertex'):
+        # See providers/__init__.py for why google-vertex → google-cloud.
```
DouweM: We really still need `google-vertex` here? We can't handle that just in the gateway path?

pydantic_ai_slim/pydantic_ai/providers/__init__.py
```diff
@@ -113,36 +111,13 @@ def __repr__(self) -> str:
...
-        if not came_from_gateway:
-            warnings.warn(
-                "The 'google-vertex:' prefix is deprecated and will be removed in v2.0. Use 'google-cloud:' instead.",
-                PydanticAIDeprecationWarning,
-                stacklevel=2,
-            )
-        provider = 'google-cloud'
+        # Gateway wire value: google-vertex → google-cloud until Gateway team renames their side.
+        if provider == 'google-vertex':
+            provider = 'google-cloud'
```
DouweM: This can't be in `normalize_gateway_provider`?

pydantic_ai_slim/pydantic_ai/providers/gateway.py
```diff
@@ -73,7 +71,7 @@ def gateway_provider(
 
 @overload
 def gateway_provider(
-    upstream_provider: Literal['gemini', 'google-cloud', 'google-vertex'],
+    upstream_provider: Literal['gemini', 'google-cloud'],
```
DouweM: It's weird that we have a mix of old and new names here

pydantic_ai_slim/pydantic_ai/providers/gateway.py
```diff
@@ -246,19 +244,11 @@ def normalize_gateway_provider(provider: str) -> str:
...
-            PydanticAIDeprecationWarning,
-            stacklevel=2,
-        )
-
     if provider in ('openai', 'openai-chat', 'chat'):
         return 'openai'
     elif provider in ('openai-responses', 'responses'):
         return 'openai-responses'
-    elif provider in ('gemini', 'google-cloud', 'google-vertex'):
+    elif provider in ('gemini', 'google-cloud'):
```
DouweM: I don't understand why the `gemini` name has to survive

pydantic_ai_slim/pydantic_ai/models/__init__.py
dsfaccini: David's AICA here: addressed in d36fe5b2b — `normalize_gateway_provider` now returns the canonical class-lookup name `google-cloud`, and the legacy `google-vertex` branches in `models/__init__.py`, `providers/__init__.py`, and `embeddings/__init__.py` are gone. The wire-value remap to `google-vertex` lives in one place (gateway_provider's default-route logic).

pydantic_ai_slim/pydantic_ai/providers/__init__.py
dsfaccini: David's AICA here: done in d36fe5b2b — the post-normalize special-case is gone; `normalize_gateway_provider` now returns `google-cloud` directly. The wire-value tweak (`google-cloud` → `google-vertex`) is applied inline only in `gateway_provider`'s default-route logic, which is the one caller that actually needs the upstream wire value.

pydantic_ai_slim/pydantic_ai/providers/gateway.py
dsfaccini: David's AICA here: fixed in d36fe5b2b — `normalize_gateway_provider` is now consistent (returns class-lookup names like `google-cloud`). The only `google-vertex` reference left in this file is the inline wire-value remap inside `gateway_provider`, with a comment pointing at the Gateway-team rename TODO.

docs/.hooks/main.py
dsfaccini: David's AICA here: `gateway/google:` isn't a valid upstream — `gateway_provider` only accepts `gemini` (Gemini-API flavor) and `google-cloud` (Vertex flavor). Pydantic AI's bare `google:` prefix uses `GoogleProvider` (Gemini Developer / GLA), so the equivalent on the gateway side is `gateway/gemini:`. Happy to rename the wire alias to `gateway/google:` in a follow-up if you want symmetry with the bare prefix, but it would need a Gateway-team coordination since `gemini` is the API-flavor name they expose.

docs/.hooks/main.py
dsfaccini: David's AICA here: agreed, this is #5469 scope — I'll push the `GATEWAY_MODEL_MAP` update there. Per the routing table in #5469's description: after the openai-default flip, `openai` → `openai-responses`, `openai-chat` → `openai`, `openai-responses` → `openai-responses`. Leaving this thread open for #5469.

pydantic_ai_slim/pydantic_ai/providers/gateway.py
dsfaccini: David's AICA here: `gemini` survives because it's the user-facing `APIFlavor` value the Gateway team exposes for the Gemini-API flavor (analogous to `chat` / `responses` / `converse` for the other providers — see the `APIFlavor` Literal a few lines above). Users can call `gateway_provider('gemini', ...)` directly to hit the Gemini API flavor on the gateway, vs `'google-cloud'` for Vertex. Dropping it would break that public alias. If the Gateway team renames their side to `google` someday we can flip both — happy to track that as a follow-up if you want.

pydantic_ai_slim/pydantic_ai/providers/gateway.py
```diff
@@ -246,22 +248,12 @@ def normalize_gateway_provider(provider: str) -> str:
...
-
     if provider in ('openai', 'openai-chat', 'chat'):
         return 'openai'
     elif provider in ('openai-responses', 'responses'):
         return 'openai-responses'
-    elif provider in ('gemini', 'google-cloud', 'google-vertex'):
-        # The Gateway API still expects `google-vertex` as the upstream-provider wire value.
-        # When the Gateway team renames their side, flip this to `google-cloud`.
-        return 'google-vertex'
+    elif provider in ('gemini', 'google-cloud'):
```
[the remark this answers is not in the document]
dsfaccini: David's AICA here: verified empirically — `gemini` is the gateway wire value for the Gemini-API flavor (parallel to `chat`/`responses`/`converse` in `APIFlavor` at gateway.py:105-112). It's exercised in [`tests/providers/test_provider_names.py:65`](https://github.com/pydantic/pydantic-ai/blob/v2-exec-google-consolidation/tests/providers/test_provider_names.py#L65) and [`tests/models/test_model.py:60,65`](https://github.com/pydantic/pydantic-ai/blob/v2-exec-google-consolidation/tests/models/test_model.py#L60-L65). Dropping it would break `gateway/gemini:...` model strings. Re docs hook: `GATEWAY_MODEL_MAP` maps `google: gemini` to render Gemini-flavor URLs (`gateway/gemini:...`) — this is intentional, not a forced workaround. Linked to the original Discuss thread on [r2163058488](https://github.com/pydantic/pydantic-ai/pull/5479#discussion_r2163058488); leaving that one open for Douwe to confirm.

docs/.hooks/main.py
dsfaccini: David's AICA here: applied in #5469 commit c58a2cb99. Also fixed a related pre-existing bug — `'openai-chat': 'openai'` was silently converting Chat-direct doc examples to Responses-gateway examples post-M6 (since `gateway/openai:` now routes to Responses); changed to `'openai-chat': 'openai-chat'` so Chat-direct stays Chat-gateway.

tests/providers/test_provider_names.py
```diff
@@ -106,14 +97,17 @@ def test_infer_provider(provider: str, provider_cls: type[Provider[Any]], except
...
         pytest.skip('Gateway providers are not supported for this test')
 
     assert infer_provider_class(provider) == provider_cls
+
+
+@pytest.mark.parametrize('removed_prefix', ['google-gla', 'google-vertex', 'vertexai'])
+def test_infer_provider_rejects_removed_google_prefixes(removed_prefix: str):
+    """The `google-gla:`, `google-vertex:`, and `vertexai:` provider prefixes were removed in v2. The `google-vertex` internal gateway wire value still maps to GoogleCloudProvider, but only when prefixed with `gateway/`."""
+    with pytest.raises(ValueError, match=f'Unknown provider: {removed_prefix}'):
+        infer_provider_class(removed_prefix)
```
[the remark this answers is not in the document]
dsfaccini: David's AICA here: Good catch — `gateway/google-vertex:` also raises (`gateway_provider` only accepts `google-cloud`/`gemini` as upstream). Rewrote the docstring to make explicit that `google-vertex` survives only as the internal Gateway route string in `gateway.py:154-157`, never as a user-facing prefix. Fixed in dad4dbf56.

docs/.hooks/main.py
DouweM: See https://github.com/pydantic/pydantic-ai/pull/5469#discussion_r3269647510, I think we should keep `google` -> `google` here (as these are user-facing strings), and handle the `google` -> `gemini` mapping at the gateway-route layer.

pydantic_ai_slim/pydantic_ai/providers/gateway.py
```diff
@@ -153,6 +151,10 @@ def gateway_provider(
     if route is None:
         # Use the implied providerId as the default route.
         route = normalize_gateway_provider(upstream_provider)
+        # The Gateway API still expects `google-vertex` as the upstream-provider wire value.
+        # When the Gateway team renames their side, drop this remap.
+        if route == 'google-cloud':
+            route = 'google-vertex'
```
DouweM: This is similar to what I meant in the OpenAI PR, in that there should be one path that deals with Gateway routes, separate from regular gateway normalization.

pydantic_ai_slim/pydantic_ai/providers/gateway.py
DouweM: This should be `google`, not `gemini` -- we need to map those below as well for the route.

pydantic_ai_slim/pydantic_ai/providers/gateway.py
```diff
@@ -199,7 +201,7 @@ def _with_http_client(provider: Provider[Any]) -> Provider[Any]:
                 anthropic_client=AsyncAnthropic(auth_token=api_key, base_url=base_url, http_client=http_client)
             )
         )
-    elif upstream_provider in ('google-cloud', 'google-vertex', 'gemini'):
+    elif upstream_provider in ('google-cloud', 'gemini'):
```
DouweM: `gemini` -> `google`

pydantic_ai_slim/pydantic_ai/providers/gateway.py
DouweM: AICA is being dumb, we are NOT forced to keep `gemini` user facing just because that's the value Gateway takes.

We should, however, consider checking what the Gateway docs say, and if we haven't explicitly deprecated e.g. `gateway/gemini:` yet, then we should maybe keep supporting it just in `normalize_gateway_provider` -- just as user input?

docs/.hooks/main.py
dsfaccini: David's AICA here: addressed in 63d8756e2 — `GATEWAY_MODEL_MAP` is now 1-to-1 identity (`'google': 'google'`), so the docs toggle shows `Agent('gateway/google:')`. The non-identity wire-value remap (`google → gemini`) lives in `_GATEWAY_ROUTE_REMAP` in `gateway.py` per the huddle decision.

pydantic_ai_slim/pydantic_ai/providers/gateway.py
dsfaccini: David's AICA here: addressed in 63d8756e2. `normalize_gateway_provider` is gone; replaced with two helpers — `_strip_gateway_prefix` (strip `gateway/` + resolve API-flavor aliases `chat`/`responses`/`converse`) and `_gateway_route` (canonical name → PAIG URL route segment via the explicit `_GATEWAY_ROUTE_REMAP` dict). Class-lookup paths and URL-route generation no longer share a single overloaded function.

pydantic_ai_slim/pydantic_ai/providers/gateway.py
dsfaccini: David's AICA here: addressed in 63d8756e2 + 37a161770. `gemini` survives only as (a) a runtime-accepted backward-compat alias for callers passing `gateway_provider('gemini', ...)`, and (b) the Gateway URL route segment for the GLA API (looked up via `_GATEWAY_ROUTE_REMAP['google'] = 'gemini'`). It is no longer in the static `APIFlavor` Literal or the `GoogleClient` overload — autocomplete now only offers `'google'` and `'google-cloud'`.

pydantic_ai_slim/pydantic_ai/providers/gateway.py
dsfaccini: David's AICA here: addressed in 63d8756e2. The two helpers now separate the concerns: `_strip_gateway_prefix` handles class lookup (strip + alias resolution), `_gateway_route` is the only path that knows about Gateway-side wire values.

pydantic_ai_slim/pydantic_ai/providers/gateway.py
dsfaccini: David's AICA here: addressed in 37a161770 — `APIFlavor` and the `GoogleClient` overload now expose `'google'` instead of `'gemini'`. `'gemini'` is kept as a runtime backward-compat alias (still accepted by `gateway_provider`'s body) but no longer in the static types.

pydantic_ai_slim/pydantic_ai/providers/gateway.py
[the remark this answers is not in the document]
dsfaccini: David's AICA here: addressed in 37a161770 — `APIFlavor` Literal updated to use `'google'` instead of `'gemini'`. Runtime still accepts `'gemini'` (via the `('google', 'gemini')` branch in `gateway_provider`) but it's no longer surfaced in autocomplete.

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -1456,7 +1449,7 @@ def infer_model(  # noqa: C901
         from .openai import OpenAIResponsesModel
 
         return OpenAIResponsesModel(model_name, provider=provider)
-    elif model_kind in ('google', 'google-gla', 'google-vertex', 'google-cloud'):
+    elif model_kind in ('google', 'gemini', 'google-cloud'):
```
dsfaccini: is `gemini` right here?

pydantic_ai_slim/pydantic_ai/providers/__init__.py
```diff
@@ -164,7 +136,7 @@ def infer_provider_class(provider: str) -> type[Provider[Any]]:  # noqa: C901
         from .azure import AzureProvider
 
         return AzureProvider
-    elif provider == 'google':
+    elif provider in ('google', 'gemini'):
```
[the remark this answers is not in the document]
dsfaccini: David's AICA here: Addressed in 7f722408f, with a different cut than suggested. After discussion + live verification (see local-notes/verify-gateway-google-prefixes.py), we collapsed `gateway/google` onto `gateway/google-cloud` (both → GoogleCloudProvider / Vertex) and dropped `gateway/gemini` entirely. Reason: `/proxy/gemini/*` returns 404 on the Gateway server today, so shipping it as a working prefix would be misleading. Both kept prefixes now hit the working `/proxy/google-vertex/*` route, and bare `gemini:` correctly raises UserError.

`'gemini'` is no longer in any match branch in `providers/__init__.py`, `models/__init__.py`, or `embeddings/__init__.py`. Will re-add `gateway/google` (or `gateway/gemini`) as a distinct GLA route if/when the Gateway team exposes one.

pydantic_ai_slim/pydantic_ai/models/__init__.py
dsfaccini: David's AICA here: No, dropped in 7f722408f. The match branch is now just `('google', 'google-cloud')`. `gateway/google` normalizes to `'google-cloud'` upstream (in `normalize_gateway_provider`) so both arrive at this branch correctly; bare `gemini:` now raises UserError.

pydantic_ai_slim/pydantic_ai/providers/gateway.py
```diff
@@ -110,7 +108,7 @@ def gateway_provider(
     'chat',
     'responses',
     'converse',
-    'gemini',
+    'google',
```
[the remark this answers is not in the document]
dsfaccini: David's AICA here: Moved `'google'` from `APIFlavor` to `ModelProvider` in bd78c87bc. It stays in `ModelProvider` after the gateway-prefix collapse in 7f722408f — users can still pass `gateway_provider('google', ...)` directly (it now routes to `GoogleCloudProvider` alongside `'google-cloud'`, since the Gateway server only exposes the Vertex route today).

PR #5480, opened 2026-05-15
fix(anthropic): support Bedrock token counting

### LINKED ISSUE #5377: Support token counting for Anthropic Bedrock models
### Description

Anthropic token counting currently cannot be used with Bedrock.

When using `AsyncAnthropicBedrock`, Pydantic AI raises before attempting token counting:

```py
UserError: AsyncAnthropicBedrock client does not support `count_tokens` api.
```

Since August 2025, Bedrock has supported CountTokens for Claude models. However, the Anthropic SDK still has not been updated to support this, except for the limited set of `bedrock-mantle` models.

Absent a Anthropic SDK update, there is a workaround - the supplied `AsyncAnthropicBedrock` client already has the AWS profile/credentials/region/http client/signing config, and its low-level `post(...)` method can call Bedrock Runtime's `/model/{modelId}/count-tokens` endpoint directly.

I tested a small fix locally against `anthropic.claude-haiku-4-5-20251001-v1:0` and got a successful token count response. I have a branch ready that:

- removes the current `AsyncAnthropicBedrock` guard
- builds the existing Anthropic Messages request body
- wraps it in Bedrock Runtime's `input.invokeModel.body` CountTokens shape
- base64-encodes the inner body for the raw HTTP JSON protocol
- returns `BetaMessageTokensCount(input_tokens=...)`

Would this be a reasonable bug-fix PR to open?

If relying on `AsyncAnthropicBedrock.post(...)` is considered too low-level, an alternative that would still solve this for downstream users would be an easy way to override the count-tokens behavior for a model. The function that needs to be overridden, `_messages_count_tokens(...)`, is currently not very atomic and is tied to a bunch of Pydantic AI internals.

### References

_No response_

dyscott:
### Summary
- Fix token counting for Anthropic models on Bedrock.
- Remove the current `AsyncAnthropicBedrock` guard now that Bedrock supports CountTokens for Claude models.
- Route Bedrock token counting through the supplied `AsyncAnthropicBedrock` client.
- Build the existing Anthropic Messages request body, wrap it in Bedrock Runtime’s `input.invokeModel.body` CountTokens shape, base64-encode the inner body for the raw HTTP JSON protocol, and return `BetaMessageTokensCount(input_tokens=...)`.
- Caveat: I could not find another model implementation in this codebase that calls through to an underlying SDK client’s low- level `post(...)` method directly. I am not sure if what I implemented would be the best pattern going forward, so looking to hear feedback on this.
- Closes #5377

### Testing Plan
- [x] New tests in `tests/models/test_anthropic.py` cover the Bedrock CountTokens request shape, including the base64-encoded
`input.invokeModel.body`, `system`, `anthropic_betas`, `extra_body`, and timeout handling.
- [x] Live-tested against `anthropic.claude-haiku-4-5-20251001-v1:0` on Bedrock and got a successful token count response.
- [x] `uv run pytest tests/models/test_anthropic.py::test_anthropic_bedrock_count_tokens tests/models/
test_anthropic.py::test_anthropic_bedrock_count_tokens_unexpected_response tests/models/
test_anthropic.py::test_anthropic_process_response_server_tool_blocks tests/models/
test_anthropic.py::test_anthropic_count_tokens_with_no_messages`
- [x] `uv run ruff check pydantic_ai_slim/pydantic_ai/models/anthropic.py tests/models/test_anthropic.py`
- [x] `uv run ruff format --check pydantic_ai_slim/pydantic_ai/models/anthropic.py tests/models/test_anthropic.py`
- [x] `uv run pyright pydantic_ai_slim/pydantic_ai/models/anthropic.py tests/models/test_anthropic.py`
### Checklist


dsfaccini:
David's AICA here: thanks @dyscott — your approach is correct, and we verified it works against **real** Bedrock: `count_tokens()` returns `input_tokens=18` via the low-level `/model/{model}/count-tokens` endpoint (the SDK blocks the high-level method on Bedrock, so this is the only path).

Cleanup we layered on top of your commits (authorship preserved):

- **Extracted** the low-level Bedrock logic into its own module, `pydantic_ai/models/_anthropic_bedrock_count_tokens.py`, so it doesn't clutter `AnthropicModel`. There's a module comment documenting *why* the SDK's high-level method can't be used, so a future reader doesn't "clean it away".
- **Replaced the mock-only test with a recorded VCR test** that exercises a real Bedrock round-trip (request envelope + parsed token count). The cassette uses `anthropic.claude-sonnet-4-20250514-v1:0` rather than the `claude-3-5-sonnet` id from the original test, because that id is now **end-of-life** on Bedrock's CountTokens endpoint. Worth noting: CRIS inference-profile ids (`us.`/`eu.`/`global.` prefixes) aren't accepted there either — only **base** foundation-model ids — so we added a short docs note about that.
- **Merged latest `main`** to resolve conflicts (your branch predated `main`'s server-tools workaround for `count_tokens`, which we kept).

The approach itself is unchanged — purely a tidy-up plus the recorded test. Thanks again for the contribution!

dsfaccini:

|---|---|---|---|---|
| 1 · code-correctness | ✅ closed | 0 | 0 | 2 |

Reviewed at 2026-06-25 · diff `31bcad04`

Two earlier informational findings have now been **addressed**:
- **Test placement** — the Bedrock `count_tokens` tests were extracted from the `test_anthropic.py` monolith into a dedicated `tests/models/test_anthropic_bedrock_count_tokens.py` feature file (cassettes moved alongside).
- **Server-tool counting (verified, not assumed)** — probed live: Bedrock's InvokeModel `/count-tokens` endpoint **rejects** server tools (`web_search`) with HTTP 400, exactly like the standard endpoint, so the strip-to-`MemoryTool` is correct on Bedrock and there is **no Bedrock-specific undercounting**. Locked in by `test_anthropic_bedrock_count_tokens_rejects_server_tools` with a recorded 400.

<details>
<summary>⚠️ 2 informational (skip)</summary>

_Automated pre-merge review — findings are pre-triage and may include false positives._

**1. `test_anthropic_bedrock_count_tokens_before_request` reuses the `_real_api` cassette by filename** · `tests/models/test_anthropic_bedrock_count_tokens.py`
`@pytest.mark.vcr('...real_api.yaml')` shares one deterministic count-tokens interaction; `input_tokens_limit=10` raises `UsageLimitExceeded` before any message request, so only the count call replays — skip: reuse is correct and documented, avoids a redundant recording.

**2. Body-key list can drift from the SDK `count_tokens()` kwargs** · `models/_anthropic_bedrock_count_tokens.py` · `count_tokens_via_bedrock`
The Bedrock path hand-marshals the body from an explicit key list while the non-Bedrock path passes kwargs to the SDK; kept in sync only by the parallel call sites — skip: inherent to bypassing the SDK's Bedrock-blocked high-level method, no shared abstraction to factor out.
</details>

tests/models/test_anthropic_bedrock_count_tokens.py
```diff
@@ -0,0 +1,242 @@
...
+    Mocks `client.post` to return a body without `inputTokens` — a shape no real Bedrock
+    CountTokens endpoint returns, so it can't be exercised through a VCR recording.
+    """
+    from anthropic import AsyncAnthropicBedrock
+
+    bedrock_client = AsyncAnthropicBedrock(
+        aws_access_key='test-access-key',
+        aws_secret_key='test-secret-key',
+        aws_region='us-east-1',
+    )
```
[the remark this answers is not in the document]
dsfaccini: David's AICA here: this is a false positive — `S106` (flake8-bandit) is not enabled in this repo's Ruff config, so `make lint` does not flag it.

`[tool.ruff.lint]` uses `select = ['E4', 'E7', 'E9', 'F']` and `extend-select = ['Q', 'RUF100', 'RUF018', 'C90', 'UP', 'I', 'D', 'TID251', 'PLW1514']`. The `S` ruleset is never selected, so `S106` never fires in `make lint` or CI's `lint` job (both green on this PR). `uv run ruff check` on this file reports `All checks passed!`; the violation only surfaces under an explicit `--select S106`, which is not the project configuration. The `'test-secret-key'` literal is a dummy value matching the existing convention across the test suite. No change needed.

tests/models/test_anthropic_bedrock_count_tokens.py
```diff
@@ -0,0 +1,242 @@
...
+    pytest.importorskip('botocore')
+
+    from anthropic import AsyncAnthropicBedrock
+
+    bedrock_client = AsyncAnthropicBedrock(
+        aws_access_key=os.environ.get('AWS_ACCESS_KEY_ID', 'test-access-key'),
+        aws_secret_key=os.environ.get('AWS_SECRET_ACCESS_KEY', 'test-secret-key'),
+        aws_session_token=os.environ.get('AWS_SESSION_TOKEN'),
+        aws_region=os.environ.get('AWS_REGION', 'us-east-1'),
+    )
```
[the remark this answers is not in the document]
dsfaccini: David's AICA here: extracted a `bedrock_client` pytest fixture (env-var-defaulted `AsyncAnthropicBedrock` + `pytest.importorskip('botocore')`) and deduped the 4 live-recorded Bedrock count-tokens tests to take it instead of constructing the client inline. The 5th test (`test_anthropic_bedrock_count_tokens_unexpected_response`, which uses hardcoded throwaway creds for a mocked `post`) is intentionally left untouched. All 5 tests stay green on cassette replay.

tests/models/test_anthropic_bedrock_count_tokens.py
```diff
@@ -0,0 +1,228 @@
...
+        ModelRequestParameters(),
+    )
+
+    assert result.input_tokens == snapshot(18)
+
+    assert len(vcr.requests) == 1  # pyright: ignore[reportUnknownMemberType,reportUnknownArgumentType]
+    assert vcr.requests[0].path == snapshot('/model/anthropic.claude-sonnet-4-20250514-v1:0/count-tokens')  # pyright: ignore[reportUnknownMemberType]
+    envelope = json.loads(vcr.requests[0].body)  # pyright: ignore[reportUnknownMemberType,reportUnknownArgumentType]
+    body = json.loads(base64.b64decode(envelope['input']['invokeModel']['body']))
+    assert body == snapshot(
```
dsfaccini: we have a utility for this

tests/models/test_anthropic_bedrock_count_tokens.py
```diff
@@ -0,0 +1,228 @@
...
+    `_messages_count_tokens` strips server tools (keeping only client-side tools like `MemoryTool`)
+    before counting, on both the standard and Bedrock paths. Posting a `web_search` server tool to the
+    low-level Bedrock endpoint directly confirms the strip is needed on Bedrock too: the endpoint 400s
+    rather than counting the tool, so Bedrock does not diverge from the standard endpoint and there is
+    no Bedrock-specific undercounting to fix.
+    """
+    from anthropic import NOT_GIVEN, BadRequestError, omit as OMIT
+
+    from pydantic_ai.models._anthropic_bedrock_count_tokens import count_tokens_via_bedrock
+
```
dsfaccini: these go up behind an import gate!!!

tests/models/test_anthropic_bedrock_count_tokens.py
dsfaccini: David's AICA here: switched to the existing `single_request_body(vcr)` helper from `tests/cassette_utils.py` (it asserts the single-request invariant and returns the parsed envelope), dropping the manual `len(vcr.requests) == 1` + `json.loads(vcr.requests[0].body)`. The inner `base64.b64decode(envelope['input']['invokeModel']['body'])` step stays — no helper covers the Bedrock-specific base64-encoded inner body, and that's the part this test is actually asserting.

tests/models/test_anthropic_bedrock_count_tokens.py
dsfaccini: David's AICA here: hoisted the function-local optional-dep imports up into the module-level `with try_import()` gate, next to the rest — `AsyncAnthropicBedrock`, `NOT_GIVEN`/`BadRequestError`/`omit as OMIT`, and `count_tokens_via_bedrock`. They're now gated like the others (and the dangling `AsyncAnthropicBedrock` `TYPE_CHECKING` import is gone since the real import resolves the annotations). The `pytest.importorskip('botocore')` stays in the fixture since botocore is only needed at request-prep time, not import time.

PR #5481, opened 2026-05-15
Convert `ModelProfile` to `TypedDict` for easier per-model overrides

### LINKED ISSUE #3796: More easily override only specific model profile fields
### Initial Checks

### Description

```python
super().__init__(settings=settings, profile=profile or provider.model_profile)
```
The above code replace the provider.model_profile with user provided one, rather than update it, which is not intuitive and cause inconvinence.
I found this because I provide a profile to OpenAIChatModel which uses DeepSeekProvider and I get
```python
openai.BadRequestError: Error code: 400 - {'error': {'message': 'Missing `reasoning_content` field in the assistant message at message index 2. For more information, please refer to https://api-docs.deepseek.com/guides/thinking_mode#tool-calls', 'type': 'invalid_request_error', 'param': None, 'code': 'invalid_request_error'}}
```

### Example Code

```Python

```

```Text
3.14, 1.37.0
```

DouweM:
Re-applies [#5340](https://github.com/pydantic/pydantic-ai/pull/5340) (which was merged then reverted in [#5449](https://github.com/pydantic/pydantic-ai/pull/5449)), rebased onto `v2-main` after the post-revert cleanup commits (drops of `builtin_tools`, Outlines, and the deprecation shim file).

- Closes #3796

### What changed since the reverted PR

1. **Lock-in snapshot test added** (`tests/profiles/test_resolution_lockin.py`) — 83 cases covering every gateway, every cross-provider routing case (Bedrock-Anthropic, OpenRouter-Google, Vercel-Vertex, etc.), and every inference provider. Snapshots use a `_normalize` helper that strips canonical defaults; result is the per-route delta the provider actually contributes. Used to verify the migration produces equivalent resolved profiles.
2. **Migration recipes promoted to `docs/changelog.md`** — v2.0.0 section with literal old→new code samples, designed both for human readers and for an LLM agent grepping the file for breakage patterns.

### Why `TypedDict` (recap from #5340)

`ModelProfile` and its subclasses were `@dataclass(kw_only=True)` while the closely-related `ModelSettings` was already a `TypedDict(total=False)`. That asymmetry produced three structural pain points:

1. **Cross-provider routing forced "downsize then upsize."** Bedrock hosting Anthropic models had to build a `BedrockModelProfile`, call `.update()` with the `AnthropicModelProfile`, then `replace(...)` to override one field. Class identity made mixing awkward.
2. **`ModelProfile.update()` was leaky.** The `f.default == value` heuristic couldn't distinguish "user explicitly set X to its default" from "user didn't pass X."
3. **Two mental models for one concept.** Users had to learn dataclass semantics for `ModelProfile` and TypedDict semantics for `ModelSettings`.

Plus the original issue: overriding a single field on a built-in profile required constructing a full `OpenAIModelProfile(...)` from scratch.

### Lock-in test findings

73 of 83 cases produce identical resolved profiles pre- and post-migration. **10 cases show drift, all from the same root cause:** v1's `ModelProfile.update()` silently filtered out fields not declared on the target class. v2's `merge_profile` (dict-spread) preserves all keys. So e.g. Bedrock-Anthropic profiles now carry `anthropic_*` fields alongside `bedrock_*` fields. Detailed in the changelog entry; no in-tree model class reads cross-class fields, so behavior is unchanged in the standard providers.

The per-field ordering classification (fallback vs upstream-wins vs override) for the 6 three-layer providers (OpenRouter, Azure, Grok, Ollama, DeepSeek, Cerebras) was verified to produce identical resolved profiles for the high-risk fields (`supports_json_schema_output` override on Bedrock-Anthropic, `openai_supports_tool_choice_required` per-model logic on DeepSeek, `json_schema_transformer` fallback in all 3-layer routes).

### Checklist

- [ ] No **breaking changes** in accordance with the [version policy](https://github.com/pydantic/pydantic-ai/blob/main/docs/version-policy.md). _(v2 PR — breaking changes are allowed and documented in `docs/changelog.md`.)_

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -40,7 +40,7 @@ In v1, `ModelProfile.update()` silently filtered out fields not declared on the 
 
 This means e.g. a Bedrock-hosted Anthropic model's resolved profile now carries the upstream `anthropic_*` fields alongside the `bedrock_*` fields, where v1 dropped them. No in-tree model class reads cross-class fields, so behavior is unchanged in the standard providers; but custom model classes that do `profile.get('anthropic_supports_adaptive_thinking', False)` on a non-Anthropic route will now see the value the upstream Anthropic profile set, where v1 always returned the default.
 
-See [PR #TBD](https://github.com/pydantic/pydantic-ai/pull/0) for the full ModelProfile redesign.
+See [PR #5481](https://github.com/pydantic/pydantic-ai/pull/5481) for the full ModelProfile redesign.
 
 ### v1.0.1 (2025-09-05)
 
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -6,6 +6,42 @@ In September 2025, Pydantic AI reached V1, which means we're committed to API st
 
 Here's a filtered list of the breaking changes for each version to help you upgrade Pydantic AI.
 
+### v2.0.0 (unreleased)
+
+#### `ModelProfile` is now a `TypedDict`
+
+`ModelProfile` and all its subclasses (`OpenAIModelProfile`, `AnthropicModelProfile`, `GoogleModelProfile`, `BedrockModelProfile`, etc.) are now `TypedDict(total=False)` instead of `@dataclass`. This unifies the mental model with [`ModelSettings`][pydantic_ai.settings.ModelSettings] (also a `TypedDict`) and enables direct dict-spread for cross-class merging.
+
+`ModelProfile.update()` and `ModelProfile.from_profile()` are removed; use the module-level [`merge_profile`][pydantic_ai.profiles.merge_profile] (later argument wins per key).
+
+Migration recipes:
+
+| v1 (dataclass) | v2 (TypedDict) |
+|---|---|
+| `OpenAIModelProfile(field=value)` | Same syntax; returns a partial `dict` instead of a fully-defaulted instance. |
+| `profile.field` (attribute read) | `profile.get('field', <default>)` |
+| `profile.field = value` (attribute write) | `profile['field'] = value` |
+| `dataclasses.replace(profile, field=value)` | `{**profile, 'field': value}` or `merge_profile(profile, ModelProfile(field=value))` |
+| `profile.update(other)` | `merge_profile(profile, other)` |
+| `OpenAIModelProfile.from_profile(p)` | Just `p` — no upcasting needed |
+| `Model(name, profile=full_profile)` (full replace) | Now merges on top of the provider's default profile — usually what you want. For a hard replace use `Model(name, profile=lambda _default: full_profile)`. |
+| `Model(name, profile=fn)` where `fn: Callable[[str], ModelProfile \| None]` | Removed — the user-passed callable is now `Callable[[ModelProfile], ModelProfile]`, receiving the resolved default and returning the final profile. The `(model_name: str) -> ModelProfile \| None` shape is still accepted internally by `Provider.model_profile`. |
+| `isinstance(profile, OpenAIModelProfile)` | Not supported by `TypedDict` at runtime — raises `TypeError`. Use `isinstance(profile, dict)` or check key presence (`'openai_chat_supports_web_search' in profile`). Pyright still narrows correctly via the TypedDict subclass annotation. |
+
+`Model.profile` is now the single source of truth for the **resolved** profile. It is composed by [`merge_profile`][pydantic_ai.profiles.merge_profile] in this order (later wins):
+
+1. [`DEFAULT_PROFILE`][pydantic_ai.profiles.DEFAULT_PROFILE] — base defaults for every documented key.
+2. `Provider.model_profile(model_name)` — provider/model-specific resolution.
+3. The user's `profile=` argument — either a partial dict (merged on top) or a `Callable[[ModelProfile], ModelProfile]` (full control: receives the resolved default, returns the final profile).
+
+#### Resolved profiles now carry cross-class fields
+
+In v1, `ModelProfile.update()` silently filtered out fields not declared on the target class. In v2, dict-spread preserves every key.
+
+This means e.g. a Bedrock-hosted Anthropic model's resolved profile now carries the upstream `anthropic_*` fields alongside the `bedrock_*` fields, where v1 dropped them. No in-tree model class reads cross-class fields, so behavior is unchanged in the standard providers; but custom model classes that do `profile.get('anthropic_supports_adaptive_thinking', False)` on a non-Anthropic route will now see the value the upstream Anthropic profile set, where v1 always returned the default.
+
+See [PR #TBD](https://github.com/pydantic/pydantic-ai/pull/0) for the full ModelProfile redesign.
+
 ### v1.0.1 (2025-09-05)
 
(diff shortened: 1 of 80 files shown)


pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -993,26 +1001,42 @@ def supported_native_tools(cls) -> frozenset[type[AbstractNativeTool]]:
...
 
-        if _profile is None:
-            _profile = DEFAULT_PROFILE
+        After resolution we compute the intersection of the profile's `supported_native_tools`
+        and the model class's implemented tools, ensuring `model.profile['supported_native_tools']`
+        is the single source of truth for what's actually usable.
+        """
+        # Step 1+2: provider default merged with base default
+        provider_profile: ModelProfile = {}
+        if (provider := getattr(self, '_provider', None)) is not None:
```
DouweM: I don't like the `getattr` one bit; should we make `_provider` "official" with a public `provider` getter on `Model`?

pydantic_ai_slim/pydantic_ai/models/ollama.py
```diff
@@ -90,6 +90,6 @@ def __init__(
         if profile is None and _routes_to_ollama_cloud(provider, model_name):
             base_profile = provider.model_profile(model_name)
             assert base_profile is not None  # OllamaProvider always returns a profile
-            profile = replace(base_profile, supports_json_schema_output=False)
+            profile = merge_profile(base_profile, ModelProfile(supports_json_schema_output=False))
```
DouweM, line 93: Could we just use `{*base_profile, '...'}` here? Does that "type check" correctly?

pydantic_ai_slim/pydantic_ai/profiles/harmony.py
```diff
@@ -9,7 +9,7 @@ def harmony_model_profile(model_name: str) -> ModelProfile | None:
 
     See <https://cookbook.openai.com/articles/openai-harmony> for more details.
     """
-    profile = openai_model_profile(model_name)
-    return OpenAIModelProfile(
-        openai_supports_tool_choice_required=False, ignore_streamed_leading_whitespace=True
-    ).update(profile)
+    return merge_profile(
+        OpenAIModelProfile(openai_supports_tool_choice_required=False, ignore_streamed_leading_whitespace=True),
+        openai_model_profile(model_name),
```
DouweM, line 13: Do we need `merge_profile` in a case like this is? Is it because of `None` handling, perhaps?

pydantic_ai_slim/pydantic_ai/providers/azure.py
```diff
@@ -67,17 +68,19 @@ def model_profile(model_name: str) -> ModelProfile | None:
...
+                    OpenAIModelProfile(json_schema_transformer=OpenAIJsonSchemaTransformer),
+                    profile,
+                    OpenAIModelProfile(openai_chat_supports_document_input=False),
+                )
+
+        # OpenAI models are unprefixed.
+        # Azure Chat Completions API doesn't support document input — gateway override wins on top of upstream.
+        return merge_profile(
+            openai_model_profile(model_name),
+            OpenAIModelProfile(openai_chat_supports_document_input=False),
```
DouweM: Code can be simplified to only set this once.

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -1949,8 +1955,8 @@ def _get_task_budget(self, model_settings: AnthropicModelSettings) -> AnthropicT
         if task_budget is None:
             return None
 
-        profile = AnthropicModelProfile.from_profile(self.profile)
-        if not profile.anthropic_supports_task_budgets:
+        profile = cast(AnthropicModelProfile, self.profile)
```
DouweM: Is it feasible to override `self.profile` with the cast, even if it costs us an "override type ignore" comment, because we know it's valid type wise?

pydantic_ai_slim/pydantic_ai/models/groq.py
```diff
@@ -710,8 +715,10 @@ async def _get_event_iterator(self) -> AsyncIterator[ModelResponseStreamEvent]:
                         for event in self._parts_manager.handle_text_delta(
                             vendor_part_id='tool_use_failed',
                             content=failed_generation,
-                            thinking_tags=self._model_profile.thinking_tags,
-                            ignore_leading_whitespace=self._model_profile.ignore_streamed_leading_whitespace,
+                            thinking_tags=self._model_profile.get('thinking_tags', ('<think>', '</think>')),
```
DouweM: It is a shame to have to repeat defaults like this across all models. Let's have a constant to import for all that are non-trivial.

docs/changelog.md
```diff
@@ -6,6 +6,42 @@ In September 2025, Pydantic AI reached V1, which means we're committed to API st
 
 Here's a filtered list of the breaking changes for each version to help you upgrade Pydantic AI.
 
+### v2.0.0 (unreleased)
+
+#### `ModelProfile` is now a `TypedDict`
+
+`ModelProfile` and all its subclasses (`OpenAIModelProfile`, `AnthropicModelProfile`, `GoogleModelProfile`, `BedrockModelProfile`, etc.) are now `TypedDict(total=False)` instead of `@dataclass`. This unifies the mental model with [`ModelSettings`][pydantic_ai.settings.ModelSettings] (also a `TypedDict`) and enables direct dict-spread for cross-class merging.
```
DouweM: Link to the existing docs section on what model profiles are.

docs/changelog.md
```diff
@@ -6,6 +6,42 @@ In September 2025, Pydantic AI reached V1, which means we're committed to API st
...
+
+`ModelProfile.update()` and `ModelProfile.from_profile()` are removed; use the module-level [`merge_profile`][pydantic_ai.profiles.merge_profile] (later argument wins per key).
```
DouweM, line 17: Have we updated the mkdocs api files etc?

docs/changelog.md
```diff
@@ -6,6 +6,42 @@ In September 2025, Pydantic AI reached V1, which means we're committed to API st
...
+
+Migration recipes:
+
+| v1 (dataclass) | v2 (TypedDict) |
+|---|---|
+| `OpenAIModelProfile(field=value)` | Same syntax; returns a partial `dict` instead of a fully-defaulted instance. |
+| `profile.field` (attribute read) | `profile.get('field', <default>)` |
```
DouweM: Specify that constants are available in the module

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -2251,7 +2260,7 @@ def _get_responses_tool_choice(
             A tuple of (filtered_function_tools, tool_choice).
             Note: builtin tools are handled separately and should be added to this list.
         """
-        openai_profile = OpenAIModelProfile.from_profile(self.profile)
+        openai_profile = cast(OpenAIModelProfile, self.profile)
```
DouweM: I'd prefer not to need this with some "type ignore" magic on an overridden method, as long as pyright handles that correclty.

PR #5482, opened 2026-05-16
fix: auto-heal invalid encrypted content responses

### LINKED ISSUE #4608: OpenAIResponsesModel: invalid_encrypted_content error when using API gateway/proxy across orgs

pragnyanramtha:
- Closes #4608

### Summary

- retry a non-streaming OpenAI Responses request once when OpenAI returns `invalid_encrypted_content`
- only retry when the original request actually sent encrypted reasoning content
- retry with `openai_send_reasoning_ids=False` so the existing history can auto-heal without disabling reasoning IDs globally
- add positive and negative regression coverage for the retry path

### Notes

This follows the approach discussed by @DouweM in #4608: keep the normal fast path unchanged, but auto-heal the rare cross-org/gateway failure by retrying without reasoning IDs. The retry is intentionally narrow and one-shot to avoid retry loops.

### Validation

- `uv run pytest tests/models/test_openai_responses.py -k 'invalid_encrypted_content or non_matching_error'`
- `uv run ruff check pydantic_ai_slim/pydantic_ai/models/openai.py tests/models/test_openai_responses.py`
- `uv run pyright pydantic_ai_slim/pydantic_ai/models/openai.py tests/models/test_openai_responses.py`
- `git diff --check`

### Checklist


pragnyanramtha:
Rebased/merged the branch on current `main` and resolved the OpenAI Responses conflict from the new `_build_responses_request_params` refactor.

Resolution kept the retry semantics intact:
- retry checks `request_params.input` for encrypted reasoning content
- retry rebuilds from the original `messages` with `openai_send_reasoning_ids=False`
- Azure content-filter handling still runs before the retry branch
- streaming remains excluded from this retry path

Validation run after resolving:
- `uv run pytest tests/models/test_openai_responses.py -k 'invalid_encrypted_content or non_matching_error'` -> 4 passed
- `uv run ruff check pydantic_ai_slim/pydantic_ai/models/openai.py tests/models/test_openai_responses.py`
- `uv run pyright pydantic_ai_slim/pydantic_ai/models/openai.py tests/models/test_openai_responses.py`
- `git diff --check`

PR #5487, opened 2026-05-16
fix(instrumentation): Prevent false positive `variable_instructions` span attribute

### LINKED ISSUE #5394: False-positive `pydantic_ai.variable_instructions=True` when using PromptedOutput or other model-injected instructions
## Bug

The OTel agent run span attribute `pydantic_ai.variable_instructions` can be incorrectly set to `True` when the agent didn't actually vary its instructions — specifically when the model layer appends instructions (e.g. for `PromptedOutput`) that aren't reflected on the per-request `ModelRequest.instructions` field.

## Mechanism

In the end-of-run span-attributes computation:

```python
last_instructions = get_instructions(message_history, model_request_parameters)
...
if any(
    (isinstance(m, ModelRequest) and m.instructions is not None and m.instructions != last_instructions)
    for m in message_history[new_message_index:]
):
    attrs['pydantic_ai.variable_instructions'] = True
```

- `last_instructions` resolves through `Model._get_instruction_parts(...)` → joined `instruction_parts`, which **includes** the prompted-output template and other model-layer instructions added during the request.
- `m.instructions` on each `ModelRequest` is the raw string set at request-construction time, which does **not** include those model-layer additions.

So with `PromptedOutput` (or any path that uses `params.prompted_output_instructions`):

- `last_instructions` = `base + prompted-output template`
- `m.instructions` = `base`
- `m.instructions != last_instructions` → flag set, even though the user's instructions never changed.

This is a false positive: the signal is documented as indicating that the agent's instructions varied between steps, but here only the model added formatting on top of the user's stable instructions.

## Affected versions

This bug exists on `main` today (logic introduced in #3131 / `2631d9b85`) and is preserved unchanged through PR #4967 (which ports instrumentation to the `Instrumentation` capability — same code, new location at `pydantic_ai_slim/pydantic_ai/capabilities/instrumentation.py` lines 194–198).

## Repro

Any agent using `output_type=PromptedOutput(...)` with instrumentation enabled and looking at the `pydantic_ai.variable_instructions` attribute on the agent run span.

## Suggested fix

Compare like with like — either:

1. Run `m.instructions` through the same `instruction_parts` synthesis path before comparing, or
2. Drop the `variable_instructions` signal entirely if it's no longer load-bearing (the v2 instrumentation pass is expected to revisit this).

madanlalit:
Closes #5394

### Summary
Resolves a bug where the `pydantic_ai.variable_instructions` OpenTelemetry span attribute was incorrectly triggering as `True` when using model-injected instructions.

### Root Cause 
When you use `PromptedOutput`, the library automatically appends a JSON schema template to the end of your instructions. The old code was doing a mismatched post-hoc check: it was comparing your original (unformatted) instructions against the final instructions (which included the new schema text). Because they didn't match, it wrongly assumed your instructions were dynamically changing on every request.

### Fix
We moved the check into the `wrap_model_request` hook so that it now compares the fully formatted instructions (including the JSON schema) request-to-request. By caching and evaluating the final payload immediately before it's sent to the model, we achieve a true "like to like" comparison. This guarantees the flag only triggers if the finalized prompt actually changes between steps, completely avoiding re-computation at span-finalization.

*Includes new tests to guarantee this works correctly for both v2 and v3 instrumentation across single and multi-step runs.*

### Checklist

- [X] **PR title** is fit for the [release changelog](https://github.com/pydantic/pydantic-ai/releases).

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/capabilities/instrumentation.py
+++ b/pydantic_ai_slim/pydantic_ai/capabilities/instrumentation.py
@@ -19,8 +19,9 @@ from pydantic_ai._instrumentation import (
     open_model_request_span,
     serialize_any,
 )
+from pydantic_ai._utils import UNSET, Unset
 from pydantic_ai.exceptions import ApprovalRequired, CallDeferred, ToolRetryError
-from pydantic_ai.messages import ModelMessage, ModelRequest, ModelResponse, ToolCallPart, tool_return_ta
+from pydantic_ai.messages import ModelMessage, ModelResponse, ToolCallPart, tool_return_ta
 from pydantic_ai.tools import ToolDefinition
 
 from .abstract import (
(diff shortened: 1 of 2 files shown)


madanlalit:
@DouweM Please Review.

madanlalit:
@DouweM you might have missed this.

DouweM:
@madanlalit Thank you sir :)

PR #5499, opened 2026-05-17
docs: add typed task receipt example

aDragon0707:
## Summary

Adds a small structured output docs example for typed task receipts.

The example shows how to model a final-state receipt with Pydantic models that separate:

- agent claims
- supporting evidence
- unsupported or unverified claims
- next owner
- human approval boundaries

## Motivation

Agent final answers can include operational claims such as tests passed, files changed, ready to merge, or ready to publish.

Those claims may require evidence from tool outputs, logs, traces, file diffs, command output, approvals, or human review before a workflow treats the task as complete.

Pydantic AI structured output is a natural fit for this because the receipt fields can be validated and typed instead of parsed from free-form final text.

## Notes

- Docs-only change.
- No runtime dependency.
- No behavior change.
- Safety-oriented structured output example.

## Verification

- Compiled the added `completion_receipt.py` docs snippet with Python `compile(...)`.
- `git diff --check`
- Not run: full docs build because `uv` is not installed in this local environment.

### DIFF
--- a/docs/output.md
+++ b/docs/output.md
@@ -29,6 +29,65 @@ print(result.usage)
 
 _(This example is complete, it can be run "as is")_
 
+## Receipts for final-state claims {#completion-receipts}
+
+Structured output is also useful when an agent's final answer contains operational claims that need to be reviewed, such as "tests passed", "files were changed", or "ready to publish". A typed receipt can separate each claim from its evidence and make human approval boundaries explicit.
+
+For example:
+
+```python {title="completion_receipt.py" line_length="90"}
+from typing import Literal
+
+from pydantic import BaseModel, Field
+
+from pydantic_ai import Agent
+
+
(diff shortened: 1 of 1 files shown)


hramezani:
Thanks for putting this together, @aDragon0707 — the "separate each claim from its evidence" idea is a genuinely nice use of structured output. 🙏

That said, we're going to pass on this one. We try to keep the core docs focused on teaching Pydantic AI's features and general primitives rather than specific agent-design patterns — and the structured-output mechanics this demonstrates (nested models, `Literal` fields, typed lists) are already covered by the existing examples on the page. The "completion receipt" framing is more of an opinionated workflow pattern, which is better suited to a blog post or a cookbook-style writeup than our reference docs.

Really appreciate the thought you put into it, and hope to see more contributions from you!

PR #5503, opened 2026-05-18
Stop requiring `httpx`; use `httpx2` for internal HTTP

dsfaccini:
### Summary

Installing `pydantic-ai` no longer installs `httpx`, and the HTTP we own end to end runs on [`httpx2`](https://github.com/pydantic/httpx2).

`httpx` survives only at the provider SDK boundary, where it is unavoidable: `openai`, `anthropic`, `google-genai`, `groq` and friends all run `isinstance(http_client, httpx.AsyncClient)` and raise `TypeError: Invalid http_client argument` for anything else. That boundary keeps taking real `httpx` clients — but they now arrive transitively, from whichever provider SDK the user installed, rather than from us.

### What changes

**Packaging**

- `pydantic-ai-slim` drops `httpx>=0.27` from its dependencies and keeps `httpx2>=2.0`. `pydantic-graph` is `httpx2`-only.
- The `retries` extra gains `httpx>=0.27` — `retries.py` genuinely subclasses `httpx` transports, for clients handed to provider SDKs, and `tenacity` doesn't bring `httpx` along.

**Internal HTTP moves to `httpx2`**

- `pydantic_ai._ssrf` — SSRF-protected `safe_download` builds its own `httpx2.AsyncClient` rather than going through `create_async_http_client()`, which still returns `httpx.AsyncClient` for the SDK boundary. It now raises plain `httpx2` errors.
- `pydantic_ai.common_tools.web_fetch` — fetches via that client.
- `pydantic_ai.ui._web.app` — the chat-UI HTML download (the `web` extra already declared `httpx2`).

**The SDK-less core stops importing `httpx`**

- `models/__init__.py`, `providers/__init__.py`, `settings.py` — annotations moved under `if TYPE_CHECKING`; the two runtime sites in `models/__init__.py` (`create_async_http_client()` and `StreamedResponse.get_stream_cancel_errors()`) import lazily. The latter wraps the import in `try`/`except ImportError`, because its caller spells the guard `except self.get_stream_cancel_errors():`, which Python evaluates for *any* exception passing through — an early `break`'s `GeneratorExit` included.
- `models/test.py` — `TestModel` no longer borrows `httpx.StreamClosed` to simulate a cancelled stream; it raises its own error and overrides `get_stream_cancel_errors()`. `Agent('test')` is the keyless path and shouldn't need an SDK's HTTP library.
- Provider modules, `mcp.py`, `models/{openai,mistral}.py` keep their top-level `import httpx`: none is reachable without a package that depends on `httpx`.

**Tests**

- `tests/test_httpx2_sdk_readiness.py` — two guards. `test_core_runs_without_httpx` blocks `httpx` on `sys.meta_path` in a subprocess and runs an agent, so the packaging claim is enforced rather than asserted. `test_sdk_still_rejects_httpx2_client` pins the SDK rejection, and so goes red the moment an SDK ships `httpx2` support — that's the signal the boundary can move.
- `tests/test_ssrf.py` asserts `safe_download`'s errors are *not* catchable as `httpx.*`, so the coupling can't come back by accident.

### What does NOT change

- Every `Provider` keeps its `http_client: httpx.AsyncClient | None` parameter and its `httpx_client` / `httpx_async_client` attribute names. `create_async_http_client()` still returns an `httpx.AsyncClient`.
- MCP, retry transports, `ModelSettings.timeout`, and every model passing `http_client` into an SDK stay on `httpx`.
- `logfire[httpx]` — an `httpx2` extra would need `opentelemetry-instrumentation-httpx2`, which doesn't exist on PyPI yet.

### Breaking change

`safe_download` (behind `ImageUrl` / `DocumentUrl` / `AudioUrl` / `VideoUrl` downloads and `web_fetch`) now raises `httpx2.HTTPStatusError` / `httpx2.RequestError`. Code catching the `httpx` equivalents around an agent run needs to catch the `httpx2` ones instead.

An earlier revision kept them catchable as both via multiply-inherited compat classes. That was dropped: it pinned `httpx` as a required dependency until v3, which is the opposite of the point.

### Checklist

</content>


pydantic_ai_slim/pydantic_ai/_ssrf.py
```diff
@@ -342,19 +342,22 @@ async def safe_download(
...
-        httpx.HTTPStatusError: If the response has an error status code.
+        httpx2.HTTPStatusError: If the response has an error status code.
     """
     current_url = url
     redirects_followed = 0
     original_hostname = urlparse(url).hostname
     effective_headers = dict(headers) if headers else {}
 
-    async with create_async_http_client(timeout=timeout) as client:
+    async with httpx2.AsyncClient(
```
dsfaccini: this needs to go back to being `create_async_http_client` when SDKs support httpx

pydantic_ai_slim/pyproject.toml
```diff
@@ -127,7 +127,7 @@ a2a = ["fasta2a>=0.6.1"]
 # AG-UI
 ag-ui = ["ag-ui-protocol>=0.1.10", "starlette>=0.45.3"]
 # Web
-web = ["starlette>=0.45.3", "httpx>=0.27.0", "uvicorn>=0.38.0"]
+web = ["starlette>=0.45.3", "httpx>=2.0", "uvicorn>=0.38.0"]
```
Kludex: This is a mistake.


```suggestion
web = ["starlette>=0.45.3", "httpx2>=2.0", "uvicorn>=0.38.0"]
```

pydantic_ai_slim/pyproject.toml
```diff
@@ -55,7 +55,7 @@ Changelog = "https://github.com/pydantic/pydantic-ai/releases"
 [tool.hatch.metadata.hooks.uv-dynamic-versioning]
 dependencies = [
     "griffelib>=2.0",
-    "httpx>=0.27",
+    "httpx>=2.0",
```
Kludex: Ok... Please review this PR. 👀

mkdocs.yml
```diff
@@ -368,7 +368,7 @@ plugins:
             - url: https://typing-extensions.readthedocs.io/en/latest/objects.inv
             - url: https://rich.readthedocs.io/en/stable/objects.inv
             - url: https://modelcontextprotocol.github.io/python-sdk/objects.inv
-            # waiting for https://github.com/encode/httpx/discussions/3091#discussioncomment-11205594
+            # waiting for an objects.inv from https://www.python-httpx.org/ (originally tracked in https://github.com/encode/httpx/discussions/3091#discussioncomment-11205594)
```
Kludex: We already have the objects.inv in httpx2, you can already use it.

tests/test_web_fetch.py
```diff
@@ -4,7 +4,7 @@
 
 from unittest.mock import AsyncMock, patch
 
-import httpx
+import httpx2
```
Kludex: I'm not sure myself, but should we do `from httpx2 import httpx` for simplicity?

dsfaccini: `import httpx2 as httpx`?

Kludex: yes. I'm not sure about this.

mkdocs.yml
dsfaccini: David's AICA here: pointed at `https://httpx2.pydantic.dev/objects.inv` in ca8b4f9e2 — but probing that URL currently returns 404. zensical [emits objects.inv to the site root](https://github.com/zensical/zensical/blob/master/crates/zensical/src/workflow.rs) so the path is correct in principle; looks like the deployed httpx2 site predates that or the deploy needs rerunning. Want me to leave the inventory entry in (forward-compatible) or remove until httpx2 redeploys?

tests/test_web_fetch.py
dsfaccini: David's AICA here: `from httpx2 import httpx` doesn't work — `httpx2` 2.2.0 doesn't expose an `httpx` submodule (verified locally). Went with `import httpx2 as httpx` instead in ca8b4f9e2 across all six touched files; gives the same simplification — `httpx.X` references stay unchanged across the diff.

pydantic_ai_slim/pydantic_ai/_ssrf.py
dsfaccini: David's AICA here: captured as inline `# NOTE:` comments at the `_create_client` definition and at the call site in `safe_download` (ca8b4f9e2). Will revert when provider SDKs accept httpx2.

pydantic_ai_slim/pyproject.toml
dsfaccini: David's AICA here: Done — applied your suggestion in ca8b4f9e2.

pydantic_ai_slim/pyproject.toml
dsfaccini: David's AICA here: Same typo, fixed in ca8b4f9e2.

pydantic_ai_slim/pydantic_ai/_ssrf.py
```diff
@@ -12,13 +12,48 @@
...
+        self.request = request
+        self.response = response
+
+
+class _SafeDownloadRequestError(  # pyright: ignore[reportIncompatibleMethodOverride, reportIncompatibleVariableOverride]
+    httpx.RequestError, _httpx_legacy.RequestError
+):
+    def __init__(self, message: str, *, request: httpx.Request) -> None:
+        Exception.__init__(self, message)
+        self._request = request
```
dsfaccini: David's AICA here:

Flagging for next week's team conversation.

These compat classes multi-inherit from both `httpx2.*` (the actual exception types raised by the underlying client) and legacy `httpx.*`, so user code wrapping `Agent.run()` with `ImageUrl` / `DocumentUrl` keeps working whether it catches via `httpx.HTTPStatusError` or `httpx2.HTTPStatusError`. v2 cleanup is one-line per class: drop the legacy parent.

What we could NOT do: emit a `PydanticAIDeprecationWarning` specifically when user code catches via the legacy `httpx.*` class. Python has no hook to observe which `except` clause matched an exception — the warning would have to fire unconditionally inside `safe_download` (noisy, fires even for users who never catch httpx exceptions at all), or never. So the deprecation lives as a docstring/release-note item rather than a runtime warning.

Question for the team: do we want to fire an unconditional warning at the `safe_download` raise site (annoying for users who don't catch httpx exceptions at all) or accept docs-only deprecation (current state)? If the latter, we'll add the migration note to the v2 cards and the changelog when this lands.

Kludex: I don't think this is the approach we want to go with. We want to allow people to migrate away from `httpx`.

Also, we should probably not include this much of comments.

dsfaccini: I don't really know what to do here bc we can no longer do the deprecate/remove dance, next release is probably gonna be v2 stable

so we either catch both (I can add a deprecation warning for the httpx branch) or not handle httpx at all but users will get an actual error

dsfaccini:
SDKs still assert the client's class (openai, anthropic, google-genai SDKs) which prevents us from fully replacing httpx2

tests/_vcr_httpcore2_compat.py
Kludex, line 1: I think they support httpx2 now?

pydantic_ai_slim/pydantic_ai/_ssrf.py
```diff
@@ -9,16 +9,54 @@
...
+# the pyright suppressions acknowledge that these only ever hold `httpx2` objects.
+
+
+class _SafeDownloadHTTPStatusError(  # pyright: ignore[reportIncompatibleMethodOverride, reportIncompatibleVariableOverride]
+    httpx.HTTPStatusError, _httpx_legacy.HTTPStatusError
+):
+    def __init__(self, message: str, *, request: httpx.Request, response: httpx.Response) -> None:
+        Exception.__init__(self, message)
+        self.request = request
+        self.response = response
```
Kludex: I don't think those structures make sense given that httpx stops being a dependency.

dsfaccini:
This comment was posted by Claude Code using claude-opus-5 on behalf of David. David reviewed this lightly.

@Kludex all three of your open comments were right, and acting on them got us further than expected: **installing `pydantic-ai` no longer installs `httpx`.** It arrives transitively, from whichever provider SDK you install. PR title and description updated to match.

That didn't need to wait for the SDKs. Going through every `httpx` import in `pydantic-ai-slim`, the only genuinely core *runtime* uses were two spots in `models/__init__.py` — everything else was a lazy annotation or already extras-gated. So: annotations under `TYPE_CHECKING`, those two lazy, `retries.py`'s `httpx` into the `retries` extra, gone from `[dependencies]`.

I kept the nominal `httpx` annotation on the `http_client` params rather than a structural `Protocol`. A `Protocol` would let an `httpx2` client type-check and then blow up inside the SDK's `isinstance` — a false signal, for no extra dependency win.

<details><summary>Also addressed from your comments</summary>

- **Compat exception classes deleted.** `safe_download` raises plain `httpx2` errors. Since v2 already shipped without them, their removal horizon was v3 — they'd have pinned `httpx` for a whole major. There's now a test asserting the errors are *not* catchable as `httpx.*`.
- **`tests/_vcr_httpcore2_compat.py` deleted** — right again, vcrpy 8.2.1 ships a native `httpx2` patcher and that's already our locked floor. Verified cassette replay with the network hard-blocked.
- `ui/_web/app.py` → `httpx2` (CDN fetch for the chat UI; its `web` extra already declared `httpx2`).

</details>

<details><summary>How the packaging claim is enforced</summary>

`tests/test_httpx2_sdk_readiness.py::test_core_runs_without_httpx` blocks `httpx` on `sys.meta_path` in a subprocess, then imports and runs an agent. It caught a real gap while I was writing it: `StreamedResponse`'s cancel guard is spelled `except self.get_stream_cancel_errors():`, and Python evaluates that for *every* exception passing through it — including the `GeneratorExit` of an ordinary early `break` — so the lazy import there has to be `try`/`except ImportError`, or an httpx-less install masks the real error.

The file's other test, `test_sdk_still_rejects_httpx2_client`, pins the `isinstance` rejection in openai / anthropic / groq. It goes red the moment one of them accepts an `httpx2` client, which is our signal that the boundary can move.

</details>

<details><summary>SDK blocker status (re-verified today)</summary>

Every provider SDK still runs `isinstance(http_client, httpx.AsyncClient)`; none references `httpx2` anywhere.

| SDK | version | assert site |
|---|---|---|
| `openai` | 2.45.0 | `openai/_base_client.py:894,1504` |
| `anthropic` | 0.109.0 | `anthropic/_base_client.py:1009,1751` |
| `google-genai` | 1.70.0 | `google/genai/_interactions/_base_client.py:879,1461` |
| `groq` | 0.25.0 | `groq/_base_client.py:837,1335` |

Anthropic has an httpx2-capable release coming, so that one resolves itself. The others have no date — but they no longer block anything here.

</details>

One thing that does need a call from a human, not from me: `safe_download` now raises `httpx2` errors, so `except httpx.HTTPStatusError` around an agent run breaks. Flagged in the description; the version-policy checkbox is unticked accordingly.
</content>

tests/_vcr_httpcore2_compat.py
dsfaccini: This reply was posted by Claude Code using claude-opus-5 on behalf of David. David reviewed this lightly.

Confirmed — vcrpy 8.2.1 has `CassettePatcherBuilder._httpx2`, and that's already our locked floor (it arrived on `main` via #6136, not this PR). Shim deleted in 3bc5e30af, along with its `conftest.py` import and coverage `omit` entry. Verified replay still works with the network hard-blocked, and with a cassette moved aside as a negative control.

pydantic_ai_slim/pydantic_ai/_ssrf.py
dsfaccini: This reply was posted by Claude Code using claude-opus-5 on behalf of David. David reviewed this lightly.

Agreed, and acted on: `httpx` is now out of `pydantic-ai-slim`'s dependencies entirely (5cab75d65), so those structures are gone rather than adjusted. `safe_download` raises plain `httpx2` errors, with a test asserting they are *not* catchable as `httpx.*`.

pydantic_ai_slim/pydantic_ai/_ssrf.py
dsfaccini: This reply was posted by Claude Code using claude-opus-5 on behalf of David. David reviewed this lightly.

You were right on both counts. The compat classes are deleted (3bc5e30af) and the comment block with them — and `httpx` has now left `[dependencies]` altogether (5cab75d65), which is the migration path you were pointing at. My "we can no longer do the deprecate/remove dance" reply was the wrong frame: keeping them would have pinned `httpx` until v3, not until the next release.

tests/test_web_fetch.py
dsfaccini: This reply was posted by Claude Code using claude-opus-5 on behalf of David. David reviewed this lightly.

Settled on `import httpx2 as httpx` — `from httpx2 import httpx` isn't a thing (no such submodule). Same simplification: every `httpx.X` reference in the diff stays as-is.

pydantic_ai_slim/pydantic_ai/_ssrf.py
dsfaccini: This reply was posted by Claude Code using claude-opus-5 on behalf of David. David reviewed this lightly.

Moot — the compat classes are deleted, so there's no legacy `httpx` catch to warn about. `safe_download` raises `httpx2` errors directly; the break is documented in the PR description instead.

pydantic_ai_slim/pydantic_ai/providers/__init__.py
```diff
@@ -9,15 +9,25 @@
...
+if TYPE_CHECKING:
+    from httpx import AsyncClient
+else:
+    try:
+        from httpx import AsyncClient
+    except ImportError:
+        # `httpx` reaches us only transitively, via the provider SDKs. `Provider` is public and
+        # subclassable, so its annotations must stay resolvable for `typing.get_type_hints()` even
+        # where no SDK — and therefore no client — is installed.
+        AsyncClient = object
```
Kludex, line 29: I didn't get this 🤔

Who calls get_type_hints on Provider?

dsfaccini:
> This comment was posted by Codex Desktop using gpt-5.6-sol on behalf of David.

Superseded by #7351. OpenAI 2.47+ now supports HTTPX2, and the replacement was rebuilt on current `main` with compatibility for legacy caller-owned clients, HTTPX2 retry transports, current bounded-download behavior, and wheel-metadata dependency checks.

PR #5509, opened 2026-05-18
Fix mid-conversation `SystemPromptPart`s being incorrectly hoisted into the top-level system prompt on Anthropic, Google, and Bedrock

### LINKED ISSUE #5437: Render mid-conversation `SystemPromptPart`s as XML-wrapped `UserPromptPart`s for Anthropic/Google
## Context

Mid-conversation `SystemPromptPart`s are mapped inconsistently across providers today:

| Provider | Behavior |
|---|---|
| OpenAI Chat | Sent inline as `role='system'` at the part's position — `models/openai.py:1410` |
| OpenAI Responses | Sent inline as `role='system'` at the part's position — `models/openai.py:2425` |
| Anthropic | **Hoisted** to top-level `system=` parameter — `models/anthropic.py:1104-1105` |
| Google | **Hoisted** to top-level `system_instruction=` parameter — `models/google.py:940-941` |

For Anthropic/Google this hoisting has two failure modes:

1. **Cache invalidation.** A `SystemPromptPart` enqueued mid-run is intended to update the model's context *without* breaking the prefix KV cache (the cited motivation in issue #700 and the dynamic-instructions use case). But because all `SystemPromptPart`s on Anthropic/Google get concatenated into the top-level system block, the system content changes byte-for-byte → prefix cache is invalidated.
2. **Lost positional intent.** The producer of a mid-conversation `SystemPromptPart` typically expects it to sit *between* a tool return and the next user-style content, or to follow a specific conversational point. Hoisting strips that ordering — the model sees the new content at the start of the conversation instead.

## Workaround in PR #4980

[PR #4980](https://github.com/pydantic/pydantic-ai/pull/4980) (`ctx.enqueue` / `agent_run.enqueue`) drops `SystemPromptPart` from the `EnqueueContent` type for exactly this reason: it would be a footgun. To inject system-style content mid-run today, you have to use a `ModelRequest` passthrough, which makes the cross-provider quirk visible at the call site.

## Proposal

For non-leading `ModelRequest`s on Anthropic/Google (i.e. anything except the very first request), render each `SystemPromptPart` as a `UserPromptPart` with an XML wrapper instead of hoisting it. Something like:

```
<system-reminder>
{system_prompt_part.content}
</system-reminder>
```

Anthropic's [`<system-reminder>`](https://docs.claude.com/en/docs/build-with-claude/prompt-engineering/prompt-improver#xml-tags) convention is already a documented pattern for this, and Claude Code itself injects mid-conversation reminders using this exact shape (within user turns).

Benefits:
- Cache prefix preserved (no top-level system block churn)
- Positional intent honored (the reminder lands where it was enqueued)
- Cross-provider consistency (matches OpenAI's inline behavior)
- Backwards compatible (existing `SystemPromptPart`s at the start of the conversation still hoist as before — that's the correct mapping for leading system content)

The leading `SystemPromptPart`s (those before any `ModelResponse`) should keep their current behavior — they're the actual system prompt and belong in the top-level slot for both providers.

## Open questions

- Wrapper tag name: `<system-reminder>`, `<system>`, `<instructions-update>`, something else?
- Should this be configurable via `ModelProfile`, or hardcoded per provider?
- Behavior for tools/builtins on `_clean_message_history` once this lands — the merge logic might want to know whether a part will be hoisted vs inlined.

## Related

- #700 (original context-injection discussion)
- #4980 (pending message queue PR)
- #5xxx (this issue's parent PR if filed)

navalprakhar:
- Closes #5437 

Non-leading `SystemPromptPart`s are now wrapped as `<system>`-tagged `UserPromptPart`s on Anthropic, Google, and Bedrock, gated by `ModelProfile.supports_inline_system_prompts`.

### Checklist


DouweM:
@navalprakhar Thanks for working on this, I'll take it from here :)

Issue #5399, opened 2026-05-13
Test flakes: 5 distinct CI flakes across capabilities, evals, temporal, ui_web, and dbos


adtyavrdhn:
Flaky tests found in recent CI runs. Each was verified by reading the failed job log; each failed on an unrelated branch/commit and either passed on rerun or self-healed without a code change.

> **Status (sweep over last ~120 CI runs, 2026-06-17 → 06-23):** items re-verified against current `main`. Items 2–5 remain latent; resolved items have been removed. See the sweep comment below for newly discovered flakes.

### 2. `tests/evals/test_online.py::test_correlated_sampling_subset_property` — ❌ still latent

Probabilistic: 100 calls at `sample_rate=0.1`, asserts at least one fires. P(zero) = 0.9¹⁰⁰ ≈ 2.7e-5.

- [run 25585653514, 3.11 standard](https://github.com/pydantic/pydantic-ai/actions/runs/25585653514/job/75113536970) — `assert 0 > 0`, passed on rerun

Confirmed latent: `call_seed = random.random()` in `pydantic_evals/pydantic_evals/_online.py` is unseeded; `tests/evals/conftest.py` does not seed it.

Fix: seed the RNG.

### 3. `tests/test_temporal.py::*` (entire suite) — ❌ still latent

When the Temporal dev-server binary download fails, ~30 tests fail at setup with `RuntimeError: Failed starting Temporal dev server: error sending request for url (https://...)`.

- [run 24815366788 attempt 1, main `c1d060d`, 3.10 all-extras](https://github.com/pydantic/pydantic-ai/actions/runs/24815366788/job/72628571556) — passed on attempt 2

Confirmed latent: no Temporal binary cache/retry in `.github/workflows/`; the fixture calls `WorkflowEnvironment.start_local(...)` with no backoff.

Fix: cache the binary via `actions/cache`, or retry with backoff on startup.

### 4. `tests/test_ui_web.py::test_chat_app_index_caching` — ❌ still latent

xdist race on the shared filesystem cache at `~/.cache/pydantic-ai/web-ui/{version}.html`. `_get_ui_html` checks `cache_file.exists()` then reads — but `exists()` flips to `True` the instant another worker's `write_bytes()` opens the file, before content lands. Concurrent reader gets `b''`.

[...]

PR #5533, opened 2026-05-20
Make retry-after transport test deterministic

FU-max-boop:
## Summary

- Replace the wall-clock upper-bound assertion in `test_sync_transport_with_wait_retry_after` with a fake `sleep` callback.
- Keep the retry flow, call-count check, and `Retry-After` cap assertion while avoiding scheduler/coverage/xdist timing jitter.

Refs #5399

## Validation

- `uv run pytest tests/test_tenacity.py::TestIntegration::test_sync_transport_with_wait_retry_after -q`
- `uv run pytest tests/test_tenacity.py -q`
- `uv run ruff check tests/test_tenacity.py`
- `uv run ruff format --check tests/test_tenacity.py`
- `git diff --check`

### DIFF
--- a/tests/test_tenacity.py
+++ b/tests/test_tenacity.py
@@ -651,26 +651,27 @@ class TestIntegration:
             if response.status_code == 429:
                 raise httpx.HTTPStatusError('Rate limited', request=request, response=response)
 
+        sleep_calls: list[float] = []
+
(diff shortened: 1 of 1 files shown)


adtyavrdhn (approved):
Yeah this is good. CC @DouweM

PR #5549, opened 2026-05-20
docs: recommend the Thinking capability

dsfaccini:
Recommends the unified `Thinking` capability first in the thinking guide, while keeping provider-specific sections focused on advanced native settings. Also updates the Google model docs to lead with `Thinking` and frame `google_thinking_config` as advanced usage.

No issue; docs-only follow-up to #5527.

### Checklist

### DIFF
--- a/docs/thinking.md
+++ b/docs/thinking.md
@@ -3,20 +3,12 @@
 Thinking (or reasoning) is the process by which a model works through a problem step-by-step before
 providing its final answer.
 
-This capability is typically disabled by default and depends on the specific model being used.
-See the sections below for how to enable thinking for each provider.
+The simplest way to enable thinking across supported providers is the [`Thinking`][pydantic_ai.capabilities.Thinking] capability.
+Provider-specific settings are available for advanced usage when you need direct access to a provider's native thinking controls.
 
 ## Unified thinking settings
 
-The simplest way to enable thinking across any supported provider is the `thinking` field in [`ModelSettings`][pydantic_ai.settings.ModelSettings]:
-
-```python {title="unified_thinking.py"}
-from pydantic_ai import Agent
-
-agent = Agent('anthropic:claude-opus-4-7', model_settings={'thinking': 'high'})
-```
-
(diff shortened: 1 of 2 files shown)


docs/thinking.md
```diff
@@ -3,26 +3,26 @@
...
 
-This capability is typically disabled by default and depends on the specific model being used.
-See the sections below for how to enable thinking for each provider.
+The simplest way to enable thinking across supported providers is the [`Thinking`][pydantic_ai.capabilities.Thinking] capability.
+Provider-specific settings are available for advanced usage when you need direct access to a provider's native thinking controls.
 
 ## Unified thinking settings
 
-The simplest way to enable thinking across any supported provider is the `thinking` field in [`ModelSettings`][pydantic_ai.settings.ModelSettings]:
+Use the [`Thinking`][pydantic_ai.capabilities.Thinking] capability to enable thinking:
```
DouweM: Is there a section in the capabilities doc we should be linking to?

docs/thinking.md
```diff
@@ -3,26 +3,26 @@
...
-```python {title="thinking_capability.py"}
+```python {title="unified_thinking.py"}
 from pydantic_ai import Agent
-from pydantic_ai.capabilities import Thinking
 
-agent = Agent('anthropic:claude-opus-4-7', capabilities=[Thinking(effort='high')])
+agent = Agent('anthropic:claude-opus-4-7', model_settings={'thinking': 'high'})
 ```
 
 The `thinking` setting accepts:
```
DouweM: The stuff below here focuses too much on the model setting again, and doesn't make it clear that these "thinking setting values" map to `effort` on teh Thinking cap. So this should also be rewritten to be `Thinking` cap forward, the model setting then becomes the "also" mentioned thing.

DouweM (changes requested):

docs/thinking.md
dsfaccini: David's AICA here: Reworked this section to frame the accepted values and provider translation table around  first, with  described as the underlying direct setting.

PR #5553, opened 2026-05-20
fix: resolve single-arg unwrap silently dropping wrapped payloads with extra=ignore

### LINKED ISSUE #5550: single-arg unwrap silently dropping wrapped payloads
### Initial Checks

### Description

Context
---
For a tool function with a single ``BaseModel`` parameter, pydantic-ai flattens the input schema sent to the model (so the LLM emits the model fields at the top level instead of inside a redundant wrapper). The ``_validate_single_arg`` wrap validator accepts BOTH shapes — unwrapped ``{name: ..., value: ...}`` and wrapped ``{<param>: {name: ..., value: ...}}`` — to support callers that re-validate already-validated args (e.g. Temporal's serialization round-trip, or LLMs that didn't honor the unwrap).

The validator tries the unwrapped path first and only unwraps the ``{<param>: {...}}`` envelope on ``ValidationError``:

```python
    def _validate_single_arg(value, handler, *, name):
        try:
            return {name: handler(value)}      # try unwrapped
        except ValidationError:
            if isinstance(value, dict) and list(value) == [name]:
                return {name: handler(value[name])}  # fallback: unwrap
            raise
```

Bug
---
When the input model uses pydantic's default ``extra="ignore"``, validating ``{<param>: {...}}`` against the unwrapped schema does NOT raise — the ``<param>`` key is silently dropped as an extra, and every field falls back to its default. The fallback never runs and the tool receives a defaulted model.

Reproduction
------------
Run with:

```sh
    uv run pytest /tmp/test_pydantic_ai_single_arg_bug.py -v
```
On pydantic-ai 1.96.0 the assertions fail: the tool receives ``MyInput(name='default_name', value=0)`` instead of the ``MyInput(name='actual_name', value=42)`` payload the model sent.

A possible fix is to attempt the wrapped form first when ``value`` is shaped like ``{<param>: dict}``, or to make the validator reject extras at this level (or make it configurable). The current implementation is very fragile.
"""

```Python

from pydantic import BaseModel, ConfigDict
from pydantic_ai import Agent
    ModelMessage,
    ModelResponse,
    TextPart,
    ToolCallPart,
)


class MyInput(BaseModel):
    # Default pydantic behavior: extras are silently ignored. This is the
    # configuration that triggers the bug.
    name: str = "default_name"
    value: int = 0


class StrictInput(BaseModel):
    # With extras forbidden, the unwrapped attempt against a wrapped payload
    # raises ValidationError and the fallback path correctly unwraps.
    model_config = ConfigDict(extra="forbid")

    name: str = "default_name"
    value: int = 0


def _build_agent(tool, wrapped_args: dict) -> Agent:
    call_count = 0

    async def model_fn(messages: list[ModelMessage], info: AgentInfo) -> ModelResponse:
        nonlocal call_count
        call_count += 1
        if call_count == 1:
            # Emit the WRAPPED shape. This is what an LLM that didn't honor
            # the unwrap hint emits, and what re-validation paths (e.g.
            # Temporal serialization round-trip) feed back through the
            # validator. Either way, it should round-trip through
            # _validate_single_arg without data loss.
            return ModelResponse(
                parts=[ToolCallPart(tool_name=tool.__name__, args=wrapped_args)]
            )

    return Agent(FunctionModel(model_fn), tools=[tool])


def test_single_arg_default_extras_loses_wrapped_payload():
    """Demonstrates the bug: with default extras=ignore, wrapped args are silently dropped."""
    received: list[MyInput] = []

    def my_tool(input: MyInput) -> str:
        received.append(input)
        return "ok"

    agent = _build_agent(
        my_tool, wrapped_args={"input": {"name": "actual_name", "value": 42}}
    )
    agent.run_sync("go")

    assert len(received) == 1
    got = received[0]
    # Both assertions FAIL on pydantic-ai 1.96.0 — got is MyInput() with defaults.
    assert got.name == "actual_name", (
        f"name was {got.name!r}: the wrapped payload was silently dropped"
    )
    assert got.value == 42, (
        f"value was {got.value}: the wrapped payload was silently dropped"
    )


def test_single_arg_forbid_extras_unwraps_correctly():
    """Control: with extra='forbid', the fallback path runs and unwrapping works."""
    received: list[StrictInput] = []

    def strict_tool(input: StrictInput) -> str:
        received.append(input)
        return "ok"

    agent = _build_agent(
        strict_tool, wrapped_args={"input": {"name": "actual_name", "value": 42}}
    )
    agent.run_sync("go")

    assert len(received) == 1
    got = received[0]
    assert got.name == "actual_name"
    assert got.value == 42
```

### Logfire Trace

_No response_

- Python: 3.12.13
- Pydantic AI: 1.96.0

algojogacor:
This is an independent contribution. This PR was not solicited, commissioned, or paid for by any party.

## Summary

Fixes a bug in `_validate_single_arg` where single-argument tool functions silently receive default-valued models when the LLM sends wrapped `{param_name: {...}}` payloads and the model uses pydantic's default `extra="ignore"`.

## Root Cause

The current validator attempts the unwrapped path first (`handler(value)`), relying on `ValidationError` to trigger the wrapped fallback. With `extra="ignore"` (pydantic default), the outer `{param_name}` key is silently dropped as an extra rather than raising — every field falls back to its default and the fallback path never executes.

## Fix

Move the wrapped-shape check (`{name: dict}`) before the unwrapped attempt. When the value is already wrapped, unwrap it immediately. When it is not, proceed with the direct handler call. This eliminates the race between `extra="ignore"` and the `ValidationError` guard.

```
Before:
    try:
        return {name: handler(value)}
    except ValidationError:
        if isinstance(value, dict) and list(...) == [name]:
            return {name: handler(value[name])}
        raise

After:
    if isinstance(value, dict) and list(...) == [name]:
        return {name: handler(value[name])}
    return {name: handler(value)}
```

## Testing

The reproducer in issue #5550 passes with this change. Existing test suite behavior is preserved: the validator now handles the wrapped case proactively instead of reactively, producing identical results for both `extra="ignore"` and `extra="forbid"` configurations.

Closes #5550

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/_function_schema.py
+++ b/pydantic_ai_slim/pydantic_ai/_function_schema.py
@@ -344,14 +344,13 @@ def _validate_single_arg(
     *,
     name: str,
 ) -> dict[str, Any]:
-    try:
-        return {name: handler(value)}
-    except ValidationError:
-        # Fall back to unwrapping the `{name: value}` shape produced by a previous validation pass
-        # (e.g. after Temporal serialization/deserialization of already-validated args).
-        if isinstance(value, dict) and list(cast(dict[Any, Any], value)) == [name]:
-            return {name: handler(value[name])}
-        raise
+    if isinstance(value, dict) and list(cast(dict[Any, Any], value)) == [name]:
+        # Unwrap the `{name: value}` shape — either from a previous validation pass
+        # or from an LLM that didn't honor the flattened single-arg schema.
+        # Checking shape first avoids the case where pydantic's extra="ignore" (default)
+        # silently drops the wrapper key and the outer handler never raises ValidationError.
+        return {name: handler(value[name])}
+    return {name: handler(value)}
 
 
 def _extract_return_schema_type(return_annotation: Any, function: Callable[..., Any]) -> Any:


DouweM:
@algojogacor Please add a regression test and fix CI.

PR #5558, opened 2026-05-21
Support `top_k` in `GoogleModel`, `AnthropicModel`, and `CohereModel`

### LINKED ISSUE #5348: Passing `top_k` for gemini models
### Question

I'm trying to use Gemini's `top_k` sampling parameter with Pydantic AI, but I can't find a supported way to pass it through.

As far as I can tell, `top_k` is currently not available in either `ModelSettings` or `GoogleModelSettings`. In addition, the current `GoogleModel._build_content_and_config(...)` [implementation](https://github.com/pydantic/pydantic-ai/blob/641989370b304fe206c30e5fdeb7ef0912cd3744/pydantic_ai_slim/pydantic_ai/models/google.py#L499) does not appear to forward `top_k` into `GenerateContentConfigDict`, even though the underlying Google Gen AI config supports that field. 

For now I  built workaround:

```python
class CustomGoogleModel(GoogleModel):
    """Forwards extra Gemini knobs from settings into `GenerateContentConfig`."""

    async def _build_content_and_config(self, messages, model_settings, model_request_parameters):  # type: ignore[override]
        contents, generation_config = await super()._build_content_and_config(
            messages, model_settings, model_request_parameters
        )
        if (top_k := (model_settings or {}).get("top_k")) is not None:
            generation_config["top_k"] = top_k
        return contents, generation_config
```

What I'm trying to understand is:

- Should `top_k` be a field on `ModelSettings` or `GoogleModelSettings`?
- Is there already another supported way to pass `top_k` through to Gemini that I'm missing?
- If this is intentionally unsupported, what is the recommended extension point for provider-specific generation parameters like this?

### Additional Context

pydantic_ai version: `1.86.1`
python version: `3.12`

madanlalit:
- Closes #5348

### Description
This PR adds support for forwarding the `top_k` setting to model providers that support it:
- `GoogleModel` (maps `top_k` to the Gemini generation config `top_k`)
- `AnthropicModel` (maps `top_k` to the Anthropic `top_k` parameter)
- `CohereModel` (maps `top_k` to the Cohere V2 `k` parameter)


DouweM:
@madanlalit Nice one, thanks!

PR #5560, opened 2026-05-21
fix: don't drop single `BaseModel` tool arguments sent in wrapped form when the model uses default `extra='ignore'`

### LINKED ISSUE #5550: single-arg unwrap silently dropping wrapped payloads

macayu17:
### Summary

Fixes #5550

A tool whose single argument is a model-like type (`BaseModel`, dataclass, `TypedDict`) has its schema flattened, so the model emits the fields at the top level. The validator accepts both that unwrapped shape *and* the already-wrapped `{argument: {...}}` shape, so that previously-validated args stay valid when re-validated (e.g. across a Temporal activity round-trip).

The bug: the validator tried the unwrapped shape first and only unwrapped `{argument: {...}}` on `ValidationError`. Under Pydantic's default `extra='ignore'`, validating the wrapped shape against the unwrapped schema does **not** raise — `argument` is dropped as an extra and every field falls back to its default — so the real payload was silently lost.

### Changes

- Disambiguate `{argument: ...}` by whether `argument` is a key the model actually **accepts** (field name *or* alias), read from the model's validation JSON schema so Pydantic resolves aliases (incl. `AliasChoices`) for us.
- If `argument` is not an accepted key, `{argument: ...}` can only be a wrapper envelope → unwrap it.
- If it is a real field/alias, validate as-is and fall back to unwrapping only on `ValidationError` (the round-trip of such a model).
- Regression tests: wrapped payload with defaults (the reported case), same-named-field round-trip, aliased field with a default, and `AliasChoices`.

### Known limitations

- **Permissive name collision (irreducible):** if the model's field shares the parameter's name *and* accepts both the payload and the wrapped payload (e.g. typed `Any`), `{name: {name: X}}` validates as either reading. We prefer the unwrapped one, so re-validation isn't idempotent for that rare case. This matches `main`'s behavior; eliminating it would require unwrapping at the re-validation boundary rather than in the validator.
- **`populate_by_name` + alias:** a bare `{field_name: value}` (using the canonical name rather than the alias) for such a model is treated as a wrapper and raises. No LLM or Temporal round-trip emits this shape — they use the alias — so it only affects hand-written `validate_python` calls.

### Checklist


macayu17:
No worries, I missed that #5553 was already open before creating mine. My branch had a regression test for the default-value case and also covered the edge case where the model itself has a top-level field with the same name as the tool parameter. Happy to leave it here if that helps with the existing PR.

DouweM:
@macayu17 Sorry about the auto-close, I'll check this out in a bit.

pydantic_ai_slim/pydantic_ai/_function_schema.py
```diff
@@ -338,20 +349,43 @@ def _build_schema(
...
 
 
+def _annotation_accepts_key(annotation: Any, name: str) -> bool:
+    """Whether a model-like `annotation` accepts `name` as a top-level validation key.
+
+    Used by `_build_schema` to tell a wrapper envelope `{name: ...}` (which we unwrap) apart from
+    genuine unwrapped input for a model with a field — or field alias — called `name` (which we don't).
+    Reading the keys from the validation JSON schema lets Pydantic resolve aliases for us.
+    """
+    properties = TypeAdapter(annotation).json_schema(mode='validation').get('properties', {})
```
DouweM: This is reasonably expensive. Don't we already generate the type adapter somewhere? Could/should we reuse that?

PR #5563, opened 2026-05-21
fix: avoid wall-clock retry test flake

he-yufeng:
﻿## Summary

This removes a wall-clock upper bound from `test_sync_transport_with_wait_retry_after`.

The behavior under test is that `Retry-After: 30` is capped to `0.1` by `wait_retry_after(max_wait=0.1)`. The old assertion measured real elapsed time and required it to stay below `0.2s`, which can fail under coverage, xdist, or a busy CI runner even when tenacity asked for the correct sleep duration.

The test now passes a fake `sleep` callable through `RetryConfig`, records the requested sleep duration, and asserts that tenacity asked to sleep for `0.1` seconds.

Refs #5399.

## To verify

- `python -m pytest tests/test_tenacity.py::TestIntegration::test_sync_transport_with_wait_retry_after -q`
- `python -m pytest tests/test_tenacity.py::TestWaitRetryAfter tests/test_tenacity.py::TestIntegration::test_sync_transport_with_wait_retry_after -q`
- `python -m pytest tests/test_tenacity.py -q`
- `python -m ruff check tests/test_tenacity.py`
- `python -m ruff format --check tests/test_tenacity.py`
- `python -m py_compile tests/test_tenacity.py`

### DIFF
--- a/tests/test_tenacity.py
+++ b/tests/test_tenacity.py
@@ -651,26 +651,22 @@ class TestIntegration:
             if response.status_code == 429:
                 raise httpx.HTTPStatusError('Rate limited', request=request, response=response)
 
+        sleep_calls: list[int | float] = []
         config = RetryConfig(
             retry=retry_if_exception_type(httpx.HTTPStatusError),
             wait=wait_retry_after(max_wait=0.1),  # Cap at 0.1 seconds for tests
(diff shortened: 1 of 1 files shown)


adtyavrdhn:
Closing because there is a PR which does the same thing

PR #5566, opened 2026-05-21
Add fail-fast sequential tool barriers

### LINKED ISSUE #5551: Fail-fast option for sequential tool execution (abort downstream tools on upstream failure)
Spun out of #3791 / #5339, where `sequential=True` becomes a per-tool *barrier* and tools execute in emission order. That gives ordering but not *failure propagation*: even with `sequential=True` on every tool, a failure in one doesn't stop the tools emitted after it, and (via the new "retry-wins" invariant) a later output tool's result is only suppressed *after* it executes — its side effects still fire.

**Idea:** a per-tool opt-in (e.g. `sequential='fail_fast'`, or a separate flag) meaning "if any tool emitted before me failed — `ModelRetry`, validation, or exception — don't run me, and don't run anything after me." Turns a run of co-emitted calls into an abort-on-error pipeline.

**Use case:** the model emits a series of actions followed by an output/"done" tool. If an early action fails you want the later actions — and the output tool's side effects — skipped, and the failure surfaced to the model.

**Open questions:**
- It's a parallel-execution concern (adjacent to `sequential=True`), not an `end_strategy`, and applies to all tools, not just output tools.
- It ties behavior to the model's emission order, which isn't always reliable — is that acceptable, given the model can be instructed to emit actions in order?
- API shape: extend `sequential` (`True | 'fail_fast'`) vs a separate `stop_on_error` / dependency flag.
- Interaction with `'exhaustive'`'s all-parallel default and existing barriers.

Not in scope for #5339 — capturing for discussion.

shusingh:
## Summary
- add `sequential='fail_fast'` as a barrier mode for function and output tools
- skip the fail-fast barrier and downstream tools when an earlier tool in the same response produced a retry
- document the mode and cover graceful/exhaustive execution with regression tests

## Motivation
This gives users an opt-in abort-on-error pipeline for co-emitted tool calls, so completion/write-style tools do not run after an upstream action failed.

Fixes #5551

## Validation
- `py -m uv run pytest tests/test_agent.py::TestMultipleToolCalls -q`
- `py -m uv run ruff check pydantic_ai_slim/pydantic_ai/_tool_execution.py pydantic_ai_slim/pydantic_ai/tools.py pydantic_ai_slim/pydantic_ai/output.py pydantic_ai_slim/pydantic_ai/tool_manager.py pydantic_ai_slim/pydantic_ai/agent/__init__.py pydantic_ai_slim/pydantic_ai/toolsets/function.py tests/test_agent.py`
- `py -m uv run ruff format --check pydantic_ai_slim/pydantic_ai/_tool_execution.py pydantic_ai_slim/pydantic_ai/tools.py pydantic_ai_slim/pydantic_ai/output.py pydantic_ai_slim/pydantic_ai/tool_manager.py pydantic_ai_slim/pydantic_ai/agent/__init__.py pydantic_ai_slim/pydantic_ai/toolsets/function.py tests/test_agent.py`
- `py -m uv run pyright pydantic_ai_slim/pydantic_ai/_tool_execution.py pydantic_ai_slim/pydantic_ai/tools.py pydantic_ai_slim/pydantic_ai/output.py pydantic_ai_slim/pydantic_ai/tool_manager.py pydantic_ai_slim/pydantic_ai/agent/__init__.py pydantic_ai_slim/pydantic_ai/toolsets/function.py`


shusingh:
Pushed a follow-up for the coverage failure. Added regression coverage for repeated fail-fast skips and deferred-call skips after an earlier retry, and marked the empty-yield paths that cannot produce events as uncovered. Locally verified the focused fail-fast tests and Ruff on the touched files.

shusingh:
The remaining failure appears unrelated to this PR: 	ests/test_ui_web.py::test_chat_app_index_caching failed only on the Python 3.13 pydantic-evals shard, while the touched fail-fast tests and the rest of the matrix passed. Could this job be rerun?

shusingh:
Closing this since `v2-main` has moved significantly and this branch is now stale/conflicting. Thanks for taking a look.

Issue #5473, opened 2026-05-15
Safe mode for `AgentSpec` / `Agent.from_file` to bound capability-config attack surface


DouweM:
## Description

`AgentSpec` instances loaded from user-provided YAML/JSON via `Agent.from_file()` / `Agent.from_spec()` can carry capability config that, when activated, executes code or makes network requests against attacker-controlled targets. Discovered in review of PR #5325 (`MCP` capability widening) — quoting DouweM's note:

> `AgentSpec`s are potentially loaded from user-provided yaml files, so passing a script path could be abused for RCE, and a local URL for SSRF. So we will at some point need to have the concept of safe and unsafe capabilities/configs, or a safe mode to run spec parsing in.

### Concrete attack surface today

- **MCP capability** (`pydantic_ai.capabilities.MCP`) — `local='/path/to/script.py'` / `local=MCPToolset('script.py')` spawns a subprocess via FastMCP's `PythonStdioTransport` → arbitrary code execution as the agent host. `local='http://internal-service'` allows SSRF against internal endpoints. (PR #5325 already validates `local=` strings as `http(s)://` URLs in `MCP.from_spec`, which mitigates the RCE-via-script-path vector specifically, but the URL → SSRF vector remains and the underlying pattern will repeat for other capabilities.)
- **`WebFetch` / `WebSearch` capabilities** — local-mode tools that hit arbitrary HTTP endpoints from the agent host (SSRF if local-network paths aren't blocked).
- **Plugins** — see #5266: dotted Python module paths in YAML get imported. This one is opt-in by design (the user adding `plugins:` is explicitly authorizing code load), so it sits at a different point on the trust spectrum, but a safe-mode design needs to take a position on it.
- **Future capabilities** — anything that takes a URL, file path, command, or import string from the spec will have this shape.

### Proposed direction (sketch, open to alternatives)

A `safe_mode` flag on `Agent.from_file` / `Agent.from_spec` (and the corresponding `AgentSpec` validation), default `True` for `from_file` (file → likely user-provided) and `False` for `from_spec` (programmatic → caller trusts the dict). When `safe_mode=True`:

- Capabilities expose an `allowed_in_safe_mode` flag (default `True` for HTTP-only fully-validated configs, `False` for anything that takes file paths, commands, or arbitrary URLs).
- `MCP(local='http(s)://...')` ✅ allowed (already URL-validated)
[...]

PR #5571, opened 2026-05-21
fix(ui): Strip `force_download` flag from client-submitted `FileUrl` parts in `UIAdapter.sanitize_messages`

### LINKED ISSUE #5535: `sanitize_messages` should strip `FileUrl` `force_download='allow-local'` on client-submitted parts
### Initial Checks

### Description

# `sanitize_messages` should strip `force_download='allow-local'` on client-submitted `FileUrl` parts

## Problem

`FileUrl.force_download='allow-local'` (declared at `messages.py:198–217`) tells the SSRF-protected downloader to allow private IPs through. That is the right default for server-authored URLs but lets a frontend probe internal services if the value is accepted from a client.

`UIAdapter.sanitize_messages` already drops `FileUrl` parts with disallowed URL schemes, so the precedent for stripping unsafe `FileUrl` attributes from client-submitted messages is there. `force_download='allow-local'` is the missing one — it survives `sanitize_messages` today.

This is defense-in-depth for the existing SSRF guard at `_ssrf.py:safe_download`, not a vulnerability — but it concretizes the recurring SSRF pattern flagged by GHSA-2jrp-274c-jhv3 (CVE-2026-25580).

## Repro

1. Frontend sends a `ModelRequest` containing `ImageUrl(url='https://...', force_download='allow-local')`.
2. `AGUIAdapter.sanitize_messages` returns the part unchanged.
3. Downstream `download_item` honors `force_download='allow-local'` and passes `allow_local=True` to `safe_download`, bypassing the private-IP block.

## Proposed fix

In `_filter_user_content`, when keeping a `FileUrl`, reset `force_download='allow-local'` to `False` and emit a `UserWarning` listing the affected URLs, mirroring the existing scheme-strip pattern.

Server-side `message_history` (passed to `Agent.run` directly, not through `sanitize_messages`) is unaffected.

I have a working patch + 3 tests (positive, negative, scheme-vs-allow-local ordering edge case). Happy to open a PR if a maintainer agrees on the approach and assigns this issue.

```Python
from ag_ui.core import RunAgentInput, UserMessage
from pydantic_ai import Agent
from pydantic_ai.messages import ImageUrl, ModelRequest, UserPromptPart
from pydantic_ai.ui.ag_ui import AGUIAdapter

agent = Agent(model=TestModel())
run_input = RunAgentInput(
    thread_id='t1', run_id='r1',
    state={}, messages=[UserMessage(id='m1', content='hi')],
    tools=[], context=[], forwarded_props={},
)
adapter = AGUIAdapter(agent=agent, run_input=run_input)

crafted = [ModelRequest(parts=[UserPromptPart(content=[
    ImageUrl(url='https://example.com/img.png', force_download='allow-local'),
])])]

sanitized = adapter.sanitize_messages(crafted)
img = sanitized[0].parts[0].content[0]
print(f'force_download after sanitize: {img.force_download!r}')
# v1.99.0 output:        force_download after sanitize: 'allow-local'
# Expected with fix:     force_download after sanitize: False
#                        (+ UserWarning naming the URL whose value was reset)
```

### Logfire Trace

N/a

Python: 3.10–3.13 (any supported version)
Pydantic AI: v1.99.0 (also affects every version since 1.34 when force_download was introduced)
LLM client: N/A — bug doesn't reach a model provider.

colesmcintosh:
- Closes #5535

## Summary

`FileUrl.force_download='allow-local'` opts a URL out of the SSRF private-IP block in `safe_download`. That is safe for server-authored URLs, but should not be honored blindly from client-submitted UI history.

This PR makes `UIAdapter.sanitize_messages` treat `FileUrl.force_download` like other client-controlled file URL trust settings:

- Adds `UIAdapter.allowed_force_download: frozenset[ForceDownloadMode]`, defaulting to `frozenset({False})`, and plumbs it through `from_request` and `dispatch_request`.
- Resets non-allowlisted `force_download` values to `False` on client-submitted `FileUrl`s.
- Applies the check to `FileUrl`s in user content and in plain tool return parts, including nested dict/list content.
- Keeps typed tool-return subclasses out of the recursive tool-return walker by gating on `part.tool_kind is None`.
- Keeps warnings free of raw client-submitted URLs; warnings report only schemes or `force_download` values.

Server-side `message_history` passed directly to `Agent.run` is unaffected (`sanitize_messages` is only called on protocol-derived input).

## Behavior

| Input | Before | After, default config |
|---|---|---|
| `FileUrl(force_download=False)` | unchanged | unchanged |
| `FileUrl(force_download=True)` | unchanged | reset to `False` + warning |
| `FileUrl(force_download='allow-local')` | unchanged | reset to `False` + warning |
| `FileUrl(scheme='s3://')` in user content | dropped (scheme) | dropped (scheme) |
| `FileUrl(scheme='s3://')` in tool return content | unchanged | dropped (scheme) |
| `allowed_force_download=frozenset({False, True})` | n/a | keeps `True`, resets `'allow-local'` |
| `allowed_force_download=frozenset({False, True, 'allow-local'})` | n/a | keeps all three values |

## Tests

- `test_sanitize_messages_resets_force_download_not_in_allowlist`
- `test_sanitize_messages_leaves_allowlisted_force_download_alone`
- `test_sanitize_messages_widened_allowlist_lets_true_through_but_resets_allow_local`
- `test_sanitize_messages_strips_disallowed_scheme_before_reset`
- `test_sanitize_messages_resets_force_download_in_tool_return_parts`
- `test_sanitize_messages_strips_disallowed_schemes_in_tool_return_parts`

### Checklist


pydantic_ai_slim/pydantic_ai/ui/_adapter.py
```diff
@@ -349,6 +361,17 @@ def sanitize_messages(
                 stacklevel=2,
             )
 
+        if reset_allow_local_count:
+            file_url_label = 'file URL' if reset_allow_local_count == 1 else 'file URLs'
+            warnings.warn(
+                f'{reset_allow_local_count} client-submitted {file_url_label} had '
+                f"`force_download='allow-local'` reset to `False`. `'allow-local'` opts the URL "
```
DouweM: Do you think we should disallow `True` as well?

pydantic_ai_slim/pydantic_ai/ui/_adapter.py
```diff
@@ -284,6 +284,10 @@ def sanitize_messages(
           Non-HTTP schemes like `s3://` or `gs://` cause the model provider to fetch
           the object using the server-side IAM role, so they should only be accepted
           from trusted frontends.
+        - [`FileUrl.force_download`][pydantic_ai.messages.FileUrl.force_download]
```
DouweM, line 318: I think this should be configurable, like the other sanitization features are (right?).

DouweM (changes requested):

dsfaccini:
David's AICA here: maintainer taking this over. Pushed a revision (8c3114af) that adopts the allowlist design from the review thread, which answers both of @DouweM's questions in one shape.

**Allowlist instead of an `'allow-local'`-only reset**

Replaced the `force_download == 'allow-local'` check with a configurable `allowed_force_download` field on `UIAdapter` — `frozenset[ForceDownloadMode]`, default `frozenset({False})` — plumbed through `from_request` and `dispatch_request` exactly like `allowed_file_url_schemes`.

- By default both `True` and `'allow-local'` are reset to `False` on client-submitted `FileUrl` parts. `'allow-local'` bypasses the SSRF private-IP block; `True` makes the server fetch the file itself. Neither is safe to honor from untrusted client input by default — so this answers "should we disallow `True` as well?": yes, by default.
- Users who trust their frontend can widen the set, e.g. `allowed_force_download=frozenset({False, True})` or `frozenset({False, True, 'allow-local'})` — so this answers the configurability question, and it matches the existing `allowed_file_url_schemes` allowlist (restrictive default + opt-in escape hatch) rather than introducing a parallel mechanism.

**Reaching `FileUrl`s in tool return parts**

The sanitizer previously only inspected user-content parts. PR #5255 introduces multimodal tool returns, which put `FileUrl` objects inside `ToolReturnPart` / native tool return part `content` — including nested in dicts and lists — that `sanitize_messages` currently doesn't reach. The revision walks tool-return content recursively and applies the same `allowed_force_download` check there, so the allowlist holds regardless of where a `FileUrl` appears in client-submitted history.

**Tests** — `tests/test_ui.py`: `True` and `'allow-local'` both reset by default, a widened allowlist lets values through, and `FileUrl`s nested in a `ToolReturnPart` (incl. inside a dict) and in a native tool return part get sanitized. Signature test covers `allowed_force_download` on `from_request` / `dispatch_request`.

pydantic_ai_slim/pydantic_ai/ui/_adapter.py
dsfaccini: David's AICA here: addressed in 8c3114af. `sanitize_messages` no longer hard-codes the reset — there is now a configurable `allowed_force_download` field on `UIAdapter` (`frozenset[ForceDownloadMode]`, default `frozenset({False})`), plumbed through `from_request` and `dispatch_request` the same way `allowed_file_url_schemes` is. Default behaviour strips disallowed values; a trusting frontend can widen the set.

pydantic_ai_slim/pydantic_ai/ui/_adapter.py
dsfaccini: David's AICA here: yes — addressed in 8c3114af. With the new `allowed_force_download` allowlist defaulting to `frozenset({False})`, `True` is reset to `False` by default alongside `'allow-local'`. `True` makes the server fetch the file on the client's behalf — a weaker concern than `'allow-local'`'s SSRF bypass, but still not something to accept from untrusted input by default. A frontend you trust can opt back in with `allowed_force_download=frozenset({False, True})`.

pydantic_ai_slim/pydantic_ai/ui/_adapter.py
```diff
@@ -284,6 +284,10 @@ def sanitize_messages(
...
+          values of `'allow-local'` on kept parts. `'allow-local'` opts the URL out
+          of the SSRF private-IP block, which is only safe for server-authored URLs;
+          on client-submitted parts it's reset to `False`.
```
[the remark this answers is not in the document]
dsfaccini: David's AICA here: implemented this design in 8c3114af. Added `allowed_force_download: frozenset[ForceDownloadMode]` defaulting to `frozenset({False})`, plumbed through `from_request` / `dispatch_request` like `allowed_file_url_schemes`; `_filter_user_content` now does a `not in self.allowed_force_download` check. One addition beyond the proposal: the sanitizer also walks `FileUrl`s nested in tool return part content recursively — PR #5255 introduces multimodal tool returns that put `FileUrl` objects inside `ToolReturnPart` / native tool return part `content` (incl. nested in dicts/lists), which `sanitize_messages` didn't previously reach, so the allowlist needs to cover that surface too.

dsfaccini:
David's AICA here: follow-up in 19ecbff7 — the tool-return content walker is now scoped to `ToolReturnPart` / `NativeToolReturnPart` and explicitly excludes `ToolSearchReturnPart` / `NativeToolSearchReturnPart`. Their `content` is a `ToolSearchReturnContent` TypedDict (not the recursive `ToolReturnContent` the walker is built for), so search-result parts no longer get needlessly rebuilt as plain dicts — they pass through unchanged.

pydantic_ai_slim/pydantic_ai/ui/_adapter.py
```diff
@@ -382,7 +441,22 @@ def _sanitize_request_parts(
                 stripped_system_prompt = True
                 continue
             if isinstance(part, UserPromptPart) and not isinstance(part.content, str):
-                new_parts.append(replace(part, content=self._filter_user_content(part.content, disallowed_schemes)))
+                filtered_content = self._filter_user_content(
+                    part.content, disallowed_schemes, reset_force_download_values
+                )
+                new_parts.append(replace(part, content=filtered_content))
+            elif isinstance(part, (ToolReturnPart, NativeToolReturnPart)) and not isinstance(
+                part, (ToolSearchReturnPart, NativeToolSearchReturnPart)
```
DouweM: Please don't exclude this specific type -- there will be more part subclasses with narrower types for e.g. `load_capability` and other well-known tools! The thing to check is whether `tool_kind` is set on the part.

pydantic_ai_slim/pydantic_ai/ui/_adapter.py
```diff
@@ -403,27 +483,88 @@ def _filter_user_content(
...
+        # but resolves to `Any` at type-check time, so pyright can't infer the element types.
+        if isinstance(content, Mapping):
+            mapping: Mapping[str, ToolReturnContent] = content  # pyright: ignore[reportUnknownVariableType]
+            return {
+                key: self._sanitize_tool_return_content(value, reset_force_download_values)
+                for key, value in mapping.items()
+            }
+        if isinstance(content, Sequence) and not isinstance(content, (str, bytes)):
+            sequence: Sequence[ToolReturnContent] = content  # pyright: ignore[reportUnknownVariableType]
+            return [self._sanitize_tool_return_content(item, reset_force_download_values) for item in sequence]
```
DouweM: I'm getting the strong feeling the real solution is to not walk these hierarchies, but to have a Pydantic validator on `FileUrl` that uses Pydantic's "validation context" feature (or a contextvar) to pass trusted/untrusted information.

Similar to #5473. Something to experiment with while I'm out? :)

pydantic_ai_slim/pydantic_ai/ui/_adapter.py
```diff
@@ -403,27 +483,88 @@ def _filter_user_content(
...
+                and part.tool_call_id not in resolved_tool_call_ids
+            ):
                 dangling_names.append(part.tool_name)
                 continue
-            new_parts.append(part)
+            if isinstance(part, (ToolReturnPart, NativeToolReturnPart)) and not isinstance(
+                part, (ToolSearchReturnPart, NativeToolSearchReturnPart)
+            ):
+                # See `_sanitize_request_parts`: only the non-search tool-return parts carry the
+                # recursive `ToolReturnContent` the walker is built for.
```
DouweM: It is very weird to me that this requires excluding `ToolSearchReturnPart`. Why would a walker made for `ToolReturnPart` not work on _its subclass_ `ToolSearchReturnPart`? `ToolReturnPart` has `content` as `ToolReturnContent`, including `Any`, while `ToolSearchReturnPart` has it as a specific `TypedDict`, which is valid for `Any`. So why would the walker fail on the tool search dict it sees?

pydantic_ai_slim/pydantic_ai/ui/_adapter.py
dsfaccini: David's AICA here: good call — gated on `part.tool_kind is None` instead of an isinstance allowlist in ba102ecb7ceaf6d49efd60776197a8dfd754f71d. Plain `ToolReturnPart`/`NativeToolReturnPart` have `tool_kind=None` and carry the recursive `ToolReturnContent`; any future narrower subclass (`load_capability`, etc.) that sets a non-None `tool_kind` is skipped automatically without touching the sanitizer. Dropped the `_tool_search` and `ToolReturnPart`/`NativeToolReturnPart` imports and restored `BaseToolReturnPart`.

dsfaccini:
David's AICA here: ran a full local review wave on `b3f50408` (5 reviewers — `review-ui-security`, `review-runtime-behavior`, `review-code-reuse`, `review-patterns`, `auto-review`). **0 blocking, 6 non-blocking findings.** Logging them here so they're not lost; happy to come back and address any on @DouweM's call. Cole has stepped back from the PR.

Trust model is intact — the security reviewer confirmed the default-deny `allowed_force_download` + recursive walker close every path a client-controlled `FileUrl` can take into the agent. The non-blocking items:

1. **Wrapper-signature symmetry** (`review-code-reuse`): `pydantic_ai/ui/vercel_ai/_adapter.py:{from_request,dispatch_request}` enumerates `allowed_file_url_schemes` explicitly but not the new `allowed_force_download`. Same for `pydantic_ai/ag_ui.py:{handle_ag_ui_request,run_ag_ui}` (though those are deprecated for 2.0). Runtime works via `**kwargs`; statically, callers of those wrappers can't configure the new opt-in surface. The security reviewer cleared this from the trust lens (secure default still applies regardless) — purely ergonomics. Cheap to fix by mirroring the `allowed_file_url_schemes` plumbing.

2. **`(keep, content)` tuple + `content=None` semantic** (`auto-review`): when a tool-return part's *sole* top-level `FileUrl` has a disallowed scheme, `_sanitize_tool_return_content` returns `(False, content)` and the caller writes `content=None`. The bool flag is only ever set to `False` at leaves; container recursion always returns `(True, ...)`. A sentinel return or dropping the whole part (mirroring how requests with no parts get dropped in `sanitize_messages`) would read more clearly. This may also become moot if we move to the `FileUrl` validator-with-context approach in https://github.com/pydantic/pydantic-ai/pull/5571#discussion_r3291616364 — flagging now so it's preserved.

3. **Unbounded recursion in `_sanitize_tool_return_content`** (`review-runtime-behavior`): a ~250-deep client-submitted nested list/dict triggers `RecursionError` (HTTP 500). Mild DoS surface. Pydantic recurses through the same structure during validation, so the depth floor is roughly the same. Easily addressed with an explicit-stack iteration if it matters.

4. **One test pattern nit** (`review-patterns`): `tests/test_ui.py::test_sanitize_messages_strips_disallowed_scheme_before_reset` uses `warnings.catch_warnings(record=True) + manual filter` rather than stacked `pytest.warns(match=...)`. Matches an older pattern elsewhere in the file, but the other new tests in this PR use `pytest.warns` correctly.

dsfaccini:
David's AICA here: addressed the two final blockers from sync today so this can land Monday.

**1. Renamed `allowed_force_download` → `allowed_file_url_force_download`** for symmetry with the sibling `allowed_file_url_schemes` field — both gate `FileUrl` fields, so they share the `allowed_file_url_*` prefix. Renamed everywhere (`UIAdapter` field + docstring, `from_request` / `dispatch_request` kwargs on the base + `VercelAIAdapter` + `AGUIAdapter` overrides, warning text, tests).

**2. Default is now `frozenset()` and `False` is always treated as safe.** The previous default `frozenset({False})` required users to keep `False` in any custom allowlist, otherwise the natural `frozenset({True, 'allow-local'})` spelling would log a confusing 'False was reset to False' warning. The check is now:

```python
if part.force_download is not False and part.force_download not in self.allowed_file_url_force_download:
    ... reset ...
```

The allowlist now means 'additional values beyond `False` to permit'. The warning only fires for actually-reset values.

Targeted pyright clean; `tests/test_ui.py`, `tests/test_vercel_ai.py`, `tests/test_ag_ui.py` green.

PR #5588, opened 2026-05-22
feat: add `DeferredToolCallEvent` and `DeferredToolResultEvent` to `AgentStreamEvent`

### LINKED ISSUE #5577: Missing `DeferredToolCallEvent` / `DeferredToolResultEvent` in `AgentStreamEvent`
### Description

When a tool is deferred (via `ApprovalRequired`, `CallDeferred`, or `kind='external'`), the agent graph emits a `FunctionToolCallEvent` like for any regular tool — but there is no way for an event stream consumer to know that this particular call is deferred and requires interaction. The `FunctionToolCallEvent` carries only a `ToolCallPart` (tool name, args, call ID); it does not carry the `DeferredToolRequests` payload, so the frontend cannot know what kind of interaction is expected without maintaining its own out-of-band mapping of tool names.

#### Problem

The `HandleDeferredToolCalls` capability resolves deferred tool calls (user questions, approval requests, externally-executed tools) but does **not** emit any event into the `AgentStreamEvent` stream.

Today, `AgentStreamEvent` is defined as:

```python
AgentStreamEvent = ModelResponseStreamEvent | HandleResponseEvent
```

`HandleResponseEvent` covers `FunctionToolCallEvent`, `FunctionToolResultEvent`, `OutputToolCallEvent`, etc. — but there is **no event type for deferred tool interactions** (neither the request going out nor the result coming back).

##### Consequence

The event stream handler (`ProcessEventStream`) already sees a `FunctionToolCallEvent` for every tool call — including deferred ones. The frontend *knows* a tool was invoked. But that's not enough:

- The `FunctionToolCallEvent` carries only a `ToolCallPart` (tool name, args, call ID). It does not tell the consumer that **this particular call requires interaction** — the frontend has no way to distinguish a tool that will resolve instantly from one that is paused waiting for human input.
- More critically, the event does not carry the `DeferredToolRequests` payload. The frontend cannot know *what kind* of interaction is expected (approval? user question? external execution?), which tools are in the batch, or what metadata the deferred handler attached.

This forces the frontend to maintain its own mapping of "which tool names are interactive" — a fragile, implicit coupling between the frontend and the tool registration.

What's needed is an event that explicitly marks the call as deferred **and** carries the full `DeferredToolRequests`, so consumers can react appropriately without out-of-band knowledge.

##### Durable execution makes this critical

This gap is particularly problematic with durable execution frameworks (Temporal, DBOS, Prefect). In a typical Temporal-based setup, the pattern for human-in-the-loop is:

1. The `HandleDeferredToolCalls` handler receives a deferred tool call (e.g. `ask_user`)
2. It needs to **signal the frontend** that the run is paused and waiting for user input — including the `DeferredToolRequests` payload so the frontend can render the right interaction
3. It calls `workflow.wait_condition(...)` to pause the workflow until a signal arrives
4. When the user responds (via a Temporal signal), it returns the result

The problem: the stream already emits a `FunctionToolCallEvent` for the deferred tool, but that event carries no `DeferredToolRequests` context. The deferred handler is forced to emit its own custom event (directly via SSE/WebSocket) to send the requests payload to the frontend, because there is no standard event type that carries it.

This means:

- The **deferred handler** (`HandleDeferredToolCalls`) becomes responsible for both resolving the tool *and* building/emitting the interaction payload — two separate concerns
- The **event stream handler** (`ProcessEventStream`) cannot distinguish deferred calls from regular ones, so it must maintain exclusion lists (`_SILENT_TOOLS`) to avoid emitting redundant notifications for tools the deferred handler already handled directly
- Both capabilities become **implicitly coupled**: the deferred handler must know about the transport layer (SSE, WebSocket), and the event stream handler must know which tools are "handled elsewhere"

I'd like to propose adding two new event types to `HandleResponseEvent`:

- **`DeferredToolCallEvent`** — emitted for each deferred tool call, carrying the `DeferredToolRequests` so consumers know what interaction is needed
- **`DeferredToolResultEvent`** — emitted when a `HandleDeferredToolCalls` capability resolves deferred calls, carrying the `DeferredToolResults`

#### Demo

With these events, a `ProcessEventStream` handler can cleanly react to deferred interactions without any coupling to the deferred handler:

```python
from pydantic_ai import Agent
from pydantic_ai.capabilities import ProcessEventStream, HandleDeferredToolCalls
from pydantic_ai.messages import DeferredToolCallEvent, DeferredToolResultEvent

async def event_handler(ctx, stream):
    async for event in stream:
        match event:
            case DeferredToolCallEvent(part=part, requests=requests):
                # Send "waiting for input" notification to the frontend
                # requests carries the full context: calls, approvals, metadata
                await notify_frontend(part.tool_name, requests)
            case DeferredToolResultEvent(results=results):
                # Notify the frontend that the interaction was resolved
                await notify_frontend_resolved(results)

async def resolve_deferred(ctx, requests):
    # Pure resolution logic — no need to emit events here
    return requests.build_results(approve_all=True)

agent = Agent(
    'openai:gpt-4o',
    capabilities=[
        ProcessEventStream(handler=event_handler),
        HandleDeferredToolCalls(handler=resolve_deferred),
    ],
)
```

Today, the same pattern requires the deferred handler to directly call `emit()` (SSE, WebSocket, etc.) to notify the frontend, because the event stream has no deferred-specific event to match on. This couples resolution and presentation.

#### Use case: durable execution with Temporal

In a Temporal workflow, the `HandleDeferredToolCalls` handler pauses via `workflow.wait_condition()` until a signal arrives from the frontend. Currently, it must also emit the interaction request to the frontend itself (via a direct `emit()` call), because `ProcessEventStream` has no visibility into deferred calls.

With `DeferredToolCallEvent`, the event stream handler handles all frontend communication, and the deferred handler becomes pure resolution logic:

```python
# Before (coupled)
class TemporalDeferredResolver:
    async def resolve(self, ctx, requests):
        await emit(AgentQuestionEvent(requests=requests))  # forced to emit directly
        await workflow.wait_condition(...)
        return answer

# After (decoupled)
class TemporalDeferredResolver:
    async def resolve(self, ctx, requests):
        # DeferredToolCallEvent is emitted by the agent graph
        # ProcessEventStream picks it up and forwards to the frontend
        await workflow.wait_condition(...)
        return answer
```

#### Proposed types

```python
@dataclass
class DeferredToolCallEvent(ToolCallEvent):
    """Emitted when a tool call is deferred for external resolution or approval."""
    requests: DeferredToolRequests
    event_kind: Literal['deferred_tool_call'] = 'deferred_tool_call'

@dataclass
class DeferredToolResultEvent:
    """Emitted when deferred tool calls have been resolved by a handler."""
    results: DeferredToolResults
    event_kind: Literal['deferred_tool_result'] = 'deferred_tool_result'
```

`DeferredToolCallEvent` inherits from `ToolCallEvent` (like `FunctionToolCallEvent`). `DeferredToolResultEvent` does not inherit from `ToolResultEvent` because the payload is a batch `DeferredToolResults`, not a single `ToolReturnPart` — the resolved calls still emit their own `FunctionToolResultEvent` when re-executed.

This is a non-breaking addition per the [version policy](https://ai.pydantic.dev/version-policy/) (adding new stream events is allowed in minor releases).

### References

- `HandleDeferredToolCalls` capability: `pydantic_ai/capabilities/deferred_tool_handler.py`
- `ProcessEventStream` capability: `pydantic_ai/capabilities/process_event_stream.py`
- `HandleResponseEvent` union: `pydantic_ai/messages.py` ~L2935
- Deferred tool processing in agent graph: `pydantic_ai/_agent_graph.py` `process_tool_calls()` ~L1700-1820
- `DeferredToolRequests` / `DeferredToolResults`: `pydantic_ai/tools.py`
- Version policy (non-breaking additions): https://ai.pydantic.dev/version-policy/
- Durable execution docs: https://ai.pydantic.dev/durable_execution/

YHallouard:
Add `DeferredToolCallEvent` and `DeferredToolResultEvent` to `AgentStreamEvent` so that stream consumers can observe deferred tool interactions without coupling to the `HandleDeferredToolCalls` handler.

- Closes #5577

### What changed

**`pydantic_ai/messages.py`**
- Added `DeferredToolCallEvent(ToolCallEvent)` — emitted for each deferred tool call (external or approval-required), carrying the full `DeferredToolRequests` batch so consumers know what interaction is expected
- Added `DeferredToolResultEvent` — emitted once per batch when a `HandleDeferredToolCalls` handler resolves deferred calls, carrying the `DeferredToolResults`
- Both added to `HandleResponseEvent` discriminated union

**`pydantic_ai/_agent_graph.py`**
- `process_tool_calls()` now yields one `DeferredToolCallEvent` per deferred call (after `DeferredToolRequests` is built, before `resolve_deferred_tool_calls()`)
- yields `DeferredToolResultEvent` after the handler returns results, before re-executing resolved tools via `_call_tools()`

**`pydantic_ai/__init__.py`**
- Re-exports both new types

### Motivation

Before this change, a `ProcessEventStream` handler had no way to distinguish a deferred tool call from a regular one — `FunctionToolCallEvent` is emitted for both, but carries only `ToolCallPart` (name, args, call ID), not the `DeferredToolRequests` payload. This forced the `HandleDeferredToolCalls` handler to directly emit SSE/WebSocket events to the frontend, coupling resolution logic to the transport layer.

With these new events, responsibilities are cleanly separated:

```python
async def event_handler(ctx, stream):
    async for event in stream:
        match event:
            case DeferredToolCallEvent(part=part, requests=requests):
                await notify_frontend(part.tool_name, requests)  # transport layer
            case DeferredToolResultEvent(results=results):
                await notify_frontend_resolved(results)

async def resolve_deferred(ctx, requests):
    await workflow.wait_condition(...)  # pure resolution logic
    return answer
```

### Backward compatibility

Adding new event types to `AgentStreamEvent` is explicitly a non-breaking change per the [version policy](https://ai.pydantic.dev/version-policy/) (adding new stream events is allowed in minor releases). Existing `match` statements without a `DeferredToolCallEvent`/`DeferredToolResultEvent` arm are unaffected.

### Checklist


DouweM:
I'll owe you a review when I'm back from parental leave: https://github.com/pydantic/pydantic-ai/issues/5577#issuecomment-4522457973

DouweM:
> [!NOTE]
> This comment is from DouweM's coding agent (Claude), writing on his behalf — he directed the changes and approved this message, but the words are the agent's, not his directly.

@YHallouard Thanks again for the issue and PR — and for keeping it rebased and green while Douwe was away! As promised, he's now had a proper look. The direction was right, so rather than going back and forth in review rounds, we've pushed our take directly to your branch (your commits are all preserved): c93268249

The main reshape: **batch-level events instead of per-call**. Since every deferred call already emits a regular `FunctionToolCallEvent` (which frontends use to render the tool call), a per-call `DeferredToolCallEvent` duplicated that — and each copy carried the same full batch. So it's now:

- `DeferredToolRequestsEvent(requests: DeferredToolRequests)` — once per batch, emitted before any `HandleDeferredToolCalls` handler runs (so your Temporal frontend hears about pending interactions while the handler waits for its signal)
- `DeferredToolResultsEvent(results: DeferredToolResults)` — once when a handler resolves inline; not on the `deferred_tool_results` resume path, where the caller already knows the results

The names now mirror the payload types exactly, and the request/result asymmetry from your proposal is gone. Your `_deferred.py` extraction is kept as-is — nice solution to the circular import.

We also wired up the downstream surfaces: `UIEventStream` gets explicit `handle_deferred_tool_requests`/`handle_deferred_tool_results` hooks (no-op by default, overridable for protocol mapping), and there's a serialization round-trip test since Temporal/Prefect ship each stream event across an activity boundary — relevant to your use case.

One edge case we decided against handling: if a handler-approved call defers *again* on re-execution, no second requests event fires (re-announcing the same call IDs seemed more confusing than helpful — there's a code comment explaining this). Shout if your Temporal setup needs it.

Since this ended up quite different from your original, we'd love your review — especially whether the batch events give your `ProcessEventStream` handler everything it needs.

YHallouard:
Hi again @DouweM :), 

Thanks for the review and for the work! I went back and forth on whether to batch the request or not, but since handle_deferred_tool_requests already exists, LGTM :)

> One edge case we decided against handling: if a handler-approved call defers again on re-execution, no second requests event fires (re-announcing the same call IDs seemed more confusing than helpful — there's a code comment explaining this). Shout if your Temporal setup needs it.

I'm trying to work out when we'd actually need this. The issue with not re-emitting a request is that if we hit the handler, we'd get stuck on a wait_condition and the agent would break. But since we guard with and not ctx.tool_call_approved: before raising ApprovalRequired, we shouldn't be raising it again in the first place.

PR #5605, opened 2026-05-23
fix: restore `ThinkingPart.id` in AG-UI adapter on round-trip

mayoka0:
Ensures ThinkingPart.id is successfully serialized and deserialized during round-trips.


dsfaccini:
David's AICA here: thanks for the contribution! Closing this as a duplicate of #5602, which addresses the same issue (`ThinkingPart.id` lost in AG-UI round-trips). Let's keep the discussion in one place over there.

PR #5608, opened 2026-05-23
fix(openrouter): map `thinking=False` to `reasoning_effort="none"`

### LINKED ISSUE #5587: Unified `thinking=False` is ignored by `OpenRouterModel`
### Initial Checks

### Description

`thinking=False` should map to `effort=none` for unified thinking and openrouter reasoning effort

https://github.com/pydantic/pydantic-ai/blob/130e85b09f164bd7193fc7daddfda8ef1a081945/pydantic_ai_slim/pydantic_ai/models/openrouter.py#L559-L573

```Python

```

### Logfire Trace

_No response_

- Python: 3.12
- Pydantic AI: latest
- LLM provider SDK: latest

Jennifer-00:
## Summary

Fixes #5587

When `thinking=False` is set on `OpenRouterModel`, it should disable reasoning by mapping to `effort="none"`. Instead, it was silently ignored because the inner guard `if thinking is not False` filtered out the `False` value, making it behave identically to `thinking=None`.

## Changes

- **Removed** the redundant `if thinking is not False:` guard
- **Added** `False: 'none'` to the effort map so `thinking=False` maps to `effort="none"`

## Verification

Unit test confirms the fix:
- `thinking=False` → `extra_body["reasoning"] = {"effort": "none"}`
- `thinking=None` → no reasoning config set
- `thinking=True` → `effort="medium"` (unchanged)

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/openrouter.py
+++ b/pydantic_ai_slim/pydantic_ai/models/openrouter.py
@@ -558,19 +558,19 @@ def _openrouter_settings_to_openai_settings(
     # Fall back to unified thinking when openrouter_reasoning is not set
     if 'openrouter_reasoning' not in model_settings and model_request_parameters.thinking is not None:
         thinking = model_request_parameters.thinking
-        if thinking is not False:
-            unified_reasoning: OpenRouterReasoning = {}
-            # OpenRouter only supports low/medium/high; map others to closest
-            effort_map: dict[ThinkingLevel, str] = {
-                True: 'medium',
-                'minimal': 'low',
-                'low': 'low',
-                'medium': 'medium',
-                'high': 'high',
-                'xhigh': 'high',
-            }
-            unified_reasoning['effort'] = effort_map[thinking]  # type: ignore[typeddict-item]
-            model_settings['openrouter_reasoning'] = unified_reasoning
+        unified_reasoning: OpenRouterReasoning = {}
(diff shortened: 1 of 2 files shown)


Jennifer-00:
The 1-2 checkout failures are intermittent and unrelated to my changes — they appear to be a known GitHub Actions issue with fork PRs where GITHUB_TOKEN read permissions aren't explicitly declared. All 20 pytest jobs passed.

dsfaccini:
David's AICA here: Thank you so much for this, @Jennifer-00 — and apologies for the slow turnaround.

Your fix is correct: `thinking=False` should map to `{'effort': 'none'}`, and removing the `if thinking is not False:` guard was the right call.

In the meantime, though, #5433 (by @sarth6) landed the same fix on `main` in a more complete form:
- a centralized `_OPENROUTER_EFFORT_MAP` (with `False: 'none'`),
- `enabled: True` for reasoning-optional routes,
- model-profile gating so always-on models (`openai/o3`, `openai/gpt-5`, `deepseek/deepseek-r1`, `x-ai/grok-3-mini`, …) silently ignore `thinking=False` to match their direct-route behavior,
- plus updated docs.

That overlap is also why this branch now conflicts with `main`. So we're going to close this as superseded — not because anything was wrong with it, but because the same behavior already shipped.

Really appreciate you taking the time to dig in and get the mapping right, and we'd love to see you contribute again. 🙏

PR #5611, opened 2026-05-23
fix(groq): emit `reasoning_effort="none"` when `thinking=False` on `qwen3`

### LINKED ISSUE #5445: Unified thinking: audit-found gaps across providers
While auditing the unified `thinking` subsystem during work on #5379 / #5433, two parallel reviews surfaced additional gaps across providers and integration concerns. Filing here so they're not lost. Each item is independently actionable.

### High-risk

- **Groq `qwen/qwen3-*`: `thinking=False` should emit `reasoning_effort='none'`, not `reasoning_format='hidden'`.** `profiles/groq.py:34-35` comments explicitly say qwen3 supports a true disable via `reasoning_effort='none'`, but `models/groq.py:264-278` only emits `reasoning_format` values. Today, `thinking=False` on `qwen/qwen3-32b` suppresses output but the model still reasons. Same Cause #2 pattern as #5379.

- **Bedrock-Anthropic adaptive models (Opus 4.7) emit budget shape that the API rejects.** `models/bedrock.py:650-654` unconditionally emits `{'type': 'enabled', 'budget_tokens': N}` for the anthropic variant. But Opus 4.7's profile flags `anthropic_disallows_budget_thinking=True` — direct Anthropic switches to `{'type': 'adaptive'}` for those models (`models/anthropic.py:655-658`). Bedrock-Anthropic has no parallel branch. May overlap PR #5326.

### Medium-risk

- **Cerebras `gpt-oss-120b` / `zai-glm-4.6`: effort granularity lost.** `models/cerebras.py:88-101` returns `omit` for `_translate_thinking` and only maps `thinking=False/True` to `disable_reasoning`. `thinking='low'` ≡ `thinking='high'` ≡ `thinking=True` today. Cerebras's OpenAI-compatible API does accept `reasoning_effort` on these models.

- **MoonshotAI direct provider drops `thinking` entirely.** `providers/moonshotai.py` returns bare `OpenAIModelProfile()`. `kimi-thinking-preview` and `kimi-k2-thinking` are reasoning models with `reasoning_content` parsing already wired up — but requests never carry thinking parameters. Mirror of the original OpenRouter bug class. Fix: set `supports_thinking=True` at the MoonshotAI provider level.

- **Heroku / Nebius / Together / Fireworks `gpt-oss-*` drop `thinking`.** `harmony_model_profile` (used by these providers) calls `openai_model_profile(model_name)` for `gpt-oss-120b`, which doesn't match `is_o_series`/`is_gpt_5_*`, so `supports_thinking=False`. Yet these providers do accept `reasoning_effort` for gpt-oss. Fix: in `harmony_model_profile`, detect gpt-oss family explicitly.

- **Bedrock-Mistral (Magistral) / Bedrock-Cohere (command-a-reasoning) have no `bedrock_thinking_variant`.** `providers/bedrock.py:130-133` — `_translate_thinking` can't emit anything for these. If either upstream needs a reasoning opt-in field in `additionalModelRequestFields`, this is wrong. Needs verification against Bedrock docs.

- **DeepSeek V4 `reasoning_effort='none'` to disable is untested.** `profiles/deepseek.py:11-15` — V4 has `supports_thinking=True, thinking_always_enabled=False`, so `thinking=False` maps to `reasoning_effort='none'` via `OPENAI_REASONING_EFFORT_MAP`. No cassette verifies DeepSeek V4 accepts `'none'`. If it doesn't, mark V4 always-on.

### Low-risk

- **Cohere direct `command-a-reasoning`: token budget unreachable.** `models/cohere.py` has no `_translate_thinking`; the unified `thinking` is set but never read. Reasoning still happens (always-on model) but users can't control token budget through the unified setting.

### High-risk methodology gap

- **Wire-body cassettes for Anthropic, Google, Bedrock-Anthropic disable signals.** Today's tests assert on mock kwargs or in-memory `_translate_thinking` return values, **not the actual wire body**. This is the exact methodology that let the OpenRouter `enabled: True` bug slip through. Add cassette tests asserting `vcr.requests[0].body` contains:
  - `\"thinking\": {\"type\": \"disabled\"}` for Anthropic on `thinking=False`
  - `thinking_config: {thinking_budget: 0}` for Google on `thinking=False`  
  - `additionalModelRequestFields.thinking: {type: 'disabled'}` for Bedrock-Anthropic on `thinking=False`

- **FallbackModel + heterogeneous thinking profiles is untested.** `models/fallback.py:228` runs `prepare_request` per inner model then calls `model.request()` which runs it again. Currently idempotent because the gate checks `'thinking' in model_settings` before mutating — but easy to break in a refactor. Add a test that exercises fallback from a reasoning model to a non-reasoning one with `thinking='high'`.

### Medium-risk integration gaps

- **Native tools + unified `thinking`**: tests exist for Anthropic / OpenAI Responses / Google, but only with provider-specific reasoning settings (`anthropic_thinking`, `openai_reasoning_effort`). No parametrized test combines `ModelSettings(thinking='high')` with a native `WebSearchTool`.

- **Multi-turn round-trip cassettes** are missing for Cerebras (decorative ThinkingPart survival), and OpenAI Responses `openai_send_reasoning_ids` is only tested with provider-specific `openai_reasoning_effort`, not unified `thinking`.

- **xAI streaming + unified `thinking=False`** has no cassette. Today it's only tested via non-streaming `model_request`.

### Suggested cross-cutting test

Add `tests/test_thinking_wire_contract.py` parametrized over `(provider, model_name, thinking_value, expected_wire_payload_predicate)`. Each row asserts on `vcr.requests[0].body` directly. This is the pattern that caught the original OpenRouter `enabled` miss; replicating it across providers is how future regressions of this bug class get caught at merge time.

---

Filed by @sarth6 after the audit during #5433. None of these are regressions introduced by #5433 — most predate it and are independent fixes.

devteamaegis:
## Summary

Fixes #5445.

When `thinking=False` is passed to a `qwen/qwen3-*` model via Groq, pydantic-ai previously sent `reasoning_format='hidden'`, which suppresses reasoning output but **does not disable the reasoning chain** — the model still spends tokens thinking in the background.

Groq's `qwen3` models support a `reasoning_effort` parameter. Setting it to `'none'` truly disables reasoning (the model skips its thinking chain entirely), which is the correct behaviour for `thinking=False`.

### Root cause (two-part bug)

1. **Wrong model profile for `qwen` prefix** (`providers/groq.py`): `GroqProvider.model_profile()` mapped the `'qwen'` prefix to `qwen_model_profile`, which returns `supports_thinking=False` for all qwen models. This meant the existing `_translate_thinking` logic could never reach the `reasoning_effort` path for qwen3.

2. **Missing `reasoning_effort` parameter** (`models/groq.py`): The Groq completions call never sent `reasoning_effort` at all, and `_translate_thinking` returned `'hidden'` (format, not effort) even for models that support disabling reasoning.

### Changes

**`providers/groq.py`**
- Add `qwen_groq_model_profile()` that overlays `groq_model_profile()`'s reasoning flags (`supports_thinking=True`, `thinking_always_enabled=False`) on top of the base `qwen_model_profile()`.
- Wire it into `GroqProvider.model_profile()` for the `'qwen'` prefix (mirrors the existing `meta_groq_model_profile` pattern).

**`models/groq.py`**
- Add `groq_reasoning_effort` to `GroqModelSettings` for explicit control.
- Add `_translate_thinking_effort()`: returns `'none'` when `thinking=False` for models where `supports_thinking=True` and `thinking_always_enabled=False` (i.e., qwen3 today).
- Update `_translate_thinking()`: return `NOT_GIVEN` (instead of `'hidden'`) for those same models, so the two parameters don't conflict — `reasoning_effort='none'` is the correct disable signal, not `reasoning_format`.
- Pass `reasoning_effort=...` in the `_completions_create` call.

**`tests/models/test_groq.py`**
- 9 new pure-unit tests (no network calls, no cassettes) covering: qwen3 `thinking=False` → `effort='none'`, qwen3 `thinking=True` → `NOT_GIVEN`, explicit `groq_reasoning_effort` override, legacy reasoning models (deepseek-r1, qwen-qwq) → `format='hidden'` / `effort=NOT_GIVEN`, and non-reasoning models.

### Behaviour table

| Model | `thinking=False` before | `thinking=False` after |
|---|---|---|
| `qwen/qwen3-32b` | `reasoning_format='hidden'` (compute wasted) | `reasoning_effort='none'` (reasoning skipped) |
| `deepseek-r1-*`, `qwen-qwq-*` | `reasoning_format='hidden'` | `reasoning_format='hidden'` (unchanged — can't disable) |
| Non-reasoning models | `NOT_GIVEN` | `NOT_GIVEN` (unchanged) |

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/groq.py
+++ b/pydantic_ai_slim/pydantic_ai/models/groq.py
@@ -139,6 +139,16 @@ class GroqModelSettings(ModelSettings, total=False):
     See [the Groq docs](https://console.groq.com/docs/reasoning#reasoning-format) for more details.
     """
 
+    groq_reasoning_effort: Literal['none', 'default', 'low', 'medium', 'high']
+    """The reasoning effort level for qwen3 models.
+
+    qwen3 models support setting this to ``'none'`` to truly disable reasoning
+    (the model will not run its reasoning chain at all, saving tokens).  Other
+    values control how much compute the model spends on reasoning.
+
+    See [the Groq docs](https://console.groq.com/docs/reasoning) for more details.
+    """
+
 
 @dataclass(init=False)
 class GroqModel(Model[AsyncGroq]):
@@ -271,10 +281,40 @@ class GroqModel(Model[AsyncGroq]):
             return fmt
         thinking = model_request_parameters.thinking
         if thinking is False:
-            # Groq has no true disable; 'hidden' suppresses reasoning output
+            if not self.profile.thinking_always_enabled:
+                # qwen3 and similar models that can truly disable reasoning: skip
+                # reasoning_format entirely and let _translate_thinking_effort handle it.
+                return NOT_GIVEN
+            # Legacy reasoning models (qwen-qwq, deepseek-r1, llama-4-maverick) cannot
+            # truly disable reasoning; 'hidden' suppresses the output.
             return 'hidden'
         if thinking is not None:
             return 'parsed'
+        return NOT_GIVEN
+
+    def _translate_thinking_effort(
+        self,
+        model_settings: GroqModelSettings,
+        model_request_parameters: ModelRequestParameters,
+    ) -> Literal['none', 'default', 'low', 'medium', 'high'] | NotGiven:
+        """Get the reasoning effort level for qwen3 models.
+
+        qwen3 is the only Groq reasoning model that supports ``reasoning_effort``.
+        For models where ``thinking_always_enabled=True`` (legacy reasoning models),
+        this parameter is not meaningful and is omitted.
+        """
+        if effort := model_settings.get('groq_reasoning_effort'):
+            return effort
+        # Only emit reasoning_effort for models that support thinking AND can disable it.
+        # - Non-thinking models: skip entirely (reasoning_effort is not a valid parameter)
+        # - Legacy always-on models (deepseek-r1, qwen-qwq, llama-4-maverick): skip;
+        #   those models cannot disable reasoning, so we leave reasoning_format='hidden'.
+        if not self.profile.supports_thinking or self.profile.thinking_always_enabled:
+            return NOT_GIVEN
+        thinking = model_request_parameters.thinking
+        if thinking is False:
+            return 'none'
+        # 'default' is the API default; only override when explicitly requested.
         return NOT_GIVEN
 
     @overload
@@ -341,6 +381,7 @@ class GroqModel(Model[AsyncGroq]):
                 seed=model_settings.get('seed', NOT_GIVEN),
                 presence_penalty=model_settings.get('presence_penalty', NOT_GIVEN),
                 reasoning_format=self._translate_thinking(model_settings, model_request_parameters),
+                reasoning_effort=self._translate_thinking_effort(model_settings, model_request_parameters),
                 frequency_penalty=model_settings.get('frequency_penalty', NOT_GIVEN),
                 logit_bias=model_settings.get('logit_bias', NOT_GIVEN),
                 extra_headers=extra_headers,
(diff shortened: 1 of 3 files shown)


dsfaccini:
we'll take over this one, thanks

PR #5618, opened 2026-05-23
fix: stream Gemini thought parts

he-yufeng:
﻿## Summary
- map Gemini streaming `thought` text parts to `ThinkingPart` deltas
- add a regression test covering streamed thinking followed by normal text output

## Why
The non-streaming Gemini path already treats `text` parts marked with `thought=True` as thinking. The streaming path treated the same part as normal text, so thinking content was either surfaced as user-visible text or lost in parity with the expected thinking stream representation.

## Tests
- `uv run pytest tests\models\test_gemini.py::test_stream_text_thought_part -q`
- `python -m py_compile pydantic_ai_slim\pydantic_ai\models\gemini.py tests\models\test_gemini.py`
- `uv run ruff check pydantic_ai_slim\pydantic_ai\models\gemini.py tests\models\test_gemini.py`
- `uv run ruff format --check pydantic_ai_slim\pydantic_ai\models\gemini.py tests\models\test_gemini.py`

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/gemini.py
+++ b/pydantic_ai_slim/pydantic_ai/models/gemini.py
@@ -480,6 +480,15 @@ class GeminiStreamedResponse(StreamedResponse):
             gemini_part: _GeminiPartUnion
             for gemini_part in candidate['content']['parts']:
                 if 'text' in gemini_part:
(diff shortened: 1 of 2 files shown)


Kludex:
Thanks, but we are removing `GeminiModel` for V2.

PR #5619, opened 2026-05-23
fix: complete stream output after early break

### LINKED ISSUE #5615: [streaming-resilience-sweep] Streaming: `stream_output()` doesn't set `is_complete=True` on early break
## Impact

Users who partially consume a stream via `stream_output()` and then break early will find `is_complete` is `False` even though the stream was properly drained during cleanup. This is inconsistent with `stream_text()` which correctly sets `is_complete = True` after early break.

## Focus Area & Code Path

`StreamedRunResult.stream_output()` method in `pydantic_ai_slim/pydantic_ai/result.py`

## Reproduction

```python
import asyncio
from pydantic_ai import Agent
from pydantic_ai.models.function import FunctionModel

async def stream_slow_text(messages, info):
    for word in ["Hello", " ", "world", "!"]:
        yield word
        await asyncio.sleep(0.01)

async def test():
    model = FunctionModel(stream_function=stream_slow_text)
    agent = Agent(model, output_type=str)

    async with agent.run_stream('test') as result:
        # Early break from stream_output
        count = 0
        async for output in result.stream_output():
            count += 1
            if count >= 1:
                break
        
        print(f"is_complete after early break: {result.is_complete}")
        # Expected: True
        # Actual: False  <-- BUG

asyncio.run(test())
```

**Command to run:**
```bash
cd /home/runner/work/pydantic-ai/pydantic-ai
source /tmp/gh-aw/uv/cache/environments-v2/pydantic-ai-runner-c87898cc2b6b2057/bin/activate
python /tmp/gh-aw/agent/test_streaming_bug.py
```

## Expected vs Actual

**Expected:** After breaking early from `stream_output()`, `is_complete` should be `True` (consistent with `stream_text()` behavior)

**Actual:** After breaking early from `stream_output()`, `is_complete` is `False`

**Comparison:**
- `stream_text(delta=True)` with early break: `is_complete = True` ✓
- `stream_output()` with early break: `is_complete = False` ✗

## Root Cause

In `StreamedRunResult.stream_output()`:

```python
async def stream_output(self, *, debounce_by: float | None = 0.1) -> AsyncIterator[OutputDataT]:
    ...
    elif self._stream_response is not None:
        async for output in self._stream_response.stream_output(debounce_by=debounce_by):
            yield output
        await self._marked_completed(self.response)  # NOT called if break happens!
```

When the user breaks early from the `async for` loop, `await self._marked_completed(self.response)` is never executed, so `is_complete` remains `False`.

In contrast, `stream_text()` works because `AgentStream.stream_text()` uses `aclosing()` to wrap the underlying iterator, which ensures proper cleanup and draining of the stream, eventually leading to `is_complete` being set to `True`.

## Evidence

- Test output shows `is_complete = False` after breaking from `stream_output()`
- Test output shows `is_complete = True` after breaking from `stream_text()`
- Code path: `result.py:537-565` (`stream_output`), `result.py:562-600` (`stream_text`)

he-yufeng:
## Summary
- drain incomplete streamed runs when `agent.run_stream()` exits normally
- keep `StreamedRunResult.is_complete` and response state complete after callers break out of `stream_output()` early
- add a regression test for the early-break cleanup path

## To verify
- `uv run pytest tests\test_streaming.py::test_run_stream_context_completes_after_stream_output_break tests\test_streaming.py::test_stream_text_early_break_cleanup -q`
- `python -m py_compile pydantic_ai_slim\pydantic_ai\agent\abstract.py pydantic_ai_slim\pydantic_ai\result.py tests\test_streaming.py`
- `uv run ruff check pydantic_ai_slim\pydantic_ai\agent\abstract.py pydantic_ai_slim\pydantic_ai\result.py tests\test_streaming.py`
- `uv run ruff format --check pydantic_ai_slim\pydantic_ai\agent\abstract.py pydantic_ai_slim\pydantic_ai\result.py tests\test_streaming.py`
- `git diff --check`

Fixes #5615


he-yufeng:
Follow-up pushed in 37aa4aeb. The remote coverage failure came from the regression test using sync for ... break, which introduced an uncovered loop branch. I changed it to wait anext(...) so it still exercises early stream-output exit without adding a coverage-only branch.\n\nValidation run locally:\n- uv run pytest tests\\test_streaming.py::test_run_stream_context_completes_after_stream_output_break tests\\test_streaming.py::test_stream_text_early_break_cleanup -q\n- uv run ruff check tests\\test_streaming.py pydantic_ai_slim\\pydantic_ai\\agent\\abstract.py\n- uv run ruff format --check tests\\test_streaming.py pydantic_ai_slim\\pydantic_ai\\agent\\abstract.py\n- uv run pyright tests\\test_streaming.py pydantic_ai_slim\\pydantic_ai\\agent\\abstract.py\n- uv run python -m py_compile tests\\test_streaming.py pydantic_ai_slim\\pydantic_ai\\agent\\abstract.py\n- git diff --check

he-yufeng:
﻿Follow-up pushed in 5847d972.

Changes:
- narrowed the implicit cleanup suppression to NotImplementedError/OSError;
- simplified the new stream generators so coverage no longer sees the loop-exit branch partial.

Validation on Windows:
- python -m py_compile pydantic_ai_slim\pydantic_ai\agent\abstract.py pydantic_ai_slim\pydantic_ai\result.py tests\test_streaming.py
- uv run pytest tests/test_streaming.py -k "run_stream_records_partial_response_after_early_break or run_stream_records_partial_response_on_exception_in_body or run_stream_early_break_when_provider_lacks_cancel_support": 4 passed
- uv run ruff check pydantic_ai_slim/pydantic_ai/agent/abstract.py pydantic_ai_slim/pydantic_ai/result.py tests/test_streaming.py
- uv run ruff format --check pydantic_ai_slim/pydantic_ai/agent/abstract.py pydantic_ai_slim/pydantic_ai/result.py tests/test_streaming.py

pydantic_ai_slim/pydantic_ai/result.py
```diff
@@ -756,16 +774,55 @@ async def cancel(self) -> None:
...
+        if self.is_complete or self._stream_response is None:
+            return
+        # Keep this tuple narrow so a real bug in `close_stream()` still surfaces.
+        # Widening to bare `Exception` would also let a body exception be shadowed
+        # by a cleanup-side failure during implicit `__aexit__` cleanup.
+        # The two-deep reach into `_raw_stream_response` is intentional — no user
+        # has asked for the cancel-error tuple at the `AgentStream` surface yet,
+        # so we don't add a public delegator for one internal helper to use.
+        cancel_errors = (
+            NotImplementedError,
```
adtyavrdhn, line 815: Why?

adtyavrdhn: Pretending that cancellations went through when they didn't doesn't seem right

he-yufeng:
﻿Good point. I changed the cancel path so failed cancels no longer mark the result complete or record an interrupted response.

The current behavior is now:

- explicit `await result.cancel()` still propagates provider cancel failures;
- implicit `run_stream.__aexit__` still swallows the narrow documented cancel errors so a clean early-break block does not raise;
- if that implicit cancel fails, the result remains incomplete instead of pretending cancellation succeeded.

Validation:

- `PYTHONUTF8=1 uv run pytest tests/test_streaming.py -k "run_stream_early_break_when_provider_lacks_cancel_support or run_stream_early_break_swallows_transport_cancel_error or run_stream_cancel or run_stream_cancel_all_messages_includes_interrupted_response" -q`
- `PYTHONUTF8=1 uv run ruff check pydantic_ai_slim/pydantic_ai/result.py tests/test_streaming.py`
- `PYTHONUTF8=1 uv run ruff format --check pydantic_ai_slim/pydantic_ai/result.py tests/test_streaming.py`
- `python -m py_compile pydantic_ai_slim\pydantic_ai\result.py tests\test_streaming.py`
- `git diff --check`

The `PYTHONUTF8=1` prefix is needed on my Windows shell because the build backend otherwise reads `pyproject.toml` with the default GBK codec and fails before tests start.

pydantic_ai_slim/pydantic_ai/result.py
dsfaccini: gonna let claude take a look at your comment

dsfaccini: I lost the session when my system crashed but claude was rather sure you were being overly cautious lol

any way, we have no rush to merge this so I'm just gonna let this one sit I guess

dsfaccini:
hey @he-yufeng thank you. lets leave this PR alone for a while, @DouweM will take a look when he's back.

we're fine deferring merging this bc the issue was raised by github actions (would be worth checking if there are other past issues referencing the same problem) so there's no urgency in fixing "a bug a user is having"

and judging by the number of issues you're opening I'm willing to bet they're mostly automated and not you trying to fix an issue you're experiencing.

so any way, we'll set a pause on this PR

dsfaccini:
David's AICA here: Reviewed this adversarially — reproduced on both `main` and this branch and ran the suite. It holds up.

- **Bug confirmed on `main`:** after an early `break` from `stream_output()` and exiting the `async with`, the partial `ModelResponse` is lost from `all_messages()` and `is_complete=False`. This branch drains + records it (`state='interrupted'`) at `run_stream.__aexit__`, reusing the existing `cancel()` recording path rather than inventing a new one. `stream_text()` / `stream_response()` had the same loss on `main` and are fixed consistently here (the original #5615 framing that `stream_text` already worked on early break was incorrect).
- **Attacked the claims — all hold:** full consumption isn't mislabeled `interrupted` (the `is_complete` guard skips the backstop), no double-recording (explicit `cancel()` then exit → exactly one `ModelResponse`), the partial response is well-formed (correct `parts`/`usage`, `finish_reason=None`), and the regression test genuinely fails on `main` without the fix (not a tautology).

**One edge worth your explicit call, @DouweM** (not a blocker): if a provider's `close_stream()` raises an error *outside* `get_stream_cancel_errors()` while the `async with` body *also* raised, the cleanup error now **shadows** the body exception (on `main` the body error surfaces; here it's demoted to `__context__`). It's narrow — it needs a custom/mis-declared provider whose `close_stream()` raises an undeclared error — and the PR's `get_stream_cancel_errors` docstring already warns about this. Widening the swallow-tuple to avoid it would be net-negative (it would re-hide real `close_stream()` bugs), so this is a "is silent body-exception demotion-to-`__context__` acceptable?" sign-off, not a fix request.

Minor non-blocking nits: a never-started stream now records `state='interrupted'` on exit (untested), and the two-deep `_raw_stream_response.get_stream_cancel_errors()` private reach is self-flagged with a `pyright: ignore`.

DouweM:
Thank you for this, @he-yufeng — and apologies it sat for a while. The behavior you fixed (a clean early `break` out of a `stream_*` leaving the response un-recorded) is now handled on `main`: the stream is drained and the interrupted response recorded when the `run_stream()` / `run_stream_events()` context manager exits, with interrupted request/response capture via #5364 and the run-cancellation stack (#6497 / #6498).

Since that resolves it upstream, we're closing this as superseded rather than asking you to rebase onto a substantially changed tree. Really appreciate the contribution and the clear repro — it fed directly into the broader cancellation semantics work tracked in #5838.

_Posted by Claude Code on behalf of @DouweM._

PR #5620, opened 2026-05-23
fix: serialize callable thinking provider details

### LINKED ISSUE #5612: AgentStreamEvent: ThinkingPartDelta with callable provider_details cannot be serialized (OpenAI gpt-oss raw CoT streaming)
## Summary

`ThinkingPartDelta.provider_details` is typed as `ProviderDetailsDelta`, which allows a `Callable` (`pydantic_ai_slim/pydantic_ai/messages.py:145`). During streaming, the parts manager resolves the callable when updating its internal `ThinkingPart`, but the `PartDeltaEvent` it yields still carries the original delta with the raw callable inside. Serializing that event with Pydantic fails:

```
PydanticSerializationError: Unable to serialize unknown type: <class 'function'>
```

## Exposed path

Verified on main (5b1216af1):

1. `models/openai.py:3707`: when streaming raw chain of thought from gpt-oss models via the Responses API (`ResponseReasoningTextDeltaEvent`), the model passes `provider_details=_make_raw_content_updater(chunk.delta, chunk.content_index)`. This is a closure (defined at `models/openai.py:4087`) because the update is "append this string to `provider_details['raw_content'][index]`", which a plain dict merge cannot express.
2. `_parts_manager.py:284-290` (`handle_thinking_delta`): builds `ThinkingPartDelta(provider_details=<callable>)`, applies it to the internal `ThinkingPart` (the callable is resolved to a dict there), then yields `PartDeltaEvent(index=..., delta=part_delta)` with the callable still on the delta.
3. Any consumer that serializes stream events fails. An in-repo example: the Temporal integration wraps every `AgentStreamEvent` in `_EventStreamHandlerParams` (`durable_exec/temporal/_agent.py:53`) and passes it across an activity boundary, where Temporal's Pydantic data converter JSON-serializes it. Streaming gpt-oss raw CoT through a `TemporalAgent` with an `event_stream_handler` crashes mid-stream.

## Reproduction

```python
from pydantic import TypeAdapter
from pydantic_ai.messages import AgentStreamEvent, PartDeltaEvent, ThinkingPartDelta

delta = ThinkingPartDelta(
    content_delta='x',
    provider_details=lambda d: {**(d or {}), 'raw_content': ['x']},
)
event = PartDeltaEvent(index=0, delta=delta)
TypeAdapter(AgentStreamEvent).dump_json(event)
# PydanticSerializationError: Unable to serialize unknown type: <class 'function'>
```

No chaining is required. A single delta carrying a callable fails.

Note: this issue originally framed the bug around chained deltas (`ThinkingPartDelta.apply()` on another `ThinkingPartDelta`). That branch exists in `messages.py` but the framework never executes it: the parts manager creates a `ThinkingPart` on the first thinking delta and only ever applies deltas to parts, never to other deltas. The framing above is the actual reachable path.

## Impact

- `TemporalAgent` streaming with gpt-oss models that emit raw CoT.
- Any application that persists or transmits raw `AgentStreamEvent`s.
- Only the OpenAI Responses raw CoT path produces callables. Anthropic and other providers pass dicts or nothing, so they are unaffected.

Severity: low to medium. Requires gpt-oss raw CoT streaming plus event serialization.

## Suggested fix

Resolve the callable in `handle_thinking_delta` before constructing the emitted event. The parts manager knows the pre-update `provider_details`, so it can emit the resolved dict on the event delta instead of the callable. This also fixes round-tripping: a deserialized event would otherwise lose the update entirely. A serializer on `ProviderDetailsDelta` that drops callables would silence the error but lose data, so it is the weaker option.

he-yufeng:
## Summary
- make callable `ThinkingPartDelta.provider_details` JSON-serializable by emitting `null`
- keep the existing callable merge behavior for runtime application to `ThinkingPart`
- add a regression test for serializing chained thinking deltas through `AgentStreamEvent`

## To verify
- `uv run pytest tests\test_messages.py::test_thinking_part_delta_callable_provider_details_serializable tests\test_messages.py::test_thinking_part_delta_apply_to_thinking_part_delta -q`
- `python -m py_compile pydantic_ai_slim\pydantic_ai\messages.py tests\test_messages.py`
- `uv run ruff check pydantic_ai_slim\pydantic_ai\messages.py tests\test_messages.py`
- `uv run ruff format --check pydantic_ai_slim\pydantic_ai\messages.py tests\test_messages.py`
- `git diff --check`

Fixes #5612

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/messages.py
+++ b/pydantic_ai_slim/pydantic_ai/messages.py
@@ -142,7 +142,17 @@ ForceDownloadMode: TypeAlias = bool | Literal['allow-local']
 - `'allow-local'`: The file is always downloaded, allowing private IPs but still blocking cloud metadata.
 """
 
-ProviderDetailsDelta: TypeAlias = dict[str, Any] | Callable[[dict[str, Any] | None], dict[str, Any]] | None
+
+def _serialize_provider_details_delta(value: Any) -> dict[str, Any] | None:
+    if callable(value):
+        return None
+    return value
+
+
+ProviderDetailsDelta: TypeAlias = Annotated[
+    dict[str, Any] | Callable[[dict[str, Any] | None], dict[str, Any]] | None,
+    pydantic.PlainSerializer(_serialize_provider_details_delta, return_type=dict[str, Any] | None),
+]
 """Type for provider_details input: can be a static dict, a callback to update existing details, or None."""
 
 
--- a/tests/test_messages.py
+++ b/tests/test_messages.py
@@ -7,6 +7,7 @@ import pytest
 from pydantic import TypeAdapter
 
 from pydantic_ai import (
+    AgentStreamEvent,
     AudioUrl,
     BinaryContent,
     BinaryImage,
@@ -22,6 +23,7 @@ from pydantic_ai import (
     MultiModalContent,
     NativeToolCallPart,
     NativeToolReturnPart,
+    PartDeltaEvent,
     RequestUsage,
     RetryPromptPart,
     TextContent,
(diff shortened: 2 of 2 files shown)


he-yufeng:
Follow-up pushed in 59c5ad0b. It fixes the pyright complaint by typing the stream-event TypeAdapter in the regression test and adds coverage for the non-callable provider_details serializer branch.\n\nValidation run locally:\n- uv run pytest tests\\test_messages.py::test_thinking_part_delta_callable_provider_details_serializable -q\n- uv run pyright tests\\test_messages.py\n- uv run ruff check tests\\test_messages.py\n- uv run ruff format --check tests\\test_messages.py\n- uv run python -m py_compile tests\\test_messages.py\n- git diff --check

hramezani:
Superseded by #6294, which rebases this fix onto current `main`, documents why callable `provider_details` serialize to `null`, and tightens the regression test with structured assertions plus a round-trip check. Original authorship is preserved in the first two commits — thanks @he-yufeng!

PR #5568, opened 2026-05-21
Fix `AnthropicModel` dropping code execution tool results when continuing a conversation

2830500285:
- Closes #5567

### Summary

- Preserve Anthropic `bash_code_execution` and `text_editor_code_execution` tool result blocks when replayed history stores list-shaped result content.
- Add a regression test covering both Anthropic code execution variants so same-provider replay does not emit a `server_tool_use` without the matching `*_code_execution_tool_result` block.

### Validation

- `PYTHONPATH=... .venv\Scripts\python.exe -m pytest tests/models/test_anthropic.py::test_anthropic_code_execution_tool_message_replay_with_list_results`
- `PYTHONPATH=... .venv\Scripts\python.exe -m pytest tests/models/test_anthropic.py::test_anthropic_text_editor_code_execution_tool_message_replay tests/models/test_anthropic.py::test_anthropic_bash_code_execution_tool_message_replay tests/models/test_anthropic.py::test_anthropic_code_execution_tool_message_replay_with_list_results tests/models/test_anthropic.py::test_anthropic_code_execution_tool_message_replay_infers_anthropic_tool_name`
- `.venv\Scripts\python.exe -m ruff check pydantic_ai_slim/pydantic_ai/models/anthropic.py tests/models/test_anthropic.py`
- `uv run --no-project --with pyright pyright pydantic_ai_slim/pydantic_ai/models/anthropic.py`

### Checklist

- [ ] Any **AI generated code** has been reviewed line-by-line by the human PR author, who stands by it.
- [x] No **breaking changes** in accordance with the [version policy](https://github.com/pydantic/pydantic-ai/blob/main/docs/version-policy.md).
- [x] **PR title** is fit for the [release changelog](https://github.com/pydantic/pydantic-ai/releases).

PR #5622, opened 2026-05-23
fix(google): handle list content in `code_execution` tool return parts

### LINKED ISSUE #5614: [bug-hunter] Google Model drops code execution tool results with list content on message replay
## Impact

When replaying a conversation history with `NativeToolReturnPart` containing code execution tool results, if the `content` field is a `list` (rather than a `dict`), the Google model raises `UnexpectedModelBehavior('Unknown native tool name: code_execution')` instead of properly serializing the result.

This affects:
- Users who serialize/deserialize message history containing code execution results with list content
- Multi-turn conversations where the model uses code execution tools

## Reproduction Steps

1. Create a `NativeToolReturnPart` with `tool_name='code_execution'` and `content` as a list
2. Call `_native_tool_return_part_dict()` with this item
3. Observe that it raises `UnexpectedModelBehavior` instead of returning a proper `PartDict`

## Expected vs Actual

**Expected:** The function should return a `PartDict` with `'code_execution_result': item.content` regardless of whether content is a dict or list.

**Actual:** The function raises:
```
UnexpectedModelBehavior: Unknown native tool name: 'code_execution'
```

This happens because the code at `google.py:1674` only checks for `dict`:
```python
if item.tool_name == CodeExecutionTool.kind and isinstance(item.content, dict):
```

When content is a list, this check fails, and the code falls through to:
```python
tool_type = _NATIVE_TOOL_NAME_TO_TOOL_TYPE.get(item.tool_name)
if tool_type is None:  # pragma: no cover
    raise UnexpectedModelBehavior(f'Unknown native tool name: {item.tool_name!r}')
```

## Failing Test

```python
import pytest
from pydantic_ai.messages import NativeToolReturnPart
from pydantic_ai.models.google import _native_tool_return_part_dict
from pydantic_ai.native_tools import CodeExecutionTool


async def test_google_code_execution_list_content_message_replay():
    """Test that list content for code execution tools is handled correctly."""
    bash_result = [
        {
            'return_code': 0,
            'stderr': '',
            'stdout': 'hello\n',
            'type': 'bash_code_execution_result',
        }
    ]
    
    item = NativeToolReturnPart(
        provider_name='google',
        tool_name=CodeExecutionTool.kind,
        content=bash_result,  # List content
        tool_call_id='test_id',
    )
    
    # This raises UnexpectedModelBehavior
    result = _native_tool_return_part_dict(
        item, 
        frozenset({'google'}), 
        None,
        supports_tool_combination=True,
    )
    
    assert result is not None
    assert 'code_execution_result' in result
    assert result['code_execution_result'] == bash_result
```

## Evidence

- **File:** `pydantic_ai_slim/pydantic_ai/models/google.py:1674`
- **Same bug was fixed in anthropic.py:** PR #5568 changed `isinstance(response_part.content, dict)` to `isinstance(response_part.content, dict | list)` at `anthropic.py:1440` and `anthropic.py:1456`
- **The fix:** Change line 1674 from:
  ```python
  if item.tool_name == CodeExecutionTool.kind and isinstance(item.content, dict):
  ```
  to:
  ```python
  if item.tool_name == CodeExecutionTool.kind and isinstance(item.content, dict | list):
  ```

## Root Cause

The `_native_tool_return_part_dict` function in `google.py` only handles `dict` content for code execution tools, but `NativeToolReturnPart.content` can be either a `dict` or a `list` (as documented by `ToolReturnContent` type alias in `messages.py`).

strawgate:
## Summary

Fix `_native_tool_return_part_dict` to accept `list` content (not just `dict`) for code execution tool returns on the Google/Gemini model.

## Issue

Closes #5614

## Changes

- Widened the `isinstance` check in `_native_tool_return_part_dict` from `dict` to `dict | list`
- Added tests for both `list` and `dict` content types

## Context

When replaying message history containing `NativeToolReturnPart` with `tool_name='code_execution'` and `content` as a `list` (e.g., bash code execution results), the function raised `UnexpectedModelBehavior` instead of properly serializing the result.

This is the same pattern already applied to the Anthropic model in PR #5568.

- [ ] AI generated code

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/google.py
+++ b/pydantic_ai_slim/pydantic_ai/models/google.py
@@ -1671,7 +1671,7 @@ def _native_tool_return_part_dict(
 ) -> PartDict | None:
     if item.provider_name not in accepted_provider_names:
         return None
-    if item.tool_name == CodeExecutionTool.kind and isinstance(item.content, dict):
+    if item.tool_name == CodeExecutionTool.kind and isinstance(item.content, dict | list):
         return _attach_signature(
             {'code_execution_result': cast(CodeExecutionResultDict, item.content)},  # pyright: ignore[reportUnknownMemberType]
             signature,
--- a/tests/models/test_google.py
+++ b/tests/models/test_google.py
@@ -69,6 +69,7 @@ from pydantic_ai.messages import (
 )
 from pydantic_ai.models import DEFAULT_HTTP_TIMEOUT, ModelRequestParameters
 from pydantic_ai.native_tools import (
+    CodeExecutionTool,
     FileSearchTool,
     ImageGenerationTool,
     UrlContextTool,  # pyright: ignore[reportDeprecated]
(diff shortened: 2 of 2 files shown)


strawgate:
Closing after deeper analysis - the premise does not hold for Google. Google's _map_code_execution_result always calls model_dump() on CodeExecutionResult which the SDK defines as {outcome, output, id} — a dict, never a list. There is no code path through which a Google-provider NativeToolReturnPart for code execution would have list content. The provider_name gate at the top of _native_tool_return_part_dict also ensures cross-provider parts never reach this check. The original isinstance check to dict was correct and the pragma no cover comment was already signalling this fallthrough was unreachable. Widening to dict|list would silently pass malformed data to the Google API instead of raising an error as intended.

PR #5642, opened 2026-05-25
Add `OrcaRouterProvider` for OrcaRouter

akf66:
## Summary

Adds a thin built-in provider for [OrcaRouter](https://orcarouter.ai), which exposes an OpenAI-compatible `/v1/chat/completions` endpoint. The provider is wired through the existing `OpenAIChatModel` path, mirroring the pattern used by `VercelProvider` / `TogetherProvider` / `DeepSeekProvider`.

- new `pydantic_ai.providers.orcarouter.OrcaRouterProvider`
- registered in `infer_provider_class`
- listed in `OpenAIChatCompatibleProvider` so `'orcarouter:<model>'` resolves to `OpenAIChatModel` with the new provider
- profile lookup splits `vendor/model` so upstream profiles (`anthropic`, `deepseek`, `google`, `meta-llama`, `mistral` / `mistralai`, `moonshotai`, `openai`, `qwen`, `x-ai` / `xai`) are picked up and then re-projected through `OpenAIJsonSchemaTransformer`
- reads the API key from `ORCAROUTER_API_KEY`

Usage:

```python
from pydantic_ai import Agent

agent = Agent('orcarouter:openai/gpt-4o-mini')
```

## Why a built-in provider rather than `OpenAIProvider(base_url=...)`

Functionally equivalent today; the built-in version is a convenience that gives users a stable `'orcarouter:'` prefix, env-var contract, and upstream-vendor profile mapping out of the box. Maintainers — if the bar for new built-in providers isn't met here, happy to close and rely on the generic `OpenAIProvider` instead.

## Tests

- new `tests/providers/test_orcarouter.py` (provider init, missing-key error, `model_profile` vendor mapping, fallback for unknown vendor / no slash)
- new parametrization entry in `tests/providers/test_provider_names.py`
- `uv run pytest tests/providers/` — 334 passed locally
- `make lint` / `make typecheck` clean; full pre-commit hook chain passes

## Docs

- `docs/models/overview.md`: OrcaRouter added under "OpenAI-compatible Providers"
- `docs/models/openai.md`: new "OrcaRouter" section with env-var and `OrcaRouterProvider` examples
- `docs/api/providers.md`: mkdocstrings entry for `OrcaRouterProvider`

- Closes #<issue>

### Checklist

### DIFF
--- a/docs/api/providers.md
+++ b/docs/api/providers.md
@@ -40,6 +40,8 @@
 
 ::: pydantic_ai.providers.openrouter.OpenRouterProvider
 
+::: pydantic_ai.providers.orcarouter.OrcaRouterProvider
+
 ::: pydantic_ai.providers.vercel.VercelProvider
 
 ::: pydantic_ai.providers.huggingface.HuggingFaceProvider
--- a/docs/models/openai.md
+++ b/docs/models/openai.md
@@ -579,6 +579,36 @@ agent = Agent(model)
 ...
 ```
 
+### OrcaRouter
+
+To use [OrcaRouter](https://orcarouter.ai), first create an API key in the OrcaRouter dashboard.
+
+You can set the `ORCAROUTER_API_KEY` environment variable and use [`OrcaRouterProvider`][pydantic_ai.providers.orcarouter.OrcaRouterProvider] by name:
+
+```python
+from pydantic_ai import Agent
+
+agent = Agent('orcarouter:openai/gpt-4o-mini')
+...
+```
+
+Or initialise the model and provider directly:
+
+```python
+from pydantic_ai import Agent
+from pydantic_ai.models.openai import OpenAIChatModel
+from pydantic_ai.providers.orcarouter import OrcaRouterProvider
+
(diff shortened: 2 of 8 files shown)


dsfaccini:
David's AICA here: Thanks for the PR, but we're going to close this as not planned.

We don't want to add a built-in provider or documentation entry for OrcaRouter at this time. Pydantic AI already supports OpenAI-compatible services through the generic `OpenAIProvider(base_url=...)` path, and adding a dedicated provider prefix plus docs listing for every OpenAI-compatible router is not something we can scale or maintain.

Including a third-party service in the built-in provider list or project docs can also be read by users as Pydantic AI vouching for that service. In the current environment, we need to be careful across dependencies, docs, examples, and integrations so we don't expand users' operational or supply-chain assumptions around services we don't operate or audit.

If OrcaRouter users want a shorter setup path, the best place is OrcaRouter's own documentation: show how to use `OpenAIProvider` with OrcaRouter's base URL and API key. No Pydantic AI code change is needed for that.

PR #5644, opened 2026-05-25
gh-aw: fix Logfire 401s, install rg/uv in sandbox, improve shim + prompts

strawgate:
Fixes several issues found in the first weeks of running the gh-aw agentic workflows.

### Logfire observability fixes
- Rename the write token secret from `LOGFIRE_TOKEN` → `LOGFIRE_WRITE_TOKEN` (with backward-compat fallback)
- Fix the OTLP endpoint from the old `logfire-api.pydantic.dev` to the correct regional URL (`logfire-eu.pydantic.dev` by default), and document how to switch regions
- Also rename `LOGFIRE_READ_EXTERNAL_VARIABLES` → `LOGFIRE_PROMPT_TOKEN` in the workflow `logfire-read-key` inputs for consistency
- Add `opentelemetry-instrumentation-httpx` to the inline deps so outbound HTTP calls made by the agent are traced

### Sandbox tool installation
Adds `.github/scripts/install-sandbox-tools.sh`, called as a `pre-agent-step` in all 9 workflows. It installs `ripgrep` via `uv tool install` and symlinks both `rg` and `uv` into `/opt/hostedtoolcache/gh-aw-tools/current/x64/bin`, which AWF auto-discovers and merges into the container PATH. This follows the pattern used by `elastic/ai-github-actions`.

Previously the agent would fail or fall back to `grep -rn` because `rg` was not reliably available. `uv` (needed to run tests/lint) also now reliably resolves from PATH.

### Shim improvements (`pydantic_ai_gh_aw_shim/`)
- Raise `TOOL_RESULT_TRIM_THRESHOLD` 2 000 → 10 000 chars — the old threshold was too aggressive for a 200k-token context window and caused unnecessary truncation
- Raise `TOOL_RESULT_HEAD_TAIL_CHARS` 800 → 4 000 chars so truncated snippets are actually useful
- Raise `MAX_TOOL_OUTPUT` 16 000 → 50 000 chars so most Python files fit in a single read
- Emit structured `system` events to the stream-json output for `compaction_trim` and `compaction_summary_*` — makes runs observable in Logfire
- Expand `INSTRUCTIONS` to cover: parallel tool calls, large file reads, `rg`/`uv` availability, dev-env setup (`make install`), and the `gh issue list` 403 workaround (use MCP tools)

### Prompt improvements
- Add a mandatory pre-exploration deduplication step to all 9 agent prompts (search for existing issues via MCP before any code exploration)
- Add sandbox environment notes (parallel calls, read size, tooling) to all prompts
- Narrow `docs-drift` scope to negative drift only (avoids AI-generated noise)
- Add a `shared/tool-hints.md` fragment imported by all workflows so the sandbox hints stay in sync

### gh-aw-actions version
Bump from v0.74.4 → v0.74.8 to match current upstream `actions-lock.json`.

## Security review (safe update mode)

New secrets added in this PR:
- `LOGFIRE_WRITE_TOKEN` — Logfire project write token, used exclusively as the `Authorization` header for OTLP ingest to `logfire-eu.pydantic.dev` and passed to the agent container for the Logfire Python SDK. No exfiltration risk beyond what `LOGFIRE_TOKEN` already did.
- `LOGFIRE_PROMPT_TOKEN` — Logfire read token for fetching managed prompts via the existing `fetch-dynamic-prompt` action. Replaces `LOGFIRE_READ_EXTERNAL_VARIABLES` (same purpose, new name for consistency). Read-only token scoped to prompt variable lookups.

Neither secret is logged, forwarded to third parties, or used in any way beyond their stated purpose. Both follow the same pattern as the secrets they replace.

- Closes #

### Checklist

### DIFF
--- a/.github/scripts/install-sandbox-tools.sh
+++ b/.github/scripts/install-sandbox-tools.sh
@@ -1,0 +1,41 @@
+#!/usr/bin/env bash
+# Install tools that the agent needs inside the AWF sandbox.
+#
+# Runs in `pre-agent-steps` (on the runner, open network, after checkout)
+# before AWF starts the firewalled container. Tools are exposed via:
+#
+#   1. /opt/hostedtoolcache/gh-aw-tools/current/x64/bin — AWF auto-scans
+#      hostedtoolcache bin dirs and merges them into the container PATH.
+#   2. $GITHUB_PATH — AWF reads this file at container startup and merges
+#      entries into AWF_HOST_PATH (the container's PATH).
+#
+# This follows the pattern used by elastic/ai-github-actions.
+set -euo pipefail
+
+toolcache_bin="/opt/hostedtoolcache/gh-aw-tools/current/x64/bin"
+sudo mkdir -p "$toolcache_bin"
+
+# --- ripgrep ---
+# The agent's native Grep tool wraps `rg` for fast code search.
+echo "[install-sandbox-tools] Installing ripgrep..."
+uv tool install ripgrep --force --quiet
+rg_path="$(uv tool dir --bin)/rg"
+if [ -x "$rg_path" ]; then
+  sudo ln -sf "$rg_path" "$toolcache_bin/rg"
+  echo "[install-sandbox-tools] rg -> $toolcache_bin/rg"
+else
+  echo "::warning::ripgrep install succeeded but rg binary not found at $rg_path"
+fi
+
+# --- uv ---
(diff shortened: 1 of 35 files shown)


strawgate:
im a maintainer please don't close

strawgate:
Reopening from pydantic/pydantic-ai directly so CI runs.

PR #5647, opened 2026-05-25
fix: update Cerebras reasoning settings and model names

### LINKED ISSUE #5606: Cerebras: multi-turn reasoning history broken for zai-glm, and `cerebras_disable_reasoning` is deprecated upstream
### Initial Checks

### Description

Running `CerebrasModel('zai-glm-4.7')` against the bare `https://api.cerebras.ai/v1` endpoint, I hit two problems that I believe come from the current `CerebrasProvider` / `CerebrasModel` integration:

**1. Multi-turn reasoning history can't be replayed for `zai-glm` (and likely `qwen`).**

Cerebras' docs say that for GLM/Qwen, the previous reasoning must be sent back inside `<think>...</think>` tags inside the assistant message's `content` ([Retain Reasoning Context for GLM/Qwen](https://inference-docs.cerebras.ai/capabilities/reasoning)). Today, `CerebrasProvider.model_profile()` sets `supports_thinking=True` for `zai`/`gpt-oss` but does not set `openai_chat_send_back_thinking_parts`, so it defaults to `'auto'`. In `'auto'` mode, if the response populated the `reasoning` field (or any custom thinking field), the assistant message is reconstructed with that field on the next turn rather than with `<think>` tags in `content` — which is not what Cerebras GLM/Qwen accepts. The result is that prior thinking is effectively lost across turns, and in tool-using flows the model can refuse to continue.

**2. `cerebras_disable_reasoning` is built on a setting Cerebras has deprecated.**

The [GLM-4.7 migration guide](https://inference-docs.cerebras.ai/resources/glm-47-migration) states that `disable_reasoning` was deprecated on 2026-03-24 in favor of the standard OpenAI `reasoning_effort="none"`. Today `_cerebras_settings_to_openai_settings` writes `extra_body['disable_reasoning']`, which is the deprecated path.

The same migration page also documents a new `clear_thinking` flag (passed via `extra_body` with the OpenAI SDK) that controls whether prior thinking is preserved across turns. It is not currently exposed by `CerebrasModelSettings`, so users wanting multi-turn / cache-friendly behavior have to drop down to raw `extra_body`.

**3. Minor: `LatestCerebrasModelNames` still lists `zai-glm-4.6`.**

That version is gone; the current GLM model on Cerebras is `zai-glm-4.7`.

### Expected behavior

- Multi-turn agent runs against `zai-glm-4.7` should preserve prior reasoning across turns out of the box.
- `cerebras_disable_reasoning` should map onto the supported, non-deprecated parameter (`reasoning_effort="none"`).
- There should be a first-class way (`cerebras_clear_thinking` setting, or similar) to set `clear_thinking` without resorting to `extra_body`.
- `LatestCerebrasModelNames` should reflect the currently-available GLM model name.

### Workaround I'm currently running

Keeping `CerebrasModel` + `CerebrasProvider`, but overriding the profile and settings:

```python

CerebrasModel(
    'zai-glm-4.7',
    provider=CerebrasProvider(...),
    profile=CerebrasProvider.model_profile('zai-glm-4.7').update(
        OpenAIModelProfile(openai_chat_send_back_thinking_parts='tags'),
    ),
    settings={'extra_body': {'clear_thinking': False}},
)
```

### Suggested direction (open to maintainer feedback)

Three separable changes:

1. **Bug fix:** have `CerebrasProvider.model_profile()` set `openai_chat_send_back_thinking_parts='tags'` for the reasoning-capable Cerebras families that the docs identify as using `<think>` tags (`zai`, `qwen`). `gpt-oss` follows Harmony rules and probably should keep its current handling.
2. **Track Cerebras' deprecation:** switch `_cerebras_settings_to_openai_settings` to emit `reasoning_effort='none'` instead of `extra_body['disable_reasoning']`. The existing `cerebras_disable_reasoning` setting can stay as a thin compat alias with a `DeprecationWarning`, or be replaced — I'd defer to a maintainer here.
3. **Expose `clear_thinking` as a first-class setting** (e.g. `cerebras_clear_thinking: bool` on `CerebrasModelSettings`, written into `extra_body['clear_thinking']`). The default value is the only thing I'd want feedback on: Cerebras defaults to `True` (strip prior thinking), but for pydantic-ai's multi-turn flows `False` is what users want most of the time.

Plus the trivial `LatestCerebrasModelNames` bump (add `zai-glm-4.7`, drop `zai-glm-4.6`).

Happy to send PRs for (1) and the model-name bump once direction is confirmed, and a separate PR for (2)+(3) once the API shape is decided.

### References

- Cerebras: Retain Reasoning Context for GLM/Qwen — https://inference-docs.cerebras.ai/capabilities/reasoning
- Cerebras: GLM-4.7 migration guide — https://inference-docs.cerebras.ai/resources/glm-47-migration

```Python
import os
import asyncio

from pydantic_ai import Agent
from pydantic_ai.models.cerebras import CerebrasModel
from pydantic_ai.providers.cerebras import CerebrasProvider


async def main() -> None:
    model = CerebrasModel(
        'zai-glm-4.7',
        provider=CerebrasProvider(api_key=os.environ['CEREBRAS_API_KEY']),
    )
    agent = Agent(model)

    result1 = await agent.run('What is 25 * 4?')
    result2 = await agent.run(
        'Now divide that by 2.',
        message_history=result1.all_messages(),
    )
    print(result2.output)


asyncio.run(main())


Observed: in the second turn the assistant message in the outgoing request does not carry the prior reasoning in `<think>...</think>` tags inside `content`, so the model loses its prior thinking. (Replace this line with the exact error / behavior you saw when you ran it — e.g. paste the 4xx body, or the printed output where the model claims it doesn't know the previous answer.)
```

### Logfire Trace

_No response_

- Python: 3.14.4
- Pydantic AI: 1.93.0
- LLM provider SDK: 
name = "cerebras-cloud-sdk"
version = "1.67.0"

kuishou68:
## Changes

- Update model name  → 
- Set  for zai models so thinking content is preserved in multi-turn conversations
- Replace deprecated  extra_body parameter with  (maps  to  for backward compatibility)
- Update tests and cassettes to match new behavior

Fixes #5606

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/cerebras.py
+++ b/pydantic_ai_slim/pydantic_ai/models/cerebras.py
@@ -54,9 +54,8 @@ class CerebrasModelSettings(ModelSettings, total=False):
 
     This setting is only supported on reasoning models: `zai-glm-4.7` and `gpt-oss-120b`.
 
-    .. deprecated::
-        Cerebras has deprecated the ``disable_reasoning`` parameter in favor of ``reasoning_effort``.
-        This setting is mapped to ``openai_reasoning_effort='none'`` when enabled.
+    **Deprecated:** Cerebras has deprecated the `disable_reasoning` parameter in favor of `reasoning_effort`.
+    This setting is mapped to `openai_reasoning_effort='none'` when enabled.
 
     See [the Cerebras docs](https://inference-docs.cerebras.ai/resources/openai#passing-non-standard-parameters) for more details.
     """
--- a/pydantic_ai_slim/pydantic_ai/models/cerebras.py
+++ b/pydantic_ai_slim/pydantic_ai/models/cerebras.py
@@ -129,15 +129,19 @@ def _cerebras_settings_to_openai_settings(
     Returns:
         An 'OpenAIChatModelSettings' object with equivalent settings.
     """
-    if (disable_reasoning := model_settings.pop('cerebras_disable_reasoning', None)) is not None:
+    # Build a plain dict so we can set openai_reasoning_effort (not in CerebrasModelSettings)
+    openai_settings: dict[str, Any] = dict(model_settings)
+    disable_reasoning = openai_settings.pop('cerebras_disable_reasoning', None)
+
+    if disable_reasoning is not None:
         # Map deprecated `disable_reasoning` to `reasoning_effort` for backward compatibility
         if disable_reasoning:
-            model_settings['openai_reasoning_effort'] = 'none'
-        else:
-            model_settings.setdefault('openai_reasoning_effort', 'low')
+            openai_settings['openai_reasoning_effort'] = 'none'
+        elif 'openai_reasoning_effort' not in openai_settings:
+            openai_settings['openai_reasoning_effort'] = 'low'
     elif model_request_parameters.thinking is False:
-        model_settings['openai_reasoning_effort'] = 'none'
+        openai_settings['openai_reasoning_effort'] = 'none'
     elif model_request_parameters.thinking:
-        model_settings.setdefault('openai_reasoning_effort', 'low')
+        openai_settings.setdefault('openai_reasoning_effort', 'low')
 
-    return OpenAIChatModelSettings(**model_settings)  # type: ignore[reportCallIssue]
+    return OpenAIChatModelSettings(**openai_settings)  # type: ignore[reportCallIssue]
--- a/pydantic_ai_slim/pydantic_ai/models/cerebras.py
+++ b/pydantic_ai_slim/pydantic_ai/models/cerebras.py
@@ -30,7 +30,7 @@ LatestCerebrasModelNames = Literal[
     'llama3.1-8b',
     'qwen-3-235b-a22b-instruct-2507',
     'qwen-3-32b',
-    'zai-glm-4.6',
+    'zai-glm-4.7',
 ]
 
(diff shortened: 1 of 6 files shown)


kuishou68:
Hi @kuishou68, the CI coverage check is failing because of an uncovered branch in `cerebras.py` (line 139→146).

The missing coverage is for the case where `disable_reasoning=False` AND `openai_reasoning_effort` is already set in the settings. In this scenario, the existing `openai_reasoning_effort` value should be preserved (the `elif 'openai_reasoning_effort' not in openai_settings` branch is not tested).

To fix this, add a test case like:
```python
# Test with disable_reasoning=False but openai_reasoning_effort already set → preserves existing value
settings_with_effort = CerebrasModelSettings(cerebras_disable_reasoning=False, openai_reasoning_effort='high')
transformed_with_effort = _cerebras_settings_to_openai_settings(settings_with_effort, params)
assert transformed_with_effort.get('openai_reasoning_effort') == 'high'
```

This should bring coverage back to 100% and fix both the `coverage` and `check` CI jobs.

dsfaccini:
David's AICA here: thanks for the PR! Closing this one because the description came through as an unfilled template — the placeholder lines rendered literally and the actual change/intent isn't described. If you'd like to pursue it, please open a fresh PR with a filled-in description and the concrete diff. Cheers!

Issue #3092, opened 2025-10-06
[Bedrock] `ValidationException`: "Thinking may not be enabled when `tool_choice` forces tool use" with interleaved thinking on Sonnet 4.5


sharon8811:
### Initial Checks

- [x] I confirm that I'm using the latest version of Pydantic AI
- [x] I confirm that I searched for my issue in https://github.com/pydantic/pydantic-ai/issues before opening this issue

### Description

# Bug Description

When using AWS Bedrock Claude Sonnet 4.5 with interleaved thinking enabled, the agent fails when attempting to use tools after the initial call. This appears to be a regression introduced between versions 1.0.1 and 1.0.15.

## Error Message

```
botocore.errorfactory.ValidationException: An error occurred (ValidationException) when calling the ConverseStream operation: The model returned the following errors: Thinking may not be enabled when tool_choice forces tool use.
```

## Configuration

**PydanticAI Version:** 1.0.15
**Model:** Claude Sonnet 4.5 via AWS Bedrock
**Provider:** `bedrock`

**Model Settings:**
```json
{
  "max_tokens": 63000,
  "temperature": 1.0,
  "timeout": 300,
  "parallel_tool_calls": true,
  "bedrock_additional_model_requests_fields": {
    "anthropic_beta": [
      "token-efficient-tools-2025-02-19",
      "interleaved-thinking-2025-05-14"
    ],
    "thinking": {
      "type": "enabled",
      "budget_tokens": 4096
    }
  }
}
```

## Steps to Reproduce

1. Configure agent with AWS Bedrock Sonnet 4.5 and thinking enabled (settings above)
2. Call agent with structured output: `agent.iter(prompt, output_type=MyOutputType)`
3. Agent responds to first call with the error

## Stack Trace

```python
File ".venv/lib/python3.12/site-packages/pydantic_ai/models/instrumented.py", line 381, in request_stream
  async with self.wrapped.request_stream(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/opt/homebrew/anaconda3/lib/python3.12/contextlib.py", line 210, in __aenter__
  return await anext(self.gen)
         ^^^^^^^^^^^^^^^^^^^^^
File ".venv/lib/python3.12/site-packages/pydantic_ai/models/bedrock.py", line 289, in request_stream
  response = await self._messages_create(messages, True, settings, model_request_parameters)
[...]

PR #5651, opened 2026-05-25
fix(bedrock): recognize `type='adaptive'` in `_is_thinking_enabled`

### LINKED ISSUE #5650: `_is_thinking_enabled` misses `type='adaptive'` in Bedrock additional fields
### Initial Checks

### Description

Follow-on from #5304 / #5326. The unified path now correctly translates to `{'type': 'adaptive'}` via `bedrock_supports_adaptive_thinking`, but `_is_thinking_enabled` was not updated to recognize the new variant:

```python
# pydantic_ai/models/bedrock.py
if (
    (additional_fields := model_settings.get('bedrock_additional_model_requests_fields'))
    and (thinking := additional_fields.get('thinking'))
    and thinking.get('type') == 'enabled'   # <-- misses 'adaptive'
):
    return True
```

Anyone passing a raw `adaptive` dict (the documented workaround for fields the unified API doesn't expose — `display`, `output_config`, …) is invisible to detection. Consequences:

- `prepare_request` doesn't switch `output_mode` away from `tool` → structured-output requests fail with the Bedrock "tool_choice + thinking" rejection.
- `_support_tool_forcing` doesn't downgrade forced tool_choice to `auto`.

The workaround is to redundantly set unified `thinking=True` alongside the raw dict, which contradicts the role of `_is_thinking_enabled` as the single source of truth.

### Proposed

```python
and thinking.get('type') in ('enabled', 'adaptive')
```

### References

- #5304 / #5326 — added the adaptive variant but did not update this detection helper

```Python

```

### Logfire Trace

_No response_

- Python: 3.12
- Pydantic AI: 1.102.0
- LLM provider SDK: bedrock

sevakva:
- Closes #5650

`_is_thinking_enabled` previously matched only `{'type': 'enabled'}` in `bedrock_additional_model_requests_fields`, missing the `'adaptive'` variant added in #5304 / #5326. When a user passed the raw adaptive dict (the documented escape hatch for fields the unified API doesn't expose, like `display` and `output_config`), detection silently failed and the downstream tool_choice + thinking guards never fired:

- `prepare_request` didn't switch `output_mode` away from `tool` → structured-output requests failed with Bedrock's "tool_choice + thinking" rejection.
- `_support_tool_forcing` didn't downgrade forced `tool_choice` to `auto`.

The forced workaround was redundantly setting unified `thinking=True` alongside the raw dict, which contradicts the role of `_is_thinking_enabled` as the single source of truth.

The Anthropic adapter already uses the `('enabled', 'adaptive')` membership pattern at `pydantic_ai/models/anthropic.py:593` and `:2728`; this aligns Bedrock with it.

The existing `test_bedrock_output_tool_with_thinking_raises` is parametrized over `enabled` / `adaptive` so the same regression is covered for both variants of #3092 and #5650.

### Checklist

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/bedrock.py
+++ b/pydantic_ai_slim/pydantic_ai/models/bedrock.py
@@ -1544,7 +1544,7 @@ def _is_thinking_enabled(
         if (
             (additional_fields := model_settings.get('bedrock_additional_model_requests_fields'))
             and (thinking := additional_fields.get('thinking'))
-            and thinking.get('type') == 'enabled'
+            and thinking.get('type') in ('enabled', 'adaptive')
         ):
             return True
     return False
--- a/tests/models/test_bedrock.py
+++ b/tests/models/test_bedrock.py
@@ -2313,19 +2313,27 @@ Mexico City is an important cultural, financial, and political center for the co
     )
 
 
-async def test_bedrock_output_tool_with_thinking_raises(allow_model_requests: None, bedrock_provider: BedrockProvider):
+@pytest.mark.parametrize(
+    'thinking_field',
+    [
+        pytest.param({'type': 'enabled', 'budget_tokens': 1024}, id='enabled'),
+        pytest.param({'type': 'adaptive'}, id='adaptive'),
+    ],
+)
+async def test_bedrock_output_tool_with_thinking_raises(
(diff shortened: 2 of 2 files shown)


dsfaccini:
this is beautiful, thank you

PR #5654, opened 2026-05-25
fix(tool-search): drop redundant `description` from `ToolSearchMatch`

### LINKED ISSUE #5653: `ToolSearchMatch.description` is redundant - duplicates tools-schema content
### Initial Checks

### Description

https://github.com/pydantic/pydantic-ai/blob/main/pydantic_ai_slim/pydantic_ai/_tool_search.py#L65-L72
```python
class ToolSearchMatch(TypedDict):
    """A single match in a tool-search result."""

    name: str
    """Name of the discovered tool, as the model will call it."""

    description: str | None
    """Human-readable description, if the tool provided one."""
```

When the model calls `search_tools`, the matched tools are marked discovered and pydantic-ai injects their **full `ToolDefinition`** (name + description + parameter schema) into the next LLM call's `tools` array. Returning the description **inside the search result** sends the same string a second time — once as a tool_result payload, once in the tools schema — and the model sees both.

For an agent with ~3-5k-char docstrings × 5-10 matches per search, the search result alone is 30-40k chars. That blob stays in context for every subsequent LLM call in the turn, paid at input-token rates (only the cached portion at 0.1×).

### Internal measurements (agent run, 3 baseline + 3 patched, GPT-5.4)

| | Baseline (`{name, description}`) | Patched (`{name}` only) |
|---|---|---|
| Input tokens / 3 runs | 355,615 | **272,027 (-23.5%)** |
| LLM requests | 19 | **19 (unchanged)** |
| Cache hit ratio | 84.6% | 91.3% |

No behavioral change in tool selection — the model picks correctly from the schema delivered on the next call.

### Suggested fix

Either:
1. Drop `description` from `ToolSearchMatch` entirely — it's structurally redundant.
2. Make it opt-in (e.g. `ToolSearchToolset(include_descriptions=False)`), preserving backward compat.

Happy to send a PR if there's interest.

```Python

```

### Logfire Trace

_No response_

- Python: 3.13.2
- Pydantic AI: 1.78.0 (pydantic-ai-slim)
- LLM provider SDK: 2.30.0 (openai)

serozhenka:
Fixes #5653.

## Why

`ToolSearchMatch.description` duplicates the description that pydantic-ai already injects on the next LLM call.

When the model calls `search_tools`, the framework flips `defer_loading=False` for the matched tools so their full `ToolDefinition` (name + description + parameter schema) lands in the next request's `tools` array. Including `description` in the search-result payload too sends the same string a second time — once as the tool-result content, once in the tools schema — and the model sees both. The cost is paid in input tokens on every subsequent call in the turn.

In an internal measurement on an agent with ~3-5k-char tool docstrings × 5-10 matches per search, dropping the field cut input tokens by **-23.5% (355k → 272k over 3 runs)** with no behavioral change in tool selection.

## What

- Remove `description: str | None` from the `ToolSearchMatch` `TypedDict` and update its docstring.
- Drop `'description': ...` from the three sites that construct matches:
  - `toolsets/_tool_search.py` — keyword and user-strategy paths.
  - `models/openai.py` — OpenAI Responses native return mapper.
  - `models/anthropic.py` — Anthropic tool-search result mapper (was always `None` anyway, since Anthropic's `tool_reference` blocks don't carry descriptions).
- Update snapshot tests in `tests/test_tool_search.py` to match.

Internal scoring against tool descriptions (`_search_terms` reading `tool_def.description` to tokenize) is **unchanged** — that path is not model-facing.

## Test plan

- [x] `uv run pytest tests/test_tool_search.py` — 136 passed.
- [x] `uv run ruff check` / `ruff format` on touched files.


pydantic_ai_slim/pydantic_ai/_tool_search.py
```diff
@@ -66,10 +66,12 @@ class ToolSearchMatch(TypedDict):
     """A single match in a tool-search result."""
 
     name: str
-    """Name of the discovered tool, as the model will call it."""
+    """Name of the discovered tool, as the model will call it.
 
-    description: str | None
-    """Human-readable description, if the tool provided one."""
```
dsfaccini, line 72: placeholder

PR #5694, opened 2026-05-27
Fix `MCPToolset(http_client=...)` crash on FastMCP's `follow_redirects` kwarg

### LINKED ISSUE #5688: `MCPToolset(url, http_client=...)` crashes: factory missing `follow_redirects` kwarg passed by `FastMCP`
AI notice: this bug was spotted and this report was generated by ANthropic's Claude Opus 4.7

### Initial Checks

### Description

Constructing an HTTP `MCPToolset` with an explicit `http_client=` fails at connect
time with a `TypeError` about an unexpected `follow_redirects` keyword argument.
The URL-only path (`MCPToolset('http://...')`) is unaffected.

### Root cause

When `http_client=` is passed, `MCPToolset` adapts it to a factory:

- `pydantic_ai/mcp.py` (~L2322): `factory = _make_httpx_client_factory(http_client)`,
  passed as `httpx_client_factory=` to the FastMCP transport.
- `_make_httpx_client_factory` (~L2344) returns a closure with signature
  `factory(headers=None, timeout=None, auth=None)`. This matches the mcp SDK's
  `McpHttpClientFactory` protocol (`mcp/shared/_httpx_utils.py`), which declares
  exactly `headers, timeout, auth`.

But the actual caller is **FastMCP**, not the mcp SDK directly. In
`fastmcp/client/transports/http.py` (~L176, fastmcp 3.3.1) it calls:

```python
http_client = self.httpx_client_factory(
    headers=headers,
    auth=self.auth,
    follow_redirects=True,  # type: ignore[call-arg]  # ty:ignore[unknown-argument]
    **({"timeout": timeout} if timeout else {}),
)
```

i.e. FastMCP passes `follow_redirects=` (intentionally — note its own
`# type: ignore[call-arg]`), which is beyond the `McpHttpClientFactory` protocol the
pydantic-ai closure was written against. The closure has no `**kwargs`, so it raises.

### Suggested fix

Make the closure tolerant of extra kwargs the transport may pass (its return type is
already `Callable[..., httpx.AsyncClient]`):

```python
def factory(
    headers: dict[str, str] | None = None,
    timeout: httpx.Timeout | None = None,
    auth: httpx.Auth | None = None,
    **_kwargs: Any,
) -> httpx.AsyncClient:
    return http_client
```

This is robust regardless of whether the module is on `httpx` or `httpx2`.

### Notes

- Related but not a fix: #5664 (`prefer httpx2 in pydantic_ai.mcp`) touches the same
  module but does not modify `_make_httpx_client_factory`.
- Arguably FastMCP is calling factories outside the documented `McpHttpClientFactory`
  protocol; but since `MCPToolset` is built on FastMCP's client, accepting `**kwargs`
  here is the pragmatic and forward-compatible fix.

No live server required — it fails while building the httpx client, before any request:

```python
import asyncio
import httpx
from pydantic_ai.mcp import MCPToolset

async def main():
    toolset = MCPToolset("http://localhost:9999/mcp", http_client=httpx.AsyncClient())
    async with toolset:
        await toolset.list_tools()

asyncio.run(main())
```

### Actual behavior

```
RuntimeError: Client failed to connect: _make_httpx_client_factory.<locals>.factory()
got an unexpected keyword argument 'follow_redirects'
```

### Expected behavior

The user-supplied `http_client` is used for the connection (it works on the
`MCPServerStreamableHTTP(url, http_client=...)` deprecated path).

- pydantic-ai 1.102.0
- fastmcp 3.3.1
- httpx 0.28.1
- Python 3.14.3

he-yufeng:
## Summary

`MCPToolset(url, http_client=...)` adapts the supplied `httpx.AsyncClient` into FastMCP's `httpx_client_factory` hook. FastMCP's HTTP transports can call that factory with extra runtime kwargs such as `follow_redirects`, so the adapter needs to accept and ignore them while still returning the caller-owned client.

This keeps the existing URL/SSE behavior and only makes the factory tolerant of the extra transport kwargs.

Fixes #5688.

## To verify

- `python -m pytest tests/test_mcp_toolset.py::TestMCPToolsetConstruction -q -p no:cacheprovider --basetemp .tmp/pytest-5688-construction2`
- `ruff format --check pydantic_ai_slim/pydantic_ai/mcp.py tests/test_mcp_toolset.py`
- `ruff check pydantic_ai_slim/pydantic_ai/mcp.py tests/test_mcp_toolset.py`
- `pyright pydantic_ai_slim/pydantic_ai/mcp.py`
- `python -m py_compile pydantic_ai_slim/pydantic_ai/mcp.py tests/test_mcp_toolset.py`
- `git diff --check`

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/mcp.py
+++ b/pydantic_ai_slim/pydantic_ai/mcp.py
@@ -127,7 +127,6 @@ __all__ = (
     'ServerCapabilities',
     'ProcessToolCallback',
     'CallToolFunc',
-    'MCPHTTPClientFactory',
     'ToolResult',
     'Prompt',
     'PromptArgument',
@@ -1942,24 +1941,6 @@ class CallToolFunc(Protocol):
     ) -> ToolResult: ...
 
 
-class MCPHTTPClientFactory(Protocol):
-    """Factory returning the `httpx.AsyncClient` an MCP HTTP transport should use.
-
-    FastMCP's HTTP transports call this with `follow_redirects` and may pass further
-    kwargs in future versions, so the contract names the known arguments and keeps a
-    `**kwargs` catch-all for forward compatibility.
-    """
-
-    def __call__(
-        self,
-        headers: dict[str, str] | None = None,
-        timeout: httpx.Timeout | None = None,
-        auth: httpx.Auth | None = None,
-        follow_redirects: bool = True,
-        **kwargs: object,
-    ) -> httpx.AsyncClient: ...
-
-
 ProcessToolCallback = Callable[
     [
         RunContext[Any],
@@ -2776,15 +2757,16 @@ def _build_transport(
 
 def _make_httpx_client_factory(
     http_client: httpx.AsyncClient,
-) -> MCPHTTPClientFactory:
+) -> Callable[..., httpx.AsyncClient]:
     """Return an `httpx_client_factory` that always returns the user-supplied `http_client`."""
 
     def factory(
         headers: dict[str, str] | None = None,
         timeout: httpx.Timeout | None = None,
         auth: httpx.Auth | None = None,
+        # FastMCP's StreamableHttpTransport calls the factory with `follow_redirects`,
+        # which the mcp SDK's `McpHttpClientFactory` protocol doesn't declare.
         follow_redirects: bool = True,
-        **_kwargs: object,
     ) -> httpx.AsyncClient:
         return http_client
 
--- a/pydantic_ai_slim/pydantic_ai/mcp.py
+++ b/pydantic_ai_slim/pydantic_ai/mcp.py
@@ -127,6 +127,7 @@ __all__ = (
     'ServerCapabilities',
     'ProcessToolCallback',
     'CallToolFunc',
+    'MCPHTTPClientFactory',
     'ToolResult',
     'Prompt',
     'PromptArgument',
@@ -1797,14 +1798,7 @@ class MCPServerSSE(_MCPServerHTTP):
             raise ValueError('`http_client` is mutually exclusive with `headers`.')
 
         if self.http_client is not None:
-
-            def httpx_client_factory(
-                headers: dict[str, str] | None = None,
-                timeout: httpx.Timeout | None = None,
-                auth: httpx.Auth | None = None,
-            ) -> httpx.AsyncClient:
-                assert self.http_client is not None
-                return self.http_client
+            httpx_client_factory = _make_httpx_client_factory(self.http_client)
 
             async with sse_client(
                 url=self.url,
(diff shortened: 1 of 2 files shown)


CostcoFanboy:
@DouweM deepest apologies for the tag I'm aware of how incredibly busy you are.
Do you think we could merge this in? 

This small bug that unlocks quite a lot of features.

dsfaccini:

| Wave | Status | Blocking | Warnings |
|---|---|---|---|
| 1 · code-correctness | ✅ closed | 0 | 0 |
| 2 · tests/cassettes/docs | ✅ closed | 0 | 0 |
| 3 · cross-cutting/siblings | ✅ closed | 0 | 0 |

**Verdict:** ✅ all 3 waves passed, 0 blocking
Reviewed at 2026-06-11T21:07:21Z · diff `56856073c571`

dsfaccini (changes requested):
we generally don't like and don't use `kwargs: Any` (precisely bc they are untyped). let me check if I can find a better approach for this (maybe one field that holds all the possbile http factory kwargs)

don't push any commits in the meantime please @he-yufeng

dsfaccini:
David's AICA here: thanks for this. We talked the typing over and want to take it one step past `**kwargs` before merging, so here's the agreed direction.

Why not just `**_kwargs: Any` (or `**_kwargs: object`): the reason `follow_redirects` has to be accepted at all is that FastMCP calls the factory with kwargs beyond the MCP SDK's `McpHttpClientFactory` protocol (which only declares `headers`, `timeout`, `auth`), and FastMCP's own docs recommend a `**kwargs` catch-all for forward compatibility. A bare `**_kwargs: Any` works but we avoid untyped `Any`; a bare `**_kwargs: object` would typecheck too, but David would rather we name the contract with a predefined type than leave an inline catch-all.

So please introduce a named `Protocol` for pyai's MCP http-client factory and type both factory sites against it. Proposed shape, placed near the existing `CallToolFunc` protocol in `mcp.py`:

```python
class MCPHTTPClientFactory(Protocol):
    """Factory returning the `httpx.AsyncClient` an MCP HTTP transport should use.

    FastMCP's HTTP transports call this with `follow_redirects` and may pass further
    kwargs in future versions, so the contract names the known arguments and keeps a
    `**kwargs` catch-all for forward compatibility.
    """

    def __call__(
        self,
        headers: dict[str, str] | None = None,
        timeout: httpx.Timeout | None = None,
        auth: httpx.Auth | None = None,
        follow_redirects: bool = True,
        **kwargs: object,
    ) -> httpx.AsyncClient: ...
```

The name is `MCPHTTPClientFactory` (pyai's all-caps `MCP`/`HTTP` style, as in `MCPServerHTTP` / `ModelHTTPError`), deliberately distinct from the SDK's `McpHttpClientFactory` so the two don't collide. It names the one kwarg #5688 needs (`follow_redirects`) plus the existing three, and keeps `**kwargs: object` for whatever FastMCP adds next.

Apply it at both factory sites:

1. `_make_httpx_client_factory` (the FastMCP path): annotate it `-> MCPHTTPClientFactory` and give the inner `factory` the full signature above (`follow_redirects: bool = True`, `**_kwargs: object`). Body stays `return http_client`.
2. The deprecated `MCPServerStreamableHTTP` SSE path currently inlines its own three-arg `httpx_client_factory`. Simplest is to drop that inline def and reuse the helper: `httpx_client_factory = _make_httpx_client_factory(self.http_client)`. That gives both sites one named, documented contract and removes the duplicate. (If you'd rather not restructure that path, giving its inner def the same full signature and annotating the local as `MCPHTTPClientFactory` also typechecks.)

A value satisfying `MCPHTTPClientFactory` stays assignable to the SDK's narrower `McpHttpClientFactory` slot, so both the FastMCP transport and the SDK's `sse_client` accept it. I confirmed the whole shape is pyright-clean at both sites.

One test cleanup while you're in there: the two assertions can drop `factory: Any` by calling the helper directly, since it now returns the typed factory:

```python
factory = _make_httpx_client_factory(client)
assert factory(follow_redirects=True) is client
```

Could you implement against this shape? Appreciate you holding off until it's in. Thanks!

dsfaccini (approved):

PR #5703, opened 2026-05-28
fix(anthropic): strip server-side tools from `count_tokens` request

### LINKED ISSUE #5702: Anthropic `count_tokens` fails when native `CodeExecutionTool` is enabled
### Initial Checks

- [x] I searched the existing issues and discussions

### Description

`AnthropicModel.count_tokens()` fails when `ModelRequestParameters` contains an Anthropic server-side/native tool such as `CodeExecutionTool`.

In current Pydantic AI this is exposed via `native_tools`:

```python
ModelRequestParameters(native_tools=[CodeExecutionTool()])
```

The legacy `builtin_tools` field shows the same behavior on older versions.

The current Anthropic token counting path prepares regular tools, calls `_add_builtin_tools(...)`, and then sends the resulting `tools` list to `client.beta.messages.count_tokens(...)`:

```python
tools, tool_choice = self._prepare_tools_and_tool_choice(model_settings, model_request_parameters)
tools, mcp_servers, builtin_tool_betas = self._add_builtin_tools(
    tools, model_request_parameters, model_settings
)
...
return await self.client.beta.messages.count_tokens(
    ...
    tools=tools or OMIT,
    ...
)
```

When code execution is enabled this attaches an Anthropic server tool, for example:

```python
{"name": "code_execution", "type": "code_execution_20250825"}
```

Anthropic rejects server tools on the `count_tokens` endpoint. In our app this surfaced as:

```text
ModelHTTPError: status_code: 400, model_name: claude-opus-4-6, body: {
  "type": "error",
  "error": {
    "message": "Server tools are not supported in the count_tokens endpoint: code_execution_20260120. Use the /v1/messages endpoint instead."
  }
}
```

This means any capability/integration that calls `count_tokens()` before a run can fail even though the actual `/v1/messages` request would support the server-side tool.

### Example Code

```python
import asyncio

from pydantic_ai import CodeExecutionTool


async def main() -> None:
    model = AnthropicModel("claude-opus-4-1-20250805")
    await model.count_tokens(
        messages=[ModelRequest(parts=[UserPromptPart(content="hello")])],
        model_settings={},
        model_request_parameters=ModelRequestParameters(
            native_tools=[CodeExecutionTool()],
        ),
    )


asyncio.run(main())
```

### Local Verification Without Hitting Anthropic

Using a fake Anthropic client with `pydantic-ai-slim==1.103.0`, `AnthropicModel.count_tokens()` still sends this to `beta.messages.count_tokens`:

```python
tools = [{"name": "code_execution", "type": "code_execution_20250825"}]
```

The same payload is produced on `pydantic-ai-slim==1.93.0` when using the older `builtin_tools=[CodeExecutionTool()]` API.

### Expected Behavior

`count_tokens()` should not fail merely because server-side/native tools are enabled for the eventual model request.

Possible behaviors that would work:

- omit Anthropic server-side/native tools from Anthropic `count_tokens` calls, or
- provide a fallback/clear handling path when Anthropic cannot count tokens for requests with server tools.

The important part is that attaching `CodeExecutionTool` for generation should not make a preflight token-count request fail.

### Affected Components

- Anthropic model provider
- Native/server-side tools
- Token counting

### Versions

Reproduced with:

- `pydantic-ai-slim==1.103.0` using `native_tools=[CodeExecutionTool()]`
- `pydantic-ai-slim==1.93.0` using legacy `builtin_tools=[CodeExecutionTool()]`

devteamaegis:
## Summary

`AnthropicModel.count_tokens()` fails with HTTP 400 when any server-side native tool
(e.g. `CodeExecutionTool`, `WebSearchTool`, `WebFetchTool`) is enabled:

```
Server tools are not supported in the count_tokens endpoint: code_execution_20260120.
Use the /v1/messages endpoint instead.
```

The root cause: `_messages_count_tokens` calls `_add_native_tools` — which appends
server-tool entries to the `tools` list — then passes that full list to
`client.beta.messages.count_tokens`. The actual `/v1/messages` endpoint accepts
those tools fine; only the token-counting endpoint rejects them.

## Fix

After `_add_native_tools`, filter `tools` down to standard function tools
(`type == "custom"` or absent) before the `count_tokens` call.
The real `messages.create` / `messages.stream` paths are untouched.

```python
# The Anthropic count_tokens endpoint does not accept server-side native tools.
# Only keep standard function tools (type == "custom" or absent).
tools = [t for t in tools if t.get('type') in {None, 'custom'}]
```

## Test

Added `test_count_tokens_strips_server_side_native_tools` which:
1. Creates an agent with `CodeExecutionTool` enabled (a server-side tool).
2. Runs it with `count_tokens_before_request=True` against `MockAnthropic`.
3. Asserts that no tool with a non-"custom" `type` appears in the kwargs
   recorded for the `count_tokens` call.

Closes #5702

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/anthropic.py
+++ b/pydantic_ai_slim/pydantic_ai/models/anthropic.py
@@ -854,6 +854,14 @@ class AnthropicModel(Model[AsyncAnthropicClient]):
         tools, tool_choice = self._prepare_tools_and_tool_choice(model_settings, model_request_parameters)
         tools, mcp_servers, native_tool_betas = self._add_native_tools(tools, model_request_parameters, model_settings)
 
+        # The Anthropic count_tokens endpoint does not accept server-side native tools
+        # (e.g. code_execution, web_search, web_fetch, tool_search, memory).  Only
+        # standard function tools — whose type field is "custom" or absent — are
+        # supported.  Strip any server tool entries so the API call succeeds; the
+        # token estimate will still reflect all prompt content and any function tools.
+        # See https://github.com/pydantic/pydantic-ai/issues/5702
+        tools = [t for t in tools if t.get('type') in {None, 'custom'}]
+
         auto_cache_control, resolved_cache_ttl = self._build_automatic_cache_control(model_settings)
         system_prompt, anthropic_messages = await self._map_message(messages, model_request_parameters, model_settings)
         self._apply_per_block_caching_fallback(resolved_cache_ttl, anthropic_messages)
--- a/tests/models/test_anthropic.py
+++ b/tests/models/test_anthropic.py
@@ -68,6 +68,7 @@ from pydantic_ai.settings import ModelSettings
 from pydantic_ai.usage import RequestUsage, UsageLimits
 
 from .._inline_snapshot import snapshot
+from ..cassette_utils import single_request_body
 from ..conftest import IsDatetime, IsInstance, IsNow, IsStr, TestEnv, raise_if_exception, try_import
 from ..parts_from_messages import part_types_from_messages
 from .mock_async_stream import MockAsyncStream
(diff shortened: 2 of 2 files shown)


kazmer97:
I am pretty sure this is AI spam, you're blocking the legit solution. either fix the pr or close it

PR #5704, opened 2026-05-28
Fix `AnthropicModel.count_tokens` with native tools

### LINKED ISSUE #5702: Anthropic `count_tokens` fails when native `CodeExecutionTool` is enabled

kazmer97:
- Closes #5702

### Summary

- Omit Anthropic native/server tools from `count_tokens` requests by clearing `native_tools` on the count-only request parameters.
- Preserve native tools for normal message creation.
- Add a regression test covering `CodeExecutionTool` and `WebFetchTool` during preflight token counting.

### Validation

- `uv run pytest tests/models/test_anthropic.py::test_anthropic_count_tokens_with_mock tests/models/test_anthropic.py::test_anthropic_count_tokens_with_no_messages tests/models/test_anthropic.py::test_anthropic_count_tokens_omits_native_tools tests/models/test_anthropic.py::test_anthropic_code_execution_tool_version_auto tests/models/test_anthropic.py::test_anthropic_code_execution_tool_version_setting`
- `uv run ruff check pydantic_ai_slim/pydantic_ai/models/anthropic.py tests/models/test_anthropic.py`
- `uv run ruff format --check pydantic_ai_slim/pydantic_ai/models/anthropic.py tests/models/test_anthropic.py`
- `PYRIGHT_PYTHON_IGNORE_WARNINGS=1 uv run pyright pydantic_ai_slim/pydantic_ai/models/anthropic.py`

### Checklist


dsfaccini:
sorry about this @kazmer97 , sometimes issues/PRs get buried under spam, feel free to ping us directly on slack when this happens so we can act more quickly

let me do a review of the PR and get back to you, I may push updates if they're needed

kazmer97:
> sorry about this @kazmer97 , sometimes issues/PRs get buried under spam, feel free to ping us directly on slack when this happens so we can act more quickly
> 
> let me do a review of the PR and get back to you, I may push updates if they're needed

Thanks for looking into it, let me see if I can setup the slack connection

dsfaccini:
hey @kazmer97 can you enable maintainers to push to your fork please?

kazmer97:
> hey @kazmer97 can you enable maintainers to push to your fork please?

Should be done now, thanks

dsfaccini:
@kazmer97 I still cant push apparently :(

kazmer97:
> @kazmer97 I still cant push apparently :(

sorry about that, try now?

kazmer97:
> > @kazmer97 I still cant push apparently :(
> 
> sorry about that, try now?
Screenshot 2026-06-06 at 11 13 21

assuming this is the tick that is needed, doing it from mobile yesterday didn't seem to work

dsfaccini (approved):

dsfaccini:
hey @adtyavrdhn can you take a look here? let me know if you can think of a better approach, otherwise we can merge and release this tomorrow

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -850,6 +850,10 @@ async def _messages_count_tokens(
         if isinstance(self.client, AsyncAnthropicBedrock):
             raise UserError('AsyncAnthropicBedrock client does not support `count_tokens` api.')
 
+        # Anthropic docs: https://platform.claude.com/docs/en/api/messages-count-tokens
+        # TODO: Remove this workaround if Anthropic starts accepting native/server tools on `count_tokens`.
+        model_request_parameters = replace(model_request_parameters, native_tools=[])
```
adtyavrdhn: It looks like `MemoryTool` is not rejected by the endpoint. It would be better to keep a tool which isn't rejected so we don't undercount for no reason.

Also worth a comment that we're going to be undercounting this because the API does not allow it. 

@dsfaccini

adtyavrdhn (approved):
Approving assuming David takes a call on `MemoryTool`.

pydantic_ai_slim/pydantic_ai/models/anthropic.py
dsfaccini: makes sense! thank you.

dsfaccini: David's AICA here: confirmed via a live probe against `count_tokens` (claude-sonnet-4-5) that `MemoryTool` is the only native tool the endpoint accepts. The server tools are rejected with a 400 — `web_search`/`code_execution`/`web_fetch`/`tool_search` all return "Server tools are not supported in the count_tokens endpoint", and `MCPServerTool` returns "The mcp_servers parameter is not supported on the count_tokens endpoint". `MemoryTool` is client-side and accepted; in the probe its definition added ~1565 tokens (1581 vs 16 baseline), so clearing it undercounts meaningfully.

Switched the omission to an allow-list that keeps `MemoryTool` (so we no longer undercount it) while still omitting the rejected server tools, with a comment noting the remaining undercount is deliberate until Anthropic supports them. Used an allow-list rather than a deny-list so a future tool defaults to a mild undercount rather than a hard 400. Added a live VCR regression test asserting the `count_tokens` request still carries the memory tool. Thanks for catching this!

PR #5680, opened 2026-05-27
Preserve `TextContent` metadata in Vercel AI adapter round-trips

serejaris:
Closes #5679

### Summary

- Preserve `TextContent.metadata` when Vercel AI user text parts round-trip through `dump_messages()` / `load_messages()`, carrying it on `TextUIPart.provider_metadata` under the established `pydantic_ai` wrapper (`dump_provider_metadata` / `load_provider_metadata`, matching the `vendor_metadata` precedent from #5571 / #5772). Plain strings stay plain strings.
- Add regression coverage for mixed plain `str` + `TextContent` user content, asserting a full dump→load round-trip.

### Scope note

This PR ships the **Vercel AI** adapter fix only, which fully closes #5679 (the issue is Vercel-specific).

The **AG-UI** half is intentionally deferred. `TextInputContent` has no protocol-defined metadata slot, so carrying metadata there means riding a loose `model_extra` extra field — which a spec-compliant (zod/TS) AG-UI client strips on parse, so it would not survive a real client-mediated round-trip the way Vercel's first-class `providerMetadata` does. AG-UI round-trip metadata is better handled by the reserved-carrier mechanism being worked out in #6127, and is left to that work.

### Checklist

- [ ] Any **AI generated code** has been reviewed line-by-line by the human PR author, who stands by it.
- [x] No **breaking changes** in accordance with the [version policy](https://github.com/pydantic/pydantic-ai/blob/main/docs/version-policy.md).
- [x] **PR title** is fit for the [release changelog](https://github.com/pydantic/pydantic-ai/releases).

PR #5710, opened 2026-05-28
fix: preserve Vercel text content metadata

he-yufeng:
﻿## Summary
- preserves `TextContent.metadata` when dumping user prompt text into Vercel `TextUIPart.provider_metadata`
- restores that metadata back to `TextContent` when loading Vercel UI messages
- adds a focused round-trip regression test

This addresses the Vercel AI adapter path described in #5679. I left the AG-UI path alone here because its text input content type does not appear to have an equivalent metadata field to round-trip through.

## To verify
- `PYTHONUTF8=1 uv run ruff check pydantic_ai_slim\pydantic_ai\ui\vercel_ai\_adapter.py tests\test_vercel_ai.py`
- `PYTHONUTF8=1 uv run pytest tests\test_vercel_ai.py -k "text_content_metadata_roundtrip or convert_user_prompt_part_text_content" -q`
- `PYTHONUTF8=1 uv run pyright pydantic_ai_slim\pydantic_ai\ui\vercel_ai\_adapter.py tests\test_vercel_ai.py`

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
+++ b/pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py
@@ -272,7 +272,13 @@ class VercelAIAdapter(UIAdapter[RequestData, UIMessage, BaseChunk, AgentDepsT, O
                 user_prompt_content: str | list[UserContent] = []
                 for part in msg.parts:
                     if isinstance(part, TextUIPart):
-                        user_prompt_content.append(part.text)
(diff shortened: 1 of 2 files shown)


dsfaccini:
David's AICA here: closing as superseded by https://github.com/pydantic/pydantic-ai/pull/5680, which makes the same `TextContent` metadata round-trip through the Vercel adapter and also covers the AG-UI adapter, with a more specific `provider_metadata` key. If #5680 stalls we can revisit.

PR #5734, opened 2026-05-30
Document Tuning Engines as an OpenAI-compatible endpoint

cerebrixos:
## Summary

- Adds a small documentation example showing how to point this project’s existing OpenAI-compatible configuration at Tuning Engines.
- Keeps this project’s APIs and runtime behavior unchanged: no new dependency, adapter, or code path.
- Shows a path for teams to keep this framework in charge of app, agent, tool, workflow, or retrieval logic while routing model calls through a governed OpenAI-compatible endpoint.

## Why this belongs here

Tuning Engines is an AI control plane for teams that want centralized model access, routing, tenant-scoped keys, policy/guardrail checks, audit logs, traces, approvals, and usage/cost visibility around existing AI applications.

This project already supports OpenAI-compatible endpoints. The docs change makes that existing capability discoverable for users who want governance and observability without rewriting their application around a separate SDK or changing this project’s runtime behavior.

## Testing

- `git diff --check`

### DIFF
--- a/docs/models/openai.md
+++ b/docs/models/openai.md
@@ -345,6 +345,21 @@ agent = Agent(model)
 ...
 ```
 
+For example, a governed OpenAI-compatible endpoint such as [Tuning Engines](https://www.tuningengines.com/) can be configured without changing your agent code:
+
+```python
+model = OpenAIChatModel(
+    'gpt-4o-mini',
+    provider=OpenAIProvider(
+        base_url='https://api.tuningengines.com/v1',
+        api_key='your-tuning-engines-key',
+    ),
+)
+agent = Agent(model)
+```
+
+This lets Pydantic AI continue to handle agent logic and tool execution while the gateway centralizes model routing, policy controls, audit logs, traces, approvals, and cost visibility.
+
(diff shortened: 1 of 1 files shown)


cerebrixos:
Thank you for taking a look at this docs-only contribution. I checked the current CI results: docs, lint, and mypy pass. The remaining Python 3.14 failures are outside the changed file (`docs/models/openai.md`), including `tests/test_streaming.py::test_run_stream_events_standalone_deprecation`. Would a maintainer please rerun the affected test jobs when convenient? I am happy to adjust the example if a docs-specific issue surfaces.

hramezani:
Thanks for the contribution, and for keeping it docs-only!

We're going to pass on this one. We generally list third-party OpenAI-compatible providers once they have some demonstrated community use, and Tuning Engines doesn't have that traction yet. Beyond that, the example duplicates the generic `OpenAIChatModel` + `OpenAIProvider(base_url=..., api_key=...)` pattern that's already documented in the section directly above it — so any user wanting to point Pydantic AI at a governed endpoint can already do so with the existing docs.

Appreciate you taking the time, and best of luck with the project! If it picks up adoption down the line, feel free to revisit.

PR #5736, opened 2026-05-30
fix: pass provider modules through Temporal sandbox

### LINKED ISSUE #5732: Anthropic models fail in Temporal workflows with `RestrictedWorkflowAccessError` on `Path.home`
Constructing an Anthropic model inside a Temporal workflow fails:

```
temporalio.worker.workflow_sandbox._restrictions.RestrictedWorkflowAccessError: Cannot access pathlib.Path.home.__call__ from inside a workflow.
```

It happens when the model is built lazily inside the workflow — a `gateway/anthropic:` or `anthropic:` model resolved via `infer_model` (dynamic model selection, or a model passed by name). Models built at module scope and passed into the agent are unaffected, which is why it doesn't show up in our existing Temporal tests.

**Root cause:** `anthropic>=0.99.0` reads `~/.config/anthropic` via `Path.home()` during `AsyncAnthropic.__init__` (credential/profile auto-discovery, added in 0.99.0 on 2026-05-05). The Temporal sandbox forbids `Path.home`, and `anthropic` is not in the `PydanticAIPlugin` sandbox passthrough list. Our floor is `anthropic>=0.105.0`, so every supported version is affected.

**Two adjacent problems on the same in-workflow construction path** (found while testing construction across providers):

- The gateway builds its own httpx client in the workflow, which imports `certifi` lazily → `Module certifi was imported after initial workflow load`.
- `google-genai` lazily imports `google.auth` submodules during client construction → same warning for `gateway/google-cloud:`.

These two are warnings by default (benign in production) but hard errors under our `filterwarnings=["error"]` test policy, and they produce log noise for users either way. Google documents the same passthrough requirement in their own Gemini + Temporal example.

**Fix:** add `anthropic`, `certifi`, and `google.auth` to the passthrough modules in `durable_exec/temporal/__init__.py`. Verified minimal: `google.auth` alone is enough (not the whole `google` namespace, not `google.genai`), and no other modules are needed for OpenAI/Groq/Gemini.

Repro: a `TemporalAgent` running a `gateway/anthropic:claude-sonnet-4-6` model resolved inside the workflow.

he-yufeng:
﻿Fixes #5732.

## Summary
- add `anthropic`, `certifi`, and `google.auth` to the Temporal sandbox passthrough list
- cover the plugin runner configuration with a small regression test

## To verify
- `PYTHONUTF8=1 uv run pytest tests/test_temporal.py::test_pydantic_ai_plugin_provider_passthrough_modules -q`
- `PYTHONUTF8=1 uv run ruff check pydantic_ai_slim/pydantic_ai/durable_exec/temporal/__init__.py tests/test_temporal.py`
- `PYTHONUTF8=1 uv run ruff format --check pydantic_ai_slim/pydantic_ai/durable_exec/temporal/__init__.py tests/test_temporal.py`
- `PYTHONUTF8=1 uv run python -m py_compile pydantic_ai_slim/pydantic_ai/durable_exec/temporal/__init__.py tests/test_temporal.py`
- `PYTHONUTF8=1 uv run pyright pydantic_ai_slim/pydantic_ai/durable_exec/temporal/__init__.py tests/test_temporal.py`
- `git diff --check`

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/durable_exec/temporal/__init__.py
+++ b/pydantic_ai_slim/pydantic_ai/durable_exec/temporal/__init__.py
@@ -80,6 +80,9 @@ def _workflow_runner(runner: WorkflowRunner | None) -> WorkflowRunner:
             'anyio',
             'sniffio',
             'httpcore',
+            'anthropic',
+            'certifi',
+            'google.auth',
             # `fastmcp` (and the `mcp` SDK it transitively imports) calls `Path.expanduser` at
             # import time when resolving its config directory — restricted by the workflow
             # sandbox. Safe to pass through: the call only happens once at module init.
--- a/tests/test_temporal.py
+++ b/tests/test_temporal.py
@@ -83,6 +83,7 @@ try:
     from temporalio.exceptions import ApplicationError
     from temporalio.testing import WorkflowEnvironment
     from temporalio.worker import Worker
+    from temporalio.worker.workflow_sandbox import SandboxedWorkflowRunner
     from temporalio.workflow import ActivityConfig
 
     from pydantic_ai.durable_exec.temporal import (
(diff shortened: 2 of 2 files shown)


he-yufeng:
Closing this because #5733 already covers the same Temporal passthrough fix. I missed that existing PR during the duplicate check, so keeping the maintainer branch as the active path.

Issue #5662, opened 2026-05-26
xAI: review support for newer SDK request and provider options


colesmcintosh:
### Description

The current xAI adapter exposes a useful subset of the SDK request surface, but the current `xai-sdk` includes several request/provider options that Pydantic AI does not account for.

SDK introspection shows `chat.create` now accepts options such as:

- `conversation_id`
- `seed`
- `search_parameters`
- `max_turns`
- `agent_count`
- `batch_request_id`
- include options such as verbose streaming and attachment-search call output

The provider constructor also supports deployment/transport options such as:

- `api_host`
- `management_api_host`
- `metadata`
- `channel_options`
- `timeout`
- `use_insecure_channel`

Today `XaiProvider` primarily accepts either an API key or a prebuilt SDK client. A prebuilt client is a valid escape hatch, but some of these options may be common enough to deserve first-class Pydantic AI support.

### Suggested scope

This should be selective, not a blanket mirror of every SDK argument.

Likely candidates to evaluate:

- `xai_seed` for reproducibility
- `xai_conversation_id` alongside `xai_store_messages` / `xai_previous_response_id`
- `xai_max_turns` for server-side native-tool loops
- `xai_agent_count` for multi-agent Grok variants, if those models are listed/supported
- include flags for verbose streaming and attachment-search output
- `api_host` and `timeout` on `XaiProvider` for deployment ergonomics

Lower-confidence candidates such as raw `search_parameters`, `batch_request_id`, or low-level channel options may be better left to a custom `xai_client` unless there is a concrete user need.

### References

- xAI models docs: https://docs.x.ai/developers/models
- Current provider: `pydantic_ai_slim/pydantic_ai/providers/xai.py`
- Current model settings/request mapping: `pydantic_ai_slim/pydantic_ai/models/xai.py`

adtyavrdhn:
Which ones according to you are most useful and requires least discussion and alignment?

NishchayMahor:
Took a pass at categorizing the candidates by how much alignment they would actually require, since the SDK introspection list is fairly broad. Posting this here so it can be picked apart before any code is written.

**Group A: ship-as-is, no design decisions**

[...]

PR #5741, opened 2026-05-31
Map base `seed` setting to xAI

colesmcintosh:
- Part of #5662 (Group A)

xAI's `seed` parameter is the same reproducibility concept already exposed by the base `ModelSettings.seed` field (used by OpenAI, Groq, Cohere, Mistral, and Gemini). Rather than introduce a redundant provider-prefixed `xai_seed` field, this maps the existing base `seed` straight through to the SDK's `seed` parameter and documents xAI as a supported provider — keeping the API consistent with every other provider that supports seeding.

### Checklist


colesmcintosh:
The failing `test on 3.x (lowest-versions)` checks are unrelated to this change. They're caused by `aiohttp 3.14.0` (published to PyPI on 2026-06-01) removing `aiohttp.streams.AsyncStreamReaderMixin`, which `vcrpy` still imports — so the `lowest-direct` resolver pulls the new aiohttp and the whole suite errors at collection. Tracked in #5751, with the fix in #5752 (`Pin aiohttp below 3.14 for vcrpy`).

This branch is already up to date with `main`, so once #5752 merges I'll re-merge `main` here to pick up the constraint and the lowest-versions matrix should go green. All other CI jobs pass.

Issue #2666, opened 2025-08-25
Tool function with `BaseModel` argument does not work with Temporal


DouweM:
### Initial Checks

- [x] I confirm that I'm using the latest version of Pydantic AI
- [x] I confirm that I searched for my issue in https://github.com/pydantic/pydantic-ai/issues before opening this issue

### Description

The object passed into the tool function is a `dict` instead of an instance of the annotated type:

```
  File "/home/DouweM/pydantic-ai/scratch/scratchpads/9c8b872dabe381702d009fe29745ee89/scratch9.py", line 21, in weather
    return f'The weather in {f.city} is sunny'
                             ^^^^^^
AttributeError: 'dict' object has no attribute 'city'
```

### Example Code

```Python
import asyncio
import uuid

from pydantic import BaseModel
from temporalio import workflow
from temporalio.client import Client
from temporalio.worker import Worker

from pydantic_ai import Agent
from pydantic_ai.durable_exec.temporal import AgentPlugin, PydanticAIPlugin, TemporalAgent

agent = Agent('openai:gpt-4.1', name='test-tool')


class WeatherInput(BaseModel):
    city: str


@agent.tool_plain
def weather(f: WeatherInput) -> str:
    return f'The weather in {f.city} is sunny'


temporal_agent = TemporalAgent(agent)


@workflow.defn
class PydanticWorkflow:
    @workflow.run
    async def run(self, prompt: str) -> str:
        result = await temporal_agent.run(prompt)
        return result.output


async def main():
    client = await Client.connect(
        'localhost:7233',
        plugins=[PydanticAIPlugin()],
    )

    async with Worker(
        client,
        task_queue='pydantic-ai-task-queue',
        workflows=[PydanticWorkflow],
        plugins=[AgentPlugin(temporal_agent)],
    ):
        result = await client.execute_workflow(
            PydanticWorkflow.run,
            args=['What is the weather in Tokyo?'],
            id=f'pydantic-hello-world-workflow-{uuid.uuid4()}',
            task_queue='pydantic-ai-task-queue',
        )
        print(f'Result: {result}')


if __name__ == '__main__':
    asyncio.run(main())
```

### Python, Pydantic AI & LLM client version

```Text
latest
```

PR #5758, opened 2026-06-02
examples: add signed tool receipts

tomjwxf:
## Summary

- Adds a PydanticAI example showing typed, hash-chained receipts for tool execution.
- Uses the Hooks capability (`before_tool_execute` and `after_tool_execute`) to emit pre/post receipts around each tool call.
- Defines the receipt and signature envelopes with Pydantic models, canonicalizes payloads, and verifies the chain locally.
- Keeps the example dependency-free with HMAC-SHA256; comments note that public third-party verification should use an asymmetric signature such as Ed25519.

## Verification

- `python3 -m py_compile examples/pydantic_ai_examples/signed_tool_receipts.py`
- `python3 -m ruff format examples/pydantic_ai_examples/signed_tool_receipts.py`
- `python3 -m ruff check examples/pydantic_ai_examples/signed_tool_receipts.py`
- `git diff --check`

Note: I could not run the full example module in this sandbox because the local checkout does not have the full workspace runtime dependencies installed (`logfire_api`, etc.).


arian-gogani:
the hook + signed-record pattern is the right shape, and the author's note about asymmetric signatures for public third-party verification is the load-bearing point worth promoting from the comments to the example's docstring.

a few additions that would make this composable with the broader ecosystem if a follow-up version lands:

1. **HMAC vs Ed25519 isn't just a key-management choice — it's a verifier-class choice.** HMAC requires the verifier to hold the same secret as the signer, so it only proves authenticity within an organization. Ed25519 lets a verifier check the signature with only the public key, so it survives the case the audit is *about* the organization. for an example positioned around tool execution receipts (regulatory framing is the natural use case), Ed25519 should probably be the default with HMAC as the in-trust-zone variant, rather than the other way around.

2. **Canonicalization is the silent failure mode.** `json.dumps(sort_keys=True, separators=(',', ':'))` is byte-identical to RFC 8785 JCS only for ASCII text + integers + booleans. The moment a Unicode tool name (`get_müşteri_data`), a non-integer confidence score, or a key requiring UTF-16 code-unit ordering shows up, the hashes diverge and the chain breaks silently. The `rfc8785` PyPI library handles this; worth importing it in the example so users who copy-paste don't hit this in production.

3. **`action_ref` derivation timing.** if the receipt computes its content-derived identifier from the canonical bytes *before* execution, the same identifier serves as both the idempotency key on retries (same intent = same ref = detect duplicate) and the join key from admission → outcome. if it's computed *after*, the pre-commitment property is lost. minor change to the hook ordering; meaningful change to what the example teaches.

the shape you're sketching here is the same one that's been converging across vaara, agent-guard, x402 settlement-receipt extension, and nobulex over the last 90 days — see x402 PR #2666 §5 for the three-issuer conformance evidence. PydanticAI getting a typed-receipts example into the docs is exactly the right move for the framework's compliance-curious audience.

hramezani:
@
Thanks for the contribution, @tomjwxf — the hooks-based auditing pattern is a genuinely interesting thing to demonstrate, and it's clear you put thought into the receipt/hash-chain design.

That said, I'm going to close this one. A few reasons:

**It's incomplete as an example.** Every documented example in the repo ships with a matching `docs/examples/*.md` page and a `mkdocs.yml` nav entry. This PR only adds the `.py` file, so the example wouldn't actually surface anywhere in the docs. The PR notes also mention the module was never run end-to-end (only `py_compile` + `ruff`), so we don't have confidence it executes as intended.

**There are correctness gaps for an audit use case:**
- `after_tool_execute` doesn't fire on tool errors, so a failed tool call leaves a silent gap in the receipt chain — which undermines the whole point of a tamper-evident audit log.
- `canonical_json` permits `NaN`/`Infinity` and isn't true canonical JSON (RFC 8785), so hashes can diverge across runtimes/unicode and break the chain silently.
- For receipts positioned around auditability, HMAC only proves authenticity to a holder of the shared secret — not useful when the audit is *about* the signing party. The example acknowledges Ed25519 is the right tool but defaults to HMAC anyway.

**Scope.** A signed, hash-chained receipts example is fairly niche for the examples set. If we want to showcase hooks for observability/auditing, a simpler logging-oriented example would likely serve more users without the cryptographic caveats.

If you'd like to revisit this, we'd be happy to look at a follow-up that: adds the docs page + nav entry, records receipts on tool errors too, tightens canonicalization, and defaults to an asymmetric signature. Appreciate the effort here regardless!
@

PR #5761, opened 2026-06-02
Reject duplicate deferred tool call IDs

White-Mouse:
## Summary

- reject duplicate `tool_call_id` values before returning deferred tool requests
- reject duplicate pending IDs when resuming from message history with deferred tool results
- add regressions for ambiguous approval/result matching

## Why

Deferred tool resume matches results back to pending calls by `tool_call_id`. If a model response or imported message history contains duplicate pending IDs, the resume state is ambiguous: one approval/result can no longer be bound to exactly one pending tool call. This change fails closed before downstream resume handling consumes that ambiguous state.

## Tests

- `uv run ruff check pydantic_ai_slim/pydantic_ai/_agent_graph.py tests/test_agent.py`
- `uv run pytest -q tests/test_agent.py::test_deferred_tool_requests_reject_duplicate_tool_call_ids tests/test_agent.py::test_deferred_tool_results_reject_duplicate_tool_call_ids_in_history`
- `uv run pytest -q tests/test_agent.py`
- `git diff --check`

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/_agent_graph.py
+++ b/pydantic_ai_slim/pydantic_ai/_agent_graph.py
@@ -243,6 +243,17 @@ def is_agent_node(
     return isinstance(node, AgentNode)
 
 
+def _duplicate_tool_call_ids(calls: Sequence[_messages.ToolCallPart]) -> list[str]:
+    """Return duplicate tool call ids while preserving first duplicate order."""
+    seen: set[str] = set()
+    duplicates: list[str] = []
+    for call in calls:
+        if call.tool_call_id in seen and call.tool_call_id not in duplicates:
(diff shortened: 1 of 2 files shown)


hramezani:
Thanks for this, @White-Mouse! The concern is valid — duplicate `tool_call_id`s produce ambiguous result binding, and on the resume path it silently mis-executes (a single approval binds to both calls, running the unintended tool).

Since this branched before the V2 tool-execution refactor, `process_tool_calls` moved out of `_agent_graph.py` into `_tool_execution.py` and the branch no longer merged cleanly. I've rebased your change onto the current structure — same guards, helper, and regression tests, with your authorship preserved — in #6241. Closing this in favor of that; full credit to you. 🙏

PR #5763, opened 2026-06-02
fix(openai): ignore post-done tool arg deltas

### LINKED ISSUE #5757: OpenAI Responses: stray post-done arguments delta corrupts final tool-call `args` (model-layer follow-up to #4733)

he-yufeng:
﻿Fixes #5757.

## Summary

- track completed OpenAI Responses function-call item IDs during streaming
- ignore any later `response.function_call_arguments.delta` for those completed items
- add a synthetic Responses stream regression test where a malformed post-done `}` delta would otherwise corrupt the final tool-call args

The guard is specific to Responses function-call argument deltas. It does not treat "another part started" as completion, so normal interleaved streaming remains handled by the existing parts manager.

## To verify

- `python -m pytest tests/models/test_openai_responses.py::test_openai_responses_ignores_post_done_function_call_delta -q`
- `python -m pytest tests/models/test_openai_responses.py::test_openai_responses_null_text_stream tests/models/test_openai_responses.py::test_openai_responses_ignores_post_done_function_call_delta -q`
- `python -m ruff check pydantic_ai_slim/pydantic_ai/models/openai.py tests/models/test_openai_responses.py`
- `python -m ruff format --check pydantic_ai_slim/pydantic_ai/models/openai.py tests/models/test_openai_responses.py`
- `git diff --check`

Note: local `pyright` could not be run cleanly in this worktree because this machine is reusing a sibling checkout's `.venv`, while pyright resolves the configured venv relative to the current worktree.

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/openai.py
+++ b/pydantic_ai_slim/pydantic_ai/models/openai.py
@@ -3410,6 +3410,7 @@ class OpenAIResponsesStreamedResponse(StreamedResponse):
             # from the `output_item.added` event and merged into the corresponding
             # `TextPart.provider_details` on `output_text.done`.
             _phase_by_item: dict[str, Literal['commentary', 'final_answer']] = {}
+            _done_function_call_item_ids: set[str] = set()
 
             if self._provider_timestamp is not None:  # pragma: no branch
                 self.provider_details = {'timestamp': self._provider_timestamp}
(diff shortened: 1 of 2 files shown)


dsfaccini:
David's AICA here: thanks for this — the approach is exactly right (track done function-call item IDs, skip later `response.function_call_arguments.delta`, synthetic regression test).

We're consolidating the fix into #5027, which already handled the AG-UI protocol side of #4733/#5757 and now also carries the same model-layer guard in `OpenAIResponsesStreamedResponse`. Since that PR is further along and keeps the whole #4733/#5757 thread in one place, I'm closing this as superseded by #5027. Much appreciated — two independent PRs converging on the same approach was a good signal the fix is correct.

PR #5777, opened 2026-06-04
Support callable `requires_approval` for conditional tool approval

### LINKED ISSUE #5617: Allow `requires_approval` to accept a callable for conditional per-tool approval
### Description

Hey Pydantic team! Tagging @DouweM since this follows up on the work from #5577 but don't look at this notification while you're on vacation! We'll get to it later!

----

Nowadays I have a problem: I want to pass metadata based on tool call args to the `DeferredToolRequests` to send it to `HandleDeferredToolCalls`, and currently I am doing it like this:

```python
class MyCapability(AbstractCapability):
    async def wrap_tool_execute(self, ctx, *, call, tool_def, args, handler):
        if tool_def.name == "transfer" and not ctx.tool_call_approved:
            raise ApprovalRequired(metadata={"amount": args["amount"]})
        return await handler(args)
```

And then something came to my mind. When you need conditional human-in-the-loop approval on a specific tool (i.e. only ask for approval based on the argument values), the current way to do it is a `wrap_tool_execute` override in a capability:

```python
class MyCapability(AbstractCapability):
    async def wrap_tool_execute(self, ctx, *, call, tool_def, args, handler):
        if tool_def.name == "transfer" and not ctx.tool_call_approved:
            if args["amount"] > 1000:
                raise ApprovalRequired(metadata={"amount": args["amount"]})
        return await handler(args)
```

Three things bother me with this:

1. **Hardcoded tool name** — `tool_def.name == "transfer"` breaks silently on rename and has nothing to do with the capability's responsibility.
2. **The check runs for every tool call** — `wrap_tool_execute` intercepts everything; you have to route manually and may add a bit of latency.
3. **Approval logic is separated from the tool** — the predicate lives in a capability file, away from the function it governs. It would feel more natural to declare it next to the tool itself, similar to how Claude Code defines `requiresUserInteraction`, `checkPermissions`, `renderToolUseMessage` as first-class properties of a tool rather than external interceptors. That's a bigger API surface than what I'm proposing here — I'm just looking at the minimal step that covers the common case: a per-tool approval predicate that inspects the validated args.

## Proposed solution

Allow `requires_approval` to accept a callable in addition to `bool`:

```python
RequiresApprovalFunc = Callable[
    [RunContext[AgentDepsT], dict[str, Any]],
    dict[str, Any] | None | Awaitable[dict[str, Any] | None],
]
```

### Usage:

```python
def needs_approval(ctx: RunContext[MyDeps], args: dict[str, Any]) -> dict[str, Any] | None:
    if args["amount"] > 1000:
        return {"amount": args["amount"], "recipient": args["to"]}
    return None  # small transfers run without approval

@agent.tool(requires_approval=needs_approval)
def transfer(ctx: RunContext[MyDeps], amount: float, to: str) -> str:
    ...
```

Internally, the framework wraps `call_func` at `get_tools()` time with a guard that checks the callable before executing the tool. The model still sees a normal `kind='function'` tool — the approval decision happens at execution time, not at scheduling time:

```python
async def _guarded(args, ctx):
    if not ctx.tool_call_approved:
        result = await approval_fn(ctx, args)  # sync or async
        if result is not None:
            raise ApprovalRequired(metadata=result)
    return await original_call_func(args, ctx)
```

### Behaviour table:

| `requires_approval` | `ToolDefinition.kind` | First-pass behaviour |
|---|---|---|
| `False` (default) | `'function'` | Runs directly |
| `True` | `'unapproved'` | Always deferred (existing behaviour) |
| callable | `'function'` | Guard evaluated with validated args; deferred only if callable returns a dict |

The metadata dict returned by the callable is available in the `HandleDeferredToolCalls` handler via `DeferredToolRequests.metadata[tool_call_id]`.

### References

_No response_

MicroYui:
- Closes #5617

### What

`requires_approval` previously only accepted a `bool`. This PR lets it also accept a
[`RequiresApprovalFunc`][pydantic_ai.tools.RequiresApprovalFunc] callable so that whether a
specific tool call needs human-in-the-loop approval can depend on the validated arguments
or run context, with the approval policy declared next to the tool:

```python
from typing import Any

from pydantic_ai import Agent, DeferredToolRequests, RunContext

agent = Agent('openai:gpt-5.2', output_type=[str, DeferredToolRequests])


def needs_approval(ctx: RunContext[None], args: dict[str, Any]) -> dict[str, Any] | None:
    if args['amount'] > 1000:
        return {'amount': args['amount'], 'recipient': args['to']}
    return None  # small transfers run without approval


@agent.tool(requires_approval=needs_approval)
def transfer(ctx: RunContext[None], amount: float, to: str) -> str:
    return f'Transferred {amount} to {to}'
```

Behaviour table:

| `requires_approval` | `ToolDefinition.kind` | First-pass behaviour |
|---|---|---|
| `False` (default) | `'function'` | Runs directly |
| `True` | `'unapproved'` | Always deferred (existing behaviour) |
| callable | `'function'` | Guard evaluated with validated args; deferred only if the callable returns a dict |

The returned dict flows through the existing approval pipeline:
`ApprovalRequired(metadata=...)` → `DeferredToolRequests.metadata[tool_call_id]` →
`DeferredToolResults.metadata` → `RunContext.tool_call_metadata`.

### How

- Added the public `RequiresApprovalFunc` type alias in `pydantic_ai.tools` (exported from
  `pydantic_ai`).
- Widened `requires_approval` to `bool | RequiresApprovalFunc[...]` on `Tool`,
  `FunctionToolset.__init__` / `tool` / `tool_plain` / `add_function`, and `Agent.tool` /
  `Agent.tool_plain`.
- `Tool.tool_def` now maps to `kind='unapproved'` only when `requires_approval is True`, so a
  callable keeps `kind='function'` and the approval decision happens at execution time.
- `FunctionToolset.call_tool` evaluates the callable (sync or async) immediately before
  execution, outside the per-tool timeout, and only when `ctx.tool_call_approved` is `False`;
  a non-`None` return raises `ApprovalRequired(metadata=...)`.

### Why not `ApprovalRequiredToolset`?

`ApprovalRequiredToolset` remains the right API for applying a cross-cutting approval policy
across a whole toolset or MCP server (its predicate is `(ctx, tool_def, tool_args) -> bool`).
This PR targets the narrower per-function-tool case where the predicate belongs next to the
tool declaration and can return per-call metadata.

### Tests & docs

- Added tests in `tests/test_tools.py` covering: defer-with-metadata, run-directly on `None`,
  resume after approval (metadata exposed via `tool_call_metadata`), async callable,
  `kind` stays `'function'` for callables (and `'unapproved'` for `True`), validated args are
  passed (not raw JSON), all registration entry points, and the denial path.
- Updated `docs/deferred-tools.md`, `docs/toolsets.md`, `docs/tools-advanced.md`, and the
  `building-pydantic-ai-agents` skill.

### Checklist


rpelevin:
Callable `requires_approval` is a useful addition because approval becomes per-call instead of only per-tool.

The boundary I would keep stable is that the callable's returned metadata becomes the approval request contract for the exact validated arguments that were about to execute.

A few invariants seem important:

- evaluate the callable after argument validation and before tool execution;
- bind the approval request to `tool_call_id` and the validated argument shape;
- preserve the returned metadata through defer/resume without reconstructing it later;
- resume should not approve a different tool call with changed args;
- callable errors should fail closed before tool execution, not fall through to execution.

Acceptance tests I would add or keep:

- callable returns metadata -> tool is deferred and metadata is present in `DeferredToolRequests`;
- callable returns `None` -> tool runs directly;
- async callable has the same approval semantics as sync callable;
- resume exposes the same metadata through `RunContext.tool_call_metadata`;
- wrong `tool_call_id` or changed args cannot reuse a prior approval;
- `requires_approval=True` still maps to the existing always-deferred path.

That keeps conditional approval as a precise decision on one proposed call, not a broad preapproval for the function.

DouweM:
Closing pending discussion in the issue to define the shape: https://github.com/pydantic/pydantic-ai/issues/5617#issuecomment-4871011075.

PR #5779, opened 2026-06-04
fix(messages): from_data_uri crashes on a valid non-base64 data URI

### LINKED ISSUE #5778: BinaryContent.from_data_uri crashes with "not enough values to unpack" on a non-base64 data URI
`BinaryContent.from_data_uri` does `media_type, data = data_uri[len(prefix):].split(';base64,', 1)`, unpacking into two variables guarded only by `startswith('data:')`. RFC 2397 data URIs are not required to be base64-encoded, so a valid non-base64 URI lacks the `;base64,` substring and the unpack raises `ValueError: not enough values to unpack (expected 2, got 1)`.

```python
from pydantic_ai.messages import BinaryContent
BinaryContent.from_data_uri('data:text/plain,Hello%20World')
```

The method is public and called on untrusted data URIs in the Vercel AI and AG-UI adapters, so this surfaces as an opaque error rather than a clear one.

devteamaegis:
## What's broken

`BinaryContent.from_data_uri` unpacks `...split(';base64,', 1)` into two variables, guarded only by `startswith('data:')`. RFC 2397 permits data URIs without base64, so a valid one crashes:

```python
>>> from pydantic_ai.messages import BinaryContent
>>> BinaryContent.from_data_uri('data:text/plain,Hello%20World')
ValueError: not enough values to unpack (expected 2, got 1)
```

The classmethod is public and is called on client-supplied data URIs in the Vercel AI and AG-UI adapters.

## Why it happens

A data URI without the `;base64,` marker splits into a single element, so the two-target unpack fails.

## Fix

Check for the `;base64,` marker first and raise a clear `ValueError` if it's missing. Keeping the same exception class (`ValueError`) means the adapter call sites that already catch `ValueError` keep their existing fallback behavior.

## Test

Added a case asserting `from_data_uri('data:text/plain,Hello%20World')` raises a descriptive `ValueError`; a valid base64 data URI still parses.

Fixes #5778


dsfaccini:
thank you @devteamaegis

PR #5785, opened 2026-06-05
fix: prevent cancel() from flipping complete state and guard against stray post-done tool-call deltas

### LINKED ISSUE #5757: OpenAI Responses: stray post-done arguments delta corrupts final tool-call `args` (model-layer follow-up to #4733)

### LINKED ISSUE #5782: [streaming-resilience-sweep] Streaming: `await result.cancel()` after fully consumed stream flips `response.state` from 'complete' to 'interrupted'
## Impact

Anyone using the `async with agent.run_stream(...) as result:` pattern who defensively calls `await result.cancel()` in a `finally`-style cleanup path (or simply after a successful run) will see `result.response.state` flip from `'complete'` to `'interrupted'` even though the stream was actually fully and successfully consumed.

This is misleading: a successful run that ends in `'complete'` is silently downgraded to `'interrupted'` by a redundant defensive `cancel()`. The public `ModelResponse.state` field is a documented lifecycle field (`pydantic_ai_slim/pydantic_ai/messages.py:2150`); its value must reflect reality, not whether `cancel()` was called.

Note: this is distinct from open issues **#5615** (early break doesn't flip `is_complete` synchronously) and **#5756** (proposed v3 async-context-manager iterators). The fix in #5619 already correctly handles the history-completeness side of this — the bug here is in the `response.state` field, not in `all_messages()`.

## Focus Area & Code Path

- `pydantic_ai_slim/pydantic_ai/result.py:755` — `StreamedRunResult.cancel()` always calls `await self._stream_response.cancel()`, which sets `_cancelled = True` on the underlying `StreamedResponse`.
- `pydantic_ai_slim/pydantic_ai/models/__init__.py:1368` — `StreamedResponse.get()` checks `self._cancelled` *before* `self._finished`, so any prior `cancel()` call forces `state` to `'interrupted'` regardless of whether the stream was fully consumed.

## Reproduction

```python
import asyncio
from pydantic_ai import Agent


async def main():
    agent = Agent(TestModel(custom_output_text='Hello world'))

    async with agent.run_stream('Hi') as result:
        # Normal successful consumption
        await result.get_output()
        assert result.response.state == 'complete'  # passes
        # Defensive cancel() — common pattern in cleanup paths
        await result.cancel()
        # BUG: state is now 'interrupted' even though the run succeeded
        assert result.response.state == 'complete'  # FAILS


asyncio.run(main())
```

Run with: `uv run python repro.py`

## Expected vs Actual

**Expected:**
```
[1] before any consume:  state='incomplete'
[2] after get_output:    state='complete'  is_complete=True
[3] after cancel():      state='complete'  cancelled=True   (or no-op on the response)
```

**Actual:**
```
[1] before any consume:  state='incomplete'
[2] after get_output:    state='complete'  is_complete=True
[3] after cancel():      state='interrupted'  cancelled=True   <-- BUG
```

## Evidence

- Root cause: `pydantic_ai_slim/pydantic_ai/models/__init__.py:1368-1377`

  ```python
  def get(self) -> ModelResponse:
      """Build a ModelResponse from the data received from the stream so far."""
      if self._cancelled:
          state: ModelResponseState = 'interrupted'   # wins over _finished
      elif self._finished:
          state = 'complete'
      else:
          state = 'incomplete'
  ```

  The `_cancelled` branch must not override a stream that was already `_finished`.

- `pydantic_ai_slim/pydantic_ai/result.py:755-766` — `StreamedRunResult.cancel()` always invokes the underlying `StreamedResponse.cancel()` (line 758), which sets `_cancelled = True` even when the stream is already done. The `is_complete` guard at line 760 only protects `_all_messages` from a double-append, not `response.state`.

- Existing test `test_run_stream_cancel_after_complete` (`tests/test_streaming.py:4690`) asserts only `result.cancelled`, not `result.response.state`, so this regression is uncovered.

- Existing test `test_completed_streamed_response_cancel_noop` (`tests/test_streaming.py:4702`) covers `CompletedStreamedResponse` (a different code path used in tool-call re-asks) and explicitly asserts `response.state == 'complete'` after cancel — proving the intended invariant for that path. The new test must assert the same invariant for the normal `run_stream` path.

## Suggested fix

Make `cancel()` a no-op on an already-finished stream (preferred), or reorder the branches in `StreamedResponse.get()`. Either way the goal is the same: a successful run must not be downgraded to `'interrupted'` by a defensive `cancel()`.

```python
# Option A — guard cancel() itself (preferred):
# In StreamedResponse.cancel() / models/__init__.py ~line 1320
if self._finished:
    return
if self.cancelled:
    return
self._cancelled = True
await self.close_stream()

# Option B — fix get() to prefer _finished:
# In models/__init__.py:1368
def get(self) -> ModelResponse:
    if self._finished:
        state: ModelResponseState = 'complete'
    elif self._cancelled:
        state = 'interrupted'
    else:
        state = 'incomplete'
```

A regression test should be added next to `test_run_stream_cancel_after_complete` asserting `result.response.state == 'complete'` after the defensive `cancel()`.

Oxygen56:
## Summary

Two streaming bug fixes:

### 1. cancel() no-op on finished stream (closes #5782)

**Problem**: A defensive `await result.cancel()` in a `finally`-style cleanup path after a fully consumed stream would flip `response.state` from `'complete'` to `'interrupted'`.

**Root cause**: `StreamedResponse.cancel()` always set `_cancelled=True`, and `StreamedResponse.get()` checked `_cancelled` before `_finished` — so any prior `cancel()` downgraded a successful run.

**Fix**: `cancel()` now returns immediately (no-op) when the stream is already `_finished`, guarding against redundant defensive cancel calls.

### 2. Drop stray post-done arguments deltas (closes #5757)

**Problem**: Non-conforming OpenAI Responses-compatible endpoints may emit a `ResponseFunctionCallArgumentsDeltaEvent` **after** `ResponseOutputItemDoneEvent` / `ResponseFunctionCallArgumentsDoneEvent`. The stray delta was applied to the part, corrupting the final tool-call args (e.g., `'{...}' + '}'` → invalid JSON).

**Root cause**: While #5027 fixed the AG-UI **protocol** layer (suppressing the stray event from stream consumers), the delta had already mutated the underlying part via `handle_tool_call_delta`, corrupting the `ModelResponse.tool_calls` args.

**Fix**: Track done function-call `item_id`s in `_done_function_call_ids`, and skip `ResponseFunctionCallArgumentsDeltaEvent` for already-done items:
- `ResponseFunctionCallArgumentsDoneEvent` → add to done set
- `ResponseOutputItemDoneEvent` (for `ResponseFunctionToolCall` items) → add to done set
- `ResponseFunctionCallArgumentsDeltaEvent` → skip if item_id is in done set

### Files changed
| File | Change |
|------|--------|
| `models/__init__.py` | +6/-1 (cancel guard) |
| `models/openai.py` | +19/-7 (done tracking + delta guard) |

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/__init__.py
+++ b/pydantic_ai_slim/pydantic_ai/models/__init__.py
@@ -1321,8 +1321,13 @@ class StreamedResponse(ABC):
         so the flag is visible to any iterator that observes the transport error
         raised when the underlying connection is torn down, even if
         `close_stream()` itself raises.
+
+        Does nothing if the stream was already fully consumed
+        (`self._finished` is ``True``) — a redundant defensive `cancel()` after
+        a successful run must not flip `response.state` from `'complete'` to
+        `'interrupted'` (fixes #5782).
         """
-        if self.cancelled:
+        if self.cancelled or self._finished:
             return
         self._cancelled = True
         await self.close_stream()
--- a/pydantic_ai_slim/pydantic_ai/models/openai.py
+++ b/pydantic_ai_slim/pydantic_ai/models/openai.py
@@ -3406,6 +3406,11 @@ class OpenAIResponsesStreamedResponse(StreamedResponse):
         with _map_api_errors(self._model_name):
             # Track annotations by item_id and content_index
             _annotations_by_item: dict[str, list[Any]] = {}
+            # Track function-call item_ids that have already received a
+            # "done" signal. Stray arguments deltas arriving after the item
+            # is done (non-conforming endpoints) are silently dropped to
+            # prevent corrupting the final tool-call args (#5757).
+            _done_function_call_ids: set[str] = set()
             # Track `phase` (commentary | final_answer) on assistant message items, captured
             # from the `output_item.added` event and merged into the corresponding
             # `TextPart.provider_details` on `output_text.done`.
@@ -3447,15 +3452,18 @@ class OpenAIResponsesStreamedResponse(StreamedResponse):
                     self._usage += self._map_usage(chunk.response)
 
                 elif isinstance(chunk, responses.ResponseFunctionCallArgumentsDeltaEvent):
-                    maybe_event = self._parts_manager.handle_tool_call_delta(
-                        vendor_part_id=chunk.item_id,
-                        args=chunk.delta,
-                    )
-                    if maybe_event is not None:  # pragma: no branch
-                        yield maybe_event
+                    # Drop stray deltas arriving after the item is done (#5757).
+                    if chunk.item_id not in _done_function_call_ids:
+                        maybe_event = self._parts_manager.handle_tool_call_delta(
+                            vendor_part_id=chunk.item_id,
+                            args=chunk.delta,
+                        )
+                        if maybe_event is not None:  # pragma: no branch
+                            yield maybe_event
 
                 elif isinstance(chunk, responses.ResponseFunctionCallArgumentsDoneEvent):
-                    pass  # there's nothing we need to do here
+                    # Mark item done so post-done deltas are silently dropped (#5757).
+                    _done_function_call_ids.add(chunk.item_id)
 
                 elif isinstance(chunk, responses.ResponseIncompleteEvent):  # pragma: no cover
                     self._usage += self._map_usage(chunk.response)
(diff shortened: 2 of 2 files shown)


dsfaccini:
David's AICA here: heads up — the post-done tool-call delta half of this PR (#5757) is now superseded by #5027, which carries the same `OpenAIResponsesStreamedResponse` guard (track done function-call item IDs, skip later `response.function_call_arguments.delta`) alongside the AG-UI protocol fix from #4733.

The `cancel()`-on-finished-stream fix for #5782, though, is unique to this PR and still needed — nothing on `main` covers it. Could you rescope this PR to just that change (drop the `models/openai.py` tool-call-delta edits plus their test, keep the `models/__init__.py` `_finished` guard and its test)? That keeps it focused on #5782 so it can move forward independently. Thanks for the contribution!

Oxygen56:
Closing in favor of a clean PR with only the #5782 cancel fix. The #5757 portion was superseded by #5027.

Issue #4491, opened 2026-03-01
LLMJudge fails with Anthropic models when max_tokens is high


moredatarequired:
### Initial Checks

- [x] I'm using the [latest version](https://github.com/pydantic/pydantic-ai/releases/latest) of Pydantic AI
- [x] I've searched for my issue in [the issue tracker](https://github.com/pydantic/pydantic-ai/issues) before opening this issue

### Description

The Anthropic SDK enforces a streaming requirement client-side when max_tokens is above ~21k tokens. Since `LLMJudge` uses `agent.run()`, this means depending on setting and model, the evaluator can get a ValueError when it tries to evaluate.

```
ValueError: Streaming is required for operations that may take longer than 10 minutes.
See https://github.com/anthropics/anthropic-sdk-python#long-requests for more details
```

The error fires before any request is made. It's checked here in the Anthropic SDK: https://github.com/anthropics/anthropic-sdk-python/blob/49d639a671cb0ac30c767e8e1e68fdd5925205d5/src/anthropic/_base_client.py#L726
(no idea whether it's also enforced server side)

### Minimal, Reproducible Example

```Python
# /// script
# requires-python = ">=3.12"
# dependencies = [
#     "pydantic-ai",
#     "pydantic-evals",
#     "anthropic",
#     "logfire[httpx]",
# ]
# ///
import asyncio

import logfire

logfire.configure()
logfire.instrument_pydantic_ai()
logfire.instrument_httpx()

from pydantic_evals.evaluators.llm_as_a_judge import judge_output


async def main():
    result = await judge_output(
        output="Hello world!",
        rubric="Output is a greeting",
        model="anthropic:claude-sonnet-4-20250514",
        model_settings={"max_tokens": 32000},
    )
    print(result)


asyncio.run(main())
```

### Logfire Trace

https://logfire-us.pydantic.dev/public-trace/1b0824fd-6075-4c59-8f26-86f8d519208d?spanId=76363d6d276ff74f

### Python, Pydantic AI & LLM client version

- Python: Python 3.13.12
- Pydantic AI: 1.63.0
- LLM provider SDK: anthropic 0.52.0

DouweM:
@moredatarequired Thanks for the report. I'd rather handle this automatically inside `AnthropicModel.request` by catching the `Streaming is required for operations` error and handling it by using the streaming approach instead. Would you be up for implementing that?

moredatarequired:
[...]

PR #5786, opened 2026-06-05
fix(anthropic): guard against `message=None` in `_map_usage` on Bedrock streams

### LINKED ISSUE #5774: `_map_usage` crashes on Bedrock streams: `RawMessageStartEvent(message=None)` from type-less chunks
### Initial Checks

### Description

```python

event = construct_type(
)
```

### Suggested fix

Oxygen56:
## Summary

Fixes an AttributeError crash in `_map_usage` when running on Bedrock (#5774).

### Problem

On Bedrock (`AsyncAnthropicBedrock` via `AnthropicProvider`), the Anthropic SDK's stream decoder drops SSE event types, so Bedrock-only chunks like `amazon-bedrock-invocationMetrics` are constructed (non-validating `construct_type`) as `BetaRawMessageStartEvent(message=None)`.

`_map_usage` unconditionally dereferenced `message.message.usage`, causing:
```
AttributeError: 'NoneType' object has no attribute 'usage'
```

This is particularly insidious because it surfaces on **non-streaming** Bedrock calls that exceed the SDK's max_tokens threshold, triggering the internal streaming fallback introduced in #4491.

### Fix

Guard `message.message is None` in the `BetaRawMessageStartEvent` branch — return the existing usage or an empty `RequestUsage` when the SDK produces contract-violating events on Bedrock.

### Reproduction (before fix)

Closes #5774

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/__init__.py
+++ b/pydantic_ai_slim/pydantic_ai/models/__init__.py
@@ -1321,8 +1321,13 @@ class StreamedResponse(ABC):
         so the flag is visible to any iterator that observes the transport error
         raised when the underlying connection is torn down, even if
         `close_stream()` itself raises.
+
+        Does nothing if the stream was already fully consumed
+        (`self._finished` is ``True``) — a redundant defensive `cancel()` after
+        a successful run must not flip `response.state` from `'complete'` to
+        `'interrupted'` (fixes #5782).
         """
-        if self.cancelled:
+        if self.cancelled or self._finished:
             return
         self._cancelled = True
         await self.close_stream()
--- a/pydantic_ai_slim/pydantic_ai/models/anthropic.py
+++ b/pydantic_ai_slim/pydantic_ai/models/anthropic.py
@@ -2098,6 +2098,11 @@ def _map_usage(
     if isinstance(message, BetaMessage):
         response_usage = message.usage
     elif isinstance(message, BetaRawMessageStartEvent):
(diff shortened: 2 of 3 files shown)


dsfaccini:
David's AICA here: thanks for digging into this @Oxygen56, and apologies it took a beat to get to it.

Two reasons we're going to route #5774 through a different PR rather than land this one:

1. **The `_map_usage` guard alone doesn't stop the crash.** The real Bedrock failure path goes through `AnthropicStreamedResponse._get_event_iterator`, which dereferences the same `message=None` start event again on the next lines — `event.message.id` and `event.message.container`. With only the `_map_usage` guard, `_map_usage` returns cleanly but `self.provider_response_id = event.message.id` still raises `AttributeError: 'NoneType' object has no attribute 'id'`. The issue's reproduction calls `_map_usage` directly so it passes, but the non-streaming `request()` fallback (the `'Streaming is required'` path) iterates the stream and still crashes. A complete fix has to skip the event in the iterator, not just guard `_map_usage`.

2. **This PR bundles two unrelated changes.** The `openai.py` edit (#5757, stray post-done tool-call deltas) and the `models/__init__.py` edit (#5782, `cancel()` flipping `response.state`) aren't part of #5774 and aren't mentioned in the description. Both look worth doing, but each should be its own PR scoped to its issue so they can be reviewed and credited independently.

Closing in favor of a focused #5774 fix with a regression test. Please do re-open the #5757 and #5782 changes as separate PRs — happy to review them there. Thanks again!

PR #5793, opened 2026-06-05
fix: keep streaming state as 'complete' if cancelled after finish

ali009-eng:
Resilience Sweep:
      > Swapped logic in StreamedResponse.get() to prioritize _finished over _cancelled. This ensures state remains 'complete' if
  cancel() is called after a successful finish, as noted in the streaming-resilience-sweep (#5782).


dsfaccini:
David's AICA here: thanks for taking a crack at this! I'm closing this one as a duplicate — #5795 targets the same issue with a more complete change, so let's consolidate the discussion there. Appreciate the contribution.

PR #5795, opened 2026-06-05
fix: prevent `cancel()` from flipping complete state on finished stream

### LINKED ISSUE #5782: [streaming-resilience-sweep] Streaming: `await result.cancel()` after fully consumed stream flips `response.state` from 'complete' to 'interrupted'

Oxygen56:
## Summary

Fixes #5782 — `cancel()` on an already-consumed stream flipped `response.state` from `'complete'` to `'interrupted'`.

### Problem

Anyone using `async with agent.run_stream(...) as result:` who defensively calls `await result.cancel()` in a cleanup path saw `response.state` silently downgraded even though the stream was fully and successfully consumed.

### Root cause

`StreamedResponse.get()` checked `_cancelled` **before** `_finished`, so a `cancel()` after completion overwrote the correct `'complete'` state.

### Fix (issue's Option B)

`get()` now prefers `_finished` over `_cancelled`: a stream that finished naturally reports `'complete'` regardless of a later `cancel()`. `cancel()` still records `_cancelled` (so `result.cancelled` stays `True` — no public-API behavior change) but skips `close_stream()` on an already-finished stream, since there is no live connection left to tear down.

We chose this over making `cancel()` a full no-op (Option A) because Option A would also flip `result.cancelled` to `False` on a finished stream — a gratuitous public-API change, and one that would make `StreamedResponse` diverge from `CompletedStreamedResponse`, where `cancelled == True` and `state == 'complete'` already coexist. With the `state` lifecycle field now the canonical completeness signal, `cancelled` is free to simply mean "was `cancel()` called".

### Files changed

| File | Change |
|------|--------|
| `models/__init__.py` | reorder `get()`; `cancel()` skips `close_stream()` when finished |
| `tests/test_streaming.py` | `test_run_stream_cancel_after_complete` now asserts `response.state == 'complete'` after the defensive `cancel()` |


dsfaccini:
David's AICA here: maintainers are now actively working on this PR. **Please hold off on pushing further commits to this branch while we iterate** — we'd like to avoid races/superseding pushes.

### Decision on approach

The fix originally made `cancel()` a full no-op on a finished stream (issue #5782's Option A). We migrated it to **Option B**: reorder `StreamedResponse.get()` to prefer `_finished` over `_cancelled`.

Why B over A:

- **No public-API behavior change.** A flipped `result.cancelled` to `False` after a post-completion `cancel()`, which broke `test_run_stream_cancel_after_complete` (it asserts `cancelled is True`) and left CI red. B keeps `cancelled == True`; only the buggy `state` is fixed.
- **Consistency.** `CompletedStreamedResponse` already has `cancelled == True` coexisting with `state == 'complete'`. A would make the two streamed-response paths diverge; B keeps them aligned.
- **`state` is the canonical signal now.** With the `incomplete | complete | interrupted` lifecycle field in place, completeness is read from `state`, so `cancelled` can simply mean "was `cancel()` called".

A regression test asserting `response.state == 'complete'` after the defensive `cancel()` is included. CI should now be green.

dsfaccini:
hey @adtyavrdhn , small PR, review passes, please have a look and tell me if you're comfortable merging

Edit: i.e. merge it if you're fine with it / doesn't overlap with your work

docs/output.md
```diff
@@ -947,6 +947,9 @@ _(This example is complete, it can be run "as is" -- you'll need to add `asyncio
 
 If you `break` out of `stream_text()` and then leave the surrounding `async with` block, the stream is cleaned up as the context exits. Use `cancel()` when you want to stop generation immediately instead of only stopping local consumption.
 
+!!! note "Cancelling an already-completed stream"
```
adtyavrdhn: We don't need to document this, that was a bug and doesn't need explaining.

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -130,7 +130,9 @@
   while iteration is in flight.
 - `'interrupted'`: streaming was explicitly stopped via
   [`StreamedRunResult.cancel()`][pydantic_ai.result.StreamedRunResult.cancel] before the model
-  finished generating.
+  finished generating. Calling `cancel()` *after* the response is fully received leaves the state
```
adtyavrdhn: Here neither

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -1321,10 +1321,19 @@ async def cancel(self) -> None:
         so the flag is visible to any iterator that observes the transport error
         raised when the underlying connection is torn down, even if
         `close_stream()` itself raises.
+
+        Skips `close_stream()` when the stream was already fully consumed
+        (`self._finished` is `True`): there is no live connection left to tear
+        down. `get()` reports a finished stream as `'complete'` regardless of
```
adtyavrdhn: I don't think it needs anything more than a comment on why the ordering exists

adtyavrdhn (changes requested):
Just docstring changes, looks good to me otherwise. 

Thank you!

Oxygen56:
@adtyavrdhn addressed your review:

- `models/__init__.py`: replaced the docstring paragraph with a brief inline comment explaining why `_finished` is checked before `_cancelled`.
- `docs/output.md` and `messages.py`: the extra note/text was already removed in earlier commits (cfdc489e, 34329ba1, 1abd9775).

adtyavrdhn (approved):

Oxygen56:
Small follow-up: dropped a stray blank line that was left at the end of the `cancel()` docstring after the earlier trim. Diff is now just the inline comment explaining the ordering plus the `get()` reorder.

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -832,6 +832,10 @@ async def cancel(self) -> None:
         if self.cancelled:
```
adtyavrdhn, line 785: Can you move the condition to here, no need of another check. I missed this

adtyavrdhn: Let us check finished first to keep consistency with the bug we just had, will be easier to reason in our minds

Oxygen56:
@adtyavrdhn done — merged the two guard clauses into `if self.cancelled or self._finished: return`. Keeps `_finished` checked first, consistent with the `get()` fix.

Oxygen56:
👋 @adtyavrdhn @dsfaccini — all review threads are resolved and CI is green. Let me know if there is anything else you would like me to address before this is merged. Thanks!

adtyavrdhn (approved):

PR #5814, opened 2026-06-06
Fix `BinaryContent` tool return round-trips

### LINKED ISSUE #5811: [roundtrip-sweep] `ToolReturnPart.content`: `BinaryContent` round-trips as a plain dict via `dump_python(mode='json')`/`validate_python` ...
## Impact

A `BinaryContent` (or `BinaryImage`) placed in `ToolReturnPart.content` is silently **demoted to a plain `dict`** when messages are round-tripped through `ModelMessagesTypeAdapter.dump_python(..., mode='json')` + `validate_python` (the same boundary used by `to_jsonable_python` and many AG-UI / Vercel AI / OTel-instrumentation paths).

After round-trip:
- `part.content` is a `dict`, not a `BinaryContent`/`BinaryImage` model object.
- `content['data']` is a **base64-encoded `str`**, not `bytes` — so the binary payload is structurally different even if callers treat the value as a mapping.
- Any code that reads `part.content.data`, `.media_type`, `.identifier`, `.is_image`, `.format`, etc. as attributes will raise `AttributeError`.
- `part.content.files` (a property on `BaseToolReturnPart` that splits multimodal content) will be empty — the dict is not recognized as `MultiModalContent` and downstream `_split_content` returns the wrong shape.

This breaks the round-trip contract for any tool that returns multimodal data (image generation, audio synthesis, screenshot tools, MCP file-fetching tools, etc.) whenever the message history is persisted and reloaded via the Python-side `dump_python(mode='json')` path (as opposed to the `dump_json`/`validate_json` path, which works correctly).

Note: this is **distinct** from #5801 (a tracking issue for moving the `BinaryContent` -> `BinaryImage` narrowing into `MultiModalContent` validation). The bug here is that the value isn't reconstructed as a model object at all — it stays a `dict` — so `BinaryImage` narrowing is moot until that is fixed.

## Boundary & Code Path

- `ModelMessagesTypeAdapter.dump_python(messages, mode='json')` -> `ModelMessagesTypeAdapter.validate_python(serialized)`
- `pydantic_ai_slim/pydantic_ai/messages.py:1087-1094` — `ToolReturnContent = TypeAliasType('ToolReturnContent', MultiModalContent | Sequence['ToolReturnContent'] | Mapping[str, 'ToolReturnContent'] | Any)`. The recursive `TypeAliasType` defeats the `MultiModalContent` `Discriminator('kind')` dispatch on `validate_python`, so the value passes through as a raw `dict` rather than being reconstructed as `BinaryContent`.
- `pydantic_ai_slim/pydantic_ai/messages.py:509-547` — `BinaryContent` config: `ser_json_bytes='base64'`, `val_json_bytes='base64'`. The `val_json_bytes='base64'` config is honored on the JSON path (`dump_json`/`validate_json` round-trips correctly) but **not** applied when validating the same base64 string through the Python-side path, so `data` stays a string.

For comparison, the same round-trip works for:
- `FilePart.content` — typed as `Annotated[MultiModalContent, AfterValidator(BinaryImage.narrow_type)]` (`messages.py:1730+`), non-recursive union.
- `UserPromptPart.content` — typed as `str | Sequence[UserContent]` where `UserContent = str | TextContent | MultiModalContent | CachePoint` (`messages.py:876`), non-recursive union.

Only the `ToolReturnPart.content` path (recursive `ToolReturnContent`) is broken.

## Reproduction

A minimal test was added to `tests/test_messages.py`:

```python
def test_tool_return_binary_content_serialization_roundtrip():
    messages: list[ModelMessage] = [
        ModelRequest(
            parts=[
                ToolReturnPart(
                    tool_name='image_tool',
                    content=BinaryContent(data=b'fake', media_type='image/jpeg'),
                    tool_call_id='call-1',
                )
            ]
        )
    ]
    serialized = ModelMessagesTypeAdapter.dump_python(messages, mode='json')
    deserialized = ModelMessagesTypeAdapter.validate_python(serialized)
    assert deserialized == messages
    part = deserialized[0].parts[0]
    assert isinstance(part, ToolReturnPart)
    assert isinstance(part.content, BinaryContent)
    assert part.content.data == b'fake'
    assert part.content.media_type == 'image/jpeg'
```

Run with:

```
uv run pytest tests/test_messages.py::test_tool_return_binary_content_serialization_roundtrip -xvs
```

## Expected vs Actual

**Expected:** `part.content` is a `BinaryContent` with `data=b'fake'` and `media_type='image/jpeg'`.

**Actual** (pytest `-xvs` output):

```
-     ModelRequest(parts=[ToolReturnPart(tool_name='image_tool', content=BinaryContent(data=b'fake', media_type='image/jpeg'), tool_call_id='call-1', timestamp=datetime.datetime(2026, 6, 6, 4, 28, 10, 894236, tzinfo=datetime.timezone.utc))]),
+     ModelRequest(parts=[ToolReturnPart(tool_name='image_tool', content={'data': 'ZmFrZQ==', 'media_type': 'image/jpeg', 'vendor_metadata': None, 'kind': 'binary', 'identifier': 'c053ec'}, tool_call_id='call-1', timestamp=datetime.datetime(2026, 6, 6, 4, 28, 10, 894236, tzinfo=TzInfo(0)))]),
```

`tests/test_messages.py:667: AssertionError`

## Evidence

- `pydantic_ai_slim/pydantic_ai/messages.py:1087-1094` — `ToolReturnContent` recursive `TypeAliasType` definition.
- `pydantic_ai_slim/pydantic_ai/messages.py:509-547` — `BinaryContent` `pydantic_dataclass` config (`ser_json_bytes='base64'`, `val_json_bytes='base64'`).
- `pydantic_ai_slim/pydantic_ai/messages.py:1730+` — `FilePart.content` typed as `Annotated[MultiModalContent, AfterValidator(BinaryImage.narrow_type)]` (working path).
- `pydantic_ai_slim/pydantic_ai/messages.py:876` — `UserContent` non-recursive union (working path).
- `tests/test_messages.py` — `test_file_part_serialization_roundtrip` (line 587) and `test_uploaded_file_serialization_roundtrip` (line 1030) are the existing round-trip patterns this issue extends.

Cross-references: complements #5801 (the planned `MultiModalContent`-level narrowing refactor — that work assumes `BinaryContent` is reconstructed at all, which it currently isn't on the `ToolReturnPart.content` Python-mode path).

william-xue:
- Closes #5811

This fixes `ToolReturnPart.content` JSON-mode round-tripping for multimodal content. `BinaryContent` serialized via `ModelMessagesTypeAdapter.dump_python(..., mode='json')` was validating back as a plain dict because the recursive tool-return content union can accept generic mappings before reconstructing multimodal content.

The change restores serialized multimodal content before normal content validation, including nested lists/dicts, while leaving ordinary user dictionaries with unrelated `kind` values unchanged.

Validation:

```console
uv run pytest tests/test_messages.py::test_tool_return_binary_content_roundtrip tests/test_messages.py::test_tool_return_dict_with_kind_roundtrip -q
uv run ruff check pydantic_ai_slim/pydantic_ai/messages.py tests/test_messages.py
uv run pyright pydantic_ai_slim/pydantic_ai/messages.py tests/test_messages.py
```

### Checklist

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/messages.py
+++ b/pydantic_ai_slim/pydantic_ai/messages.py
@@ -6,7 +6,7 @@ import mimetypes
 import os
 import warnings
 from abc import ABC, abstractmethod
-from collections.abc import Callable, Mapping, Sequence
+from collections.abc import Callable, ItemsView, Mapping, Sequence
 from dataclasses import KW_ONLY, dataclass, field, replace
 from datetime import datetime
 from mimetypes import MimeTypes
@@ -867,6 +867,10 @@ MultiModalContent = Annotated[
 # Explicit tuple for readability; validated against MultiModalContent in tests
 MULTI_MODAL_CONTENT_TYPES: tuple[type, ...] = (ImageUrl, AudioUrl, DocumentUrl, VideoUrl, BinaryContent, UploadedFile)
 
+MULTI_MODAL_CONTENT_KINDS = frozenset(
+    {'image-url', 'audio-url', 'document-url', 'video-url', 'binary', 'uploaded-file'}
+)
+
 
 def is_multi_modal_content(obj: Any) -> TypeGuard[MultiModalContent]:
     """Check if obj is a MultiModalContent type, enabling type narrowing."""
@@ -1094,6 +1098,25 @@ else:
     )
 
 
+multi_modal_content_ta: pydantic.TypeAdapter[MultiModalContent] = pydantic.TypeAdapter(
+    MultiModalContent, config=pydantic.ConfigDict(defer_build=True, ser_json_bytes='base64', val_json_bytes='base64')
+)
+
+
+def _restore_multi_modal_content(content: Any) -> Any:
+    if isinstance(content, Mapping):
+        items = cast(ItemsView[str, Any], content.items())
+        content_dict = dict(items)
+        if content_dict.get('kind') in MULTI_MODAL_CONTENT_KINDS:
+            return multi_modal_content_ta.validate_python(content)
+        return {key: _restore_multi_modal_content(value) for key, value in content_dict.items()}
+    elif isinstance(content, Sequence) and not isinstance(content, str):
+        values = cast(Sequence[Any], content)
+        return [_restore_multi_modal_content(value) for value in values]
+    else:
+        return content
+
+
 ToolPartKind: TypeAlias = Literal['tool-search', 'capability-load']
 """Discriminator value for the typed call/return-part subclass associated with a tool.
 
@@ -1151,6 +1174,11 @@ class BaseToolReturnPart:
     - `'failed'`: The tool raised an error during execution.
     - `'denied'`: The tool call was denied by the approval mechanism.
     """
+
+    @pydantic.field_validator('content', mode='before')
+    @classmethod
+    def _validate_content(cls, content: Any) -> Any:
+        return _restore_multi_modal_content(content)
 
     def _split_content(self) -> tuple[list[Any], list[MultiModalContent], bool]:
         """Split content into non-file and file parts.
--- a/tests/test_messages.py
+++ b/tests/test_messages.py
@@ -341,6 +341,38 @@ def test_binary_content_base64():
     assert bc.data_uri == 'data:image/png;base64,SGVsbG8sIHdvcmxkIQ=='
 
 
+def test_tool_return_binary_content_roundtrip():
+    messages: list[ModelMessage] = [
+        ModelRequest(
+            parts=[
+                ToolReturnPart(tool_name='get_image', content=BinaryContent(data=b'image-data', media_type='image/png'))
+            ]
+        )
+    ]
+
+    payload = ModelMessagesTypeAdapter.dump_python(messages, mode='json')
+    round_tripped = ModelMessagesTypeAdapter.validate_python(payload)
+
+    part = round_tripped[0].parts[0]
+    assert isinstance(part, ToolReturnPart)
+    assert isinstance(part.content, BinaryContent)
+    assert part.content.data == b'image-data'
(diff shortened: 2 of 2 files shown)


william-xue:
Quick follow-up: this PR is clean with all checks passing and is labeled `bug` / `size: S`. The fix is limited to preserving `BinaryContent` during tool-return message round-trips, with regression tests added. Please let me know if you want any changes before merge.

william-xue:
I rechecked this after the latest status refresh: all CI checks are still passing, and the diff remains limited to preserving `BinaryContent` through tool-return message round-trips with regression coverage. Happy to adjust if there is anything you would like changed.

dsfaccini:
David's AICA here: the bug you found here is real — `BinaryContent` returned from a tool doesn't survive the `dump_python(mode='json')` → `validate_python` round-trip — and your repro pinned it correctly.

We're going to close this in favor of #5255, which is already approved and fixes the same root cause one level up: it places a `pydantic.Discriminator` on the core `ToolReturnContent` alias, so multimodal content re-hydrates into its subclass everywhere `ModelMessagesTypeAdapter` runs (both the JSON and the `dump_python(mode='json')` paths) rather than via a separate load-time walker. We've verified #5255 closes #5811's exact repro.

One note that may be useful: the discriminator approach also sidesteps an edge the load-time walker runs into — a non-multimodal dict whose `kind` happens to collide with a reserved value (e.g. `{'kind': 'binary', 'label': 'foo'}`) would raise on reload under the walker, whereas the discriminator only dispatches when a type-specific field (`media_type`/`url`/`file_id`) is also present, so such dicts stay plain mappings.

Closing as superseded by #5255. Thank you for the report and the fix — it's appreciated, and please do keep contributing.

PR #5817, opened 2026-06-07
fix(openai): merge LiteLLM leading system prompts

### LINKED ISSUE #5812: vLLM backend rejects PromptedOutput requests with 'System message must be at the beginning'
### Description

```
```

### Reproduction

```python
from pydantic_ai import Agent

class MyOutput(BaseModel):
    answer: str

model = OpenAIChatModel(
    model_name="my-vllm-model",
)

agent = Agent(
    model=model,
)

```

### Expected behavior

### Actual behavior

### Workaround

```python

model = OpenAIChatModel(
    model_name="my-vllm-model",
    profile=OpenAIModelProfile(
    ),
)
```

### Suggested improvements

### Environment

- pydantic-ai version: latest
- Model: custom hosted model

he-yufeng:
## Summary

Fixes #5812.

`LiteLLMProvider` is commonly used in front of strict OpenAI-compatible servers such as vLLM. Those servers can reject requests that start with more than one `system` message, which happens when ordinary agent instructions are combined with extra `PromptedOutput` instructions.

The OpenAI model already has the correct compatibility path via `openai_chat_supports_multiple_system_messages=False`; this changes the LiteLLM provider profile to use that path by default and documents the behavior in the LiteLLM section.

## Validation

```text
PYTHONUTF8=1 PYTHONIOENCODING=utf-8 uv run pytest tests/providers/test_litellm.py -q
PYTHONUTF8=1 PYTHONIOENCODING=utf-8 uv run ruff check pydantic_ai_slim/pydantic_ai/providers/litellm.py tests/providers/test_litellm.py
PYTHONUTF8=1 PYTHONIOENCODING=utf-8 uv run ruff format --check pydantic_ai_slim/pydantic_ai/providers/litellm.py tests/providers/test_litellm.py
python -m py_compile pydantic_ai_slim/pydantic_ai/providers/litellm.py tests/providers/test_litellm.py
git diff --check
```

Note: `PYTHONUTF8=1` is needed in my Windows shell because dependency metadata generation otherwise reads `pyproject.toml` with the local GBK default encoding.

### DIFF
--- a/docs/models/openai.md
+++ b/docs/models/openai.md
@@ -748,6 +748,9 @@ To use [LiteLLM](https://www.litellm.ai/), set the configs as outlined in the [d
 
 To use custom LLMs, use `custom/` prefix in the model name.
 
+`LiteLLMProvider` merges consecutive leading system messages by default, which is required by stricter
+OpenAI-compatible backends such as some vLLM deployments.
+
 Once you have the configs, use the [`LiteLLMProvider`][pydantic_ai.providers.litellm.LiteLLMProvider] as follows:
 
 ```python
--- a/pydantic_ai_slim/pydantic_ai/providers/litellm.py
+++ b/pydantic_ai_slim/pydantic_ai/providers/litellm.py
@@ -78,9 +78,15 @@ class LiteLLMProvider(Provider[AsyncOpenAI]):
         if profile is None:
             profile = openai_model_profile(model_name)
 
-        # As LiteLLMProvider is used with OpenAIModel, which uses OpenAIJsonSchemaTransformer,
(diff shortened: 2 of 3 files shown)


he-yufeng:
The only failing check currently looks unrelated to this PR. The `test on 3.11 (lowest-versions)` job reports network/cache errors while setting up `tests/test_embeddings.py::TestSentenceTransformers::{test_documents,test_query}`:

```text
OSError: We couldn't connect to 'https://huggingface.co' to load the files, and couldn't find them in the cached files.
```

The LiteLLM-specific tests and the other CI jobs are passing.

PR #5818, opened 2026-06-07
fix(anthropic): guard `message=None` Bedrock start events in stream path

### LINKED ISSUE #5774: `_map_usage` crashes on Bedrock streams: `RawMessageStartEvent(message=None)` from type-less chunks

Bartok9:
## Summary

- Guards the three `BetaRawMessageStartEvent(message=None)` dereference sites that crash Anthropic-on-Bedrock streaming.
- A non-streaming Bedrock call with high `max_tokens` no longer dies with `AttributeError: 'NoneType' object has no attribute ...`.

## Motivation

Closes #5774.

On Bedrock (`AsyncAnthropicBedrock` via `AnthropicProvider`), the Anthropic SDK's stream decoder drops SSE event types, so Bedrock-only chunks like `amazon-bedrock-invocationMetrics` are constructed via non-validating `construct_type` into `BetaRawMessageStartEvent(message=None)` — violating the type annotation. pydantic-ai is the first consumer to dereference them.

This surfaces in a non-obvious way: a *non-streaming* `model.request()` with `max_tokens` above the SDK threshold (~21.3k) trips the SDK's "Streaming is required" guard, which pydantic-ai catches and falls back to internal streaming (the #4491 fix). On Anthropic-direct that's fine; on Bedrock the fallback walks straight into the broken `message=None` events.

## Root cause

The same contract-violating start event is dereferenced at **three** sites in `models/anthropic.py`:

1. `_process_streamed_response` — `first_chunk.message.model` (crashes during stream setup if the *first* chunk is the type-less Bedrock metrics event).
2. `_map_usage` — `message.message.usage`.
3. `AnthropicStreamedResponse._get_event_iterator` — `event.message.id` / `event.message.container`.

## Fix

- `_process_streamed_response`: fall back to the configured `self.model_name` when `first_chunk.message is None` (the leading Bedrock chunk carries no model name; the iterator skips it anyway).
- `_map_usage`: return `existing_usage or RequestUsage()` instead of dereferencing `.usage`.
- `_get_event_iterator`: `continue` past the event before touching `.id` / `.container`.

The SDK violates its own type annotation here (`message` is declared required, non-`None`), so each guard carries `# pyright: ignore[reportUnnecessaryComparison]` with a comment, matching the codebase's existing convention for SDK-contract-violation guards.

## Tests

Two regression tests in `tests/models/test_anthropic.py`, using `BetaRawMessageStartEvent.model_construct(message=None)` to mirror the SDK's `construct_type` on Bedrock:

- `test_map_usage_bedrock_start_event_without_message` — direct `_map_usage` unit assertion (existing + empty usage paths).
- `test_streaming_bedrock_start_event_without_message_is_skipped` — drives the full streaming path via the mock client with a leading `message=None` chunk; the run completes instead of raising.

## Verification

- `uv run pytest tests/models/test_anthropic.py` — 233 passed
- `uv run ruff check` / `ruff format --check` — clean
- `uv run pyright pydantic_ai_slim/pydantic_ai/models/anthropic.py tests/models/test_anthropic.py` — 0 errors
- `uv run mypy` — clean

## Relationship to prior PRs

#5787 (@xlyoung) was identified by the maintainer as the correct two-site approach but was auto-closed as a duplicate of #5786 (@Oxygen56), which was then closed for being incomplete (only `_map_usage` guarded) and bundling unrelated changes. Neither landed; #5774 currently has no open PR.

This PR is scoped to #5774 only, adds the regression test the maintainer asked for, and additionally guards the **third** dereference site (`first_chunk.message.model` in `_process_streamed_response`) — surfaced by the full-streaming-path regression test and not covered by either prior PR.

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/anthropic.py
+++ b/pydantic_ai_slim/pydantic_ai/models/anthropic.py
@@ -976,7 +976,7 @@ class AnthropicModel(Model[AsyncAnthropicClient]):
         # On Bedrock the SDK drops SSE event types, so a leading Bedrock-only chunk
         # (e.g. `amazon-bedrock-invocationMetrics`) is non-validating `construct_type`d
         # into `BetaRawMessageStartEvent(message=None)`. Fall back to the configured model
-        # name rather than dereference `first_chunk.message.model` (issue #5774). The
+        # name rather than dereference `first_chunk.message.model` (https://github.com/pydantic/pydantic-ai/issues/5774). The
         # iterator below skips these `message=None` events.
         model_name = first_chunk.message.model if first_chunk.message is not None else self.model_name  # pyright: ignore[reportUnnecessaryComparison]
 
--- a/pydantic_ai_slim/pydantic_ai/models/anthropic.py
+++ b/pydantic_ai_slim/pydantic_ai/models/anthropic.py
@@ -2109,6 +2109,8 @@ def _map_usage(
             # On Bedrock the Anthropic SDK drops SSE event types, so Bedrock-only chunks
             # (e.g. `amazon-bedrock-invocationMetrics`) are non-validating `construct_type`d
             # into `BetaRawMessageStartEvent(message=None)`, violating the type annotation.
+            # The metrics chunk's token counts duplicate the canonical `message_start` /
+            # `message_delta` usage, so dropping them here avoids double-counting.
             return existing_usage or usage.RequestUsage()
         response_usage = message.message.usage
     elif isinstance(message, BetaRawMessageDeltaEvent):
--- a/pydantic_ai_slim/pydantic_ai/models/anthropic.py
+++ b/pydantic_ai_slim/pydantic_ai/models/anthropic.py
@@ -973,9 +973,16 @@ class AnthropicModel(Model[AsyncAnthropicClient]):
 
         assert isinstance(first_chunk, BetaRawMessageStartEvent)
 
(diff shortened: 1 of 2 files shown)


dsfaccini:
David's AICA here: thanks for the solid fix @Bartok9 — it guards all the right dereference sites. I ran a full review pass and it came back clean (no blocking issues); the only thing worth tightening was the streaming regression test, so I pushed one commit to your branch (edits-from-maintainers was on):

- **Strengthened `test_streaming_bedrock_start_event_without_message_is_skipped`** — it now asserts the skipped `message=None` chunk contributes no usage and no response id: usage comes from the real `message_start`/`message_delta` (`input_tokens=4, output_tokens=2`), `provider_response_id` from the real event, and `model_name` falls back to the configured id (which also pins your `_process_streamed_response` fallback, since the malformed chunk is peeked first). Previously it only asserted `output == 'hello'`, so a regression that swallowed the real event's usage/id would have passed green.
- **Fixed the docstring** to describe the actual `run_stream` path it drives, and noted the leading-chunk ordering is a constructed worst case (real Bedrock trails the metrics chunk) that deliberately also reaches the model-name fallback.
- **Added a code comment** at the `_map_usage` guard explaining why dropping the Bedrock metrics token counts is correct (they duplicate the canonical `message_start`/`message_delta` usage), plus a note on why the `_map_usage` test is a unit test rather than VCR.

Feel free to push back on any of it. Thanks again for getting to this first!

PR #5827, opened 2026-06-08
fix(anthropic): ensure count_tokens preserves replay history and drops native tools

### LINKED ISSUE #5780: `AnthropicModel.count_tokens` diverges on `ToolSearchTool` replay history
### Summary

After #5704, `AnthropicModel.count_tokens` strips native tools from the token-counting request (`AnthropicModel._messages_count_tokens` does `model_request_parameters = replace(model_request_parameters, native_tools=[])` before `_add_native_tools` runs). This correctly fixes the API 400 for server tools, but `native_tools` is read by a **second** consumer in the same count path, and clearing it has an unintended side effect on how tool-search replay history is rendered.

### Failing code path

- `AnthropicModel._messages_count_tokens` (`pydantic_ai_slim/pydantic_ai/models/anthropic.py`) clears `native_tools` to `[]`, then calls `_map_message`.
- `_map_message` derives `tool_search_active = any(isinstance(t, ToolSearchTool) for t in model_request_parameters.native_tools)`. With `native_tools` cleared, `tool_search_active` is forced to `False` in the count path even when the real request path (`_messages_create`, which receives `model_request_parameters` unmodified) has it `True`.
- `tool_search_active` feeds `_build_custom_tool_search_replay_blocks(request_part, tool_search_active, available_tool_names)`. When `False`, the custom tool-search replay returns no `tool_reference` blocks, so a `ToolReturnPart` from a prior tool-search exchange is rendered as a plain `tool_result` (text) block instead of the `tool_reference` array that the actual `/v1/messages` request emits.

### Why this is a latent bug

The purpose of `count_tokens_before_request` is to estimate the token count of the request that will actually be sent. When an agent has a `ToolSearchTool` capability **and** the message history contains a tool-search replay exchange (a `ToolReturnPart` for the local `search_tools` tool), the count path serializes that history turn to a **different wire shape** than the real request. So the estimate diverges from the real request body — not just by the (already-accepted, API-forced) omission of native-tool definitions, but in the history body itself.

Note the `tool_reference` entries point at **function tools** (`available_tool_names = {t.name for t in model_request_parameters.function_tools}`), which are *not* cleared by the #5704 fix — so they would still be valid in a count_tokens payload. That means preserving `tool_search_active` in the count path is feasible and would be more accurate.

### Severity

Low / narrow. This is a divergence in the estimate, not an error: pre-#5704, any `ToolSearchTool` + `count_tokens` combination was a hard 400, so #5704 is a strict improvement. This issue tracks closing the remaining accuracy gap for the `ToolSearchTool` + tool-search-replay-history + `count_tokens_before_request=True` combination.

### Suggested fix (one of)

- Clear native tools **after** `_map_message` runs (so `tool_search_active` is still derived from the real `native_tools`), or
- Clear only the non-`ToolSearchTool` native tools in the count path, preserving the `ToolSearchTool` discriminator while still dropping the server-tool definitions / betas / mcp_servers that the endpoint rejects.

Either way, confirm the `tool_search_tool_*` server-side entry itself is still excluded from the count_tokens `tools` list (the endpoint rejects it), while the history `tool_reference` blocks (which reference function tools) are preserved.

### Acceptance criteria

- [ ] With a `ToolSearchTool` capability and a message history containing a tool-search replay `ToolReturnPart`, `count_tokens` renders that history turn with the same `tool_reference` wire shape as the real `/v1/messages` request.
- [ ] The server-side `tool_search_tool_*` tool entry is still omitted from the `count_tokens` `tools` list.
- [ ] A regression test pairs a tool-search-replay history with `count_tokens_before_request=True` and asserts the count-path `messages` kwarg matches the create-path shape for the tool-search turn.

### Cross-reference

Discovered during review of #5704 (the fix that introduced the `native_tools=[]` clearing). See `AnthropicModel._messages_count_tokens` and `_map_message`'s `tool_search_active` in `pydantic_ai_slim/pydantic_ai/models/anthropic.py`.

Sujit-1509:
- Addresses #5702 (Replaces inactive PR #5704 which introduces the #5780 regression)
- Closes #5780

### Summary

- Fixes the Anthropic token counting payload to omit native/server tools (which fixes API 400 errors).
- Defers native tool removal until after message mapping completes, which fixes an issue where the \ToolSearch\ replay history was diverging from the real execution payload by incorrectly serializing as plain text.

### Checklist

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/anthropic.py
+++ b/pydantic_ai_slim/pydantic_ai/models/anthropic.py
@@ -850,12 +850,15 @@ class AnthropicModel(Model[AsyncAnthropicClient]):
         if isinstance(self.client, AsyncAnthropicBedrock):
             raise UserError('AsyncAnthropicBedrock client does not support `count_tokens` api.')
 
-        # standalone function to make it easier to override
-        tools, tool_choice = self._prepare_tools_and_tool_choice(model_settings, model_request_parameters)
-        tools, mcp_servers, native_tool_betas = self._add_native_tools(tools, model_request_parameters, model_settings)
-
         auto_cache_control, resolved_cache_ttl = self._build_automatic_cache_control(model_settings)
         system_prompt, anthropic_messages = await self._map_message(messages, model_request_parameters, model_settings)
+
+        # Omit native/server tools from count_tokens payload to avoid API 400 errors,
+        # but only after mapping messages so that history replay formatting is preserved.
+        count_params = replace(model_request_parameters, native_tools=[])
+        tools, tool_choice = self._prepare_tools_and_tool_choice(model_settings, count_params)
+        tools, mcp_servers, native_tool_betas = self._add_native_tools(tools, count_params, model_settings)
+
         self._apply_per_block_caching_fallback(resolved_cache_ttl, anthropic_messages)
         self._apply_explicit_message_caching(model_settings, anthropic_messages)
         self._limit_cache_points(
--- a/tests/models/test_anthropic.py
+++ b/tests/models/test_anthropic.py
@@ -10630,9 +10630,10 @@ async def test_anthropic_count_tokens_omits_native_tools(allow_model_requests: N
             parts=[
                 ToolSearchReturnPart(
                     tool_name='search_tools',
-                    content=[{'name': 'my_tool'}],
-                    discovered_tools=[{'name': 'my_tool', 'description': ''}],
-                    message='Found 1 tool',
+                    content={
+                        'discovered_tools': [{'name': 'my_tool', 'description': ''}],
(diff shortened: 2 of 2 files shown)


dsfaccini:
David's AICA here: thank you for taking a crack at #5780! I'm closing this in favor of #5855, which supersedes it. For reference, the root cause is that the count_tokens path stripped `ToolSearchTool` before `_map_message`, forcing `tool_search_active` off — so tool-search replay history rendered a different wire shape than the real `/v1/messages` request. #5855 splits the params (the wire `tools` list still omits the server-side tool, but `_map_message` keeps the correct `tool_reference` shape), and adds a wire-shape guard plus a live VCR regression test. Appreciate the effort here!

PR #5839, opened 2026-06-09
Document snapshot matcher-vs-concrete-value guidance in `tests/AGENTS.md`

dsfaccini:
Adds a decision rule for when a snapshot value should stay a concrete number vs. become a dirty-equals matcher (`IsInt()`, `IsFloat()`, `IsStr()`, ...).

Motivation: when an existing concrete snapshot value changes (after a re-record, dependency bump, or refactor), swapping it for a matcher to make the test pass silently discards the drift-detection signal the snapshot existed to provide. The new guidance frames the call as one question — *did the concrete value provide stability?* — keep small, meaningful scalars (token counts, costs, ids) concrete and root-cause why they changed; reserve matchers for genuinely non-deterministic values (timestamps, random ids) or large opaque noise (a static base64 blob). It also notes that a value differing only because CI runs two dependency versions should be fixed by aligning the versions, not by masking the split with a matcher.

Lands in `tests/AGENTS.md` (the canonical testing-philosophy doc, loaded for every test-writing agent) and the VCR `testing-skill` snapshot-review step.

### Checklist

### DIFF
--- a/.claude/skills/testing-skill/SKILL.md
+++ b/.claude/skills/testing-skill/SKILL.md
@@ -48,6 +48,15 @@ If tests use [`snapshot()`](https://github.com/15r10nk/inline-snapshot) assertio
 - You only review - don't manually write snapshot contents
 - Snapshots capture what the test actually produced, additional to explicit assertions
 
+## Snapshots: prefer a concrete value over a dirty-equals matcher
+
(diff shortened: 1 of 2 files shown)


dsfaccini:
David's AICA here: Superseding this in favor of a single woven Best-Practices line in `tests/AGENTS.md` (folded into #5847), rather than the duplicated standalone section this PR added to both `tests/AGENTS.md` and the testing skill. Closing.

PR #5840, opened 2026-06-09
Map `seed` and `top_k` ModelSettings for GeminiModel

ly-wang19:
## Summary

`GeminiModel` (the Gemini GLA API backend) silently ignores the `seed` and `top_k` model settings. `_settings_to_generation_config` maps `temperature`, `top_p`, `presence_penalty`, `frequency_penalty`, etc., but never `seed` or `top_k` — so a user who sets either on a `GeminiModel` has it dropped without warning.

This is inconsistent with `GoogleModel`, which already forwards both, and with the recent provider-by-provider `seed` support (#2842 added it to `GoogleModel`, #5741 to xAI). The Gemini `generationConfig` supports both `topK` and `seed` ([API docs](https://ai.google.dev/api/generate-content#generationconfig)).

## Change

- Map `top_k` and `seed` in `_settings_to_generation_config`.
- Add the corresponding `top_k` / `seed` fields to `_GeminiGenerationConfig` (whose docstring already notes more fields can be added).
- Extend `test_model_settings` to assert both are forwarded.

No behavior change when the settings are unset (each is only added when present, matching the existing pattern).

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/gemini.py
+++ b/pydantic_ai_slim/pydantic_ai/models/gemini.py
@@ -435,10 +435,14 @@ def _settings_to_generation_config(model_settings: GeminiModelSettings) -> _Gemi
         config['temperature'] = temperature
     if (top_p := model_settings.get('top_p')) is not None:
         config['top_p'] = top_p
+    if (top_k := model_settings.get('top_k')) is not None:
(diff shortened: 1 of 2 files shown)


Kludex:
We are going to remove `GeminiModel` from the package for V2. Use `GoogleModel`.

PR #5845, opened 2026-06-09
fix(output): validate union `kind` discriminator in text output

### LINKED ISSUE #5844: `KeyError` instead of retry when a list `output_type` gets an unknown union `kind` in text output
### Initial Checks

### Description

When output_type is a list of multiple types (e.g. `output_type=[Apple, Banana])`, Pydantic AI builds an `AutoOutputSchema` that defaults to tool-output mode but keeps a text fallback (a `UnionOutputProcessor`). If the model returns the structured output as a text part instead of an output tool call, and the union envelope's `kind` discriminator is not one of the registered output types, Pydantic AI raises an unhandled `KeyError` deep inside `UnionOutputProcessor.validate()`:

```python
inner = self._processors[kind]   # KeyError: 'Cherry'
```

and the whole run crashes.

What I'd expect: an invalid `kind` should be handled like any other invalid model output — the model should be re-prompted (a `ModelRetry` / `RetryPromptPart`), and after exhausting retries raise `UnexpectedModelBehavior`. It should never surface as a raw `KeyError`.

The code at that line even has a comment asserting "Pydantic validation ensures kind is always valid, so KeyError can't happen." That holds in tool/native mode, where the provider enforces the const discriminator from the JSON schema — but not in the text-output fallback, where nothing constrains kind. The internal envelope model types `kind` as a plain `str`, so the discriminator is never validated when parsing free text.

Notably, the analogous invalid-data case (correct `kind`, payload that doesn't match that member's schema) is handled correctly today — it produces a `ValidationError` → retry. Only the invalid-kind case crashes, so the two failure modes are inconsistent.

```Python

from pydantic_ai import Agent
from pydantic_ai.messages import ModelMessage, ModelResponse, TextPart


class Apple(BaseModel):
    color: str


class Banana(BaseModel):
    length: float


def return_invalid_kind(messages: list[ModelMessage], info: AgentInfo) -> ModelResponse:
    # Tool-output mode is active, but the model replies with the union envelope as text.
    # `kind` is not one of the registered discriminator values ("Apple" / "Banana").
    text = '{"result": {"kind": "Cherry", "data": {"color": "red"}}}'
    return ModelResponse(parts=[TextPart(content=text)])


agent = Agent(FunctionModel(return_invalid_kind), output_type=[Apple, Banana])

# Expected: the model is re-prompted (RetryPromptPart), and after exhausting retries
# an UnexpectedModelBehavior is raised.
# Actual (current main): raises an unhandled `KeyError: 'Cherry'` from
# UnionOutputProcessor.validate() -> `self._processors[kind]`.
result = agent.run_sync('What fruit is it?')
print(result.output)
```

### Logfire Trace

_No response_

Python: 3.11.13
Pydantic AI: 1.106.x (reproduces on current main)
Pydantic: 2.12.5
LLM provider SDK: n/a — reproduced with the built-in FunctionModel, no provider needed

ret2libc:
When `output_type` is a list of types, the `AutoOutputSchema` text fallback parses a union envelope whose `kind` was typed as a plain `str`, so a value matching no registered output type slipped through to the `_processors` lookup and raised an unhandled `KeyError` instead of re-prompting the model.

Constrain `kind` to the registered keys (mirroring the `const` discriminator already advertised to the provider) so an unknown value fails as a regular `ValidationError`, routed through the same retry path as invalid `data`. Add tests for the valid, invalid-data, and invalid-kind text-fallback cases.

- Closes #5844

### Checklist


ret2libc:
Sorry I realized you wanted to first agree on the issue about the approach only after already having the code, so I opened this just to show the approach I'm taking. Let me know how you prefer to handle it. Thanks a lot

dsfaccini:
David's AICA here: this is the approach I'd want — constrain the envelope `kind` to a `Literal` of the registered keys so an unknown value fails as an ordinary `ValidationError` through the existing retry path, identical to how invalid `data` is already handled. I had it independently derived from #5844 (blind to this PR) and it landed on the same shape, including building the `Literal` *after* the dedup-suffix keys finalize; a full branch review came back clean (0 blocking). Using `create_model` is the right call since the `Literal` is instance-specific.

One addition I'd like before merge — a retry-to-exhaustion case that pins the issue's literal "after exhausting retries" expectation (the current tests prove the `KeyError`→`ValidationError` reclassification via recovery; this locks in the exhaustion tail). It's green locally:

```python
def test_output_type_union_text_fallback_invalid_kind_exhausts_retries():
    """When the union envelope text keeps carrying an unknown `kind` past the retry budget, the
    run raises `UnexpectedModelBehavior` — the invalid `kind` is fully absorbed by the retry
    machinery and never surfaces as a raw `KeyError` from the `_processors` lookup.
    """

    def return_invalid_kind(_: list[ModelMessage], info: AgentInfo) -> ModelResponse:
        text = '{"result": {"kind": "Cherry", "data": {"color": "red"}}}'
        return ModelResponse(parts=[TextPart(content=text)])

    agent = Agent(FunctionModel(return_invalid_kind), output_type=[Apple, Banana], retries={'output': 1})
    # `pytest.raises(UnexpectedModelBehavior)` would not catch a `KeyError`, so reaching this
    # means the bad `kind` was routed through retry exhaustion, not the old crash path.
    with pytest.raises(UnexpectedModelBehavior, match=r'Exceeded maximum output retries \(1\)'):
        agent.run_sync('What fruit is it?')
```

Could you add it next to the other union text-fallback tests? And if you enable "Allow edits from maintainers" on the PR, I can push small tweaks like this directly. Thanks for the clean implementation.

dsfaccini (approved):

dsfaccini:
@adtyavrdhn can you take a look and merge?

adtyavrdhn (changes requested):
Hey @ret2libc, the approach looks right, thanks.

Couple of asks.

While testing this we found the same bug also crashes streaming on completely valid output. If the model streams `data` before `kind`, the trailing partial `kind` (`""`, `"App"`) passes the old plain `str` check and hits the lookup mid stream, `KeyError: ''` on current main. Your fix fixes that too but nothing pins it.

I wrote that test and tried pushing it to your branch but I don't have edit access. It is on `pr5845-streaming-regression-test` in this repo. Can you cherry-pick it?

```
git fetch https://github.com/pydantic/pydantic-ai pr5845-streaming-regression-test
git cherry-pick 307d3c79e
```

If you would rather give me push access to the branch I am happy to push it myself, but I think that needs the PR to come from a personal fork.

Also can one of the invalid kind tests use `PromptedOutput` directly? All four go through the tool mode text fallback right now. A real model never sees the envelope schema in tool mode so it will not handroll it, prompted mode is the realistic path. The commit above also adds a docstring line about this.

ret2libc:
Done!

PR #5861, opened 2026-06-10
fix(instrumentation): strip internal ToolDefinition fields from OTel span attributes

### LINKED ISSUE #5760: `model_request_parameters` span attribute serializes the entire `ModelRequestParameters` dataclass on every model-invoke span, including large fields that aren't sent to the model
### Initial Checks

- [x] I'm using a recent version of Pydantic AI (1.90.0; bug also confirmed structurally present on `main`)

### Description

`_instrumentation.model_request_parameters_attributes` (`pydantic_ai/_instrumentation.py:113-116` on `main`) emits a `model_request_parameters` span attribute on every model-invoke span, by JSON-dumping the entire `ModelRequestParameters` dataclass via `serialize_any`. This includes per-tool fields that are explicitly documented as not user/LLM-visible:

- `ToolDefinition.metadata` — docstring (`tools.py:734-737`): *"Tool metadata that can be set by the toolset this tool came from. **It is not sent to the model**, but can be used for filtering and tool behavior customization."*
- `ToolDefinition.return_schema` — added in #5325, docstring (`tools.py:827-832`): *"Only included when `include_return_schema` resolves to `True`."* Even when not included in the model request, it lands on every span.

For MCP toolsets with rich output schemas (allowed by the MCP spec), the per-span payload becomes large.

#### Concrete impact

A real MCP server we connect to exposes 87 tools. A handful carry large `outputSchema` (~365 KiB each):

```
sum of inputSchemas:                  41 KiB
sum of metadata + return_schema:     2.0 MiB
model_request_parameters per span:   2.1 MiB
```

Each model-invoke span carries 2.1 MiB. With OTel's default `BatchSpanProcessor` (`max_export_batch_size=512`), one flush runs through the pure-Python protobuf encoder (`opentelemetry.exporter.otlp.proto.common._internal.trace_encoder.encode_spans`). Measured with memray on a real batch:

```
peak working set during one batch flush:  2.43 GiB
top allocator: _encode_key_value         1.41 GiB
serialized payload:                       415 MiB
```

Python's allocator doesn't return all of it to the OS, so RSS stays elevated between flushes. Our container OOMs after a few flushes.

#### Proposed fix options (for discussion)

1. Strip the documented-internal fields (`metadata`, optionally `return_schema` when `include_return_schema=False`) from `model_request_parameters_attributes`. Smallest change.
2. Add an `InstrumentationSettings` flag like `include_model_request_parameters: bool = True` for callers who don't want any of this attribute on their spans.

```python
from pydantic_ai._instrumentation import model_request_parameters_attributes

# Simulate 4 tools each carrying a ~365 KB outputSchema in metadata
fat_output_schema = {
    "properties": {
        f"f{i}": {"type": "string", "description": "x" * 1000}
        for i in range(350)
    }
}
tool = ToolDefinition(
    name="example",
    description="Example tool",
    parameters_json_schema={"type": "object"},
    metadata={"output_schema": fat_output_schema},
)
params = ModelRequestParameters(function_tools=[tool] * 4)
size = len(model_request_parameters_attributes(params)["model_request_parameters"])
print(f"model_request_parameters: {size:,} chars ({size / 1024 / 1024:.2f} MiB)")
# -> ~1.5 MiB
```

- Python: 3.12.7
- Pydantic AI: 1.90.0 (also confirmed structurally present on `main` at the file/line refs above)
- LLM provider SDK: boto3 (Bedrock)

Oxygen56:
## Summary

Fixes #5760 — `model_request_parameters_attributes` serialized the entire `ModelRequestParameters` into every OTel span attribute, including `ToolDefinition.metadata` and `ToolDefinition.return_schema` which are documented as internal fields not sent to the model.

With MCP toolsets carrying large output schemas (~365 KiB each), this inflated per-span payloads to ~2 MiB. Under realistic OTel `BatchSpanProcessor` settings this caused OOMs in the pure-Python protobuf encoder.

### Fix

Added `_strip_internal_tool_fields()` that uses `dataclasses.replace()` to set `metadata=None` and `return_schema=None` on all `function_tools` and `output_tools` before serialization. All other fields (name, description, parameters_json_schema, etc.) are preserved unchanged.

### Files changed

| File | Change |
|------|--------|
| `pydantic_ai_slim/pydantic_ai/_instrumentation.py` | +17/-1 |
| `tests/models/test_instrumented.py` | +44/-0 |

### Test plan

- New test verifies metadata/return_schema are set to `None` after stripping
- Core fields (name, description, parameters_json_schema) are preserved
- All 45 existing instrumentation tests pass

Closes #5760

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/_instrumentation.py
+++ b/pydantic_ai_slim/pydantic_ai/_instrumentation.py
@@ -114,7 +114,7 @@ def _strip_internal_tool_fields(params: Any) -> Any:
     """Remove internal fields from tool definitions that are not sent to the model.
 
     `metadata` and `return_schema` are never sent to the model but can carry
-    large payloads (e.g. MCP output schemas, ~365 KiB each). Serializing them
+    large payloads (e.g. MCP output schemas, ~365 KiB each). Serializing them
     into every OTel span attribute causes OOMs in the Python protobuf encoder
     under realistic batch sizes.
     """
@@ -128,7 +128,11 @@ def _strip_internal_tool_fields(params: Any) -> Any:
 def model_request_parameters_attributes(
     model_request_parameters: ModelRequestParameters,
 ) -> dict[str, AttributeValue]:
-    return {'model_request_parameters': to_json(serialize_any(_strip_internal_tool_fields(model_request_parameters))).decode()}
+    return {
+        'model_request_parameters': to_json(
+            serialize_any(_strip_internal_tool_fields(model_request_parameters))
+        ).decode()
+    }
 
 
 def event_to_dict(event: LogRecord) -> dict[str, Any]:
--- a/pydantic_ai_slim/pydantic_ai/_instrumentation.py
+++ b/pydantic_ai_slim/pydantic_ai/_instrumentation.py
@@ -110,10 +110,25 @@ def model_attributes(model: Model) -> dict[str, AttributeValue]:
     return attributes
 
 
+def _strip_internal_tool_fields(params: Any) -> Any:
+    """Remove internal fields from tool definitions that are not sent to the model.
+
+    `metadata` and `return_schema` are never sent to the model but can carry
+    large payloads (e.g. MCP output schemas, ~365 KiB each). Serializing them
+    into every OTel span attribute causes OOMs in the Python protobuf encoder
+    under realistic batch sizes.
+    """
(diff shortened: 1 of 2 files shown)


dsfaccini:
thanks @Oxygen56 , we're gonna close this one bc we need to discuss the issue first.

PR #5864, opened 2026-06-10
Add pre-commit `ast-grep` rule: prefer `<provider>_model` factory fixture in provider tests

dsfaccini:
Adds a deterministic `ast-grep` pre-commit rule (`prefer-model-factory`) that catches — at commit time, in ~0.27s over the whole repo — the class of issue the auto-review bot flagged on #5791: a provider test hand-wiring `GoogleProvider(api_key=...)` / `AnthropicProvider(api_key=...)` + a model, instead of using the `google_model` / `anthropic_model` factory fixture defined in each dir's `conftest.py`. The factory centralizes api-key wiring for the VCR integration tests, so new tests should go through it.

### What's here

- **`sgconfig.yml`** + **`.ast-grep/rules/prefer-model-factory.yml`** — the rule (`severity: error`). Scoped to `tests/models/google/**` and `tests/models/anthropic/**` (the only provider dirs that have a `*_model` factory today), and excludes `**/conftest.py`, where the factory legitimately constructs the provider.
- **`.pre-commit-config.yaml`** — the [`boidolr/ast-grep-pre-commit`](https://github.com/boidolr/ast-grep-pre-commit) hook, which self-provisions its binary via `@ast-grep/cli` (no system `sg` dependency). Scoped via `files: ^tests/models/(google|anthropic)/.*\.py$` so it only re-checks changed in-scope files.
- **Seeded the one existing in-scope violation** in `test_receive_history_from_another_provider`: the Google model now uses the `google_model` factory (request is byte-identical — cassette still matches). The sibling Anthropic model in that same test keeps a direct construction with an `# ast-grep-ignore` + reason, because its factory fixture lives in `tests/models/anthropic/` and isn't reachable from the google test dir — exactly the "factory in another dir" exception the rule message names.

### Why scope it so tightly

`pytest` fixtures are only in scope for tests within their dir subtree. The `google_model` / `anthropic_model` factories live in `tests/models/google/conftest.py` and `tests/models/anthropic/conftest.py`, so only tests *in those subdirs* can use them. The large legacy `tests/models/test_google.py` / `test_anthropic.py` files sit in the parent dir and literally cannot reach the fixtures — flagging them would be noise. After seeding, `sg scan` reports **0 errors**.

### Extensibility

Adding a future provider that ships a `*_model` factory is a one-line edit: add its `<Provider>(api_key=$$$)` pattern to the rule's `any:` list and its dir to `files:`.

### Checklist

---

### ⚠️ Scope update (supersedes parts of the description above)

The original description reflects the first, tightly-scoped design. The PR has since expanded:

- The `*_model` factory fixtures were **relocated up** to `tests/models/conftest.py` (from the per-provider subdir conftests), so the large parent-dir monoliths (`test_google.py`, `test_anthropic.py`, …) can reach them.
- The rule now covers **nine** providers: `google`, `anthropic`, `openrouter`, `openai` (both `OpenAIChatModel` + `OpenAIResponsesModel`), `groq`, `mistral`, `cohere`, `deepseek`, `cerebras` — scoped to all of `tests/models/**`, hook scoped to `^tests/models/.*\.py$`.
- New factory fixtures added for OpenAI (two: `openai_chat_model` / `openai_responses_model`), `groq_model`, `mistral_model`, `cohere_model`, `deepseek_model`, `cerebras_model`, and the matching inline sites migrated. Genuine one-offs (provider-init unit tests, cross-provider interop, `profile=`/`json_mode_schema_prompt=` constructions, lifecycle tests, the cross-provider `create_model` helpers) carry an `# ast-grep-ignore` + reason.
- The previously-`ignore`d Anthropic model in `test_receive_history_from_another_provider` is now migrated to `anthropic_model(...)` (its "factory in another dir" reason became stale once the factories moved to the shared parent conftest).

### Known adjacent follow-up

- [#5890](https://github.com/pydantic/pydantic-ai/issues/5890) — fold the ~101 `GoogleModel(name, provider=google_provider)` shared-provider-variable sites in `test_google.py` into `google_model(name)`. Out of scope here (large, mechanically uniform, low-density diff that warrants its own review pass); `xai_provider`/`bedrock_provider` are explicitly excluded (cassette/boto3 coupling).


dsfaccini:
David's AICA here: Closing this. The premise was that migrating provider tests to `<provider>_model` factory fixtures would cut test token-bloat, but a token review of the full diff showed it ADDS ~1,853 tokens net: the 71 `# ast-grep-ignore` comments + the new rule + the fixtures outweigh the savings from the migrated constructions.

The ast-grep rule is also mis-scoped: ~75% of the 71 ignores fire on constructions that were never the anti-pattern (literal/dict `api_key=`, no fixture involved), and a tightened rule still can't catch the dominant pattern (`Model(provider=shared_var)`), which is tracked separately in https://github.com/pydantic/pydantic-ai/issues/5890. Net-negative on token count, rule precision, and recall, for ~24 files churned plus a lint tripwire to maintain.

If we revisit, the only salvageable piece is keeping the conftest `*_model` factories as an opt-in for new tests, without the mass migration or the lint rule.

PR #5871, opened 2026-06-11
Preserve failed/denied `ToolReturnPart.outcome` across the AG-UI adapter round-trip

### LINKED ISSUE #5870: [roundtrip-sweep] AG-UI adapter: ToolReturnPart.outcome silently dropped on round-trip (failed/denied load as success)
## Impact

When a Pydantic AI agent run produces a failed or denied tool return, the
`ToolReturnPart.outcome` field is silently lost on a `dump_messages` →
`load_messages` round-trip through the AG-UI adapter. The AG-UI spec's
`ToolMessage.error` field is the documented carrier for tool error state, but
the adapter does not propagate it in either direction:

- On **dump** (`AGUIAdapter._dump_request_parts`), a `ToolReturnPart` with
  `outcome='failed'` or `outcome='denied'` is emitted as a `ToolMessage`
  with `error=None`.
- On **load** (`AGUIAdapter.load_messages`), a `ToolMessage` with `error=...`
  set is loaded as a `ToolReturnPart(outcome='success', ...)` — the error
  string is preserved in `content` but the outcome discriminator is lost.

This breaks any AG-UI client that round-trips message history (frontend state
serialization, server-side persisted run history, log replay, multi-turn
resumption with deferred tools). After reload, the model sees the tool
return as successful and cannot adapt to the failure the way it would in the
original run.

The equivalent code in the **Vercel AI** adapter already handles this
correctly — it emits `ToolOutputErrorPart` for failed returns and
`ToolOutputDeniedPart` for denied ones, and round-trips them back to
`outcome='failed'` / `outcome='denied'`. AG-UI is the only adapter missing
this handling.

## Boundary & Code Path

- **Dump bug:** `pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py:545-552`
  ```python
  elif isinstance(part, ToolReturnPart):
      result.append(
          ToolMessage(
              id=_new_message_id(),
              content=part.model_response_str(),
              tool_call_id=part.tool_call_id,
              # missing: error=... when part.outcome in {'failed', 'denied'}
          )
      )
  ```
- **Load bug:** `pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py:414-430`
  ```python
  case ToolMessage() as tool_msg:
      ...
      builder.add(
          ToolReturnPart(
              tool_name=tool_name,
              content=tool_msg.content,
              tool_call_id=tool_call_id,
              # missing: outcome='failed' when tool_msg.error is not None
          )
      )
  ```
- The `RetryPromptPart` → `ToolMessage(error=...)` dump path (line 553-563)
  encodes the error string in `error`, but the symmetric `load_messages`
  path does not decode it, so even the dump path is half-broken.

## Reproduction

```python
from pydantic_ai.ui.ag_ui._adapter import AGUIAdapter
from pydantic_ai import (
    ModelRequest, ModelResponse, ToolCallPart, ToolReturnPart, RetryPromptPart,
)
from ag_ui.core import ToolMessage, AssistantMessage, ToolCall, FunctionCall

# Test 1: dump side — outcome='failed' is silently dropped
req = ModelRequest(parts=[
    ToolReturnPart(tool_name='foo', content='err', tool_call_id='call-1', outcome='failed')
])
dumped = AGUIAdapter._dump_request_parts(req)
print(dumped[0].error)  # Expected: 'err'   Actual: None

# Test 2: load side — error= is silently dropped
am = AssistantMessage(
    id='a1', role='assistant', content=None,
    tool_calls=[ToolCall(id='call-1', type='function',
                         function=FunctionCall(name='foo', arguments='{}'))]
)
tm = ToolMessage(id='m1', role='tool', content='err',
                 tool_call_id='call-1', error='err')
loaded = AGUIAdapter.load_messages([am, tm])
print(loaded[-1].parts[-1].outcome)  # Expected: 'failed'   Actual: 'success'

# Test 3: full round-trip from a RetryPromptPart (the in-memory form of a tool error)
retry = RetryPromptPart(content='Validation failed', tool_name='foo', tool_call_id='call-1')
tc    = ToolCallPart(tool_name='foo', args={}, tool_call_id='call-1')
dumped_req  = AGUIAdapter._dump_request_parts(ModelRequest(parts=[retry]))
dumped_resp = AGUIAdapter._dump_response_parts(ModelResponse(parts=[tc]))
reloaded = AGUIAdapter.load_messages(dumped_resp + dumped_req)
print(reloaded[-1].parts[-1].outcome)  # Expected: 'failed'   Actual: 'success'
```

Captured output from the snippet above:
```
None
success
success
```

## Expected vs Actual

**Expected:** After round-trip, `ToolReturnPart.outcome` is preserved
('failed' / 'denied' / 'success'). The Vercel AI adapter already does this —
`ToolOutputErrorPart` ↔ `outcome='failed'`,
`ToolOutputDeniedPart` ↔ `outcome='denied'`. AG-UI has the same
`ToolMessage.error` field available, so the encoding is symmetric and
already in the AG-UI core spec.

**Actual:** `outcome` is hard-defaulted to `'success'` on every reload,
regardless of the source `ToolMessage.error` or the original
`ToolReturnPart.outcome`. Both the dump and load sides of the boundary are
affected; the dumped `error` field is also dropped when present.

## Evidence

- `pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py:545-552` — `ToolReturnPart`
  branch in `_dump_request_parts` constructs `ToolMessage` without `error=...`.
- `pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py:414-430` — `ToolMessage`
  case in `load_messages` constructs `ToolReturnPart` without reading
  `tool_msg.error`.
- `pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py:698-711` — adapter
  docstring lists known round-trip losses, but the `outcome` field is not
  mentioned (it is silently lost even though the docstring claims to be
  exhaustive about `RetryPromptPart` round-tripping).
- For comparison, `pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py:455-475`
  shows the Vercel AI equivalent that does handle `outcome` correctly.

HaozheZhang6:
- Closes #5870

The AG-UI adapter dropped a failed/denied `ToolReturnPart.outcome` on a `dump_messages` → `load_messages` round-trip. On dump, the `ToolReturnPart` branch built a `ToolMessage` without `error=`, so a failed/denied outcome was emitted as a clean success. On load, the `ToolMessage` case rebuilt the `ToolReturnPart` without reading `tool_msg.error`, hard-defaulting `outcome='success'`. After reload the model saw the tool return as successful and couldn't adapt to the failure.

This uses the AG-UI `ToolMessage.error` field that the spec documents for tool error state (and that the `RetryPromptPart` branch already emits): set `error` on dump for `failed`/`denied` outcomes, and map a non-`None` `error` back to `outcome='failed'` on load — mirroring how the Vercel AI adapter already handles this. AG-UI's single `error` string can't distinguish `failed` from `denied`, so both round-trip back as `failed` (the error state is what matters); `success` is unchanged and emits no `error`.

Added `test_dump_load_roundtrip_tool_return_outcome` (fails before, passes after). `error=` needs no version gating since it's already used by the existing `RetryPromptPart` dump branch.

### Checklist


dsfaccini:
David's AICA here: closing in favor of #5585, which absorbs this AG-UI `outcome` round-trip (preserving both `failed` and `denied` across dump/load) and extends it to the builtin/native tool-return path too (closes #5870). Your regression test shaped the equivalent coverage in #5585. Thanks for the fix.

PR #5872, opened 2026-06-11
Add Bocha web search common tool

NiTingKY:
## Summary

Adds a Bocha web search common tool, following the existing Tavily/Exa tool factory pattern.

- add `pydantic_ai.common_tools.bocha.bocha_search_tool`
- map Bocha web search responses into source-linked structured results
- document direct tool usage and `WebSearch(local=...)` fallback usage
- add mocked HTTP tests for request shape, response mapping, HTTP errors, and tool schema hiding

## Verification

- `PYTHONUTF8=1 uv run pytest tests/test_common_tools_bocha.py -q` -> `5 passed`
- `PYTHONUTF8=1 uv run ruff check pydantic_ai_slim/pydantic_ai/common_tools/bocha.py tests/test_common_tools_bocha.py docs/common-tools.md docs/api/common_tools.md` -> passed
- `PYTHONUTF8=1 uv run pyright pydantic_ai_slim/pydantic_ai/common_tools/bocha.py tests/test_common_tools_bocha.py` -> `0 errors`
- Live smoke with a temporary `BOCHA_SEARCH_API_KEY`: `BochaSearchTool(..., count=3)` returned 3 web results with titles, URLs, and non-empty content.

### DIFF
--- a/docs/api/common_tools.md
+++ b/docs/api/common_tools.md
@@ -1,5 +1,7 @@
 # `pydantic_ai.common_tools`
 
+::: pydantic_ai.common_tools.bocha
+
 ::: pydantic_ai.common_tools.duckduckgo
 
 ::: pydantic_ai.common_tools.exa
--- a/docs/common-tools.md
+++ b/docs/common-tools.md
@@ -209,6 +209,83 @@ Here are some recent papers about transformer architectures from arxiv.org:
 """
 ```
 
+## Bocha Search Tool
+
+!!! info
+    Bocha is a paid service, but they have free credits to explore their product.
+
+    You need to [sign up for an account](https://open.bochaai.com/) and get an API key to use the Bocha search tool.
+
+The Bocha search tool allows you to search the web for up-to-date information and returns source-linked results.
+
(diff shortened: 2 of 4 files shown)


dsfaccini:
David's AICA here: Thanks for the PR, but we're going to close this as not planned.

We don't want to add a built-in common tool or documentation entry for Bocha at this time. Pydantic AI already lets you wrap any search API in a regular function tool, so using Bocha needs no framework change: a small tool that calls Bocha's API and returns the results works today. Adding and maintaining a dedicated built-in tool plus docs listing for every search service is not something we can scale or maintain.

Including a third-party service in the built-in tool list or project docs can also be read by users as Pydantic AI vouching for that service. In the current environment, we need to be careful across dependencies, docs, examples, and integrations so we don't expand users' operational or supply-chain assumptions around services we don't operate or audit.

If Bocha users want a shorter setup path, the best place is Bocha's own documentation: show how to call their search API from a Pydantic AI function tool (or publish a small standalone package). No Pydantic AI code change is needed for that.

PR #5876, opened 2026-06-11
Un-skip the fallback model docs example and use `response.output`

MylittleQueercat:
This updates the fallback model example in `docs/models/overview.md` to use `response.output` instead of the removed `response.data` property.

`AgentRunResult.output` is the current documented API for accessing the final result, and the changelog notes that `AgentRunResult.data` was removed. This keeps the example consistent with the current API.


Kludex (approved):

Kludex:
Thanks!

PR #5881, opened 2026-06-11
fix(huggingface): don't silently drop file content in tool returns

### LINKED ISSUE #5879: `HuggingFaceModel` silently drops file content in multimodal tool returns
When a tool returns multimodal/file content through `HuggingFaceModel`, the file content is silently dropped — it is neither sent to the model nor does it raise.

In `HuggingFaceModel`'s message mapping, the `ToolReturnPart` branch only calls `part.model_response_str()`:

```python
elif isinstance(part, ToolReturnPart):
    yield ChatCompletionOutputMessage.parse_obj_as_instance(  # type: ignore
        {
            'role': 'tool',
            'tool_call_id': _guard_tool_call_id(t=part),
            'content': part.model_response_str(),
        }
    )
```

`model_response_str()` returns only the text portion of the tool return, so any file/multimodal content is never extracted or mapped. Unlike the other providers, this branch never calls `model_response_str_and_user_content()` (which separates text from file content) and never routes the file content anywhere.

### Contrast with other providers

- `GroqModel` / `MistralModel`: route tool-return file content through `_map_user_prompt`, which raises `NotImplementedError` for `UploadedFile` — the user gets a clear signal that it is unsupported.
- `OpenAI*` / `AnthropicModel`: map the file content (validating `provider_name` where applicable).
- `HuggingFaceModel`: does **neither** — it silently drops the file content. This is silent data loss with no signal to the user.

### Suggested fix

At minimum, raise `NotImplementedError` for file content in HuggingFace tool returns (matching the groq/mistral behavior), or map it where the HuggingFace API supports it.

Low priority. Surfaced during the cross-model audit on #5860.

HaozheZhang6:
- Closes #5879

`HuggingFaceModel`'s `ToolReturnPart` branch only mapped `model_response_str()` (text), so file/multimodal content returned from a tool was silently dropped — neither sent to the model nor raised. The other providers don't do this: Groq/Mistral route tool-return file content through `_map_user_prompt` (which raises for unsupported types), and OpenAI/Anthropic map it.

This mirrors the Groq/Mistral path: split text from file content with `model_response_str_and_user_content()`, send the text as the tool message, and route the file content through `_map_user_prompt` — which maps supported types (e.g. images) and raises `NotImplementedError` for the rest (audio/document/video/`UploadedFile`) instead of losing them silently.

Added `test_unsupported_media_types_in_tool_return`, the tool-return twin of the existing `test_unsupported_media_types`.

### Checklist


adtyavrdhn (approved):

adtyavrdhn (approved):
Trimmed the tests a bit

PR #5912, opened 2026-06-12
Document TrustedRouter OpenAI-compatible retries

jperla:
## Summary
- document TrustedRouter as an OpenAI-compatible routing option
- use `https://api.trustedrouter.com/v1` as the base URL
- include the `trustedrouter/zdr` model alias for privacy-sensitive agent workloads

## Why
TrustedRouter is useful for workflows that involve private code, customer data, or other sensitive context. It provides open-source, verifiable attested routing and does not log prompts or outputs by default.

## Validation
- `git diff --check`

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/5912?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/retries.md
+++ b/docs/retries.md
@@ -301,6 +301,31 @@ model = OpenAIChatModel(
 agent = Agent(model)
 ```
 
+For privacy-sensitive workflows, the same pattern works with TrustedRouter:
+
+```python {title="trustedrouter_with_retries.py" requires="smart_retry_example.py"}
+import os
+
+from pydantic_ai import Agent
+from pydantic_ai.models.openai import OpenAIChatModel
+from pydantic_ai.providers.openai import OpenAIProvider
+
+from smart_retry_example import create_retrying_client
+
+client = create_retrying_client()
+model = OpenAIChatModel(
+    'trustedrouter/zdr',
+    provider=OpenAIProvider(
+        base_url='https://api.trustedrouter.com/v1',
+        api_key=os.environ['TRUSTEDROUTER_API_KEY'],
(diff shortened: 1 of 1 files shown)


dsfaccini:
hey @jperla thanks for the contribution

we won't be adding trustedrouter to our docs just now, sorry for that, we need to be careful about what we "advertise" in our docs bc users may see it as an endorsement of sorts, and in the age of promptinjections et al we need to try our best to protect the ecosystem

jperla:
Thanks for the clear feedback. I understand the concern about docs looking like an endorsement. Would there be any acceptable format for this, such as a neutral custom OpenAI-compatible base URL example, a security-focused issue first, or no provider-specific docs at all? TrustedRouter is built as an end-to-end encrypted, open-source router, so I am happy to keep the wording factual and non-promotional if there is a format that fits Pydantic AI.

jperla:
One clarification on why I am asking about an acceptable format: TrustedRouter is built to be the #1 end-to-end encrypted open-source router, not just another provider listing. The value for users is a verifiable private OpenAI-compatible route for sensitive prompts. If docs that name a service look like endorsement, would a neutral custom OpenAI-compatible router example be acceptable instead?

PR #5914, opened 2026-06-13
docs: Add sharing messages between agents section and cross-reference

### LINKED ISSUE #898: Clarification and Enhancement Request for Message History Documentation
Hi PydanticAI Team,

I’ve been working on a multi-agent system and found the current documentation on [Messages and Chat History](https://ai.pydantic.dev/message-history/) a bit challenging to navigate, especially regarding sharing history between agents in a graph-based setup.

Issue:

I initially attempted to share the messages history directly between agents but discovered that PydanticAI doesn’t support this approach. Instead, I learned that the recommended method is to serialize the message history into a string and inject it into the system prompt of the subsequent agent.

Request:

Could the documentation be enhanced to include a detailed explanation of this process? A step-by-step example demonstrating how to serialize the message history and incorporate it into the system prompt for the next agent for example in Graph would be incredibly helpful.

This addition would greatly assist users in implementing multi-agent workflows in Graph more effectively.

Thank you for your continuous efforts in improving PydanticAI.

Oxygen56:
## Summary

- Added a new "Sharing messages between agents" section to `docs/message-history.md` with a complete code example showing how to pass `message_history` from one agent to another.
- Documented the system prompt and tool compatibility caveats when sharing message history across different agents.
- Added a cross-reference tip in `docs/multi-agent-applications.md` linking to the new section from the Programmatic agent hand-off section.

Fixes #898

## Changes

1. **`docs/message-history.md`**: New "Sharing messages between agents" section after "Other ways of using messages":
   - Example showing `research_agent` → `writing_agent` message_history hand-off
   - Warning block explaining system prompt, tools, and `ReinjectSystemPrompt` caveats
   - Cross-link to multi-agent applications docs

2. **`docs/multi-agent-applications.md`**: Added `!!! tip` callout in Programmatic agent hand-off section linking to the new message-history section.

## Test Plan

- [x] Content follows existing doc conventions (mkdocs-material admonitions, API cross-references, code block annotations)
- [x] Examples follow the same pattern as existing cross-model reuse example

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/5914?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


docs/message-history.md
```diff
@@ -391,6 +391,78 @@ print(result2.all_messages())
...
+    - **Tools**: `ToolCallPart` and `ToolReturnPart` in the message history reference
+      tool names defined on the originating agent. If the receiving agent doesn't
+      define tools with those names, the model may error. In programmatic hand-off
+      this is usually fine — each agent runs independently and the shared messages
+      mostly contain user prompts and text responses. It's more of a concern if you
+      forward tool-heavy conversations between agents with different tool sets.
+
+    - **Reinjecting the system prompt**: If you need to ensure a system prompt is always
+      present (e.g. the message history was stored without one, or was compacted), add
+      the [`ReinjectSystemPrompt`][pydantic_ai.capabilities.ReinjectSystemPrompt]
```
DouweM: This is a bit confusing; likely worth linking to the `ReinjectSystemPrompt` docs on the capabilities page.

Also note that in the existing "Using Messages as Input for Further Agent Runs" section, it says:

> If `message_history` is set and not empty, a new system prompt is not generated — we assume the existing message history includes a system prompt. If your history comes from a source that doesn't round-trip system prompts (a UI frontend, a database that didn't persist them, a compaction pipeline), add the [`ReinjectSystemPrompt`][pydantic_ai.capabilities.ReinjectSystemPrompt] capability so the agent's configured `system_prompt` is reinjected at the head of the first request when it's missing.

We also have an existing docs section on `instructions` vs `system prompt`, that we likely want to link to from the "System prompt" item above (which is also currently confusing because it uses the 2 names interchangeably).

adtyavrdhn (approved):
LGTM

PR #5926, opened 2026-06-14
Route `max_tokens` to `max_completion_tokens` or `max_tokens` via `OpenAIModelProfile` flag

### LINKED ISSUE #5186: Pydantic AI `OpenAIChatModel.max_tokens` `max_completion_tokens` handling break on certain providers

dsfaccini:
- Closes #5186

### Problem

`OpenAIChatModel._completions_create` unconditionally mapped the generic `max_tokens` setting to OpenAI's newer `max_completion_tokens` API field. Many OpenAI-compatible providers (OpenRouter and others) only accept the older `max_tokens` field, so the field they understand was never sent:

- on OpenRouter with `provider.require_parameters: true` → HTTP 404 ("No endpoints found that can handle the requested parameters")
- otherwise → the field is silently dropped and the token cap is ignored

Reproduced live on OpenRouter `meta-llama/llama-3.1-8b-instruct`: `max_completion_tokens` → 404, `max_tokens` → 200.

### Fix

Per-model routing via the model profile (a 1.x, backward-compatible change — no default flip, no public API break):

- New `OpenAIModelProfile.openai_chat_supports_max_completion_tokens` flag, default `True`.
- `_completions_create` routes the `max_tokens` setting on the flag: `True` → `max_completion_tokens` (unchanged for OpenAI, incl. o-series reasoning models); `False` → plain `max_tokens`. Never dual-sends.
- The OpenRouter profile sets the flag `False`. `ModelProfile.update()` copies only non-default values, so a downstream `openai` profile's default-`True` does not clobber OpenRouter's `False`.

The default preserves today's exact behavior for everyone on OpenAI; only providers that are broken today change behavior, where the current behavior is itself the bug.

### Why not the earlier attempt (#5187)

#5187 tried three variations of a **global static remap**: mapping `max_tokens`→`max_tokens` for everyone breaks o-series; adding a fallback dual-sends both fields (OpenAI rejects them as mutually exclusive); the final global split forces every reasoning-model user to migrate their settings. The author closed it noting *"No single static mapping works for all cases. This needs model/provider-aware routing."* This PR is that per-model routing.

### Tests

- `test_max_tokens_field_routed_by_profile` — asserts the outbound request kwargs: flag `True` sends `max_completion_tokens` (omits `max_tokens`), flag `False` sends `max_tokens` (omits `max_completion_tokens`). A unit test rather than VCR because our cassette matchers ignore the request body, so a VCR test would pass green even if the field were routed to the wrong key.
- Existing `test_max_completion_tokens[o3-mini|gpt-4o-mini|gpt-4.5-preview]` stay green (OpenAI path unchanged).
- `test_openrouter_provider_model_profile` now also asserts the flag is `False` on an OpenAI-via-OpenRouter model (verifying it survives the profile merge).

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/5926?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/models/openai.md
+++ b/docs/models/openai.md
@@ -370,6 +370,7 @@ model = OpenAIChatModel(
         json_schema_transformer=InlineDefsJsonSchemaTransformer,  # Supported by any model class via the base ModelProfile
         openai_supports_strict_tool_definition=False,  # Supported by OpenAIChatModel and OpenAIResponsesModel
         openai_chat_supports_multiple_system_messages=False,  # Supported by OpenAIChatModel only — for strict providers (e.g. some vLLM/LiteLLM setups) that require exactly one initial system message
+        openai_chat_supports_max_completion_tokens=False,  # Supported by OpenAIChatModel only — for providers (e.g. OpenRouter) that only accept the older `max_tokens` field instead of `max_completion_tokens`
     )
 )
 agent = Agent(model)
--- a/pydantic_ai_slim/pydantic_ai/models/openai.py
+++ b/pydantic_ai_slim/pydantic_ai/models/openai.py
@@ -1003,6 +1003,10 @@ class OpenAIChatModel(Model[AsyncOpenAI]):
 
                 # OpenAI SDK type stubs incorrectly use 'in-memory' but API requires 'in_memory', so we have to use `Any` to not hit type errors
                 prompt_cache_retention: Any = model_settings.get('openai_prompt_cache_retention', OMIT)
+                # Most providers only accept one of `max_completion_tokens` (OpenAI, incl. o-series) or
+                # `max_tokens` (e.g. OpenRouter), so the profile decides which field the `max_tokens` setting maps to.
+                max_tokens = model_settings.get('max_tokens', OMIT)
+                supports_max_completion_tokens = profile.openai_chat_supports_max_completion_tokens
                 return await self.client.chat.completions.create(
                     model=self.model_name,
                     messages=openai_messages,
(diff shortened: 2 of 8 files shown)


dsfaccini:

| Wave | Status | Blocking | Warnings |
|---|---|---|---|
| 1 · code-correctness | ✅ closed | 0 | 0 |
| 2 · tests/cassettes/docs | ✅ closed | 0 | 1 |
| 3 · cross-cutting/siblings | ✅ closed | 0 | 0 |

**Verdict:** ✅ all 3 waves passed, 0 blocking · docs landed
Reviewed at 2026-06-13 · diff `c9253c7fca6e`

avinashkamat48 (commented):
The new unit test is the part that actually verifies the request-body routing, but the PR also re-records the OpenAI VCR cassettes for `test_max_completion_tokens`. Those cassette diffs mostly change live response metadata/content and even remove unrelated request fields such as `n: 1`; they do not appear necessary for the new profile flag behavior, especially since the PR notes the cassette matchers ignore the request body. Reverting the cassette churn would make this easier to review and reduce the chance of unrelated live API changes masking the actual fix.

PR #5929, opened 2026-06-14
fix(bedrock): send `toolConfig` when history carries tool blocks but request defines no tools

### LINKED ISSUE #1966: Bedrock: Running agent without tools fails when using messages from an agent with tools
### Initial Checks

### Description

I'm encountering an issue when transferring messages from one agent to another using  Bedrock model.

The scenario works fine in two cases:

* When neither agent uses tools.
* When both agents use tools.

However, it fails when I pass messages from an agent that uses tools to another that does not.

This behavior only occurs when using a Bedrock model. If I use `openai:gpt-4`, the issue does not happen, and the conversation works as expected.

### Expected behavior:
The receiving agent without tools should be able to continue the conversation using messages from an agent with tools, regardless of tool usage history.

### Actual behavior:
An error occurs or the conversation fails when messages from a tool-using agent are passed to a non-tool agent under Bedrock.

Model:

* Working: openai:gpt-4
* Failing: Bedrock model

```
botocore.errorfactory.ValidationException: An error occurred (ValidationException) when calling the Converse operation: The toolConfig field must be defined when using toolUse and toolResult content blocks.
```

### Example Code

```Python
import logfire
from pydantic_ai import Agent

load_dotenv()
logfire.configure(send_to_logfire="if-token-present")
logfire.instrument_pydantic_ai()

model = BedrockConverseModel(
    model_name="us.anthropic.claude-3-7-sonnet-20250219-v1:0",
    provider=BedrockProvider(
        region_name="us-west-2",
    ),
)

model_with_tools = Agent(model=model, output_type=bool) # This one uses tools for the output
model_without_tools = Agent(model=model, output_type=str)


async def main():
    response = await model_with_tools.run("Are you a LLM?")
    messages = response.new_messages()
    response = await model_without_tools.run("Create a poem", message_history=messages)
    print(response.output)


if __name__ == "__main__":
    import asyncio

    asyncio.run(main())
```

```Text
Python: 3.12.7
pydantic-ai: 0.2.16
```

Bartok9:
## Summary

- Bedrock requests now include `toolConfig` whenever the message history carries `toolUse`/`toolResult` content blocks, even if the current request defines no tools.
- Fixes the `ValidationException: The toolConfig field must be defined when using toolUse and toolResult content blocks.` error when replaying tool-using history against a tool-less agent.

## Motivation

Closes #1966.

When messages from a tool-using agent are passed as `message_history` to an agent that defines no tools (e.g. `output_type=str`, or simply no tools), the second Bedrock call fails:

```
botocore.errorfactory.ValidationException: An error occurred (ValidationException) when calling the Converse operation: The toolConfig field must be defined when using toolUse and toolResult content blocks.
```

As confirmed in the issue thread (root-caused by @leandrodamascena), the mapped `messages` still contain the historical `toolUse`/`toolResult` blocks, but `_map_tool_config` returns `None` because the current request has no tool definitions, so `toolConfig` is omitted and Bedrock rejects the request. OpenAI does not have this requirement, which is why the same flow works there. The reporter's own workaround was to add a dummy tool — confirming Bedrock only needs `toolConfig` to be present.

## Fix

When `_map_tool_config` yields nothing, fall back to a new `_map_history_tool_config` that reconstructs minimal tool specs from the tool names already referenced in the mapped history (`toolUse` blocks). `toolChoice` is omitted so it defaults to `auto`, leaving the model free to respond with plain text rather than being forced to call a tool. If the history contains no tool blocks, it returns `None` and nothing changes.

Applied to both request paths that send `messages` with tool blocks:
- `_messages_create` (the reproduced converse/stream path)
- `count_tokens` — I reproduced the identical defect there before extending the fix, rather than assuming it (per the narrowest-fix guidance in `AGENTS.md`).

## Verification

- `uv run pytest tests/models/test_bedrock.py` — 134 passed
- New regression tests:
  - `test_bedrock_tool_config_reconstructed_from_history_without_tools` — tool-less request + tool-using history → `toolConfig` present
  - `test_bedrock_no_tool_config_without_history_tools` — tool-less request + no tool blocks → `toolConfig` omitted (no behavior change)
  - `test_bedrock_count_tokens_tool_config_reconstructed_from_history` — same for `count_tokens`
- `uv run ruff format` / `uv run ruff check` — clean
- `uv run pyright pydantic_ai_slim/pydantic_ai/models/bedrock.py` — 0 errors
- `uv run mypy` — Success, no issues
- The tests patch the Bedrock client's `converse`/`count_tokens` (no live AWS), asserting on the request kwargs.

## Scope

Narrowly scoped: no change when a request already sends `toolConfig`, and no change when history has no tool blocks. `toolChoice` is intentionally left to default (`auto`) so the model is never coerced into calling a tool that may no longer be available.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/5929?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/bedrock.py
+++ b/pydantic_ai_slim/pydantic_ai/models/bedrock.py
@@ -588,6 +588,9 @@ class BedrockConverseModel(Model[BaseClient]):
         # count-tokens-capable models (Claude) don't support native tools, and native-tool-capable models
         # (Nova-2) don't support count_tokens, so a `systemTool` can never reach this request.
         tool_config = self._map_tool_config(model_request_parameters, settings)
+        if not tool_config:
+            # Bedrock requires `toolConfig` whenever the history carries tool blocks.
+            tool_config = self._map_history_tool_config(bedrock_messages)
         if tool_config:
             converse['toolConfig'] = tool_config
         tools: list[ToolTypeDef] = list(tool_config['tools']) if tool_config else []
@@ -793,6 +796,9 @@ class BedrockConverseModel(Model[BaseClient]):
         }
 
         tool_config = self._map_tool_config(model_request_parameters, settings)
+        if not tool_config:
+            # Bedrock requires `toolConfig` whenever the history carries tool blocks.
+            tool_config = self._map_history_tool_config(bedrock_messages)
         if tool_config:
             params['toolConfig'] = tool_config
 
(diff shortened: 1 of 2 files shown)


DouweM:
Closing unless you can verify that the problem is still present today (the issue is very old).

PR #5936, opened 2026-06-15
fix(xai): drop non-native ThinkingParts from history to prevent reasoning leak

### LINKED ISSUE #5927: Thinking-part re-render leak parity: `groq` / `huggingface` / `xai`
## Summary

[#5920](https://github.com/pydantic/pydantic-ai/pull/5920) fixed [#5869](https://github.com/pydantic/pydantic-ai/issues/5869) for **Anthropic** (direct) and **Anthropic-via-Bedrock**: unsigned / foreign-provider `ThinkingPart`s in history were re-rendered as literal `<thinking>…</thinking>` text in the assistant turn, which the model then mimics into its user-visible answer (reasoning leak). The fix drops such parts by default (`anthropic_send_back_thinking_parts` / `bedrock_send_back_thinking_parts`).

A wave-3 sibling-consistency review of #5920 flagged that **`groq`, `huggingface`, and `xai` share the same unguarded re-render shape** and — unlike OpenAI Chat, which re-absorbs `<think>` tags from response text via `split_content_into_text_and_thinking` — read reasoning from native fields, so a mimicked tag would reach the user. This issue tracks verifying and fixing the parity gap; #5920 intentionally stays scoped to the reported+confirmed Anthropic/Bedrock surface.

## Verification (local probes, mirroring the #5869 repro)

Two parts per provider: **(A)** does the mapper inject `<think>` text into history from an unsigned/foreign `ThinkingPart`? **(B)** does the provider fail to re-absorb a `<think>` block from response text (so it would leak into the visible answer)?

| Provider | (A) injects `<think>` from history | (B) re-absorbs response `<think>`? | Verdict |
|---|---|---|---|
| **xai** (`grok-3-mini`) | ✅ yes (deterministic) | ❌ no — a `<think>…</think>` block in the response stays in the visible `TextPart` | **CONFIRMED** — both halves of the #5869 mechanism present |
| **groq** (`openai/gpt-oss-20b`) | ✅ yes (deterministic) | ✅ absorbed for this model/default `groq_reasoning_format` (tags extracted to `ThinkingPart`) | **CONDITIONAL** — injection confirmed, but the leak depends on model + `groq_reasoning_format`; needs per-model/format analysis (reviewer notes `raw` format does not re-absorb) |
| **huggingface** (`Qwen/Qwen3-235B-A22B`) | ✅ yes (deterministic) | ❓ unverified — inference endpoints returned `400` / base-URL routing errors in the probe | **UNVERIFIED on (B)** — injection confirmed; re-absorption needs a reachable endpoint |

Out of scope (correctly): **OpenAI Chat** re-absorbs `<think>` tags (no leak; already guarded by `openai_chat_send_back_thinking_parts`); **Outlines** never sends `ThinkingPart`s back on the wire.

Part-A evidence (the wire injection, identical to #5869) reproduced deterministically for all three via `_map_messages` on an unsigned/foreign `ThinkingPart` — e.g. xai emits `content { text: "<think>\nThe user is asking about X...\n</think>" } role: ROLE_ASSISTANT`. Live multi-turn mimicry (Part B end-to-end) did not reproduce in a short 3-turn probe for groq/xai, consistent with #5869's note that the leak is stochastic and builds up over many persisted turns (the reporter observed 90+ turns).

## Affected code

- `pydantic_ai/models/xai.py` — `XaiModel._map_thinking_part` (the `elif item.content:` → `<think>` text fallback)
- `pydantic_ai/models/groq.py` — `GroqModel._map_messages` `ThinkingPart` branch (unconditional `<think>` re-render)
- `pydantic_ai/models/huggingface.py` — `HuggingFaceModel._map_messages` `ThinkingPart` branch (unconditional `<think>` re-render)

## Proposed fix (future PR)

For the providers confirmed to leak (xai for sure; groq/huggingface pending (B) verification), mirror #5920:

- Add a per-provider `*_send_back_thinking_parts: Literal['auto', 'tags'] = 'auto'` profile setting (`xai_*`, `groq_*` — `huggingface_*` as applicable), defaulting to **drop** unsigned/foreign parts, with `'tags'` as the opt-in to restore the current re-render.
- Keep signed same-provider parts unchanged (native reasoning round-trip).
- For groq, account for `groq_reasoning_format` (the `raw` format re-absorbs; `parsed`/`hidden` do not).
- Add body-level mapper tests mirroring `test_anthropic_send_back_thinking_parts` / `test_bedrock_send_back_thinking_parts`.

Refs: #5869, #5920.

Julien-ser:
Fixes #5927.

## Problem

`XaiModel._map_thinking_part` unconditionally re-wrapped any `ThinkingPart` with content in `<think>…</think>` tags when replaying it in multi-turn history — even if the part originated from a different provider (e.g. Anthropic) or was produced by xAI itself without a signature.

The model sees those tags in the request and mimics that format in its next output, leaking reasoning content into user-visible text.

## Fix

Guard the fallback branch (`elif item.content`) with the new `GrokModelProfile.grok_send_back_thinking_parts: bool = False` flag, so unsignatured and foreign `ThinkingPart`s are silently dropped instead of being re-rendered.

The pattern mirrors the existing `bedrock_send_back_thinking_parts` field in `BedrockModelProfile`. Operators who genuinely need the old re-rendering behaviour can opt in by passing a profile with `grok_send_back_thinking_parts=True`.

## Changes

| File | Change |
|------|--------|
| `profiles/grok.py` | Add `grok_send_back_thinking_parts: bool = False` to `GrokModelProfile` |
| `models/xai.py` | Guard `elif item.content` with the flag in `_map_thinking_part` |
| `tests/models/test_xai.py` | Update snapshot for existing test (was asserting the leaky behaviour); add two new regression tests |

## Tests

- `test_xai_foreign_thinking_part_dropped_from_history` — verifies that a `ThinkingPart` with `provider_name='anthropic'` is silently excluded from the API call.
- `test_xai_foreign_thinking_part_preserved_when_flag_enabled` — verifies that the flag opts back in to the old wrapping behaviour when explicitly set.
- All 111 existing xAI model tests pass.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/5936?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/xai.py
+++ b/pydantic_ai_slim/pydantic_ai/models/xai.py
@@ -456,7 +456,8 @@ class XaiModel(Model[AsyncClient]):
         """Map a `ThinkingPart` into a single xAI assistant message.
 
         - Native xAI thinking (with optional signature) is sent via `reasoning_content`/`encrypted_content`
-        - Non-xAI (or non-native) thinking is preserved by wrapping in the model profile's thinking tags
+        - Non-xAI (or unsignatured) thinking is silently dropped by default; set
+          `GrokModelProfile(grok_send_back_thinking_parts=True)` to re-wrap it in thinking tags instead
         """
         if item.provider_name in _XAI_PROVIDER_NAMES and (item.content or item.signature):
             msg = assistant('')
--- a/pydantic_ai_slim/pydantic_ai/models/xai.py
+++ b/pydantic_ai_slim/pydantic_ai/models/xai.py
@@ -465,7 +465,7 @@ class XaiModel(Model[AsyncClient]):
             if item.signature:
                 msg.encrypted_content = item.signature
             return msg
-        elif item.content:
+        elif item.content and self.profile.get('grok_send_back_thinking_parts', False):
             start_tag, end_tag = self.profile.get('thinking_tags', DEFAULT_THINKING_TAGS)
             return assistant('\n'.join([start_tag, item.content, end_tag]))
         else:
--- a/pydantic_ai_slim/pydantic_ai/profiles/grok.py
+++ b/pydantic_ai_slim/pydantic_ai/profiles/grok.py
@@ -46,8 +46,8 @@ class GrokModelProfile(ModelProfile, total=False):
     grok_reasoning_efforts: frozenset[GrokReasoningEffort]
     """Native `reasoning_effort` values supported by the Grok model. Default: empty (`frozenset()`)."""
 
-    grok_send_back_thinking_parts: bool = False
-    """Whether to re-render non-native ThinkingParts as thinking tags in history.
+    grok_send_back_thinking_parts: bool
+    """Whether to re-render non-native ThinkingParts as thinking tags in history. Default: `False`.
 
     When False (default), ThinkingParts produced by other providers are dropped from history
     rather than being wrapped in thinking tags — which the model would otherwise repeat
--- a/pydantic_ai_slim/pydantic_ai/profiles/grok.py
+++ b/pydantic_ai_slim/pydantic_ai/profiles/grok.py
@@ -46,6 +46,14 @@ class GrokModelProfile(ModelProfile, total=False):
     grok_reasoning_efforts: frozenset[GrokReasoningEffort]
     """Native `reasoning_effort` values supported by the Grok model. Default: empty (`frozenset()`)."""
 
+    grok_send_back_thinking_parts: bool = False
+    """Whether to re-render non-native ThinkingParts as thinking tags in history.
+
+    When False (default), ThinkingParts produced by other providers are dropped from history
+    rather than being wrapped in thinking tags — which the model would otherwise repeat
+    verbatim in subsequent turns (reasoning leak).
+    """
+
 
 def grok_model_profile(model_name: str) -> ModelProfile | None:
     """Get the model profile for a Grok model."""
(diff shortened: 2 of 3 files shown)


Julien-ser:
Hi, just checking in on this one — happy to make any changes if there's feedback. Thanks!

Julien-ser:
Friendly ping — all CI checks are green (lint, mypy, full test matrix on Python 3.10–3.14). Happy to address any feedback or make changes if needed!

dsfaccini:
David's AICA here: Thanks @Julien-ser — folding this into #5920, which consolidates the whole #5927 thinking-leak family (anthropic/bedrock already there, plus groq #6056 and huggingface #6122) into one PR so the cross-provider flag stays consistent. Your fix is preserved with a `Co-authored-by` credit; the only change is the flag shape — `grok_send_back_thinking_parts` is now `Literal['auto','tags']` (default `'auto'`) to match `anthropic_send_back_thinking_parts` and `openai_chat_send_back_thinking_parts` rather than a `bool`, so the value names say which one is leak-safe and the API can extend without a breaking type change. Closing in favor of #5920 — thank you for the contribution!

PR #5944, opened 2026-06-15
fix(openai): add tool_search when a deferred tool is sent on Responses

### LINKED ISSUE #5938: OpenAI Responses adapter sends `defer_loading` without `tool_search` for deferred capabilities
### Initial Checks

### Description

PydanticAI’s OpenAI Responses adapter sends an invalid tool payload for a deferred capability on gpt-5.2.

When a deferred Capability is loaded via `load_capability`, the next request includes the revealed function tool with `defer_loading: true`, but the request does not include the native `tool_search` tool. OpenAI rejects this with:

```
Invalid Value: tools.defer_loading. Deferred tools require tools.tool_search.
```

```Python
import asyncio

from pydantic_ai import Agent
from pydantic_ai.capabilities import Capability


foo = Capability(
    id="foo",
    description="Use this capability when the user asks for foo.",
    defer_loading=True,
)


@foo.tool_plain
def bar(x: int):
    """Return x."""
    return x


agent = Agent(
    "openai-responses:gpt-5.2",
    instructions="First load capability id 'foo', then call bar.",
    capabilities=[foo],
)


async def main():
    await agent.run("Use foo with x=1.")


asyncio.run(main())
```

### Logfire Trace

_No response_

- Python: 3.12
- Pydantic AI: 1.107.0
- LLM provider SDK: openai 2.41.1

joaquinhuigomez:
When a deferred capability is revealed, its function tool is sent with `defer_loading=True`, but the `ToolSearchTool` that originally contributed the native `tool_search` entry may no longer be in `native_tools`. OpenAI then rejects the request with `Invalid Value: tools.defer_loading. Deferred tools require tools.tool_search`. This appends a `tool_search` tool whenever any function tool carries `defer_loading` and none is already present. Adds a unit test on the Responses request path covering the deferred-tool case. Fixes #5938.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/5944?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/openai.py
+++ b/pydantic_ai_slim/pydantic_ai/models/openai.py
@@ -2259,6 +2259,15 @@ class OpenAIResponsesModel(Model[AsyncOpenAI]):
         tools: list[responses.ToolParam] = (
             self._get_native_tools(model_request_parameters) + list(extra_native_tools) + function_tools
         )
+        # OpenAI rejects a request that carries a `defer_loading` tool without an
+        # accompanying `tool_search` tool ("Deferred tools require tools.tool_search").
+        # A revealed deferred capability sets `defer_loading` on its function tool (see
+        # `_map_tool_definition`), but the `ToolSearchTool` that originally contributed the
+        # native `tool_search` entry may no longer be present, so add one back when needed.
+        if any(t.get('defer_loading') for t in function_tools) and not any(
+            t.get('type') == 'tool_search' for t in tools
+        ):
+            tools.append(ToolSearchToolParam(type='tool_search'))
         if not tools:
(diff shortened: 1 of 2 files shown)


adtyavrdhn:
Hey @joaquinhuigomez 

This isn't the approach we should take for this, adding native tool when the model doesn't support it

PR #5978, opened 2026-06-18
feat: add `xai_max_turns` to `XaiModelSettings`

### LINKED ISSUE #5960: xAI: add `xai_max_turns` model setting
### Description

The xAI SDK accepts `max_turns` on `chat.create(...)`. Pydantic AI does not currently expose a provider-specific setting for this value.

This should be a narrow request-setting passthrough for xAI server-side native-tool loops. It should not change Pydantic AI's own agent loop behavior.

### Suggested scope

- Add `xai_max_turns` to `XaiModelSettings`.
- Pass it through to `chat.create(max_turns=...)` in the xAI model request mapping.
- Add focused tests for the outgoing request kwargs.
- Update xAI model settings docs to make clear this controls xAI's server-side native-tool loop, not the Pydantic AI agent loop.

### Related

- Umbrella issue: #5662
- Group A request-setting/provider work: #5741 / #5742

manan-tech:
- Closes #5960

### Summary

Exposes the xAI SDK's `max_turns` parameter through the existing `XaiModelSettings` pattern:

- Adds `xai_max_turns: int` field to `XaiModelSettings` (with docstring matching the SDK's description)
- Adds `'xai_max_turns': 'max_turns'` to `_XAI_MODEL_SETTINGS_MAPPING` — the existing `_map_model_settings()` pipeline forwards it to `chat.create()` with no further changes
- Updates `test_xai_specific_model_settings` to assert `max_turns` appears in the outgoing SDK kwargs
- Adds a short "Agentic turns" section to `docs/models/xai.md`

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/5978?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


dsfaccini:
David's AICA here: thanks @manan-tech for the PR — your mechanism is exactly right, so we kept it as-is: the `xai_max_turns` field on `XaiModelSettings`, the `'xai_max_turns': 'max_turns'` mapping, and the test asserting it reaches the SDK kwargs are all untouched.

We pushed one commit on top that only tightens the docstring and the "Agentic turns" docs section. The previous wording implied `xai_max_turns` caps the agent's tool loop generally, but per the xAI SDK it *"has no effect on non-agentic requests (i.e. requests that do not use server-side tools)"* — Pydantic AI calls `chat.sample()` once per step and runs its own agent loop, so `xai_max_turns` only governs xAI's **server-side native-tool loop** (web search / code execution / X search). The docs now say that plainly and point to [`UsageLimits`](https://ai.pydantic.dev/api/usage/#pydantic_ai.usage.UsageLimits) for bounding Pydantic AI's own loop — which is what #5960 asked for. We also fixed the example slug to `grok-4.3` to match the rest of the page (the retired `grok-4` slug redirects).

Your commit authorship is intact; this is a follow-up commit. Leaving it open for maintainer review.

manan-tech:
Thanks for the docs refinement @dsfaccini — scoping it to the server-side native-tool loop and pointing to `UsageLimits` for the agent loop is a much clearer framing than what I had. Appreciate the review.

pydantic_ai_slim/pydantic_ai/models/xai.py
```diff
@@ -245,6 +245,19 @@ class XaiModelSettings(ModelSettings, total=False):
...
     """
 
+    xai_max_turns: int
+    """Maximum number of agentic turns xAI's server-side tool loop may take.
+
+    Only affects requests that use xAI's server-side native tools (e.g. web search, code
+    execution, X search): xAI iterates up to this many turns — calling those server-side tools
+    and processing their results — before returning a final response. It has no effect on ordinary
+    client-side tools or on Pydantic AI's own agent loop; use [`UsageLimits`][pydantic_ai.usage.UsageLimits]
+    to bound those.
```
dsfaccini, line 255: why did we choose this name? is it consistent with convention? bc if it is tightly related with server-side tool calls (native tools) the name can be clearer

dsfaccini: David's AICA here: `xai_max_turns` follows the established `XaiModelSettings` convention where scalar SDK passthroughs mirror the xAI SDK kwarg name with an `xai_` prefix — `xai_user`→`user`, `xai_store_messages`→`store_messages`, `xai_previous_response_id`→`previous_response_id`, `xai_reasoning_effort`→`reasoning_effort`. The SDK param is `max_turns` on `chat.create()`, so `xai_max_turns` is the exact mirror, and `_XAI_MODEL_SETTINGS_MAPPING` forwards it 1:1. Keeping the mirror maximizes discoverability for users who already know the SDK param.

The server-side-native-tool scope (the source of the ambiguity you flagged) is documented in the docstring rather than encoded in the name.

Contrast: the `xai_include_*` settings are descriptive precisely because they do *not* map 1:1 to a single kwarg — they build the `include` list. Where there's a 1:1 SDK kwarg, we mirror it; where there isn't, we name it descriptively.

Issue #414, opened 2024-12-18
 Incorrect Class Type for `result` in `result_validator` Callback


ishswar:
Incorrect Class Type for `result` in `result_validator` Callback

**Pydantic AI Version:** 0.0.13

---

#### Description:

I am experiencing an issue with the `result_validator` decorator in Pydantic AI when validating the result of a dice game. I expected the `result` parameter in the `validate_result` function to be of type `Game`, but instead, it is of type `pydantic_ai.tools.RunContext`.

Below is the code snippet I am using:

```python
@dice_game_agent.result_validator
def validate_result(result: Game, ctx: RunContext[PlayerData]) -> Game:
    """Validate the result of the dice game."""
    print("Printing class of result")
    print(result.__class__)

    return result
```

**Expected Behavior:**
- I expect the `result` parameter to be an instance of the `Game` model class.

**Actual Behavior:**
- When I run the code, the `result` is incorrectly identified as `pydantic_ai.tools.RunContext`.

**Output:**
```
Printing class of result
<class 'pydantic_ai.tools.RunContext'>
```

#### Steps to Reproduce:
1. Create an agent with a `result_validator` callback like the one above.
2. Pass the `result` and `ctx` parameters as described in the code.
3. Observe the type of `result`, which incorrectly returns `RunContext` instead of the expected `Game` model.

#### Additional Information:
- When I swap the position of `ctx` and `result` in the parameter order (i.e., `ctx` first, then `result`), the issue is resolved, and the `result` is correctly identified as `Game`.

```python
@dice_game_agent.result_validator
def validate_result(ctx: RunContext[PlayerData],result: Game,) -> Game:
    """Validate the result of the dice game."""
    print("Printing class of result")
    print(result.__class__)

    return result
```
#### Output : 

```bash
pydantic_ai version: 0.0.13
Printing class of result
<class '__main__.Game'>
```

- Is this expected behavior, or is this a bug in the way Pydantic AI handles result validation?

---

#### Expected Action:
- Please clarify if this behavior is expected and, if so, document it.
- If it is a bug, It would be good to address it 

### Full Code 

```python
from dataclasses import dataclass
import random
from typing import Optional

import logfire
import pydantic_ai
from devtools import debug
from groq import BaseModel
from pydantic_ai import Agent, RunContext
import dotenv

dotenv.load_dotenv()
# # 'if-token-present' means nothing will be sent (and the example will work) if you don't have logfire configured
[...]

PR #5986, opened 2026-06-19
fix(groq): surface reasoning_tokens in usage details

### LINKED ISSUE #5981: [groq] `cached_tokens` from groq response is dropped and not set in `ModelResponse.usage`
### Initial Checks

### Description

Cached tokens metric is missing in usage returned by `agent.run` however it is returned by the internal groq client.

Usage in groq response (`completion.usage`) contains `prompt_tokens_details.cached_tokens`.
Usage pydantic models in groq lib: https://github.com/groq/groq-python/blob/main/src/groq/types/completion_usage.py#L20 
Example:
```json
{
  "completion_tokens": 135,
  "prompt_tokens": 19038,
  "total_tokens": 19173,
  "completion_time": 0.409281709,
  "completion_tokens_details": {
    "reasoning_tokens": 94
  },
  "prompt_time": 0.031360741,
  "prompt_tokens_details": {
    "cached_tokens": 18944
  },
  "queue_time": 0.018796632,
  "total_time": 0.44064245
}
```

[groq._map_usage](https://github.com/pydantic/pydantic-ai/blob/e356f32cf20fab8fd0c50df7526a43892402b51d/pydantic_ai_slim/pydantic_ai/models/groq.py#L766) [call](https://github.com/pydantic/pydantic-ai/blob/e356f32cf20fab8fd0c50df7526a43892402b51d/pydantic_ai_slim/pydantic_ai/models/groq.py#L402) that sets `ModelResponse.usage` returns `RequestUsage` without cached tokens: 
```python
RequestUsage(input_tokens=19038, output_tokens=135)
```

Sidenote: it would be nice to have reasoning/thinking tokens metric reported in `usage` as well.

**UPD**: afaik the reasoning tokens should be set in `ModelResponse.usage.details["reasoning_tokens"]`
**UPD2**: Created an issue in genai-prices: https://github.com/pydantic/genai-prices/issues/414 as the issue originates there, as far as I understand correctly.

```Python

```

### Logfire Trace

_No response_

- Python: 3.12.7
- Pydantic AI: 1.107.0
- LLM provider SDK: groq 1.4.0

tcconnally:
## Summary

Fixes an issue where `reasoning_tokens` from Groq API responses were silently dropped from `RequestUsage.details`.

## Problem

The Groq API returns usage like:

```json
{
  "prompt_tokens": 19038,
  "completion_tokens": 135,
  "completion_tokens_details": {"reasoning_tokens": 94},
  "prompt_tokens_details": {"cached_tokens": 18944}
}
```

`_map_usage()` only extracted flat integer fields from the top level, discarding `completion_tokens_details.reasoning_tokens`.

`cached_tokens` (via `prompt_tokens_details`) is handled separately by the `genai-prices` extractors invoked in `RequestUsage.extract`; the correct fix there is the upstream extractor config change tracked in pydantic/genai-prices#414, not a duplicate mapping here.

## Fix

Walks `completion_tokens_details` and promotes any integer values into the `details` dict passed to `RequestUsage.extract`, matching the approach in `openai._map_usage`. Result:

```python
RequestUsage(
    input_tokens=19038,
    output_tokens=135,
    details={"reasoning_tokens": 94}
)
```

## Testing

- Existing Groq model tests continue to pass
- Snapshots updated to show `details={'reasoning_tokens': N}`

Closes #5981


VladyslavHl:
@tcconnally I think the issue with cached tokens is not in the `_map_ussage` itself.
I noticed that when using groq through openrouter the usage is set correctly (at least in a way I was used to), despite `prompt_tokens_details` key not being hardcoded in the `openai._map_usage` that is callsed from openrouter.`_map_usage`: 
`prompt_tokens_details.cached_tokens` maps to `details.cache_read_tokens` of the pydantic-ai `RequestUsage`.

When debugging I noticed that these mappings are defined in `provider.extractors` that are invoked in the [RequestUsage.extract](https://github.com/pydantic/pydantic-ai/blob/892fe6b419c563df8e01ac3cc08802604a0c514f/pydantic_ai_slim/pydantic_ai/usage.py#L174C47-L174C73) that `_map_usage` calls.

Mapping in these extractors differ for groq and openrouter: 
openrouter:
SCR-20260622-iung

groq:
SCR-20260622-ivcv

There mappings are part of `genai_prices` package 
image

So I assume the right change would be here under groq key: https://github.com/pydantic/genai-prices/blob/main/prices/data.json 

---

However for `reasoning_tokens` the change `_map_usage` is probably needed as the `openai._map_usage` sets all keys from `completion_tokens_details` to `details` : https://github.com/pydantic/pydantic-ai/blob/892fe6b419c563df8e01ac3cc08802604a0c514f/pydantic_ai_slim/pydantic_ai/models/openai.py#L4300

--- 

But thanks for opening PR as it made me think more! 
Posted this finding into the original issue and created a new issue in the genai_prices repo: https://github.com/pydantic/genai-prices/issues/414

tcconnally:
Thanks @VladyslavHl, that's a great catch and you're right. `RequestUsage.extract` is where the `prompt_tokens_details.cached_tokens -> details.cache_read_tokens` mapping happens via the provider extractors in `genai-prices`, so doing it in `groq._map_usage` is the wrong layer for cached tokens; it would also collide with the canonical mapping rather than complement it. The proper fix is the `genai-prices` data/extractor change you opened in pydantic/genai-prices#414.

`reasoning_tokens` is the part that genuinely needs handling here, since the extractors don't surface it and `openai._map_usage` already lifts `completion_tokens_details` into `details` the same way.

So I've narrowed this PR to just that: it now lifts the integer fields of `completion_tokens_details` (i.e. `reasoning_tokens`) into `details` and leaves `cached_tokens` entirely to the extractors once #414 lands. Updated the affected snapshots accordingly, and they now show `details={'reasoning_tokens': N}` with no `cached_tokens`. Let me know if you'd rather fold this into the genai-prices work too and close this one.

tcconnally:
@VladyslavHl thanks for taking a look! 

You're right that the `cached_tokens` path is handled through the provider extractors — that's a separate concern. The scope of this PR is specifically `completion_tokens_details.reasoning_tokens`, which was being silently dropped by `_map_usage()` because it only extracted flat integer fields from the top level.

The `prompt_tokens_details.cached_tokens` → `details.cache_read_tokens` mapping you linked goes through `RequestUsage.extract()` which uses the extractors — that path works as intended. The gap this PR fixes is that `reasoning_tokens` (an integer inside a nested `details` object) wasn't being surfaced at all.

I've kept the change minimal — just bubbling `reasoning_tokens` through `_map_usage()` so users consuming Groq responses through PydanticAI get accurate token breakdowns. Let me know if you see anything else that should be adjusted!

PR #5999, opened 2026-06-21
fix(litellm): default openai_chat_supports_multiple_system_messages=False for models without recognized provider prefix

### LINKED ISSUE #5812: vLLM backend rejects PromptedOutput requests with 'System message must be at the beginning'

citizen204:
## Summary

Fixes #5812.

When `LiteLLMProvider` is used with a model that has no recognized provider prefix (e.g. a bare model name like `my-vllm-model` for a custom vLLM deployment), the backend often rejects multiple consecutive system messages with `400 Bad Request: System message must be at the beginning`. The `OpenAIModelProfile.openai_chat_supports_multiple_system_messages` docstring already notes this: _"Set to `False` for strict OpenAI-compatible backends (e.g. some LiteLLM/vLLM deployments)"_.

## Changes

**`pydantic_ai_slim/pydantic_ai/providers/litellm.py`**
- Track whether the model name had a recognized provider prefix (e.g. `openai/`, `anthropic/`)
- For models **without** a recognized prefix, override `openai_chat_supports_multiple_system_messages=False` after building the profile, causing consecutive leading system messages to be merged before the request is sent
- Models **with** a recognized prefix (e.g. `openai/gpt-4o`, `anthropic/claude-3`) keep their provider-specific setting unchanged

**`tests/providers/test_litellm.py`**
- New test `test_model_profile_multiple_system_messages` verifying the correct value for bare names, unknown prefixes, and recognized prefixes

## Behaviour

| Model name | `openai_chat_supports_multiple_system_messages` | Why |
|---|---|---|
| `my-vllm-model` | `False` (new) | No prefix — conservative default for strict backends |
| `unknown-provider/model` | `False` (new) | Unrecognized prefix — treated as custom |
| `openai/gpt-4o` | `True` (unchanged) | Known prefix, OpenAI profile preserved |
| `anthropic/claude-3` | `True` (unchanged) | Known prefix, Anthropic profile preserved |

The manual workaround from the issue (passing `OpenAIModelProfile(openai_chat_supports_multiple_system_messages=False)`) still works and takes precedence over the provider default.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/5999?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/providers/litellm.py
+++ b/pydantic_ai_slim/pydantic_ai/providers/litellm.py
@@ -1,5 +1,6 @@
 from __future__ import annotations as _annotations
 
+from dataclasses import replace
 from typing import overload
 
 from httpx import AsyncClient as AsyncHTTPClient
@@ -67,12 +68,14 @@ class LiteLLMProvider(Provider[AsyncOpenAI]):
         }
 
         profile = None
+        has_recognized_prefix = False
 
         # Check if model name contains a provider prefix (e.g., "anthropic/claude-3")
         if '/' in model_name:
             provider_prefix, model_suffix = model_name.split('/', 1)
             if provider_prefix in provider_to_profile:
                 profile = provider_to_profile[provider_prefix](model_suffix)
+                has_recognized_prefix = True
 
         # If no profile found, default to OpenAI profile
         if profile is None:
@@ -80,7 +83,15 @@ class LiteLLMProvider(Provider[AsyncOpenAI]):
 
         # As LiteLLMProvider is used with OpenAIModel, which uses OpenAIJsonSchemaTransformer,
         # we maintain that behavior
-        return OpenAIModelProfile(json_schema_transformer=OpenAIJsonSchemaTransformer).update(profile)
+        result = OpenAIModelProfile(json_schema_transformer=OpenAIJsonSchemaTransformer).update(profile)
+
+        # Models without a recognized provider prefix are often custom backends (e.g. vLLM deployments)
+        # that require exactly one system message at the start of the conversation. Default to merging
+        # consecutive system messages to avoid "System message must be at the beginning" errors.
+        if not has_recognized_prefix:
+            result = replace(result, openai_chat_supports_multiple_system_messages=False)
+
+        return result
 
     @overload
     def __init__(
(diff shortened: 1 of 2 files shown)


pydantic_ai_slim/pydantic_ai/providers/litellm.py
```diff
@@ -67,20 +68,30 @@ def model_profile(model_name: str) -> ModelProfile | None:
...
 
         # As LiteLLMProvider is used with OpenAIModel, which uses OpenAIJsonSchemaTransformer,
         # we maintain that behavior
-        return OpenAIModelProfile(json_schema_transformer=OpenAIJsonSchemaTransformer).update(profile)
+        result = OpenAIModelProfile(json_schema_transformer=OpenAIJsonSchemaTransformer).update(profile)
+
+        # Models without a recognized provider prefix are often custom backends (e.g. vLLM deployments)
+        # that require exactly one system message at the start of the conversation. Default to merging
+        # consecutive system messages to avoid "System message must be at the beginning" errors.
+        if not has_recognized_prefix:
```
adtyavrdhn, line 91: This will leave the base_url convention broken still

citizen204: Good point — the  pattern (e.g.  + custom  pointing to vLLM) would still be affected since  is a recognized prefix.

I also saw your comment on the issue preferring a dedicated vLLM provider or documentation update over auto-detection. Happy to pivot: should I close this PR and open one that either (a) adds a  wrapper with the flag set, or (b) adds a documentation note in the LiteLLM provider page? Let me know which direction you'd prefer.

adtyavrdhn: I'll do it in two phases, one with a doc change and one with a vLLM provider. I'll close this one. Thanks for taking the time.

PR #6004, opened 2026-06-21
Convert `ExceptionGroup`-wrapped tool errors to `ModelRetry` in `MCPToolset`

dmontagu:
## What

A tool call through `MCPToolset` (the FastMCP-based client, and the only MCP client `load_mcp_toolsets` builds) can fail with the underlying error **wrapped in an `ExceptionGroup`** (`"unhandled errors in a TaskGroup"`) rather than as a bare `ToolError`/`McpError`, because the MCP client session runs inside an anyio task group. The existing `except ToolError` in `MCPToolset._call_tool` doesn't match an `ExceptionGroup`, so the error escaped as a fatal exception out of `agent.run()` instead of being converted to a recoverable `ModelRetry`.

This was discovered by this repo's own **Pydantic AI PR Review** workflow (gh-aw), whose runner *is* a Pydantic AI agent: when the reviewer had nothing to flag it submitted an empty-bodied review, the gh-aw safe-outputs server rejected it with `ERR_VALIDATION`, and the resulting error came back wrapped in an `ExceptionGroup` — crashing the agent on every retry (`is_error`, 0 tokens) instead of being a recoverable retry the model could resolve.

## Fix

Add an `except BaseExceptionGroup` alongside the existing `except ToolError` in `MCPToolset._call_tool`, around `await self.client.call_tool(...)`:

- `eg.split((ToolError, McpError))` separates tool/protocol errors from anything else.
- Group of **only** `ToolError`/`McpError` → descend to the leaf and convert to `ModelRetry` (matching the bare case; honors `tool_error_behavior='error'`, which still propagates).
- Otherwise re-raise unchanged, so a concurrent `CancelledError` (or anything else) grouped alongside is never swallowed.

3.10-compatible via the repo's existing `_utils.BaseExceptionGroup`.

> **Note — earlier revision.** A previous version of this PR placed the guard on `MCPServer.direct_call_tool`. That class is **deprecated** and `load_mcp_toolsets` never instantiates it, so the guard never ran on the path the workflow actually hits. It has been moved to `MCPToolset`, the live FastMCP path.

## What's proven vs. defended

The production run artifact proves the failure shape — `ExceptionGroup([McpError])` escaping `agent.run()` on the `MCPToolset` path. The **exact frame it unwinds from** (the guarded `await self.client.call_tool` vs. the session-context exit) and the **task-group race that births it** (an empty-bodied tool error colliding with the session's GET-stream teardown) are **inferred, not reproduced** — every faithful local reproduction, at the exact failing versions, converted cleanly. So this is a deliberately **defensive** guard, strictly safer than the status quo. Capturing the real traceback in the gh-aw shim (a separate change) would let us confirm the placement.

## Tests

`tests/test_mcp_toolset.py::TestMCPToolsetIntegration::test_call_tool_unwraps_real_exception_group_to_model_retry` — drives the in-process FastMCP toolset, injecting a failure at the real seam (`self.client.call_tool`) and letting a **genuine `anyio` task group** produce the `ExceptionGroup` (parametrized over `ToolError` and the production-realistic `McpError`), then asserts `ModelRetry`. Unit test because the wrapping race can't be triggered deterministically; verified it fails without the guard.

No public API or behavior change a user configures, so no docs update.

### Checklist

### DIFF
--- a/tests/test_mcp.py
+++ b/tests/test_mcp.py
@@ -17,12 +17,14 @@ from tempfile import TemporaryDirectory
 from typing import Any
 from unittest.mock import AsyncMock
 
+import anyio
 import httpx
 import pytest
 from inline_snapshot import snapshot
 
 from pydantic_ai import models
 from pydantic_ai._run_context import RunContext
+from pydantic_ai._utils import BaseExceptionGroup
 from pydantic_ai.exceptions import ModelRetry
 from pydantic_ai.models.test import TestModel
 from pydantic_ai.usage import RunUsage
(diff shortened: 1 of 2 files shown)


tests/test_mcp.py
```diff
@@ -2110,6 +2111,42 @@ async def test_mcp_server_raises_mcp_error(
...
+    mcp_error = McpError(error=ErrorData(code=400, message='Wrapped MCP error conversion'))
+    wrapped = BaseExceptionGroup('unhandled errors in a TaskGroup', [mcp_error])
+
+    async with agent:
+        with patch.object(
+            mcp_server._get_client(),  # pyright: ignore[reportPrivateUsage]
+            'send_request',
+            new=AsyncMock(side_effect=wrapped),
+        ):
+            with pytest.raises(ModelRetry, match='Wrapped MCP error conversion'):
```
adtyavrdhn: Did Claude test that this can happen by the actual chain? 

Looking at the MCP sdk code it looks like it can only be raised on task group closure which happens when the session is closed and never at tool call. I think this is fixing a phantom bug, this doesn't exist.

adtyavrdhn: Claude is not able to reproduce this and I don't see it happening either. I'll add a defensive check for this in the correct place and add traceback logging in the shim so that we don't move blind here

adtyavrdhn (approved):
It was catching the error at the incorrect boundary but I am not able to still reproduce this. I'm keeping it around as a defensive measure and I'll change the shim logs to print tracebacks in case it happens again

PR #6009, opened 2026-06-22
Fix map `downstream_join_id` double-firing the join on a non-empty iterable

### LINKED ISSUE #6008: [graph] map with downstream_join_id double-fires the join on a non-empty iterable
### Initial Checks

### Description

> **Disclaimer:** This issue was found by an AI agent (Claude Opus 4.8); I'm filing it on the agent's behalf and don't have deep pydantic-graph internals expertise to independently judge it. The evidence below is what the agent produced and ran. Please verify rather than take it at face value.

With the beta/builder graph API, `add_mapping_edge(..., downstream_join_id=...)` makes the node downstream of the join fire **twice** when the mapped iterable is non-empty: once with the reducer's initial value, once with the actual reduced result. The empty initial-value emission wins, so the run returns the initial value instead of the reduced output — a silently wrong result.

`downstream_join_id` is documented only as a safeguard for mapping an empty iterable, so on a non-empty iterable it should be a no-op. It isn't: the same graph with a plain `.map()` (no `downstream_join_id`) returns the correct reduced value in a single firing.

**Observed** (full runnable script in the "Minimal, Reproducible Example" field below):

```text
run output  : []
join emitted: [[], [1, 2, 3]]
```

`after_join` (the node wired after the join) runs twice — first with `[]`, then with `[1, 2, 3]` — and the run output is the empty one.

The single line that causes it is `downstream_join_id=collect.id`. Replacing the mapping edge with a plain map and no `downstream_join_id`:

```python
g.add(g.edge_from(produce).map().to(identity))
```

gives the correct result, with `after_join` firing once:

```text
run output  : [1, 2, 3]
join emitted: [[1, 2, 3]]
```

**Expected**

With `downstream_join_id` set and a non-empty iterable, the join should fire once and emit the fully reduced list — run output `[1, 2, 3]`, `after_join` once. The basis for this isn't an assumption: the identical graph with a plain `.map()` already returns `[1, 2, 3]` in a single firing, and `downstream_join_id` is documented purely as an empty-iterable safeguard ("Specifying this ensures correct handling if you try to map an empty iterable."), so it should be a no-op when the iterable is non-empty.

**Notes**

- Empty-iterable behavior, for context: **with** `downstream_join_id`, mapping `[]` works correctly (one firing with `[]`). **Without** it, mapping `[]` raises `RuntimeError: Graph run completed, but no result was produced. This is either a bug in the graph or a bug in the graph runner.`. So the two are a lose-lose — the flag is required to survive an empty map but corrupts a non-empty one.
- The `downstream_join_id` example in the docs only exercises the empty-iterable case; the non-empty + `downstream_join_id` combination appears untested.
- Reproduces identically on 2.0.0b7 (latest beta) and 1.107.0 (latest stable); no agent or model call is involved.

```Python
import asyncio
from pydantic_graph import GraphBuilder, StepContext, reduce_list_append

emitted: list[list[int]] = []  # what the join hands to the node after it

g = GraphBuilder(output_type=list)

@g.step
async def produce(ctx: StepContext[None, None, None]) -> list[int]:
    return [1, 2, 3]                          # non-empty input

@g.step
async def identity(ctx: StepContext[None, None, int]) -> int:
    return ctx.inputs

collect = g.join(reduce_list_append, initial_factory=list)

@g.step
async def after_join(ctx: StepContext[None, None, list]) -> list:
    emitted.append(sorted(ctx.inputs))
    return ctx.inputs

g.add(g.edge_from(g.start_node).to(produce))
g.add_mapping_edge(produce, identity, downstream_join_id=collect.id)
g.add(
    g.edge_from(identity).to(collect),
    g.edge_from(collect).to(after_join),
    g.edge_from(after_join).to(g.end_node),
)
graph = g.build()

result = asyncio.run(graph.run())
print("run output  :", result)      # expected [1, 2, 3], actual []
print("join emitted:", emitted)      # expected [[1, 2, 3]], actual [[], [1, 2, 3]]
```

### Logfire Trace

_No response_

- Python: 3.12.3
- pydantic-graph: 2.0.0b7 (ships with pydantic-ai 2.0.0b7); reproduces identically on 1.107.0, the latest stable
- LLM provider SDK: N/A — no model or provider is involved (pure pydantic-graph)

clavedeluna:
- Closes #6008

## Summary

With the builder graph API, `add_mapping_edge(..., downstream_join_id=...)` made the node downstream of the join fire **twice** on a non-empty iterable — once with the reducer's initial value and once with the actual reduced result — and the empty initial-value emission could win, silently returning the wrong result. `downstream_join_id` is documented purely as a safeguard for mapping an *empty* iterable, so on a non-empty iterable it should be a no-op.

### Root cause

A join's reducer is keyed by the run id of the join's **parent (dominating) fork**, and live `JoinItem` reductions look that up via `get_parent_fork(join_id)`. But the eager reducer created for a map's `downstream_join_id` (the empty-iterable safeguard) was keyed by the **map fork's own run id** instead. These only coincide when the map fork *is* the join's dominating fork; otherwise you get two reducers for the same join run:

- the eager one, stuck at the initial value, and
- the real one, accumulating the mapped items,

both of which get finalized → the downstream node fires twice.

### Fix

Extract the parent-fork-resolution logic into a shared `_resolve_join_fork_run` helper and use it from both the `JoinItem` path and the eager-create path, so the eager reducer lands on the same key and a non-empty iterable accumulates into a single reducer. The empty-iterable safeguard behavior is unchanged.

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6009?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_graph/pydantic_graph/graph_builder.py
+++ b/pydantic_graph/pydantic_graph/graph_builder.py
@@ -706,20 +706,9 @@ class _GraphIterator(Generic[StateT, DepsT, OutputT]):
                             elif isinstance(maybe_overridden_result, JoinItem):
                                 result = maybe_overridden_result
                                 parent_fork_id = self.graph.get_parent_fork(result.join_id).fork_id
-                                for i, x in enumerate(result.fork_stack[::-1]):
-                                    if x.fork_id == parent_fork_id:
-                                        # For non-final joins (those that are intermediate nodes of other joins),
-                                        # preserve the fork stack so downstream joins can still associate with the same fork run
-                                        if self.graph.is_final_join(result.join_id):
-                                            # Final join: remove the parent fork from the stack
-                                            downstream_fork_stack = result.fork_stack[: len(result.fork_stack) - i]
-                                        else:
-                                            # Non-final join: preserve the fork stack
-                                            downstream_fork_stack = result.fork_stack
-                                        fork_run_id = x.node_run_id
-                                        break
-                                else:  # pragma: no cover
-                                    raise RuntimeError('Parent fork run not found')
+                                fork_run_id, downstream_fork_stack = self._resolve_join_fork_run(
+                                    result.join_id, result.fork_stack
+                                )
 
                                 join_node = self.graph.nodes[result.join_id]
                                 assert isinstance(join_node, Join), f'Expected a `Join` but got {join_node}'
@@ -964,6 +953,22 @@ class _GraphIterator(Generic[StateT, DepsT, OutputT]):
         else:
             assert_never(next_node)
 
+    def _resolve_join_fork_run(self, join_id: JoinID, fork_stack: ForkStack) -> tuple[NodeRunID, ForkStack]:
+        """Determine which parent fork run a join item belongs to, and the fork stack its output should run under."""
+        parent_fork_id = self.graph.get_parent_fork(join_id).fork_id
+        for i, x in enumerate(fork_stack[::-1]):
+            if x.fork_id == parent_fork_id:
+                if self.graph.is_final_join(join_id):
+                    # Final join: keep the stack up to and including the parent fork, dropping everything
+                    # below it (the fork that directly produced this item, and any deeper forks).
+                    downstream_fork_stack = fork_stack[: len(fork_stack) - i]
+                else:
+                    # Non-final join (an intermediate node of another join): preserve the fork stack so
+                    # downstream joins can still associate with the same fork run.
+                    downstream_fork_stack = fork_stack
+                return x.node_run_id, downstream_fork_stack
+        raise RuntimeError('Parent fork run not found')  # pragma: no cover
+
     def _get_completed_fork_runs(
         self,
         t: GraphTask,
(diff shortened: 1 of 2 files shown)


itxaiohanglover:
Nice refactor! Extracting the fork run resolution into a helper method makes the logic much clearer. The fix for empty map iterables — eagerly creating join state upfront — is important for correctness when no items ever reach the join.

slavashvets:
Thanks for the fix, @clavedeluna. It's only conflicting because the V2 merge (#5451) reworked the surrounding `JoinItem` path a day after you opened this. The actual fix isn't affected - the buggy eager-reducer line is still byte-identical on `main`, so #6008 still reproduces post-V2.

Could you rebase onto `main`?

dmontagu (approved):

PR #6016, opened 2026-06-23
fix(openai): raise ContentFilterError for plain-text content_filter responses

ikatyal2110:
OpenAIResponsesModel only raised ContentFilterError when the block came back as a structured ResponseOutputRefusal. When a content-policy block arrives as plain text plus incomplete_details.reason='content_filter' (happens with Azure + web search), the parts stayed non-empty so the agent loop never raised and the refusal text was returned as normal output. Now clear the parts in that case too, like we already do for structured refusals. Added a regression test. Fixes #5221.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6016?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


DouweM:
This is the net-negative path in https://github.com/pydantic/pydantic-ai/issues/5221#issuecomment-4700095640.

PR #6017, opened 2026-06-23
ci: try ubicloud-standard-8-arm in place of ubicloud-premium-4

strawgate:
## Why try this

89% of CI minutes on this repo (n=30 recent main runs) are on the `test` + `test-lowest-versions` matrix, which currently uses the `ubicloud-premium-4` conditional. Swapping to `ubicloud-standard-8-arm` is interesting on three axes:

1. **~same price/min** — 8-core arm is roughly the same cost as 4-core x86 premium on ubicloud.
2. **2× cores** — should offset slower per-core arm performance for parallelizable variants (`pytest -n auto`).
3. **First arm test coverage** — a meaningful chunk of pydantic-ai users run on Mac M-series (developer) and AWS Graviton (production). Today the test matrix has zero arm signal; an arm-specific regression in the dep tree (`pydantic-core` SIMD path, etc.) wouldn't surface until users hit it.

The `|| 'ubuntu-latest'` fork-PR fallback stays as the still-x86 leg, so fork contributors still get free runners and the matrix retains x86 coverage on every untrusted PR.

## What to look at

This PR's own CI is the measurement. Compare wall-clock on:

| variant | premium-4 today (avg) | -8-arm here (TBD) |
|---|---|---|
| `test on 3.12 (lowest-versions)` | ~9:00 | ? |
| `test on 3.12 (all-extras)` | ~8:30 | ? |
| `test on 3.12 (standard)` | ~5:00 | ? |
| `test on 3.12 (pydantic-evals)` | ~2:10 | ? |
| `test on 3.12 (pydantic-ai-slim)` | ~2:10 | ? |

If the long variants land within ~30% of x86 wall-clock, this is a clean win (~30-40% cost reduction on the bulk of CI plus arm coverage). If `lowest-versions` doubles, we revert — or split the matrix to keep `lowest-versions` on premium and the others on arm.

## Revert

One file (`ci.yml`), 2 lines, easy.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6017?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -152,7 +152,7 @@ jobs:
       # under-subscribes Ubicloud's hyperthreaded vCPUs (premium-4 -> 2 workers
       # instead of 4). `logical` uses all vCPUs; ~39% faster on premium-4, no-op
       # on ubuntu-latest (already 4). See PR benchmark.
-      - run: uv run ${{ matrix.install.command }} coverage run -m pytest --durations=100 -n $(($(nproc) * 2)) --dist=loadgroup
+      - run: uv run ${{ matrix.install.command }} coverage run -m pytest --durations=100 -n logical --dist=loadgroup
         env:
           COVERAGE_FILE: .coverage/.coverage.${{ matrix.python-version }}-${{ matrix.install.name }}
 
@@ -220,7 +220,7 @@ jobs:
           UV_FROZEN: "0"
 
       # `-n logical` (see note on the `test` job): use all vCPUs on Ubicloud.
-      - run: uv run --no-sync coverage run -m pytest --durations=100 -n $(($(nproc) * 2)) --dist=loadgroup
+      - run: uv run --no-sync coverage run -m pytest --durations=100 -n logical --dist=loadgroup
         env:
           COVERAGE_FILE: .coverage/.coverage.${{matrix.python-version}}-lowest-versions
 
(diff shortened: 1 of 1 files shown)


strawgate:
Closing — collected the data we wanted, not pursuing the migration here.

## What we measured

**`ubicloud-standard-8-arm` vs `ubicloud-premium-4` (n=30 amd baseline, n=1 arm)**: essentially flat. Across the 25 test matrix variants, arm-8 lands at 92-106% of premium-4 wall-clock for the variants that matter most (lowest-versions, all-extras). The 2× core count compensates for slower per-core arm perf. Two outliers were both py3.13 small variants (likely n=1 noise — py3.13 amd numbers are unusually low).

**`-n logical*2` oversubscribe (16 workers on 8 cores)**: clean loss. Every single variant got 14-41% slower than `-n logical` (8 workers). Tests aren't I/O-bound enough to fill idle worker slots; 16 venvs on 8 cores thrash on context switching. Reverted.

## Takeaways for if you ever revisit

- arm-8 + `-n logical` is roughly cost-flat vs premium-4 at similar wall-clock.
- The bonus would have been arm test coverage for end-users on Mac M-series / Graviton — currently zero in this repo.
- The fork-PR `|| 'ubuntu-latest'` fallback already keeps an x86 leg for free contributor PRs, so a one-line flip is also a one-line revert.
- Don't try `-n logical*2`. CPU-bound, doesn't help.

No urgency to land this on the platform side; the data point stands either way.

PR #6041, opened 2026-06-24
Add `pytest.mark.eval` integration for `pydantic-evals`

Kludex:
Closes: N/A

### Summary

Adds a pytest plugin for `pydantic-evals` so datasets can be run with `pytest.mark.eval(dataset)`.

The plugin:

- registers a `pytest.mark.eval` marker and pytest entry point
- treats a one-argument marked test function as the eval task
- supports `task_factory=True` for pytest fixture-driven task factories
- fails tests based on configurable eval report criteria
- renders eval reports in the pytest terminal summary
- automatically configures Logfire when available and links to the dataset eval page
- exposes `assert_evaluation_passes()` and `assert_evaluation_passes_sync()` helpers

### Validation

- `uv run pytest tests/evals/test_pytest.py -q`
- `uv run ruff check pydantic_evals/pydantic_evals/pytest.py tests/evals/test_pytest.py docs/evals/how-to/pytest.md`
- `PYRIGHT_PYTHON_IGNORE_WARNINGS=1 uv run pyright pydantic_evals/pydantic_evals/pytest.py tests/evals/test_pytest.py`
- `uv run pytest test_scratch.py`

The full pre-commit typecheck hook currently fails on unrelated optional dependency/import issues outside this change, including `sentence_transformers`, `voyageai`, and AG-UI symbols. The focused typecheck for the changed plugin and tests passes.

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6041?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


strawgate:
One gap worth considering: datasets that use **span-based evaluators** (`HasMatchingSpan` — the documented way to assert tool calls / control flow) don't work under `@pytest.mark.eval` unless `logfire` is installed *and* configured, but the `pytest` extra doesn't pull in `logfire`.

With just `pip install 'pydantic-evals[pytest]'`, a case with `HasMatchingSpan(...)` produces:

```
assertions: {}
evaluator_failures: ['SpanTreeRecordingError: ... you need to call logfire.configure(...) before running an evaluation']
```

So with the default `require_assertions` / `fail_on_evaluator_failures` the test fails with a confusing "evaluator failure", and the suggested fix (`logfire.configure`) isn't even importable from the advertised extra.

Proposed fix: when no `TracerProvider` is configured (and `logfire` isn't present), have the plugin set up an in-process OpenTelemetry SDK `TracerProvider` so spans are recorded out of the box. Failing that, a docs note + having `[pytest]` suggest the `logfire` extra would at least make it discoverable.

---
_Generated by [Claude Code](https://claude.ai/code)_

Kludex:
Thanks, this was valid. Addressed in `11cdaac5c` by adding `opentelemetry-sdk` to the `pydantic-evals[pytest]` extra and having the pytest plugin install an in-process OpenTelemetry SDK `TracerProvider` when Logfire is not importable and no provider is configured.

I also made Logfire eval URL rendering tolerant of older Logfire versions that do not expose `url_from_eval`, and added regression coverage for both cases.

Kludex:
We should document the way to not run this on every run e.g. `pytest --no-eval`.

PR #5486, opened 2026-05-16
test(evals): seed RNG in correlated/independent sampling tests (#5399)

Quratulain-bilal:
Addresses sub-item #2 of #5399.
  ▎
  ▎ Three sampling tests in tests/evals/test_online.py depend on the global RNG drawn by
  ▎ pydantic_evals._online.sample_evaluators, so they can flake on CI:
  ▎ - test_correlated_sampling_subset_property fails when no call fires the low-rate (0.1) evaluator — P ≈ 0.9¹⁰⁰ ≈ 2.7e-5
  ▎  (this is the case logged in #5399).
  ▎ - test_correlated_sampling_max_overhead asserts 5 < calls < 40 for 200 draws at rate 0.1; binomial tails can still
  ▎ land outside.
  ▎ - test_independent_sampling_is_default asserts >0 fires for two 0.5-rate evaluators over 100 draws.
  ▎
  ▎ Each test now monkey-patches _online.random with a seeded random.Random(0). pytest restores the attribute on teardown
  ▎ so no state leaks. Ran the file 5× locally  102/102 pass each time.

PR #6046, opened 2026-06-24
Fix two flaky tests in `test_online.py` and `test_dbos.py`

### LINKED ISSUE #5399: Test flakes: 5 distinct CI flakes across capabilities, evals, temporal, ui_web, and dbos

Oxygen56:
## Changes

Fix two flaky tests tracked in #5399:

1. **`test_correlated_sampling_subset_property`** — seed `random` before the test to eliminate the probabilistic assertion failure (P(zero) ≈ 2.7e-5 at `sample_rate=0.1` with 100 calls)

2. **`test_complex_agent_run_in_workflow`** — sort `span.children` by `content` in `_normalize_json_spans` to fix non-deterministic child ordering from the async span tree

## Related issue

Closes #5399 (partial — items 2 and 5)

## Checklist

- [x] Tests pass locally for the evals test fix
- [ ] AI generated code (please check this box in the UI)

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6046?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/ui/_web/app.py
+++ b/pydantic_ai_slim/pydantic_ai/ui/_web/app.py
@@ -47,6 +47,18 @@ def _get_cache_dir() -> Path:
     return cache_dir
 
 
+def _write_cache_atomically(cache_file: Path, content: bytes) -> None:
+    """Write `content` to `cache_file` atomically.
+
+    `Path.write_bytes` truncates the destination before writing, so a concurrent reader can
+    observe the file existing but empty. Writing to a sibling temp file and `os.replace`-ing it
+    into place makes the swap atomic, so readers only ever see the complete file.
+    """
+    tmp_file = cache_file.with_name(f'{cache_file.name}.{os.getpid()}.tmp')
+    tmp_file.write_bytes(content)
+    os.replace(tmp_file, cache_file)
+
+
 async def _get_ui_html(html_source: str | Path | None = None) -> bytes:
     """Get UI HTML content from the specified source or default CDN.
 
@@ -73,7 +85,7 @@ async def _get_ui_html(html_source: str | Path | None = None) -> bytes:
             response.raise_for_status()
             content = response.content
 
-        cache_file.write_bytes(content)
+        _write_cache_atomically(cache_file, content)
         return content
 
     # Handle Path instances
@@ -97,7 +109,7 @@ async def _get_ui_html(html_source: str | Path | None = None) -> bytes:
             response.raise_for_status()
             content = response.content
 
-        cache_file.write_bytes(content)
+        _write_cache_atomically(cache_file, content)
         return content
 
     # Handle local file paths (strings)
--- a/tests/evals/test_online.py
+++ b/tests/evals/test_online.py
@@ -3,6 +3,7 @@
 from __future__ import annotations as _annotations
 
 import asyncio
+import random
 from collections.abc import Sequence
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any
@@ -1811,8 +1812,6 @@ async def test_sampling_context_input_based_sampling():
 @pytest.mark.anyio
 async def test_correlated_sampling_subset_property():
     """In correlated mode, lower-rate evaluator calls are a subset of higher-rate ones."""
-    import random
-
     random.seed(0)
     collector_high = Collector()
     collector_low = Collector()
--- a/tests/evals/test_online.py
+++ b/tests/evals/test_online.py
@@ -1811,6 +1811,9 @@ async def test_sampling_context_input_based_sampling():
 @pytest.mark.anyio
 async def test_correlated_sampling_subset_property():
     """In correlated mode, lower-rate evaluator calls are a subset of higher-rate ones."""
+    import random
+
+    random.seed(0)
     collector_high = Collector()
     collector_low = Collector()
     config = OnlineEvalConfig(sampling_mode='correlated')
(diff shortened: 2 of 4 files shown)


dsfaccini:
David's AICA here — we're taking this one over to land it, so I've pushed some changes to the branch. Quick rundown of what changed and why:

- **`test_online.py`** — kept your `random.seed(0)` fix (just lifted `import random` to module scope). This is a genuine latent flake; good catch.
- **`test_dbos.py`** — reverted this hunk. I pulled the CI log for the run #5399 cites (`25457470942`), and that failure was actually three stray `POST api.openai.com` httpx-instrumentation spans in an older snapshot — a span-*presence* mismatch, not non-deterministic child ordering. Sorting `span.children` can't fix a multiset difference, and today's snapshot has no POST spans at all, so there's nothing to fix there.
- **`test_ui_web.py`** — swapped the module-wide `xdist_group` marker for per-test cache isolation (monkeypatching the cache dir + stubbing the CDN fetch on the two tests that actually hit the real cache), matching the pattern the other cache tests already use. As a bonus this drops a live network request from the suite.

Still looking at the underlying non-atomic cache-write race those `ui_web` tests exposed — will follow up here. Thanks for getting this started!

dsfaccini:
David's AICA here — an update on the takeover.

A closer look turned up two earlier open PRs that already cover the same two #5399 items this PR targets, both more complete:

- **#5490** (cache-write race) — adds the atomic write **plus** a read-side empty-cache guard and temp-file cleanup-on-error, across both cache branches.
- **#5486** (RNG seed) — seeds the sampler via a module-scoped `Random(0)` (no global-RNG-state leak) across all three sampling tests.

Rather than keep three PRs racing on the same flakes, we've consolidated onto those two: both have been refreshed against v2, reviewed, and are green. The one thing this PR's branch had that #5490 didn't — a deterministic cache-write-**atomicity** regression test — has been carried over into #5490 (re-expressed to fit its `os.replace`-based implementation).

So I'm closing this as superseded, with thanks — the work here directly drove the atomicity test that's now part of #5490. 🙏

(For the record: the `test_dbos.py` change was reverted earlier — that flake was misdiagnosed, as noted above; the cited CI failure was stray httpx spans, not child ordering.)

PR #6047, opened 2026-06-24
Add return type annotations to model schema helpers

Oxygen56:
## Changes

Add explicit return type annotations to two internal model schema helper functions in `pydantic_ai_slim/pydantic_ai/models/__init__.py`:

- `_customize_tool_def` returns `ToolDefinition`
- `_customize_output_object` returns `OutputObjectDefinition`

Both helpers already return `replace(...)` results for their corresponding input dataclass types, so this is a typing-only cleanup with no intended behavior change.

The earlier `_function_schema.py` change has been removed from this PR.

## Checklist

- [x] Targeted typecheck: `PYRIGHT_PYTHON_IGNORE_WARNINGS=1 uv run pyright pydantic_ai_slim/pydantic_ai/models/__init__.py`
- [ ] AI generated code (please check this box in the UI)


DouweM:
> * **`_function_schema.py`** — Replace `# TODO warn?` with an actual `UserWarning` when a tool function parameter is missing a type annotation. This helps users catch accidental untyped parameters early.

This is not actually currently happening? Nor should we just pull that in in a chore PR, that requires more discussion (and tests etc). I'm fine with the missing type annotation fixes.

adtyavrdhn (approved):
Removed unrelated bits to what the PR said it would do. I wasn't sure about getting the type_hints like that and if needed can be followed up as a separate PR

PR #6056, opened 2026-06-24
fix(groq): drop non-native ThinkingParts from history to prevent reasoning leak

Julien-ser:
Closes the groq surface of #5927 (xAI surface fixed in #5936, Anthropic/Bedrock fixed in #5920).

## Problem

`GroqModel._map_messages` unconditionally re-wrapped **any** `ThinkingPart` in multi-turn history with `<think>…</think>` tags — including parts produced by other providers (e.g. Anthropic) or native groq parts with no provenance information.

The model may then mimic those tags verbatim in its next response, leaking reasoning content into the user-visible answer. The leak is stochastic and builds up over many persisted turns (as noted in the #5927 original report).

## Fix

Guard the re-wrap behind a new `GroqModelProfile` field:

```python
groq_send_back_thinking_parts: bool  # default False
```

When `False` (default), `ThinkingPart`s without a provider signature are silently dropped from the assistant turn. Operators who explicitly want the old behaviour can opt in:

```python
profile = GroqModelProfile(groq_send_back_thinking_parts=True, ...)
m = GroqModel('qwen/qwen3-32b', profile=profile)
```

## Changes

| File | Change |
|------|--------|
| `profiles/groq.py` | Add `groq_send_back_thinking_parts: bool` to `GroqModelProfile` |
| `models/groq.py` | Guard `ThinkingPart` re-render in `_map_messages` with the flag |
| `tests/models/test_groq.py` | Add `call_args_list` to `MockGroq`; add two regression tests |

## Tests

- `test_groq_foreign_thinking_part_dropped_from_history` — verifies that a `ThinkingPart` with `provider_name='anthropic'` is silently excluded from the API call by default.
- `test_groq_foreign_thinking_part_preserved_when_flag_enabled` — verifies the opt-in flag restores the old wrapping behaviour.

This mirrors the pattern used for xAI (`grok_send_back_thinking_parts`) in #5936 and Anthropic/Bedrock in #5920.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6056?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/groq.py
+++ b/pydantic_ai_slim/pydantic_ai/models/groq.py
@@ -522,8 +522,9 @@ class GroqModel(Model[AsyncGroq]):
                     elif isinstance(item, ToolCallPart):
                         tool_calls.append(self._map_tool_call(item))
                     elif isinstance(item, ThinkingPart):
-                        start_tag, end_tag = self.profile.get('thinking_tags', DEFAULT_THINKING_TAGS)
-                        texts.append('\n'.join([start_tag, item.content, end_tag]))
+                        if item.content and self.profile.get('groq_send_back_thinking_parts', False):
+                            start_tag, end_tag = self.profile.get('thinking_tags', DEFAULT_THINKING_TAGS)
+                            texts.append('\n'.join([start_tag, item.content, end_tag]))
                     elif isinstance(item, NativeToolCallPart | NativeToolReturnPart):  # pragma: no cover
                         # These are not currently sent back
                         pass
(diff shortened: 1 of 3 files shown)


itxaiohanglover:
Nice fix! Defaulting to drop ThinkingParts from Groq history prevents reasoning-content leak. The opt-in profile flag groq_send_back_thinking_parts gives users control when they explicitly want it. Good test coverage.

dsfaccini:
David's AICA here: Thanks @Julien-ser — folding this into #5920 alongside the rest of the #5927 family (anthropic/bedrock + xai #5936 + huggingface #6122). Your fix is preserved with a `Co-authored-by` credit; `groq_send_back_thinking_parts` is now `Literal['auto','tags']` (default `'auto'`) instead of a `bool`, to match the rest of the family. Closing in favor of #5920 — thank you!

PR #6066, opened 2026-06-25
fix(test): set `provider_name` on `TestModel` non-streaming responses

### LINKED ISSUE #6062: Inconsistent metadata in `TestModel`: `agent.run()` vs `agent.run_stream()` results
### Initial Checks

### Description

When using `pydantic_ai` with `TestModel`, the result returned by `agent.run()` does not expose `provider_name`, but the streamed result from `agent.run_stream()` does.

This suggests that result metadata is not consistent between the two execution paths, and other fields may also differ between them.

```Python
import asyncio

from pydantic_ai import Agent, AgentRunResult, ModelResponse


def print_response(result: AgentRunResult):
    for message in result.all_messages():
        if isinstance(message, ModelResponse):
            print(message)
            print(message.provider_name)
            print()


async def main():
    agent = Agent(model=TestModel())

    print("with `agent.run`:")
    result = await agent.run("hello")
    print_response(result)

    print("with `agent.run_stream`:")
    async with agent.run_stream("hello") as stream:
        await stream.get_output()
        print_response(stream)


asyncio.run(main())
```

### Logfire Trace

_No response_

- Python: 3.13.12
- Pydantic AI: 2.0.0
- LLM provider SDK: -

syf2211:
Fixes #6062.

## Summary

- Set `provider_name` on `TestModel.request()` responses so non-streaming runs expose the same provider metadata as streamed runs.
- Update affected snapshots and docs to include `provider_name='test'`.
- Adds a test covering the `ModelResponse` returned by both run paths and comparing.

## Tests

- `uv run pytest tests/models/test_model_test.py::test_response_metadata_consistent_between_run_and_run_stream -q`
- `uv run ruff check tests/models/test_model_test.py`
- `PYRIGHT_PYTHON_IGNORE_WARNINGS=1 uv run pyright tests/models/test_model_test.py`


itxaiohanglover:
Good fix! Setting provider_name on TestModel responses makes it consistent with real model behavior. The docs update to include provider_name in test expectations is the right call.

adtyavrdhn (approved):
Fixed the test structure, looks good now

PR #6076, opened 2026-06-26
fix(dbos,prefect): preserve runtime `toolsets` in durable runs

### LINKED ISSUE #6075: DBOSAgent and PrefectAgent drop runtime `toolsets` in durable runs
## Bug

Two durable wrappers silently drop per-run `toolsets=` inside durable runs:

- `DBOSAgent`
- `PrefectAgent`

That means runtime toolsets (for example an `ExternalToolset` supplied by `AGUIAdapter`) never reach the model.

## Reproducer

```python
from pydantic_ai import Agent, ExternalToolset
from pydantic_ai.tools import DeferredToolRequests, ToolDefinition
from pydantic_ai.durable_exec.dbos import DBOSAgent
from pydantic_ai.durable_exec.prefect import PrefectAgent

agent = Agent('openai:gpt-4o', name='repro', output_type=[str, DeferredToolRequests])
dbos_agent = DBOSAgent(agent)
prefect_agent = PrefectAgent(agent)

frontend = ExternalToolset([ToolDefinition(name='frontend_lookup', parameters_json_schema={'type': 'object'})])

# Plain agent: returns DeferredToolRequests(calls=[ToolCallPart('frontend_lookup', ...)])
result = await agent.run('run frontend tool', toolsets=[frontend])
assert isinstance(result.output, DeferredToolRequests)  # passes

# DBOS agent: runtime toolsets are dropped
result = await dbos_agent.run('run frontend tool', toolsets=[frontend])
assert isinstance(result.output, DeferredToolRequests)  # fails

# Prefect agent: runtime toolsets are dropped
result = await prefect_agent.run('run frontend tool', toolsets=[frontend])
assert isinstance(result.output, DeferredToolRequests)  # fails
```

## Cause

Both wrappers set an override toolset list containing only constructor-time durable toolsets:

- `DBOSAgent._dbos_overrides()`
- `PrefectAgent._prefect_overrides()`

The base agent only appends per-run `additional_toolsets` when `self._override_toolsets.get() is None`.

Once the durable wrapper sets an override toolset list, the runtime `toolsets=` argument is ignored. The model never sees the runtime `ExternalToolset`, so it cannot emit the expected deferred/external tool call.

## Why Temporal Is Different

`TemporalAgent` explicitly rejects runtime toolsets inside workflows today, and its docs/tests reflect that.

That looks intentional rather than buggy: the Temporal wrapper pre-registers durable activities for its model/tool execution surface up front, so it expects that durable surface to be known at agent/worker startup.

So this issue is about wrappers where runtime toolsets appear intended to work, but were being dropped accidentally.

## Impact

This breaks the documented `ExternalToolset` / deferred tool flow in durable runs, including AG-UI frontend tool integration where tools are supplied per-run by the adapter.

Trollgeir:
- Closes #6075

### Summary

This PR fixes the same runtime `toolsets=` override bug in two durable wrappers:

- `DBOSAgent`
- `PrefectAgent`

In both wrappers, the durable override context replaced the agent's effective toolsets with only the constructor-time durable toolsets. Because the base agent ignores per-run `additional_toolsets` whenever toolsets are already overridden, runtime `toolsets=` were silently dropped inside durable runs.

That broke cases like runtime `ExternalToolset` usage, where the model should be able to return `DeferredToolRequests` for externally-executed tools.

The fix is the same in both wrappers:

- merge constructor-time durable toolsets with per-run `toolsets=` inside the wrapper override context
- keep passing `toolsets=toolsets` through `run()` / `run_sync()` so the runtime toolsets still reach `iter()`
- pass `toolsets=None` from the durable wrapper's `iter()` into `super().iter()` because the merged override is already active there, avoiding double-application

### Why Temporal Is Not Changed

`TemporalAgent` is intentionally different.

Inside a Temporal workflow, runtime toolsets are explicitly rejected today and covered by tests/docs. That wrapper pre-registers durable activities for its model/tool execution surface up front, and the worker/plugin model depends on those activities being known ahead of workflow execution.

So this PR only fixes the wrappers where runtime toolsets are intended to work but were being dropped accidentally.

### Verification

- Added a DBOS regression test for runtime `ExternalToolset` support in `tests/test_dbos.py`
- Confirmed that test fails without the DBOS fix and passes with it restored
- Added a matching Prefect regression test in `tests/test_prefect.py`
- Verified the Prefect behavior with a direct runtime repro returning `DeferredToolRequests(...)`
- `tests/test_prefect.py` is currently blanket-skipped upstream, so the Prefect verification could not rely on the module test runner yet

### Checklist


Trollgeir:
@qianl15 FYI :)

Trollgeir:
@desertaxle FYI :)

qianl15:
> @qianl15 FYI :)

Thanks for the fix!

dsfaccini:
would be superseded by #4977 (https://pydanticlogfire.slack.com/archives/C083V7PMHHA/p1782724473026419) so priority should be there, and this PR can be closed when that merges

PR #6084, opened 2026-06-27
test: use running loop for tenacity timing

Ghraven:
Problem: an async tenacity integration test measures elapsed retry time with `asyncio.get_event_loop().time()`. Since the test is already running inside an active event loop, `asyncio.get_running_loop()` is the more explicit API.

Before / after: before, the test asked the event loop policy for a loop before and after the async request. After, it reads time from the currently running loop while preserving the same retry timing assertion.

Verification:
- `python -m py_compile tests\test_tenacity.py`
- Tried `python -m pytest tests\test_tenacity.py -k "async_transport_with_wait_retry_after"`, but local collection stops before executing tests because OpenTelemetry emits a deprecation warning that this environment treats as an error during import.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6084?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


hramezani:
Thanks @Ghraven

PR #6088, opened 2026-06-27
fix(utils): run sync callbacks inline on Pyodide/emscripten by default

### LINKED ISSUE #6086: Run sync tool/callback functions inline on Pyodide/emscripten instead of offloading to a thread
David's AICA here:

### Summary

Running an agent in the browser via Pyodide (`sys.platform == 'emscripten'`, which has no thread support) crashes whenever a **synchronous** tool function (or other sync callback) is invoked:

```
RuntimeError: can't start new thread

Pyodide cannot start Python threads.
```

Reproduced live on [play.pydantic.work](https://play.pydantic.work) (pydantic-ai 2.0.0): an agent with a sync `@agent.tool_plain` / `Tool(fn)` fails the moment the model calls the tool. Making the tool `async def` works around it. The playground already prints a helpful hint pointing at the async workaround.

### Cause

Sync callbacks are dispatched through `_utils.run_in_executor`, whose default branch calls `anyio.to_thread.run_sync`, which starts a worker thread:

`_function_schema.py:FunctionSchema.call` (sync branch) → `_utils.run_in_executor` → `anyio.to_thread.run_sync` → `threading.Thread.start()` → `RuntimeError: can't start new thread`.

Async callbacks are awaited directly and are unaffected. The same crash applies to every user-supplied sync surface that routes through `run_in_executor`: sync tools, sync output validators/functions, sync dynamic system prompts (`_system_prompt.py`), sync history processors (`capabilities/process_history.py`), and the UI `on_complete` callback (`ui/_event_stream.py`).

### Proposed fix

pyai already has an inline-execution path: the `_disable_threads` ContextVar (added for Temporal in #3704), which `run_in_executor` honors by calling the function inline. Default it to `True` on emscripten:

```python
# _utils.py  (sys already imported)
_disable_threads: ContextVar[bool] = ContextVar(
    '_disable_threads', default=sys.platform == 'emscripten'
)
```

That's the whole functional change. On CPython the default stays `False`, so behavior is byte-for-byte unchanged; on Pyodide, sync callbacks run inline. Inline execution blocks the single Pyodide thread for the callback's duration, but that is unavoidable in a single-threaded runtime — the current alternative is a hard crash. It reuses the existing, already-tested code path and is gated to emscripten, so `ThreadExecutor` / `using_thread_executor` keep composing normally on CPython.

### Out of scope (note for reviewers)

A few provider-internal call sites use `anyio.to_thread.run_sync` **directly** rather than via `run_in_executor` (`models/bedrock.py`, `embeddings/bedrock.py`, `common_tools/duckduckgo.py`) and would still start threads — but they aren't on the browser sync-tool path. A separate public `DisableThreads` capability (mirroring `ThreadExecutor` / `using_thread_executor`) could be added later for non-emscripten threadless hosts, but isn't needed for this fix.

Happy to open a PR with the one-line change plus a regression test alongside the existing `disable_threads` tests in `tests/test_utils.py`.

syf2211:
## Summary

Default `_disable_threads` to `True` when `sys.platform == 'emscripten'`, so sync tools and callbacks use the existing inline execution path instead of `anyio.to_thread.run_sync`, which crashes on Pyodide with `RuntimeError: can't start new thread`.

## Motivation

Running agents in the browser via Pyodide fails whenever a synchronous tool (or other sync callback routed through `run_in_executor`) is invoked. Async tools work around the issue, but sync tools are a common authoring style and should not hard-crash on emscripten.

Fixes #6086

## Changes

| File | Change |
|------|--------|
| `pydantic_ai_slim/pydantic_ai/_utils.py` | Default `_disable_threads` to `sys.platform == 'emscripten'`; clarify `disable_threads()` docstring |
| `tests/test_utils.py` | Add regression tests for emscripten default inline execution and non-emscripten default |

## Tests

- `uv run pytest tests/test_utils.py::test_disable_threads_defaults_false_on_non_emscripten tests/test_utils.py::test_run_in_executor_runs_inline_by_default_on_emscripten tests/test_utils.py::test_run_in_executor_with_disable_threads tests/test_utils.py::test_disable_threads_takes_priority_over_custom_executor` — all passed
- `uv run ruff check` on changed files — passed
- `uv run pyright pydantic_ai_slim/pydantic_ai/_utils.py` — passed

## Notes

- CPython behavior is unchanged (`default=False` on non-emscripten platforms).
- Reuses the existing `_disable_threads` / `disable_threads()` path already used by Temporal workflows.
- Provider-internal call sites that call `anyio.to_thread.run_sync` directly (e.g. Bedrock) are out of scope for this fix, as noted in the issue.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6088?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -13,6 +13,7 @@ from importlib.metadata import distributions
 
 import pytest
 
+import pydantic_ai._utils as utils_module
 from pydantic_ai import UserError
 from pydantic_ai._utils import (
     UNSET,
@@ -25,7 +26,6 @@ from pydantic_ai._utils import (
     strip_markdown_fences,
     using_thread_executor,
 )
-import pydantic_ai._utils as utils_module
 
 from ._inline_snapshot import snapshot
 from .models.mock_async_stream import MockAsyncStream
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -251,14 +251,23 @@ async def test_disable_threads_takes_priority_over_custom_executor() -> None:
         executor.shutdown(wait=True)
 
 
-def test_disable_threads_defaults_false_on_non_emscripten(monkeypatch: pytest.MonkeyPatch) -> None:
+async def test_disable_threads_defaults_false_on_non_emscripten(monkeypatch: pytest.MonkeyPatch) -> None:
     import importlib
 
     import pydantic_ai._utils as utils_module
 
     monkeypatch.setattr('sys.platform', 'linux')
     importlib.reload(utils_module)
-    assert utils_module._disable_threads.get() is False
+    try:
+        main_thread = threading.current_thread()
+
+        def check_thread() -> threading.Thread:
+            return threading.current_thread()
+
+        result = await utils_module.run_in_executor(check_thread)
+        assert result is not main_thread
+    finally:
+        importlib.reload(utils_module)
 
 
 async def test_run_in_executor_runs_inline_by_default_on_emscripten(monkeypatch: pytest.MonkeyPatch) -> None:
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -3,7 +3,9 @@ from __future__ import annotations as _annotations
 import asyncio
 import contextvars
 import functools
+import importlib
 import os
+import sys
 import threading
 from collections.abc import AsyncIterator
 from concurrent.futures import ThreadPoolExecutor
@@ -23,6 +25,7 @@ from pydantic_ai._utils import (
     strip_markdown_fences,
     using_thread_executor,
 )
+import pydantic_ai._utils as utils_module
 
 from ._inline_snapshot import snapshot
 from .models.mock_async_stream import MockAsyncStream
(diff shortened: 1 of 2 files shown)


itxaiohanglover:
Clean fix! Defaulting _disable_threads to True on emscripten is the right approach — Python threads are unavailable there. The test with monkeypatch + importlib.reload verifies both emscripten and non-emscripten behavior.

tests/test_utils.py
```diff
@@ -251,6 +251,45 @@ def check_thread() -> threading.Thread:
         executor.shutdown(wait=True)
 
 
+async def test_disable_threads_defaults_false_on_non_emscripten(monkeypatch: pytest.MonkeyPatch) -> None:
+    import importlib
+
+    import pydantic_ai._utils as utils_module
```
dsfaccini: imports go up top

dsfaccini (changes requested):

syf2211:
Addressed review feedback: moved `importlib`, `sys`, and `pydantic_ai._utils` imports to the top of `tests/test_utils.py`. Pushed in 7dddd6b.

syf2211:
Hi @dsfaccini — pushed import cleanup in 5082dae (moved `importlib`, `sys`, and `pydantic_ai._utils` imports to the top of `tests/test_utils.py`). Could you take another look when you have a moment? Thanks!

syf2211:
Thanks for the review! Moved `import importlib` and `import pydantic_ai._utils as utils_module` to the top of `tests/test_utils.py` in 5082dae. Ready for another look when you have a moment.

syf2211:
@dsfaccini Thanks for the review — imports have been moved to the module top in the latest commit. Could you take another look when you have a moment?

PR #6052, opened 2026-06-24
fix(`GoogleModel`): include `CodeExecutionTool` in `include_server_side_tool_invocations` check

aarsh767:
- Closes #6051

## Summary

Combining `CodeExecutionTool` with function tools on Gemini 3 raised an HTTP 400:

```
google.genai.errors.ClientError: 400 INVALID_ARGUMENT
Please enable tool_config.include_server_side_tool_invocations
to use Built-in tools with Function calling.
```

`_get_tool_config` already set `include_server_side_tool_invocations` for `WebSearchTool`, `WebFetchTool`, and `FileSearchTool`, but `CodeExecutionTool` was missing from the `isinstance` guard. One-line fix: add it to the tuple.

## Changes

- `pydantic_ai_slim/pydantic_ai/models/google.py` — add `CodeExecutionTool` to the `emits_tool_call_invocations` check; update comment to reflect that the flag applies to all built-in tools (not just the three previously listed)
- `tests/models/test_tool_choice_unit.py` — new parametrized unit-test case `code-execution-gemini-3-sets-server-side-flag` asserting the request shape directly (VCR can't catch a malformed request since it replays a recorded response without re-validating against the API)

### Checklist

- [x] Any **AI generated code** has been reviewed line-by-line by the human PR author, who stands by it.
- [x] No **breaking changes** in accordance with the [version policy](https://github.com/pydantic/pydantic-ai/blob/main/docs/version-policy.md).
- [x] **PR title** is fit for the [release changelog](https://github.com/pydantic/pydantic-ai/releases).

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6052?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #6096, opened 2026-06-27
fix(google): enable `include_server_side_tool_invocations` for `CodeExecutionTool`

Hasnaathussain:
On Gemini 3 models, combining the built-in `CodeExecutionTool` with any user-defined function tool in the same request causes the Gemini API to reject the request with HTTP 400:
`Please enable tool_config.include_server_side_tool_invocations to use Built-in tools with Function calling.`

This happens because `GoogleModel._get_tool_config` only sets `include_server_side_tool_invocations` when one of `(WebSearchTool, WebFetchTool, FileSearchTool)` is present, missing `CodeExecutionTool`.

This PR fixes the bug by:
1. Adding `CodeExecutionTool` to the tuple of native tools checked for `include_server_side_tool_invocations`.
2. Updating the unit tests in `tests/models/test_tool_choice_unit.py` to assert that `CodeExecutionTool` correctly triggers this configuration.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6096?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/google.py
+++ b/pydantic_ai_slim/pydantic_ai/models/google.py
@@ -712,11 +712,11 @@ class GoogleModel(Model[Client]):
             tool_config['function_calling_config'] = function_calling_config
 
         # `include_server_side_tool_invocations` makes Gemini emit explicit `tool_call`/`tool_response`
-        # parts for WebSearchTool, WebFetchTool, FileSearchTool. Pre-Gemini-3 models reject the field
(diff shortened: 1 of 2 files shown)


DouweM:
Dupe of #6052

PR #6098, opened 2026-06-27
fix(bedrock): handle `toolResult` attachment co-location per-model via `bedrock_tool_result_colocatable_content`

### LINKED ISSUE #6081: Bedrock: an attachment (`DocumentUrl`/`ImageUrl`) merged into the same user message as a `tool_result` → Converse `ValidationException`
### Initial Checks

### Description

When a tool call's ToolReturnPart is the last message before a new user turn that carries an attachment (DocumentUrl / ImageUrl / BinaryContent), BedrockConverseModel merges them into a single Converse user message containing both a toolResult block and a document block. Anthropic Claude models on Bedrock reject it with a ValidationException whose wording is misleading — it reports a missing tool_result, but the tool_result is present; it's just co-located with an attachment:

```
botocore.errorfactory.ValidationException: An error occurred (ValidationException) when calling the Converse operation: The model returned the following errors: messages.N: `tool_use` ids were found without `tool_result` blocks immediately after: tooluse_xxx. Each `tool_use` block must have a corresponding `tool_result` block in the next message.
```

Cause. _clean_message_history merges consecutive ModelRequests and BedrockConverseModel._map_messages merges consecutive Bedrock user messages. A trailing ToolReturnPart request followed by an attachment-bearing UserPromptPart request therefore become one user message → content = [toolResult, text, document].

Empirically the constraint is specifically attachment-next-to-tool-result:

- [toolResult, text] → accepted
- [toolResult, text, document] → rejected

This shape arises naturally in multi-turn apps: the model ends a turn with a tool call, the tool result is the last message, then the user's next turn includes a file. It works on OpenAI (id-based, order-independent tool matching) but fails on Bedrock.

Expected. The Bedrock adapter should not place an attachment block in the same Converse message as a tool_result — i.e. keep the attachment turn out of the merged tool_result message.

Note: this "no tool_result + document in one message" rule is a model-level constraint and isn't documented in the AWS Converse API reference (the ContentBlock array schema permits the mix), which is likely why the adapter doesn't guard against it.

```Python
import asyncio
import os

os.environ.setdefault("AWS_REGION", "us-east-1")

from pydantic_ai import DocumentUrl
    ModelRequest,
    ModelResponse,
    UserPromptPart,
    ToolCallPart,
    ToolReturnPart,
)
from pydantic_ai._agent_graph import _clean_message_history

model = BedrockConverseModel("us.anthropic.claude-opus-4-6-v1")

# A completed tool-call turn (the tool result is the last message),
# followed by a new user turn that carries a document attachment.
history = [
    ModelRequest(parts=[UserPromptPart(content="show me my expenses")]),
    ModelResponse(parts=[ToolCallPart(tool_name="get", args={}, tool_call_id="t1")]),
    ModelRequest(parts=[ToolReturnPart(tool_name="get", content="ok", tool_call_id="t1")]),
    ModelRequest(parts=[UserPromptPart(content=[
        "what accounts are in this?",
        DocumentUrl(url="s3://bucket/file.csv", media_type="text/csv"),
    ])]),
]


async def main():
    _system, messages = await model._map_messages(
        _clean_message_history(history), ModelRequestParameters(), None
    )
    for i, msg in enumerate(messages):
        print(i, msg["role"], [next(iter(block)) for block in msg["content"]])


asyncio.run(main())

# Output — the final user message co-locates a toolResult with a document:
#   0 user ['text']
#   1 assistant ['toolUse']
#   2 user ['toolResult', 'text', 'document']
# Sending this via client.converse(...) raises the ValidationException in the description.
```

### Logfire Trace

_No response_

Python: 3.11.15
pydantic-ai-slim: 1.60.0 (reproduced here; the responsible merge logic — _clean_message_history + Bedrock _map_messages consecutive-user merge — is unchanged on main/2.0)
boto3: 1.43.0, botocore: 1.43.0
Model: us.anthropic.claude-opus-4-6-v1 (Anthropic Claude on Amazon Bedrock)

Hasnaathussain:
Closes #6081

### Problem

On Bedrock, pydantic-ai merges consecutive user turns into one Converse message. When a tool-result turn is followed by a user turn carrying an attachment, they merge into one user message co-locating a `toolResult` block with the attachment — which some models reject with a misleading `ValidationException` about a missing `tool_result`.

The original fix split *every* such turn with a synthetic assistant message for *all* models. Live testing against Bedrock (`us-east-1`) shows the constraint is per-model, so a blanket split is both incomplete (misses models with a stricter rule) and over-broad (needlessly pollutes models with no rule).

### Evidence (live Converse, what may share a user message with a `toolResult`)

| Family | text | image | document (pdf) | document (csv/txt/docx) | video |
|---|---|---|---|---|---|
| **Anthropic** | ✅ | ✅ | ✅ | ❌ | ❌ (video unsupported) |
| **Amazon Nova** | ✅ | ✅ | ✅ | ✅ | ✅ |
| **Meta Llama 4** | ❌ | ❌ | ❌ | ❌ | ❌ |
| **Mistral (pixtral)** | ❌ | ❌ | ❌ | ❌ | ❌ |
| **Qwen** | ✅ | ✅ | — (no doc support) | — | — |

- **Anthropic** rejects documents and video beside a `toolResult`, but accepts text and images. (PDF happens to pass, but the pdf-vs-other-document split is undocumented and brittle, so we conservatively treat all documents as non-co-locatable.)
- **Nova** has no constraint at all — the old blanket fix inserted a pointless dummy turn here.
- **Llama and Mistral** reject *any* content sharing a `toolResult` turn, even plain text — a broader form of the same bug that the attachment-only guard missed.

Mechanism notes, also verified: a bare split into two consecutive user turns does **not** work (Bedrock re-merges same-role turns), so an alternating assistant turn is required; the separator must be non-whitespace (`.` works everywhere); block order within the message is irrelevant. AWS does not document the `toolResult`+document constraint — it's model behavior — which is why this is driven by empirical per-family testing.

### Fix

Add a `bedrock_tool_result_colocatable_content` model-profile flag listing the content-block kinds a model accepts alongside a `toolResult`. The merge pass splits the turns (separating them with a synthetic `.` assistant message) only when the combined content isn't co-locatable for that model; otherwise it merges as before.

| Family | flag value | effect |
|---|---|---|
| Anthropic | `{'text', 'image'}` | splits out documents + video |
| Meta, Mistral | `frozenset()` | `toolResult` isolated from all other content |
| everyone else (default) | all kinds | merges freely, no dummy turn |

The default is permissive, so no behavior change for uncovered families.

### Tests

- Per-family mapping unit tests pinning the exact reshaped sequence (Anthropic split, Nova merged, Llama isolated).
- VCR cassettes recorded against live Bedrock proving each family accepts its reshaped history: Anthropic with a **`text/plain`** document (the original PDF recording is accepted even *without* the fix, so it proved nothing — a text document actually reproduces the rejection), Nova co-located with no dummy turn, and Llama with text isolated.
- Provider-profile assertions for the new flag values.

### Follow-ups (separate PRs)

Live testing surfaced two adjacent, distinct defects that are handled in their own PRs rather than widening this one:
- `bedrock_supported_media_kinds_in_tool_returns` has wrong per-family values (Mistral/Writer accept `document` but reject `image` inside tool results; Qwen accepts neither; Nova rejects binary documents).
- The "conversation must start with a user message" workaround is provider-specific (Anthropic and Qwen accept a leading assistant turn) and currently applied as a blanket rule.


DouweM:
There's some related logic here:

https://github.com/pydantic/pydantic-ai/blob/e2bdafb8151ccad52ee6fb73d7a9fe612a28d552/pydantic_ai_slim/pydantic_ai/models/bedrock.py#L1297-L1303

Worth checking to see if we can make it more consistent / fix it in one place.

Hasnaathussain:
I have addressed this! I've updated the logic at the end of `_map_messages` to perform a clean pass over all final processed user messages. Any message that contains a `document` block but lacks a `text` block will now have a "See attached document(s)." prefix automatically inserted. This makes the text-field injection completely consistent and robust across user prompts, tool returns, and split/merged user turns. All 138 Bedrock tests pass locally.

adtyavrdhn (approved):

pydantic_ai_slim/pydantic_ai/providers/bedrock.py
```diff
@@ -141,6 +141,19 @@ class BedrockModelProfile(ModelProfile, total=False):
     """Default: `False`."""
     bedrock_supported_media_kinds_in_tool_returns: frozenset[str]
     """Default: `frozenset({'image'})`."""
+    bedrock_tool_result_colocatable_content: frozenset[str]
```
DouweM: Let's use a `Literal` union with the actual supported values, instead of `str`. That also goes for `supported_media_kinds...`.

PR #6108, opened 2026-06-28
Handle malformed tool search call args

### LINKED ISSUE #6106: search_tools malformed args raise uncaught ValidationError during ToolCallPart narrowing instead of ModelRetry
### Summary

When a model emits a `search_tools` (tool-search) call with the wrong argument shape, pydantic-ai raises an uncaught `pydantic_core.ValidationError` during **response-part narrowing**, which aborts the whole agent run. Narrowing happens while the model response is being assembled — *before* tool execution — so the normal "invalid tool args → `ToolRetryPromptPart` / `ModelRetry`" recovery never engages, and a recoverable model slip becomes a fatal error shown to the end user.

This is easy for a model to hit: `search_tools` takes `queries: list[str]`, but most ordinary tools take `query: str`, so models frequently emit `{"query": "..."}` (singular) or `{"queries": "noop"}` (a bare string instead of a list).

### Version

- pydantic-ai 1.99.0
- Python 3.12

### Minimal repro

```python

part = ToolCallPart(
    tool_name="search_tools",
    args={"query": "find a tool"},  # singular `query` instead of `queries: [...]`
    tool_call_id="c1",
)
ToolCallPart.narrow_type(part, tool_kind="tool-search")
```

raises:

```
pydantic_core._pydantic_core.ValidationError: 2 validation errors for nullable[union[str,ToolSearchArgs]]
str
  Input should be a valid string [type=string_type, input_value={'query': 'find a tool'}, input_type=dict]
ToolSearchArgs.queries
  Field required [type=missing, input_value={'query': 'find a tool'}, input_type=dict]
```

The same occurs for `{"queries": "noop"}` (a string where a `list[str]` is required).

### Where

`ModelResponsePartsManager._typed_call_part` → `ToolCallPart.narrow_type(..., tool_kind="tool-search")` → `pydantic_ai._tool_search._narrow_tool_search_call` → `_TOOL_SEARCH_CALL_ARGS_TA.validate_python(part.args)` (a `TypeAdapter[str | ToolSearchArgs | None]`).

`_typed_call_part` runs during streamed/processed-response assembly, so a `ValidationError` here propagates out of the model request before the agent graph reaches the tool-validation/execution node — i.e. it bypasses the retry machinery that recovers bad arguments for ordinary tools.

### Impact

Any agent using tool-search (the `ToolSearch` capability / `defer_loading=True`) will surface a fatal "unexpected error" to the end user whenever the model gets the `search_tools` arg shape wrong. In our production traffic this happens to multiple distinct users per day.

### Expected behavior

A malformed `search_tools` call should be **recoverable** the same way bad arguments to any other tool are: the model gets a retry prompt (`ModelRetry`) describing the expected `queries: list[str]` shape, rather than the run crashing.

### Possible fixes

1. In `_narrow_tool_search_call` / `_narrow_native_tool_search_call`, on `ValidationError` leave the part as a base `ToolCallPart` (don't promote) so it flows into the normal tool-arg validation → `ModelRetry` path; and/or
2. Parse leniently — coerce a singular `query` → `queries`, and wrap a scalar/string `queries` into a one-element list — before validating.

Happy to send a PR if you can point me at the preferred approach.

ft4453080:
- Closes #6106

### Summary

- Keep malformed tool-search call args as the original base call part during typed narrowing instead of raising a ValidationError.
- Apply the same fallback to local search_tools calls and native tool_search calls.
- Add regression coverage for malformed args shaped as {"query": ...} and {"queries": "noop"}.

### Checklist

- [x] No breaking changes in accordance with the version policy.
- [x] PR title is fit for the release changelog.

### Tests

- .venv/bin/python -m pytest tests/test_tool_search.py::test_narrow_type_promotes_builtin_call_to_tool_search tests/test_tool_search.py::test_narrow_type_builtin_call_with_invalid_args_returns_input_unchanged tests/test_tool_search.py::test_narrow_type_local_promotes_with_tool_kind_set tests/test_tool_search.py::test_narrow_type_local_with_invalid_args_returns_input_unchanged -q
- .venv/bin/python -m pytest tests/test_tool_search.py -k narrow_type -q
- .venv/bin/python -m ruff check pydantic_ai_slim/pydantic_ai/_tool_search.py tests/test_tool_search.py
- .venv/bin/python -m ruff format --check pydantic_ai_slim/pydantic_ai/_tool_search.py tests/test_tool_search.py

Note: full tests/test_tool_search.py -q was not run successfully in this minimal local environment because optional test dependencies/types are missing (brotli, ResponseToolSearchCall).

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/_tool_search.py
+++ b/pydantic_ai_slim/pydantic_ai/_tool_search.py
@@ -368,10 +368,13 @@ _TOOL_SEARCH_RETURN_CONTENT_TA: pydantic.TypeAdapter[ToolSearchReturnContent] = 
 )
 
 
-def _narrow_native_tool_search_call(part: NativeToolCallPart) -> NativeToolSearchCallPart:
+def _narrow_native_tool_search_call(part: NativeToolCallPart) -> NativeToolCallPart:
     if isinstance(part, NativeToolSearchCallPart):
         return part
-    validated_args = _TOOL_SEARCH_CALL_ARGS_TA.validate_python(part.args)
+    try:
+        validated_args = _TOOL_SEARCH_CALL_ARGS_TA.validate_python(part.args)
+    except pydantic_core.ValidationError:
+        return part
     return copy_dataclass_fields(part, NativeToolSearchCallPart, args=validated_args, tool_kind='tool-search')
 
 
@@ -382,10 +385,13 @@ def _narrow_native_tool_search_return(part: NativeToolReturnPart) -> NativeToolS
     return copy_dataclass_fields(part, NativeToolSearchReturnPart, content=validated_content, tool_kind='tool-search')
 
 
-def _narrow_tool_search_call(part: ToolCallPart) -> ToolSearchCallPart:
+def _narrow_tool_search_call(part: ToolCallPart) -> ToolCallPart:
     if isinstance(part, ToolSearchCallPart):
         return part
-    validated_args = _TOOL_SEARCH_CALL_ARGS_TA.validate_python(part.args)
+    try:
+        validated_args = _TOOL_SEARCH_CALL_ARGS_TA.validate_python(part.args)
+    except pydantic_core.ValidationError:
+        return part
     return copy_dataclass_fields(part, ToolSearchCallPart, args=validated_args, tool_kind='tool-search')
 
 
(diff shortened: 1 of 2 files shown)


DouweM:
Superseded by https://github.com/pydantic/pydantic-ai/pull/5873.

PR #6111, opened 2026-06-28
Align LLMJudge `_build_prompt` section order with its few-shot examples

### LINKED ISSUE #6110: LLMJudge: `_build_prompt` section order doesn't match the judge prompts' own few-shot examples
### Description

The default `LLMJudge` evaluator builds its judge prompt with `pydantic_evals.evaluators.llm_as_a_judge._build_prompt`. Each judge agent's system prompt primes the model with a section layout via few-shot examples, but `_build_prompt` emits the runtime sections in a **different order**, so the model is shown a format it was never primed on — and the rubric (the instruction) is placed *before* the expected output it has to reason about.

### Few-shot format (system prompts) vs. runtime format (`_build_prompt`)

| Judge function | Few-shot examples teach | `_build_prompt` actually emits |
|---|---|---|
| `judge_input_output_expected` | `Input → ExpectedOutput → Output → Rubric` | `Input → Output → Rubric → ExpectedOutput` |
| `judge_output_expected` | `ExpectedOutput → Output → Rubric` | `Output → Rubric → ExpectedOutput` |

The `judge_output` and `judge_input_output` variants (no expected output) already match their examples (`… → Rubric` last); only the two expected-output variants diverge, because `_build_prompt` appends `ExpectedOutput` *after* the `Rubric` instead of before `Output`.

### Reproduction

```python
from pydantic_evals.evaluators.llm_as_a_judge import _build_prompt

print(_build_prompt(output='O', rubric='R', inputs='I', expected_output='E'))
```

Output (note `ExpectedOutput` lands after `Rubric`):

```
<Input>
I
</Input>
<Output>
O
</Output>
<Rubric>
R
</Rubric>
<ExpectedOutput>
E
</ExpectedOutput>
```

…whereas the `judge_input_output_expected` system prompt's examples are formatted `<Input> … <ExpectedOutput> … <Output> … <Rubric>`.

### Expected behavior

`_build_prompt` should emit sections in the same order the few-shot examples demonstrate — `Input → ExpectedOutput → Output → Rubric` — so the runtime prompt matches what the model was primed with and the rubric (instruction) comes last, after all the context it applies to.

### Impact

Affects the framework-default judge used by `LLMJudge(include_expected_output=True)` and the `judge_*_expected` helpers. A self-inconsistent prompt (examples in one layout, real data in another) can degrade judging quality, particularly for expected-output comparisons.

I'm happy to send a PR reordering `_build_prompt` (and updating the affected snapshot tests).

immuhammadfurqan:
Fixes #6110.

### Problem

The default `LLMJudge` evaluator builds its prompt with `_build_prompt`, and each expected-output judge agent primes the model with a section layout via few-shot examples. The two were inconsistent: the few-shot examples put `ExpectedOutput` before `Output`, while `_build_prompt` emitted `ExpectedOutput` after the `Rubric`. So the model saw a layout it was never primed on, and the rubric (the instruction) came before the expected output it has to reason about.

### Fix

Following maintainer feedback, the section order is aligned to the function names (`judge_input_output_expected`): `Input → Output → ExpectedOutput → Rubric`, with the rubric last so the instruction follows all the context it applies to. Both the few-shot examples (in `judge_input_output_expected` and `judge_output_expected`) and `_build_prompt` were reordered to that layout. The `judge_output` / `judge_input_output` variants (no expected output) were already correct and are unchanged.

### Tests

- Updated the affected snapshot tests in `test_llm_as_a_judge.py` to the new order.
- Added `test_build_prompt_section_order_matches_few_shot_examples`, a parametrized regression test pinning `_build_prompt`'s section order for all four variants.

`tests/evals/test_llm_as_a_judge.py` + `tests/evals/test_evaluator_common.py` pass; `ruff format`/`ruff check` and `pyright` are clean.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6111?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_evals/pydantic_evals/evaluators/llm_as_a_judge.py
+++ b/pydantic_evals/pydantic_evals/evaluators/llm_as_a_judge.py
@@ -250,16 +250,22 @@ def _build_prompt(
     inputs: Any | None = None,
     expected_output: Any | None = None,
 ) -> str | Sequence[str | UserContent]:
-    """Build a prompt that includes input, output, expected output, and rubric."""
+    """Build a prompt that includes input, expected output, output, and rubric.
+
+    Sections are emitted in the same order the judge agents' system-prompt few-shot
+    examples demonstrate — `Input → ExpectedOutput → Output → Rubric` — so the runtime
+    prompt matches the format the model was primed with and the rubric (the instruction)
+    comes last, after all the context it applies to.
+    """
     sections: list[str | UserContent] = []
     if inputs is not None:
         sections.extend(_make_section(inputs, 'Input'))
 
-    sections.extend(_make_section(output, 'Output'))
-    sections.extend(_make_section(rubric, 'Rubric'))
-
     if expected_output is not None:
         sections.extend(_make_section(expected_output, 'ExpectedOutput'))
+
+    sections.extend(_make_section(output, 'Output'))
+    sections.extend(_make_section(rubric, 'Rubric'))
     if all(isinstance(section, str) for section in sections):
         return '\n'.join(sections)  # type: ignore[arg-type]
     return sections
--- a/pydantic_evals/pydantic_evals/evaluators/llm_as_a_judge.py
+++ b/pydantic_evals/pydantic_evals/evaluators/llm_as_a_judge.py
@@ -121,14 +121,14 @@ _judge_input_output_expected_agent = Agent(
         Examples:
 
         <Input>What color is the sky?</Input>
-        <ExpectedOutput>Blue</ExpectedOutput>
         <Output>Cerulean</Output>
+        <ExpectedOutput>Blue</ExpectedOutput>
         <Rubric>The output is consistent with the expected output but doesn't have to match exactly</Rubric>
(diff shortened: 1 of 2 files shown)


DouweM:
@immuhammadfurqan Any particular reason you're changing the order in the prompt, rather than order in the the few-shot examples? The latter matches the function names (`judge_input_output_expected`).

immuhammadfurqan:
Good call — you're right. The `judge_input_output_expected` name is the source of truth for the intended order, so I've flipped the fix: instead of reshuffling the runtime prompt to match the examples, I reordered the few-shot examples (and `_build_prompt`) to `Input → Output → ExpectedOutput → Rubric`. Rubric stays last so the instruction follows all the context it applies to. Snapshots regenerated.

PR #6114, opened 2026-06-28
Mention GEMINI_API_KEY in GoogleProvider missing key error

onyx679:
﻿## Summary

- Add coverage for the `GoogleProvider` missing API key error.
- Update the error message and constructor docs to mention both `GOOGLE_API_KEY` and backwards-compatible `GEMINI_API_KEY`.

## Why

`GoogleProvider` reads the API key from `GOOGLE_API_KEY` or `GEMINI_API_KEY`, keeping the Gemini variable for backwards compatibility. When neither variable is present, the missing-key error only mentioned `GOOGLE_API_KEY`, which could mislead users who still use the compatible `GEMINI_API_KEY` configuration.

## Tests

- `python -m pytest tests/providers/test_google.py -q`
- `python -m pytest tests/providers/test_google.py tests/providers/test_provider_names.py -k "google" -q`
- `python -m ruff check pydantic_ai_slim\pydantic_ai\providers\google.py tests\providers\test_google.py`
- `python -m py_compile pydantic_ai_slim\pydantic_ai\providers\google.py tests\providers\test_google.py`
- `git diff --check`

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6114?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/providers/google.py
+++ b/pydantic_ai_slim/pydantic_ai/providers/google.py
@@ -108,7 +108,8 @@ class GoogleProvider(BaseGoogleProvider):
 
         Args:
             api_key: The [API key](https://ai.google.dev/gemini-api/docs/api-key) to
-                use for authentication. It can also be set via the `GOOGLE_API_KEY` environment variable.
+                use for authentication. It can also be set via the `GOOGLE_API_KEY` or `GEMINI_API_KEY`
+                environment variable.
             client: A pre-initialized client to use.
             http_client: An existing `httpx.AsyncClient` to use for making HTTP requests.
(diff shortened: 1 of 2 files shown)


dsfaccini:
we prefer users only use one key, which is the google variant

PR #6120, opened 2026-06-29
fix(google): handle list content in code execution tool return parts

Quratulain-bilal:
## Summary

  Fix Google native tool return mapping so `CodeExecutionTool` results are handled when the provider
  returns list-shaped content instead of only dict-shaped content.

  This prevents code execution responses with list payloads from failing to map into
  `code_execution_result`, which could break model response processing.

  ## Changes

  - Support both list and dict content when mapping `code_execution_result` parts.
  - Add regression tests for both list-shaped and dict-shaped code execution results.

  ## Verification

  - `pytest tests/models/test_google.py -k code_execution_list_content -q`
  - `pytest tests/models/test_google.py -k code_execution_dict_content -q`

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6120?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/google.py
+++ b/pydantic_ai_slim/pydantic_ai/models/google.py
@@ -1661,7 +1661,7 @@ def _native_tool_return_part_dict(
 ) -> PartDict | None:
     if item.provider_name not in accepted_provider_names:
         return None
-    if item.tool_name == CodeExecutionTool.kind and isinstance(item.content, dict):
+    if item.tool_name == CodeExecutionTool.kind and isinstance(item.content, dict | list):
         return _attach_signature(
             {'code_execution_result': cast(CodeExecutionResultDict, item.content)},  # pyright: ignore[reportUnknownMemberType]
             signature,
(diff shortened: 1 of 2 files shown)


adtyavrdhn:
Thanks for the contribution! Closing this as the underlying premise doesn't hold — this was investigated in #5614 (closed as invalid) and the same change was previously proposed in #5616 and #5622, both closed after review.

In short: Google's `CodeExecutionResult` is always an object (`{id, outcome, output}`) — `_map_code_execution_result()` calls `.model_dump()` on the SDK's Pydantic model, which always yields a dict, and the `provider_name` gate at the top of `_native_tool_return_part_dict()` prevents parts from other providers (where list content *is* possible, e.g. Anthropic's bash code execution) from ever reaching this check. The list-shaped payload in the tests (`bash_code_execution_result`) is Anthropic's shape, which Google's SDK never produces. Widening the guard to `dict | list` would silently pass malformed data to the Gemini API instead of failing fast.

See the closing analysis on #5614 for the full reasoning.

PR #6134, opened 2026-06-29
Exclude first-class token counts from Anthropic `RequestUsage.details` to fix OTel double-counting

### LINKED ISSUE #6131: Anthropic adapter leaks first-class usage fields into details, causing `output_tokens` double-counting in OTel consumers
### Initial Checks

### Description

The Anthropic adapter's `_extract_usage_details` copies raw API fields — including `output_tokens` and `input_tokens` — into the `details` dict on `RequestUsage`. When `opentelemetry_attributes()` later emits both first-class attributes (`gen_ai.usage.output_tokens`) and detail attributes (`gen_ai.usage.details.output_tokens`), downstream consumers like Langfuse sum them — **doubling reported output tokens and cost**.

This is distinct from #4364 (cache input token semantics). The input-token cache debate was about whether `input_tokens` should include cache — a semantic question where the OTel spec takes a clear position. This issue is about **emitting the same value twice under two different OTel attribute namespaces**, which no consumer can reasonably deduplicate.

**Code path:**

1. Anthropic API returns `usage.output_tokens = 3946`
2. `_extract_usage_details` (anthropic.py, ~line 1764) copies it into `details`:
   ```python
   for key in _COMPACTION_TOKEN_KEYS:  # includes 'output_tokens', 'input_tokens'
       if isinstance((value := getattr(response_usage, key, None)), int):
           details[key] = value
   ```
3. `_map_usage` passes `details=details` to `RequestUsage.extract()`, which also extracts `output_tokens=3946` as a first-class field via genai-prices
4. `opentelemetry_attributes()` (usage.py) emits:
   - `gen_ai.usage.output_tokens = 3946` (from `self.output_tokens`)
   - `gen_ai.usage.details.output_tokens = 3946` (from `self.details`)
5. Langfuse ingests both → `usage.output = 7892` (2x the actual value)

**The OpenAI adapter already prevents this** — it explicitly filters first-class keys from details before passing them to `extract()`:

```python
# openai.py, ~line 3670
{
    k: v
    for k, v in usage_data.items()
    if k not in {'prompt_tokens', 'completion_tokens', 'input_tokens', 'output_tokens', 'total_tokens'}
    if isinstance(v, int)
}
```

The Anthropic adapter doesn't do this filtering — it's an inconsistency between the two adapters.

**Evidence from production Langfuse data (self-hosted, post-PR langfuse/langfuse#13110):**

| Model | Provider | `usage.output` | `usageDetails.output` | Ratio |
|---|---|---|---|---|
| claude-sonnet-4-6 | Direct Anthropic | 7892 | 3946 | 2.0x |
| claude-sonnet-4-6 | Bedrock | 3946 | 3946 | 1.0x |

Bedrock is unaffected because `BedrockConverseModel` constructs `RequestUsage` directly without a `details` dict.

### Suggested Fix

Strip first-class `RequestUsage` field names from the `details` dict in `_extract_usage_details` (or in `_map_usage` before passing to `extract()`), mirroring the OpenAI adapter's approach. The `details` dict should only contain supplementary information (compaction iteration counts, thinking tokens, etc.), not duplicates of fields already on `RequestUsage`.

```python
_FIRST_CLASS_USAGE_KEYS = frozenset({
    'input_tokens', 'output_tokens',
    'cache_creation_input_tokens', 'cache_read_input_tokens',
})

# In _map_usage, before calling extract():
details = {k: v for k, v in details.items() if k not in _FIRST_CLASS_USAGE_KEYS}
```

### Workaround

Monkey-patch `RequestUsage.opentelemetry_attributes` to strip overlapping detail keys before emission:

```python

_FIRST_CLASS_KEYS = frozenset({"input_tokens", "output_tokens", "cache_creation_input_tokens", "cache_read_input_tokens"})
_original = RequestUsage.opentelemetry_attributes

def _patched(self):
    original_details = self.details
    self.details = {k: v for k, v in self.details.items() if k not in _FIRST_CLASS_KEYS}
    try:
        return _original(self)
    finally:
        self.details = original_details

RequestUsage.opentelemetry_attributes = _patched
```

### Related

- #4364 — Anthropic cache tokens double-counted (closed as not-a-bug; different issue — cache semantics vs. pure duplication)
- langfuse/langfuse#13110 — Langfuse fix for input-side cache double-counting (merged; does not address output duplication)
- langfuse/langfuse#12306 — Langfuse issue tracking the input-side problem

- Python: 3.12
- Pydantic AI: 1.85.1 (also verified on latest main)
- anthropic SDK, Langfuse (self-hosted, latest)

linyp:
- Closes #6131

## Summary

The Anthropic adapter's `_map_usage` copied the raw API token fields (`input_tokens`, `output_tokens`, `cache_creation_input_tokens`, `cache_read_input_tokens`) into `RequestUsage.details`. Since these duplicate the first-class `RequestUsage` fields, `opentelemetry_attributes()` emitted each value twice — once as `gen_ai.usage.output_tokens` and once as `gen_ai.usage.details.output_tokens` — so consumers like Langfuse summed them and doubled reported output tokens and cost. (`RunUsage.incr`, which sums `details`, was affected too.)

The OpenAI adapter already filters these keys; this brings Anthropic in line.

## Why it's not a one-line strip

`details` did double duty as the **streaming merge carrier**: Anthropic `message_delta` events may omit `input_tokens`, so the raw counts were carried forward from `message_start` via `existing_usage.details` (added in #3111). Simply stripping the stored `details` regresses that carry-forward (the existing `test_streaming_usage`, where the delta has no `input_tokens`, would drop to `input_tokens=0`). genai-prices also folds cache tokens into `input_tokens`, so the raw per-event values can't be reliably reconstructed from the first-class fields.

So the fix **decouples the two roles**: `_map_usage` now returns the `RequestUsage` (with the raw token keys excluded from `details`) alongside the raw counts, and `AnthropicStreamedResponse` threads those raw counts in a separate field. `self._usage.details` is always clean, while streaming carry-forward still works. Pricing is unaffected — it reads the `usage=` field, not `details`. Compaction/iteration keys remain in `details` as they're genuinely supplementary.

## Tests

- New `test_usage_otel_attributes_have_no_first_class_token_duplicates` asserts `opentelemetry_attributes()` (the method emitted on the model-request span) no longer contains `gen_ai.usage.details.{input,output}_tokens`.
- Updated the direct `_map_usage` unit tests; `test_streaming_usage` now explicitly pins carry-forward (delta without `input_tokens` → result keeps `input_tokens=1`, `details` empty).
- Public-API snapshots updated to drop the duplicated detail keys; first-class token counts are unchanged. The end-to-end `agent.run_stream` test `test_streaming_bedrock_start_event_without_message_is_skipped` covers the real streaming path.

Scoped to Anthropic only — OpenAI already filters these keys and Bedrock builds `RequestUsage` directly.

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6134?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/anthropic.py
+++ b/pydantic_ai_slim/pydantic_ai/models/anthropic.py
@@ -2195,10 +2195,16 @@ class AnthropicCompaction(AbstractCapability[AgentDepsT]):
 
 
 _COMPACTION_TOKEN_KEYS = ('input_tokens', 'output_tokens', 'cache_creation_input_tokens', 'cache_read_input_tokens')
-# The same raw Anthropic token fields all map onto first-class `RequestUsage` fields, so they're
-# excluded from the exposed `RequestUsage.details` to stop OTel consumers double-counting them against
-# the `gen_ai.usage.*` attributes (see `_map_usage` and <https://github.com/pydantic/pydantic-ai/issues/6131>).
-_FIRST_CLASS_TOKEN_KEYS = frozenset(_COMPACTION_TOKEN_KEYS)
+# Raw Anthropic token fields that map onto first-class `RequestUsage` fields, so they're excluded from the
+# exposed `RequestUsage.details` to stop OTel consumers double-counting them against the `gen_ai.usage.*`
+# attributes (see `_map_usage` and <https://github.com/pydantic/pydantic-ai/issues/6131>). Kept as its own
+# literal rather than `frozenset(_COMPACTION_TOKEN_KEYS)`: the two sets share these four keys only because the
+# compaction iteration totals happen to cover exactly the first-class token fields today, but they answer
+# independent questions ("which keys need compaction summing" vs "which keys duplicate a first-class field")
+# and must be free to diverge if Anthropic's usage schema grows.
+_FIRST_CLASS_TOKEN_KEYS = frozenset(
+    {'input_tokens', 'output_tokens', 'cache_creation_input_tokens', 'cache_read_input_tokens'}
+)
 
 
 def _extract_usage_details(response_usage: BetaUsage | BetaMessageDeltaUsage) -> dict[str, int]:
@@ -2278,8 +2284,11 @@ def _map_usage(
     # Note: genai-prices already extracts cache_creation_input_tokens and cache_read_input_tokens
     # from the Anthropic response and maps them to cache_write_tokens and cache_read_tokens.
     # The raw token counts are summed into the first-class `RequestUsage` fields above, so they're
-    # excluded from `details` (see docstring); `compaction_*`/`*_iterations` keys stay as they're
-    # supplementary information not represented by any first-class field.
+    # excluded from `details` (see docstring). The `compaction_iterations`/`message_iterations` counts
+    # stay as they're genuinely supplementary. The `compaction_*` token sub-totals also stay for
+    # transparency, but note they ARE folded into the first-class totals above, so a consumer summing
+    # them against `gen_ai.usage.*` still double-counts under compaction — a known, narrower residual
+    # scoped out of this fix.
     request_usage = usage.RequestUsage.extract(
         dict(model=model, usage=usage_for_extraction),
         provider=provider,
--- a/pydantic_ai_slim/pydantic_ai/models/anthropic.py
+++ b/pydantic_ai_slim/pydantic_ai/models/anthropic.py
@@ -2195,13 +2195,12 @@ class AnthropicCompaction(AbstractCapability[AgentDepsT]):
 
 
 _COMPACTION_TOKEN_KEYS = ('input_tokens', 'output_tokens', 'cache_creation_input_tokens', 'cache_read_input_tokens')
-# Raw Anthropic token fields that map onto first-class `RequestUsage` fields, so they're excluded from the
-# exposed `RequestUsage.details` to stop OTel consumers double-counting them against the `gen_ai.usage.*`
-# attributes (see `_map_usage` and <https://github.com/pydantic/pydantic-ai/issues/6131>). Kept as its own
-# literal rather than `frozenset(_COMPACTION_TOKEN_KEYS)`: the two sets share these four keys only because the
-# compaction iteration totals happen to cover exactly the first-class token fields today, but they answer
-# independent questions ("which keys need compaction summing" vs "which keys duplicate a first-class field")
-# and must be free to diverge if Anthropic's usage schema grows.
+# Raw Anthropic token fields that a consumer normalizes onto a first-class `RequestUsage` metric (input,
+# output, cache read/write). Excluded from the exposed `RequestUsage.details` so consumers don't add them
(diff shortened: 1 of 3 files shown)


adtyavrdhn:
Thanks for the PR @linyp 

A one line strip weirdly enough doesn't cause a regression in the test suite so either we have a false understanding or we don't have good tests for the exposed scenario. Looking into it.

adtyavrdhn:
Turns out it was breaking but the snapshot was being updated

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -2194,6 +2195,10 @@ def get_serialization_name(cls) -> str | None:
 
 
 _COMPACTION_TOKEN_KEYS = ('input_tokens', 'output_tokens', 'cache_creation_input_tokens', 'cache_read_input_tokens')
```
adtyavrdhn, line 2197: compaction_input_tokens is also included in input_tokens so it is not really complementary in that sense. We should drop that too after verification

adtyavrdhn (changes requested):
We should have a test assert the behaviour with a comment for streaming carry forward, it is difficult to read with just the snapshot and easy to miss it

pydantic_ai_slim/pydantic_ai/models/anthropic.py
adtyavrdhn: not likely to be overcounted because no name collision so let us keep it

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -2252,24 +2271,32 @@ def _map_usage(
...
-        if compaction_value := details.get(f'compaction_{key}'):
+        if compaction_value := raw_details.get(f'compaction_{key}'):
             usage_for_extraction[key] = usage_for_extraction.get(key, 0) + compaction_value
 
     # Note: genai-prices already extracts cache_creation_input_tokens and cache_read_input_tokens
-    # from the Anthropic response and maps them to cache_write_tokens and cache_read_tokens
-    return usage.RequestUsage.extract(
+    # from the Anthropic response and maps them to cache_write_tokens and cache_read_tokens.
+    # The raw token counts are summed into the first-class `RequestUsage` fields above and excluded from
+    # `details`: consumers (e.g. Langfuse) map a detail key like `output_tokens`, or the well-known Anthropic
```
adtyavrdhn, line 2286: So verbose, trim this

adtyavrdhn:
Going with another approach in #6244

PR #6135, opened 2026-06-29
Add turn index to agent chat spans

### LINKED ISSUE #5687: Add per-turn index to GenAI span names for agent/model requests
### Description

I’d like to propose adding a per-turn index to the GenAI span naming emitted by PydanticAI agent/model tracing, similar in spirit to the per-LLM-turn spans described in Promptfoo’s tracing documentation. Today the model spans are emitted with names like `chat {gen_ai.request.model}` and the corresponding semantic attributes include `gen_ai.operation.name="chat"` and `gen_ai.request.model="gpt-4.1-mini"` (or similar), which is useful, but it does not expose the turn number in a way that is easy to assert against during evaluation. 

My use case is evaluation with Promptfoo on top of OpenTelemetry traces outputed by PydanticAI agent run. I need to verify not just that the correct tools were called, but that the agent batched tool calls in the correct turn order. For example, in a workflow where the agent must paginate sequentially and then fan out tool calls in parallel, it is important to distinguish:  
1. a single LLM turn that emits multiple tool calls in parallel, versus  
2. multiple LLM turns that emit tools sequentially across generations. 

Right now, that distinction is hard to make reliably from the existing trace shape because the visible model span name is model-based and the agent span name is separate. In practice, the trace is great for observability, but it is not sufficient for clean automated assertions about turn structure when tool execution order matters. I ended up needing custom wrapper spans in my provider just to create a stable span name for counting turns, which feels like a workaround for something that could be built into the framework itself. 

I’d like to suggest that PydanticAI emit an incrementing turn index on every model request, for example by extending the current span naming or attributes to include something like:  
- span name: `gen_ai.operation.name + gen_ai.operation.turn + gen_ai.request.model`  

That would make trace-based evaluation much cleaner because it would let tools like Promptfoo distinguish turn boundaries directly from the trace, instead of inferring them indirectly from parent-child spans or from model-specific span titles. It would also make it easier to assert sequential versus parallel tool calling behavior, which is a common requirement for latency-sensitive agent workflows. 

### References

- Promptfoo per-LLM-turn spans documentation: <https://github.com/promptfoo/promptfoo/blob/feat/claude-agent-sdk-turn-spans/site/docs/tracing.md#per-llm-turn-spans>
- PydanticAI issue about span titles: <https://github.com/pydantic/pydantic-ai/issues/2925>
- PydanticAI issue about updating OpenTelemetry span and attribute names: <https://github.com/pydantic/pydantic-ai/issues/2964>
- OpenTelemetry GenAI span conventions: <https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/> 
- PydanticAI instrumented model docs: <https://pydantic.dev/docs/ai/api/models/instrumented/>
- PydanticAI span-based evaluation docs: <https://pydantic.dev/docs/ai/evals/evaluators/span-based/>

HarperZ9:
## Summary

- add a stable `gen_ai.turn.index` attribute to agent chat/model request spans
- keep span names unchanged, so programmatic consumers can use attributes instead of parsing names
- add a regression test covering a tool-call run with two `chat test` spans indexed as `0, 1`
- update existing trace snapshots that assert chat span attributes

Closes #5687.

## Verification

- `uv run pytest -q tests/test_logfire.py::test_instrumentation_model_request_spans_include_turn_index tests/test_logfire.py::test_logfire tests/test_logfire.py::test_instrumentation_capability_explicit tests/test_logfire.py::test_instrument_all_skipped_when_capability_already_present`
- `uv run pytest -q tests/models/test_fallback.py::test_first_failed_instrumented`
- `uv run ruff check pydantic_ai_slim/pydantic_ai/_instrumentation.py pydantic_ai_slim/pydantic_ai/capabilities/instrumentation.py tests/test_logfire.py tests/models/test_fallback.py`
- `uv run ruff format --check pydantic_ai_slim/pydantic_ai/_instrumentation.py pydantic_ai_slim/pydantic_ai/capabilities/instrumentation.py tests/test_logfire.py tests/models/test_fallback.py`
- `git diff --check`

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6135?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


adtyavrdhn:
Converting to draft, please align the approach with Douwe's comment on the issue

HarperZ9:
Thanks, Douwe. I agree that a custom capability in `pydantic-ai-harness` is the right place to prove this before asking core instrumentation to carry a non-standard attribute.

I will prototype the counter in a capability hook that updates the current OpenTelemetry span, keep span names unchanged, and retain the evaluation cases that motivated the change: sequential turns, parallel tool calls within one turn, and retry behavior without conflating attempts with new turns. I will document the attribute as harness-scoped/non-standard rather than presenting it as GenAI semantic convention.

I will leave this core draft unchanged while that path is validated, then link and supersede/close #6135 once there is a concrete harness result to review.

HarperZ9:
Closing this draft.

Per DouweM's guidance on #5687, `gen_ai.turn.index` is a Promptfoo-specific field rather than a settled convention, so hardcoding it into the core `Instrumentation` capability is the wrong place for it. The agreed direction is a separate custom capability in pydantic-ai-harness that tracks the turn counter and sets the attribute on the current span from a hook. This PR takes the core approach that was set aside, so it should not merge.

Apologies for leaving this as a dangling draft since my 19 July note; I said I would prototype the harness capability and then close this, and I did not follow through in time.

#5687 stays open to track the attribute-first, capability-based approach. Thanks @DouweM and @adtyavrdhn for the review and the redirection, and @sipa-echo-ngbm for the well-scoped use case.

PR #6139, opened 2026-06-29
fix: skip thinking-only OpenAI chat history messages

### LINKED ISSUE #6138: [provider-mapping-sweep] openai: chat completions `_into_message_param` emits assistant message with `content=null` (no `tool_calls`) when a `ModelRespon
[Content truncated due to length]
## Impact

A `ModelResponse` that contains only `ThinkingPart`s — no `TextPart`, no `ToolCallPart` — is mapped to an OpenAI Chat Completions assistant message with `content: null` and no `tool_calls`, which the API rejects with HTTP 400 ("`content` must be of type string or array, not null"; per the spec `content` is *Required unless `tool_calls` or `function_call` is populated*).

This breaks multi-turn runs against any model that uses a custom reasoning field (DeepSeek `reasoning_content`, Moonshot `reasoning_content`, gpt-oss `reasoning` via OpenRouter/Ollama, etc.) whenever the model emits a reasoning-only turn. The existing `test_empty_response_skipped_in_history` guard covers the empty-`parts` case but NOT the thinking-only case, because the early-return condition only fires when **all three** of `texts`, `thinkings`, `tool_calls` are empty.

Reproduction failure message:

```
Assistant message with only ThinkingParts must have non-null content OR tool_calls, otherwise OpenAI Chat Completions rejects with 400.
Got: {'role': 'assistant', 'reasoning_content': 'Let me think about this...', 'content': None}
```

## Provider & Code Path

**Provider:** `openai` (Chat Completions adapter, `OpenAIChatModel`)

**File:** `pydantic_ai_slim/pydantic_ai/models/openai.py`

**Faulty method:** `OpenAIChatModel._MapModelResponseContext._into_message_param` (lines 1262–1289).

```python
if not self.texts and not self.thinkings and not self.tool_calls:
    return None                                              # line 1274 — guards empty parts
message_param = chat.ChatCompletionAssistantMessageParam(role='assistant')
if self.thinkings:
    for field_name, contents in self.thinkings.items():
        message_param[field_name] = '\n\n'.join(contents)   # sets e.g. reasoning_content
if self.texts:
    message_param['content'] = '\n\n'.join(self.texts)
else:
    message_param['content'] = None                         # line 1285 — null content w/ no tool_calls
if self.tool_calls:
    message_param['tool_calls'] = self.tool_calls
return message_param
```

The early-return condition is too narrow: it requires **all three** collections to be empty, so a thinking-only turn (`thinkings` populated, `texts`/`tool_calls` empty) bypasses it and falls through to the `content: None` branch.

This is hit when `openai_chat_send_back_thinking_parts` is set to `'field'` OR is `'auto'` with the ThinkingPart's `id` matching the profile's `openai_chat_thinking_field` (and `provider_name == self.system`). Both paths are exercised by `_map_response_thinking_part` (lines 1294–1325) — `'field'` always populates `self.thinkings`, and `'auto'` does so when the field/id/provider-name triad matches.

The docstring on `_into_message_param` (lines 1263–1272) explicitly states the goal — *"Returning `None` ensures we don't emit an assistant message with `content=None` and no `tool_calls`, which the Chat Completions API rejects with a 400 error."* — so the intent is clear; the guard is just incomplete.

## Reproduction

```python
# repro_thinking_only.py

import asyncio

    ModelRequest, ModelResponse, ThinkingPart, UserPromptPart,
)


def build_model(*, send_mode: str) -> OpenAIChatModel:
    profile = OpenAIModelProfile(
        openai_chat_send_back_thinking_parts=send_mode,
    )
    return OpenAIChatModel(
        'deepseek-reasoner',
        provider=OpenAIProvider(api_key='sk-test'),
        profile=profile,
    )


def make_thinking_only_response() -> ModelResponse:
    return ModelResponse(
        parts=[
            ThinkingPart(
                content='Let me think about this...',
                id='reasoning_content',          # matches the configured custom field
                provider_name='openai',
            ),
        ],
        model_name='deepseek-reasoner',
        provider_name='openai',
    )


async def main():
    history = [
        make_thinking_only_response(),
        ModelRequest(parts=[UserPromptPart(content='Continue')]),
    ]

    for mode in ('field', 'auto'):
        model = build_model(send_mode=mode)
        mapped = await model._map_messages(history, model_request_parameters=ModelRequestParameters())
        assistant = next(m for m in mapped if m.get('role') == 'assistant')
        assert assistant.get('content') or assistant.get('tool_calls'), (
            f'BUG ({mode} mode): assistant message has content=None and no tool_calls; '
            f'OpenAI Chat Completions rejects with 400. Got: {assistant!r}'
        )


asyncio.run(main())
```

Run with `python repro_thinking_only.py`.

## Expected vs Actual

**Expected:** when a `ModelResponse` has only `ThinkingPart`s, `_map_messages` either (a) drops the turn entirely (mirroring the empty-`parts` case via `_into_message_param` returning `None`) or (b) emits a syntactically valid assistant message — i.e., one with non-null `content` OR a `tool_calls` field. Per the OpenAI Chat Completions spec, the assistant message MUST have `content` populated unless `tool_calls`/`function_call` is populated.

**Actual:** both the `'field'` mode and the matching-`'auto'` mode emit `{'role': 'assistant', 'reasoning_content': '...', 'content': None}`. The OpenAI API server rejects this with `400 Bad Request` ("`content` must be of type string or array, not null").

## Evidence

- OpenAI Chat Completions API spec, `messages[].content` for `role: 'assistant'`: *"The contents of the assistant message. Required unless `tool_calls` or `function_call` is populated."* (https://platform.openai.com/docs/api-reference/chat/create — `ChatCompletionAssistantMessageParam`).
- Existing test `test_empty_response_skipped_in_history` (`tests/models/test_openai.py:4118`) documents the same bug class for the empty-`parts` case; the fix for that case is the early-return at `openai.py:1274`, which is exactly the line whose predicate needs to be tightened.
- Recent PR #5842 added the `backfill_field` empty-string injection for tool-call turns missing the reasoning field (`openai.py:1416–1423`), which protects the *tool-call* leg but does nothing for the *thinking-only* leg because there is no tool_call to attach the field to.
- `OpenAIResponsesModel._map_messages` (around `openai.py:2688–2707`) does NOT have this defect: a thinking-only turn is serialized as a `ResponseReasoningItemParam` without an empty `EasyInputMessageParam`, which the Responses API accepts.

## Suggested fix (one-line predicate tightening)

```python
# openai.py:1274
if not self.texts and not self.tool_calls:
    return None
```

This restores the docstring's stated invariant ("Returning `None` ensures we don't emit an assistant message with `content=None` and no `tool_calls`") for the thinking-only case, mirroring the existing empty-`parts` guard. The current `not self.thinkings` clause in the predicate is what lets the bug through — it's the only collection that can be non-empty while still leaving `content: None` and `tool_calls` absent.

Regression test: add a `test_thinking_only_assistant_message_dropped` parallel to `test_empty_response_skipped_in_history`, asserting that the thinking-only assistant turn is omitted from the wire `messages[]` array sent to the chat completions endpoint (not just from the in-memory `mapped` list — the wire assertion catches a future regression where a different code path emits it).

pradeep-ramola:
- Closes #6138

### Summary

This updates the OpenAI Chat Completions message mapping to skip model responses that only contain mapped thinking parts and no text or tool calls. Custom reasoning fields such as `reasoning_content` do not satisfy Chat Completions' requirement for assistant messages to include `content` unless `tool_calls` is populated, so sending those responses back in history can produce `content=null` request errors.

### Tests

- `py -3.11 -m uv run --python 3.11 pytest tests/models/test_openai.py::test_empty_response_skipped_in_history tests/models/test_openai.py::test_thinking_only_response_skipped_in_history`
- `py -3.11 -m uv run --python 3.11 ruff format --check pydantic_ai_slim/pydantic_ai/models/openai.py tests/models/test_openai.py`
- `py -3.11 -m uv run --python 3.11 ruff check pydantic_ai_slim/pydantic_ai/models/openai.py tests/models/test_openai.py`

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6139?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


hramezani (approved):

adtyavrdhn:
Thanks @pradeep-ramola 

Needed to trim the tests but PR didn't have maintainer edit access so pulled #6182.

PR #6144, opened 2026-06-30
Don't send `defer_loading` without `tool_search` on models without native tool search

### LINKED ISSUE #5938: OpenAI Responses adapter sends `defer_loading` without `tool_search` for deferred capabilities

adtyavrdhn:
- Closes #5938

## What was going wrong

Loading a deferred capability (or discovering a `defer_loading=True` tool via tool search) on an OpenAI model that **doesn't support native tool search** (e.g. `gpt-5.2`) failed with:

```
Invalid Value: 'tools.defer_loading'. Deferred tools require tools.tool_search.
```

## Why

OpenAI's wire-side `defer_loading` flag only works **paired with** a native `tool_search` tool, and native tool search only exists on `gpt-5.4+`.

On older models we already do the right thing and fall back to Pydantic AI's **local** tool search (a plain `search_tools` function), so no `tool_search` tool goes on the wire. The bug: once a deferred tool was *discovered*, it kept an internal `with_native='tool_search'` marker, and the OpenAI adapter stamped the wire-side `defer_loading: true` flag purely off that marker, without checking whether a `tool_search` tool was actually being sent. So the request went out with `defer_loading` and nothing to pair it with, and the provider rejected it.

In short:

| | `gpt-5.4+` (native) | `gpt-5.2` (local fallback) |
|---|---|---|
| `tool_search` tool on wire | ✅ yes | ❌ no |
| `defer_loading` stamped | ✅ correct | ❌ **stamped anyway, so 400** |

## The fix

In the provider-agnostic native-tool swap, when the model doesn't support native tool search, clear the stale `with_native` marker on tools that have already been discovered (they stay callable as plain function tools). With the marker gone, the adapter no longer stamps `defer_loading`, and the request is accepted.

Models that **do** support native tool search (`gpt-5.4+`) are completely unaffected: the marker is kept and `defer_loading` + `tool_search` still go out together. The change only fires on the path that had no native tool search (and therefore no native cache) to begin with.

## Tests

- **Unit test** pinning the wire payload: a discovered tool-search tool on `gpt-5.2` ships with no `defer_loading` and no `tool_search` (the cassette matcher isn't body-sensitive, so this is the real guard).
- **VCR test** against live `gpt-5.2` reproducing the issue exactly (deferred `Capability` + `load_capability` + tool call). All three recorded requests return `200`, where the second one (carrying the revealed tool) used to `400`.

### Checklist

### DIFF
--- a/tests/test_tool_search.py
+++ b/tests/test_tool_search.py
@@ -2230,73 +2230,22 @@ async def test_openai_deferred_capability_tool_reveal_uses_client_tool_search(al
     assert replay_outputs and all(item.get('execution') == 'client' for item in replay_outputs)
 
 
-async def test_openai_discovered_tool_without_native_tool_search_omits_defer_loading(
-    allow_model_requests: None,
-):
-    """A tool-search tool discovered in a prior turn must not carry the wire-side `defer_loading`
-    flag on a model without native `tool_search` (e.g. `gpt-5.2`).
-
-    `defer_loading` only travels alongside a native `tool_search` tool; without one, pydantic-ai
-    falls back to local search and OpenAI rejects a lone `defer_loading`. Minimal form of #5938;
-    the deferred-capability path is covered by
-    `test_openai_deferred_capability_runs_on_model_without_native_tool_search`.
-    """
-    pytest.importorskip('openai')
-
-    final = response_message(
-        [
-            ResponseOutputMessage(
-                id='msg',
-                content=[ResponseOutputText(text='Sunny.', type='output_text', annotations=[])],
-                role='assistant',
-                status='completed',
-                type='message',
-            )
-        ]
-    )
-    mock_client = MockOpenAIResponses.create_mock(final)
-    model = OpenAIResponsesModel('gpt-5.2', provider=OpenAIProvider(openai_client=mock_client))
-    agent: Agent[None, str] = Agent(model=model, capabilities=[ToolSearch()])
-
-    @agent.tool_plain(defer_loading=True)
-    def get_weather(city: str) -> str:  # pragma: no cover
-        return f'Weather in {city}.'
-
-    # `get_weather` was discovered last turn, so it now rides along as a callable tool
-    # (`defer_loading=False`, but `with_native='tool_search'` still set).
-    history: list[ModelMessage] = [
-        ModelRequest(parts=[UserPromptPart(content='I might want the weather later.')]),
-        ModelResponse(parts=[ToolSearchCallPart(args={'queries': ['weather']}, tool_call_id='loc_1')]),
-        ModelRequest(
-            parts=[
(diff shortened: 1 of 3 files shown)


tests/test_tool_search.py
```diff
@@ -2221,6 +2230,106 @@ def lookup_refund_policy(order_id: str) -> str:  # pragma: no cover
...
+    # `bar` (the one that used to 400) was sent and accepted.
+    assert any(isinstance(p, LoadCapabilityReturnPart) for m in result.all_messages() for p in m.parts)
+    bar_returns = [
+        p for m in result.all_messages() for p in m.parts if isinstance(p, ToolReturnPart) and p.tool_name == 'bar'
+    ]
+    assert [p.content for p in bar_returns] == [1]
+
+    # gpt-5.2 runs the local fallback: no request may carry `defer_loading` or a `tool_search` tool.
+    for request in vcr.requests:
+        request_tools = cast(list[dict[str, Any]], json.loads(request.body).get('tools', []))
```
adtyavrdhn: No need to test the payload, let us just assert the `all_messages` post run which will show if it worked

tests/test_tool_search.py
```diff
@@ -2221,6 +2230,106 @@ def lookup_refund_policy(order_id: str) -> str:  # pragma: no cover
     assert replay_outputs and all(item.get('execution') == 'client' for item in replay_outputs)
 
 
+async def test_openai_discovered_tool_without_native_tool_search_omits_defer_loading(
```
adtyavrdhn: Remove this one

adtyavrdhn:
All anthropic models(non-deprecated) support tool search so there isn't an equivalent anthropic test for this scenario we could have.

PR #6148, opened 2026-06-30
Suppress `ClosedResourceError` when cancelling a graph run mid-send (v1)

conradlee:
**v1 backport of #6149** (do not auto-close — this targets the maintained 1.x line, a different base branch than #6149).

#3675 added `except BrokenResourceError` at the two `iter_stream_sender.send(...)` sites in `_GraphIterator._run_tracked_task`, but run cleanup closes the memory-object-stream **sender**, so an in-flight send raises the sibling `anyio.ClosedResourceError` (not `BrokenResourceError`), which the narrow `except` misses — leaving runs cancelled before completion to crash out of `GraphRun.__aexit__` instead of cancelling cleanly. This broadens both excepts to `(BrokenResourceError, ClosedResourceError)`.

The `_run_tracked_task` teardown is byte-identical on `v1` and `main`, so this is the same one-line change as #6149, targeted at the 1.x line (where many users — including us — are still pinned). Validated with the repro from the linked issue: ~40 `ClosedResourceError` per 2000 cancelled runs on `v1` (1.107) → **0** with this change.

🤖 Prepared with Claude Code and validated against the repro; opened by @conradlee.

### Checklist

### DIFF
--- a/pydantic_graph/pydantic_graph/graph_builder.py
+++ b/pydantic_graph/pydantic_graph/graph_builder.py
@@ -38,7 +38,7 @@ from typing import (
     overload,
 )
 
-from anyio import BrokenResourceError, CancelScope, create_memory_object_stream, create_task_group
+from anyio import BrokenResourceError, CancelScope, ClosedResourceError, create_memory_object_stream, create_task_group
 from anyio.abc import TaskGroup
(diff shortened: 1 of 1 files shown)


dsfaccini:
closing this for the moment as the user has been able to patch the bug it their setup and we don't envision backporting minor bugfixes

Issue #3655, opened 2025-12-06
anyio.BrokenResourceError surfacing on agent/task cancellation


vikigenius:
### Initial Checks

- [x] I confirm that I'm using the latest version of Pydantic AI
- [x] I confirm that I searched for my issue in https://github.com/pydantic/pydantic-ai/issues before opening this issue

### Description

When cancelling a running Agent.run(...) task, I sometimes see an anyio.BrokenResourceError bubble out of pydantic-graph instead of a CancelledError.

My typical usage looks roughly like this (simplified):

```python
async with asyncio.TaskGroup() as tg:
    web_context_task = tg.create_task(
        web_context_agent.run(web_context_prompt, deps=ctx.deps)
    )

    # ... do some other work, possibly early returns ...

    # At some point I decide I don't need this result anymore:
    _ = web_context_task.cancel()

    # I may continue doing other async work here to avoid burning tokens
    # ...
```

In rare cases, cancelling the task results in this traceback instead of a clean cancellation:

```
Traceback (most recent call last):
  File "/var/lang/lib/python3.12/site-packages/opentelemetry/trace/__init__.py", line 589, in use_span
    yield span
  File "/var/lang/lib/python3.12/site-packages/pydantic_graph/beta/graph.py", line 271, in iter
    async with GraphRun[StateT, DepsT, OutputT](
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/lang/lib/python3.12/site-packages/pydantic_graph/beta/graph.py", line 400, in __aexit__
    await self._async_exit_stack.__aexit__(exc_type, exc_val, exc_tb)
  File "/var/lang/lib/python3.12/contextlib.py", line 754, in __aexit__
    raise exc_details[1]
  File "/var/lang/lib/python3.12/contextlib.py", line 735, in __aexit__
    cb_suppress = cb(*exc_details)
                  ^^^^^^^^^^^^^^^^
  File "/var/lang/lib/python3.12/contextlib.py", line 158, in __exit__
    self.gen.throw(value)
  File "/var/lang/lib/python3.12/site-packages/pydantic_graph/beta/graph.py", line 939, in _unwrap_exception_groups
    raise exception
  File "/var/lang/lib/python3.12/asyncio/tasks.py", line 314, in __step_run_and_handle_result
    result = coro.send(None)
             ^^^^^^^^^^^^^^^
[...]

PR #6149, opened 2026-06-30
Suppress `ClosedResourceError` when cancelling a graph run mid-send

### LINKED ISSUE #6146: Cancelling a run mid-send raises uncaught `anyio.ClosedResourceError` (sibling missed by #3675)
### Initial Checks

### Description

This is the remaining half of #3655, which #3675 ("Suppress broken resource errors if cancelling") only partially fixed.

When an `Agent.run` / graph run is **cancelled before it reaches its End node** (an `asyncio.wait_for` / `anyio.fail_after` deadline, a websocket disconnect, or a parent task cancelling the run), the run intermittently crashes with `anyio.ClosedResourceError` bubbling out of `GraphRun.__aexit__` instead of cancelling cleanly.

**Root cause.** #3675 added `except BrokenResourceError` at the two `iter_stream_sender.send(...)` sites in `_GraphIterator._run_tracked_task` (`pydantic_graph/graph_builder.py`). But run cleanup closes the memory-object-stream **sender**, and an in-flight `send()` to a *closed sender* raises `anyio.ClosedResourceError`, **not** `BrokenResourceError` (a closed *receiver* is what raises `BrokenResourceError`). So the `except` misses it and the error escapes:

```
  File ".../pydantic_graph/graph_builder.py", line 874, in _run_tracked_task
    await self.iter_stream_sender.send(_GraphTaskResult(t_, result))
  File ".../anyio/streams/memory.py", line 249, in send
    self.send_nowait(item)
  File ".../anyio/streams/memory.py", line 218, in send_nowait
    raise ClosedResourceError
anyio.ClosedResourceError
```

The fix is a one-line broadening at both sites: `except (BrokenResourceError, ClosedResourceError)`. The author of #3675 noted *"I am not sure if this fully addresses that issue"* — this is the missing piece.

> Note: the same teardown-under-cancellation also occasionally surfaces the second-form anyio `RuntimeError: Attempted to exit a cancel scope that isn't the current task's current cancel scope`, when a per-node `CancelScope` exits out of LIFO order. That one is a harder structural case; the `ClosedResourceError` is the dominant, cleanly-fixable variant and is what the example below triggers.

**Both maintained branches are affected** — the `_run_tracked_task` teardown is identical on `main` (2.x) and `v1` (1.x). I have a one-line fix validated for both and will open PRs against each.

```python
import asyncio
import random
import traceback
from collections import Counter

from pydantic_ai import Agent
from pydantic_ai.messages import ModelResponse, TextPart, ToolCallPart
from pydantic_ai.models.function import FunctionModel


# A model that asks for several parallel tool calls per step, for a few steps,
# then finishes — enough concurrent graph node tasks to widen the teardown race.
def fn(messages, info):
    n = sum(1 for m in messages if isinstance(m, ModelResponse))
    if n >= 4:
    return ModelResponse(
        parts=[ToolCallPart(tool_name="work", args={"i": i}, tool_call_id=f"c{n}_{i}") for i in range(16)]
    )


agent = Agent(FunctionModel(fn))


@agent.tool_plain
async def work(i: int) -> str:
    await asyncio.sleep(random.uniform(0, 0.004))
    return f"r{i}"


res: Counter = Counter()
sample = None


async def one(cancel: bool):
    global sample
    try:
        if cancel:
            # Cancel mid-run with a tight, randomized deadline.
            await asyncio.wait_for(agent.run("go"), timeout=random.uniform(0.001, 0.02))
        else:
            await agent.run("go")
        res["ok"] += 1
    except (asyncio.TimeoutError, asyncio.CancelledError):
        res["clean_cancel"] += 1
    except BaseException as e:  # anything else here is the leak
        res[f"LEAK:{type(e).__name__}"] += 1
        if sample is None:
            sample = traceback.format_exc()


async def main():
    random.seed(7)
    for _ in range(50):
        await asyncio.gather(*[asyncio.create_task(one(random.random() < 0.5)) for _ in range(40)])
    for k, v in sorted(res.items()):
        print(f"{v:5d} {k}")
    if sample:
        print("\n--- sample leak traceback (tail) ---\n" + sample[-900:])


asyncio.run(main())
```

On current `main` (`2.1.2.dev`) and `v1` (`1.107`) this prints ~40 `LEAK:ClosedResourceError` per 2000 runs. With `except (BrokenResourceError, ClosedResourceError)` at the two send sites, the leaks go to **0** (all cancellations surface cleanly as `TimeoutError`).

- Python: 3.12.11
- Pydantic AI: 2.1.2.dev (`main`) and 1.107 (`v1`) — both affected
- LLM provider SDK: n/a (reproduced with `FunctionModel`, no provider involved)

conradlee:
- Closes #6146

Follow-up to #3655 / #3675. #3675 added `except BrokenResourceError` at the two `iter_stream_sender.send(...)` sites in `_GraphIterator._run_tracked_task` to swallow the benign error from sending into a stream closed during run teardown. But run cleanup closes the memory-object-stream **sender**, so an in-flight send raises the sibling `anyio.ClosedResourceError`, which the narrow `except` misses — leaving runs cancelled before completion to crash out of `GraphRun.__aexit__` instead of cancelling cleanly. This broadens both excepts to `(BrokenResourceError, ClosedResourceError)`.

**Validation:** the minimal repro in #6146 leaks ~40 `ClosedResourceError` per 2000 cancelled runs on `main`; with this change, **0** (every cancellation surfaces cleanly as `TimeoutError`/`CancelledError`). No deterministic unit test is added: this is the same hard-to-trigger teardown race the sibling `BrokenResourceError` branch already carries (and which #3675 marked `# pragma: no cover`), so the repro in the linked issue is the validation.

A matching backport against `v1` is in #6148.

🤖 Prepared with Claude Code and validated against the repro; opened by @conradlee.

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6149?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_graph/pydantic_graph/graph_builder.py
+++ b/pydantic_graph/pydantic_graph/graph_builder.py
@@ -874,10 +874,10 @@ class _GraphIterator(Generic[StateT, DepsT, OutputT]):
                     await self.iter_stream_sender.send(_GraphTaskResult(t_, result))
             except (BrokenResourceError, ClosedResourceError):
                 # Can happen when an asyncio task is cancelled mid-send: the run's
-                # cleanup closes ``iter_stream_sender`` in another task while this
-                # task is still in-flight on ``send``. Closing the sender raises
-                # ``ClosedResourceError`` (closing the receiver would raise
-                # ``BrokenResourceError``); both are benign here — the result/error
+                # cleanup closes `iter_stream_sender` in another task while this
+                # task is still in-flight on `send`. Closing the sender raises
+                # `ClosedResourceError` (closing the receiver would raise
+                # `BrokenResourceError`); both are benign here — the result/error
                 # is no longer needed because the run is being torn down.
                 pass
 
--- a/pydantic_graph/pydantic_graph/graph_builder.py
+++ b/pydantic_graph/pydantic_graph/graph_builder.py
@@ -30,7 +30,7 @@ from typing import (
     overload,
 )
 
-from anyio import BrokenResourceError, CancelScope, create_memory_object_stream, create_task_group
+from anyio import BrokenResourceError, CancelScope, ClosedResourceError, create_memory_object_stream, create_task_group
 from anyio.abc import TaskGroup
 from anyio.streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStream
 from typing_extensions import Never, TypeAliasType, TypeVar, assert_never
@@ -862,7 +862,7 @@ class _GraphIterator(Generic[StateT, DepsT, OutputT]):
                 # or ExceptionGroup). This preserves the original exception for the caller.
                 try:
                     await self.iter_stream_sender.send(_GraphTaskResult(t_, [], error=exc))
-                except BrokenResourceError:
+                except (BrokenResourceError, ClosedResourceError):
                     pass  # pragma: no cover
                 return
             try:
@@ -872,8 +872,13 @@ class _GraphIterator(Generic[StateT, DepsT, OutputT]):
                     await self.iter_stream_sender.send(_GraphTaskResult(t_, []))
                 else:
                     await self.iter_stream_sender.send(_GraphTaskResult(t_, result))
-            except BrokenResourceError:
-                # Can happen when an asyncio task is cancelled mid-send.
+            except (BrokenResourceError, ClosedResourceError):
+                # Can happen when an asyncio task is cancelled mid-send: the run's
+                # cleanup closes ``iter_stream_sender`` in another task while this
+                # task is still in-flight on ``send``. Closing the sender raises
+                # ``ClosedResourceError`` (closing the receiver would raise
+                # ``BrokenResourceError``); both are benign here — the result/error
+                # is no longer needed because the run is being torn down.
                 pass
 
     async def _run_task(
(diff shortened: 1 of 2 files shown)


pydantic_graph/pydantic_graph/graph_builder.py
```diff
@@ -872,8 +872,13 @@ async def _run_tracked_task(self, t_: GraphTask):
                     await self.iter_stream_sender.send(_GraphTaskResult(t_, []))
                 else:
                     await self.iter_stream_sender.send(_GraphTaskResult(t_, result))
-            except BrokenResourceError:
-                # Can happen when an asyncio task is cancelled mid-send.
+            except (BrokenResourceError, ClosedResourceError):
+                # Can happen when an asyncio task is cancelled mid-send: the run's
+                # cleanup closes ``iter_stream_sender`` in another task while this
```
dsfaccini: we don't use double backticks in this codebase

conradlee: Good catch — switched these to single backticks to match the codebase convention (and it was tripping the `no-rst-syntax` pre-commit hook, which was the `lint` CI failure). Fixed in 63f18759.

PR #6152, opened 2026-06-30
docs: document `openai_chat_supports_multiple_system_messages` for backends that reject multiple system messages

### LINKED ISSUE #5812: vLLM backend rejects PromptedOutput requests with 'System message must be at the beginning'

adtyavrdhn:
Documents `openai_chat_supports_multiple_system_messages` for models whose chat template accepts only one leading system message (e.g. newer Qwen, Mistral, Gemma, Command-R served via vLLM or LiteLLM), so users who hit `System message must be at the beginning.` can find the fix.

Adds a "Models that accept only one leading system message" subsection under Model Profile and links to it from the LiteLLM section.

Fixes #5812

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6152?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/models/openai.md
+++ b/docs/models/openai.md
@@ -367,12 +367,28 @@ model = OpenAIChatModel(
     profile=OpenAIModelProfile(
         json_schema_transformer=InlineDefsJsonSchemaTransformer,  # Supported by any model class via the base ModelProfile
         openai_supports_strict_tool_definition=False,  # Supported by OpenAIChatModel and OpenAIResponsesModel
-        openai_chat_supports_multiple_system_messages=False,  # Supported by OpenAIChatModel only — for strict providers (e.g. some vLLM/LiteLLM setups) that require exactly one initial system message
+        openai_chat_supports_multiple_system_messages=False,  # Supported by OpenAIChatModel only — for models whose chat template accepts only one leading system message (see note below)
         openai_chat_supports_max_completion_tokens=False,  # Supported by OpenAIChatModel only — for providers (e.g. OpenRouter) that only accept the older `max_tokens` field instead of `max_completion_tokens`
     )
 )
 agent = Agent(model)
 ```
+
+!!! note "Backends that accept only one leading system message"
+    Some models are served with a chat template (applied server-side, for example by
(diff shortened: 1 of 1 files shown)


hramezani (approved):
Comment from Claude:
`
One tiny non-blocking note: the merge only kicks in when system_prompt_role is 'system' or 'developer' — with system_prompt_role='user', _merge_leading_system_messages returns early, so the flag is effectively a no-op there. Not worth complicating the prose over (the common case is covered), just flagging in case you want a half-sentence caveat.
`

docs/models/openai.md
```diff
@@ -374,6 +374,12 @@ model = OpenAIChatModel(
 agent = Agent(model)
 ```
 
+#### Models that accept only one leading system message
+
+Some models are served with a chat template (applied server-side, for example by [vLLM](https://docs.vllm.ai/), [LiteLLM](#litellm), or TGI) that accepts only a single system message at the start of the conversation and rejects additional ones. Sending more than one fails with a `400` error such as `System message must be at the beginning.` or `Conversation roles must alternate ...`, seen with some newer Qwen, Mistral, Gemma, and Command-R models. It's easy to hit without intending to: combining agent `instructions` with [`PromptedOutput`][pydantic_ai.output.PromptedOutput] can produce two consecutive leading system messages.
```
adtyavrdhn: get rid of `PromptedOutput` example, no need to mention

PR #6170, opened 2026-07-01
Strip nested tool-return files by default (outbound hardening)

dsfaccini:
Stacked on #5255 — one of three PRs splitting that work. **The base of this PR is the `ui-thread-multimodal-toolreturn` branch (PR #5255), not `main`**, so the diff shows only this PR's changes. It will be **retargeted to `main` once #5255 merges**.

### What this adds

Outbound secure-default file-strip hardening, split out of #5255. On the `preserve_file_data=False` default, file data can currently leak to the client through file content nested inside tool-return dicts/lists. This adds `strip_tool_return_files`, which recursively strips nested file data on that default, closing the server→client exposure gap.

The leak is pre-existing on `main` and **not urgent** — this is framed as GHSA hardening, not an emergency fix.

Included:

- `strip_tool_return_files` (recursive nested-file strip) wired into the AG-UI and Vercel AI outbound paths on the `preserve_file_data=False` default
- AG-UI orphan-sidecar guard (`raise ValueError` when a sidecar has no owning tool message)
- `media_type`-gated Vercel `_coerce_js_binary_data` fix (only coerces recognized binary shapes, preserves non-multimodal `kind`/`type` dicts)
- Tests for all of the above
- Docs update (`docs/ui/vercel-ai.md`)

### The #5255 split

- **A** — #5255 itself, reduced to the core multimodal tool-return round-trip
- **B** — this PR: outbound `strip_tool_return_files` secure-default hardening
- **C** — inbound sanitizer (`sanitize_message_history`), independent of A and B

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6170?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


dsfaccini:
waiting for #5255 to merge to resolve conflicts here and do RB

dsfaccini:
David's AICA here: thanks for the review — addressed in `8ca095b`.

**1. MEDIUM — streaming path leaked nested files (confirmed, fixed in-PR).** Reproduced it: `model_response_str()` / `model_response_object()` only drop *top-level* files, so a `BinaryContent` nested inside a mapping or a deeper list was serialized inline (base64) into the emitted event content. This is the same defect this PR fixes on the non-streaming path, surviving in the sibling streaming path — so per the parallel-path principle it belongs in this PR rather than as a follow-up. Applied `strip_tool_return_files(..., keep_top_level_files=True)` symmetrically before serialization in `ag_ui/_event_stream.py:_tool_return_content` and `vercel_ai/_event_stream.py:_tool_return_with_files`, and corrected the now-inaccurate `tool_return_output` docstring. Regression tests added on both streaming paths (`test_event_stream_tool_return_nested_multimodal_stripped`, `test_run_stream_tool_return_nested_multimodal_stripped`) asserting no file bytes reach the wire.

**2. LOW — outbound strip doesn't mirror the inbound `tool_kind` skip (intentional, dismissed).** The two paths are asymmetric by design, and mirroring the skip here would be a regression, not a fix. The inbound sanitizer skips typed subclasses so their `TypedDict` `content` stays schema-valid for *re-validation*. Outbound, the stripped content is only serialized to a wire string — there's no schema to keep valid — and skipping typed parts would instead *leak* a file-bearing typed field (the very thing this PR closes). Confirmed inert today: none of the registered typed subclasses (`LoadCapabilityReturn`, `ToolSearchReturnContent`) carries a file-bearing field. Documented the asymmetry in the `strip_tool_return_files` docstring so it isn't re-litigated.

**3. NITPICK — `AGUIAdapter.load_messages` docstring (fixed).** Added a `Raises:` clause documenting the orphan-sidecar `ValueError`.

dsfaccini:

|---|---|---|---|---|
| 1 · code-correctness | ✅ closed | 0 | 0 | 3 |
| 3 · cross-cutting/siblings | ✅ closed | 0 | 0 | 1 |

**Verdict:** ✅ all 3 waves passed, 0 blocking, 0 required · 2 applied
Reviewed at 2026-07-02 · HEAD `79d4f8efe`

<details>
<summary>✅ 2 applied since the review</summary>

**1. Vercel `failed`/`denied` outcome paths leaked nested tool-return files** · `ui/vercel_ai/_adapter.py` · `_process_tool_call_part` / builtin-return path, `ui/vercel_ai/_event_stream.py` · `ToolOutputErrorChunk` sinks
The success sinks were stripped, but `failed`/`denied` serialized `error_text`/`approval.reason` via unwrapped `model_response_str()`/`model_response_object()` (top-level-only), so a `BinaryContent` nested in a dict/deeper list still leaked inline on the `preserve_file_data=False` default — reproduced via public `VercelAIAdapter.dump_messages`. Fixed by adding `tool_return_error_text` (strips nested files, mirrors the success paths) and routing all 4 adapter error sinks + 3 `_event_stream` twins through it, with a parametrized `failed`/`denied` regression test. AG-UI needed no change (single outcome-agnostic content sink, already wrapped).

**2. `docs/ui/ag-ui.md` nested-file wording** · `docs/ui/ag-ui.md`
Mirrored the "including files nested inside dicts or lists" clause onto the AG-UI trust-model sentence to match the `docs/ui/vercel-ai.md` clarification this PR already carries (the PR strips nested files symmetrically in AG-UI too).
</details>

<details>
<summary>⚠️ 4 informational (defer/skip)</summary>

**1. Mirror recursion convention drift** · `ui/_adapter.py` · `strip_tool_return_files` vs `_sanitize_tool_return_content` — skip: the two walkers share only the tree-walk skeleton (different leaf predicate, drop-signal, direction); consolidating is net-negative and the mirror is author-documented.
**2. `replace`+strip idiom repeated across sinks** · `ui/vercel_ai/` — skip: the sinks produce different wire shapes per adapter; `strip_tool_return_files`/`tool_return_error_text` are the correctly-extracted shared primitives.
**3. `_coerce_js_binary_data` recursion depth on client input** · `ui/vercel_ai/_adapter.py` — skip: consistent with the pre-existing inbound `_sanitize_tool_return_content` recursion; bounded by the interpreter limit into a clean `RecursionError`, not new attack surface.
**4. Empty `Uint8Array` → `{}` passthrough** · `ui/vercel_ai/_adapter.py` · `_js_binary_to_bytes` — skip: correct disambiguation from an empty user mapping, not a conformance miss.
</details>

DouweM:
Closing: the premise here — stripping agent-produced file content from **outbound** tool returns as a security measure — turns out to be a non-problem. Outbound file content is meant for the client (that's the point of a multimodal UI); there's no confidentiality boundary crossed server→client, so `strip_tool_return_files` was guarding against something that isn't a leak, and stripping just silently breaks multimodal frontends.

The genuine file-security concern is **inbound** (client-supplied `UploadedFile` / non-HTTP `FileUrl` references that the model provider resolves with the server's own credentials), which is handled by #6169 (`sanitize_message_history`) and the flag clarification in #6232 (`allow_uploaded_files`).

The two independently-correct fixes bundled here — the Vercel `_coerce_js_binary_data` `media_type` gate and the AG-UI orphan-sidecar guard — will be carried into #5255 (where the AG-UI sidecar guard may become moot, since #5255 is moving multimodal tool-return files inline into `ToolMessage.content` and dropping the tool-return sidecar entirely).

See #6232 for the flag cleanup this supersedes.

PR #6198, opened 2026-07-02
Clean up the pending task and streams on `KeyboardInterrupt` in synchronous `run_until_complete` wrappers

### LINKED ISSUE #5975: Ctrl-C during synchronous `run_sync`/`run_stream_sync` leaves pending tasks and open connections on the event loop
### Initial Checks

### Description

When a user presses Control-C, the event loop remains stuffed with old events and open sockets.
Many people have complained about it, and, well, I have a fix.
Please consider it as a constructive criticism.
Sorry if the comments in the fix burns a bit, but, this is a horrifying oversight.
I expected better from such a good library.
Hope someone fixes this mess.

```Python
try:
            response = self.agent.run_stream_sync(...)
            final_output = response.get_output()
            raw_history=None
            if (self.history is None):
                raw_history=response.all_messages()
            else:
                raw_history=self.history+response.new_messages()
            if(self.keep_records):
                processed_history=self.prune_tools_history(raw_history, SYSTEM_HISTORY_TTL)
                self.history = processed_history # Why the handover? I don't trust the language to handle it properly
                return final_output
        except KeyboardInterrupt as e:
            print("\n**BREAK PRESSED**")
            # The following code is needed because...
            # the pydantic-ai devs act like noobs when it comes to understanding threading or any sort of event system.
            # Control-C means we don't care about a clean exit, we need to get out NOW.
            # We do not care if anything was corrupted, we already know that, do not care, and normal flow is not affected.
            # Technically this correction SHOULD be either:
            # 1: caught, processed and propagate the KeyboardInterrupt
            # 2: the Agent class should have a function to call that does this.
            # There is no excuse for such lack-luster code. It's embarrassing.
            loop=asyncio.get_event_loop()
            pending = asyncio.all_tasks(loop) # get all pending tasks
            for task in pending:
                task.cancel()
            loop.run_until_complete(asyncio.gather(*pending, return_exceptions=True)) # reap those tasks
            loop.run_until_complete(loop.shutdown_asyncgens()) # this should slam shut all the open sockets...
            # Finally, propagate the break signal to the caller.
            raise KeyboardInterrupt
```

### Logfire Trace

_No response_

- Python: 3.12.3
- pydantic-2.13.3.dist
- pydantic_ai-1.60.0.dist
- pydantic_ai_middleware-0.2.1.dist
- pydantic_ai_slim-1.60.0.dist
- pydantic_core-2.46.3.dist
- pydantic_evals-1.60.0.dist
- pydantic_graph-1.60.0.dist
- pydantic_settings-2.14.0.dist
- LLM provider SDK: ??? what SDK? I write my own dogfood/run locally!

DouweM:
- Closes #5975

## Problem

When a `KeyboardInterrupt` (e.g. pressing Ctrl-C) interrupts `loop.run_until_complete(coro)`, asyncio leaves the coroutine's task suspended with its `async with`/`finally` blocks un-run. The task lingers on the loop and any open model streams and HTTP connections are never closed — exactly the "event loop remains stuffed with old events and open sockets" the reporter hit with `run_stream_sync`.

The reporter's own workaround cancels *every* task on the loop via `asyncio.all_tasks()` and calls `shutdown_asyncgens()`. That's unsafe for a library: it runs on a loop it doesn't own and would cancel the user's unrelated tasks and every async generator in the process.

## Fix

Add `_utils.run_until_complete(coro)` that wraps the coroutine in its own task and, on any `BaseException`, cancels **only that task** and drives its cleanup to completion before re-raising — so the coroutine's `async with`/`finally` blocks close their streams and connections, without touching any other task on the loop.

Route the synchronous wrappers through it:
- `Agent.run_sync`, `run_stream_sync`, `to_cli_sync`, `StreamedRunResultSync.get_output`/`validate_response_output`
- `direct.model_request_sync`
- `Embedder.*_sync`
- `Graph.run_sync` (helper copied into `pydantic_graph._utils`)
- `Dataset.evaluate_sync`, `Evaluator.evaluate_sync` (helper copied into `pydantic_evals._utils`)

`sync_async_iterator` now also `aclose()`s the underlying async iterator in a `finally`, so breaking out of a synchronous stream early closes the model stream and connection instead of leaking them until garbage collection.

The threaded `direct.model_request_stream_sync` is intentionally left as-is: it runs its loop in a background thread, so a main-thread `KeyboardInterrupt` never interrupts its `run_until_complete` and this defect doesn't apply.

## Follow-up

This fixes interrupts that land *inside* a synchronous call. `run_stream_sync` additionally keeps the `run_stream(...)` context open on a suspended `_consume_stream()` generator between calls, which is only closed at GC; making `StreamedRunResultSync` a proper (context-manager) owner of that stream is tracked separately in #3716.

## Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6198?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/_utils.py
+++ b/pydantic_ai_slim/pydantic_ai/_utils.py
@@ -410,11 +410,11 @@ def sync_async_iterator(async_iter: AsyncIterator[T]) -> Iterator[T]:
         # Close the underlying async iterator so its `async with`/`finally` blocks (which close
         # model streams and HTTP connections) run even when the consumer breaks out early or is
         # interrupted (Ctrl-C closing this generator with `GeneratorExit`), not just when the
-        # stream is exhausted.
-        aclose: Callable[[], Awaitable[None]] | None = getattr(async_iter, 'aclose', None)
-        if aclose is not None:  # pragma: no branch
+        # stream is exhausted. The `stream_*` methods always return async generators at runtime even
+        # though they're typed as `AsyncIterator`, so this narrows to the closable case.
+        if isinstance(async_iter, AsyncGenerator):  # pragma: no branch
             with suppress(BaseException):
-                run_until_complete(aclose())
+                run_until_complete(async_iter.aclose())
 
 
 def now_utc() -> datetime:
--- a/pydantic_ai_slim/pydantic_ai/_utils.py
+++ b/pydantic_ai_slim/pydantic_ai/_utils.py
@@ -45,7 +45,10 @@ from typing_extensions import ParamSpec, TypeIs, TypeVar, is_typeddict
 from typing_inspection import typing_objects
 from typing_inspection.introspection import is_union_origin
 
-from pydantic_graph._utils import AbstractSpan, run_until_complete
+from pydantic_graph._utils import (
+    AbstractSpan,
+    run_until_complete as run_until_complete,  # re-exported for the sync wrappers
+)
 from pydantic_graph.util import get_callable_name
 
 from .exceptions import UserError
--- a/pydantic_ai_slim/pydantic_ai/_utils.py
+++ b/pydantic_ai_slim/pydantic_ai/_utils.py
@@ -45,7 +45,7 @@ from typing_extensions import ParamSpec, TypeIs, TypeVar, is_typeddict
 from typing_inspection import typing_objects
 from typing_inspection.introspection import is_union_origin
 
-from pydantic_graph._utils import AbstractSpan
+from pydantic_graph._utils import AbstractSpan, run_until_complete
 from pydantic_graph.util import get_callable_name
 
 from .exceptions import UserError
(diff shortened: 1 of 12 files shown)


pydantic_ai_slim/pydantic_ai/_utils.py
```diff
@@ -397,12 +400,21 @@ def sync_anext(iterator: Iterator[T]) -> T:
...
+            try:
+                yield run_until_complete(anext(async_iter))
+            except StopAsyncIteration:
+                break
+    finally:
+        # Close the underlying async iterator so its `async with`/`finally` blocks (which close
+        # model streams and HTTP connections) run even when the consumer breaks out early or is
+        # interrupted (Ctrl-C closing this generator with `GeneratorExit`), not just when the
+        # stream is exhausted.
+        aclose: Callable[[], Awaitable[None]] | None = getattr(async_iter, 'aclose', None)
```
adtyavrdhn: Do we need this because sync_async_iterator is not typed as an AsyncGenerator? 

Can we change the type to `AsyncGenerator` if it is at runtime always a generator which it should be

pydantic_ai_slim/pydantic_ai/_utils.py
```diff
@@ -397,12 +400,21 @@ def sync_anext(iterator: Iterator[T]) -> T:
...
-        except StopAsyncIteration:
-            break
+    try:
+        while True:
+            try:
+                yield run_until_complete(anext(async_iter))
+            except StopAsyncIteration:
+                break
+    finally:
+        # Close the underlying async iterator so its `async with`/`finally` blocks (which close
```
adtyavrdhn, line 410: We cannot use both interchangeably (iterator and generator) it is confusing and not swappable

pydantic_ai_slim/pydantic_ai/_utils.py
adtyavrdhn: Could probably do an isinstance() check instead

PR #6211, opened 2026-07-02
fix(google): retain file search grounding with native tool parts

### LINKED ISSUE #6207: Gemini File Search citations/grounding dropped on Gemini 3 when combined with function tools
## Summary

When Gemini's built-in **File Search** tool (`FileSearchTool`) is combined with user **function tools** on a **Gemini 3+** model, File Search citations are silently lost: `NativeToolReturnPart.content` for the `file_search` tool is `None`, so the retrieved-context `custom_metadata` (e.g. a document `source_url`) never reaches the model response. The grounding data **is** returned by the API — pydantic-ai just doesn't surface it in this code path.

This is the File Search analogue of #2890 (web search grounding) and shares the root mechanism with #6051.

## Repro

Register `FileSearchTool` **and** at least one `@agent.tool` function tool, on `gemini-3.5-flash` (Gemini 3+), against a File Search store whose documents were uploaded with `custom_metadata` (e.g. `source_url`). Ask a question that retrieves a document.

```python
from pydantic_ai.capabilities import NativeTool
from pydantic_ai.native_tools import FileSearchTool
from pydantic_ai.messages import ModelResponse, NativeToolReturnPart

agent = Agent(
    "google-gla:gemini-3.5-flash",
    capabilities=[NativeTool(FileSearchTool(file_store_ids=["fileSearchStores/<id>"]))],
)

@agent.tool
def noop(ctx: RunContext[None]) -> str:  # any function tool forces the combination
    return "ok"

result = await agent.run("<question that hits a document with source_url in custom_metadata>")

for m in result.new_messages():
    if isinstance(m, ModelResponse):
        for p in m.parts:
            if isinstance(p, NativeToolReturnPart) and p.tool_name == "file_search":
                print("content:", p.content)   # -> None  (expected: list of retrieved contexts)
```

A raw `google.genai` `generate_content` call with the same tool returns `candidate.grounding_metadata.grounding_chunks[].retrieved_context` populated with `text`, `title`, and `custom_metadata` (including `source_url`). So the data is available from the API; it's dropped by pydantic-ai.

## Root cause

On Gemini 3+, `GoogleModel._get_tool_config` sets `tool_config.include_server_side_tool_invocations=True` (required — the API otherwise returns `400 INVALID_ARGUMENT: "Please enable tool_config.include_server_side_tool_invocations to use Built-in tools with Function calling."`). Gemini then returns **explicit** `tool_call`/`tool_response` parts.

`_process_response_from_parts` guards grounding reconstruction behind `if not _has_native_tool_invocations(parts):`, so on Gemini 3 `_map_file_search_grounding_metadata` is **skipped** (to avoid duplicate parts). But the explicit file_search `tool_response` part carries `content=None` — the retrieved contexts only exist in `candidate.grounding_metadata`, which is discarded. The streaming path has the equivalent guard (`if not self._has_tool_invocations:` around `_handle_file_search_grounding_metadata_streaming`, which additionally requires `_file_search_tool_call_id`, never set in this mode).

Net effect: with File Search + function tools on Gemini 3, `file_search` `NativeToolReturnPart.content` is always `None`, and no citation/grounding metadata is recoverable from the response — for both `request` and `request_stream`.

## Expected

The file_search retrieved contexts (with `custom_metadata`) from `grounding_metadata` should be surfaced into `NativeToolReturnPart.content` even when explicit tool-invocation parts are present — merged into the explicit part (or emitted as an additional grounding-derived part), so downstream code can read citations. Pre-Gemini-3 (no explicit invocations) already works.

## Environment

- `pydantic-ai-slim` 1.107.0 (behaviour also present on `main` as of 2026-07 — same `_has_native_tool_invocations` guard).
- `google-genai` (current), provider `google-gla`, model `gemini-3.5-flash`.

## Related

- #2890 — same idea for the built-in web search tool (`google_search`) grounding; does not cover File Search or the explicit-invocations case.
- #6051 / #6052 — same `include_server_side_tool_invocations` mechanism (built-in tool + function tools on Gemini 3), different symptom.
- #4115 — File Search query-time `metadata_filter` (unrelated to returning `custom_metadata`).

Happy to send a PR if you agree on the approach (surface grounding-derived file_search contexts alongside the explicit tool parts).

Sehlani042:
Closes #6207

### Summary

When Gemini returns explicit native File Search `tool_call` / `tool_response` parts, the response mapper skips metadata-based native tool reconstruction to avoid duplicate parts. That is correct for call reconstruction, but Gemini can still leave the explicit File Search `tool_response.response` empty while putting the retrieved contexts in `grounding_metadata`.

This PR keeps the explicit native tool parts and only fills an empty File Search `NativeToolReturnPart.content` from `grounding_metadata.grounding_chunks`. It applies to both non-streaming and streaming mapping paths and does not reopen full metadata reconstruction when native tool invocations are present.

### Verification

- RED: `uv run pytest tests/models/google/test_native_tools.py::test_file_search_tool_response_uses_grounding_metadata_when_response_is_empty -q` failed with `NativeToolReturnPart.content is None`.
- GREEN: `uv run pytest tests/models/google/test_native_tools.py::test_file_search_tool_response_uses_grounding_metadata_when_response_is_empty -q`
- `uv run pytest tests/models/google/test_native_tools.py tests/models/test_google.py::test_google_model_file_search_tool tests/models/test_google.py::test_google_model_file_search_tool_stream -q`
- `uv run ruff check pydantic_ai_slim/pydantic_ai/models/google.py tests/models/google/test_native_tools.py`
- `uv run ruff format --check pydantic_ai_slim/pydantic_ai/models/google.py tests/models/google/test_native_tools.py`
- `PYRIGHT_PYTHON_IGNORE_WARNINGS=1 uv run pyright pydantic_ai_slim/pydantic_ai/models/google.py tests/models/google/test_native_tools.py`
- `git diff --check`

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6211?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/google.py
+++ b/pydantic_ai_slim/pydantic_ai/models/google.py
@@ -1416,6 +1416,7 @@ class GeminiStreamedResponse(StreamedResponse):
                         yield self._parts_manager.handle_part(vendor_part_id=uuid4(), part=tool_call_part)
                     elif part.tool_response:
                         tool_response_part = _map_tool_response(part.tool_response, self.provider_name)
+                        _fill_empty_file_search_return_content(tool_response_part, candidate.grounding_metadata)
                         tool_response_part.provider_details = provider_details
                         yield self._parts_manager.handle_part(vendor_part_id=uuid4(), part=tool_response_part)
                     elif part.executable_code is not None:
@@ -1786,6 +1787,8 @@ def _process_response_from_parts(
     for part in parts:
         item, code_execution_tool_call_id = _process_part(part, code_execution_tool_call_id, provider_name)
         if item is not None:
+            if isinstance(item, NativeToolReturnPart):
+                _fill_empty_file_search_return_content(item, grounding_metadata)
             items.append(item)
 
     return ModelResponse(
@@ -1936,6 +1939,18 @@ def _map_tool_response(tool_response: ToolResponse, provider_name: str) -> Nativ
         tool_call_id=tool_response.id or _utils.generate_tool_call_id(),
         content=tool_response.response,
     )
+
+
+def _fill_empty_file_search_return_content(
+    item: NativeToolReturnPart, grounding_metadata: GroundingMetadata | None
+) -> None:
+    if item.tool_name != FileSearchTool.kind or item.content is not None:
+        return
+
+    grounding_chunks = grounding_metadata.grounding_chunks if grounding_metadata else None
+    retrieved_contexts = _extract_file_search_retrieved_contexts(grounding_chunks)
+    if retrieved_contexts:
+        item.content = retrieved_contexts
 
 
 def _map_grounding_metadata(
--- a/tests/models/google/test_native_tools.py
+++ b/tests/models/google/test_native_tools.py
@@ -29,11 +29,13 @@ from pydantic_ai.native_tools import (
 from ...conftest import try_import
 
 with try_import() as imports_successful:
-    from google.genai.types import ToolType
+    from google.genai.types import GroundingMetadata, Part, ToolType
 
     from pydantic_ai.models.google import (
         _content_model_response,  # pyright: ignore[reportPrivateUsage]
+        _process_response_from_parts,  # pyright: ignore[reportPrivateUsage]
     )
+    from pydantic_ai.usage import RequestUsage
 
 pytestmark = pytest.mark.skipif(not imports_successful(), reason='google-genai not installed')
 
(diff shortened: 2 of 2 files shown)


adtyavrdhn:
Thanks for the PR, easier for me to fix it with the changes in #6216

PR #6221, opened 2026-07-02
Document why `_clean_message_history` merges consecutive `ModelRequest`s regardless of `conversation_id`/`metadata`

adtyavrdhn:
- Refs #5731
- Refs #5629

An earlier version of this PR made `_clean_message_history` guard the merge of consecutive `ModelRequest`s on `conversation_id`/`metadata` (skipping the merge when they conflicted) and preserve `run_id`/`conversation_id`/`metadata` across the merge via `dataclasses.replace`.

We decided **not** to do that. Refusing to merge on mismatched `conversation_id` or `metadata` would leave two consecutive requests where a provider expects a single request per turn — a real regression for providers (and provider-side conversation state) — just to preserve fields the model request node never reads. Those fields are only bookkeeping for callers and are never part of what gets sent to the model.

This PR therefore reverts the behavioral change and instead documents the reasoning with a comment in `_clean_message_history`, so the next person who looks at it doesn't re-introduce the guard.

### Checklist


adtyavrdhn:
Close this, add a comment explaining the reason we don't do this.

Issue #3138, opened 2025-10-11
Framework-level testing utilities for unit testing agents


amnox:
### Description

PydanticAI has excellent testing infrastructure (`TestModel`, `FunctionModel`, `capture_run_messages()`), but lacks framework-level utilities that make unit testing agents ergonomic and maintainable.

This proposal suggests adding testing helpers that complement existing tools and reduce boilerplate in agent unit tests.

---

## Problem

Currently, testing an agent requires manual inspection of captured messages:

```python
from pydantic_ai import Agent, capture_run_messages
from pydantic_ai.models.test import TestModel

agent = Agent('openai:gpt-4o', ...)

def test_weather_agent():
    with capture_run_messages() as messages:
        with agent.override(model=TestModel()):
            result = agent.run_sync('Weather in London?')

    # Manual inspection - verbose and brittle
    assert len(messages) == 4
    assert messages[1].parts[0].tool_name == 'get_forecast'
    assert messages[1].parts[0].args['location'] == 'a'  # TestModel generates 'a'
    assert isinstance(messages[3].parts[0], TextPart)
```

**Pain points:**

1. **Verbose**: Every test needs to manually index into `messages` list
2. **Brittle**: If message order changes, tests break
3. **Hard to read**: Test intent is buried in assertions
4. **No semantic assertions**: Can't easily assert "tool X was called with args Y"

---

## Proposed Solution

Add a `pydantic_testing` module (or extend existing test utilities) with:

### 1. Assertion Helpers

```python
from pydantic_ai.testing import AgentAsserter

def test_weather_agent():
    asserter = AgentAsserter(agent)

    with asserter.override(model=TestModel()):
        result = asserter.run_sync('Weather in London?')

    # Framework-level assertions
    asserter.expect_tool_called('get_forecast', location='London')
    asserter.expect_no_errors()
    asserter.expect_output_contains('weather')
```

### 2. Fixture Support with Real Data

Currently `TestModel` generates schema-valid but meaningless data (`location='a'`, `date='2024-01-01'`).

Proposal: Support YAML fixtures with realistic data:

```yaml
# fixtures/weather_scenario.yaml
model_responses:
  - on_message: "Weather in London?"
    call_tool:
      name: get_forecast
      args:
        location: London  # Extracted from prompt
    return: "Sunny, 22°C"

  - on_tool_return: "Sunny, 22°C"
    respond: "The weather in London is sunny, 22°C."
```

```python
[...]

PR #6225, opened 2026-07-02
chore: add testing helpers for tool calls

Sehlani042:
- Refs #3138

### Summary

This adds a small `pydantic_ai.testing` helper module for deterministic assertions over function tool calls without introducing fixture files, recording/replay, or snapshot machinery.

The first slice is intentionally narrow:

- `tool_calls(source, across_history=False)` collects `ToolCallPart`s from either a `ModelResponse` or an `AgentRunResult`
- `assert_tool_call(source, name, args=...)` asserts that a call exists and returns the matching part for further checks
- `assert_tool_call_sequence(source, [...])` asserts exact tool-call order
- `across_history=True` scans all `ModelResponse`s in an `AgentRunResult` when the final response no longer contains the tool call

This follows the issue discussion direction of staying Python-native and building on `result.response.tool_calls` / `result.all_messages()` rather than adding YAML fixtures or new snapshot behavior.

### Tests

- `uv run pytest tests/test_testing.py tests/test_messages.py::test_model_response_convenience_methods -q`
- `uv run ruff check pydantic_ai_slim/pydantic_ai/testing.py tests/test_testing.py`
- `uv run ruff format --check pydantic_ai_slim/pydantic_ai/testing.py tests/test_testing.py`
- `PYRIGHT_PYTHON_IGNORE_WARNINGS=1 uv run pyright pydantic_ai_slim/pydantic_ai/testing.py tests/test_testing.py`

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6225?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/testing.py
+++ b/pydantic_ai_slim/pydantic_ai/testing.py
@@ -1,0 +1,79 @@
+from __future__ import annotations as _annotations
+
+from collections.abc import Sequence
+from typing import Any, TypeAlias
+
+from .messages import ModelResponse, ToolCallPart
+from .run import AgentRunResult
+
+ToolCallSource: TypeAlias = AgentRunResult[Any] | ModelResponse
+
+_UNSET = object()
+
+
+def tool_calls(source: ToolCallSource, *, across_history: bool = False) -> list[ToolCallPart]:
+    """Return function tool calls from a model response or agent run result.
+
+    By default this reads the final response, matching `result.response.tool_calls`.
+    Set `across_history=True` to collect calls from every model response in an
+    agent run's message history.
+    """
+    if isinstance(source, ModelResponse):
+        return source.tool_calls
+
+    if not across_history:
+        return source.response.tool_calls
+
+    return [
+        call for message in source.all_messages() if isinstance(message, ModelResponse) for call in message.tool_calls
+    ]
+
+
+def assert_tool_call(
+    source: ToolCallSource,
+    tool_name: str,
+    *,
+    args: Any = _UNSET,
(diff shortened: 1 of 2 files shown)


dsfaccini:
we'll consider this if we can use it to noticeable reduce boilerplate in our own test suite, I'll have claude check

dsfaccini:
This reply was posted by Claude Code using claude-opus-4-8 on behalf of David.

Thanks for this. The bar for pulling test helpers into our public API is whether they noticeably cut boilerplate in our own suite, so we checked.

Most of it overlaps with what we already ship. `tool_calls(response)` is `ModelResponse.tool_calls`, and the cross-history variant is our existing `iter_message_parts()` conftest helper. `assert_tool_call_sequence` also cuts against our house style: we assert tool-call order with inline snapshots (`tool_names == snapshot([...])`), auto-maintained by `--inline-snapshot=fix`, which a bespoke assert would lose.

What we did find is that `iter_message_parts()` was underused (hand-rolled ~80x across `tests/`), so we are adopting it more widely in https://github.com/pydantic/pydantic-ai/pull/6522 instead of adding a new public `testing.py` module. Every public symbol is a lasting backward-compat commitment.

Appreciate the contribution and the nudge to make our tests less repetitive.

Issue #3220, opened 2025-10-22
How to handle multi-turn conversation evals with Pydantic AI Evals


xjose97x:
### Question

Hey everyone 👋

I’m trying to figure out how to handle **multi-turn conversations** in Pydantic AI Evals.

Are there any examples or docs that demonstrate this? I’d love to see how others are approaching it.

Thanks a lot!

DouweM:
@xjose97x You can have your evaluated function return the entire run result rather than just the output, and then use custom evaluators that check `result.all_messages()`, `result.response` etc. Does that help?

xjose97x:
> [@xjose97x](https://github.com/xjose97x) You can have your evaluated function return the entire run result rather than just the output, and then use custom evaluators that check `result.all_messages()`, `result.response` etc. Does that help?

Yeah, I think I'll follow that approach.
Thanks for the tip, @DouweM !

PR #6226, opened 2026-07-02
feat(evals): prototype multi-turn eval primitives

EgonFerri:
## Summary

This PR adds a small prototype for multi-turn conversation evals, related to #3220.

The issue asks how to handle multi-turn conversations in Pydantic AI Evals. This is already possible with the existing primitives, but each project currently has to rebuild the same scaffolding: simulator and target message history, max-turn stopping, transcript capture, per-turn timings, completion state, and basic evaluators.

I recently implemented this pattern for a production eval use case, and the existing primitives were flexible enough to support it. However, the same structure seems broadly useful for testing agents that need to interact with a simulated user over multiple turns, so I think it may be worth considering as a small built-in primitive.

The goal here is not to propose this exact API as final. This is a lightweight implementation to make the idea concrete and testable. If maintainers think this would be useful, I would be happy to champion the feature and iterate together on the best shape and location for it in `pydantic_evals`.

## What this includes

- A reusable multi-turn conversation runner.
- A simulator-agent factory that turns a scenario into a structured simulator agent.
- Conversation output models with transcript, turn count, stop reason, and timing metrics.
- Small evaluators for completion, max turns, and average target-agent latency.
- A minimal animal-sounds example showing how concise the client-side eval can become.

## Why this may be useful

Multi-turn evals are a common pattern when testing agents that need to interact with a user, clarify intent, or complete a task over several messages. A small primitive could make this easier and more consistent while still keeping the existing eval system flexible.

## How to test

```bash
uv run --env-file .env python examples/pydantic_ai_examples/evals/multi_turn/examples.py
```

The live example requires `OPENAI_API_KEY`; `LOGFIRE_TOKEN` is optional but useful to inspect the eval run in Logfire.

```bash
uv run pytest tests/evals/test_multi_turn_conversation_example.py
```

The pytest coverage is intentionally focused on the prototype helpers.

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6226?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


adtyavrdhn:
@dmontagu Have a glance here if it is something you might consider useful, lots of public API though

dmontagu:
I like the idea in principle. However, it would help to have a bit more exposition about how you would expect end users to make use of this. In particular, I'd like to better understand exactly which parts you would want to upstream and have be public (everything in `multi_turn`? Just a subset?), and a little bit more exposition about the scenario of a user going in.

This adds a lot of surface area of structs/etc. that I would ideally like to either reduce or generalize if possible (and it's not clear to me if it is). But it would be a lot easier to determine what the requirements should be in a world where we want to refactor the API with a bit more exposition about what the example is doing and which parts _matter_ vs. are just implementation details.

To be clear, I'm fine if you just have an AI generate that exposition, as long as you look just closely enough to be confident you agree with its assessment. And I'll be happy to review more / iterate on this once we have that.

EgonFerri:
Thanks @dmontagu, this makes sense. I have reworked the proposal around a much smaller public surface and added two examples to make the intended user journey and the boundary between reusable primitives and application code more explicit.

The proposed public surface is now limited to three types in `pydantic_evals.multiturn`:

- `ConversationTask`, which owns the conversation loop and can be passed to the existing `Dataset.evaluate()` as `conversation.run`;
- `ConversationTurn`, the typed record passed to simulator callbacks and exposed to evaluators; and
- `ConversationResult`, the typed dataset output containing the completed turns and stop reason.

`ConversationTask` has two construction paths. `ConversationTask.from_agents()` is the convenience path for the common case where both participants are Pydantic AI agents; it manages separate message histories and creates fresh sessions for every case, repeat, and retry. The generic constructor accepts factories that return simple message-in/output-out callbacks, so the target can instead be a deterministic function, a remote service, or another agent implementation. The factories are important because any participant-specific state must be isolated across concurrent dataset runs.

The two examples are intended to validate different properties of this proposal:

1. `example_05_multiturn.py` demonstrates the basic case and the convenience of the abstraction. When the target and simulator are both Pydantic AI agents, the user only defines the scenario, the simulator's structured decision, and a few callbacks. `ConversationTask.from_agents()` provides the conversation loop, message-history management, isolation, stopping, trajectory capture, metrics, and integration with `Dataset.evaluate()`.
2. `example_06_multiturn_veterinary.py` is an extensibility test. It starts from a use case with additional application requirements—external text context, structured patch evaluation, tool-use assertions, and comparison with a non-Agent deterministic target—and shows that they can be implemented on top of the same primitive using its prompt hooks, generic factories, typed result, and the existing evaluator API. The veterinary domain is just a concrete vehicle for testing whether the proposed API remains generic; none of those domain concepts needs to move into the library.

The end-user flow in the first, common case is intended to be roughly:

```python
conversation = ConversationTask.from_agents(
    target_agent=target_agent,
    simulator_agent=simulator_agent,
    first_message=lambda scenario: scenario.first_message,
    next_message=lambda decision: None if decision.done else decision.message,
    max_turns=lambda scenario: scenario.max_turns,
)

report = await dataset.evaluate(conversation.run)
```

The application still owns the scenario and simulator-output models. `Scenario` and `SimulatorDecision` in the basic example are therefore not proposed as library types; they only describe that particular simulated user and its completion protocol.

The reusable behavior in `ConversationTask` is deliberately limited to:

- alternating the target and simulator;
- creating isolated sessions for each evaluation run;
- enforcing a maximum number of target turns;
- returning a typed trajectory and explicit stop reason; and
- emitting per-role spans and target-turn timing metrics through the existing evaluation run.

There is no separate dataset type or evaluation runner, and this does not require changes to `Dataset`. Conversation correctness is still evaluated with the existing evaluator API.

The second example deliberately requires more application-specific behavior than the core abstraction provides. Its purpose is to show that users can add that behavior without reimplementing the multi-turn mechanics or requiring more public multi-turn types.

To run the examples from the repository root, create a `.env` file containing:

```dotenv
OPENAI_API_KEY=your-openai-api-key

LOGFIRE_TOKEN=your-logfire-write-token
```

Both examples use `openai:gpt-4.1-mini` to keep the live runs inexpensive, so `OPENAI_API_KEY` is required. `LOGFIRE_TOKEN` is optional; without it, the examples still run and print their reports locally.

Run the basic example, which demonstrates the concise two-Agent abstraction, with:

```bash
uv run --env-file .env python examples/pydantic_ai_examples/evals/example_05_multiturn.py
```

Run the advanced example, which demonstrates application-level extensions and compares an Agent target with a deterministic target, with:

```bash
uv run --env-file .env python examples/pydantic_ai_examples/evals/example_06_multiturn_veterinary.py
```

The examples are invoked by file path because running them with `python -m pydantic_ai_examples.evals...` first executes the existing `pydantic_ai_examples.evals.__init__` module and its unrelated example brokens imports.

So the distinction I would propose is:

- **Public and reusable:** the conversation task, turn/result types, the Agent convenience constructor, and their callback/factory contracts.
- **Example implementation details:** scenario schemas, simulator decisions, file handling, patch models, tool assertions, and the deterministic target.

There is an unavoidable tension here between making the common case simple and immediate by adopting some opinionated decisions, and keeping the primitive generic enough to support targets, simulators, and application requirements that we have not anticipated. The proposed split—an opinionated `from_agents()` convenience path over a generic factory-based constructor—seems like a reasonable compromise: the basic case stays concise, while applications can drop down one level without reimplementing the multi-turn mechanics.

That said, I do not consider these exact names, signatures, or even this overall split final. This is the best balance I have found so far between convenience, a small public surface, and extensibility, but I am completely open to revisiting or restructuring the proposal if you think a different abstraction would fit Pydantic Evals better.

EgonFerri:
Hi, @dmontagu let me know if this could make sense, otherwise I'll close it

PR #6238, opened 2026-07-03
feat(instrumentation): opt-in `gen_ai.agent.name` on client metrics

### LINKED ISSUE #6204: Option to emit `gen_ai.agent.name` as a dimension on GenAI client metrics (token usage, cost, TTFT)
## Context

#5967 adds the `gen_ai.client.operation.time_to_first_chunk` histogram for streaming requests. Like the existing `gen_ai.client.token.usage` and `operation.cost` histograms, its data points carry the OTel GenAI semconv metric attribute set: `gen_ai.operation.name`, `gen_ai.provider.name`/`gen_ai.system`, `gen_ai.request.model`, `gen_ai.response.model` (plus resource-level `service.name` / `deployment.environment.name`).

That means the metrics can be broken down by model, provider, service, and environment — but **not by agent**. In a multi-agent app, a planner and a summarizer calling the same model are indistinguishable in the metric series, so per-agent cost/latency/TTFT views and alerts can't be built from metrics alone.

## Why we should NOT emit `gen_ai.agent.name` by default

- The semconv metric attribute group for GenAI client metrics deliberately excludes agent identity today; `gen_ai.agent.name` is only specified for spans (`invoke_agent`).
- No major framework stamps agent identity on the standard client metric series by default: Microsoft Agent Framework / Semantic Kernel follow the semconv metric set as-is, OpenLLMetry keeps entity/association context on spans, and OpenLIT extends metrics only with app-level dimensions (`gen_ai.application_name`, `gen_ai.environment`) configured once at init.
- Turning it on by default would split every existing user's token/cost series (any dashboard keyed on the current dimensions), and adds cardinality that not everyone wants.
- The OTel GenAI SIG is actively working on standardizing exactly this — shipping a default now risks diverging from whatever shape lands.

## Why an opt-in is still worth offering

- **OTel Views can drop metric attributes downstream, but can never add them.** If the SDK doesn't stamp the attribute, a metrics-only consumer (Prometheus/Grafana, no span pipeline) has *no path at all* to per-agent series. The SDK is the only place this option can live.
- `gen_ai.agent.name` is a low-cardinality logical label (bounded by code, not user input), so the cardinality cost is modest and the user opts into it knowingly.
- There is an open upstream proposal to standardize `gen_ai.agent.name` as a documented dimension on GenAI **client metrics** for precisely this use case: open-telemetry/semantic-conventions-genai#91. Related SIG discussions: open-telemetry/semantic-conventions-genai#249 (span + metric definition principles for agents), open-telemetry/semantic-conventions-genai#243 (representing multiple agents on the same telemetry), and the [AI agent observability blog post](https://opentelemetry.io/blog/2025/ai-agent-observability/). An opt-in keeps us aligned-by-default while letting users get the breakdown today; if #91 lands, flipping the default (or removing the flag) becomes a straightforward follow-up.

## Proposed shape

A boolean on `InstrumentationSettings` (precedent: `use_aggregated_usage_attribute_names`), e.g.:

```python
InstrumentationSettings(include_agent_name_in_metric_attributes=True)
```

When enabled, stamp `gen_ai.agent.name` on **all three** client histograms (token usage, cost, time-to-first-chunk) — consistently, not just TTFT — whenever the model request is made on behalf of a named agent (available from baggage / run context). Omit the attribute for direct model use and unnamed agents. Changelog should note that enabling it splits existing series.

Note: Logfire itself doesn't need this — per-agent aggregation on the platform comes from spans (`model_calls` joined to agent runs), which is higher fidelity. This option is for users consuming pydantic-ai metrics in metrics-only OTel backends.

Bartok9:
## Summary

- Adds an opt-in `include_agent_name_in_metric_attributes` flag to `InstrumentationSettings`.
- When enabled, `gen_ai.agent.name` (read from OTel baggage) is stamped on the GenAI **client metric** data points — `gen_ai.client.token.usage` and `operation.cost` — so users can break cost/usage metrics down by agent.

## Motivation

Closes #6204.

GenAI client metric data points currently carry only the semconv metric attribute set (`gen_ai.operation.name`, provider, request/response model). That means metrics can be sliced by model/provider/service/environment but **not by agent** — in a multi-agent app, a planner and a summarizer calling the same model are indistinguishable in the metric series.

OTel Views can drop metric attributes downstream but can never add them, so a metrics-only consumer (Prometheus/Grafana with no span pipeline) has no path to per-agent series unless the SDK stamps the attribute. This makes the SDK the only place the option can live.

## Fix

`open_model_request_span` already captures the agent-run baggage (`gen_ai.agent.name`) into the span `attributes` dict. The new flag reuses that captured value in the shared `metric_attributes` dict passed to `record_metrics`, so every client histogram recorded through that path picks it up consistently.

Design choices (matching the issue and semconv guidance):
- **Off by default** — the current OTel GenAI semconv specifies `gen_ai.agent.name` only for spans (`invoke_agent`), not client metrics. Enabling by default would split every existing user's token/cost series and add cardinality not everyone wants. Precedent: `use_aggregated_usage_attribute_names`.
- **Omitted for direct model use and unnamed agents** — the attribute is only added when the request runs on behalf of a named agent (present in baggage).
- Low-cardinality logical label (bounded by code, not user input), opted into knowingly.

## Tests

- `test_metrics_omit_agent_name_by_default` — flag off + baggage set → attribute absent on all data points.
- `test_metrics_include_agent_name_when_enabled` — flag on + baggage set → `gen_ai.agent.name` present on token-usage and cost data points.
- `test_metrics_omit_agent_name_when_enabled_but_unnamed` — flag on, no agent name in baggage → attribute omitted (direct model use).

## Verification

- `uv run pytest tests/models/test_instrumented.py` — 48 passed
- `uv run ruff format --check` / `ruff check` — clean on changed files
- `uv run pyright` — 0 errors on the changed files
- Did NOT change default behavior: the attribute is only emitted when the new flag is explicitly enabled AND an agent name is present in baggage.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6238?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/_instrumentation.py
+++ b/pydantic_ai_slim/pydantic_ai/_instrumentation.py
@@ -247,13 +247,18 @@ def open_model_request_span(
                 price_calculation = None
 
                 def _record_metrics() -> None:
-                    metric_attributes = {
+                    metric_attributes: dict[str, AttributeValue] = {
                         GEN_AI_PROVIDER_NAME_ATTRIBUTE: system,
                         GEN_AI_SYSTEM_ATTRIBUTE: system,
                         'gen_ai.operation.name': operation,
                         'gen_ai.request.model': request_model,
                         'gen_ai.response.model': response_model,
                     }
+                    if (
+                        settings.include_agent_name_in_metric_attributes
+                        and (agent_name := attributes.get(AGENT_NAME_BAGGAGE_KEY)) is not None
+                    ):
+                        metric_attributes[AGENT_NAME_BAGGAGE_KEY] = agent_name
                     settings.record_metrics(response, price_calculation, metric_attributes)
 
                 record_metrics = _record_metrics
--- a/pydantic_ai_slim/pydantic_ai/models/instrumented.py
+++ b/pydantic_ai_slim/pydantic_ai/models/instrumented.py
@@ -66,6 +66,7 @@ class InstrumentationSettings:
     include_content: bool = True
     version: Literal[2, 3, 4, 5] = DEFAULT_INSTRUMENTATION_VERSION
     use_aggregated_usage_attribute_names: bool = True
+    include_agent_name_in_metric_attributes: bool = False
 
     def __init__(
         self,
@@ -76,6 +77,7 @@ class InstrumentationSettings:
         include_content: bool = True,
         version: Literal[2, 3, 4, 5] = DEFAULT_INSTRUMENTATION_VERSION,
         use_aggregated_usage_attribute_names: bool = True,
+        include_agent_name_in_metric_attributes: bool = False,
     ):
         """Create instrumentation options.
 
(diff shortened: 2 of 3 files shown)


adtyavrdhn:
Closing pending internal discussions. Please do not file community PRs for this.

PR #5775, opened 2026-06-04
Reject conflicting native MCP tool IDs

mqy9787:
## Summary

- reject conflicting native tool IDs within the same agent-level or run-level capability layer
- preserve the existing runtime override behavior for agent-level native tools
- add regression coverage for conflicting `MCPServerTool` instances and the existing override path

## Rationale

Native tools are ultimately keyed by `tool.unique_id`. The final merged request parameters intentionally use last-wins behavior so run-level native tools can override agent-level defaults. Within a single capability layer, though, two different native tools with the same ID are ambiguous and can silently bind one stable ID to a different definition.

For native MCP tools, that ambiguity can bind one stable tool ID to an unexpected server URL or authorization token. This PR keeps the existing cross-layer override behavior and fails early only when the duplicate ID is ambiguous within the same layer.

## Tests

- `uv run pytest tests/test_agent.py::test_agent_native_tools_runtime_vs_agent_level tests/test_agent.py::test_agent_rejects_conflicting_agent_level_native_tool_ids tests/test_agent.py::test_agent_rejects_conflicting_run_level_native_tool_ids tests/models/test_model_request_parameters.py -q`
- `PYRIGHT_PYTHON_IGNORE_WARNINGS=1 uv run pyright tests/models/test_model_request_parameters.py`
- `uv run ruff format --check pydantic_ai_slim/pydantic_ai/models/__init__.py pydantic_ai_slim/pydantic_ai/agent/__init__.py tests/models/test_model_request_parameters.py tests/test_agent.py`
- `uv run ruff check pydantic_ai_slim/pydantic_ai/models/__init__.py pydantic_ai_slim/pydantic_ai/agent/__init__.py tests/models/test_model_request_parameters.py tests/test_agent.py`
- `uv run python -m py_compile pydantic_ai_slim/pydantic_ai/models/__init__.py pydantic_ai_slim/pydantic_ai/agent/__init__.py tests/models/test_model_request_parameters.py tests/test_agent.py`
- `git diff --check`

PR #6242, opened 2026-07-03
Reject conflicting native tool IDs within a capability layer

hramezani:
## Summary

Reject native tools that share a `unique_id` but carry **conflicting** definitions within a single capability layer, failing fast with a clear `UserError` instead of silently binding a stable id to an unexpected definition. Alternative to the stale #5775.

## Rationale

Native tools are keyed by `unique_id` when request parameters are deduplicated in `Model.prepare_request`. That dedup is intentionally **last-wins across layers**, so a run-level native tool can override an agent-level default with the same id. **Within a single layer**, though, two *different* tools sharing an id are ambiguous — the silent last-wins would bind a stable id (e.g. an `MCPServerTool` id) to an unexpected definition such as a different server URL or authorization token.

## Approach

Since #6332, the agent resolves each capability layer's `for_run` exactly once and keeps the resolved layers addressable, so this PR is now validation-only: add `_validate_native_tool_ids(...)` (mirroring `_validate_capability_ids`) and call it per layer, each read from its **resolved** form so that native tools contributed dynamically (e.g. by capability functions) are covered too:

- **base layer** — the agent's own capabilities or their `override(spec=...)` replacement, validated per run post-resolution; a construction-time check remains as a fail-fast for static agent-level conflicts
- **run layer** — per-run `capabilities=[...]` and spec-resolved capabilities
- **`override(native_tools=...)`** — the override list as provided

**Identical** duplicates within a layer are still allowed and collapsed by the existing dedup; only *conflicting* definitions raise.

An earlier revision of this PR pre-resolved the per-run capabilities and let them be resolved a second time, violating `for_run`'s once-per-run contract — that mechanism was superseded by #6332 (see [discussion](https://github.com/pydantic/pydantic-ai/pull/6242#issuecomment-4909468007) and #6331).

## Tests

Regression tests per layer in `tests/test_agent.py`: conflicting ids at agent level (construction), agent level via capability functions (resolved base layer), run level (static and via capability functions), `override(spec=...)`, and `override(native_tools=...)`, plus identical-duplicates-allowed.

### Checklist


DouweM:
@hramezani Looks reasonable to me so I'm reopening, lmk if there was a particular reason you closed it!

hramezani:
> @hramezani Looks reasonable to me so I'm reopening, lmk if there was a particular reason you closed it!

Thanks @DouweM for finding this. I closed it by mistake

DouweM:
@hramezani I've taken a closer look and I don't think we can merge this in its current shape — the problem is the pre-resolution block added in 6ab2496 (and its "idempotent" justification comment), not the validation idea itself.

The block calls `cap.for_run(initial_ctx)` on every per-run capability up front, and then `effective_capability.for_run(initial_ctx)` below resolves them all a *second* time. That violates `AbstractCapability.for_run`'s documented contract ("Called once per run... Override to return a fresh instance for per-run state isolation"):

- Fresh-instance-per-run is the blessed pattern, so "the resolved capabilities are idempotent under `for_run`" isn't something we can assume — and per-run side effects in `for_run` (which e.g. the durable-exec integrations rely on for deterministic replay) now run twice.
- When a capability's second `for_run` returns a fresh instance, the id validation and the `override(native_tools=...)` merge operate on the first-pass instance while the run actually uses the second-pass one.
- The `DynamicCapability` filter only special-cases a bare wrapper whose factory returned `None`; a factory that returns a capability still gets that capability's `for_run` called again by the second pass.

The commit history shows this was already understood: 4f3dcac restructured to resolve everything exactly once, then ff0ad71 reverted it because `test_complex_agent_run_in_workflow` regressed. That test failure is telling us the restructure needs more care (the recomposed capability structure differs from what the Temporal plugin observes), not that the double resolution is fine.

The pre-resolution was only added to fix the review finding that `override(native_tools=...)` drops native tools contributed by capability functions — but that's a pre-existing bug on `main`, independent of the validation this PR is about. I've filed it separately as #6331 and will fix it properly (single `for_run`, no pre-resolution) in its own PR.

I'm converting this to a draft in the meantime. Once #6331 is fixed, this PR can drop the pre-resolution block entirely and go back to the much simpler validation-only shape of e03eedb, rebased on top of that fix.

PR #6247, opened 2026-07-03
Add `RunContext.output` for capability final output commits

### LINKED ISSUE #6243: Expose a public capability API to commit validated tool results as final output
## Summary

Capabilities that orchestrate tools, such as a Code Mode capability, sometimes produce a structured object that already matches the agent final output schema. It should be possible for the capability to validate and commit that object as the final result without asking the model for another response whose only job is to copy the same data into an output tool call or text output.

## Use case

A Code Mode tool executes Python that calls other tools and returns a structured result. That result may contain exact IDs or other opaque values from tool responses. Requiring an extra model round trip to restate the result adds latency and cost, and it can corrupt or omit long identifiers.

The desired flow is:

1. Model calls a Code Mode function tool.
2. Code Mode returns a structured candidate final output.
3. A capability validates that candidate against the current agent output schema.
4. If it validates, the run ends immediately with that processed final output.
5. The transcript still records the Code Mode tool return, but there is no later model output-tool call or text response.

If the candidate does not validate, the graph should continue normally.

## Current gap

`AbstractCapability.before_output_process` receives `OutputContext`, so it can replace or normalize output cleanly. However, it only runs after the model has already produced a final output response.

`AbstractCapability.after_node_run` can return `End(FinalResult(...))`, so it can avoid the additional model turn. However, from the public capability API it does not expose a way to validate and process a candidate through the current output schema, output processor, and output validators.

The workaround requires private/internal access such as `agent._output_schema`, `agent._output_validators`, `run_output_process_hooks`, `FinalResult`, and graph `End`/node internals. That works, but it is a brittle boundary for a capability that otherwise fits the public hook model.

## Feature request

Please expose a public capability-facing API for committing a candidate final output from a tool or node hook. For example, an API that can:

- validate a Python candidate against the current output schema
- run the same output processing path and output validators used by normal model output
- preserve message history for completed tool returns
- return or create the final result without another model request

The shape could be a helper on `RunContext`, the tool manager, or a public output pipeline object, for example conceptually:

```python
processed = await ctx.output.process_candidate(candidate)
return ctx.output.end(processed)
```

or a single helper that returns a final-result node/object when validation succeeds.

The important contract is that capabilities can finish a run from a validated tool result without requiring the model to restate the same object.

kazmer97:
- Closes #6243

## Summary

Adds `RunContext.output`, an `OutputController` that lets capabilities validate and commit a candidate final output using the agent's configured output schema, output functions, and output validators.

The controller can preserve pending `ToolReturnPart`s when it turns a `ModelRequestNode` into an `End`, which is the missing primitive for capabilities that produce a final answer as a tool result.

This also documents the capability pattern and covers valid, invalid, validator, output-function, and message-history behavior in focused tests.

## Validation

- `UV_CACHE_DIR=/tmp/uv-cache uv run ruff check pydantic_ai_slim/pydantic_ai/_output.py pydantic_ai_slim/pydantic_ai/output.py pydantic_ai_slim/pydantic_ai/_run_context.py pydantic_ai_slim/pydantic_ai/_agent_graph.py pydantic_ai_slim/pydantic_ai/__init__.py tests/test_capabilities.py`
- `PATH=/Users/kazmer.nagy-betegh/.nvm/versions/node/v24.14.1/bin:$PATH UV_CACHE_DIR=/tmp/uv-cache PYRIGHT_PYTHON_IGNORE_WARNINGS=1 uv run pyright pydantic_ai_slim/pydantic_ai/_output.py pydantic_ai_slim/pydantic_ai/output.py pydantic_ai_slim/pydantic_ai/_run_context.py pydantic_ai_slim/pydantic_ai/_agent_graph.py`
- `UV_CACHE_DIR=/tmp/uv-cache uv run pytest tests/test_capabilities.py::test_run_context_output_try_commit_candidate_preserves_tool_return tests/test_capabilities.py::test_run_context_output_candidate_runs_output_function_and_validator tests/test_capabilities.py::test_run_context_output_try_commit_candidate_invalid_continues_normally`
- `git diff --check`

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6247?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


DouweM:
Marking as draft per https://github.com/pydantic/pydantic-ai/issues/6243#issuecomment-4879916248.

DouweM:
Superseded by https://github.com/pydantic/pydantic-ai/pull/6257

PR #6252, opened 2026-07-03
Retain dropped usage fields across OpenAI and xAI streaming chunks

### LINKED ISSUE #5889: Verify streaming usage retention across cumulative-snapshot chunks in OpenAI + xAI models
### Background

#5886 (closes #5205) fixed a streaming usage-retention bug in `GoogleModel`: `GeminiStreamedResponse` overwrote `self._usage` on every chunk, so when a later cumulative-snapshot usage chunk dropped a field an earlier chunk had carried (e.g. `cached_content_token_count` through a gateway), the typed `RequestUsage` field (`cache_read_tokens`) was zeroed. The fix threads `existing_usage` through `_metadata_as_usage`, unions the `details` dict, and backfills any typed token field the latest snapshot dropped.

The same overwrite-then-rebuild shape exists in two other streamed-response paths, where the per-chunk `details` dict is built with `exclude_none` / conditional includes. If a provider ever drops a previously-reported sub-field on a later cumulative chunk, those paths would zero it the same way. Neither is confirmed to reproduce, so this issue is to verify the behavior before deciding whether a parity fix is warranted.

### Surfaces to verify

**OpenAI (`models/openai.py`)** — `OpenAIStreamedResponse._get_event_iterator`, `openai_continuous_usage_stats` branch, assigns `self._usage = chunk_usage` per chunk. `_map_usage` builds `details` via `model_dump(exclude_none=True)` plus conditionally added `completion_tokens_details`. Verify whether a continuous-usage stream can deliver `prompt_tokens_details.cached_tokens` or `completion_tokens_details.reasoning_tokens` on an intermediate chunk and omit it on a later one. Suggested check: record a continuous-usage stream where an intermediate chunk reports `reasoning_tokens=N` and the final usage chunk omits it, then assert `result.usage().details['reasoning_tokens']` is retained.

**xAI (`models/xai.py`)** — `XaiStreamedResponse._update_response_state`, assigns `self._usage = _extract_usage(...)` per chunk. `_extract_usage` only includes `reasoning_tokens` / `cached_prompt_text_tokens` / `server_side_tools_used` when truthy. Verify whether the xAI SDK can revert one of these to falsy on a later cumulative `Response`. Suggested check: instrument `_extract_usage` across a reasoning/cached xAI stream and confirm whether the optional fields are monotonic.

If either reproduces, the fix is the same shape as #5886 (retain dropped fields across chunks); if confirmed monotonic, no change is needed and we can document that.

### Out of scope

`models/gemini.py` (legacy `GeminiModel`) has its own `_metadata_as_usage` with the same per-chunk overwrite, but it is a deprioritized surface and assigns usage only from the final buffered snapshot, so its exposure window is narrower. Mentioned for completeness; not part of this verification.

Related: #5205, #5886.

SuperMarioYL:
- Closes #5889

## Summary

#5886 fixed a streaming usage-retention bug in `GoogleModel`: the streamed response overwrote `self._usage` on every cumulative snapshot, so when a later chunk dropped a sub-field an earlier chunk had reported (e.g. cached tokens through a gateway), the typed field was zeroed. #5889 asks to verify the same overwrite-then-rebuild shape in the OpenAI and xAI streamed paths.

Both share the defect:

- `OpenAIStreamedResponse` (with `openai_continuous_usage_stats`) does `self._usage = chunk_usage`
- `XaiStreamedResponse._update_response_state` does `self._usage = _extract_usage(...)`

Because each per-chunk usage is built with `exclude_none`/conditional includes, a field reported on an earlier cumulative chunk (e.g. `reasoning_tokens` or `cache_read_tokens`) is lost when a later snapshot omits it.

## Change

- Add a small provider-agnostic `merge_cumulative_usage` helper in `usage.py` that keeps the latest cumulative snapshot but backfills any typed token field or `details` entry the latest snapshot dropped, from the usage accumulated so far. Safe because cumulative snapshots are monotonic — a real drop to zero never overwrites a prior non-zero.
- Use it in the OpenAI `continuous_usage_stats` branch and in xAI's `_update_response_state`.
- Add a regression test for each provider: an intermediate cumulative chunk reports `reasoning_tokens` and cached tokens; later chunks drop them; the test asserts both survive. Each is red on `main` and green with the fix.

## Notes

The regression tests demonstrate the data loss with synthetic cumulative chunks — mirroring #5886's own tests (a later chunk with `cached=None`) rather than a live recording, since the field-drop is a defensive branch a recorded stream doesn't reliably produce. The fix is the same shape #5886 applied to `GoogleModel`.

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6252?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/openai.py
+++ b/pydantic_ai_slim/pydantic_ai/models/openai.py
@@ -100,6 +100,7 @@ from ..profiles.openai import (
 from ..providers import Provider, infer_provider
 from ..settings import ModelSettings
 from ..tools import AgentDepsT, ToolDefinition
+from ..usage import merge_cumulative_usage
 from . import (
     Model,
     ModelRequestContext,
@@ -3178,8 +3179,9 @@ class OpenAIStreamedResponse(StreamedResponse):
                 chunk_usage = self._map_usage(chunk)
                 if self._model_settings and self._model_settings.get('openai_continuous_usage_stats'):
                     # When continuous_usage_stats is enabled, each chunk contains cumulative usage,
-                    # so we replace rather than increment to avoid double-counting.
-                    self._usage = chunk_usage
+                    # so we replace rather than increment to avoid double-counting. Merge with the usage
+                    # so far to retain any field a later cumulative chunk dropped (#5889, same shape as #5886).
+                    self._usage = merge_cumulative_usage(self._usage, chunk_usage)
                 else:
                     self._usage += chunk_usage
 
--- a/pydantic_ai_slim/pydantic_ai/models/xai.py
+++ b/pydantic_ai_slim/pydantic_ai/models/xai.py
@@ -58,7 +58,7 @@ from ..profiles.grok import GrokModelProfile, GrokReasoningEffort
 from ..providers import Provider, infer_provider
 from ..settings import ModelSettings, ThinkingLevel
 from ..tools import ToolDefinition
-from ..usage import RequestUsage
+from ..usage import RequestUsage, merge_cumulative_usage
 from ._tool_choice import resolve_tool_choice
 
 try:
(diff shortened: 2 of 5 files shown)


DouweM:
@dsfaccini Assigning to you as the author of the issue.

dsfaccini:
David's AICA here: We checked whether the OpenAI and xAI streaming paths actually lose usage fields the way the Google path did in #5886 — against real API responses, not just the code shape. Result: OpenAI and xAI never drop a usage sub-field once it's been reported on an earlier chunk; the values only accumulate. So the overwrite this guards against isn't currently triggerable by these providers — the Google case was only reachable because a gateway stripped a field mid-stream. The added tests exercise the `merge_cumulative_usage` helper against a hand-built stream that drops a field, but the real APIs don't emit that shape.

Since we can't reproduce the underlying bug with a real request, we're closing #5889 as not-currently-reproducible, and this PR with it. If a real drop surfaces later (e.g. through a gateway, as happened for Google), we'll reopen the question. Thanks for the contribution, @SuperMarioYL.

PR #6279, opened 2026-07-05
fix(agent): honor `end_strategy="early"` for `NativeOutput`

### LINKED ISSUE #6277: `NativeOutput` with `end_strategy="early"` still executes tool calls
### Initial Checks

### Description

When using `NativeOutput` (native structured output) with `Agent(end_strategy="early")`, the agent still executes tool calls even when a valid structured output is already available in the model response.

## Expected behavior

With `end_strategy="early"`, once a final result is available (the native structured output in the text portion of the response), function tool calls should be skipped — matching the behavior for output tools.

## Actual behavior

Tool calls are always executed first, regardless of `end_strategy`, because the code unconditionally prioritizes tool calls over text output.

## Root cause

Location: `pydantic_ai/_agent_graph.py`, `CallToolsNode._run_stream()` (inner `_run_stream` function)

```python
# Lines ~1277-1286
if tool_calls:
        yield event
    return
elif output_schema.toolset:
# ... later, text is processed
```

The issue is that:
1. When using `NativeOutput`, the structured output appears in the **text** portion of the response (accumulated from `TextPart`s)
2. The code checks `if tool_calls:` first and unconditionally processes them
3. The `final_result` check inside `process_tool_calls` (line ~1622) only handles **output tools** (`kind='output'`), not the case where output comes from native structured output mode
4. The text processing that would extract the native output happens **after** tool calls are handled

## Reproduction scenario

When a model (e.g., Claude, GPT with native structured output) returns a response containing both:
1. Valid native structured output in `TextPart` 
2. Function tool calls

The tool calls will be executed even with `end_strategy="early"`, when they should be skipped since the final result is already available.

## Suggested fix

Before processing tool calls, check if:
1. The output mode is `native` 
2. `end_strategy == "early"`
3. Valid text content exists that can be parsed as the native output

If all conditions are met, process the native output as the final result and skip tool execution.

```python
if tool_calls:
    # Check if we have valid native output that should be treated as final result
    # before executing any tool calls (respects end_strategy='early')
    if (
        text
        and output_schema.mode == 'native'
        and ctx.deps.end_strategy == 'early'
        and (text_processor := output_schema.text_processor)
    ):
        try:
            self._next_node = await self._handle_text_response(ctx, text, text_processor)
            return  # Skip tool calls - native output is the final result
        except ToolRetryError:
            pass  # Native output validation failed, fall through to tool calls
        yield event
    return
```

```Python

import asyncio
from pydantic_ai import Agent
from pydantic_ai.models.function import FunctionModel, AgentInfo
    ModelMessage,
    ModelResponse,
    TextPart,
    ToolCallPart,
)


class Response(BaseModel):
    topic: str
    response: str


class Clarification(BaseModel):
    question: str


# Track if tool was called
tool_was_called = False
request_count = 0


def search_documents(query_paraphrases: list[str]) -> str:
    """Search for documents."""
    global tool_was_called
    tool_was_called = True
    print(f"  TOOL CALLED with: {query_paraphrases}")
    print("  ^^^ This should NOT happen with end_strategy='early' when output is already available")
    return "some documents"


def model_function(messages: list[ModelMessage], info: AgentInfo) -> ModelResponse:
    """
    Simulates a model that returns BOTH structured output AND a tool call on first request.
    On subsequent requests (after tool result), returns only structured output.
    """
    global request_count
    request_count += 1

    # First request: return both output AND tool call (this is where the bug manifests)
    if request_count == 1:
        return ModelResponse(
            parts=[
                # Native structured output - this IS the final result
                TextPart(
                    content='{"result": {"kind": "Response", "data": {"topic": "contact info", "response": "You can contact us at support@example.com"}}}'
                ),
                # Tool call - should be SKIPPED with end_strategy="early" since we already have output
                ToolCallPart(
                    tool_name="search_documents",
                    args={"query_paraphrases": ["contact information", "how to reach support"]},
                    tool_call_id="tooluse_abc123",
                ),
            ]
        )
    
    # Subsequent requests: return only structured output (no tool call)
    return ModelResponse(
        parts=[
            TextPart(
                content='{"result": {"kind": "Response", "data": {"topic": "contact info", "response": "You can contact us at support@example.com"}}}'
            ),
        ]
    )


agent = Agent(
    model=FunctionModel(model_function),
    output_type=NativeOutput(Response | Clarification),
    end_strategy="early",  # Should stop as soon as output is available
    tools=[search_documents],
)


async def main():
    global tool_was_called, request_count
    tool_was_called = False
    request_count = 0

    print("=" * 70)
    print("Testing: NativeOutput + end_strategy='early' + tool calls in response")
    print("=" * 70)
    print()
    print("Model will return BOTH:")
    print("  1. Valid structured output (Response) in TextPart")
    print("  2. A tool call (search_documents)")
    print()
    print("Expected: Tool should NOT be called (end_strategy='early' + output available)")
    print("Actual:   Tool IS called (bug)")
    print()
    print("-" * 70)

    result = await agent.run("How can I contact support?")

    print("-" * 70)
    print()
    print(f"Tool was called: {tool_was_called}")
    print()

    if tool_was_called:
        print("BUG CONFIRMED: Tool was called despite end_strategy='early'")
        print("               and valid native output being available.")
    else:
        print("OK: Tool was correctly skipped.")


if __name__ == "__main__":
    asyncio.run(main())
```

### Logfire Trace

_No response_

- Python: 3.14
- Pydantic AI: `main`
- LLM provider SDK: Bedrock

Oxygen56:
- Closes #6277

### Summary

This fixes `NativeOutput` handling when `end_strategy="early"` is used and the model response contains both:

- valid native structured output text
- regular tool calls

Before this change, `CallToolsNode` prioritized the tool calls and executed them before considering the native output text. With `end_strategy="early"`, valid native output should be enough to finish the run without executing additional regular tools.

### Changes

- In `CallToolsNode`, when tool calls and text are both present, attempt to process valid native output text first if `end_strategy="early"` is configured.
- Preserve skipped tool returns in message history when native output wins, so later runs do not inherit dangling tool calls.
- If native output parsing/validation fails, keep the existing behavior and fall through to tool-call handling.
- Add regression coverage for both valid native output winning over tool calls and invalid native output falling back to tool-call handling.

### Testing

- `uv run pytest tests/test_agent.py::TestMultipleToolCalls::test_early_strategy_prefers_native_output_text_over_tool_calls tests/test_agent.py::TestMultipleToolCalls::test_early_strategy_falls_back_to_tools_when_native_output_text_is_invalid tests/test_agent.py::TestMultipleToolCalls`
- `PYRIGHT_PYTHON_IGNORE_WARNINGS=1 uv run pyright pydantic_ai_slim/pydantic_ai/_agent_graph.py tests/test_agent.py`
- `uv run ruff check pydantic_ai_slim/pydantic_ai/_agent_graph.py tests/test_agent.py`
- `uv run ruff format --check pydantic_ai_slim/pydantic_ai/_agent_graph.py tests/test_agent.py`
- `env -u HTTP_PROXY -u HTTPS_PROXY -u ALL_PROXY -u http_proxy -u https_proxy -u all_proxy uv run pytest tests/test_agent.py`

Note: running the full `tests/test_agent.py` with my local SOCKS proxy env vars enabled failed in provider lifecycle tests because `httpx` required the optional `socksio` dependency. Re-running the same test file with proxy env vars removed passed: `344 passed, 3 skipped`.

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6279?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/_agent_graph.py
+++ b/pydantic_ai_slim/pydantic_ai/_agent_graph.py
@@ -1294,7 +1294,7 @@ class CallToolsNode(AgentNode[DepsT, NodeRunEndT]):
                             async for event in self._handle_tool_calls_after_final_result(
                                 ctx, tool_calls, final_result, output_parts
                             ):
-                                yield event
+                                yield event  # pragma: no cover
                             self._next_node = self._handle_final_result(ctx, final_result, output_parts)
                             return
                         except ToolRetryError:
@@ -1416,29 +1416,16 @@ class CallToolsNode(AgentNode[DepsT, NodeRunEndT]):
 
         ctx.deps.tool_manager = await ctx.deps.tool_manager.for_run_step(run_context)
 
-        try:
-            async for event in process_tool_calls(
-                tool_manager=ctx.deps.tool_manager,
-                tool_calls=tool_calls,
-                tool_call_results=self.tool_call_results,
-                tool_call_metadata=self.tool_call_metadata,
-                final_result=final_result,
-                ctx=ctx,
-                output_parts=output_parts,
-            ):
-                yield event
-        except BaseException:
-            if output_parts:
-                ctx.state.message_history.append(
-                    _messages.ModelRequest(
-                        parts=list(output_parts),
-                        run_id=ctx.state.run_id,
-                        conversation_id=ctx.state.conversation_id,
-                        timestamp=now_utc(),
-                        state='interrupted',
-                    )
-                )
-            raise
+        async for event in process_tool_calls(
+            tool_manager=ctx.deps.tool_manager,
+            tool_calls=tool_calls,
+            tool_call_results=self.tool_call_results,
+            tool_call_metadata=self.tool_call_metadata,
+            final_result=final_result,
+            ctx=ctx,
+            output_parts=output_parts,
+        ):
+            yield event  # pragma: no cover
 
     @staticmethod
     def _recover_text_from_message_history(message_history: list[_messages.ModelMessage]) -> str | None:
--- a/pydantic_ai_slim/pydantic_ai/_agent_graph.py
+++ b/pydantic_ai_slim/pydantic_ai/_agent_graph.py
@@ -1290,15 +1290,13 @@ class CallToolsNode(AgentNode[DepsT, NodeRunEndT]):
                     ):
                         try:
                             final_result = await self._process_text_response(ctx, text, text_processor)
-                            output_parts: list[_messages.ModelRequestPart] = []
-                            async for event in self._handle_tool_calls_after_final_result(
-                                ctx, tool_calls, final_result, output_parts
-                            ):
-                                yield event  # pragma: no cover
-                            self._next_node = self._handle_final_result(ctx, final_result, output_parts)
-                            return
                         except ToolRetryError:
                             pass
+                        else:
+                            # Record the tool calls as skipped so history has no dangling calls.
+                            async for event in self._handle_tool_calls(ctx, tool_calls, final_result):
+                                yield event
+                            return
 
                     if tool_calls:
                         async for event in self._handle_tool_calls(ctx, tool_calls):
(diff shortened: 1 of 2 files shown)


Oxygen56:
Addressed the review feedback in 6e554f99: the native early-exit path now records skipped tool returns by reusing `process_tool_calls(..., final_result=...)`, and the regression test asserts the skipped `ToolReturnPart` is present in `result.all_messages()`.

hramezani (approved):

PR #6282, opened 2026-07-05
Give a clear error for non-positive concurrency `max_running`

### LINKED ISSUE #6281: Concurrency limiter should reject non-positive max_running
`ConcurrencyLimit` and the `max_running` argument on `ConcurrencyLimiter` accept `0` and negative values, which have no useful meaning: `None` already means no limiting, and positive integers mean the number of concurrent operations allowed.

`anyio.CapacityLimiter` already rejects these at construction (`total_tokens must be >= 1`), so `Agent(..., max_concurrency=0)` raises rather than hanging. Two things are worth improving:

- the surfaced error is anyio's internal `total_tokens must be >= 1`, which does not mention `max_running`, `max_concurrency`, or the `None` opt-out, and
- `ConcurrencyLimit(max_running=0)` is constructed silently and only fails later when it is turned into a limiter.

Expected: reject `max_running < 1` eagerly at construction with a clear `UserError`, consistent with the `OnlineEvaluator.max_concurrency` check in #6267.

VectorPeak:
- Closes #6281

Eagerly reject non-positive concurrency `max_running` with a clear `UserError` at construction, across every path: `ConcurrencyLimit`, `ConcurrencyLimiter`, `from_limit`, `normalize_to_limiter`, and `Agent(..., max_concurrency=...)`. `None` remains the explicit no-limit value. Wording matches the `OnlineEvaluator.max_concurrency` validation from #6267.

## Why

Since anyio 4.12, `CapacityLimiter(0)` is allowed by design (agronholm/anyio#1019 — `total_tokens` is a mutable pause/resume "valve"; allowed in sync construction too since 4.14, agronholm/anyio#1183). pydantic-ai never mutates the limiter, so `Agent(..., max_concurrency=0)` constructs silently and the first `run()` hangs forever — reproduced on `main` with anyio 4.13. On anyio <= 4.11 the failure is anyio's opaque `total_tokens must be >= 1` instead. This PR turns both into an immediate, actionable error, uniform across the supported range (`anyio>=4.5.0`).

## Tests

`tests/test_concurrency.py` covers `0` and negatives through every path above, plus `None` still meaning no limiting.

### Checklist


dsfaccini:

|---|---|---|---|---|

**Verdict:** ✅ all 3 waves passed — 0 blocking, 0 required (8 reviewers ran; independent runtime probe confirmed the bug and the fix)
Reviewed at 2026-07-05T20:34Z · diff `27bfa1d07fc3`

<details>
<summary>⚠️ 3 informational (skip / maintainer-decision)</summary>

**1. Exception type + message differ from evals siblings** · `pydantic_ai_slim/pydantic_ai/concurrency.py` · `_validate_max_running`
This PR raises `UserError` (`'max_running must be a positive integer, got N. Use None for no concurrency limiting.'`); the sibling guards in `pydantic_evals` — #6228 (merged) and #6267 (open) — raise `ValueError` (`'max_concurrency must be >= 1, got N'`), and issue #6281 requested `ValueError`. The precedent reviewer's verdict: **not a defect — correct per-package idiom.** `UserError` is `pydantic_ai_slim`'s convention for developer-misconfiguration; `pydantic_evals` doesn't import `UserError` at all. Deliberate per commit `d33c6099f`. Flagged as a maintainer alignment call, not changed.

**2. `"Use None"` remedy not literally actionable at the `ConcurrencyLimiter.__init__` boundary** · `pydantic_ai_slim/pydantic_ai/concurrency.py` · `_validate_max_running`
The message is accurate at the outer API (`Agent(max_concurrency=None)`, `normalize_to_limiter(None)`) but `ConcurrencyLimiter(max_running: int)` can't accept `None`. Minor wording — skip.

**3. `max_queued` (0 / negative) left unvalidated** · `pydantic_ai_slim/pydantic_ai/concurrency.py` · `ConcurrencyLimit`
Pre-existing, outside this PR's reproduced `max_running` scope; `max_queued=0` may be legitimate intent. Correct narrow-fix scoping — skip (own issue if ever pursued). Maintainer already documented this as out of scope.
</details>

<details>
<summary>💡 1 optional docs nit (skip-safe)</summary>

**Docstrings don't state the positive-integer constraint** · `pydantic_ai_slim/pydantic_ai/agent/__init__.py` (`max_concurrency`) + `concurrency.py` (`ConcurrencyLimit.max_running`, `ConcurrencyLimiter.__init__`)
The three docstrings for this param describe it but don't say the integer must be positive / that `0` is rejected. One-line docstring additions would close it; the docs reviewer rates it skip-safe (the eager `UserError` self-documents). Not applied here — this review ran read-only while another maintainer edit was in flight; left for that pass.
</details>

dsfaccini:
David's AICA here: thanks for the fix (and for reporting the issue). I pushed one commit to this branch as a maintainer edit with three small refinements:

- **`ValueError` -> `UserError`.** `UserError` is pydantic-ai's convention for a developer-misconfiguration of a public API surface (see the model / output-type validations in `agent/__init__.py` and `_output.py`). A bare `ValueError` reads as an internal error, and it's also what users already get opaquely from anyio today -- the point of this fix is to replace that with a typed, self-explanatory pydantic-ai error.
- **Message points to the remedy.** Now `max_running must be a positive integer, got {n}. Use None for no concurrency limiting.` -- naming `None` addresses the exact 0-vs-None confusion the issue calls out.
- **Regression test** that `None` still means "no limiting" (via `normalize_to_limiter(None)` and `Agent(max_concurrency=None)`), plus a `-5` case.

The shared `_validate_max_running` helper and the two guard sites (`ConcurrencyLimit.__post_init__` + `ConcurrencyLimiter.__init__`) were already right -- every live-limiter path funnels through `__init__`, and the `__post_init__` guard fail-fasts a standalone config. Lint, pyright, and `tests/test_concurrency.py` (44) pass locally.

Note: `max_queued` (0 / negative) is unvalidated and remains a separate latent gap -- intentionally left out of scope here, since it's a backpressure-semantics decision rather than part of this hang fix.

VectorPeak:
@dsfaccini Thanks for the review and the follow-up commit. The `UserError` change makes sense to me. This is a bad user config value, not really an internal limiter error.

For future PRs, is there already a short doc or rule for this kind of case? For example: if a public pydantic-ai API gets a config value that would create invalid internal state, should we fail early with `UserError`?

If there is no rule yet, would it be useful to write it down somewhere small, maybe in the contributing docs, a simple check where possible, or the repo context used by AI assistants? That would make future PRs easier to keep consistent.

Since the PR is green now and CodeRabbit is happy, do you think it is ready to merge?

adtyavrdhn:
anyio rejects it anyway so I don't think this is needed apart from maybe a friendlier error message?

VectorPeak:
@adtyavrdhn Thanks a lot for taking a look, and for tightening the wording/docstrings here. I agree with your framing that this is mainly about giving a clearer public API error instead of letting lower-level limiter behavior leak through.

I rechecked one detail locally though: `anyio.CapacityLimiter(0)` does not seem to reject at construction time. With real AnyIO, I get:

```text
CapacityLimiter(0) constructed: total_tokens= 0 available_tokens= 0
CapacityLimiter(-1) construct ERROR: ValueError total_tokens must be >= 0
CapacityLimiter(1) constructed: total_tokens= 1 available_tokens= 1
```

So negative values are rejected by AnyIO, but `0` appears to create a zero-capacity limiter. A later `acquire()` then waits because no token can become available.

Could you reproduce a different result on your side? If not, I think the `>= 1` check you added is the right public API boundary: positive integers mean a bounded concurrency limit, and `None` is the explicit no-limit value.

Thanks again for the maintainer edits. CI is green now after the formatting fix, and I don't think there is anything else I need to change on my side.

PR #6289, opened 2026-07-06
fix(azure): disable encrypted reasoning content for Azure OpenAI

### LINKED ISSUE #2915: Encrypted content is not supported for GPT-5 though Azure
### Initial Checks

### Description

when running gpt-5 through Azure you get the following error 

```
ModelHTTPError: status_code: 400, model_name: gpt-5, body: {'message': 'Encrypted content is not supported with this model.', 'type': 'invalid_request_error', 'param': 'include', 'code': None}
```

### Example Code

```Python
import os

from openai import AsyncAzureOpenAI
from pydantic_ai import Agent, NativeOutput


class CalendarEvent(BaseModel):
    name: str
    date: str
    participants: list[str]


client = AsyncAzureOpenAI(
    azure_endpoint=os.environ["AZURE_OPENAI_ENDPOINT"],
    api_version=os.environ["OPENAI_API_VERSION"],
    api_key=os.getenv("AZURE_OPENAI_API_KEY"),
)

model = OpenAIResponsesModel(
    model_name="gpt-5",
    provider=OpenAIProvider(openai_client=client),
)

agent = Agent(
    model=model,
    tools=[],
    system_prompt="Extract the event information.",
    output_type=NativeOutput(CalendarEvent),
)
result = await agent.run(
    [
        "Alice and Bob are going to a science fair on Friday.",
    ]
)
```

```Text
python -V
Python 3.13.7

import pydantic_ai
pydantic_ai.__version__
'1.0.2'
```

Bartok9:
## Summary

Disables the `include=['reasoning.encrypted_content']` request option for Azure OpenAI, which the provider rejects.

## Motivation

Closes #2915.

Running a reasoning model (e.g. `gpt-5`) through Azure OpenAI with `OpenAIResponsesModel` fails with:

```
ModelHTTPError: status_code: 400, body: {'message': 'Encrypted content is not supported with this model.', 'type': 'invalid_request_error', 'param': 'include', 'code': None}
```

### Root cause

`AzureProvider.model_profile` derives OpenAI model profiles from `openai_model_profile`, which sets `openai_supports_encrypted_reasoning_content=True` for reasoning models (gpt-5, o-series). The Responses model then always requests encrypted reasoning content via `include`, and Azure OpenAI rejects it with a 400.

The reporter confirmed the failure occurs on **both** the legacy `AsyncAzureOpenAI` client and the new v1 GA API (`.../openai/v1`), and that non-reasoning models (gpt-4.1) work fine — so this is Azure-wide for reasoning models, not client-specific.

### Fix

Override `openai_supports_encrypted_reasoning_content=False` in `AzureProvider.model_profile`, alongside the existing `openai_chat_supports_document_input=False` Azure override. This matches the per-model workaround @DouweM suggested in the issue thread.

### Scope / intended resolution

The reproductions in #2915 use `OpenAIProvider` with an Azure client (or `.../openai/v1` base URL). The intended resolution is to use **`AzureProvider`**, which with this change disables encrypted reasoning content out of the box (no manual profile override needed). `OpenAIProvider.model_profile` is a `@staticmethod` that only receives `model_name` and can't inspect the base URL to detect Azure, so covering the raw `OpenAIProvider`+Azure path would require a larger design change and can be a follow-up if preferred.

## Tests

- Extended `test_azure_provider_model_profile` to assert `openai_supports_encrypted_reasoning_content is False` for a reasoning model (`o4-mini`), where the base OpenAI profile sets it to `True`.

## Verification

- `uv run pytest tests/providers/test_azure.py` — 12 passed
- `uv run ruff check` / `ruff format --check` — clean
- `uv run mypy pydantic_ai_slim/pydantic_ai/providers/azure.py` — no issues

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6289?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/providers/azure.py
+++ b/pydantic_ai_slim/pydantic_ai/providers/azure.py
@@ -76,8 +76,16 @@ class AzureProvider(Provider[AsyncOpenAI]):
             # OpenAI models are unprefixed.
             base = openai_model_profile(model_name)
 
-        # Azure Chat Completions API doesn't support document input.
-        return merge_profile(base, OpenAIModelProfile(openai_chat_supports_document_input=False))
+        # Azure Chat Completions API doesn't support document input, and Azure OpenAI rejects the
+        # `include=['reasoning.encrypted_content']` request option that reasoning models otherwise send
+        # (see https://github.com/pydantic/pydantic-ai/issues/2915).
+        return merge_profile(
+            base,
+            OpenAIModelProfile(
+                openai_chat_supports_document_input=False,
+                openai_supports_encrypted_reasoning_content=False,
+            ),
+        )
 
     @overload
     def __init__(self, *, openai_client: AsyncAzureOpenAI) -> None: ...
--- a/tests/profiles/test_resolution_matrix.py
+++ b/tests/profiles/test_resolution_matrix.py
@@ -926,7 +926,6 @@ def test_azure_openai_gpt_5():
             'supported_native_tools': frozenset(
                 {CodeExecutionTool, FileSearchTool, ImageGenerationTool, MCPServerTool, ToolSearchTool, WebSearchTool}
             ),
-            'openai_supports_encrypted_reasoning_content': True,
             'openai_supports_reasoning': True,
             'openai_supports_reasoning_effort_none': True,
             'openai_supports_phase': True,
--- a/tests/providers/test_azure.py
+++ b/tests/providers/test_azure.py
@@ -150,6 +150,9 @@ def test_azure_provider_model_profile(mocker: MockerFixture):
     openai_model_profile_mock.assert_called_with('o4-mini')
     assert openai_profile is not None
     assert openai_profile.get('json_schema_transformer', None) == OpenAIJsonSchemaTransformer
+    # Azure OpenAI rejects `include=['reasoning.encrypted_content']`, so even reasoning models
+    # must not send it. See https://github.com/pydantic/pydantic-ai/issues/2915.
+    assert openai_profile.get('openai_supports_encrypted_reasoning_content', None) is False
 
     unknown_profile = provider.model_profile('unknown-model')
     openai_model_profile_mock.assert_called_with('unknown-model')


hramezani:
Thanks for the fix! The change and tests look correct for the `AzureProvider` path.

One thing to confirm: both reproductions in #2915 use `OpenAIProvider` (with an `AsyncAzureOpenAI` client / an `.../openai/v1` base URL), not `AzureProvider`. Since this fix only touches `AzureProvider.model_profile`, that reported code would still hit the 400 unless the user switches to `AzureProvider`.

Could you either:
- confirm that pointing people to `AzureProvider` is the intended resolution (and note it in the PR description), or
- extend the fix to also cover the `OpenAIProvider` + Azure path?

Just want to make sure the reporter's exact scenario is covered. Thanks!

Bartok9:
Thanks @hramezani, good catch — you're right that the reproductions use `OpenAIProvider` (Azure client / `/openai/v1` base URL), so this PR alone doesn't cover that exact path.

Confirming **option 1**: pointing Azure users to `AzureProvider` is the intended resolution, and with this PR it works out of the box (no manual profile override needed). I've updated the PR description to state this explicitly.

On extending the fix to the `OpenAIProvider` + Azure path directly: `OpenAIProvider.model_profile` is a `@staticmethod` that only receives `model_name`, so it can't inspect the client/base_url to detect an Azure endpoint. Auto-detecting Azure there (e.g. sniffing `*.openai.azure.com` / `*.cognitiveservices.azure.com` / `.../openai/v1` in the base URL) would mean turning it into an instance method and reconciling the class-level call in `infer_model_profile` — a bigger design change I didn't want to make unilaterally. Happy to take that on in a follow-up if you'd prefer that route over steering Azure users to `AzureProvider`.

dsfaccini:

|---|---|---|---|---|
| 1 · code-correctness | ✅ closed | 0 | 0 | 5 |

Reviewed at 2026-07-06 · diff `b9efb759`

<details>
<summary>⚠️ 5 informational (defer/skip)</summary>

**1. Scope: fix covers `AzureProvider` only, not the reporter's `OpenAIProvider`+Azure path** · `providers/azure.py` · `AzureProvider.model_profile`
Both #2915 reproductions reach Azure via `OpenAIProvider` (Azure client / `.../openai/v1` base URL), so the reporter's exact code still 400s unless they switch to `AzureProvider`. — skip: already raised by @hramezani and resolved in-thread (option 1: steer Azure users to `AzureProvider`, works out of the box); `OpenAIProvider.model_profile` is a `@staticmethod(model_name)` that can't sniff the base URL, so covering it needs an instance-method rework — a deliberate, documented follow-up.

**2. Breadth: disables encrypted reasoning for ALL Azure reasoning models, only gpt-5 confirmed** · `providers/azure.py` · `AzureProvider.model_profile`
o-series (o1/o3/o4-mini) on Azure was never empirically tested. — skip (deliberate scope boundary): disabling an *optional* feature is the safe default; Azure's error is model-level ("this model"); the narrower verified-only scope isn't reachable from `model_profile`; and @DouweM's "maybe only last-gen API" narrowing was disproven in-thread (reporter confirmed the `.../openai/v1` next-gen API also 400s), so Azure-wide is the justified breadth.

**3. No end-to-end VCR test that the Azure request omits `include=['reasoning.encrypted_content']`** · `tests/providers/test_azure.py`
— skip (net-neutral): recording needs Azure credentials a contributor lacks, and the cassette matchers here aren't request-body-sensitive anyway. The mocker-based profile-flag assertion (`is False`) plus the pre-existing gate at `models/openai.py` (`if profile.get('openai_supports_encrypted_reasoning_content')`) together demonstrate the behavior — the documented no-Azure-creds unit-test carve-out.

**4. No docs note for the new Azure limitation** · `docs/`
The sibling override `openai_chat_supports_document_input=False` (#4048) carried a doc note. — skip: that sibling is user-facing (pass a document → it fails); this fix is transparent (an internal multi-turn reasoning optimization, no user knob, no gotcha, restores working behavior). Docs' own conventions say to omit implementation details with no user-observable effect.

**5. Forward-compat staleness if Azure fixes their 400** · `providers/azure.py` · `AzureProvider.model_profile`
If Azure later accepts the `include` option, Azure users silently stop getting encrypted-reasoning continuity with no error. — skip: mitigated by the code comment + issue link, and the profile is user-overridable; not a present defect.

</details>

<sub>Independent verification: `AzureProvider.model_profile` now yields `openai_supports_encrypted_reasoning_content=False` for reasoning models (o4-mini/gpt-5/gpt-5.4) vs `True` in the base `openai_model_profile` — confirmed by profile probe. The live Azure 400 was NOT reproduced here (no Azure creds); live behavior rests on the #2915 reporter's confirmation on both the legacy client and the `.../openai/v1` API.</sub>

Bartok9:
Thanks @hramezani! Confirming: **pointing people to `AzureProvider` is the intended resolution** for this PR. With this change, `AzureProvider` disables encrypted reasoning content out of the box, so the reporter's scenario is fixed by switching from `OpenAIProvider`+Azure-client to `AzureProvider` (which is the recommended way to target Azure anyway).

Covering the raw `OpenAIProvider`+Azure path directly isn't feasible without a larger change: `OpenAIProvider.model_profile` is a `@staticmethod` that only receives `model_name` and can't inspect the client/base URL to detect Azure. Auto-detecting Azure there (e.g. sniffing `azure.com` / `/openai/v1` in the base URL) would be a reasonable follow-up but is a separate design decision. Happy to open that as a follow-up PR if you'd prefer.

I've kept the PR description's 'Scope / intended resolution' section aligned with this.

dsfaccini:
David's AICA here: This disables `encrypted_content` only for the last-generation Azure path — `AzureProvider` builds an `AsyncAzureOpenAI` client — matching what you offered in #2915: disable it "only for the last gen API, and not for Azure OpenAI as a whole." @hramezani accepted that scope. Good to merge on that basis?

One thing this PR deliberately leaves out: the reporter's own repro used the **next-gen** path — `OpenAIProvider(base_url='…cognitiveservices.azure.com/openai/v1')` — where gpt-5 **still** returns `400 Encrypted content is not supported with this model` (gpt-4.1 is fine), and they suspect an Azure-side bug that'll get fixed. So the next-gen path stays uncovered, and whether to cover it depends on whether Azure v1 rejects this permanently or temporarily.

If you do want to cover it, the cleanest shape is a base_url-host check (`*.openai.azure.com` + `*.cognitiveservices.azure.com`) applied in `OpenAIResponsesModel.__init__` as a profile-flag override rather than at the `include` site — the flag also gates sending reasoning back on multi-turn (`_map_messages`), so it has to stay profile-level to keep both coherent. That would also subsume this PR's `AzureProvider` override. But it assumes "all Azure ⇒ off," which cuts against both your "not Azure-as-a-whole" note and the reporter's "they'll fix it" — so it's probably best held until the v1 behavior is pinned down. How do you want to handle the next-gen path?

Bartok9:
Thanks @dsfaccini — yes, good to merge on that basis. This intentionally scopes to the last-gen Azure path (`AzureProvider` → `AsyncAzureOpenAI`), matching exactly what I offered in #2915 ("only for the last-gen API, not Azure as a whole") and what @hramezani accepted.

On the next-gen path (`OpenAIProvider(base_url='…/openai/v1')`): I'd prefer to leave it uncovered here. Two reasons line up:

1. The reporter believes it's an Azure-side bug on v1 that they expect to be fixed, so a blanket "all Azure ⇒ off" override risks permanently disabling encrypted reasoning content for a path that may work correctly soon.
2. The clean shape you describe — a host-based flag override in `OpenAIResponsesModel.__init__` (so it stays profile-level and keeps `_map_messages` multi-turn behavior coherent) — is the right design *if* we decide to cover it, but it hardcodes the "all Azure ⇒ off" assumption that cuts against both my #2915 note and the reporter's expectation.

So: merge this for the `AzureProvider` fix now, and hold the next-gen coverage until the Azure v1 behavior is pinned down (permanent vs. transient). If it turns out permanent, I'm happy to follow up with the `__init__` host-check override you sketched, gated so it subsumes this PR's `AzureProvider` path cleanly. Sound good?

DouweM:
The issue is _very_ old so we don't know if this is still a current issue.

PR #6308, opened 2026-07-07
Fix `RetryPromptPart.content` round-trip when `ErrorDetails` entries omit `input`

### LINKED ISSUE #5987: [roundtrip-sweep] `ModelMessagesTypeAdapter`: `RetryPromptPart.content` list of error details fails to round-trip when any item lacks `input`
## Impact

Any code that round-trips message history containing a `RetryPromptPart` whose `content` is a list of `ErrorDetails`-shaped dicts missing the `input` key will silently produce a JSON payload that the framework **cannot reload**:

- `Agent.all_messages_json()` → `validate_json` raises `ValidationError`.
- `Temporal` / durable-exec workflow replay of an in-flight run that had a structured retry → workflow cannot resume.
- Persisted message history that survives a Pydantic version bump and is then re-fed as `message_history=` to a later run → permanent data loss.

The serialization step also emits a `PydanticSerializationUnexpectedValue` `UserWarning`, which the framework codebase has not silenced — every `dump_json()` of such a part hits the noise path.

The `pydantic_core.ErrorDetails` `TypedDict` makes every key (`type`, `loc`, `msg`, `input`, `ctx`, `url`) **required**; but the framework's own helpers build these lists with `error.errors(include_url=False, include_context=False)` — which already omits two of the required keys (`url`, `ctx`) — and `pydantic_core.ValidationError.errors(include_input=False)` is a documented public option. Third-party validators that build their own dicts (or that lift the dicts from a plain `ValueError` with structured `loc`/`msg`) routinely do not include `input` either. Once a list like that is fed into `RetryPromptPart`, the data is irrecoverable after a `dump_json` round-trip.

## Boundary & Code Path

`pydantic_ai_slim/pydantic_ai/messages.py` — `RetryPromptPart` is declared with `content: list[pydantic_core.ErrorDetails] | str`.

`ModelMessagesTypeAdapter` is built at module top with `defer_build=True` and is used (via `Agent.all_messages_json`, `result.new_messages_json`, and elsewhere) to (de)serialize `list[ModelMessage]`. Because the inner type is `list[ErrorDetails]`, the `validate_json` step enforces strict TypedDict shape, which makes the round-trip a *one-way door* the moment any element is missing a required key.

## Reproduction

A new test was added: `tests/test_retry_prompt_roundtrip.py::test_retry_prompt_round_trip_with_minimal_error_details`. It fails on the current `main`:

```python
    ModelMessagesTypeAdapter, ModelRequest, RetryPromptPart,
)

# Three-key error details, as produced by `error.errors(include_url=False, include_context=False, include_input=False)`
# or by a custom validator / older replay log.
content: list[dict[str, object]] = [
    {'type': 'missing', 'loc': ('x',), 'msg': 'Field required'},
]
part = RetryPromptPart(content=content, tool_name='foo', tool_call_id='call_1')  # type: ignore[arg-type]
messages = [ModelRequest(parts=[part])]

# Serializes (with PydanticSerializationUnexpectedValue warning)
serialized = ModelMessagesTypeAdapter.dump_json(messages)

# Deserialization FAILS — `input` is required by `ErrorDetails`
ModelMessagesTypeAdapter.validate_json(serialized)
```

Run with `uv run python -m pytest tests/test_retry_prompt_roundtrip.py -v`. Captured output:

```
Pydantic serializer warnings:
  PydanticSerializationUnexpectedValue: Expected 4 fields but got 3: ...
  PydanticSerializationUnexpectedValue: Expected `str` - serialized value may not be as expected ...

FAILED tests/test_retry_prompt_roundtrip.py::test_retry_prompt_round_trip_with_minimal_error_details
pydantic_core._pydantic_core.ValidationError: 2 validation errors for list[tagged-union[ModelRequest,ModelResponse]]
0.request.parts.0.retry-prompt.content.list[ErrorDetails].0.input
  Field required [type=missing, input_value={'type': 'missing', 'loc': ('x',), 'msg': 'Field required'}, input_type=dict]
0.request.parts.0.retry-prompt.content.str
  Input should be a valid string [type=string_type, input_value=[...], input_type=list]
```

The JSON payload written to disk is `{"content": [{"type": "missing", "loc": ["x"], "msg": "Field required"}], ...}` — a valid-looking, hand-editable list of error details — but the framework refuses to round-trip it back into a `RetryPromptPart`.

## Expected vs Actual

**Expected:** `validate_json(dump_json(messages)) == messages` for any `messages` that the dataclass type accepts on construction. The `content` field should round-trip whether each entry is the full `ErrorDetails` shape, the framework's `include_url=False, include_context=False` shape (no `url`/`ctx`), the `include_input=False` shape (no `input`), or a hand-built `{type, loc, msg}` dict.

**Actual:** Only the canonical `ErrorDetails` shape round-trips. Anything missing a required key fails hard at `validate_json` time and emits a warning at `dump_json` time. Once a list with a 3-key entry has been serialized, the message is unrecoverable from its JSON.

## Evidence

- `pydantic_ai_slim/pydantic_ai/messages.py:1403` — `error_details_ta = pydantic.TypeAdapter(list[pydantic_core.ErrorDetails], config=pydantic.ConfigDict(defer_build=True))`
- `pydantic_ai_slim/pydantic_ai/messages.py:1422` — `content: list[pydantic_core.ErrorDetails] | str` (the strict type that's the root cause)
- `pydantic_ai_slim/pydantic_ai/messages.py:2381` — `ModelMessagesTypeAdapter = pydantic.TypeAdapter(list[ModelMessage], ...)` (the boundary used by `Agent.all_messages_json` / `result.new_messages_json`)
- `pydantic_ai_slim/pydantic_ai/tool_manager.py:187` — framework call site that produces a 4-key list (`type/loc/msg/input`) and stores it as `content`
- `pydantic_ai_slim/pydantic_ai/_output.py:123` — same pattern, in output-tool validation
- `tests/test_retry_prompt_roundtrip.py` — new test (full code) demonstrating the round-trip failure

## Suggested Fix

Loosen the field's runtime type to `list[dict[str, Any]] | str` (keep the `ErrorDetails` shape as the documented ideal for tooling) so that:

1. Construction accepts dicts of any subset of the canonical keys (as it does today at the dataclass level — `pydantic_core.ErrorDetails` is a `TypedDict`, not enforced at runtime).
2. The `ModelMessagesTypeAdapter` (de)serialization round-trips all of those subsets.
3. The `PydanticSerializationUnexpectedValue` warning stops firing for the framework's own `error.errors(include_url=False, include_context=False)` output (which is missing `url` and `ctx`).

If stricter validation is desired, gate it behind a `model_config` / `model_validator` so users can opt in. The default — the path that runs for `all_messages_json` — must round-trip.

adtyavrdhn:
- Closes #5987

`RetryPromptPart.content` is typed `list[pydantic_core.ErrorDetails] | str`, and `ErrorDetails` requires the `input` key. Construction never validates (plain dataclass), and `ModelMessagesTypeAdapter.dump_json` serializes entries missing `input` with only a `UserWarning`, but `validate_json` then rejects the payload. Any history containing such a part (e.g. built with `ValidationError.errors(include_input=False)`, or with `input` redacted by a history processor) serializes fine and can never be loaded back, breaking `message_history=` reload and durable-exec replay.

Fix: validate and serialize `content` against a runtime `TypedDict` identical to `pydantic_core.ErrorDetails` except `input` is `NotRequired`. At type-check time the annotation stays `pydantic_core.ErrorDetails`, so the public static type, rendered API docs, and the `$defs.ErrorDetails` JSON schema name are unchanged; the schema only drops `input` from `required`. `type`/`loc`/`msg` stay required so non-`ErrorDetails` garbage still fails loudly on load instead of validating to empty dicts. `ToolRetryError._format_error_details` now renders entries without `input` instead of raising `KeyError` (reachable with a user-supplied `RetryPromptPart` via `DeferredToolResults`).

Deliberate trade-off: constructing partial entries still needs a `cast` because the static type keeps `input` required. Making it `NotRequired` statically would break existing typed user code that passes `part.content` to `list[pydantic_core.ErrorDetails]`-annotated functions (verified with pyright), so the loosening is runtime-only.

Framework-produced histories were never affected; both internal producers keep `input`. Tests cover the `include_input=False` shape, the framework shape, the full shape (JSON and python modes, `model_response()` warning-free), garbage rejection, and `ToolRetryError` formatting.

Supersedes #5990.

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6308?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/exceptions.py
+++ b/pydantic_ai_slim/pydantic_ai/exceptions.py
@@ -301,7 +301,9 @@ class ToolRetryError(Exception):
         for e in errors:
             loc = '.'.join(str(x) for x in e['loc']) if e['loc'] else '__root__'
             lines.append(loc)
-            lines.append(f'  {e["msg"]} [type={e["type"]}, input_value={e["input"]!r}]')
+            # `input` may be omitted, e.g. when the part was built with `errors(include_input=False)`.
+            input_value = f', input_value={e["input"]!r}' if 'input' in e else ''
+            lines.append(f'  {e["msg"]} [type={e["type"]}{input_value}]')
         return '\n'.join(lines)
 
 
--- a/pydantic_ai_slim/pydantic_ai/messages.py
+++ b/pydantic_ai_slim/pydantic_ai/messages.py
@@ -19,7 +19,7 @@ import pydantic
 import pydantic_core
 from genai_prices import calc_price, types as genai_types
 from pydantic.dataclasses import dataclass as pydantic_dataclass
-from typing_extensions import TypeAliasType, TypedDict, TypeVar, assert_never
+from typing_extensions import NotRequired, TypeAliasType, TypedDict, TypeVar, assert_never
 
 from . import _otel_messages, _utils
 from ._instrumentation import serialize_any
@@ -1496,30 +1496,26 @@ class NativeToolReturnPart(BaseToolReturnPart):
 
 
 if TYPE_CHECKING:
-    _ErrorDetails: TypeAlias = pydantic_core.ErrorDetails
+    from pydantic_core import ErrorDetails
 else:
 
-    class _ErrorDetails(TypedDict, total=False):
-        """Runtime stand-in for [`ErrorDetails`][pydantic_core.ErrorDetails] with every key optional.
+    class ErrorDetails(TypedDict):
+        """[`pydantic_core.ErrorDetails`][pydantic_core.ErrorDetails] with an optional `input` key.
 
-        `ErrorDetails` marks `type`, `loc`, `msg`, and `input` as required, but nothing enforces that when a
-        [`RetryPromptPart`][pydantic_ai.messages.RetryPromptPart] is constructed, and `content` is commonly
-        built from dicts that omit some of them, e.g. `ValidationError.errors(include_input=False)` or
-        history processors that redact `input`. Validating and serializing against this all-optional shape
-        keeps such message histories round-trippable through
-        [`ModelMessagesTypeAdapter`][pydantic_ai.messages.ModelMessagesTypeAdapter], instead of failing on
-        load after serializing with a `PydanticSerializationUnexpectedValue` warning.
+        `input` is not always present, e.g. when [`RetryPromptPart.content`][pydantic_ai.messages.RetryPromptPart.content]
+        was built with `ValidationError.errors(include_input=False)`, so it must not be required for a
+        serialized message history to load back.
         """
 
         type: str
         loc: tuple[int | str, ...]
         msg: str
-        input: Any
-        ctx: dict[str, Any]
-        url: str
+        input: NotRequired[Any]
+        ctx: NotRequired[dict[str, Any]]
+        url: NotRequired[str]
 
 
-error_details_ta = pydantic.TypeAdapter(list[_ErrorDetails], config=pydantic.ConfigDict(defer_build=True))
+error_details_ta = pydantic.TypeAdapter(list[ErrorDetails], config=pydantic.ConfigDict(defer_build=True))
 
 
 @dataclass(repr=False)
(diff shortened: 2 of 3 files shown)


adtyavrdhn:
I don't think we should do this unless a user actually complains about this. The framework paths always add input and it can only happen in a hand rolled path which I am not sure why a user would end up with. 

CC @DouweM

Issue #5964, opened 2026-06-17
`AGUIAdapter.dump_messages` reorders `ToolReturnPart` after `UserPromptPart`


mlorthiois:
### Initial Checks

- [x] I'm using the [latest version](https://github.com/pydantic/pydantic-ai/releases/latest) of Pydantic AI
- [x] I've searched for my issue in [the issue tracker](https://github.com/pydantic/pydantic-ai/issues) before opening this issue

### Description

`AGUIAdapter.dump_messages()` appears to reorder messages when a `ModelRequest` contains both a `ToolReturnPart` and a `UserPromptPart`.

  If the original Pydantic AI history has:

  1. assistant message with a tool call
  2. tool return
  3. user prompt

  then dumping to AG-UI produces:

  1. assistant message with a tool call
  2. user prompt
  3. tool return

This breaks providers that require tool results to immediately follow tool use blocks. For example, Anthropic with Bedrock rejects the next request with an error like:

  ```text
status_code: 400, model_name: eu.anthropic.claude-haiku-4-5-20251001-v1:0, body: {'message': 'messages.6: `tool_use` ids were found without `tool_result` blocks immediately after: toolu_bdrk_01Wd1vxxYxjAvGHhm1kUoxSf. Each `tool_use` block must have a corresponding `tool_result` block in the next message.'}
```

The issue occurs after loading an AG-UI input where the order is valid, running the agent, then sending a snapshot using `AGUIAdapter.dump_messages(result.all_messages())` and using this snapshot as input in a future agent run.

### Minimal, Reproducible Example

```Python
from pydantic_ai.messages import (
    ModelRequest,
    ModelResponse,
    TextPart,
    ToolCallPart,
    ToolReturnPart,
    UserPromptPart,
)
from pydantic_ai.ui.ag_ui import AGUIAdapter

messages = [
    ModelResponse(
        parts=[
            TextPart(content="Please answer the question."),
            ToolCallPart(
                tool_name="suggest_answer",
                args={"suggestions": ["Yes", "No"]},
                tool_call_id="toolu_1",
            ),
        ]
    ),
    ModelRequest(
        parts=[
            ToolReturnPart(
                tool_name="suggest_answer",
                content="['Yes', 'No'] suggested.",
                tool_call_id="toolu_1",
            ),
            UserPromptPart(content="Yes, I confirm."),
        ]
    ),
]

print("Original Pydantic AI message parts:")
for msg in messages:
[...]

Issue #5836, opened 2026-06-09
Define replay semantics for interrupted message histories


dsfaccini:
## Problem

Interrupted responses and requests can now be captured in message history, but we still need to define what callers may safely do with histories ending in `state='interrupted'`.

Capturing partial state is useful for persistence and observability. Replaying that state into a future model request is a separate contract.

## Scope

Decide:

- whether `message_history` containing `ModelResponse(state='interrupted')` is accepted, rejected, repaired, or handled via an explicit policy
- whether `ModelRequest(state='interrupted')` is replayable when it contains partial tool returns
- how incomplete tool-call arguments, missing tool returns, and provider-specific history constraints are handled
- whether callers need an explicit argument such as a replay policy when continuing after cancellation
- what docs should recommend for persistence followed by resume/continue

## Non-goals

- Do not duplicate [#3219](https://github.com/pydantic/pydantic-ai/issues/3219) / [#5364](https://github.com/pydantic/pydantic-ai/pull/5364), which are about capturing interrupted messages.
- Do not duplicate [#5615](https://github.com/pydantic/pydantic-ai/issues/5615) / [#5619](https://github.com/pydantic/pydantic-ai/pull/5619), which are about clean early-break stream history completion.
- Do not design the whole-agent cancellation API here except where it determines replay policy.

## Acceptance Criteria

- Tests showing the supported behavior when interrupted history is passed back as `message_history`.
- Provider-safe behavior for incomplete tool-call args and missing tool results.
- Documentation showing how to persist interrupted history and either resume safely or intentionally discard/repair interrupted parts.

## References

- [#3219](https://github.com/pydantic/pydantic-ai/issues/3219) - user report that interrupted streamed output was not persisted.
- [#5364](https://github.com/pydantic/pydantic-ai/pull/5364) - added interrupted message capture for exception/external-cancellation paths.
- [#5615](https://github.com/pydantic/pydantic-ai/issues/5615) / [#5619](https://github.com/pydantic/pydantic-ai/pull/5619) - clean early break should record partial response after stream context exit.
- [#5756](https://github.com/pydantic/pydantic-ai/issues/5756) - future v3 synchronous cleanup API discussion.

urbantech:
I would strongly prefer an explicit replay policy here rather than implicit repair.

A practical contract for interrupted durable runs:

- default: reject interrupted histories for replay;
[...]

Issue #5637, opened 2026-05-24
Expose public `message_history` validator for provider-valid histories


dsfaccini:
## Proposal

Add a public helper to `pydantic_ai.messages`:

```python
def is_provider_valid(messages: Sequence[ModelMessage]) -> bool:
    """Return True when `messages` can be safely passed to `Agent.run(message_history=...)`."""
```

The check: every `ToolCallPart` has a matching later `ToolReturnPart` or tool-bound `RetryPromptPart` (`tool_name is not None`), and every such return / tool-bound retry resolves a currently-open tool call (no orphans, duplicates, or out-of-order returns). Non-tool `RetryPromptPart` (output-validation retries) is fine because providers map it as a regular user message.

## Why

Every third-party capability that touches `message_history` is reinventing this walk. `pydantic_ai._agent_graph` enforces the invariant by construction, but consumers (harness capabilities, durable_exec wrappers, custom UIs that hand-build histories) can't easily reuse that. The closest existing helpers are provider-specific (`_collect_orphan_tool_search_call_ids` in `models/anthropic.py`).

## Caller already in tree

The new `StepPersistence` capability in `pydantic-ai-harness#251` ships its own `is_provider_valid` ([source](https://github.com/pydantic/pydantic-ai-harness/blob/step-persistence/pydantic_ai_harness/step_persistence/_helpers.py)) gating `ContinuableSnapshot` writes. We'd drop ours and call into pyai once this lands.

## Open questions

- Live in `pydantic_ai.messages` (data-shape utilities) or `pydantic_ai.utils` (cross-cutting)?
- Strict mode only, or also a `repair` variant that returns a fixed-up list? The orphan-repair concept already appears in passing in `_agent_graph` and harness review threads.

dsfaccini:
David's AICA here:

Quick implementation sketch.

### Location

`pydantic_ai/messages.py`. It's a pure structural check over message-part shapes that already live in that module (`ToolCallPart`, `ToolReturnPart`, `RetryPromptPart`, `ModelResponse`), and `messages.py` is the public re-export surface. `pydantic_ai.utils` is private and `_history_processor.py` is too narrow.

### Signature

Start with the simple bool. Defer the structured-report variant until a second caller asks for it — the only in-tree consumer today (`StepPersistence` snapshot gate) just needs the bool.

```python
from collections.abc import Sequence

def is_provider_valid(messages: Sequence[ModelMessage]) -> bool: ...
```

[...]

PR #6319, opened 2026-07-07
Make `message_history` provider-valid out of the box (repair tool-call/result pairing)

### LINKED ISSUE #3461: How to handle follow-up actions when the agent is aborted during a tool invocation.
### Question

First, thank you for providing the ag_ui integration feature—I’m using it very well.

However, while recently testing the abort function, I encountered the following error:
```
data: {"type":"RUN_STARTED","threadId":"553632f0-09a6-4053-8c95-61b85b03a1b9","runId":"8735d904-7b17-4adc-b7fe-3dca71bf0fb8"}

data: {"type":"RUN_ERROR","message":"status_code: 400, model_name: gpt-4, body: {'message': 'No tool output found for function call call_tYFQBQ4RTmvnbPkV40KWCKCh.', 'type': 'invalid_request_error', 'param': 'input', 'code': None}"}
```
Right after the error occurred, agent.messages is as follows.
```json
[
  {
    "id": "019a9502-57fb-7209-a94f-c47d2b2f3647",
    "role": "user",
    "content": "test"
  },
  {
    "id": "d9d864d5-3adb-4c92-b42d-489b65e68639",
    "role": "assistant",
    "toolCalls": [
      {
        "id": "call_tYFQBQ4RTmvnbPkV40KWCKCh",
        "type": "function",
        "function": {
          "name": "retrieve_documents",
          "arguments": "{\n  \"search_query\": \"test\"\n}"
        }
      }
    ]
  },
  {
    "id": "019a9502-d49b-7159-bfc5-dc3142d2255b",
    "role": "user",
    "content": "test"
  }
]
```

I executed abort while a tool was running.  
After it was aborted, when I entered a new message, there was still a message indicating a tool had been called, but there was no result from the tool in the message, so all subsequent workflows seemed to fail.

It seems to be related to this issue #2050.  
Is there a good way to resolve this?

For a new message, is the best approach to check whether a tool call has results; and if not, to explicitly enter a failure result?
If this is an issue that should be handled on the frontend rather than by pydantic-ai, just let me know.

### Additional Context

_No response_

### LINKED ISSUE #4728: Built-in HistoryProcessor for orphaned tool call/result repair

DouweM:
- Fixes #4728
- Fixes #3461
- Related to #5637, #5836, #4882

A run that is cancelled or crashes mid-tool — or a hand-built, truncated, or context-evicted history — can leave the message history with broken tool-call/tool-result pairing that strict providers (Anthropic, Gemini, OpenAI-compatible) reject on the next request: a tool call with no result (OpenAI `400 No tool output found for function call`, the AG-UI abort-mid-tool repro in #3461) or a result with no call (OpenAI Responses `No tool call found for function call output`, #4882). Today every application that persists and resumes histories has to reinvent the same repair walk (see #5637), or worse, monkeypatch `_clean_message_history` to get such histories through.

This PR makes the history **provider-valid out of the box**: `_clean_message_history` in `_agent_graph` becomes an ordered pipeline of repair passes that run before every model request, so interrupted and hand-built histories can be passed straight back into a run. It resolves the canonical, maintainer-endorsed request in #4728 via the automatic-internal path DouweM called out there. Two earlier attempts implemented the same idea as a history processor (#5021) and then as a capability (#5056) — both closed; this PR lands the repair internally instead, per DouweM's direction. This is the "repair" half of #5637 implemented internally. The public `is_provider_valid()` validator proposed there is complementary, not obviated: repair *fixes* a history to send to a model, whereas a validator lets a caller *judge* a history and branch on it — e.g. snapshot/replay gating that must detect a crashed-mid-tool point and refuse to persist it as a resumable continuation, rather than silently synthesizing a tool return for it. That predicate remains a separate follow-up (#5637).

The guiding principle (DouweM): *massage the history however we can to make the model API accept it, and drop only what's fundamentally unsendable* — never silently removing content that was semantically meaningful. So the passes only **ADD** (synthesize) or **REMOVE** parts no provider could ever accept; nothing meaningful is dropped.

### The pipeline

`_clean_message_history(messages, *, repair_last_response)` runs three pure `list[ModelMessage] -> list[ModelMessage]` passes in order over **regular, locally-executed** tool call/result pairing (order matters — matching-graph repairs before the merge that changes message boundaries). Native/builtin parts are left entirely untouched (see Scope):

1. **`_drop_orphaned_tool_results`** (REMOVE) — a `ToolReturnPart` or tool-bound `RetryPromptPart` (in a request) whose `tool_call_id` never appeared as a regular `ToolCallPart` in any *preceding* message. No provider accepts a `tool_result`/`function_call_output` without its call. A `RetryPromptPart` with no `tool_name` is validation feedback (plain user text, not a tool result) and is left alone. Emptied interior request → dropped; emptied *last* request → kept empty so history still ends on a `ModelRequest`.
2. **`_repair_dangling_tool_calls`** (ADD synthesized results) — the existing dangling-call pass. Every call with no result gets a synthesized `ToolReturnPart` (`'The tool call was interrupted before a result was produced.'`, `outcome='interrupted'`, `metadata={'pydantic_ai_synthesized_tool_return': True}`). This includes a call whose args were cut off mid-stream (unparsable JSON): the call is kept verbatim and closed out like any other dangling call — malformed args are already sendable (serializers degrade them via `ToolCallPart.args_as_dict`'s `INVALID_JSON` wrapper, which the tool-call retry flow has always relied on), and removing the call would disturb the response's shape (e.g. leave a thinking-only response whose signature was computed over a turn that included the call). Frontier-gated by `repair_last_response` (below).
3. **`_merge_consecutive_messages`** (normalize) — the existing consecutive-same-role merge, extracted unchanged; runs last, once pairing is valid, so it never separates a result from its call. Changes message boundaries but preserves all content.

Matching in passes 1 and 2 keys on the concrete regular `ToolCallPart`/`ToolReturnPart` types; `NativeToolCallPart`/`NativeToolReturnPart` are *sibling* `Base*Part` subclasses, so native parts are automatically ignored by every pass.

**Matching is an ordered walk** (per the sketch in #5637): a result only answers a call that is *open* — produced by an earlier response and not already answered — at that point. A result preceding its call, a duplicate, or a `tool_call_id` reused across responses therefore can't mask a genuinely dangling call, and an out-of-place result is recognized as an orphan.

**Repair is silent.** Like the existing consecutive-message merge in `_clean_message_history`, repair is quiet cleanup — it doesn't warn or raise. The synthesized `ToolReturnPart`'s `pydantic_ai_synthesized_tool_return` metadata marker makes every repair inspectable in the resulting history, and a Logfire/OTel span attribute is the intended non-noisy observability surface as a follow-up. (A Python `UserWarning` was tried but, under the test suite's `filterwarnings=["error"]`, it fired inside a legitimate internal flow — an `after_model_request` hook raising `ModelRetry` leaves the just-streamed tool call dangling until the wire-level pass synthesizes a return — so it cried wolf on the framework's own retry loops and was dropped.)

**The live frontier is protected.** The *last* `ModelResponse`'s tool calls may still legitimately receive results, so repair only touches them when the interruption is evident:

- History ending on a response with tool calls + no new prompt → run resumption executes them (unchanged). A trailing response ending on an unparsable-args call also stays intact on promptless resume, so local args validation still turns it into a `RetryPromptPart` the model can react to.
- Deferred/approval flows: a run ending in `DeferredToolRequests` persists `resp(A executed, B deferred) + req(return A)` with a `'complete'` trailing request — that shape is never repaired at run start (gated on `ModelRequest.state == 'interrupted'`), so `deferred_tool_results` matching keeps working and a forgotten-results resume doesn't close the approval-pending call in the persisted history. The wire-level pass (below) answers it on the copy sent to the model only.
- History ending on a response with tool calls + a new user prompt → still raises `Cannot provide a new user prompt when the message history contains unprocessed tool calls.` for `state='complete'` responses, but for `state='interrupted'` responses (e.g. a cancelled stream) the calls will never be executed, so they are repaired instead of refused — the "cancel a stream, then continue the conversation" flow now just works.
- A trailing `ModelRequest` with `state='interrupted'` (a run that crashed during tool execution, leaving partial tool returns) means the preceding response's unanswered calls can never be executed, so they are closed out at run start.
- As a final safety net, the per-step wire-level cleaning (after history processors and `Model.prepare_messages`) repairs everything including the last response, since at that point the history is definitively being sent. These wire-level repairs are not persisted into run state.

### Determinism, idempotency, cache stability

Synthesized parts derive their timestamp from the response they repair and contain no wall-clock or random data: repairing the same history always produces the same bytes, and re-running a repaired history is a no-op (pinned by tests), so repair never churns provider prompt-cache prefixes across turns.

### Security interaction (GHSA-jpr8 / #6169)

The AG-UI `sanitize_messages` tail-strip removes unresolved *trailing* tool calls from client-supplied history so a forged call can't reach the promptless-resume execute path. Core repair composes with (and never undoes) that: it doesn't touch the trailing response on the execute paths, and for mid-history dangling calls it only *answers* them with a synthetic return — a repaired call is inert and can never execute, so repair strictly reduces what forged history can do.

### Scope — reproduced-and-included vs. deliberately excluded

Every repair was validated against the provider rule and message shape it satisfies. Included (each with a focused test asserting the exact repaired history a capturing `FunctionModel` receives): dangling-call synthesize (including calls with mid-stream-truncated args) and orphaned-result drop (#4882/#4728).

Excluded, with reasons (left as follow-ups so the surface stays scoped):

- **Dangling builtin/native tool calls, incl. Anthropic `code_execution` (#6401)** — *investigated and excluded from the core pipeline.* Native/builtin parts (`NativeToolCallPart`/`NativeToolReturnPart`) are fundamentally different from locally-executed tool calls: the provider produces the call **and** its result inline, they're co-located in one `ModelResponse` (and a tool-search result can even arrive in a *later* response — see `models/anthropic.py` `_collect_orphan_tool_search_call_ids`), and they're contractually **preserved** in `all_messages()` (`test_xai_builtin_tool_without_tool_call_id` pins that an empty-id native call stays in history while xAI's serializer skips it on the wire at `models/xai.py:493`). A generic per-message "dangling native call" drop in the core graph is both redundant (serializers already repair these — Anthropic repairs tool-search orphans) and harmful (it would surgically remove user-constructed content from persisted history, and would strand a legitimate tool-search call whose result lands in a later response). So #6401's Anthropic `code_execution` 400 should be fixed in the Anthropic model's serialization as a separate scoped change, not here. The pipeline leaves all native parts exactly where they are (regression test: a history with a dangling native call passes through `_clean_message_history` byte-identical).
- **Reorder out-of-place tool results (#5964, #4827)** — *investigated and excluded.* Both are **downstream serialization/mapping bugs, not history-shape bugs**: in #5964 the `ModelRequest.parts` were already correctly ordered `[ToolReturnPart, UserPromptPart]` and the AG-UI adapter (`_dump_request_parts`) moved the user part ahead of the result (fixed by #5972); in #4827 the OpenAI mapping inserts the `instructions` system message between an assistant `tool_calls` message and its `tool` result because it counts *all* system messages including history-embedded ones. Neither is fixable by reordering parts *within* the history, and the existing merge already hoists tool results to the front of a merged request — so a history-level reorder pass would be redundant here and risk reordering user/text content. No clean reproduction at the history level → excluded (CLAUDE.md scope discipline).
- **Duplicate tool results for one call** — a second `tool_result` for an already-answered call. Removing "the redundant one" is a which-is-meaningful judgment; left for the broader validity pass.
- **Provider-specific ordering rules** beyond call/result pairing — part of #5637's fuller validity definition and the cross-model replay class.

### Design decisions for review

- **Keep-and-synthesize (not drop) for unparsable-args calls**: originally these were dropped, on #4053's premise that incomplete args must never reach a provider. That premise turned out to be false: `ToolCallPart.args_as_dict` deliberately degrades malformed JSON to `{'INVALID_JSON': '<raw args>'}` precisely so it "can still be sent to a model API (e.g. during a retry flow)", and the retry flow has always replayed such calls (a complete-but-invalid-args call answered by a `RetryPromptPart` round-trips to the provider today). Dropping was also silent surgery on the model's self-model (phantom-action references — deepagents' `PatchToolCallsMiddleware` synthesizes for this reason) and, worse, could leave a thinking-only response whose signature was computed over a turn that included the call — the exact signed-thinking risk that ruled out drop for regular dangling calls. Now every dangling call gets the same synthesized return, and the REMOVE set shrinks to exactly one thing: orphaned results, the one shape that cannot be truthfully repaired (the missing call's args are unknowable, so any synthesized call would fabricate an assistant action).
- **Marker**: `metadata={'pydantic_ai_synthesized_tool_return': True}` on the synthesized `ToolReturnPart`, following the `DEFERRED_CAPABILITY_TOOL_METADATA_KEY` precedent (`pydantic_ai_`-prefixed key, module-level constant `SYNTHESIZED_TOOL_RETURN_METADATA_KEY` in `_agent_graph`). No new public API in this PR, so the key is documented as a literal in the message-history docs.
- **`outcome='interrupted'`**: a new neutral `BaseToolReturnPart.outcome` value — we don't know whether the tool ran, failed, or never started, so the synthesized return must not claim `'failed'` (or success). It is never sent to any LLM provider (the content string carries the wording) and UI adapters render it neutrally rather than as an error. It also survives Vercel AI and AG-UI dump/load round-trips: Vercel carries the claim in the `pydantic_ai` `callProviderMetadata` channel (`failed`/`denied` already ride its typed part states), and AG-UI carries every non-`success` outcome — regular and native returns, previously all reloading as `success` — in the `encrypted_value` carrier alongside `tool_kind` (>= 0.1.11). Anything that can affect provider serialization must round-trip, or a reloaded history changes the request bytes and breaks prompt-cache prefix stability. Coordinated with #5585, which maps only `outcome='failed'` to provider native error channels, so `'interrupted'` correctly stays out of them.
- **Wording**: `'The tool call was interrupted before a result was produced.'`, styled after `'The tool call was denied.'` (`ToolDenied.message`).

### Open questions

1. Should the `UserError` for a *complete* trailing response with unanswered tool calls + new prompt also be replaced by repair? Kept for now as an explicit-policy guard for deferred/approval flows (cf. urbantech's argument in #5836), at the cost of hand-built truncated histories without `state='interrupted'` still erroring in that one spot.
2. Should `SYNTHESIZED_TOOL_RETURN_METADATA_KEY` (and/or an `is_provider_valid()` companion, #5637) be exposed publicly in a follow-up?

### Checklist


DouweM:
_Review by Claude (Fable) against the transcript-repair design notes in pydantic-ai-notes, posted from Douwe's session; the follow-up agent should treat the numbered items as the work list._

Deep-reviewed this against the transcript-repair design notes (`is_provider_valid` + main-loop repair), the cache-prefix-stability contract (repair determinism arm), and the harness-comparison observability findings. The core of this is right and lands exactly where we wanted it:

- Placement adjacent to `_clean_message_history` with the `repair_last_response` frontier flag is the design we settled on, and the wire-level safety net in `_prepare_request` covering history-processor output is a good catch.
- The determinism/idempotency work is done properly: timestamps derived from the repaired response, `ModelRequest.timestamp=None` default keeping the synthesized-request path stable, repair persisted into run state at ingestion (so turn N+1 replays identical bytes), and both pinned via `dump_json` byte-equality tests. This satisfies the cache-stability contract as written.
- Test coverage of the mechanism itself (parallel partials, drop-vs-synthesize, byte-identical no-op, round-trips from real cancelled streams) is strong.

Concerns to address, in order:

1. **The run-start trailing-`ModelRequest` repair clobbers the deferred-approval shape.** A run ending in `DeferredToolRequests` persists exactly `resp(A executed, B deferred) + req(return A)` — see `_handle_deferred_tool_results`, which walks back to that trailing request. The new block in `UserPromptNode.run` fires `repair_last_response=True` whenever history ends on a `ModelRequest`, gated only on `deferred_tool_results is None` — so resuming that history with a new prompt (or forgetting `deferred_tool_results`) silently closes the approval-pending call as "interrupted" instead of failing loudly. This contradicts open question 2's own rationale, which keeps the `UserError` on the resp-tail shape *precisely* as a guard for deferred/approval flows: the req-tail shape is the actual persisted deferred shape, and it gets no guard. Suggested fix: key this branch on `messages[-1].state == 'interrupted'` (the crash-mid-tool marker, symmetric with the resp-tail branch — the `capture_run_messages` history in `test_interrupted_tool_execution_round_trips` already has it), and add a regression test resuming a deferred run's history without results.

2. **The incomplete-args drop isn't gated by `repair_last_response`, breaking the "live frontier is protected" claim.** In `_repair_dangling_tool_calls`, `incomplete_ids` filtering runs before the `repair_last_response or index != last_response_index` check. So resuming a history ending on a response with an unparsable-args call and *no new prompt* — the case the PR body says "run resumption executes them (unchanged)" — silently drops the call at the run-start `_clean_message_history` pass, where previously it went through local args validation → `RetryPromptPart` → the model could retry. #4053 forbids incomplete args reaching a *provider*; resumption executes locally. Move the incomplete-args handling behind the same frontier gate.

3. **Make repair observable to the developer.** Repair currently fires with no warning, span event, or Logfire attribute — the metadata key only helps if you already suspect it. A `warnings.warn` (or span event) when repair changes anything is cheap and would surface exactly the bug class in point 1 (forgotten `deferred_tool_results`, over-aggressive history processors). Relatedly, on open question 1: the design notes lean toward observable-over-silent (silent surgery corrupts the model's self-model — it said "let me look that up" and the call vanishes), and deepagents' `PatchToolCallsMiddleware` synthesizes rather than drops for the same reason. Dropping per #4053 is defensible for v1, but at minimum emit the developer-facing signal; a `repair_note` option can be a follow-up.

4. **Use an order-aware walk instead of the global `answered_ids` set.** A return preceding its call, a duplicate return, or a `tool_call_id` reused across responses (realistic in hand-built histories) masks a genuinely dangling call. The open-calls walk sketched in #5637 is the same complexity, strictly more correct, and positions this code to be shared with the future public `is_provider_valid`.

5. **Scope the docs claim.** "Provider-valid by construction" overclaims: orphaned returns, duplicates, and out-of-order returns — all part of #5637's validity definition — aren't repaired and providers reject those too. Say "dangling tool calls are repaired" and note the rest as future work. (Keeping the *implementation* scoped is fine; `_repair_dangling_tool_calls` being private leaves room for the broader repair framework, including the cross-model replay class.)

On the other open questions: deferring `is_provider_valid` and the public metadata-key export to follow-ups is fine; keeping the `UserError` for *complete* trailing responses is right (and point 1 makes the two tails consistent about it).

DouweM:
Reminder to check out vstorm's `PatchToolCallsCapability`

DouweM:
_Analysis by Claude (Claude Code), posted on behalf of DouweM to open a design question before this merges._

## Open question: how should we repair a **dangling tool call** (a tool call with no result)?

This PR makes message history provider-valid before each request. For a dangling tool call it currently **synthesizes** a `ToolReturnPart` (content `"The tool call was interrupted before a result was produced."`, `outcome='failed'`). Two things are worth settling before merge:

### 1. We must not assert an outcome we don't know
`outcome='failed'` is a structured claim that the tool **failed**. We don't know that — the tool may have executed its side effect (created a file, sent a request) before the result was lost, or it may never have run. Claiming `failed` is as wrong as claiming `succeeded`. If we synthesize at all, the return must convey **"outcome unknown / interrupted"**, never success or failure. This part seems uncontroversial.

### 2. Synthesize vs. drop vs. truncate — the real debate
For the same dangling call we have three repair strategies, with different effects on the model's world-model:

- **Synthesize (unknown):** keep the call, add an "outcome unknown" return. *Pro:* preserves the record that a call was attempted; structurally safe (doesn't disturb the response's internal shape — thinking/signature pairing, partial-parallel siblings — which this PR already handles). *Con:* fabricates a message the model never actually received.
- **Drop the call:** remove the dangling `ToolCallPart`. *Pro:* never fabricates anything; cleanest for the "cancel the stream, then redirect" flow (model re-plans from a clean slate); matches Vercel AI's `convertToModelMessages({ ignoreIncompleteToolCalls: true })`. *Con:* erases that the model tried the call, and removing a part from *inside* a response can create new edge cases (e.g. a response left as thinking-block-only may not serialize cleanly for Anthropic).
- **Truncate:** cut back to the last valid boundary. Only viable for a **trailing** dangling call — you can't truncate a call buried mid-history without discarding the real conversation after it.

The core tension: an honest *omission* (drop) vs. an honest *fabrication* (synthesize-unknown). For a model trying to maintain an accurate world-model, it's not obvious which is less misleading.

### 3. Trailing vs. mid-history probably deserve different treatment
A **trailing** dangling call (the frontier — nothing after it) can be dropped or truncated cleanly, and is often the "interrupted, user is about to redirect" case. A **mid-history** dangling call has real conversation after it, so truncate is off the table and it's drop-or-synthesize only. Treating the frontier differently from a buried orphan may be the right design rather than one uniform rule.

### Where we're taking this
Rather than decide in the abstract, we're (a) surveying what other agent frameworks treat as the standard, always-safe repair tricks (Pi / the TS ecosystem, Vercel AI SDK, LangChain/LangGraph, deepagents, etc.), and (b) prototyping the drop variant to see whether it's actually clean against the tricky response shapes. Input very welcome — especially concrete "here's what breaks" scenarios for any of the three strategies.

DouweM:
_Follow-up analysis by Claude (Claude Code) — resolving the synthesize-vs-drop question above._

Surveyed the field (Vercel AI, pi-mono, Mastra, deepagents; LangChain, LangGraph, OpenAI Agents SDK, Anthropic guidance) and prototyped the drop variant end-to-end. **Verdict: keep synthesize; make the synthesized return's outcome honest.**

**The field splits by data model, not taste.** Frameworks that model tool state *on the tool-call part* (Vercel's UIMessage) drop the incomplete call; frameworks with **separate call/result messages** — pi-mono, LangChain, and pydantic-ai — **synthesize** a result, specifically to preserve thinking-block signatures (pi-mono's own comment: *"insert synthetic empty tool results for orphaned tool calls — this preserves thinking signatures and satisfies API requirements"*). OpenAI Agents does drop, but has to *also* strip the reasoning items tied to the dropped call — a cost that comes with drop. Our own closed prior art (#5021/#5056) already landed on synthesize + neutral content + drop-orphaned-results + a trailing-vs-mid distinction.

**Prototyping drop confirmed the risk empirically.** The make-or-break shape — the one a cancelled stream actually produces — is a `[ThinkingPart(signature), ToolCallPart]` response. Dropping the call leaves a **thinking-only assistant turn** whose signature was computed over a turn that *included* the tool call. That shape appears in **zero** recorded cassettes, and Anthropic requires signed thinking blocks to be replayed unchanged — so it's a one-directional 400 risk; for OpenAI/Google the thinking degrades to a lossy text tag. Synthesize keeps the canonical `[thinking, tool_use] → tool_result` shape and never touches the signed turn. (Drop is clean for text+call, call-only, and partial-parallel — just not for the thinking case, which is the one that matters here.)

**On the "don't pretend it failed" concern:** on the wire, only the content *string* reaches the model, and ours is already honest ("The tool call was interrupted before a result was produced." — no success/failure claim). The overstatement is purely `outcome='failed'`, which today only affects UI rendering and history inspection. The fix is narrow: give the synthesized return a new neutral **`outcome='interrupted'`** (better than `'failed'`), distinct from a tool that ran and errored. Re #5585 (`ToolFailed`): that PR maps only `outcome='failed'` to a provider's native error channel, so `'interrupted'` correctly stays out of it — the content string already carries the wording.

**Trailing vs mid-history:** no provider-level difference surfaced between the two positions; the distinction that actually matters is *"can this call still be executed"*, which is exactly what the existing frontier gate (`repair_last_response`) already encodes. So no position-based drop is needed — the semantic version is already here.

Net change from this: keep the synthesize pipeline as-is, switch the synthesized return from `outcome='failed'` to `outcome='interrupted'`.

DouweM:
Douwe's AICA here: Review by OpenAI Codex.

I would not merge this as currently published. The overall direction is right—central repair is substantially better than requiring every caller to sanitize history—but three semantics need resolving first.

## 1. High: orphan-result repair silently deletes potentially true world state

`_drop_orphaned_tool_results` simply removes an unmatched `ToolReturnPart`. Its content may be the only surviving evidence that a side effect happened, for example:

```text
tool result: "Transfer completed, transaction 123"
assistant: "The transfer is complete."
```

If context eviction removed the call, repair deletes the result but retains the assistant's conclusion. The resulting history is provider-valid yet causally inconsistent. This also contradicts the documentation's assertion that repair never discards anything meaningful.

There is no unambiguously correct automatic reconstruction here: the missing call's arguments and provenance cannot be inferred. I recommend keeping automatic repair focused on dangling calls and treating orphaned results as a separate policy decision—warning/error, explicit processor, or a deliberately lossy conversion that preserves the result as an uncertainty-marked historical note. At minimum, do not silently mutate persisted run state.

## 2. High: incomplete-argument handling is neither complete nor always provider-valid

`_tool_call_args_incomplete` is only applied to calls returned by `_dangling_tool_calls_by_response`. Therefore a malformed call with an existing result passes through unchanged:

```python
[
    ModelResponse(parts=[ToolCallPart('x', '{', tool_call_id='1')]),
    ModelRequest(parts=[ToolReturnPart('x', 'done', tool_call_id='1')]),
]
```

I reproduced `_clean_message_history(..., repair_last_response=True)` returning this unchanged. A provider can still reject it, despite the documented claim that unparsable calls are removed before every request.

Conversely, when a malformed call is dangling, only the `ToolCallPart` is removed. For a cancelled Anthropic-style response shaped like:

```text
[signed ThinkingPart, incomplete ToolCallPart]
```

this leaves a thinking-only response whose signature was produced for a different response shape. That is precisely the signed-thinking risk used in the PR discussion to reject dropping normal dangling calls.

This needs a defined policy for the whole response, not merely filtering one part. It may need provider-aware preparation, or a conservative truncation rule for interrupted responses.

## 3. Medium: synthesized returns falsely record `outcome='failed'`

`_repair_dangling_tool_calls` records `outcome='failed'`. Interruption only means that no result was recorded; the tool may have completed its side effect, failed, or never started.

This is not merely cosmetic. Vercel UI adapters interpret `failed` as an error outcome. It therefore presents invented information to application/UI consumers even though the textual result is appropriately neutral.

The local follow-up commit `8edf85c22` already implements the sounder design: add `outcome='interrupted'`, keep it neutral in UI serialization, and let the content explain the uncertainty. That commit is not on the currently published PR head.

## Overall assessment

The best part of the design is dangling-call synthesis:

- It preserves signed model responses.
- It honestly tells the model that no result was recorded.
- It is deterministic and cache-stable.
- The live-frontier and deferred-approval protections are thoughtful.
- Ordered matching and reused-ID handling are considerably better than a global answered-ID set.

The questionable leap is treating every structurally invalid history as automatically and silently repairable. Provider validity and historical truth are different invariants:

| Invalid shape | Safe default |
|---|---|
| Dangling call with complete arguments | Synthesize an uncertainty-marked interrupted result |
| Dangling call with truncated arguments | Needs a whole-response/provider-aware policy |
| Orphaned result | Cannot be repaired truthfully without an explicit lossy policy |
| Duplicate/out-of-order result | Leave out of scope or diagnose explicitly |
| Native/server-side parts | Correctly left to provider adapters |

I would narrow the PR's promise to "automatically close safely repairable dangling regular tool calls before model requests," land neutral `outcome='interrupted'`, and defer orphan deletion plus malformed-call surgery until their world-model semantics are explicitly settled.

Verification:

- `uv run pytest tests/test_transcript_repair.py -q`: 24 passed
- `git diff --check origin/main...HEAD`: clean
- Published CI's core test, lint, docs, coverage, and CodeQL checks pass.
- The harness compatibility failure is unrelated: its own README is missing a guardrails link.

DouweM:
_Response by Claude (Claude Code), on behalf of DouweM — verifying each finding against the code._

Thanks — finding 2 caught something real, and the fix is now pushed (`f77833b9e`). Verdicts on all three:

## 2 (incomplete-argument handling): confirmed, fixed — by *keeping* the call, not more surgery

Both sub-claims were verified, but they pointed at the same root cause from opposite sides: **the drop itself was the mistake**, resting on a false premise.

- The premise "incomplete args must never reach a provider" is wrong: `ToolCallPart.args_as_dict()` deliberately degrades malformed JSON to `{'INVALID_JSON': '<raw args>'}` precisely so it *"can still be sent to a model API (e.g. during a retry flow) without crashing"* — and the retry flow has always replayed exactly this shape (a complete-but-invalid-args call answered by a `RetryPromptPart` round-trips to Anthropic/Google/Bedrock via that wrapper, and to OpenAI as the raw string). So your case of "a malformed call with an existing result passes through unchanged" is not a hole — it's the framework's own working retry-flow shape, and *removing* it would orphan the `RetryPromptPart` and break the retry loop. What was wrong was the docs claiming unparsable calls "can't be replayed".
- Your second case — dropping a dangling incomplete-args call from `[signed ThinkingPart, incomplete ToolCallPart]` leaves an unattested thinking-only response — was exactly right, and it's the shape a cancelled stream actually produces. The fix removes the special case entirely: a dangling call with truncated args is now kept verbatim and closed out with the same synthesized `outcome='interrupted'` return as any other dangling call. No response surgery, signatures untouched, and the pipeline gets simpler (the REMOVE set shrinks to orphaned results only). Tests and docs updated accordingly.

## 1 (orphaned-result drop): staying as-is, deliberately

The scenario is real but the drop remains the least-bad option, and it was an explicit maintainer decision in this PR's design review:

- An orphaned result is the one shape that *cannot* be truthfully repaired: the missing call's args are unknowable, so synthesizing the call would fabricate an assistant action — a bigger lie than the omission. Preserving the result as an "uncertainty-marked historical note" (a text part) fabricates a user-turn message that was never sent, trading a structural lie for a provenance lie; no surveyed framework does this (Vercel, LangChain-ecosystem middleware, and our own prior art #5021/#5056 all drop).
- The realistic sources are naive slicing (`history[-n:]`) and hand-built histories, where the orphan is a consequence of the caller's own cut; a summarizer that evicts a response while keeping its paired results is buggy upstream, and preserving its bug as fake user text is worse DX than the clean drop. The alternative — leaving it in — is a guaranteed 400 that kills the whole run.
- It's not silent in the sense that matters: the docs enumerate exactly this removal (the "Making histories provider-valid" section documents both REMOVE cases — now just this one — rather than claiming nothing is ever removed; "nothing meaningful is silently dropped" is scoped by the fundamentally-unsendable rule stated right above it).

A warn/error/policy knob stays a possible follow-up if real-world reports show meaningful orphans, but out of the box the history must send.

## 3 (`outcome='failed'`): already fixed before your review

`8edf85c22` (neutral `outcome='interrupted'`, UI adapters render it neutrally, coordinated with #5585) was the published PR head at review time — the checkout reviewed was stale. Nothing further to do.

PR #3335, opened 2025-11-04
Proposal: AdaptiveModel

ChuckJonas:
_WIP: submitted code in POC stage_
_Based on discussions from https://github.com/pydantic/pydantic-ai/issues/3023_

`AdaptiveModel` is a new model type that provides full control over model selection at runtime. Unlike `FallbackModel` which tries models sequentially, `AdaptiveModel` allows custom logic to select the next model based on rich context including attempts, exceptions, and agent dependencies.

## Core API

```python
class AdaptiveModel[AgentDepsT](Model):
    def __init__(
        self,
        models: Sequence[Model],
        selector: Callable[[AdaptiveContext[AgentDepsT]], Model | None] 
                  | Callable[[AdaptiveContext[AgentDepsT]], Awaitable[Model | None]],
        *,
        max_attempts: int | None = None,
    ):
        """
        Args:
            models: Pool of models to choose from
            selector: Sync or async function that selects the next model to try
            max_attempts: Maximum total attempts across all models (None = unlimited)
        """
```

## Execution Flow

The `AdaptiveModel` automatically handles fallback and retry by:

1. Calling `selector(context)` to get the next model to try
2. Attempting the request with that model
3. If the request succeeds, returning the result
4. If the request fails:
   - Recording the attempt (model + exception) in `context.attempts`
   - Calling `selector(context)` again with updated context
   - If selector returns a `Model`, goto step 2 (retry/fallback)
   - If selector returns `None`, raise exception group with all failures
5. If `max_attempts` is reached, stop and raise exception group

The selector has **full control** over retry/fallback logic:
- Return the **same model** that failed → retry with same model
- Return a **different model** → fallback to another model  
- Return `None` → stop trying
- Inspect `ctx.attempts[-1].exception` to make decisions
- Use `time.sleep()` or `await asyncio.sleep()` for backoff/waiting (selector can be sync or async)

## Context

```python
@dataclass
class AdaptiveContext[AgentDepsT]:
    """Context provided to the selector function."""
    
    run_context: RunContext[AgentDepsT] | None  # Access to agent dependencies (None for non-streaming)
    models: Sequence[Model]  # Available models
    attempts: list[AttemptResult]  # History of attempts in this request
    attempt_number: int  # Current attempt number (1-indexed)
    messages: list[ModelMessage]  # The original request
    model_settings: ModelSettings | None
    model_request_parameters: ModelRequestParameters

@dataclass
class AttemptResult:
    """Record of a single attempt."""
    model: Model
[...]

PR #6333, opened 2026-07-07
Add `get_model`, `resolve_model_id`, and `for_agent` capability hooks

DouweM:
## Motivation

Capabilities can contribute almost every part of an agent runtime configuration, but could not bind reusable capability instances to an agent or choose and construct models. That blocks dependency-aware tenant, region, credential, adaptive-routing, and durability use cases, including model-less agents whose model is supplied by a capability.

Related design work: #3335 and #4977.

## Design

This adds three composable capability hooks:

- `for_agent(agent)` binds a reusable capability to constructor or per-run agent configuration. Constructor model strings remain raw during binding so capabilities can define their own interpretation and credentials. Static run/spec capabilities bind once per run; `CapabilityFunc` results bind before their own `for_run()`, while specialized run-bound values returned by an ordinary `for_run()` are not rebound.
- `get_model()` supplies a static `Model`/model ID, or a sync/async `ModelSelectionContext` callable evaluated once per new logical request step. `SelectModel(selector)` is the convenience capability for adaptive policies.
- `resolve_model_id(ctx, *, model_id)` turns a selected string into a `Model`, with the `ResolveModelId` convenience capability for user callables.

`ModelSelectionContext` exposes the agent, run dependencies, lower-precedence model, run step, messages, and usage without pretending a full `RunContext` exists before the model has been selected. A model-less agent can select its first model using live dependencies during run setup.

Precedence is:

`override(model=...)` / explicit run model › run-spec model › capability `get_model()` › constructor model.

Later model contributions win. Model ID resolvers instead form a chain of responsibility, where the first non-`None` result in user order wins. Explicit models skip selection but still use capability resolution when supplied as strings.

Bootstrap IDs use the post-`for_agent`, pre-`for_run` resolver chain. If `for_run()` replaces the capability, step-one and later selections use the replacement resolver chain. Static selections are resolved and entered once per run. Dynamic selections happen before model-dependent settings, profile, tool, message, token-counting, instrumentation, and request preparation. Same-step provider continuation remains pinned; ambiguous cross-run continuation is rejected.

Run-spec capabilities participate in bootstrap. Deferred capabilities do not contribute model hooks, even after loading; post-load model selection is explicitly deferred pending demand and a clear continuation contract. Dynamic selection under durable execution is also deferred. #4977 can carry raw constructor IDs through its durable proxy and resolve them worker-side.

Fallback models compose with selection today. Replacing the `FallbackModel` wrapper with a capability in v3 is plausible, but requires moving per-model message preparation and actual-attempt tracking into the request handler first; the current wrapper hook runs after initial-model message preparation.

## Tests and documentation

Coverage includes binding identity and replacement, raw constructor IDs, per-run and `CapabilityFunc` binding, precedence and ordering, dependency-aware resolution, first-step adaptive selection, async/per-step selection, streaming, model lifecycle cleanup, static pinning, `for_run()` replacement, run-spec bootstrap, continuation safety, fallback composition, non-run consumers, and the web UI handling of resolver-owned IDs.

The capabilities documentation and bundled Pydantic AI agent skill describe the lifecycle order, adaptive-selection helper, fallback composition, and current deferred/durable limitations.

### Checklist


docs/capabilities.md
```diff
@@ -1159,6 +1159,42 @@ print(result.output)
...
+[`get_model`][pydantic_ai.capabilities.AbstractCapability.get_model] lets a capability supply the [model](models/overview.md) for the agent. It's the one hook that can provide a model an agent doesn't otherwise have, because the model is resolved during run setup before any per-request hook runs:
+
+```python {title="model_from_capability.py"}
+from dataclasses import dataclass
+
+from pydantic_ai import Agent
+from pydantic_ai.capabilities import AbstractCapability
+from pydantic_ai.models import KnownModelName, Model
+
+
```
adtyavrdhn, line 1257: We should explain what happens if multiple capabilities contribute a model

pydantic_ai_slim/pydantic_ai/capabilities/abstract.py
```diff
@@ -293,6 +293,29 @@ def get_model_settings(self) -> AgentModelSettings[AgentDepsT] | None:
...
+    def get_model(self) -> Model | KnownModelName | str | None:
+        """Return the model this capability supplies for the agent, or None.
+
+        This is the one way a capability can supply the `model` for an agent that has none, since
+        the model is resolved during run setup before any per-request hook fires. Return a
+        [`Model`][pydantic_ai.models.Model] instance or a model name string (e.g.
+        `'openai:gpt-5.2'`); strings flow through the same inference path as
+        [`Agent(model=...)`][pydantic_ai.Agent.__init__].
+
+        The first capability (in user order) that returns a non-None model wins, and its model slots
```
adtyavrdhn: Way too verbose lol

DouweM:
Design note for review — interaction with nameless managed capabilities (harness pydantic-ai-harness#331).

The harness managed capabilities are getting an **optional `name`**: when omitted, a capability derives its backing variable from the agent's own `name` (e.g. `ManagedPrompt()` on an agent named `weather_agent` → `prompt__weather_agent`), so the name isn't repeated across every capability. Run-time surfaces can do this today via `RunContext.agent.name` (in `for_run`/`wrap_run`/the `get_instructions`/`get_model_settings` callables/`get_wrapper_toolset`).

The one hole is **`get_model`**: as designed here it takes no args and runs at `iter()` setup before a `RunContext` exists, so a *nameless* `ManagedSettings`/`ManagedAgentSpec` can't derive its variable name there and thus can't source the model for a model-less agent. Three ways to close it:

1. **Give `get_model` a `RunContext`** — invasive: model resolution deliberately happens before the run context is built, so this means restructuring the agent loop. Rejected.
2. **Add `for_agent(agent)` construction-time binding** (the mechanism from #4977) into this PR — works, but pulls in a separate, larger mechanism.
3. **Have `get_model` receive the agent**, i.e. `get_model(*, agent: AbstractAgent) -> Model | KnownModelName | str | None` — cleanest. The agent *is* available at `iter()` setup (it's the one calling `get_model` on its leaves), only the `RunContext` isn't. This mirrors #4977's `resolve_model_id(model_id, *, agent)` signature exactly, needs no new mechanism, and lets a nameless capability read `agent.name` to build its variable and source the model.

Recommendation: **option 3** — it composes with #4977's `resolve_model_id` (same `*, agent` shape) and fully removes the nameless-`get_model` limitation. Until it lands, harness#331 ships a run-time-derivation interim where a nameless `get_model` returns `None` (model override still works via the run-time path; only *sourcing* a model for a model-less + nameless agent needs an explicit name).

Leaving this here rather than changing the signature unilaterally since it's your call at review.

DouweM:
TODO: Give access to `deps` (and look at platform repo where we determine model based on env vars / enterprise / regions)

DouweM:
David's AICA here: I have picked this back up and, after reviewing #3335, #4977, the current agent/model lifecycle, and the platform enterprise-model routing case, I am going to broaden the implementation from the current static setup-time fallback into a complete capability model-selection surface.

Planned shape:

- `get_model()` may return a static `Model`/model ID, a sync or async selector callable, or `None`.
- Dynamic selectors receive a dedicated `ModelSelectionContext` containing the agent, actual run deps, lower-precedence/default model, run step, and the other safe pre-selection fields. A full `RunContext` cannot be used honestly because it requires the model being selected.
- Selection runs once per new logical model-request step, before model-dependent tool/native-tool preparation, settings, message preparation, token counting, instrumentation, and request hooks.
- `override(model=...)`, an explicit run model, and a run-spec model remain authoritative and skip capability selection. Capability layers then follow normal ordering/configuration precedence, with later contributions winning; the constructor model is the fallback.
- A selected string flows through an async, deps-aware, keyword-only `resolve_model_id(..., *, ctx=...)` hook extracted from #4977, together with its callable convenience capability.
- Continuation polling within the same run remains pinned to the already-selected model and does not re-run the selector. Cross-run continuation will use persisted selection provenance where it can be resolved unambiguously and otherwise raise rather than silently switching models.
- Dynamically selected model instances will participate in per-run async lifecycle management, including identity deduplication and cleanup.
- `FallbackModel` remains a model abstraction. Adaptive selection and within-request fallback are separate axes, and the tests will cover selectors returning fallback models.
- Non-run consumers such as `system_prompt_parts()`, MCP sampling setup, and `Agent.__aenter__()` will get explicit semantics rather than an `iter()`-only special case.
- The docs and tests will cover multiple-capability precedence, deps/run-step adaptive selection, streaming parity, settings/profile effects, continuations, lifecycle, resolution, instrumentation, and fallback composition.

The main remaining implementation-level question is how much selection provenance can be reconstructed for arbitrary model instances across a separate resumed run. The agreed safety rule is to reject ambiguous cases. I will keep that behavior narrow and explicit rather than introduce a generic model registry in this PR.

DouweM:
Notes from the #4977 (durability capabilities) side, since this PR extracts `resolve_model_id` / `ModelResolutionContext` / `ResolveModelId` (which #4977 currently carries its own copy of) and #4977 will drop its copy and consume this version after it merges. Flagging what #4977 depends on so this surface composes with durable execution. Mostly good news — one interaction to be aware of.

### Already compatible — please keep ✅

- **`ModelResolutionContext(agent=..., deps=...)`** carries both `agent` and `deps`. #4977 reconstructs exactly this inside the activity/step/task to re-run the resolver chain deps-aware on the worker — that's our per-request-provider-credentials story (it's what replaces the deprecated wrapper agents' `provider_factory`). Keeping `deps` + `agent` on it is the one hard requirement.
- **Resolver signature** `(ctx, model_id) -> Model | None`, sync-or-async, is identical to #4977's, so user resolvers port unchanged.
- **`has_resolve_model_id`** and its "keep the raw model string instead of inferring eagerly" gate are preserved. #4977 relies on this: under durability the model must stay a **string** through run setup so only the string crosses the boundary and the `Model` is built on the worker (see next section). Please keep that deferral.
- **First-non-`None`-wins aggregation** with `CombinedCapability`/`WrapperCapability` delegating: #4977's `BaseDurabilityCapability.resolve_model_id` participates as the `models=` registry backstop (returns the registered instance or `None`), so it needs the chain to keep delegating through wrappers. The PR body says it does; just confirming it's a hard dependency.

### The invariant that matters for durability

Model-id → `Model` **resolution must be deferrable to inside the durable boundary**: #4977 sends only the model-id **string** across the activity/step/task and re-runs the construction-time `resolve_model_id` chain *there*, because (1) the Temporal workflow sandbox can't do the provider I/O a deps-aware resolver may perform, and (2) determinism/checkpointing require that work in the activity, not in replayed workflow code. For the mainstream case (agent has its own model, optionally + a `ResolveModelId` capability) this PR preserves it via the `has_resolve_model_id` string-deferral above — so that path is unaffected. 👍

### `get_model()` + durability (the one interaction to note)

Two coupled reasons `get_model()`-sourced models don't yet work under durability — flagging so this PR doesn't assume they do; both are **#4977-side follow-ups**, and it's fine for that combination to be unsupported initially (I'll track it on #4977):

1. **Registry**: #4977's durability capabilities build their model registry from `agent.model` at `for_agent()` (bind) time and wrap those instances. A model contributed by a capability's `get_model()` (e.g. a model-less agent) isn't in that registry, so it wouldn't be wrapped for durable execution. Wiring `get_model()` into `for_agent` is non-trivial since `get_model()` can be per-run/per-step and isn't always resolvable at bind time.

2. **Resolution locus**: the new `_select_model` graph step (gated on a capability `model_selector`, so it only fires when `get_model()` is in play) does `resolve_model_id(selected)` in the graph — i.e. **workflow-side** under durability. That's the resolution #4977 needs to happen in the boundary. The thing that would make the eventual wiring clean: **keep "select the model id" (produce a string/id — deterministic, workflow-safe) cleanly separable from "resolve the id to a `Model`" (may do I/O — deferrable)**, so a durability capability can let selection run in the graph but take over resolution inside the activity/step/task (the same shape as today's string-deferral, just extended to the `_select_model` path). It already reads as two steps — mostly asking that they stay decoupled and that `_select_model` not hard-resolve when a downstream capability signals it will. A note on which direction you'd prefer (graph leaves the id unresolved for durability vs. a documented seam to intercept) would let me plan the #4977 rebase.

### Per-step selection vs. continuation segments

`get_model()`'s per-step `ModelSelectionContext` callable is "evaluated before each new model request step." Under durability, one model turn can span multiple **continuation segments** (Anthropic `pause_turn` / OpenAI background mode) — each its own activity/step/task — that must all stay on the **same** model (a suspended response continues on the model that produced it). So `run_step` should increment per request step, not per continuation segment, and the selector should key off request steps; #4977 pins the model for a continued turn. Just worth confirming `run_step` semantics so the two notions don't collide.

### Minor / mechanical (my side, no action here)

- This PR makes `ctx` keyword-only (`resolve_model_id(model_id, *, ctx=...)`); #4977 currently calls it positionally and defines its registry-backstop override with a positional `ctx`. I'll adapt the call sites and the override to `*, ctx` on rebase.

Happy to review the reconciliation from the #4977 side once this lands, or to pair on the select/resolve seam if that's the cleaner path.

PR #6341, opened 2026-07-08
feat(`RunContext`): expose the run's `usage_limits` to tools and capabilities

DouweM:
### What

Adds a read-only `usage_limits: UsageLimits | None` field to [`RunContext`][pydantic_ai.tools.RunContext], so tools and capabilities can read the run's [`UsageLimits`][pydantic_ai.usage.UsageLimits] that were already passed to `run` / `run_sync` / `run_stream`. It sits alongside the existing `ctx.usage`, so a budget-aware tool or capability can compute the remaining budget (`limit - usage`) directly.

### How

The limits live on the agent graph's `GraphAgentDeps`, not on the `RunContext` that tools and capabilities see. This threads them through at every `RunContext` construction site:

- `build_run_context` (the central builder for tools, hooks, history processors, streaming, and capability dispatch) reads `ctx.deps.usage_limits`.
- The initial `RunContext` built in `Agent.iter` for `for_run` lifecycle hooks reads the resolved `usage_limits` (which defaults to `UsageLimits()` when the user passes none, so during a run the field reflects the effective limits, never `None`).
- `TemporalRunContext` now serializes `usage_limits` by default (like `usage`), so it survives the activity boundary; the existing exhaustiveness guard test already forces every field to be categorized.
- The synthetic context from `Agent.system_prompt_parts` has no run and correctly leaves the field at its `None` default.

It's read-only by convention (a plain dataclass field like its siblings): mutating it does not change what the run enforces — enforcement stays on `GraphAgentDeps`.

### Motivation

Capabilities currently cannot read the run's limits, so every budget-aware capability has to be configured with a **duplicate** `limits=` param — the `BudgetDisclosure` capability in [pydantic-ai-harness #334](https://github.com/pydantic/pydantic-ai-harness/pull/334) re-declares limits the run already knows, exactly like the older `LimitWarner` (the VStorm duplication pattern). Fixing this at the source lets those capabilities drop the duplicate param and read `ctx.usage_limits`.

### Deliberate scope

No `remaining()` convenience helper — `ctx.usage_limits` alongside `ctx.usage` is enough to compute remaining budget, and keeping the surface minimal avoids committing to a helper shape prematurely. Could be a follow-up if demand appears.

### Tests

- `tests/test_run_context_usage_limits.py`: limits visible in a tool and in a capability `before_run` hook across `run` / `run_sync` / `run_stream`; effective default limits (not `None`) when the user passes none; `None` on a bare/synthetic `RunContext`.
- `tests/test_temporal.py`: `usage_limits` round-trips through Temporal serialization; covered by the existing exhaustiveness guard.

### Checklist


adtyavrdhn:
This is nice, it'll allow me to do some good stuff in dynamic workflows.

PR #6349, opened 2026-07-08
Add Temporal coverage for nested multimodal tool returns

### LINKED ISSUE #6059: Temporal: cover dict-nested multimodal `ToolReturnContent` round-trip across the activity boundary
### Summary

PR #5255 lifts a callable `pydantic.Discriminator` onto the core `ToolReturnContent` type alias (`messages.py`) so that multimodal items (`BinaryContent`, `FileUrl` subclasses, `UploadedFile`) **nested inside a mapping** re-hydrate to their subclasses on load instead of being left as plain dicts by smart-union (closes #5811).

Because `ToolReturnContent` is also embedded in the Temporal durable-exec activity return type — `durable_exec/temporal/_toolset.py` defines `_ToolReturnResult = Annotated[ToolReturn | ToolReturnContent, Discriminator(...)]` → `CallToolResult`, the return type of `call_tool_activity` — the same discriminator now governs how a multimodal tool return deserializes when it crosses the Temporal workflow/activity boundary (validated by `temporalio.contrib.pydantic.pydantic_data_converter`).

### What's already covered

`tests/test_temporal.py::test_multimodal_content_serialization_in_workflow` already round-trips a tool returning a **top-level list** `['test', BinaryImage, DocumentUrl]` through a real `Worker`, and snapshots the reloaded `ToolReturnPart.content` as `BinaryContent`/`DocumentUrl` subclasses. So list/top-level multimodal tool returns through Temporal are covered.

### The gap

There is no Temporal workflow test for a multimodal item **nested inside a mapping** returned from a tool (e.g. `{'caption': 'see attached', 'attachment': BinaryImage(...)}`) — which is precisely the shape the new discriminator was added to fix. The top-level-list case worked even before the discriminator (list items validate against the union member); the dict-nested case is the one the discriminator newly fixes, and it crosses the Temporal boundary untested.

### Verification (it works today — this is a coverage gap, not a bug)

Proxying the Temporal data-converter with the actual activity return type confirms dict-nested multimodal re-hydrates correctly post-#5255:

```python
from pydantic import TypeAdapter
from pydantic_ai.durable_exec.temporal._toolset import CallToolResult, _ToolReturn
from pydantic_ai.messages import BinaryContent

ta = TypeAdapter(CallToolResult)
img = BinaryContent(data=b'\x00\x01\x02', media_type='image/png')
wrapped = _ToolReturn(result={'caption': 'see attached', 'attachment': img})
reloaded = ta.validate_python(ta.dump_python(wrapped, mode='json'))
# reloaded.result == {'caption': 'see attached', 'attachment': BinaryContent(...)}  ✅ subclass restored
```

### Proposed work

Add a `test_temporal.py` workflow test mirroring `test_multimodal_content_serialization_in_workflow` with a tool that returns a **mapping** containing a nested `BinaryContent`/`FileUrl`, asserting the reloaded `ToolReturnPart.content` keeps the multimodal subclass rather than a plain dict. This pins the discriminator's behavior across the real Temporal activity boundary (the `TypeAdapter` proxy above is a stand-in, not the worker path).

Out of scope for #5255 (UI-adapter PR); filing separately so the Temporal coverage lands with the durable-exec harness.

lxingy3:
﻿Summary
- add a Temporal workflow test for dict-nested multimodal tool return content
- assert BinaryImage and DocumentUrl survive the activity boundary inside a mapping

Tests
- `python -m ruff format --check tests/test_temporal.py`
- `python -m ruff check tests/test_temporal.py`
- `pyright tests/test_temporal.py`
- `python -m pytest tests/test_temporal.py::test_nested_multimodal_tool_return_survives_temporal -q`

Fixes #6059

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6349?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/tests/test_temporal.py
+++ b/tests/test_temporal.py
@@ -4686,6 +4686,67 @@ async def test_multimodal_content_serialization_in_workflow(client: Client):
         ]
 
 
+nested_multimodal_tool_return_agent = Agent(TestModel(), name='nested_multimodal_tool_return_agent')
+
+
+@nested_multimodal_tool_return_agent.tool
+def get_nested_multimodal_content(ctx: RunContext) -> dict[str, str | MultiModalContent]:
+    """Return multimodal content nested inside a mapping."""
+    return {
+        'caption': 'see attached',
+        'attachment': BinaryImage(data=b'\x89PNG', media_type='image/png'),
+        'source': DocumentUrl(url='https://example.com/doc/12345', media_type='application/pdf'),
+    }
+
+
+nested_multimodal_tool_return_temporal_agent = TemporalAgent(
+    nested_multimodal_tool_return_agent, activity_config=BASE_ACTIVITY_CONFIG
+)
+
+
+@workflow.defn
+class NestedMultiModalToolReturnWorkflow:
+    @workflow.run
+    async def run(self, prompt: str) -> list[ModelMessage]:
+        result = await nested_multimodal_tool_return_temporal_agent.run(prompt)
+        return result.all_messages()
+
+
+async def test_nested_multimodal_tool_return_survives_temporal(client: Client):
+    """Nested multimodal values in tool returns survive the Temporal activity boundary."""
+    async with Worker(
+        client,
+        task_queue=TASK_QUEUE,
+        workflows=[NestedMultiModalToolReturnWorkflow],
+        plugins=[AgentPlugin(nested_multimodal_tool_return_temporal_agent)],
+    ):
+        messages = await client.execute_workflow(
+            NestedMultiModalToolReturnWorkflow.run,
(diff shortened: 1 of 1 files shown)


hramezani (approved):

hramezani:
I am investigating

PR #6352, opened 2026-07-08
fix: Shorthand azure-responses:[model-id] support

### LINKED ISSUE #6335: Shorthand azure-responses:[model-id] support
### Description

It would be great if there were a shorthand for using the Azure Responses API.

Similar to the existing `openai-responses:[model-id]` shorthand, it would be useful to have an Azure equivalent, for example:

`Agent("azure-responses:gpt-5.5")`

This would provide a more concise way to use OpenAIResponsesModel with AzureProvider, without requiring users to manually instantiate the model and provider for what is otherwise a common configuration.

### References

_No response_

Phoenix1454:
Here's a concise GitHub PR description:

### Description

Adds shorthand `azure-responses:[model-id]` support for Azure Responses API.

### Problem
Current configuration requires manual instantiation of model and provider, which can be cumbersome.

### Root cause
Lack of shorthand syntax for Azure Responses API.

### Fix
Introduces `azure-responses:[model-id]` shorthand, mirroring existing `openai-responses:[model-id]` syntax.

Fixes #6335

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6352?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/examples/pydantic_ai_examples/evals/models.py
+++ b/examples/pydantic_ai_examples/evals/models.py
@@ -3,7 +3,6 @@ from __future__ import annotations as _annotations
 from pydantic import AwareDatetime, BaseModel
 from typing_extensions import TypedDict
 
-
 class TimeRangeBuilderSuccess(BaseModel, use_attribute_docstrings=True):
     """Response when a time range could be successfully generated."""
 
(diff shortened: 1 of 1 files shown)


hramezani:
Closing this PR as it doesn't address #6335.

The issue asks for a real `azure-responses:[model-id]` shorthand so that `Agent("azure-responses:gpt-5.5")` resolves to an `OpenAIResponsesModel` backed by an `AzureProvider`, mirroring the existing `openai-responses:` handling.

This PR instead only adds unrelated stub classes (`AzureResponsesModel`, a fake `AzureProvider`, and an `Agent` class that shadows the real one) to `examples/pydantic_ai_examples/evals/models.py`. Those stubs are never wired into model inference, so the shorthand does not actually work.

A real fix would live in `infer_model` (`pydantic_ai_slim/pydantic_ai/models/__init__.py`) and provider inference (`pydantic_ai_slim/pydantic_ai/providers/__init__.py`), plus `KnownModelName`, docs, and tests. Happy to see a follow-up along those lines.

Phoenix1454:
orry for the noise on this one looking back at it, the fix clearly needed to go into infer_model and infer_provider_class to map azure-responses to OpenAIResponsesModel backed by AzureProvider, not stub classes in the examples file. I can see @hramezani implemented it correctly in #6355. Apologies for the misdirected PR.

PR #6354, opened 2026-07-08
Add `ToolOutput(buffered=True)` for incremental output tools

### LINKED ISSUE #4723: New end_strategy='review' — let the model review and patch output before finalizing
### Description

# Problem

Today, output_type tools are terminal — when the model calls final_result, the data is validated and the loop ends immediately. The model never sees its own structured output and has no opportunity to self-correct. For extraction tasks (documents, forms, etc.), this is a significant quality gap: schema validation passing doesn't mean the data is correct.
The current workarounds are:
- Output validators — can reject and retry, but the model resubmits the entire schema blind rather than seeing what it got wrong
- Regular tools for draft/patch + function-as-output-type for confirm — works, but requires the user to manually wire up state management in deps, draft tools, and a JSON patch tool for every extraction agent

# Proposal

Add end_strategy='review' (name open to bikeshedding). When set, the output tool behaves differently:
Model calls final_result(MyModel(...))
  → schema validation runs (as today)
  → if invalid: retry prompt (as today)
  → if valid: result stored in state, serialized back to model as tool result
    → model sees its extraction and can reason about correctness
    → model can call an auto-generated `patch_result` tool (JSON Patch RFC 6902)
    → model calls `confirm_result` (or similar) to finalize
    → loop ends with the validated + patched result
```
Pseudocode for how this differs from 'early':
# early (current): valid output → FinalResult → End
# review (proposed): valid output → store in state → ToolReturn → model reviews
#                    → optional patch → confirm → FinalResult → End
Agent(
    'openai:gpt-4o',
    output_type=MyModel,
    end_strategy='review',  # that's it — no manual tool wiring needed
)
```

What the framework would auto-generate
When end_strategy='review', the framework would automatically create (alongside the existing output tool):
1. A patch_result tool accepting list[JsonPatchOp] — applies patches to the stored result, re-validates against the output type, returns the updated data to the model
2. A confirm_result tool (or reuse the output tool with empty args) — moves the stored result into FinalResult and ends the loop
The model's output tool (final_result) would become non-terminal in this mode: instead of producing End(FinalResult(...)), it would produce a ToolReturnPart containing the serialized result, giving the model a chance to review.
Use case
Document/form extraction (invoices, contracts, medical records, etc.) where:
- The schema validates fine but the model misread a value, swapped two fields, or hallucinated a default
- Targeted JSON patches are cheaper and less error-prone than full re-extraction
- The model is the best judge of whether its extraction matches the source
Prior art

---
References:
- Existing EndStrategy definition: _agent_graph.py:63
- Output tool processing: _agent_graph.py:972-1064
- EndRun comment: result.py:481
- RFC 6902 JSON Patch: https://datatracker.ietf.org/doc/html/rfc6902

### References

_No response_

kazmer97:
- Closes #4723

### Summary

- Adds `ToolOutput(..., buffered=True)` for tool-output agents that need draft/patch/submit behavior.
- Keeps the normal output tool name, schema, typing, and final validation path inherited from `output_type`.
- Routes buffering per output tool, so mixed output specs can opt in only the structured outputs that should be buffered while leaving text, deferred output, and other output tools unchanged.
- Exposes generated `read_*_buffer` and `patch_*_buffer` tools for inspecting/updating the draft, and treats an empty call to the same buffered output tool as submit.
- Updates output docs, the bundled Pydantic AI agent-building skill, and focused tests.

### Notes for maintainers

This is intentionally still a draft PR for API and internals discussion. The design has moved away from a new agent-level `BufferedOutputStrategy` public API. The opt-in now lives on `ToolOutput`, which keeps the capability attached to the output tool that advertises the final result shape.

The main behavior to review is that buffered output widens only the opted-in output tool schema to also allow an empty object as the submit signal. Staged and patched buffer contents are still validated through normal output-tool validation and final submit still runs through the normal output pipeline.

I have not claimed streaming support in this pass; the implementation and tests target normal `run`/`run_sync` tool execution.

### Validation

- `UV_CACHE_DIR=/tmp/uv-cache uv run pytest tests/test_agent.py::test_buffered_tool_output_patches_and_submits_buffer tests/test_agent.py::test_buffered_tool_output_ignores_deferred_tool_requests_output_type tests/test_agent.py::test_buffered_tool_output_inherits_multiple_output_tool_names -vv`
- `UV_CACHE_DIR=/tmp/uv-cache uv run ruff check pydantic_ai_slim/pydantic_ai/output.py pydantic_ai_slim/pydantic_ai/_output.py pydantic_ai_slim/pydantic_ai/_agent_graph.py pydantic_ai_slim/pydantic_ai/_tool_execution.py pydantic_ai_slim/pydantic_ai/agent/__init__.py pydantic_ai_slim/pydantic_ai/__init__.py tests/test_agent.py`
- `PYRIGHT_PYTHON_IGNORE_WARNINGS=1 UV_CACHE_DIR=/tmp/uv-cache uv run pyright pydantic_ai_slim/pydantic_ai/output.py pydantic_ai_slim/pydantic_ai/_output.py pydantic_ai_slim/pydantic_ai/_agent_graph.py pydantic_ai_slim/pydantic_ai/_tool_execution.py pydantic_ai_slim/pydantic_ai/agent/__init__.py pydantic_ai_slim/pydantic_ai/__init__.py tests/test_agent.py`

### Checklist


kazmer97:
Updated this draft after the API-design discussion:

- removed the proposed agent-level `BufferedOutputStrategy` public API
- moved the opt-in to `ToolOutput(..., buffered=True)`
- made buffering per output tool, so mixed output specs can buffer only selected structured outputs
- left final submit on the normal output-tool validation/processing path

The PR body has been refreshed with the current shape and focused validation commands.

PR #6357, opened 2026-07-08
Allow `override(model=...)` on an agent with no model set

### LINKED ISSUE #6316: Why does overriding an unset model not set the model?
### Question

For reasons I can't quite understand overriding the model fails when the original model is None. This is the responsible code

https://github.com/pydantic/pydantic-ai/blob/49fa9daf8923326b83b574b12f4d62fc8c57fe4a/pydantic_ai_slim/pydantic_ai/agent/__init__.py#L2373-L2377

This surprised me a bit. I stumbled upon the error when I was explicitly trying to separate the agent definition from the model definition using something like

```py
with agent.override(model=model):
    await agent.to_cli()
```
but this failed because the agent had no model (by design). I managed to work around the issue with the following set up. This works but I still don't understand *why* the error exists and why it might not be a good idea to work around it.

```py
with agent.override(model=model or agent.model): # type: ignore
    if agent.model is None:
        agent.model = NotNone # type: ignore
    await agent.to_cli()
    if agent.model is NotNone:
        agent.model = None
```

### Additional Context

_No response_

hramezani:
Closes #6316

## Problem

`override(model=...)` raised a `UserError` when the agent had no model set (neither on `Agent(...)` nor passed to the run method), even though the override itself supplies a perfectly usable model:

```py
agent = Agent()  # no model, by design
with agent.override(model=model):
    await agent.to_cli()  # UserError: `model` must either be set on the agent or included when calling it.
```

The guard existed to stop a test-time `override` from masking the fact that an agent was never given a model. But as discussed in the issue, `override` is documented as "particularly useful when testing", **not** "only to be used in testing". Deliberately defining a model-less agent and injecting the model exclusively via `override` is a legitimate pattern, and this guard forced an ugly sentinel workaround.

## Change

Drop the guard in `Agent._get_model`. When an override model is present it is simply used. The genuine "no model set anywhere" mistake is still caught by the existing non-override path, so forgetting a model entirely still raises `UserError` as before.

## Tests

`test_override_model_no_model` previously asserted the `UserError` was raised; it now asserts the override succeeds and produces output. The no-model-anywhere error path remains covered by `test_model_arg`.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6357?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/agent/__init__.py
+++ b/pydantic_ai_slim/pydantic_ai/agent/__init__.py
@@ -2404,12 +2404,6 @@ class Agent(AbstractAgent[AgentDepsT, OutputDataT]):
         """
         model_: models.Model
         if some_model := self._override_model.get():
-            # we don't want `override()` to cover up errors from the model not being defined, hence this check
-            if model is None and self.model is None:
-                raise exceptions.UserError(
-                    '`model` must either be set on the agent or included when calling it. '
-                    '(Even when `override(model=...)` is customizing the model that will actually be called)'
-                )
             model_ = some_model.value
         elif model is not None:
             model_ = models.infer_model(model)
--- a/tests/test_agent.py
+++ b/tests/test_agent.py
@@ -4242,11 +4242,11 @@ def test_set_model(env: TestEnv):
 
 
 def test_override_model_no_model():
-    agent = Agent()
+    agent = Agent(output_type=tuple[int, str])
 
-    with pytest.raises(UserError, match=r'`model` must either be set.+Even when `override\(model=...\)` is customiz'):
-        with agent.override(model='test'):
-            agent.run_sync('Hello')
+    with agent.override(model='test'):
+        result = agent.run_sync('Hello')
+    assert result.output == snapshot((0, 'a'))
 
 
 async def test_agent_name():


adtyavrdhn (approved):
LGTM

godququ5-code:
Nice, this covers the first action item from #6316 (`override(model=...)` on an agent with no default model).

One small scope note: DouweM also called out a second action item in the issue — adding `model` (and other run arguments) to `to_cli()` for completeness. I had opened #6358 with that additional `to_cli(model=...)` propagation and regression coverage, but the duplicate-PR bot closed it because this PR already references #6316.

I’m happy either way:
- keep this PR focused on the `override()` behavior and do `to_cli(model=...)` as a follow-up after this lands, or
- if maintainers prefer one PR for both action items, the relevant `to_cli()` changes/tests are in #6358 and can be folded in.

No need to block this PR on my side; just wanted to flag the remaining issue scope so it doesn’t get lost.

PR #6363, opened 2026-07-08
fix: skip empty ModelResponse in Groq and HuggingFace message history

### LINKED ISSUE #6364: Groq and HuggingFace adapters send an invalid empty assistant message (400) on the empty-response retry
### Summary

`GroqModel._map_messages` and `HuggingFaceModel._map_messages` append an assistant message for every `ModelResponse`, including an empty one with no content and no tool calls. A bare `{'role': 'assistant'}` is rejected by the Chat Completions API with a 400.

This breaks the empty-response retry flow: when a model returns an empty response, the agent graph records `ModelResponse(parts=[])` and retries by appending a `RetryPromptPart`, relying on the adapter to omit the empty response from the next request. On the retry, `_map_messages` turns the empty response into the invalid assistant message, so the request 400s and the self-correction crashes instead of recovering.

The OpenAI, Anthropic, Mistral, and Cohere adapters already guard this (OpenAI's `_into_message_param` returns `None`; Mistral and Cohere were added in #6302, "mirroring the OpenAI and Anthropic adapters", tracked from #6138). Groq and HuggingFace have their own `_map_messages` and were missed by that sweep.

### Reproduction

Mapping a history that contains an empty `ModelResponse` (what the graph produces on an empty-response retry) yields the invalid message:

```python
from pydantic_ai.messages import ModelRequest, ModelResponse, UserPromptPart
from pydantic_ai.models.groq import GroqModel
from pydantic_ai.providers.groq import GroqProvider

history = [
    ModelRequest(parts=[UserPromptPart(content='hi')]),
    ModelResponse(parts=[]),                                # empty response the graph retries
    ModelRequest(parts=[UserPromptPart(content='retry')]),
]
model = GroqModel('llama-3.3-70b-versatile', provider=GroqProvider(api_key='x'))
mapped = await model._map_messages(history, ModelRequestParameters())
# mapped contains a {'role': 'assistant'} entry with no content and no tool_calls,
# which the Groq/HuggingFace Chat Completions API rejects with a 400 on the retry request.
```

The OpenAI adapter, given the same history, omits the empty response.

### Expected

An assistant message with neither content nor tool calls is omitted from the request payload, matching the OpenAI, Anthropic, Mistral, and Cohere adapters, so the empty-response retry can recover.

### Environment

pydantic-ai `main`; affects the `groq` and `huggingface` extras.

---

A fix with regression tests for both adapters is ready in #6363 (closed by the issue-first bot); happy to reference this issue and reopen it.

AmirF194:
Fixes #6364

## Problem

`GroqModel._map_messages` and `HuggingFaceModel._map_messages` append an assistant message for every `ModelResponse`, even an empty one with no content and no tool calls. A bare `{'role': 'assistant'}` is rejected by the Chat Completions API with a 400.

This breaks the empty-response retry flow. When a model returns an empty response, the agent graph records `ModelResponse(parts=[])` and retries by appending a `RetryPromptPart`, relying on the model adapter to omit the empty response from the next request. On the retry, `_map_messages` turns the empty response into the invalid assistant message and the request 400s, so the self-correction crashes instead of recovering.

This is the same bug (and the same fix) as the already-merged sweep for the other Chat Completions adapters: #6138 (OpenAI) and #6302 (Mistral + Cohere, "mirroring the OpenAI and Anthropic adapters"). Groq and HuggingFace have their own `_map_messages` and were missed by that sweep, so they still lack the guard.

## Fix

Skip an assistant message that has neither content nor tool calls, matching the OpenAI (`_into_message_param` returns `None`), Anthropic, Mistral, and Cohere adapters.

```python
if not texts and not tool_calls:
    continue
```

(`# noqa: C901` is added to each `_map_messages` because the one extra branch pushes the existing function one over the complexity limit; this matches the `# noqa: C901` convention used across the codebase for these mapping functions.)

## Tests

Adds `test_empty_response_skipped_in_history` to `test_groq.py` and `test_huggingface.py`, mirroring the existing OpenAI test: an empty response is followed by a valid one, and the retry request payload must contain no assistant message. The Groq test needed request-kwargs recording on `MockGroq` (a few lines), which `MockHuggingFace` and `MockOpenAI` already have.

## Verification

Clean container (`uv sync --frozen --all-extras --all-packages`):

- Both new tests fail on `main` (the empty response is sent as `{'role': 'assistant'}`) and pass with the fix.
- The offline (`-m "not vcr"`) suites for both files stay green (26 passed).
- `ruff check` and `ruff format --check` on the changed files pass.

## Notes

- No public API change and no breaking change; only an empty, API-invalid message is dropped from the request payload.
- This change was prepared with AI assistance and reviewed and verified as described above. Happy to move it to an issue first if you'd prefer that route for the two remaining adapters.

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/groq.py
+++ b/pydantic_ai_slim/pydantic_ai/models/groq.py
@@ -527,7 +527,7 @@ class GroqModel(Model[AsyncGroq]):
                 )
         return tools, search_settings
 
-    async def _map_messages(
+    async def _map_messages(  # noqa: C901
         self, messages: list[ModelMessage], model_request_parameters: ModelRequestParameters
     ) -> list[chat.ChatCompletionMessageParam]:
         """Just maps a `pydantic_ai.Message` to a `groq.types.ChatCompletionMessageParam`."""
@@ -558,6 +558,12 @@ class GroqModel(Model[AsyncGroq]):
                         pass
                     else:
                         assert_never(item)
+                if not texts and not tool_calls:
+                    # An assistant message must carry content or tool calls; a bare
+                    # `{'role': 'assistant'}` is rejected by the Chat Completions API. Skip an empty
+                    # `ModelResponse` (e.g. one the agent graph recorded before retrying) rather than
+                    # send it, matching the OpenAI, Anthropic, Mistral, and Cohere adapters.
+                    continue
                 message_param = chat.ChatCompletionAssistantMessageParam(role='assistant')
                 if texts:
                     # Note: model responses from this model should only have one text item, so the following
--- a/pydantic_ai_slim/pydantic_ai/models/huggingface.py
+++ b/pydantic_ai_slim/pydantic_ai/models/huggingface.py
@@ -377,7 +377,7 @@ class HuggingFaceModel(Model[AsyncInferenceClient]):
         tools = [HuggingFaceModel._map_tool_definition(r) for r in tool_defs.values()]
         return tools, tool_choice
 
-    async def _map_messages(
+    async def _map_messages(  # noqa: C901
         self, messages: list[ModelMessage], model_request_parameters: ModelRequestParameters
     ) -> list[ChatCompletionInputMessage | ChatCompletionOutputMessage]:
         """Just maps a `pydantic_ai.Message` to a `huggingface_hub.ChatCompletionInputMessage`."""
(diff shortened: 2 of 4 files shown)


AmirF194:
Reopened as #6365, now referencing #6364 (opened a bug report with a repro per the issue-first policy). This one was auto-closed before it had an issue reference; the fix and tests are unchanged.

PR #6367, opened 2026-07-08
Reduce Google streamed response overhead

### LINKED ISSUE #6396: Reduce overhead in Google streamed response handling
### Description

`GeminiStreamedResponse._get_event_iterator` does repeated per-chunk attribute lookups and small dict copies while processing Google streamed responses. This sits on the per-stream-chunk path, so the overhead compounds on longer streams.

I have an existing patch in #6367 that keeps the current streaming behavior unchanged while reducing this overhead by:

- caching stable instance attributes locally during iteration
- updating provider details in place instead of repeatedly rebuilding small dicts
- passing only the fields needed by `RequestUsage.extract`

### Measured impact

Focused evaluator, repeated runs:

- upstream: `2127.071475 us`, `2136.656975 us`
- patched branch: `2060.2042125 us`, `2049.346975 us`

This is a modest but measurable improvement on the Google streaming path.

### Validation

The evaluator checks:

- text chunk merging
- file-search deferred returns
- empty grounding flush behavior
- code execution parts
- provider metadata updates across later stream chunks:
  - `provider_response_id`
  - `service_tier`
  - `traffic_type`

Existing focused tests also pass:

- `tests/models/google/test_native_tools.py`
- `tests/models/google/test_code_execution.py`

### Existing patch

Closed PR with the patch: #6367

Autoresearch / optimization run:
https://dashboard.weco.ai/share/Aj5ACG8Fya3OT_IYxHCkhWA8XEjQQcp3

dexhunter:
## Summary

This reduces overhead in Google streamed response handling by avoiding repeated per-chunk attribute lookups and dict copies in `GeminiStreamedResponse._get_event_iterator`, while preserving the existing stream control flow and `uuid4()` vendor part IDs.

It also narrows the usage extraction payload so `_metadata_as_usage` passes only the response model version and usage metadata into `RequestUsage.extract` instead of dumping the whole response object slice.

This closes #6396.

Supplementary autoresearch: [public dashboard](https://dashboard.weco.ai/share/Aj5ACG8Fya3OT_IYxHCkhWA8XEjQQcp3).

## Validation

I used a focused evaluator that checks text chunk merging, file-search deferred returns, empty grounding flush behavior, code execution parts, and provider metadata updates across later stream chunks (`provider_response_id`, `service_tier`, and `traffic_type`). The final patch uses the conservative implementation that passes this strengthened metadata gate.

Strengthened metric repeats:

- Upstream: `2127.071475 us`, `2136.656975 us`
- This branch: `2060.2042125 us`, `2049.346975 us`

Checks run:

- `uvx ruff check --fix pydantic_ai_slim/pydantic_ai/models/google.py`
- `git diff --check`
- `python -m pytest tests/models/google/test_native_tools.py -q` (`9 passed`)
- `python -m pytest tests/models/google/test_code_execution.py -q` (`2 passed, 1 skipped`)

Fixes #6396.

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/google.py
+++ b/pydantic_ai_slim/pydantic_ai/models/google.py
@@ -1258,39 +1258,48 @@ class GeminiStreamedResponse(StreamedResponse):
                 raise
 
     async def _get_event_iterator(self) -> AsyncIterator[ModelResponseStreamEvent]:  # noqa: C901
+        stream_provider_details = self.provider_details
         if self._provider_timestamp is not None:
-            self.provider_details = {'timestamp': self._provider_timestamp}
+            stream_provider_details = {'timestamp': self._provider_timestamp}
+            self.provider_details = stream_provider_details
+        provider_name = self._provider_name
+        provider_url = self._provider_url
+        parts_manager = self._parts_manager
+
         try:
             async for chunk in self._response:
-                self._usage = _metadata_as_usage(chunk, self._provider_name, self._provider_url, self._usage)
+                self._usage = _metadata_as_usage(chunk, provider_name, provider_url, self._usage)
 
                 if (
                     chunk.sdk_http_response
                     and chunk.sdk_http_response.headers
                     and (service_tier := chunk.sdk_http_response.headers.get('x-gemini-service-tier'))
                 ):
-                    self.provider_details = {**(self.provider_details or {}), 'service_tier': service_tier.lower()}
+                    if stream_provider_details is None:
+                        stream_provider_details = {}
+                        self.provider_details = stream_provider_details
+                    stream_provider_details['service_tier'] = service_tier.lower()
 
                 # Capture traffic_type before the candidates guard, since usage_metadata
                 # may be present on chunks without candidates.
                 if chunk.usage_metadata and chunk.usage_metadata.traffic_type:
-                    self.provider_details = {
-                        **(self.provider_details or {}),
-                        'traffic_type': chunk.usage_metadata.traffic_type.value,
-                    }
+                    if stream_provider_details is None:
+                        stream_provider_details = {}
+                        self.provider_details = stream_provider_details
+                    stream_provider_details['traffic_type'] = chunk.usage_metadata.traffic_type.value
 
                 if not chunk.candidates:
                     if chunk.prompt_feedback and chunk.prompt_feedback.block_reason:
                         self._has_content_filter = True
                         block_reason = chunk.prompt_feedback.block_reason
-                        self.provider_details = {
-                            **(self.provider_details or {}),
-                            'block_reason': block_reason.value,
-                        }
+                        if stream_provider_details is None:
+                            stream_provider_details = {}
+                            self.provider_details = stream_provider_details
+                        stream_provider_details['block_reason'] = block_reason.value
                         if chunk.prompt_feedback.block_reason_message:
-                            self.provider_details['block_reason_message'] = chunk.prompt_feedback.block_reason_message
+                            stream_provider_details['block_reason_message'] = chunk.prompt_feedback.block_reason_message
                         if chunk.prompt_feedback.safety_ratings:
-                            self.provider_details['safety_ratings'] = [
+                            stream_provider_details['safety_ratings'] = [
                                 r.model_dump(by_alias=True) for r in chunk.prompt_feedback.safety_ratings
                             ]
                         self.finish_reason = 'content_filter'
@@ -1305,13 +1314,13 @@ class GeminiStreamedResponse(StreamedResponse):
 
                 raw_finish_reason = candidate.finish_reason
                 if raw_finish_reason and not self._has_content_filter:
-                    self.provider_details = {
-                        **(self.provider_details or {}),
-                        'finish_reason': raw_finish_reason.value,
-                    }
+                    if stream_provider_details is None:
+                        stream_provider_details = {}
+                        self.provider_details = stream_provider_details
+                    stream_provider_details['finish_reason'] = raw_finish_reason.value
 
                     if candidate.safety_ratings:
-                        self.provider_details['safety_ratings'] = [
+                        stream_provider_details['safety_ratings'] = [
                             r.model_dump(by_alias=True) for r in candidate.safety_ratings
                         ]
 
@@ -1342,11 +1351,11 @@ class GeminiStreamedResponse(StreamedResponse):
                 # tool_call/tool_response parts *instead of* the metadata, never both.
                 if not self._has_tool_invocations:
                     web_fetch_call, web_fetch_return = _map_url_context_metadata(
-                        candidate.url_context_metadata, self.provider_name
+                        candidate.url_context_metadata, provider_name
                     )
                     if web_fetch_call and web_fetch_return:
-                        yield self._parts_manager.handle_part(vendor_part_id=uuid4(), part=web_fetch_call)
-                        yield self._parts_manager.handle_part(vendor_part_id=uuid4(), part=web_fetch_return)
+                        yield parts_manager.handle_part(vendor_part_id=uuid4(), part=web_fetch_call)
+                        yield parts_manager.handle_part(vendor_part_id=uuid4(), part=web_fetch_return)
 
                 if candidate.content is None or candidate.content.parts is None:
                     continue
@@ -1356,7 +1365,10 @@ class GeminiStreamedResponse(StreamedResponse):
                     continue  # pragma: no cover
 
                 if not self._has_tool_invocations:
-                    self._has_tool_invocations = _has_native_tool_invocations(parts)
+                    for response_part in parts:
+                        if response_part.tool_call or response_part.tool_response:
+                            self._has_tool_invocations = True
+                            break
 
                 for part in parts:
                     provider_details: dict[str, Any] | None = None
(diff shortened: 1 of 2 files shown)


dexhunter:
Added the requested VCR-test rationale to all four provider-details regression tests in `4697860a`. Focused validation: 4 passed, 185 deselected; Ruff check and format pass. The standalone whole-file Pyright invocation still reports pre-existing unknown-type findings elsewhere in `test_google.py`; this comment-only test change adds no new typing surface.

dsfaccini:

_This comment was posted by Claude Code using claude-opus-4-8 on behalf of David. It is automated and no maintainer has signed off yet._

|---|---|---|---|---|
| 1 · code-correctness | ✅ closed | 0 | 3 | 3 |
| 2 · tests/cassettes/docs | ✅ closed | 0 | 3 | 1 |
| 3 · cross-cutting/siblings | ✅ closed | 0 | 1 | 0 |

**Verdict:** 🔧 REVIEW — 0 blocking, but ~5 non-blocking required actions (code/test quality) before this is merge-ready. The refactor is **behavior-preserving** (verified: identical event stream + final `ModelResponse`; existing #5886/#6216 regression tests + 4 new tests all pass).
Reviewed at 2026-07-21 · diff `870c9906`

<details>
<summary>🔧 6 required (do)</summary>

**Test quality (the weakest part of the PR — coverage is complete and tests pass, but the shape is off):**

**1. New tests reach into private `_get_event_iterator`** · `tests/models/test_google.py` · `test_gemini_streamed_response_initializes_provider_details_*`
Each injects a fake `PeekableAsyncStream` into the private `_response` field (`cast(Any, …)` + `# pyright: ignore[reportPrivateUsage]`) and drives the private iterator. Sibling tests in the **same file** (`test_google_service_tier_streamed_response_extraction`, `test_google_prompt_feedback_streaming`) prove every one of these behaviors is observable through `agent.run_stream(...)` by mocking only the SDK stream boundary — fix: use that public pattern, which also drops the private-access coupling for free.

**2. Unit-test rationale docstring is inaccurate** · `tests/models/test_google.py` · same 4 tests
All four carry `"""Pin internal provider details that a VCR cassette would not expose reliably."""`, which is false: `service_tier` is a response **header** and `traffic_type`/`block_reason` are response **body** fields — all VCR-recordable, and sibling public tests already assert them. The real reason a live cassette is awkward is that FLEX-tier / `PROHIBITED_CONTENT` can't be triggered deterministically against the live API — fix: state the true reason if any unit test survives.

**3. 3 of 4 new tests duplicate existing public coverage** · `tests/models/test_google.py`
`service_tier` → `test_google_service_tier_streamed_response_extraction`; `traffic_type` → `test_google_vertex_service_tier_flex_stream` (real VCR); `block_reason` → `test_google_prompt_feedback_streaming`. Only `_adds_prompt_feedback_to_existing_provider_details` (blocked stream **+** `traffic_type` coexistence — the accumulation the in-place refactor actually risks) is additive, and even it belongs as a case on the existing public test.

**Code quality:**

**4. Inlined `for`/`break` duplicates a still-used documented helper** · `pydantic_ai_slim/pydantic_ai/models/google.py` · `GeminiStreamedResponse._get_event_iterator`
The inline replaced `_has_native_tool_invocations(parts)`, whose body is `any(p.tool_call or p.tool_response for p in parts)` and which is **still defined and used** on the non-streaming path (`:1818`). `any()` short-circuits identically, so the inline saves ~one function call per chunk while reintroducing a divergent copy of a predicate whose docstring documents the duplicate-`NativeToolCallPart` rationale — fix: keep the helper call. (Converged on by 2 independent reviewers.)

**5. Verbose 4× None-guard vs the concise sibling idiom** · `pydantic_ai_slim/pydantic_ai/models/google.py` · `GeminiStreamedResponse._get_event_iterator`
The in-place `provider_details` mutation itself is fine and **sibling-consistent** — `AnthropicStreamedResponse` already does it. But it's expressed here as a 3-line `if … is None: … = {}; self.provider_details = …` block repeated 4×, where the family idiom is one line: `stream_provider_details = self.provider_details = stream_provider_details or {}`. Fix: adopt the concise form (halves each block, ~12 fewer lines in an already-`# noqa: C901` method).

**6. Hand-built usage payload hardcodes SDK aliases** · `pydantic_ai_slim/pydantic_ai/models/google.py` · `_metadata_as_usage`
`response.model_dump(include={'model_version','usage_metadata'}, by_alias=True)` → literal `{'modelVersion': …, 'usageMetadata': metadata.model_dump(by_alias=True)}`. Verified byte-equal (incl. `model_version=None`) and regression-tested, so it's **safe** — but the literal keys re-encode the SDK's `by_alias` mapping and won't track an upstream alias rename. This is the PR's single biggest per-chunk win (~0.56µs), so reverting to the `model_dump(include=…)` form trades that perf for SDK-owned aliases — a **maintainer tradeoff call**, not a clear-cut fix.
</details>

<details>
<summary>✅ 3 verified (precedent-guarded contracts held)</summary>

Two precedent reviewers flagged hardened contracts on this exact surface; I ran the guarding regression tests and all pass, so no change is needed:
- **#5886 `_metadata_as_usage` raw-key contract** — `RequestUsage.extract` derives typed token fields from the raw `usage_metadata` keys; the payload rebuild reproduces them exactly (verified byte-equal + `test_google.py -k usage/cumulative` green).
- **#6216 file_search deferred `PartStartEvent` flush ordering** — the alias-only refactor preserves the load-bearing yield order (`test_native_tools.py` green).
- **Behavioral equivalence** — the emitted `ModelResponseStreamEvent` stream is identical (same `handle_*` calls, same order); the only delta is that a mid-stream `get()` now returns a live-mutating `provider_details` dict rather than a frozen snapshot — but no first-party code reads it mid-stream, it converges to the identical final value, and it matches the Anthropic sibling. Benign.
</details>

<details>
<summary>⚠️ 4 informational (skip)</summary>

**1. Mid-stream `provider_details` snapshot semantics** · `models/google.py` · `_get_event_iterator` — skip: real semantic change (snapshot → live view) but no first-party consumer, benign convergence, sibling-consistent with Anthropic. Residual actionable part is the concise-idiom item above.
**2. `any()` vs manual loop** · `models/google.py` — skip: superseded by required item 4 (keep the helper call).
**3. Perf-vs-readability of the local hoists** · `models/google.py` — skip: caching `self._parts_manager`/`_provider_name`/`_provider_url` to locals is a standard hoist; sub-µs gain on an I/O-bound path is a maintainer judgment call, no mechanical fix.
**4. Tests appended to the provider monolith** · `tests/models/test_google.py` — skip: net-neutral; the whole Gemini-streaming `provider_details` cluster already lives in this file, splitting would fragment.
</details>

**Premise note (this is a perf PR):** the targeted overhead is **real** (per-chunk `{**dict}` copies, `model_dump(include=…)`, repeated property/attr lookups on the streaming hot path) and the change **does** reduce it — but by measurement it's **sub-microsecond per chunk** (~0.56µs from the payload rebuild + ~0.1µs from the dict change), i.e. the PR's own ~3% figure is on a CPU-bound microbenchmark and is negligible against network-bound streaming where chunks arrive milliseconds apart. This is a legitimate `p:4-low` micro-optimization; issue #6396 notes the contributor asked for maintainer alignment before polishing further, and that ask is unanswered.

dsfaccini:

Thanks for digging into this, and for the care that went into the benchmark and the tests. I profiled the patch before deciding, so this is not a drive-by close.

Measured result: about 3% of per-chunk CPU, roughly 1 microsecond per chunk. Google streaming chunks arrive tens of milliseconds apart on a network-bound path, so the total effect is on the order of 0.01% of stream wall time.

The bar we want to hold for optimization PRs is that the change either fixes a user-visible symptom (a reported stall, memory growth, measurable slowness) or comes with a profile showing it targets a real hot spot, and that it does not add brittleness or complexity for a gain that is invisible next to network latency. This one does not clear that bar, and it has costs:

- Roughly half the gain comes from `_metadata_as_usage`, where `response.model_dump(include={'model_version', 'usage_metadata'}, by_alias=True)` becomes a hand-built `{'modelVersion': ..., 'usageMetadata': ...}` payload. It is output-identical today, but those camelCase keys were previously derived from the SDK's model config and are now hardcoded with no test pinning them. That makes the fastest part of the patch more fragile rather than simpler.
- The streaming-loop rewrite mutates `provider_details` in place where `main` rebuilds it (`{**(self.provider_details or {}), ...}`). That is a real behavior change and it is untested: a mid-stream snapshot taken through the public `StreamedRunResult.response` would then alias live stream state instead of being an isolated copy.
- Inlining `_has_native_tool_invocations` into a for/flag/break loop conflicts with `agent_docs/code-simplification.md` (use `any()` instead of for-loops with boolean flags) and with `agent_docs/index.md` on reusing shared helpers instead of maintaining a parallel implementation.
- The 117 added test lines pass unmodified against current `main`, so they characterize pre-existing behavior rather than protecting the change; their practical function is to cover branches the optimization itself introduces. They also land in `tests/models/test_google.py`, one of the legacy monolith test files we are trying to shrink.

The genuinely valuable target is one you came close to. Profiling the Google streaming path shows the dominant per-chunk cost is `_metadata_as_usage` calling `RequestUsage.extract`, which runs full genai-prices extraction (`extract_usage` / `_extract_path`) on every chunk. That is roughly half of the total per-chunk cost, and because Google's usage metadata is cumulative, nearly all of it is redundant work. Hoisting price extraction out of the per-chunk loop is a real win rather than a microsecond one. I have filed #6641 for it with the profiling context. It needs a maintainer decision on the approach (once per stream, on the final chunk, or cached by model id, and whether it generalizes to other providers) before implementation, and you would be very welcome to pick it up once that is settled.

Closing this one. Thanks again for the time you put into it.

PR #6378, opened 2026-07-09
fix: preserve deferred calls and output retries on sibling task exception

### LINKED ISSUE #6370: Bug: deferred tool calls and output retries silently dropped on sibling task exception
## Description

In both `_call_tools` and `_ExhaustiveProcessor._run_strategy`, `_populate_deferred_calls` and `output_retries_increment` are placed after the try/finally block. When a sibling task in the same parallel segment raises a non-CancelledError exception, the deferred calls and retry state are silently lost.

## Steps to reproduce

When multiple tool tasks run in parallel:
1. Task A completes and defers a call (CallDeferred/ApprovalRequired)
2. Task B crashes with RuntimeError
3. The deferred call from Task A is never committed

## Expected behavior

Deferred calls accumulated by completed tasks should survive sibling exceptions, same as how output_parts are already preserved in the finally block.

## Impact

Agent may hang because deferred tool calls are silently dropped, with no way to recover.

isheng-eqi:
Fixes #6370

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6378?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/_tool_execution.py
+++ b/pydantic_ai_slim/pydantic_ai/_tool_execution.py
@@ -1083,12 +1083,12 @@ class _ExhaustiveProcessor(_ToolCallProcessor[DepsT, NodeRunEndT]):
                 for i in executable_indices:  # pragma: no branch
                     if i in function_user_parts:
                         self.output_parts.append(function_user_parts[i])
-                # Preserve deferred calls and output retries on exception.
-                self._populate_deferred_calls(
-                    self.tool_calls,
-                    deferred_calls_by_index=deferred_by_index,
-                    deferred_metadata_by_index=deferred_meta_by_index,
-                    deferred_calls=self.deferred_calls,
-                    deferred_metadata=self.deferred_metadata,
-                )
-                self.ctx.state.output_retries_used += self.output_retries_increment
+            # Always preserve deferred calls and output retries.
+            self._populate_deferred_calls(
+                self.tool_calls,
+                deferred_calls_by_index=deferred_by_index,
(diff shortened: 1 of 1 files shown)


dsfaccini:
David's AICA here: thanks for digging into the deferred-tools concurrency path.

We validated this with a TDD repro and it turns out the change doesn't have the intended effect. The deferred calls aren't lost because of where `_populate_deferred_calls` sits relative to the `finally`. They're lost because when a sibling task raises, the exception propagates out of `_ToolCallProcessor.run()` before `_finalize_deferred()` runs, and `_finalize_deferred()` is the only step that turns accumulated deferred calls into a `DeferredToolRequests`. So moving the population into the `finally` fills `self.deferred_calls`, but that state is still discarded, and `main` and this branch behave identically at the public-API level.

Separately, moving `output_retries_used += output_retries_increment` into the `finally` double-counts it on the retry-exhausted path: the exhausted-retries branch already increments right before `raise r.raise_exc`, and the `finally` now increments again as the exception unwinds.

The real fix is a design decision, so we're taking that to https://github.com/pydantic/pydantic-ai/issues/6370 and will follow up there. Closing this for now, but thanks again for the report and the careful investigation.

PR #6381, opened 2026-07-09
fix(mcp): preserve structured MCP error code/data in tool-call ModelRetry messages

### LINKED ISSUE #6209: MCP tool-call errors flatten McpError to a string, discarding code/data (asymmetric with resource methods)
`direct_call_tool` converts a tool error to `ModelRetry(message=str(e))` (in both the `ToolError` branch and the `ExceptionGroup` branch), discarding the structured `McpError.error.code` and `.data`. The resource methods (`read_resource`, `list_resources`, `list_resource_templates`) instead preserve them via `MCPError.from_mcp_sdk(e)`, which carries `message`, `code`, and `data`. The tool-call path should be symmetric.

### Impact

Downstream consumers can only classify tool failures by substring-matching the human-readable message, which is brittle across SDK/server wording changes.

Concrete case: a client read-timeout surfaces as an `mcp` SDK `McpError` with `code = 408` (`REQUEST_TIMEOUT`, from `mcp/shared/session.py`) and message `"Timed out while waiting for response to ClientRequest. Waited N seconds."`. Because only the message survives the conversion to `ModelRetry`, to treat timeouts as non-retryable we're forced to match that exact phrase rather than branch on `code == 408`.

### Current behavior (main / 2.3.0)

```python
# direct_call_tool
except ToolError as e:
    if self.tool_error_behavior == 'retry':
        raise exceptions.ModelRetry(message=str(e)) from e
    raise
except _utils.BaseExceptionGroup as eg:
    ...
    raise exceptions.ModelRetry(message=str(error)) from eg
```

vs. the resource path, which preserves the structure:

```python
except mcp_exceptions.McpError as e:
    raise MCPError.from_mcp_sdk(e) from e   # keeps message + code + data
```

### Ask

When raising `ModelRetry` (or `ToolError`) from a tool-call failure, surface the underlying `code`/`data` (e.g. attach the structured error, mirroring `MCPError.from_mcp_sdk`) so callers can classify by code instead of parsing the message string.

Related but distinct from #1989, which added error *surfacing* (convert to `ModelRetry` so the agent can self-correct); this is about preserving the *structured code/data* on that conversion.

Confirmed present in 2.3.0 and `main` as of 2026-07-02.

isheng-eqi:
Fixes #6209

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6381?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


adtyavrdhn:
Please respect the issue discussion before filing a PR

PR #6389, opened 2026-07-09
fix: count tool calls in usage even when ModelRetry is raised

### LINKED ISSUE #6388: Bug: usage.tool_calls counter skipped on ModelRetry — tool_calls_limit bypassable
## Description

In _raw_execute(), usage.tool_calls += 1 is placed after the try/except block. When a tool raises ModelRetry (which triggers retry), the exception propagates before the counter is incremented. Tools that repeatedly retry can bypass tool_calls_limit.

## Fix

Move usage.tool_calls += 1 before the try block. The tool is invoked regardless of outcome.

isheng-eqi:
Fixes #6388

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6389?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/tool_manager.py
+++ b/pydantic_ai_slim/pydantic_ai/tool_manager.py
@@ -745,6 +745,7 @@ class ToolManager(Generic[AgentDepsT]):
 
         name = validated.call.tool_name
 
+        usage.tool_calls += 1
         try:
             tool_result = await self.toolset.call_tool(
                 name,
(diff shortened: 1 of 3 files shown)


hramezani:
Thanks for the PR and the careful write-up @isheng-eqi!

To explain the close: this is actually working as intended. `usage.tool_calls` is meant to count **successful tool invocations** only — a call that raises `ModelRetry` did not successfully complete, so by design it doesn't consume the `tool_calls_limit`. Retries are independently bounded by the tool's `max_retries`, so they can't loop unbounded.

Moving the increment before the `try` would count failed/retrying calls against the limit, which is the behavior we specifically want to avoid (and it's why the provider-test snapshots shift here). Appreciate the contribution nonetheless! Closing along with #6388.

PR #6391, opened 2026-07-09
fix: don't replace empty tool_call_id with generated ID

### LINKED ISSUE #6390: Bug: tool_call_id or _generate_tool_call_id swallows empty string from providers
### Bug

In ToolCallPartDelta.as_part() and _apply_to_delta(), tool_call_id or _generate_tool_call_id() uses Python's or operator which treats empty string '' as falsy. Providers that send tool_call_id="" would get their ID silently replaced with a generated one.

### Fix

Use is not None check instead: tool_call_id if tool_call_id is not None else _generate_tool_call_id()

isheng-eqi:
Fixes #6390

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6391?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/messages.py
+++ b/pydantic_ai_slim/pydantic_ai/messages.py
@@ -3115,7 +3115,7 @@ class ToolCallPartDelta:
         return ToolCallPart(
             self.tool_name_delta,
             self.args_delta,
-            self.tool_call_id or _generate_tool_call_id(),
+            self.tool_call_id if self.tool_call_id is not None else _generate_tool_call_id(),
             provider_name=self.provider_name,
             provider_details=self.provider_details,
         )
@@ -3190,7 +3190,7 @@ class ToolCallPartDelta:
             return ToolCallPart(
                 delta.tool_name_delta,
                 delta.args_delta,
-                delta.tool_call_id or _generate_tool_call_id(),
+                delta.tool_call_id if delta.tool_call_id is not None else _generate_tool_call_id(),
                 provider_name=delta.provider_name,
                 provider_details=delta.provider_details,
             )


hramezani:
Thanks for the careful writeup and the playground repro, @isheng-eqi! 🙏 I traced where an empty `tool_call_id` actually flows before it reaches a provider, and I don't think we should fix this by preserving `""` — I believe the current `... or _generate_tool_call_id()` behavior is correct and intended. Sharing the reasoning so it's transparent:

**1. The empty id never reaches the wire.** When a request is assembled, every `tool_call_id` goes through `guard_tool_call_id()`, which is literally `t.tool_call_id or generate_tool_call_id()`. See `openai.py` for both the assistant tool-call echo (`_map_tool_call`, `id=_guard_tool_call_id(t=t)`) and the tool result (`tool_call_id=_guard_tool_call_id(t=part)`). So an empty id is regenerated at request-assembly time regardless of what the part stores internally — the request/response correlation this issue worries about can't actually break on the wire.

**2. Streaming correlation doesn't use `tool_call_id`.** Streaming deltas are matched by `vendor_part_id` (the stream index) in `PartsManager`, not by `tool_call_id`, so preserving `""` in `as_part()` / `_apply_to_delta()` has no effect on how a streamed call is assembled.

**3. It's the documented behavior.** `ToolCallPart.tool_call_id` is documented as *"In case the tool call id is not provided by the model, Pydantic AI will generate a random one."* An empty string is effectively "not provided," and the framework relies on this id being **non-empty and unique** — it's used as a dict key when correlating tool results in the agent graph. Preserving `""` would actually weaken that invariant and let two parallel empty-id calls collide.

For what it's worth, the empty-string case is real (Gemini's OpenAI-compatible endpoint does send `id: ""`, and we already have a passing test for it via `test_compatible_api_with_tool_calls_without_id`), so the observation is a good catch — it just turns out to be handled correctly today.

There *is* one real but minor wart nearby: the non-streaming OpenAI path (`_process_response`) currently **keeps** `""` on the `ToolCallPart`, while streaming and the other providers regenerate it — so the internal representation is inconsistent. I'll open a small separate PR to normalize that toward regenerating (the opposite direction to this PR) so empty ids are handled consistently everywhere.

Given the above I don't think we should merge this one, but thank you again for the thorough investigation!

hramezani:
Closing this — as detailed in the comment above, this is working as intended: empty `tool_call_id`s are regenerated at request-assembly time via `guard_tool_call_id`, streaming correlation is keyed on `vendor_part_id` rather than `tool_call_id`, and an empty id is even used as a deliberate sentinel in some native-tool paths (so we don't want to blanket-normalize it either way). Full writeup on the issue: #6390.

Genuinely appreciate the careful investigation and the playground repro, @isheng-eqi — the empty-string case is real, it's just already handled correctly. 🙏

PR #6395, opened 2026-07-09
fix: use `get_running_loop` instead of deprecated `get_event_loop`

### LINKED ISSUE #1196: `pydantic_graph`: `asyncio.get_event_loop()` is deprecated since python 3.12

gautamkishore:
Fix #1196

Replace asyncio.get_event_loop() (deprecated since 3.12, warns on 3.14+) with asyncio.get_running_loop() and a new_event_loop() fallback, removing the intermediate get_event_loop_policy() call that is also slated for removal in Python 3.16.

The previous approach (using get_event_loop_policy()) was rejected because that function is itself deprecated.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6395?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/_utils.py
+++ b/pydantic_ai_slim/pydantic_ai/_utils.py
@@ -806,15 +806,6 @@ def get_union_args(tp: Any) -> tuple[Any, ...]:
         return ()
 
 
-def get_event_loop() -> asyncio.AbstractEventLoop:
-    try:
-        return asyncio.get_running_loop()
-    except RuntimeError:  # pragma: lax no cover
-        event_loop = asyncio.new_event_loop()
-        asyncio.set_event_loop(event_loop)
-        return event_loop
-
-
 def is_str_dict(obj: Any) -> TypeGuard[dict[str, Any]]:
     """Check if obj is a dict, narrowing the type to `dict[str, Any]`."""
     return isinstance(obj, dict)
--- a/pydantic_ai_slim/pydantic_ai/_utils.py
+++ b/pydantic_ai_slim/pydantic_ai/_utils.py
@@ -808,11 +808,11 @@ def get_union_args(tp: Any) -> tuple[Any, ...]:
 
 def get_event_loop() -> asyncio.AbstractEventLoop:
     try:
-        event_loop = asyncio.get_event_loop()
+        return asyncio.get_running_loop()
     except RuntimeError:  # pragma: lax no cover
         event_loop = asyncio.new_event_loop()
         asyncio.set_event_loop(event_loop)
-    return event_loop
+        return event_loop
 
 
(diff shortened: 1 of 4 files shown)


dsfaccini:
David's AICA here: thanks @gautamkishore, and for the rebase. Quick note on why CI hangs. `get_running_loop()` with a `new_event_loop()` fallback but no cache mints a fresh loop on every `run_sync` call. pydantic-ai holds a persistent httpx client across `run_sync`, so that trips the #748 keep-alive case (client bound to a different loop), which surfaces as the whole test matrix hanging at teardown rather than failing cleanly. The same helper also lives in `pydantic_evals/_utils.py` and still needed migrating. I pushed a module-cached version on top of your commits that preserves loop reuse (checked against #748, warning-free on 3.10 to 3.14) and migrates evals too. The `asyncio.Runner`/portal "proper sync support" shape is intentionally left out of scope here.

dsfaccini:
David's AICA here: thanks again @gautamkishore, and for the rebase. Digging in surfaced that the event-loop swap is harder than it looks. The uncached shape breaks loop reuse (the #748 keep-alive case, which hangs the suite at teardown), and the cached shape I pushed here fixes that but reuses a loop whose default executor can be torn down mid-session (`cannot schedule new futures after shutdown`). Neither is a clean minimal fix, and the robust shape (a held `asyncio.Runner` / anyio portal) is a bigger change.

I've written up the full findings and guidance on https://github.com/pydantic/pydantic-ai/issues/1196 and tagged it `v3` + `needs-discussion` so it gets picked up properly. Closing this for now. Really appreciate the work here.

Issue #6334, opened 2026-07-08
Capability-contributed toolsets have no `id`, so they can't be used with durable execution


DouweM:
## Problem

Toolsets contributed by capabilities via `get_toolset()` are built without an `id`:

- `Capability(tools=[...])` builds `FunctionToolset(tools)` with `id=None`
- `MCP(url=...)` builds its local `MCPToolset(url, ...)` with `id=None`

Temporal requires every leaf toolset to have a unique `id` to register its activities, so combining one of these capabilities with durable execution fails at agent construction:

```python
agent = Agent(
    'openai:gpt-5',
    name='my_agent',
    capabilities=[Capability(tools=[my_tool]), TemporalDurability()],
)
# UserError: Toolsets that are 'leaves' (i.e. those that implement their own tool listing
# and calling) need to have a unique `id` in order to be used with Temporal. ...
```

The error is correct but unactionable: neither `Capability` nor `MCP` exposes a way to set the contributed toolset's `id`. (The same applies on the deprecated `TemporalAgent` wrapper path.) The workaround today is to pass a pre-built toolset with an explicit id, e.g. `Toolset(FunctionToolset(tools, id='billing'))` or `MCP(local=MCPToolset(url, id='docs'))`.

On the DBOS and Prefect paths the same missing `id` means the toolset is silently never swapped for its step/task wrapper, which is arguably worse than the loud Temporal error.

## Proposal

Derive the contributed toolset's `id` from the capability:

- `AbstractCapability.id` already exists, must be unique within a run, and already has a class-name-derived fallback — so `Capability(id='billing', tools=[...])` can stamp `billing` (or a derived form) onto its `FunctionToolset`.
- `MCP` already derives `_resolved_id` from its explicit `id` / native tool id / URL host+slug for the native tool path; `_build_local` can pass the same value to the local `MCPToolset`. (Careful with the `local='https://...'` override path where `url` may be `None` — `_resolved_id`'s assertion doesn't hold there.)

A reasonable place to centralize this is where capability toolsets are extracted/wrapped (`CapabilityOwnedToolset`), so any capability's contributed toolset gets a stable derived id without each capability reimplementing it — but stamping ids onto someone else's toolset instance has side effects (labels, uniqueness validation), so per-capability derivation may be the safer first step.

PR #6406, opened 2026-07-10
Guard against id-less `MCPToolset` on the `DBOSAgent` wrapper path

### LINKED ISSUE #5885: DBOS id-less MCP leaf toolsets collide on the per-run tool-defs cache key
> **Note:** please do not open automated or AI-generated PRs for this issue — it is tracked and will be addressed deliberately by a maintainer.

### Summary

Two **id-less** DBOS MCP leaf toolsets collide on the run-scoped tool-defs cache key `''` introduced in #5883, so the second toolset silently returns the first's tool definitions and never lists its own.

### Mechanism

#5883 keys the per-run cache by `self.id or ''` in `DBOSMCPToolsetBase.get_tools`. DBOS does not require leaf-toolset ids (unlike Temporal, which raises `UserError` for a `None` id). Two id-less MCP toolsets with disjoint tool names pass `CombinedToolset`'s uniqueness check (it dedupes on tool *names*, not toolset *ids*), then share cache key `''` within a run. The second toolset reads the first's cached defs and returns the wrong tools.

### Repro (cache logic, no live DBOS needed)

```python
"""Attack 4: two id-less DBOS MCP toolsets collide on cache key ''.

Simulates the get_tools cache logic from DBOSMCPToolsetBase.get_tools without
needing a live DBOS workflow, to show the per-run cache keyed by `self.id or ''`
makes two id-less toolsets read each other's tool defs.
"""

# Mirror of the cache read/write logic in DBOSMCPToolsetBase.get_tools
class FakeToolset:
    def __init__(self, id, tool_defs, cache_tools=True):
        self._id = id
        self._tool_defs = tool_defs  # {name: tool_def}
        self._cache_tools = cache_tools
        self.fetches = 0

    @property
    def id(self):
        return self._id

    def _fetch(self):
        self.fetches += 1
        return dict(self._tool_defs)

    def get_tools(self, run_cache):
        cache_key = self.id or ''
        if self._cache_tools and (cached := run_cache.get(cache_key)) is not None:
            return dict(cached)
        tool_defs = self._fetch()
        if self._cache_tools:
            run_cache[cache_key] = tool_defs
        return dict(tool_defs)


# Two id-less toolsets, disjoint tool names (passes CombinedToolset uniqueness)
ts_a = FakeToolset(id=None, tool_defs={'alpha': 'DEF_ALPHA'})
ts_b = FakeToolset(id=None, tool_defs={'beta': 'DEF_BETA'})

run_cache = {}  # one per run (GraphAgentState.tool_defs_cache)

# Model request 1: combined toolset calls get_tools on both, in order
a1 = ts_a.get_tools(run_cache)
b1 = ts_b.get_tools(run_cache)

print('after request 1:')
print('  ts_a sees:', a1, 'fetches=', ts_a.fetches)
print('  ts_b sees:', b1, 'fetches=', ts_b.fetches)
print('  run_cache:', run_cache)

assert a1 == {'alpha': 'DEF_ALPHA'}, a1
# BUG: ts_b reads ts_a's cached defs under key '' and returns alpha, never fetching beta
print()
print('ts_b returned ts_a tools?', b1 == {'alpha': 'DEF_ALPHA'})
print('ts_b ever fetched its own tools?', ts_b.fetches > 0)
```

Output:

```
ts_b returned ts_a tools? True
ts_b ever fetched its own tools? False
```

### Scope / severity

- **Temporal is immune** — it rejects `None`-id leaf toolsets at construction.
- This configuration is **already broken on `main`** independently: two id-less leaf toolsets build identical DBOS step names (`{agent}__mcp_server.get_tools`), which DBOS keys by name + sequence, so duplicate step names already break recovery. #5883 changes the failure mode (loud → silent-wrong-tools) for an already-unsupported config rather than breaking a working one.

### Suggested direction

Require DBOS leaf-toolset ids (mirroring Temporal's guard), which both prevents this collision and fixes the pre-existing duplicate-step-name problem. Alternatively, skip the run-cache for id-less toolsets.

dsfaccini:
- Closes #5885

## Problem

On the `DBOSAgent` **wrapper** path, an id-less [`MCPToolset`][pydantic_ai.mcp.MCPToolset] has `wrapped.id is None`, which makes two of them collide on:

1. **The per-run tool-defs cache key** — `DBOSMCPToolsetBase.get_tools` keys the cache by `self.id or ''` (introduced in #5883), so a second id-less toolset silently returns the **first's** tool definitions and never lists its own.
2. **DBOS step names** — `DBOSMCPToolsetBase.__init__` derives its step-name prefix as `{agent}__mcp_server` (empty `id_suffix`), so two id-less toolsets register identically-named `@DBOS.step`s (`.get_tools`/`.get_instructions`/`.call_tool`), which DBOS can't distinguish in the workflow journal — breaking recovery.

Both stem from the same root cause named in the issue: unlike Temporal, DBOS doesn't require ids on MCP leaf toolsets. Temporal's `temporalize_toolset` raises `UserError` for any leaf with `id is None`; the DBOS wrapper's `dbosify_toolset` had no such guard. Thanks to @DouweM for [documenting the sibling step-name symptom](https://github.com/pydantic/pydantic-ai/issues/5885#issuecomment-4921054731).

## Fix

Mirror Temporal: raise `UserError` in the `dbosify_toolset` closure (`durable_exec/dbos/_agent.py`) for an id-less `MCPToolset`. **One guard closes both symptoms** — an id-less MCP toolset can no longer reach `DBOSMCPToolset.__init__`, so neither the cache key nor the step name can collide. `_mcp.py` is intentionally untouched (the guard sits upstream of it).

**Scope is deliberately narrow:**
- **DBOS wraps only `MCPToolset` at construction**, so the guard covers only that type — not all leaf toolsets the way Temporal does. `FunctionToolset` runs inline (an id-less one stays allowed, by design), and runtime `mcp`/`dynamic` toolsets are already rejected by `_reject_unsupported_runtime_toolsets`.
- **Not Prefect.** #6334 lumps Prefect in, but Prefect names tasks by tool name (`Call MCP Tool: {name}`) with no id-keyed cache — the mechanism here is DBOS-specific.
- **#4977 (`DBOSDurability` capability) is unmerged**, and already handles this at bind time on that path. The scope here is the `DBOSAgent` wrapper path on `main`.

## ⚠️ Design decision for a sanity pass

The guard raises for **any** id-less `MCPToolset` (Temporal-consistent), **not only** when 2+ id-less toolsets would actually collide.

The trade-off: a **single** id-less MCP toolset currently works (it gets the generic name, no collision). Raising for it is a minor **backward-compatibility change** — those users now need to add an `id=` (e.g. `MCPToolset(url, id='...')`).

I went with matching Temporal for engine consistency and because the fix is trivial for users, but this single-toolset back-compat point is your call — happy to switch to collision-only if you'd prefer to preserve it.

### Checklist

- [ ] No **breaking changes** in accordance with the [version policy](https://github.com/pydantic/pydantic-ai/blob/main/docs/version-policy.md). _Intentional minor back-compat change: a single id-less `MCPToolset` now raises, matching Temporal (which requires an `id` on leaf toolsets). Pending @DouweM's decision on the raise-for-any-id-less vs. only-on-collision fork flagged above._

### DIFF
--- a/docs/durable_execution/dbos.md
+++ b/docs/durable_execution/dbos.md
@@ -113,6 +113,8 @@ When using DBOS with Pydantic AI agents, there are a few important consideration
 
 Each agent instance must have a unique `name` so DBOS can correctly resume workflows after a failure or restart.
 
+Each [`MCPToolset`][pydantic_ai.mcp.MCPToolset] must have a unique [`id`][pydantic_ai.toolsets.AbstractToolset.id], as DBOS derives its step names and per-run tool-defs cache key from it. This field is normally optional, but is required when using DBOS. It should not be changed once the durable agent has been deployed to production, as this would break active workflows.
+
 Tools and event stream handlers are not automatically wrapped by DBOS. You can decide how to integrate them:
 
 * Decorate with `@DBOS.step` if the function involves non-determinism or I/O.
--- a/pydantic_ai_slim/pydantic_ai/durable_exec/dbos/_agent.py
+++ b/pydantic_ai_slim/pydantic_ai/durable_exec/dbos/_agent.py
@@ -121,6 +121,11 @@ class DBOSAgent(WrapperAgent[AgentDepsT, OutputDataT], DBOSConfiguredInstance):
                 pass
             else:
                 if isinstance(toolset, MCPToolset):
+                    if toolset.id is None:
+                        raise UserError(
+                            'MCP toolsets need to have a unique `id` in order to be used with DBOS. '
+                            "The ID will be used to identify the MCP server's steps within the workflow."
+                        )
                     return DBOSMCPToolset(
                         wrapped=toolset,
                         step_name_prefix=dbosagent_name,
--- a/tests/test_dbos.py
+++ b/tests/test_dbos.py
@@ -809,6 +809,20 @@ async def test_toolset_without_id():
     DBOSAgent(Agent(model=model, name='test_agent', toolsets=[FunctionToolset()]))
 
 
+async def test_mcp_toolset_without_id():
+    # Unlike `FunctionToolset`, an `MCPToolset` is wrapped in a `DBOSMCPToolset` whose step names and per-run
+    # tool-defs cache key both derive from the toolset's `id`; without one, two id-less MCP toolsets would
+    # collide on both. Require an `id` up front, like Temporal does for all leaf toolsets.
+    with pytest.raises(
+        UserError,
+        match=re.escape(
+            'MCP toolsets need to have a unique `id` in order to be used with DBOS. '
+            "The ID will be used to identify the MCP server's steps within the workflow."
+        ),
+    ):
+        DBOSAgent(Agent(model=model, name='test_agent', toolsets=[MCPToolset('https://example.com/mcp')]))
+
+
 async def test_dbos_agent():
     assert isinstance(complex_dbos_agent.model, DBOSModel)
     assert complex_dbos_agent.model.wrapped == complex_agent.model


dsfaccini:

|---|---|---|---|---|
| 1 · code-correctness | ✅ closed | 0 | 0 | 1 |

**Verdict:** ✅ all 3 waves passed — 0 blocking, 0 required. Guard, regression test, and docs note all reviewed clean; sibling engines (Temporal already guarded, Prefect structurally N/A) verified consistent.
Reviewed at 2026-07-09T21:20:00Z · diff `645a58aa5d6f`

<details>
<summary>⚠️ 1 informational (defer/skip)</summary>

**1. Guard checks `id is None`, not uniqueness — two *same-explicit-`id`* MCP toolsets still collide** · `durable_exec/dbos/_agent.py` · `dbosify_toolset`
Two independent reviewers (precedent + suspend/resume) noted the error text says "unique" while the check only tests presence. **Verified on the spot** (`local-notes/repro_same_id.py`): two `MCPToolset(id='mcp')` on one `DBOSAgent` → DBOS logs "Duplicate registration" per step but does **not** raise; both get step-name prefix `agent__mcp_server__mcp` + cache key `'mcp'` → the same collision class as #5885. **Affects Temporal identically** (its guard also checks presence only). — **defer (out of scope for the id-*less* fix; deliberately mirrors Temporal's guard one-to-one).** Tracking artifact: a cross-engine follow-up issue is **drafted** at `local-notes/followup_issue.md`; autonomous public-issue creation was blocked by policy, so filing is pending the maintainer.
</details>

PR #6426, opened 2026-07-10
Fix cancellation deadlock when a wrapped model request or agent run survives cancellation

### LINKED ISSUE #6422: Streaming handoff deadlocks when the model request survives cancellation, wedging cancelled TemporalAgent runs
### Initial Checks

### Description

When an agent run with an `event_stream_handler` is cancelled while the model request is in flight, and the awaitable inside `request_stream` **absorbs the cancellation and completes normally**, the run task deadlocks forever instead of unwinding.

An await that survives cancellation may sound like misbehaving user code, but it is the *documented, by-design* behavior of temporalio — and therefore of pydantic-ai's own first-party Temporal integration. Under `WAIT_CANCELLATION_COMPLETED` (the cancellation type users set via `TemporalAgent`'s `activity_config`), Temporal defines activity cancellation as cooperative; per the enum's own definition in sdk-core: "This can block the cancellation for a long time if activity doesn't heartbeat **or chooses to ignore the cancellation request**" ([sdk-core `workflow_commands.proto#L150-L153`](https://github.com/temporalio/sdk-core/blob/2cf2c103ee566148981a00c71f5fb2702d41fff6/crates/protos/protos/local/temporal/sdk/core/workflow_commands/workflow_commands.proto#L150-L153)). The Python SDK implements this in `run_activity()`: on `CancelledError` it emits the activity-cancel command and **re-awaits the shielded result future**; if the activity then completes successfully (e.g. the user cancels right as the model finishes), the await **returns the successful result** and the `CancelledError` is consumed ([sdk-python `_workflow_instance.py#L1901-L1930`](https://github.com/temporalio/sdk-python/blob/4ec9ab05b41a14cf8bf37a83f09fbfeb59628f7a/temporalio/worker/_workflow_instance.py#L1901-L1930)).

So inside a `TemporalAgent` workflow, "workflow cancelled while `model_request_stream` completes successfully" — an everyday race when a user hits stop — produces exactly the survivor condition, with no bug in user code.

The resulting deadlock in `_agent_graph.py` (`ModelRequestNode._make_request`, streaming path):

1. The run task is cancelled while parked on `asyncio.wait({ready_waiter, wrap_task})` ([_agent_graph.py#L727](https://github.com/pydantic/pydantic-ai/blob/ab79636233fd3e4bb386306994da09ba53dd5654/pydantic_ai_slim/pydantic_ai/_agent_graph.py#L727)).
2. The `except BaseException` cleanup calls `cancel_and_drain(ready_waiter, wrap_task)` ([#L729](https://github.com/pydantic/pydantic-ai/blob/ab79636233fd3e4bb386306994da09ba53dd5654/pydantic_ai_slim/pydantic_ai/_agent_graph.py#L729)), which cancels `wrap_task` and then awaits it inside a shielded scope ([_utils.py `cancel_and_drain`](https://github.com/pydantic/pydantic-ai/blob/ab79636233fd3e4bb386306994da09ba53dd5654/pydantic_ai_slim/pydantic_ai/_utils.py)).
3. The `CancelledError` lands on `wrap_task`'s innermost await — the in-flight model request inside `request_stream.__aenter__`. Per the SDK behavior above, that await swallows it and returns the completed response, so **`wrap_task` survives its own cancellation**.
4. The surviving `_streaming_handler` proceeds: sets `stream_ready`, then `await stream_done.wait()` ([#L694](https://github.com/pydantic/pydantic-ai/blob/ab79636233fd3e4bb386306994da09ba53dd5654/pydantic_ai_slim/pydantic_ai/_agent_graph.py#L694)). But `stream_done.set()` only happens on the normal consumption path ([#L775](https://github.com/pydantic/pydantic-ai/blob/ab79636233fd3e4bb386306994da09ba53dd5654/pydantic_ai_slim/pydantic_ai/_agent_graph.py#L775)), which the cancelled caller never reaches.
5. `cancel_and_drain` waits for `wrap_task`; `wrap_task` waits for `stream_done`. Mutual deadlock.

**Expected:** the run task ends cancelled (`CancelledError` propagates after cleanup).
**Actual:** the run task stays pending forever.

Inside a Temporal workflow the consequences are nasty: the workflow task completes with zero pending commands, so the workflow just idles in Running forever; workflow cancellation is one-shot state and has already been consumed, so further cancel requests are no-ops and only Terminate ends the execution. We diagnosed this from a production workflow that had been wedged for 18 days: history shows `WorkflowExecutionCancelRequested` → `ActivityTaskCancelRequested` → `ActivityTaskCompleted` (successful, 0.37s later) → a final `WorkflowTaskCompleted` with no commands, then nothing.

**Suggested fix direction:** in the cleanup paths that drain `wrap_task` before the stream has been consumed (the `except BaseException` around the `asyncio.wait` handoff, and `_cancel_task(wrap_task)` on stream error), set `stream_done` before awaiting `wrap_task`. A `wrap_task` that survived its cancel can then exit the `async with` and finish (its result is discarded by `cancel_and_drain`), the drain completes, and the original `CancelledError` re-raises as intended. Setting an extra event on a task that *did* die cancelled is harmless.

**About the MRE:** it deliberately has no Temporal dependency. A live reproduction needs a Temporal server, a worker, and winning a sub-second race between the cancel request and the activity completion — not minimal, and not deterministic. Instead the `WrapperModel` below simulates the exact observable behavior at the await boundary that the SDK links above document (absorb one `CancelledError`, return the completed result), which makes the deadlock reproduce deterministically on both 2.8.0 and 1.95.0. Verified output on both: `DEADLOCK: run task still pending 5s after cancellation`. The same handoff pattern exists at the run level (`wrap_run`/`_run_done` in `agent/__init__.py`), where a durable-execution step inside `before_run` produces the same hang.

Related but distinct: #5835 (run-level cancellation semantics) and #5837 (tool cancellation semantics) discuss the *contract*; this issue is a concrete liveness bug in the existing cleanup path. Workaround we're shipping meanwhile: `TemporalAgent(model_activity_config={"cancellation_type": ActivityCancellationType.TRY_CANCEL})`, which shrinks the race window from the full model-call duration to one server round-trip but cannot eliminate it (an activity that completed just before the cancel command is processed still resolves as completed and swallows the cancellation). We have a working fix with regression tests for both handoff sites on a branch and are happy to open a PR if a maintainer agrees with the approach.

```Python
import asyncio
from contextlib import asynccontextmanager

from pydantic_ai import Agent
from pydantic_ai.models.wrapper import WrapperModel


class SwallowOneCancelModel(WrapperModel):
    """request_stream absorbs one CancelledError, then completes normally.

    This is the observable behavior of the Temporal integration when the
    workflow is cancelled while the `model_request_stream` activity is in
    flight and the activity then completes successfully: temporalio's
    activity handle catches the CancelledError, re-awaits the shielded
    result future, and returns the successful result.
    """

    @asynccontextmanager
    async def request_stream(self, *args, **kwargs):
        try:
            await asyncio.sleep(30)  # the in-flight "activity"
        except asyncio.CancelledError:
            pass  # activity completed successfully; cancellation consumed
        async with self.wrapped.request_stream(
            *args, **kwargs
        ) as streamed_response:
            yield streamed_response


async def event_stream_handler(ctx, events):
    async for _ in events:
        pass


async def main() -> None:
    agent = Agent(SwallowOneCancelModel(TestModel()))

    run_task = asyncio.create_task(
        agent.run("hello", event_stream_handler=event_stream_handler)
    )
    await asyncio.sleep(0.5)  # let the run reach the model request
    run_task.cancel()

    try:
        await asyncio.wait_for(asyncio.shield(run_task), timeout=5)
    except asyncio.CancelledError:
        print("OK: run task ended cancelled")
    except TimeoutError:
        print("DEADLOCK: run task still pending 5s after cancellation")
    else:
        print("OK: run task completed")


asyncio.run(main())
```

### Logfire Trace

_No response_

- Python: 3.12
- Pydantic AI: reproduced on pydantic-ai-slim 2.8.0 (latest) and 1.95.0
- LLM provider SDK: n/a — reproduces with TestModel; observed in production with temporalio 1.26.0 + OpenAI

0ywfe:
## Summary

Closes #6422.

Cancelling an agent run deadlocks when an inner `await` **survives its own cancellation** and returns normally instead of propagating the `CancelledError`. This is the documented behavior of Temporal's `WAIT_CANCELLATION_COMPLETED` activity cancellation, routed through the first-party durable-execution integration — so "run cancelled while the activity completes successfully" (an everyday race when a user hits stop) wedges the run forever. The reporter diagnosed it from a production Temporal workflow stuck for 18 days.

## Root cause

Three cleanup paths share the failure: each tears down a task that has parked on a synchronization primitive the cancelled caller now never releases. The consumer-side cleanup paths already release it before draining; the early "cancelled mid-teardown" paths did not.

## Fix

- **Streaming handoff** — `ModelRequestNode.stream` (`_agent_graph.py`): if the model request inside `wrap_model_request` absorbs the cancellation, `_streaming_handler` is parked on `stream_done.wait()`. Set `stream_done` before draining `wrap_task`.
- **Run-level handoff** — `Agent.iter` (`agent/__init__.py`): if a durable step inside `before_run` absorbs the cancellation, `_do_run` is parked on `_run_done.wait()`. Set `_run_error` (so the survivor re-raises the caught error rather than asserting on a not-yet-produced result) and `_run_done` before draining `_wrap_task`.
- **`run_stream_events` teardown** — `_RunStreamEventsIterator.aclose()` (`agent/abstract.py`): if an inline tool absorbs the cancellation, the run resumes and the internal event handler blocks forever pushing the next event into the zero-buffer stream. Cancel the run first (unchanged for a normally-cancelled run), then close the receive end before draining so the surviving `send` fails with `BrokenResourceError` and the drain completes.

In every case the survivor can then unwind, its result/exception is discarded during the drain, and the original cancellation propagates. The extra `set()` / early close is a harmless no-op for a task that actually died cancelled.

## Tests

Regression tests for all three sites (`test_streaming_handoff_survives_absorbed_cancellation`, `test_run_handoff_survives_absorbed_cancellation`, `test_run_stream_events_aclose_survives_absorbed_cancellation`), each verified to deadlock without its fix and pass with it. No Temporal dependency — they reproduce the survivor condition with the "absorb one `CancelledError`" simulation from the reporter's MRE.


DouweM:
Thanks for jumping on this @0ywfe! I pushed a follow-up commit to complete it.

Your change handles the streaming handoff in `ModelRequestNode`, but as the issue calls out, the *identical* deadlock exists at the run level (`wrap_run`/`_run_done` in `Agent.iter`): when a durable step inside `before_run` absorbs the injected `CancelledError` and returns (Temporal's cooperative `WAIT_CANCELLATION_COMPLETED` cancellation), `_do_run` parks on `_run_done.wait()` that the cancelled caller never sets, so draining `_wrap_task` hangs forever. The follow-up sets `_run_error` + `_run_done` before the drain there too, mirroring your streaming fix — the survivor re-raises the caught error, which `cancel_and_drain` discards.

I also added regression tests for *both* handoff sites. Each was verified to deadlock (10s timeout) without its respective fix and to pass with it, with no Temporal dependency — using the same "absorb one `CancelledError`" simulation from the issue's MRE.

Should be complete now.

Issue #6436, opened 2026-07-13
`run_stream` / `run_stream_sync` hangs until the request timeout (~600s) when `message_history` is passed (Anthropic); `run_sync` is unaffected


mathewdaubney:
### Initial Checks

- [x] I'm using the [latest version](https://github.com/pydantic/pydantic-ai/releases/latest) of Pydantic AI
- [x] I've searched for my issue in [the issue tracker](https://github.com/pydantic/pydantic-ai/issues) before opening this issue

### Summary

Streaming a run that carries `message_history=` stalls on the first model request until the
HTTP client timeout elapses (the Anthropic SDK's ~600s default), and only then does the run
complete (apparently via an internal retry). A `run_sync` call with **identical** inputs returns
in ~1.2s, and streaming **without** `message_history` is also fast (~1.8s). So the trigger is
specifically *streaming + message history*.

With the default timeout this presents as a ~10-minute freeze with no output. Setting a short
`model_settings={"timeout": N}` makes the run complete in ~`N`+1.5s instead — strong evidence the
stall is timeout-bound (a stuck first request rescued by a retry), not real work.

### Reproduction

Requires a real Anthropic API key (the behavior is network/streaming timing, so `TestModel`
can't reproduce it).

```python
import time
from pydantic_ai import Agent

agent = Agent("anthropic:claude-sonnet-5", instructions="Be concise.")

# Produce a message history with one normal (non-streaming) turn.
r1 = agent.run_sync("Name a colour.")
hist = r1.all_messages()

# 1) run_sync WITH history -> fast
t = time.perf_counter()
agent.run_sync("Another one?", message_history=hist)
print(f"run_sync + history        : {time.perf_counter()-t:5.1f}s")

# 2) run_stream_sync WITH history -> stalls until the client timeout (~600s), then completes
t = time.perf_counter()
with agent.run_stream_sync("Another one?", message_history=hist) as s:
    _ = "".join(s.stream_text())
print(f"run_stream_sync + history : {time.perf_counter()-t:5.1f}s")

# 3) control: run_stream_sync WITHOUT history -> fast
t = time.perf_counter()
with agent.run_stream_sync("Name a colour.") as s:
    _ = "".join(s.stream_text())
print(f"run_stream_sync NO history: {time.perf_counter()-t:5.1f}s")
```

> Tip for reviewers: to avoid waiting ~600s, add `model_settings={"timeout": 20}` to the streamed
[...]

PR #6454, opened 2026-07-13
Fix event-loop affinity in `run_stream_sync`

adtyavrdhn:
Addresses the concrete synchronous reproduction in #6436.

## Problem

`run_sync()` and `run_stream_sync()` reused the same Anthropic/HTTPX client but ran it on different event loops. The blocking portal introduced in #6199 moved synchronous streaming to a new loop, so HTTPX could reuse a keep-alive connection whose asyncio transport belonged to the loop used by the preceding `run_sync()` call. On Windows this could hang until the Anthropic timeout and retry; with retries disabled it fails immediately as a cross-loop transport error.

`message_history` is not causal. It made the common preceding-request sequence, and therefore keep-alive connection reuse, visible.

## Fix

Keep synchronous streaming on the caller-owned event loop while retaining stable task lifecycles:

- one persistent task owns the async context manager;
- one persistent pump task owns each async iterator;
- short-lived calls run on the same loop with run-owned context variables merged into the current caller context;
- shutdown drains pumps before closing the owner context, including interrupt and garbage-collection paths;
- `run_stream_sync()` and `model_request_stream_sync()` document their same-thread lifecycle requirements.

The concrete issue reproduction, `run_sync()` followed by `run_stream_sync()` on the same provider client, now reuses the same TCP connection successfully. Fully async `run()` followed by `run_stream()` on one event loop is covered as a control and remains healthy. Moving one pooled async client across unrelated event loops remains unsupported.

The issue also reports an async `run_stream()` failure without an executable async reproduction or loop-ownership details. This PR deliberately does not broaden the fix to cross-loop async client reuse. The issue should remain open until that path is clarified.

## Verification

- Real local Anthropic-compatible keep-alive server, retries disabled and one-second timeouts, covering provider-owned and user-injected clients, agent and direct APIs, both request orders, and message history.
- Regression was confirmed against the pre-fix implementation.
- Focused sync-stream suite: 166 passed, 3 xfailed.
- Full repository suite before the final lifecycle hardening: 8,144 passed, 535 skipped, 3 xfailed.
- Full pre-commit hooks, Ruff, Pyright, and 100% line and branch coverage for `_sync_stream.py` pass.
- Three independent adversarial reviews plus the automated whole-diff review returned no actionable findings.

### Checklist


pydantic_ai_slim/pydantic_ai/_sync_stream.py
```diff
@@ -93,117 +192,151 @@ def __init__(self, cm: AbstractAsyncContextManager[StreamT], *, async_alternativ
...
-            # `portal.call` accepts sync or async `func` (`Awaitable[T] | T`); the union defeats the
-            # return-type inference, so we narrow it back the same way anyio does internally.
-            return cast(T, self._portal.call(func, *args))
+            return self._run(self._call(func, *args))
         except (KeyboardInterrupt, SystemExit) as exc:
             self.shutdown((type(exc), exc, exc.__traceback__))
             raise
 
     def call_with_otel_context(self, func: Callable[[], Awaitable[T]]) -> T:
         """Like [`call`][pydantic_ai._sync_stream.SyncStreamBridge.call], with the run span's OTel context attached."""
```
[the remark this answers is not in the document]
adtyavrdhn: Updated the docstring to state that the task context propagates the run span. Addressed in 53e86da4.

tests/test_streaming.py
```diff
@@ -293,62 +304,774 @@ async def ret_a(x: str) -> str:
...
+    try:
+        with pytest.raises(error_type) as exc_info:
+            SyncStreamBridge(FailingContextManager(), async_alternative='`async_method`')
+        assert exc_info.value is error
+        assert not forced_stop
+        assert loop.run_until_complete(asyncio.sleep(0)) is None
+    finally:
+        stop_handle.cancel()
+        loop.close()
+        asyncio.set_event_loop(original_loop)
```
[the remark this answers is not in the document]
adtyavrdhn: Increased the defensive hang-detector timeout to one second while preserving the forced-stop assertion. Addressed in 53e86da4.

tests/test_streaming.py
```diff
@@ -434,7 +1130,143 @@ def test_run_stream_sync_early_break_tears_down_pump():
...
+        with pytest.raises(error_type) as exc_info:
+            while True:
+                next(stream)
+        assert exc_info.value is error
+        assert not forced_stop
+        assert bridge._owner_task.done()  # pyright: ignore[reportPrivateUsage]
+        assert loop.run_until_complete(asyncio.sleep(0)) is None
+    finally:
+        stop_handle.cancel()
+
```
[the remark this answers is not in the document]
adtyavrdhn: Increased the defensive pump hang-detector timeout to one second while preserving the forced-stop assertion. Addressed in 53e86da4.

DouweM:
Thanks @adtyavrdhn — this is a careful piece of work and the keep-alive HTTP-server test genuinely proves the fix (same client port reused across `run_sync` → `run_stream_sync`), which VCR structurally can't.

Before merging I ran two independent AI reviews (OpenAI Codex / GPT-5.6 and Claude Fable) plus a manual pass with the #6199 history in mind (that was the PR that deliberately introduced the portal thread this PR now removes). Sharing everything below so you can point your agent at it. **None of the High-severity items are code bugs** — they're conscious behavioral changes vs #6199 that I need to sign off on, not fix. The one thing I'd actually like changed before merge is a single Medium (string-matched teardown). Full list, most to least important:

## Confirmed sound (no action needed)

Both reviewers independently traced and confirmed the fix re-earns everything #6199 bought:
- **Task stability**: the `run_stream()` CM stays in the one long-lived owner task, and each `async for` stays in one pump task; `run_until_complete()` suspending/restarting the loop between calls never re-homes those tasks, so the "cancel scope exited in a different task" `RuntimeError` cannot recur.
- **OTel parenting**: `copy_context()` at the CM yield snapshots the run-span contextvar, and `_task_context()` merges the run-owned deltas into every call/pump task, so child spans still parent under the run span and it closes in the task that opened it. The explicit attach/detach removal is a clean simplification.
- **Interrupt/teardown paths** (init pre/post entry, `call`, `shutdown`, pump-drain-before-owner-exit ordering) are correct, and the test coverage for them is unusually thorough.

## Should fix before merge

**[Medium] `_run_task_to_completion` recovers from a stale loop-stop by matching an exact English string.** It matches `'Event loop stopped before Future completed.'` and retries exactly once. Two problems: (a) brittle across CPython patch releases / uvloop / custom loops, and (b) the single retry assumes at most one queued stop callback. To be clear, this is *not* the string-match-and-suppress pattern from #3716 that I rejected — this only *detects* a well-understood asyncio artifact and the re-drive is semantically sound. But it doesn't need to be message-dependent: `_wait_for_task` only awaits `asyncio.wait` + `sleep(0)` and can't itself raise `RuntimeError`, so any `RuntimeError` while `waiter` is not done (and the loop isn't closed) is an early stop. Suggest replacing the string check with a bounded `while not waiter.done()` re-drive loop (raise if `loop.is_closed()`), which is loop-implementation-independent and handles multiple queued stops. ~3 lines; existing tests should pass unchanged.

## Lower priority

**[Low] Stale stop callbacks can escape the bridge and poison a later unrelated `run_sync`.** The bridge now shares the caller's default loop with `run_sync` (`pydantic_graph._utils.run_until_complete` has no stale-stop handling). Narrow window: a KeyboardInterrupt landing in `stream_sync`'s post-exhaustion `self._run(pump_task)` (not routed through `call`, so no `shutdown()`) while the pump is already done with its stop callback queued, in no-`with` usage → the stale stop survives to the user's next `run_sync`, which fails with a confusing `RuntimeError`. The bounded-loop fix above shrinks this window.

**[Low] Dropping the `_disable_threads` guard is right for Temporal but unverified for emscripten/Pyodide.** Temporal is fine — workflow code runs inside a running loop, so the existing running-loop guard still rejects `run_stream_sync` with a clear message, and the thread-free design is strictly better for `disable_threads()` contexts (good call adding the `works_with_disabled_threads` test). The residual risk is Pyodide only: `_disable_threads` defaults on for `sys.platform == 'emscripten'`, where `WebLoop.run_until_complete` can't block, so `loop.run_until_complete(ready)` may fail cryptically instead of #6199's clear "use `run_stream` instead" error. `run_sync` already has this limitation there, so it's an error-quality regression, not a capability loss — worth either a targeted clear error or a note that emscripten is out of scope.

**[Nit] Docstring thread-wording.** `direct.py` says "the thread where its `with` block is entered", but the bridge records the thread at construction; these diverge for no-`with` `__iter__`. `abstract.py`/`result.py` ("where it was created") is accurate. Align on one precise phrase, e.g. "the thread where the synchronous stream is created."

**[Nit] `_task_context` captures run contextvars once at the CM yield**, so contextvar mutations by run-scoped middleware *after* the yield won't propagate to later calls. This was raised as a Medium but on review it's parity with the portal impl (which also snapshotted once at yield and gave each `portal.call` an isolated copy) — and the new merge actually improves on it by flowing the caller's live context in. No change needed beyond making sure the "run-owned contextvars propagate" comment doesn't overclaim.

**[Nit] `_finalize_loop` early-returns on `owner_task.done()`** without draining pumps / retrieving the owner exception. Reachable only via private-API interference (external cancel of `_owner_task`), not any public path — noting for completeness.

## Behavioral changes vs #6199 — for my decision, not code changes

These two are inherent to moving off the dedicated thread onto the caller's loop, and are already disclosed in the updated docstrings. Flagging so we're explicit about them (they likely warrant a changelog/version-policy note):

1. **Non-`with` (GC) cleanup is now best-effort.** #6199 preserved reliable no-`with` cleanup via the live portal thread; the caller-loop design can't drive a stopped loop from a foreign thread or under another running loop, so those cases only *queue* cleanup on the next drive of the loop — or never (owner task, pumps, HTTP socket, OTel run span leak until process exit). The common case (refcount-death on the owner thread while the loop is idle) still cleans up fully and synchronously, and deferred cleanup piggybacks on any later sync call, so real exposure is narrower than "GC is broken" — but it is a real reduction from #6199's guarantee.

2. **The wrapper is now single-thread-affine.** The portal was thread-safe by construction (`portal.call` from any thread), so since #6199 constructing a stream on thread A and consuming/closing it on worker thread B *worked*. Now the first off-thread `next()`/`get_output()`/`__exit__` raises `RuntimeError('A synchronous stream must be used and closed on the thread where it was created.')`, and off-thread `__exit__` skips cleanup (thread check precedes `finalizer.detach()`), falling back to the best-effort GC path above. Unavoidable given the design and the error is clear, but it's a behavioral regression on a public API rather than just a doc tweak.

I'm comfortable with both trades — fixing the actual transport/loop-affinity root cause of #6436 is worth it, and the portal could never fix that. I'll add a changelog note. Leaving #6436 open for the async path is the right call too.

Net: address the Medium (+ the Low it subsumes), decide on the emscripten error-quality nit, and the doc nits, and I think this is good to merge.

pydantic_ai_slim/pydantic_ai/_sync_stream.py
```diff
@@ -36,52 +38,165 @@
 _ExcInfo = tuple[type[BaseException] | None, BaseException | None, TracebackType | None]
 
 
-@asynccontextmanager
-async def _capture_otel_context(
-    cm: AbstractAsyncContextManager[StreamT], captured: list[otel_context.Context]
-) -> AsyncGenerator[StreamT]:
-    """Enter `cm` and capture the OTel context that's active at its yield point.
+def _is_awaitable(value: T | Awaitable[T]) -> TypeIs[Awaitable[T]]:
```
adtyavrdhn, line 40: I would assume this is something that exists already in _function_schema for Code Mode function signatures? Can we not reuse?

pydantic_ai_slim/pydantic_ai/_sync_stream.py
```diff
@@ -36,52 +38,165 @@
...
+
+async def _hold_context_manager(
+    cm: AbstractAsyncContextManager[StreamT],
+    ready: asyncio.Future[tuple[StreamT, Context]],
+    exit_requested: asyncio.Future[_ExcInfo],
+) -> None:
+    """Enter and exit `cm` in one task, remaining parked while sync code uses the yielded stream."""
+    try:
+        stream = await cm.__aenter__()
+    except (KeyboardInterrupt, SystemExit):
```
adtyavrdhn, line 53: I am not sure why we need to capture these two specifically

tests/test_streaming.py
```diff
@@ -293,62 +304,774 @@ async def ret_a(x: str) -> str:
...
+        # Finishing the call queues its `run_until_complete()` stop callback behind this interrupt.
+        loop.call_soon(interrupt)
+
+    with pytest.raises(KeyboardInterrupt) as exc_info:
+        bridge.call(finish_then_interrupt)
+
+    assert exc_info.value is original_error
+    assert cleanup_complete
+    assert bridge._owner_task.done()  # pyright: ignore[reportPrivateUsage]
+    assert loop.run_until_complete(asyncio.sleep(0)) is None
```
adtyavrdhn, line 450: These look flaky and not obvious what we're even testing here

pydantic_ai_slim/pydantic_ai/_sync_stream.py
```diff
@@ -36,52 +38,165 @@
...
-    stream_cm: AbstractContextManager[Any],
+
+def _run_task_to_completion(loop: asyncio.AbstractEventLoop, task: asyncio.Task[None]) -> None:
+    """Drive a task and consume a stale `run_until_complete()` stop callback if one is queued."""
+    waiter = loop.create_task(_wait_for_task(task))
+    try:
+        try:
+            loop.run_until_complete(waiter)
+        except RuntimeError as exc:
+            if str(exc) != 'Event loop stopped before Future completed.':
```
adtyavrdhn: Brittle, should not do this

pydantic_ai_slim/pydantic_ai/direct.py
```diff
@@ -333,7 +334,8 @@ def __exit__(
     def __iter__(self) -> Iterator[messages.ModelResponseStreamEvent]:
         """Stream the response as an iterable of [`ModelResponseStreamEvent`][pydantic_ai.messages.ModelResponseStreamEvent]s."""
         bridge = self._ensure_bridge()
-        return bridge.stream_sync(lambda: aiter(bridge.stream))
+        stream = bridge.stream
+        return bridge.stream_sync(lambda: aiter(stream))
```
adtyavrdhn, line 340: What does this achieve?

pydantic_ai_slim/pydantic_ai/_sync_stream.py
```diff
@@ -93,117 +208,194 @@ def __init__(self, cm: AbstractAsyncContextManager[StreamT], *, async_alternativ
...
-            raise RuntimeError(
-                f'Synchronous streaming runs on a dedicated event-loop thread, which is unavailable in this '
-                f'environment (e.g. emscripten or a Temporal workflow); use {async_alternative} instead.'
-            )
 
-        captured: list[otel_context.Context] = []
-        portal_cm = start_blocking_portal()
-        portal = portal_cm.__enter__()
+        loop = _utils.get_event_loop()
+        caller_context = copy_context()
```
adtyavrdhn, line 215: What warrants a `copy_context` change?

pydantic_ai_slim/pydantic_ai/_sync_stream.py
```diff
@@ -93,117 +208,194 @@ def __init__(self, cm: AbstractAsyncContextManager[StreamT], *, async_alternativ
...
+        ready = loop.create_future()
+        exit_requested = loop.create_future()
```
adtyavrdhn: Do we care about the data this will populate? Why a future and not an event?

pydantic_ai_slim/pydantic_ai/_sync_stream.py
```diff
@@ -36,52 +38,165 @@
...
-    """
-    async with cm as stream:
-        captured.append(otel_context.get_current())
-        yield stream
 
+async def _wait_for_task(task: asyncio.Task[None]) -> None:
+    """Wait for a task without propagating its result, then remain pending for one extra loop turn."""
+    if not task.done():
+        await asyncio.wait((task,))
+    await asyncio.sleep(0)
```
adtyavrdhn, line 80: This feels too brittle but somehow codex has written failing tests for this which I'll need to validate

pydantic_ai_slim/pydantic_ai/_sync_stream.py
```diff
@@ -36,52 +38,165 @@
...
+    except (KeyboardInterrupt, SystemExit):
+        # Futures deliberately do not stop `run_until_complete()` for these exceptions because tasks
+        # normally re-raise them directly from the event loop. Preserve that behavior instead of
+        # forwarding them through `ready`, which would leave the loop running indefinitely.
+        raise
+    except BaseException as exc:
+        ready.set_exception(exc)
+        return
+
+    ready.set_result((stream, copy_context()))
```
adtyavrdhn: I see it does matter that these are futures

pydantic_ai_slim/pydantic_ai/_sync_stream.py
```diff
@@ -36,52 +38,165 @@
...
+    try:
+        exc_info = await exit_requested
```
adtyavrdhn: It might not be an exception? Maybe the variable name should not imply that then?

pydantic_ai_slim/pydantic_ai/_sync_stream.py
```diff
@@ -36,52 +38,165 @@
...
+
+    if loop.is_running() or get_ident() != owner_thread_id:
+        with suppress(RuntimeError):
+            loop.call_soon_threadsafe(request_exit)
+        return
     try:
-        stream_cm.__exit__(*exc_info)
-    finally:
-        portal_cm.__exit__(None, None, None)
+        asyncio.get_running_loop()
```
adtyavrdhn: We already have a utils._get_event_loop helper

pydantic_ai_slim/pydantic_ai/_sync_stream.py
```diff
@@ -36,52 +38,165 @@
...
+        with suppress(BaseException):
+            _shutdown_loop(loop, owner_task, exit_requested, pump_tasks, (None, None, None))
+    else:
+        # Another event loop is running in this thread, so the owner loop cannot be driven synchronously.
+        with suppress(RuntimeError):
+            loop.call_soon_threadsafe(request_exit)
+
+
+@dataclass(frozen=True, slots=True)
+class _Received(Generic[T]):
```
adtyavrdhn: What even is this?

pydantic_ai_slim/pydantic_ai/_sync_stream.py
```diff
@@ -36,52 +38,165 @@
...
+    """Drive a task to completion despite stale `run_until_complete()` stop callbacks."""
+    waiter = loop.create_task(_wait_for_task(task))
+    try:
+        while not waiter.done():
+            try:
+                loop.run_until_complete(waiter)
+            except RuntimeError:
+                if loop.is_closed() or waiter.done():
+                    raise
+    except BaseException:
```
adtyavrdhn, line 95: Catching BaseExceptions everywhere

pydantic_ai_slim/pydantic_ai/_sync_stream.py
adtyavrdhn: _function_schema classifies callable signatures; this checks a returned value that may be awaitable. The local TypeIs preserves T in both branches. Added a docstring in 417f0e70.

pydantic_ai_slim/pydantic_ai/_sync_stream.py
adtyavrdhn: asyncio.run_until_complete intentionally does not stop for futures containing KeyboardInterrupt or SystemExit. Forwarding either through entered would hang, so they must escape from the owner task directly. The focused tests cover both.

tests/test_streaming.py
adtyavrdhn: The ordering is deterministic: the interrupt callback is queued before the run_until_complete stop callback. The docstring now states that invariant, and the focused bridge suite passes. Keeping the regression coverage.

pydantic_ai_slim/pydantic_ai/_sync_stream.py
adtyavrdhn: Addressed in 9127ddde. The exact exception-string match is gone; cleanup now re-drives while the waiter is pending. 417f0e70 also documents why an early stop is the only RuntimeError handled here.

pydantic_ai_slim/pydantic_ai/direct.py
adtyavrdhn: The pump task retains its iterator factory. Capturing bridge there creates a reference cycle that delays the bridge finalizer, so the local intentionally captures only stream. Added the rationale in 417f0e70.

pydantic_ai_slim/pydantic_ai/_sync_stream.py
adtyavrdhn: __aenter__ changes context only inside the owner task. Later call and pump tasks need that run context for OTel and capability contextvars while preserving live caller values. Added the rationale beside the snapshot in 417f0e70.

pydantic_ai_slim/pydantic_ai/_sync_stream.py
adtyavrdhn: Both signals carry data, so events are insufficient: entered carries the stream and Context or an entry error, and exit_requested carries the three __exit__ arguments. Renamed and explicitly typed them in 417f0e70.

pydantic_ai_slim/pydantic_ai/_sync_stream.py
adtyavrdhn: This waits one extra loop turn so a stale run_until_complete stop callback cannot interrupt cleanup. The docstring and progress invariant are now explicit in 417f0e70, and the focused interrupt tests pass.

pydantic_ai_slim/pydantic_ai/_sync_stream.py
adtyavrdhn: Confirmed. The futures carry the entered stream and context plus the exit arguments. Renamed ready to entered and added explicit types in 417f0e70.

pydantic_ai_slim/pydantic_ai/_sync_stream.py
adtyavrdhn: Agreed. The tuple is the context manager __exit__ arguments and is often all None. Renamed _ExcInfo and exc_info to _ExitInfo and exit_info in 6053d6aa.

pydantic_ai_slim/pydantic_ai/_sync_stream.py
adtyavrdhn: This call is checking whether another event loop is already running in the current thread. _utils.get_event_loop retrieves or creates the owner loop, so it is not interchangeable here.

pydantic_ai_slim/pydantic_ai/_sync_stream.py
adtyavrdhn: Removed in 417f0e70. _Received only distinguished a real None item from end-of-stream; the bridge now reuses _utils.UNSET and _utils.is_set.

pydantic_ai_slim/pydantic_ai/_sync_stream.py
adtyavrdhn: These cleanup blocks must also handle KeyboardInterrupt, SystemExit, and CancelledError; catching Exception would strand tasks on the exact interrupt paths this bridge protects. Added a comment explaining the narrow cleanup purpose in 6053d6aa.

DouweM:
Round 2 — re-ran the two independent AI reviews (Codex/GPT-5.6 and Claude Fable) against the current HEAD (`6053d6aa`), since the lifecycle refactor touched real teardown code. Good news first: **the refactor resolved everything substantive from round 1**, and both reviewers independently re-verified the core invariants still hold. There's a small, concrete to-do list to close it out, plus two items I've adjudicated below so you don't have to.

## Resolved since round 1 ✅

- **The brittle string-match is gone and the replacement is correct.** Both reviewers traced the new bounded `while not waiter.done()` re-drive in `_run_task_to_completion` against CPython's `run_until_complete` machinery and confirmed it consumes a finite queue of stale `loop.stop()` callbacks, preserves and re-raises the initiating interrupt, and can't swallow a genuine RuntimeError from the awaited work (`_wait_for_task` provably can't raise `RuntimeError`). The new `retries_multiple_early_stops` test nails the multi-stop case. 👍
- **Renames** (`_ExcInfo`→`_ExitInfo`, `exc_info`→`exit_info`, `ready`→`entered`) are purely cosmetic, no tuple-order changes.
- **`_Received`→`_utils.UNSET`/`is_set`** is clean — identity check on a private singleton, so a genuine `None` stream item still flows through.
- **Rationale comments** for the interrupt-escape, extra `asyncio.sleep(0)` turn, and context snapshot are accurate.
- **Emscripten concern dissolved** — this is now a *functional improvement*, not a regression: sync streaming works under `disable_threads()`, and the running-loop guard still emits the clear "use `run_stream`" error wherever a live loop exists (Temporal sandbox, Pyodide with a browser loop). Nice.

Core invariants re-confirmed after the refactor: task-stability (CM in one owner task, each `async for` in one pump task), OTel parenting via the contextvar snapshot, and shutdown ordering (pumps drain before the owner CM's `__aexit__`).

## To close it out

### 1. [Blocker — CI] Uncovered `raise` at `_sync_stream.py:92`
This is the only thing actually red (coverage 99.99% vs the `fail-under=100` gate; `check` fails only because `coverage` does). Both reviewers concluded this branch is **effectively unreachable via any public path on a standard CPython loop** — `loop.is_closed()` can't become true between `create_task` and the drive single-threaded, and the `waiter` coroutine can't raise `RuntimeError`. So either:
- `# pragma: no cover` with a one-line justification ("defensive: reachable only via a non-standard loop, or the loop being closed/run from another thread"), **or**
- combine with fix #2 below, which gives a deterministically testable second arm.

### 2. [Should fix — hardening] `_run_task_to_completion` retry isn't bounded on a *persistent* `RuntimeError`
The two reviewers split on severity here (Codex: High/"hang risk"; Fable: Low/"pure misuse") — my read is Fable's right on likelihood but Codex has the sharper structural point: **the old code was bounded (retry once, then re-raise) and the refactor silently removed that bound.** If `run_until_complete()` raises a *persistent* RuntimeError — most plausibly `'This event loop is already running'` because another thread is concurrently driving the owner loop — the current guard (`if loop.is_closed() or waiter.done(): raise`) neither trips, so it spins CPU-bound. One-token fix both reviewers endorse:

```python
if loop.is_closed() or loop.is_running() or waiter.done():
    raise
```

This closes the `'already running'` case and, conveniently, gives you a deterministically testable arm for finding #1 (monkeypatch `run_until_complete` to drive the waiter once via the original, then raise `RuntimeError` with the loop marked running). It does *not* defend a pathological self-rescheduling `loop.stop()` callback — I'd accept that as a user footgun that would break far more than this bridge, though a belt-and-suspenders iteration cap is your call.

### 3. [Nit — polish] `stream_sync` creates the pump task before the already-closed check
Minor: on a closed-but-loop-still-open bridge, the first `next()` creates a task, then immediately errors and cancels it in `finally` (harmless — cancelled before its first step). An `if not self._finalizer.alive: raise` at the top of the generator would be tidier.

### 4. [Nit — docs] Thread-wording still inconsistent
`direct.py`'s `StreamedResponseSync` docstring says "the thread where the `with` block is entered" while the bridge records the thread at construction; `abstract.py`/`result.py` ("where it was created") are accurate. One-word alignment.

## Two items I've adjudicated so you don't have to

**`exit_info` path discards `__aexit__`'s return value (Codex flagged as a suppression gap).** I checked the actual chain: both `StreamedRunResultSync.__exit__` and `StreamedResponseSync.__exit__` are typed `-> None` and don't return `shutdown()`'s result — and `shutdown()` returned `None` in the old portal code too. So an async CM returning `True` from `__aexit__` to suppress the sync `with`-body exception was **never** honored through the sync wrapper — this is **pre-existing behavior unchanged by this PR, not a regression**, and the real CMs (`run_stream` / `model_request_stream`, both `@asynccontextmanager` generators) never suppress anyway. **No action needed**; at most a "suppression-on-shutdown is not supported" note if you want it explicit.

**`UNSET` as an unrepresentable stream item (Codex, Low).** A stream emitting the *exact* `_utils.UNSET` singleton would be truncated as end-of-stream. Theoretical only — the real event types never emit it — but it does narrow the generic `SyncStreamBridge[T]` contract vs the removed `_Received` wrapper. Fine to leave; noting for the record.

## Verdicts
- **Fable: mergeable** — single most important item is settling line 92 (± the `is_running()` hardening).
- **Codex: not yet** — primarily on the unbounded retry (which the `is_running()` guard addresses).

Net from me: this is close. The honest pre-merge list is **#1 (coverage)** and **#2 (the `is_running()` guard)** — both tiny — plus the two nits if you're in there anyway. Everything else is theoretical or by-design. Really nice work on the teardown machinery; the deterministic stale-stop-injection tests in particular are exactly the right kind of coverage for this.

DouweM:
### Decision record: the `__aexit__`-suppression "gap" — intentionally *not* fixing here

Capturing this so it isn't re-flagged later as an oversight.

**The observation** (from the round-2 Codex review): in `_hold_context_manager`, the caller-driven exit path
```python
else:
    await cm.__aexit__(*exit_info)
```
discards `__aexit__`'s return value, so if the sync `with` body raises and the async CM's `__aexit__` returns `True` to *suppress* that exception, the suppression isn't honored — the exception still propagates out of the sync `with`.

**Decision: leave as-is, add a clarifying comment, don't thread the bool back.** Rationale:

1. **It's pre-existing, not a regression.** Both `StreamedRunResultSync.__exit__` and `StreamedResponseSync.__exit__` are typed `-> None` and don't return `shutdown()`'s result, and the old portal `shutdown()`/`_shutdown_portal` returned `None` too. So async `__aexit__` suppression has *never* been honored through the sync wrapper. Preserving that is the status quo; "fixing" it would be net-new behavior.

2. **No real caller can observe it.** `SyncStreamBridge` is a private helper that only ever wraps `run_stream` / `model_request_stream`, both `@asynccontextmanager` generators that don't swallow the with-body exception. You can't write a failing test against the real API — you'd have to inject a synthetic suppressing CM. That's the tell that this is a protocol-completeness itch, not a user-facing bug, and it doesn't meet the "reproduce the defect before extending a fix" bar.

3. **The cost lands in the most dangerous place.** Threading the `__aexit__` bool back through `shutdown()` → `__exit__` means touching the interrupt / cancellation / GC teardown paths — the most safety-sensitive code in this PR — and changing the public `__exit__` signature to `-> bool | None`, which *advertises* suppression support and thereby obligates it to work correctly across all of those paths. Poor risk/reward: complicating the trickiest code for zero observable benefit.

**Action taken:** none required for merge. Worth a one-line comment at the `else: await cm.__aexit__(*exit_info)` branch stating that suppression is intentionally not propagated on the sync path, so a future reader doesn't "fix" it. If a concrete need ever surfaces (an actual suppressing CM), it becomes a separate, well-scoped PR with its own test.

**Residual caveat:** the "real CMs never suppress" claim rests on `run_stream` / `model_request_stream` being non-swallowing generators. I haven't exhaustively traced whether an inner `anyio.CancelScope` could ever surface a `True` out of `run_stream`'s `__aexit__` for a `CancelledError`; if that turns out possible it's worth a second look, but it wouldn't change the disposition here (the sync path never honored suppression regardless).

pydantic_ai_slim/pydantic_ai/direct.py
```diff
@@ -302,14 +303,14 @@ def _prepare_model(
...
-    [`SyncStreamBridge`][pydantic_ai._sync_stream.SyncStreamBridge]) rather than pumping a shared event loop
-    from a background thread, so exiting the `with` block cancels the underlying request promptly and closes
-    the connection instead of waiting for the whole response to arrive.
+    The stream uses [`SyncStreamBridge`][pydantic_ai._sync_stream.SyncStreamBridge] to keep context-manager
+    and iterator lifecycles in stable tasks. Exiting the `with` block cancels the underlying request promptly
+    and closes the connection instead of waiting for the whole response to arrive.
 
-    This class must be used as a context manager with the `with` statement.
+    This class must be used as a context manager with the `with` statement. It must be used and closed on the
+    thread where the `with` block is entered.
```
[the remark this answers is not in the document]
adtyavrdhn: Addressed in 75cffc3d. The docstring now uses the shared “synchronous stream is created” terminology while retaining the precise lifecycle: `StreamedResponseSync.__enter__` creates the bridge, and the stream must be used and closed on that thread.

PR #6459, opened 2026-07-13
Fix `RenamedToolset` silently dropping a tool on a name collision

### LINKED ISSUE #6461: `RenamedToolset` silently drops a tool when a rename collides with an existing tool name
`RenamedToolset.get_tools` writes its output dict without any collision check, so renaming a tool onto a name another tool already occupies silently overwrites one of them — the model just loses access to a tool, with no error, and which one survives depends on iteration order.

```python
from pydantic_ai.tools import Tool
from pydantic_ai.toolsets.function import FunctionToolset

def a(x: int) -> int: ...
def b(x: int) -> int: ...

toolset = FunctionToolset(tools=[Tool(a), Tool(b)])
renamed = toolset.renamed({'b': 'a'})   # rename `a` -> `b`, but `b` already exists
# renamed.get_tools(...) returns just {'b'} — tool `a` is silently gone
```

The sibling toolsets already raise `UserError` on this same conflict — `FunctionToolset`, `CombinedToolset`, and `PreparedToolset` — so `RenamedToolset` is the only wrapper that swallows it.

I have a small fix ready that mirrors `FunctionToolset`'s guard (reusing its two existing messages) plus a test.

sean-kim05:
Closes #6461

`RenamedToolset.get_tools` writes its output dict unconditionally, so renaming a tool onto a name another tool already occupies silently overwrites one of them — the model loses access to a tool with no error, and which one survives depends on iteration order.

Every sibling wrapper already raises `UserError` on this exact conflict — `FunctionToolset`, `CombinedToolset`, and `PreparedToolset` — so `RenamedToolset` was the lone outlier that swallowed it. This adds the same guard and reuses `FunctionToolset`'s two existing error messages.

Before/after: a toolset with tools `a` and `b` wrapped with `.renamed({'b': 'a'})` returned just `{'b'}` (tool `a` silently gone); it now raises `UserError`. A genuine swap `.renamed({'b': 'a', 'a': 'b'})` is unaffected and still returns both tools.

### Checklist

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/toolsets/renamed.py
+++ b/pydantic_ai_slim/pydantic_ai/toolsets/renamed.py
@@ -4,6 +4,7 @@ from dataclasses import dataclass, replace
 from typing import Any
 
 from .._run_context import AgentDepsT, RunContext
+from ..exceptions import UserError
 from .abstract import ToolsetTool
 from .wrapper import WrapperToolset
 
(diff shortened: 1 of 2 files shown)


sean-kim05:
Superseded by #6462, which references issue #6461 (`Closes #6461`). Continuing there.

PR #6470, opened 2026-07-14
Backport #6161 to v1: Add Claude Sonnet 5 (`claude-sonnet-5`) support

basedinesh:
## Summary

Backports #6161 (`beaefc07`, released in v2.2.0) to the `v1` branch: adds `claude-sonnet-5` to the Anthropic model profile — most importantly to `models_that_support_json_schema_output`, so v1 users get native structured outputs (`output_config.format`) instead of the unconstrained tool-call fallback.

## Motivation

On 1.107.1, `anthropic_model_profile('claude-sonnet-5').supports_json_schema_output` is `False`, so structured output for Sonnet 5 rides the tool-call path. We hit this in production-eval work: tool-mode structured output occasionally returns nested array fields as JSON-*encoded strings* (`facts='[{"claim": ...}]'`), which native constrained decoding makes impossible. Sonnet 5 has supported structured outputs since GA ([docs](https://platform.claude.com/docs/en/build-with-claude/structured-outputs)), confirmed live via the Models API:

```
GET /v1/models/claude-sonnet-5 → capabilities.structured_outputs.supported: true
```

The v1 branch got the same treatment for Fable 5 / Mythos 5 in #5849, so this follows that precedent for the remaining Claude 5-generation model.

## Changes vs the original commit

Clean cherry-pick (`-x`) with two v1 adaptations:

- Dropped the `supports_dynamic_filtering` block — that profile field is v2-only.
- `test_model_profile_sonnet_5` converted from v2's `profile.get(...)` style to v1's attribute-access style, minus the dynamic-filtering assertion.

## Verification

- `pytest tests/profiles/test_anthropic.py tests/test_capabilities.py` — 739 passed
- `ruff check` / `ruff format --check` clean on touched files

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6470?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


adtyavrdhn:
Hey @basedinesh 

Any reason you cannot move over to v2?

adtyavrdhn:
https://github.com/pydantic/pydantic-ai/pull/5849 

This was not back ported to v1 but it was just v1 at the time

basedinesh:
hey @adtyavrdhn — fair question, two-part answer:

**why we're on v1**: production agent-execution system mid-deployment, inference transport pinned `<2`. Our layer is built on the v1 surface (adapters, streaming, usage normalization, `anthropic_cache` mechanics), so v2 is a real scoped migration for us — it's on the roadmap, but not something we'll rush mid-push. So "can you move to v2" — yes, planned; "instead of this PR" — not this week.

**on #5849** — fair point, it predates the v1/v2 split so there's no backport precedent; entirely your call. I'd just frame the ask differently: this isn't really new-model *support*, it's that on 1.107.x (which is still cutting releases) `claude-sonnet-5` + structured output silently rides the unconstrained tool path — degraded output with no error, which is closer to a correctness gap than a feature. Cherry-pick is clean and green.

That said, zero pressure: we've since unblocked ourselves with a runtime capability probe on our side (Models API `capabilities.structured_outputs`), so if v1 is strictly fixes-only, feel free to close — it becomes moot when we migrate. The ordering bug I MRE'd over in #6471 (profile-default native mode skips the `strict=True` forcing) affects v2 as well and is probably the more valuable thing to land.

adtyavrdhn:
The releases being cut are only for security vulnerabilities and not for ongoing feature work so closing this.

PR #6478, opened 2026-07-14
Fix Mistral streamed structured output rejecting integer-valued JSON numbers

### LINKED ISSUE #6477: Mistral streamed structured output rejects integer-valued JSON `number` fields
### Description

When using **streamed** structured output with the Mistral model, a schema-valid response can be silently dropped whenever a required JSON-Schema `number` field holds an integer.

`MistralStreamedResponse._try_get_output_tool_from_text` parses the accumulated stream text with `pydantic_core.from_json` and gates the result through `_validate_required_json_schema`, which checks required fields with `isinstance(value, VALID_JSON_TYPE_MAPPING[type])` where `number` maps to `float`. JSON has a single `number` type, and `pydantic_core.from_json` (like `json.loads`) parses `20` to a Python `int`, so `isinstance(20, float)` is `False`. The value is rejected, the validator returns `False`, `_try_get_output_tool_from_text` returns `None`, and the structured-output `ToolCallPart` is never emitted — diverging from the non-streaming path and from pydantic itself (which coerces `int` → `float`). Nested `number` fields and arrays of `number` items are affected too. Secondarily, because `bool` subclasses `int`, a JSON boolean is wrongly accepted for an `integer` field.

### Minimal reproduction

```python
from pydantic_core import from_json

# JSON has one number type; a fraction-less number parses to int:
print(type(from_json('{"temperature": 20}')["temperature"]))  # <class 'int'>

# the validator's check for a JSON-Schema `number` field:
print(isinstance(20, float))   # False  -> schema-valid `number` value rejected
print(isinstance(True, int))   # True   -> JSON boolean wrongly accepted for an integer field
```

End to end, a streamed Mistral run whose structured output contains e.g. `{"temperature": 20}` for a `{"type": "number"}` field yields no output tool call, while the same payload works on the non-streaming path.

### Expected

Streaming and non-streaming should agree, and both should follow JSON-Schema/pydantic numeric semantics: a `number` field accepts `int` or `float` (not `bool`); an `integer` field accepts `int` (not `bool`).

I have a fix ready (a small `_matches_json_type` helper) with regression tests and am happy to finish it.

sohumt123:
## Problem

When using streamed structured output with the Mistral model, a schema-valid response could be silently dropped.

`MistralStreamedResponse._try_get_output_tool_from_text` parses the accumulated stream text with `pydantic_core.from_json` and then gates the result through `MistralStreamedResponse._validate_required_json_schema`. That validator checks each required field with `isinstance(value, VALID_JSON_TYPE_MAPPING[type])`, where `number` maps to `float`.

JSON has a single `number` type, and `pydantic_core.from_json` (like `json.loads`) parses a fraction-less number such as `20` to a Python `int`:

```python
import pydantic_core
pydantic_core.from_json('{"value": 20}')['value']  # 20, a Python int
isinstance(20, float)                               # False  -> schema-valid `number` rejected
```

Because `isinstance(20, float)` is `False`, a schema-valid `number` value was rejected, `_validate_required_json_schema` returned `False`, `_try_get_output_tool_from_text` returned `None`, and the structured-output `ToolCallPart` was never emitted. This diverges from the non-streaming path and from pydantic itself, which coerces `int` to `float` for `float` fields. Nested `number` fields and arrays of `number` items were affected the same way.

There is also a secondary correctness issue in the same check: Python's `bool` is a subclass of `int`, so `isinstance(True, int)` is `True`, meaning a JSON boolean was wrongly accepted for an `integer` field.

## Fix

Add a small module-private helper `_matches_json_type(value, json_type)` and use it in both the array-items branch and the scalar branch of `_validate_required_json_schema`:

- `number` accepts `int` or `float`
- `integer` accepts `int`
- `boolean` accepts `bool`
- `bool` is excluded from the numeric types (since it subclasses `int`), so a JSON boolean is no longer accepted for an `integer`/`number` field
- all other types keep the existing `VALID_JSON_TYPE_MAPPING` behavior

This is a behavior-tightening bug fix: a JSON boolean is no longer accepted for a required `integer`/`number` field. That is strictly more correct per JSON Schema, and it only means a malformed payload now withholds the streamed `ToolCallPart` instead of emitting one that pydantic would reject downstream anyway. No public API changes.

## Testing

- Extended the existing parametrized `test_validate_required_json_schema` with cases covering: `number` accepts an int, `number` accepts a float, `integer` rejects a bool, `number` rejects a bool, a nested `number` accepts an int, an array of `number` accepts ints, and an array of `integer` rejects a bool item. The 6 pre-existing cases still pass unchanged.
- Added `test_try_get_output_tool_from_text_number_accepts_int`, which builds a `ToolDefinition` with a required `number` field, calls `_try_get_output_tool_from_text('{"value": 20}', ...)`, and asserts it returns a `ToolCallPart` with `args_as_dict() == {'value': 20}` rather than `None` — proving the user-visible symptom is fixed.

Before the fix these new tests fail (6 failing), after the fix the whole module passes:

```
uv run pytest tests/models/test_mistral.py -q
64 passed
```

`ruff format`, `ruff check`, and `pyright` are clean on both changed files.

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6476?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Fixes #6477

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/mistral.py
+++ b/pydantic_ai_slim/pydantic_ai/models/mistral.py
@@ -817,9 +817,9 @@ class MistralStreamedResponse(StreamedResponse):
                 if not isinstance(json_dict[param], list):
                     return False
                 for item in json_dict[param]:
-                    if not isinstance(item, VALID_JSON_TYPE_MAPPING[param_items_type]):
+                    if not _matches_json_type(item, param_items_type):
                         return False
-            elif param_type and not isinstance(json_dict[param], VALID_JSON_TYPE_MAPPING[param_type]):
+            elif param_type and not _matches_json_type(json_dict[param], param_type):
                 return False
 
             if isinstance(json_dict[param], dict) and 'properties' in param_schema:
@@ -828,6 +828,23 @@ class MistralStreamedResponse(StreamedResponse):
                     return False
 
         return True
+
+
+def _matches_json_type(value: Any, json_type: str) -> bool:
+    """Check whether a Python value parsed from JSON matches a JSON Schema type.
+
+    JSON has a single `number` type, and `pydantic_core.from_json` (like `json.loads`) parses a
+    fraction-less number such as `20` to a Python `int`, so a `number` field must accept both `int`
+    and `float`. Python's `bool` is a subclass of `int`, so it is excluded from the numeric types to
+    avoid accepting a JSON boolean for an `integer` or `number` field.
+    """
+    if json_type == 'number':
+        return isinstance(value, (int, float)) and not isinstance(value, bool)
+    elif json_type == 'integer':
+        return isinstance(value, int) and not isinstance(value, bool)
+    elif json_type == 'boolean':
+        return isinstance(value, bool)
+    return isinstance(value, VALID_JSON_TYPE_MAPPING[json_type])
 
 
 VALID_JSON_TYPE_MAPPING: dict[str, Any] = {
--- a/tests/models/test_mistral.py
+++ b/tests/models/test_mistral.py
@@ -75,6 +75,7 @@ with try_import() as imports_successful:
     from pydantic_ai.models.openai import OpenAIResponsesModel, OpenAIResponsesModelSettings
     from pydantic_ai.providers.mistral import MistralProvider
     from pydantic_ai.providers.openai import OpenAIProvider
+    from pydantic_ai.tools import ToolDefinition
 
     MockChatCompletion = MistralChatCompletionResponse | Exception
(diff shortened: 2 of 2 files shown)


adtyavrdhn:
Hey @sohumt123 

Thanks, we'll take over in #6479

PR #6490, opened 2026-07-14
fix(groq,huggingface): skip bare assistant message for empty ModelResponse

### LINKED ISSUE #6364: Groq and HuggingFace adapters send an invalid empty assistant message (400) on the empty-response retry

Mukller:
## Summary

Fixes #6364.

`GroqModel._map_messages` and `HuggingFaceModel._map_messages` unconditionally appended
an assistant message for every `ModelResponse`, even when `parts` was empty.
A bare `{'role': 'assistant'}` is rejected with HTTP 400 by the Chat Completions API.

This breaks the empty-response retry flow: the agent records `ModelResponse(parts=[])`
and on the next request the adapter turns it into an invalid bare message, crashing
instead of recovering.

The OpenAI, Anthropic, Mistral, and Cohere adapters already guard this (tracked
in #6302 and #6138). This PR mirrors the same fix for the two adapters that were missed.

## Fix

Wrap the `message_param` construction and `append` in `if texts or tool_calls:`.
Empty `ModelResponse` objects produce neither, so they are silently skipped — matching
the OpenAI adapter behaviour.

## Files changed
- `pydantic_ai_slim/pydantic_ai/models/groq.py`
- `pydantic_ai_slim/pydantic_ai/models/huggingface.py`

Closes #6364.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6490?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/groq.py
+++ b/pydantic_ai_slim/pydantic_ai/models/groq.py
@@ -558,14 +558,15 @@ class GroqModel(Model[AsyncGroq]):
                         pass
                     else:
                         assert_never(item)
-                message_param = chat.ChatCompletionAssistantMessageParam(role='assistant')
-                if texts:
-                    # Note: model responses from this model should only have one text item, so the following
-                    # shouldn't merge multiple texts into one unless you switch models between runs:
-                    message_param['content'] = '\n\n'.join(texts)
-                if tool_calls:
-                    message_param['tool_calls'] = tool_calls
-                groq_messages.append(message_param)
+                if texts or tool_calls:
+                    message_param = chat.ChatCompletionAssistantMessageParam(role='assistant')
+                    if texts:
+                        # Note: model responses from this model should only have one text item, so the following
+                        # shouldn't merge multiple texts into one unless you switch models between runs:
+                        message_param['content'] = '\n\n'.join(texts)
+                    if tool_calls:
+                        message_param['tool_calls'] = tool_calls
+                    groq_messages.append(message_param)
             else:
                 assert_never(message)
         if instruction_parts := self._get_instruction_parts(messages, model_request_parameters):
--- a/pydantic_ai_slim/pydantic_ai/models/huggingface.py
+++ b/pydantic_ai_slim/pydantic_ai/models/huggingface.py
@@ -408,14 +408,15 @@ class HuggingFaceModel(Model[AsyncInferenceClient]):
                         pass
                     else:
                         assert_never(item)
-                message_param = ChatCompletionInputMessage(role='assistant')
-                if texts:
-                    # Note: model responses from this model should only have one text item, so the following
-                    # shouldn't merge multiple texts into one unless you switch models between runs:
-                    message_param['content'] = '\n\n'.join(texts)
-                if tool_calls:
-                    message_param['tool_calls'] = tool_calls
-                hf_messages.append(message_param)
+                if texts or tool_calls:
+                    message_param = ChatCompletionInputMessage(role='assistant')
+                    if texts:
+                        # Note: model responses from this model should only have one text item, so the following
+                        # shouldn't merge multiple texts into one unless you switch models between runs:
+                        message_param['content'] = '\n\n'.join(texts)
+                    if tool_calls:
+                        message_param['tool_calls'] = tool_calls
+                    hf_messages.append(message_param)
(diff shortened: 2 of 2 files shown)


Mukller (commented):
## Self-Review

### Root Cause
Both `GroqModel._map_messages` and `HuggingFaceModel._map_messages` unconditionally append
an assistant message for every `ModelResponse`. When `parts=[]`, the message is a bare
`{'role': 'assistant'}` — rejected with HTTP 400 by Chat Completions APIs.

### Fix
Guard with `if texts or tool_calls:`. Empty responses contribute neither and are silently
skipped. This matches the existing OpenAI adapter behaviour (`_into_message_param` returns
`None` for empty responses) and the fix applied in #6302 for Mistral and Cohere.

### Correctness
- Non-empty responses are unaffected — the guard passes and the existing logic runs unchanged.
- Empty responses are skipped — the retry flow proceeds without injecting an invalid message.
- Two-adapter sweep: both Groq and HuggingFace are fixed in the same PR.

### No new API surface
Both changes are internal to private `_map_messages` methods.

dsfaccini:

Thanks for the PR. We checked both providers against the live APIs and neither rejects an empty assistant message, so this isn't a 400 bug and the guard isn't needed:

- Groq: verified live by a maintainer, who closed the identical fix in https://github.com/pydantic/pydantic-ai/pull/6365 — "Groq is fine with empty message".
- HuggingFace: verified live via the `InferenceClient.chat_completion` path the adapter uses, on `meta-llama/Llama-3.3-70B-Instruct` and `Qwen/Qwen2.5-7B-Instruct`, for both a bare `{'role': 'assistant'}` and `{'role': 'assistant', 'content': None}`. All return 200.

The reproduction in #6364 pinned the mapped request payload, not an actual API 400. Closing this and #6364 as not planned.

PR #6512, opened 2026-07-15
feat(anthropic): add AnthropicJsonSchemaTransformer for native structured output

### LINKED ISSUE #6471: Anthropic profile-default native structured output skips schema transformation
This issue was posted by Codex Desktop using gpt-5.6-terra on behalf of David.

### Description

Anthropic's profile does include `AnthropicJsonSchemaTransformer`; the original report was wrong about that. Explicit `NativeOutput(...)` also works, because `AnthropicModel.prepare_request()` forces the output schema to `strict=True` before the transformer runs.

The actual bug is narrower: setting native structured output as the model profile's default causes a plain Pydantic output type to select native output *after* schema customization. The transformer consequently sees `strict=None`, leaves the Pydantic schema unchanged, and Anthropic rejects it because it requires the transformed strict-schema dialect.

This is supported usage. In v2, `profile=` accepts a partial `ModelProfile` dict which is merged onto the provider's resolved profile; `default_structured_output_mode` is the public setting that controls how an otherwise `auto` structured-output request is resolved.

### MRE

```python
from pydantic import BaseModel, Field
from pydantic_ai import Agent


class Decision(BaseModel):
    confidence: float = Field(ge=0.0, le=1.0)
    title: str = Field(min_length=1)


model = AnthropicModel(
    'claude-opus-4-8',
    provider=AnthropicProvider(),
    profile={'default_structured_output_mode': 'native'},
)
agent = Agent(model, output_type=Decision)
result = agent.run_sync('Ship it? Confidence in [0,1] + short title.')
```

On v2.9.1, Anthropic returns a 400 such as:

```
output_config.format.schema: For 'object' type, 'additionalProperties' must be explicitly set
```

The equivalent `output_type=NativeOutput(Decision)` path succeeds because it reaches the strict-schema transformation before the request is sent.

### Expected behavior

When native structured output is selected by `default_structured_output_mode='native'`, it should receive the same Anthropic schema transformation as explicit `NativeOutput(...)`: the schema must be transformed before being sent, including the required `additionalProperties: false` and conversion of unsupported constraints.

### Root cause

`AnthropicModel.prepare_request()` checks whether the output mode is native before `Model.prepare_request()` resolves `auto` using the profile default. The base method performs schema customization before that resolution. Therefore the Anthropic strict forcing and transformation are skipped for the profile-default-native path.

### Scope

This is a main/v2 regression test and ordering fix. It is unrelated to the closed v1-only model-profile backport request in #6470.

Mukller:
## Summary

Fixes #6471.

The Anthropic model profile had no `json_schema_transformer`, so native structured
output (`NativeOutput` / `default_structured_output_mode='native'`) sent the raw
Pydantic-generated JSON schema to the API. Anthropic's constrained decoder rejects
several keywords that Pydantic emits routinely, returning HTTP 400
"unsupported feature".

## Root Cause

Common Pydantic model constraints produce schema keywords that Anthropic doesn't support:

| Pydantic | JSON Schema keyword | Anthropic?  |
|----------|---------------------|-------------|
| `ge` / `le` / `gt` / `lt` | `minimum`, `maximum`, `exclusiveMinimum/Maximum` | ❌ rejected |
| `multiple_of` | `multipleOf` | ❌ rejected |
| `min_length` / `max_length` | `minLength`, `maxLength` | ❌ rejected |
| list with `max_length` | `maxItems` | ❌ rejected |
| `min_length > 1` on list | `minItems > 1` | ❌ rejected (only 0 or 1 allowed) |
| arbitrary `format` | `format: binary` etc. | ❌ rejected unless in allowlist |
| object without `additionalProperties` | (omitted) | ❌ rejected (must be `false`) |

`OpenAIJsonSchemaTransformer` and `GoogleJsonSchemaTransformer` already handle
exactly this class of dialect mismatch. This PR adds the missing Anthropic version.

## Fix

`AnthropicJsonSchemaTransformer` (new class in `profiles/anthropic.py`):

1. **Strips unsupported constraint keywords** (`minimum/maximum/exclusiveMin/Max`,
   `multipleOf`, `minLength`, `maxLength`, `maxItems`) and moves their values into
   the `description` field so the model still sees the semantic intent.
2. **Clamps `minItems > 1` to `1`** (the maximum allowed).
3. **Filters `format`** to Anthropic's documented allowlist; unsupported values are
   moved to `description`.
4. **Sets `additionalProperties: false`** on object schemas (required by the API).

The transformer is injected into the profile via
`json_schema_transformer=AnthropicJsonSchemaTransformer` when
`supports_json_schema_output=True`. This mirrors how the OpenAI and Google profiles
apply their transformers.

## Example

```python
class Decision(BaseModel):
    confidence: float = Field(ge=0.0, le=1.0)
    title: str = Field(min_length=1)

# Before: POST /v1/messages → 400 "unsupported feature" (minimum, maximum, minLength)
# After:  POST /v1/messages → 200 OK, model-constrained JSON response
result = await agent.run("...", output_type=NativeOutput(Decision))
```

Closes #6471.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6512?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/profiles/anthropic.py
+++ b/pydantic_ai_slim/pydantic_ai/profiles/anthropic.py
@@ -2,6 +2,8 @@ from __future__ import annotations as _annotations
 
 from typing import Literal, TypeAlias
 
+from dataclasses import dataclass
+
 from ..native_tools import (
     CodeExecutionTool,
     MCPServerTool,
@@ -12,6 +14,7 @@ from ..native_tools import (
 from ..native_tools._tool_search import ToolSearchTool
 from ..settings import ThinkingEffort, ThinkingLevel
 from . import ModelProfile
+from .._json_schema import JsonSchema, JsonSchemaTransformer
 
 _ANTHROPIC_BASE_BUILTINS = frozenset({WebSearchTool, CodeExecutionTool, WebFetchTool, MemoryTool, MCPServerTool})
 """Native tool types Anthropic generally supports across the model line. Mirrors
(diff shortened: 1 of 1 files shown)


Mukller (commented):
## Self-Review

### Root Cause
The Anthropic profile had no `json_schema_transformer`. When `supports_json_schema_output=True`
and the user uses `NativeOutput` (or a profile with `default_structured_output_mode='native'`),
pydantic-ai sends the raw Pydantic JSON schema to the API. Anthropic's constrained decoder
rejects numeric/string/array constraint keywords and any object missing
`additionalProperties: false`, returning HTTP 400.

### Fix
`AnthropicJsonSchemaTransformer` strips/transforms all the documented unsupported keywords
while preserving semantic intent in `description`. It follows the same pattern as
`OpenAIJsonSchemaTransformer` and `GoogleJsonSchemaTransformer`.

### Key design decisions
- **Description annotation**: stripped constraints are appended to `description` rather than
  silently dropped, so the model still sees the constraint intent even though it's not enforced
  by the constrained decoder. Pydantic validates the final output against the full schema
  pydantic-ai-side anyway.
- **`additionalProperties: false`**: set with `setdefault` so user-provided values are not
  overwritten.
- **`minItems` clamping**: values > 1 are clamped to 1 (the max Anthropic allows) rather than
  dropped entirely, preserving the "at least one element" intent.

### Correctness
- All non-Anthropic code paths are unaffected — the transformer is only injected when
  `supports_json_schema_output=True`.
- The transformer is pure (reads `schema`, returns modified `schema`) — consistent with
  the `JsonSchemaTransformer` protocol.

### Scope
One new class + two small additions to the existing file. No changes to other providers or
the base transformer.

adtyavrdhn:
Please let @dsfaccini conclude what the issue is, there is a `json_schema_transformer` on Anthropic 

https://github.com/pydantic/pydantic-ai/blob/806fa7acbe2466278fc9d1da51be1483a2d85018/pydantic_ai_slim/pydantic_ai/providers/anthropic.py#L66

Issue #3885, opened 2025-12-30
Provider Citation/Annotation/Grounding Type Definitions for Unified Citation Design


certainly-param:
## Related PR

This issue was created as part of [PR #3729](https://github.com/pydantic/pydantic-ai/pull/3729) based on @DouweM feedback to research and propose a unified citation design.

## Summary

This issue documents research on all provider-specific citation/annotation/grounding data types and proposes a unified citation representation. The goal is to design a unified interface so users can deal with citations from all providers in one consistent way, without needing to know provider-specific details.

### OpenAI (Chat Completions + Responses API)

**SDK Types:**
- `openai.types.chat.chat_completion_message.Annotation`
  - `type: Literal["url_citation"]`
  - `url_citation: UrlCitation | None`
- `openai.types.chat.chat_completion_message.UrlCitation`:
  - `url: str`
  - `title: str | None`
  - `start_index: int` (character offset, inclusive)
  - `end_index: int` (character offset, exclusive)
- `openai.types.responses.ResponseOutputTextAnnotationAddedEvent` (streaming events)
- `openai.types.responses.annotation.ContainerFileCitation`:
  - `type: Literal["container_file_citation"]`
  - `container_id: str`
  - `file_id: str`
  - `filename: str | None`
  - `start_index: int`
  - `end_index: int`

**Sample Data (from cassettes):**

OpenAI Responses API `url_citation`:
```json
{
  "type": "url_citation",
  "url_citation": {
    "url": "https://www.accuweather.com/en/us/st-louis/63102/current-weather/349084",
    "title": "St. Louis, MO Current Weather | AccuWeather",
    "start_index": 541,
    "end_index": 659
  }
}
```

OpenAI Responses API `container_file_citation`:
```json
{
  "type": "container_file_citation",
  "container_id": "cntr_68dd936a4cfc81908bdd4f2a2f542b5c0a0e691ad2bfd833",
  "file_id": "cfile_68dd937c234081919ee5238a16b4fd87",
  "filename": "cfile_68dd937c234081919ee5238a16b4fd87.png",
  "start_index": 0,
  "end_index": 0
}
```

### Anthropic

**SDK Types:**
- `anthropic.types.beta.BetaCitationsDelta` (streaming)
  - `citation: BetaCitationsWebSearchResultLocation | BetaCitationSearchResultLocation | None`
- `anthropic.types.beta.BetaCitationsWebSearchResultLocation`:
[...]

PR #6516, opened 2026-07-15
fix(bedrock): handle citationsContent blocks instead of silently dropping them (#6455)

### LINKED ISSUE #6455: [provider-mapping-sweep] bedrock: add citationsContent block / citation delta mapping (currently unmapped, not yet reachable via public API)
> **Note (2026-07-15):** This was originally filed by the `provider-mapping-sweep` automation as a `p:2-high` bug. After investigation ([details](https://github.com/pydantic/pydantic-ai/issues/6455#issuecomment-4982252847)) it's been reclassified as a **`feature` / `p:4-low` gap**: it is not reachable through the public API today. The original report is preserved below the line, with corrections inline.

## Summary (corrected)

The Bedrock Converse adapter has no mapping for `citationsContent` content blocks (non-streaming) or `citation` deltas (streaming). This is a **real but latent gap**, not a production bug:

- **Not reachable via the public API.** Bedrock only emits `citationsContent` when the *request* opts in per-document via `document.citations.enabled = true`. Our document builder (`_make_document_block`) never sets that flag, and there is no public API (nor `bedrock_additional_model_requests_fields` path) to enable it. So no current caller can receive a response containing `citationsContent`. The reproduction below feeds a **hand-constructed** `ConverseResponseTypeDef` directly to `_process_response` — it proves the mapping is missing, but does not correspond to anything the public API can currently produce.
- **Streaming text is *not* lost.** In the streaming path the generated text still arrives via the normal `delta['text']` and is emitted today. The separate `citation` delta (`ContentBlockDeltaTypeDef.citation`, carrying only citation *metadata*) is what's dropped. So the "lost generated text" symptom applies only to the non-streaming path, where the text is nested under `citationsContent.content[].text`.
- **Precedent caveat.** The original report cited Anthropic/OpenAI as handling citations. Note that even for Anthropic — where citation enablement *is* wired up and user-reachable (`WebFetchTool.enable_citations`) — the response side is currently an intentional no-op (`# TODO(Marcelo): We need to handle citations.` / `elif isinstance(event.delta, BetaCitationsDelta): pass`). So the codebase treats citation *response*-parsing as deferred work even where it's reachable.

## Suggested direction

Close this end-to-end rather than response-only: add request-side citation support for Bedrock (enable `citations` on document blocks), then map the response (`citationsContent.content[].text` → `TextPart`, with `citations` on `TextPart.provider_details`), ideally alongside finishing the Anthropic citation TODO so the two stay consistent. Unified cross-provider citation typing is tracked in #3126 / #3885.

---

<sub>Original automated report (retained for reference; see corrections above):</sub>

## Impact

When Bedrock Converse returns a `citationsContent` content block (the wire-format wrapper for **document-grounded** Claude responses, also used by Knowledge Base / RAG flows that enable citations), **the response parser silently drops the entire block — including the model's generated text and the citation list**. The user loses both the answer snippet and the source/location metadata.

The same omission also drops standalone `text` deltas inside `citation` deltas during streaming, so streamed citation-grounded responses are partial as well. *(Correction: streaming generated text is not dropped — only the citation metadata delta is; see Summary above.)*

Per `botocore/data/bedrock-runtime/2023-09-30/service-2.json`, `ContentBlock` is a union whose members include (among others) `text`, `image`, `document`, `toolUse`, `toolResult`, `cachePoint`, `reasoningContent`, `citationsContent`, and `searchResult`. The pydantic-ai Bedrock adapter handles `text` / `toolUse` / `toolResult` / `cachePoint` / `reasoningContent` and silently ignores the rest.

This is in scope for the provider-mapping-sweep audit (`Dropped or mis-serialized parts: ...citations...`).

## Provider & Code Path

Provider: `bedrock`

- `pydantic_ai_slim/pydantic_ai/models/bedrock.py`
  - `_process_response` (response parser for non-streaming), lines ~708–762.
  - `BedrockStreamedResponse._get_event_iterator` (streaming response parser), `contentBlockDelta` branch, lines ~1592–1637.
- `botocore/data/bedrock-runtime/2023-09-30/service-2.json` defines `CitationsContentBlock` and `CitationsDelta` as valid wire shapes.

## Reproduction (synthetic)

> **Correction:** this constructs a response object by hand and calls `_process_response` directly. It demonstrates the missing mapping, but is not reachable through the public API, because the request side never enables `document.citations`.

```python
import asyncio
from typing import Any, cast
import boto3

client = boto3.client('bedrock-runtime', region_name='us-east-1',
                      aws_access_key_id='x', aws_secret_access_key='y')
provider = BedrockProvider(bedrock_client=client)
m = BedrockConverseModel('us.anthropic.claude-sonnet-4-5-20250929-v1:0',
                         provider=provider)

async def main() -> None:
    response: dict[str, Any] = {
        'output': {'message': {'role': 'assistant', 'content': [
            {'text': 'According to multiple sources,'},
            {'citationsContent': {
                'content': [{'text': ' the Eiffel Tower is in Paris.'}],
                'citations': [{
                    'title': 'Wikipedia',
                    'source': 'en.wikipedia.org/redacted',
                    'sourceContent': [{'text': 'The Eiffel Tower ...'}],
                    'location': {'documentChar': {'documentIndex': 0, 'end': 56, 'start': 0}},
                }],
            }},
        ]}},
        'usage': {'inputTokens': 10, 'outputTokens': 5},
        'stopReason': 'end_turn',
        'ResponseMetadata': {'RequestId': 'test-id'},
    }
    result = await m._process_response(
        cast('pydantic_ai.models.bedrock.ConverseResponseTypeDef', response)
    )
    texts = [p.content for p in result.parts
             if hasattr(p, 'content') and isinstance(p.content, str)]
    print('parts:', len(result.parts))
    print('texts:', texts)

asyncio.run(main())
```

**Observed:** `parts: 1` / `texts: ['According to multiple sources,']` — the `citationsContent` block's text and citations array are dropped by the (currently unreachable) parse path.

## Related

- #3126 — Support structured citations (generic across providers).
- #3885 — Design RFC for a unified citation type.

0ywfe:
## Summary

Closes #6455 — Bedrock Converse's `citationsContent` blocks (document-grounded responses with citation metadata) are silently dropped by both the non-streaming and streaming response parsers. The model's generated text and the citation list are lost.

## Fix

Two sites, matching the issue's suggested fix shape:

**Non-streaming** (`_process_response`, line ~755): when an item has `citationsContent`, emit the `content[*].text` as a `TextPart` with the citation metadata preserved on `provider_details={'citations': [...]}`.

**Streaming** (`_get_event_iterator`, `contentBlockDelta` branch): handle `delta['citation']` by extracting the cited text and emitting it through `handle_text_delta`.

The raw Bedrock citation structure (title, source, sourceContent, location) is stashed on `provider_details` rather than mapped to a unified type, per the issue's out-of-scope note — the cross-provider `Citation` type is tracked on #3126 / #3885.

## Before / after

```
Before — parts: 1, texts: ['According to multiple sources,']
After  — parts: 2, texts: ['According to multiple sources,', ' the Eiffel Tower is in Paris.']
         citation_parts[0].provider_details['citations'][0]['title'] == 'Wikipedia'
```

## Test plan

- [x] 162 existing `test_bedrock.py` tests pass
- [x] No existing test asserts the current drop behavior (confirmed by grep for `citation`/`CitationsContent`/`searchResult` — zero matches)
- [x] Verified the `botocore` service model (`ContentBlock` union) lists `citationsContent` as a valid shape; `ContentBlockDelta` lists `citation`

## Files

- `pydantic_ai_slim/pydantic_ai/models/bedrock.py` (+20)

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6516?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/bedrock.py
+++ b/pydantic_ai_slim/pydantic_ai/models/bedrock.py
@@ -763,6 +763,17 @@ class BedrockConverseModel(Model[BaseClient]):
                                 provider_details={'status': tool_result['status']} if 'status' in tool_result else {},
                             )
                         )
+                elif citations_content := item.get('citationsContent'):
+                    cited_text = ' '.join(
+                        c['text'] for c in citations_content.get('content', []) if 'text' in c
+                    )
+                    if cited_text:
+                        items.append(
+                            TextPart(
+                                content=cited_text,
+                                provider_details={'citations': citations_content.get('citations', [])},
+                            )
+                        )
 
         u = _map_usage(response['usage'], self._provider.name, self.base_url, self.model_name)
         response_id = response.get('ResponseMetadata', {}).get('RequestId', None)
@@ -1631,6 +1642,15 @@ class BedrockStreamedResponse(StreamedResponse):
                                     return_part.content = tr_content[0].get('json')
 
                                 # Don't yield anything yet - we wait for content block end
+                        if citation_delta := delta.get('citation'):
+                            cited_text = ' '.join(
+                                c.get('text', '') for c in citation_delta.get('content', [])
+                            )
+                            if cited_text:
+                                for event in self._parts_manager.handle_text_delta(
+                                    vendor_part_id=index, content=cited_text
+                                ):
+                                    yield event
 
                     case {'contentBlockStop': content_block_stop}:
                         index = content_block_stop['contentBlockIndex']


hramezani:
Thanks for putting this together. Closing this along with #6455 — as detailed in https://github.com/pydantic/pydantic-ai/issues/6455#issuecomment-4982252847, this isn't a bug reachable through the public API today: Bedrock only returns `citationsContent` when the request opts in via `document.citations.enabled = true`, and we never set that flag (no request-side plumbing exists). So there's no code path that can currently produce the response this parses.

It's a legitimate latent gap, but the right shape is to add request-side citation support for Bedrock first (and ideally finish the existing Anthropic citation TODO for consistency), then map the response — rather than a response-only parser for input we never generate. Happy to revisit as part of that larger enhancement.

PR #6324, opened 2026-07-07
Add realtime speech-to-speech support with `Agent.realtime()`

DouweM:
- Closes #1447
- Closes #6622

> **Updated since the last review rounds** (2026-07-22):
> - **API**: the entry point is now `agent.realtime(model, ...).session()` — an `AgentRealtime` accessor that binds the agent's realtime config once (mirroring `iter`) so a session, and (in the stacked WebRTC PR) browser signaling, reuse the same instructions and tools without re-passing them. `Agent.realtime_session()` is removed.
> - **Typed frame parsing**: the OpenAI-protocol codec now parses every inbound WebSocket frame through the OpenAI SDK's typed `openai.types.realtime.*` models (via the SDK's lenient recursive `construct()`), instead of raw dict reads — dispatching on the wire `type` so clone/beta events aren't mis-classified. Transport and every cassette are unchanged.
> - **Azure**: this PR now ships **Azure OpenAI GA realtime** as a single `AzureRealtimeModel` (`provider='azure'` / `AzureProvider`, `api-key` auth, `/openai/v1/realtime`), live-verified. **Azure AI Voice Live** moved to a stacked follow-up, #6642 (additive `azure_voice_live` setting), so nothing here is breaking.
> - **New events / usage**: `InputTranscriptionFailedEvent` (recoverable), input-transcription (ASR) usage reported in `RunUsage.details` separate from response tokens, and `ReconnectedEvent.state_restored` distinguishing providers that restore vs lose conversation state on reconnect.
> - **Gateway**: realtime detects gateway providers via a registry (works for both `provider='gateway/openai'` and a `gateway_provider('openai')` instance) so it adds the required `/v1` WebSocket path either way.
> - **Docs**: `docs/realtime.md` restructured into a `Realtime` sidebar section — an overview (positioning, one-agent-many-modalities, choose-your-path, connecting a frontend, troubleshooting, provider matrix) plus per-provider pages (OpenAI, Azure, xAI, Gemini).
> - Plus review-blocker fixes from a deep review pass (see commits).

Adds realtime (speech-to-speech) support: `Agent.realtime()` opens a bidirectional session against OpenAI Realtime, Gemini Live, xAI Grok Voice, or Azure OpenAI realtime, with the agent's tools, instructions, and capabilities wired through the existing framework infrastructure.

This builds directly on #6042 by @DEENUU1 (thank you Kacper! your commits are preserved in this branch) and reshapes it per our internal design work on realtime integration:

[...]

PR #6570, opened 2026-07-17
Add experimental Amazon Nova Sonic realtime support (`BedrockRealtimeModel`)

### LINKED ISSUE #1447: Support for OpenAI Realtime API and equivalents (AWS Nova Sonic, etc)
### Description

The OpenAI Agents SDK is probably one of the first SDKs that supports voice agents OOTB and I think that it would be a nice addition to PydanticAI.

Essentially, these are native speech-to-speech models that can perform tool calling, transcription, interruption detection, etc. 

The implementation differs slightly between providers (Google, AWS, OpenAI...) but the idea is similar with normal LLMs: send user input, capture all the steps that happen (tool calls, partial transcriptions) and then returning output.

OpenAI Agents SDK have implemented them using a `VoicePipeline`, as shown here:

https://openai.github.io/openai-agents-python/voice/quickstart/

Although probably not as pressing as supporting Prompt Caching, this is definitely a use case that people are willing to switch framework for, especially companies looking to augment customer service with AI Agent triage.

Does it sound like something that Pydantic would be interested in supporting?

### References

_No response_

DouweM:
Relates to #1447 (which lists "AWS Nova Sonic" as one of the realtime equivalents people asked for). The base realtime PR #6324 already closes #1447; this is an **additive provider** on top of it.

## Status: draft / exploratory — do not merge yet

This is a feasibility spike, opened so the work isn't lost and so we can decide **whether we want it at all** before investing in finishing it. It targets the realtime PR's branch (`realtime-v1`), not `main`, so it rides along with that work rather than blocking it.

**The good news:** Amazon Nova Sonic maps cleanly onto the `RealtimeModel` / `RealtimeConnection` abstractions introduced in #6324, even though its transport is completely different — Bedrock's `InvokeModelWithBidirectionalStream`, an **HTTP/2 bidirectional JSON-event stream, not WebSockets**. The connection ABC turned out to be transport-agnostic enough that it slots in, and the wire events map almost one-to-one onto our codec (`audioOutput`→`AudioDelta`, USER/ASSISTANT `textOutput`→`InputTranscript`/`Transcript`, `toolUse`→`ToolCall`, `usageEvent`→`SessionUsageEvent`, `completionEnd`→`TurnCompleteEvent`). So this doubles as evidence the abstraction generalises past WebSocket providers.

## The decision that gates everything: the experimental Smithy SDK

Nova Sonic's bidirectional API has **no boto3 support**. The only Python path is the experimental [`aws-sdk-bedrock-runtime`](https://github.com/awslabs/aws-sdk-python) (a new Smithy-generated async SDK), pinned here to `0.4.0`. Before anything else, we need a maintainer call on whether we're willing to take that dependency:

- It's a **separate SDK stack** from the boto3 one our existing `BedrockConverseModel` uses — different client, config, and auth machinery. There's no reusing `BedrockProvider` (it owns a boto3 client whose transport can't invoke the bidirectional operation).
- It's **`0.x` and explicitly experimental**, with a churny public surface.
- It requires **Python 3.12+**, so the `realtime-bedrock` extra is marker-empty on 3.10/3.11 — a user there can `pip install` it but not import the module.

If the answer is "not worth it," this PR should close and we note in #1447 that Nova Sonic is blocked on upstream SDK maturity. Everything below only matters if we say yes.

## Blockers before this could go non-draft

1. **Auth is unverified live.** The environment I tested in has a Bedrock **bearer token** (`AWS_BEARER_TOKEN_BEDROCK`), but `aws-sdk-bedrock-runtime` only exposes **SigV4** through its public config — no bearer auth scheme. The live smoke reached the real SDK invocation and then timed out with no events and no auth error, so **no live transcript/audio behaviour has actually been verified**. We need to confirm the bidirectional endpoint accepts Bedrock API keys (and get the SDK to expose a bearer scheme) or test with real SigV4 credentials before trusting any of this end to end.
2. **No transparent reconnect / session continuation.** Nova has an ~8-minute connection limit; AWS documents continuing by opening a new stream and re-seeding prior turns as text. `reconnect` is currently a stub — a drop surfaces as a non-recoverable `SessionErrorEvent`. Real support means snapshotting normalised text history, re-dialling before the limit, resending the handshake + history, and emitting `ReconnectedEvent` only after a successful reseed.
3. **No cassette transport → thin, non-recorded tests.** The WS cassette engine (`ws_cassettes.py`) patches `websockets` and doesn't apply to HTTP/2 Smithy streams. Tests here run against a **fake bidirectional stream** and cover the handshake, settings merge, and speculative-vs-final transcript selection — but tool round-trips, interruption, stream exceptions, history seeding, and the unsupported-input guards still need coverage, and none of it is backed by a real recording. `RealtimeCassette`'s JSON-frame format could gain a Smithy adapter (patching `invoke_model_with_bidirectional_stream`, recording input `send()`s and replaying `await_output()` receivers) — sketched, not built.
4. **The `bedrock:` inference test can't join the shared suite yet.** Because the SDK is 3.12+, unconditionally constructing a `BedrockRealtimeModel` in `test_inference.py` would break CI on 3.10/3.11. Needs a skip/guard story before it's safe in the common path.
5. **`BedrockProvider` reconciliation is unresolved.** To avoid a second public provider class, `BedrockRealtimeModel(provider=...)` currently accepts a structurally-typed Smithy client and otherwise builds the default. A future shared Bedrock auth-config object could reconcile the boto3 and Smithy stacks without pretending the client types are interchangeable — but that's undesigned.

## What's implemented

- `BedrockRealtimeModel` + `BedrockRealtimeConnection` over the Smithy duplex stream: session/prompt/content handshake, 16 kHz LPCM audio in / 24 kHz out, text turns, tools + tool results, USER/ASSISTANT transcripts, usage, provider errors, and text/transcript history seeding.
- `BedrockRealtimeModelSettings` following the migrated settings pattern (`voice`, `max_tokens`, `temperature`, `top_p`, and Nova 2's `bedrock_endpointing_sensitivity`), merged model-level + per-connect.
- `bedrock:` prefix in `infer_realtime_model` / `KnownRealtimeModelName` for both `amazon.nova-2-sonic-v1:0` (default) and `amazon.nova-sonic-v1:0`.
- Speculative/final handling: only **FINAL** assistant text becomes a `Transcript` (speculative text describes planned, not-yet-spoken audio; forwarding it would duplicate or retain unspoken text). Automatic server-side barge-in is still supported even though explicit client `interrupt()` is not.
- A dedicated `realtime-bedrock` optional extra, docs in `docs/realtime.md` + `docs/api/realtime.md`, and a `live_bedrock_smoke.py` (no `test_` prefix, never runs in CI).

## Review note

I reviewed the generated code line-by-line and fixed one real bug before opening this: the usage adapter forwarded `usageEvent.details.total` (cumulative), but the session does `usage.incr()` per `SessionUsageEvent`, so a multi-turn session would have re-added the whole running total every turn. It now forwards `details.delta` (the per-turn increment), with the test fixture carrying both keys to prove `total` is ignored.

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6570?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


DouweM:
### Reconciliation notes for after the realtime SDK refactor (`realtime-v1`) merges

Heads-up from the realtime-v1 work (the big SDK-typed-parse + `RealtimeCodecEvent`/`RealtimeSession` reshape): this PR predates that codec, so it'll need reconciliation before it can come out of exploratory status. I also did a comparison against Mastra's `voice/aws-nova-sonic` — good news, **our provider-codec/session separation is cleaner than theirs** (they execute tools inside the voice object; we keep the adapter thin and let `RealtimeSession` own tool execution/history/usage). Mastra mainly corroborates the AWS event sequence; it does **not** solve our real blocker (the experimental Python SDK + a live SigV4 recording).

Concrete reconciliation checklist against the current `realtime/` codec:

1. **Adopt the new two-layer event model.** Map into the current `RealtimeCodecEvent` union in `_base.py` (audio delta, `Transcript`/`InputTranscript`, `ToolCall`, speech start/end, `TurnCompleteEvent`, `SessionUsageEvent`, `SessionErrorEvent`, plus handshake/session identity where Nova exposes it). Match the current `RealtimeModel.connect()` signature and keep tool execution **out** of `bedrock.py` — `_session.py` already owns concurrent calls, retries, cancellation, usage accumulation, and normalized history.
2. **Fix the profile.** `ClearAudio`/`CreateResponse` currently raise `NotImplementedError` and `CommitAudio` just closes the audio container, so `supports_manual_turn_control` must be **`False`** (closing an audio block ≠ full manual turn control). `supports_interruption` can be `True` (server-side barge-in); `supports_output_truncation` `False`.
3. **Content-scoped mapping state.** Tie `_role`/`_generation_stage` to Nova content blocks (key by `contentName` if present), clear at `contentEnd`, reset at `completionEnd`. Parse the interruption sentinel and `additionalModelFields` as **structured JSON**.
4. **Incremental usage.** Keep mapping `usageEvent.details.delta` (the session calls `RunUsage.incr()`), not cumulative totals. Add a two-turn regression; handle the tool-only-completion → final-completion ordering like the OpenAI mapper's `response_usage_follows`.
5. **Redo seeding via the Gemini pattern.** Current OpenAI/Gemini seeding is exhaustive (multimodal, tool call/result pairing, retry/deferred, compaction, explicit rejection of unrepresentable parts). Since Nova only accepts USER/ASSISTANT text, follow `google.py`: render a documented lossy text form for tool rounds, reject what can't be represented, use the shared seed helpers, and test incomplete call/result pairs. Don't silently drop parts.
6. **Cassettes.** Needs a Smithy/bidi-stream cassette adapter that records logical frames (secret-scrubbed) + at least one real SigV4 recording. Reconnect should stay unsupported until it actually opens a new authenticated stream and reseeds normalized history.

Two specifics from Mastra worth **adopting**: structural-JSON interruption detection, and handling all documented interruption locations (`completionEnd.stopReason == INTERRUPTED` *and* `contentEnd`). Two to **avoid**: their broad substring `"interrupted"` fallback (false-positives on ordinary text) and their `||` defaults (which clobber legitimate zero values).

Happy to help wire this onto the new codec once realtime-v1 lands.

DouweM:
Reconciliation status against the current `realtime-v1` (this branch is ~106 commits behind it now).

I tried merging the latest `realtime-v1` in: the merge itself is nearly clean (only the provider-list error message, the inference test, and the now-restructured docs conflict — all trivial to combine), but `bedrock.py` no longer typechecks against the evolved codec (12 pyright errors). I deliberately did **not** push a red merge or rewrite the spike's `connect`/`__aiter__` — those need your design intent — so the branch is left clean at its current head. Here's the concrete adaptation roadmap for when this leaves draft:

**Codec API drift to adapt to (the merge's type errors):**
- **Two-layer event split.** `RealtimeConnection.__aiter__` now returns `AsyncIterator[RealtimeCodecEvent]` (the connection→session vocabulary: `AudioDelta`, `Transcript`, `InputTranscript`, `ToolCall`, `TurnCompleteEvent`, `SessionUsageEvent`, `SessionErrorEvent`, `SpeechStart/StopEvent`, `ReconnectedEvent`), **not** `RealtimeEvent` (the session→user vocabulary). `bedrock.py`'s `__aiter__` is annotated `AsyncIterator[RealtimeEvent]`; it should yield codec events and be re-typed. The session now owns the translation to `PartStart/Delta/End`, tool-execution, and history.
- **`connect` signature.** `RealtimeModel.connect` is now `connect(*, messages, model_settings, model_request_parameters)` — instructions come from `get_instructions(messages)` and tools from `model_request_parameters.function_tools` (see `openai.py`), rather than the `instructions=`/`tools=`/`native_tools=` params the spike takes.
- **New codec events available** (worth mapping from Nova's vocabulary): `InputTranscriptionFailedEvent` (recoverable), `InputSpeechEndEvent.item_id`, `ReconnectedEvent.state_restored`, and `SessionUsageEvent.response_scoped` for run-vs-response usage attribution.
- **Profiles now resolve per-model-name** via `Provider.realtime_model_profile(model_name)` folded into `RealtimeModel.profile` with `merge_realtime_profile` — so the Bedrock profile should move to that shape.

**Semantic reconciliations from the earlier comparison review (still open):**
- Reset all content-scoped state (`_role`/`_generation_stage`/current item) at `contentEnd`/turn completion, not just on `completionEnd`.
- Barge-in: keep the structural (`completionEnd.stopReason == INTERRUPTED` + JSON marker) detection; do **not** add a substring fallback (ordinary text containing "interrupted" false-positives).
- Fix the profile over-claim: Nova's automatic endpointing + closing an audio content block is **not** the full manual-turn contract (`commit` + `create_response` + `clear`), so it shouldn't advertise `supports_manual_turn_control` until all session controls work.
- Usage: the current per-`delta` mapping is correct for the new codec (the session increments events); don't switch to cumulative totals.

**Still the gating blocker (unchanged):** a real SigV4 run + a Smithy-stream cassette, and the decision on the experimental `aws-sdk-bedrock-runtime` Python SDK dependency.

PR #2970, opened 2025-09-19
Support image generation and output with Google and OpenAI

DouweM:
- Closes https://github.com/pydantic/pydantic-ai/issues/2968
- Closes https://github.com/pydantic/pydantic-ai/issues/2391
- Closes https://github.com/pydantic/pydantic-ai/issues/2794
- Replaces https://github.com/pydantic/pydantic-ai/pull/1130

## Google

```py
from pydantic_ai import Agent
from pydantic_ai.messages import BinaryImage

agent = Agent('google-gla:gemini-2.5-flash-image-preview', output_type=BinaryImage)

result = agent.run_sync('Generate an image of a programmer axolotl wearing Pydantic swag.')

output = result.output
assert isinstance(output, BinaryImage)
assert output.media_type == 'image/png'
assert isinstance(output.data, bytes)
assert result.response.images == [output]
with open('axolotl-google.png', 'wb') as f:
    f.write(output.data)
```

axolotl-google

## OpenAI

```py
from pydantic_ai import Agent
from pydantic_ai.messages import BinaryImage

agent = Agent('openai-responses:gpt-5', output_type=BinaryImage)

result = agent.run_sync('Generate an image of a programmer axolotl wearing Pydantic swag.')

output = result.output
assert isinstance(output, BinaryImage)
assert output.media_type == 'image/png'
assert isinstance(output.data, bytes)
assert result.response.images == [output]
with open('axolotl-openai.png', 'wb') as f:
    f.write(output.data)
```

axolotl-openai

## To do

- [x] Image as `result.output` via `output_type=Image`
- [x] `output_type=Image | str` or other combinations
- [x] Streaming
- [x] OpenAI image generation tool options
- [x] Store OpenAI `revised_prompt` as built-in tool call result
- [x] Vertex AI
- [x] OTel
- [x] Temporal (doesn't work because of 2MB size limit)
- [ ] Docs

PR #6571, opened 2026-07-17
feat: distinguish no-output retries from validation feedback in `RetryPromptPart`

### LINKED ISSUE #6404: Render `RetryPromptPart` so the model can tell validation feedback from a user message
When a `RetryPromptPart` has no `tool_name`, it goes out to the provider as a plain user message, so the model can't tell our validation feedback from something the human typed.

For `gpt-5` (Chat Completions) and `claude-sonnet-4-5`, `RetryPromptPart(content='Please return text or call a tool.')` maps to:

```json
{"role": "user", "content": "Validation feedback:\nPlease return text or call a tool.\n\nFix the errors and try again."}
```

The only signal that this isn't the user is the wording. Retries *with* a `tool_name` don't have this problem: they map to a `tool` message (OpenAI) or a `tool_result` block with `is_error` (Anthropic), which is structurally unambiguous.

The `tool_name is None` case covers retries from output validators on `TextOutput`, `NativeOutput`, and `PromptedOutput` (`_output.py`, where `tool_name` comes from `run_context.tool_name`), plus the retry issued when a response carries no actionable output (`_agent_graph.py`).

### Why now

Two reasons this is worth revisiting:

1. Removing the backward-looking text recovery (#6403) means the no-output retry fires in cases that previously ended the run, so this rendering is now on a hotter path.
2. `Validation feedback:` / `Fix the errors and try again.` is a poor fit for the no-output retry. Nothing failed validation and there are no errors to fix — the model just didn't produce output. The wrapper is right for `NativeOutput` schema failures and wrong here.

### Proposal

We already solved the adjacent problem in #5509 (closing #5437): non-leading `SystemPromptPart`s are wrapped as `<system>`-tagged `UserPromptPart`s in `Model.prepare_messages`, gated on `ModelProfile.supports_inline_system_prompts`, so providers that accept an inline system message get a real one and the rest get a tagged user message. Tagging framework-authored turns inside a user message is also what Claude Code and other harnesses do.

The same shape should work here:

- Where the provider accepts an inline system/developer message, send the retry as one. OpenAI Responses already routes `SystemPromptPart` through `openai_system_prompt_role`, which can be `developer` — plausibly the right vehicle.
- Otherwise, wrap the content in a tag so the model can attribute it, rather than relying on the `Validation feedback:` prefix.
- Separately, reconsider `RetryPromptPart.model_response()` so the framing matches the retry's cause. `Fix the errors and try again.` is not appropriate when no output was produced.

Worth deciding whether the tag is `<system>` (reusing the existing convention) or something narrower like `<validation_feedback>`, and whether this reuses `supports_inline_system_prompts` or wants its own profile flag.

Any change here alters what every provider sends for existing retries, so it needs cassette updates and a look at whether the new framing actually improves recovery rates.

Juyotal:
- Closes #6404

This is the first of the two PRs discussed on the issue: the part-level cause split, independent of the renderer/provenance work.

### What

`RetryPromptPart.model_response()` rendered every `tool_name is None` retry with a `Validation feedback:` prefix and a `Fix the errors and try again.` suffix. That framing is wrong for the no-actionable-output retry (thinking-only responses), where nothing failed validation and there is nothing to fix.

This adds a `cause` discriminator to `RetryPromptPart` so the two `tool_name is None` causes are distinguishable at the part level, and branches `model_response()` accordingly:

- `cause: Literal['error', 'no_output'] = 'error'`. The default `'error'` reproduces today's behavior for every existing and deserialized part; `'no_output'` is set only at the no-actionable-output site in `_agent_graph.py`.
- For `'no_output'`, `model_response()` returns the content verbatim (e.g. `Please return text or call a tool.`), without the validation prefix/suffix.

Naming: `cause` rather than `kind` to avoid confusion with the existing `part_kind`. Happy to rename if you'd prefer `reason` or something else.

### Scope

Deliberately excludes the renderer (per-profile developer/system role + tag fallback at `Model.prepare_messages`) and the model-neutral `source`/provenance field, which are the follow-up PR still pending the tag-name and profile-flag decisions.

### Backward compatibility

Per the version policy, adding an optional field with a default to an existing message part is not a breaking change. A test round-trips a `cause`-less history through `ModelMessagesTypeAdapter` and asserts it deserializes to `'error'` and renders identically.

### Notes

The snapshot updates show `input_tokens` dropping by ~8 on the follow-up request, since the no-output retry no longer carries the validation boilerplate.

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6571?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


DouweM:
@Juyotal Thanks for the PR, I plan to get to this in the next few days!

dsfaccini:

Heads-up for whoever's driving this workstream: we closed https://github.com/pydantic/pydantic-ai/pull/4869 (Vercel `errorText` leaking the "Fix the errors and try again." retry suffix) in favor of the provenance direction here. That PR restructured the same `RetryPromptPart.model_response()` this one touches and stashed the full suffixed text in Vercel `call_provider_metadata`, which cuts against the provenance-as-data direction in #6404. Flagging so the Vercel `errorText` rendering is accounted for as the renderer/provenance work lands. Full reasoning: https://github.com/pydantic/pydantic-ai/pull/4869#issuecomment-5146760213

dsfaccini:
gonna start pushing stuff into here, leaving a note so other claude's don't push at the same time

dsfaccini:

🔍 Local review suite (automated, pre-merge) — 14 reviewers across 3 waves, each wave adjudicated by an independent claim-check.

|---|---|---|---|---|
| 1 · code-correctness | ✅ closed | 0 | 0 | 11 |
| 2 · tests/cassettes/docs | ✅ closed | 0 | 0 | 3 |
| 3 · cross-cutting/siblings | ✅ closed | 0 | 0 | 4 |

**Verdict:** ✅ PASS — was 🛑 1 blocking + 🔧 5 required; **all 6 resolved in `de7184f`** · 6 applied. The suite flagged these findings and then fixed them, so this is a PASS *reached by* the review rather than one it started from. The underlying design was sound throughout — `cause` sits in the right layer, all 11 provider mappers and both UI adapters inherit the new framing through the single `model_response()` renderer, and no sibling construction site was left unstamped.

Reviewed at 2026-08-05T06:20Z · diff `de7184f8`

<details>
<summary>✅ 6 required action(s) resolved</summary>

_The 1 blocking and 5 required items the review flagged, all resolved in `de7184f`._

**1. Re-recorded the two drifted cassettes** · `test_anthropic/test_anthropic_model_retrying_after_empty_response.yaml`, `test_bedrock/test_bedrock_thinking_high_qwen_variant.yaml`
Both stored `Validation feedback:\nPlease return text.\n\nFix the errors and try again.` for a retry now rendered as bare `Please return text.` A full tape-vs-live body diff showed the framing string was the **only** delta in each — a clean single-variable re-record. They passed CI regardless, because `tests/conftest.py::vcr_config` sets no `match_on`, so vcrpy never inspects the request body. Re-recorded live with `--record-mode=rewrite` (not hand-edited), then verified replaying with no credentials present. Anthropic's `input_tokens` drops 26 → 13, which is the boilerplate coming off the wire.

**2. `cause` contract edit** · `messages.py` · `RetryPromptPart.cause`
One docstring edit absorbing six independent raises across all three waves — the field promised more than the code delivers, and mis-named what it marks:
- **definition** — "no actionable output at all" / "e.g. only thinking content" was wrong for a path the framework hits routinely: with a tool or structured-output schema set, plain text also stamps `cause='no_output'` (reproduced via `test_plain_response_then_tuple`, where the model returned `TextPart('hello')`). The class-docstring bullet describing that path was left reading as an `'error'` cause and is now labelled `'no_output'`.
- **list content** — the branch is nested inside `isinstance(self.content, str)`, so error details still render as validation errors.
- **`tool_name` set** — the check runs before the `tool_name` branch, so a tool-scoped `no_output` retry drops the `Fix the errors and try again.` suffix…
- **…while the wire still says error** — every sink flags it regardless (Anthropic `is_error=True`, Google `{'error': …}`, Vercel `ToolOutputErrorChunk`, AG-UI `ToolMessage(error=…)`, Vercel reload → `ToolReturnPart(outcome='failed')`). Reproduced end-to-end on two providers via a `DeferredToolResults` entry, which is how a user-built part legitimately acquires a `tool_name`. Worded after the sibling precedent — `BaseToolReturnPart.outcome` states exactly this invariant ("Only `'failed'` is mapped to a provider's native error channel"). **No sink change; documentation only.**

**3. Pinned the tool-scoped `no_output` rendering** · `tests/test_messages.py` · `test_retry_prompt_no_output_renders_verbatim_for_tool_scoped_retry`
No test constructed `cause='no_output'` together with a `tool_name`. Asserts both arms so the contrast is visible.

**4. Round-tripped the non-default `cause`** · `tests/test_messages.py` · `test_retry_prompt_cause_survives_serialization_round_trip`
The back-compat test only ever dumped a part whose `cause` was the **default**, so it structurally could not catch the canonical regression in this class — a non-default discriminator collapsing to the default on reload (the #5870 shape).

**5. Dropped two redundant asserts** · `test_retry_prompt_no_output_renders_content_verbatim`
`assert 'Validation feedback:' not in response` / `assert 'Fix the errors and try again.' not in response` sat beside a `snapshot()` already pinning the exact full string — `tests/AGENTS.md`: "Don't add field-by-field asserts alongside a snapshot."

**6. Rewrote two test docstrings** · `tests/test_messages.py`
They argued to a reviewer of this PR and referenced the pre-change state ("could pass green on the old wording"), against the repo rule on referencing past code state.
</details>

<details>
<summary>⚠️ 4 informational (defer/skip) — all tracked</summary>

**1. Five cassettes with pre-existing retry-content drift** — their recorded retry *content* diverged from the code before this branch (the `alternatives` block changed upstream in #2970 without the tapes being re-cut), so re-recording is not single-variable; one has additionally drifted at SDK level. Tracked at **#7178**.

**2. `docs/retries.md` in open, approved #7145** — says a `RetryPromptPart` "renders for the model with `'Fix the errors and try again.'` appended", false for `cause='no_output'`. Neither PR can fix it alone: the file doesn't exist at this branch's base, and #7145 can't reference a field that doesn't exist yet. Tracked at **[#7145 (comment)](https://github.com/pydantic/pydantic-ai/pull/7145#issuecomment-5192696637)** — whichever merges second carries the qualifier.

**3. PR 2's renderer must also cover the UI adapter dump paths** — `ui/vercel_ai/_adapter.py` and `ui/ag_ui/_adapter.py` call `model_response()` directly, bypassing `Model.prepare_messages` where the parked provenance design puts its step. Correctly out of scope here. Tracked at **[#6404 (comment)](https://github.com/pydantic/pydantic-ai/issues/6404#issuecomment-5192700366)**.

**4. `DeferredToolResults.calls` collapses serialized parts to plain dicts** — the `| Any` arm wins the smart union, so a resumed run gets a `dict` and the `isinstance(…, RetryPromptPart)` branch never fires. Pre-dates this PR entirely. Tracked at **#7179**.

Skipped with stated reasons (not listed): inline `Literal` vs named alias, no `assert_never` (matches the `ToolReturnPart.outcome` precedent), `cause` vs `outcome`/`interrupted` vocabulary, no `ModelRetry` passthrough, VCR rationale on 1 of 3 tests, durable-exec rollback direction, instrumentation surface.
</details>

<details>
<summary>🔎 What the claim-check refuted</summary>

_Recorded so these don't get re-raised. Four reviewer findings were dropped after adjudication:_

1. **"Get DouweM's sign-off before merging PR 1 without PR 2"** — already his instruction: "That split … is a well-scoped first PR **that doesn't need to wait on the mechanism decision above**."
2. **"Both UI adapters' round-trip docstrings are now stale"** — they already document the loss of the whole part type, which subsumes any field on it. Still true post-change.
3. **"`test_openai_model_cerebras_provider_qwen_3_coder` is a validation retry, so not stale"** — it records the same no-output site, from before the message was rewritten. It *is* stale, just not by this branch.
4. **"`test_openai_responses_image_and_text_output` needs re-recording here"** — its tape text isn't what this branch emits either, so it isn't a single-variable re-record. Deferred to #7178 instead.
</details>

dsfaccini:

Heads-up only, no action implied — just flagging a docs surface that didn't exist when this PR was opened.

[#7145](https://github.com/pydantic/pydantic-ai/pull/7145) adds `docs/retries.md`, which says a `RetryPromptPart`:

> renders for the model with `'Fix the errors and try again.'` appended

The new `cause='no_output'` path returns `self.content` before that suffix is added, so whichever of the two lands second will need that sentence adjusted.

DouweM:
Does not address the issue correctly.

PR #6572, opened 2026-07-17
Fix spurious OTel `Failed to detach context` error when a streamed run is interrupted mid-segment

### LINKED ISSUE #6569: OTel context-detach `ValueError` when a streamed run is interrupted mid-segment (continuation teardown)
This issue was posted by Claude Code using claude-opus-4-8 on behalf of David.

## Summary

When a streamed run is interrupted mid-segment by an exception (e.g. `UsageLimitExceeded` from `UsageLimits(output_tokens_limit=N)`), tearing down the parked `_ContinuationStreamedResponse` generator raises an OpenTelemetry context-detach `ValueError: <Token ...> was created in a different Context`.

The original exception (`UsageLimitExceeded`) still surfaces and is caught correctly — this is a spurious *secondary* traceback during generator finalization. In CPython it goes through OTel's "Failed to detach context" log path (non-fatal, easy to miss). Under Pyodide (e.g. play.pydantic.work) the same traceback surfaces in the output panel, so a streamed-run demo looks like it crashed / like the exception wasn't caught.

## Reproduction (keyless, deterministic — `FunctionModel`, no provider key)

Playground: https://play.pydantic.work/p/vyvAb1mx

A `FunctionModel` streaming many text deltas under `UsageLimits(output_tokens_limit=10)` trips the limit mid-segment while the inner `request_stream` generator is parked at its `yield`; the consumer's raise + the graph's `finally: await sr.aclose()` then tears the parked generator down.

Observed traceback (both CPython and Pyodide):

```
Failed to detach context
  File ".../pydantic_ai/_instrumentation.py", line 353, in attach_captured_context
    yield
  File ".../pydantic_ai/models/_continuation.py", line 424, in _get_event_iterator
    yield self._reindex(event, segment_offset)
GeneratorExit

During handling of the above exception, another exception occurred:

  File ".../opentelemetry/context/__init__.py", line 135, in detach
    _RUNTIME_CONTEXT.detach(token)
  File ".../opentelemetry/context/contextvars_context.py", line 42, in detach
    self._current_context.reset(token)
ValueError: <Token ...> was created in a different Context
```

## Root cause

`_ContinuationStreamedResponse._get_event_iterator` opens each segment's `request_stream` inside `with self.segment_context():` (`pydantic_ai/models/_continuation.py`), which enters `attach_captured_context` in `pydantic_ai/_instrumentation.py`:

```python
@contextmanager
def attach_captured_context() -> Generator[None]:
    token = otel_context.attach(captured)
    try:
        yield
    finally:
        otel_context.detach(token)
```

When the run is torn down mid-segment, the generator finalizes via `GeneratorExit` in a *different* async `Context` than the one the token was attached in, so OpenTelemetry's `detach` fails resetting the contextvar token in the `finally`.

## Impact

- Any streamed run interrupted mid-segment (usage limits, cancellation, downstream errors), any provider — the continuation wrapper is not provider-specific.
- CPython: logged, non-fatal. Pyodide/browser: the traceback surfaces in output and breaks streamed-run demos/consumers that capture stdout/stderr.
- Independent of #6546 (that fix is correct; this is the continuation-teardown path, not the UI-adapter error path).

dsfaccini:
This pull request was posted by Claude Code using claude-opus-4-8 on behalf of David. David has not reviewed this change; it was validated by an automated `/review-branch` pass.

- Closes #6569

## Problem

When a streamed run is interrupted mid-segment (e.g. a usage limit trips), an `ERROR opentelemetry.context Failed to detach context` is logged. In Pyodide/the playground this surfaces verbatim in the output panel as a scary secondary traceback.

Root cause: `capture_current_context().attach_captured_context` (in `_instrumentation.py`) is held across the `yield` in `_ContinuationStreamedResponse._get_event_iterator` (`with self.segment_context():`). On interruption the async generator is finalized (`GeneratorExit`) in a *different* contextvars `Context` than the one that entered the context manager, so `otel_context.detach(token)` → `ContextVar.reset(token)` raises `ValueError: <Token> was created in a different Context`. OTel swallows the error but logs it at ERROR.

## Fix

Restore the previous OTel context by re-`attach`ing it instead of `detach`ing the token:

```python
previous = otel_context.get_current()
otel_context.attach(captured)
try:
    yield
finally:
    otel_context.attach(previous)
```

`attach()` is a plain `ContextVar.set`, which never fails cross-context, so it restores `previous` silently. In the well-nested (non-interrupted) path it restores the exact same value `detach(token)` would.

## "Swapping `detach` → `attach` in a `finally` — can it mask the in-flight exception?"

No. `attach()` cannot raise (it's a `ContextVar.set`, not a `reset`), so the `finally` can't throw and can't discard the propagating exception. The regression test asserts the original `UsageLimitExceeded` still surfaces (`pytest.raises(UsageLimitExceeded)`) *and* that no `Failed to detach context` ERROR is logged.

## Alternative considered

The zero-deviation alternative is structural: don't straddle the `yield` at all — restructure `_get_event_iterator`'s per-segment `with self.segment_context()` so each attach/detach stays within one contextvars `Context`. That's a materially larger change than this one-symptom bug warrants; happy to take that route if a reviewer prefers no OTel-idiom deviation. (Wrapping `detach` in `try/except ValueError` was rejected — it keeps the fragile `detach` and swallows a real exception class.)

## Test

`tests/models/test_continuation_stream.py::test_streamed_interruption_no_context_detach_error` — a public-API regression (the teardown only fires through the full agent stack): it interrupts a streamed run via an output-token limit and asserts no `Failed to detach context` ERROR on the `opentelemetry.context` logger. Verified red-before-green — it fails against the pre-fix `detach(token)` body.

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6572?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/_instrumentation.py
+++ b/pydantic_ai_slim/pydantic_ai/_instrumentation.py
@@ -354,11 +354,19 @@ def capture_current_context() -> Callable[[], AbstractContextManager[None]]:
 
     @contextmanager
     def attach_captured_context() -> Generator[None]:
-        token = otel_context.attach(captured)
+        # Restore the previous context by re-`attach`ing it rather than `detach`ing the token: this CM is
+        # held across the `yield` in `_ContinuationStreamedResponse._get_event_iterator`, so when a streamed
+        # run is interrupted mid-segment the async generator is finalized (`GeneratorExit`) in a different
+        # contextvars `Context`, where `otel_context.detach(token)` -> `ContextVar.reset` raises
+        # `ValueError: ... created in a different Context`. OTel swallows it but logs a noisy
+        # 'Failed to detach context' (surfaced verbatim in the Pyodide output panel). `attach()` is a plain
+        # `set`, which never fails cross-context, so it restores `previous` silently. See #6569.
+        previous = otel_context.get_current()
+        otel_context.attach(captured)
         try:
             yield
         finally:
-            otel_context.detach(token)
+            otel_context.attach(previous)
 
     return attach_captured_context
 
--- a/tests/models/test_continuation_stream.py
+++ b/tests/models/test_continuation_stream.py
@@ -1,8 +1,10 @@
 """Unit tests for the streamed-continuation composite `_ContinuationStreamedResponse`.
 
-These drive the composite directly with a scripted fake model (no HTTP, no cassettes) to
+Most drive the composite directly with a scripted fake model (no HTTP, no cassettes) to
 pin down the provider-agnostic stitching behavior: part-index reindexing across segments,
 merged snapshots/usage, live state transitions, cancellation, and the continuation limit.
+A few exercise the composite through the full agent stack where the behavior under test
+only fires end-to-end (e.g. OTel context teardown on mid-segment interruption).
 
 These are unit tests rather than VCR tests for two reasons: `FunctionModel` can't emit a
 *suspended streaming* segment (the input a real continuation needs), and a cassette wouldn't
@@ -14,6 +16,7 @@ existing recording and pass green. Asserting the stitched indices directly is wh
 from __future__ import annotations
 
 import asyncio
+import logging
 from collections.abc import AsyncGenerator, AsyncIterator, Awaitable, Callable
 from contextlib import asynccontextmanager
 from dataclasses import dataclass, field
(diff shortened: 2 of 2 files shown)


dsfaccini:
verified fix :) https://play.pydantic.work/p/w9F23cyD/v/1/s/s1?version=2.12.0

PR #6585, opened 2026-07-19
docs(models): correct `FallbackModel` finish-reason note to match code

### LINKED ISSUE #6565: [Docs/Bug]: Docs claim length/content_filter raise and are caught by default FallbackModel
### Describe the bug

`docs/models/overview.md` states that responses with a `'length'` or `'content_filter'` finish reason raise exceptions which **`FallbackModel` catches by default**.

Local checks with `FunctionModel` (no API keys) show this is not accurate:

1. **Raise only when the response is empty / non-actionable**  
   - Empty + `content_filter` → `ContentFilterError`  
   - Non-empty text + `content_filter` → **no raise**, text is returned  
   - Same pattern for `length` / `UnexpectedModelBehavior`

2. **`ContentFilterError` is not a `ModelAPIError`**  
   - MRO: `ContentFilterError` → `UnexpectedModelBehavior` → `AgentRunError`  
   - Default `fallback_on=(ModelAPIError,)` would not match even if raised inside `model.request`.

3. **Errors are raised in the agent graph after `model.request` succeeds**  
   So `FallbackModel` never sees them — even when `fallback_on` explicitly includes `ContentFilterError` / `UnexpectedModelBehavior`.

This is related to product discussion in #5221 (whether non-empty `content_filter` should hard-fail), but this report is specifically about **docs claiming a reliability behavior that the current architecture does not implement**.

### Debug information

- Checkout @ `8481c72789374a5af071be2864b1c8b566a0ff5b`
- Files: `docs/models/overview.md`, `pydantic_ai/models/fallback.py`, exceptions + agent graph finish-reason handling

### Repro steps

Use a `FunctionModel` that returns non-empty text with `finish_reason="content_filter"` / `"length"`, and an empty response with `content_filter`. Wrap with default `FallbackModel` and observe:

- non-empty content_filter/length: no raise  
- empty content_filter: `ContentFilterError` escapes FallbackModel (`primary_calls` stays 1)

### Expected behavior (per docs)

Finish reasons raise and default FallbackModel switches models.

### Actual behavior

Partial responses succeed; empty filter/length errors are not caught by default FallbackModel.

### Suggested direction

Rewrite the overview note to match code: when raises happen, that FallbackModel does not catch graph-level `UnexpectedModelBehavior` / `ContentFilterError` by default, and point users at response handlers / explicit `fallback_on` if that becomes supported.

Happy to open a docs PR.

sergioperezcheco:
Closes #6565

The finish-reason note in docs/models/overview.md claimed that length and content_filter finish reasons raise exceptions which FallbackModel catches by default. That does not match the code:

- The agent loop only raises on empty or thinking-only responses (_agent_graph.py, `if is_empty or is_thinking_only:`). A length finish reason raises UnexpectedModelBehavior, and an empty response with content_filter raises ContentFilterError. Non-empty responses with either finish reason are returned as-is.
- Both exceptions are raised from the agent graph, after model.request() returns successfully, so FallbackModel never sees them — its exception handlers only wrap the model request.
- ContentFilterError inherits from UnexpectedModelBehavior then AgentRunError, not from ModelAPIError, so even a fallback_on that explicitly listed ModelAPIError would not match it.

The rewritten note states the real behavior and points users at the response-handler example just above (for falling back on finish_reason) and at the RaiseContentFilterError capability (for raising on non-empty content_filter responses). All claims were verified against the current code on main; this is a docs-only change.

This contribution was prepared with assistance from an AI agent (Claude / glm-5.2).

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6585?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/models/overview.md
+++ b/docs/models/overview.md
@@ -407,14 +407,16 @@ print(result.output)
     To keep exception-based fallback alongside a response handler, pass them together as a list — see the [mixed example below](#combining-handlers).
 
 !!! note
-    Note that the [agent loop](../agent.md) raises on empty or thinking-only responses: a
-    `'length'` finish reason raises [`UnexpectedModelBehavior`][pydantic_ai.exceptions.UnexpectedModelBehavior]
+    Note that the [agent loop](../agent.md) can raise on empty, thinking-only, or non-empty
+    responses: a `'length'` finish reason raises [`UnexpectedModelBehavior`][pydantic_ai.exceptions.UnexpectedModelBehavior]
     when the model produced no actionable output (typically because it hit the token limit mid-thinking),
-    and an empty response with a `'content_filter'` finish reason raises
-    [`ContentFilterError`][pydantic_ai.exceptions.ContentFilterError]. Non-empty responses with either
-    finish reason are returned as-is. These exceptions are raised from the agent loop rather than from
-    `model.request()`, so they are **not** caught by `FallbackModel`'s default `fallback_on=(ModelAPIError,)`
-    (and `ContentFilterError` does not inherit from [`ModelAPIError`][pydantic_ai.exceptions.ModelAPIError]).
+    an empty response with a `'content_filter'` finish reason raises
+    [`ContentFilterError`][pydantic_ai.exceptions.ContentFilterError], and a non-empty response truncated
+    mid-tool-call raises [`IncompleteToolCall`][pydantic_ai.exceptions.IncompleteToolCall] (also a subclass
+    of `UnexpectedModelBehavior`). Otherwise non-empty responses are returned as-is. These exceptions are
+    raised from the agent loop rather than from `model.request()`, so they are **not** caught by
+    `FallbackModel`'s default `fallback_on=(ModelAPIError,)` (and `ContentFilterError` does not inherit
+    from [`ModelAPIError`][pydantic_ai.exceptions.ModelAPIError]).
     To fall back on them, add a response handler (see the [example above](#finish-reason-example)) that
     inspects [`finish_reason`][pydantic_ai.messages.ModelResponse.finish_reason]. To also raise on
     `content_filter` responses that still carry partial or refusal text, add the
--- a/docs/models/overview.md
+++ b/docs/models/overview.md
@@ -407,11 +407,18 @@ print(result.output)
     To keep exception-based fallback alongside a response handler, pass them together as a list — see the [mixed example below](#combining-handlers).
 
 !!! note
-    Note that Pydantic AI already handles some finish reasons automatically in the [agent loop](../agent.md):
-    responses with a `'length'` or `'content_filter'` finish reason raise exceptions (which `FallbackModel`
-    catches by default), and empty responses are retried. A response handler is useful for custom
-    checks beyond these built-in behaviors. To also raise on `content_filter` responses that still carry
-    partial or refusal text, add the [`RaiseContentFilterError`][pydantic_ai.capabilities.RaiseContentFilterError] capability.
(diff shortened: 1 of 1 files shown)


dsfaccini:

Reopening. This PR was auto-closed by duplicate-detection 12 seconds after you opened it, in favor of #6566. Having read all four PRs on this issue, yours is the only one that states the behavior correctly:

- it scopes the `content_filter` raise to strictly-empty responses (the code guard is `is_empty and finish_reason == 'content_filter'`, so thinking-only does not raise) — the other PRs get this wrong
- it correctly covers empty *and* thinking-only for `'length'`
- it correctly explains these are raised from the agent loop after `model.request()` returns, and that `ContentFilterError` does not inherit from `ModelAPIError` (the two independent reasons `FallbackModel` doesn't catch them)
- it preserves the `RaiseContentFilterError` pointer

#6566 has been closed: it was factually wrong on the `content_filter` + thinking-only case (CodeRabbit flagged it, never addressed), it dropped the `RaiseContentFilterError` sentence, and it was conflicting and stale.

One small change requested: in "the agent loop only raises on empty or thinking-only responses", please drop the word "only" — `IncompleteToolCall` can also be raised for a non-empty response whose last part is a truncated tool call with unparseable args.

Sorry for the automated closure. Thanks for the careful reading here.

sergioperezcheco:
Good catch @dsfaccini — dropped "only" in 174af83. You are right that a non-empty response whose last part is a truncated tool call with unparseable args also raises (`IncompleteToolCall`), so "only" was too strong. Sorry about the auto-close noise on your end too. Let me know if anything else needs adjusting.

dsfaccini:
This comment was posted by Claude Code using claude-opus-4-8 on behalf of David. David reviewed the findings below and directed this change, but has not reviewed the final wording line by line.

@sergioperezcheco thanks for turning this around same-day, and for going further than we asked — your rewrite is a real improvement over what `main` said (which flatly claimed `FallbackModel` catches these). Rather than send you a third round, I pushed the corrected wording straight to your branch (2b231e024). Three things a follow-up pass caught, verified by running the code rather than reading it:

1. **`IncompleteToolCall` doesn't raise on the first truncation** (the important one). A non-empty response truncated mid-tool-call is re-prompted like any other invalid-arguments failure; it only surfaces as `IncompleteToolCall` once the retry budget is exhausted — with the default `retries=1` that's after 2 model calls, with `retries=0` after 1. As written it also contradicted the note further up the page ("Validation errors … use the retry mechanism instead").

2. **The "empty responses are retried" fact got dropped.** Your version covers what raises, but a reader can no longer tell what happens to an empty response that *doesn't* raise — those are re-prompted up to the output retry limit (or accepted as `None` when the output schema allows it).

3. **"To fall back on them" misframes the mechanism.** A response handler runs inside `FallbackModel.request()` on the *successful* response and rejects it **before** the agent loop's guards run — it prevents these exceptions rather than catching them. The old phrasing invites `fallback_on=(ContentFilterError,)`, which was verified not to work (the secondary model is never tried). The note now says so explicitly, and flags that a handler rejects *every* response with that finish reason, including non-empty ones the agent loop would have accepted.

Nothing for you to do — just flagging it so the reasoning is on the record. Thanks again for the contribution.

PR #6592, opened 2026-07-19
Preserve `$ref` sibling keywords when inlining `$defs`

### LINKED ISSUE #6591: InlineDefsJsonSchemaTransformer drops `$ref` sibling keywords (description/default) when inlining $defs
### Description

`InlineDefsJsonSchemaTransformer` (and any transformer with `prefer_inlined_defs=True`, used for Bedrock, OpenRouter, etc.) inlines a `$ref` by replacing the whole schema with the referenced definition, discarding any JSON Schema keywords that sit **alongside** the `$ref`.

Pydantic emits `{"$ref": ..., "description": ...}` for nested-model fields (and adds `default` for fields with defaults), so a **field-level description or default is silently dropped** from the schema sent to the model.

### Reproducible example

```python
from pydantic_ai._json_schema import InlineDefsJsonSchemaTransformer

schema = {
    'type': 'object',
    'properties': {
        'field': {'$ref': '#/$defs/Foo', 'description': 'field-level description', 'default': None},
    },
    '$defs': {'Foo': {'type': 'object', 'properties': {'x': {'type': 'integer'}}}},
}

field = InlineDefsJsonSchemaTransformer(schema).walk()['properties']['field']
print(field)
# actual:   {'type': 'object', 'properties': {'x': {'type': 'integer'}}}
# expected: {..., 'description': 'field-level description', 'default': None}
```

### Expected behavior

Keywords next to the `$ref` (notably `description` and `default`) should be preserved on the inlined schema, with the field-level value taking precedence over the referenced model's. Dropping the description means the model loses the field's documentation; dropping `default` loses optionality/default information.

### Root cause

In `pydantic_ai_slim/pydantic_ai/_json_schema.py`, the `$ref`-following loop does `schema = def_schema`, replacing the schema (and its siblings) entirely.

Happy to submit the fix (merge the non-`$ref` siblings over the inlined definition) — draft PR at #6590.

lntutor:
Fixes #6591

### What's broken

`InlineDefsJsonSchemaTransformer` (Bedrock, OpenRouter, any `prefer_inlined_defs=True`) inlines `$ref`s by replacing the schema outright (`schema = def_schema`), dropping keywords that sat next to the `$ref`. Pydantic emits `{"$ref": ..., "description": ...}` for nested-model fields, so a field-level `description`/`default` is silently dropped from the schema sent to the model. Full reproducible example in #6591.

### Fix

Merge the non-`$ref` sibling keywords over the inlined definition, siblings taking precedence:

```python
siblings = {k: v for k, v in schema.items() if k != '$ref'}
schema = {**def_schema, **siblings} if siblings else def_schema
```

### Tests

Adds `test_inline_defs_preserves_ref_sibling_keywords` to `tests/test_json_schema.py`; the full file passes (12). The no-sibling path is unchanged.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6592?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lntutor:
Good catch — addressed in the latest push. `{**def_schema, **siblings}` was letting a sibling override the definition's *validation* keywords too. I've restricted the override to annotation keywords (`description`, `title`, `default`, `examples`, `deprecated`, `readOnly`, `writeOnly`, `$comment`); any validation keyword present in the referenced definition now keeps the definition's value (sibling-only keys are still added). Added `test_inline_defs_does_not_let_ref_siblings_weaken_validation` covering the `maxLength` case.

lntutor:
@coderabbitai review

dsfaccini:

Thanks @lntutor — the core fix is correct and resolves the real bug from #6591: `JsonSchemaTransformer._handle` dropped keywords sitting next to a `$ref` when inlining `$defs`, so a field-level `description`/`default` on a nested-model field was silently lost on every `prefer_inlined_defs` path.

I pushed one forward commit (`fd0aa33`) that reverts the later `_OVERRIDABLE_REF_SIBLING_KEYS` whitelist back to your original simple merge:

```python
siblings = {k: v for k, v in schema.items() if k != '$ref'}
schema = {**def_schema, **siblings} if siblings else def_schema
```

Why the revert: the whitelist kept the definition's validation keywords on conflict to avoid "weakening" them, but that reasoning only holds when the definition is the stricter side — it weakens the constraint just as often when the sibling is stricter (e.g. sibling `maxLength: 3` vs def `maxLength: 5`). A single flattened schema can't represent the conjunction either way; the fully-correct answer would wrap conflicts in `allOf`, which is out of scope for this bug. And in practice Pydantic never emits validation keywords as `$ref` siblings — only annotations like `description`/`default` — so the simple merge handles every real case and the extra machinery wasn't earning its keep. I also dropped the test that pinned the whitelist's def-wins behavior and kept your strengthened annotation-sibling test.

The reverted diff matches your PR description again, so nothing else needs updating.

David directed this revert and reviewed the approach; the code change itself was written by Claude Code.

tests/test_json_schema.py
```diff
@@ -253,3 +253,34 @@ def test_typeless_anyof_member_still_recursed():
 
     # Single-member union collapses into the member, which is still transformed.
     assert result == {'type': 'integer'}
+
+
+def test_inline_defs_preserves_ref_sibling_keywords():
+    """Test internal schema walking, which has no provider request to cover with VCR."""
+    from pydantic_ai._json_schema import InlineDefsJsonSchemaTransformer
```
dsfaccini: move up this import please [claude]

dsfaccini: This reply was posted by Claude Code using claude-opus-4-8 on behalf of David.

Done — hoisted `InlineDefsJsonSchemaTransformer` to the module-level import block (`b9bad90`). Left the other in-function transformer imports in this file as-is (pre-existing convention, and your comment was anchored here). `make lint` + `tests/test_json_schema.py` green.

PR #6655, opened 2026-07-22
Route Gemini Live realtime through the Pydantic AI Gateway

adtyavrdhn:
Stacked on #6324 (base: `realtime-v1` — retarget to `main` after it merges). Follow-up to the scoping deferred there: the router comment previously read "only `gateway/openai` is proxied for realtime"; this PR adds the Gemini Live path.

> [!IMPORTANT]
> **Draft until the companion platform relay change lands** — this code is inert on its own. The gateway's realtime relay currently exact-matches `/proxy/{gemini,google-vertex}/v1/realtime` and requires `?model=`; the `google-genai` SDK dials `/proxy/google-vertex/ws/google.cloud.aiplatform.v1beta1.LlmBidiService/BidiGenerateContent` with no query param. The platform side (scoped, cross-ref pydantic/platform#25528 discussion context) will accept the SDK-shaped path, derive the billed model from the first `setup` frame, and rewrite `setup.model` to the full Vertex resource path from the provider row's service account + region.

## What this does

- `infer_realtime_model` accepts `gateway/google:...` and `gateway/google-cloud:...`, dispatching to `GoogleRealtimeModel` with the provider string passed through (resolved via `gateway_provider`, base URL `…/proxy/google-vertex`).
- `GoogleRealtimeModel` gains a `_gateway` flag from the existing `is_gateway_provider` registry (mirroring `OpenAIRealtimeModel`).
- **No URL rewriting**: the SDK's own derivation from the provider base URL produces exactly the URL the relay will accept.
- The one real gap was auth: the gateway authenticates on `Authorization: Bearer <key>`, normally added by its httpx request hook — which can't reach the SDK's `websockets`-library handshake, where the key travels only as `x-goog-api-key` (which the relay ignores). New `_ws_gateway_auth` contextmanager mirrors the neighbouring `_single_ws_user_agent`/`_ws_trace_context` helpers: inject the bearer header into the shared HTTP-options headers inside the `_ws_connect_lock` window, restore after, defensive no-ops for missing key / pre-existing `Authorization` / fake clients.

## Tests

- Positive dispatch tests for both routes (mirroring the `gateway/openai` ones); the obsolete `gateway/google`-is-rejected guard is replaced with a `gateway/groq` case.
- Handshake test drives the public `connect()` with `google.genai.live.ws_connect` patched (the cassette engine's seam), asserting the exact dialed URI, the bearer header, single user-agent, and post-connect header restoration; plus the negative (direct providers carry no `Authorization`) and unit coverage of every defensive branch.
- `tests/realtime/`: 502 passed; changed files at 100% coverage; lint/typecheck clean.

## Verification plan (blocked on the platform PR)

The gateway's Gemini Live relay has no e2e coverage and the built-in Google provider's region (`global`) doesn't serve the Live API — so live verification uses a BYOK `google-vertex` provider row with a service account in a Live-capable region (`us-central1`, base_url `https://us-central1-aiplatform.googleapis.com`), running a full session (audio both ways, tool round-trip, usage metering) through the gateway.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6655?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/realtime/gemini.md
+++ b/docs/realtime/gemini.md
@@ -71,6 +71,29 @@ custom Google Cloud credentials, project, or region.
 Gemini uses `google_input_transcription`, enabled by default; there is no separate transcription
 model. See [Transcribing user input](index.md#transcribing-user-input).
 
+### Routing through the gateway
+
+Route a session through the [Pydantic AI Gateway](../gateway.md) by naming the upstream provider. The
+gateway credentials come from [`gateway_provider`][pydantic_ai.providers.gateway.gateway_provider]:
+
+```python {test="skip" lint="skip"}
+from pydantic_ai.realtime.google import GoogleRealtimeModel
+
+model = GoogleRealtimeModel('gemini-live-2.5-flash', provider='gateway/google')
+```
+
+Or infer it from a gateway-prefixed identifier:
+
+```python {test="skip" lint="skip"}
+from pydantic_ai.realtime import infer_realtime_model
+
+model = infer_realtime_model('gateway/google:gemini-live-2.5-flash')
+```
+
+The gateway proxies Gemini Live over **Vertex** (the `google-vertex` route), so the provider row it
+routes to must be in a region that supports the Live API. `gateway/google-cloud` is an alias for the
+same route. See the shared [gateway notes](index.md#gateway) for trace propagation over the handshake.
+
 ### Session resumption
 
 Set `google_enable_session_resumption=True` together with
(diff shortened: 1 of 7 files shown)


DouweM:
Folded into https://github.com/pydantic/pydantic-ai/pull/6324

PR #6656, opened 2026-07-22
fix(google): defer genai-prices usage extraction until end of Gemini stream

### LINKED ISSUE #6641: Google streaming runs full genai-prices usage extraction on every chunk
This issue was posted by Claude Code using claude-opus-4-8 on behalf of David.

## Problem

On the Google streaming path, full price/usage extraction runs once per streamed chunk instead of once per stream.

`GeminiStreamedResponse._get_event_iterator` in `pydantic_ai_slim/pydantic_ai/models/google.py` calls `_metadata_as_usage(chunk, ...)` at the top of its `async for chunk in self._response` loop. `_metadata_as_usage` (same module) ends in `usage.RequestUsage.extract(...)`, and `RequestUsage.extract` in `pydantic_ai_slim/pydantic_ai/usage.py` loops over candidate provider identities calling `get_snapshot().find_provider(...)` and `provider_obj.extract_usage(...)` from genai-prices, which walks its extraction paths for each attempt.

## Evidence

Profiling the Google streaming path shows `_metadata_as_usage` -> `RequestUsage.extract` -> genai-prices `extract_usage` / `_extract_path` dominating per-chunk cumulative time, roughly half of the whole per-chunk cost. This is by far the largest single item on that path.

Google's `usage_metadata` is cumulative: each chunk carries a running snapshot rather than a delta, so only the last snapshot's extraction result actually contributes to the final usage. The extraction work on every intermediate chunk is largely redundant.

## Design question (needs a maintainer decision before implementation)

There are several viable shapes and they differ in observable behavior, so this should not be implemented until the approach is settled:

- Extract once per stream, at the end. Cheapest, but `StreamedRunResult`'s in-flight `usage()` would no longer reflect priced usage mid-stream.
- Extract only on the final chunk that carries `usage_metadata`. Similar effect, but needs a definition of "final" that holds when a stream is cancelled or errors mid-way.
- Keep per-chunk extraction but cache the provider/model lookup (for example by model id) so the repeated `find_provider` / path-walking cost collapses while per-chunk numbers stay live.

Additional open question: whether this generalizes. Any provider whose streamed usage is cumulative rather than incremental has the same redundancy, so a fix may belong in the shared streaming/usage layer rather than in `google.py` alone. Note that `_metadata_as_usage` also carries deliberate merge logic for fields that a later cumulative chunk can drop (see #5205), so any change here must preserve that behavior and stay covered by tests.

## Where this was found

Surfaced while profiling #6367 (closed), which was filed against #6396. That PR optimized surrounding per-chunk bookkeeping for roughly 1 microsecond per chunk while leaving this, the actual hot spot, untouched.

Diwak4r:
Closes #6641

## Summary

On the Google streaming path, `_metadata_as_usage()` (which ends in `RequestUsage.extract()` → genai-prices `find_provider` + `extract_usage`) was called on every streamed chunk. Google's `usage_metadata` is cumulative — each chunk carries a running snapshot — so only the last chunk's extraction actually contributes to final usage.

This defers the expensive extraction to once after the stream loop completes, using a lightweight tracker variable that only captures the last chunk carrying `usage_metadata`.

## Changes

- Added `_last_usage_chunk: GenerateContentResponse | None = None` tracker
- Inside the loop: only stash the chunk if `chunk.usage_metadata is not None` (no extraction)
- After the loop: single `_metadata_as_usage()` call with the final chunk
- Fallback to `usage.RequestUsage()` if no usage data arrived

## Checklist

- [ ] No breaking changes in accordance with the [version policy](https://github.com/pydantic/pydantic-ai/blob/main/docs/version-policy.md).
- [ ] PR title is fit for the release changelog.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6656?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/hooks.md
+++ b/docs/hooks.md
@@ -78,32 +78,33 @@ Both sync and async hook functions are accepted. Sync functions are automaticall
 
 ```python {title="deferred_hooks_capability.py"}
 from pydantic_ai import Agent, RunContext, ToolDefinition
-from pydantic_ai.capabilities import Hooks, ValidatedToolArgs
-from pydantic_ai.messages import ToolCallPart
+from pydantic_ai.capabilities import Hooks
 
-approval_hooks = Hooks(
-    id='approval-hooks',
-    description='Use when a workflow needs approval before destructive actions.',
+tracing_hooks = Hooks(
+    id='workflow-tracing',
+    description='Use when a workflow needs step-level tracing and metrics.',
     defer_loading=True,
 )
 
 
-@approval_hooks.on.before_tool_execute
-async def require_approval(
-    ctx: RunContext[None],
+@tracing_hooks.on.before_tool_execute
+async def trace_tool(
+    ctx: RunContext,
     *,
-    call: ToolCallPart,
+    call,
     tool_def: ToolDefinition,
-    args: ValidatedToolArgs,
-) -> ValidatedToolArgs:
-    # Runs only after the model loads `approval-hooks`.
-    return args
+    args,
+) -> dict:
+    # Runs only after the model loads `workflow-tracing`.
+    return {'traced': True, 'tool': tool_def.name}
 
 
-agent = Agent('openai-responses:gpt-5.4', capabilities=[approval_hooks])
+agent = Agent('openai-responses:gpt-5.4', capabilities=[tracing_hooks])
 ```
 
 You do not need to guard hooks owned by a deferred `Hooks` instance with `ctx.capability_loaded`; Pydantic AI skips those hooks until the model calls the `load_capability` tool for that capability. Once the hook runs, `ctx.capability_loaded` is true for that hook's owning capability. To check a different capability, inspect `ctx.loaded_capability_ids` or `ctx.available_capability_ids`.
+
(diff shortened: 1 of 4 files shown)


dsfaccini:

Thanks for picking this up, and sorry to close it after you put work in. This is about process on our side, not the quality of what you wrote.

We filed #6641 ourselves off a profiling session, and the issue says the approach needs a maintainer decision before implementation. That was not a formality. There are three candidate shapes and they differ in observable behavior:

- extract once at the end of the stream
- extract on the final chunk that carries `usage_metadata`
- keep per-chunk extraction but cache the provider/model lookup

This PR takes the first one. That changes what `StreamedRunResult.usage()` reports mid-stream, which is a user-visible semantics change we have not agreed to yet, and it is exactly the call the issue was waiting on.

We are going to implement this one ourselves with the design settled first, so I am closing this rather than leaving it open against a spec that is still moving.

One thing worth knowing if you look at this area again: the ceiling here is small. The entire per-chunk loop measures around 29 microseconds, so even removing it completely saves roughly 0.3% of stream wall time on a network-bound path. The redundancy is real and worth fixing, but it is not a large win.

Genuinely appreciate the interest in the codebase. Issues without an open design question are a better bet for a smooth contribution, and I am happy to point you at some if useful.

PR #6682, opened 2026-07-23
fix: preserve `TextPart` and `ToolCallPart` metadata across AG-UI dump/load round-trip

### LINKED ISSUE #5937: [roundtrip-sweep] AG-UI adapter: cluster of part-level fields (id, provider_name, provider_details, dynamic_ref) silently dropped on dump→load rou
[Content truncated due to length]
## Impact

The `AGUIAdapter.dump_messages` / `AGUIAdapter.load_messages` boundary
silently drops a wide cluster of fields on `BaseModelResponsePart`s.
The losses are listed in the adapter's own docstring
(`pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py:738-743`) but
have never been filed as a single tracking issue, and the cluster
straddles several existing `[roundtrip-sweep]` issues that only cover
their headline field:

- `#5870` covers `ToolReturnPart.outcome` only.
- `#5599` covers `ThinkingPart.id` only.
- `#5873` covers `tool_kind` only.

This means frontends / persistence layers that round-trip an agent
run's message history through the AG-UI adapter lose:

- `TextPart.id`, `TextPart.provider_name`, `TextPart.provider_details`
- `ToolCallPart.id`, `ToolCallPart.provider_name`, `ToolCallPart.provider_details`
  (and `ToolCallPart.args` changes type — `dict | None` → `str`)
- `NativeToolCallPart.id`, `NativeToolCallPart.provider_details`
  (only `provider_name` survives via the prefixed tool-call ID)
- `NativeToolReturnPart.provider_details`
- `SystemPromptPart.dynamic_ref` (used to re-evaluate cached system
  prompts; see #4865 for why `dynamic_ref` matters)
- `RetryPromptPart` (no-tool / validation-error form) is silently
  reloaded as `UserPromptPart` with the content re-wrapped in
  `"Validation feedback:\n...\n\nFix the errors and try again."` —
  a **type-discriminator drift** that the discriminator guard from
  #5834 does not protect against because the part is silently
  re-tagged instead of raising.

The Vercel AI adapter handles all of these via `provider_metadata` /
`call_provider_metadata` (an opaque dict on each `UIMessagePart` that
`load_messages` decodes back into the corresponding field). The AG-UI
adapter has the **same** round-trip contract; the AG-UI core spec just
does not have a `provider_metadata` field on its message types, so
either (a) the spec needs an extension point, or (b) the adapter
needs to use a pydantic-ai-private `ActivityMessage` (the same trick
it already uses for `FilePart` at line 707-711) to carry the
provider fields.

## Boundary & Code Path

- **Dump:** `pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py:620-723`
  (`AGUIAdapter._dump_response_parts`):
  - `TextPart` (line ~652): only `content` is forwarded; `id`,
    `provider_name`, `provider_details` are dropped.
  - `ToolCallPart` (line ~670): `id`, `provider_name`,
    `provider_details` dropped; `args` is JSON-stringified into
    `ToolCall.function.arguments` so the type drifts from
    `dict[str, Any] | str | None` to `str`.
  - `NativeToolCallPart` (line ~677): `id` and `provider_details`
    dropped; `provider_name` is encoded as a prefix in the
    `tool_call_id` (`BUILTIN_TOOL_CALL_ID_PREFIX | provider_name | orig`,
    line 677) and decoded back on load, so it is the only one of the
    three that survives.
  - `NativeToolReturnPart`: same loss pattern.
- **Load:** `pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py:405-407`
  (the `AssistantMessage` case) only reads `content` and `tool_calls`
  and calls `builder.add(TextPart(content=content))` — `id` and
  provider fields are never reconstructed.
- **SystemPromptPart.dynamic_ref:** loaded in
  `pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py:402-403` (the
  `SystemMessage` / `DeveloperMessage` case) which constructs
  `SystemPromptPart(content=content)` without `dynamic_ref`. The
  `dynamic_ref` field is silently discarded; pydantic-ai uses it to
  decide whether a system prompt is cache-eligible on Anthropic /
  Bedrock (#4865), so losing it defeats the cache.
- **RetryPromptPart → UserPromptPart drift:** the `RetryPromptPart`
  dump branch at line 553-563 emits a `ToolMessage(error=...)`; when
  the part has no `tool_name` (i.e. it is the validation-error form,
  not the tool-error form), the dump falls through to the
  `UserMessage` branch (line 351-400) which wraps the content in
  `"Validation feedback:\n...\n\nFix the errors and try again."` —
  on reload it becomes a plain `UserPromptPart`, silently downgrading
  the message kind.

## Reproduction

A minimal test that exercises the cluster in one pass:

```python
from pydantic_ai import (
    ModelRequest, ModelResponse, TextPart, ToolCallPart,
    UserPromptPart, SystemPromptPart, NativeToolCallPart, NativeToolReturnPart,
    RetryPromptPart,
)
from pydantic_ai.ui.ag_ui import AGUIAdapter


def _sync_timestamps(o, n):
    for oo, nn in zip(o, n):
        if isinstance(nn, ModelResponse) and isinstance(oo, ModelResponse):
            nn.timestamp = oo.timestamp


def assert_round_trip(name, original):
    dumped = AGUIAdapter.dump_messages(original)
    reloaded = AGUIAdapter.load_messages(dumped)
    _sync_timestamps(original, reloaded)
    assert reloaded == original, f"{name}: LOST {[(o, r) for o, r in zip(original, reloaded) if o != r]!r}"


ts = datetime(2026, 1, 1, 12, 0, 0, tzinfo=timezone.utc)

# 1. TextPart.id / provider_name / provider_details
assert_round_trip("TextPart", [
    ModelRequest(parts=[UserPromptPart(content="hi", timestamp=ts)]),
    ModelResponse(parts=[TextPart(
        content="hello",
        id="text-1",
        provider_name="anthropic",
        provider_details={"anthropic_meta": "yes"},
    )], timestamp=ts),
])

# 2. ToolCallPart.id / provider_name / provider_details
assert_round_trip("ToolCallPart providers", [
    ModelRequest(parts=[UserPromptPart(content="call foo", timestamp=ts)]),
    ModelResponse(parts=[ToolCallPart(
        tool_name="foo",
        args={"q": "x"},
        tool_call_id="c1",
        id="tc-1",
        provider_name="anthropic",
        provider_details={"anthropic_meta": "yes"},
    )], timestamp=ts),
])

# 3. ToolCallPart.args type drift dict -> str
assert_round_trip("ToolCallPart.args drift", [
    ModelRequest(parts=[UserPromptPart(content="call foo", timestamp=ts)]),
    ModelResponse(parts=[ToolCallPart(
        tool_name="foo",
        args={"q": "x", "n": 1},
        tool_call_id="c1",
    )], timestamp=ts),
])

# 4. NativeToolCallPart.id / provider_details
assert_round_trip("NativeToolCallPart providers", [
    ModelRequest(parts=[UserPromptPart(content="hi", timestamp=ts)]),
    ModelResponse(parts=[NativeToolCallPart(
        tool_name="web_search",
        args={"q": "x"},
        tool_call_id="c1",
        id="nc-1",
        provider_name="openai",
        provider_details={"openai_meta": "yes"},
    )], timestamp=ts),
])

# 5. NativeToolReturnPart.provider_details
assert_round_trip("NativeToolReturnPart.provider_details", [
    ModelRequest(parts=[UserPromptPart(content="hi", timestamp=ts)]),
    ModelResponse(parts=[NativeToolReturnPart(
        tool_name="web_search",
        content={"result": "ok"},
        tool_call_id="c1",
        provider_name="openai",
        provider_details={"openai_meta": "return"},
    )], timestamp=ts),
])

# 6. SystemPromptPart.dynamic_ref
assert_round_trip("SystemPromptPart.dynamic_ref", [
    ModelRequest(parts=[SystemPromptPart(content="You are helpful", dynamic_ref="ref-1", timestamp=ts)]),
])

# 7. RetryPromptPart (no-tool) -> UserPromptPart type drift
assert_round_trip("RetryPromptPart no-tool", [
    ModelRequest(parts=[UserPromptPart(content="hi", timestamp=ts)]),
    ModelResponse(parts=[TextPart(content="hi")], timestamp=ts),
    ModelRequest(parts=[RetryPromptPart(content="Output validation error: bad", timestamp=ts)]),
])
```

Run with `uv run pytest tests/test_ag_ui.py::test_dump_load_roundtrip_provider_fields -q`.

## Expected vs Actual

**Expected:** Every field on the source part is preserved by
`dump_messages` → `load_messages`. The Vercel AI adapter is the
reference — it preserves all of these via `provider_metadata` /
`call_provider_metadata` on `UIMessagePart` and a `_sync` helper
in `load_messages`.

**Actual:** 7/7 assertions fail. Captured diff (per case):

| Case | Original | Reloaded |
|------|----------|----------|
| 1 | `TextPart(content='hello', id='text-1', provider_name='anthropic', provider_details={'anthropic_meta': 'yes'})` | `TextPart(content='hello')` |
| 2 | `ToolCallPart(tool_name='foo', args={'q': 'x'}, tool_call_id='c1', id='tc-1', provider_name='anthropic', provider_details={'anthropic_meta': 'yes'})` | `ToolCallPart(tool_name='foo', args='{"q":"x"}', tool_call_id='c1')` |
| 3 | `ToolCallPart(args={'q': 'x', 'n': 1}, ...)` | `ToolCallPart(args='{"q":"x","n":1}', ...)` |
| 4 | `NativeToolCallPart(..., id='nc-1', provider_name='openai', provider_details={'openai_meta': 'yes'})` | `NativeToolCallPart(..., args='{"q":"x"}', provider_name='openai')` |
| 5 | `NativeToolReturnPart(..., provider_name='openai', provider_details={'openai_meta': 'return'})` | `NativeToolReturnPart(..., provider_name='openai')` |
| 6 | `SystemPromptPart(content='You are helpful', dynamic_ref='ref-1')` | `SystemPromptPart(content='You are helpful')` |
| 7 | `RetryPromptPart(content='Output validation error: bad', ...)` | `UserPromptPart(content='Validation feedback:\nOutput validation error: bad\n\nFix the errors and try again.')` |

## Evidence

- `pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py:620-723`
  (`_dump_response_parts` body) — every `if isinstance(part, ...)`
  branch constructs an AG-UI message from the part's narrow
  `content` / `tool_calls` fields, never reading `part.id`,
  `part.provider_name`, or `part.provider_details`.
- `pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py:405-407`
  (`load_messages` `AssistantMessage` case) — the symmetric read
  side also ignores those fields.
- `pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py:402-403`
  (`SystemMessage` / `DeveloperMessage` case) — `dynamic_ref` is
  not passed through.
- `pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py:731-748` —
  adapter docstring explicitly lists the losses but does not
  reference a tracking issue for any of them.
- For comparison, `pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_adapter.py`
  uses `provider_metadata` / `call_provider_metadata` on
  `UIMessagePart` to carry `id`, `provider_name`, `provider_details`
  and decodes them back in `load_messages` — the same shape is
  achievable in AG-UI via a `pydantic_ai_*`-prefixed `ActivityMessage`
  (the same trick the adapter already uses for `FilePart` at
  line 707-711).

rkfshakti:
## What this PR does

Extends the AG-UI adapter's existing carrier mechanisms to preserve part-level metadata (`id`, `provider_name`, `provider_details`) across `dump_messages()` → `load_messages()` round-trips.

## Why it's needed

When messages are round-tripped through `AGUIAdapter.dump_messages()`/`load_messages()`, seven categories of part-level fields were silently dropped because AG-UI has no generic per-part metadata carrier. The Vercel AI adapter handles all of these correctly via `provider_metadata`.

## What changed

**ToolCallPart / NativeToolCallPart** (`id`, `provider_name`, `provider_details`): carried inside the existing `pydantic_ai`-namespaced `encrypted_value` blob alongside `tool_kind`/`outcome`. On load, `parse_encrypted_part_metadata()` extracts them back.

**TextPart** (`id`, `provider_name`, `provider_details`): carried as a preceding `ActivityMessage` with `activity_type='pydantic_ai_text_part'`, the same pattern already used for `FilePart` and `UploadedFile`. On load, the metadata is buffered in `pending_text_metadata` and applied to the next `TextPart`.

### How to verify

1. Create a `TextPart` with `id`, `provider_name`, and `provider_details` set
2. Round-trip through `AGUIAdapter.dump_messages()` → `AGUIAdapter.load_messages()`
3. Before the fix: all three fields are `None` on the reloaded part
4. After the fix: all three fields are preserved

### Tested on

| OS | Status |
| :-- | :----: |
| macOS | ✅ |
| Windows | ⚠️ |
| Linux | ⚠️ |

## Risk & Scope

- **Main risk or tradeoff**: The `encrypted_value` carrier is gated on `ENCRYPTED_VALUE_VERSION` (0.1.11), so pre-0.1.11 clients continue to lose metadata (same as before). The `ActivityMessage` carrier is also gated on `use_encrypted_value` for consistency.
- **Not validated / out of scope**: `NativeToolReturnPart.provider_details` — the return part has no `encrypted_value` slot in the AG-UI protocol.
- **Breaking changes / migration notes**: None. Existing behavior for parts without metadata is unchanged.

## Linked Issues

Fixes #5937

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6682?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
+++ b/pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
@@ -88,11 +88,13 @@ try:
         FILE_ACTIVITY_TYPE,
         MULTIMODAL_VERSION,
         REASONING_VERSION,
+        TEXT_PART_ACTIVITY_TYPE,
         UPLOADED_FILE_ACTIVITY_TYPE,
         dump_tool_return_content,
         parse_ag_ui_version,
         parse_builtin_tool_call_id,
         parse_encrypted_outcome,
+        parse_encrypted_part_metadata,
         parse_encrypted_tool_kind,
         rehydrate_tool_return_content,
         thinking_encrypted_metadata,
@@ -364,6 +366,8 @@ class AGUIAdapter(UIAdapter[RunAgentInput, Message, BaseEvent, AgentDepsT, Outpu
         builder = MessagesBuilder()
         tool_calls: dict[str, str] = {}  # Tool call ID to tool name mapping.
         tool_kinds: dict[str, ToolPartKind] = {}  # Tool call ID to `tool_kind` claim mapping.
+        # Pending TextPart metadata from a preceding ActivityMessage (TEXT_PART_ACTIVITY_TYPE).
+        pending_text_metadata: dict[str, Any] = {}
         # `ToolCall`/`ToolMessage.encrypted_value` only exists on the installed model from 0.1.11
         # onward; older versions drop the client's claim, so the field is only read when present.
         use_encrypted_value = parse_ag_ui_version(DEFAULT_AG_UI_VERSION) >= ENCRYPTED_VALUE_VERSION
@@ -425,7 +429,16 @@ class AGUIAdapter(UIAdapter[RunAgentInput, Message, BaseEvent, AgentDepsT, Outpu
 
                 case AssistantMessage(content=content, tool_calls=tool_calls_list):
                     if content:
-                        builder.add(TextPart(content=content))
+                        text_meta = pending_text_metadata
+                        pending_text_metadata = {}
+                        builder.add(
+                            TextPart(
+                                content=content,
+                                id=text_meta.get('id'),
+                                provider_name=text_meta.get('provider_name'),
+                                provider_details=text_meta.get('provider_details'),
+                            )
+                        )
                     if tool_calls_list:
                         for tool_call in tool_calls_list:
                             tool_call_id = tool_call.id
@@ -441,6 +454,10 @@ class AGUIAdapter(UIAdapter[RunAgentInput, Message, BaseEvent, AgentDepsT, Outpu
                             if tool_kind is not None:
                                 tool_kinds[tool_call_id] = tool_kind
 
+                            # Read part-level metadata (id, provider_name, provider_details) from
+                            # the encrypted_value carrier, fixing the silent field loss in #5937.
+                            part_meta = parse_encrypted_part_metadata(tool_call.encrypted_value) if use_encrypted_value else {}
+
                             builtin_id = parse_builtin_tool_call_id(tool_call_id)
                             if builtin_id is not None:
                                 provider_name, original_id = builtin_id
@@ -451,6 +468,8 @@ class AGUIAdapter(UIAdapter[RunAgentInput, Message, BaseEvent, AgentDepsT, Outpu
                                         tool_call_id=original_id,
                                         provider_name=provider_name,
                                         tool_kind=tool_kind,
+                                        id=part_meta.get('id'),
+                                        provider_details=part_meta.get('provider_details'),
                                     )
                                 )
                             else:
@@ -460,6 +479,9 @@ class AGUIAdapter(UIAdapter[RunAgentInput, Message, BaseEvent, AgentDepsT, Outpu
                                         tool_call_id=tool_call_id,
                                         args=tool_call.function.arguments,
                                         tool_kind=tool_kind,
+                                        id=part_meta.get('id'),
+                                        provider_name=part_meta.get('provider_name'),
+                                        provider_details=part_meta.get('provider_details'),
                                     )
                                 )
                 case ToolMessage() as tool_msg:
@@ -583,6 +605,9 @@ class AGUIAdapter(UIAdapter[RunAgentInput, Message, BaseEvent, AgentDepsT, Outpu
                                 ]
                             )
                         )
+                    elif activity_msg.activity_type == TEXT_PART_ACTIVITY_TYPE:
+                        # Store metadata for the next TextPart in the following AssistantMessage.
+                        pending_text_metadata = dict(activity_msg.content)
 
                 case _:
                     if TYPE_CHECKING:
@@ -753,6 +778,23 @@ class AGUIAdapter(UIAdapter[RunAgentInput, Message, BaseEvent, AgentDepsT, Outpu
             if isinstance(part, TextPart):
                 if tool_calls_list:
                     flush()
+                # Carry TextPart metadata (id, provider_name, provider_details) as a preceding
+                # ActivityMessage so it survives a dump/load round-trip.  Same pattern as FilePart.
+                if use_encrypted_value and (part.id is not None or part.provider_name is not None or part.provider_details is not None):
+                    text_meta: dict[str, Any] = {}
+                    if part.id is not None:
+                        text_meta['id'] = part.id
+                    if part.provider_name is not None:
+                        text_meta['provider_name'] = part.provider_name
+                    if part.provider_details is not None:
+                        text_meta['provider_details'] = part.provider_details
+                    result.append(
+                        ActivityMessage(
+                            id=_new_message_id(),
+                            activity_type=TEXT_PART_ACTIVITY_TYPE,
+                            content=text_meta,
+                        )
+                    )
                 text_content.append(part.content)
             elif isinstance(part, ThinkingPart):
                 if use_reasoning:
@@ -772,7 +814,10 @@ class AGUIAdapter(UIAdapter[RunAgentInput, Message, BaseEvent, AgentDepsT, Outpu
                     ToolCall(
                         id=part.tool_call_id,
                         function=FunctionCall(name=part.tool_name, arguments=part.args_as_json_str()),
-                        **tool_kind_encrypted_value_kwargs(part.tool_kind, supported=use_encrypted_value),
+                        **tool_kind_encrypted_value_kwargs(
+                            part.tool_kind, supported=use_encrypted_value,
+                            part_id=part.id, provider_name=part.provider_name, provider_details=part.provider_details,
+                        ),
                     )
                 )
             elif isinstance(part, NativeToolCallPart):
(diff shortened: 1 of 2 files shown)


rkfshakti:
Thanks for the review. Both flagged concerns were addressed in `ba1b2856`:

1. **Metadata loss with multiple TextParts** — the dump path now flushes an `AssistantMessage` before each metadata-bearing `TextPart` so each keeps its own metadata instead of overwriting a single slot.

2. **Stale metadata with empty TextPart content** — `pending_text_metadata` is now cleared unconditionally on each `TextPart` so stale metadata can't attach to a later response.

Would appreciate a human review when time allows.

dsfaccini:

Closing alongside #5937, which we've closed as not planned. The underlying issue came from our automated `roundtrip-sweep` rather than a real-world report, and the fields this PR preserves (`id`/`provider_name`/`provider_details`) aren't sent to the model, so their loss across an AG-UI round-trip isn't a bug we're treating as one.

No reflection on the implementation — appreciate the effort here, @rkfshakti.

PR #6685, opened 2026-07-23
Preserve arbitrary `RequestUsage` fields across serialization

alexmojaki:
Stacks on #6683.

`RequestUsage` and `RunUsage` can accept usage dimensions introduced by newer `genai-prices` releases, but Pydantic's generated dataclass serializer previously dropped those runtime fields. This PR preserves them through the documented serialization paths while keeping the current object and wire formats intact.

- Wrap the generated Pydantic schema so declared fields keep their existing validation, aliases, defaults, filters, and canonical serialization.
- Preserve arbitrary top-level fields through direct type adapters, `ModelMessagesTypeAdapter`, and the documented `pydantic_core` helpers, including falsey and non-numeric values.
- Continue loading pre-usage-refactor message histories, including nullable `request_tokens`, `response_tokens`, and `details`, while re-emitting only canonical current keys.
- Keep the existing message-history serialization examples working with arbitrary usage fields.

Verification:

- 209 message and usage tests pass.
- 18 relevant message-history docs examples pass; 2 are skipped as configured.
- Ruff and focused Pyright checks pass.
- All commit-time pre-commit checks pass.

- Follow-up to #6683

### Checklist


pydantic_ai_slim/pydantic_ai/usage.py
```diff
@@ -70,6 +141,52 @@ def __init__(self, *, details: dict[str, int] | None = None, **kwargs: Any):
...
+                value_dict = None
+                input_value = cast(object, value)
+
+            result = inner(input_value)
+            assert isinstance(result, UsageBase)
+            if value_dict is not None:
+                for key, item in value_dict.items():
+                    if key not in non_arbitrary_keys:
+                        setattr(result, key, item)
+            return result
```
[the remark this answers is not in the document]
alexmojaki: (AI) Addressed in 2374ba69

pydantic_ai_slim/pydantic_ai/usage.py
```diff
@@ -20,6 +21,76 @@
...
+            continue
+        included, item_include = _usage_field_filter(info.include, key, include=True)
+        if not included:
+            continue
+        included, item_exclude = _usage_field_filter(info.exclude, key, include=False)
+        if not included:
+            continue
+        if info.exclude_none and item is None:
+            continue
+        if item_include is not None or item_exclude is not None:
```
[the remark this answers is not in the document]
alexmojaki: (AI) Thanks, no action needed. The mapping returned by the wrap serializer is traversed by Pydantic's outer serializer: I reproduced both `dump_python(..., mode='json')` and `dump_json()` with arbitrary nested `BaseModel` and `datetime` values, and both were converted correctly. The explicit `Any` serializer is only needed when applying field-specific nested include/exclude selectors; leaving unfiltered values for the outer pass also preserves caller-provided fallback handling.

docs/message-history.md
```diff
@@ -335,9 +335,8 @@ result2 = agent.run_sync(  # (3)!
    ```
 2. Alternatively you can serialize to/from JSON directly:
    ```python {test="skip" lint="skip" format="skip"}
-   from pydantic_core import to_json
    ...
-   as_json_objects = to_json(history_step_1)
+   as_json_objects = ModelMessagesTypeAdapter.dump_json(history_step_1)
```
alexmojaki: why is this change needed? this means existing user code out there will break. does to_json not pick up the custom `__get_pydantic_core_schema__` etc?

tests/test_temporal.py
```diff
@@ -3800,12 +3800,33 @@ def test_temporal_run_context_serializes_usage():
...
             details={'foo': 1},
+            future_tokens=7,
+            label='original',
+            zero_tokens=0,
         ),
         run_id='run-123',
     )
 
     serialized = TemporalRunContext.serialize_run_context(ctx)
-    assert serialized['usage'] == ctx.usage
```
alexmojaki, line 3898: why was this assertion removed

docs/message-history.md
alexmojaki: (AI) Tracking as `## preserve-pydantic-core-usage-serialization`: confirmed that bare `pydantic_core.to_json()` and `to_jsonable_python()` do not consult the custom core schema for this nested stdlib dataclass. Declared fields still serialize exactly as before, but arbitrary usage fields are dropped. I will preserve the existing documented path, add direct regression coverage, and revert this docs replacement.

tests/test_temporal.py
alexmojaki: (AI) Tracking as `## retain-temporal-preconverter-assertion`: the old object-equality assertion no longer applies because this method intentionally returns a JSON-ready usage dictionary before the untyped Temporal boundary. I will retain an immediate assertion on that dictionary as well as the post-converter assertion.

docs/message-history.md
alexmojaki: (AI) Addressed in b2abf6ed. Bare `to_json()` and `to_jsonable_python()` now use the same per-concrete-class serializer as `TypeAdapter`; the original docs path is restored and covered directly.

tests/test_temporal.py
alexmojaki: (AI) Addressed in 893944c3. The test now asserts the immediate JSON-ready usage dictionary before the Temporal converter and the same value after the converter round-trip.

alexmojaki: (AI) Follow-up addressed in 85ab8e5b. `serialize_run_context()` again returns the original `RunUsage` object; the class-level serializer lets Temporal preserve flexible fields when the payload converter creates the wire dictionary. The test now pins object identity before encoding and the complete flat representation after encoding.

tests/test_temporal.py
```diff
@@ -3800,14 +3800,38 @@ def test_temporal_run_context_serializes_usage():
...
+            'tool_calls': 1,
+            'future_tokens': 7,
+            'label': 'original',
+            'zero_tokens': 0,
+        }
+    )
+    assert decoded['usage'] == expected_usage
 
-    reconstructed = TemporalRunContext.deserialize_run_context(serialized, deps=None)
+    reconstructed = TemporalRunContext.deserialize_run_context(decoded, deps=None)
```
alexmojaki: why did this line change

alexmojaki: (AI) Tracking as `## preserve-both-temporal-deserialization-paths`: this changed so the reconstruction assertion exercised the actual decoded wire dictionary, but that replaced coverage of the existing in-memory `RunUsage` path. I will restore this line and assert the decoded path separately.

alexmojaki: (AI) Addressed in 62d5252e. The original in-memory deserialization line is restored, and the decoded payload is reconstructed and asserted separately.

tests/test_temporal.py
```diff
@@ -3800,16 +3800,42 @@ def test_temporal_run_context_serializes_usage():
...
+            'tool_calls': 1,
+            'future_tokens': 7,
+            'label': 'original',
+            'zero_tokens': 0,
+        }
+    )
+    assert decoded['usage'] == expected_usage
+
+    reconstructed_from_payload = TemporalRunContext.deserialize_run_context(decoded, deps=None)
+    assert reconstructed_from_payload.usage == ctx.usage
```
alexmojaki: do we actually want these extra lines? it looks like they're adding a new contract that i'm not sure we need, but i might be wrong.

alexmojaki: (AI) Tracking as `## avoid-redundant-temporal-payload-reconstruction`: agreed. The wire snapshot proves the flexible fields reached Temporal, and the existing adapter tests cover dictionary rehydration. I will remove these two redundant reconstruction lines and keep the original in-memory reconstruction assertion.

alexmojaki: (AI) Addressed in 6125928e. The redundant decoded-payload reconstruction assertion is removed; the test retains the original in-memory reconstruction and the flexible-field wire snapshot.

alexmojaki: same question for the remaining lines added at the end. are they adding a contract?

alexmojaki: (AI) Tracking as `## remove-temporal-contract-expansion`: yes, the remaining payload encoding and snapshot explicitly add a Temporal wire-shape contract. That is unnecessary now that the general usage serializer works, so I will restore this entire test to `main` and remove the Temporal claim from the PR description.

alexmojaki: (AI) Addressed in 5439bc11. The entire Temporal test change is removed, so this PR now has zero cumulative Temporal code or test diff from `main`; the PR description has been narrowed accordingly.

alexmojaki: i think that adding fields to the usage had some value

alexmojaki: (AI) Addressed in 3d2db61b. The arbitrary `future_tokens`, `label`, and zero-valued field are restored to the existing usage fixture. Temporal still returns the original `RunUsage` object and the test keeps only its original assertions, so this exercises flexible usage without adding a wire-format contract.

alexmojaki:
@cubic-dev-ai review

Issue #5581, opened 2026-05-22
BedrockModel incorrectly enables native structured output for Claude Opus 4.7


shailendher:
### Initial Checks

- [x] I'm using the [latest version](https://github.com/pydantic/pydantic-ai/releases/latest) of Pydantic AI
- [x] I've searched for my issue in [the issue tracker](https://github.com/pydantic/pydantic-ai/issues) before opening this issue

Relates to #4209, #5579

#### Description

BedrockProvider appears to derive Anthropic Bedrock structured-output support from the direct Anthropic model profile, then [only excludes](https://github.com/pydantic/pydantic-ai/blob/9871f756992fdab68fa0b103022a494a5aaf0ebf/pydantic_ai_slim/pydantic_ai/providers/bedrock.py#L171) claude-opus-4-1. That denylist approach lets newly added Anthropic models inherit supports_json_schema_output=True before Bedrock Runtime/Converse support has been confirmed.

This now affects Claude Opus 4.7. anthropic_model_profile() correctly marks Opus 4.7 as structured-output capable for the direct Anthropic API, but BedrockConverseModel uses Bedrock Runtime Converse with outputConfig.textFormat, not the bedrock-mantle Messages API path.

The supporting docs:

- [Anthropic's structured-outputs docs](https://platform.claude.com/docs/en/build-with-claude/structured-outputs) explicitly enumerate the Bedrock-supported models — Opus 4.6, Sonnet 4.6/4.5, Opus 4.5, and Haiku 4.5 — and note that Opus 4.7 and Mythos Preview are only supported through [Claude in Amazon Bedrock](https://platform.claude.com/docs/en/build-with-claude/claude-in-amazon-bedrock) (the Messages-API endpoint), not the Bedrock Runtime path Pydantic AI uses.
- AWS's [Opus 4.6 model card](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-opus-4-6.html) lists "Structured outputs" as a supported capability; the [Opus 4.7 model card](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-opus-4-7.html) does not.
[...]

PR #6723, opened 2026-07-24
Add Claude Opus 5 (`claude-opus-5`) support

dsfaccini:
This pull request was posted by Claude Code using claude-opus-4-8 on behalf of David.

David steered the scope decisions in this PR (how to model the Opus 5 divergence, and to ground-truth the pre-existing claims rather than trust the docs) but has **not** reviewed the diff line by line.

Adds `claude-opus-5`, released 2026-07-24. No issue — same as the Claude Sonnet 5 add (#6161).

The `/v1/models` endpoint exposes exactly one new id, `claude-opus-5`, with no dated snapshot; Anthropic's model docs confirm that is intentional ("starting with the Claude 4.6 generation, model IDs use a dateless format that is also a pinned snapshot").

## Capabilities — probed live, not mirrored

Every profile flag was verified against the API by probing `claude-opus-5` side by side with `claude-opus-4-8` rather than copying 4.8's profile. It matches 4.8 on: adaptive thinking (and rejects budget thinking), rejects `temperature`/`top_p`, `effort` incl. `xhigh`, task budgets, fast mode, json-schema output, forced `tool_choice`, tool search, advisor (as executor **and** as advisor), code execution `20260120`, and web search/fetch `20260209`.

**It is not a pure 4.8 mirror.** Opus 5 returns a 400 for `xhigh`/`max` effort while thinking is explicitly disabled, where 4.8 accepts the same request:

| `thinking={'type':'disabled'}` + effort | `claude-opus-5` | `claude-opus-4-8` |
|---|---|---|
| `xhigh` | 400 `output_config.effort 'xhigh' is not supported when thinking is disabled on this model` | accepted |
| `max` | 400 | accepted |
| `high` / unset | accepted | accepted |

Modeled as `anthropic_disallows_top_effort_when_thinking_disabled` with a fail-fast `UserError`, mirroring how `_validate_task_budget_vs_context_management` already handles a known-bad combination instead of forwarding an opaque 400.

Two probe results worth calling out, because reading the response alone would mislead:

- **Fast mode** — both `claude-opus-5` and `claude-opus-4-8` return an identical 429 (this org has zero fast-mode quota), which on its own proves nothing. The signal is the contrast with models that genuinely lack it: `claude-opus-4-5`, `claude-sonnet-5`, and `claude-opus-4-1` all return a clean `400 does not support the 'speed' parameter`. A 429 means the request reached quota enforcement, i.e. the parameter is supported.
- **Advisor** — the first probe failed with `'claude-opus-4-8' cannot be used as an advisor when the request model is 'claude-opus-5'`, which is a *pairing* rejection, not an executor-capability one. Re-probed across pairs: `claude-opus-5` is valid as an executor (advisor `claude-opus-5`), and valid as an advisor for `claude-opus-4-8`, `claude-sonnet-5`, and `claude-haiku-4-5` executors.

## Also in this PR

Stale copy that this change would otherwise have contradicted:

- `AnthropicModelSettings.anthropic_speed` docstring didn't list Opus 5, while its profile twin and `docs/models/anthropic.md` did. The same miss was caught by review on the Opus 4.8 add (#5709).
- The task-budget docstring and `_get_task_budget`'s `UserError` enumerated `claude-opus-4-7`/`4-8` only — already stale for Sonnet 5 and Fable 5. Rather than extend a fourth copy of the list, both now point at the `anthropic_supports_task_budgets` profile flag and the Anthropic docs, so there's one source of truth.
- Guarded Bedrock structured output on `claude-opus-5` in `bedrock_structured_output_unsupported`. Without it, `BedrockProvider.model_profile('us.anthropic.claude-opus-5')` flips `supports_json_schema_output` to `True` — reachable today, since `BedrockModelName` is `Literal[...] | str`. This matches #5709, which added `claude-opus-4-8` to that tuple before any Bedrock id existed, and is exactly the defect #5581 filed.

## Deliberate omissions

- **No Bedrock/Heroku model ids.** Only the structured-output guard above. `heroku:` is enforced against Heroku's live catalog by `test_known_model_names`, so a speculative entry would fail CI; Bedrock ids land in their own refresh PR (#6318), not in per-model adds.
- **No SDK floor bump.** `anthropic` 0.109.0 lacks the id, so it goes on the existing `AnthropicModelName` bridge `Literal`, per the established bridge-then-drop pattern (#5849).
- **Not added to the `claude-opus-4-7`/`4-8` parametrize lists.** `test_model_profile_opus_5` asserts every flag those cover, so adding the id would be duplicate coverage.

Two claims flagged during review as provider-doc contradictions were **ground-truthed and left alone**, because the API disagrees with the docs: Sonnet 5 *does* accept task budgets (200, though the docs table says "Not supported"), and Opus 4.7 fast mode is *still* live (429 quota, not the documented removal). Changing either would have broken working behavior.

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6723?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tests/profiles/test_anthropic.py
```diff
@@ -406,3 +408,47 @@ def test_model_profile_sonnet_5():
...
+    advisor tool, code execution `20260120`, and web search/fetch `20260209`; it rejects
+    budget-based thinking and sampling settings. Unlike Sonnet 5 it also supports
+    `anthropic_speed='fast'` (the API returns a fast-mode quota error rather than the
+    `does not support the 'speed' parameter` 400 that unsupported models return).
+
+    The divergence from 4.8 is `anthropic_disallows_top_effort_when_thinking_disabled`: Opus 5
+    returns a 400 for `xhigh`/`max` effort while thinking is disabled, where 4.8 accepts it.
+    """
+    profile = anthropic_model_profile('claude-opus-5')
+    assert profile is not None
```
dsfaccini: why not "assert profile == snapshot"?

dsfaccini: This reply was posted by Claude Code using claude-opus-4-8 on behalf of David.

Done in c08d9382c — `test_model_profile_opus_5` now asserts the full profile via `snapshot({...})`, replacing the per-flag `profile.get(...)` asserts (kept the Opus 4.8 divergence contrast).

dsfaccini:

|---|---|---|---|---|
| 1 · code-correctness | ✅ closed | 0 | 0 | 2 |
| 3 · cross-cutting/siblings | ✅ closed | 0 | 0 | 2 |

**Verdict:** ✅ PASS — 0 blocking, 0 open required · 1 applied. The suite flagged one real gap (Bedrock registration, 3 reviewers); confirmed against AWS docs and resolved in `691692a5f`, which is why this now passes rather than having been clean throughout.
Reviewed at 2026-07-28T04:11:22Z · diff `f66d2ce60338`

<details>
<summary>✅ 1 applied since the review</summary>

**1. `claude-opus-5` missing from the Bedrock name sinks** · `models/_known_model_names.py`, `models/bedrock.py` · MODEL_PROFILE
opus-5 was in `providers/bedrock.py::bedrock_structured_output_unsupported` but absent from every Bedrock name sink, while siblings opus-4-8/sonnet-5 were registered. AWS's model card confirms opus-5 is GA on `bedrock-runtime` (launched 2026-07-24), so this was a real omission. Added the 4 `_known_model_names.py` sinks + 2 MODEL_PROFILE entries (mirroring opus-4-8) in `691692a5f`; `BedrockProvider.model_profile('us.anthropic.claude-opus-5')` now resolves with structured-output correctly gated off.
</details>

<details>
<summary>⚠️ 3 informational (defer/skip)</summary>

**1. `_validate_effort_vs_disabled_thinking` is single-use** · `models/anthropic.py` · `AnthropicModel._validate_effort_vs_disabled_thinking`
Repo rule "don't create single-use utilities" vs. symmetry with the sibling `_validate_*`/`_get_*` validator family — skip (maintainer chose to keep the method: the name + docstring carry the live-verified Opus-5-vs-4.8 rationale at the call site).

**2. Client-side effort/thinking guard** · `models/anthropic.py` · `_validate_effort_vs_disabled_thinking`
Tension with the "don't preemptively reject provider-namespaced settings" rule — skip (net-neutral): resolved by live API verification + profile-flag gating + a direct sibling precedent (`_validate_task_budget_vs_context_management`) + a recorded maintainer decision.

**3. Guard is native-Anthropic-only by construction** · `models/anthropic.py` · `_validate_effort_vs_disabled_thinking`
The flag rides along on Bedrock/Vertex Anthropic profiles but the guard only fires on direct-API requests — skip: consistent with the established per-backend enforcement pattern (sibling `anthropic_disallows_*` guards), and no backend routes opus-5 where it would need to fire pre-request.
</details>

Issue #6559, opened 2026-07-16
Emit OTel `gen_ai.usage.reasoning.output_tokens` and `gen_ai.request.reasoning.level`


pedrurrego:
### Summary
Two reasoning-related GenAI semantic-convention attributes are defined in the spec but not emitted on spans:
  - `gen_ai.usage.reasoning.output_tokens`
  - `gen_ai.request.reasoning.level`

Both are included in the [gen-ai attribute registry](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/registry/attributes/gen-ai.md).

### Current behavior
Reasoning tokens are captured under the custom `details` namespace instead of the spec key, and the key name is inconsistent across providers:

| Provider  | Detail key         | Surfaces as                              |
|-----------|--------------------|------------------------------------------|
| OpenAI    | `reasoning_tokens` | `gen_ai.usage.details.reasoning_tokens`  |
| xAI       | `reasoning_tokens` | `gen_ai.usage.details.reasoning_tokens`  |
| Google    | `thoughts_tokens`  | `gen_ai.usage.details.thoughts_tokens`   |
| Anthropic | — (not captured)   | — (thinking billed within output_tokens) |

Source lines: `models/openai.py` (`output_tokens_details.reasoning_tokens`),
`models/xai.py` (`usage.reasoning_tokens`), `models/google.py`
(`usage_metadata.thoughts_token_count`, added as `thoughts_tokens`). Anthropic's
`_extract_usage_details` only reads input/output/cache token keys.

The requested reasoning effort (`ModelSettings.thinking`) is not emitted as a span attribute for any provider.

### Anthropic now exposes thinking tokens
As of anthropic-sdk-python [v0.105.0](https://github.com/anthropics/anthropic-sdk-python/blob/main/CHANGELOG.md) (2026-05-28), `Usage.output_tokens_details.thinking_tokens` reports reasoning tokens directly (stable and beta types). 
Its docstring *"Always ≤ output_tokens; output_tokens - thinking_tokens approximates the non-reasoning output"* matches OTel note [30] exactly (a subset of `gen_ai.usage.output_tokens`, not additive).
So Anthropic's `_extract_usage_details` can now capture this, unblocking #3249.

### Proposed behavior
- Emit `gen_ai.usage.reasoning.output_tokens`, mapped from `reasoning_tokens`
(OpenAI/xAI), `thoughts_tokens` (Google), and the newly available
[...]

Issue #5205, opened 2026-04-25
Google streaming: `cached_content_token_count` not populated in `Usage.cache_read_tokens`


georg-ort:
## Summary

When using Google Gemini models via streaming (`agent.iter()`), `Usage.cache_read_tokens` is always `0` even when Google's implicit caching is active and the OTel span shows cached tokens.

## Evidence

We confirmed via Logfire that Google IS caching (Gemini 3 Flash Preview, ~76k input tokens):

- **OTel span attribute** `gen_ai.usage.cache_read.input_tokens` = **73,673** ✅
- **OTel span attribute** `gen_ai.usage.details.cached_content_tokens` = **73,673** ✅  
- **Python `Usage.cache_read_tokens`** = **0** ❌

The `details` dict passed to OTel also has the value (it's populated at line 955-956 of `models/google.py`), but the `RequestUsage` object returned to the caller has `cache_read_tokens=0`.

## Root Cause Analysis

We tested `RequestUsage.extract()` directly with a mock `GenerateContentResponse` containing `cachedContentTokenCount=73673` — it correctly returns `cache_read_tokens=73673`. So `genai-prices` mapping works.

The issue appears to be in **streaming mode**: `_metadata_as_usage()` is called on every chunk (line 662), and the final chunk's `usage_metadata` may not include `cached_content_token_count`. The OTel instrumentation captures it from a different path (likely aggregated across chunks), but the `Usage` object only reflects the last chunk.

## Reproduction

1. Use a Google Gemini model with streaming (`agent.iter()`)
2. Send a multi-turn conversation with >1024 input tokens (to trigger implicit caching)
3. On the second turn, check `result.usage().cache_read_tokens` — it will be `0`
4. Check the Logfire/OTel span `gen_ai.usage.cache_read.input_tokens` — it will show the correct cached count

## Environment

- `pydantic-ai==1.84.1`
- `genai-prices==0.0.54`
- Google Gemini 3 Flash Preview via PAIG gateway (Vertex AI backend)
- Streaming mode (`agent.iter()`)

## Impact

Applications relying on `Usage.cache_read_tokens` for cost calculation will overcharge for Google models — treating all input tokens as full-price when a large portion may be cached (we observed 96% cache hit rates).

adtyavrdhn:
Hey @georg-ort 

Thanks for reporting and the details, appreciate it. I'll have a look!

georg-ort:
[...]

PR #6753, opened 2026-07-27
Capture Anthropic thinking tokens in usage details

### LINKED ISSUE #3249: Track Anthropic thinking tokens separate from other tokens
### Description

It would be nice to be able to track thinking tokens separate from other tokens, both in input and output. For my use case, it would help inform how we use thinking and understand it's value vs cost tradeoff.

### References

_No response_

jakub-orsula-parrot:
Fixes #3249

_Supersedes #6752, which the PR-checker auto-closed for not carrying a closing keyword. The GitHub API refused to reopen it (422), hence a fresh PR — same branch, same commit._

Also unblocks the Anthropic half of #6559 (referenced, not closed by this PR — the OTel `gen_ai.usage.reasoning.output_tokens` / `gen_ai.request.reasoning.level` normalization proposed there is out of scope here).

Anthropic reports reasoning tokens at `usage.output_tokens_details.thinking_tokens` (anthropic-sdk-python [v0.105.0](https://github.com/anthropics/anthropic-sdk-python/blob/main/CHANGELOG.md), 2026-05-28), but `_extract_usage_details` reads only the top-level and compaction/advisor token keys, so the count is dropped. Anthropic is currently the only frontier provider whose reasoning tokens are unobservable:

| Provider | Detail key | Source |
|---|---|---|
| OpenAI | `reasoning_tokens` | `models/openai.py` |
| xAI | `reasoning_tokens` | `models/xai.py` |
| Google | `thoughts_tokens` | `models/google.py` |
| Anthropic | — | — |

This adds the missing capture — the prerequisite #6559 identifies ("So Anthropic's `_extract_usage_details` can now capture this, unblocking #3249"). It deliberately stops there: the OTel normalization in #6559 is `needs-discussion`, and this stands alone without it.

### Notes on the implementation

- **Not additive.** Thinking tokens are billed within `output_tokens` (the SDK field docstring: *"Always ≤ output_tokens; output_tokens - thinking_tokens approximates the non-reasoning output"*), so this is a readable subset and nothing is folded into request totals or reaches genai-prices as new cost. The test asserts `total_tokens == 103` for a 3/100 response carrying 40 thinking tokens.
- **Placement matters.** The extraction sits before the `iterations` early returns, so it applies to responses without compaction or advisor iterations — i.e. the common case. Appending at the end of the function would silently skip those.
- **Covers streaming for free.** `_extract_usage_details` is the mapper both `_process_response` and the two streaming `_map_usage` call sites funnel through, and `BetaMessageDeltaUsage` declares `output_tokens_details` too. Both paths are tested.
- **Omitted when zero,** following Google's `thoughts_tokens` precedent. This keeps `thinking_tokens: 0` off every non-thinking request and leaves all existing snapshots and cassettes untouched. Happy to switch to OpenAI's always-write-`0` shape if you'd prefer the explicit form.
- No dependency change needed: the repo floor is already `anthropic>=0.108.0` (lock: 0.109.0).

### Testing

- `tests/models/test_anthropic.py`: 273 passed, including two new tests (non-streaming via `MockAnthropic`, and the cumulative streaming delta merge through `_map_usage`).
- Also ran every other test file whose cassettes contain `thinking_tokens` — `test_cache_prefix_stability.py`, `test_code_execution_files_vcr.py`, `test_thinking_native_tools.py`, plus `test_usage_limits.py`: 44 passed.
- Full suite: identical failure set before and after (195 pre-existing Bedrock/AWS-credential failures in my sandbox, unrelated), passing count +2 — exactly the new tests.
- `ruff format` / `ruff check` / `pyright` clean on both changed files.

### Context

I hit this in production: Anthropic is the primary model in a document-processing pipeline where we track per-call token accounting, and reasoning cost is the one component we can't see. We're currently carrying a local monkeypatch of this exact function as a workaround, which is what prompted looking upstream.

I know CONTRIBUTING asks for maintainer alignment before code on anything non-trivial — I've kept this to the narrow provider-parity fix rather than the broader OTel work in #6559 on the assumption that puts it near the "obvious small fix" end. Entirely happy for you to rewrite, supersede, or close it if you'd rather take the change forward yourselves or fold it into #6559; I can also validate whatever lands against our real Anthropic traffic.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6753?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/capabilities/thinking.md
+++ b/docs/capabilities/thinking.md
@@ -110,6 +110,8 @@ agent = Agent(model, model_settings=settings)
 ...
 ```
 
+Anthropic reports how many thinking tokens it used in [`RunUsage.details`][pydantic_ai.usage.RunUsage.details] under the `thinking_tokens` key. They are billed within `output_tokens`, so they are a readable subset of the output total rather than an addition to it, and the key is omitted entirely when a response used no thinking tokens.
+
 ### Interleaved Thinking
 
 To enable [interleaved thinking](https://docs.anthropic.com/en/docs/build-with-claude/extended-thinking#interleaved-thinking), you need to include the beta header in your model settings:
--- a/pydantic_ai_slim/pydantic_ai/models/anthropic.py
+++ b/pydantic_ai_slim/pydantic_ai/models/anthropic.py
@@ -2441,6 +2441,13 @@ def _extract_usage_details(response_usage: BetaUsage | BetaMessageDeltaUsage) ->
         if isinstance((value := getattr(response_usage, key, None)), int):
             details[key] = value
 
+    # Anthropic bills thinking tokens inside `output_tokens`, so this is a readable subset of the
+    # output total rather than an additive one, matching `reasoning_tokens` on OpenAI and
+    # `thoughts_tokens` on Google.
+    output_tokens_details = response_usage.output_tokens_details
+    if output_tokens_details is not None and (thinking_tokens := output_tokens_details.thinking_tokens):
+        details['thinking_tokens'] = thinking_tokens
+
     iterations = response_usage.iterations
     if not iterations:
         return details
--- a/tests/models/test_anthropic.py
+++ b/tests/models/test_anthropic.py
@@ -4265,6 +4265,9 @@ async def test_anthropic_opus_48_features(allow_model_requests: None, anthropic_
         }
     )
     assert any(isinstance(p, TextPart) for p in response.parts)
+    # Thinking was enabled but the recording reports `thinking_tokens: 0`, which is omitted
+    # rather than written as a zero.
+    assert 'thinking_tokens' not in result.usage.details
 
 
 _REFUSAL_CASE_PARAMS = [
@@ -4844,7 +4847,11 @@ def test_streaming_usage():
 
 
 def test_streaming_usage_thinking_tokens():
-    """Streaming usage is cumulative, so `thinking_tokens` from a delta event replaces the running value."""
+    """`thinking_tokens` carried by a `message_delta` lands in the merged streaming usage.
+
+    A unit test rather than a VCR one: it pins how a delta merges into the running usage, which a
+    cassette cannot protect because the matcher replays whatever delta was recorded regardless.
+    """
     start = BetaRawMessageStartEvent(message=anth_msg(BetaUsage(input_tokens=1, output_tokens=1)), type='message_start')
     initial_usage = _map_usage(start, 'anthropic', '', 'unknown')
     delta = BetaRawMessageDeltaEvent(
@@ -7493,6 +7500,9 @@ async def test_anthropic_advisor_tool(allow_model_requests: None, anthropic_api_
     assert details['advisor_output_tokens'] > 0
     # Advisor tokens bill at the advisor model's rates and are excluded from the request totals.
     assert result.usage.input_tokens == details['input_tokens']
+    # Recorded top-level `output_tokens_details.thinking_tokens`, billed within `output_tokens`.
+    assert details['thinking_tokens'] == 28
+    assert details['thinking_tokens'] < details['output_tokens']
 
 
 @pytest.mark.vcr()
@@ -7525,6 +7535,8 @@ async def test_anthropic_advisor_tool_stream(
     content = returns[0].content
     assert isinstance(content, dict)
     assert content['type'] == 'advisor_result'
+    # Recorded on a streaming `message_delta`, which is the only place the field appears mid-stream.
+    assert agent_run.result.usage.details['thinking_tokens'] == 47
 
 
 @pytest.mark.vcr()
--- a/tests/models/test_anthropic.py
+++ b/tests/models/test_anthropic.py
@@ -132,6 +132,7 @@ with try_import() as imports_successful:
         BetaMessageDeltaUsage,
         BetaMessageIterationUsage,
         BetaMessageTokensCount,
+        BetaOutputTokensDetails,
         BetaRawContentBlockDeltaEvent,
         BetaRawContentBlockStartEvent,
         BetaRawContentBlockStopEvent,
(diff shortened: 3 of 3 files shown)


dsfaccini:
This comment was posted by Claude Code using claude-opus-5 on behalf of David. David made the two judgement calls below (the zero-shape and the fix scope); he has not read the diff line by line.

🔍 Local review suite (automated, pre-merge)

|---|---|---|---|---|

**Verdict:** ✅ PASS — was 🔧 7 required; 6 resolved in `1a827f364`, 1 deliberately skipped · 6 applied

Reviewed at 2026-07-28 · diff `8656cf2d` (waves ran against `060d4e1bc`; the fixes touch tests and docs only, `models/anthropic.py` is unchanged)

<details>
<summary>✅ 6 required action(s) resolved</summary>

_The 6 required actions the review flagged, now resolved in `1a827f364`._

**1. Wire parsing was guarded only by tautological tests** · `tests/models/test_anthropic.py`
Both new tests constructed the `BetaUsage` / `BetaMessageDeltaUsage` they then asserted on, so they could not fail if Anthropic moved or stopped populating the field. Three cassettes already in the repo carry the real value, so live coverage cost no new recording: `test_anthropic_advisor_tool` (28, non-streaming), `test_anthropic_advisor_tool_stream` (47, on a `message_delta`), `test_anthropic_opus_48_features` (0).

**2. Omit-when-zero was unpinned** · `tests/models/test_anthropic.py` · `test_anthropic_opus_48_features`
`thinking_tokens` is a required `int` in the SDK, so `0` is the common value, not an edge case — line coverage read 100% while the decision sat unprotected. Now asserted as absent.

**3. Load-bearing facts lived only in the cassettes** · `tests/models/test_anthropic.py`
Per `tests/AGENTS.md`, a recorded fact asserted nowhere in a test body is effectively unreviewed. The values above are now pinned in the test bodies.

**4. A docstring claimed behavior its test never exercised** · `tests/models/test_anthropic.py` · `test_streaming_usage_thinking_tokens`
It read *"`thinking_tokens` from a delta event replaces the running value"*, but the start event carries no `output_tokens_details`, so the test only ever demonstrated *set*. Rewritten to describe what it actually pins.

**5. A unit test didn't say why it isn't a VCR test** · `tests/models/test_anthropic.py` · `test_streaming_usage_thinking_tokens`
`tests/AGENTS.md` requires the justification. It's now stated: the test pins how a delta merges into the running usage, which a cassette can't protect because the matcher replays whatever delta was recorded regardless.

**6. The new `details` key was undocumented** · `docs/capabilities/thinking.md` · `## Anthropic`
The PR's entire value is observability, and no page told a user the key exists. `grep` for `reasoning_tokens|thoughts_tokens|thinking_tokens` across `docs/` returned zero hits, so this was a net-new docs gap, not a stale table. One prose sentence added to the section intro — it applies to extended, interleaved and adaptive thinking alike, so it deliberately does not sit under `### Adaptive Thinking & Effort`.

</details>

<details>
<summary>⚠️ 1 informational (skip)</summary>

**1. New tests sit in the provider monolith rather than the split-out package** · `tests/models/test_anthropic.py`
`tests/AGENTS.md` prefers feature-centric files over appending to the 12,740-line `test_anthropic.py`, and `tests/models/anthropic/` already exists with a `conftest.py` re-exporting `MockAnthropic`. Skipped as a deliberate scope boundary: relocating a contributor's tests widens the diff without changing behavior. Worth doing whenever this file is next split.

</details>

<details>
<summary>On the <code>thinking_tokens == 0</code> shape, since it was raised as an open question</summary>

The PR omits the key when zero (matching Google's `thoughts_tokens` and xAI's `reasoning_tokens`) rather than always writing `0` like OpenAI's `reasoning_tokens`. Keeping the omit shape, for a reason stronger than precedent-matching:

`_map_usage` merges `details = existing | _extract_usage_details(...)`. Because a zero is omitted, a later delta **cannot reset** an earlier non-zero `thinking_tokens` — it carries forward. Always writing `0` would let a `0`-bearing delta clobber the running value mid-stream. This is the same deliberate carry-forward Google relies on (added for #5205). Verified by running the two-delta sequence, not just by reading.

Cross-provider key-name normalization (`gen_ai.usage.reasoning.output_tokens`) remains #6559's scope.

</details>

<details>
<summary>What else the suite checked and cleared</summary>

- **Totals and pricing unaffected** — `total_tokens` is `input + output`, not a sum over `details`; genai-prices' Anthropic extractor is a path allow-list with no `thinking_tokens` entry, and extraction output is byte-identical with and without the key.
- **Lowest-versions safe** — the field landed in anthropic-sdk 0.105.0, floor is `>=0.108.0`; both `BetaUsage` and `BetaMessageDeltaUsage` declare `output_tokens_details`, verified by installing 0.108.0 in a throwaway venv.
- **Bedrock `construct_type` path safe** — that path skips validation, so a missing field is the classic `AttributeError` source. It returns `None`, not absent, and the `is not None` guard handles it.
- **Sibling parity** — 21 model files plus the four Anthropic client hosting surfaces (`AsyncAnthropicBedrock` / `Vertex` / `Foundry` / `Mantle`) audited. All Anthropic surfaces share the single `_map_usage` sink, so one edit covers them; every reasoning-capable sibling already captures its own key; `bedrock.py`'s Converse API has no reasoning field to capture. 0 parity gaps.
- **No snapshot churn needed** — five committed cassettes already carry non-zero `thinking_tokens`, so the new key activates on already-recorded traffic. Confirmed harmless by running the affected tests, which assert individual keys rather than whole-dict snapshots.

</details>

dsfaccini:

FYI on #3249, now shipped in #6753: it long looked closed-off, but the original blocker went away. The Anthropic API didn't expose reasoning-token counts when this was last discussed (2025-10-24) — `anthropic` v0.105.0 added `output_tokens_details.thinking_tokens` after that. #6753 captures it into `usage.details['thinking_tokens']`, mirroring Google's `thoughts_tokens` and xAI's `reasoning_tokens`; cross-provider normalization stays with #6559.

PR #6774, opened 2026-07-28
Fix `ALLOW_MODEL_REQUESTS` guards for embeddings

### LINKED ISSUE #6763: `ALLOW_MODEL_REQUESTS` is not enforced on the embeddings path (or by `BedrockConverseModel`), so test suites can reach live providers
### Summary

`pydantic_ai.models.ALLOW_MODEL_REQUESTS` is the documented way to make live
model calls impossible in a test suite. It is enforced by convention: each
concrete model is expected to call `check_allow_model_requests()` itself.
Nothing enforces that convention, and it has drifted in two places:

- **No concrete `EmbeddingModel` calls it.** `Embedder.embed_*` reaches the
  provider even when the flag is `False`.
- **`BedrockConverseModel` does not call it either**, so the gap is not
  limited to embeddings.

The failure is silent and fails open. A suite that sets the flag and believes
it is offline makes real, billed requests and sends whatever text it is
embedding to a third party. That is the outcome the setting exists to prevent,
and nothing in the API surface suggests it does not apply.

### Versions

Reproduced on **v2.18.0** (latest release) with `openai` 2.48.0 on Python
3.12, and confirmed by inspection on **`main`** at `eccfb56b68c7`. Line numbers
below are `main`'s.

### Reproduction

No API key and no third party needed: point a provider at a dead local port
and compare the errors. A refused connection proves a socket was attempted;
`RuntimeError` proves the guard fired first.

```python
import pydantic_ai.models as models
models.ALLOW_MODEL_REQUESTS = False   # "make a live call impossible"

from pydantic_ai import Agent, Embedder
from pydantic_ai.embeddings.openai import OpenAIEmbeddingModel

DEAD = "http://127.0.0.1:1/v1"   # nothing listening

try:
    Agent(OpenAIChatModel("gpt-4o",
        provider=OpenAIProvider(base_url=DEAD, api_key="x"))).run_sync("hi")
except Exception as e:
    print("AGENT:     ", type(e).__name__, e)

try:
    Embedder(OpenAIEmbeddingModel("text-embedding-3-small",
        provider=OpenAIProvider(base_url=DEAD, api_key="x"))).embed_query_sync("hi")
except Exception as e:
    print("EMBEDDINGS:", type(e).__name__, e)
```

Observed on v2.18.0:

```
AGENT:      RuntimeError Model requests are not allowed, since ALLOW_MODEL_REQUESTS is False
EMBEDDINGS: ModelAPIError Connection error.
```

The agent never opens a socket. The embedder does. Against a real base URL
that request would have been served, and billed.

### Where it diverges

`check_allow_model_requests()` (defined at `models/__init__.py:1200`) is called
by eight chat-model modules: `models/openai.py`, `anthropic.py`, `cohere.py`,
`google.py`, `groq.py`, `huggingface.py`, `mistral.py`, `xai.py` — for example
`models/openai.py:956` in `OpenAIChatModel.request`. `CerebrasModel`,
`OllamaModel`, `OpenRouterModel` and `ZaiModel` subclass `OpenAIChatModel`, and
`BedrockMantleResponsesModel` subclasses `OpenAIResponsesModel` (which guards
its own `request` and `request_stream`), so all five inherit it.

Two places do not call it:

**1. Every concrete embedding model.** No module under
`pydantic_ai/embeddings/` references it at all. Each `embed()` goes straight to
its client:

| file | `async def embed` |
| --- | --- |
| `embeddings/openai.py` | 121 |
| `embeddings/cohere.py` | 154 |
| `embeddings/google.py` | 212 |
| `embeddings/bedrock.py` | 572 |
| `embeddings/voyageai.py` | 147 |
| `embeddings/sentence_transformers.py` | 124 |

**2. `BedrockConverseModel`** (`models/bedrock.py:478`), whose `request`
(`:657`) and `request_stream` (`:710`) both omit it, unlike its eight peers.
This one is source inspection rather than a live repro, since I have no
Bedrock credentials to hand.

This reads as omission rather than intent, since the docstring on
`check_allow_model_requests` already states the rule that would cover both:

> If you're defining your own models that have costs or latency associated
> with their use, you should call this in `Model.request` and
> `Model.request_stream`.

An embedding request has both costs and latency, and so does Bedrock.

There is also some evidence the gap has already cost you time from the inside:
#6253 fixed `sentence-transformers` embedding tests reaching the HF Hub on
CI by setting offline env vars for that one test class. `ALLOW_MODEL_REQUESTS`
is the mechanism that would otherwise have covered it.

### Suggested fix

Follow the existing convention rather than adding a new mechanism: call
`check_allow_model_requests()` at the top of each concrete
`EmbeddingModel.embed()`, and in `BedrockConverseModel.request` /
`request_stream`. `TestEmbeddingModel` stays exempt, exactly as `TestModel` and
`FunctionModel` are, so injected stubs keep working offline.

Two smaller points either way:

- `SentenceTransformerEmbeddingModel` arguably should not be guarded, since it
  runs in process and costs nothing external. Per-model placement handles that
  naturally; a single check in `Embedder.embed` would not.
- Since this is a convention that several implementations have now missed, it
  may be worth a test that walks the concrete model and embedding-model
  classes and asserts each refuses under `ALLOW_MODEL_REQUESTS = False`, with
  an explicit exempt-list for the local ones. That turns the convention into
  something enforced rather than remembered, and would have caught both of
  these.

Happy to open a PR if the approach looks right, though equally happy to leave
it to you if you would rather shape it yourselves.

### How this surfaced

Found while building an embeddings-backed feature in an application that
handles unpublished research, where a test accidentally reaching a provider
would be a data-disclosure incident rather than an inconvenience. We worked
around it by calling `check_allow_model_requests()` in our own adapter before
delegating to `Embedder`, skipping the check when a stub model is injected so
the offline path still runs. That works for us, but every user of the
embeddings API has the same exposure and would each need the same workaround.

zcxGGmu:
## Summary

- Guard external embedding model requests when `ALLOW_MODEL_REQUESTS` is disabled.
- Guard `BedrockConverseModel` request and streaming entry points consistently with other model providers.
- Add regression coverage for each affected embedding implementation and both Bedrock request modes.

Closes #6763

## Testing

- `PYTHONUTF8=1 uv run pytest tests/test_embeddings.py::test_openai_embedding_model_blocks_requests_when_disabled tests/test_embeddings.py::test_cohere_embedding_model_blocks_requests_when_disabled tests/test_embeddings.py::test_google_embedding_model_blocks_requests_when_disabled tests/test_embeddings.py::test_bedrock_embedding_model_blocks_requests_when_disabled tests/test_embeddings.py::test_voyageai_embedding_model_blocks_requests_when_disabled tests/models/test_bedrock.py::test_bedrock_model_blocks_requests_when_disabled -q -rA`
- `PYTHONUTF8=1 uv run ruff check pydantic_ai_slim/pydantic_ai/embeddings/openai.py pydantic_ai_slim/pydantic_ai/embeddings/cohere.py pydantic_ai_slim/pydantic_ai/embeddings/google.py pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py pydantic_ai_slim/pydantic_ai/embeddings/voyageai.py pydantic_ai_slim/pydantic_ai/models/bedrock.py tests/test_embeddings.py tests/models/test_bedrock.py`
- `PYTHONUTF8=1 uv run ruff format --check pydantic_ai_slim/pydantic_ai/embeddings/openai.py pydantic_ai_slim/pydantic_ai/embeddings/cohere.py pydantic_ai_slim/pydantic_ai/embeddings/google.py pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py pydantic_ai_slim/pydantic_ai/embeddings/voyageai.py pydantic_ai_slim/pydantic_ai/models/bedrock.py tests/test_embeddings.py tests/models/test_bedrock.py`
- `PYTHONUTF8=1 PYRIGHT_PYTHON_IGNORE_WARNINGS=1 uv run pyright pydantic_ai_slim/pydantic_ai/embeddings/openai.py pydantic_ai_slim/pydantic_ai/embeddings/google.py tests/models/test_bedrock.py`

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6774?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/embeddings.md
+++ b/docs/embeddings.md
@@ -840,6 +840,10 @@ async def test_my_rag_system():
         assert test_model.last_settings is not None
 ```
 
+Setting [`ALLOW_MODEL_REQUESTS`][pydantic_ai.models.ALLOW_MODEL_REQUESTS] to `False` also blocks embedding requests, so an embedder you forgot to override raises instead of quietly calling the provider. [`TestEmbeddingModel`][pydantic_ai.embeddings.TestEmbeddingModel] and [`SentenceTransformerEmbeddingModel`][pydantic_ai.embeddings.sentence_transformers.SentenceTransformerEmbeddingModel] are unaffected, as neither reaches a provider.
+
+This covers [`count_tokens()`][pydantic_ai.embeddings.Embedder.count_tokens] as well, but only where tokenization happens server-side: Google and Cohere count tokens through an API call and are blocked, while OpenAI tokenizes locally with `tiktoken` and is not.
+
 ## Instrumentation
 
 Enable OpenTelemetry instrumentation for debugging and monitoring:
--- a/pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
+++ b/pydantic_ai_slim/pydantic_ai/embeddings/bedrock.py
@@ -12,6 +12,7 @@ import anyio
 import anyio.to_thread
 
 from pydantic_ai.exceptions import ModelAPIError, ModelHTTPError, UnexpectedModelBehavior, UserError
+from pydantic_ai.models import check_allow_model_requests
 from pydantic_ai.providers import Provider, infer_provider
 from pydantic_ai.providers.bedrock import remove_bedrock_geo_prefix
 from pydantic_ai.usage import RequestUsage
@@ -572,6 +573,7 @@ class BedrockEmbeddingModel(EmbeddingModel):
     async def embed(
         self, inputs: str | Sequence[str], *, input_type: EmbedInputType, settings: EmbeddingSettings | None = None
     ) -> EmbeddingResult:
+        check_allow_model_requests()
         inputs_list, settings_dict = self.prepare_embed(inputs, settings)
         settings_typed = cast(BedrockEmbeddingSettings, settings_dict)
 
--- a/pydantic_ai_slim/pydantic_ai/embeddings/cohere.py
+++ b/pydantic_ai_slim/pydantic_ai/embeddings/cohere.py
@@ -3,6 +3,7 @@ from dataclasses import dataclass, field
 from typing import Any, Literal, cast
 
 from pydantic_ai.exceptions import ModelAPIError, ModelHTTPError, UnexpectedModelBehavior
+from pydantic_ai.models import check_allow_model_requests
 from pydantic_ai.providers import Provider, infer_provider
 from pydantic_ai.usage import RequestUsage
 
@@ -154,6 +155,7 @@ class CohereEmbeddingModel(EmbeddingModel):
     async def embed(
         self, inputs: str | Sequence[str], *, input_type: EmbedInputType, settings: EmbeddingSettings | None = None
     ) -> EmbeddingResult:
+        check_allow_model_requests()
         inputs, settings = self.prepare_embed(inputs, settings)
         settings = cast(CohereEmbeddingSettings, settings)
 
--- a/pydantic_ai_slim/pydantic_ai/embeddings/cohere.py
+++ b/pydantic_ai_slim/pydantic_ai/embeddings/cohere.py
@@ -216,9 +216,11 @@ class CohereEmbeddingModel(EmbeddingModel):
         return _MAX_INPUT_TOKENS.get(self.model_name)
 
     async def count_tokens(self, text: str) -> int:
-        check_allow_model_requests()
+        # The guard goes below the capability check, not above it: without a v1 client this path never reaches
+        # Cohere, so it should report that token counting is unsupported regardless of `ALLOW_MODEL_REQUESTS`.
         if self._v1_client is None:
             raise NotImplementedError('Counting tokens requires the Cohere v1 client')
+        check_allow_model_requests()
         try:
             result = await self._v1_client.tokenize(
                 model=self.model_name,
--- a/pydantic_ai_slim/pydantic_ai/embeddings/cohere.py
+++ b/pydantic_ai_slim/pydantic_ai/embeddings/cohere.py
@@ -216,6 +216,7 @@ class CohereEmbeddingModel(EmbeddingModel):
         return _MAX_INPUT_TOKENS.get(self.model_name)
 
     async def count_tokens(self, text: str) -> int:
+        check_allow_model_requests()
         if self._v1_client is None:
             raise NotImplementedError('Counting tokens requires the Cohere v1 client')
         try:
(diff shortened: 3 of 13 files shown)


dsfaccini:
This comment was posted by Claude Code using claude-opus-5 on behalf of David. David approved the decision to widen scope; he has not read the diff line by line.

Pushed three commits to your branch — heads up, they extend your fix past what #6763 asked for.

`count_tokens` is the third provider-reaching entry point alongside `request` / `request_stream` / `embed`, and `GoogleModel` / `OpenAIResponsesModel` already guard it — so the four that didn't were a miss, not a scope line: `BedrockConverseModel`, `AnthropicModel`, `GoogleEmbeddingModel`, `CohereEmbeddingModel`.

The two embedding ones are why this didn't wait for a follow-up PR: they ship the user's text to the provider — the exact harm #6763 describes — and `Embedder.count_tokens` reaches them directly, without passing through the `embed()` you guarded.

Your `test_bedrock.py` change was the tell: `test_bedrock_count_tokens_non_http_error` now needs `allow_model_requests` for the same reason its `request` / `request_stream` siblings did — and two more tests turned out to need it too.

<details><summary>The sinks, and three smaller changes from review</summary>

| | sink |
|---|---|
| `BedrockConverseModel.count_tokens` | `self.client.count_tokens` |
| `AnthropicModel.count_tokens` | `client.beta.messages.count_tokens` |
| `GoogleEmbeddingModel.count_tokens` | `client.aio.models.count_tokens(contents=text)` |
| `CohereEmbeddingModel.count_tokens` | `_v1_client.tokenize(text=text, offline=False)` |

1. On Cohere the guard sits *below* the `_v1_client is None` check — without a v1 client that path never reaches Cohere, so it should report token counting as unsupported regardless of the flag.
2. Added an `Embedder.embed_query` guard test — none of the existing ones went through the public surface the issue actually names.
3. `docs/embeddings.md` and the `check_allow_model_requests` / `ALLOW_MODEL_REQUESTS` docstrings now describe the widened scope; neither mentioned embeddings before.

</details>

Nothing wrong with your patch — it did exactly what the issue described.

PR #6776, opened 2026-07-28
Add `pydantic_ai.history_repair.*` instrumentation attributes

### LINKED ISSUE #6495: Add OTel observability for silent message-history mutations (repair, sanitize, prepare_messages)
### Description

_Issue drafted by Claude (Claude Code), on behalf of DouweM. Follow-up from #6319._

#6319 made message-history repair automatic and silent: before every model request, `_clean_message_history` synthesizes `ToolReturnPart`s for dangling tool calls, drops orphaned tool results, and merges consecutive same-role messages. Silence is the right default for the *framework's own* transient shapes (a `UserWarning` was tried and fired on the framework's legitimate `ModelRetry`-from-hook flow under `filterwarnings=["error"]`, so it was removed) — but it means a user whose history is being repaired on every request has no signal unless they know to look for the `pydantic_ai_synthesized_tool_return` metadata marker.

That signal matters in practice: a repair firing every turn usually points at an upstream bug — a history processor that trims calls but not results, a persistence layer that loses the tail, forgotten `deferred_tool_results` — and today it's invisible.

The intended non-noisy surface is OpenTelemetry: attributes/events on the existing model-request span, visible in Logfire (or any OTel backend), no stdout noise, no `filterwarnings` interaction.

### Proposal

On the model-request span, when the wire-level repair pass changed anything, record what it did — e.g. an attribute set or a span event like:

- `pydantic_ai.history_repair.synthesized_tool_returns`: count (+ tool names / tool_call_ids in an event payload)
- `pydantic_ai.history_repair.dropped_orphaned_results`: count
- `pydantic_ai.history_repair.merged_messages`: count (arguably noise — merging is normal shape-normalization and may deserve exclusion)

Zero-change requests record nothing, so dashboards/alerts can key on presence.

(Attribute naming should follow whatever conventions `_otel_*.py` already establishes; per the repo guideline those modules implement only spec-defined features, so this likely belongs as pydantic-ai-specific attributes on our own spans rather than in the GenAI semconv mapping.)

### Sibling silent mutations worth including (or spinning off)

While the repair pipeline is the motivating case, it's not the only place the framework silently rewrites what the model sees; a single "history mutation observability" pass could cover:

- **AG-UI `sanitize_messages` tail-strip** (GHSA-jpr8 hardening): unresolved trailing tool calls from client-supplied history are dropped — security-relevant and currently only a `warnings.warn`.
- **`Model.prepare_messages` translations**: e.g. typed `NativeToolSearch*Part` downgraded to local-shape parts when the active profile doesn't support tool search — the model sees a different history than what's persisted.
- **UI-adapter round-trip degradations**: `tool_kind` loss below AG-UI 0.1.11 (currently a `UserWarning`), `RetryPromptPart` → `ToolReturnPart` reload, part-level `metadata` loss — these happen outside a request span, so they may need a different surface, but they're part of the same "the transcript changed and nobody saw it" family.
- **History processors** are user code and don't need instrumentation themselves, but the repair attributes above effectively surface their mistakes (a processor that breaks pairing shows up as nonzero repair counts on the next request).

### Non-goals

- No warnings/stdout output (see above — tried and reverted in #6319).
- No behavior change to the repair itself.

chakshu-dhannawat:
Closes #6495

This exposes the silent history-repair work from `_clean_message_history` as OpenTelemetry span attributes on the model request span.

- `pydantic_ai.history_repair.synthesized_tool_returns` - synthesized tool returns for dangling calls.
- `pydantic_ai.history_repair.dropped_orphaned_results` - orphaned tool results dropped.
- `pydantic_ai.history_repair.merged_messages` - consecutive compatible messages merged.

The counters are collected in `_clean_message_history` and passed to `open_model_request_span` via a new context var, then emitted only when non-zero. Existing call sites and tests are updated to unpack the new return value.

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6776?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/message-history.md
+++ b/docs/message-history.md
@@ -235,6 +235,12 @@ After the invalid parts are handled, consecutive compatible messages are **merge
 
 The repair is deterministic and idempotent: repairing the same history always produces the same output, running a repaired history through another run leaves it untouched, and synthesized parts contain no wall-clock data, so reuse doesn't invalidate provider prompt caches.
 
+When [OpenTelemetry instrumentation](logfire.md) is active, the model request span records how much repair happened on that request, so you can see at a glance whether a history was already clean or needed intervention. Non-zero counters are reported under the `pydantic_ai.history_repair.*` span attributes:
+
+- `pydantic_ai.history_repair.synthesized_tool_returns` - how many dangling [`ToolCallPart`][pydantic_ai.messages.ToolCallPart]s were closed with a synthesized [`ToolReturnPart`][pydantic_ai.messages.ToolReturnPart].
+- `pydantic_ai.history_repair.dropped_orphaned_results` - how many orphaned tool-result parts were removed.
+- `pydantic_ai.history_repair.merged_messages` - how many merge operations combined consecutive compatible messages.
+
 Tool calls that can still receive a real result are left alone: when the history ends on a `ModelResponse` with tool calls, running without a new `user_prompt` executes them, and [deferred tool calls](deferred-tools.md) are matched to their `deferred_tool_results` — including when a 'complete' `ModelRequest` with the already-executed results follows the response. Repair of that live frontier only happens when the interruption is evident: a final response with [`state='interrupted'`][pydantic_ai.messages.ModelResponse.state] or a trailing request with [`state='interrupted'`][pydantic_ai.messages.ModelRequest.state] (e.g. from a [cancelled stream](output.md#cancelling-streams) or a crash during tool execution) whose tool calls will never be executed.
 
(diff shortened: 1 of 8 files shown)


chakshu-dhannawat:
Closing to recreate with the correct author/email on the commits.

PR #6778, opened 2026-07-28
Document when an unsupported option should warn rather than raise `UserError`

dsfaccini:
This pull request was posted by Claude Code using claude-opus-5 on behalf of David. David reached the conclusion this rule states; Claude wrote it up and verified it against the codebase and version policy.

One rule in `agent_docs/api-design.md`, no code change. Opening it separately from #6738 so the principle can be argued on its own.

**The rule:** when an upstream API or protocol removes a capability we already expose, the option that configured it should **warn and degrade**. `UserError` is for a requested *operation* that can't return a result.

### Why this came up

FastMCP 4 / MCP SDK v2 removes three things `MCPToolset` exposes today: `logging/setLevel`, and (via SEP-2577) server-initiated sampling and elicitation. So `log_level=`, `sampling_model=`, and `elicitation_handler=` become options we accept and can never honour.

The obvious move is `raise UserError`, matching ~18 `raise UserError('... is not supported ...')` sites in `models/`. That reading is wrong, and the difference is worth naming: **every one of those precedents has raised since the surface was introduced.** No user's working code ever started failing because of them. Raising here is different — it takes code that runs today and breaks it on a minor upgrade, which the [version policy](https://github.com/pydantic/pydantic-ai/blob/main/docs/version-policy.md) rules out.

### The test

Can the application carry on without the capability?

- **Yes → warn and continue.** An unapplied log level, or a sampling handler the server can no longer invoke, doesn't stop anything from working. The user needs to know, not to crash.
- **No → raise.** An operation with nothing to return has no degraded form: a call the server can't run, an output mode the model can't produce.

Both sides show up in #6738. `log_level` on a modern session warns and skips; `direct_call_tool(..., use_task=True)` against a session with no task path still raises, because there is no result to hand back.

@DouweM — flagging you since this constrains how every future provider/protocol removal gets handled, not just the MCP one.

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6778?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


agent_docs/api-design.md
```diff
@@ -26,3 +26,5 @@
 - Keep `NativeToolReturnPart.content` flat and non-redundant — avoid duplicating part fields, repeating `return_value` data, single-key wrappers, or unnecessary lists — Reduces API surface area, prevents inconsistencies between duplicate fields, and simplifies consumption for both users and AI assistants
 <!-- rule:265 -->
 - Use `'provider:model'` format (e.g., `'openai:gpt-4'`, `'anthropic:claude-3'`) and `infer_model()` for instantiation — Ensures consistent model reference syntax across code, docs, and CLI; provides unified instantiation interface that prevents fragmentation
+<!-- rule:1187 -->
```
dsfaccini, line 29: using tje word "capability" here is confusing, bc we're ostly talking about settings and profile flags, API fields. this could have been phrased more humanely instead of getting into this awfully unspecific LLM talk

PR #6070, opened 2026-06-26
fix(heroku): route model names through family profiles so thinking isn't dropped

Bartok9:
## Summary

- `HerokuProvider.model_profile()` ignored the model name and returned a bare `OpenAIModelProfile` for every model.
- As a result, `supports_thinking` defaulted to unset/`False`, so the unified `thinking` setting was **silently dropped** for every reasoning-capable model Heroku serves (Claude, DeepSeek-R1, …).

## Motivation

Closes #6022.

Heroku Managed Inference serves models from several families under **bare model names** (no provider prefix), e.g. `claude-3-7-sonnet`, `deepseek-r1`, `gpt-oss-120b`. `model_profile()` hard-coded a bare OpenAI profile and never inspected `model_name`, so capabilities resolved from the family profile (most importantly `supports_thinking`) were never applied. Users could pass `ModelSettings(thinking=True)` against `heroku:claude-3-7-sonnet`, get no warning, and have it never reach the wire — while direct `anthropic:claude-3-7-sonnet` correctly emits the thinking block.

## Root Cause

**Symptom** — `thinking=True` accepted but no extended-thinking is requested for Heroku-hosted reasoning models.

**Root cause** — `HerokuProvider.model_profile()` (`providers/heroku.py`) returned `OpenAIModelProfile(json_schema_transformer=OpenAIJsonSchemaTransformer)` unconditionally; `model_name` was unused, so `supports_thinking` (set only by the per-family profiles like `anthropic_model_profile`/`deepseek_model_profile`) was never picked up.

**Evidence** — `tests/test_capabilities.py` confirms Heroku hosts reasoning-capable Claude/DeepSeek/gpt-oss models; the old `test_heroku_model_profile` only asserted the transformer, so the drop was invisible to CI. Real-run output below.

**Fix + why this level** — Fix at the provider's `model_profile`, mirroring the existing `NebiusProvider` routing pattern: map the bare name's prefix to the matching family profile and `merge_profile` it onto the OpenAI-compatible base. This is the canonical layer for per-model capability resolution; patching downstream (e.g. the model adapter) would not generalize across families.

**Scope / risk** — Only touches `HerokuProvider.model_profile`. Unmapped models fall back to the prior bare base profile unchanged (verified). The OpenAI `json_schema_transformer` base is preserved for all models.

## Verification

- `uv run pytest tests/providers/test_heroku.py -q` — 7 passed
- `uv run pytest tests/providers/ -q` — 320 passed, 227 skipped
[...]

PR #6815, opened 2026-07-29
fix(providers): pass lowercased model name to family profiles in SambaNova/Heroku/Fireworks

### LINKED ISSUE #6816: SambaNova/Heroku/Fireworks pass mixed-case model names to case-sensitive family profiles, silently dropping thinking
### Description

`SambaNovaProvider.model_profile` matches the family-profile prefix against the **lowercased** model name but passes the **original casing** to the profile function:

https://github.com/pydantic/pydantic-ai/blob/main/pydantic_ai_slim/pydantic_ai/providers/sambanova.py#L67-L73

```python
profile = None
model_name_lower = model_name.lower()

for prefix, profile_func in prefix_to_profile.items():
    if model_name_lower.startswith(prefix):
        profile = profile_func(model_name)   # <- original casing
        break
```

Every family profile it dispatches to is case-sensitive:

| function | match |
|---|---|
| `deepseek_model_profile` | `startswith('deepseek-r1')`, `startswith('deepseek-v4-')` |
| `qwen_model_profile` | `startswith('qwen-3-coder')`, `_QWEN_3_5_RE` |
| `mistral_model_profile` | `startswith('magistral')` |

So a mixed-case model ID selects the right profile function and then detects nothing inside it. SambaNova's model IDs **are** mixed case — `docs/models/openai.md` uses `sambanova:Meta-Llama-3.1-8B-Instruct`, and `DeepSeek-R1-0528` / `Qwen3-32B` are real IDs — so this fires in ordinary use.

### Reproduction

```python
from pydantic_ai.providers.sambanova import SambaNovaProvider

for name in ('DeepSeek-R1-0528', 'deepseek-r1-0528'):
    profile = SambaNovaProvider.model_profile(name)
    model = OpenAIChatModel(name, provider=SambaNovaProvider(api_key='x'))
    _, params = model.prepare_request({'thinking': True}, ModelRequestParameters())
    print(
        f'{name:18} supports_thinking={profile.get("supports_thinking")!s:6} '
        f'always={profile.get("thinking_always_enabled")!s:6} '
        f'ignore_ws={profile.get("ignore_streamed_leading_whitespace")!s:6} '
        f'-> params.thinking={params.thinking!r}'
    )
```

Output on `main` (`251de3d`):

```
DeepSeek-R1-0528   supports_thinking=False  always=False  ignore_ws=False  -> params.thinking=None
deepseek-r1-0528   supports_thinking=True   always=True   ignore_ws=True   -> params.thinking=True
```

The same model, spelled the way the docs spell SambaNova model names, loses three profile flags.

### Why it's silent

`Model.prepare_request` (`models/__init__.py:418-426`) only forwards `thinking` when the profile advertises support, but strips it from `model_settings` either way:

```python
if model_settings and 'thinking' in model_settings:
    thinking_value = model_settings['thinking']
    supports_thinking = self.profile.get('supports_thinking', False)
    thinking_always_enabled = self.profile.get('thinking_always_enabled', False)
    if supports_thinking or thinking_always_enabled:
        if not (thinking_value is False and thinking_always_enabled):
            params = replace(params, thinking=thinking_value)
    stripped = {k: v for k, v in model_settings.items() if k != 'thinking'}
    model_settings = cast(ModelSettings, stripped) if stripped else None
```

So `thinking=True` is consumed and discarded — no error, no warning, nothing on the wire. `ignore_streamed_leading_whitespace` is lost the same way, which shows up as leading whitespace in streamed DeepSeek-R1 output.

This is the same failure mode as #6022 / #6070, one layer down: Heroku's routing was added so capabilities "are detected instead of silently dropped", and the original casing defeats the detection that routing enables.

### Two more providers with the same defect

**`providers/heroku.py:69-74`** — identical shape (lowercases into `lower_model_name`, then passes `model_name`):

```
DeepSeek-R1   supports_thinking=False
deepseek-r1   supports_thinking=True
```

**`providers/fireworks.py:57-63`** — never lowercases at all, so a mixed-case ID matches no prefix whatsoever and gets the bare OpenAI-compatible profile (`None` rather than `False`):

```
accounts/fireworks/models/DeepSeek-R1   supports_thinking=None
accounts/fireworks/models/deepseek-r1   supports_thinking=True
```

Heroku and Fireworks both document lowercase IDs, so those two are latent rather than currently-firing — but they're the same one-line defect.

### Expected behaviour

`model_profile` should be case-insensitive end to end: the dispatched name should be lowercased, as it already is in the four providers that get this right.

| provider | how it gets this right |
|---|---|
| `cerebras.py:59` | `profile_func(model_name_lower)` |
| `azure.py:52` | `model_name = model_name.lower()` — rebinds, so the dispatch can't diverge |
| `ollama.py:58`, `ovhcloud.py:46` | same rebinding |
| `together.py:57` | same rebinding, then `split('/', 1)` |

Note the pattern: the four correct ones either pass the lowercased variable explicitly or rebind `model_name` itself; all three broken ones keep the lowercased name in a *separate* variable and then forget to use it.

An alternative would be to make each family profile defend itself with an internal `.lower()` — `moonshotai_model_profile` and `zai_model_profile` already do, which is why they're immune. But the majority convention is clearly "the caller passes lowercase", so fixing the three callers is the smaller and more consistent change. Happy to go the other way if you'd prefer the profiles to be robust independently.

### Note on test coverage

`tests/providers/test_sambanova_provider.py` already uses the real mixed-case IDs (`DeepSeek-R1-0528`, `Qwen3-32B`, `Meta-Llama-3.1-8B-Instruct`) but only asserts `profile is not None` — it never inspects a single flag, which is why the bug was invisible to the suite.

### Branch

Fix plus 4 tests: #6815 (currently auto-closed pending this issue). Reverting only the three provider files and keeping the tests fails exactly the four new tests and nothing else; `tests/providers/` + `tests/profiles/` is 404 passed / 110 skipped, unchanged from `main`; ruff and pyright clean; Macroscope's automated review approved it.

```
pydantic-ai main @ 251de3d
Python 3.12, Windows
```

LHMQ878:
Fixes #6816

## Summary

Three providers match their family-profile prefix against the **lowercased** model name but pass the **original casing** to the profile function. Every family profile keys off lowercase prefixes, so a mixed-case model ID selects the right profile function and then detects nothing inside it — `supports_thinking` comes back `False`/unset and a `thinking=True` setting is **silently dropped**.

This is the same failure mode as #6022 / #6070, one layer down: Heroku's routing was added so capabilities "are detected instead of silently dropped", but the original casing defeats the detection it enables.

## The defect

`providers/sambanova.py`:

```python
model_name_lower = model_name.lower()

for prefix, profile_func in prefix_to_profile.items():
    if model_name_lower.startswith(prefix):
        profile = profile_func(model_name)   # <- original casing
        break
```

The profile functions it dispatches to are all case-sensitive:

| function | match |
|---|---|
| `deepseek_model_profile` | `startswith('deepseek-r1')`, `startswith('deepseek-v4-')` |
| `qwen_model_profile` | `startswith('qwen-3-coder')`, `_QWEN_3_5_RE` |
| `mistral_model_profile` | `startswith('magistral')` |
| `google_model_profile` | `'gemini-3' in`, `'gemini-2.5' in` |
| `anthropic_model_profile` | ~20 × `startswith('claude-…')` |

(`moonshotai_model_profile` and `zai_model_profile` lowercase internally, so they're immune — which is what makes the contract "the caller passes lowercase" rather than "each profile defends itself".)

SambaNova's model IDs **are** mixed case. `docs/models/openai.md` uses `sambanova:Meta-Llama-3.1-8B-Instruct`, and `DeepSeek-R1-0528` / `Qwen3-32B` are real IDs, so this fires in ordinary use:

```
model name         supports_thinking  thinking_always_enabled  ignore_streamed_leading_whitespace
DeepSeek-R1-0528   False              False                    False
deepseek-r1-0528   True               True                     True
```

## Why it's silent

`Model.prepare_request` (`models/__init__.py:418-426`) only forwards `thinking` when the profile advertises support, and strips it from `model_settings` either way:

```python
if model_settings and 'thinking' in model_settings:
    thinking_value = model_settings['thinking']
    supports_thinking = self.profile.get('supports_thinking', False)
    thinking_always_enabled = self.profile.get('thinking_always_enabled', False)
    if supports_thinking or thinking_always_enabled:
        if not (thinking_value is False and thinking_always_enabled):
            params = replace(params, thinking=thinking_value)
    stripped = {k: v for k, v in model_settings.items() if k != 'thinking'}
    model_settings = cast(ModelSettings, stripped) if stripped else None
```

So the setting is consumed and discarded — no error, no warning, nothing on the wire:

```python
for name in ('DeepSeek-R1-0528', 'deepseek-r1-0528'):
    m = OpenAIChatModel(name, provider=SambaNovaProvider(api_key='x'))
    _, params = m.prepare_request({'thinking': True}, ModelRequestParameters())
    print(name, params.thinking)

# DeepSeek-R1-0528  None      <- silently dropped
# deepseek-r1-0528  True
```

`ignore_streamed_leading_whitespace` is lost the same way, which shows up as leading whitespace in streamed DeepSeek-R1 output.

## The other two

**Heroku** has the identical shape (`profile_func(model_name)` after `lower_model_name = model_name.lower()`):

```
DeepSeek-R1   supports_thinking=False
deepseek-r1   supports_thinking=True
```

**Fireworks** never lowercases at all, so a mixed-case ID matches no prefix whatsoever and gets the bare OpenAI-compatible profile — `None` rather than `False`:

```
accounts/fireworks/models/DeepSeek-R1   supports_thinking=None
accounts/fireworks/models/deepseek-r1   supports_thinking=True
```

Heroku and Fireworks both document lowercase IDs, so those two are latent rather than currently-firing. They're the same one-line defect and fixing them together keeps the providers consistent.

## The fix

Pass the lowercased name, which is what the rest of the codebase already does:

| provider | how it gets this right |
|---|---|
| `cerebras.py:59` | `profile_func(model_name_lower)` |
| `azure.py:52` | `model_name = model_name.lower()` — rebinds, so the dispatch can't diverge |
| `ollama.py:58`, `ovhcloud.py:46` | same rebinding |
| `together.py:57` | same rebinding, then `split('/', 1)` |

Note the pattern: the four that are correct either pass the lowercased variable explicitly or rebind `model_name` itself. The three that are wrong all keep the lowercased name in a *separate* variable and then forget to use it. Rebinding would arguably be the more robust idiom, but I've kept each provider's existing structure and just corrected the argument, to keep the diff minimal and reviewable.

After the fix, all three providers give identical results for every casing of every model name.

## Tests

`tests/providers/test_sambanova_provider.py` already used the real mixed-case IDs (`DeepSeek-R1-0528`, `Qwen3-32B`, `Meta-Llama-3.1-8B-Instruct`) but only asserted `profile is not None` — it never inspected a single flag, which is why the bug was invisible to it. Added:

- **sambanova** — asserts `DeepSeek-R1-0528` resolves `supports_thinking`/`thinking_always_enabled`/`ignore_streamed_leading_whitespace`, and that the profile is *equal* to the lowercase variant's.
- **sambanova** — the user-visible symptom: `prepare_request({'thinking': True})` must yield `params.thinking is True`.
- **heroku** — `.upper()` of each model name must give an identical profile.
- **fireworks** — `assert_called_with('deepseek-r1')` on the mocked profile function, pinning the exact argument (matching the style of the existing `test_fireworks_provider_model_profile`).

## Validation

- `tests/providers/test_{sambanova_provider,heroku,fireworks}.py`: 30 → **34 passed**
- `tests/providers/ tests/profiles/`: **404 passed, 110 skipped**. One pre-existing failure, `test_anthropic_provider_pass_anthropic_client` (`AsyncAnthropicBedrock` rejecting `api_key` alongside AWS credentials) — verified to fail identically on unmodified `main`, unrelated to this change.
- **Control experiment** — reverting only the three provider files and keeping the tests fails exactly the four new tests, and nothing else:
  ```
  test_sambanova_provider.py:104: AssertionError: assert False is True
  test_sambanova_provider.py:121: assert None is True
  test_heroku.py:118: AssertionError: deepseek-r1
  test_fireworks.py:119: AssertionError: expected call not found.
  4 failed, 30 passed
  ```
- `tests/profiles/test_resolution_matrix.py` snapshots unchanged. The two entries there that use mixed-case names are `test_sambanova_llama` (`Meta-Llama-3.3-70B-Instruct`, which routes through `meta_model_profile` — it ignores the model name entirely, so casing can't affect it) and `test_together_qwen` (`Qwen/Qwen2.5-72B-Instruct-Turbo`, on `together.py`, which was already correct).
- `ruff check` / `ruff format --check` clean; `pyright` **0 errors**.
- Coverage on the three providers unchanged from baseline — the one uncovered line in each is the pre-existing `_set_http_client`.

Happy to split this into three PRs, or to narrow it to SambaNova (the only one currently reachable with documented model names), if you'd prefer.

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/providers/fireworks.py
+++ b/pydantic_ai_slim/pydantic_ai/providers/fireworks.py
@@ -56,7 +56,9 @@ class FireworksProvider(Provider[AsyncOpenAI]):
 
         profile = None
         if model_name.startswith(prefix):
-            model_name = model_name[len(prefix) :]
+            # Match and dispatch on the lowercased name: the family profile functions all key off
+            # lowercase prefixes, so a mixed-case model ID would match nothing at all.
+            model_name = model_name[len(prefix) :].lower()
             for provider, profile_func in prefix_to_profile.items():
                 if model_name.startswith(provider):
                     profile = profile_func(model_name)
--- a/pydantic_ai_slim/pydantic_ai/providers/heroku.py
+++ b/pydantic_ai_slim/pydantic_ai/providers/heroku.py
@@ -67,10 +67,12 @@ class HerokuProvider(Provider[AsyncOpenAI]):
         }
 
         profile = None
+        # The family profile functions match on lowercase prefixes, so they get the lowercased name
+        # too -- passing the original casing would defeat the detection described above.
         lower_model_name = model_name.lower()
         for prefix, profile_func in prefix_to_profile.items():
             if lower_model_name.startswith(prefix):
-                profile = profile_func(model_name)
+                profile = profile_func(lower_model_name)
                 break
 
         # As the Heroku API is OpenAI-compatible, we keep the OpenAIJsonSchemaTransformer as the base
--- a/pydantic_ai_slim/pydantic_ai/providers/sambanova.py
+++ b/pydantic_ai_slim/pydantic_ai/providers/sambanova.py
@@ -65,11 +65,13 @@ class SambaNovaProvider(Provider[AsyncOpenAI]):
         }
 
         profile = None
+        # SambaNova's model names are mixed case (`DeepSeek-R1-0528`, `Qwen3-32B`), and the family
+        # profile functions match on lowercase prefixes, so they need the lowercased name too.
         model_name_lower = model_name.lower()
 
         for prefix, profile_func in prefix_to_profile.items():
             if model_name_lower.startswith(prefix):
-                profile = profile_func(model_name)
+                profile = profile_func(model_name_lower)
                 break
 
         # Wrap into OpenAIModelProfile since SambaNova is OpenAI-compatible
--- a/tests/providers/test_fireworks.py
+++ b/tests/providers/test_fireworks.py
@@ -102,3 +102,19 @@ def test_fireworks_provider_model_profile(mocker: MockerFixture):
     unknown_profile = provider.model_profile('unknown-model')
     assert unknown_profile is not None
     assert unknown_profile.get('json_schema_transformer', None) == OpenAIJsonSchemaTransformer
+
+
+def test_fireworks_provider_model_profile_is_case_insensitive(mocker: MockerFixture):
+    """The family prefixes are lowercase, so a mixed-case model ID must be lowercased first.
+
+    Otherwise `DeepSeek-R1` matches no prefix at all and gets the bare OpenAI-compatible
+    profile -- `supports_thinking` unset, so a `thinking=True` setting is silently dropped.
+    """
+    provider = FireworksProvider(api_key='api-key')
+    deepseek_mock = mocker.patch('pydantic_ai.providers.fireworks.deepseek_model_profile', wraps=deepseek_model_profile)
+
+    profile = provider.model_profile('accounts/fireworks/models/DeepSeek-R1')
+    deepseek_mock.assert_called_with('deepseek-r1')
+    assert profile is not None
+    assert profile.get('supports_thinking') is True
+    assert profile == provider.model_profile('accounts/fireworks/models/deepseek-r1')
--- a/tests/providers/test_heroku.py
+++ b/tests/providers/test_heroku.py
@@ -105,6 +105,23 @@ def test_heroku_model_profile_routes_thinking_capable_families():
     assert fallback.get('supports_thinking') is None
 
 
+def test_heroku_model_profile_is_case_insensitive():
+    """The prefix match is case-insensitive, so the dispatched name has to be too.
+
+    Matching on the lowercased name but handing the family profile the original casing means a
+    mixed-case model ID selects the right profile function and then detects nothing inside it --
+    every family profile keys off lowercase prefixes.
+    """
+    provider = HerokuProvider(api_key='api-key')
+
+    for model_name in ('deepseek-r1', 'claude-3-7-sonnet'):
+        assert provider.model_profile(model_name.upper()) == provider.model_profile(model_name), model_name
+
+    deepseek_profile = provider.model_profile('DeepSeek-R1')
+    assert deepseek_profile is not None
+    assert deepseek_profile.get('supports_thinking') is True
+
+
 async def test_heroku_model_provider_claude_3_7_sonnet(allow_model_requests: None, heroku_inference_key: str):
     provider = HerokuProvider(api_key=heroku_inference_key)
     m = OpenAIChatModel('claude-3-7-sonnet', provider=provider)
--- a/tests/providers/test_sambanova_provider.py
+++ b/tests/providers/test_sambanova_provider.py
@@ -8,6 +8,8 @@ from ..conftest import TestEnv, try_import
 with try_import() as imports_successful:
     import openai
 
+    from pydantic_ai.models import ModelRequestParameters
+    from pydantic_ai.models.openai import OpenAIChatModel
     from pydantic_ai.providers import infer_provider
     from pydantic_ai.providers.sambanova import SambaNovaProvider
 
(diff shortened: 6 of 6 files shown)


LHMQ878:
Thanks — filed #6816 with the reproduction, and I have updated this description to reference it with `Fixes #6816`.

I do not have permission to reopen this PR myself (the API returns 422 for a non-collaborator reopening a bot-closed PR), so this needs a maintainer to reopen it, or I can open a fresh PR against the same branch if that is easier — just say which you prefer.

For context on what is already verified here: Macroscope approved it (`ebf86cc`), `test on 3.13 (pydantic-ai-slim)` and `test on 3.13 (pydantic-evals)` passed before the auto-close, and reverting only the three provider files while keeping the tests fails exactly the four new tests and nothing else.

LHMQ878:
Superseded by #6817 — same branch, same commit (`ebf86cc`), now with `Fixes #6816` in the description so the issue-link guard passes. Closing the loop here so the history is traceable; no need for a maintainer to reopen this one.

PR #6841, opened 2026-07-29
Map xAI finish reasons from proto enums

### LINKED ISSUE #6812: xAI streamed responses always report finish_reason='stop' (map is keyed on lowercase strings, SDK returns proto enum names)
### Description

Every **streamed** xAI response reports `finish_reason='stop'`, whatever the model actually said. `XaiStreamedResponse._update_response_state` does:

https://github.com/pydantic/pydantic-ai/blob/main/pydantic_ai_slim/pydantic_ai/models/xai.py#L966-L967

```python
# Handle finish reason (SDK Response always provides a finish_reason)
self.finish_reason = _FINISH_REASON_MAP.get(response.finish_reason, 'stop')
```

with

```python
_FINISH_REASON_MAP: dict[str, FinishReason] = {
    'stop': 'stop',
    'length': 'length',
    'content_filter': 'content_filter',
    'max_output_tokens': 'length',
    'cancelled': 'error',
    'failed': 'error',
}
```

But the xAI SDK's `Response.finish_reason` is not a lowercase string — it's the proto enum **name** (`xai_sdk/chat.py:1287`, xai-sdk 1.17.0):

```python
@property
def finish_reason(self) -> str:
    """Returns the finish reason of this response."""
    return sample_pb2.FinishReason.Name(self._get_output(sync=False).finish_reason)
```

`sample_pb2.FinishReason` has exactly these members:

```
REASON_INVALID, REASON_MAX_LEN, REASON_MAX_CONTEXT, REASON_STOP, REASON_TOOL_CALLS, REASON_TIME_LIMIT
```

So `response.finish_reason` returns `'REASON_STOP'`, `'REASON_MAX_LEN'`, `'REASON_TOOL_CALLS'`, … — **no key in `_FINISH_REASON_MAP` can ever match**. The map is entirely dead code and the `.get(..., 'stop')` default is taken on every single chunk. The default being a *valid-looking* value is why this passed unnoticed.

Two consequences:

1. **Streaming and non-streaming disagree on the same response.** `_process_response` (line ~870) reads the proto ints via `_FINISH_REASON_PROTO_MAP` and gets this right, so a truncated response is `'length'` unstreamed and `'stop'` streamed.
2. **Intermediate chunks claim the response is finished.** A chunk mid-stream carries `REASON_INVALID` (the proto default = "not finished yet"), which also falls through to `'stop'`, so every intermediate `ModelResponse` snapshot asserts a clean stop. Every other model guards this — e.g. `models/groq.py`:

   ```python
   if raw_finish_reason := choice.finish_reason:
       self.provider_details = {'finish_reason': raw_finish_reason}
       self.finish_reason = _FINISH_REASON_MAP.get(raw_finish_reason)
   ```

### Impact

`'length'` is not cosmetic — two separate diagnostics in `_agent_graph.py` test for it, and both are silently defeated when streaming from xAI:

- `check_incomplete_tool_call` (`_agent_graph.py:337`) raises `IncompleteToolCall` when the model was cut off mid-tool-call. With `'stop'`, the truncated JSON arguments instead surface as an ordinary validation error, or worse, a silently wrong tool call.
- `_agent_graph.py:1749` — the `'Model returned no content, and the token limit was exceeded…'` message. Streaming users get the generic "no content" error with no hint that they hit the token limit.

`REASON_TOOL_CALLS` also never maps to `'tool_call'`, and `REASON_TIME_LIMIT` never maps to `'error'`.

### Reproduction

Part 1 needs nothing but `xai-sdk` — it builds real SDK `Response` objects and tabulates every value the SDK can return against the map the streaming path uses.

<details>
<summary><code>repro_xai_stream_finish_reason.py</code></summary>

```python
from xai_sdk.chat import Response
from xai_sdk.proto import chat_pb2, sample_pb2

from pydantic_ai.models.xai import _FINISH_REASON_MAP

print(f'_FINISH_REASON_MAP keys: {sorted(_FINISH_REASON_MAP)}\n')
print(f'{"SDK Response.finish_reason":<22} {"map hit":<10} {"reported":<12} correct?')
for name, want in {
    'REASON_STOP': 'stop',
    'REASON_MAX_LEN': 'length',
    'REASON_MAX_CONTEXT': 'length',
    'REASON_TOOL_CALLS': 'tool_call',
    'REASON_TIME_LIMIT': 'error',
}.items():
    proto = chat_pb2.GetChatCompletionResponse()
    out = proto.outputs.add()
    out.message.role = chat_pb2.MessageRole.ROLE_ASSISTANT
    out.finish_reason = sample_pb2.FinishReason.Value(name)
    raw = Response(proto, index=0).finish_reason
    reported = _FINISH_REASON_MAP.get(raw, 'stop')
    print(f'{raw:<22} {str(_FINISH_REASON_MAP.get(raw)):<10} {reported!r:<12} '
          f'{"yes" if reported == want else f"NO (want {want!r})"}')
```

</details>

Output on `main` (`251de3d`), xai-sdk 1.17.0:

```
_FINISH_REASON_MAP keys: ['cancelled', 'content_filter', 'failed', 'length', 'max_output_tokens', 'stop']

SDK Response.finish_reason map hit    reported     correct?
REASON_STOP            None       'stop'       yes
REASON_MAX_LEN         None       'stop'       NO (want 'length')
REASON_MAX_CONTEXT     None       'stop'       NO (want 'length')
REASON_TOOL_CALLS      None       'stop'       NO (want 'tool_call')
REASON_TIME_LIMIT      None       'stop'       NO (want 'error')
```

`map hit` is `None` in every row, including `REASON_STOP` — that one is only "correct" by coincidence of the default.

End-to-end with the existing mocks, same response streamed and not:

```
non-streaming finish_reason: 'length'
streaming     finish_reason: 'stop'
```

### Why the test suite doesn't catch it

`tests/models/mock_xai.py` only ever passes `'stop'`, and `create_response` additionally forces `REASON_STOP` onto the accumulated `Response` objects that represent intermediate chunks:

```python
finish_reason=finish_reason if finish_reason else 'stop',  # type: ignore[arg-type]
```

so the mocks can't express the wire behaviour the bug depends on — a chunk whose finish reason is `REASON_INVALID`. That `# type: ignore` is the tell: `''` was being coerced to `'stop'` to satisfy the signature. Fixing this is part of any fix.

### Proposed fix

Delete the dead string map, extend the proto map to every reason the enum can carry, and share one helper between the two paths:

```python
_FINISH_REASON_PROTO_MAP: dict[int, FinishReason] = {
    sample_pb2.FinishReason.REASON_MAX_LEN: 'length',
    sample_pb2.FinishReason.REASON_MAX_CONTEXT: 'length',
    sample_pb2.FinishReason.REASON_STOP: 'stop',
    sample_pb2.FinishReason.REASON_TOOL_CALLS: 'tool_call',
    sample_pb2.FinishReason.REASON_TIME_LIMIT: 'error',
}


def _map_finish_reason(response: chat_types.Response) -> FinishReason | None:
    """Map an xAI response's finish reason, or `None` if it has not finished yet."""
    outputs = response.proto.outputs
    if not outputs:  # pragma: no cover
        return None
    return _FINISH_REASON_PROTO_MAP.get(outputs[-1].finish_reason)
```

`_process_response` becomes `finish_reason = _map_finish_reason(response) or 'stop'`, and the streaming path only assigns when the response has actually finished:

```python
if (finish_reason := _map_finish_reason(response)) is not None:
    self.finish_reason = finish_reason
```

Reading the proto ints on both paths means there's one set of values to keep in step with the SDK, and it keeps the existing `_process_response` behaviour of preferring the *last* output (in multi-output responses an intermediate `TOOL_CALLS` output would otherwise win over the final `STOP`).

Two notes on scope, in case you'd rather it went differently:

- `REASON_MAX_CONTEXT` → `'length'`: the response stopped because the context was full, which is the same class of truncation as `REASON_MAX_LEN`. The alternative is `'error'`.
- `content_filter` has no proto counterpart — xAI's enum has no such member — so the old map's `'content_filter'` entry was unreachable twice over. I've dropped it rather than inventing a mapping; `capabilities/content_filter.py:34` tests for it, so if xAI does signal filtering some other way that's a separate issue.

I deliberately did **not** add `provider_details['finish_reason']` on the streaming path (which groq/anthropic do), because `_process_response` doesn't set it either — adding it on one side only would trade this asymmetry for a new one.

### Branch

Fix plus tests: https://github.com/LHMQ878/pydantic-ai/tree/fix/xai-stream-finish-reason

Three tests: a 4-way parametrized one asserting streaming and non-streaming agree (`stop`/`length`/`tool_call`/`error`), one asserting intermediate snapshots stay `None` until the final chunk, and one asserting every member of the SDK enum is mapped except `REASON_INVALID` — so this can't silently rot when xAI adds a reason. `tests/models/test_xai.py` goes 116 → 122 passed; 730 passed / 260 skipped across every module that touches xAI, unchanged from `main`; `models/xai.py` coverage unchanged from baseline; ruff and pyright clean.

Reverting only `models/xai.py` and keeping the tests fails exactly the new ones, with the wrong values named:

```
test_xai.py:1435: AssertionError: assert 'stop' == 'length'
test_xai.py:1435: AssertionError: assert 'stop' == 'tool_call'
test_xai.py:1435: AssertionError: assert 'stop' == 'error'
test_xai.py:1466: AssertionError: assert {'REASON_INVALID', 'REASON_MAX_CONTEXT', 'REASON_TOOL_CALLS', 'REASON_TIME_LIMIT'} == {'REASON_INVALID'}
5 failed, 117 passed
```

Happy to open a PR if you'd like it, or to leave it with you — per the contributing guide I'll wait to be assigned rather than opening one cold.

```
pydantic-ai main @ 251de3d
xai-sdk 1.17.0
Python 3.12, Windows
```

### LINKED ISSUE #6840: Streamed xAI responses always report finish_reason="stop" (map keyed on the wrong strings)
### Initial Checks

### Description

Every **streamed** xAI response reports `finish_reason='stop'`, regardless of what actually happened. A stream truncated at the token limit, one that ended in a tool call, and one that hit the server-side time limit are all indistinguishable from a clean finish.

`pydantic_ai_slim/pydantic_ai/models/xai.py` keeps two maps:

```python
_FINISH_REASON_MAP: dict[str, FinishReason] = {
    'stop': 'stop',
    'length': 'length',
    'content_filter': 'content_filter',
    'max_output_tokens': 'length',
    'cancelled': 'error',
    'failed': 'error',
}

# `GetChatCompletionResponse.outputs[*].finish_reason` uses the proto enum (ints), not the string values returned by
# `Response.finish_reason`.
_FINISH_REASON_PROTO_MAP: dict[int, FinishReason] = {
    sample_pb2.FinishReason.REASON_STOP: 'stop',
    sample_pb2.FinishReason.REASON_MAX_LEN: 'length',
    sample_pb2.FinishReason.REASON_TOOL_CALLS: 'tool_call',
}
```

The comment is right that `Response.finish_reason` is a string and the proto field is an int — but the string is not what `_FINISH_REASON_MAP` expects. In the xAI SDK:

```python
@property
def finish_reason(self) -> str:
    """Returns the finish reason of this response."""
    return sample_pb2.FinishReason.Name(self._get_output(sync=False).finish_reason)
```

It's `FinishReason.Name(...)`, so the values are `'REASON_STOP'`, `'REASON_MAX_LEN'`, `'REASON_MAX_CONTEXT'`, `'REASON_TOOL_CALLS'`, `'REASON_TIME_LIMIT'`, `'REASON_INVALID'`. **None of the six keys in `_FINISH_REASON_MAP` can ever match**, and `.get(..., 'stop')` swallows the miss:

```python
# XaiStreamedResponse._update_response_state
self.finish_reason = _FINISH_REASON_MAP.get(response.finish_reason, 'stop')
```

`_process_response` is unaffected — it reads the int off the proto and only touches the string map on a branch marked `# pragma: no cover`. So this is a streaming-only defect.

### Reproduction

```python
from xai_sdk import chat as chat_types
from xai_sdk.proto import chat_pb2, sample_pb2

from pydantic_ai.models.xai import _FINISH_REASON_MAP

for reason in sample_pb2.FinishReason.values():
    out = chat_pb2.CompletionOutput(index=0, finish_reason=reason)
    proto = chat_pb2.GetChatCompletionResponse(id='x', outputs=[out])
    proto.created.GetCurrentTime()
    sdk_str = chat_types.Response(proto, index=None).finish_reason
    print(f'{sdk_str:20} -> {_FINISH_REASON_MAP.get(sdk_str, "MISS (defaults to stop)")}')
```

```
REASON_INVALID       -> MISS (defaults to stop)
REASON_MAX_LEN       -> MISS (defaults to stop)
REASON_MAX_CONTEXT   -> MISS (defaults to stop)
REASON_STOP          -> MISS (defaults to stop)
REASON_TOOL_CALLS    -> MISS (defaults to stop)
REASON_TIME_LIMIT    -> MISS (defaults to stop)
```

### Impact

`finish_reason` is user-visible on `ModelResponse` and is exported as an OpenTelemetry attribute, so anything relying on it to detect truncation gets the wrong answer:

```python
async with agent.run_stream(prompt, model_settings=ModelSettings(max_tokens=50)) as result:
    ...
# response.finish_reason == 'stop' even though the model was cut off at 50 tokens
```

Two of the enum values also have no mapping at all, so they'd be wrong even with the string bug fixed: `REASON_MAX_CONTEXT` (context window exceeded — other providers map their equivalent to `'length'`, e.g. `anthropic.py`'s `'model_context_window_exceeded': 'length'`) and `REASON_TIME_LIMIT`.

Separately, `REASON_INVALID` is the proto default that every chunk *before* the last one carries, so a partial stream currently reports `'stop'` while still generating.

### Why the tests didn't catch it

`tests/models/mock_xai.py` builds mocks from a pydantic-ai `FinishReason` and translates it to the proto enum, which is correct — but every streaming test passes `'stop'`:

```python
get_grok_text_chunk('.', 'stop')          # test_xai_stream_text_finish_reason
get_grok_tool_chunk(None, '}', finish_reason='stop', ...)
```

Since the fallback is also `'stop'`, the assertions pass either way. No streaming test asserts a non-`stop` finish reason.

```Text
main @ 3d66418, xai-sdk 1.17.0 (pydantic-ai requires >=1.14.0)
```

LHMQ878:
Fixes #6840
Fixes #6812

Map xAI proto finish reasons for streamed and non-streamed responses. `REASON_INVALID` remains unset; length, context, tool-call, and time-limit endings map to Pydantic AI finish reasons.

Focused tests cover each mapped streaming result. The linked issues report the same defect.

### Checklist

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/xai.py
+++ b/pydantic_ai_slim/pydantic_ai/models/xai.py
@@ -154,9 +154,16 @@ _FINISH_REASON_PROTO_MAP: dict[int, FinishReason | None] = {
 }
 
 
-def _map_finish_reason(reason: int) -> FinishReason | None:
-    """Map an xAI proto finish reason to a pydantic-ai one, or `None` if it's unset/unrecognised."""
-    return _FINISH_REASON_PROTO_MAP.get(reason)
+def _map_finish_reason(response: chat_types.Response) -> FinishReason | None:
+    """Map an xAI response's finish reason to a pydantic-ai one, or `None` if it's unset/unrecognised.
+
+    `response.finish_reason` reflects a single output, and in multi-output responses (e.g. server-side tools) that
+    can be an intermediate `REASON_TOOL_CALLS` output rather than the final `REASON_STOP` one, so read the last
+    output instead. Streamed responses start out with a single empty output, whose `REASON_INVALID` keeps the
+    finish reason unset until the stream actually finishes.
+    """
+    outputs = response.proto.outputs
+    return _FINISH_REASON_PROTO_MAP.get(outputs[-1].finish_reason) if outputs else None
 
 
 class XaiModelSettings(ModelSettings, total=False):
@@ -867,12 +874,7 @@ class XaiModel(Model[AsyncClient]):
         # Convert usage with detailed token information
         usage = _extract_usage(response, self._model_name, self._provider.name, self._provider.base_url)
 
-        # Map finish reason.
-        #
-        # `response.finish_reason` reflects a single output, and in multi-output responses (e.g.
-        # server-side tools) that can be an intermediate TOOL_CALLS output rather than the final STOP
-        # one, so we derive the finish reason from the last output.
-        finish_reason = _map_finish_reason(outputs[-1].finish_reason) if outputs else None
+        finish_reason = _map_finish_reason(response)
 
         return ModelResponse(
             parts=parts,
@@ -959,12 +961,8 @@ class XaiStreamedResponse(StreamedResponse):
         if response.id and self.provider_response_id is None:
             self.provider_response_id = response.id
 
-        # Handle finish reason. Read it off the proto rather than `response.finish_reason`, which
-        # is the enum *name* of the same value, and take the last output for the same reason as
-        # `_process_response`. Chunks before the final one carry `REASON_INVALID`, which maps to
-        # `None` so the reason stays unset until the stream actually finishes.
-        if outputs := response.proto.outputs:
-            self.finish_reason = _map_finish_reason(outputs[-1].finish_reason)
+        # Handle finish reason
+        self.finish_reason = _map_finish_reason(response)
 
     def _collect_reasoning_events(
         self,
--- a/pydantic_ai_slim/pydantic_ai/models/xai.py
+++ b/pydantic_ai_slim/pydantic_ai/models/xai.py
@@ -140,22 +140,23 @@ def _map_reasoning_effort(thinking: ThinkingLevel, profile: GrokModelProfile) ->
         assert_never(thinking)
 
 
-_FINISH_REASON_MAP: dict[str, FinishReason] = {
-    'stop': 'stop',
-    'length': 'length',
-    'content_filter': 'content_filter',
-    'max_output_tokens': 'length',
-    'cancelled': 'error',
-    'failed': 'error',
-}
-
-# `GetChatCompletionResponse.outputs[*].finish_reason` uses the proto enum (ints), not the string values returned by
-# `Response.finish_reason`.
-_FINISH_REASON_PROTO_MAP: dict[int, FinishReason] = {
+# The xAI SDK reports finish reasons as the proto enum, both as ints on
+# `GetChatCompletionResponse.outputs[*].finish_reason` and as the enum *name* (`'REASON_STOP'`, …) from
+# `Response.finish_reason`, which is `sample_pb2.FinishReason.Name(...)` of the same value. `REASON_INVALID`
+# is the proto default for "unset" and maps to `None` so we don't claim the model stopped normally.
+_FINISH_REASON_PROTO_MAP: dict[int, FinishReason | None] = {
+    sample_pb2.FinishReason.REASON_INVALID: None,
     sample_pb2.FinishReason.REASON_STOP: 'stop',
     sample_pb2.FinishReason.REASON_MAX_LEN: 'length',
+    sample_pb2.FinishReason.REASON_MAX_CONTEXT: 'length',
     sample_pb2.FinishReason.REASON_TOOL_CALLS: 'tool_call',
+    sample_pb2.FinishReason.REASON_TIME_LIMIT: 'error',
 }
+
+
+def _map_finish_reason(reason: int) -> FinishReason | None:
+    """Map an xAI proto finish reason to a pydantic-ai one, or `None` if it's unset/unrecognised."""
+    return _FINISH_REASON_PROTO_MAP.get(reason)
 
 
 class XaiModelSettings(ModelSettings, total=False):
@@ -868,15 +869,10 @@ class XaiModel(Model[AsyncClient]):
 
         # Map finish reason.
         #
-        # The xAI SDK exposes `response.finish_reason` as a *string* for the overall response, but in
-        # multi-output responses (e.g. server-side tools) it can reflect an intermediate TOOL_CALLS
-        # output rather than the final STOP output. We derive the finish reason from the final output
-        # when available.
-        if outputs:
-            last_reason = outputs[-1].finish_reason
-            finish_reason = _FINISH_REASON_PROTO_MAP.get(last_reason, 'stop')
-        else:  # pragma: no cover
-            finish_reason = _FINISH_REASON_MAP.get(response.finish_reason, 'stop')
+        # `response.finish_reason` reflects a single output, and in multi-output responses (e.g.
+        # server-side tools) that can be an intermediate TOOL_CALLS output rather than the final STOP
+        # one, so we derive the finish reason from the last output.
+        finish_reason = _map_finish_reason(outputs[-1].finish_reason) if outputs else None
 
         return ModelResponse(
             parts=parts,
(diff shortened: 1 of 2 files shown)


adtyavrdhn:
Superseded by #6814

PR #6857, opened 2026-07-29
Correct provider support documentation for `ModelSettings`

### LINKED ISSUE #6843: `ModelSettings.timeout` and `ModelSettings.extra_headers` document xAI support but are silently dropped
### Initial Checks

### Description

`ModelSettings.timeout` and `ModelSettings.extra_headers` both list xAI under `Supported by:`, but `models/xai.py` never reads either one. Both are silently dropped.

The `Supported by:` lists are the only place a user can learn whether a shared setting does anything on their provider, so a wrong entry costs real debugging time: the setting is accepted, typechecks, and no-ops.

`XaiModel` forwards shared settings through a single declarative mapping:

```python
_XAI_MODEL_SETTINGS_MAPPING: dict[str, str] = {
    'temperature': 'temperature',
    'top_p': 'top_p',
    'max_tokens': 'max_tokens',
    'stop_sequences': 'stop',
    'parallel_tool_calls': 'parallel_tool_calls',
    'presence_penalty': 'presence_penalty',
    'frequency_penalty': 'frequency_penalty',
    'seed': 'seed',
    'xai_logprobs': 'logprobs',
    ...
}


def _map_model_settings(model_settings: XaiModelSettings) -> dict[str, Any]:
    return {
        _XAI_MODEL_SETTINGS_MAPPING[key]: value
        for key, value in model_settings.items()
        if key in _XAI_MODEL_SETTINGS_MAPPING
    }
```

Neither `timeout` nor `extra_headers` is a key, and neither appears anywhere else in the module — `grep -n 'timeout\|Timeout' models/xai.py` returns nothing, and the one `extra_headers` hit is an unrelated kwarg on the `mcp()` builtin-tool helper.

### Reproduction

```python
from pydantic_ai.models.xai import _map_model_settings

for key in ('temperature', 'timeout', 'extra_headers'):
    print(f'{key:15} -> {_map_model_settings({key: "X"})}')
```

```
temperature     -> {'temperature': 'X'}
timeout         -> {}
extra_headers   -> {}
```

### Why this is a docs fix rather than a missing feature

Neither setting is implementable per request with the current SDK. `xai_sdk.AsyncClient` bakes both into the gRPC channel at construction:

```python
def _init(self, api_key, management_api_key, api_host, management_api_host, metadata, channel_options, timeout, use_insecure_channel):
    self._api_channel = self._make_grpc_channel(api_key, api_host, metadata, channel_options, timeout, use_insecure_channel)
```

and the per-request surface has no hook for either — `Chat.sample()` and `Chat.stream()` take no arguments, and `Client.create()` accepts neither. (`Chat.defer()` has a `timeout`, but that's the deferred-completion poll deadline, not a request timeout.)

`XaiProvider` already exposes the client-level equivalent for the timeout:

```python
provider = XaiProvider(api_key='...', timeout=30)
```

and `metadata` is reachable by constructing your own `AsyncClient` and passing it as `xai_client=`. So the correct outcome is to drop xAI from both lists and say where the equivalent lives.

The provider docstring and `docs/models/xai.md` compound the problem — both currently tell users the provider-level `timeout` "is distinct from `ModelSettings.timeout`, which overrides the timeout for an individual request", pointing at the setting that does nothing.

### Scope

I checked every `Supported by:` list in `settings.py` against the implementations: `max_tokens`, `temperature`, `top_p`, `top_k`, `parallel_tool_calls`, `tool_choice`, `seed`, `presence_penalty`, `frequency_penalty`, `logit_bias`, `stop_sequences`, `thinking`, `service_tier` and the non-xAI entries of `timeout`/`extra_headers` all check out. These two xAI entries are the only wrong ones.

```Text
main @ 3d66418, xai-sdk 1.17.0 (pydantic-ai requires >=1.14.0)
```

### LINKED ISSUE #6856: Twelve of `ModelSettings`' `Supported by:` lists disagree with the code, in both directions
### Initial Checks

### Description

Each field of `ModelSettings` carries a `Supported by:` list. That list is the only place a user can find out whether a general setting actually reaches a given provider — and a setting that doesn't reach one is *silently dropped*, not rejected, so a stale list is indistinguishable from a broken provider.

Twelve of the sixteen lists have drifted from the code.

#### Read by a provider the list doesn't name

`models/huggingface.py` forwards nine general settings and is named by **none** of them. Verified against a fake client that records the outgoing kwargs:

```
  max_tokens -> 7
  temperature -> 0.1
  top_p -> 0.2
  seed -> 42
  presence_penalty -> 0.3
  frequency_penalty -> 0.4
  logit_bias -> {'1': 5}
  stop -> ['X']
  extra_body -> {'k': 'v'}
```

`models/mistral.py:304` and `:355` both pass `parallel_tool_calls=model_settings.get('parallel_tool_calls')`, and `tests/models/test_mistral.py::test_parallel_tool_calls_sent` already asserts it reaches the SDK — but the field's list is `OpenAI`, `Groq`, `Anthropic`, `xAI`.

`extra_body` is read by `cerebras.py:151`, `huggingface.py:288`, `openrouter.py:670` and `zai.py:162`; the list names only `OpenAI`, `Anthropic`, `Groq`. On the three that merge (Cerebras, OpenRouter, Z.AI) the model's own derived keys win over the user's, which is worth documenting too.

`temperature` and `stop_sequences` are read by `mcp_sampling.py:63`/`:65`; only `max_tokens` names MCP Sampling.

`thinking` reaches Z.AI via `_translate_thinking` → `extra_body['thinking']`, which the list doesn't mention.

`tool_choice` reaches Cohere via `resolve_tool_choice` (`cohere.py:224`), also unlisted.

#### Named by a list but not read at all

`xai.py` forwards settings through a fixed mapping table:

```python
_XAI_MODEL_SETTINGS_MAPPING: dict[str, str] = {
    'temperature': 'temperature',
    'top_p': 'top_p',
    'max_tokens': 'max_tokens',
    'stop_sequences': 'stop',
    'parallel_tool_calls': 'parallel_tool_calls',
    ...
}
```

`timeout` and `extra_headers` are not in it, and `chat.create` accepts neither parameter — but both fields list xAI as supported:

```
mapped -> {'temperature': 0.5}          # from XaiModelSettings(timeout=5, extra_headers={'X-A': '1'}, temperature=0.5)
'timeout' in mapping: False
'extra_headers' in mapping: False
```

`XaiProvider.__init__` compounds this: its docstring says the client-level timeout "is distinct from `ModelSettings.timeout`, which overrides the timeout for an individual request" — there is no such override on this provider.

### Suggested fix

Correct the twelve lists, and add a test that derives each list from the source and compares it against the modules that actually read the field, so the next provider to grow or drop a setting can't leave the docs behind. `tool_choice` and `thinking` can't be machine-checked this way — every model receives them through `resolve_tool_choice` / `ModelRequestParameters.thinking` whether it honours them or not — so those two stay excluded and hand-maintained.

### Example Code

```python
from pydantic_ai.models.xai import XaiModelSettings, _map_model_settings

# `timeout` and `extra_headers` both document xAI support; neither survives the mapping
print(_map_model_settings(XaiModelSettings(timeout=5, extra_headers={'X-A': '1'}, temperature=0.5)))
#> {'temperature': 0.5}
```

```Text
pydantic-ai-slim 2.20.0 (main @ 3d66418)
Python 3.12
```

LHMQ878:
Fixes #6856
Fixes #6843

Correct the `ModelSettings` provider support lists and document that xAI timeout and metadata are configured on the SDK client, not per request.

Documentation only. No runtime behavior changes.

### Checklist


LHMQ878:
The `coverage` failure was mine, not a flake: `_supported_by`'s `return None` and the matching branch in `_documented_support` were unreachable, because every one of the 16 `ModelSettings` fields does have a `Supported by:` list. The 100% gate was right to catch it.

Rather than delete the defensive path I turned it into the assertion it was implying: `_documented_support` now collects every annotated field, and the test rejects an empty list — a general setting whose provider support is documented nowhere is the same gap this test exists to catch. `tests/test_settings.py` is now 100.00% with 0 missed statements and 0 partial branches locally (40 passed, 2 skipped).

PR #6862, opened 2026-07-29
docs: show approval-bound deferred tool calls

jw-ond:
## Summary

Adds a Deferred Tools documentation section showing how to bind a human approval decision to the exact tool name and arguments that were reviewed.

The example uses existing Pydantic AI concepts only:

- DeferredToolRequests approvals
- DeferredToolResults metadata
- RunContext.tool_call_metadata
- ToolCallPart.args_as_dict

The goal is to make the approval result valid only for the specific tool call that produced the fingerprint, instead of treating approval as a reusable session flag if the action changes before execution.

## Test plan

- git diff --check
- confirmed the new docs section is present with a lightweight Python file check

I could not run the full mkdocs build locally because uv is not installed in this shell.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6862?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/deferred-tools.md
+++ b/docs/deferred-tools.md
@@ -310,6 +310,81 @@ _(This example is complete, it can be run "as is")_
 !!! note "Tool result ordering"
     Tool results follow the order in which the model emitted the corresponding tool calls. In the message history above, `delete_file`'s denied result appears before `update_file`'s result for `.env` because the model emitted `delete_file` first. This is an intentional behavior change in v2: results are no longer grouped by tool kind, so the ordering you see reflects the model's emission order.
 
+### Binding approval to tool-call arguments
+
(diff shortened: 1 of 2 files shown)


jw-ond:
Following up on this one. The PR is approved and the current check set is green from my side. If there is any final wording or placement change needed before merge, I can adjust it quickly.

Issue #3994, opened 2026-01-13
OpenRouterModel discards valid responses when metadata fields are null


ZacharyHampton:
## Bug Description

`OpenRouterModel._validate_completion` discards valid LLM responses when OpenRouter returns non-standard response structures.

## Error

```
pydantic_ai.exceptions.UnexpectedModelBehavior: Invalid response from openrouter chat completions endpoint: 5 validation errors for _OpenRouterChatCompletion
id: Input should be a valid string [input_value=None]
choices: Input should be a valid list [input_value=None]
model: Input should be a valid string [input_value=None]
object: Input should be 'chat.completion' [input_value=None]
provider: Field required [type=missing]
```

## Root Cause

OpenRouter intermittently returns two non-standard response formats:

### 1. Nested Response in `provider` Field
```json
{
    "id": null, "choices": null, "model": null, "object": null,
    "provider": {
        "id": "gen-123",
        "choices": [{"message": {"content": "..."}, "finish_reason": "stop"}],
        "model": "google/gemini-3-flash-preview",
        "provider": "Google"
    }
}
```

### 2. Error Response with Null Fields
```json
{
    "id": null, "choices": null,
    "error": {"message": "...", "code": 400},
    "user_id": "org_xxx"
}
```

The OpenAI SDK wraps these as `ChatCompletion` objects (via `model_construct`), preserving extra fields but with null standard fields. pydantic-ai's strict validation then fails.

## Suggested Fix

In `_validate_completion`, before validation:

1. **Extract nested responses** from `provider` field if it contains `choices`
2. **Handle error responses** by checking for `error` field
3. **Provide defaults** for null metadata fields

```python
def _validate_completion(self, response):
    response_dict = response.model_dump()
    
    # Extract nested response from provider field
    provider_data = response_dict.get("provider")
    if isinstance(provider_data, dict) and provider_data.get("choices"):
        nested_provider = provider_data.pop("provider", "unknown")
        response_dict.update(provider_data)
        response_dict["provider"] = nested_provider
    
    # Handle error responses
    if error := response_dict.get("error"):
        raise ModelHTTPError(status_code=error.get("code", 500), ...)
    
    # Provide defaults for null metadata
    response_dict.setdefault("id", f"gen-{uuid.uuid4().hex[:12]}")
    response_dict.setdefault("provider", "unknown")
    # ... etc
    
    return _OpenRouterChatCompletion.model_validate(response_dict)
```

## Environment

[...]

PR #6885, opened 2026-07-29
Raise ModelAPIError for OpenRouter responses with null choices and no error envelope

### LINKED ISSUE #6900: OpenRouter: null `choices` with no error envelope raises a bare ValidationError instead of ModelAPIError
### Description

OpenRouter intermittently returns a response body with **null `choices` and no error envelope at all** — a transient downstream-provider hiccup that carries no completion. `OpenRouterModel._validate_completion` tries the error-envelope shape and the nested-provider shape (both added for #3994), then re-raises the original pydantic `ValidationError`.

That makes a transient indistinguishable from a malformed contract for anything that classifies on exception type:

- `FallbackModel` defaults to `fallback_on=(ModelAPIError,)`, so it does **not** fall back — a bare `ValidationError` isn't in that family.
- Any retry layer keyed on `ModelAPIError` / `ModelHTTPError` treats it as fatal and fails the run fast.

We hit this in production regularly enough that it's the last surviving guard of what used to be a ~574-line local OpenRouter patch module (v2 absorbed everything else — thank you). #3994 covered the sibling shapes that *do* carry an `error` key; this is the shape that carries nothing.

### Example Code

```python
from openai.types.chat import ChatCompletion

model = OpenRouterModel('openai/gpt-4.1-mini', provider=OpenRouterProvider(api_key='test-key'))

# The real shape, minus the `error` key: no completion, no envelope.
completion = ChatCompletion.model_construct(
    id=None,
    choices=None,
    model=None,
    object=None,
    provider=None,
    created=1234567890,
    usage=None,
)

model._process_response(completion)
# pydantic_core._pydantic_core.ValidationError: 5 validation errors for ChatCompletion
#   id      Input should be a valid string  [input_value=None]
#   choices Input should be a valid list    [input_value=None]
#   model   Input should be a valid string  [input_value=None]
#   object  Input should be 'chat.completion' [input_value=None]
#   ...
#
# Expected: an error in the ModelAPIError family, so retry/fallback can act on it.
```

### Expected vs actual

- **Actual:** bare `ValidationError` (surfacing as `UnexpectedModelBehavior`), not retryable, no fallback.
- **Expected:** `ModelAPIError` — the body is a provider hiccup, not a contract break.

A malformed *nested-provider* body (`choices=None`, no `error`, but `provider={'some_key': 'some_value'}`) should stay fatal — that one tried to be a structured response and failed, so retrying it forever would mask a real break.

```
pydantic-ai-slim 2.18.0 (and reproduces on main)
python 3.12.11
openai (via pydantic-ai-slim[openai])
```

---

Note for triage: @DouweM has already reviewed the fix for this in #6885 (which was auto-closed for lacking a linked issue — this is that issue). His review shaped the current approach: `ModelAPIError` rather than a faked HTTP 502, and modelling the shape as a `_OpenRouterNoCompletionResponse` sibling of `_OpenRouterErrorResponse` rather than inspecting the response dict.

ZacharyHampton:
Fixes #6900

OpenRouter intermittently returns a body with **null `choices` and no error envelope at all** — a transient provider hiccup carrying no completion. `_validate_completion` tries the error-envelope and nested-provider shapes (#3994), then re-raises the bare `ValidationError`, so consumers that classify on exception type treat a transient as fatal. Nothing retries it and `FallbackModel` doesn't fall back on it.

This models that shape and raises `ModelAPIError` for it.

### The shape

`_OpenRouterNoCompletionResponse`, a sibling of `_OpenRouterErrorResponse` — the same family of bodies, minus the error envelope:

```python
class _OpenRouterNoCompletionResponse(BaseModel, extra='allow'):
    choices: None
    error: None = None
    provider: str | None = None
    model: str | None = None
```

Tried in the existing `except ValidationError` chain, after the error-envelope and nested-provider attempts. Deliberate details:

- **`provider: str | None`** — a body whose `provider` is a *dict* is a malformed nested-provider response, not this shape, so it fails validation here and stays fatal. Declarative, no `isinstance` check.
- **`choices: None` is required, not defaulted** — a body that merely omits `choices` doesn't silently match.
- **`_OpenRouterChatCompletion` is untouched**: `id`, `model`, `object`, `provider` all stay required and non-null, so the happy path loses no guarantees and a `ValidationError` from it still means "a shape we don't account for."

### The exception

`ModelAPIError`, not a faked HTTP status. It's `FallbackModel`'s default `fallback_on`, so the transient now reaches retry/fallback machinery that a bare `ValidationError` never did.

### Tests

Both boundaries pinned, and all existing quirky-shape tests unchanged:

- `test_openrouter_null_choices_without_error_envelope_is_retryable` — the envelope-less body raises `ModelAPIError`, and asserts it is *not* a `ModelHTTPError` so nobody re-adds a fake status.
- `test_openrouter_null_choices_with_malformed_provider_dict_stays_fatal` — the malformed-`provider` body still raises `UnexpectedModelBehavior`.

Full `tests/models/test_openrouter.py`: 58 passed.

---

We've run the workaround for this as a monkey-patch in production — it's the last survivor of a once-574-line OpenRouter patch module, since v2 absorbed the rest. Discussed with @DouweM on a call; the approach here follows his review (`ModelAPIError` over a faked 502, and modelling the shape rather than inspecting the response dict).

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/openrouter.py
+++ b/pydantic_ai_slim/pydantic_ai/models/openrouter.py
@@ -8,7 +8,7 @@ from pydantic import BaseModel, Discriminator, ValidationError, field_validator
 from typing_extensions import TypedDict, assert_never, override
 
 from .. import usage
-from ..exceptions import ModelHTTPError, UserError
+from ..exceptions import ModelAPIError, ModelHTTPError, UserError
 from ..messages import (
     BinaryContent,
     CachePoint,
@@ -1030,15 +1030,15 @@ class OpenRouterModel(OpenAIChatModel):
                     and not isinstance(response_dict.get('provider'), dict)
                 ):
                     # A null-`choices` body with no error envelope and no nested-provider
-                    # payload is a transient provider hiccup, not a malformed request:
-                    # surface it as a retryable upstream error rather than a bare
-                    # ValidationError. A `provider` dict that failed to validate above is
-                    # excluded on purpose — that body tried to be a nested-provider
-                    # response and was malformed, which must stay fatal.
-                    raise ModelHTTPError(
-                        status_code=502,
+                    # payload is a transient provider hiccup, not a malformed request.
+                    # `ModelAPIError` is `FallbackModel`'s default `fallback_on`, so this
+                    # reaches retry/fallback machinery that a bare `ValidationError` never
+                    # did. A `provider` dict that failed to validate above is excluded on
+                    # purpose — that body tried to be a nested-provider response and was
+                    # malformed, which must stay fatal.
+                    raise ModelAPIError(
                         model_name=response_dict.get('model') or self.model_name,
-                        body='OpenRouter returned a response with null `choices` and no error envelope',
+                        message='OpenRouter returned a response with null `choices` and no error envelope',
                     ) from exc
                 raise exc
 
--- a/pydantic_ai_slim/pydantic_ai/models/openrouter.py
+++ b/pydantic_ai_slim/pydantic_ai/models/openrouter.py
@@ -1024,10 +1024,17 @@ class OpenRouterModel(OpenAIChatModel):
             try:
                 nested = _OpenRouterNestedProviderResponse.model_validate(response_dict)
             except ValidationError:
-                if response_dict.get('choices') is None and response_dict.get('error') is None:
-                    # A null-`choices` body with no error envelope at all is a transient
-                    # provider hiccup, not a malformed request: surface it as a retryable
-                    # upstream error rather than a bare ValidationError.
+                if (
+                    response_dict.get('choices') is None
+                    and response_dict.get('error') is None
+                    and not isinstance(response_dict.get('provider'), dict)
+                ):
+                    # A null-`choices` body with no error envelope and no nested-provider
+                    # payload is a transient provider hiccup, not a malformed request:
+                    # surface it as a retryable upstream error rather than a bare
+                    # ValidationError. A `provider` dict that failed to validate above is
+                    # excluded on purpose — that body tried to be a nested-provider
+                    # response and was malformed, which must stay fatal.
                     raise ModelHTTPError(
                         status_code=502,
                         model_name=response_dict.get('model') or self.model_name,
--- a/pydantic_ai_slim/pydantic_ai/models/openrouter.py
+++ b/pydantic_ai_slim/pydantic_ai/models/openrouter.py
@@ -1024,6 +1024,15 @@ class OpenRouterModel(OpenAIChatModel):
             try:
                 nested = _OpenRouterNestedProviderResponse.model_validate(response_dict)
             except ValidationError:
+                if response_dict.get('choices') is None and response_dict.get('error') is None:
+                    # A null-`choices` body with no error envelope at all is a transient
+                    # provider hiccup, not a malformed request: surface it as a retryable
+                    # upstream error rather than a bare ValidationError.
+                    raise ModelHTTPError(
+                        status_code=502,
+                        model_name=response_dict.get('model') or self.model_name,
+                        body='OpenRouter returned a response with null `choices` and no error envelope',
+                    ) from exc
                 raise exc
 
             validated = nested.provider
--- a/pydantic_ai_slim/pydantic_ai/models/openrouter.py
+++ b/pydantic_ai_slim/pydantic_ai/models/openrouter.py
@@ -588,6 +588,25 @@ class _OpenRouterErrorResponse(BaseModel, extra='allow'):
     model: str | None = None
 
 
+class _OpenRouterNoCompletionResponse(BaseModel, extra='allow'):
+    """OpenRouter response that carries no completion at all: `choices` is null and there is no error envelope.
+
+    A sibling of `_OpenRouterErrorResponse` — the same family of bodies OpenRouter
+    returns when a downstream provider hiccups, minus the error envelope. Modelling it
+    keeps a `ValidationError` from `_OpenRouterChatCompletion` meaning "a shape we do not
+    account for" rather than doubling as a transient signal.
+
+    `provider` is typed `str | None` on purpose: a body whose `provider` is a *dict* is a
+    malformed nested-provider response, not this shape, so it fails validation here and
+    stays fatal.
+    """
+
+    choices: None
+    error: None = None
+    provider: str | None = None
+    model: str | None = None
+
+
 class _OpenRouterNestedCompletion(_OpenRouterChatCompletion):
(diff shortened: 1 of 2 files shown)


pydantic_ai_slim/pydantic_ai/models/openrouter.py
```diff
@@ -1024,6 +1024,15 @@ def _validate_completion(self, response: chat.ChatCompletion) -> _OpenRouterChat
...
                 nested = _OpenRouterNestedProviderResponse.model_validate(response_dict)
             except ValidationError:
+                if response_dict.get('choices') is None and response_dict.get('error') is None:
+                    # A null-`choices` body with no error envelope at all is a transient
+                    # provider hiccup, not a malformed request: surface it as a retryable
+                    # upstream error rather than a bare ValidationError.
+                    raise ModelHTTPError(
+                        status_code=502,
+                        model_name=response_dict.get('model') or self.model_name,
+                        body='OpenRouter returned a response with null `choices` and no error envelope',
```
DouweM: - `ModelAPIError` would be more appropriate than faking a HTTP 502.
- `_OpenRouterNestedProviderResponse` is meant to represent the actual shapes that the API returns -- a validation error should mean that it has some value that we should be accounting for but aren't. So I suggest updating the model to support this shape, and then to verify that `choices` is not `None` in the non-exception path, just like we already validate that `choices` is non-null.

ZacharyHampton:
CI caught a real regression in the first pass — my fault for running only the null/malformed/metadata subset locally instead of the whole file.

`test_openrouter_provider_dict_without_choices_raises` has `choices=None` and no `error`, but **does** carry `provider={'some_key': 'some_value'}` — a body that tried to be a nested-provider response and failed to validate. My original condition swept that into the retryable 502, which would mask a real contract break behind infinite retries.

Fixed in 5adbae3 by excluding it: the 502 now requires `choices is None` **and** no `error` **and** no `provider` dict. Only the genuinely envelope-less, provider-less shape (the transient we actually see in production) becomes retryable; everything that looks like a malformed structured body stays fatal.

Added `test_openrouter_null_choices_with_malformed_provider_dict_stays_fatal` to pin the distinction so the boundary can't drift back. Full `tests/models/test_openrouter.py` now passes locally — 58 passed.

ZacharyHampton:
Both fair. First one done in dd7e490 — `ModelAPIError` instead of the faked 502, and it's strictly better than what I proposed: it's `FallbackModel`'s default `fallback_on`, so the transient now reaches retry/fallback machinery that a bare `ValidationError` never did. (On our side it means adapting our own retry classification off "`ModelHTTPError` with a 5xx" — which was the wrong shape to key on anyway.) Test asserts `ModelAPIError` and explicitly *not* `ModelHTTPError`, so nobody re-adds a fake status. 58 passed locally.

On the second, I want to check how far you want the models opened up before I touch them, because the shape is broader than just `choices`.

The real transient body we see in production (same family as #3994, minus the error envelope):

```python
{'id': None, 'choices': None, 'created': 1768361801, 'model': None,
 'object': None, 'service_tier': None, 'system_fingerprint': None, 'usage': None}
```

Against `_OpenRouterChatCompletion` that's **five** fields that fail, not one: `id: str`, `choices: list[...]`, `model: str`, `object: Literal['chat.completion']` all arrive `None`, and `provider: str` is absent entirely. So "update the model to support this shape, then verify `choices` is not `None` in the non-exception path" means making `id`, `model`, `object` and `provider` all optional on the model that also validates every **healthy** response — which weakens the contract on the happy path in exchange for representing a body that carries no completion at all.

Two ways to read your suggestion, happy to take either:

1. **Represent it on the shared model** — relax those five on `_OpenRouterChatCompletion`, then check `choices is None` (and presumably `provider`) in the non-exception path alongside the existing `error` check. Honest about the API, but the happy path loses those guarantees.
2. **A dedicated model for the no-completion shape** — e.g. `_OpenRouterEmptyResponse` with everything optional, tried in the `except ValidationError` chain right after `_OpenRouterErrorResponse` (which is what it's a sibling of — same body, error envelope absent). Keeps `_OpenRouterChatCompletion` strict, and the "shape we didn't account for" becomes an explicit model rather than a dict inspection.

I lean 2 — it's the same structure you already use for the error and nested-provider variants, and it keeps a `ValidationError` from `_OpenRouterChatCompletion` meaning what it means today. But it's your model surface, so say which and I'll push it.

Worth noting the `provider`-dict case that CI caught either way: `{'choices': None, 'provider': {'some_key': 'some_value'}}` must stay fatal, since that body *tried* to be a nested-provider response and was malformed. Under (2) that falls out naturally — it fails `_OpenRouterEmptyResponse` too and re-raises. Under (1) it needs the explicit `provider` check I have now.

ZacharyHampton:
Went ahead and pushed option 2 (e4bdd79) rather than leave the design comment open — happy to switch to option 1 if you'd rather relax the shared model, but this way there's code to react to.

`_OpenRouterNoCompletionResponse`, a sibling of `_OpenRouterErrorResponse`:

```python
class _OpenRouterNoCompletionResponse(BaseModel, extra='allow'):
    choices: None
    error: None = None
    provider: str | None = None
    model: str | None = None
```

Tried in the same `except ValidationError` chain, after the error-envelope and nested-provider attempts. Three things fall out of it:

- **The shape is modelled, not sniffed.** My dict inspection (`response_dict.get('choices') is None and ... not isinstance(..., dict)`) is gone. This is your point 2: a `ValidationError` from `_OpenRouterChatCompletion` goes back to meaning "a shape we don't account for," and the body that carries no completion is an accounted-for shape with a name.
- **`provider: str | None` does the malformed-nested exclusion declaratively.** A body whose `provider` is a *dict* fails validation here, so it re-raises and stays fatal — no explicit check needed. That's what CI caught in my first pass, now handled by the type rather than an `isinstance`.
- **`choices: None` is required, not defaulted**, so a body that simply omits `choices` doesn't silently match either.

`_OpenRouterChatCompletion` keeps every guarantee it has today on the happy path — `id`, `model`, `object`, `provider` all stay required and non-null.

Full `tests/models/test_openrouter.py` green (58 passed), including the two boundary tests: the envelope-less body raises `ModelAPIError`, and the malformed-`provider` body still raises `UnexpectedModelBehavior`.

ZacharyHampton:
The linked-issue bot auto-closed this when I edited the description — the rewrite dropped the issue reference, my mistake. I couldn't reopen from a fork, so:

- Issue filed with a reproducible example: #6900
- Same branch and commits continue in **#6901**

Your review is what shaped the final shape (`ModelAPIError` over the faked 502, and modelling the body as a `_OpenRouterNoCompletionResponse` sibling of `_OpenRouterErrorResponse` instead of inspecting the response dict) — both are in #6901, and I've referenced this thread there so the discussion stays findable. Sorry for the churn.

Issue #6883, opened 2026-07-29
Temporal cache-eviction of a workflow running an agent livelocks the event loop (anyio re-arms cancellation delivery for already-`_must_cancel` tasks)


ZacharyHampton:
### Summary

Cancelling an in-flight `agent.run()` executing in a Temporal **workflow body** (the supported `TemporalDurability` pattern) deterministically livelocks the workflow event loop. The loop spins at ~17k iterations/sec inside a single `_run_once`, the workflow task hits the deadlock detector (TMPRL1101), the task fails, Temporal retries the identical activation, and the run makes zero progress — we've measured single runs burning hours. A wedged workflow-task thread also blocks graceful worker shutdown indefinitely.

Reproduces on **2.18.0, 2.19.0 and 2.20.0** with `temporalio==1.30.0` (matrix below). #6496's `raise_if_cancelling()` backstop does not reach it — the spin never arrives at a run-step boundary.

### Mechanism

- pydantic-graph runs nodes under **anyio cancel scopes**; anyio delivers cancellation *level-triggered* (`_backends/_asyncio.py` — `should_retry` per task, re-armed via `call_soon`, deduped only on `task._must_cancel`).
- temporalio's `run_activity` (`_workflow_instance.py`) swallows the `CancelledError`, emits `request_cancel_activity`, calls `t.uncancel()` and re-shields — clearing `_must_cancel` and leaving an undone `_fut_waiter`, so anyio's dedup guard never trips and it re-arms forever.
- The activity resolution that would end the wait can only arrive in a **later activation**, which can never start because the current one never yields.

temporalio's `uncancel()` was added (their #1523) to fix a cancellation-*counter* livelock; it doesn't consider anyio's cancel scopes. Both behaviors are individually correct and jointly pathological.

### Two ways the cancellation arrives (both measured in production)

**A. Application-issued** — e.g. `asyncio.wait_for(agent.run(...), timeout=...)` in the workflow body. cProfile of a real wedged activation: `_deliver_cancellation` 140,541 calls / `Task.cancel`+`run_activity` 140,538 / `shield` 140,536 — all 1:1, one activity handle spun 140k times, 8.30 s in one `_run_once`. We've since removed every workflow-body cancel (cooperative `UsageLimits(request_limit=...)` + child-workflow `execution_timeout` instead), so this half is avoidable by convention — but nothing warns you, and `timeout=` is the first thing anyone reaches for.

[...]

PR #6892, opened 2026-07-29
Fix Temporal workflow livelock when anyio scope cancellation hits an in-flight activity await

DouweM:
Fixes the application-issued trigger (A) of #6883 — independently validated by the reporter on his unmodified repro. The issue stays open until his staging run delivers a verdict on the cache-eviction trigger (B), whose zero-delivery spin our analysis places downstream of the same absorption this PR removes.

## Summary

A user reported (Critical) a deterministic livelock: inside a workflow body, `asyncio.wait_for(agent.run(...), timeout=...)` cancelling an in-flight run wedges the workflow task at ~17k spins/sec until Temporal's deadlock detector fails it — and the retried activation fails identically, forever.

**Mechanism (all three parts verified empirically):**
- anyio delivers scope cancellation **level-triggered**: `CancelScope._deliver_cancellation` re-arms itself via `call_soon` for as long as a task remains in a cancelled scope, deduplicating only on `task._must_cancel` or a done `_fut_waiter`.
- temporalio's `run_activity` is a `while True` that swallows `CancelledError`, emits `request_cancel_activity`, and re-parks on a shielded result future. `task.cancel()` on a task awaiting an activity handle **delegates** to the handle task, so the scoped task stays parked, uncancelled, with an undone waiter.
- anyio re-cancels every event-loop iteration (appending another cancel command each time); the activity's resolution can only arrive in a **later activation**, which can never start because the loop never goes idle: `[TMPRL1101] Potential deadlock detected`.

Pydantic AI puts user code in this position: the graph engine runs steps as anyio task-group children, and every Temporal activity await (model request, tools, MCP, event dispatch) happens beneath them.

## The fix

A cancellation-safe activity executor (`durable_exec/temporal/_activity_execution.py`) now wraps **all 15** `workflow.execute_activity` call sites: it shields the activity handle so the cancellation is delivered to *our* task, delegates exactly one cancel to temporalio's graceful machinery (honoring the configured `ActivityCancellationType`), waits for the resolution inside an `anyio.CancelScope(shield=True)` so re-delivery stops and the activation can complete, then re-raises the original `CancelledError`.

Re-raising `CancelledError` (instead of letting the activity's `ActivityError` leak) also fixes a second bug found while reproducing: `asyncio.wait_for(agent.run(...), t)` in a workflow used to **fail the workflow** with `ActivityError: Activity cancelled` instead of raising `TimeoutError` — the `except asyncio.TimeoutError` in workflow code never matched. Both are pinned by tests:

- `test_anyio_scope_cancel_of_activity_await_does_not_wedge` — the primitive wedge shape through our executor (unit-style by necessity: the exact timing can't be forced through the agent API; before the fix this exact workflow shape produced `[TMPRL1101]` task failures, verified live).
- `test_wait_for_agent_timeout_in_workflow_does_not_livelock` — public-API: `wait_for` over a run with a slow streaming activity returns `'timed out cleanly'`.

External cancellation semantics are unchanged: cancelling the Temporal workflow still delivers `CancelledError` and ends the workflow as *Cancelled* (`test_temporal_cancellation_backstop_survives_absorbed_activity_cancel` still passes).

Note: the full local `test_temporal.py` run flakes on the shared dev server dying mid-suite; the same batch fails identically on unmodified `main`, so that instability is pre-existing and unrelated (focused runs pass consistently).

## #6883's two triggers

- **Trigger A (application-issued, e.g. `asyncio.wait_for`)**: fixed and pinned — including the issue's exact MRE shape (`test_wait_for_nonstreaming_agent_timeout_does_not_livelock`) alongside the streaming and primitive variants.
- **Trigger B (SDK cache-eviction cancel)**: the zero-delivery spin (`_must_cancel` set on every task while anyio re-arms) is downstream of tasks *absorbing* cancellation instead of unwinding. With this fix, an eviction's direct `task.cancel()` is not blocked by the executor's shield (`anyio.CancelScope(shield=True)` only stops **anyio's** delivery; a direct cancel reaches the shielded `asyncio.wait` and propagates), so activity-awaiting tasks unwind promptly and eviction can complete. That analysis wants validation against the reporter's production shape — @ZacharyHampton, if you can run your eviction profile against this branch, that would be the definitive check.

## Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6892?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_activity_execution.py
+++ b/pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_activity_execution.py
@@ -1,0 +1,51 @@
+"""Cancellation-safe activity execution for workflow code that runs under anyio cancel scopes.
+
+Awaiting `workflow.execute_activity()` directly from a task inside an anyio cancel scope can
+livelock the workflow when that scope is cancelled (e.g. by `asyncio.wait_for` cancelling an
+in-flight agent run — the graph engine runs its steps as anyio task-group children):
+
+- anyio delivers scope cancellation level-triggered: it re-arms `task.cancel()` via `call_soon`
+  for as long as a task remains in the cancelled scope, deduplicating only on `task._must_cancel`
+  or a done `_fut_waiter`.
+- `task.cancel()` on a task awaiting an activity handle DELEGATES to the handle task, whose
+  `run_activity` loop swallows the `CancelledError`, emits a `request_cancel_activity` command,
+  and re-parks on a shielded result future — leaving the scoped task uncancelled with an undone
+  waiter, so anyio cancels it again on the next loop iteration, appending another cancel command.
+- The activity's resolution can only arrive in a later activation, which can never start because
+  the event loop never goes idle: the spin continues until Temporal's deadlock detector fails the
+  workflow task, which then retries identically forever.
+
+This executor keeps the cancellation delivery on OUR task instead (via `asyncio.shield`), forwards
+exactly one cancellation to Temporal's graceful machinery, and waits for the configured
+`ActivityCancellationType` resolution inside an anyio-shielded scope so re-delivery stops and the
+activation can complete. Re-raising the original `CancelledError` also keeps standard asyncio
+semantics at the caller: `asyncio.wait_for` produces `TimeoutError` instead of leaking the
+activity's `ActivityError`, and cancelling the Temporal workflow still ends it as *Cancelled*.
+"""
+
+from __future__ import annotations
+
+import asyncio
+from collections.abc import Sequence
+from typing import Any
+
+import anyio
+from temporalio import workflow
+from temporalio.workflow import ActivityConfig
+from typing_extensions import Unpack
+
+
+async def execute_activity(activity: Any, *, args: Sequence[Any], **config: Unpack[ActivityConfig]) -> Any:
+    """Drop-in replacement for `workflow.execute_activity()` — see the module docstring for why."""
+    handle = workflow.start_activity(activity, args=args, **config)
+    try:
+        return await asyncio.shield(handle)
+    except asyncio.CancelledError:
+        # The cancellation hit this task because the shield kept the activity handle alive.
+        # Delegate exactly one cancellation to Temporal, then wait for its configured
+        # cancellation behavior without anyio redelivering cancellation on every loop turn.
+        if not handle.done():
+            handle.cancel()
+            with anyio.CancelScope(shield=True):
+                await asyncio.wait([handle])
+        raise
--- a/pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_activity_execution.py
+++ b/pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_activity_execution.py
@@ -44,7 +44,9 @@ async def execute_activity(activity: Any, *, args: Sequence[Any], **config: Unpa
         # The cancellation hit this task because the shield kept the activity handle alive.
         # Delegate exactly one cancellation to Temporal, then wait for its configured
         # cancellation behavior without anyio redelivering cancellation on every loop turn.
-        if not handle.done():
+        # The already-done arm is a real race (cancel landing in the same tick the activity
+        # resolves) that cannot be timed deterministically through the workflow API.
+        if not handle.done():  # pragma: no branch
             handle.cancel()
             with anyio.CancelScope(shield=True):
                 await asyncio.wait([handle])
--- a/pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
+++ b/pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_agent.py
@@ -45,6 +45,7 @@ from pydantic_ai.tools import (
 )
 
 from .._runtime_toolsets import reject_unsupported_runtime_toolsets
+from ._activity_execution import execute_activity
 from ._model import TemporalModel, TemporalProviderFactory
 from ._run_context import TemporalRunContext, deserialize_run_context
 from ._toolset import temporalize_toolset, toolset_temporal_activities
@@ -274,7 +275,7 @@ class TemporalAgent(WrapperAgent[AgentDepsT, OutputDataT]):
         serialized_run_context = self.run_context_type.serialize_run_context(ctx)
         async for event in stream:
             activity_config: ActivityConfig = {'summary': f'handle event: {event.event_kind}', **self.activity_config}
-            await workflow.execute_activity(
+            await execute_activity(
                 activity=self.event_stream_handler_activity,
                 args=[
                     _EventStreamHandlerParams(
--- a/pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_durability.py
+++ b/pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_durability.py
@@ -45,6 +45,7 @@ from pydantic_ai.settings import ModelSettings
 from pydantic_ai.tools import AgentDepsT, RunContext
 from pydantic_ai.toolsets import AbstractToolset, WrapperToolset
 
+from ._activity_execution import execute_activity
 from ._run_context import TemporalRunContext, deserialize_run_context
 from ._toolset import (
     TemporalWrapperToolset,
@@ -433,7 +434,7 @@ class TemporalDurability(BaseDurabilityCapability[AgentDepsT]):
             'summary': f'handle event: {event.event_kind}',
             **self._event_stream_handler_activity_config,
         }
-        await workflow.execute_activity(
+        await execute_activity(
             activity=self.event_stream_handler_activity,
             args=[
                 _EventStreamHandlerParams(event=event, serialized_run_context=serialized_run_context),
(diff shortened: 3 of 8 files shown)


ZacharyHampton:
Ran this branch (`dd73e7a`) against the #6883 repro. **Trigger A is fixed.**

| pydantic-ai | temporalio | result |
|---|---|---|
| 2.18.0 | 1.30.0 | livelock (TMPRL1101, repeated) |
| 2.19.0 | 1.30.0 | livelock |
| 2.20.0 | 1.30.0 | livelock |
| **this branch** | 1.30.0 | **CLEAN — `asyncio.TimeoutError`, no TMPRL1101, worker shuts down normally** |

Same script, same shape as the issue; the only variable is the install. Also confirms the `wait_for` side-effect you mentioned — the timeout surfaces as `TimeoutError` rather than leaking `ActivityError`.

**On trigger B I don't have a verdict for you yet, and I'd rather say that than hand you a number I don't trust.** Two synthetic shapes both failed to isolate it:

1. **Cache pressure** (`max_cached_workflows=1`, two workflows each parked on an in-flight model-request activity, second one forcing the first out): completes cleanly on **2.18.0** — so it never reproduced the bug in the first place and proves nothing about the fix. Eviction driven purely by cache pressure appears to unwind fine even before your change.
2. **Ceiling breach → eviction** (CPU burn in the workflow body past the deadlock timeout while an agent run is parked on an activity — the actual production sequence): reproduces a wedge, but the harness can't *attribute* it. A deterministic CPU burn re-breaches on every replay, so "workflow never completes" is the expected outcome whether eviction livelocks or unwinds perfectly and Temporal just retries a task that always overruns. Fixed and broken code look identical through that lens.

What separated the two in production was the profile shape, not the outcome: the eviction wedge showed **2,101,038 `_deliver_cancellation` calls inside a single `_run_once` with zero `Task.cancel` frames** — every task hitting `if task._must_cancel: continue` after `should_retry` was already set, so not one cancellation delivered in 2.1M iterations. Your reasoning that this is downstream of tasks absorbing cancellation matches that signature: with the executor keeping the cancel on the awaiting task, there should be nothing left re-arming.

So the definitive check is the one you asked for — the real workflow, in the environment that broke. We'll run this branch on staging against the classify workflow that produced that profile and report the activation profile back here. If it'd help in the meantime, I can share the eviction-side cProfile from the production wedge.

One incidental datapoint from the A runs: a wedged workflow-task thread also blocks graceful **worker shutdown** indefinitely on the unfixed versions, and shuts down normally on this branch.

DouweM:
Thanks for the fast, rigorous validation — the version matrix on your unmodified repro is exactly the confirmation we wanted for trigger A, and the worker-shutdown datapoint is a nice bonus (it follows from the same mechanism: the wedged activation thread could never finish its `_run_once`).

On trigger B, your two-shapes analysis matches ours: cache-pressure eviction of a *healthy* parked workflow unwinds fine even unfixed (the activity-awaiting task there is parked on temporalio's own machinery at eviction time, not mid-absorption), and the ceiling-breach harness can't attribute because the burn re-breaches deterministically. The 2.1M-zero-delivery signature is the discriminating observable, so a staging activation profile from the classify workflow is the right definitive check — and yes, the production eviction-side cProfile would be useful in the meantime (mainly to confirm every spinning frame traces to a task parked under an anyio scope on an activity await, i.e. the population this executor empties).

Note the branch moved two small commits since `dd73e7a` — a test pinning your MRE shape and coverage pragmas; no behavior change, so your results stand.

ZacharyHampton:
Here's the eviction-side profile. It's a cProfile of one real wedged activation (our own in-tree activation profiler, armed on one lane), taken on pydantic-ai 2.18.0 / temporalio 1.30.0:

| frame | calls |
|---|---:|
| anyio `_deliver_cancellation` | 2,101,038 |
| `call_soon` | 2,101,038 |
| `current_task` | 2,101,038 |
| `Handle.__init__` | 2,101,038 |
| `_run_once` | **1** |
| `Task.cancel` | **absent** |
| `isinstance` | **absent** |
| `_task_started` | **absent** |
| `Future.done` | **absent** |

All inside a **single** `_run_once`, which is what makes it a livelock rather than a slow loop — the activation never yields, so the resolution that would end it can never arrive.

**What this establishes:** every iteration did nothing but re-arm. `_deliver_cancellation` at exactly the `call_soon`/`Handle.__init__` count is a scope re-scheduling itself once per pass, and the **absence of `Task.cancel`** means not one cancellation was actually delivered in 2.1M iterations — so each pass took `if task._must_cancel: continue` with `should_retry` already set. Contrast the trigger-A profile, where `_deliver_cancellation` / `Task.cancel` / `shield` / `_apply_cancel_command` all sit at ~140,540 in a 1:1 ratio: there, delivery *was* attempted every pass and swallowed. Different signature, same root.

**What it does not establish**, to be precise about the question you asked: this is aggregate per-function counts, so it does not carry a per-task attribution — I can't show you from this profile that *every* spinning task was parked under an anyio scope on an activity await. What it does say is that no other machinery was running (no `Task.cancel`, no `Future.done`, no `_task_started`), and the only scopes in that workflow's tree belong to pydantic-graph's node execution around the activity awaits, which is consistent with your population claim but doesn't independently prove it. The staging run against your branch is what closes that gap, and I'll report the activation profile from it.

Thanks for the note on the two extra commits — `dd73e7a` is what I ran, and since they're a test plus coverage pragmas I'll leave the matrix as reported rather than re-run it.

DouweM:
That profile is a tight characterization — thanks. Three things it pins down for us:

1. **Pure re-arm, zero work**: `_deliver_cancellation` == `call_soon` == `Handle.__init__` with `Task.cancel`/`isinstance`/`Future.done`/`_task_started` absent means every one of the 2.1M passes exited each task's check at `if task._must_cancel: continue` — anyio re-scheduling itself once per pass with nothing to deliver and nothing that could change. (Which also makes the zero-delivery variant a candidate for an upstream anyio report — re-arming for tasks that already carry `_must_cancel` can't ever accomplish anything; we'll raise it once your staging run characterizes the fixed-code shape.)
2. **The `_must_cancel`-forever detail is itself informative**: CPython only sets `_must_cancel` when a task is cancelled while it has no cancellable waiter — including the case of being parked on an *already-completed* future. Tasks staying in that state across a 2.1M-iteration drain of the same `_run_once` means their wakeups were never processed — consistent with eviction cancelling the tree at a point where the completion callbacks that would step those tasks are behind the self-re-arming anyio handles in the ready queue, or dropped by eviction's teardown ordering. Either way: tasks that can't step, scopes that never empty, a drain that never ends.
3. **The discriminator for staging**: with the executor in place, cancellation stays on the awaiting task and is re-raised promptly, so no anyio scope should retain residents once eviction (or any cancellation) sweeps through. Concretely we'd expect a healthy activation profile to show `_deliver_cancellation` in the tens-to-hundreds (one or two passes per scope teardown), 1:1 with delivered or no-op exits — and `_run_once` returning. If staging still shows a growing zero-delivery tail, that would point at a resident population *outside* the activity awaits, and we'd want the per-task attribution then, not before.

And understood on the per-task attribution caveat — 'no other machinery was running' plus 'the only scopes in the tree are pydantic-graph's around activity awaits' is the right epistemic line, and the staging profile closes it. Looking forward to it.

ZacharyHampton:
Status on the staging trigger-B profile, so you're not waiting on it blind — and one finding that may change what it's worth measuring.

**Where it is.** The run is configured and validated end to end against our staging harness (cheap model, small scale). What it still needs is a worker whose deadlock ceiling is low enough to actually breach, which is a deploy-shaped step rather than a script argument. I'd rather say that than keep it silently queued.

**The finding, which is the interesting part.** Our harness estimates activation cost from requirements mass, and its own fit is ~0.078 CPU-s per KiB (measured points: 9,948 B → 0.65 s, 19,885 B → 1.43 s). Extrapolated, ~110 KiB would be needed to reach our 8 s ceiling. **But that fit predates our own fix for the discarded second schema walk** — the one where `customize_request_parameters` walks `output_tools` *and* `output_object` before `prepare_request` resolves the output mode, so each request paid the transform twice and threw one away (138,516 transform calls per request on a ~100 KB dynamic schema, 32.2 s activations). We memoized it locally.

So on current code we may not be able to breach the ceiling from input size at all, which is a good problem: **the production trigger for trigger B is largely gone on our side.** That doesn't tell you whether your executor fixes the eviction spin — it means our fastest route to a real wedge is now to lower the ceiling deliberately rather than to make the work expensive, and that a natural recurrence is unlikely to hand us one soon.

**What that implies for you.** If you're holding the anyio report on our number, I wouldn't block on it indefinitely. The zero-delivery signature is well characterized from the production capture (2,101,038 `_deliver_cancellation` == `call_soon` == `current_task` == `Handle.__init__`, `Task.cancel`/`isinstance`/`Future.done`/`_task_started` absent, one `_run_once`), and your reasoning about `_must_cancel`-forever tasks parked behind self-re-arming handles is consistent with every frame of it. If anyio wants a reproduction rather than a characterization, the re-arm-for-already-`_must_cancel`-tasks argument stands on its own without our staging run — it can never accomplish anything regardless of who cancelled.

I'll still get you the profile. Flagging the schema-walk fix now because it's also a candidate issue on your side (transform after output-mode resolution, memoized on schema content) and because it explains why a natural trigger-B recurrence may not show up in our fleet again.

DouweM:
Understood on the staging timeline — a deploy-shaped step is a deploy-shaped step, and the honest status is more useful than a silently queued number. Agreed on the anyio-report posture too: the re-arm-for-already-`_must_cancel`-tasks argument is self-contained (re-arming can never accomplish anything for a task that will cancel itself at its next step, regardless of who cancelled), so we won't hold it on your run; your production capture serves as the motivating field report rather than the proof.

And thank you for flagging the discarded schema walk — verified in our source and filed as #6939 with your numbers and both remedies (transform after output-mode resolution + content-memoization). That it also de-fangs your natural trigger-B recurrence is a satisfying kind of collateral: the wedge needed an expensive activation to breach the ceiling, and the expense was itself a bug.

tapetersen:
I just handled the issue on anyio's side. The isolation you do here with cancelling the task and then waiting on it to finish in a shielded scope looks correct and should handle a task ignoring cancellations repeatedly. 

As I noted after some investigation I'm not sure your B case can actually be triggered as a stuck loop as opposed to just a couple of extra spins with normal Api's respecting documented asyncio invariants.

DouweM:
@tapetersen Thanks for looking into this Tobias!

Issue #5165, opened 2026-04-23
OpenAI streaming iterator crashes when streamed chunk has choices=None


alexandreczg:
### Initial Checks

- [x] I'm using the [latest version](https://github.com/pydantic/pydantic-ai/releases/latest) of Pydantic AI
- [x] I've searched for my issue in [the issue tracker](https://github.com/pydantic/pydantic-ai/issues) before opening this issue

### Description

We're seeing an intermittent crash in the OpenAI streaming path on the latest PydanticAI release.

The exception is:

```text
TypeError: 'NoneType' object is not subscriptable
```

Traceback points to:

```python
# pydantic_ai/models/openai.py
choice = chunk.choices[0]
```

Specifically, this happens in `OpenAIStreamedResponse._get_event_iterator()` when a streamed chunk has `choices=None`.

Observed behavior:

- the model stream is otherwise healthy and has already yielded many prior chunks
- the crash happens mid-stream while converting raw streamed OpenAI-compatible chunks into PydanticAI native events
- downstream UI code only sees the propagated error, but the root cause is earlier in the PydanticAI OpenAI stream parser

Relevant traceback:

```text
File "/workspaces/bossclaw/.venv/lib/python3.13/site-packages/pydantic_ai/models/openai.py", line 2289, in _get_event_iterator
    choice = chunk.choices[0]
             ~~~~~~~~~~~~~^^^
TypeError: 'NoneType' object is not subscriptable
```

Observed result on `pydantic-ai==1.68.0`:

```text
Traceback (most recent call last):
  File ".../pydantic_ai/models/openai.py", line 2289, in _get_event_iterator
    choice = chunk.choices[0]
             ~~~~~~~~~~~~~^^^
TypeError: 'NoneType' object is not subscriptable
```

The relevant implementation currently handles `choices=[]` via `IndexError`, but not `choices is None`:

```python
try:
    choice = chunk.choices[0]
except IndexError:
    continue
```

A guard like `if not chunk.choices: continue` appears to fix the failure mode.

Current code appears to handle `choices=[]` via `IndexError`, but not `choices is None`:

```python
try:
    choice = chunk.choices[0]
except IndexError:
    continue
```

Expected behavior:

- if a provider returns a streamed chunk with `choices=None`, PydanticAI should skip or explicitly handle that chunk instead of raising `TypeError`
- a guard like `if not chunk.choices: continue` seems more robust here

Related prior issue:

[...]

PR #3571, opened 2025-11-27
Fix error when OpenRouter API returns `None` tool call arguments

barp:
This only happens if you have a tool with all optional parameters.

for example the following response can return from open router:
```
{'id': 'gen-1764223032-90is29WrhzbOG2XawxiT', 'choices': [{'finish_reason': 'tool_calls', 'index': 0, 'logprobs': None, 'messag
e': {'content': '', 'refusal': None, 'role': 'assistant', 'annotations': None, 'audio': None, 'function_call': None, 'tool_call
s': [{'id': 'toolu_vrtx_0142STg84ZeUdDEVtucWMjys', 'function': {'arguments': None, 'name': 'find_education_content'}, 'type': '
function', 'index': 0}], 'reasoning': None}, 'native_finish_reason': 'tool_calls'}], 'created': 1764223033, 'model': 'anthropic
/claude-sonnet-4.5', 'object': 'chat.completion', 'service_tier': None, 'system_fingerprint': None, 'usage': {'completion_token
s': 23, 'prompt_tokens': 8146, 'total_tokens': 8169, 'completion_tokens_details': None, 'prompt_tokens_details': None}, 'provid
er': 'Google'}
```

as you can see the function arguments can be None.

The following code reproduces the issue:
```
from pydantic import BaseModel
from pydantic_ai import RunContext, Agent
from pydantic_ai.models.openrouter import OpenRouterModel

model = OpenRouterModel(
    'anthropic/claude-sonnet-4.5',
)

agent = Agent[None, str](
    model=model,
    output_type=str,
    instrument=True,
    retries=5,
)

class FindEducationContentFilters(BaseModel):
    title: str | None = None

@agent.tool
async def find_education_content(
    ctx: RunContext[None], filters: FindEducationContentFilters
) -> str:
    return "This is a content about education"

print(agent.run_sync("Can you find me education content?"))
```

PR #6901, opened 2026-07-30
Raise `ModelAPIError` for OpenRouter responses with null choices and no error envelope

### LINKED ISSUE #6900: OpenRouter: null `choices` with no error envelope raises a bare ValidationError instead of ModelAPIError

ZacharyHampton:
Fixes #6900

Replaces #6885, which the linked-issue bot auto-closed when I edited its description (my fault — the rewrite dropped the issue reference, and I couldn't reopen from a fork). Same branch, same commits; @DouweM's review and the discussion of both design changes are on #6885.

OpenRouter intermittently returns a body with **null `choices` and no error envelope at all** — a transient provider hiccup carrying no completion. `_validate_completion` tries the error-envelope and nested-provider shapes (#3994), then re-raises the bare `ValidationError`, so consumers that classify on exception type treat a transient as fatal. Nothing retries it and `FallbackModel` doesn't fall back on it.

This models that shape and raises `ModelAPIError` for it.

### The shape

`_OpenRouterNoCompletionResponse`, a sibling of `_OpenRouterErrorResponse` — the same family of bodies, minus the error envelope:

```python
class _OpenRouterNoCompletionResponse(BaseModel, extra='allow'):
    choices: None
    error: None = None
    provider: str | None = None
    model: str | None = None
```

Tried in the existing `except ValidationError` chain, after the error-envelope attempt and before the nested-provider one. Position is inert — no body can match both this shape and the nested-provider shape, since `provider` is typed `str | None` here — but the description should match the code. Deliberate details:

- **`provider: str | None`** — a body whose `provider` is a *dict* is a malformed nested-provider response, not this shape, so it fails validation here and stays fatal. Declarative, no `isinstance` check.
- **`choices: None` is required, not defaulted** — a body that merely omits `choices` doesn't silently match.
- **`_OpenRouterChatCompletion` is untouched**: `id`, `model`, `object`, `provider` all stay required and non-null, so the happy path loses no guarantees and a `ValidationError` from it still means "a shape we don't account for."

### The exception

`ModelAPIError`, not a faked HTTP status. It's `FallbackModel`'s default `fallback_on`, so the transient now reaches retry/fallback machinery that a bare `ValidationError` never did.

### Tests

Both boundaries pinned, and all existing quirky-shape tests unchanged:

- `test_openrouter_null_choices_without_error_envelope_is_retryable` — the envelope-less body raises `ModelAPIError`, and asserts it is *not* a `ModelHTTPError` so nobody re-adds a fake status.
- `test_openrouter_null_choices_with_malformed_provider_dict_stays_fatal` — the malformed-`provider` body still raises `UnexpectedModelBehavior`.

Full `tests/models/test_openrouter.py`: 58 passed.

---

We've run the workaround for this as a monkey-patch in production — it's the last survivor of a once-574-line OpenRouter patch module, since v2 absorbed the rest. Discussed with @DouweM on a call; the approach here follows his review (`ModelAPIError` over a faked 502, and modelling the shape rather than inspecting the response dict).

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6901?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/openrouter.py
+++ b/pydantic_ai_slim/pydantic_ai/models/openrouter.py
@@ -1241,10 +1241,13 @@ class OpenRouterStreamedResponse(OpenAIStreamedResponse):
                     validated = _OpenRouterChatCompletionChunk.model_validate(chunk_dict)
                 except ValidationError as exc:
                     # Parity with `OpenRouterModel._validate_completion`: the same no-completion body can
-                    # arrive mid-stream, and this strict chunk model rejects it before `OpenAIStreamedResponse`
-                    # can reach its tolerant `if not chunk.choices` guard, so classify it here instead.
-                    # Classification only: `FallbackModel` catches failures while *opening* a stream, so
-                    # unlike the non-streamed path this does not reach fallback.
+                    # arrive mid-stream. `_OpenRouterChatCompletionChunk.choices` is required and non-null,
+                    # so a null-`choices` chunk has always been terminal here — unlike `OpenAIStreamedResponse`,
+                    # which skips such chunks (https://github.com/pydantic/pydantic-ai/issues/5165). This only
+                    # improves the exception, from a bare `ValidationError` to `ModelAPIError`; it does not
+                    # newly terminate any stream that used to succeed.
+                    # Classification only: `FallbackModel`'s window is `Model.request_stream`'s own
+                    # `__aenter__`, which returns before any chunk is validated, so this never falls back.
                     _raise_for_no_completion(chunk_dict, self._model_name, exc)
                     raise
                 yield validated
--- a/pydantic_ai_slim/pydantic_ai/models/openrouter.py
+++ b/pydantic_ai_slim/pydantic_ai/models/openrouter.py
@@ -8,7 +8,7 @@ from pydantic import BaseModel, Discriminator, ValidationError, field_validator
 from typing_extensions import TypedDict, assert_never, override
 
 from .. import usage
-from ..exceptions import ModelHTTPError, UserError
+from ..exceptions import ModelAPIError, ModelHTTPError, UserError
 from ..messages import (
     BinaryContent,
     CachePoint,
@@ -1030,15 +1030,15 @@ class OpenRouterModel(OpenAIChatModel):
                     and not isinstance(response_dict.get('provider'), dict)
                 ):
                     # A null-`choices` body with no error envelope and no nested-provider
-                    # payload is a transient provider hiccup, not a malformed request:
-                    # surface it as a retryable upstream error rather than a bare
-                    # ValidationError. A `provider` dict that failed to validate above is
-                    # excluded on purpose — that body tried to be a nested-provider
-                    # response and was malformed, which must stay fatal.
-                    raise ModelHTTPError(
-                        status_code=502,
+                    # payload is a transient provider hiccup, not a malformed request.
+                    # `ModelAPIError` is `FallbackModel`'s default `fallback_on`, so this
+                    # reaches retry/fallback machinery that a bare `ValidationError` never
+                    # did. A `provider` dict that failed to validate above is excluded on
+                    # purpose — that body tried to be a nested-provider response and was
+                    # malformed, which must stay fatal.
+                    raise ModelAPIError(
                         model_name=response_dict.get('model') or self.model_name,
-                        body='OpenRouter returned a response with null `choices` and no error envelope',
+                        message='OpenRouter returned a response with null `choices` and no error envelope',
                     ) from exc
                 raise exc
 
--- a/pydantic_ai_slim/pydantic_ai/models/openrouter.py
+++ b/pydantic_ai_slim/pydantic_ai/models/openrouter.py
@@ -589,20 +589,17 @@ class _OpenRouterErrorResponse(BaseModel, extra='allow'):
 
 
 class _OpenRouterNoCompletionResponse(BaseModel, extra='allow'):
-    """OpenRouter response that carries no completion at all: `choices` is null and there is no error envelope.
-
-    A sibling of `_OpenRouterErrorResponse` — the same family of bodies OpenRouter
-    returns when a downstream provider hiccups, minus the error envelope. Modelling it
-    keeps a `ValidationError` from `_OpenRouterChatCompletion` meaning "a shape we do not
-    account for" rather than doubling as a transient signal.
+    """OpenRouter body carrying no completion: null `choices` and no error envelope (see https://github.com/pydantic/pydantic-ai/issues/6900).
 
     `provider` is typed `str | None` on purpose: a body whose `provider` is a *dict* is a
     malformed nested-provider response, not this shape, so it fails validation here and
     stays fatal.
 
     `Literal[None]` rather than a bare `None` annotation: this module uses
-    `from __future__ import annotations`, so annotations are strings, and resolving the
-    string `'None'` raises `PydanticUserError` on the lowest supported pydantic (2.12).
+    `from __future__ import annotations`, so annotations are strings, and on Python 3.14
+    (PEP 649) resolving the string `'None'` raises `PydanticUserError` under pydantic
+    below 2.13 — the `3.14 (lowest-versions)` CI job. It resolves fine on 3.13 and on
+    pydantic 2.13+, so a local check on either will not reproduce it.
     """
 
     choices: Literal[None]
@@ -627,6 +624,26 @@ class _OpenRouterNestedProviderResponse(BaseModel, extra='allow'):
     """OpenRouter response where the real completion is nested in `provider` (see https://github.com/pydantic/pydantic-ai/issues/3994)."""
 
     provider: _OpenRouterNestedCompletion
+
+
+def _raise_for_no_completion(response_dict: dict[str, Any], model_name: str, exc: ValidationError) -> None:
+    """Raise `ModelAPIError` if the body carries no completion and no error envelope.
+
+    Returns normally for any other shape so the caller can keep trying the shapes it knows.
+    Shared by the non-streamed and streamed paths so both classify this body the same way.
+    """
+    try:
+        no_completion = _OpenRouterNoCompletionResponse.model_validate(response_dict)
+    except ValidationError:
+        return
+
+    # A body with no completion and no error envelope is a transient provider hiccup, not a
+    # malformed request, so it belongs in the `ModelAPIError` family rather than surfacing as a
+    # bare `ValidationError` (which reaches callers as `UnexpectedModelBehavior`).
+    raise ModelAPIError(
+        model_name=no_completion.model or model_name,
+        message='OpenRouter returned a response with null `choices` and no error envelope',
+    ) from exc
 
 
 def _map_openrouter_provider_details(
@@ -1044,22 +1061,14 @@ class OpenRouterModel(OpenAIChatModel):
                     body=error_response.error.message,
                 )
 
+            # `ModelAPIError` is `FallbackModel`'s default `fallback_on`, so on this non-streamed
+            # path the transient now falls back where a bare `ValidationError` never did.
+            _raise_for_no_completion(response_dict, self.model_name, exc)
+
             try:
                 nested = _OpenRouterNestedProviderResponse.model_validate(response_dict)
             except ValidationError:
-                try:
-                    no_completion = _OpenRouterNoCompletionResponse.model_validate(response_dict)
-                except ValidationError:
-                    raise exc
-                else:
-                    # A body with no completion and no error envelope is a transient
-                    # provider hiccup, not a malformed request. `ModelAPIError` is
-                    # `FallbackModel`'s default `fallback_on`, so this reaches retry and
-                    # fallback machinery that a bare `ValidationError` never did.
-                    raise ModelAPIError(
-                        model_name=no_completion.model or self.model_name,
-                        message='OpenRouter returned a response with null `choices` and no error envelope',
-                    ) from exc
+                raise exc
 
             validated = nested.provider
             if not validated.created:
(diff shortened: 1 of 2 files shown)


ZacharyHampton:
Fixed the `test on 3.14 (lowest-versions)` failure in 4c7287c — it was mine, and a genuine portability bug rather than a flake.

`_OpenRouterNoCompletionResponse` declared `choices: None` / `error: None = None`. Because this module uses `from __future__ import annotations`, those annotations are the *string* `'None'`, and resolving that raises `PydanticUserError: unevaluable-type-annotation` on the lowest supported pydantic (2.12) — so the module failed at import and the whole file errored during collection. It passed on my machine because the default resolution pulls a newer pydantic that resolves it fine.

Now spelled `Literal[None]`, which is string-resolvable and means the same thing, with a comment on the class saying why so nobody simplifies it back.

Verified both ways this time rather than just the default env:
- `uv sync --all-extras --resolution lowest-direct` on **3.14** (pydantic 2.12.0) — 58 passed
- default resolution on 3.12 — 58 passed

dsfaccini:
This comment was posted by Claude Code using claude-opus-5 on behalf of David. David decided the scope of this change but has not reviewed the code itself.

Thanks for this — picking it up to get it over the line. Heads up on one change we made rather than asking you to.

**We extended the fix to the streaming path.** Same body, same classification: `OpenRouterStreamedResponse._validate_response` now maps a null-`choices`/no-envelope chunk to `ModelAPIError` too, sharing one `_raise_for_no_completion` helper with `_validate_completion` so both paths can't drift.

Why it wasn't optional: `_OpenRouterChatCompletionChunk.choices` is required and non-null, so that chunk was rejected *before* `OpenAIStreamedResponse` could reach its tolerant `if not chunk.choices: continue` guard — the one added for #5165 precisely because OpenAI-compatible providers emit `choices: null`. So the same body was a retryable `ModelAPIError` non-streamed and a raw unmapped `ValidationError` streamed. `models/AGENTS.md` requires identical processing across `request()` and `request_stream()`, and DouweM has asked for that parity on this exact file three times (#3571, #4889, #4528).

Also in: two streaming tests, a `FallbackModel` test proving the non-streamed body actually falls back (nothing pinned that before), your `Literal[None]` docstring corrected to name **Python 3.14 + pydantic <2.13** as the trigger rather than pydantic 2.12 alone, and the flattened quirky-shape ladder.

**One limitation worth knowing if you rely on fallback in production:** on the streaming path this is classification only, not fallback. `FallbackModel.request_stream` catches failures while *opening* a stream, and the chunk is only rejected once chunks are consumed — so a streamed no-completion body raises `ModelAPIError` instead of a raw `ValidationError`, but it propagates rather than falling back. That asymmetry is `FallbackModel`'s, not OpenRouter's, and we left it alone here. Our first attempt at the test asserted fallback *did* happen on the stream; it failed, which is how we found it.

That last one is worth your time: your `Literal[None]` fix is correct and load-bearing, but four independent reviewers here tested it on 3.10/3.12/3.13, found it "didn't reproduce", and recommended reverting to bare `None` — which would have re-broken the `3.14 (lowest-versions)` job you fixed in `4c7287c7c`. The docstring now says a local check on 3.13 or pydantic 2.13+ won't reproduce it, so the next reader doesn't fall in.

One duplicate removed: `test_openrouter_null_choices_with_malformed_provider_dict_stays_fatal` was AST-identical to the existing `test_openrouter_provider_dict_without_choices_raises`, so its rationale moved into that test's docstring instead.

No action needed from you — just didn't want the streaming commit to look like it appeared out of nowhere.

dsfaccini:

|---|---|---|---|---|
| 1 · code-correctness | ✅ closed | 0 | 0 | 10 |
| 3 · cross-cutting/siblings | ✅ closed | 0 | 0 | 2 |

**Verdict:** ✅ PASS — was 🔧 10 required; all 10 resolved in `d91f12c05` (+ PR-body correction). The suite flagged them and the same pass fixed them; it did not start clean. · 10 applied
Reviewed at 2026-07-30T16:40:00Z · diff `6ff468d863`

<details>
<summary>✅ 10 required action(s) resolved</summary>

_The 10 required actions this suite flagged, all resolved in `d91f12c05` (+ the PR-body correction). Verified: 63/63 tests in `tests/models/test_openrouter.py` pass, targeted pyright clean, pre-commit (incl. whole-repo pyright) green. The suite was **not** re-run end-to-end against the new diff hash — the changes are two added tests plus comment/docstring edits, each the fix its reviewer specified._

**1. ✅ Streamed no-completion is pinned only as the *first* chunk** · `tests/models/test_openrouter.py` · `test_openrouter_null_choices_streaming_raises_model_api_error`
Both streaming tests pass a single chunk, so the raise always lands with zero accumulated state. Verified: when the bad chunk arrives *mid-stream*, `model_name` resolves to `first_chunk.model` (`google/gemini-2.5-flash`), **not** the configured name the first-chunk test asserts — because `OpenAIChatModel._process_streamed_response` sets `_model_name = first_chunk.model or self.model_name`. That coordinate is pinned nowhere. Fix: `create_mock_stream([<good chunk>, _null_choices_chunk()])` + snapshot `model_name`. (Two reviewers converged: ambient-matrix, suspend-resume.)

**2. ✅ `FallbackModel` × streamed is asserted only in a code comment** · `tests/models/test_openrouter.py` · `test_openrouter_null_choices_triggers_fallback`
The PR's first attempt at this test asserted fallback and **failed**; the corrected claim (streamed raises, does not fall back) now survives only as prose. Fix: streamed twin asserting `pytest.raises(ModelAPIError)` and that the output is not the fallback's.

**3. ✅ `pytest.raises(ValidationError)` cannot prove its own docstring** · `tests/models/test_openrouter.py` · `test_openrouter_streaming_malformed_chunk_stays_fatal`
Docstring says the chunk "keeps re-raising the original error", but the bare catch passes for any rejection reason. Verified discriminator: the provider-dict chunk carries a 4th error `provider / Input should be a valid string`. Fix: `match='provider'`.

**4. ✅ The `error: Literal[None]` boundary is guarded only incidentally** · `tests/models/test_openrouter.py` · `test_openrouter_malformed_error_fallthrough`
This diff amended the sibling test's docstring to pin the `provider: str | None` boundary but left `error` asymmetric. Verified load-bearing: widening `error` to `Any` flips that test from `UnexpectedModelBehavior` to a retryable `ModelAPIError`. Fix: one docstring line.

**5–6. ✅ `Any` on two new test helpers** · `tests/models/test_openrouter.py` · `_null_choices_chunk`, `_null_choices_completion`
Repo rule bans `Any` where a narrower type exists. Call sites give the exact types: `provider: dict[str, str] | None`, and explicit `model`/`provider` kwargs instead of `**kwargs: Any`. (Two reviewers converged: code-reuse, patterns.)

**7. ✅ Past-state comment** · `pydantic_ai_slim/pydantic_ai/models/openrouter.py` · `OpenRouterModel._validate_completion`
"…now falls back where a bare `ValidationError` never did" narrates before/after rather than stating the invariant — the exact shape `CLAUDE.md` gives as its bad example. Fix: state the invariant.

**8. ✅ `_OpenRouterNoCompletionResponse` docstring: say why `choices` can't be dropped** · `pydantic_ai_slim/pydantic_ai/models/openrouter.py`
On #4528 DouweM twice asked to delete a `choices: None` field from this exact model family in favour of `extra='allow'`. Here `choices` **is** the discriminator, so it can't go — but nothing says so, and he'll read it as the pattern he already rejected. Fix: one docstring line.

**9. ✅ Scope note beside the streamed call site** · `pydantic_ai_slim/pydantic_ai/models/openrouter.py` · `OpenRouterStreamedResponse._validate_response`
`models/AGENTS.md` parity was flagged on the predecessor #4528 and left open. State that only the no-completion shape is mapped here because it's the only one reproduced on a stream.

**10. ✅ PR body contradicts the diff** · PR description
Body says the new model is tried "**after** the error-envelope and nested-provider attempts"; the code places it **before** the nested-provider attempt. Behaviorally inert (the shapes are mutually exclusive), but DouweM personally specified this chain's ordering on #4528.
</details>

<details>
<summary>⚠️ 12 informational (defer/skip)</summary>

**Two BLOCKING claims were raised and then REFUTED by orchestrator verification** — recorded so they aren't re-raised:
- *"Streamed error-envelope reaches the caller as an unclassifiable `ValidationError`"* — **refuted**. The probe used `create_mock_stream`, which bypasses SSE decoding. Through a real SSE stream the openai SDK raises `APIError` at decode (`_streaming.py`, `if is_mapping(data) and data.get("error")`), caught by the pre-existing `except APIError` arm → `ModelHTTPError`.
- *"The documented mid-stream SSE error event is silently swallowed; a 429 surfaces as retry-exhaustion"* — **refuted**. Same cause. Verified end-to-end: the documented payload yields `ModelHTTPError: status_code: 429, body: Rate limit exceeded`, message intact.

**The #5165 "benign skip" precedent does not apply to OpenRouter** — `OpenAIStreamedResponse` skips null-`choices` chunks because some OpenAI-compat proxies send a usage-only final chunk. Verified: **zero** `"choices":null` / `"choices":[]` frames across all 8 OpenRouter streaming cassettes. And if OpenRouter did emit one, every existing OpenRouter stream test would already fail today — `_OpenRouterChatCompletionChunk.choices` is required non-null pre-PR.

**No sibling provider shares the defect** — reproduced across `OpenAIChatModel` (openai + deepseek providers), `CerebrasModel`, `OllamaModel`, `ZaiModel`: all raise `UnexpectedModelBehavior`, none raises `ModelAPIError`. That is shared *handling*, not a shared *defect* — the emission is router-specific (OpenRouter proxies downstream providers). Correctly left out of scope per `CLAUDE.md`.

**No cassette is warranted, and a hand-written one would be barred** — the body is an unprovokable intermittent fault, so `--record-mode=rewrite` cannot capture it. Mock fidelity verified on **both** paths: a raw `{"choices": null}` 200 parsed by the real SDK (and via real SSE framing) yields a `model_dump()` identical to the tests' `model_construct`. Matches the #3994 sibling convention (all five are `model_construct`-based).

**Exception-type narrowing is version-policy exempt** — users catching `UnexpectedModelBehavior` (non-streamed) or `ValidationError` (streamed) stop catching this body. Permitted as an undocumented-assumption bug fix; the changelog-bearing PR title carries the signal.

**Temporal/Prefect will now retry this transient** — the body moves off the `non_retryable_error_types` allow-list. Intended, and precedented by the sibling error-envelope branch already raising `ModelHTTPError` (⊂ `ModelAPIError`). DBOS does no exception-type classification.

**Streamed parity is 1 of 3 shapes** — deliberate scope boundary; extending to the other two rests on an unreproduced hunch.

**Docs: 0 gaps** — both new symbols are private (mkdocstrings excludes them); the fallback contract is already in `docs/models/overview.md`, and the streamed-vs-non-streamed asymmetry is `FallbackModel`'s, documented in `FallbackModel.request_stream`'s own docstring.

**Coverage: 100% on every added line**, both branches of `_raise_for_no_completion` and both sides of `no_completion.model or model_name`.
</details>

PR #6945, opened 2026-07-30
Don't reset an existing Logfire configuration in `LogfirePlugin`

### LINKED ISSUE #6915: LogfirePlugin's default setup calls bare logfire.configure(), resetting an already-configured host
### Description

`LogfirePlugin`'s default setup callable calls a bare `logfire.configure()`:

```python
# durable_exec/temporal/_logfire.py
def _default_setup_logfire() -> Logfire:
    instance = logfire.configure()
    ...
```

In a worker process that has **already** configured logfire (its own token, scrubbing, exporters, `include_content=False`, console output off), that re-configures it and resets the host configuration. For us the visible symptom was workflow threads console-exporting whole span trees, which pushed activations past the deadlock-detection ceiling: **41+ TMPRL1101 workflow-task failures in one day** from what is nominally just telemetry setup.

It's easy to hit precisely because `LogfirePlugin` looks like the thing you're supposed to use — you add it for the Temporal instrumentation and sandbox passthrough, and it silently takes over your logging configuration as a side effect.

### Suggestion

Default to *not* reconfiguring when logfire is already configured — either use the existing instance (`logfire.DEFAULT_LOGFIRE_INSTANCE` / whatever the current accessor is) or make the default callable a no-op that returns the existing instance, with explicit configuration left to the caller who passes `setup_logfire=`.

We pass our own callable, so this is not blocking us; filing it because the failure mode (telemetry setup causing deadlock-detector failures) is very hard to attribute back to the plugin, and the fix looks like a one-liner.

### Notes

Environment: `pydantic-ai-slim` 2.18.0 → 2.20.0 (verified unfixed on main), `temporalio` 1.30.0, AWS Lambda workers against Temporal Cloud, logfire configured at process start by our own bootstrap.

davidpavlovschi:
- Closes #6915

`LogfirePlugin`'s default setup (`pydantic_ai/durable_exec/temporal/_logfire.py::_default_setup_logfire`) called bare `logfire.configure()`. Because that resets `logfire.DEFAULT_LOGFIRE_INSTANCE`'s config, an app that had already called `logfire.configure(...)` itself lost its settings the moment a Temporal client connected with the plugin: service name, sampling and exporters were replaced by defaults, and the console exporter came back on (which is what produces console writes inside Temporal workflow threads).

The default setup now reuses the already-configured default instance and only calls `logfire.configure()` when Logfire has not been configured yet, so behaviour is unchanged for apps that never configure Logfire themselves. Passing a custom `setup_logfire` callable keeps working exactly as before.

Reproduced before the change with a small script: after `logfire.configure(send_to_logfire=False, console=False, service_name='host-app')`, calling `_default_setup_logfire()` left `service_name=''` and `console=ConsoleOptions(...)`; after the change the host config survives untouched.

### Tests

Two tests in `tests/test_temporal.py` (neither needs a Temporal server):

- `test_default_setup_logfire_keeps_existing_configuration` — with Logfire configured via the `capfire` fixture, asserts the default setup returns the same `DEFAULT_LOGFIRE_INSTANCE`, leaves `send_to_logfire`/`console` untouched, and that the app's span processors still export. This test fails on `main` (console options reinstated) and passes with the fix.
- `test_default_setup_logfire_configures_when_not_configured` — asserts `logfire.configure()` is still called exactly once when Logfire has not been configured.

Docs: one sentence added to `docs/durable_execution/temporal.md` noting the default setup reuses an existing configuration.

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6945?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/durable_execution/temporal.md
+++ b/docs/durable_execution/temporal.md
@@ -410,7 +410,7 @@ async def main():
     )
 ```
 
-By default, the `LogfirePlugin` will instrument Temporal (including metrics) and Pydantic AI and send all data to Logfire. To customize Logfire configuration and instrumentation, you can pass a `logfire_setup` function to the `LogfirePlugin` constructor and return a custom `Logfire` instance (i.e. the result of `logfire.configure()`). To disable sending Temporal metrics to Logfire, you can pass `metrics=False` to the `LogfirePlugin` constructor.
+By default, the `LogfirePlugin` will instrument Temporal (including metrics) and Pydantic AI and send all data to Logfire. If you already called `logfire.configure()` yourself, the default setup reuses that configuration instead of resetting it. To customize Logfire configuration and instrumentation, you can pass a `logfire_setup` function to the `LogfirePlugin` constructor and return a custom `Logfire` instance (i.e. the result of `logfire.configure()`). To disable sending Temporal metrics to Logfire, you can pass `metrics=False` to the `LogfirePlugin` constructor.
 
 ## Known Issues
 
--- a/pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_logfire.py
+++ b/pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_logfire.py
@@ -14,7 +14,11 @@ if TYPE_CHECKING:
 def _default_setup_logfire() -> Logfire:
     import logfire
 
-    instance = logfire.configure()
+    instance = logfire.DEFAULT_LOGFIRE_INSTANCE
+    # If the app already called `logfire.configure()`, calling it again would reset its configuration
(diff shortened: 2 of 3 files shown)


davidpavlovschi:
Closing — #6962 landed the same fix (DEFAULT_LOGFIRE_INSTANCE + the _initialized guard) an hour after I opened this, so this is superseded.

Issue #6908, opened 2026-07-30
`override(toolsets=...)` inside a durable container silently un-durables the run on all three engines


DouweM:
### Description

`agent.override(toolsets=[...])` inside a Temporal workflow, DBOS workflow, or Prefect flow **silently runs the overridden toolset's tools in container code** — no activity, no step, no task, no error, no warning. On Temporal that means I/O inside the workflow sandbox and non-determinism on replay; on DBOS and Prefect it means un-checkpointed execution that re-runs on recovery/retry.

The sibling entry point is correctly guarded, which is what makes this a hole rather than a design: `run(toolsets=[...])` raises a `UserError`. Verified on both engines I can run locally.

**Prefect** (`FunctionToolset`, which Prefect wraps in tasks):

```
construction-time toolset  -> ['construction:task']
override(toolsets=[...])   -> ['override:FLOW-CODE']
run(toolsets=[...])        -> rejected: UserError FunctionToolset cannot be passed to `run(toolsets=...)` at runtime with Prefect, because toolsets that execute their own tools or resolve dynamically must be registered for durable execution when the agent is constructed. ...
```

**DBOS** (`DynamicToolset`, which DBOS wraps in steps):

```
construction DynamicToolset -> ['construction:step']
override(toolsets=[dyn])    -> ['override:WORKFLOW-CODE']
run(toolsets=[dyn])         -> rejected: UserError DynamicToolset cannot be passed to `run(toolsets=...)` at runtime with DBOS, because toolsets ...
```

The deprecated wrapper agent guarded this. `TemporalAgent` raises:

> `Toolsets cannot be contextually overridden inside a Temporal workflow, they must be set at agent creation time.`

No durability capability has an equivalent.

### Root cause: the guard's own input is contaminated by what it is meant to catch

`BaseDurabilityCapability._reject_runtime_toolsets` builds its set of known-good leaves from `self._agent.toolsets` **at run time** — `durable_exec/_base.py:126-128`:

```python
construction_leaves: set[int] = set()
if self._agent is not None:
    for agent_toolset in self._agent.toolsets:
        agent_toolset.apply(lambda leaf: construction_leaves.add(id(leaf)))
```

[...]

PR #6952, opened 2026-07-30
Honor per-tool `TaskConfig` on Prefect MCP tools and add `PydanticUserError` to its non-retryable set

### LINKED ISSUE #6912: Three durable-exec parity inconsistencies: Prefect non-retryable set, override(tools=...) rejection wording, and silently-ignored per-tool config
### Description

Three small cross-engine inconsistencies found while auditing the durable-execution integrations for parity. Each is independently small; they are grouped because they are all "one engine does X, another silently doesn't, and nothing records the decision". Happy to split if you'd rather have three.

---

## 1. Prefect's non-retryable error set is missing `PydanticUserError`

**Bucket: genuine gap — plain drift.**

Temporal marks three framework error types non-retryable on every activity config, `temporal/_toolset.py:94-100`:

```python
def with_non_retryable_errors(retry_policy: RetryPolicy | None) -> RetryPolicy:
    """Return a copy of `retry_policy` with the framework's non-retryable errors ensured."""
    retry_policy = copy.copy(retry_policy) if retry_policy else RetryPolicy()
    existing = retry_policy.non_retryable_error_types or []
    additional = [UserError.__name__, PydanticUserError.__name__, UnexpectedModelBehavior.__name__]
```

and repeats the same three at the workflow level (`temporal/__init__.py:147`: `workflow_failure_exception_types=[UserError, PydanticUserError, AgentRunError]`).

Prefect's identically-named helper checks only two, `prefect/_toolset.py:26-31`:

```python
async def retry_condition(task: Any, task_run: Any, state: Any) -> bool:
    result = state.result(raise_on_failure=False)
    if inspect.isawaitable(result):
        result = await result
    if isinstance(result, (UserError, UnexpectedModelBehavior)):   # <- PydanticUserError missing
        return False
```

So a `PydanticUserError` (a schema-construction error — a framework misconfiguration that retrying cannot fix) is retried on Prefect according to the configured policy, and not on Temporal. The comment directly above `prefect/_durability.py:110-112` states the intent that is not met:

> `# Model and event-handler tasks compose the same non-retryable condition as tool tasks: a`
> `# UserError/UnexpectedModelBehavior raised inside them (e.g. a model that can't be`
> `# rebuilt on the worker) is a framework misconfiguration that retrying can't fix.`

Fix: add `PydanticUserError` to the `isinstance` tuple. One line.

DBOS has no equivalent mechanism at all, which is a genuine engine limitation, recorded in code comments (`dbos/_dynamic_toolset.py:35`, `dbos/_mcp_toolset.py:44`: *"DBOS has no selective non-retryable-exception support"*). Harmless at the default (`retries_allowed` off), but a user who enables step retries gets `UserError` retried, and nothing user-visible says so. Bucket **deliberate divergence, undocumented** — worth a sentence in `docs/durable_execution/dbos.md` next to the existing "No retries if omitted" notes.

**Verified how:** read-provable (source comparison). Not probed.

---

## 2. `override(tools=...)` inside a flow is rejected with a message about `run(toolsets=...)`, and only by accident

**Bucket: unknown intent.**

```
plain in flow          -> 'done'
override(tools=[...])  -> UserError: FunctionToolset cannot be passed to `run(toolsets=...)` at runtime with Prefect,
                          because toolsets that execute their own tools or resolve dynamically must be registered ...
```

Three problems:

1. **Wrong message.** The user wrote `override(tools=[...])`; the error names an API they didn't call and advises "pass them to the agent constructor instead", which is not the fix for a `tools` override.
2. **It fires by accident.** `_build_toolset_list` (`agent/__init__.py:2911`) constructs a fresh `_AgentFunctionToolset` per call, so its identity is never in `_reject_runtime_toolsets`' `construction_leaves` set and it trips the runtime-toolset guard. Nothing decided that `override(tools=...)` should be rejected; it falls out of object identity.
3. **It is a behaviour tightening against the wrapper.** `tests/test_prefect.py:1116` `test_prefect_agent_override_tools` asserts that `override(tools=...)` **succeeds** inside a flow on the `PrefectAgent` path.

Worth deciding deliberately, because the current state is inconsistent in the dangerous direction: `override(tools=...)` (harmless — the tools would still be wrapped if the guard recognised them) is rejected, while `override(toolsets=...)` (silently un-durables the run, #6908) is allowed.

**Verified how:** probed against real Prefect 3.7.7; the `test_prefect_agent_override_tools` assertion is read-provable.

---

## 3. Per-tool durable config is silently ignored in three places; one of them is contradicted by the docs

**Bucket: (a) for Prefect MCP — the docs promise it works. (b) undocumented for DBOS.**

Current state, `main` @ 855a8e6cb:

| Engine | Function tools | Dynamic tools | MCP tools |
|---|---|---|---|
| Temporal | honored; `False` + sync tool → `UserError` (`temporal/_function_toolset.py:53-62`) | honored, but the config is degraded by the JSON round trip (#6896); `False` → `UserError` (`temporal/_dynamic_toolset.py:110-119`) | honored; `False` → `UserError` (`temporal/_mcp_toolset.py:39-44`, `:76-85`) |
| DBOS | n/a (function tools run inline) | **silently ignored** `dbos/_dynamic_toolset.py:59` | **silently ignored** `dbos/_mcp_toolset.py:72` |
| Prefect | honored `prefect/_function_toolset.py:73` | honored `prefect/_dynamic_toolset.py:64` | **silently ignored** `prefect/_mcp_toolset.py:94` |

### 3a. Prefect MCP tools — actively contradicted by the docs

`prefect/_mcp_toolset.py:94`:

```python
resolve_tool_config=lambda tool, name: {},
```

so `tool.tool_def.metadata['prefect']` is never read for an MCP tool. And the `config` argument is discarded a **second** time at `prefect/_mcp_toolset.py:46`:

```python
task_config = with_non_retryable_errors(base_config)   # no `| dict(config)`
```

Compare Prefect's own function-tool and dynamic-tool operations, which do merge it (`_function_toolset.py:42`, `_dynamic_toolset.py:50`: `base_config | dict(config)`), and Temporal's MCP operation, which merges it too (`temporal/_mcp_toolset.py:114`).

`docs/durable_execution/prefect.md:169` promises this works, with no MCP carve-out:

> *"A default `TaskConfig` for all tools can be passed as `tool_task_config` … **Per-tool config lives on the tool's `metadata` field — `PrefectDurability` looks for a `'prefect'` key.** You can set the metadata directly on the tool definition, or apply it across a selection of tools via the `SetToolMetadata` capability."*

So a user who sets `metadata={'prefect': TaskConfig(timeout_seconds=30)}` on an MCP tool — or `metadata={'prefect': False}` to opt it out — gets silence. Temporal raises a `UserError` for the `False` case specifically because MCP tools require I/O and cannot run outside a unit; Prefect accepts the opt-out and ignores it.

### 3b. DBOS — undocumented rather than misdocumented

Both DBOS resolvers return `{}` with a code comment (`dbos/_dynamic_toolset.py:58`, `dbos/_mcp_toolset.py:71`: *"DBOS takes no per-tool config; tool metadata is ignored"*). `docs/durable_execution/dbos.md` never mentions per-tool config at all, so nothing is contradicted — but a user coming from the Temporal or Prefect page has no way to learn that `metadata={'dbos': ...}` (or any per-tool config) is not a thing on DBOS. `DBOSDurability` also leaves `_tool_config_key` at its `None` default, so the runtime-toolset rejection message omits the opt-out sentence that the other two engines emit.

Minimum fix for both: say so in the docs. Better: make the ignore loud — a `UserError` when a tool carries a durable-config key the engine cannot honor, which is option 3 in #6896 and would cover all three cells above with one mechanism.

**Verified how:** read-provable in all three cells — the `lambda tool, name: {}` resolvers, the discarded `config` argument, and the docs line are plain source facts. **I did not spin up an MCP server to observe the Prefect no-op end to end**, so I am claiming the code path and the docs contradiction, not a captured incident. The Temporal side of the comparison is already reproduced in #6896.

---

- Python: 3.14.3
- Pydantic AI: 2.19.1.dev2+855a8e6cb (`main`)
- LLM provider SDK: n/a (`FunctionModel`); prefect 3.7.7, dbos on SQLite

### Related

- #6896 — Temporal per-tool `ActivityConfig` silently ignored for `DynamicToolset` tools. Same class as item 3; that issue's option 3 ("reject it loudly") would fix all of these together, and its option 2 (carry the durable config as a typed field on `DynamicToolInfo`) is the shared-base version.
- #6908 — `override(toolsets=...)` silently un-durables the run; the inconsistency item 2 is measured against.

DouweM:
- Closes #6912

Three cross-engine inconsistencies from the durable-exec parity audit. Two are fixed here; the third is a **question for a maintainer** rather than a change (see below).

### 1. Prefect's non-retryable set was missing `PydanticUserError`

Temporal marks three framework error types non-retryable — `UserError`, `PydanticUserError`, `UnexpectedModelBehavior` — on every activity config and again at the workflow level. Prefect's identically-named `with_non_retryable_errors` checked only two, so a `PydanticUserError` (a schema-construction error that retrying can't fix) burned the configured retry budget. Plain drift against the intent the comment above `prefect/_durability.py:110` already states. One line, plus two assertions in the existing `test_prefect_with_non_retryable_errors_condition`.

**Verified:** read-provable, and I confirmed it on current `main` before changing anything — the `isinstance` tuple really did have two entries.

### 2. `override(tools=...)` inside a flow — a question, not a change

I reproduced this against real Prefect on the capability path:

```
plain in flow          -> 'success (no tool calls)'
override(tools=[...])  -> UserError: FunctionToolset cannot be passed to `run(toolsets=...)` at runtime with
                          Prefect, because toolsets that execute their own tools or resolve dynamically must be
                          registered ... Pass them to the agent constructor instead.
override outside flow  -> 'success (no tool calls)'
```

and confirmed the two read-provable parts:

- `_build_toolset_list` constructs a fresh `_AgentFunctionToolset` per call, so its identity is never in `_reject_runtime_toolsets`' `construction_leaves` set. Nothing decided that `override(tools=...)` should be rejected — it falls out of object identity.
- `tests/test_prefect.py::test_prefect_agent_override_tools` asserts that `override(tools=...)` **succeeds** inside a flow on the deprecated `PrefectAgent` path. Making the capability path's rejection deliberate would be a **behaviour tightening against an existing passing test**, which is why I have not touched it.

Also worth noting for whoever decides: Temporal has the same `'function'` in `_unsupported_runtime_toolset_kinds` and so behaves the same way; DBOS omits `'function'` and allows it.

**@DouweM — which of these do you want?**

- **(a) Allow it.** Teach `_reject_runtime_toolsets` to recognise the per-run `_AgentFunctionToolset` as a construction leaf, so `override(tools=...)` passes the guard. The tools would still be durably wrapped, and `test_prefect_agent_override_tools` keeps passing on both paths.
- **(b) Reject it deliberately, with an accurate message.** Keep the rejection but say `override(tools=...)`, not `run(toolsets=...)`, and drop the "pass them to the agent constructor instead" advice, which isn't the fix for a `tools` override. This tightens behaviour against `test_prefect_agent_override_tools`, so that test would need updating.

I didn't guess, partly because the current state is inconsistent in the dangerous direction — `override(tools=...)` (harmless) is rejected while `override(toolsets=...)` (silently un-durables the run, #6908) is allowed — and partly because #6908 is being worked on separately and the two decisions should agree.

### 3. Per-tool durable config that was silently ignored

**Prefect MCP tools — the docs promised this worked.** `docs/durable_execution/prefect.md:169` says "Per-tool config lives on the tool's `metadata` field — `PrefectDurability` looks for a `'prefect'` key", with no MCP carve-out, but `prefectify_mcp_toolset` passed `resolve_tool_config=lambda tool, name: {}` and `_call_tool_operation` discarded the resolved `config` a second time when building the task. Both are fixed, matching Prefect's own function-tool and dynamic-tool operations (`base_config | dict(config)`).

`False` now opts an MCP tool out of task wrapping and runs the call inline in flow code. That deliberately diverges from Temporal, which raises for the same input: Temporal's reason is that "MCP tools require the use of IO and so cannot be run outside of an activity" — a workflow-determinism constraint that doesn't apply to a Prefect flow, which can do I/O itself. This also makes MCP tools behave like Prefect's function and dynamic tools, which already honour `False`. The `DurableMCPToolset.call_tool` inline branch existed already and was `# pragma: no cover`d as unreachable; it's now reachable and covered.

The deprecated `PrefectMCPToolset` is left frozen — it never read per-tool metadata, and the docs section this fixes is about `PrefectDurability`.

**DBOS — undocumented rather than misdocumented.** DBOS genuinely takes no per-tool config and has no selective non-retryable-exception support. Both were recorded only in code comments (`dbos/_dynamic_toolset.py:58`, `dbos/_mcp_toolset.py:71`, `dbos/_utils.py`), so a user arriving from the Temporal or Prefect page had no way to learn it. Two sentences added to `docs/durable_execution/dbos.md`, next to the existing step-config and step-retry notes.

**Verified:** the issue marked all of item 3 read-provable and did not spin up an MCP server. Both new tests were confirmed to fail against unpatched `main` (the retries one with the untouched `RuntimeError`, the opt-out one with `ran_in_task == [True]`), so the no-op is now reproduced, not just read.

### Tests

- `test_prefect_with_non_retryable_errors_condition` — extended to the full three-type set
- `test_prefect_mcp_tool_metadata_configures_task`
- `test_prefect_mcp_tool_metadata_false_runs_inline`

Full `tests/test_prefect.py` + `tests/test_dbos.py` pass (221 tests).

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6952?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


DouweM:
Merged current `main` in `0fe69d2d8` and preserved the approved Prefect parity changes. Thanks, Macroscope and CI Review — both approvals contained no findings requiring code changes. There are currently no Devin comments or unresolved review threads.

DouweM:
The `override(tools=...)` question in section 2 of the description is now filed separately as #6981, with the reproduction, the per-engine table, and both options — so it survives this PR merging.

Nothing in this PR's diff touches `override(tools=...)`; `git grep` over the patch returns no hits. What's here is exactly the two fixes in the title: `PydanticUserError` added to Prefect's non-retryable set, and per-tool `TaskConfig` honored on Prefect MCP tools. Both are independently mergeable and don't depend on #6954.

PR #6696, opened 2026-07-23
Add `@durable_operation` for capabilities and a public backend API for third-party durable execution engines

DouweM:
Closes #5477. Supersedes #6906 (absorbed in full, rebuilt on the new base).

## What

Rebuilds `durable_exec` around one idea: **every durable unit is a declared operation** that the base assembles and each engine binds, instead of six engines each hand-rolling naming, serialization, dispatch, and caching. On top of that core, two new extension surfaces:

- **Capabilities contribute durable operations.** Mark an async method with `@durable_operation` and calling it normally runs it as an activity/step/task under any durability engine, with parameter/result serialization derived from type annotations the same way `@agent.tool` does it. Without a durability engine, calls pass straight through. Marking a base hook declaration makes every override durable through the MRO scan; subclasses repeat nothing. A method that declares a `ModelRequestContext`-annotated parameter gets the model round-trip treatment by TYPE, so custom methods extracted from `before_model_request` work identically. Acceptance proof: making the harness `SummarizingCompaction`'s LLM summarization replay-safe is 3 lines ([harness #707](https://github.com/pydantic/pydantic-ai-harness/pull/707)). (A dynamic declaration flow existed in earlier revisions and was cut on review: decorator-only ships now.)
- **Third-party durable execution engines get a public API.** Everything an engine needs is exported from `pydantic_ai.durable_exec`, and nothing an engine overrides is private: subclass `BaseDurabilityCapability`, set one declarative `engine_spec` (`DurabilityEngineSpec`: codec, toolset lifecycles, wrapped kinds, engine vocabulary), and override `get_durable_operation_backend()` returning one small backend class, either `CallableOperationBackend` (implement `execute`: run this body in a named durable unit) for flow/journal engines, or `RegisteredOperationBackend` (implement `register`) for Temporal-style pre-registration. The base owns naming, config resolution, cache identity, and serialization. For the registered tier the base binds the four model operations during agent assembly, so `registrations()` is complete before a worker can start, and unknown future operation ids bind through a generic fallback, so new operation families (e.g. sandboxes, #6492) need zero per-engine edits. Acceptance proof: the three out-of-tree harness engines (#443 Lambda, #444 Restate, #445 Absurd) import **zero** private names. Guide: `docs/durable_execution/backends.md`.

[...]

PR #6955, opened 2026-07-30
Reject per-run `capabilities=` on DBOS, matching Temporal

DouweM:
Split out of #6894, which bundled this with a custom-toolset-leaf change you weren't sold on. The docs half is #6956.

No existing issue — found by auditing the durable-execution integrations against the rule that anything silently skipping durability treatment is a bug.

### The defect

`_validate_runtime_capabilities` was implemented on `TemporalDurability` and nowhere else, so `agent.run(capabilities=[...])` inside a `@DBOS.workflow` was silently accepted even though DBOS registers its steps while binding — the capability's toolsets and its own `@durable_operation` methods had no registered steps.

### Which engines this applies to, and why

#6696 split engines into two backend tiers, and that split is exactly what this check depends on:

| Engine | Backend | Registers units up front? |
|---|---|---|
| Temporal | `RegisteredOperationBackend` | Yes — activities registered with the worker |
| DBOS | `RegisteredOperationBackend` | Yes — *"Register typed durable-operation handlers as DBOS steps during agent binding"* |
| Prefect | `CallableOperationBackend` | **No** — `bind()` registers nothing; `execute()` builds the `@task` per call |

So the check is gated on `isinstance(self.get_durable_operation_backend(), RegisteredOperationBackend)` rather than applied to every engine. That's derived rather than declared: `_base.py` already uses this same isinstance test, and a new `DurabilityEngineSpec` field would be one more thing a third-party engine has to set, with a default that would be wrong for the callable journal engines (Restate, Lambda and Absurd are all callable-tier).

**On Prefect nothing changes.** A per-run capability contributing an executing toolset is still rejected — by `_reject_runtime_toolsets`, which already lists `function`, `mcp` and `dynamic` as unsupported runtime kinds there. What's now accepted is a capability that contributes no executing toolset. Note that includes fewer cases than it sounds: the concrete `Capability` class always carries a `FunctionToolset`, empty or not, so in practice this is `AbstractCapability` subclasses.

### The message was blaming the wrong thing

The previous wording said a per-run capability "would run its hooks in workflow code instead of durable activities". That isn't the defect — capability hooks run in workflow/flow code by design, construction-time ones included. The message and docs now lead with registration timing instead, which is the thing that actually breaks.

### Also here

`durable_unit_plural` on `DurabilityEngineSpec`, because the shared wording rendered "activitys" for Temporal. Orthogonal but adjacent.

### Behaviour change worth seeing plainly

**`run(capabilities=[...])` inside a DBOS workflow now raises where it was previously accepted.** Temporal already rejected it; Prefect is unaffected. That's the intended alignment — the accepted-but-broken DBOS behaviour produced un-registered steps — but it can break code that "worked", so it should be an explicit decision. I've left the breaking-changes box unchecked for that reason.

### Tests

Temporal's `test_durability_validates_only_resolved_runtime_capability_layers` passes unchanged, which is the useful signal that moving the check off `TemporalDurability` preserved its behaviour. DBOS keeps its rejection tests. Prefect gains one test that a hook-only per-run capability runs to completion inside a flow, and one that a toolset-contributing one is still rejected by the toolset guard.

`tests/durable_exec/test_prefect.py` 141 passed / 6 failed, `test_dbos.py` 139 passed. The 6 are pre-existing Prefect result-persistence failures in this environment (cache/replay/event-dispatch tests); I baselined them by reverting `_base.py` and they fail identically without this change. Temporal tests need the `.venv313` + full `--all-extras` setup and were not run locally — CI covers them.

Docs: "Capabilities at Runtime" reworded in `temporal.md` and `dbos.md`, and rewritten in `prefect.md` to say what Prefect actually does. Skill reference updated to match.

### Checklist

- [ ] Any permitted **compatibility impact** has a label, warning, migration, release note, and exact API-check waiver.


DouweM:
Addressed the CI Review description nits: the body no longer claims DBOS/Prefect had byte-identical prior rejection text, and now accurately says the shared wording preserves Temporal’s prior rejection semantics while explaining the registration boundary consistently across all three engines. The merge with `main` also preserves the reviewed behavior alongside #6943’s guarded DBOS/Prefect model units and #6947’s Temporal `ActivityConfig` restoration. 👍

Issue #6920, opened 2026-07-30
customize_request_parameters transforms both output_tools and output_object, then discards one — and never memoizes the walk


ZacharyHampton:
### Description

`Model.customize_request_parameters` transforms **both** `output_tools` and `output_object` through the profile's `JsonSchemaTransformer`, and `prepare_request` then discards whichever one the resolved output mode doesn't use. So every model request pays the transform twice and throws one result away. On top of that, the transform is a pure function of `(transformer, schema, strict)` and is never memoized, so a run with a fixed output schema recomputes an identical value on every request.

Verified on installed 2.20.0 (`models/__init__.py`):

```python
# :385-395 — both halves transformed
if transformer := self.profile.get('json_schema_transformer'):
    model_request_parameters = replace(
        model_request_parameters,
        function_tools=[_customize_tool_def(transformer, t) for t in ...],
        output_tools=[_customize_tool_def(transformer, t) for t in ...],
    )
    if output_object := model_request_parameters.output_object:
        model_request_parameters = replace(
            model_request_parameters,
            output_object=_customize_output_object(transformer, output_object),
        )

# :438-441 — and one is then discarded
if params.output_tools and params.output_mode != 'tool':
    params = replace(params, output_tools=[])
if params.output_object and params.output_mode not in ('native', 'prompted'):
    params = replace(params, output_object=None)
```

`_customize_tool_def` (`:1527`) and `_customize_output_object` construct a fresh transformer and call `walk()` per invocation — no cache anywhere on that path.

### Why it matters at scale

For static agents this is microseconds. For a dynamically-built output schema it was the most expensive thing in our workflow process.

Measured on a ~100 KB `final_result` schema (12 sections × 93 per-item fields) whose per-section union is shared by `$ref`, on a vendor profile carrying `prefer_inlined_defs` (which expands every `$ref`):

| | value |
|---|---|
| `transform` calls per walk | **69,258** (vs 765 uninlined) |
| per model request (two walks) | **138,516** |
| observed in our staging profiler | **138,527** — the extra 11 are two small orchestration tool schemas |
| wall time | **32.2 s**, across three consecutive activations |

Those activations each blew past our 8 s workflow-task deadlock ceiling, which is how we found it.

### The durable case is strictly worse

[...]

PR #6521, opened 2026-07-15
fix(anthropic): transform native output schema when reached via `profile={"default_structured_output_mode": "native"}`

dsfaccini:
This pull request was posted by Claude Code using claude-sonnet-5 on behalf of David.

- Closes #6471

## What

`AnthropicModel.prepare_request` force-sets `strict=True` on the output object when `output_mode == 'native'`, which makes `customize_request_parameters()` (called by the base `Model.prepare_request()`) apply Anthropic's `transform_schema()` and strip constraints (`minimum`, `maximum`, `minLength`, etc.) that Anthropic's schema dialect doesn't support.

That check ran *before* the base class resolved `output_mode='auto'` to the profile's `default_structured_output_mode`. An explicit `NativeOutput(...)` already has `output_mode='native'` at construction time, so it was unaffected. But `profile={'default_structured_output_mode': 'native'}` left `output_mode` as `'auto'` at the point of the check, so `strict` stayed `None`, the transformer left the schema untouched, and Anthropic rejected the raw Pydantic constraints with a 400.

The fix resolves `'auto'` to the profile default right after the existing thinking-specific output-mode resolution (which must still take precedence — thinking forces `native`/`prompted` over the profile default to avoid `tool_choice=required`, which Anthropic doesn't support with thinking) and before the strict-forcing check, mirroring the precedent already established in that same function.

## Testing

- `tests/models/anthropic/test_output.py::test_native_output_profile_default_transforms_schema` — mock-based unit test asserting the schema sent for profile-default native mode has constraints transformed away (`additionalProperties: False`, no raw `minimum`/`maximum`/`minLength`).
- `tests/models/anthropic/test_output.py::test_no_tools_profile_default_native_output` — VCR round-trip test against the live API reproducing the reporter's MRE (constrained `Decision` model, `profile={'default_structured_output_mode': 'native'}`), confirming no 400.

### Checklist

- [ ] Any **AI generated code** has been reviewed line-by-line by the human PR author, who stands by it.
- [x] No **breaking changes** in accordance with the [version policy](https://github.com/pydantic/pydantic-ai/blob/main/docs/version-policy.md).
- [x] **PR title** is fit for the [release changelog](https://github.com/pydantic/pydantic-ai/releases).

[...]

Issue #6948, opened 2026-07-30
Inlined `$defs` silently become `{}` for the second and later `$ref` to a union definition


DouweM:
### Initial Checks

- [x] I confirm that I'm using the latest version of Pydantic AI

### Description

`InlineDefsJsonSchemaTransformer` (and any profile with `prefer_inlined_defs=True`) **silently sends `{}` to the model** for the second and every subsequent `$ref` to a `$def` whose schema is a union (`anyOf` / `oneOf` / `allOf`).

`{}` is a valid JSON Schema meaning "anything", so the model receives no type information at all for those fields. There is no error, no warning, and nothing in the response indicates the schema was truncated — the model just gets a worse schema than the one you declared, and validation failures show up later as `ModelRetry` churn.

### Root cause

Two mutations combine, in `pydantic_ai_slim/pydantic_ai/_json_schema.py`:

1. `JsonSchemaTransformer._handle` binds `schema` to the **live** `self.defs[key]` object when the `$ref` has no sibling keywords:

   https://github.com/pydantic/pydantic-ai/blob/main/pydantic_ai_slim/pydantic_ai/_json_schema.py#L109-L110

   ```python
   siblings = {k: v for k, v in schema.items() if k != '$ref'}
   schema = {**def_schema, **siblings} if siblings else def_schema   # <-- no copy
   ```

2. `_handle_union` then **pops** the union key off that live object:

   https://github.com/pydantic/pydantic-ai/blob/main/pydantic_ai_slim/pydantic_ai/_json_schema.py#L164-L167

   ```python
   def _handle_union(self, schema, union_kind):
       try:
           members = schema.pop(union_kind)   # <-- permanently strips it from self.defs[key]
       except KeyError:
           return schema
   ```

So the first `$ref` inlines correctly and destroys the definition as a side effect; every later `$ref` to the same `$def` finds an empty dict and inlines `{}`.

`_handle_object` mutates in place too (`schema['properties'] = handled_properties`), which is why inlined object subtrees also come out object-aliased — the same `dict` instance appears at multiple positions in the output. That aliasing is not currently harmful, but it is the same missing-copy bug and it constrains the fix.

### Example Code

```python
from typing import Union
from typing_extensions import TypeAliasType
from pydantic import BaseModel
from pydantic_ai import InlineDefsJsonSchemaTransformer
import json


class Cat(BaseModel):
    meow: str


class Dog(BaseModel):
    woof: str


[...]

PR #6963, opened 2026-07-30
Resolve the output mode before `customize_request_parameters` transforms JSON schemas

DouweM:
Part of #6920 — the smallest of the four levers identified in [the investigation comment](https://github.com/pydantic/pydantic-ai/issues/6920#issuecomment-5136186866). The other three (intra-walk memoization, cross-request memoization, and the `Instrumentation` double-prepare) are deliberately left open.

### The problem

`Model.prepare_request` called `customize_request_parameters()` — which walks every tool and output JSON schema through the profile's `JsonSchemaTransformer` — and only *then* resolved an `'auto'` output mode and dropped whichever of `output_tools` / `output_object` the resolved mode doesn't use.

The agent graph always populates *both* channels while `output_mode` is still `'auto'`, so exactly one of the two output schemas was transformed and immediately thrown away, on every single request.

### The fix

```
prepare_request:
    merge settings
    resolve output mode + drop the losing channel   # hoisted above
    customize_request_parameters()                  # transforms only what will be sent
    ...
    resolve output mode + drop the losing channel   # KEPT — now normally a no-op
```

### Why the late resolve/reset block is deliberately retained

Both `prepare_request` and `customize_request_parameters` are public and overridable. Nothing in this repo changes the output mode inside `customize_request_parameters` — `AnthropicModel`, `BedrockConverseModel` and `GoogleModel` all resolve it in their own `prepare_request` override *before* calling `super()`, and `CerebrasModel`/`ZaiModel`/`OpenRouterModel` call `super()` first and return its parameters unmodified — but a third-party model may.

Keeping the second pass means such an override still can't produce parameters where `output_mode`, `allow_text_output` and the surviving output channel disagree. The cost is two idempotent dataclass `replace`s per request; the resolve/reset logic is now a single `_resolve_output_mode` helper called from both places.

`customize_request_parameters`' docstring now says not to change `output_mode` there, and points at the `prepare_request`-override pattern the three in-repo models use.

### Measured

Counting `JsonSchemaTransformer.walk()` calls for a request with one function tool and a structured output type:

| | before | after |
|---|---|---|
| per `prepare_request` (`tool`, `native` or `prompted` profile default) | 3 walks / 9 `transform` calls | **2 walks / 6 `transform` calls** |
| per model request under `Instrumentation()`, which prepares twice (Finding 1 in #6920, not addressed here) | 6 walks | **4 walks** |

### Tests

- `test_prepare_request_transforms_only_the_output_channel_the_resolved_mode_uses` — a recording `JsonSchemaTransformer` proves that a profile-defaulted `native`/`prompted` mode never walks the output tool's schema, and a `tool` mode never walks the output object's. Fails on `main` for all three modes.
- `test_prepare_request_re_resolves_output_mode_changed_by_customize_request_parameters` — a model whose `customize_request_parameters` *does* change the output mode (switching to `prompted`, and resetting to `'auto'`) still ends up with consistent parameters. Both cases fail if the retained late block is deleted, so the safety property is genuinely covered.
- The existing #6521 / #6539 regression tests (Anthropic and Bedrock with `profile={'default_structured_output_mode': 'native'}`) still pass; those hoists run before `super()` entirely and `with_default_output_mode` is idempotent.

- Part of #6920

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6963?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/__init__.py
+++ b/pydantic_ai_slim/pydantic_ai/models/__init__.py
@@ -396,6 +396,13 @@ class Model(ABC, Generic[InterfaceClient]):
         This method can be overridden by subclasses to modify the request parameters before sending them to the model.
         In particular, this method can be used to make modifications to the generated tool JSON schemas if necessary
         for vendor/model-specific reasons.
+
+        Don't change `output_mode` here: `prepare_request` resolves it before calling this method, so that only the
+        schemas the resolved mode actually sends are transformed. A `prepare_request` override that resolves the mode
+        before calling `super().prepare_request()` is the place for that, as
+        [`AnthropicModel`][pydantic_ai.models.anthropic.AnthropicModel],
+        [`BedrockConverseModel`][pydantic_ai.models.bedrock.BedrockConverseModel] and
+        [`GoogleModel`][pydantic_ai.models.google.GoogleModel] do.
         """
         if transformer := self.profile.get('json_schema_transformer'):
             model_request_parameters = replace(
@@ -426,7 +433,11 @@ class Model(ABC, Generic[InterfaceClient]):
         """
         model_settings = merge_model_settings(self.settings, model_settings)
 
-        params = self.customize_request_parameters(model_request_parameters)
+        # Resolve the output mode and drop the losing channel *before* `customize_request_parameters` runs, so a
+        # JSON schema transformer only walks the output schema the resolved mode actually sends, instead of both.
+        params = _resolve_output_mode(model_request_parameters, self.profile)
+
+        params = self.customize_request_parameters(params)
         params = _prepare_return_schemas(params, self.profile)
 
         # Resolve unified thinking setting and strip from model_settings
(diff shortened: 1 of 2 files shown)


DouweM:
Closing this without merging. The optimization is sound in principle but fundamentally in tension with `customize_request_parameters` being a public, overridable hook, and it is the smallest of the four levers on #6920.

### What we tried, and why each version was wrong

**v1 — resolve the mode and drop the losing channel before `customize_request_parameters`.** Macroscope caught this: `_resolve_output_mode` drops but never rebuilds, so an override that changes only `output_mode` receives `output_object=None` for the mode it just selected. Prompted mode silently loses its schema instructions. The test on this branch missed it because the fixture re-supplied `output_object` itself.

**v2 — resolve non-destructively before the hook, transform only the selected channel, drop after.** Better, but still a regression against `main`. Verified both ways with an override that changes only the mode after calling `super().customize_request_parameters()`:

```
main:      output_object = {'type': 'object', 'properties': {}, 'x-transformed': True}
v2 branch: output_object = {'type': 'object', 'properties': {}}
```

The channel survives but was never transformed, because the skip decision was made against the pre-override mode.

### Why not a v3

Skipping the transform requires knowing the final output mode, but the only hook that can change the mode runs afterwards. Re-transforming when the mode changed needs an "already transformed" signal we don't have — a subclass that overrides `customize_request_parameters` without calling `super()` would get double-transformed. The alternative, moving the schema transform out of `customize_request_parameters` and into `prepare_request` after the final resolve, is a public behaviour change for every subclass that calls `super()`.

None of that is worth it for this particular win.

### Where the actual cost went instead

Measured on #6920, this reorder was the smallest of four levers — 3 → 2 walks per `prepare_request`.

- The dominant one, the ~62× intra-walk redundancy, is fixed in #6960 (1,037× fewer `transform` calls than a naive correct inline, and it fixes the #6948 corruption at the same time).
- The next-biggest is the double `prepare_request` under instrumentation: `capabilities/instrumentation.py:312` passes the *unprepared* context to the handler after `open_model_request_span` already prepared it at `_instrumentation.py:321`, while `models/instrumented.py:333-335` forwards the prepared one. That's 4 walks per request instead of 2, entirely in the workflow process on the durable path — where it's charged against Temporal's deadlock-detection budget.
- Cross-request memoization remains open, and needs `JsonSchemaTransformer.transform` declared contractually pure plus a bounded cache.

Worth revisiting only if those don't move the reported numbers enough.

---

_Analysis by @DouweM's coding agent (Claude Code); both behaviours above reproduced against `main` @ `270ef2b94` and this branch._

PR #6984, opened 2026-07-31
Preserve `SystemPromptPart` semantics across caching and Temporal

dsfaccini:
This pull request was posted by Codex Desktop using gpt-5.6-sol on behalf of David.

Follow-up to #6765, fixing two regressions it introduced. Both fixes are demonstrated by red-before-green tests.

## Anthropic: cache boundaries around mid-conversation `SystemPromptPart`s

A `CachePoint` whose preceding parts include an inline mid-conversation `SystemPromptPart` no longer silently narrows or widens the requested cache prefix:

- When the boundary is representable natively, the instruction stays a real system message and the cache boundary lands exactly where it was requested.
- When it isn't, the existing tagged-user-part fallback is used — with the boundary kept exact rather than dropped.
- Tool results that run in parallel keep their original ordering relative to cache boundaries.
- A boundary that no Anthropic request shape can represent (e.g. between a tool call and its required result) now raises an explanatory `UserError` locally instead of letting the provider reject the request.

The new integration test records a cache write and a subsequent cache read for the same prefix (`test_inline_system_prompt_cache_prefix_is_reused`), proving the prefix is actually reusable on the wire.

## Temporal: runtime model strings prepare history against the wrong profile

`prepare_messages` decides from the model's profile. For a *registered* model the workflow side holds the same instance as the activity, so its pass sees the right profile — but for an unregistered runtime model **string**, the workflow infers a profile without constructing a provider, and the concrete client can narrow it (e.g. `AsyncAnthropicFoundry` turns off `supports_inline_system_prompts`). Whatever the workflow prepared was decided against a profile the request is not sent under.

For unregistered runtime strings, the workflow-side `prepare_messages` pass is now the identity, and both the request and streaming activities run the single authoritative preparation against the concrete resolved model, followed by the same history-cleanup pass the agent graph applies — so the durable path renders history identically to the non-durable path. A single pass also removes the irreversible-downgrade hazard of preparing twice under two profiles: a mid-conversation `SystemPromptPart` wrapped into the tagged-user fallback by a wrongly-inferred profile could never be restored to a native system message by the activity.

This gap is exclusive to the legacy `TemporalModel` string+registry path: the durability capabilities (`TemporalDurability`, DBOS, Prefect) resolve a runtime string to the concrete model *before* graph preparation through the shared `BaseDurabilityCapability` contract, so their preparation pass already sees the right profile. `prepare_request` needs no equivalent treatment — the workflow side's prepared settings/parameters are used only for telemetry; the activity payload carries the unprepared originals, which the concrete model prepares once.

## Not in this PR

- The suspended-response cancellation activity still resolves its model without `deps`/`run_context`, so a `provider_factory` isn't consulted there. Closing that requires a second activity argument, which changes the recorded activity command and needs a story for in-flight workflows — tracked in #6992 (the code comments point there).

### Checklist


DouweM (commented):
Thanks for this — the gap is real and it's one I left in #6765. A `CachePoint` means "cache everything up to here", and #6765 slides a `{'role': 'system'}` entry forward past user turns to keep it legal, so the boundary the user asked for and the boundary that lands on the wire can genuinely disagree. Worth fixing.

I'd like to push back on *how*, though, because the implementation reverses three things #6765 settled deliberately.

### 1. It reinstates the `<system>` rendering inside the adapter — in three places

`models/anthropic.py` now builds `f'<system>{...}</system>'` at the pre-scan fallback and twice more in the reconciliation loop. #6765 removed exactly that: the tagging lives in `Model.prepare_messages` (`_wrap_non_leading_system_prompts`), and the adapter narrows `supports_inline_system_prompts` in `AnthropicModel.profile` so parts that can't be served never reach `_map_message` at all. That was a review request on #6765 precisely so there's one place to change if the format ever moves.

After this PR there are four sites, three of which are Anthropic-specific. If the fallback shape has to be reachable from inside the adapter, I'd rather export the single renderer from `models/__init__.py` and call it than re-inline the f-string.

### 2. The parallel-index data model comes back

`mid_conversation_system_positions: list[tuple[int, str]]` is the `system_message_indexes` structure #6765 removed for being hard to follow across ~400 lines of mapping — it was replaced by a backwards scan for `role == 'system'`, since the mapping loop is the only thing that emits that role. This PR adds it back and then adds a second one: `cached_inline_system_messages`, a list of 4-tuples, reconciled afterwards by identity (`next(... if message is system_message)`) with `assert`s on list positions.

That's a lot of state to keep in step, and the `assert anthropic_messages[system_index - 1] is user_message` is load-bearing in production rather than a test. Could the same decision be made during mapping, where the parts are still in hand?

### 3. It trades the part's authority for the cache boundary, which inverts the documented rule

This is the one I'd most like to talk through. `ModelProfile.supports_inline_system_prompts` says:

> Preserving the part's authority is worth more than preserving the exact position it was authored at — an instruction only governs the generation that follows it, and that's the same generation either way — so prefer adjusting placement over falling back to the `<system>...</system>` rendering, which the model reads as user-authored.

This PR makes the presence of a `CachePoint` silently flip a mid-conversation instruction from a real `system` entry to user-authored text. That's measurable, not theoretical: on the recordings behind #6765, asked to lift a restriction the top-level prompt set, Opus 5 acts on the native entry and refuses the identical text in a `<system>` tag. So adding a cache marker can change whether the model obeys the instruction — a surprising coupling between two unrelated features.

Two alternatives that keep authority:

- **Don't slide an entry that carries a cache point.** The slide exists to satisfy "a `system` entry must precede an assistant turn"; the anchor mechanism already handles the other rule. If the entry stays where it was authored, the boundary is exact and it stays a system message.
- **Put `cache_control` on the entry and accept the sliding boundary**, documenting that the marker covers everything before the generation it governs. The PR already does this in the non-sliding case, and Anthropic accepts `cache_control` on the entry — so the only open question is whether the slid position is an acceptable reading of the user's intent. I'd argue it is, for the same reason the profile doc gives: the instruction governs the same generation either way.

Either keeps one rendering path and drops most of the reconciliation machinery.

### Smaller things

- `_INLINE_SYSTEM_TOOL_PAIR_CACHE_POINT_ERROR` is raised from two places with the same text but genuinely different causes (pre-scan detection vs. unmatched `tool_use` ids after reconciliation). Worth distinguishing, since the second is a shape the user can't easily reason about from the message.
- The `assert not isinstance(original_user_content, str)` and the identity `assert` should be real errors or unreachable-by-construction, not bare asserts on a request path.

### Not blocking

The Temporal half (re-preparing serialized history against the concrete provider profile, and resolving suspended-response cancellation through the same runtime provider) looks right to me and is independent of the above — it'd be easy to split out if the Anthropic half needs another round.

Happy to be wrong on any of this, particularly (3) — if there's a case where the sliding boundary is genuinely incorrect rather than just imprecise, that changes the calculus.

DouweM:
I've pushed a rework of the Anthropic half to this branch (`37d35c3`). The regression you found is real and worth fixing — a `CachePoint` that ends a request stopped covering the instruction once #6765 moved the instruction into its own entry. I kept that diagnosis and your `test_inline_system_prompt_cache_prefix_is_reused` recording, which is the piece that proves Anthropic actually honors a boundary on a `system`-role block.

What changed is how the cases the native placement rules can't express exactly are handled. The branch resolved them by re-rendering the instruction as `<system>`-tagged user text. That inverts what `supports_inline_system_prompts` promises: on a model and transport that serve the role, a performance hint would decide whether an operator instruction keeps operator authority. The recordings in `test_mid_conversation_system_prompt` measure that difference — given the native entry Claude lifts a restriction its top-level prompt set, and given the identical text in a `<system>` tag it refuses. A `CachePoint` shouldn't be able to flip that.

The rule now is exact-or-narrow, and the instruction never changes form:

- A marker that ends the request lands on the entry's final block, so the instruction is inside the boundary. (Your fix, kept.)
- A marker with content after it stays on the block it followed. The entry renders after that content, so a boundary covering the instruction would have to cover the content the marker was placed in front of. The instruction falls outside instead: the cached prefix is always a prefix of what preceded the marker, never more.
- An entry that has to slide past a later user turn leaves its boundary behind on the message it was sitting behind, rather than dragging it over turns nobody marked.

That removes `_INLINE_SYSTEM_TOOL_PAIR_CACHE_POINT_ERROR`, the `cached_inline_system_messages` reconciliation pass and its identity lookups, and the three `<system>` f-strings — about 100 lines, and it leaves `_map_message` with one rendering path again.

On the tool-pair `UserError` specifically: the shape it guarded isn't caused by a cache boundary. `[assistant(two tool_use), request([ToolReturnPart(call_1), UserPromptPart([CachePoint()])]), request([ToolReturnPart(call_2)])]` renders to two consecutive single-result user messages on `main` today with no `SystemPromptPart` anywhere — I checked. If that split is invalid it's a pre-existing bug in how consecutive `ModelRequest`s render, and it deserves its own issue rather than an error raised only when an inline instruction happens to be nearby.

Two things I'd still like your read on:

1. The Temporal half (re-preparing messages in the activity, and threading `run_context`/`deps` into the cancel activity) is a genuinely separate fix from the Anthropic one, with its own risk surface. It's easier to review, and safer to revert, as its own PR. Happy to split it out if you agree.
2. The docs said the fallback and the `UserError` were the behavior; I've rewritten those paragraphs to describe exact-or-narrow. Worth a read to check it still says what you wanted it to say.

DouweM:
Follow-up on the Temporal half (`d4ebebb`). I had it adversarially reviewed, and the split I asked about above turns out to be load-bearing rather than just tidier.

**The `prepare_messages` fix is kept, and it's the strongest red-before-green in the PR.** Reverting `_model.py` to `main` and running the new tests:

```
Snapshot mismatch:
  [{'system': 'leading\n\nmid', 'messages': [..., {'role': 'user', 'content': [{'text': 'second'}]}]}]
!=
  [{'system': 'leading',       'messages': [..., {'role': 'user', 'content': [{'text': '<system>mid</system>'}, {'text': 'second'}]}]}]
```

Without the fix the mid-conversation instruction gets hoisted into the top-level `system` parameter — which is the exact cached-prefix invalidation the whole feature exists to prevent. Under Temporal the feature was silently undoing itself. Both parametrizations are genuinely red.

I did add one thing to it. The graph re-runs `_clean_message_history` whenever `prepare_messages` returns a new list, because tool-search synthesis can split a `ModelResponse(call+return)` into a response plus a request and leave two same-role messages adjacent (`_agent_graph.py:1483`). Preparing in the activity and skipping that merge would make the durable path render history differently from every non-durable path — the opposite of what re-preparing is for. The two copies of the ternary are now one `_reprepare_messages` helper.

**The cancellation change is out, as #6992.** Not because it's wrong about the bug — it isn't, `_resolve_model_id(model_id)` with no run context really does take the no-factory branch and infer a client from the environment. The problem is the shape of the fix: it changes cancellation from scheduling one activity argument to two, and the argument count is part of the `ScheduleActivityTask` command recorded in workflow history. A workflow that started on the old code and replays on the new one sees a different command and can fail with a non-determinism error before anything reaches the worker.

The `_CancelParams | dict[str, Any]` fallback doesn't cover that. It helps the *activity worker* decode a task that was already scheduled with one argument — Temporal discards type hints when the counts disagree — but workflow replay never gets that far. Fixing it properly means either carrying the serialized context and deps inside `_CancelParams` so the command stays at one argument, or gating the new shape behind `workflow.patched(...)`.

Related: `test_temporal_model_runtime_provider_cancels_on_resolved_client` was red against old code, but for the wrong reason — old `_CancelParams` simply rejects the new `serialized_run_context=` keyword, so the red run never reached provider resolution. And its "legacy" half constructs the new typed dataclass and hands it straight to `ActivityEnvironment`, which isn't the dictionary Temporal actually produces for an old one-argument payload. Whatever lands for #6992 needs a test that replays an old history.

One thing the review cleared, since it looked adjacent to a live problem: `_cancel_params_adapter = TypeAdapter(_CancelParams)` is *not* affected by #6988. That one is about a module-level adapter over `ActivityConfig`, which newer Temporal defines as a `typing.TypedDict` and which blows up on Python 3.10/3.11. `_CancelParams` is a plain dataclass.

`tests/test_temporal.py` is 231 passed / 2 skipped on 3.13.

DouweM:
Three things about my own rework that I'd rather have written down than carried in my head. None of them block merging; all three are places where I made a judgement call that a reviewer might make differently.

**1. Exact-or-narrow narrows silently.** A `CachePoint` with content after it leaves the instruction outside the boundary, and nothing tells the caller they got a smaller cached prefix than they asked for. I chose narrowing because the alternative — letting the boundary widen to cover the instruction — caches content the marker was deliberately placed in front of, and because once the instruction renders after that content it genuinely cannot be both inside the boundary and where it was authored. But it is a silent downgrade of a performance hint, and "warn when a marker can't be honoured exactly" is a defensible alternative I did not take.

**2. `_leave_cache_boundary_behind` rides inside the placement pass.** `_place_system_messages_before_generation` now does two jobs: move the entry, and hand its `cache_control` back to the message it was sitting behind. They're coupled for a real reason — only the placement pass knows an entry moved, and it has to be decided on final positions — but it is a second responsibility inside a function whose name promises one, and the next person to touch placement has to know about it.

**3. The Temporal re-prepare is convergent, not reversible.** `_reprepare_messages` runs `prepare_messages` a second time inside the activity. That's safe for today's transforms — a second pass finds no `SystemPromptPart` left to wrap and no native tool-search parts left to translate — but it cannot *undo*. If the workflow-side pass already downgraded a system part, the activity cannot restore its authority.

It happens to work because client narrowing only ever *removes* support (`AsyncAnthropicFoundry` turns `supports_inline_system_prompts` off; nothing turns it on), so the only reachable direction is the recoverable one: string-inferred profile says supported, concrete client says no, activity downgrades. That's a property of how profile narrowing is written today rather than something the code states or enforces. If a client narrowing ever went the other way, this would silently do nothing and the failure would look like a provider bug.

PR #6999, opened 2026-07-31
Require `ctx` in `tool_for_tool_def` and fix the `FunctionToolset` retry budget

DouweM:
### Summary

Fixes an unreported retry-budget mismatch introduced by #6890. For a function tool with no explicit `max_retries`, `FunctionToolset.get_tools` falls back through the toolset setting to `ctx.max_retries`, while `FunctionToolset.tool_for_tool_def` used a hardcoded `1`. The two paths therefore disagreed, and a durable function tool rebuilt inside an activity could receive the wrong retry budget.

The regression test prepares and rebuilds the same renamed function tool with no explicit retry setting, then verifies that `get_tools` and `tool_for_tool_def` both use the run context's budget.

This also requires the keyword-only `ctx` argument in both `FunctionToolset.tool_for_tool_def` and `MCPToolset.tool_for_tool_def`. Both implementations need the run context to reconstruct the same retry budget as `get_tools`.

`original_name` deliberately remains specific to `FunctionToolset`; the two method signatures are not otherwise unified. `FunctionToolset` is the only leaf toolset whose own `get_tools` can emit a `tool_def.name` that differs from the key in its internal `tools` dictionary, because per-tool `Tool.prepare` / `prepare_tool_def` runs inside `FunctionToolset.get_tools`. `MCPToolset.get_tools` keys by `mcp_tool.name` and gives `tool_def.name` that same value, so its names cannot differ. `PreparedToolset` explicitly refuses to rename: “Prepare function cannot add or rename tools. Use `FunctionToolset.add_function()` or `RenamedToolset` instead.”

`RenamedToolset` can rename tools, but it is a wrapper. Durable execution wraps leaf toolsets, so `RenamedToolset` remains outside the durable boundary and maps names back in `call_tool`; the activity only sees native names. `original_name` is therefore a consequence of one leaf toolset performing its own per-tool preparation, not a general toolset contract. It is deliberately not declared on the widely subclassed `AbstractToolset`, where only one implementation could meaningfully use it.

Neither #6449 nor #6890 shipped in v2.21.0, so no deprecation or compatibility shim is needed.

### Tests

- `PYRIGHT_PYTHON_IGNORE_WARNINGS=1 uv run pyright` on changed files
- `uv run ruff check`
- `uv run ruff format --check`
- `uv run pytest tests/test_toolsets.py tests/test_mcp.py`

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/6999?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in Cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_function_toolset.py
+++ b/pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_function_toolset.py
@@ -42,7 +42,7 @@ def temporalize_function_toolset(
                 if params.tool_def is not None:
                     # Rebuild the tool from the definition the workflow prepared, so a tool's `prepare`
                     # function isn't run a second time here against the activity's limited run context.
-                    tool = toolset.tool_for_tool_def(params.tool_def, params.original_name)
+                    tool = toolset.tool_for_tool_def(params.tool_def, ctx=ctx, original_name=params.original_name)
                 else:
                     # Only reachable for an activity scheduled by a worker predating `tool_def` on these
                     # params; re-prepare so in-flight executions still complete across the upgrade.
--- a/pydantic_ai_slim/pydantic_ai/mcp.py
+++ b/pydantic_ai_slim/pydantic_ai/mcp.py
@@ -1169,7 +1169,17 @@ class MCPToolset(AbstractToolset[AgentDepsT]):
             )
         return tools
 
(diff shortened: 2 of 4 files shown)


DouweM:
Thanks — confirmed the signature unification and `ctx.max_retries` fallback are both covered by the focused regression test and the full local verification suite.

Issue #7002, opened 2026-07-31
Agent `max_concurrency` hangs workflows under `TemporalDurability`


DouweM:
### Initial Checks

- [x] I confirm that I'm using the latest version of Pydantic AI

### Description

`Agent(max_concurrency=...)` hangs indefinitely when the agent is run inside Temporal workflows and more than one workflow contends for the limiter.

`max_concurrency` is normalized once in `Agent.__init__` (`agent/__init__.py:508`) into a single process-wide `ConcurrencyLimiter`, and acquired in `iter()` via `get_concurrency_context(self._concurrency_limiter, f'agent:{agent_name}')` (`agent/__init__.py:~1640`). Under `TemporalDurability` the agent run — and therefore that acquire — executes in **workflow** code. Awaiting an `anyio` semaphore whose resolution order depends on contention from *other* workflow executions on the same worker is not deterministic, and in practice the workflows simply never make progress.

### Reproduction

Four workflows started concurrently on one worker, each running the same module-level agent, which makes six tool calls.

- Without `max_concurrency`: completes in ~2s and all four histories replay cleanly with `Replayer`.
- With `max_concurrency=2` added to the same agent: **never completes.** Timed out at 200s (`timeout` exit code 124); the workflows stall right after the Temporal server and worker start.

The only variable between the two runs is the `max_concurrency=2` kwarg.

### Scope

This is **not** specific to the tool-call path — it is the run-level limiter, so it affects any agent constructed with `max_concurrency` and run under `TemporalDurability`.

The same hazard applies to any explicitly shared `AbstractConcurrencyLimiter`, since sharing runtime scheduling state across workflow executions is inherently non-deterministic in workflow code. The distinction that matters is whether the limiter instance is shared *between concurrent workflow executions*, not what it limits.

### Notes

- Nothing in `docs/durable_execution/` or `pydantic_ai_slim/pydantic_ai/durable_exec/` mentions `max_concurrency` today, so there is no warning for users and no guard in the durability capability.
- Found while implementing the per-turn tool concurrency bound for #6884. That work deliberately avoids the problem by normalizing an `int`/`ConcurrencyLimit` into a **fresh limiter per run**, so each run has an independent budget; only an explicitly passed limiter instance is shared, and that is documented as the caller's responsibility. The same treatment is *not* available to `max_concurrency`, whose entire purpose is to bound concurrent runs — so it needs a different answer (a guard, a documented incompatibility, or a durability-aware limiter).

[...]

PR #7007, opened 2026-07-31
Add `max_tool_concurrency` to `Agent`

### LINKED ISSUE #6884: No way to bound per-turn tool-call concurrency — one model turn fanned out 792 concurrent executions
### Problem

There is no way to bound how many tool calls from a single model turn execute concurrently. Run-level concurrency controls bound *runs*; within one turn, pydantic-ai executes every tool call of the turn concurrently, so an N-wide turn is N concurrent executions with no ceiling. Under `TemporalDurability` that means N `ScheduleActivity` commands in one workflow-task completion.

Two production incidents from unbounded turns (Temporal + Lambda workers, but only the second is Temporal-specific):

1. A wide turn's completions exceeded gRPC's 4 MiB workflow-task-completion cap → sdk-core treats it as non-retryable → **104 child workflows terminated in one day**.
2. Single streaming turns fanning out to **792 / 798 / 799 concurrent tool-call activities** (healthy peak for the same agent: 1). The activation resolving ~792 results breached the worker's deadlock-detection ceiling, the workflow task failed, Temporal retried the identical batch, and one run made **zero progress for 2 h 42 m**. The same burst drove ~800 concurrent queries into the database behind the tools.

Downstream systems (DB pools, rate-limited APIs) get whatever the model decides a turn should be. Queue-level rate limits can't help — they meter dispatch over time, not one turn's batch width.

### Ask

A first-class per-toolset (or per-agent) concurrency limit on tool-call execution within a turn — e.g. `max_concurrent_tool_calls` on the toolset or on the durability capability's toolset wrapping.

### Prior art (running in production)

A `WrapperToolset` holding an `asyncio.Semaphore(8)`, recreated per run step so the limit is per-turn and replay-safe under Temporal. It works, but as an opt-in wrapper it has to be threaded into every agent (we've missed some), and under durability it changes how activity commands batch across workflow tasks — which is exactly why it belongs inside the library where the durability capability can own it consistently. Related surface: #6510 applies tool *timeouts* in `ToolManager` so all toolsets honor them — the same argument applies to a concurrency bound. Happy to PR the toolset-wrapper shape if you'd take it.

DouweM:
- Closes #6884

There is no way to bound how many tool calls from a single model turn execute concurrently. You can bound to 1 (`parallel_tool_call_execution_mode('sequential')`) or to unbounded — not to N. `max_concurrency` is a *run* limiter, not a tool limiter.

Production motivation from the reporter: a wide turn's completions exceeded the 4 MiB gRPC workflow-task cap and Temporal terminated **104 child workflows in one day**; single streaming turns fanned out to **792/798/799** concurrent tool activities (healthy peak for the same agent: 1), breaching the worker's deadlock-detection ceiling and stalling one run for **2 h 42 m**, while driving ~800 concurrent queries into the database behind the tools.

### API

`Agent(max_tool_concurrency=...)`, alongside the existing `max_concurrency` and accepting the same `AnyConcurrencyLimit` values — `int`, `ConcurrencyLimit` (backpressure), or a shared `AbstractConcurrencyLimiter`. Enforcement lives in `ToolManager` (per @dsfaccini's direction on the issue), where every toolset funnels, so the durable paths inherit it without opting in. Acquisitions use source `tool:<tool_name>` for telemetry.

### Per-run limiters, and why that matters

An `int`/`ConcurrencyLimit` normalizes to a **fresh limiter per run**. That keeps the bound per-turn as asked, and it is what makes it safe under Temporal: tool-call scheduling happens in workflow code, so a process-wide semaphore contended by other workflows would be non-deterministic. Passing an explicit limiter instance opts into sharing and is documented as the caller's responsibility.

`test_max_tool_concurrency_temporal_replay` starts four workflows concurrently on one worker and asserts each sees a peak of exactly 2 while the global peak exceeds 2 — proving independent limiters rather than accidental serialization — then replays all four histories with `Replayer`.

### Measured

| Configuration | Peak concurrent tool bodies |
|---|---:|
| default `'parallel'`, no bound | 6 |
| `'parallel'`, `max_tool_concurrency=2` | 2 |
| `'parallel_ordered_events'`, `max_tool_concurrency=2` | 2 |
| `'sequential'`, `max_tool_concurrency=2` | 1 |

### Deliberately not included

The issue also asked for `parallel_execution_mode=` on `TemporalDurability` for DBOS parity. On investigation that isn't a real gap: `DBOSDurability` exposes it because DBOS needs a **non-default default** (`'parallel_ordered_events'`) for its checkpoint model, whereas Temporal's would be `'parallel'` — identical to the global default — and `Agent.parallel_tool_call_execution_mode()` already lets anyone pick another mode on any backend. Adding the kwarg would duplicate configuration without adding capability.

Also uncovered while testing this: `Agent(max_concurrency=...)` hangs workflows under `TemporalDurability`. Pre-existing and out of scope here — filed as #7002.

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7007?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/agent.md
+++ b/docs/agent.md
@@ -862,6 +862,22 @@ a [`ConcurrencyLimitExceeded`][pydantic_ai.exceptions.ConcurrencyLimitExceeded] 
 When instrumentation is enabled, waiting operations appear as "waiting for concurrency" spans
 with attributes showing queue depth and limits.
 
+Use `max_tool_concurrency` to limit how many tool calls can execute at once within each agent run:
+
+```python
+agent = Agent('openai:gpt-5', max_tool_concurrency=8)
+```
+
+This accepts the same values as `max_concurrency`. An integer or
+[`ConcurrencyLimit`][pydantic_ai.ConcurrencyLimit] creates an independent limiter for each run, so
+concurrent runs do not contend with one another. Passing an
+[`AbstractConcurrencyLimiter`][pydantic_ai.AbstractConcurrencyLimiter] shares that limiter across
+runs or agents instead.
+
+When using [durable execution](durable_execution/overview.md), prefer an integer or
+`ConcurrencyLimit`. A shared custom limiter runs in workflow code and must itself satisfy the
+durable runtime's determinism requirements.
+
 ### Model specific settings
 
 If you wish to further customize model behavior, you can use a subclass of [`ModelSettings`][pydantic_ai.settings.ModelSettings], like
--- a/docs/agent.md
+++ b/docs/agent.md
@@ -864,7 +864,9 @@ with attributes showing queue depth and limits.
 
 Use `max_tool_concurrency` to limit how many tool calls can execute at once within each agent run:
 
-```python
+```python {title="agent_tool_concurrency.py"}
+from pydantic_ai import Agent
+
 agent = Agent('openai:gpt-5', max_tool_concurrency=8)
 ```
 
--- a/docs/durable_execution/dbos.md
+++ b/docs/durable_execution/dbos.md
@@ -202,6 +202,10 @@ It's equivalent to the behavior of [`with agent.parallel_tool_call_execution_mod
 
 If you prefer strict ordering, you can configure the agent to run tools sequentially by setting `parallel_execution_mode='sequential'` on [`DBOSDurability`][pydantic_ai.durable_exec.dbos.DBOSDurability].
 
+Set `max_tool_concurrency` on [`Agent`][pydantic_ai.Agent] to bound parallel step fan-out within
+each run. This limit is independent of `parallel_execution_mode`; sequential mode remains a
+stricter limit of one.
+
 ### Toolsets at Runtime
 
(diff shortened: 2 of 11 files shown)


DouweM:
Good catch — the doc claim was unprotected. Added `test_dbos_durability_max_tool_concurrency`: six tool calls under `DBOSDurability` with `max_tool_concurrency=2`, asserting a measured peak of 2 concurrent tool bodies.

Confirmed it actually detects the regression rather than passing vacuously — removing `max_tool_concurrency=2` from the agent fails it with `assert 6 == 2`, so the shared `acquire_tool_call_capacity` path is genuinely what's being exercised through the DBOS step wrapper.

Issue #6976, opened 2026-07-30
Temporal: allow passing `deps` by reference instead of serializing them into every activity dispatch


DouweM:
### Description

Under Temporal, `ctx.deps` is serialized into **every** activity dispatch. A wide tool fan-out multiplies one deps object across every concurrent call in a single workflow task, and there is no library-level way to pass deps by reference instead.

Split out of #6919 as a feature request rather than a bug — the current behaviour follows from Temporal's per-activity-input model — because @ZacharyHampton's production numbers make a good case that the library should offer something better.

### Measured

With the real `pydantic_data_converter` on a 200 KB deps object:

| fan-out | bytes in one workflow-task completion |
|---|---|
| 1 | 200,658 |
| 5 | 1,003,290 |
| 10 | 2,006,580 |
| **20** | **4,013,160 — exceeds the 4 MiB gRPC completion limit** |

So ~20 concurrent tool calls at 200 KB deps kills the workflow task. Worth noting beyond the ceiling: **every copy is also written to workflow history permanently**, so this is history bloat, not just a transient limit.

### Every dispatch site passes it

`durable_exec/temporal/_durability.py` (event stream, model request, stream, cancel), `_function_toolset.py`, `_mcp_toolset.py` (×3), `_dynamic_toolset.py` (×2).

### No existing hook trims it

`TemporalRunContext.serialize_run_context` (`durable_exec/temporal/_run_context.py`) governs the run context only — `deps` is a *separate* activity argument and never passes through it. `deps_type` sets the deserialization annotation, not what gets serialized.

A user `payload_codec` **is** preserved (`durable_exec/temporal/__init__.py`, `_data_converter`), so a claim-check codec can address the size ceiling today. That's the current best answer and worth documenting either way. What it doesn't do is dedup N identical payloads within one fan-out.

### What's being asked for

Deps-by-reference as a first-class option: a way to say "this deps object is worker-resolvable, send an identifier instead of the value". Design questions this raises:

- Where does it live — a `TemporalDurability` option, something on `RunContext`, or a general durable-execution concept?
- DBOS and Prefect pass the live object in-process and don't have this problem, so a Temporal-only option is defensible, but a shared concept may be cleaner.
- Who owns rehydration on the worker side, and how does that interact with `deps_type`?
- Does it compose with the existing `payload_codec` claim-check approach or replace it?

Filing so the requirement isn't lost; not proposing a specific API.

### Python, Pydantic AI & LLM client version

```Text
[...]

PR #7009, opened 2026-07-31
Document the Temporal large-payload claim-check codec

DouweM:
Under Temporal, `ctx.deps` is serialized into **every** activity dispatch. On a 200 KB deps object a 20-wide tool fan-out produces 4,013,160 bytes in one workflow-task completion — over the 4 MiB gRPC limit — and every copy is written to workflow history permanently.

A user-supplied `payload_codec` is already preserved by `TemporalDurability` (`durable_exec/temporal/__init__.py`, `_data_converter`), so a claim-check codec solves the size ceiling **today**. Nothing documented it, so nobody finds it.

This adds the recipe to `docs/durable_execution/temporal.md`: the problem, a worked codec example, and its limits — notably that it does not dedup N identical payloads within a single fan-out.

Context: #6976. That issue asks for deps-by-reference as first-class API; this is the part that works today and is worth documenting regardless of what we decide there.

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7009?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/durable_execution/temporal.md
+++ b/docs/durable_execution/temporal.md
@@ -224,6 +224,86 @@ The activity's `RunContext` is rebuilt from the serialized payload, so its field
 
 A tool's [`prepare`](../tools-advanced.md#tool-prepare) function is not affected by these limitations: for tools in a [`FunctionToolset`][pydantic_ai.toolsets.FunctionToolset] (including those defined on the agent itself), it runs in workflow code with the complete `RunContext`, once per run step like outside a workflow. The tool definition it returns is sent to the tool-call activity, which uses it as-is, so the tool the model saw is the tool that runs, down to its [`timeout`](../tools-advanced.md#tool-timeout). Tools from a `DynamicToolset` are the exception: as the toolset is re-resolved inside activities, their `prepare` functions run there as well and see the limited `RunContext`.
 
+### Large Payloads
+
+Temporal records activity arguments in workflow history and enforces payload-size limits. As a result, a
+large [dependencies object](../dependencies.md) is copied into history every time Pydantic AI schedules a
+model, tool, MCP, or event-handler activity. A [Temporal payload codec](https://docs.temporal.io/develop/python/converters-and-encryption#custom-payload-codec)
+can apply the [claim check pattern](https://www.enterpriseintegrationpatterns.com/patterns/messaging/StoreInLibrary.html):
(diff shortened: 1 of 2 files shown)


DouweM:
Both addressed, and the section has been reworked more fundamentally after @DouweM asked whether Temporal documents this already.

**It does, and it supersedes what this PR originally taught.** Temporal ships first-class [external storage](https://docs.temporal.io/external-storage) — `DataConverter(external_storage=ExternalStorage(drivers=[...]))` with a `StorageDriver` ABC, a `payload_size_threshold` defaulting to 256 KiB, and driver selection. Verified available at our `temporalio>=1.24.0` floor (`temporalio.converter.ExternalStorage`, `StorageDriver` import fine on the locked 1.24.0).

So the hand-written 80-line `ClaimCheckCodec` has been dropped. Teaching users to reinvent a facility the SDK provides — without its thresholds, driver selection, or parallel upload/download — was the wrong recommendation, and it also meant shipping a large block of copyable code we neither test nor stand behind.

What replaced it: a short section pointing at Temporal's external storage, with Temporal's own [claim-check codec cookbook](https://docs.temporal.io/ai-cookbook/claim-check-pattern-python) mentioned as the fallback for anyone who can't adopt a public-preview API.

On the two nits specifically:

- The replay caveat is now a `!!! warning` admonition, matching the "Persisted payload schemas" warning above it.
- The "replace `payload_store` in production" guidance is gone entirely along with the example.

**And the one claim that is ours is now tested.** The section states that `PydanticAIPlugin` preserves your `external_storage` while swapping in the Pydantic-aware payload converter. That holds because `_data_converter` uses `dataclasses.replace`, but nothing pinned it — `test_pydantic_ai_plugin_preserves_external_storage` now does, mirroring the existing `..._preserves_custom_payload_codec` test.

dsfaccini:
coordination required https://github.com/pydantic/pydantic-ai/pull/7157#issuecomment-5187188521

PR #7010, opened 2026-07-31
Pass Temporal `deps` by reference (design spike)

DouweM:
**Design spike for #6976 — not for merge as-is.** Opening it so the design and its measurements are reviewable.

Adds `TemporalDurability(dependency_resolver=...)` backed by a generic `TemporalDependencyResolver[DepsT, RefT]`: `to_reference` runs synchronously in workflow code, `from_reference` runs async worker-side, and activities carry `RefT` while all Pydantic AI callbacks still receive `RunContext[DepsT]`. Applied to all dispatch sites (model request, streamed request, cancellation, event handler, function tool, dynamic toolset, all three MCP paths).

### Measured

| fan-out | current full `DepsT` | prototype `RefT` |
|---:|---:|---:|
| 1 | 200,658 B | 32 B |
| 10 | 2,006,580 B | 320 B |
| 20 | 4,013,160 B | 640 B |

### Why I recommend holding it

- Temporal 1.24 exposes `DataConverter.external_storage`, and Temporal shipped External Storage in public preview in May 2026. That is the upstream claim-check primitive.
- The resolver still emits N references for N activities — it does **not** dedup within a fan-out, which was the one gap a codec couldn't close.
- The documented codec recipe (separate PR) already addresses the size ceiling and history bloat.

So the marginal value over docs + upstream external storage is narrow: natural `DepsT` worker-side with application-versioned references, and discoverability.

Full design write-up including rejected alternatives and failure modes is in the issue thread.

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7010?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/durable_exec/temporal/__init__.py
+++ b/pydantic_ai_slim/pydantic_ai/durable_exec/temporal/__init__.py
@@ -25,6 +25,7 @@ from pydantic_graph.exceptions import UnsupportedEventLoopError
 from ...agent.abstract import AbstractAgent
 from ...exceptions import AgentRunError, UserError
 from ._agent import TemporalAgent  # pyright: ignore[reportDeprecated]
+from ._dependencies import TemporalDependencyResolver
 from ._durability import TemporalDurability
 from ._logfire import LogfirePlugin
 from ._payload_converter import PydanticAIPayloadConverter
@@ -35,6 +36,7 @@ from ._workflow import PydanticAIWorkflow
 __all__ = [
     'TemporalAgent',
     'TemporalDurability',
+    'TemporalDependencyResolver',
     'PydanticAIPlugin',
     'LogfirePlugin',
     'AgentPlugin',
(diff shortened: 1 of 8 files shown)


DouweM:
Agreed, and noted for if this graduates — the drift risk is real given how many dispatch sites re-inline the same two conditionals.

Leaving it as-is while it's a spike, because the recommendation is currently **not** to merge this: Temporal's own [external storage](https://docs.temporal.io/external-storage) is a first-class claim-check facility available at our `temporalio>=1.24.0` floor, and #7009 now documents it. This resolver still emits N references for N activities, so it doesn't dedup a fan-out either — which was the one gap a codec couldn't close, and the main argument for a bespoke API.

If it does graduate, hosting both directions on the resolver so a `None` collapses to identity — and folding `reference_type if resolver else deps_type` in with it — is the right shape.

PR #7022, opened 2026-08-01
Expose `tool_name` on `ToolResultEvent`

### LINKED ISSUE #6971: `ToolResultEvent` exposes `tool_call_id` but not `tool_name`
Split out of #6639 per [maintainer request](https://github.com/pydantic/pydantic-ai/pull/6639#issuecomment-5136835456).

### The gap

A UI labeling tool activity gets the name on the call side (`FunctionToolCallEvent.part.tool_name` — required `str` on `ToolCallPart`) but not on the result side: `ToolResultEvent` ([`messages.py:3471` @ v2.21.0](https://github.com/pydantic/pydantic-ai/blob/v2.21.0/pydantic_ai_slim/pydantic_ai/messages.py#L3471)) exposes only a `tool_call_id` property over `part: ToolReturnPart | RetryPromptPart`.

So every consumer that wants to label results picks one of:

- carry a `tool_call_id → tool_name` map for the lifetime of the stream,
- reach through `.part.tool_name` (works, but `.part` is the model-facing message part, not the event surface), or
- ship a placeholder.

Ours currently does the last: the Temporal-side consumer hardcodes `"unknown"` into the result chunk it emits, and the frontend renders that literal string as the user-visible tool label. With #6639's `event_stream_topic` re-emitting these events verbatim, the gap ships to every subscriber of every non-HTTP transport.

### The wrinkle (why this is a deliberate call, not a drive-by)

`ToolReturnPart` inherits `tool_name: str` (required — [`messages.py:1246`](https://github.com/pydantic/pydantic-ai/blob/v2.21.0/pydantic_ai_slim/pydantic_ai/messages.py#L1246)), but `RetryPromptPart.tool_name` is `str | None = None` ([`messages.py:1605`](https://github.com/pydantic/pydantic-ai/blob/v2.21.0/pydantic_ai_slim/pydantic_ai/messages.py#L1605)) because retry parts also serve output-validation retries. A mirroring property is therefore `str | None`:

```python
@property
def tool_name(self) -> str | None:
    """The name of the tool that was called, if known."""
    return self.part.tool_name
```

Two observations from the consumer seat:

- In practice, a `RetryPromptPart` reaching `FunctionToolResultEvent` came from a tool call, so the name is expected present — the `None` in the type flows from `RetryPromptPart`'s other job. If that expectation is actually an invariant on this path, the property docstring can say so.
- The retry branch is exactly where the label matters most: our consumer maps `RetryPromptPart` → an "error" status chunk, and that is precisely the row currently rendering "unknown". `str | None` with a consumer-side fallback is still strictly better than reaching through `.part`.

There's precedent for both halves already in `messages.py`: `FinalResultEvent` exposes `tool_name: str | None` at the event level ([`messages.py:3386`](https://github.com/pydantic/pydantic-ai/blob/v2.21.0/pydantic_ai_slim/pydantic_ai/messages.py#L3386)). Symmetrically, `ToolCallEvent` could grow the same property next to its existing `tool_call_id` (there it's a plain `str`).

Verified on v2.21.0.

Related: #6639 (origin, where this was flagged), #6619.

dgvj-work:
Approach matches the shape proposed in #6971 and called out by @brianstrauch on #6639 as a real core gap.

- Closes #6971

### Checklist

### Notes

- Additive property only: `ToolResultEvent.tool_name` mirrors existing `tool_call_id` and returns `self.part.tool_name` as `str`.
- Asserts `tool_name` is set — `ToolResultEvent` never wraps a `RetryPromptPart` without one.
- No change to event construction, streaming, or serialization.
- Test covers success (`ToolReturnPart`) and tool/output retries (`RetryPromptPart` with name).

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/messages.py
+++ b/pydantic_ai_slim/pydantic_ai/messages.py
@@ -3483,6 +3483,16 @@ class ToolResultEvent:
         """An ID used to match the result to its original call."""
         return self.part.tool_call_id
 
+    @property
+    def tool_name(self) -> str | None:
+        """The name of the tool that was called, if known.
+
+        Always set for [`ToolReturnPart`][pydantic_ai.messages.ToolReturnPart]. May be `None` for a
+        [`RetryPromptPart`][pydantic_ai.messages.RetryPromptPart] used for output-validation retries
+        (not a tool call).
+        """
+        return self.part.tool_name
+
     __repr__ = _utils.dataclasses_no_defaults_repr
 
 
--- a/tests/test_messages.py
+++ b/tests/test_messages.py
@@ -2210,3 +2210,21 @@ def test_narrow_message_parts_promotes_valid_claims_and_leaves_plain_parts():
     assert type(narrowed[0].parts[0]) is LoadCapabilityCallPart
     assert narrowed[0].parts[1] is messages[0].parts[1]
     assert type(narrowed[1].parts[0]) is LoadCapabilityReturnPart
+
+def test_tool_result_event_tool_name():
+    """ToolResultEvent exposes tool_name next to tool_call_id for stream consumers."""
+    return_part = ToolReturnPart(tool_name='search', content='ok', tool_call_id='call_1')
+    result_event = FunctionToolResultEvent(part=return_part)
+    assert result_event.tool_call_id == 'call_1'
+    assert result_event.tool_name == 'search'
+
+    retry_part = RetryPromptPart(content='bad args', tool_name='search', tool_call_id='call_2')
+    retry_event = FunctionToolResultEvent(part=retry_part)
+    assert retry_event.tool_call_id == 'call_2'
+    assert retry_event.tool_name == 'search'
+
+    output_retry = RetryPromptPart(content='invalid output', tool_call_id='call_3')
+    output_event = OutputToolResultEvent(part=output_retry)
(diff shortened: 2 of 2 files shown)


pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -3483,6 +3483,16 @@ def tool_call_id(self) -> str:
...
 
+    @property
+    def tool_name(self) -> str | None:
+        """The name of the tool that was called, if known.
+
+        Always set for [`ToolReturnPart`][pydantic_ai.messages.ToolReturnPart]. May be `None` for a
+        [`RetryPromptPart`][pydantic_ai.messages.RetryPromptPart] used for output-validation retries
+        (not a tool call).
+        """
+        return self.part.tool_name
```
DouweM, line 3494: I think we can drop the `| None` and assert that `tool_name` is set -- because we should never have a `ToolResultEvent` with a `RetryPromptPart` that doesn't have a `tool_name`.

dgvj-work: Good call — updated. `tool_name` is now `str`, with an assert that it’s set. Dropped the `None` case from the test and covered an output-tool retry that includes a name instead.

dgvj-work:
Superseded by #7295 — applied the `str` + assert feedback there.

PR #7023, opened 2026-08-01
Test cancellation of concurrent `PeekableAsyncStream` pulls

### LINKED ISSUE #7017: `PeekableAsyncStream.aclose()` can block indefinitely behind an in-flight provider pull
### Description

`PeekableAsyncStream.aclose()` waits for `_source_lock`, which `__anext__()` and `peek()` hold across `await anext(self._source_iter)`. If one task is parked pulling from a provider that has stalled, another task calling `aclose()` blocks until that provider produces an item — which may be never.

`pydantic_ai_slim/pydantic_ai/_utils.py`:

```python
async def __anext__(self) -> T:
    ...
    async with self._source_lock:
        try:
            return await anext(self._source_iter)   # <- lock held across a provider read
        ...

async def aclose(self) -> None:
    ...
    # Wait for any in-flight `__anext__`/`peek()` (e.g. a `group_by_temporal` prefetch task) to
    # release the source before closing it, so we don't close a generator that's still running.
    async with self._source_lock:                    # <- waits for that read
        await aclose()
```

The comment shows the wait is deliberate — it avoids `RuntimeError: aclose(): asynchronous generator is already running`. The problem is that waiting is unbounded, and nothing cancels the task holding the lock.

### Reproduction

```python
async def source():
    try:
        yield 'first'
        await asyncio.sleep(30)
        yield 'second'
    finally:
        print('source_finalized')

stream = PeekableAsyncStream(source())
await anext(stream)
pull = asyncio.create_task(anext(stream))        # parks inside the source
await asyncio.sleep(0)
await asyncio.wait_for(stream.aclose(), 0.2)     # never returns
```

```text
first: first
close: TIMEOUT
pull_done_after_close: False
source_exhausted_flag: True
source_finalized          <- only after the probe cancels the blocked pull
cleanup_close: returned
```

Cancelling the parked pull first makes the identical `aclose()` return immediately, which isolates the lock holder as the blocker rather than the source's own close.

### Reach

`PeekableAsyncStream` wraps the provider iterator in the streamed responses of Function, OpenAI, Anthropic, Google, Mistral, Groq, Hugging Face and xAI.

### Status: latent, not yet a demonstrated user-facing hang

I have not found a caller that produces the interleaving today — `group_by_temporal` cancels its own prefetch task before closing, which avoids it. So this is a hazard in a shared primitive rather than a reproducible user report. It is worth fixing because the primitive permits independent pull and close tasks and gives `aclose()` no way to cancel the one holding the lock, so any future caller that closes concurrently with a pull inherits an unbounded hang.

### Pre-existing

Not introduced by #6871. `git show origin/main:pydantic_ai_slim/pydantic_ai/_utils.py` has the same `_source_lock` held around `anext()` and re-acquired by `aclose()`.

### Suggested fix

Same shape as the fix applied to `AgentStream.aclose_events()` in #6871: don't wait unboundedly. Either acquire without blocking and skip the close when a pull owns the lock, or give the stream a way to cancel the in-flight pull before closing. The second is stronger — the first trades a hang for a leak.

Note the trade-off is real: skipping the close leaves the provider stream and its HTTP connection open, so for a primitive this widely used, cancelling the holder is probably the right answer.

### Context

Found by an automated defect hunt over the review findings on #6871, and verified there.

mikemikimike:

The production fix for `PeekableAsyncStream.aclose()` landed in #7095. It cancels every registered pull scope before closing the source, preventing `aclose()` from waiting indefinitely behind a stalled provider pull.

This PR adds regression coverage for the concurrent case that the existing test does not exercise:

- `peek()` holds the source lock while stalled inside the provider;
- `__anext__()` registers a second pull scope while waiting for that lock;
- `aclose()` cancels both scopes, finalizes the source, and lets both consumers finish with normal end-of-stream behavior.

This guards against replacing the scope set with a single last-writer-wins cancel scope, which leaves the first pull stalled and makes `aclose()` hang.

### Checklist


mikemikimike:
Thanks for catching this race. I moved the per-pull cancel scope outside `_source_lock`, so it covers both lock acquisition and the provider pull. `aclose()` still marks the stream exhausted before cancelling the published scope; a pull that races after that observes `_exhausted` once it acquires the lock and exits without entering the provider. This closes both timing windows without closing the source while another pull is active.

Validation after commit `506451701`:
- `PYTHONUTF8=1 uv run ruff check pydantic_ai_slim/pydantic_ai/_utils.py tests/test_utils.py` — passed
- `PYTHONUTF8=1 uv run pyright pydantic_ai_slim/pydantic_ai/_utils.py tests/test_utils.py` — 0 errors, 0 warnings
- `PYTHONUTF8=1 uv run pytest tests/test_utils.py -q` — 33 passed
- `PYTHONUTF8=1 uv run pytest tests/test_streaming.py -q` — 180 passed, 3 xfailed

mikemikimike:
The updated run completed with all code-quality and matrix checks passing, including the full standard test matrix on Python 3.10–3.14. The only failure is unrelated to this change:

`tests/test_temporal.py::test_durability_removed_tool_still_raises_user_error@temporal` under Python 3.13 lowest-versions — expected `UserError`, received `KeyError` (`9409 passed, 795 skipped, 3 xfailed, 1 failed`).

The failing test does not exercise `PeekableAsyncStream` or the changed files. I also attempted to rerun the failed jobs, but GitHub reports that reruns require repository admin rights. Could a maintainer rerun that job if desired?

dsfaccini (dismissed):
This review was posted by Claude Code using claude-opus-5 on behalf of David. David has not reviewed these findings.

Cancelling the in-flight pull is the right call, but the hang still reproduces with your patch applied. That's the blocker.

**1. Two pulls in flight and `aclose()` still hangs.** `_source_pull_cancel_scope` is a single slot, last writer wins:

1. Pull A takes `_source_lock` and parks in the provider.
2. Pull B publishes its scope over A's, then blocks on `_source_lock`.
3. `aclose()` cancels B — a no-op — then waits on `_source_lock`, which A still holds. Forever.

That's the prefetch-alongside-consumer interleaving `_source_lock`'s own comment describes. Two ways out: cancel *all* in-flight scopes (keep a set), or wrap only `anext(self._source_iter)` with the lock already held, so at most one can exist.

**2. Coverage is red.** The `_exhausted`-after-lock fallbacks in both `peek()` and `__anext__` — the race fix you added — plus `peek()`'s whole cancellation path are never exercised.

<details>
<summary>Reproduced output — your head commit, patched into the real class</summary>

`PeekableAsyncStream` methods from `7535066` loaded into the real class, one stalling source that never yields again:

```
single pull -> aclose OK; source cancelled: True
   consumer of the cancelled pull saw: CancelledError
   _source_pull_cancel_scope reset to None? False
two pulls   -> aclose TIMED OUT (hang NOT fixed); source cancelled: False
```
</details>

<details>
<summary>Other findings</summary>

- **The published scope is never cleared.** The `finally` does `async with self._source_pull_cancel_scope_lock` *inside* the scope that was just cancelled, and `anyio.Lock.acquire()` opens with `checkpoint_if_cancelled()`, so it re-raises immediately and the reset never runs (third line of the output above). This falls out for free if you drop that lock — a plain assignment can't be interrupted.
- **Drop `_source_pull_cancel_scope_lock` entirely.** Publish and read happen with no `await` between check and use on a single-threaded loop, and the actual defense against Macroscope's race is the `_exhausted` re-check under `_source_lock`, not this lock. Meanwhile it adds two extra `anyio.Lock` acquire/release cycles — real event-loop checkpoints — to every chunk of every streamed response across eight providers.
- **`CancelledError` vs `StopAsyncIteration`.** `raise anyio.get_cancelled_exc_class()` fires outside the scope that absorbed the cancellation, so the consuming task looks cancelled to its own task group. A pull that instead loses the race and sees `_exhausted` gets `StopAsyncIteration` for the identical event. Since `aclose()` sets `_exhausted = True` first, `StopAsyncIteration` is the coherent answer for both.
- **The regression test doesn't assert what broke.** It checks that the pull was cancelled, but never that the source actually closed, and it swallows the racing consumer's exception behind `task_group.cancel_scope.cancel()` — so it passes regardless of what `__anext__` raises.
</details>

Separately, every real provider closes past this method: `close_stream` in `models/google.py`, `models/huggingface.py` and `models/xai.py` calls `self._response.source.aclose()` directly, so only `FunctionModel.request_stream` reaches `PeekableAsyncStream.aclose` today. Filed as #7116; not your problem.

PR #7024, opened 2026-08-01
Use workspace package in agent runner

### LINKED ISSUE #6998: The `agent` check fails on any PR that adds a `pydantic_ai.messages` symbol
### Description

The `agent` check fails on any PR that adds a symbol to `pydantic_ai.messages` and uses it in the gh-aw shim. Current example, on #6793:

```
ImportError: cannot import name 'ToolAvailabilityDeltaPart' from 'pydantic_ai.messages'
  (/tmp/gh-aw/uv/cache/environments-v2/pydantic-ai-runner-c87898cc2b6b2057/lib/python3.12/site-packages/pydantic_ai/messages.py)
```

### Cause

The shim's *code* is the checkout; the *library it imports* is the last PyPI release.

`.github/scripts/pydantic_ai_gh_aw_shim/` is checked out at the PR's revision, and imports from the library directly:

```python
    ...
    ToolAvailabilityDeltaPart,
)
```

but `.github/scripts/pydantic-ai-runner` resolves that library from PyPI via its PEP 723 block:

```
# dependencies = [
#   "pydantic-ai-slim[anthropic,mcp]>=1.105.0",
```

So the two halves are at different revisions, and the shim can only ever use API that has already shipped.

### Why it's worth fixing rather than tolerating

The failure lands on exactly the PRs the AI review workflows are most useful on — the ones changing the message model — and it stays red until the symbol reaches PyPI, which is after merge. So those PRs get no `agent` review at all, and a genuinely red `agent` check becomes indistinguishable from this one. It also trains everyone to ignore that check.

### Suggested fix

Point the script at the workspace instead of PyPI. `uv run --script` honours `[tool.uv.sources]` inside the PEP 723 block, and the script sits at `.github/scripts/`, so the path is stable relative to it:

```
# [tool.uv.sources]
# pydantic-ai-slim = { path = "../../pydantic_ai_slim" }
```

That also makes the agentic workflows exercise the branch's own library, which is what you'd want from a check running on a PR. `prewarm-pydantic-ai-runner.sh` does `uv sync --script` on the open network before the firewall, so a path dependency is built there rather than fetched.

Worth confirming while implementing:

- the workflows that run on a schedule rather than a PR still resolve the path (they check out the repo too, so this should hold, but it should be checked rather than assumed);
- the lock next to the script (`the adjacent script lock pins the production dependency graph and hashes`) can still be generated with a path source;
- `pydantic-ai-harness==0.7.0` stays on PyPI — it is genuinely an external pin and not part of this repo.

The alternative — defensive imports in the shim — is worse: it would hide real incompatibilities between the shim and the library, which is the one thing this check ought to catch.

hxaxd:
## Summary

- resolve the agent runner's `pydantic-ai-slim` dependency from the checked-out workspace
- regenerate the adjacent PEP 723 lock file with the directory source
- add a regression test that keeps the workspace source declaration in place

This keeps the gh-aw shim and the library it imports at the same revision, so PRs that add symbols to `pydantic_ai.messages` can run the `agent` check before those symbols are published to PyPI. `pydantic-ai-harness==0.7.0` remains an external pinned dependency.

- Closes #6998

## Checks

- [x] Focused runner metadata tests (`2 passed`)
- [x] Ruff check and format check
- [x] Script lock consistency check
- [x] Git whitespace check

The full runner test file produced 104 passed, 1 skipped, and 16 pre-existing Windows-specific failures involving shell execution, path separators, and mocked command behavior.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7024?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


dsfaccini:

Thanks @hxaxd — your diagnosis is right and the workspace-source approach is the correct fix. We're carrying it forward in #7107.

Closing this one for a policy reason, not a quality one: we don't accept changes to `.github/` (CI workflows, runner config, actions) from outside the organization. Those files run with repository credentials, so edits to them are a supply-chain boundary we keep internal — that's on us to enforce with tooling, not something you could have known. #7107 ships the same fix and credits your catch.

Really appreciate the contribution.

PR #7026, opened 2026-08-01
Document `ToolsetTool` in the toolsets API reference

ayaangazali:
`ToolsetTool` is exported in `pydantic_ai.toolsets.__all__` (and from `pydantic_ai` itself), and the custom-toolset example in `docs/toolsets.md` has users import it and annotate their `call_tool` signature with it. But it is the one entry in that `__all__` missing from the members list in `docs/api/toolsets.md`, so it has no rendered API docs: before this change there is no `id="pydantic_ai.toolsets.ToolsetTool"` anchor on the page, and it only shows up incidentally inside other members' signatures.

Added it to the members list right after `AbstractToolset`, since the two go together when you are implementing a toolset. Ran `uv run mkdocs build --strict`: the anchor is present afterwards and no new warnings appear.

### Checklist

---

quick note: built this with Claude Code's help and read the diff myself before pushing. happy to move it further down the list if you would rather keep the concrete toolsets together at the top.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7026?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/api/toolsets.md
+++ b/docs/api/toolsets.md
@@ -4,6 +4,7 @@
     options:
         members:
         - AbstractToolset
+        - ToolsetTool
         - CombinedToolset
         - ExternalToolset
         - ApprovalRequiredToolset


ayaangazali:
Re-verified on current main: `ToolsetTool` still has no rendered anchor, and it is the only entry in `pydantic_ai.toolsets.__all__` missing from the members list in `docs/api/toolsets.md`. The custom-toolset example in `docs/toolsets.md` still has users import it and annotate `call_tool` with it.

Green and mergeable. If it is meant to stay out of the reference, happy to close this instead.

ayaangazali:
Folding this into #7483, which carries all four of these API-reference additions in one PR. Four separate PRs for four one-line additions was more review overhead than the change is worth, and the description here cited `mkdocs build --strict` output, which stopped being meaningful when #7416 removed the MkDocs pipeline. #7483 verifies the same claim against the files on main instead. Closing this in favour of that one.

PR #7045, opened 2026-08-03
Document `OutputObjectDefinition` in the output API reference

ayaangazali:
`OutputContext` and `OutputObjectDefinition` are exported from `pydantic_ai.output` but neither renders anywhere in the docs, so they have no API reference entries.

`OutputContext` is the object passed to output hooks, so it is something users receive and read fields off. Its own docstring even cross-links `[`tool_call`][pydantic_ai.output.OutputContext.tool_call]`, which could not resolve while the class was unrendered. It also exposes `object_def: OutputObjectDefinition | None`, so the second type is reachable straight from the first and had no page either.

Added both to the members list in `docs/api/output.md`. Ran `uv run mkdocs build --strict`: the `pydantic_ai.output.OutputContext` and `pydantic_ai.output.OutputObjectDefinition` anchors exist afterwards where there were none before, and no new warnings appear.

### Checklist

---

quick note: put this together with Claude Code and checked the diff myself before pushing. if these two are meant to stay out of the output page because it focuses on the specs you pass to `Agent`, say the word and I will close it.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7045?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/api/output.md
+++ b/docs/api/output.md
@@ -11,3 +11,5 @@
             - TextOutput
             - StructuredDict
             - DeferredToolRequests
+            - OutputContext
+            - OutputObjectDefinition
--- a/docs/api/output.md
+++ b/docs/api/output.md
@@ -11,5 +11,4 @@
             - TextOutput
             - StructuredDict
             - DeferredToolRequests
-            - OutputContext
             - OutputObjectDefinition


ayaangazali:
Correction, and I have narrowed this in 46429cab3.

I claimed both symbols render nowhere. That was wrong for `OutputContext`: it is re-exported through `pydantic_ai.capabilities`, and `docs/api/capabilities.md` renders that whole module, so it already has a page at `pydantic_ai.capabilities.OutputContext`. Adding it to the output page would have documented the same class twice. I checked the wrong path when I first looked, matching on the exact `pydantic_ai.output.*` anchor rather than the class.

`OutputObjectDefinition` is genuinely unrendered on current main, and it is reachable from `OutputContext.object_def` which users read in output hooks, so this now adds that one entry only. Verified after the change that the `pydantic_ai.output.OutputObjectDefinition` anchor exists and no new build warnings appear.

Sorry for the noise in the original description.

ayaangazali:
Folding this into #7483, which carries all four of these API-reference additions in one PR. Four separate PRs for four one-line additions was more review overhead than the change is worth, and the description here cited `mkdocs build --strict` output, which stopped being meaningful when #7416 removed the MkDocs pipeline. #7483 verifies the same claim against the files on main instead. Closing this in favour of that one.

PR #7060, opened 2026-08-03
fix(messages): degrade malformed tool-call args in `args_as_json_str` like `args_as_dict`

### LINKED ISSUE #7042: Malformed tool-call args are degraded on dict transports but sent verbatim on JSON-string transports
### Initial Checks

### Description

`ToolCallPart.args_as_dict()` deliberately degrades malformed JSON so a poisoned tool call can still be sent to a provider. `args_as_json_str()` — used by every OpenAI-compatible transport — does not, and returns the malformed string verbatim. When the OpenAI-compatible endpoint fronts a provider whose tool-call input is **object-typed** (Anthropic via a gateway, e.g. OpenRouter → Vertex/Bedrock), the raw string is forwarded into `tool_use.input` and the provider rejects the **entire request**:

```
messages.N.content.M.tool_use.input: Input should be a valid dictionary
```

Because the malformed `ToolCallPart` stays in the message history, **every subsequent request in the run replays it**, so the run cannot recover — including the retry that the malformed args themselves triggered. In our case a whole verification stage failed this way daily, silently, after the model had already done the work.

The intent is stated in your own docstring on `args_as_dict` (`pydantic_ai/messages.py`):

> `raise_if_invalid`: If `True`, a `ValueError` or `AssertionError` caused by malformed JSON in `args` will be re-raised. When `False` (the default), malformed JSON is handled gracefully by returning `{'INVALID_JSON': '<raw args>'}` **so that the value can still be sent to a model API (e.g. during a retry flow) without crashing.**

That is exactly the situation here — a retry flow — but the JSON-string path never gets the protection.

**Where they diverge (2.22.0):**

| helper | used by | malformed args |
|---|---|---|
| `args_as_dict()` | `models/anthropic.py`, `models/bedrock.py`, parts of `models/openai.py` (Responses) | → `{'INVALID_JSON': <raw>}`, provider accepts |
| `args_as_json_str()` | `models/openai.py` `_map_tool_call` (Chat Completions), `models/groq.py`, `models/huggingface.py` | → raw string, verbatim |

So the same history succeeds on the native Anthropic transport and fails through an OpenAI-compatible gateway to the same model.

### Example Code

```python
"""Standalone MRE — no network. pydantic-ai 2.22.0."""

import json

import pydantic_core

# Shape emitted by a real model: a complete object, an extra closing brace,
# then trailing content.
MALFORMED = '{"collection": "documents", "filter": {"a": 1}}}, "limit": 20}'

part = ToolCallPart(tool_name="find", args=MALFORMED, tool_call_id="toolu_01")

try:
    pydantic_core.from_json(MALFORMED)
except ValueError as exc:
    print("raw args do not parse:", exc)
    #> raw args do not parse: trailing characters at line 1 column 48

# dict transports: degraded, provider-safe
print(part.args_as_dict())
#> {'INVALID_JSON': '{"collection": "documents", "filter": {"a": 1}}}, "limit": 20}'}

# JSON-string transports: raw string, verbatim
as_str = part.args_as_json_str()
print(as_str == MALFORMED)
#> True

try:
    json.loads(as_str)
except ValueError as exc:
    print("what goes on the wire does not decode:", exc)
    #> what goes on the wire does not decode: Extra data: line 1 column 48 (char 47)
```

Live confirmation through a gateway (OpenRouter → `anthropic/claude-haiku-4.5`), replaying a history containing exactly that part:

```
ModelHTTPError: status_code: 400, model_name: anthropic/claude-haiku-4.5,
body: {'message': 'Provider returned error', 'code': 400, 'metadata': {'raw':
'{"type":"error","error":{"type":"invalid_request_error","message":
"messages.1.content.1.tool_use.input: Input should be a valid dictionary"},...'
```

Sanitizing only that part — replacing `args` with `part.args_as_dict()` before the request — makes the identical request succeed.

### Suggested fix

Have `args_as_json_str()` honour the same contract as `args_as_dict()`: when `args` is a `str` that does not decode to a JSON **object**, emit the degraded object instead of the raw string, e.g.

```python
def args_as_json_str(self) -> str:
    if not self.args:
        return '{}'
    if isinstance(self.args, str):
        try:
            if isinstance(pydantic_core.from_json(self.args), dict):
                return self.args
        except ValueError:
            pass
        return pydantic_core.to_json(self.args_as_dict()).decode()
    return pydantic_core.to_json(self.args).decode()
```

Two things worth weighing, since I don't know which you'd prefer:

1. **Round-tripping is a real cost** — well-formed args are returned unchanged above, so only broken ones pay the parse, but it is no longer a pure pass-through. If that is unacceptable on the hot path, doing it in `models/openai.py::_map_tool_call` (and the Groq/HF equivalents) instead keeps `args_as_json_str` free.
2. **This is arguably not OpenAI's bug** — OpenAI's `arguments` is a free-form string and tolerates anything, so the raw pass-through is correct *for OpenAI proper*. The breakage only appears when an OpenAI-compatible endpoint fronts an object-typed provider. If you'd rather not change the shared helper, a gateway-aware fix (or simply documenting that OpenAI-compatible transports don't carry the `INVALID_JSON` guarantee) would still have saved us — the surprise was that the two transports disagree at all.

I'm happy to open a PR for whichever direction you prefer.

```
pydantic-ai-slim 2.22.0 (openai, google, logfire, temporal, ui)
Python 3.12
Transport: OpenRouter (OpenAI-compatible) → anthropic/claude-haiku-4.5,
routed upstream to both Google Vertex and Amazon Bedrock (fails identically on both)
```

Diwak4r:
Fixes #7042

## Problem

`ToolCallPart.args_as_json_str()` — used by every OpenAI-compatible transport (`openai.py` Chat Completions + Responses, `groq.py`, `huggingface.py`, plus `anthropic.py` gateway paths and `cohere.py`) — returns a malformed `args` string verbatim, while `args_as_dict()` deliberately degrades it to `{'INVALID_JSON': '<raw>'}` (the Anthropic-documented pattern referenced in `messages.py:33-37`).

When the OpenAI-compatible endpoint fronts an **object-typed** provider (e.g. Anthropic via OpenRouter → Vertex/Bedrock), the raw string is forwarded into `tool_use.input` and the provider rejects the entire request:

```
messages.N.content.M.tool_use.input: Input should be a valid dictionary
```

Because the malformed `ToolCallPart` stays in message history, **every subsequent request in the run replays it** — the run cannot recover, including the retry the malformed args themselves triggered.

| helper | used by | malformed args (before) |
|---|---|---|
| `args_as_dict()` | dict transports | `{'INVALID_JSON': <raw>}` — provider accepts |
| `args_as_json_str()` | JSON-string transports | raw string, verbatim — object-typed backend rejects the whole request |

## Fix

`args_as_json_str()` now applies the same graceful degradation as `args_as_dict()`: if the args string doesn't parse, or parses to something that isn't an object, it returns `'{"INVALID_JSON": "<raw args>"}'` (JSON-encoded). Valid object JSON strings and dict args are unchanged — the wire format is identical for every well-formed tool call.

## Tests

Added five tests in `tests/test_messages.py` mirroring the existing `args_as_dict` block:

- valid object JSON string → verbatim (unchanged)
- dict args → serialized JSON (unchanged)
- empty args → `'{}'` (unchanged)
- malformed JSON string → `{"INVALID_JSON": <raw>}` (new)
- valid non-object JSON (e.g. `'[1, 2, 3]'`) → `{"INVALID_JSON": <raw>}` (new; an object-typed backend also rejects a non-dict `tool_use.input`)

Full `tests/test_messages.py` + `tests/models/test_openai.py`, `test_groq.py`, `test_huggingface.py`, `test_cohere.py`: 405 passed, 0 regressions. (Two pre-existing Windows-only failures in `test_messages.py` — `.csv` mimetype via the Windows registry and CRLF in `Path.write_text` — reproduce on a clean tree and are unrelated.)

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7060?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/messages.py
+++ b/pydantic_ai_slim/pydantic_ai/messages.py
@@ -2018,10 +2018,22 @@ class BaseToolCallPart:
         """Return the arguments as a JSON string.
 
         This is just for convenience with models that require JSON strings as input.
+
+        Malformed or non-object JSON is handled gracefully by returning
+        `'{"INVALID_JSON": "<raw args>"}'` so that the value can still be sent to a
+        model API (e.g. during a retry flow) without crashing — the same degradation
+        `args_as_dict()` applies for dict-based transports. Without it, OpenAI-compatible
+        transports forward the raw string into object-typed backends (e.g. Anthropic via
+        a gateway), which reject the entire request.
         """
         if not self.args:
             return '{}'
         if isinstance(self.args, str):
+            try:
+                parsed = pydantic_core.from_json(self.args)
+                assert isinstance(parsed, dict), 'args should be a dict'
+            except (ValueError, AssertionError):
+                return pydantic_core.to_json({INVALID_JSON_KEY: self.args}).decode()
             return self.args
         return pydantic_core.to_json(self.args).decode()
 
--- a/tests/test_messages.py
+++ b/tests/test_messages.py
@@ -1956,6 +1956,46 @@ def test_args_as_dict_raise_if_invalid_non_dict_json():
         part.args_as_dict(raise_if_invalid=True)
 
 
+def test_args_as_json_str_valid_json_verbatim():
+    """args_as_json_str should return valid object JSON strings verbatim."""
+    part = ToolCallPart(tool_name='test_tool', args='{"key": "value"}')
+    assert part.args_as_json_str() == '{"key": "value"}'
+
+
+def test_args_as_json_str_dict_args():
+    """args_as_json_str should serialize dict args to a JSON string."""
+    part = ToolCallPart(tool_name='test_tool', args={'key': 'value'})
+    assert json.loads(part.args_as_json_str()) == {'key': 'value'}
+
+
+def test_args_as_json_str_empty_args():
+    """args_as_json_str should return '{}' when args is None/empty."""
+    part = ToolCallPart(tool_name='test_tool', args=None)
+    assert part.args_as_json_str() == '{}'
+
+
+def test_args_as_json_str_malformed_json_returns_invalid_json_wrapper():
+    """args_as_json_str should degrade malformed JSON to INVALID_JSON, like args_as_dict.
+
+    Without this, OpenAI-compatible transports forward the raw malformed string into
+    object-typed backends (e.g. Anthropic via a gateway), which reject the entire
+    request — and since the part stays in history, every retry replays the failure.
+    """
+    malformed = '{"query": "bad", "ids":[4556]</parameter>\n<parameter name="limit": 8}'
+    part = ToolCallPart(tool_name='test_tool', args=malformed)
+    assert json.loads(part.args_as_json_str()) == {INVALID_JSON_KEY: malformed}
+
+
+def test_args_as_json_str_non_dict_json_returns_invalid_json_wrapper():
+    """args_as_json_str should degrade valid non-object JSON to INVALID_JSON, like args_as_dict.
+
+    An object-typed backend also rejects a non-dict `tool_use.input` (e.g. a JSON array),
+    so the same degradation applies for parity between the two transports.
+    """
+    part = ToolCallPart(tool_name='test_tool', args='[1, 2, 3]')
+    assert json.loads(part.args_as_json_str()) == {INVALID_JSON_KEY: '[1, 2, 3]'}
+
+
 def test_user_prompt_part_with_text_content():
     part = UserPromptPart(
         content=[


dsfaccini:
This comment was posted by Codex Desktop using gpt-5.6-terra on behalf of David.

Thanks for the contribution. [Issue #7042](https://github.com/pydantic/pydantic-ai/issues/7042) was not assigned to you when this PR opened, so it does not meet the contribution policy and is being closed. Please comment on the issue and wait for an assignment before opening a PR.

dsfaccini:

Thanks @Diwak4r, this is a real bug. We've taken it forward in #7092, which adds the same `INVALID_JSON` degradation plus a wire-level test through the OpenAI-compatible transport (proving the provider actually receives valid object JSON) and parity coverage across the other `args_as_json_str()` consumers (OpenAI Responses, Vercel AI, AG-UI). Closing this in favor of that one. Appreciate the report.

PR #7070, opened 2026-08-03
Populate `span_tree` and metrics for `run_on_errors` evaluators when a task raises

### LINKED ISSUE #6927: `run_on_errors` evaluators lose captured spans and retain exporter state after task exceptions
### Initial Checks

### Description

When a decorated function or agent fails and an online evaluator has `run_on_errors=True`, the evaluator receives an empty `EvaluatorContext.span_tree` even if spans completed before the exception.

The cleanup and tree-finalization statements in `context_subtree()` and `_context_subtree_spans()` are placed after their `yield` statements without `finally` blocks. When the wrapped code raises:

1. completed spans are not added to the yielded `SpanTree`;
2. the exporter context is not cleared, leaving the completed spans stored under its generated context ID;
3. span-derived metrics in `run_task()` are not extracted on the error path.

This means error evaluators cannot inspect the telemetry leading up to a failure, which is specifically where `run_on_errors=True` is most useful. Repeated failures in a long-running process can also accumulate retained exporter entries.

A narrow fix appears to be finalizing and clearing the span context in `finally`, and ensuring `run_task()` extracts the finalized tree before building the error evaluator context. I would be happy to submit the fix with regression tests if a maintainer agrees with the approach and assigns the issue.

```python
import asyncio
from dataclasses import dataclass

import logfire

from pydantic_evals.evaluators import Evaluator, EvaluatorContext
from pydantic_evals.online import OnlineEvalConfig, OnlineEvaluator, wait_for_evaluations

logfire.configure(send_to_logfire=False, inspect_arguments=False)


@dataclass
class PrintSpans(Evaluator):
    def evaluate(self, ctx: EvaluatorContext) -> bool:
        print([node.name for node in ctx.span_tree])
        return True


config = OnlineEvalConfig(default_sample_rate=1.0)


@config.evaluate(OnlineEvaluator(evaluator=PrintSpans(), run_on_errors=True))
async def fail() -> None:
    with logfire.span('before-error'):
        pass
    raise RuntimeError('boom')


async def main() -> None:
    try:
        await fail()
    except RuntimeError:
        pass
    await wait_for_evaluations()


asyncio.run(main())
```

Current evaluator output:

```text
[]
```

Expected evaluator output to include the completed span:

```text
['before-error']
```

Inspecting the context exporter after the same failure also shows the completed span still retained in `_finished_spans` instead of being cleared.

### Logfire Trace

Not included because the bug is in local in-memory span finalization and is reproduced without sending data to Logfire.

- Python: 3.12.8
- Pydantic AI: current `main` at `6877fc5b`
- LLM provider SDK: not applicable

Diwak4r:
> [!NOTE]
> This PR depended on https://github.com/pydantic/pydantic-ai/pull/7667: the new tests assert spans recorded via `context_subtree()`, which the unfixed exporter cache could silently drop (a stopped exporter returns FAILURE from `export()` and drops spans silently). That dependency is now satisfied. The fix is merged and carried into this branch, the post-merge re-run of `uv run pytest tests/evals/test_otel.py tests/evals/test_online_capability.py` passes (66 tests), and each of the three regression tests was re-confirmed to still fail with the production change reverted.

Fixes #6927

## Problem

When a task raises inside a `run_on_errors=True` evaluator context, the error evaluator sees an empty `span_tree` and no metrics — the spans recorded before the exception were never delivered.

Three independent leaks caused this, each verified to reproduce on its own:

1. **`context_subtree`** — `tree.add_readable_spans(spans)` sat after the `yield`, so it never ran when the body raised.
2. **`_context_subtree_spans`** — `exporter.get_finished_spans(...)` / `exporter.clear(...)` were also post-`yield`; on exception the exporter was never drained or cleared, leaving every span recorded under that context id retained under its now-dead key.
3. **`run_task`** — `extract_span_tree_metrics(task_run, span_tree)` was after the `finally`, so it was skipped entirely on exception.

Leak 3 is not downstream of leak 1: with the tree finalization fixed but `run_task` left alone, an error evaluator still receives `metrics == {}` while `span_tree` is populated.

## Fix

Each cleanup step moves into a `finally`, placed so the ordering between the two nested context managers still holds:

- **`context_subtree`** wraps the whole `with _context_subtree_spans()` block, because `spans` is only populated by that context manager's own teardown — finalizing the tree any earlier would read an empty list. The `isinstance` guard skips the `SpanTreeRecordingError` case, where there is no tree to finalize.
- **`_context_subtree_spans`** nests its `try`/`finally` *inside* `with _set_exporter_context_id()`, so `context_id` is always a bound `str` and no sentinel or guard is needed.
- **`run_task`** calls `extract_span_tree_metrics` from its existing `finally`, guarded by `isinstance(span_tree, SpanTree)` so a `SpanTreeRecordingError` is left alone.

`_ContextInMemorySpanExporter` itself is unchanged.

## Tests

Three regression tests, each confirmed to fail with the production change reverted and to pass with it applied:

- `test_context_subtree_finalizes_tree_on_exception` — spans completed before the exception are reachable on the yielded tree.
- `test_context_subtree_clears_exporter_on_exception` — the exporter retains nothing. This asserts on the `_finished_spans` mapping rather than on `get_finished_spans()`: the flattened view is also empty when a context's key survives holding an empty list, which is the unbounded-growth half of #6927.
- `test_run_on_errors_evaluator_sees_spans_and_metrics_on_exception` — a `run_on_errors=True` evaluator observes both the pre-error spans and the metrics derived from them.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7070?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


dsfaccini:

I pushed 46865c8b3 to this branch and updated the title and description.

All three leaks reproduce on current `main`. Leak 3 is an independent defect rather than a symptom of leak 1: with tree finalization fixed and `run_task` left alone, an error evaluator still gets `metrics == {}` while `span_tree` is populated. Separating those was the right call.

What changed:

1. Merged `origin/main`, which the branch was 33 commits behind.
2. Removed the `context_id is not None` guard in `_context_subtree_spans` and moved the `try`/`finally` inside the `with`. `_set_exporter_context_id` returns `Generator[str]` and assigns `context_id = context_id or str(uuid.uuid4())`, so `context_id` is always bound inside the `with` and the sentinel was unreachable.
3. Deleted `test_context_subtree_skips_teardown_when_context_id_is_unset`. It passed on unfixed code, on fixed code, and with the guard deleted.
4. `_ContextInMemorySpanExporter` is now byte-identical to `main`. `test_context_subtree_clears_exporter_on_exception` asserts on `_finished_spans` rather than `get_finished_spans()`, which needs no no-context-id call and is also the stronger check: the flattened view is empty when a context key survives holding an empty list, which is the unbounded-growth half of https://github.com/pydantic/pydantic-ai/issues/6927. `_task_run.py` is untouched, so your leak-3 fix stands as you wrote it.
5. This should merge after https://github.com/pydantic/pydantic-ai/pull/7667, because the regression tests here assert spans that an unfixed exporter can silently drop.

Each of the three remaining tests fails with the production change reverted.

dsfaccini:

Dismissing this: the branch gate is not broken. The `coverage` check passes on head `77c3fc322` with the pragma already removed: https://github.com/pydantic/pydantic-ai/actions/runs/33809074367/job/100829151391

The finding scoped its measurement to `pytest tests/evals` and grepped only `docs/examples/` for runnable examples, but CI measures the combined session, and `tests/test_examples.py` executes the python blocks embedded in `docs/evals/online-evaluation.md` itself.

Those examples enter `run_task` with no tracer provider configured, so `context_subtree()` yields a `SpanTreeRecordingError` and line 57 takes its False arc. Running only those nodes, `_task_run.py` reports line 57 executed and line 58 missing, and `57->exit` does not appear in the Missing column.

No test added.

PR #7076, opened 2026-08-03
fix(durable-exec): make the payload type-adapter cache unbounded

### LINKED ISSUE #7027: TemporalDurability: the payload TypeAdapter memo's maxsize=128 gives a 0% hit rate on workers whose key space exceeds it
Follow-up to #7001, which added `PydanticAIPayloadConverter` and memoised the per-payload `TypeAdapter` build. Tested on **pydantic-ai-slim 2.22.0**.

The memo is `lru_cache(maxsize=128)`. For a worker whose distinct payload type-hint count exceeds 128, cyclic access over the working set gives an LRU a **0% hit rate** — so the memo does nothing precisely on the workers large enough to have needed it. And the count that matters is not just the application's own activities: `TemporalDurability` generates several activities per agent, and their annotations land in the same key space.

On the worker that motivated the original report: **153 distinct payload type hints** — 59 from application activities, 53 more from workflow run/signal/query/update slots, and **41 more from the 188 generated activities across 37 agent plugins**. The generated ones carry `CallToolResult`, which is the annotation the whole thing is about.

### Repro (2.22.0)

```python
import sys
from pydantic_ai.durable_exec.temporal import PydanticAIPayloadConverter
from pydantic_ai.durable_exec.temporal._payload_converter import _type_adapter
from temporalio.converter import DataConverter

N = int(sys.argv[1])
models = [type(f"Result{i}", (BaseModel,), {"__annotations__": {"v": int}}) for i in range(N)]

pc = DataConverter(payload_converter_class=PydanticAIPayloadConverter).payload_converter
json_conv = pc.converters[b"json/plain"]
payloads = [pc.to_payload(m(v=1)) for m in models]

for m, p in zip(models, payloads):          # warm every hint once
    json_conv.from_payload(p, m)

before = _type_adapter.cache_info().misses
for _ in range(10):                          # steady state: 10 more full passes
    for m, p in zip(models, payloads):
        json_conv.from_payload(p, m)
print(f"N={N} maxsize={_type_adapter.cache_info().maxsize} "
      f"rebuilds={_type_adapter.cache_info().misses - before} (want 0)")
```

```
N=120 maxsize=128 rebuilds=0        # fits
N=153 maxsize=128 rebuilds=1530     # every single decode rebuilds
```

Not degraded — inverted. Once past the bound the hit rate is zero, so it is exactly as expensive as no cache at all, plus the bookkeeping.

### Why the magnitude matters

The synthetic models above are trivial (`v: int`), so a rebuild is ~0.14 ms. The annotation this actually applies to is not. Measured in our production workers on the real `CallToolResult` (`Annotated[...]` discriminated union): **2.818 ms to build vs 0.005 ms to validate, 521x**. A single workflow activation resolving a 792-result tool fan-out spent **11.935 s of 12.089 s in `TypeAdapter.__init__`** — past Temporal's deadlock ceiling, so the workflow task failed, the identical batch was redelivered, and it failed identically. One run made zero progress for 2 h 42 m. That is the failure #7001 fixes for workers under 128 hints and does not fix for workers over it.

### Suggested shape

1. Simplest: drop the bound (`lru_cache(maxsize=None)`). The key space is closed — it is the set of annotations reachable from registered workflows and activities, fixed by the code, not by traffic. It cannot grow with load, which is the usual reason to bound a cache.
2. If an unbounded cache is unacceptable for the dynamically-constructed-hints case the docstring mentions, make it configurable — a `maxsize` argument on `PydanticAIPayloadConverter`, or a module-level constant a host can raise before the converter is built.
3. Failing either, a one-line note in the docstring that the bound exists and what to do when a registry exceeds it would at least make it discoverable. Right now a host that inherits from `PydanticAIPayloadConverter` "to retain the adapter cache" (as the docstring suggests) inherits a cache that may be silently doing nothing.

For now we override the lookup with an unbounded memo and inherit everything else from `PydanticAIPayloadConverter`. Happy to send a PR for (1) or (2) if you have a preference.

cc @DouweM

Diwak4r:
Fixes #7027

## Problem

`_type_adapter` is memoized with `lru_cache(maxsize=128)`. Once a worker's distinct payload type-hint count exceeds 128, cyclic access over the working set gives the LRU a **0% hit rate**: every decode misses and rebuilds its `TypeAdapter`, so the memo provides no benefit exactly on the workers large enough to have needed it — it becomes pure overhead on the hottest path.

The count that matters isn't just the application's own activities: `TemporalDurability` generates several activities per agent, and their annotations land in the same key space. The report behind #7027 measured **153 distinct hints** (59 app activities + 53 workflow run/signal/query/update slots + 41 generated activities), with real `CallToolResult` adapter builds taking ~2.8 ms each — a single activation blew the Temporal deadlock ceiling and made zero progress for hours.

## Change

Make the memo unbounded (`functools.cache`). This is safe because the key space is **closed**: it is the set of annotations reachable from the registered workflows and activities, which is fixed by the code, not by traffic. Unhashable hints — the main dynamic-construction case — already bypass the cache entirely (they take the uncached `TypeAdapter(hint)` path in `from_payload`), so the cache cannot grow with load.

Any fixed bound would only relocate the bug; the failure mode (silent 0% hit rate on the largest registries) is invisible until it costs a workflow task.

## Verification

- New test `test_pydantic_ai_payload_converter_cache_holds_more_than_128_hints`: builds 200 distinct dynamically-created hints, warms the cache, then runs 5 full decode passes under a `TypeAdapter` spy — asserts zero rebuilds. Red before the fix (1000 rebuilds, i.e. every decode missed), green after.
- Full `tests/test_temporal.py`: 244 passed, 2 skipped.
- `ruff check` / `ruff format --check` clean on both touched files.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7076?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_payload_converter.py
+++ b/pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_payload_converter.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from functools import lru_cache
+from functools import cache
 from typing import Any
 
 from pydantic import TypeAdapter
@@ -13,14 +13,17 @@ from temporalio.contrib.pydantic import (
 from temporalio.converter import CompositePayloadConverter, DefaultPayloadConverter, JSONPlainPayloadConverter
 
 
-@lru_cache(maxsize=128)
+@cache
 def _type_adapter(type_hint: Any) -> TypeAdapter[Any]:
-    """Build an adapter once for each recently used type hint.
+    """Build an adapter once per type hint.
 
     The cache is replay-safe: a `TypeAdapter` is a pure function of its type hint, so cache hits and
-    misses validate identically and cannot change workflow history. The 128-entry bound comfortably
-    covers the code-defined set of hints used by a worker while preventing accidental growth if an
-    application constructs hints dynamically.
+    misses validate identically and cannot change workflow history. It is unbounded because its key
+    space is closed: it is the set of annotations reachable from the registered workflows and
+    activities, which is fixed by the code, not by traffic. A bounded LRU below that working set
+    (the original 128 entries) degrades to a 0% hit rate on large registries — exactly the workers
+    the memo exists for — and every decode then rebuilds its adapter. Unhashable hints, the main
+    dynamic-construction case, bypass the cache entirely.
     """
     return TypeAdapter(type_hint)
 
--- a/tests/test_temporal.py
+++ b/tests/test_temporal.py
@@ -5612,6 +5612,34 @@ async def test_pydantic_ai_payload_converter_separates_type_hints() -> None:
     assert type_adapter.call_count == 2
 
 
+async def test_pydantic_ai_payload_converter_cache_holds_more_than_128_hints() -> None:
+    """A working set larger than the old 128-entry bound must not thrash the adapter cache.
+
+    Workers that register many activities and agents generate well over 128 distinct payload
+    type hints (the report behind the unbounded cache counted 153). With an LRU sized below
+    the working set every decode misses, so the memo provides no benefit exactly where it is
+    needed most.
+    """
+    temporal_payload_converter._type_adapter.cache_clear()  # pyright: ignore[reportPrivateUsage]
+    converter = DataConverter(payload_converter_class=PydanticAIPayloadConverter)
+
+    models = [type(f'Result{i}', (BaseModel,), {'__annotations__': {'v': int}}) for i in range(200)]
+    values = [m(v=1) for m in models]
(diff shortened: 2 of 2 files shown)


DouweM:
Superseded by https://github.com/pydantic/pydantic-ai/pull/7085

PR #7079, opened 2026-08-03
fix(vercel-ai): require a valid approval on approval-responded parts

### LINKED ISSUE #7041: Vercel AI: an `approval-responded` part with a non-`ToolApprovalResponded` approval is silently dropped

A Vercel AI `approval-responded` message part whose `approval` isn't a valid `ToolApprovalResponded` is silently dropped: the client's approval turn vanishes, the pending tool call is stripped from history, and the run finishes normally with nothing on the wire to distinguish it from success.

`ToolApproval` is an **undiscriminated** union (`ToolApprovalRequested | ToolApprovalResponded`), so an approval object that omits `approved` matches the *requested* member and parses cleanly — no 422. `iter_tool_approval_responses`' `isinstance(part.approval, ToolApprovalResponded)` check then skips it.

### Repro

Driving `VercelAIAdapter(agent, run_input, sdk_version=6)` with a `state='approval-responded'` part:

| `approval` value | outcome |
|---|---|
| `{'id':'a1','approved':false,'reason':'nope'}` | `approvals={'delete_1': ToolDenied('nope')}`, call kept |
| `{'id':'a1'}` (no `approved`) | **`deferred_tool_results is None`**, pending `ToolCallPart` stripped |
| `null` / omitted | **`deferred_tool_results is None`**, pending `ToolCallPart` stripped |
| `{'id':'a1','reason':'nope'}` | 422 (`reason` is extra on `ToolApprovalRequested`) |

The only signal in the dropped cases is a server-side `UserWarning` ("Client-submitted history ended with unresolved tool call(s) …, which were stripped").

### Where

`pydantic_ai_slim/pydantic_ai/ui/vercel_ai/_utils.py` → `iter_tool_approval_responses`, and `request_types.py` → `ToolApproval` (the undiscriminated union).

### Relationship to other work

- **Not** what [#7038](https://github.com/pydantic/pydantic-ai/pull/7038) fixed. That was AG-UI's `except ValidationError: return ToolDenied()` against a `responseSchema` we advertise, under a spec clause naming `RunError` as the outcome. Vercel has no `responseSchema`, no such clause, and no `except` — same *class* of defect (a malformed client approval degrading silently), different mechanism, and the degradation here is fail-safe since the tool never executes.
- **Not** closed by [#6995](https://github.com/pydantic/pydantic-ai/pull/6995). That makes `ToolApprovalResponded.approved` a `StrictBool`, closing the lax-coercion half (a silent *approval* from `{'approved': 'yes'}`). With `StrictBool` in place, the shapes above still match `ToolApprovalRequested` and are still skipped.

A fix probably means discriminating the union rather than relying on `isinstance` after a permissive parse.

Diwak4r:
Fixes #7041

## Problem

A Vercel AI v6 client that sends an `approval-responded` part whose `approval` is missing or malformed (e.g. `{"id": "a1"}` with no `approved` flag) is silently accepted and then dropped:

- `ToolApproval = ToolApprovalRequested | ToolApprovalResponded` is an undiscriminated union, so `{"id": "a1"}` parses as the first member, `ToolApprovalRequested`.
- `iter_tool_approval_responses` skips anything that isn't a `ToolApprovalResponded`, so the pending `ToolCallPart` is stripped and the run looks like a success — the approval just vanishes.

A response that isn't a response should be a loud error, not a silent no-op.

## Change

Make `approval` a required `ToolApprovalResponded` on the two parts that carry an approval response:

- `ToolApprovalRespondedPart`
- `DynamicToolApprovalRespondedPart`

Malformed approvals now produce a 422 `ValidationError` at `build_run_input` parse time instead of being silently dropped:

| Client payload | Before | After |
| --- | --- | --- |
| `{"id": "a1"}` | silently dropped | 422 `ValidationError` |
| `null` / omitted | silently dropped | 422 `ValidationError` (required) |
| `{"id": "a1", "reason": "nope"}` | 422 `ValidationError` | 422 `ValidationError` |
| `{"id": "a1", "approved": false, "reason": "nope"}` | valid, call kept | valid, call kept |

## Tests

Added parametrized tests covering both part shapes (`dynamic-tool` and `tool-<name>`) with malformed approvals (missing `approved`, `null`) asserting a `ValidationError`, plus a positive test asserting a valid `ToolApprovalResponded` parses and round-trips.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7079?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/ui/vercel_ai/request_types.py
+++ b/pydantic_ai_slim/pydantic_ai/ui/vercel_ai/request_types.py
@@ -216,7 +216,7 @@ class ToolApprovalRespondedPart(BaseUIPart):
     input: Any | None = None
     provider_executed: bool | None = None
     call_provider_metadata: ProviderMetadata | None = None
-    approval: ToolApproval | None = None
+    approval: ToolApprovalResponded
 
 
 class ToolOutputDeniedPart(BaseUIPart):
@@ -329,7 +329,7 @@ class DynamicToolApprovalRespondedPart(BaseUIPart):
     input: Any
     provider_executed: bool | None = None
     call_provider_metadata: ProviderMetadata | None = None
-    approval: ToolApproval | None = None
+    approval: ToolApprovalResponded
 
 
 class DynamicToolOutputDeniedPart(BaseUIPart):
--- a/pydantic_ai_slim/pydantic_ai/ui/vercel_ai/request_types.py
+++ b/pydantic_ai_slim/pydantic_ai/ui/vercel_ai/request_types.py
@@ -230,7 +230,7 @@ class ToolApprovalRespondedPart(BaseUIPart):
     input: Any | None = None
     provider_executed: bool | None = None
     call_provider_metadata: ProviderMetadata | None = None
-    approval: ToolApproval | None = None
+    approval: ToolApprovalResponded
 
 
 class ToolOutputDeniedPart(BaseUIPart):
@@ -343,7 +343,7 @@ class DynamicToolApprovalRespondedPart(BaseUIPart):
     input: Any
     provider_executed: bool | None = None
     call_provider_metadata: ProviderMetadata | None = None
-    approval: ToolApproval | None = None
+    approval: ToolApprovalResponded
 
 
 class DynamicToolOutputDeniedPart(BaseUIPart):
--- a/tests/test_vercel_ai.py
+++ b/tests/test_vercel_ai.py
@@ -172,7 +172,9 @@ def test_build_run_input_allows_regenerate_without_message_id():
     ],
 )
 @pytest.mark.parametrize('malformed_approval', [{'id': 'a1'}, None])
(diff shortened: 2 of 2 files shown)


faceWang753:
Independent verification from the original reproduction path: I checked the current PR head (`f0d9ba5`). Both static and dynamic `approval-responded` parts without `approved` now fail validation, while a valid denial round-trips correctly.

Verification: `223 passed, 1 skipped`; Ruff, format, and Pyright checks are clean. The focused responded-part change is appropriately narrower than my prototype because it fixes #7041 without changing `approval-requested` wire compatibility. No blocking findings.

PR #7084, opened 2026-08-03
Document OpenRouter app attribution environment variables

ayaangazali:
`OpenRouterProvider` falls back to `OPENROUTER_APP_URL` and `OPENROUTER_APP_TITLE` when `app_url` and `app_title` are not passed, which is documented in the provider docstring but not on the OpenRouter page. The same page already tells you `OPENROUTER_API_KEY` can be set through the environment, so the attribution section was the odd one out for anyone configuring through the environment rather than in code.

Added one sentence to the existing App Attribution section naming both variables. No other changes.

Ran `uv run mkdocs build --strict`: no new warnings.

### Checklist

---

quick note: built with Claude Code's help and read the diff myself. small one, but it tripped me up while reading the provider source, so it seemed worth writing down.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7084?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/models/openrouter.md
+++ b/docs/models/openrouter.md
@@ -40,7 +40,7 @@ agent = Agent(model)
 
 OpenRouter has an [app attribution](https://openrouter.ai/docs/app-attribution) feature to track your application in their public ranking and analytics.
 
-You can pass in an `app_url` and `app_title` when initializing the provider to enable app attribution.
+You can pass in an `app_url` and `app_title` when initializing the provider to enable app attribution. Both fall back to the `OPENROUTER_APP_URL` and `OPENROUTER_APP_TITLE` environment variables when omitted.
 
 ```python
 from pydantic_ai.providers.openrouter import OpenRouterProvider
--- a/docs/models/openrouter.md
+++ b/docs/models/openrouter.md
@@ -42,6 +42,11 @@ OpenRouter has an [app attribution](https://openrouter.ai/docs/app-attribution) 
 
 You can pass in an `app_url` and `app_title` when initializing the provider to enable app attribution. Both fall back to the `OPENROUTER_APP_URL` and `OPENROUTER_APP_TITLE` environment variables when omitted.
 
+!!! note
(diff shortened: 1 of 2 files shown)


dsfaccini (changes requested):
@ayaangazali please revise the review comment by macroscope, perhaps re-use an existing test with an existing cassette to assert or snapshot the relevant fields to prove this is the case, or otherwise what the actual behavior is

ayaangazali:
thanks @dsfaccini, revised in 1850fcf34.

macroscope's finding checks out. In `OpenRouterProvider.__init__` the `attribution_headers` dict is built first, but the `openai_client is not None` branch assigns `self._client = openai_client` and never passes them, so only clients the provider builds itself get `HTTP-Referer` and `X-Title`.

One thing I would narrow from the bot's wording though: the overloads already stop a caller passing `app_url`/`app_title` alongside `openai_client` (the second overload types `openai_client: None`), so that combination fails typechecking rather than silently doing nothing. What can silently go missing is the environment fallback, since `OPENROUTER_APP_URL` and `OPENROUTER_APP_TITLE` apply without the caller writing any attribution argument at all. I scoped the note to that.

Proof rather than assertion, as you asked, in `tests/providers/test_openrouter.py`: one test sets both env vars and asserts the built client carries the two headers, and a second sets the same env vars, passes a prebuilt `openai_client`, and asserts the returned client is that same object with neither header on it. Both go through `default_headers` because the headers are fixed at client construction and cassette matching is not sensitive to them.

Ran `uv run pytest tests/providers/test_openrouter.py` (21 passed), pyright and ruff clean on both files. Happy to reshape the note or move the tests if you would rather they live elsewhere.

PR #7106, opened 2026-08-03
Document reading AG-UI `context` and `forwardedProps` off `run_input`

### LINKED ISSUE #7105: AG-UI `RunAgentInput.context` is silently dropped

AG-UI's `RunAgentInput.context` — a list of `description`/`value` pairs — is never read by `AGUIAdapter`, so every entry a client sends is silently dropped.

It's the protocol's slot for "what's relevant to this run but isn't a message": the originating platform, the requesting user, a channel's standing instructions. CopilotKit's `useCopilotReadable` maps onto it on the web side, and their new Channels SDK uses it to carry channel context into Slack/Teams agents. Today none of it reaches us.

`forwardedProps` and `parentRunId` are likewise unread, though `forwardedProps` is at least reachable as `adapter.run_input.forwarded_props`. `context` has no accessor at all.

### Proposed

Expose `AGUIAdapter.context` as a cached property, and document the wiring.

Deliberately **not** auto-injected into the prompt: it's client-submitted text, so injecting it would put it on the same footing as the server's own instructions — the same trust boundary that gives us `manage_system_prompt` and `allow_uploaded_files`. The supported path is to read it, pass what you trust into `deps`, and render it through `@agent.instructions`.

`Context` has been stable in `ag-ui-protocol` since 0.1.10, so no version gating is needed.

dsfaccini:
> This pull request was posted by Claude Code using claude-opus-5 on behalf of David.

David reviewed the direction; not the diff line-by-line.

- Closes #7105

#7105 asked for an `AGUIAdapter.context` property because AG-UI's `RunAgentInput.context` looked unreachable. It isn't — `run_input` is public, so `adapter.run_input.context` has always worked. The issue was filed on a false premise, and [DouweM explained why the property was left out on purpose](https://github.com/pydantic/pydantic-ai/pull/7106#discussion_r3723844005).

So this PR now closes the issue by **documenting** the thing, not by adding API.

### What's here

- `docs/ui/ag-ui.md` gains a **Context** section: what `context` is, how to read it off `run_input`, and — the part that matters — why client-submitted entries must reach the model as *data*, never as instructions. A worked example shows the split: the server-authenticated workspace is the instruction, the frontend's entries arrive as tool output.
- `ui/AGENTS.md` gains the rule, so this doesn't get re-proposed: a protocol field earns an adapter property only when the adapter *consumes* it **and** it's a concept every UI protocol shares.
- One test pins the negative guarantee — with `context` populated, nothing from it reaches the first model request.

`context`, `forwardedProps` and `parentRunId` stay on `run_input`.

<details><summary>Why not normalize it onto the base adapter instead</summary>

The alternative on the table was lifting `context` to `UIAdapter` with a normalized type so Vercel AI could implement it too. Three things argue against doing that now:

1. **There's no second implementer.** Vercel AI's `useChat` transport sends `id`, `messages`, `requestMetadata`, `body`, `trigger`, `messageId` — [no `context`](https://ai-sdk.dev/docs/reference/ai-sdk-ui/use-chat). Their nearest equivalent is the untyped `body` passthrough, which `SubmitMessage` already accepts via `extra='allow'`. (`experimental_context` is a server-side `streamText` setting, not a wire field.) A base abstraction derived from one protocol is a guess.
2. **`dict` would be lossy.** AG-UI types it `List[Context]`, not a map — `description` is not a key, and CopilotKit lets you register many readables each with their own description. Collapsing to `dict[str, str]` silently drops duplicates.
3. **There's no safe default sink in the run either.** The only run-scoped dict is `RunContext.metadata`, which `docs/agent.md` documents as "added to the run span attributes for observability" — so auto-routing client text there would put unbounded untrusted content into every user's traces.

If a second protocol grows the same concept, the normalized base property can be added then, against two real shapes instead of one.

</details>

### Checklist


docs/ui/ag-ui.md
```diff
@@ -239,6 +239,44 @@ uvicorn ag_ui_state:app --host 0.0.0.0 --port 9000
...
+
+
+@agent.instructions
+def frontend_context(ctx: RunContext[ChannelDeps]) -> str:
+    return '\n'.join(f'{entry.description}: {entry.value}' for entry in ctx.deps.context)
+
+
+async def run_agent(request: Request) -> Response:
+    adapter = await AGUIAdapter.from_request(request, agent=agent)
+    deps = ChannelDeps(context=adapter.context)
```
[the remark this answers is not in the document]
dsfaccini: Valid, fixed in 66a1edb44 — and the fix went further than filtering.

An allowlist wouldn't have helped: `description` is client-controlled too, so a hostile client just sends `description='Requesting user'` and passes it. The real rule is in `docs/message-history.md` — instructions carry operator authority, so nothing the client authored may be built into them at all.

The example now puts a **server-authenticated** fact in `@agent.instructions` and delivers the client's entries as **tool output**, which is the shape `docs/index.md`'s canonical instructions example uses.

docs/ui/ag-ui.md
```diff
@@ -239,6 +239,44 @@ uvicorn ag_ui_state:app --host 0.0.0.0 --port 9000
 AG-UI frontend tools are seamlessly provided to the Pydantic AI agent, enabling rich
 user experiences with frontend user interfaces.
 
+### Context
+
+Alongside messages, an AG-UI client can send a `context` array of `description`/`value` pairs describing things it considers relevant to the run: the originating platform, the requesting user, or a channel's standing instructions. It's how a frontend or a chat-platform gateway tells the agent *who is asking, from where*.
+
+These entries are not passed to the model automatically. They're client-submitted text, so injecting them would put them on the same footing as your own instructions — see the [trust model](overview.md#trust-model-for-client-submitted-messages). Instead, read them off [`AGUIAdapter.context`][pydantic_ai.ui.ag_ui.AGUIAdapter.context], pass what you trust into `deps`, and render them through [`@agent.instructions`][pydantic_ai.agent.AbstractAgent.instructions]:
```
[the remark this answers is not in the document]
dsfaccini: Incorrect — the suggested change would break the link.

This file is `docs/ui/ag-ui.md`, so `overview.md` resolves to `docs/ui/overview.md`, which exists. `../overview.md` would resolve to `docs/overview.md`, which does not.

I did switch it to `./overview.md` to match the existing convention elsewhere in this file (the Trust model section already uses `./overview.md#trust-model-for-client-submitted-messages`), and repointed the other link at a real heading anchor rather than an admonition title.

tests/test_ag_ui.py
```diff
@@ -321,6 +327,92 @@ async def test_agui_adapter_state_none() -> None:
...
+        request = messages[-1]
+        assert isinstance(request, ModelRequest)
+        instructions.append(request.instructions)
+        yield 'ok'
+
+    agent = Agent(model=FunctionModel(stream_function=capture_instructions), deps_type=ContextDeps)
+
+    @agent.instructions
+    def frontend_context(ctx: RunContext[ContextDeps]) -> str:
+        return '\n'.join(f'{entry.description}: {entry.value}' for entry in ctx.deps.context)
```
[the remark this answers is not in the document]
dsfaccini: This reply was posted by Claude Code using claude-opus-5 on behalf of David.

Valid — the test contradicted the docs revised in this PR. Fixed in `4ebc58dfc`.

It built `@agent.instructions` out of the client's `context` entries and its docstring called that "the supported way to pass it along", which is backwards: instructions carry operator authority, so client-authored text must reach the model as data. It now pins the shape the docs example teaches — the server-established workspace is the instruction, the frontend's entries arrive as tool output — and is renamed `test_agui_adapter_context_reaches_model_as_tool_output_not_instructions`.

Swept the rest of `tests/test_ag_ui.py` for the same shape (client `context` / `forwarded_props` / `state` / message content wired into `@agent.instructions`, `instructions=`, `@agent.system_prompt`, or `system_prompt=`): no other occurrence — every other `system_prompt=` in the file is a hardcoded literal.

dsfaccini:
This reply was posted by Claude Code using claude-opus-5 on behalf of David.

Both findings valid, both fixed in `605850562` — applied as suggested.

Verified against the gate itself, not by reading: `uv run mkdocs build --strict` emitted 3 warnings attributable to this diff before the fix and 0 after.

<details><summary>Before / after</summary>

Before (`3c8bfbb33`):

    WARNING - Doc file 'api/ui/ag_ui.md' contains a link '../ui/ag-ui.md#context', but the target 'api/ui/ag-ui.md' is not found among documentation files.
    WARNING - mkdocs_autorefs: api/ui/ag_ui.md: from _adapter.py:362: (pydantic_ai.ui.ag_ui.AGUIAdapter.context) Could not find cross-reference target 'pydantic_ai.agent.AbstractAgent.instructions'
    WARNING - mkdocs_autorefs: ui/ag-ui.md: Could not find cross-reference target 'pydantic_ai.agent.AbstractAgent.instructions'

After (`605850562`): none of the three.

The `../../` depth is confirmed by `models/openai.py:749`, the one other docstring that renders onto a two-level API page — it uses `../../message-history.md`.

Remaining strict warnings (`DynamicToolset`, `ToolSearchTool`, `AGENTS.md`/`CLAUDE.md` absent from nav) are byte-identical pre- and post-fix — pre-existing, not touched here.
</details>

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
```diff
@@ -356,6 +357,21 @@ def conversation_id(self) -> str | None:
         """Conversation ID from the AG-UI `RunAgentInput.threadId`."""
         return self.run_input.thread_id
 
+    @cached_property
+    def context(self) -> list[Context]:
```
DouweM: The reason I hadn't added this before is because it wasn't a shared concept across the 2 adapters and wasn't used in any of the run-args or event logic. And it could already be accessed through the run input.

If it's here with such a generic name, what happens in Vercel AI also adds `context` (they may even have it already - haven't checked) with a different type? Then we'd want to add it to the base adapter with a normalized type, but we wouldn't be able to without breaking the AG-UI version.

So, we either add it now with a type generic enough to upgrade to the base class eventually, or we don't add it at all / keep requiring people to go through run input which is very little effort.

docs/ui/ag-ui.md
```diff
@@ -239,6 +239,63 @@ uvicorn ag_ui_state:app --host 0.0.0.0 --port 9000
 AG-UI frontend tools are seamlessly provided to the Pydantic AI agent, enabling rich
 user experiences with frontend user interfaces.
 
+### Context
+
+Alongside messages, an AG-UI client can send a `context` array of `description`/`value` pairs describing things it considers relevant to the run: the originating platform, the requesting user, or a channel's standing instructions. Every entry is a claim the client made — it can send any `description`/`value` it likes — so they describe a request, they never establish who is making it.
```
DouweM, line 244: It does seem reasonable to include this in the AG-UI adapter run context by default as it's just supporting their feature faithfully. But that doesn't necessarily require a new `context` field, right? I'm not against turning it into a `dict` and lifting it up to the base class, and give VercelAI (and other future adapters?) a chance to implement it as well.

pydantic_ai_slim/pydantic_ai/ui/ag_ui/_adapter.py
dsfaccini: > This reply was posted by Claude Code using claude-opus-5 on behalf of David.

Agreed — property dropped in `861acc363`. The PR now just documents the wiring and closes #7105 that way.

All three of your points check out: `context` feeds no run-args or event logic (every other `AGUIAdapter` property does), it's not shared with Vercel AI, and `run_input` is public so `adapter.run_input.context` already worked. That last one is why the issue exists — it claimed "`context` has no accessor at all", which is simply wrong.

One correction to your guess: **Vercel AI doesn't have `context`.** `useChat`'s transport sends `id`, `messages`, `requestMetadata`, `body`, `trigger`, `messageId` ([reference](https://ai-sdk.dev/docs/reference/ai-sdk-ui/use-chat)). Their equivalent is the untyped `body` passthrough, which `SubmitMessage` already accepts via `extra='allow'`. `experimental_context` is a server-side `streamText` setting, not a wire field.

The rule is now in `ui/AGENTS.md` so an agent doesn't re-file this.

docs/ui/ag-ui.md
dsfaccini: > This reply was posted by Claude Code using claude-opus-5 on behalf of David.

Right, it doesn't require a new field — dropped it, docs only now.

On lifting a `dict` to the base class, two things that argue for waiting:

**`dict` is lossy.** AG-UI types it `List[Context]`, not a map. `description` isn't a key, and CopilotKit lets you register many readables each with its own description, so `dict[str, str]` silently drops duplicates. A non-lossy normalized type exists (list of pairs), but then we'd be defining it from one protocol — and Vercel AI has no `context` today to check the shape against.

**"In the run context by default" has no safe sink.** The only run-scoped dict is `RunContext.metadata`, which `docs/agent.md` documents as "added to the run span attributes for observability" — so defaulting client-submitted text into it would put unbounded untrusted content in every user's traces.

Happy to do the base-class version the moment a second protocol grows the concept — two real shapes beat one guess. Until then the docs carry it.

PR #7136, opened 2026-08-04
Omit server attributes for a malformed `base_url` port and forward `base_url` through `WrapperModel`

### LINKED ISSUE #7121: Malformed `base_url` port fails the request instead of omitting `server.port` in instrumentation

`urlparse` defers authority validation to the `hostname` and `port` properties, so a `base_url` with a non-numeric port parses cleanly and then raises `ValueError` when the properties are read. In both `_instrumentation.py::model_attributes` and `embeddings/instrumented.py::model_attributes` the `try` wraps only the `urlparse` call; the property reads happen in the `else` block, unguarded.

Reproduced:

```python
>>> from urllib.parse import urlparse
>>> parsed = urlparse('https://host:notaport/v1')   # no raise
>>> parsed.port
ValueError: Port could not be cast to integer value as 'notaport'
```

Consequence: instead of omitting `server.port`, the exception propagates out of attribute-building and fails the request with instrumentation enabled. `_instrumentation.py` is the core path used by every `Model`, so the blast radius is every instrumented provider, not one adapter.

Reachability is through the public surface: `Model.base_url` is an overridable property returning an arbitrary `str | None`, so a model implementation (or a provider with an unusual base URL) can supply a value that only fails at the property read. It is not reachable through SDK-backed providers that validate their base URL at construction.

Fix: move the `hostname`/`port` reads inside the guard and narrow `except Exception` to `except ValueError`, which is all these calls raise. Note both sites carry `# pragma: no cover` / `# pragma: no branch` markers that a test reaching the malformed-authority path would invalidate, so those need removing alongside.

The same defect was fixed for the image-generation adapter in #5357 (`images/instrumented.py`); that PR deliberately left these two alone rather than widening into core instrumentation. This issue is that tracking artifact.

dsfaccini:
This pull request was posted by Claude Code using claude-opus-5 on behalf of David, who reviewed it lightly.

- Closes #7121

`urlparse` defers authority validation to the `hostname`/`port` properties, so a `base_url` like `https://example.com:notaport/v1` parses cleanly and then raises `ValueError` when the properties are read. At both `model_attributes` sites the `try` wrapped only the `urlparse` call and the property reads sat in the unguarded `else`, so the exception propagated and failed every instrumented request instead of omitting the server attributes.

Both sites now read `hostname`/`port` inside the guard and catch only `ValueError`, which is all `urlparse` and those two properties raise. Same shape as the fix already in `images/instrumented.py` on #5357, which deliberately left these two alone.

- `pydantic_ai_slim/pydantic_ai/_instrumentation.py` — the core path for every `Model`
- `pydantic_ai_slim/pydantic_ai/embeddings/instrumented.py` — same defect on the embeddings path

A malformed port now drops `server.address` along with `server.port`, rather than keeping the address: the two reads share one guard, matching the prescribed fix and the #5357 precedent.

### Tests

One regression per site — `test_instrumented_model_malformed_base_url_port` and `test_instrumented_embedding_model_malformed_base_url_port` — asserting no `server.*` attributes are emitted and the request completes. Unit rather than VCR: no real provider produces a `base_url` that survives client construction and then fails at attribute-building. The `# pragma: no cover` on each `except` line is dropped since the new tests reach it.

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7136?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


dsfaccini:
This comment was posted by Claude Code using claude-opus-5 on behalf of David. 🔍 Local review suite (automated, pre-merge)

|---|---|---|---|---|
| 1 · code-correctness | ✅ closed | 0 | 0 | 9 |

**Verdict:** ✅ PASS — was 🛑 1 blocking + 🔧 4 required across two rounds; all resolved in `1242c9cd`, `7febaecb4` and `bb585014`. The blocking finding was a regression this suite's own second round caught in the fix applied by its first round — see the applied section.
Reviewed at 2026-08-05 · diff `cd4b9c2e`

<details>
<summary>✅ 5 required action(s) resolved</summary>

_Two review rounds: the second ran because round one's fixes widened the diff, and it caught a defect introduced by round one._

**1. 🛑 `TemporalModel` reported the wrong host once `base_url` forwarded** · `pydantic_ai_slim/pydantic_ai/durable_exec/temporal/_model.py` · `TemporalModel.base_url`
`TemporalModel` resolves `model_name`/`system`/`profile` through `_current_model()` (the `using_model()` selection) but not `base_url`, so it inherited the *wrapped default's* URL. Verified: inside `using_model('alt')` a span read `gen_ai.request.model: alt-model` alongside `server.address: default.example.com`. These attributes were merely **absent** before finding 2 below, so shipping that fix alone would have replaced a silent gap with silently incorrect telemetry. Now resolves through `_current_model()`; a raw model ID reports no URL rather than naming the wrong server. Confirmed the other four `WrapperModel` subclasses have no dynamic-model selection, so inheriting the plain forward is correct for them.

**2. `WrapperModel` didn't forward `base_url`** · `pydantic_ai_slim/pydantic_ai/models/wrapper.py` · `WrapperModel.base_url`
`Model.base_url` defaults to `None`, so normal attribute lookup succeeded and the existing `__getattr__` forwarder never fired — every wrapped model emitted spans with no `server.*`. `WrapperEmbeddingModel` already forwarded it. Verified end-to-end before and after: `InstrumentedModel(WrapperModel(concrete))` went from `{}` to the full attribute set.

**3. The pre-request pricing path changed too, undocumented** · `pydantic_ai_slim/pydantic_ai/models/wrapper.py` · `WrapperModel.base_url`
`best_effort_price(provider_api_url=...)` under `UsageLimits.count_tokens_before_request` is the second consumer of `base_url`, and `_cost` prefers the URL over the provider name. Wrapped models previously always passed `None`. The direction is a convergence with the unwrapped path, but `usage.cost` sits outside the version policy's OTel-attribute carve-out, so the comment now names it.

**4. Embeddings `server.port` was unpinned and kept a stale `# pragma: no cover`** · `pydantic_ai_slim/pydantic_ai/embeddings/instrumented.py` · `InstrumentedEmbeddingModel.model_attributes`
Converged on independently by `review-spec-conformance` and `review-tests`. `server.port` appeared in zero tests in `tests/test_embeddings.py` vs five sites on the models side, so the line this PR rewrote had no positive control. Added an explicit-port case and deleted the pragma; the file is now 100% with no partial branches.

**5. The `urlparse` comment overstated the deferral** · `pydantic_ai_slim/pydantic_ai/_instrumentation.py`, `pydantic_ai_slim/pydantic_ai/embeddings/instrumented.py` · `model_attributes`
It claimed `urlparse` defers authority validation to *both* properties. Verified false: `.hostname` never raises on a bad port, and invalid IPv6 / NFKC-invalid netlocs raise *eagerly* at parse. Reworded at both sites.
</details>

<details>
<summary>⚠️ 5 informational (defer/skip)</summary>

**1. `WrapperModel` doesn't forward `model_id` / `label`** · `pydantic_ai_slim/pydantic_ai/models/wrapper.py`
Same root cause as finding 2 above — a non-raising default on `Model` shadows `__getattr__`. Verified: `WrapperModel(FallbackModel(...)).model_id` yields `'fallback:sysA,test:fallback:mA,test'` instead of `'fallback:sysA:mA,test'`. [PRE-EXISTING]; fixing it reaches durable-exec registration and the web UI, so it needs its own change — defer, pending maintainer sign-off on filing.

**2. `server.port` omitted for default-port `base_url`s** · `pydantic_ai_slim/pydantic_ai/_instrumentation.py` · `model_attributes`
GenAI semconv marks `server.port` `Conditionally Required` when `server.address` is set, and `urlparse('https://api.openai.com/v1').port` is `None`. Investigated and **recommended against changing**: the requirement is a known upstream defect ([semantic-conventions-genai#16](https://github.com/open-telemetry/semantic-conventions-genai/issues/16)), the spec's own reference implementation omits the port, and `opentelemetry-instrumentation-openai-v2` actively strips an explicit 443. Current behavior matches the reference implementation.

**3. The `base_url` → `server.*` block is duplicated across the `model_attributes` sites** · `pydantic_ai_slim/pydantic_ai/_instrumentation.py`, `pydantic_ai_slim/pydantic_ai/embeddings/instrumented.py`
Now identical except one coverage pragma; the same defect needed three fixes counting #5357's `images/instrumented.py`. Not refactored here per the bug-fix-narrowness rule, and the third copy is in-flight on #5357 — defer, to land after it merges.

**4. Malformed-port assertions keep a `server.`-filtered projection** · `tests/models/test_instrumented.py`, `tests/test_embeddings.py`
`review-precedent` proposed widening to the full attribute dict as #5357's mirror test does. Skipped: the claim-check **refuted** its mechanism — emptying `model_attributes` raises `KeyError` downstream, so both tests fail loudly rather than passing silently. The residual gap is already pinned by eight snapshots in the same two files.

**5. `GEN_AI_PROVIDER_NAME_ATTRIBUTE` re-declared instead of imported** · `pydantic_ai_slim/pydantic_ai/embeddings/instrumented.py`
The file already imports its sibling constant from `pydantic_ai._instrumentation`, where this one is defined identically. [PRE-EXISTING], outside every diff hunk — folds into item 3.
</details>

dsfaccini:

#5357 just centralized this block, so two of your four hunks are now redundant.

`base_url` → `server.*` is one shared `_instrumentation.server_attributes(base_url)`, called from all three sites (chat model, embeddings, images), in https://github.com/pydantic/pydantic-ai/commit/0184a1c4c. It carries your fix — `hostname`/`port` read inside the `try`, `except ValueError` — because #5357's copy already had it.

Redundant once that lands: the `_instrumentation.py` and `embeddings/instrumented.py` hunks.

Still yours and still wanted: `WrapperModel.base_url`, the temporal `_model.py` override, and all the tests. `test_instrumented_model_wrapped_model_server_attributes` in particular covers something #5357 doesn't touch.

Whichever merges second takes the conflict; resolving it is "keep the centralized helper, drop the in-place edit."

One coverage note if you land first: three `# pragma` markers came off in the centralization, because #5357's parametrized case set (no-authority / default-port / explicit-port / malformed-port) covers every branch of the single helper. If your version lands first with the pragmas still on the two copies, `strict-no-cover` may flag them once your new tests cover those lines.

dsfaccini:
david here: #5357 is a lot bigger, this one will merge first, lets make sure we get those small things right in this one

PR #6510, opened 2026-07-15
Apply tool timeouts in `ToolManager` so all toolsets honor them

adtyavrdhn:
## Summary

- enforce prepared tool deadlines in `ToolManager`, covering MCP, code-mode, and custom toolsets
- preserve timeout ownership through composed and dynamically resolved toolsets
- start durable function, MCP, and dynamic-tool deadlines inside tasks, activities, and steps
- propagate tool-raised `TimeoutError` while converting only configured deadlines to `ModelRetry`

## Scope

No public API changes. One internal helper owns ordinary and durable deadlines; durable dispatch behavior is otherwise unchanged.

## Verification

Tool/toolset coverage and focused Prefect, DBOS, and Temporal tests pass. Real MCP coverage proves timeout recovery leaves the session usable. Focused lint, formatting, and type checks pass.

### Checklist

- [ ] Any **AI generated code** has been reviewed line-by-line by the human PR author, who stands by it.
- [x] No **breaking changes** in accordance with the [version policy](https://github.com/pydantic/pydantic-ai/blob/main/docs/version-policy.md).
- [x] **PR title** is fit for the [release changelog](https://github.com/pydantic/pydantic-ai/releases).

PR #7145, opened 2026-08-04
Add `retries` and `timeouts` docs pages and a V1 → V2 migration map

dsfaccini:
This pull request was posted by Claude Code using claude-opus-5 on behalf of David.

Written by Claude Code with light review from David.

## Why these three pages

pydantic.dev/docs asks visiting agents to declare a `goal` query param. An
analysis of 6,691 goal-carrying requests from 1,617 users showed what they come
looking for and where we have nothing to give them. This closes the three
content gaps and one navigation defect that analysis surfaced.

Notably, the analysis's own top recommendation ("write the durable-execution
documentation, there is no docs section") was **wrong** -- seven pages exist and
are published. The defect there was placement, not absence.

## `docs/cancellation-and-timeouts.md` (Core Concepts)

The sharpest pain in the data, and invisible inside any one topic because it
surfaced independently across streaming, MCP, tools, retries and debugging:
"understand cancelling streams", "verify streaming cancel handle claims", "per
tool timeout and di", "evaluate timeouts and token limits", "use a streamable
http mcp server with explicit timeouts and lifecycle".

Documents what actually holds on `main`: there is no `cancel()` API, what unwinds
when the task is cancelled, that `run_sync` is not cancellable, what message
history survives, a matrix of which stream state each way of stopping leaves
behind, and every knob that bounds a step -- including that nothing bounds a
whole run.

## `docs/retries.md` (Core Concepts)

Retries read as one feature but are three layers, and the recurring question is
which layer fired and what it did to message history: "separate transport retries
from structured output validation retries", "research retrypromptpart tool result
history", "audit unknown tool retry semantics".

This is a layer map that routes into the existing detail pages rather than
restating them -- `tools-advanced.md` already has the precedence ladder, and
`agent.md` already splits the text and tool paths. Duplicating those would create
a second source of truth to drift. What it adds is the part nothing documented: a
`RetryPromptPart` **replaces** the `ToolReturnPart` rather than joining it, the
per-tool-name counter resets on success, an invented tool name gets its own
budget under the invented key, and `prepare` callbacks are never retried.

## `docs/migration.md` (Overview)

Migration questions arrive phrased as "what replaced X": "finding the v2
replacement for graph state persistence", "plan pydantic ai v2 migration",
"migrate mcpserverstreamablehttp to mcptoolset". `changelog.md` already had the
prose under an "Upgrade Guide" H1; this is the symbol-keyed lookup table that
answers that shape of question, linking back to it as canonical. Every
replacement symbol and every removal in it was verified against this branch.

## Durable execution: promoted, not rewritten

`docs/nav.json` had Durable Execution as the 26th item inside Capabilities,
between "Process Event Stream" and "Third-Party Capabilities", while 169 users
with 246 distinct goals arrive hunting for exactly it, naming Temporal and DBOS
unprompted. It is now its own top-level section, ordered after Capabilities.

**Page URLs are unchanged.** This was verified by building the docs site against
this branch: the site's config selects those pages by explicit include, so the
sidebar group moves while the URLs stay put and no redirects are needed.

`mkdocs.yml` gets the equivalent moves. `Project > changelog` also gains its real
label, "Upgrade Guide", which is the page's H1.

## One correction to an existing page

`tools-advanced.md` claimed `tool_timeout` applies "to all tools". Enforcement
lives in `FunctionToolset.call_tool`, and `Agent.__init__` passes it only to its
own `_AgentFunctionToolset`, so MCP, external and custom toolsets silently ignore
both it and per-tool `timeout`. The claim is corrected and the real scope stated.
There is an open PR (#6510) moving enforcement into `ToolManager`.

## Verification

`uv run pytest tests/test_examples.py` passes -- that suite lints and executes
every fence, so all snippets here run. `mkdocs build` with `validation.anchors`
reports no new warnings; every reference link and anchor on the new and edited
pages resolves.

## Flagged, not fixed

Found while verifying, each needing a code decision rather than a docs edit:

- `ModelSettings.timeout`'s docstring lists xAI as supported, but `XaiModel` has
  no `timeout` handling at all. Bedrock, Cohere and HuggingFace also ignore it
  but do not claim otherwise. **Verified and already tracked** -- #6843 covers
  exactly this, #6856 the broader `Supported by:` drift, and #6846 is an open,
  mergeable fix touching all three sites that repeat the false claim
  (`settings.py`, `providers/xai.py`, `docs/models/xai.md`). Left alone here so
  this docs PR doesn't collide with that fix. The new page states the correct
  list and omits xAI.
- `ctx.max_retries` is wrong inside an output validator on the tool path:
  `execute_output_tool_call` builds the context with the agent-level default
  while `_check_max_retries` enforces the per-tool override, so `ctx.last_attempt`
  fires early. **Verified and reproduced**; tracked as #5238, where the findings
  are now posted. The defect is symmetric -- with a per-tool limit *below* the
  agent default, `last_attempt` never becomes `True` at all. `output.md` asserts
  the not-yet-true contract and is left untouched, because its fate depends on
  which way #5238 resolves.
- ~~`changelog.md` lists `Agent.set_mcp_sampling_model()` as removed in V2.~~
  **Fixed in this PR** (7a927d4b3) -- the method still exists and works, so the
  Upgrade Guide row is corrected rather than merely omitted from the migration map.
- Output-tool retry exhaustion re-wraps the error message, dropping the wording
  (and docs link) that the tool-path error carries. **Verified and reproduced**;
  filed as #7203. The rich message is built by `_check_max_retries` on the output
  path too, then replaced in `_run_output_tool_call` and dropped from the
  `__cause__` chain. The number it prints is also `tool.max_retries` labelled as
  the run-level budget, pointing users at the wrong knob.

## Note on #6497

`Add first-party run cancellation` is open and unmerged. It adds `RunCancelled`,
`AgentRun.cancel()`, `CancellationToken` and a `cancellation_token=` parameter,
none of which exist on `main`, so none appear here. When it lands, the
"Cancelling a run" section of the new page needs a rewrite.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7145?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


docs/cancellation-and-timeouts.md
```diff
@@ -0,0 +1,129 @@
...
+2. The prompt and the model's tool call were recorded before the cancellation; the tool never returned, so nothing was recorded for it.
+
+_(This example is complete, it can be run "as is" — you'll need to add `asyncio.run(main())` to run `main`)_
+
+To put a deadline on a run rather than cancelling it on a signal, wrap the same `await agent.run(...)` in `asyncio.timeout()` (Python 3.11+) or `anyio.fail_after()`. Both cancel the task when the deadline passes, so everything below applies unchanged.
+
+While the run unwinds, Pydantic AI guarantees that:
+
+- Toolsets entered for the run are exited, so MCP sessions and other resources are closed.
+- Tool calls still running in parallel are cancelled and then awaited, so their own `finally` blocks complete before the cancellation propagates to your code.
```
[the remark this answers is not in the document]
dsfaccini: This reply was posted by Claude Code using claude-opus-5 on behalf of David.

Confirmed and fixed in 7139231.

`run_in_executor` (`_utils.py`) dispatches sync tools to `anyio.to_thread.run_sync`, or to the executor from `using_thread_executor()`. Neither can interrupt a function already running in a thread, and `fail_after` in `FunctionToolset.call_tool` wraps the await rather than the thread, so a `def` tool runs to completion past its deadline and its side effects still land.

Both the unwind guarantee and the tool-timeout row now say so explicitly, and scope the original guarantee to `async def` tools.

docs/cancellation-and-timeouts.md
```diff
@@ -0,0 +1,129 @@
...
+## Ending a run from inside a tool
+
+What a tool raises decides whether the run continues, and what the model gets to see:
+
+| Raise | Run continues? | The model sees |
+|---|---|---|
+| [`ModelRetry`][pydantic_ai.exceptions.ModelRetry] | Yes | A [retry prompt](retries.md#tool-retries) asking it to correct the call — consumes that tool's retry budget |
+| [`ToolFailed`][pydantic_ai.exceptions.ToolFailed] | Yes | A [failed tool result](tools-advanced.md#tool-failed) to adapt to — does not consume the retry budget |
+| [`ApprovalRequired`][pydantic_ai.exceptions.ApprovalRequired] / [`CallDeferred`][pydantic_ai.exceptions.CallDeferred] | Ends the run with a [`DeferredToolRequests`][pydantic_ai.tools.DeferredToolRequests] output, unless a [`HandleDeferredToolCalls`][pydantic_ai.capabilities.HandleDeferredToolCalls] handler resolves the call inline | Nothing yet — see [Deferred Tools](deferred-tools.md) |
+| Any other exception | No | Nothing — it propagates out of `agent.run()` |
```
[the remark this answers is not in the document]
dsfaccini: This reply was posted by Claude Code using claude-opus-5 on behalf of David.

Confirmed and fixed in 7139231.

`tool_manager.py:444` passes the error to `cap.on_tool_execute_error`, which can return a replacement tool result or raise `ModelRetry`. The row now states that propagation is the default and names the hook that changes it.

docs/retries.md
```diff
@@ -0,0 +1,132 @@
...
+- **`prepare` callbacks.** An exception raised by a per-tool `prepare=`, by [`PrepareTools`](capabilities/prepare-tools.md), or by a [dynamic toolset](toolsets.md) propagates out of the run unchanged — including `ModelRetry`, which is *not* turned into a retry prompt there. To hide a tool for a turn, return `None` from the callback rather than raising.
+- **The `before_model_request` hook.** It runs while the request is still being assembled, before the model is called, so a `ModelRetry` raised there propagates out of the run instead of becoming a retry prompt — there is no response to retry yet. Raise it from the [wrapping or error hook](hooks.md#model-request-hooks) instead: `hooks.on.model_request` (`wrap_model_request`) or `hooks.on.model_request_error` (`on_model_request_error`).
+- **Exceptions other than `ModelRetry` and `ToolFailed`.** Anything else a tool raises ends the run, *unless* a [capability](capabilities/overview.md) implements `on_tool_execute_error`: it sees the exception first and can return a replacement tool result or raise `ModelRetry` to keep the run going. See [Ending a run from inside a tool](cancellation-and-timeouts.md#ending-a-run-from-inside-a-tool).
```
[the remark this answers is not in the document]
dsfaccini: This reply was posted by Claude Code using claude-opus-5 on behalf of David.

Valid for the deferred exceptions — fixed in 2392a9985.

`ApprovalRequired` and `CallDeferred` are tool-raised control flow that ends the run with a `DeferredToolRequests` output instead of propagating, so "anything else ends the run" did mislead. The bullet now says non-`ModelRetry`/`ToolFailed` exceptions *propagate* rather than *end the run*, and names those two as the control-flow case.

Scoped out `SkipToolExecution`: its docstring is "Exception to raise in before/wrap tool execute hooks", so it isn't something a tool raises, and this bullet is about tool-raised exceptions.

dsfaccini:

Heads-up on a collision with #6571 — whichever of these two merges **second** owes a one-sentence amendment.

`docs/retries.md` here says:

> A `RetryPromptPart` carries the failure as either a string (from `ModelRetry`) or a list of Pydantic error details (from a `ValidationError`), and renders for the model with `'Fix the errors and try again.'` appended.

#6571 adds `RetryPromptPart.cause`, and a `cause='no_output'` retry renders its content **verbatim** — no `Validation feedback:` prefix, no `Fix the errors and try again.` suffix — because nothing failed validation. That's exactly the "responses with no usable output" case this page lists two paragraphs later.

Neither PR can fix it alone: `docs/retries.md` doesn't exist at #6571's base, and this PR can't reference a field that doesn't exist yet.

Suggested amendment for whoever lands second:

> ...and renders for the model with `'Fix the errors and try again.'` appended, unless the retry was raised because the model returned nothing actionable ([`cause`][pydantic_ai.messages.RetryPromptPart.cause] `== 'no_output'`), in which case the content renders verbatim since nothing failed validation.

No change requested on this PR as it stands.

docs/cancellation-and-timeouts.md
```diff
@@ -0,0 +1,133 @@
+# Cancellation and Timeouts
```
DouweM: This is going to be extremely outdated when #6497 lands in the near future (should be coming week or so), so this will likely need to be rewritten entirely / its own docs need to be extended to a dedicated page. I'll feed it this link + the "agent intent" stuff!

docs/migration.md
```diff
@@ -0,0 +1,176 @@
+# V1 → V2 Migration Map
+
+A lookup index for upgrading from Pydantic AI V1 to V2: find the V1 name you have in your code, read off the V2 name to replace it with.
+
+The [Upgrade Guide](changelog.md) is the canonical source for *why* each change was made, the behavior changes that come with it, and the recommended upgrade path. This page is the fast path for the one question the guide answers in prose: **what replaced what.**
```
DouweM, line 5: This does seem duplicative with the changelog; can we drop more content there, now that we have this?

docs/migration.md
```diff
@@ -0,0 +1,176 @@
...
+| `StreamedRunResult.stream_responses()` (plural, yielding `(response, is_last)`) | `stream_response()` (singular, yielding a bare `ModelResponse`; read the old `is_last` as `response.state != 'incomplete'`) |
+| `StreamedRunResult.validate_structured_output` | [`validate_response_output`][pydantic_ai.result.StreamedRunResult.validate_response_output] |
+| `async for event in agent.run_stream_events(...)` | `async with agent.run_stream_events(...) as events:` then iterate — it is an async context manager only |
+
+## Pydantic Graph
+
+| V1 | V2 |
+| --- | --- |
+| `from pydantic_graph.beta import GraphBuilder` | `from pydantic_graph import GraphBuilder` |
+| `pydantic_graph.persistence` | Removed, no replacement — there is no V2 equivalent for graph state persistence |
```
DouweM: This is false! The recommendation is to use the `StepPersistence` cap, right?

docs/cancellation-and-timeouts.md
dsfaccini: > This reply was posted by Claude Code using claude-opus-5 on behalf of David.

Agreed — merging this as-is so the page is correct for `main` today, and #6497 owns the rewrite.

To narrow what you'd need to redo, I diffed your file list against the page:

- **"Cancelling a run" — rewrite.** It opens with "there is no `cancel()` method" and has a `run_sync()` cannot be cancelled warning. `CancellationToken` inverts both.
- **"Bounding how long a step takes" — should survive.** It covers `ModelSettings['timeout']`, `tool_timeout`, MCP `init_timeout`/`read_timeout` and `UsageLimits`. #6497 touches none of `settings.py`, `toolsets/`, `mcp/` or `tool_manager.py`.
- **Worth a look:** `_tool_execution.py` is in your diff, and the page makes specific claims about in-flight tool cancellation — that `async def` tools are cancelled then awaited so their `finally` blocks run, and that a `def` tool's behavior differs between `anyio.to_thread.run_sync` and `Agent.using_thread_executor()`.

The PR description's "Note on #6497" lists the symbols I deliberately kept out (`RunCancelled`, `AgentRun.cancel()`, `CancellationToken`, `cancellation_token=`).

Leaving this thread open as your marker for the rewrite.

DouweM (dismissed):

docs/migration.md
dsfaccini: > This reply was posted by Claude Code using claude-opus-5 on behalf of David.

You're right, fixed in 3318dbfc8 — in both this page and `changelog.md:119`, which carried the same false claim.

I kept a distinction in the wording, because `StepPersistence` isn't a like-for-like swap for everything the old package did:

> No `pydantic_graph` equivalent — the builder API doesn't snapshot graph state. To save, resume, and fork **agent run** state, [Pydantic AI Harness](https://pydantic.dev/docs/ai/harness/) ships [`StepPersistence`](https://pydantic.dev/docs/ai/harness/step-persistence/)

Reasoning: #5470's own description says the builder API "has no equivalent and isn't planning to", and `docs/graph/builder/index.md` carries a "No Native Persistence" note. So for a non-agent graph there genuinely is nothing — what was false was implying that's the whole story. Shout if you'd rather it just point at `StepPersistence` flat.

(Also confirmed the cap graduated out of `experimental` — canonical import is `pydantic_ai_harness.step_persistence`.)

docs/migration.md
dsfaccini: > This reply was posted by Claude Code using claude-opus-5 on behalf of David.

Dropped it in 3318dbfc8. The `Changes covered by deprecation warnings` rename list went from 27 before → after bullets to a provenance table (group | announced in | removed in).

I mapped every bullet against `migration.md` before cutting. The what → what was fully duplicated; two things weren't, so they stayed:

1. **PR provenance** — kept as the table's two link columns, so the audit trail survives.
2. **Four caveats the name mapping can't convey** — kept as bullets under the table: `set_mcp_sampling_model()` is *not* removed, `instrument` only loses the constructor args, instrumentation `version=2`–`4` still work but warn, and the Outlines "file an issue upstream" pointer.

Left the `Changes not covered by deprecation warnings` section alone — that's the *why* prose the map explicitly defers to.

Happy to cut harder (e.g. the caveats into the map too) if you'd rather the guide be thinner.

docs/changelog.md
```diff
@@ -187,42 +187,50 @@ See [#5523](https://github.com/pydantic/pydantic-ai/pull/5523).
 
 #### Changes covered by deprecation warnings
 
-These changes were announced in the latest V1 releases via deprecation warnings that name the replacement API. If you upgraded to the latest V1 and resolved every warning, you've already made them; they're listed here with full before → after for reference.
+These changes were announced in the latest V1 releases via deprecation warnings that name the replacement API. If you upgraded to the latest V1 and resolved every warning, you've already made them. The [V1 → V2 migration map](migration.md) carries the full before → after for every symbol; this section records the behavior that changes with them and the PRs each one landed in.
```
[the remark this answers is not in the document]
dsfaccini: > This reply was posted by Claude Code using claude-opus-5 on behalf of David.

Valid — fixed in 392ec609b. The `Announced` column was mine and it invented a claim the source never made.

I took the first branch, not the second: those rows **were** deprecation-announced, so moving them to "not covered" would swap one false statement for another. Verified on the `v1` branch:

- `GrokProvider` — `@deprecated` at `pydantic_ai_slim/pydantic_ai/providers/grok.py:61`
- `instrument=` / `history_processors=` — routed through `consume_deprecated_instrument` / `consume_deprecated_history_processors_as_capabilities` in `agent/__init__.py`

The real defect: the original bullets used *"See #X"* when one PR was cited and *"Announced via #X; removed in #Y"* when two were. Splitting that into two columns turned "only one PR cited" into "never announced". Collapsed to a single `PRs` column, which carries the same provenance with no implication either way.

docs/cancellation-and-timeouts.md
dsfaccini: > This reply was posted by Claude Code using claude-opus-5 on behalf of David.

#6497 landed, so I acted on it rather than leaving the page stale: **`cancellation-and-timeouts.md` is now `timeouts.md`** (2009d6349).

The cancellation half was no longer just outdated, it was contradicted — the page warned "`run_sync()` cannot be cancelled" while your new `agent.md` section says a token from another thread is *the* way to interrupt one.

Split it the way we called it above:

| Section | Now |
|---|---|
| Cancelling a run | **deleted** — `agent.md#cancelling-a-run` owns it |
| Message history after cancellation | **deleted** — covered by `RunCancelled` |
| Cancelling a stream | **deleted** — covered in your section |
| Bounding how long a step takes | **kept** — nothing else covers the layering |
| Ending a run from inside a tool | **kept**, plus a `RunContext.cancel()` row pointing at `tools-advanced.md` |

So no rewrite needed from you — the page no longer overlaps yours. Nav, `mkdocs.yml` and all four inbound links updated; strict `mkdocs build` shows no new warnings.

Leaving this open in case you'd rather the timeout content fold into `agent.md` too, in which case the page can go entirely.

DouweM: Revisiting this now that the whole cancellation stack has merged to `main`. Good news: your rewrite already handles it — `timeouts.md`, `tools-advanced.md` ("Cancelling the Run from a Tool"), and the `agent.md` section all use `CancellationToken` / `RunContext.cancel()` / `RunCancelled` correctly, and the stale "there is no `cancel()` method" / "`run_sync()` cannot be cancelled" claims the original comment flagged are gone. Nothing here is *wrong*.

The one remaining gap is that this branch is based at #6497 (`a19b68ade`), and one thing merged **after** it that touches the exact section you edited:

- **#6498 — `AgentRunEvents.cancel()`** (a 4th first-party surface, on the `run_stream_events()` handle). On `main` it added three things to "Cancelling a Run":
  1. a 4th row to the **"Which mechanism, and which exception"** table — `Consuming run_stream_events() | AgentRunEvents.cancel() on the yielded handle | RunCancelled`;
  2. the prose below the table now says **"The first four are first-party"** (this branch still says "first three");
  3. a new subsection, *"When consuming `run_stream_events()`… `AgentRunEvents.cancel()`…"*, with a `run_cancel_stream_events.py` example.

(Also merged, but no doc impact for your pages: #7256 and #7265, two internal callback-dispatch fixes.)

**Suggested update:** when you merge `main`, that section will collide with these additions. Please make sure the merged result keeps **main's** current version of the "Cancelling a Run" section — the 4-row table, "first four", and the `AgentRunEvents.cancel()` subsection — rather than this branch's #6497-era 3-row / "first three" version. Everything else in your diff (the `timeouts.md` table, `tools-advanced.md`) already lines up with the merged behavior and needs no change.

Happy to push the merge + reconciliation to this branch directly if that's easier for you — just say the word. Otherwise this should be a quick `merge main` + keep-theirs on that one section.

docs/retries.md
```diff
@@ -0,0 +1,132 @@
...
+|---|---|---|---|
+| [Transport](#transport-retries) | The same HTTP request to the provider | [`AsyncTenacityTransport`][pydantic_ai.retries.AsyncTenacityTransport] on your HTTP client | Nothing — the agent never sees the attempts |
+| [Model fallback](#model-fallback-is-not-a-retry) | The same request against a *different* model | [`FallbackModel`][pydantic_ai.models.fallback.FallbackModel] | Only the winning response |
+| [Tool](#tool-retries) | One tool call, by asking the model to correct it | `retries={'tools': N}` and per-tool limits | A [`RetryPromptPart`][pydantic_ai.messages.RetryPromptPart] in place of the tool's result |
+| [Output](#output-retries) | The model's final answer, by asking it to correct it | `retries={'output': N}` and [`ToolOutput(max_retries=N)`][pydantic_ai.output.ToolOutput.max_retries] | A `RetryPromptPart` — see [below](#output-retries) for where it lands |
+| [Model-request hooks](hooks.md) | The model request, from `wrap_model_request` or `on_model_request_error` raising `ModelRetry` | The hook itself; it draws on the **output** budget | A new request carrying a `RetryPromptPart` |
```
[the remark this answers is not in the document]
dsfaccini: > This reply was posted by Claude Code using claude-opus-5 on behalf of David.

Valid — fixed in 8a931ef3b, both places.

Verified before changing: `_agent_graph.py` catches `ModelRetry` from `root_capability.after_model_request`, appends the response, and builds a retry node — so it does consume the output budget, exactly as described.

Two extras beyond the literal ask:

- `docs/hooks.md` already grouped `after_model_request` with `wrap_model_request`/`on_model_request_error` as retry sources, so `retries.md` was inconsistent with existing docs, not just with the code.
- The line 130 fix names the response-rejection case specifically, since that's the hook's natural use: the rejected response stays in the message history so the model can see what it said.

docs/cancellation-and-timeouts.md
dsfaccini: > This reply was posted by Claude Code using claude-opus-5 on behalf of David.

Merged in 9dc1c7b15 — all three #6498 additions are present and main's version won.

Verified each point you listed:

1. 4th table row — `Consuming run_stream_events() | AgentRunEvents.cancel() ... | RunCancelled` ✅
2. `agent.md:645` now reads "The first four are **first-party**" ✅
3. `agent.md:689` "When consuming `run_stream_events()`…" subsection with `run_cancel_stream_events.py` ✅

No reconciliation was needed, and worth flagging in case it changes how you review the next one: **there was no conflict.** This branch never touched "Cancelling a Run" — our only `agent.md` edit is a single cross-link line at ~1406 (`This is one of [several layers that can retry](retries.md)…`), so `git merge` took your section wholesale. `git diff origin/main -- docs/agent.md` is exactly that one added line.

No change needed in `timeouts.md` either: `AgentRunEvents.cancel()` is a cancellation surface, and that page now only bounds *steps* — its single cancellation touchpoint is the `RunContext.cancel()` row, which points at your sections rather than restating them.

Thanks for the offer to push it yourself — wasn't needed this time.

PR #7202, opened 2026-08-06
Cover and document DeepSeek V4 Flash via `OpenAIResponsesModel` + `DeepSeekProvider`

### LINKED ISSUE #7169: Support and document DeepSeek V4 Flash 0731 via the Responses API
### Feature request

DeepSeek has updated `deepseek-v4-flash` to the V4 Flash 0731 version and now exposes an OpenAI-compatible Responses API.

Pydantic AI currently supports `deepseek-v4-flash` through `DeepSeekProvider`, but the existing integration and documentation appear to focus on `OpenAIChatModel` / Chat Completions. I could not find explicit tests or documentation confirming that the new DeepSeek Responses API is supported.

It would be helpful to officially support and document this combination:

```python
from pydantic_ai import Agent

model = OpenAIResponsesModel(
    'deepseek-v4-flash',
    provider=DeepSeekProvider(),
)
agent = Agent(model)
```

In particular, could the integration cover:

- streamed and non-streamed responses;
- thinking/reasoning content across multi-turn tool calls;
- function tools and structured output;
- response IDs / `previous_response_id` continuation;
- the V4 Flash 0731 backend while the public API model ID remains `deepseek-v4-flash`.

If this already works through the generic OpenAI-compatible Responses implementation, dedicated tests and a short documentation example would still make the compatibility status clear.

Is support for the DeepSeek Responses API planned, and is there an expected release timeline?

### References

- DeepSeek API change log: https://api-docs.deepseek.com/updates/
- DeepSeek V4 release documentation: https://api-docs.deepseek.com/news/news260424/
- Existing Pydantic AI V4 support: https://github.com/pydantic/pydantic-ai/pull/5195
- Pydantic AI OpenAI model documentation: https://ai.pydantic.dev/models/openai/

dsfaccini:
> This pull request was posted by Claude Code using claude-opus-5 on behalf of David. David has not reviewed it yet.

- Closes #7169

DeepSeek's OpenAI-compatible Responses API (`deepseek-v4-flash`, the V4-Flash-0731 backend) already works through `OpenAIResponsesModel` + `DeepSeekProvider`, but it was undocumented and untested, so nothing stopped it from silently breaking. This PR pins it and documents it. Every claim here was verified live against `api.deepseek.com` before it was written down; **no runtime logic changes** (the one production edit is a comment).

### Tests

New `tests/models/test_deepseek_responses.py`: one case-based parametrized VCR test, recorded live, covering non-streamed, streamed, multi-turn, function tools (non-streamed and streamed), structured output, and `openai_reasoning_effort`. Each case snapshots the resulting messages *and* the request bodies that actually went out, captured via an httpx event hook so the wire assertions run against what the client builds rather than what the cassette holds.

Three facts the request-body snapshots pin, each of which a refactor could otherwise break silently:

- `openai_reasoning_effort` reaches DeepSeek as `reasoning.effort`
- DeepSeek's raw chain-of-thought is sent back as `reasoning_text` content on a follow-up request
- the `phase` DeepSeek labels its output with is surfaced in `provider_details` but never sent back

### Docs

A Responses example in the DeepSeek subsection, plus the list of things DeepSeek accepts and then ignores — the failure mode worth documenting, since its API silently drops what it doesn't support instead of erroring.

### Scope decisions

<details><summary>Why no <code>openai_supports_reasoning</code> on the DeepSeek profile</summary>

The issue and its follow-up comment both suggest adding `openai_responses_*` profile fields. Verified live that this isn't needed and would be a regression:

1. `_translate_thinking` sets `reasoning.effort` from `openai_reasoning_effort` (and from the unified `thinking` setting) **without** gating on `openai_supports_reasoning`, so effort already forwards — the `reasoning_effort` case's recorded request body shows `{'reasoning': {'effort': 'high'}}`.
2. `openai_supports_reasoning` is *not* Responses-scoped. Setting it `True` would make `_drop_sampling_params_for_reasoning` strip `temperature`/`top_p` with a `UserWarning` on the **Chat Completions** path too, for `deepseek-reasoner` and every `deepseek-v4-*` model. DeepSeek accepts `temperature` today (verified live), so that's a behavior regression for existing users, not a fix.

`openai_supports_minimal_reasoning_effort` also stays at its default: `minimal`, `none` and `xhigh` were each verified live to return 200.
</details>

<details><summary>Why <code>openai_supports_phase</code> stays off</summary>

DeepSeek stamps `phase: final_answer` on every response, which we ingest unconditionally but only send back when this flag is set — so the field looked like a candidate to flip. Sending it back does return 200. But DeepSeek's [compatibility table](https://api-docs.deepseek.com/guides/responses_api) states that unsupported fields are *silently ignored rather than rejected*, and `phase` is absent from its documented input items, so the 200 proves acceptance, not that the model reads it. Flipping the flag would add an undocumented field to every request for no observable gain. The reasoning is recorded as a comment in `providers/deepseek.py`, and the `text_multi_turn` case pins the current behavior: the replayed assistant turn carries no `phase` key.
</details>

`previous_response_id` continuation is **out of scope by design** — DeepSeek's Responses API is stateless. It's worse than merely unavailable, so the docs say to leave `openai_previous_response_id` unset: setting it makes Pydantic AI trim the turns it assumes are server-side while DeepSeek stores nothing, and a two-turn run then answered `'Unknown'` to a fact stated one turn earlier, on a 200.

### Checklist

### DIFF
--- a/docs/models/openai.md
+++ b/docs/models/openai.md
@@ -416,7 +416,7 @@ async def main():
 
 _(This example is complete, it can be run "as is" -- you'll need to add `asyncio.run(main())` to run `main`)_
 
-On models that don't label their output, per [`OpenAIModelProfile.openai_supports_phase`][pydantic_ai.profiles.openai.OpenAIModelProfile.openai_supports_phase], `provider_details` has no `'phase'` key and nothing is sent back.
+A `'phase'` key appears in `provider_details` whenever the model labels its output, but it is only sent back on models that [`OpenAIModelProfile.openai_supports_phase`][pydantic_ai.profiles.openai.OpenAIModelProfile.openai_supports_phase] marks as accepting it. On every other model the label is surfaced to you and dropped from follow-up requests.
 
 ### Background mode
 
@@ -568,7 +568,7 @@ agent = Agent(model)
 ...
 ```
 
-As an alternative to the Chat Completions API shown above, DeepSeek also serves an OpenAI-compatible [Responses API](#responses-api-features), currently for the `deepseek-v4-flash` model only. Use it by pairing [`OpenAIResponsesModel`][pydantic_ai.models.openai.OpenAIResponsesModel] with `DeepSeekProvider`:
+As an alternative to the Chat Completions API shown above, DeepSeek also serves an OpenAI-compatible [Responses API](#responses-api-features), [currently for the `deepseek-v4-flash` model only](https://api-docs.deepseek.com/guides/responses_api). Use it by pairing [`OpenAIResponsesModel`][pydantic_ai.models.openai.OpenAIResponsesModel] with [`DeepSeekProvider`][pydantic_ai.providers.deepseek.DeepSeekProvider]:
 
 ```python
 from pydantic_ai import Agent
@@ -583,7 +583,11 @@ agent = Agent(model)
 ...
 ```
 
-DeepSeek's Responses API is stateless, so [`openai_previous_response_id`](#referencing-earlier-responses) and [`openai_conversation_id`](#using-durable-conversations) are not available: pass [message history](../message-history.md) back on each run instead.
+DeepSeek [documents](https://api-docs.deepseek.com/guides/responses_api) which parts of the Responses API it implements, and unsupported fields are silently ignored rather than rejected, so it's worth knowing what does nothing:
+
+- The API is stateless, so [`openai_previous_response_id`](#referencing-earlier-responses), [`openai_conversation_id`](#using-durable-conversations), [background mode](#background-mode) and [message compaction](#message-compaction) are unavailable. Pass [message history](../message-history.md) back on each run instead.
+- Of the [native tools](../native-tools.md), only [`WebSearchTool`][pydantic_ai.native_tools.WebSearchTool] is executed; the others are ignored rather than refused.
+- Reasoning is configured with `openai_reasoning_effort` (or the unified [`thinking`](../capabilities/thinking.md) setting); `openai_reasoning_summary` is accepted but produces no summary.
 
 ### Alibaba Cloud Model Studio (DashScope)
 
--- a/docs/models/openai.md
+++ b/docs/models/openai.md
@@ -568,6 +568,23 @@ agent = Agent(model)
 ...
 ```
 
+As an alternative to the Chat Completions API shown above, DeepSeek also serves an OpenAI-compatible [Responses API](#responses-api-features), currently for the `deepseek-v4-flash` model only. Use it by pairing [`OpenAIResponsesModel`][pydantic_ai.models.openai.OpenAIResponsesModel] with `DeepSeekProvider`:
+
+```python
+from pydantic_ai import Agent
+from pydantic_ai.models.openai import OpenAIResponsesModel
+from pydantic_ai.providers.deepseek import DeepSeekProvider
+
+model = OpenAIResponsesModel(
+    'deepseek-v4-flash',
+    provider=DeepSeekProvider(api_key='your-deepseek-api-key'),
+)
+agent = Agent(model)
+...
+```
+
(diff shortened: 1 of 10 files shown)


dsfaccini:

|---|---|---|---|---|
| 1 · code-correctness | ✅ closed | 0 | 0 | 3 |

**Verdict:** ✅ all 3 waves passed, 0 blocking, 0 required · 8 applied — was 🔧 8 required; all 8 resolved in `689609b5e` and `506a5b59b`
Reviewed at 2026-08-05 · diff `a116643e`

<details>
<summary>✅ 8 required action(s) resolved</summary>

_The 8 required actions the review raised, all resolved in `689609b5e` and `506a5b59b`._

**1. `openai_supports_phase` unaccounted for** · `pydantic_ai_slim/pydantic_ai/providers/deepseek.py` · `DeepSeekProvider.model_profile`
Two reviewers flagged that DeepSeek stamps `phase` on every response, which we surface but never send back. Live-tested: sending it back returns 200 — but DeepSeek's compat table says unsupported fields are *silently ignored*, and `phase` is absent from its documented input items, so a 200 proves acceptance, not that the model reads it. Left at the `False` default, with the reasoning recorded as a comment so it isn't re-litigated.

**2. `phase` off-side untested** · `tests/models/test_deepseek_responses.py` · `text_multi_turn`
Added a two-turn case whose follow-up request body pins the assistant turn going back out with no `phase` key. Flipping the flag would fail the snapshot.

**3. Streamed tool-call parity untested** · `tests/models/test_deepseek_responses.py` · `function_tool_stream`
Added a streamed counterpart to the function-tool case.

**4. `Case` was a mutable dataclass** · `tests/models/test_deepseek_responses.py` · `Case`
Now `frozen=True`, matching every sibling case-table and the `tests/AGENTS.md` template.

**5. `previous_response_id` framed as merely unavailable** · `docs/models/openai.md` · DeepSeek
Live-verified it's a silent-data-loss trap, not a no-op: with `'auto'`, Pydantic AI trims the turns it assumes are server-side and DeepSeek stores nothing, so the model answered `'Unknown'` to a fact stated one turn earlier — on a 200. Docs now say to leave it unset and why.

**6. Native-tool claim was wrong in both directions** · `docs/models/openai.md` · DeepSeek
"the others are ignored rather than refused" conflated DeepSeek ignoring built-ins with Pydantic AI refusing tools outside `supported_native_tools()`. Reworded to the provider-specific fact only. Image and document inputs degrading to placeholder text were also missing.

**7. Text phases intro contradicted its own closing sentence** · `docs/models/openai.md` · Text phases
The opener said `phase` is sent back unconditionally; the send-back is gated on `openai_supports_phase` while ingest is not. Scoped the claim.

**8. Docstring misdescribed the recorded evidence** · `tests/models/test_deepseek_responses.py`
Said requests carry `store: false`; it's the responses that report it. Reworded.
</details>

<details>
<summary>⚠️ 4 informational (defer/skip)</summary>

**1. `openai_supports_reasoning` deliberately not set** · `pydantic_ai_slim/pydantic_ai/providers/deepseek.py` · `DeepSeekProvider.model_profile`
skip — net-negative. `reasoning.effort` already reaches the wire without it; setting it would make `_drop_sampling_params_for_reasoning` strip `temperature`/`top_p` on the Chat Completions path too, which DeepSeek accepts today. Independently re-verified by the architecture lane.

**2. `openai_supports_minimal_reasoning_effort` left at its default** · `pydantic_ai_slim/pydantic_ai/providers/deepseek.py`
skip — live-verified `minimal`, `none`, `xhigh` and unified `thinking='minimal'` all return 200, so the unsupported-enum-value concern doesn't apply.

**3. `previous_response_id` documented, not enforced** · `pydantic_ai_slim/pydantic_ai/providers/deepseek.py`
skip — no server-side-state profile flag exists, and inventing one for a single provider is the narrow one-off the library's philosophy warns against. A deliberate scope boundary.

**4. Per-case httpx hook repeated across cases** · `tests/models/test_deepseek_responses.py`
skip — extracting a shared fixture would turn a provider-coverage PR into a test-infra refactor; the repetition is the shape `tests/AGENTS.md` prescribes.
</details>

PR #7204, opened 2026-08-06
fix(tool-search): omit empty search corpus

### LINKED ISSUE #7189: ToolSearch prevents rediscovery of tools discovered before CompactionPart
### Initial Checks

### Description

When message history contains a tool discovery followed by `CompactionPart`, Pydantic AI continues treating the pre-compaction tool as discovered.

This prevents the model from finding the tool again after compaction:

1. A deferred tool is discovered through tool search.
2. A later `CompactionPart` replaces the earlier conversation context.
3. On the next run, `parse_discovered_tools` still reconstructs the tool from the search result before compaction.
4. `ToolSearchToolset` treats the tool as revealed and excludes it from the custom search strategy's corpus.
5. The model searches for that exact tool again (since the tool discovery has been compacted). The strategy receives an empty corpus and returns `No matching tools found`.

The model may no longer have the structured tool reference after compaction, but the runtime prevents it from rediscovering the tool.

Expected behavior:

A discovery that occurs only before `CompactionPart` should not make the tool unsearchable after that boundary.

```Python

```

### Logfire Trace

_No response_

- Python: 3.12.10
- Pydantic AI: 2.24.0
- LLM provider SDK: Anthropic API

daleselaji-dev:
## Problem

After tool discovery, a `CompactionPart` can remove the discovery conversation while the runtime still keeps the discovered tool callable. The local `search_tools` function remained exposed with an empty corpus, so a model could retry a search that can only return no matches.

## Root Cause

`ToolSearchToolset.get_tools()` checked whether any deferred tools were searchable before building the corpus. It did not check whether all searchable tools had already been revealed.

## Solution

Build the local search tool first and only expose it when its undiscovered corpus is non-empty.

## Changes

- Omit `search_tools` when all searchable deferred tools are already revealed.
- Keep revealed tools callable after compaction.
- Add a regression test covering discovery followed by `CompactionPart`.

## Testing

- `PYTHONPATH=pydantic_ai_slim;. py -3 -m pytest --import-mode=importlib tests/test_tool_search.py -k "tool_search_toolset or parse_discovered_tools" -q` ? 37 passed.
- `py -3 -m ruff format --check` ? passed.
- `py -3 -m ruff check` ? passed.
- `py -3 -m pyright --pythonpath D:\Python310\python.exe pydantic_ai_slim\pydantic_ai\toolsets\_tool_search.py` ? passed.
- `git diff --check` ? passed.
- Full `tests/test_tool_search.py` was attempted; 138 passed, 66 unrelated provider/cassette failures and 21 environment errors occurred because this incremental checkout lacks the full test helper/cassette surface and provider fixtures. Those failures are not counted as passed.

## Compatibility / Risk

This is a narrow behavior change for the empty-corpus case only. Revealed tools remain available; searches remain available whenever at least one deferred tool is still undiscovered. Native tool-search behavior is unchanged.

## Notes for Reviewer

The test models the reported sequence: a typed discovery, then a `CompactionPart`, with `parse_discovered_tools()` preserving the callable tool names. The search surface is then omitted because no searchable corpus remains.

## Linked Issue

Fixes #7189

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7204?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/toolsets/_tool_search.py
+++ b/pydantic_ai_slim/pydantic_ai/toolsets/_tool_search.py
@@ -369,7 +369,9 @@ class ToolSearchToolset(WrapperToolset[AgentDepsT]):
         # "unsupported builtin" raise AND leave a redundant function tool on the wire
         # alongside the native builtin on providers that DO support it.
         if self.enable_fallback and searchable:
-            result[_SEARCH_TOOLS_NAME] = self._build_search_tool(ctx, searchable, revealed_tool_names)
+            search_tool = self._build_search_tool(ctx, searchable, revealed_tool_names)
+            if search_tool.corpus:
+                result[_SEARCH_TOOLS_NAME] = search_tool
 
         return result
 
--- a/tests/test_tool_search.py
+++ b/tests/test_tool_search.py
@@ -39,6 +39,7 @@ from pydantic_ai.capabilities.combined import CombinedCapability
 from pydantic_ai.exceptions import ModelRetry, UnexpectedModelBehavior, UserError
 from pydantic_ai.messages import (
     AgentStreamEvent,
+    CompactionPart,
     LoadCapabilityCallPart,
     LoadCapabilityReturnPart,
     ModelMessage,
@@ -898,8 +899,8 @@ async def test_tool_search_toolset_discovered_tools_keep_defer_loading():
     assert tools['crypto_price'].tool_def.defer_loading is True
 
 
-async def test_tool_search_toolset_keeps_search_tool_after_all_discovered():
-    """`search_tools` stays in the request even when every deferred tool is discovered.
+async def test_tool_search_toolset_omits_search_tool_after_compaction_discovery():
+    """A compacted discovery stays callable but is removed from the empty search corpus.
 
     Dropping it would invalidate the cached request prefix on the next turn — keeping
     it preserves prompt caching across discovery steps. The local tool's body is a no-op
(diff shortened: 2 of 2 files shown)


pydantic_ai_slim/pydantic_ai/toolsets/_tool_search.py
```diff
@@ -369,7 +369,9 @@ async def get_tools(self, ctx: RunContext[AgentDepsT]) -> dict[str, ToolsetTool[
         # "unsupported builtin" raise AND leave a redundant function tool on the wire
         # alongside the native builtin on providers that DO support it.
         if self.enable_fallback and searchable:
-            result[_SEARCH_TOOLS_NAME] = self._build_search_tool(ctx, searchable, revealed_tool_names)
+            search_tool = self._build_search_tool(ctx, searchable, revealed_tool_names)
+            if search_tool.corpus:
+                result[_SEARCH_TOOLS_NAME] = search_tool
```
pingxia23, line 374: Thanks for working on this. I don’t think this addresses the case reported in the issue.

The reported case has a partially remaining corpus, not an empty one. For example:

- The initial deferred tools are A, B, C, D, and E.

- The model discovers A, B, and C.

- The conversation is compacted, so the model loses that discovery context.

- parse_discovered_tools() still treats A, B, and C as discovered. 

- The search corpus now contains only D and E. Because D and E remain, the corpus is not empty and this PR still exposes search_tools. 

- However, the model cannot rediscover A, B, or C.

The new test discovers every deferred tool, producing an empty corpus. That is a different case from the one reported.
Could we add a test where only some tools were discovered before CompactionPart and make those pre-compaction tools searchable again?

pingxia23: Thinking more

Could all eligible tools remain searchable, with undiscovered matches returned first and previously discovered matches appended when the result limit allows? This may occasionally return a tool that's been discovered before. But it prevents the runtime from blocking the LLM in edge cases like this.

DouweM:
Thanks for the PR but I'm working on related stuff in https://github.com/pydantic/pydantic-ai/pull/7104 so I'll pick this up myself.

Issue #7224, opened 2026-08-06
`OpenAICompaction` stateful mode is defeated by full-history replay: OpenAI does not ignore input items before a `compaction` item


DouweM:
### Description

While verifying compaction wire semantics for #7189, a live probe showed that **OpenAI's Responses API does not treat a `compaction` input item as a truncation marker**: input items that appear before it are fully visible to the model and billed as input tokens.

Probe: compact a small conversation via `/responses/compact`, then send a follow-up whose `input` contains a fabricated pre-boundary user/assistant exchange holding a magic word that cannot be in the encrypted state, followed by the compaction item and a question asking for the word. The model answered with the magic word (`input_tokens` billed accordingly). Anthropic behaves the opposite way — its docs state that when a `compaction` block is received, "all content blocks before it are ignored", and the same probe there returns UNKNOWN with identical `input_tokens` whether or not the pre-boundary content is included.

This matters for `OpenAICompaction`'s **stateful mode** (server-side `context_management`) with default settings: the server returns a compaction item mid-response, we store it as a `CompactionPart` in history, and on later requests `_map_messages` replays the **full** history inline with the compaction item in place. Without `openai_previous_response_id='auto'`, every pre-compaction item is re-sent, re-processed, and re-billed — so compaction saves nothing and context keeps growing. It also means the model sees both the original content *and* the compaction state.

Stateless mode is not affected: the capability already replaces the request history with `[compacted_response, last_request]` client-side. Anthropic is not affected: the server ignores (and does not bill) pre-boundary blocks, matching OpenAI's own guidance that you "can drop items that came before the most recent compaction item to keep requests smaller".

### Proposed fix

In `OpenAIResponsesModel._map_messages` (or `prepare_messages`), drop messages/items that precede the most recent same-provider `CompactionPart` when building `input`, mirroring what the server does on Anthropic and what OpenAI's compaction guide describes. Alternatively (or additionally), document that stateful `OpenAICompaction` should be combined with `openai_previous_response_id='auto'`.

### References

- OpenAI compaction guide: "After appending output items to the previous input items, you can drop items that came before the most recent compaction item to keep requests smaller."
- Anthropic compaction docs: "When the API receives a `compaction` block, all content blocks before it are ignored."
- Related: #7189 (tool rediscovery across compaction; fix in progress)

PR #7225, opened 2026-08-06
Keep deferred tools searchable after discovery and reset reveal state at `CompactionPart` boundaries

### LINKED ISSUE #7189: ToolSearch prevents rediscovery of tools discovered before CompactionPart

DouweM:
- Closes #7189

Follow-up to #7104 (based on that branch; retargets to `main` when it merges). After compaction, the model "starts anew": this PR makes derived tool state agree with what the model can actually see, in two independent, mutually reinforcing changes.

### 1. The search corpus never shrinks with discovery

`ToolSearchToolset` used to subtract already-revealed names from the corpus, so a tool discovered before a `CompactionPart` became unsearchable exactly when the model could no longer remember discovering it — `search_tools` (and any custom `search_fn`, as in the report) answered "No matching tools found" for a tool that exists. Now every searchable deferred tool stays in the corpus for the whole run; the default keyword strategy ranks not-yet-revealed matches first when trimming to `max_results` (as suggested by @pingxia23 on #7204). Rediscovery is idempotent — `parse_discovered_tools` is a set union — and a fresh search return regenerates reveal evidence the model can see, so discovery is self-healing under any history truncation: server-side compaction, history processors, or harness compaction strategies.

### 2. A `CompactionPart` is a visibility boundary for derived state

`parse_discovered_tools` and `parse_loaded_capabilities` now reset their accumulated state at every `CompactionPart`, at its exact part position (Anthropic ignores earlier parts *within the same response* too). Without this, on #7104's lazy-advertisement wire designs the runtime kept claiming "revealed" for tools whose schema carrier (`additional_tools` item / replay block) sits before the boundary — schema-withheld in `tools`, carrier invisible to the model, and unsearchable per the old corpus rule. The reset is deliberately provider-agnostic: over-counting blocks rediscovery and ships invisible "reveals", while under-counting merely permits a redundant, idempotent re-search or capability re-load. Tools remain callable throughout — the reset hides schemas, it never revokes execution.

### 3. The boundary is public API: `pydantic_ai.messages.post_compaction_window`

Capabilities keep needing the same rule — harness `CodeMode`'s dynamic-catalog announcements and `DynamicWorkflow`'s reveal announcements both dedupe against in-memory state that compaction silently invalidates (pydantic-ai-harness#562, pydantic-ai-harness#563). `post_compaction_window(messages)` returns the messages from the latest `CompactionPart` onward at part-level precision; both parsers here are rebuilt on it as the single definition of the boundary (the reverse boundary-locate keeps the scan cheap), and the compaction docs tell capability authors to derive model-visible state from the window rather than instance attributes. The default search ranking also puts undiscovered matches unconditionally first now, so an already-available tool never displaces an undiscovered match under `max_results` — and `test_repeated_searches_paginate_through_a_large_corpus` pins that repeating a query still enumerates a large corpus page by page (the explicit pagination signal is #7232).

### 4. A capability's own tools always activate its validate/execute hooks

Flagged by Veria on this PR and confirmed by reproduction: a capability-owned tool stays callable while its capability counts as unloaded, and `CombinedCapability`'s loaded-gate then skipped the owner's `before/after/wrap_tool_validate`/`_execute` hooks — approval gates included — for that call. The boundary reset made this reachable in legitimate runs (load pair pre-boundary, model calls the tool from summary memory); the fabricated-history route (reveal evidence with no load pair) already had the same bypass on the base branch, and its regression test fails there too. The tool-scoped dispatchers now force the owning capability's hooks for its own tool (`_ctx_for_tool_hook`), with `capability_loaded` kept truthfully `False`; cross-cutting hooks of unrelated unloaded capabilities remain skipped as before.

### Verified against the live APIs

- **Anthropic** (docs + live probe): "when the API receives a `compaction` block, all content blocks before it are ignored" — confirmed semantically (planted marker unrecoverable) and in billing (identical `input_tokens` with and without pre-boundary content). Server-side `tool_search` re-discovers a deferred tool after compaction from the request's declarations, and the API accepts a dangling pre-boundary reveal without error — the boundary-reset semantics here match the server's own.
- **OpenAI** (live probe): input items before a compaction item are **not** ignored — a planted marker that cannot be in the encrypted state was recovered, and the items are billed. For this PR that only makes the reset err on the benign side; the resulting inefficiency in `OpenAICompaction`'s stateful full-replay path is filed separately as #7224. (Stateless mode already replaces history client-side and is unaffected.)

### Tests

Each fix is independently load-bearing (mutation-verified: reinstating the corpus subtraction or neutering either reset fails tests):

- the reporter's exact partial-corpus scenario: A–C discovered, compaction, custom `search_fn` still sees A–E and A is rediscovered and executed
- corpus invariance with no compaction at all, and undiscovered-first ordering when `max_results` trims
- part-level boundary reset for all four discovery representations (typed, native, legacy metadata, delta) and for capability load pairs, including a pair split across the boundary
- capability tools re-hide after the boundary and re-reveal on a fresh `load_capability`
- a pre-boundary-revealed tool called directly post-compaction (no re-search) still executes
- a capability-owned tool called post-compaction (and via fabricated reveal with no load at all) runs its owner's execute hooks with `capability_loaded=False`
- on the OpenAI wire, a pre-boundary reveal resolves as `withheld` again and emits no `additional_tools` carrier

### Checklist


pydantic_ai_slim/pydantic_ai/_deferred_capabilities.py
```diff
@@ -136,12 +137,21 @@ def _narrow_load_capability_return(part: ToolReturnPart) -> LoadCapabilityReturn
...
+    under-counting only permits a redundant, idempotent load.
+    """
     call_id_by_tool_call_id: dict[str, str] = {}
     loaded: set[str] = set()
     for msg in messages:
         for part in msg.parts:
-            if isinstance(part, LoadCapabilityCallPart):
+            if isinstance(part, CompactionPart):
+                call_id_by_tool_call_id.clear()
+                loaded.clear()
```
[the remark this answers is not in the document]
DouweM: Confirmed and fixed in 4f04637fb — thanks, this was a real find. Verified before fixing: with the reset in place, the tool stays dispatchable while `_ctx_for_available_cap` returns `None`, so the owner's `before/after/wrap_tool_validate`/`_execute` hooks (approval gates included) were skipped for its own tool call. Notably, the same bypass already existed on the base branch via the fabricated-history route (reveal evidence with no load pair) — the regression test for that route fails on base too; this PR just made the state reachable in legitimate runs.

Went with your second option, scoped to where it matters: the tool-scoped dispatchers in `CombinedCapability` now force the owning capability's hooks for its *own* tool (`_ctx_for_tool_hook`), with `capability_loaded` kept truthfully `False` so hooks can distinguish the states. Rejecting the call outright (your first option) would contradict the deliberate keep-callable semantics pinned by the trust-boundary tests, and cross-cutting hooks of *unrelated* unloaded capabilities remain skipped as before. Both routes are pinned by mutation-verified tests.

pydantic_ai_slim/pydantic_ai/toolsets/_tool_search.py
```diff
@@ -221,7 +227,9 @@ def parse_discovered_tools(messages: Sequence[ModelMessage]) -> set[str]:
                     _collect_legacy(part.metadata, discovered)
         elif isinstance(msg, ModelResponse):
             for part in msg.parts:
-                if isinstance(part, NativeToolSearchReturnPart):
+                if isinstance(part, CompactionPart):
```
DouweM: Would it be more efficient to scan from the end? Similar to #7225

DouweM: Done in dee422c78 — both parsers (`parse_discovered_tools` and `parse_loaded_capabilities`) now scan from the end and stop at the boundary instead of parsing history the boundary would reset anyway. The discovery collector is a set so order doesn't matter; the capability pairing works in reverse by collecting returns first and completing pairs when their call appears, which admits exactly the same fully-post-boundary pairs as the forward scan (the split-pair test pins it).

Same commit also addresses the result-slot concern you raised: undiscovered-first is now the PRIMARY sort key (relevance is the tiebreak), so an already-available tool can never displace an undiscovered match when `max_results` trims — it only fills leftover slots, which is pingxia's original "appended when the result limit allows". The "how does the model know there are more matches" half predates this PR (pure undiscovered overflow was always silently trimmed) — filed as #7232 to add a truncation hint to the search return content.

DouweM:
Follow-up to the review discussion: the compaction boundary is now public API — [`pydantic_ai.messages.compacted_window`][1] returns the messages from the latest `CompactionPart` onward (part-level precision), and both parsers are rebuilt on it as the single definition of the boundary. Capability authors are pointed at it in the compaction docs for deriving model-visible state (announcements, disclosures, catalogs) instead of instance attributes — the pattern that bit `CodeMode` (pydantic-ai-harness#562) and `DynamicWorkflow` (pydantic-ai-harness#563).

Also pinned: `test_repeated_searches_paginate_through_a_large_corpus` proves repeat-to-paginate enumeration of a >`max_results` corpus still works under the new ranking — each page's results sink after discovery, so identical searches walk the whole corpus; the explicit truncation/pagination signal is #7232.

[1]: https://github.com/pydantic/pydantic-ai/blob/tool-reveal-compaction/pydantic_ai_slim/pydantic_ai/messages.py

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -2559,6 +2559,38 @@ def otel_message_parts(self, settings: InstrumentationSettings) -> list[_otel_me
...
+    (announcements, disclosures, catalogs) should be recomputed from this window rather than
+    remembered in instance attributes, so it self-heals when compaction replaces the history that
+    carried it.
+
+    Deliberately provider-agnostic: a compaction part another provider would skip on the wire
+    still counts as a boundary. Treating too little as visible only permits a redundant,
+    idempotent re-disclosure; treating too much as visible hides state the model can no longer
+    see.
+    """
+    for message_index in range(len(messages) - 1, -1, -1):
```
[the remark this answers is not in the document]
DouweM: Confirmed and fixed in 8afa223d6 — fair catch on a public API: typeshed's `Sequence` promises the slice overload (which is why this typechecked), but the runtime ABC contract only requires integer `__getitem__`, so a minimal conforming implementation would have exploded only when a compaction part was present. The message suffix is now built by indexed iteration (`message.parts` is always a `list`, so its slice stays), pinned by `test_compacted_window_accepts_a_minimal_sequence` — an int-only `Sequence` whose typing declares the slice overload and whose runtime refuses it, which is exactly the gap.

pydantic_ai_slim/pydantic_ai/messages.py
```diff
@@ -2561,6 +2561,41 @@ def otel_message_parts(self, settings: InstrumentationSettings) -> list[_otel_me
 """Pydantic [`TypeAdapter`][pydantic.type_adapter.TypeAdapter] for (de)serializing messages."""
 
 
+def post_compaction_window(messages: Sequence[ModelMessage]) -> list[ModelMessage]:
```
DouweM, line 2564: Shouldn't the post-compaction window be system-specific?

DouweM: Considered and deliberately not — the two provider-sensitivity choices answer different questions, and I've expanded the docstring to say so (086a2eccf):

- **The wire trim (#7228) is system-specific** because it must be exact for the one request it renders: a foreign compaction blob can't replay on this provider, so its pre-boundary content is still needed on the wire (`test_openai_responses_foreign_compaction_does_not_trim` pins that).
- **This window is provider-agnostic** because it feeds *run-level* state — `RunContext.discovered_tool_names`, loaded capabilities — that must stay valid across `FallbackModel` failover and mid-run model switches. At parse time there is no "current" provider to resolve against (the parsers run without a model), and the state outlives whichever provider serves the next request, so the boundary has to be the conservative intersection: any provider's compaction counts.

The asymmetry is what makes the conservative choice safe rather than sloppy: under-counting visibility (a cross-provider handoff where the *next* provider still renders the pre-boundary reveal) costs at most a redundant, idempotent re-disclosure; over-counting (treating a pre-boundary reveal as visible when the same provider's trim just dropped its carrier) hides a tool the model genuinely can't see — the failure class this PR exists to close. A system-specific variant would also need the parse→render provider to match, which model switching breaks by construction.

PR #7227, opened 2026-08-06
Document Bedrock client timeouts and their environment fallbacks

ayaangazali:
`BedrockProvider` takes `aws_read_timeout` and `aws_connect_timeout`, each falling back to `AWS_READ_TIMEOUT` / `AWS_CONNECT_TIMEOUT` and then to 300 and 60 seconds, but the Bedrock page never mentions timeouts. Long generations are exactly where people hit the read timeout and go looking for the knob.

Adds a Timeouts section to `docs/models/bedrock.md` with a table of argument, environment variable and default, a short example, and a note that a prebuilt `bedrock_client` is used as-is so its timeouts belong on the client.

This reopens #7222, which the PR guard closed automatically. That one also touched the two argument docstrings in `providers/bedrock.py`, which put it outside the `docs/*|*.md|mkdocs.yml` docs-only exemption in `.github/workflows/pr-guard.yml`, so it was correctly caught by the rule. This version is documentation only. Happy to follow up on the docstrings behind an issue if that information is worth having in the API reference too.

The documented defaults and fallbacks are the ones `test_bedrock_provider_timeout` in `tests/providers/test_bedrock.py` already pins, by setting both environment variables and asserting they reach `config.read_timeout` and `config.connect_timeout`. Ran that test, `uv run pytest tests/test_examples.py -k bedrock` (21 passed, covering the new example), and `uv run mkdocs build --strict` with no new warnings.

### Checklist

---

quick note: built with Claude Code's help and read the diff myself. my mistake bundling a docstring change into a docs PR the first time, that is what tripped the guard.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7227?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/models/bedrock.md
+++ b/docs/models/bedrock.md
@@ -87,7 +87,8 @@ agent = Agent(model)
 ```
 
 These apply to clients the provider builds itself. A prebuilt `bedrock_client` is used as-is, so
-configure its timeouts on the client.
+configure its timeouts on the client. See [Timeouts](../timeouts.md) for how request timeouts work
+across model classes, including why `ModelSettings['timeout']` does not reach Bedrock.
 
 ### Customizing Bedrock Runtime API
 
--- a/docs/models/bedrock.md
+++ b/docs/models/bedrock.md
@@ -62,6 +62,33 @@ agent = Agent(model)
 ...
 ```
 
+### Timeouts
+
+Long generations can outlast the Bedrock client's default timeouts. Both are configurable on
+[`BedrockProvider`][pydantic_ai.providers.bedrock.BedrockProvider], and each falls back to an
+environment variable when omitted:
+
+| Argument | Environment variable | Default |
+| --- | --- | --- |
+| `aws_read_timeout` | `AWS_READ_TIMEOUT` | 300 seconds |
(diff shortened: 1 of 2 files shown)


ayaangazali:
Rebased on main and added a cross-link in c31ab768c, since #7145 landed the new Timeouts page while this was open.

The two do not overlap. `docs/timeouts.md` says `BedrockProvider` takes `aws_read_timeout` and `aws_connect_timeout`; it does not give the `AWS_READ_TIMEOUT` / `AWS_CONNECT_TIMEOUT` fallbacks or the 300 and 60 second defaults, which is what this section adds. Provider-specific config on the provider page with the general page linking out also matches the convention in `docs/CLAUDE.md`.

Happy to move the table onto the Timeouts page instead if you would rather keep all timeout details in one place.

ayaangazali:

PR #7256, opened 2026-08-07
Forward events from `UIEventStream` callbacks that return an async iterator without being an async generator function

DouweM:
A self-found bug from Macroscope's re-review of #6324; the affected code shipped on `main` via #6497.

`UIEventStream`'s `on_complete` / `on_cancel` callbacks are typed as `_CallbackFunc`, whose third arm is `Callable[[arg], AsyncIterator[EventT]]` — a callback that emits protocol events. `_dispatch_callback` only forwarded those events when `inspect.isasyncgenfunction(callback)` was true, which holds for the common `async def ... yield` form but **not** for two other shapes the type admits:

- a plain `def` that *returns* an async iterator (`def cb(arg): return agen()`), and
- a callable instance whose `__call__` is an async generator.

Both fell through to the executor branch, where the returned async iterator was constructed and then discarded — the callback's protocol events were silently dropped.

The fix inspects the returned value in the fallback branch and forwards an async iterator (or awaits a returned awaitable), mirroring the existing probe pattern in `ProcessEventStream.wrap_run_event_stream`. Genuinely-sync callbacks still run in the executor as before.

### Tests

Regression tests in `tests/test_ui.py`, parametrized over both dropped shapes (plain `def` returning an async generator, and a callable instance with an async-generator `__call__`), plus a case for a plain `def` returning an awaitable. Verified failing before / passing after.

### Follow-ups

- A related but distinct sibling class — dispatch sites that use `is_async_callable` and mishandle a plain `def` returning a coroutine — is catalogued in #7257 rather than widened into this fix.
- Worth considering a shared dispatch helper (the "call, then inspect the returned value" pattern) so `_dispatch_callback`, `ProcessEventStream`, and the #7257 sites don't each re-implement it. Left for the #7257 follow-up.

### Checklist

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/ui/_event_stream.py
+++ b/pydantic_ai_slim/pydantic_ai/ui/_event_stream.py
@@ -320,12 +320,24 @@ class UIEventStream(ABC, Generic[RunInputT, EventT, AgentDepsT, OutputDataT]):
         self, callback: _CallbackFunc[_CallbackArgT, EventT], arg: _CallbackArgT
     ) -> AsyncIterator[EventT]:
         if inspect.isasyncgenfunction(callback):
+            # Fast path for the common `async def ... yield` form.
             async for event in callback(arg):
                 yield event
         elif _utils.is_async_callable(callback):
+            # `async def ... return None`, or a callable object with a coroutine `__call__`.
             await callback(arg)
         else:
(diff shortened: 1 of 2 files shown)


DouweM:
Sibling dispatch sites sharing the related `is_async_callable` variant are catalogued in #7257 (kept out of this PR to avoid widening it across core paths + a contract decision).

PR #7277, opened 2026-08-07
Reject `Agent.run_sync()` from synchronous callbacks inside agent runs

### LINKED ISSUE #3899: Tool output function with delegated another Agent inside, does not work reliably in sync environment. The `run_sync` freezes.
### Initial Checks

### Description

I've noticed the weird type of bug while working with run_sync agents.

Suppose we have two agents. Something like this:

    response_agent = Agent(
        deps_type=SomeKindOfDataclass,
        output_type=str,
        system_prompt="Make human readable response. ",
    )

    def get_humanized_response(context: RunContext, instructions: str) -> str:
        result = response_agent.run_sync(
            instructions,
            usage=context.usage,
            deps=context.deps,
        )
        return result.output

    orchestrator_agent = Agent(
        deps_type=SomeKindOfDataclass,
        output_type=[get_humanized_response],
    )

So I have main orchestrator agent, that makes all complicated work with different tools, and then it must pass instructions to the response agent, that will generate human readable text that will be shown to the user.

The problem is that when orchestrator agent is being run with `run_sync` sometimes it just freezes. This makes overall delegation to the different agents as output tool unreliable in sync environment.

---

P.S:

I have also noticed that GLM 4.7 through openrouter breaks in configuration above. It returns this error:

     {'error': {'message': "No endpoints found that support the provided 'tool_choice' value. To learn more about provider routing, visit: https://openrouter.ai/docs/guides/routing/provider-selection", 'code': 404}}

- Python: 3.12
- Pydantic AI: 1.39.0
- LLM provider SDK: Openrouter

dsfaccini:
- Closes #3899

Synchronous tools and output functions normally run in worker threads. Calling `Agent.run_sync()` there can create a second event loop while the parent run is waiting, then deadlock if the nested run touches an async resource bound to the parent loop.

This PR turns that hang into an actionable `UserError` and directs users to an `async def` callback with `await agent.run(...)`.

### Implementation

- Set a reset-safe `ContextVar` around functions dispatched through Pydantic AI's executor helper.
- Check `run_sync()` and `run_stream_sync()` before starting synchronous agent execution.
- Apply the same `run_sync()` check to the DBOS and Prefect overrides.
- Document the supported async delegation pattern.

### Tests

- Reject `run_sync()` from a synchronous output function, matching the reported failure.
- Reject `run_stream_sync()` from a synchronous tool.

### Checklist


vedaant00:
The docs read accurately and the mechanism matches what we pinned down, nice writeup. One optional thought, entirely your call: docs-only leaves the silent hang itself in place, so someone who doesn't land on these pages still gets a frozen process with no error. If you ever want to turn that into a signal, the non-breaking version is a ContextVar set during a run, checked at the top of run_sync, emitting a warning (not a raise) that points at the async pattern. It only fires when a run is already in progress, so top-level run_sync and serverless are untouched. Happy to open it as a separate PR if useful, otherwise this is good to merge as-is @dsfaccini.

dsfaccini:
assigned to @adtyavrdhn per our pyai team call on Fri Aug 7 @ 11:40

adtyavrdhn:
@macroscope-app Review

PR #7321, opened 2026-08-09
test: bound model name HTTP lookups

### LINKED ISSUE #7322: Bound live model-name HTTP lookups in tests
### Description

`tests/models/test_model_names.py` makes live HTTP requests to the Heroku and Cerebras model-list endpoints while checking known model names. Those calls do not currently pass an explicit timeout.

If either external endpoint accepts a connection but stalls, the test helper can wait longer than intended during local or CI runs. Adding a small explicit timeout keeps these live-network checks bounded while preserving the existing behavior that skips when the endpoint returns a non-200 status.

### Example

The affected call sites are:

- `httpx.get("https://us.inference.heroku.com/available-models")`
- `httpx.get("https://api.cerebras.ai/v1/models", ...)`

### Suggested fix

Use a shared timeout constant for both model-list HTTP lookups.

Ghraven:
Closes #7322

## Problem

`tests/models/test_model_names.py` reaches the Heroku and Cerebras model-list endpoints with `httpx.get()` calls that did not pass a timeout. If either external service accepts a connection but stalls, this test helper can wait longer than intended during local or CI model-name checks.

## Before / after behavior

Before, the live model-name lookups relied on the client default behavior at each call site.

After, both lookups share a small `MODEL_LIST_HTTP_TIMEOUT` constant, so the test helper has an explicit network bound while preserving the existing status-code skip behavior.

## Verification

- `python -m ruff format --check tests/models/test_model_names.py`
- `python -m ruff check tests/models/test_model_names.py`
- `python -m compileall tests\models\test_model_names.py`
- Attempted `python -m pytest tests/models/test_model_names.py::test_known_model_names_accessor` with this checkout on `PYTHONPATH`; the test is skipped locally because optional model packages are not installed in this environment.
- `git diff --check`

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7321?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/tests/models/test_model_names.py
+++ b/tests/models/test_model_names.py
@@ -278,8 +278,11 @@ class HerokuModel(TypedDict):
     type: list[str]
 
 
+MODEL_LIST_HTTP_TIMEOUT = 10
+
+
 def get_heroku_model_names():
-    response = httpx.get('https://us.inference.heroku.com/available-models')
+    response = httpx.get('https://us.inference.heroku.com/available-models', timeout=MODEL_LIST_HTTP_TIMEOUT)
 
     if response.status_code != 200:
         pytest.skip(f'Heroku AI returned status code {response.status_code}')  # pragma: lax no cover
@@ -306,6 +309,7 @@ def get_cerebras_model_names():  # pragma: lax no cover
     response = httpx.get(
         'https://api.cerebras.ai/v1/models',
         headers={'Authorization': f'Bearer {api_key}', 'Accept': 'application/json', 'Accept-Encoding': 'identity'},
+        timeout=MODEL_LIST_HTTP_TIMEOUT,
     )
 
     if response.status_code != 200:


Ghraven:
Thank you for the guidance. I created tracking issue #7322 and updated the PR description with `Closes #7322`.

For local verification, I ran:
- `python -m ruff format --check tests/models/test_model_names.py`
- `python -m ruff check tests/models/test_model_names.py`
- `python -m compileall tests\models\test_model_names.py`
- `git diff --check`

I also attempted the narrow pytest invocation with this checkout on `PYTHONPATH`, but it is skipped in my local environment because optional model packages are not installed. The change is limited to passing an explicit timeout into the two existing live HTTP lookups. I will leave the PR closed unless maintainers confirm this issue is worth reopening. Thank you!

PR #7360, opened 2026-08-10
Use ASCII units for microsecond eval durations

### LINKED ISSUE #7291: ASCII fallback in `EvaluationReport.print()` misses the `µs` duration unit, so sub-millisecond durations still raise `UnicodeEncodeError`

`EvaluationReport.print()` still raises `UnicodeEncodeError` on a non-UTF-8 console when any duration is sub-millisecond: `_render_duration` emits `µ` (U+00B5), which the `ascii_only` fallback added in #7290 never reaches.

#7290 covers `✔` / `✗` / `→` and fixes the reported cp1252 case in #7281 (cp1252 encodes `µ` at 0xB5, so that environment renders — just with a stray `µ` in an otherwise ASCII table). Code pages that **cannot** encode `µ` still crash: `ascii`, `cp932` (JP), `cp936` (zh-Hans), `cp949` (KO), `cp950` (zh-Hant), `cp866`, `cp874`, `koi8-r`.

Sub-millisecond `task_duration` is routine, not contrived — a plain `Dataset(...).evaluate_sync(...)` over a pure function gives `497µs`.

```
print() → cp1252: renders "| v | 0.4µs |"   (half-degraded)
print() → cp932 : UnicodeEncodeError: 'cp932' codec can't encode character '\xb5' in position 275
print() → ascii : UnicodeEncodeError: 'ascii' codec can't encode character '\xb5' in position 275
```

Origin: the `µs` unit was introduced in #935.

### Why this needs its own PR

The fix has to cross a public contract, which is why #7290 scoped it out:

1. `default_render_duration` and `default_render_duration_diff` are in `pydantic_evals.reporting.render_numbers.__all__` and are wired in as `_DEFAULT_DURATION_CONFIG`'s `value_formatter` / `diff_formatter`. Threading a flag through them changes their public signatures.
2. `duration_config` is user-supplied on `print()` / `console_table()`, so any fix that only swaps the *default* config leaves a user-supplied formatter still emitting `µ`. Whatever shape is chosen has to say what happens to user-supplied formatters — degrade their output, or document that they own it.
3. `tests/test_examples.py` monkeypatches `render_numbers._render_duration` with a two-arg mock (`mock_render_duration(seconds, force_signed)`), which must be updated in step with any signature change.

<details>
<summary>Repro</summary>

```python
from io import BytesIO, TextIOWrapper

from rich.console import Console

from pydantic_evals.reporting import EvaluationReport, ReportCase

case = ReportCase(
    name='fast_case',
    inputs={},
    output={},
    expected_output={},
    metadata={},
    metrics={},
    attributes={},
    scores={},
    labels={},
    assertions={},
    task_duration=5e-7,
    total_duration=5e-7,
    trace_id='',
    span_id='',
)
report = EvaluationReport(cases=[case], name='r')

for encoding in ('cp1252', 'cp932', 'ascii'):
    stream = TextIOWrapper(BytesIO(), encoding=encoding, errors='strict', newline='')
    try:
        report.print(console=Console(file=stream, width=120))
        print(encoding, 'ok')
    except UnicodeEncodeError as e:
        print(encoding, 'RAISED', e)
```
</details>

uuzzrm:
- Fixes #7291

### Summary

This follow-up applies the ASCII-only console fallback to the default microsecond duration formatters. `EvaluationReport.print()` now renders `us` instead of `µs` when Rich reports an ASCII-only console, while direct formatter calls and custom duration formatters keep their existing behavior.

The change is based on #7290 and targets its head branch so the follow-up stays separate from the already-reviewed glyph fallback.

### Verification

- Added direct coverage for ASCII-only duration values and diffs.
- Added a `cp932` `EvaluationReport.print()` regression test for sub-millisecond durations.
- `git diff --check` passes against the #7290 head.

### Checklist

AI assistance was used for the implementation and test drafting. Please review the diff line by line before checking the first item above.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7360?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_evals/pydantic_evals/reporting/__init__.py
+++ b/pydantic_evals/pydantic_evals/reporting/__init__.py
@@ -3,6 +3,7 @@ from __future__ import annotations as _annotations
 from collections import defaultdict
 from collections.abc import Callable, Mapping, Sequence
 from dataclasses import dataclass, field
+from functools import partial
 from io import StringIO
 from typing import Any, Generic, Literal, Protocol
 
@@ -480,9 +481,8 @@ class EvaluationReport(Generic[InputsT, OutputT, MetadataT]):
         """Print this report to the console, optionally comparing it to a baseline report.
 
         On a console `rich` reports as ASCII-only — a non-UTF-8 stream, as Windows produces when stdout is
-        redirected to a file or a pipe — `v`, `x` and `->` stand in for the `✔`, `✗` and `→` glyphs. The `µs`
-        unit sub-millisecond durations render with is not covered yet, so a stream that can't encode `µ` can
-        still raise: see [#7291](https://github.com/pydantic/pydantic-ai/issues/7291).
+        redirected to a file or a pipe — `v`, `x`, `->` and `us` stand in for the `✔`, `✗`, `→` and `µs`
+        glyphs and units.
 
         If you want more control over the output, use `console_table` instead and pass it to `rich.Console.print`,
         forwarding `ascii_only=console.options.ascii_only` from the console you print to.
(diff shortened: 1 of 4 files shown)


uuzzrm:
This is ready for maintainer merge at head `7997687`. All applicable lint, mypy, docs, and test checks are green, and Macroscope has approved the current head. The `uuzzrm` account has pull-only permission in this repository, so I cannot execute the upstream merge.

Issue #7109, opened 2026-08-03
Durable-exec leaf-toolset `id` error never mentions the capability `id` that would fix it


dsfaccini:
This issue was posted by Claude Code using claude-opus-5 on behalf of David.

`durable_exec/_base.py::BaseDurabilityCapability._wrap_and_register_leaf` raises two variants of the same "leaves need a unique `id`" error. The `DynamicToolset` branch tells the user how to fix it; the generic branch — the one capability-contributed `FunctionToolset`s actually hit — doesn't.

Generic branch, verbatim:

> Toolsets that are 'leaves' (i.e. those that implement their own tool listing and calling) need to have a unique `id` in order to be used with Temporal. The ID will be used to identify the toolset's activitys within the workflow.

The `DynamicToolset` branch additionally says: "Set the dynamic toolset ID with `DynamicToolset(id=...)`, or, when it is contributed by a capability, set the capability's `id` (for example, `DynamicCapability(..., id='user-tools')`)."

Repro:

```python
Agent('openai:gpt-5', name='a', capabilities=[WebSearch(local='duckduckgo'), TemporalDurability(name='x')])
```

Same for `ImageGeneration(local=...)` and `ImageGeneration(fallback_model='openai:gpt-5')` — any `NativeOrLocalTool` contributing a local toolset. `capabilities/native_or_local.py::NativeOrLocalTool.get_toolset` builds `FunctionToolset([...], id=self.id)`, and `AbstractCapability.id` defaults to `None`, so the explicit-id contract established in #6408 is exactly the right answer; the message just never points there. `WebSearch(local='duckduckgo', id='search')` constructs fine.

Engine split: Temporal and Prefect raise (Prefect's wording ends "tasks within the flow"). DBOS passes, because `durable_exec/dbos/_durability.py::DBOSDurability._wrap_leaf_toolset` only wraps `DynamicToolset`/`MCPToolset` and returns `None` otherwise.

Adding the capability hint to the generic branch's message repairs every affected capability at once. (`activitys` in the Temporal wording is a typo worth fixing in the same pass.)

Found while reviewing #5357; not fixed there because it pre-dates that PR.

PR #7370, opened 2026-08-11
Experiment: give single-purpose capabilities stable default `id`s

dsfaccini:

This is a ready-for-review experiment comparing an alternative design with #7248. It is intentionally not a draft: the purpose is to get normal reviewer feedback on the behavioral trade-off before choosing which implementation to keep.

## Problem

Single-purpose capabilities such as `WebSearch(local='duckduckgo')` contribute a leaf toolset without an `id`. Durable execution requires stable, unique leaf IDs, so a normal capability configuration currently needs an otherwise-redundant explicit `id`.

## Experimental design

This branch gives the four single-purpose built-ins stable default capability IDs:

| Capability | Default `id` |
|---|---|
| `WebSearch` | `'web_search'` |
| `WebFetch` | `'web_fetch'` |
| `ImageGeneration` | `'image_generation'` |
| `XSearch` | `'x_search'` |

Their local fallback toolsets inherit the same ID. `MCP` is deliberately excluded because its identity depends on the configured server; an explicitly supplied empty ID remains distinct from an unset ID.

Unlike #7248, this makes the capability itself semantically identifiable. Consequences are intentional:

- two instances using the same default ID are rejected as duplicate capabilities;
- callers can assign distinct explicit IDs when the instances represent distinct roles;
- callers can pass `id=None` to opt out of stable identity and retain stackable, auto-deduplicated occurrences;
- `defer_loading=True` works for these built-ins without an extra `id=` argument.

## Shared durability fix

Both implementations need to preserve construction-time capability toolsets when assembling a run. Re-extracting an unchanged capability can create a new leaf object with the same ID; treating that object as either a registered static leaf or a genuine runtime replacement by ID alone is unsafe.

The agent therefore caches construction capability/toolset occurrences and reuses an entry only when the exact unchanged construction capability occurrence survives into the run. Replaced, overridden, dynamic, and runtime occurrences are normalized afresh. Durable wrapper selection remains strict: only the registered leaf object can select its wrapper, and a distinct leaf with the same ID raises instead of silently substituting tools.

## A/B comparison

| Scenario | #7248 | this experiment |
|---|---|---|
| one built-in with durable execution | works | works |
| two default instances on a plain agent | allowed; generated occurrence IDs | rejected as duplicate semantic IDs |
| two instances with distinct explicit IDs | allowed | allowed |
| two instances with `id=None` | allowed; generated public toolset IDs | allowed; id-less public toolsets |
| `defer_loading=True` without explicit ID | rejected | allowed |
| unchanged static capability plus safe runtime instrumentation | works | works |
| rebuilt leaf with an already-registered ID | rejected before model execution | rejected before model execution |

The main review question is whether these one-purpose built-ins should have stable semantic identity by default, even though that makes duplicate default instances an explicit configuration error, or whether preserving the existing stacking behavior is more important and only their contributed toolsets should receive stable IDs.

Related to #7109 and #7248.

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7370?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


DouweM:
Closing as superseded by #7248, which is the chosen lane for default capability ids (this PR was the A/B experiment). The duplicate-stacking rejection explored here fed the design discussion, and the requirement it served is now also enforced from the durable side: #6696's contributed durable operations require exactly one capability per stable id, with bind-time errors. Remaining identity work continues on #7695 -> #7248 -> #7144.

*This comment was posted by Claude Code on behalf of Douwe.*

PR #7399, opened 2026-08-12
fix: honor cancellation during setup hooks

### LINKED ISSUE #7386: First-party cancellation during the setup-phase hooks: `for_run` `ctx.cancel()` raises, and a pre-cancelled token cannot interrupt setup
### Summary

Two related gaps in first-party cancellation, both rooted in the same thing: the run's cancellation machinery isn't wired up yet when the setup-phase hooks (`_resolve_run_capabilities`, `toolset.for_run(initial_ctx)`) run. Both are design consequences of #6497 (not regressions), and both are best resolved inside the setup-phase **hook-contract rework** (#7053) rather than as standalone patches — filing here so they aren't lost. Surfaced by Macroscope's re-review of #6324 (threads linked below); the code is byte-identical to `main`.

### 1. `ctx.cancel()` from a capability/toolset `for_run` hook raises `UserError`

- **Where:** `pydantic_ai_slim/pydantic_ai/_agent_graph.py` — the initial `RunContext` construction / `GraphAgentDeps.cancellation`.
- **What:** the `RunContext` passed to `for_run()` hooks has `_cancellation=None`, because the `RunCancellation` controller is constructed on `GraphAgentDeps` *after* the initial context and capability resolution. So `ctx.cancel()` during run initialization raises `UserError` ("no run to cancel") instead of requesting cancellation. `RunCancellation.bind()` can already deliver a pre-bind cancellation later — the controller just needs to exist and be wired into the initial context.
- **Fix direction:** create the `RunCancellation` before `initial_ctx`/capability resolution and share the one instance between `RunContext._cancellation` and `GraphAgentDeps.cancellation`.
- Thread: https://github.com/pydantic/pydantic-ai/pull/6324#discussion_r3732493116

### 2. A pre-cancelled `cancellation_token` can't interrupt the setup phase

- **Where:** `pydantic_ai_slim/pydantic_ai/agent/__init__.py` `iter()` — `attach_token` runs *after* `_resolve_run_capabilities` and `toolset.for_run(initial_ctx)` have awaited.
- **What:** if a capability's or toolset's `for_run` blocks, cancelling the token has no effect (the controller isn't registered with the token yet), so the caller hangs; and a pre-cancelled token doesn't prevent setup from running at all. #6497's "bind before potentially-blocking setup" comment covers the concurrency limiter and model entry, but capability/toolset resolution happens even earlier.
- **Fix direction:** attach the token before those setup hooks run.
- Thread: https://github.com/pydantic/pydantic-ai/pull/6324#discussion_r3732493122

### Why together, and why #7053

Both are the "a hook runs before the run's resolved state exists" shape (there: model / tool_manager; here: the cancellation controller and token). Whether a `for_run`/setup hook *should* be able to observe or trigger cancellation is a contract question that #7053's "which hooks wrap what" rework owns. Fixing the ordering in isolation would pre-empt that decision, so these should be resolved as part of it.

### Links
- #6497 (first-party cancellation), #6460 (contract), #7053 (setup-phase hook-contract rework)

_Filed by Claude Code on behalf of @DouweM._

jbdrak:
Fixes #7386.

## Summary

The run cancellation controller was attached after setup-phase capability and toolset hooks ran. That caused two failures:

- `RunContext.cancel()` from a `for_run` hook raised `UserError`.
- A pre-cancelled token, or a token cancelled while setup was blocked, couldn't stop setup before the run began.

This change creates and binds the run controller before setup, exposes it through the initial `RunContext`, translates setup-phase cancellation into `RunCancelled`, and explicitly skips setup for an already-cancelled token. It doesn't add a public API and leaves realtime-session behavior unchanged because that surface doesn't accept a cancellation token.

## Verification

- `uv run pytest tests/test_run_cancellation.py` - 69 passed, 1 skipped
- `uv run ruff format --check pydantic_ai_slim/pydantic_ai/agent/__init__.py tests/test_run_cancellation.py`
- `uv run ruff check pydantic_ai_slim/pydantic_ai/agent/__init__.py tests/test_run_cancellation.py`
- `uv run pyright pydantic_ai_slim/pydantic_ai/agent/__init__.py tests/test_run_cancellation.py` - 0 errors
- `git diff --check`

The repository-wide `make typecheck-pyright` remains blocked by existing missing optional dependencies and unrelated errors in the full tree; the changed files pass standalone Pyright.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7399?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/agent/__init__.py
+++ b/pydantic_ai_slim/pydantic_ai/agent/__init__.py
@@ -1308,6 +1308,11 @@ class Agent(AbstractAgent[AgentDepsT, OutputDataT]):
         # run would otherwise consume it and attach the outer handle to the wrong run.
         binding = take_run_binding()
 
+        # The run's cancellation controller: owned by the `AgentRunEvents` binding when one started
+        # this run, otherwise fresh. Created up front so it can be bound and the token attached
+        # before any setup-phase hook runs (`_resolve_run_capabilities` / `toolset.for_run` below).
+        run_cancellation = binding.cancellation if binding is not None else RunCancellation()
+
         # A bare `int` overrides both budgets; a partial `retries={'tools': ...}` / `{'output': ...}`
         # dict overrides only the named budget for this run (riding `ToolManager.default_max_retries`).
         retry_overrides = _normalize_agent_retry_overrides(retries)
@@ -1548,6 +1553,7 @@ class Agent(AbstractAgent[AgentDepsT, OutputDataT]):
             pending_messages=state.pending_messages,
             run_id=state.run_id,
             conversation_id=state.conversation_id,
+            _cancellation=run_cancellation,
         )
 
         # Resolve run metadata up front so capability and toolset `for_run` hooks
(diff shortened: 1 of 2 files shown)


adtyavrdhn:
I merged a PR for this already. Thank you!

PR #7407, opened 2026-08-12
Add `blockbuster` to the test suite to detect blocking calls in the event loop

Kludex:

Runs tests under a [BlockBuster](https://github.com/cbornet/blockbuster) guard that raises `BlockingError` when a blocking call occurs inside the event loop with a `pydantic_ai`, `pydantic_graph`, `pydantic_evals`, or `clai` frame on the stack. Test-only and third-party-only stacks are ignored, while user callbacks and tools remain covered when a scanned library frame is below them.

The initial run surfaced 177 violations, resolved as follows:

- **Library fix**: web chat UI cache and local-file access is offloaded as complete synchronous operations. Cache reads and atomic writes are serialized to keep Windows replacement safe, and cancellation cannot abandon local filesystem workers beyond AnyIO's thread limit.
- **Documented exemptions** (`BLOCKBUSTER_EXEMPTIONS`): sync-by-design APIs, setup-time SDK construction, source introspection, and coverage.py 7.13's Python 3.14 `sys.monitoring` source reads. Anthropic's async Bedrock credential/signing I/O is temporarily exempted at its exact SDK auth frames and tracked by [anthropic-sdk-python#1770](https://github.com/anthropics/anthropic-sdk-python/issues/1770).
- **Narrow integration boundaries**: CLI and Prefect exclude only their intentionally synchronous integration modules, one VoyageAI docs example disables detection around pytest-examples' synchronous output reader, and Temporal remains disabled because its worker/integration setup both blocks and becomes unusably slow under BlockBuster's current stack inspection.
- **Lower fixture overhead**: each pytest worker configures one detector per exclusion set, then activates and deactivates it for each test. Redundant Python-version, lowest-version, and all-extras CI lanes disable the guard; Python 3.13 slim/evals/standard lanes and unique compatibility/live jobs retain it so every shard stays within the seven-minute budget.
- **Fixture isolation**: configuration failures, partial activation failures, test failures, and teardown all restore process-global wrappers, including immutable built-ins patched through ForbiddenFruit.

The fixture contract and exemption policy are documented in `tests/AGENTS.md` and covered by focused regression tests. The remaining upstream cleanup and performance work is tracked in [BlockBuster#59](https://github.com/cbornet/blockbuster/issues/59) and [BlockBuster#54](https://github.com/cbornet/blockbuster/issues/54).

### Checklist

## AI Disclaimer

This PR was developed with the assistance of either Claude or Codex. I've reviewed and verified the changes.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7407?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


pydantic_ai_slim/pydantic_ai/ui/_web/app.py
Kludex, line 1: Can we use anyio file API instead of to_thread on std lib files?

Kludex: Done in 25e5648 - `anyio.Path` everywhere except the atomic tempfile+`os.replace` write, which stays sync behind one `to_thread` call: `anyio.NamedTemporaryFile` needs anyio 4.9 and our floor is 4.7.

Kludex:
The failing `agent` check (`Pydantic AI UI Security Review`) is broken infra, not this PR: the shim dies at startup because the `Checkout PR branch` step fetches the PR head with `--depth=3` (branch commit count), leaving no release tag reachable from the head - `uv-dynamic-versioning` then computes `pydantic-ai-slim==0.0.1.dev3+25e564872`, which conflicts with `pydantic-ai-harness==0.7.0`'s `pydantic-ai-slim>=2.1.0` and the script resolution fails before the reviewer runs.

Evidence: identical failure on both pushes ([run 1](https://github.com/pydantic/pydantic-ai/actions/runs/31606797807), [run 2](https://github.com/pydantic/pydantic-ai/actions/runs/31607940329): `0.0.1.dev2`/`0.0.1.dev3` tracking my commit count), while the last real execution ([capability-prepares-own-tools, Aug 12 00:48](https://github.com/pydantic/pydantic-ai/actions?query=workflow%3A%22Pydantic+AI+UI+Security+Review%22), fetched `--depth=34`) resolved `pydantic-ai-slim==2.27.2.dev37` fine. Every 'success' since is the `agent` job being skipped on non-UI PRs - this is the first UI-touching PR since. No workflow/shim files changed since Aug 3, and this PR touches neither.

.github/workflows/ci.yml
```diff
@@ -325,6 +328,7 @@ jobs:
     env:
       CI: true
       COVERAGE_PROCESS_START: ./pyproject.toml
+      BLOCKBUSTER_ENABLED: false
```
Kludex, line 333: Why do we have this?

pydantic_ai_slim/pydantic_ai/ui/_web/app.py
```diff
@@ -37,7 +40,10 @@
 OutputDataT = TypeVar('OutputDataT')
 
 
-def _get_cache_dir() -> Path:
+_CACHE_FILE_LOCK = threading.Lock()
```
Kludex, line 45: Why do we need this lock?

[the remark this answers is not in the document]
dsfaccini: > This reply was posted by Codex Desktop using gpt-5.6-sol on behalf of David.

Good observation. The global lock does serialize distinct cache files, but I am keeping that tradeoff deliberately: it covers only short local reads and writes, not the network fetch, while a per-file registry would add synchronization and lifecycle state without observed cache contention. Commit `0b435f9aa` documents the Windows invariant that requires same-file exclusion.

pydantic_ai_slim/pydantic_ai/ui/_web/app.py
```diff
@@ -71,19 +78,23 @@ def _write_cached_file(cache_file: Path, content: bytes) -> None:
...
+        tmp_path = Path(tmp_file.name)
+        try:
+            # Close the handle before the rename: Windows refuses to replace a file that still has an
+            # open handle, which would break the atomic write there.
+            with tmp_file:
+                tmp_file.write(content)
+                tmp_file.flush()
+            os.replace(tmp_path, cache_file)
+        except BaseException:
+            tmp_path.unlink(missing_ok=True)
```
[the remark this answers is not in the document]
dsfaccini: > This reply was posted by Codex Desktop using gpt-5.6-sol on behalf of David.

This path cannot occur: both `NamedTemporaryFile(...)` and the `tmp_path` assignment are before the `try`. If either operation fails, execution never enters this `except`; whenever the cleanup handler runs, `tmp_path` is already bound. Keeping the code unchanged.

.github/workflows/ci.yml
dsfaccini: > This reply was posted by Codex Desktop using gpt-5.6-sol on behalf of David.

The fixture defaults on, so this explicit `false` is required to exclude the lowest-version matrix. With detection, the Python 3.13 lowest-version job took 7m41s; without it, 5m38s. Detection still runs in the Python 3.13 slim, evals, and standard jobs. Commit `0b435f9aa` adds the rationale here and records that dependency-version-specific blocking behavior is outside CI detection.

pydantic_ai_slim/pydantic_ai/ui/_web/app.py
dsfaccini: > This reply was posted by Codex Desktop using gpt-5.6-sol on behalf of David.

Offloading cache reads and writes to worker threads allows a read and `os.replace()` to overlap. On Windows, replacement may fail while the destination is still open for reading, so the lock preserves the previous same-file serialization. Commit `0b435f9aa` documents that invariant beside the lock; `test_cache_read_and_write_do_not_overlap_on_windows` covers it.

cbornet:
> The remaining upstream cleanup and performance work is tracked in https://github.com/cbornet/blockbuster/issues/59 and https://github.com/cbornet/blockbuster/issues/54.

I published blockbuster 1.5.27 which fixes these issues.

Issue #6713, opened 2026-07-24
Make heavy imports lazy: `import pydantic_ai` eagerly loads `mcp` + `fastmcp`, provider modules load their SDKs at module import


Kludex:
### Description

`import pydantic_ai` currently imports the full MCP client stack even when the application never uses MCP:

```
pydantic_ai/agent/__init__.py
  -> pydantic_ai/capabilities/__init__.py   (from .mcp import MCP)
    -> pydantic_ai/capabilities/mcp.py      (from pydantic_ai.mcp import MCPToolset, ...)
      -> pydantic_ai/mcp.py                 (imports `mcp` SDK + `fastmcp` client/transports)
```

The `try/except ImportError` in `capabilities/mcp.py` only guards the case where the `mcp` extra is not installed - when it is installed, it is always imported.

Measured cold RSS (psutil, after `gc.collect()`, Python 3.14.5, macOS; v2.14.1, reproduced on current main @ f1c7305):

- `import pydantic_ai`: ~113 MB
- of which the `mcp` + `fastmcp` subtree alone: ~80 MB

Separately, provider modules import their vendor SDKs at module level, e.g. `pydantic_ai/providers/google_cloud.py` does `from google.genai.client import Client` at import time. Any app that imports `GoogleCloudProvider` at module level (the normal thing for typed code) pays the full `google.genai` import (~100+ MB with its grpc/proto deps) even if the provider is constructed rarely or never on a given process.

For long-running server apps that import `pydantic_ai` at startup, this is a per-process RSS and cold-start cost for features that may never run, multiplied across workers/pods.

### Proposal

- In `pydantic_ai/capabilities/__init__.py`, re-export `MCP` via a PEP 562 module `__getattr__` instead of an eager import, so `mcp`/`fastmcp` only load when the MCP capability (or `pydantic_ai.mcp`) is actually touched. Same pattern for any other capability with a heavy optional dependency.
- In provider modules, keep SDK imports under `TYPE_CHECKING` and perform the runtime import inside `__init__` (construction time), so importing the provider class for annotations/wiring stays free.

Public API and type checking are unchanged in both cases; the ImportError-with-hint behavior for missing extras can be preserved at first use instead of first import.

Happy to send a PR if the approach sounds right.

Related in spirit to #3916 (dependency slimming), but this is about import-time laziness rather than package requirements.

Kludex:
This is not what I wanted, and I'm also not happy to send a PR.

PR #7408, opened 2026-08-12
Resolve deferred `openai` resource imports and genai-prices data loading at `Model` construction time

### LINKED ISSUE #7405: First model request pays ~0.5s of lazy imports on the event loop (OpenAI resource tree, genai-prices snapshot)
### Description

The first `agent.run()` of a fresh process executes ~490 module imports inline on the event loop, inside the model-request path. Measured on Python 3.14 with `pydantic-ai-slim[openai]` 2.21.0 and 2.28.0 (same behavior on both):

- first run: **486 ms, 491 imports**
- second run: 6 ms, 0 imports

The breakdown (via `sys.addaudithook`):

| source | modules | cost |
| --- | --- | --- |
| `openai.resources` lazy tree (triggered by the first `client.chat.completions.create` attribute access) | ~483 | ~440 ms |
| `genai_prices` data snapshot (`get_snapshot()` on first usage/cost computation) | 2 | ~50 ms |
| `openai._base_client.get_platform` (`platform.mac_ver` pulls `plistlib`/`xml`) | 3 | ~20 ms (runs via `asyncify`, so off-loop) |

Module imports run whole on the event loop (no await points), so a latency-sensitive service pays this as a single stall on its first agent turn. Under a container CPU limit (`--cpus=0.5`) the same work stretches to 1-2 s. We found it via event-loop stall monitoring in a long-running worker: the largest stalls clustered right after pod restarts, on the first request of each fresh process.

### Repro

```python
import asyncio, collections, json, sys, threading, time
from http.server import BaseHTTPRequestHandler, HTTPServer

imported = []
sys.addaudithook(lambda e, a: imported.append(str(a[0])) if e == 'import' else None)

from pydantic_ai import Agent

class Mock(BaseHTTPRequestHandler):
    def do_POST(self):
        self.rfile.read(int(self.headers.get('Content-Length', 0)))
        body = json.dumps({'id': 'x', 'object': 'chat.completion', 'created': 1, 'model': 'm',
            'choices': [{'index': 0, 'message': {'role': 'assistant', 'content': 'hi'}, 'finish_reason': 'stop'}],
            'usage': {'prompt_tokens': 10, 'completion_tokens': 2, 'total_tokens': 12}}).encode()
        self.send_response(200)
        self.send_header('Content-Type', 'application/json')
        self.send_header('Content-Length', str(len(body)))
        self.end_headers()
        self.wfile.write(body)
    def log_message(self, *a): pass

srv = HTTPServer(('127.0.0.1', 0), Mock)
threading.Thread(target=srv.serve_forever, daemon=True).start()

model = OpenAIChatModel('m', provider=OpenAIProvider(api_key='x', base_url=f'http://127.0.0.1:{srv.server_port}/v1'))
agent = Agent(model, instructions='hi')

for label in ('first', 'second'):
    imported.clear()
    t0 = time.perf_counter()
    asyncio.run(agent.run('hello'))
    roots = collections.Counter(m.split('.')[0] for m in imported)
    print(f'{label} run: {(time.perf_counter() - t0) * 1000:.0f} ms, {len(imported)} imports, {roots.most_common(5)}')
```

Output on my machine:

```
first run: 486 ms, 491 imports, [('openai', 483), ('jiter', 3), ('xml', 2), ('genai_prices', 2), ('plistlib', 1)]
second run: 6 ms, 0 imports, []
```

### Workaround

Warming at startup removes it entirely (first turn drops to ~27 ms, 0 imports):

```python
import openai.resources  # noqa: F401
from genai_prices.data_snapshot import get_snapshot
get_snapshot()
```

### Suggestion

Options, roughly in order of preference:

1. `pydantic_ai.models.openai` eagerly imports `openai.resources` at module import time: anyone importing the model class has already opted into the OpenAI SDK, and it moves the cost to import time where it is expected.
2. Same for the `genai_prices` snapshot: call `get_snapshot()` at import time in `pydantic_ai.usage`, or wrap the first load in `anyio.to_thread`.
3. A documented `warm_up()` hook that services can call at boot.

Happy to send a PR for whichever direction you prefer.

```
Python 3.14, pydantic-ai-slim[openai] 2.21.0 and 2.28.0, openai 2.45.0-2.53.0
```

adtyavrdhn:
- Closes #7405

The first `agent.run()` of a fresh process paid ~0.5s of module imports inline on the event loop (1–2s under container CPU limits): the OpenAI SDK defers importing its ~480-module resource tree until the first attribute access on `client.chat.completions`, and genai-prices defers loading its pricing data until the first usage extraction. Both fired mid-request, with no `await` points, stalling every task in the process.

This moves both costs to model construction time, which normally runs at service startup:

- `OpenAIChatModel.__init__` / `OpenAIResponsesModel.__init__` touch the SDK's lazy `cached_property` resource chain (`client.chat.completions` / `client.responses`), so the imports resolve when the model is built.
- `Model.__init__` calls a new private `_cost.preload_pricing_data()` helper, so the genai-prices snapshot (cached process-wide) loads at construction for every provider, since `RequestUsage.extract` is shared.
- `docs/troubleshooting.md` gains a short "Slow First Run After Process Start" note: construct agents at startup, not in request handlers.

With the issue's repro script, the first run drops from **486 ms / 491 imports** to **43 ms / 3 imports** — the remaining three (`plistlib`/`xml`) come from `get_platform`, which the SDK already runs off-loop via `asyncify`.

The regression test constructs both models in a subprocess and asserts the deferred modules are present in `sys.modules` afterwards; a fresh interpreter is required because import state is process-global. Verified it fails with the fix reverted.

Notes for review:

- **Residual case**: constructing a model *inside* async code still pays the one-time cost at construction, on the loop. That's the user's placement choice and construction is where config errors are expected; documented via the troubleshooting note rather than adding thread offloading.
- **Scope**: only the two hot spots #7405 measured. Other provider SDKs (anthropic, google, bedrock) may have similar lazy trees — follow-up issue to audit them with the same `sys.addaudithook` harness.
- This is deliberately construction-time rather than module-import-time (the issue's suggestion 1) to stay consistent with #6713's push to keep `import pydantic_ai` light.

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7408?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -34,6 +34,12 @@ result = agent.run_sync('Who let the dogs out?')
 
 Synchronous methods like [`Agent.run_sync()`][pydantic_ai.agent.AbstractAgent.run_sync] reuse the thread's current event loop, and install a fresh one if other code closed it. If this error is raised from inside `httpx` or `httpcore` during a model request, the agent was already used before its event loop was closed: the provider's HTTP connection pool still holds connections bound to the dead loop. Recreate the agent together with its model and provider (or pass a fresh `http_client` to the provider); reusing an existing `Model` instance keeps the dead connection pool. Avoid closing an event loop that other code is still using.
 
+## Slow First Run After Process Start
+
+Some model SDK dependencies defer heavy imports and data loading until first use. Pydantic AI resolves these when a [`Model`][pydantic_ai.models.Model] is constructed, so the one-time cost is paid at startup rather than inline on the event loop during the first model request.
+
+To keep this cost out of request handling entirely, construct your agent and model once at application startup — at module level or in your framework's startup/lifespan hook — instead of inside a request handler.
+
 ## API Key Configuration
 
 ### [`UserError`][pydantic_ai.exceptions.UserError]: Set the `[PROVIDER]_API_KEY` environment variable or pass it via the provider's `api_key=...` argument
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -34,12 +34,6 @@ result = agent.run_sync('Who let the dogs out?')
 
 Synchronous methods like [`Agent.run_sync()`][pydantic_ai.agent.AbstractAgent.run_sync] reuse the thread's current event loop, and install a fresh one if other code closed it. If this error is raised from inside `httpx` or `httpcore` during a model request, the agent was already used before its event loop was closed: the provider's HTTP connection pool still holds connections bound to the dead loop. Recreate the agent together with its model and provider (or pass a fresh `http_client` to the provider); reusing an existing `Model` instance keeps the dead connection pool. Avoid closing an event loop that other code is still using.
 
-## Slow First Run After Process Start
-
-Pydantic AI loads deferred pricing data when a [`Model`][pydantic_ai.models.Model] is constructed, and deferred OpenAI SDK resource modules when an OpenAI model is constructed. This moves the one-time work out of the first model request, where it would otherwise block the event loop.
-
-To keep this cost out of request handling entirely, construct your agent and model once at application startup — at module level or in your framework's startup/lifespan hook — instead of inside a request handler.
-
 ## API Key Configuration
 
 ### [`UserError`][pydantic_ai.exceptions.UserError]: Set the `[PROVIDER]_API_KEY` environment variable or pass it via the provider's `api_key=...` argument
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -36,7 +36,7 @@ Synchronous methods like [`Agent.run_sync()`][pydantic_ai.agent.AbstractAgent.ru
 
 ## Slow First Run After Process Start
 
-Some model SDK dependencies defer heavy imports and data loading until first use. Pydantic AI resolves these when a [`Model`][pydantic_ai.models.Model] is constructed, so the one-time cost is paid at startup rather than inline on the event loop during the first model request.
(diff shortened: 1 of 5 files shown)


adtyavrdhn:
Fresh-process event-loop probe (macOS, Python 3.14.2, openai 2.45.0): I ran a 1 ms heartbeat concurrently with the first mocked agent request, five fresh processes per revision.

| | Base | This PR |
| --- | ---: | ---: |
| Imports during first run | 490 | 3 |
| First-run request time | 167–274 ms | 23–29 ms |
| Maximum heartbeat gap | 145–239 ms | 5.6–7.0 ms |
| Model construction time | 13–27 ms | 154–180 ms |

This confirms the expensive work moves to construction and removes the large event-loop stall rather than merely improving total elapsed time. The remaining three request-time imports were plistlib/xml from the OpenAI SDK platform lookup, which is already offloaded by the SDK. These timings are diagnostic evidence, not assertions added to the test suite.

Commit 0fe567d8d also names the workaround in a private interface-aware helper, documents why any provider client can trigger the process-wide module load (important for Bedrock Mantle subclasses), makes the troubleshooting wording specific to pricing/OpenAI, and stops the regression test from requiring Responses resources to remain unloaded after Chat construction.

PR #7437, opened 2026-08-13
Add `allowed_hosts` to `Agent.to_web()` and `clai web`

DouweM:
Adds an `allowed_hosts` argument to `create_web_app()` and `Agent.to_web()`, and a repeatable `--allowed-host` flag to `clai web`, controlling which `Host` headers the development chat UI answers to.

By default the app answers to IP addresses (`127.0.0.1`, `[::1]`, or a LAN address like `192.168.1.5`) and `localhost`, including names under it like `my-app.localhost`. Any other `Host` gets a `421 Misdirected Request`.

That covers how the UI is normally reached — `to_web()` under uvicorn, `clai web`, LAN access, container or SSH port forwarding — but a deployment that serves the UI under a real hostname, behind a reverse proxy or through a tunnel, now has to name it:

```python
app = agent.to_web(allowed_hosts=['ui.example.com'])
```

`*.example.com` matches subdomains, the same way that pattern works in Starlette's `TrustedHostMiddleware`. `['*']` accepts anything, for a deployment that already authenticates in front of the app.

### Implementation notes

- Raw ASGI middleware rather than `BaseHTTPMiddleware`, which buffers and would break the chat SSE stream. It wraps the whole app, so the `/api` mount, both index routes, and any route added later are covered.
- Not `TrustedHostMiddleware`: its `host.split(':')[0]` turns `[::1]:7932` into `[`, so IPv6 loopback can't be expressed at all.
- A single trailing dot is stripped from both the header and the configured patterns, because a browser navigating to `http://localhost./` keeps the DNS root dot and `localhost.` names the same host as `localhost`.
- Hostnames compare in ASCII, so an internationalized name goes in the list as punycode — which is what the browser sends.

### Tests

Cover the accepted and rejected forms (including bracketed IPv6 and label-boundary near-misses), wildcard semantics, normalization of configured entries, and WebSocket routes. Every mutation of the matcher I tried now fails at least one test.

Existing `TestClient(app)` call sites now pass an explicit `base_url`: `TestClient`'s default `Host: testserver` is not an accepted host.

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7437?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -190,6 +190,7 @@ clai web --agent my_module:my_agent -i 'Always respond in Spanish'
 | `--host` | Host to bind server (default: 127.0.0.1) |
 | `--port` | Port to bind server (default: 7932) |
 | `--html-source` | URL or file path for the chat UI HTML. |
+| `--allowed-host` | Hostname to answer to in addition to IP addresses and `localhost` (repeatable). See [Reaching the UI under a hostname](web.md#reaching-the-ui-under-a-hostname). |
 
 When using `--agent`, the agent's configured model becomes the default. CLI models (`-m`) are additional options. Without `--agent`, the first `-m` model is the default.
 
--- a/docs/web.md
+++ b/docs/web.md
@@ -101,11 +101,38 @@ Tools that [require approval](deferred-tools.md#human-in-the-loop-tool-approval)
 !!! warning
     The chat endpoint executes tool approvals relayed by the client, including for tools marked `requires_approval=True`. The server trusts the approval decision it receives, so any client that can reach the endpoint can approve any pending call.
 
-    Binding to localhost is not on its own a security boundary here: a web page open in the same browser can also reach `http://127.0.0.1:7932`. The chat endpoint therefore only accepts `Content-Type: application/json`, which a browser cannot send cross-origin without a preflight that the server refuses. Treat approval prompts as a convenience for the developer driving the UI rather than an authorization control, and don't expose `to_web()` to untrusted clients without putting authentication in front of it.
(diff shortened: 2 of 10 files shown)


DouweM:
Fixed in 482ec973d — the `['*.']` finding was real, and self-inflicted.

Reproduced first:

```python
app = create_web_app(Agent('test'), allowed_hosts=['*.'])
# GET /api/health  Host: evil.example  ->  200
```

The cause was a regression introduced earlier in this branch. Stripping a root dot from configured patterns (added so `allowed_hosts=['ui.example.com.']` matches) also rewrote `'*.'` into `'*'`, the documented sentinel that disables validation. So a plausible typo for `'*.example.com'` produced the one setting nobody wants, silently. Before that normalization existed `'*.'` was inert, which is why an earlier review pass dismissed it.

Patterns are now validated: a `*` prefix must be `*.` followed by a domain, so `'*.'`, `'*..'` and `'*example.com'` raise `UserError`. The last of those could only ever have matched a literal host containing `*`, so it was already a silent no-op.

Validation runs in `create_web_app()` rather than only in `HostValidationMiddleware.__init__`, because Starlette builds its middleware stack lazily — leaving it to the middleware would surface the error as a 500 on the first request instead of from the `to_web()` call:

```
>>> Agent('test').to_web(allowed_hosts=['*.'])
UserError: Invalid `allowed_hosts` pattern '*.'. Use a hostname, `*.example.com` to match its subdomains, or `*` to allow any host.
```

Mutation-tested: dropping the validation, and the narrower mutation that lets `'*.'` specifically through, both now fail tests.

Also in this push, from the CI Review doc note: the 421 body now mentions `.localhost` names, which it accepts but didn't name.

PR #7440, opened 2026-08-13
Declare the compaction trim's behavior on the model adapter

### LINKED ISSUE #7255: Lift the compaction trim + standing-prompt hoist out of the adapters into `prepare_messages`, keyed by profile flags
### Description

#7228 implements the compaction trim inside each adapter (Anthropic `_map_message`, OpenAI Responses `_build_responses_request_params` / `_map_messages` / count-tokens paths), which leaves near-duplicate juggling in two adapters: each decides where to trim, each carries the standing-prompt hoist, and OpenAI additionally interleaves it with `_resolve_server_side_state` ordering and (interim, until #7225) a lost-reveal redeclaration. Maintainer review verdict: this lives at the wrong level of abstraction.

Lift it into the shared pipeline, keyed by profile flags:

- **`prepare_messages` does the trim + standing-prompt hoist** for every adapter, gated on profile facts instead of adapter code, e.g. `compaction_mode: 'encrypted' | 'text' | None` (OpenAI Responses trims only when the `CompactionPart` carries `encrypted_content` — that's data, not adapter logic) plus the same-provider check it already has precedent for (`synthesize_local_tool_search_messages` is provider-name-keyed there today).
- **Why the standing-prompt hoist exists at all** (and must move with the trim): the provider's compaction blob covers *conversation content only* — on the wire, system/instructions are per-request channels sent fresh every turn, so providers never fold them into the compaction state. But our history representation stores the standing prompt *inside the first `ModelRequest`'s parts*, so a naive trim deletes it. The hoist re-prepends the opening `SystemPromptPart`s + latest `instructions` as a synthetic leading request. It's an impedance mismatch between our history model and the wire's (system, messages) split — pipeline-level by nature, not per-adapter.
- **Ordering constraint to design around**: OpenAI's `_resolve_server_side_state` recovers conversation/response IDs from responses the trim drops, so it must read the *untrimmed* history. Either resolve server-side state before `prepare_messages`, or pass the untrimmed view alongside.
- **Sequencing**: after #7225 merges, its boundary-aware reveal state makes #7228's interim `via_history` redeclaration branch inert — delete it in the same move, which is what makes the lift clean.

Out of scope: changing what the trim does (#7228 defines that, probe-backed); this is purely about where it lives.

Related: #7224, #7228, #7225, #6528.

DouweM:
Option (d) from the [discussion on #7255](https://github.com/pydantic/pydantic-ai/issues/7255#issuecomment-5260790699), as agreed: take the decision half of that issue and leave the trim itself adapter-side, because lifting it into `prepare_messages` breaks two things (evidence in that comment; the blocked attempt is preserved on `lift-compaction-trim-wip`).

Both adapters that honor `CompactionPart`s restated what their API's compaction carrier implies — OpenAI Responses hardcoded `requires_encrypted_content=True` at three call sites, Anthropic leaned on the helper's defaults — so the mapping from carrier to trim behavior was spread across the adapters rather than stated once.

`Model` now declares the two facts about its API's carrier, and `Model._trim_before_compaction` is the single place they become trim behavior:

```python
compaction_requires_encrypted_content: ClassVar[bool] = False
compaction_retains_standing_prompt: ClassVar[bool] = False
```

They're deliberately **two flags and not one enum**. Today's two adapters answer both the same way — OpenAI Responses `True`/`True`, Anthropic `False`/`False` — but that's a coincidence of the sample, not a rule: an API could send a plaintext summary that still stood in for the compacted window's system items, or an encrypted blob whose provider rebuilt the system channel per request. Neither field predicts the other, and `models/AGENTS.md` now says so for whoever adds a third adapter.

**Declared on the adapter, not the model profile**, which is the one deliberate deviation from the issue's wording. Which form an API uses is a property of the *API*, not the model behind it — the same model answers differently through OpenAI's Chat Completions and Responses APIs, which share one `OpenAIModelProfile`. This matches the `standing_prompt_retained` call in #7264 ("API-implementation behavior, not a model quality, so no profile key").

Concretely, eight providers other than `openai` route through `OpenAIResponsesModel` (`azure`, `deepseek`, `fireworks`, `nebius`, `openrouter`, `ovhcloud`, `sambanova`, `together`), each supplying a profile of its own:

| Provider | `openai_*` keys in the profile it returns |
|---|---|
| `OpenAIProvider` | 11 |
| `OpenRouterProvider` | 5 |
| `DeepSeekProvider` | 3 |
| `TogetherProvider` | **0** |

An `openai_compaction_*` profile key would be absent for exactly the models whose wire format is *least* in doubt — they're going through the Responses API, so the item is an encrypted blob regardless of which model is behind it. Absent means `False`, which is the dangerous direction for `requires_encrypted_content`: a `CompactionPart` carrying no blob would count as a wire boundary, so the history before it gets dropped with nothing sent in its place. Keeping the declaration on the adapter class is the same reason `supported_tool_deferral_modes` lives there.

The declaration also doesn't help a reader holding only a `CompactionPart`: it lives on the adapter class, while a part records a `provider_name`, and one adapter serves many of those (`OpenAIResponsesModel` over Azure reports `'azure'`), so there's no name → carrier lookup. Recording the carrier on the part itself is #7449, which depends on #7442 and is deliberately not taken here.

Behavior is unchanged and no existing test moved. Where the trim runs is untouched — OpenAI Responses still keeps its separate trimmed view so `_resolve_server_side_state` reads the untrimmed history. `AnthropicModel` restates both `False`s even though they're the inherited defaults, so its trim reads as a decision rather than an accident.

Left open on #7255, and deliberately out of scope here: whether the `'via_history'` lost-reveal redeclaration branch is genuinely inert post-#7225 — `test_openai_responses_pre_compaction_revealed_deferred_tool_is_redeclared_with_schema` still pins it at the adapter level, and deciding whether that test should instead assert the state is unreachable is a separate call.

`models/AGENTS.md` gains the adapter-authoring rule. Note that editing it makes the `Review` job fail by design.

- Closes #7255

### Checklist


DouweM:
Filed #7449 out of the review on #7442, and DouweM has said this PR can expand to take it if you want it.

**The gap:** `compaction_mode` declares the carrier on the adapter class, which is right for the adapter's own trim decision. But a reader holding only a `CompactionPart` can't get to it. The callability gate in #7442 has the `provider_name` recorded on the serving response, and one adapter serves many provider names — an Azure-backed `OpenAIResponsesModel` reports `'azure'` — so there's no name → mode lookup. The declaration and the recorded fact never meet.

The gate therefore infers the carrier from `'encrypted_content' in part.provider_details`, the same magic key `models/openai.py:3606-3611` branches on when rendering. That inference is wrong exactly when the payload has been lost in a round-trip through a client we don't control: the part still names an encrypted-carrier provider but carries only text, and the gate counts it as a boundary the adapter would have skipped. It errs toward refusing a call, which is why #7442 ships it as a documented inconsistency rather than a blocker.

**Two changes, if you take it:**

1. `CompactionPart.encrypted_content: str | None` as a real field. It's a standard field of this part type for one of the two carriers, which is what `provider_details` explicitly says it isn't for. Needs a migration path for histories already persisted with the key.
2. Record the carrier the producing adapter used, reusing the `CompactionMode` literal this PR introduces, so "would this provider have honored this part?" is answerable from the part alone.

With both, `_compaction_part_is_wire_boundary` reads the carrier instead of inferring it, the trim and the gate stop being able to disagree, and the `requires_encrypted_content` parameter #7442 threads through can go away — which also removes the `mode == 'encrypted'` translation `_trim_before_compaction` currently does.

Worth weighing against just landing this PR as-is: it's approved and green, #7449 is a follow-up either way, and the field migration is the kind of thing that's easier to review on its own. Your call — but if it does land here, update the description to close #7449.

DouweM:
Read #7449 — the analysis holds, including the load-bearing bit: `system` is `self._provider.name`, so one adapter really does report many provider names (`OpenAIResponsesModel` over Azure reports `'azure'`), and there is no name → mode lookup for a reader holding only a part. `compaction_mode` was never going to close that gap, and the PR description shouldn't imply otherwise.

**Not taking it here**, for the reason #7449's own "Order" section gives: it depends on #7442's `_compaction_part_is_wire_boundary` — the predicate the recorded carrier would simplify — and #7442 is itself stacked on #7271, so neither has landed. Taking it in this PR would mean either pre-empting that predicate or building the field against a shape that doesn't exist yet, and then re-reviewing it when #7442 lands anyway.

The scope test in CLAUDE.md also points the same way: a verified adjacent concern rides along only when it doesn't explode scope *or delay an already mergeable PR*. This one is approved and green, and #7449 adds a public field on `CompactionPart` plus a `provider_details` → field migration for already-persisted histories — exactly the kind of change that deserves its own review rather than an amendment to a behavior-neutral refactor.

So: land this as-is, and #7449 lands after #7271 → #7442, closed by whichever PR takes it. Nothing here blocks that — `CompactionMode` is already exported from `pydantic_ai.models` for it to reuse, and the `mode == 'encrypted'` translation in `_trim_before_compaction` is deliberately the only place that inference lives, so deleting it later is a one-line change once the carrier is recorded on the part.

One correction for whoever picks up #7449: it lists `messages.py:1990,2018` as documenting the magic key in prose. Those line numbers moved with the `STANDING_PROMPT_PLANTED_KEY` work in #7285 — the prose is now in the `CompactionPart` class docstring and the `content`/`provider_details` field docstrings. Worth grepping rather than trusting the line refs (per the repo's rule against line numbers in prose).

DouweM:
Agreed on not taking it here — the scope test cuts the way you read it, and #7449 wanting a public field plus a `provider_details` migration is not something to bolt onto a behavior-neutral refactor. Land this as-is.

Two updates:

**#7442 has merged**, so the dependency you cited is gone: `_compaction_part_is_wire_boundary` is on `main` now. The only remaining dependency is this PR, for the `mode == 'encrypted'` translation that #7449 deletes.

**DouweM has since scoped #7449 down further** (comment on the issue): we don't need to support a part that keeps its `provider_name` but loses its payload — in a round-trip you'd expect both to survive or both to go. So recording the carrier drops out entirely, `compaction_mode` never needs to reach a reader holding only a part, and the issue reduces to promoting `encrypted_content` to a real field. Worth knowing before you read the issue, since the version you reviewed argued the opposite.

On the line-reference correction: I checked, and `messages.py:1990` and `:2018` are accurate on current `main` — that prose didn't move with #7285. But your underlying point is right and is the repo's own rule, so I've replaced them in the issue with names rather than line numbers.

DouweM:
Merged `main` and fixed the CI break — both were real, and the second one was mine.

**CI was red on my own test file**, not on the refactor: `NameError: name 'OpenAIProvider' is not defined` on the lanes that don't install openai/anthropic. I built the models in a `parametrize` argument, which is evaluated at collection time, before `skipif` applies. Fixed in `1500d9269` by building the adapter inside the test, and the no-mode case now runs everywhere since it only needs `FunctionModel`. `main` is merged in; 568 tests pass locally across the OpenAI, Anthropic, compaction-mode and tool-availability suites.

**My line-reference correction was wrong** — thanks for checking rather than taking it. I verified against `origin/main`: `messages.py:1989` does carry "For OpenAI, `content` is `None` and the encrypted data is stored in `provider_details`", and `:2018` the `provider_details` docstring. I was reasoning from my branch's state instead of looking, which is exactly the failure mode the no-line-numbers rule exists to prevent. Replacing them with names in the issue was the right call anyway.

**On the reduced #7449:** dropping payload-loss support and therefore the recorded carrier makes sense to me — if `provider_name` survives a round-trip and `provider_details` doesn't, that's a lossy client, not a shape we should model. Two notes for whoever picks it up, from the #7287 work:

- The UI adapters *do* round-trip `provider_details` verbatim, including `encrypted_content` (`compaction_payload` in `ui/_adapter.py`), so the both-survive-or-both-go assumption holds for the protocol paths we control. Promoting `encrypted_content` to a real field means updating that payload and `compaction_part_from_payload` alongside the migration, or client-held compacted histories silently lose the blob.
- `sanitize_messages` strips `STANDING_PROMPT_PLANTED_KEY` out of `provider_details` and would keep doing so; a promoted `encrypted_content` field just stops being adjacent to it, which is fine, but worth a glance so the stamp-stripping test still covers what it means to.

With #7442 landed, the only thing left blocking #7449 is this PR, and it's green now.

pydantic_ai_slim/pydantic_ai/models/__init__.py
```diff
@@ -456,6 +478,39 @@ def tool_addition_mode(self) -> ToolAdditionMode | None:
...
+    ) -> list[ModelMessage]:
+        """Drop history before the latest compaction boundary this adapter's API honors.
+
+        The one place [`compaction_mode`][pydantic_ai.models.Model.compaction_mode] is turned into
+        trim behavior, so adapters call this instead of restating what their own mode implies. A
+        `None` mode returns the history unchanged. See `_trim_messages_before_compaction` for what
+        the trim preserves.
+
+        `standing_prompt_retained` defaults to what the mode implies: an `'encrypted'` carrier
+        retains the compacted window's leading system items, so re-inserting them would duplicate
```
DouweM: This doesn't sound like a valid "a + b" go together; it just _happens_ to be the case that OpenAI holds system parts _and_ uses encrypted content, while Ant does not.

DouweM: So sounds like we need 2 profile flags?

DouweM: Agreed, and done in d4b5ba90c — the two facts are now declared separately:

```python
compaction_requires_encrypted_content: ClassVar[bool] = False
compaction_retains_standing_prompt: ClassVar[bool] = False
```

The enum was asserting a rule from a sample of two. An API could send a plaintext summary that still stood in for the window's system items, or an encrypted blob whose provider rebuilt the system channel per request — neither field predicts the other, and the `AGENTS.md` note now says so explicitly for whoever adds a third adapter.

**On `Model` rather than the profile**, though — I want to flag why, since a profile key looks like the natural home. Eight providers other than `openai` route through `OpenAIResponsesModel` (`azure`, `deepseek`, `fireworks`, `nebius`, `openrouter`, `ovhcloud`, `sambanova`, `together`), each supplying its own profile. Checked what they actually produce:

| Provider | `openai_*` keys in profile |
|---|---|
| `OpenAIProvider` | 11 |
| `OpenRouterProvider` | 5 |
| `DeepSeekProvider` | 3 |
| `TogetherProvider` | **0** |

So an `openai_compaction_*` profile key would be absent for exactly the models whose wire format is *least* in doubt — they're going through the Responses API, so the item is an encrypted blob regardless of which model is behind it. And absent means `False`, which is the dangerous direction for `requires_encrypted_content`: a `CompactionPart` with no blob would count as a wire boundary, so the history before it gets dropped with nothing sent in its place. A profile key would have to be set defensively by every current and future Responses-compatible provider to stay correct, which is the failure mode the `supported_tool_deferral_modes` ClassVar already avoids for reveal renderers.

Happy to move them if you'd rather, but that's the case for keeping them next to `supported_tool_deferral_modes`.

Two other things in that commit:

- Dropped the "no declaration → no trim" third state I'd invented. It only existed to give the enum a `None`; only adapters that render compaction call `_trim_before_compaction`, exactly as before this PR. The synthetic `FunctionModel` test that pinned it went with it — the provider suites cover the real behavior.
- `AnthropicModel` restates both `False`s even though they're the inherited defaults, so its trim reads as a decision rather than an accident.

PR #7183, opened 2026-08-05
Document the `GROQ_BASE_URL` environment variable

ayaangazali:
`GroqProvider` reads `GROQ_BASE_URL` when `base_url` is not passed, which is documented in the provider docstring but not on the Groq page. The page's Environment variable section only covers `GROQ_API_KEY`, so someone pointing Groq at a proxy or gateway through the environment has no way to find it from the docs.

Added it to that section, following the same shape as the Ollama page, which documents `OLLAMA_BASE_URL` alongside `OLLAMA_API_KEY`.

The behavior is already covered by `test_groq_provider_with_env_base_url` in `tests/providers/test_groq.py`, which sets `GROQ_BASE_URL` and asserts the resulting `provider.base_url`, so this documents what that test already pins. Ran it (passes) plus `uv run mkdocs build --strict`, no new warnings.

### Checklist

- [ ] Any **AI generated code** has been reviewed line-by-line by the human PR author, who stands by it.
- [x] No **breaking changes** in accordance with the [version policy](https://github.com/pydantic/pydantic-ai/blob/main/docs/version-policy.md).
- [x] **PR title** is fit for the [release changelog](https://github.com/pydantic/pydantic-ai/releases).

---

quick note: built with Claude Code's help and read the diff myself. freshman contributor, happy to adjust the wording or drop the proxy example if you would rather keep it to one line.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7183?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #7488, opened 2026-08-15
Document provider environment variables that are only in docstrings

ayaangazali:
Four provider settings are read from the environment but documented only in a docstring, so a user configuring through the environment cannot find them from the docs:

| Page | Variable | What reads it |
| --- | --- | --- |
| `models/groq.md` | `GROQ_BASE_URL` | `GroqProvider` falls back to it when `base_url` is omitted. The page's Environment variable section covers only `GROQ_API_KEY`. |
| `models/cohere.md` | `CO_BASE_URL` | `CohereProvider` passes it to both the v1 and v2 clients. It takes no `base_url` argument at all, so the environment is the only way to point it elsewhere. |
| `gateway.md` | `PYDANTIC_AI_GATEWAY_BASE_URL` | `_infer_base_url` raises a `UserError` telling the user to "set the `PYDANTIC_AI_GATEWAY_BASE_URL` environment variable explicitly" when the key encodes no region. The variable appeared nowhere in the Gateway docs, so that error pointed at something unsearchable. |
| `models/bedrock.md` | `AWS_READ_TIMEOUT`, `AWS_CONNECT_TIMEOUT` | `BedrockProvider` falls back to these, then to 300 and 60 seconds. Long generations are where people hit the read timeout and go looking. |

Each is verified against main: the variable is still read by its provider, and still absent from the page. The Bedrock section also notes that a prebuilt `bedrock_client` is used as-is, and links to the Timeouts page added in #7145.

This replaces #7183, #7184, #7223 and #7227, which carried these one at a time. Their descriptions cited `mkdocs build --strict` output, which stopped meaning anything when #7416 removed the MkDocs pipeline, so the evidence here is stated against the files instead. Ran `uv run pytest tests/test_examples.py -k "bedrock or gateway"`, 26 passed.

### Checklist

---

quick note: built with Claude Code's help and read the diff myself. happy to drop any individual page from this if one of these is meant to stay undocumented.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7488?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/gateway.md
+++ b/docs/gateway.md
@@ -92,6 +92,14 @@ Set the `PYDANTIC_AI_GATEWAY_API_KEY` environment variable to your Gateway API k
 export PYDANTIC_AI_GATEWAY_API_KEY="pylf_v..."
 ```
 
+The Gateway base URL is inferred from the region encoded in your API key, so you normally do not
+need to set it. If your key predates region-encoded keys, Pydantic AI raises a `UserError` asking
+you to generate a new key or set the URL yourself, which you do with `PYDANTIC_AI_GATEWAY_BASE_URL`:
+
+```bash
+export PYDANTIC_AI_GATEWAY_BASE_URL="https://gateway-us.pydantic.dev/proxy"
+```
+
 You can access multiple models with the same API key, as shown in the code snippet below.
 
 ```python {title="hello_world.py"}
--- a/docs/models/bedrock.md
+++ b/docs/models/bedrock.md
@@ -62,6 +62,34 @@ agent = Agent(model)
 ...
 ```
 
+### Timeouts
+
+Long generations can outlast the Bedrock client's default timeouts. Both are configurable on
+[`BedrockProvider`][pydantic_ai.providers.bedrock.BedrockProvider], and each falls back to an
+environment variable when omitted:
+
+| Argument | Environment variable | Default |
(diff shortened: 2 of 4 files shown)


ayaangazali:
Re-checked this against current main (4fda38988) since it has been open a week.

All five variables are still undocumented and still read by the source: `GROQ_BASE_URL`, `CO_BASE_URL`, `AWS_READ_TIMEOUT`, `AWS_CONNECT_TIMEOUT`, `PYDANTIC_AI_GATEWAY_BASE_URL`. The Bedrock section still matches the code too, `aws_read_timeout` and `aws_connect_timeout` are still constructor arguments on `BedrockProvider`, the model id in the example is still a known name, and the `../timeouts.md` link still resolves.

No rush from my side. If the bundle is the problem and you would rather review these as separate per-provider PRs, or drop the Bedrock timeouts part and keep just the base URLs, tell me which and I will split it.

PR #7495, opened 2026-08-16
Fix diff rendering pointing the wrong way for negative baselines

### LINKED ISSUE #7496: pydantic-evals: diff rendering points the wrong way for negative baselines
### Initial Checks

### Description

In `pydantic_evals`, the relative half of a rendered diff cell points the wrong way whenever the baseline is negative, because `_render_relative` in `pydantic_evals/reporting/render_numbers.py` divides the delta by the **signed** base:

```python
rel_change = (delta / base) * 100
```

The result is that the two halves of a single cell disagree, and in the multiplier branch a regression is rendered as a positive-looking multiple:

| old | new | true direction | rendered today |
|---|---|---|---|
| -2.0 | -1.0 | improved | `+1.0 / -50.0%` |
| -1.0 | -3.0 | regressed | `-2.0 / 3.0x` |
| -1.0 | +3.0 | improved | `+4.0 / -3.0x` |

Row 2 is the one that worries me: `new / base` is `-3 / -1 = 3`, so a score that got twice as bad renders as `3.0x` with no sign at all.

This feeds `EvaluationRenderer` through `diff_formatter = default_render_number_diff`, so it affects the score and metric columns of a report diff, which is the view someone reads to decide whether a run regressed. Negative scores and metrics are legitimate and not rare: reward margins, signed deltas against a baseline, log-scores, and any custom evaluator returning a signed quantity.

Expected: the sign of the relative change matches the direction of the change, so the absolute and relative halves of a cell never contradict each other.

```python
from pydantic_evals.reporting.render_numbers import default_render_number_diff

# An evaluator score improving from -2.0 to -1.0
print(default_render_number_diff(-2.0, -1.0))  # '+1.0 / -50.0%'  <- halves disagree

# The same score regressing from -1.0 to -3.0
print(default_render_number_diff(-1.0, -3.0))  # '-2.0 / 3.0x'    <- regression reads as 3x
```

`pydantic-evals` on `main` (reproduced at commit ebd2054). No LLM client involved; the function is pure formatting.

arthi-arumugam-git:
Fixes #7496

## The wrong number

`default_render_number_diff` renders the relative half of a diff cell with an inverted sign whenever the baseline is negative, because `_render_relative` divides the delta by the signed base:

| old | new | true direction | rendered on main |
|---|---|---|---|
| -2.0 | -1.0 | improved | `+1.0 / -50.0%` |
| -1.0 | -3.0 | regressed | `-2.0 / 3.0x` |
| -1.0 | +3.0 | improved | `+4.0 / -3.0x` |

In the first row the absolute and relative halves of one cell point in opposite directions. In the second, a regression renders as a threefold multiplier with no sign at all, since `new / base` is `-3 / -1 = 3`. These cells feed the report diff via `diff_formatter = default_render_number_diff`, which is exactly the view someone reads to decide whether a run regressed, and scores and metrics can legitimately be negative (reward margins, deltas, signed log-scores).

## The fix

The relative change is now computed against `abs(base)`, so its sign always matches the direction of the change. For negative bases the multiplier style is not used, because `new / base` is arithmetically true but reads backwards there; percentages above 100% are used instead:

| old | new | rendered with this change |
|---|---|---|
| -2.0 | -1.0 | `+1.0 / +50.0%` |
| -1.0 | -3.0 | `-2.0 / -200.0%` |
| -1.0 | +3.0 | `+4.0 / +400.0%` |

Behaviour for positive baselines is unchanged: every existing snapshot in `tests/evals/test_render_numbers.py` still passes, including the sign-crossing cases with positive bases such as `(2.0, -25.0) -> '-27.0 / -12.5x'`.

## Tests

Five parametrized negative-baseline cases are added to `test_default_render_number_diff`, plus `test_negative_base_diff_sign_matches_direction`, which asserts the absolute and relative halves of a rendered cell never disagree in sign. The added cases fail on `main` and pass with this change; the full `test_render_numbers.py` module passes (66 tests).

## What this does not change

Zero bases (relative indicator already dropped), the small-base drop heuristic, integer diffs, and duration diffs are untouched.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7495?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_evals/pydantic_evals/reporting/render_numbers.py
+++ b/pydantic_evals/pydantic_evals/reporting/render_numbers.py
@@ -140,15 +140,20 @@ def _render_relative(new: float, base: float, small_base_threshold: float) -> st
     if abs(base) < small_base_threshold and abs(delta) > MULTIPLIER_DROP_FACTOR * abs(base):
         return None
 
-    # Compute the relative change as a percentage.
-    rel_change = (delta / base) * 100
+    # Compute the relative change as a percentage of the base's magnitude. Dividing by the
+    # signed base would invert the sign for negative bases, so an improvement from -2.0 to
+    # -1.0 would render as '-50.0%' next to an absolute diff of '+1.0'.
+    rel_change = (delta / abs(base)) * 100
     perc_str = f'{rel_change:+.{PERC_DECIMALS}f}%'
     # If the percentage rounds to 0.0%, return only the absolute difference.
     if perc_str in ('+0.0%', '-0.0%'):
         return None
 
-    # Decide whether to use percentage style or multiplier style.
-    if abs(delta) / abs(base) <= 1:
+    # Decide whether to use percentage style or multiplier style. A multiplier is only used
+    # for positive bases: for a negative base, `new / base` is arithmetically true but reads
+    # backwards (-1.0 -> -3.0 would render as '3.0x'), so percentages above 100% are used
+    # there instead.
+    if abs(delta) / abs(base) <= 1 or base < 0:
         # Percentage style.
         return perc_str
     else:
--- a/tests/evals/test_render_numbers.py
+++ b/tests/evals/test_render_numbers.py
@@ -56,10 +56,34 @@ def test_default_render_number(value: float | int, expected: str):
         (0.02, -25.0, snapshot('-25.0 / -1,250x')),
         (0.001, 25.0, snapshot('+25.0')),
         (0.0, 25.0, snapshot('+25.0')),
+        # Negative baselines: the relative indicator's sign must match the direction of the
+        # change, and multipliers (which read backwards for negative bases) are not used.
+        (-2.0, -1.0, snapshot('+1.0 / +50.0%')),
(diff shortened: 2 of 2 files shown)


arthi-arumugam-git:
Apologies, I missed the linked-issue requirement. I have opened #7496 with a reproducible example, and this PR's description now starts with `Fixes #7496`.

Per the bot's instructions I have not reopened this myself, since it asks for a maintainer to confirm the issue first. Happy to reopen the moment someone acknowledges #7496, or to close this and open a fresh PR if that is tidier on your side.

For what it is worth while the issue waits: the fix is confined to `_render_relative`, every existing snapshot in `test_render_numbers.py` still passes unchanged, and the added cases fail on main.

PR #7497, opened 2026-08-16
Fix diff rendering pointing the wrong way for negative baselines

### LINKED ISSUE #7496: pydantic-evals: diff rendering points the wrong way for negative baselines

arthi-arumugam-git:
Fixes #7496

`_render_relative` divides by the signed base, so when the base is negative the sign of the
relative change is inverted with respect to the absolute change. The two halves of the same
diff cell then point in opposite directions:

- `-2.0 -> -1.0` renders as `+1.0 / -50.0%`. The value improved, and the percentage says it got worse.
- `-1.0 -> -3.0` renders as `-2.0 / 3.0x`. The value got worse, and the multiplier reads as growth.

The multiplier branch has the same root cause seen from a different angle: `new / base` is
arithmetically true for a negative base but cannot be read the way a multiplier is meant to be
read.

## Change

- Compute the relative change against `abs(base)`, so the percentage sign tracks the direction
  of the change rather than the sign of the baseline.
- Use percentage style rather than multiplier style when the base is negative.

## Tests

Added parametrised cases covering negative baselines, plus an assertion that the absolute and
relative halves of a diff never disagree in sign. Both describe behaviour that `main` does not
satisfy today.

## Scope

Only `_render_relative` changes. Positive-base behaviour, the zero-base guard, the small-base
multiplier drop and the `+0.0%` suppression are all untouched.

This replaces #7495, which was closed automatically for not referencing an issue.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7497?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_evals/pydantic_evals/reporting/render_numbers.py
+++ b/pydantic_evals/pydantic_evals/reporting/render_numbers.py
@@ -68,12 +68,15 @@ def default_render_number_diff(old: float | int, new: float | int) -> str | None
             _default_format_number_diff(3, 4) -> '+1'
       - For floats (or a mix of float and int):
           * Compute the raw delta = new - old and format it with ABS_SIG_FIGS significant figures.
-          * If `old` is nonzero, compute a relative change:
+          * If `old` is nonzero, compute a relative change against |old|, so that its sign
+            tracks the direction of the change rather than the sign of the baseline:
               - If |delta|/|old| ≤ 1, render the relative change as a percentage with
                 PERC_DECIMALS decimal places, e.g. '+0.7 / +70.0%'.
-              - If |delta|/|old| > 1, render a multiplier (new/old). Use one decimal place
-                if the absolute multiplier is less than MULTIPLIER_ONE_DECIMAL_THRESHOLD,
-                otherwise no decimals.
+              - If |delta|/|old| > 1 and `old` is positive, render a multiplier (new/old).
+                Use one decimal place if the absolute multiplier is less than
+                MULTIPLIER_ONE_DECIMAL_THRESHOLD, otherwise no decimals.
+              - If `old` is negative, keep the percentage form: a multiplier over a negative
+                base reads backwards, e.g. -1.0 -> -3.0 would show as '3.0x'.
           * However, if the percentage rounds to 0.0% (e.g. '+0.0%'), return only the absolute diff.
           * Also, if |old| is below BASE_THRESHOLD and |delta| exceeds MULTIPLIER_DROP_FACTOR×|old|,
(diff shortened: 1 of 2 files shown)


arthi-arumugam-git:
Checking in on this one, mostly to make sure it is waiting on a reviewer and not on something from me.

It is small: when the baseline value is negative the diff arrow points the wrong way, so a result that got worse renders as an improvement. Two files, and there is a test covering the negative-baseline case.

Checks are green and it still merges cleanly. Happy to rebase if that helps.

PR #7499, opened 2026-08-16
fix: preserve all conditions in pruned span queries

### LINKED ISSUE #7484: pydantic_evals: SpanQuery with stop_recursing_when silently mis-evaluates every descendant/ancestor condition after the first
### Initial Checks

### Description

In `pydantic_evals`, a `SpanQuery` that carries `stop_recursing_when` evaluates only the **first** of its descendant (or ancestor) conditions correctly. Every later condition in the same query is evaluated against an empty sequence, so:

- `no_descendant_has` / `no_ancestor_has` pass **vacuously** — an error span goes undetected
- `all_*_have` passes vacuously
- `some_*_has` fails vacuously

The false-negative case is the dangerous one: an eval asserting "no descendant span has status error" reports success on a trace that does contain one, as long as some other descendant condition was listed first.

**Cause.** In `SpanNode._matches_query`, the nested helpers `pruned_descendants()` / `pruned_ancestors()` are decorated with `@cache`, and when `stop_recursing_when` is present they return `self._filter_descendants(...)` / `self._filter_ancestors(...)`, which are **generators** (`-> Iterator[SpanNode]`), not lists. `@cache` memoizes the single generator object, and each helper is consumed by three separate conditions — so the second and third consumers iterate an already-exhausted iterator.

Without `stop_recursing_when` the same helpers return `descendants()` / `ancestors()`, which are real lists, so the bug only appears when pruning is requested. The existing tests each use exactly one descendant/ancestor condition per query, which is why it hasn't shown up.

Fix is to materialise both, i.e. wrap the two `self._filter_*(...)` calls in `list(...)`. I have that change plus a regression test ready and would be glad to open a PR if you'd like to assign this to me — flagging first since the PR template notes unassigned non-trivial PRs may be auto-closed.

```python

from pydantic_evals.otel.span_tree import SpanNode

TS = datetime(2026, 1, 1, tzinfo=timezone.utc)


def node(name: str, span_id: int, parent_span_id: int | None) -> SpanNode:
    return SpanNode(
        name=name,
        trace_id=1,
        span_id=span_id,
        parent_span_id=parent_span_id,
        start_timestamp=TS,
        end_timestamp=TS,
        attributes={},
    )


root = node('root', 1, None)
level1 = node('level1', 2, 1)
leaf = node('leaf', 3, 2)
root.add_child(level1)
level1.add_child(leaf)

plain = {
    'all_ancestors_have': {'name_matches_regex': '.*'},
    'no_ancestor_has': {'name_equals': 'root'},
}
# identical query, plus a prune condition that matches nothing at all
with_stop = {**plain, 'stop_recursing_when': {'name_equals': 'this-never-matches'}}

print([a.name for a in leaf.ancestors])  # ['level1', 'root']
print(leaf.matches(plain))               # False  (correct: 'root' IS an ancestor)
print(leaf.matches(with_stop))           # True   (wrong: same query, inert prune)
```

Output:

```
['level1', 'root']
False
True
```

Both must be `False`. `root` is an ancestor of `leaf`, so `no_ancestor_has: {'name_equals': 'root'}` cannot be satisfied — but with an inert `stop_recursing_when` added, the query reports a match.

The same happens on the descendant side (`some_descendant_has` / `all_descendants_have` / `no_descendant_has`).

### Logfire Trace

N/A — reproduces from a hand-built `SpanNode` tree with no tracing involved.

```
pydantic-ai 2.31.0
pydantic-evals 2.31.0
pydantic-ai-slim 2.31.0
repo commit 25a7092
Python 3.13
```

mikemikimike:
- Closes #7484

Materialize pruned ancestor and descendant traversal iterators so multiple `SpanQuery` conditions evaluate the same snapshot. Adds regression coverage for descendant and ancestor conditions with inert and effective `stop_recursing_when` pruning.

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7499?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


mpfaffenberger:
Hey there! Thank you for this P/R, and sorry about the runaround with the earlier auto-closes. This is the right fix, the materialization matches what the other `_filter_descendants` / `_filter_ancestors` callers already do, and CI is green.

One ask before we merge: could you broaden the regression test a little? Right now it covers a single ancestor-side case with an inert prune. The descendant side has the same bug, and there's a sneakier class of cases too: `any()` short-circuits, so a later condition can accidentally see the right node in the leftover generator and pass for the wrong reason. To really pin the bug down, the later condition needs to reference a node the earlier condition already consumed past.

I opened #7858 before spotting yours (my mistake, it's closed now in favor of this one), and the test there covers both sides plus an effective-prune case and the `and_` sub-query workaround. Feel free to lift it wholesale.

Last thing, the P/R description picked up some literal `` `n `` characters (looks like PowerShell escaping). Mind cleaning that up and filling in the template checklist?

PR #7531, opened 2026-08-17
Don't honor serialized Temporal HITL approval as ambient run-context state

### LINKED ISSUE #5536: HITL approval signal is a plain bool — hardening proposal (narrow vs proper fix)
## Problem

`ApprovalRequiredToolset` (`pydantic_ai_slim/pydantic_ai/toolsets/approval_required.py`) gates a tool call on `ctx.tool_call_approved`, a plain `bool` declared at `_run_context.py:76`. There is no nonce, no signature, no binding to `tool_call_id`, no TTL.

That is fine in-process where the deployer controls who flips the flag. It is less fine on the Temporal durable-exec wire: `durable_exec/temporal/_run_context.py` `serialize_run_context` ships `tool_call_approved` across activity boundaries as a plain JSON value, and `deserialize_run_context` is `return cls(**ctx, deps=deps)` with no integrity check. Anyone who can tamper with the Temporal activity event history can flip the bool from `False` to `True` on the next activity.

This requires privileged access to the Temporal cluster, so it is **not a CVE** — but it is a defense-in-depth weakness on a Tier-1 feature (HITL on irreversible actions,OWASP LLM06 Excessive Agency).

## Demo — wire-replay reproduction (verified on v1.99.0)

```python
from pydantic_ai._run_context import RunContext
from pydantic_ai.durable_exec.temporal._run_context import TemporalRunContext
from pydantic_ai.usage import RunUsage

# Server-side: approval has NOT been granted yet.
ctx = RunContext(
    deps=None,
    model=TestModel(),
    usage=RunUsage(),
    tool_call_id='call-1',
    tool_name='charge_card',
    tool_call_approved=False,
)

# Serialize for the Temporal activity wire — plain dict, no signing.
wire = TemporalRunContext.serialize_run_context(ctx)
print(f'on the wire: tool_call_approved={wire["tool_call_approved"]!r}')

# Attacker tampers with the activity input/output dict between activities.
# Requires privileged write access to the Temporal cluster event history.
wire['tool_call_approved'] = True

# Receiving worker re-hydrates with no signature check.
resumed = TemporalRunContext.deserialize_run_context(wire, deps=None)
print(f'after replay: tool_call_approved={resumed.tool_call_approved!r}')

# Output:
#   on the wire: tool_call_approved=False
#   after replay: tool_call_approved=True
```

## Two approaches

**Narrow fix (~3 LOC + docs).** Drop `tool_call_approved` from the default `TemporalRunContext.serialize_run_context` payload; document that approval state does not survive an activity boundary by default. `tool_manager.py` already re-derives the flag per-call from `DeferredToolResults`, so `ApprovalRequiredToolset.call_tool` continues to work inside activities. Only deployers who read `ctx.tool_call_approved` directly in their tool bodies see a change.

**Proper fix (~150–300 LOC, breaking).** Replace `tool_call_approved: bool` with an HMAC-bound `tool_call_approval_token: str | None` containing `(run_id, tool_call_id, expiry)`. `ApprovalRequiredToolset` issues the token on raise; `tool_manager.py` verifies on resume. Best done in a major version bump.

## What I'd like

Maintainer input on which direction (or neither — close as wontfix if the threat model is out of scope). I can champion either approach and send the PR once aligned.

### References

_No response_

AAH20:
Fixes #5536

This is the shippable Temporal slice of that issue (not HMAC / store-as-authority). Please assign if you want it in the review queue.

### What this PR does

Temporal tool activities currently serialize `tool_call_approved` on the generic run-context blob. Anyone who can flip that bit on the activity payload (or, more commonly, anyone whose approved activity times out after the side effect has already committed) gets a second execution of the same tool. `ApprovalRequiredToolset` is a stateless gate: once the bool is true, `call_tool` runs straight through, and Temporal's default retry policy is at-least-once.

This PR keeps HITL working and removes the ambient bool from the trust path:

1. `TemporalRunContext.serialize_run_context` no longer carries `tool_call_approved` / `tool_call_metadata`. Deserialize ignores those keys if they show up (a flipped blob cannot elevate).
2. Tool-call activities apply the gate from `CallToolParams` (computed workflow-side from `DeferredToolResults`). `None` on that field is the pre-change payload so in-flight HITL resumes still complete.
3. `RunContext.idempotency_key` is a replay-stable `run_id:tool_call_id` string for tool authors to pass to Stripe / provisioners / any idempotent API. It is not an approval token.

HMAC, signing, and store-as-authority stay out of scope (DouweM's threat-model question on Temporal-cluster compromise). The retry double-execution case does not need an attacker.

### Tests

- Mutating `serialized_run_context['tool_call_approved']=True` while `CallToolParams.tool_call_approved=False` stays unapproved (through the Pydantic data converter).
- Omitting the new `CallToolParams` fields still honors the legacy blob keys.
- The same serialized payload deserialized twice yields the same `idempotency_key`; changing `tool_call_id` on the wire changes the key.
- Exhaustive serialization categorization updated.

Ran those assertions against the Temporal Pydantic converter locally. `tests/test_temporal.py` skips the whole module on Python 3.14 because of the Temporal sandbox issue; CI on 3.10–3.13 is the real Temporal worker suite.

### Checklist

- [x] No **breaking changes** in accordance with the [version policy](https://github.com/pydantic/pydantic-ai/blob/main/docs/version-policy.md). Optional `CallToolParams` fields with defaults; in-flight HITL continues via the `None` path.

If this slice is useful in production Gate/Prove reviews (SOC 2 / insurance for agent tool policy), the commercial evaluation is the [$499 Instant Audit](https://a2zsoc.com/productized-services#instant-audit-tripwire) and [consultation](https://a2zsoc.com/consultation). The code here has no vendor dependency.


AAH20:
Updated this PR’s description to `Fixes #5536` as the issue-gate bot asked.

`gh pr reopen` and `PATCH .../pulls/7531` with `state=open` both fail from this account (GraphQL reopenPullRequest error, REST 422). A maintainer reopen from the GitHub UI should be enough if the bot will accept the closing keyword.

PR #7559, opened 2026-08-18
Add `openai_cache_instructions` to place a prompt cache breakpoint on the instructions

### LINKED ISSUE #7128: Add `openai_cache_instructions`: place `prompt_cache_breakpoint` on the developer/system message for OpenAI GPT-5.6
## Summary

For GPT-5.6 models, pydantic-ai can place an explicit `prompt_cache_breakpoint` on **user content** via `CachePoint` (added in #6423), but there is no way to put one on the **instructions** message. Instructions are always serialized to the Responses API top-level `instructions` field (or a `developer`/`system` message in Chat Completions) with no breakpoint. Meanwhile Anthropic (`anthropic_cache_instructions`), Bedrock (`bedrock_cache_instructions`), and OpenRouter (`openrouter_cache_instructions`) all have a dedicated setting for caching instructions. OpenAI is the missing counterpart.

The raw OpenAI/Bedrock Mantle Responses API does support this exact request shape — a `developer`-role message inside `input` carrying `prompt_cache_breakpoint: {"mode": "explicit"}` (see AWS blog in references). The gap is that pydantic-ai's `Agent` cannot produce it.

## Requested API

```python
from pydantic_ai import Agent
from pydantic_ai.models.openai import OpenAIResponsesModelSettings

agent = Agent(
    'openai:gpt-5.6-luna',
    instructions=long_static_instructions,  # > 1024 tokens
    model_settings=OpenAIResponsesModelSettings(
        openai_prompt_cache_key='support-app:kb-v1',
        openai_cache_instructions=True,  # <-- requested setting
        openai_prompt_cache_options={'mode': 'explicit', 'ttl': '30m'},
    ),
)
```
Expected wire shape (Responses API), mirroring anthropic_cache_instructions and the AWS blog:

```python
input=[
    {"type": "message", "role": "developer",
     "content": [{"type": "input_text", "text": instructions,  "prompt_cache_breakpoint": {"mode": "explicit"}}]},
    {"type": "message", "role": "user", "content": [{"type": "input_text", "text": question}]},
]
```

## Use case
Agentic tool loops (exactly my agent setup here: ~2,300-token static system prompt + 10 toolsets) resend the same instructions and tools every turn. On GPT-5.6, cache writes cost 1.25x input and reads are discounted 90%, so a single explicit breakpoint after the static instructions gives deterministic write-once/read-many caching for the largest repeated prefix, which CachePoint in user content cannot achieve because the instructions are the longest stable prefix and precede all user content.

## Implementation sketch
When openai_cache_instructions is set and the profile supports prompt-cache breakpoints (openai_supports_prompt_cache_breakpoints):

- Responses API: emit instructions as the first input message (role='developer') with prompt_cache_breakpoint instead of the top-level instructions field.
- Chat Completions: emit them as the leading developer/system message with prompt_cache_breakpoint on the content block.
- Reuse the existing _add_openai_prompt_cache_breakpoint helper. Anchor at index 0 so the prefix position never shifts with history length.

### References

- [AWS blog where ystem prompt is cached:](https://aws.amazon.com/blogs/machine-learning/introducing-explicit-prompt-caching-for-openai-gpt-5-6-models-on-amazon-bedrock/) : Introducing explicit prompt caching for OpenAI GPT-5.6 models on Amazon Bedrock in **system prompt**
- [OpenAI prompt caching guide](https://developers.openai.com/api/docs/guides/prompt-caching#caching-behavior-changes-when-migrating-to-gpt-56): (prompt_cache_breakpoint, prompt_cache_key, prompt_cache_options)
- pydantic/pydantic-ai#6423 (merged) : explicit prompt caching for OpenAI user content (CachePoint) this builds on
- pydantic/pydantic-ai#6419 (closed by #6423) — original "Support explicit prompt caching for OpenAI GPT-5.6" request
- pydantic/pydantic-ai#4865 (merged) : smart instruction caching for Anthropic/Bedrock; the exact analog
- pydantic/pydantic-ai#4543 (closed by #4865) — issue behind #4865
- pydantic/pydantic-ai#4604 (merged) :  OpenRouter cache settings family + CachePoint (precedent)
- pydantic/pydantic-ai#4392 (closed by #4604) — OpenRouter CachePoint request
- pydantic/pydantic-ai#3678 (merged) : openai_prompt_cache_key / openai_prompt_cache_retention settings (precedent)

adtyavrdhn:
- Closes #7128

OpenAI will not put a cache breakpoint on the top-level `instructions` field, so this setting moves them into `input`. `input` is replayed by `previous_response_id`. If we then omit the current agent's instructions (to avoid sending them twice), a later agent that continues the transcript keeps running the previous agent's policy.

That is the opposite of what [`instructions`](https://ai.pydantic.dev/agents/#instructions) are for: only the *current* agent’s instructions go to the model. `system_prompt` is the thing that is meant to stick around in history.

### Layer by layer

**1. What Anthropic does.** `anthropic_cache_instructions` still sends the instructions every request. It only marks them cacheable. There is no server-side replay, so there is no stale-policy problem.

**2. What OpenAI actually stores.**

```python
# Request 1 (this PR, caching on)
instructions: omitted
input: [
  {role: "system", content: "Refunds under $50 are automatic.", breakpoint},
  {role: "user", content: "Can I get a refund?"},
]
# → stored as resp_abc
```

OpenAI’s own contract: the top-level `instructions` field does **not** carry over with `previous_response_id`. `input` **does**.

**3. Why a boolean stamp is not enough.** After request 1, `resp_abc`’s stored `input` contains the refunds policy. Request 2 with `'auto'` used to see `instructions_relocated=True` and omit whatever the *current* agent wanted to say:

```python
refunds = Agent(model, instructions='Refunds under $50 are automatic.', model_settings=settings)
first = await refunds.run('Can I get a refund?')

collections = Agent(model, instructions='Never offer a refund.', model_settings=settings)
await collections.run('What next?', message_history=first.all_messages())
```

Before: request 2 sent `previous_response_id=resp_abc` and no instructions. OpenAI replayed the refunds policy. Collections never spoke.

A stamp only answers “did we relocate *something*?”. It does not answer “is that something this agent’s instructions?”

**4. What we do now.** Compare the current joined instruction text to `ModelRequest.instructions` on the request that produced the chained response (same join the framework already stores, including prompted-output text).

- Match (same agent, same text, tool loop / next turn): keep the chain, omit, reuse the cache prefix.
- Differ (other agent, or you changed the text): drop `previous_response_id`, relocate *this* request’s instructions onto full local history.

```python
# Request 2 after the collections handoff
previous_response_id: omitted
input: [
  {role: "system", content: "Never offer a refund.", breakpoint},
  {role: "user", content: "Can I get a refund?"},
  {role: "assistant", content: "..."},
  {role: "user", content: "What next?"},
]
```

Chat Completions is unchanged: instructions are still sent every time, we only add a breakpoint.

An explicit `openai_previous_response_id` that is not in `message_history` still raises (nothing to compare). If that response *is* in history, it uses the same match/differ rule as `'auto'`.

### New public surface

`OpenAIChatModelSettings.openai_cache_instructions` / `OpenAIResponsesModelSettings.openai_cache_instructions` (`bool`, opt-in, ignored unless `openai_supports_prompt_cache_breakpoints`).

### Verification

- `test_openai_responses_cache_instructions_mismatch_drops_chain` — the collections/refunds hole
- `test_openai_responses_cache_instructions_dynamic_mismatch_drops_chain` — date changed, do not replay yesterday
- `test_openai_responses_cache_instructions_reused_across_chained_requests` — same agent still chains
- `test_openai_responses_cache_instructions_explicit_previous_response_id_raises`
- `test_openai_responses_cache_instructions_explicit_previous_response_id_reuses_when_in_history`

Nothing unless they opt into `openai_cache_instructions=True`.

### Checklist


sipa-echo-ngbm:
Hi @adtyavrdhn  could you review the failed checks, and closed this merge request. thank you :)

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -2666,7 +2731,23 @@ async def _build_responses_request_params(
...
+            and profile.get('openai_supports_prompt_cache_breakpoints', False)
+            # A `'user'` system prompt role can't be told apart from a real user turn.
+            and system_prompt_role != 'user'
+        )
+        # Instructions live on the current request, which survives the server-side-state trim, so
+        # their static/dynamic makeup can be read from the untrimmed history before resolving.
+        instruction_parts = (
+            self._get_instruction_parts(messages, wire_request_parameters) or [] if cache_instructions else []
+        )
+        cache_instructions_static = _instructions_are_static(instruction_parts) if cache_instructions else None
```
adtyavrdhn: Inline this please, only one consumer

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -621,13 +623,46 @@ def _add_openai_prompt_cache_breakpoint(
...
+    """Whether every instruction part is a literal string, i.e. identical on every request."""
+    return all(not part.dynamic for part in instruction_parts)
+
+
+def _add_instruction_cache_breakpoint(
+    message: chat.ChatCompletionMessageParam | responses.ResponseInputItemParam,
+    text_type: Literal['text', 'input_text'],
+) -> None:
+    """Move a leading message's text into a content block carrying an explicit cache breakpoint."""
+    message_dict = cast('dict[str, Any]', message)
```
adtyavrdhn: Ouch what?

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -1686,6 +1756,19 @@ async def _map_messages(
...
+        if (
+            model_settings
+            and model_settings.get('openai_cache_instructions')
+            and profile.get('openai_supports_prompt_cache_breakpoints', False)
+            # A `'user'` system prompt role can't be told apart from a real user turn, and merging
+            # the leading messages collapses the boundary into one block, so neither can carry it.
+            and system_prompt_role != 'user'
+            and profile.get('openai_chat_supports_multiple_system_messages', True)
+            # A dynamic system prompt changes between requests, so it can't sit in the cached prefix.
+            and not _has_dynamic_system_prompt(messages)
```
adtyavrdhn, line 1772: Should we not break it apart and place the `CachePoint` ?

pydantic_ai_slim/pydantic_ai/models/openai.py
```diff
@@ -2676,6 +2757,28 @@ async def _build_responses_request_params(
...
+                        openai_messages[system_prompt_count:system_prompt_count] = [
+                            responses.EasyInputMessageParam(role=system_prompt_role, content=part.content)
+                            for part in instruction_parts
+                        ]
+                        instructions = OMIT
+                    _add_instruction_cache_breakpoint(openai_messages[index], 'input_text')
+            elif previous_response_id:
+                # A static-instruction chain: the relocated instructions are already replayed from
+                # the chained response's input, so resending them here would send them twice.
+                instructions = OMIT
```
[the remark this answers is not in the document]
adtyavrdhn: Addressed by comparing the current joined instruction text to `ModelRequest.instructions` on the request that produced the chained response.

Same text → keep `previous_response_id` and omit (the cache reuse path). Different text → drop the chain and relocate the current agent's instructions onto full local history. `test_openai_responses_cache_instructions_mismatch_drops_chain` is the collections-continues-refunds case.

sipa-echo-ngbm:
Hi @adtyavrdhn i found that in current on going branch, currently going on is only caching the system prompt, but would be great if caching also for developer role as openai recommends to use it more than the system prompt. (https://developers.openai.com/api/docs/guides/text#message-roles-and-instruction-following, https://model-spec.openai.com/2025-02-12.html#chain_of_command)

Issue #6886, opened 2026-07-29
Temporal: `ctx.usage` mutations inside an activity are silently discarded, so delegate agent usage is lost


DouweM:
### Summary

Under Temporal, anything a tool accrues on `ctx.usage` inside an activity is silently discarded. That breaks the [agent delegation](https://ai.pydantic.dev/multi-agent-applications/#agent-delegation) pattern we document (`agent.run(..., usage=ctx.usage)` inside a tool): the delegate's tokens and requests never reach the parent run, so `result.usage` under-reports and `UsageLimits` (`total_tokens_limit`, `request_limit`, …) is never charged for delegate work.

`TemporalRunContext` carries `usage`/`usage_limits` across the activity boundary and rehydrates them into **new** objects (`durable_exec/temporal/_run_context.py`), which is correct and necessary for the inbound direction (tools and capabilities inside an activity read the run's budget), but there is no return channel for the outbound direction.

### Verified repro

Test in `tests/test_temporal.py` (`FunctionModel`s, no HTTP): a parent agent whose tool delegates with `usage=ctx.usage`. Delegate response usage is `input_tokens=100, output_tokens=10`; each parent response is `input_tokens=5, output_tokens=1`.

```python
@usage_delegation_agent.tool
async def delegate(ctx: RunContext[None]) -> str:
    result = await _usage_delegate_agent.run('delegate this', usage=ctx.usage)
    return result.output
```

| | `result.usage` |
| --- | --- |
| in-process (same agent, no workflow) | `RunUsage(requests=3, input_tokens=110, output_tokens=12, tool_calls=1)` |
| inside a Temporal workflow | `RunUsage(requests=2, input_tokens=10, output_tokens=2, tool_calls=1)` |

The delegate's request and all 110 of its tokens are gone. At the converter level:

```python
serialized = TemporalRunContext.serialize_run_context(ctx)                 # ctx.usage = RunUsage(input_tokens=10, requests=1)
decoded = await pydantic_data_converter.decode(await pydantic_data_converter.encode([serialized]))
activity_ctx = TemporalRunContext.deserialize_run_context(decoded[0], deps=None)
activity_ctx.usage.input_tokens += 100
# activity-side: RunUsage(input_tokens=110, requests=1)
# workflow-side: RunUsage(input_tokens=10, requests=1)   <- unchanged
```

### Engine asymmetry

[...]

PR #7563, opened 2026-08-18
fix: accrue subagent tool usage into the outer run's RunUsage

### LINKED ISSUE #7147: Capability-contributed local tools drop their token usage from `RunUsage` and `UsageLimits`

### Summary

A local tool contributed by a capability that internally spends tokens does not accrue that spend into `RunContext.usage`. `AgentRunResult.usage` and `UsageLimits` therefore see only the outer conversational requests, and the tool's own model spend is invisible to both.

This is not specific to any one tool — it is the current convention across every local subagent-backed tool in the library.

### Verified repro

`ImageGenerationSubagentTool` (`common_tools/image_generation.py`, present on `main` before the image-generation work) receives a `RunContext`, runs a whole nested `Agent.run`, and returns only `result.output`. With `FunctionModel`s where the outer agent bills 5/1 per request and the subagent bills 1000/500:

```
outer run usage : RunUsage(input_tokens=10, output_tokens=2, requests=2, tool_calls=1)

subagent really spent: input_tokens=1000 output_tokens=500 requests=1
accrued into RunUsage: 0 input tokens
```

The limit consequence follows directly — with ~1512 tokens genuinely spent:

```
with total_tokens_limit=100: RunUsage(input_tokens=10, output_tokens=2, requests=2, tool_calls=1) -> no UsageLimitExceeded
```

### Scope: all three local paths, not one

- `ImageGenerationSubagentTool.__call__` (`common_tools/image_generation.py`) — takes `ctx: RunContext[Any]`, runs a nested agent, drops its usage.
- `XSearchSubagentTool.__call__` (`common_tools/x_search.py`) — identical shape, identical drop.
- `_DirectImageGenerationTool.__call__` (`capabilities/image_generation.py`) — drops the spend the same way. Worth noting for anyone implementing: unlike the two above it does **not** take a `RunContext` at all, so wiring it up needs a signature change, not just an added statement.

No library code anywhere accrues into `ctx.usage` today. The convention is a documented **user opt-in**: `docs/multi-agent-applications.md` tells users to pass `usage=ctx.usage` into the delegate run themselves. (`capabilities/reinject_system_prompt.py` passes `ctx.usage` to `Agent.system_prompt_parts`, but that is handing the object to dynamic system-prompt functions to read — it is not an accrual and does not spend tokens.)

### Implementation note: a correct increment is two statements

`RunUsage.incr(RequestUsage)` does **not** bump `requests` — the `self.requests += incr_usage.requests` branch only runs when the argument is itself a `RunUsage` (`usage.py::RunUsage.incr`), and `_agent_graph.py` bumps `requests` separately at its own call sites. So the increment is:

```python
ctx.usage.incr(result.usage)
ctx.usage.requests += 1
```

Getting only the first line right would leave `request_limit` uncharged while token limits started working — a partial fix that is arguably worse than none.

### Likely gated on #6886

Please read #6886 before picking this up. It documents that mutating `ctx.usage` inside a tool is **silently discarded under Temporal** and behaves differently per durable engine (DBOS/Prefect pass the live `RunContext`, so mutations land there; Temporal rehydrates a copy), and it is awaiting a maintainer design decision on a return channel for the outbound direction.

That makes the two issues complementary rather than duplicate — #6886 is about the durable-exec return channel for *user code that already opts in*; this one is about library-owned tools never accruing in the first place. But a fix here that only works in-process would silently do nothing under Temporal, so the ordering matters.

Whatever the decision, the fix should land across all three local paths at once rather than per-capability, so the tools stay consistent with each other.

### Mitigation today

The spend is not invisible everywhere: `images/instrumented.py` emits `gen_ai.usage.*` attributes on its own span (`result.usage.opentelemetry_attributes()` plus explicit input/output token attributes), so image-generation spend does show up in telemetry. It just never reaches `RunUsage` / `UsageLimits`.

aayushbaluni:
Closes #7147

## The problem

Both subagent-backed local tools run a whole nested `Agent.run` and return only `result.output`, so the nested run's spend is dropped. Reproduced on `main` (`fa9561e`) with `FunctionModel`s billing 5/1 on the outer agent and 1000/500 on the subagent:

```
outer run usage      : RunUsage(input_tokens=10, output_tokens=2, requests=2, tool_calls=1)

subagent really spent: input_tokens=1000 output_tokens=500 requests=1
accrued into RunUsage: 0 extra input tokens

with total_tokens_limit=100: NO UsageLimitExceeded
  (~1512 tokens genuinely spent)
```

The limit consequence is the sharp end: a run can burn 1500 tokens inside a tool under a 100-token budget and finish successfully.

## The fix

Pass the outer run's usage into the nested run, in `XSearchSubagentTool.__call__` and `ImageGenerationSubagentTool.__call__`:

```python
result = await agent.run(query, usage=ctx.usage)
```

After:

```
outer run usage      : RunUsage(input_tokens=1010, output_tokens=502, requests=3, tool_calls=1)
accrued into RunUsage: 1000 extra input tokens

with total_tokens_limit=100 -> UsageLimitExceeded The next request would exceed the
total_tokens_limit of 100 (total_tokens=1506).
```

## Three corrections to the issue

I checked each of the issue's implementation claims against `main` rather than following them, and all three have moved:

**1. The third path no longer exists.** `_DirectImageGenerationTool` in `capabilities/image_generation.py` is gone — that file now only defines the `ImageGeneration` capability. So this is two call sites, not three, and no signature change is needed. The issue's warning that the third one "does not take a `RunContext` at all" is moot.

**2. The recommended two-statement increment would double-count `requests`.** The issue says `RunUsage.incr` doesn't bump `requests` for a `RunUsage` argument, so you must also write `ctx.usage.requests += 1`. On current `main` that branch does run (`usage.py:377-379`):

```python
if isinstance(incr_usage, RunUsage):
    self.requests += incr_usage.requests
    self.tool_calls += incr_usage.tool_calls
```

So the prescribed fix would have counted the subagent's request twice. Verified: with `usage=`, `requests` goes 2 → 3, which is exactly the one extra request the subagent made.

**3. `usage=` is the documented mechanism, so no manual accrual is needed.** `Agent.run(usage=...)` is described as *"useful for resuming a conversation or agents used in tools"*, and it is the same thing `docs/multi-agent-applications.md` already tells users to do by hand. I verified the object is shared rather than copied — a tool mutating `ctx.usage` is mutating the run's live `RunUsage` (same `id()`), and the mutation lands in `result.usage`. Using `usage=` therefore makes these library tools do what the docs already ask user code to do, instead of introducing a second accrual convention.

## On the #6886 gating concern

The issue says this is *"likely gated on #6886"* because a fix that only works in-process would silently do nothing under Temporal. That concern is real and I'm not claiming to resolve it. Two things worth separating:

- **In-process, DBOS and Prefect**: the live `RunContext` is passed through, so this works today.
- **Temporal**: `TemporalRunContext` rehydrates `usage` into a new object, so the accrual is discarded — the same limitation that already applies to the documented `usage=ctx.usage` delegation pattern user code is told to write.

So this doesn't make Temporal worse, and it doesn't fix it either: it makes the library's own tools behave exactly like the delegation pattern the docs prescribe, on every engine. When #6886's return channel lands, these two call sites benefit from it automatically, with no change here. If you'd rather hold this until #6886 is decided, that's a reasonable call — but the in-process bug is real today and the fix is two characters shy of a one-liner.

## Tests

Three tests in `tests/test_capabilities.py`:

| Test | Asserts |
|---|---|
| `..._x_search_subagent_usage_accrues_into_run_usage` | `input_tokens == 1010`, `output_tokens == 502`, `requests == 3` |
| `..._x_search_subagent_usage_counts_against_usage_limits` | `UsageLimitExceeded` raised under `total_tokens_limit=100` |
| `..._image_generation_subagent_usage_accrues_into_run_usage` | same accrual for the image tool |

The exact token counts matter here rather than a "greater than" assertion: they're what distinguishes a correct accrual from one that double-counts `requests`, which is the specific way the issue's suggested fix would have gone wrong.

**Mutation-checked.** With both call sites reverted and the tests unchanged, all three fail (`assert 10 == 1010`, `DID NOT RAISE UsageLimitExceeded`); restored, all three pass.

**Baseline.** `tests/test_capabilities.py` 898 passed, `tests/models/xai/test_search_tools.py` + `tests/test_usage_limits.py` 58 passed. `ruff check` and `ruff format --check` clean.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7563?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/common_tools/image_generation.py
+++ b/pydantic_ai_slim/pydantic_ai/common_tools/image_generation.py
@@ -102,7 +102,7 @@ class ImageGenerationSubagentTool:
             instructions=self.instructions,
         )
         try:
-            result = await agent.run(prompt)
+            result = await agent.run(prompt, usage=ctx.usage)
         except UnexpectedModelBehavior as e:
             raise ModelRetry(str(e)) from e
         return result.output
--- a/pydantic_ai_slim/pydantic_ai/common_tools/x_search.py
+++ b/pydantic_ai_slim/pydantic_ai/common_tools/x_search.py
@@ -72,7 +72,7 @@ class XSearchSubagentTool:
             instructions=self.instructions,
         )
         try:
-            result = await agent.run(query)
+            result = await agent.run(query, usage=ctx.usage)
         except UnexpectedModelBehavior as e:
             raise ModelRetry(str(e)) from e
         return result.output
--- a/tests/test_capabilities.py
+++ b/tests/test_capabilities.py
@@ -73,6 +73,7 @@ from pydantic_ai.exceptions import (
     SkipToolValidation,
     ToolFailed,
     UnexpectedModelBehavior,
+    UsageLimitExceeded,
     UserError,
 )
 from pydantic_ai.messages import (
@@ -139,7 +140,7 @@ from pydantic_ai.toolsets._deferred_capability_loader import (
     LOAD_CAPABILITY_ALREADY_AVAILABLE_MESSAGE_TEMPLATE,
     LOAD_CAPABILITY_TOOL_NAME,
 )
-from pydantic_ai.usage import RequestUsage, RunUsage
+from pydantic_ai.usage import RequestUsage, RunUsage, UsageLimits
 from pydantic_graph import End
 
 from ._inline_snapshot import snapshot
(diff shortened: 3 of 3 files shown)


aayushbaluni:
@macroscopeapp and @veria-ai both flagged that `usage_limits` should be propagated alongside `usage`. **They're right, and I've pushed the fix.** Thanks — this was a genuine gap, not a nit.

## Confirming the finding

`usage=` alone accrues the spend but doesn't *bound* it, so the subagent issues its paid request and is only caught once it returns. Measured on this branch with `request_limit=1`, counting actual subagent model requests:

| | subagent requests actually made |
|---|---|
| `usage=` only | **1** |
| `usage=` + `usage_limits=` | **0** |

Both raise `UsageLimitExceeded`, so a test asserting only "did it raise" passes either way. The difference is whether the request was *paid for* before anyone noticed.

It gets worse when the subagent iterates. With a subagent that loops internally on a 2500-token budget:

```
usage= only                : 11 subagent requests, 11001 tokens
usage= + usage_limits=     :  3 subagent requests,  3001 tokens
```

So the original PR fixed the accounting but left the enforcement gap open. Fixed in both tools:

```python
result = await agent.run(query, usage=ctx.usage, usage_limits=ctx.usage_limits)
```

`ctx.usage_limits` is documented for exactly this — *"Combine it with `usage` to compute the remaining budget"* — and it's `None` only outside a run, so it degrades to today's behaviour when unset.

## Test added

`test_x_search_subagent_is_bound_by_the_outer_request_limit` asserts `inner_requests == 0`, i.e. the subagent never issued a request, rather than just that an exception surfaced. That's the assertion that actually separates the two behaviours above.

**Re-mutation-checked with two distinct mutations**, since there are now two independent halves:

| Mutation | Result |
|---|---|
| A: drop both (`agent.run(query)`) — the original bug | **4 failed**, 3 passed |
| B: `usage=` but no `usage_limits=` — precisely your finding | **1 failed**, 6 passed |
| restored | **7 passed** |

Mutation B failing exactly one test, the new one, is the point: the pre-existing tests could not tell the difference, which is why the gap survived the first round.

Baseline: `tests/test_capabilities.py` 899 passed. `ruff check` and `ruff format --check` clean.

---

@Kludex @DouweM — this was auto-closed because #7147 isn't assigned to me, and CI was green (42 checks) before the reviewers' finding. Happy to reopen if the approach looks right; the two commits are separable if you'd prefer only the accrual.

One judgement call worth your input: the issue notes this is *"likely gated on #6886"* because the accrual is discarded under Temporal. My view is this doesn't make Temporal worse — the library's tools now behave exactly like the `usage=ctx.usage` delegation pattern the docs already tell users to write, on every engine, and they'll pick up #6886's return channel for free when it lands. But if you'd rather wait for that design decision, that's a reasonable call and I won't push back.

aayushbaluni:
Follow-up on the coverage failure, and a note that this PR's diff is now stale through no fault of the branch.

**`coverage` was a real failure and is fixed.** The repo enforces `fail-under=100`, and my tests left two lines unreachable by construction:

```
tests/test_capabilities.py   10695   1   770   1   99.98%   8102
Coverage failure: total of 99.99 is less than fail-under=100.00
```

1. In the `total_tokens_limit` test the run aborts *inside* the tool, so the outer model is never asked for a second response — that `if len(messages) == 1:` branch was dead. Removed it.
2. In the request-limit test the subagent model function is never called, which is the entire point of the test (`inner_requests == 0`). Moved the `# pragma: no cover` from the `return` to the function, with a comment saying why.

Reproduced CI's number locally first (`coverage run -m pytest` → `combine` → `report`), rather than guessing:

```
before: 10705 stmts, 1 miss, 99.99%  -> Coverage failure
after : 10703 stmts, 0 miss, 100.00%
```

899 tests pass, `ruff check` and `ruff format --check` clean. Re-ran the mutation check afterwards, since I'd edited the tests: mutation A (drop both) fails 3, mutation B (`usage=` without `usage_limits=`) fails exactly the new test, restored passes 7.

**The stale-diff problem.** GitHub does not sync new commits onto a closed PR, so the diff above still shows only the first commit (`2f9cbea`). The branch actually has three:

```
7d0694b test: restore 100% coverage in the new subagent usage tests
87120e8 fix: propagate the outer run's usage limits into subagent tools
2f9cbea fix: accrue subagent tool usage into the outer run's RunUsage
```

and the current state of the fix is:

```python
result = await agent.run(query, usage=ctx.usage, usage_limits=ctx.usage_limits)
```

I tried to reopen to force a resync and can't — `Could not open the pull request`, since I'm not a maintainer. So **the two failing checks visible on this PR are against `2f9cbea` only**; they don't reflect the branch. Reopening (or a maintainer assigning #7147, which is what closed it) will pull in the other two commits and re-run CI against the real state.

Sorry for the noise — flagging it explicitly because "PR shows failing checks" is misleading here, and I'd rather say so than let it sit looking broken.

PR #7607, opened 2026-08-19
Fix the PR template link in the `add-new-model` skill

ayaangazali:
`.agents/skills/add-new-model/SKILL.md:135` links the PR template as `.github/pull_request_template.md`. The identical sentence lives in `AGENTS.md:71`, where that path is right because AGENTS.md sits at the repo root. The skill file is three directories down, so the link resolves to `.agents/skills/add-new-model/.github/pull_request_template.md` and 404s when you click it on GitHub.

Changed it to `../../../.github/pull_request_template.md`, which I verified resolves to the real file.

Two reasons it went unnoticed, in case they're useful: it's the only relative markdown link anywhere under `.agents/skills/` (every other link in those files is an absolute GitHub URL), and the weekly lychee run can't see it because `.github/workflows/lychee.toml` sets `scheme = ["https", "http"]`, which skips relative links deliberately.

No issue linked, since the template lists broken links as a trivial fix.

### Checklist

---

A bit about me: I'm a freshman in college, still getting comfortable in a codebase this big. I read this diff line by line myself and used Claude Code as a second opinion on the reasoning before opening it. If the convention you'd prefer is a plain backticked path (the way every other repo file is referenced in that skill) rather than a working relative link, tell me and I'll switch it.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7607?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.agents/skills/add-new-model/SKILL.md
+++ b/.agents/skills/add-new-model/SKILL.md
@@ -132,7 +132,7 @@ Terse, not botty. Structure:
 - Flag pre-existing latent bugs found but deliberately not fixed.
 - Link the prior add-model PR for context.
 
-Include the [PR template](.github/pull_request_template.md), fill in the issue number, and check the "AI generated code" box in the GitHub UI yourself — `gh pr create` cannot set it.
+Include the [PR template](../../../.github/pull_request_template.md), fill in the issue number, and check the "AI generated code" box in the GitHub UI yourself — `gh pr create` cannot set it.
 
 ## Provider-specific landmines
 


ayaangazali:
Closing this, #7623 got there a better way.

That PR made `pushing-commits-to-the-repo` the canonical PR guide and dropped the sentence this was fixing, so the broken link is gone from `.agents/skills/add-new-model/SKILL.md` entirely rather than repaired. I checked the guidance did not just move: the only `pull_request_template` reference left is `AGENTS.md:71`, and that one is root-relative so it resolves correctly, and the new canonical skill has no relative links at all.

This branch is conflicting against main now anyway, and deleting the line beats fixing it.

Issue #4458, opened 2026-02-28
Improve OpenTelemetry naming documentation clarity


Quirson:
### Description

While reviewing the OpenTelemetry span naming updates, I noticed that the documentation could benefit from a clearer explanation of:

- The difference between InstrumentationSettings version 2 and version 3.
- The reasoning behind duplicating attributes (e.g. agent_name and gen_ai.agent.name).
- Migration recommendations for users upgrading dashboards.

I would be happy to submit a PR to improve the documentation if this is useful.

adtyavrdhn:
I am not sure how much of this belongs in our documentation vs Otel's documentation of these conventions.

I agree with you on the migrations parts though I can see that being useful for users.

@DouweM What do you think?

DouweM:
@adtyavrdhn The instrumentation versions at https://ai.pydantic.dev/logfire/#configuring-data-format are ours, so this definitely concerns our docs.

@Quirson PR welcome! Since you know better what you found confusing than we do :)

ayaangazali:
hey @DouweM, since this went quiet and the docs have moved on a bit since March, I dug into where things stand now and I'd love to carry the PR if that works.

what's already covered: https://ai.pydantic.dev/logfire/#configuring-data-format now explains the version system well (v5 default, v2/3/4 deprecated, per-version sections). so the first bullet of this issue is mostly addressed.

what's still missing, as far as I can tell: the v2 to v3+ renames are never listed in one place. looking at `_instrumentation.py` (`InstrumentationConfig.for_version`), v2 emits `agent run` / `agent_name` / `running tool` / `tool_arguments` / `tool_response`, while v3+ emits `invoke_agent` / `gen_ai.agent.name` / `execute_tool` / `gen_ai.tool.call.arguments` / `gen_ai.tool.call.result`. the docs only mention the `invoke_agent` rename in passing, and there's no guidance for migrating dashboards or alerts keyed on the old names (the second and third bullets here).

my proposal: add a small rename table to the versions section of `docs/logfire.md` plus a short dashboard migration note. happy to open that PR if you're good with it, and no worries if you'd rather handle it differently since this one's been sitting a while.

PR #7609, opened 2026-08-19
Document the `running output function` span rename and the duplicated agent name attribute

ayaangazali:
Two gaps in the "Configuring data format" section of `docs/logfire.md`, both from #4458, which asked for a clearer account of what changes between instrumentation versions.

The version 3 list presents itself as the full set of renames, but it covers four of the six fields that actually differ. `InstrumentationNames.for_version` (`pydantic_ai_slim/pydantic_ai/_instrumentation.py:691`) also changes the output-tool span from `running output function` to `execute_tool`, applied through `get_output_tool_span_name` at `capabilities/instrumentation.py:585`. `tests/test_logfire.py::test_logfire_output_function_v2_v3` asserts exactly that: `running output function` on version 2, `execute_tool final_result` on version 3. Someone upgrading and grepping for their output-function spans currently has nothing in the docs to explain where they went. Added it as a third bullet, using `{gen_ai.tool.name}` because the span name and that attribute are both set from the same `span_target` (`capabilities/instrumentation.py:557,585`).

The issue also asked why `agent_name` and `gen_ai.agent.name` are duplicated. They are both set unconditionally on the run span at `capabilities/instrumentation.py:170-171`, at every version, and `realtime/_instrumentation.py:111` mirrors it with a comment explaining the intent. That is deliberate and useful to know before you change versions, so I stated it once in the intro rather than inside a single version's section.

The sixth differing field is `agent_name_attr`, and I left the docs alone there on purpose: it is set per version but never read, so `agent_name` is not actually renamed and documenting a rename would be wrong. I also left #4458's dashboard-migration bullet alone rather than inventing guidance I cannot verify, so this does not close the issue.

Ran `uv run pytest tests/test_examples.py -k logfire`. Four failures there reproduce identically on unmodified main in my environment (Python 3.14 local), so they are not from this change.

### Checklist

---

Context on me: I'm a college freshman still learning how a project this size fits together. I traced both claims to the code and the tests before writing the prose, and used Claude Code to pressure-test my reasoning, but I read every line myself and stand behind it. If you'd rather keep the deprecated-version sections lean and skip the extra bullet, I'm happy to trim it down to just the duplicated-attribute note.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7609?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/logfire.md
+++ b/docs/logfire.md
@@ -299,6 +299,8 @@ Versions 2, 3, and 4 are deprecated compatibility formats. Passing one of these 
 
 Version 6 is opt-in: it changes the role of messages you already receive, so pass it explicitly once your telemetry consumer is ready for the new role.
 
+Agent run spans carry the agent name under both `agent_name` and `gen_ai.agent.name` at every version, so a dashboard keyed on either name keeps working when you change versions.
+
 #### Version 2 (deprecated)
 
 Uses the newer OpenTelemetry GenAI spec and stores messages in the following attributes:
@@ -316,6 +318,7 @@ Builds on version 2 with the following improvements:
 - **Spec-compliant span names:**
     - `agent run` becomes `invoke_agent {gen_ai.agent.name}` (with the agent name filled in)
     - `running tool` becomes `execute_tool {gen_ai.tool.name}` (with the tool name filled in)
+    - `running output function` becomes `execute_tool {gen_ai.tool.name}`, so output tools and output functions share the span name used for other tools
 - **Spec-compliant attribute names:**
     - `tool_arguments` becomes `gen_ai.tool.call.arguments`
     - `tool_response` becomes `gen_ai.tool.call.result`


ayaangazali:
Re-checked this against current main (ada2a2d3d) since it has been open a week.

Both parts still apply. The version 3 list in `docs/logfire.md` still covers four of the six fields `InstrumentationNames.for_version` changes, and `output_tool_span_name` is still `running output function` on version 2 versus `execute_tool` on version 3 and later (`_instrumentation.py:697` and `:706`), so the output-tool rename is still missing from that list. `tests/test_logfire.py::test_logfire_output_function_v2_v3` still asserts both sides of it.

No urgency from my side. If you would rather keep the deprecated-version sections lean and skip the extra bullet, I am happy to trim this down to just the duplicated-attribute note, or to close it.

Issue #7620, opened 2026-08-19
Unrecognized YouTube hosts silently send the watch page instead of the video


dsfaccini:
> This issue was posted by Claude Code using claude-opus-5 on behalf of David.

`VideoUrl` silently sends a YouTube watch page as the video when the host isn't a recognized YouTube host.

`VideoUrl.is_youtube` is an allowlist of hosts Gemini accepts as a `file_uri`. Any other `*.youtube.com` host falls through to `download_item`, which fetches the **watch page HTML** and sends it as the video content — on every provider, with no error.

Measured on `https://music.youtube.com/watch?v=lCdaVNyHtjU`:

| | |
|---|---|
| `download_item` fetches | `200`, `text/html`, ~5 KB SPA shell |
| sent to the model as | `inline_data`, `mimeType: text/html` |
| same URL as a `file_uri` | Gemini `400 INVALID_ARGUMENT` (recorded, reproduced 3×) |

So there is no working path for these hosts, and the user gets no signal that anything went wrong.

**Open question — what should happen instead?** It affects every provider, not just Gemini, so it wants a decision rather than a patch:

1. Raise a `UserError` naming the host as an unsupported YouTube surface.
2. Emit a warning and keep downloading.
3. Leave as-is and document it.

<details><summary>Recorded Gemini rejection</summary>

```yaml
request:
  parts:
  - text: Explain me this video in a few sentences
  - fileData:
      fileUri: https://music.youtube.com/watch?v=lCdaVNyHtjU
      mimeType: video/mp4
  uri: https://generativelanguage.googleapis.com/v1beta/models/gemini-2.5-flash:generateContent
response:
  error:
    code: 400
    message: Request contains an invalid argument.
    status: INVALID_ARGUMENT
```

Control: `youtu.be`, `youtube.com`, `www.youtube.com` and `m.youtube.com` all return `200` for the same
video with identical `VIDEO` prompt-token counts (15780), so the rejection is specific to the host.
</details>

Found while scoping #7592 / #7591, which adds `m.youtube.com` to the allowlist. Deliberately left out of that PR: fixing this changes behavior for all providers and needs the decision above.

PR #7592, opened 2026-08-19
Recognize `m.youtube.com` URLs in `VideoUrl`

uczltw6:
> This pull request body was written by Claude Code using claude-opus-5 on behalf of David.
> David decided the scope of the change and has not reviewed the implementation.

Closes #7591

## Why we make these changes

`VideoUrl.is_youtube` matched an exact host tuple that omitted `m.youtube.com` — the host YouTube's own mobile share sheet produces. Unrecognized hosts don't take the YouTube `file_uri` path; they fall through to `download_item`, which fetches the **watch page** and sends that HTML as the video. Measured on `main`: `m.youtube.com/watch?v=…` 302s to `www.youtube.com/watch?app=desktop&v=…` and returns 1.1 MB of `text/html`. The model receives a web page, and nothing errors.

This adds `m.youtube.com` to the allowlist, verified against the provider rather than against the URL grammar.

**It deliberately does not generalize to `*.youtube.com`.** Gemini rejects `music.youtube.com` as a `file_uri`, so a suffix match would sweep in a host the provider cannot resolve — trading a silent failure for a hard 400. `music.youtube.com` therefore stays unrecognized and its behavior is **byte-identical to `main`**. The residual bug it represents is filed as #7620.

## New public surface

None. `VideoUrl.is_youtube` already existed; only which hosts it matches changed.

<details>
<summary><b>Goal:</b> <code>m.youtube.com</code> links reach Gemini as a <code>file_uri</code></summary>

**Before → after**

```python
from pydantic_ai import VideoUrl

VideoUrl(url='https://m.youtube.com/watch?v=lCdaVNyHtjU').is_youtube
# before -> False   (watch page downloaded, 1.1 MB text/html sent as the video)
# after  -> True    (URL passed straight through as file_uri)

VideoUrl(url='https://music.youtube.com/watch?v=lCdaVNyHtjU').is_youtube
# before -> False
# after  -> False   (unchanged — Gemini 400s this host as a file_uri)

VideoUrl(url='https://youtube.com.example.com/video.mp4').is_youtube
# before -> False
# after  -> False   (unchanged — now covered by a regression test)
```

<details>
<summary>Playground</summary>

```python
import os

from pydantic_ai import Agent, VideoUrl
from pydantic_ai.models.google import GoogleModel
from pydantic_ai.providers.google import GoogleProvider

agent = Agent(GoogleModel('gemini-2.5-flash', provider=GoogleProvider(api_key=os.environ['GOOGLE_API_KEY'])))

[...]

Issue #7591, opened 2026-08-19
`VideoUrl` does not recognize YouTube subdomains


uczltw6:
### Initial checks

- [x] Reproduced on the current `main` branch.
- [x] Searched the issue and pull request history for the same behavior.

### Description

`VideoUrl.is_youtube` recognizes `youtu.be`, `youtube.com`, and `www.youtube.com`, but returns `False` for official subdomains such as `m.youtube.com` and `music.youtube.com`.

This also affects `GoogleModel`: documented YouTube URLs should be passed directly to Google, but an unrecognized YouTube subdomain instead follows the ordinary `FileUrl` download path.

Expected behavior: any `youtube.com` subdomain is treated as YouTube, while unrelated suffixes such as `youtube.com.example.com` remain excluded.

### Minimal reproducible example

```python
from pydantic_ai import VideoUrl

assert VideoUrl(url='https://m.youtube.com/watch?v=lCdaVNyHtjU').is_youtube
assert VideoUrl(url='https://music.youtube.com/watch?v=lCdaVNyHtjU').is_youtube
```

Both assertions fail on `main` at `4167d2c8daf338e85d782c8048ffdec859f869cc`.

### Versions

- Python: 3.14.4
- Pydantic AI: `0.0.1.dev1+4167d2c`
- LLM provider SDK: not required for this reproduction

fszcd:
I reproduced this on `main` and prepared a focused fix in [PR #7638](https://github.com/pydantic/pydantic-ai/pull/7638): `VideoUrl.is_youtube` now recognizes official `youtube.com` subdomains while excluding deceptive suffixes, with regression coverage for `m.youtube.com`, `music.youtube.com`, and `youtube.com.example.com`.

The PR was auto-closed by the duplicate-effort guard because this issue is currently assigned to a maintainer. Please assign the issue to `fszcd` if you would like me to carry this patch through review.

PR #7280, opened 2026-08-07
Route Vertex-vs-Gemini API branches in `GoogleModel` by client transport instead of provider name

thejamesgore:
- Closes #6792

## Why we make these changes

`GoogleModel` currently decides between Google Cloud (Vertex AI) and the Gemini Developer API from the provider name. A pre-built client can disagree with that name, so `GoogleProvider(client=Client(vertexai=True))` takes Gemini API-only branches and can fail before sending the request.

This changes transport-specific routing to use the client's public `vertexai` property while preserving the provider name stored in message history. It also fixes the mirrored `GoogleCloudProvider(client=Client(vertexai=False))` case and keeps custom provider names scoped to their existing allow-list.

## New public surface

None.

## User-visible behavior

```python
client = Client(vertexai=True, project='project', location='us-central1')
model = GoogleModel('gemini-3-flash-preview', provider=GoogleProvider(client=client))

# Before: Gemini API routing; Vertex-only requests could fail during request construction.
# After: Google Cloud routing, matching the client's actual transport.
```

The inverse construction, `GoogleCloudProvider(client=Client(vertexai=False, api_key='...'))`, now follows the Gemini Developer API transport as well.

## Verification

- [Transport routing, service-tier headers, uploaded-file validation, and custom provider names](https://github.com/pydantic/pydantic-ai/blob/c79335ca156aa72fdd60cbc78c2d53e44cd7a44c/tests/models/google/test_transport.py#L41-L174)
- [History replay for providers whose name and transport disagree](https://github.com/pydantic/pydantic-ai/blob/c79335ca156aa72fdd60cbc78c2d53e44cd7a44c/tests/v2/test_google_history_replay.py#L49-L181)
- Existing Vertex-specific tests now construct the reported client/provider combination instead of mocking `system`.

## What changes for existing users

Normal `GoogleProvider(...)` and `GoogleCloudProvider(...)` construction is unchanged. Models given a pre-built client now consistently follow that client's actual transport instead of a potentially conflicting provider name.

### Checklist

- [x] Any **AI generated code** has been reviewed line-by-line by the human PR author, who stands by it.
- [x] No **breaking changes** in accordance with the [version policy](https://github.com/pydantic/pydantic-ai/blob/main/docs/version-policy.md).
- [ ] Any permitted **compatibility impact** has a label, warning, migration, release note, and exact API-check waiver.
- [x] **PR title** is fit for the [release changelog](https://github.com/pydantic/pydantic-ai/releases).

PR #7649, opened 2026-08-20
Document the YouTube hosts `GoogleModel` accepts as video URLs

ayaangazali:
`docs/models/google.md` shows a YouTube example using `www.youtube.com`, but never says which YouTube hosts actually work. The allowlist is exact and lives only in a source docstring: `VideoUrl.is_youtube` (`pydantic_ai_slim/pydantic_ai/messages.py:360`) matches `youtu.be`, `youtube.com`, `www.youtube.com` and `m.youtube.com`, and nothing else.

The part worth writing down is what happens when a host misses. `GoogleModel._resolve_file` only passes a video straight through as a `file_uri` when `is_youtube` is true (`models/google.py:1258`). Otherwise it falls to the generic `FileUrl` branch, and on the Gemini API that calls `download_item`, which fetches the URL. For a YouTube link that means the model gets the watch page instead of the video. It fails quietly too: the `UserError('Downloading YouTube videos is not supported.')` guard in `download_item` only fires when `is_youtube` is true, so an unrecognized host never reaches it.

I kept the download sentence scoped to the Gemini API on purpose. On Vertex the same URL takes the `else` at `models/google.py:1269` and goes out as a `file_uri` instead, so saying it is always downloaded would be wrong.

`music.youtube.com` gets named explicitly because it is the case people keep hitting: #7591 and #7620 both reported YouTube hosts not being recognized, and #7593 and #7638 both proposed a `*.youtube.com` suffix match before #7592 landed the exact-host list instead. The reason it cannot be a suffix match is already in the code comment, that Google rejects `music.youtube.com` as a `file_uri` with a 400 on the Gemini API and on Vertex alike, so I reused that rather than invent a new explanation.

Docs only, no behavior change. Ran `uv run pytest tests/test_examples.py -k google` (19 passed, 16 skipped).

### Checklist

---

About me: I'm a freshman in college, so I traced this through `messages.py` and `models/google.py` rather than assume I knew the behavior, and I used Claude Code to check my reasoning before writing it up. My first draft claimed the fallback always downloads, which reading the Vertex branch showed was wrong, so I narrowed it. If you'd rather this live in `docs/input.md` next to the other file-URL caveats, or be trimmed to one sentence, just say and I'll move it.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7649?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/models/google.md
+++ b/docs/models/google.md
@@ -337,6 +337,8 @@ result = agent.run_sync(
 print(result.output)
 ```
 
+Only `youtu.be`, `youtube.com`, `www.youtube.com`, and `m.youtube.com` are recognized as YouTube hosts and handed to the model as a video URL. Any other YouTube domain falls back to the ordinary file URL path, which on the Gemini API downloads the URL, so the model receives the watch page rather than the video. `music.youtube.com` is excluded deliberately: Google rejects it as a `file_uri` with a 400, on the Gemini API and on Vertex alike.
+
 Files can be uploaded via the [Files API](https://ai.google.dev/gemini-api/docs/files) and passed as URLs:
 
 ```py {title="file_upload.py" test="skip"}


ayaangazali:
Re-checked this against current main (a5eae608b) now that it has been open a week.

Still applies unchanged. `VideoUrl.is_youtube` at `messages.py:361` still matches exactly `youtu.be`, `youtube.com`, `www.youtube.com` and `m.youtube.com`, and `docs/models/google.md` still says nothing about which hosts work, so the `music.youtube.com` case that #7591 and #7620 both ran into is still undocumented.

No urgency. If you would rather this lived in `docs/input.md` next to the other file-URL caveats, or were cut down to a single sentence, say which and I will change it.

ayaangazali:
Line-number correction, since #7280 landed after I wrote this and moved the code I cited.

The behavior and the diff are unchanged, but if you go looking for the references in the description:

- the pass-through branch is now `models/google.py:1280`, not `:1258`
- the fallback branch is now `:1283` to `:1285`, not `:1269`
- that fallback no longer keys on `_GEMINI_API_PROVIDER_NAMES`, it keys on `self._is_google_cloud`, which #7280 derives from the client transport rather than the provider name

The sentence this PR adds still holds under that routing: a host outside the `is_youtube` set is not passed through as a `file_uri`, and on the Gemini API path it gets downloaded, which is how the watch page ends up in front of the model instead of the video.

PR #7684, opened 2026-08-21
Document the AG-UI dojo `Tool Approval` example

ayaangazali:
The AG-UI dojo backend mounts seven feature endpoints, but `docs/examples/ag-ui.md` documents six. `tool_approval` is the one with no section.

It is a real served endpoint, not a leftover: `examples/pydantic_ai_examples/ag_ui/__init__.py:38` mounts it as `Tool Approval (interrupts)`, and `api/__init__.py` exports `tool_approval_app` alongside the other six. It arrived with #5441 (`Map AG-UI interrupts to DeferredTools in AGUIAdapter`), which added and wired the example but did not add it to the page. Every one of its six siblings is pulled into the docs with a `snippet` directive; this one is the only file in that directory that no page references.

I put the section between `Shared State` and `Tool Based Generative UI`, because the page's section order already matches the mount order in `__init__.py` exactly, and that is where `/tool_approval` sits. The section follows the same shape as the others: description, tools, prompt, then the code snippet. The description is taken from the example's own module docstring rather than written fresh, so the interrupt lifecycle it describes matches what the code does.

One thing I deliberately left alone: the module docstring at the top of `examples/pydantic_ai_examples/ag_ui/__init__.py` also lists the dojo features and also omits Tool Approval, so it says six while the app mounts seven. That is a `.py` file, and including it would make this PR no longer docs-only, which trips the contributor guard on an unassigned change. Happy to send it as a separate PR if you would rather have both, or you may prefer to fix it in passing.

Verified before opening: the snippet target exists, `../deferred-tools.md#human-in-the-loop-tool-approval` resolves to a real heading, and `DeferredToolRequests` is rendered on the output API page so the cross-reference resolves. Ran `uv run pytest tests/test_examples.py -k ag_ui` (9 passed).

### Checklist

---

About me: I'm a college freshman still learning this codebase, so I checked how the dojo actually wires these endpoints before assuming the gap was real, and I used Claude Code as a second opinion on the wording. I read the diff line by line and stand behind it. If the omission was intentional, say so and I'll close this without any fuss.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7684?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ayaangazali:
Rebased onto be329c694 and pinned the heading anchor.

#7693 landed after I opened this and gave every `#### <Feature> - Code` heading on this page an explicit id, so mine was the only one without one. Added `{#tool-approval-code}` to match. Per the rule that PR added to `docs/AGENTS.md`, pinning belongs before anything links to the heading rather than after, since an explicit id replaces the generated one, and unpinned this would have generated `#tool-approval---code` with the tripled dash.

No other heading in the section needs pinning, `Tool Approval`, `Approval Tools` and `Approval Prompt` all slugify cleanly.

ayaangazali:
Re-checked against current main (4f5222649) now that this has been open a week.

Still accurate. `ag_ui/__init__.py:38` still mounts `/tool_approval` as `Tool Approval (interrupts)`, so the dojo backend still serves seven feature endpoints while `docs/examples/ag-ui.md` documents six, and `tool_approval.py` is still the only file in `ag_ui/api/` that no page references. The module docstring at the top of that file still lists six features too, which is the `.py` half I left out of this PR to keep it docs-only.

Clean against main and CI is green, including the `docs-assets` anchor check that #7693 added after I opened this, so the pinned `{#tool-approval-code}` id resolves.

No rush. If you would rather not grow that page, I am happy to close this.

ayaangazali:
Rebased onto current main and fixed one thing I got wrong.

The paragraph linked `[DeferredToolRequests][pydantic_ai.output.DeferredToolRequests]`, and that target does not resolve. The class lives in `pydantic_ai._deferred` and is re-exported by `pydantic_ai.tools` and by `pydantic_ai` itself, but not by `pydantic_ai.output`. Every other reference in the docs uses `pydantic_ai.tools.DeferredToolRequests`, sixteen of them, so this one was the outlier. Now it matches.

Re-checked the rest on current main while I was here: the dojo still mounts seven apps and `docs/examples/ag-ui.md` still documents six, with Tool Approval the one missing, so the gap this PR fills is unchanged. `uv run pytest tests/test_examples.py -k ag_ui` passes, 11 tests.

Issue #7725, opened 2026-08-24
Stabilize tests that assert timing, global context, or concurrent sibling order


adtyavrdhn:
Recent `main` CI failures expose five test-design problems rather than one production regression:

- `test_sync_stream_bridge_init_propagates_base_exception[SystemExit]` uses a one-second anti-hang callback, so a slow job can let the test's own watchdog stop the loop ([run](https://github.com/pydantic/pydantic-ai/actions/runs/32498362182/job/96822063593)). This path is a necessary regression guard, but its watchdog should count event-loop progress rather than elapsed time.
- `test_span_tree_any` builds pure query-test data through live Logfire/OTel state and sometimes receives an empty tree ([run](https://github.com/pydantic/pydantic-ai/actions/runs/32305486908/job/96237211517)). This victim can be made deterministic without claiming the wider async OTel context-isolation class is solved.
- `test_durability_complex_agent_logfire_span_tree` treats exporter arrival order among concurrent sibling tool spans as contractual ([run](https://github.com/pydantic/pydantic-ai/actions/runs/32468135964/job/96729058334)). Cross-tool interleaving should be flexible, while tool discovery, model-turn boundaries, per-tool lifecycle, and output lifecycle remain explicitly ordered.
- `test_sync_entry_points_keep_async_client_on_one_event_loop` uses a one-second response timeout against a local background server even though its contract is event-loop affinity ([run](https://github.com/pydantic/pydantic-ai/actions/runs/32701242939/job/97353021899)). A persistent in-process transport can expose the calling loop directly, eliminating the socket, server thread, and deadline.
- `test_xai_pass_xai_client` constructs an asyncio-bound SDK client from a synchronous test and fails if a prior test leaves no ambient default loop ([run](https://github.com/pydantic/pydantic-ai/actions/runs/31712722673/job/94489565365)). The client should be created and closed inside the async test's running loop; a separate missing-default-loop fixture is then redundant.

The fix should preserve the real behavioral contracts: use deterministic loop-turn anti-hang guards, construct deterministic `SpanNode` query fixtures while retaining separate live-capture tests, allow only proven concurrent sibling interleaving while asserting causal phases, directly assert loop identity at the persistent transport boundary, and create/close the xAI client inside a running loop.

Temporal deadlock-detector false positives, Prefect server startup, coverage-fragment loss, `setup-uv` downloads, and interpreter crashes have different ownership and should not be hidden by weakening these tests.

[...]

PR #7726, opened 2026-08-24
Stabilize timing- and concurrency-sensitive tests

adtyavrdhn:
Five recent flakes came from tests asserting accidental timing, ambient state, or exporter order instead of their intended contract. This keeps the behavioral coverage while removing only those unrelated dependencies.

- Exercise both sync-stream BaseException paths with an event-loop-turn watchdog. A stalled worker cannot consume turns, so the anti-hang guard no longer depends on elapsed time.
- Build the predefined `SpanTree` query fixture from deterministic `SpanNode`s. The pure query tests no longer depend on live global OTel context; the capture-specific tests still exercise Logfire/OTel integration.
- Compare concurrent Temporal agent-run children structurally, then separately assert the real causal contract: tool discovery precedes model work, model turns remain sequential, each tool keeps call/run/result order within its turn, and the output call precedes its result. Negative tests prove invalid causal order is still rejected.
- Construct and close the xAI `AsyncClient` inside the test's running AnyIO loop; no ambient default-loop fixture is needed.
- Replace the local Anthropic server and response deadline with an in-process persistent transport that records the running loop for every request. The test now directly requires both sync entry points to use the exact same loop, with no socket, server thread, or timeout.

Failure evidence:

- Sync bridge watchdog: [run 32498362182](https://github.com/pydantic/pydantic-ai/actions/runs/32498362182/job/96822063593)
- Empty `SpanTree`: [run 32305486908](https://github.com/pydantic/pydantic-ai/actions/runs/32305486908/job/96237211517)
- Concurrent Temporal span order: [run 32468135964](https://github.com/pydantic/pydantic-ai/actions/runs/32468135964/job/96729058334)
- Local keep-alive timeout: [run 32701242939](https://github.com/pydantic/pydantic-ai/actions/runs/32701242939/job/97353021899)
- xAI ambient loop: [run 31712722673](https://github.com/pydantic/pydantic-ai/actions/runs/31712722673/job/94489565365)

Validation:

- 47 focused tests across all five modified files passed together.
- The no-timeout loop/bridge/xAI targets passed twenty repeated runs: 320 executions.
- A deliberate pre-fix loop-affinity reproduction fails immediately when the second sync operation reaches the transport on a different loop; the current bridge passes.
- A deliberate broken BaseException-forwarding reproduction reaches the loop-turn watchdog; the current bridge propagates immediately without reaching it.
- The Temporal targets passed five repeated runs (15 executions total).
- Focused branch coverage reports no uncovered lines or branches in the new deterministic helpers and negative tests.
- Ruff check/format and targeted Pyright passed for all five modified files.
- The full local pre-commit typecheck was skipped only because this focused environment lacks unrelated optional provider/harness packages; all other hooks passed, and the fully provisioned CI typecheck remains required.

Scope note: deterministic input fixes the pure `SpanTree` query-test flake, but this PR does not claim to solve every possible async OTel context-isolation failure. Temporal deadlock-detector false positives already mitigated by #7451, Prefect startup, coverage-fragment loss, `setup-uv` downloads, and interpreter crashes are also intentionally excluded because weakening tests would hide those harness/infra failures.

Refs #7725

### Checklist

### DIFF
--- a/tests/evals/test_otel.py
+++ b/tests/evals/test_otel.py
@@ -98,7 +98,10 @@ async def test_context_subtree_concurrent():
 
 @pytest.fixture
 def span_tree() -> SpanTree:
-    """Fixture that creates a span tree with a predefined structure and attributes."""
+    """Build deterministic input for pure tree queries, which have no provider request to record with VCR.
+
+    Live Logfire/OTel capture is exercised separately by the `context_subtree` tests above.
+    """
 
     def make_span(
         name: str,
--- a/tests/evals/test_otel.py
+++ b/tests/evals/test_otel.py
@@ -1,6 +1,7 @@
 from __future__ import annotations as _annotations
 
 import asyncio
+from datetime import datetime, timedelta, timezone
 
 import pytest
 from pytest_mock import MockerFixture
@@ -12,7 +13,7 @@ with try_import() as imports_successful:
     from pydantic_evals.otel._context_subtree import (
         context_subtree,
     )
-    from pydantic_evals.otel.span_tree import SpanQuery, SpanTree
+    from pydantic_evals.otel.span_tree import AttributeValue, SpanNode, SpanQuery, SpanTree
 
 with try_import() as logfire_import_successful:
     import logfire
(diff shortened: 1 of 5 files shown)


adtyavrdhn:
Superseded by five independently mergeable PRs:

- #7736: deterministic `SyncStreamBridge` watchdog
- #7739: deterministic `SpanTree` query fixture
- #7740: Temporal causal-order assertions
- #7737: xAI client loop lifecycle
- #7738: direct sync-client loop-affinity assertion

Each targets `main` and can be reviewed or merged in any order.

PR #7731, opened 2026-08-24
Speed up OpenAI-compatible provider import guards

adtyavrdhn:
Part of the CI performance series: #7731, #7732, and #7733 are independent and can merge in any order. Draft #7734 → #7735 carry the separate Torch cache proposal.

### What changed

One explicitly batched subprocess now checks all 20 OpenAI-compatible missing-SDK import guards. The child blocks `openai`, imports every provider in deterministic order, and accumulates provider-named failures for missing or incorrect installation hints. The parent verifies both successful exit and an explicit completion sentinel, so clean early termination cannot silently skip later providers.

This deliberately uses one honestly named pytest test. It does not use a shared fixture, JSON IPC, parametrized result facade, or xdist scheduling group.

### Why

The workflow enables coverage subprocess tracing, so the old shape paid interpreter, Pydantic AI, OpenAI SDK, and coverage startup 20 times. A clean coverage-enabled A/B on the same machine measured **56.92s before and 5.88s after** for the original batching experiment. The final focused coverage-enabled test completed in **3.40s**; an earlier CI-shaped `-n logical --dist=loadgroup` invocation completed in **7.19s**.

Coverage JSON comparison for the batching approach found identical executed-line and executed-branch sets across all 23 `pydantic_ai.providers` files. This keeps subprocess coverage enabled; removing `COVERAGE_*` was separately rejected because it loses covered lines and branches.

### Deliberate trade-off

The old test used one fresh interpreter per provider. This test uses one fresh child for the complete import-guard family, so provider imports share `sys.modules`. All current providers fail at their guarded `openai` import before entering their success path. The invariant is documented beside the batch: guarded imports must remain ahead of success-path side effects and cross-provider imports. If pristine interpreter state per provider becomes an intended invariant, this optimization should be reverted rather than hidden behind additional test machinery.

### Adversarial review

Two independent GPT-5.6 SOL reviews covered standards/maintainability and spec/correctness. SOL found that the initial simplified version trusted exit code zero, which could falsely pass after `SystemExit(0)`. The final version adds a completion sentinel, combines stdout/stderr diagnostics, quotes provider identifiers, and documents the shared-child invariant. The sentinel's `SystemExit(0)` negative path was reproduced. Both SOL reviewers approved the final diff without blockers.

### Verification

- 64 tests passed in `tests/providers/test_openai_compatible_http_clients.py`.
- Focused test passed normally and with coverage subprocess tracing.
- A deliberately incorrect Alibaba hint failed with the provider name, expected hint, and actual error.
- A deliberately injected `SystemExit(0)` failed because the completion sentinel was absent.
- Ruff, focused Pyright, and applicable pre-commit hooks passed.
- The final SOL-reviewed five-PR stack passed the complete [GitHub CI workflow](https://github.com/pydantic/pydantic-ai/actions/runs/32809658078), including full coverage aggregation.

This is deliberately test-only. It does not change providers, product behavior, CI configuration, or coverage policy.

No linked issue: isolated CI maintenance following [the current external-runner profile](https://github.com/pydantic/pydantic-ai/actions/runs/32713769565/job/97390593387).

### Checklist

### DIFF
--- a/tests/providers/test_openai_compatible_http_clients.py
+++ b/tests/providers/test_openai_compatible_http_clients.py
@@ -223,11 +223,15 @@ for module, error_hint in {IMPORT_GUARD_CASES!r}:
 
 if failures:
     raise AssertionError("\\n".join(failures))
+
+print("all import guards checked")
 """
     result = subprocess.run([sys.executable, '-c', code], text=True, capture_output=True)
     diagnostics = f'stdout:\n{result.stdout}\nstderr:\n{result.stderr}'
 
     assert result.returncode == 0, diagnostics
+    # The sentinel guards against a clean early exit (e.g. `SystemExit(0)`) skipping later providers.
+    assert 'all import guards checked' in result.stdout, diagnostics
 
 
 @pytest.mark.anyio
--- a/tests/providers/test_openai_compatible_http_clients.py
+++ b/tests/providers/test_openai_compatible_http_clients.py
@@ -194,11 +194,11 @@ IMPORT_GUARD_CASES = [
 
 
 def test_openai_compatible_provider_import_guards() -> None:
-    # One child amortizes interpreter and coverage startup. Keep each provider's guarded `openai` import ahead of
-    # success-path side effects and cross-provider imports so sharing the child does not change the behavior under test.
+    # One child amortizes interpreter and coverage startup; the assertion prevents cases from preloading one another.
     code = f"""
 import builtins
 import importlib
+import sys
 
 original_import = builtins.__import__
 
@@ -211,8 +211,10 @@ builtins.__import__ = import_without_openai
 
 failures = []
 for module, error_hint in {IMPORT_GUARD_CASES!r}:
+    provider_module = f"pydantic_ai.providers.{{module}}"
+    assert provider_module not in sys.modules, f"{{provider_module!r}} was already imported"
     try:
-        importlib.import_module(f"pydantic_ai.providers.{{module}}")
+        importlib.import_module(provider_module)
     except ImportError as exc:
         if error_hint not in str(exc):
             failures.append(f"{{module!r}}: expected {{error_hint!r}} in {{str(exc)!r}}")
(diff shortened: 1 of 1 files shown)


adtyavrdhn:
Full-stack validation is green: [CI run 32727385861](https://github.com/pydantic/pydantic-ai/actions/runs/32727385861). On the same `ubicloud-premium-4` runner class, the [baseline all-extras pytest step](https://github.com/pydantic/pydantic-ai/actions/runs/32706412447/job/97368356980) took 582 s and the [stacked pytest step](https://github.com/pydantic/pydantic-ai/actions/runs/32727385861/job/97431789486) took 333 s. This is a single-run cumulative-stack comparison, so it supports the direction but should not be read as an isolated guaranteed saving from this PR.

adtyavrdhn:
Fresh adversarial review led to a simpler implementation. The fixture, JSON IPC, parametrized facade, and `xdist_group` have been removed; one honestly named subprocess test now checks all providers and accumulates provider-specific failures. The focused coverage-enabled test completed in 4.20s locally, its deliberately broken-hint path named the provider and expected/actual messages, and the rewritten full stack passed [CI run 32808403997](https://github.com/pydantic/pydantic-ai/actions/runs/32808403997), including coverage aggregation.

adtyavrdhn:
Correction: the final adversarial review was rerun with GPT-5.6 SOL on both standards/maintainability and spec/correctness axes. SOL found a real clean-early-exit hole in the first simplified version: exit code zero did not prove all providers ran. The final commit adds a completion sentinel, combined stdout/stderr diagnostics, quoted provider identifiers, and an inline shared-child invariant. A deliberately injected `SystemExit(0)` now fails on the absent sentinel. Both SOL reviewers approved the final diff.

adtyavrdhn:
The final GPT-5.6 SOL-reviewed stack passed [CI run 32809658078](https://github.com/pydantic/pydantic-ai/actions/runs/32809658078): 36 jobs succeeded, four conditional jobs were skipped, and full coverage aggregation passed.

adtyavrdhn:
Updated the final test design in 2e4d760eb. The shared child now asserts that each target provider module is absent from `sys.modules` before importing it, making the non-preloading invariant executable rather than relying on the earlier architectural comment. The completion sentinel has been removed: a zero exit status plus the loop assertions is the simpler contract, and defending against imported provider code deliberately calling `SystemExit(0)` was not worth a separate stdout protocol. This supersedes my earlier comment defending the sentinel. The focused test passes in 1.44s; targeted Ruff, formatting, and diff checks also pass.

PR #7732, opened 2026-08-24
Use a shallow checkout for coverage reporting

adtyavrdhn:
Part of the CI performance series: #7731, #7732, and #7733 are independent and can merge in any order. Draft #7734 → #7735 carry the separate Torch cache proposal.

### What changed

Let `actions/checkout` use its default shallow checkout in the coverage job.

### Why

The full-history override is accompanied by a stale “needed for diff-cover” comment, but the coverage job no longer invokes `diff-cover` or any Git history consumer. In [a current coverage job](https://github.com/pydantic/pydantic-ai/actions/runs/32706412447/job/97371011669), checkout occupied 14 seconds before coverage work began.

`actions/checkout` documents that the default fetches only the triggering commit; `fetch-depth: 0` fetches all branches and tags ([official checkout documentation](https://github.com/actions/checkout#fetch-all-history-for-all-tags-and-branches)).

### Verification and scope

- Inspected every remaining coverage-job command for Git history use.
- YAML and zizmor checks passed.
- Full pre-commit passed.

This does not change coverage collection, reporting, branch protection, tests, or the matrix. It is separate from coverage-process scheduling so either change can be evaluated or reverted alone.

No linked issue: isolated CI maintenance.

### Checklist

- [ ] Any **AI generated code** has been reviewed line-by-line by the human PR author, who stands by it.
- [x] No **breaking changes** in accordance with the [version policy](https://github.com/pydantic/pydantic-ai/blob/main/docs/version-policy.md).
- [x] **PR title** is fit for the release changelog.

PR #7733, opened 2026-08-24
Drop the coverage HTML report and run coverage checks concurrently

adtyavrdhn:
Part of the CI performance series with #7731 and #7732; all three are independent.

### What changed

- Removed the coverage HTML report from CI, as suggested by @Kludex [in this comment](https://github.com/pydantic/pydantic-ai/pull/7733#issuecomment-5406068659): the `coverage html` step, its artifact upload, and the `smokeshow` job in `after-ci.yml` that published it.
- Replaced the coverage badges in the READMEs and docs with a static 100% badge.
- The two remaining commands, `coverage report` and `strict-no-cover`, now run concurrently after `coverage combine`; the step fails if either fails.

### Why

The published report was always empty: `fail_under = 100` means only fully-covered runs upload it, and `skip_covered = true` then hides every file ([example](https://smokeshow.helpmanual.io/306q213u184s5q4s4724/)). The badge was likewise a constant 100.00%, so a static badge says the same thing — CI enforces 100%.

The smokeshow job had to go with the HTML: its artifact download uses `if_no_artifact_found: warn`, so it would have kept running and silently publishing nothing. Logfire dropped its coverage HTML the same way (it uses Codecov for the human-facing view — the natural replacement here if we ever want per-PR line coverage back).

Speed: the coverage job's report steps ran serially at 25s + 32s + 25s ([current job](https://github.com/pydantic/pydantic-ai/actions/runs/32706412447/job/97371011669)). Dropping the HTML removes 25s plus the artifact upload; running the remaining two concurrently cuts 57s to ~32s. The step sets `UV_NO_SYNC=1` rather than `--no-sync` flags because `strict-no-cover` spawns a nested `uv run` that only the environment variable reaches.

The `SMOKESHOW_AUTH_KEY` repository secret becomes unused and can be deleted.

No linked issue: isolated CI maintenance.

### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -575,7 +575,7 @@ jobs:
       - run: uv run coverage combine
       - name: generate coverage reports
         # UV_NO_SYNC rather than `--no-sync` flags: strict-no-cover spawns its own nested
-        # `uv run`, and a sync there could modify the venv under the two concurrent readers.
+        # `uv run`, and a sync there could modify the venv under the concurrent reader.
         env:
           UV_NO_SYNC: '1'
         run: |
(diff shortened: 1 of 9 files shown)


adtyavrdhn:
Observed full-stack result: the [coverage job](https://github.com/pydantic/pydantic-ai/actions/runs/32727385861/job/97433772823) passed in 70 s versus 133 s for the [baseline](https://github.com/pydantic/pydantic-ai/actions/runs/32706412447/job/97371011669). The baseline report/strict/HTML steps occupied 82 s in series; the new concurrent report block occupied 32 s and preserved all gates. The complete [CI run](https://github.com/pydantic/pydantic-ai/actions/runs/32727385861) is green.

Kludex:
Just drop the html part with show context, it will run faster.

Kludex:
Generate .coverage instead of html

.github/workflows/ci.yml
```diff
@@ -573,18 +573,21 @@ jobs:
...
+        run: |
+          uv run coverage report &
+          report_pid=$!
+          COVERAGE_FILE=.coverage/.coverage uv run strict-no-cover &
+          strict_pid=$!
+
+          status=0
+          wait "$report_pid" || status=1
+          wait "$strict_pid" || status=1
+          exit "$status"
```
Kludex, line 590: I don't think you need this 🤔

PR #7734, opened 2026-08-24
Use CPU-only PyTorch for Linux development

adtyavrdhn:
Draft policy proposal in the CI performance series, independent of #7731–#7733 and based on `main`. #7735 is stacked on this branch. Parked as draft pending the policy decision below.

### What changed

Pin `torch` to PyTorch's explicit CPU index only when `sys_platform == 'linux'`, then regenerate the universal lock. Non-Linux platforms continue to resolve PyPI Torch.

### Why

Current Linux all-extras jobs restore a pruned ~546KiB uv cache and download the PyPI Torch/CUDA/Triton graph on every run. Three representative jobs each downloaded about **2.529GiB** of those wheels: [same-repository PR](https://github.com/pydantic/pydantic-ai/actions/runs/32712148640/job/97385659963), [main](https://github.com/pydantic/pydantic-ai/actions/runs/32706412447/job/97368356980), and [external PR](https://github.com/pydantic/pydantic-ai/actions/runs/32713769565/job/97390593387).

The regenerated lock removes 18 NVIDIA/CUDA packages plus Triton and adds `torch==2.13.0+cpu`. The CPython 3.12 x86-64 CPU wheel is 191,817,609 bytes, approximately a 93% reduction from the previous Torch/CUDA/Triton payload. uv documents package-to-index pinning and this exact PyTorch source pattern in its [index documentation](https://docs.astral.sh/uv/concepts/indexes/#pinning-a-package-to-an-index) and [PyTorch guide](https://docs.astral.sh/uv/guides/integration/pytorch/). One ongoing cost to weigh: `download.pytorch.org` can lag PyPI by a few days on new Torch releases, so future `torch` bumps may briefly fail to resolve on Linux until the CPU index catches up.

### Policy decision

@DouweM This changes Linux **checkout development installs** to CPU-only Torch. Published package metadata and downstream consumers are unaffected. Please confirm that GPU-enabled Torch does not need to remain the default for contributors; GPU development can still opt into an explicit environment, but this PR should not merge without that policy decision.

### Verification

- `uv lock --check` passed.
- Linux CPython 3.12 dry-run resolves `torch==2.13.0+cpu` and no CUDA/Triton packages.
- Windows CPython 3.12 dry-run resolves PyPI `torch==2.13.0`.
- `tests/test_embeddings.py`: 127 passed, 2 environment-dependent skips.
- Full pre-commit passed.

The cache-policy change is intentionally absent; it is isolated in the next PR.

[...]

PR #7735, opened 2026-08-24
Retain uv wheels for dependency-heavy CI jobs

adtyavrdhn:
Stacked on #7734 (`codex/ci-linux-cpu-torch`) and must not merge before it. Draft: to be promoted only if warm CI runs after #7734 lands show all-extras jobs still spending meaningful time re-downloading wheels.

### What changed

Make setup-uv retain built wheels only for the dependency-heavy `all-extras` test lane. Other matrix caches, including `lowest-versions`, remain pruned to limit pressure on the 10GB repository cache quota.

### Why

The pinned setup-uv v8.2.0 defaults `prune-cache` to true. setup-uv documents that pruning removes pre-built wheels, causing them to be downloaded again on subsequent runs ([official caching documentation](https://github.com/astral-sh/setup-uv/blob/main/docs/caching.md#enable-cache-pruning)). Current all-extras jobs demonstrate this directly: a ~546KiB exact cache hit is followed by the same Torch/dependency downloads in [main](https://github.com/pydantic/pydantic-ai/actions/runs/32706412447/job/97368356980) and [an external PR](https://github.com/pydantic/pydantic-ai/actions/runs/32713769565/job/97390593387).

This PR is stacked after CPU-only Linux Torch because retaining the old 2.529GiB CUDA graph per cache key would be unsafe. At audit time the repository held **9,625,949,139 bytes across 298 cache entries**, close to GitHub's documented [10GB repository cache limit](https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#usage-limits-and-eviction-policy).

### Safety and measurement plan

- The all-extras caches keep their existing per-lane suffix; no other lane's cache grows.
- Small lint/docs/standard/slim/evals caches keep the current pruning behavior.
- YAML and zizmor checks passed.
- Full pre-commit passed.

This PR should be evaluated with cold then warm CI runs and cache-size telemetry. It can be dropped independently if retention does not improve warm runs or causes quota churn.

No linked issue: isolated CI cache-policy maintenance.

### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -262,6 +262,8 @@ jobs:
           python-version: ${{ matrix.python-version }}
           enable-cache: true
           cache-suffix: ${{ matrix.install.name }}
+          # Retain wheels only for all-extras; keep smaller matrix caches pruned.
+          prune-cache: ${{ matrix.install.name != 'all-extras' }}
 
       - run: mkdir .coverage
 


adtyavrdhn:
Cold-key validation is green in [run 32727385861](https://github.com/pydantic/pydantic-ai/actions/runs/32727385861). The [all-extras job](https://github.com/pydantic/pydantic-ai/actions/runs/32727385861/job/97431789486) saved an unpruned uv cache of 500,727,200 bytes (~478 MiB), after downloading the 182.9 MiB CPU Torch wheel and no CUDA wheels. This validates the sequencing with the CPU-Torch PR; warm-hit and eviction telemetry remain intentional post-merge follow-up.

PR #6258, opened 2026-07-04
Let application code and capabilities emit typed `CustomEvent`s and `CapabilityEvent`s into the run event stream and subscribe with `@on_event`

DouweM:
This PR implements the whole **run event design** — application `CustomEvent`s, capability `CapabilityEvent`s, and the `on_event` subscription surface — as one stack. Closes #2382, supersedes #3114 and #7794 (whose design discussion and description this one carries forward). It is the core-native answer to [the capability-domain-hooks proposal](https://github.com/pydantic/pydantic-ai-notes/blob/mpfaffenberger/capability-domain-hooks/features/capability-domain-hooks/2026-08-20%20proposal%20-%20unified%20domain%20events%20and%20intervention%20hooks%20for%20harness%20capabilities.md), scoped in [this comment](https://github.com/pydantic/pydantic-ai/pull/6258#issuecomment-5411935203) and audited against Code Puppy's full callback inventory in [pydantic-ai-notes#15](https://github.com/pydantic/pydantic-ai-notes/pull/15).

Harness PRs follow on top: `FileSystem` emits read/traversal events with `RepoContext` subscribing (replacing its `traversal_tool_names` tool-name sniffing), compaction lifecycle events with a cancellable `CompactionStartEvent`, and deprecation of the ad-hoc `on_spend`/`on_usage`/`on_fire` constructor callbacks. A stacked core PR adds the compaction event family to core itself (emitted by `OpenAICompaction`/`AnthropicCompaction` as well as the Harness strategies).

## Two event families

Both families are **typed dataclass subclasses only** — there is no name + untyped-payload shorthand. A quick event is a two-line class; forcing the class keeps every event `isinstance`-matchable, serializable, and self-documenting:

- **`CustomEvent`** — *application-owned* events, emitted from tools, history processors, output validators (`await ctx.emit(event)`) or code driving `agent.iter()` (`AgentRun.emit`). Forwarded to UI frontends.
- **`CapabilityEvent`** — events *capabilities* emit, namespaced and stamped with the emitting instance's run id. **Not** forwarded to frontends by default.

```python
@dataclass(kw_only=True)
class SyncProgressEvent(CustomEvent):
    done: int
    total: int

await ctx.emit(SyncProgressEvent(done=3, total=9))
```

A capability family declares a namespace; the wire identity `kind` is `'{namespace}.{snake_case(class name)}'`, so a bare, collision-prone `'file_read'` is unrepresentable:

```python
FILE_SYSTEM_EVENTS = 'file_system'

[...]

PR #7794, opened 2026-08-26
Let capabilities emit typed `CapabilityEvent`s and subscribe to run events with `@on_event`

DouweM:
Stacked on #6258. Together with it and one more PR on top, this implements the design for **capability events**: typed events that capabilities emit and that other capabilities, application code, and instrumentation can subscribe to — the core-native answer to [the capability-domain-hooks proposal](https://github.com/pydantic/pydantic-ai-notes/blob/mpfaffenberger/capability-domain-hooks/features/capability-domain-hooks/2026-08-20%20proposal%20-%20unified%20domain%20events%20and%20intervention%20hooks%20for%20harness%20capabilities.md), scoped in [this comment](https://github.com/pydantic/pydantic-ai/pull/6258#issuecomment-5411935203). This description doubles as an introduction to the whole design; the pieces land as a stack:

1. **#6258** — `CustomEvent`: events emitted by *application* code (name + payload, or typed subclasses), forwarded to UI frontends.
2. **This PR** — the capability side, whole: `CapabilityEvent` (events emitted by *capabilities*, namespaced and stamped with the emitting instance's id, *not* forwarded to frontends by default) plus the subscription surface — an `on_event` capability hook, an `@on_event` method marker, `hooks.on.event` app parity, and dispatch semantics including inline-dispatched "decision" events.
3. **Harness PRs** — `FileSystem` emits read/traversal events, `RepoContext` subscribes (replacing its `traversal_tool_names` tool-name sniffing), compaction lifecycle events with a cancellable `BeforeCompactionEvent`, and deprecation of the ad-hoc `on_spend`/`on_usage`/`on_fire` constructor callbacks in favor of this mechanism.

## Defining and emitting events

A capability family declares a namespace constant and defines its events as dataclass subclasses whose fields are the payload:

```python
# pydantic_ai_harness/filesystem/_events.py
FILE_SYSTEM_EVENTS = 'file_system'

@dataclass(kw_only=True)
class FileReadEvent(CapabilityEvent, namespace=FILE_SYSTEM_EVENTS):
    path: str          # the normalized, authorization-checked path
    content_hash: str
```

```python
# inside the capability's hook or contributed tool
await ctx.emit_event(FileReadEvent(path=normalized, content_hash=digest))
```

- The wire identity `kind` is `'{namespace}.{snake_case(class name)}'` (`'file_system.file_read'`; name overridable via `name='...'`), so a bare, collision-prone `'file_read'` is unrepresentable, and duplicates are rejected at class definition. Registration is automatic via `__init_subclass__` — one class statement is the entire authoring surface.
- **`capability_id` is stamped at emit time** with the emitting *instance*'s run id: resolved from the per-dispatch `RunContext` copy for hooks, and through the contributing toolset's id for capability-contributed tools. The union tag stays class-stable while instances remain distinguishable — two MCP capability instances share `kind='mcp.tools_listed'` but carry `capability_id='docs_server'` / `'internal'`. `tool_call_id`/`tool_name` are stamped like custom events.
- Events surface on the run's event stream (`event_stream_handler`, `run_stream_events()`, `iter()` streaming) alongside the framework's own events.

## Subscribing

Capabilities subscribe with a method marker; filtering is by explicit event type (isinstance semantics, statically checked against the parameter annotation — no casts, no name-string matching):

```python
class RepoContext(AbstractCapability[Any]):
    @on_event(FileReadEvent, DirectoryListedEvent)
    async def _on_traversal(self, ctx: RunContext[Any], event: FileReadEvent | DirectoryListedEvent) -> None:
        if note := self._context_file_note(event.path):
            ctx.enqueue(note)  # lands in history before the next model request
```

This replaces the tool-name sniffing `RepoContext` does today (`traversal_tool_names` + guessing the args dict + re-deriving path validation): it reacts to `FileSystem`'s *processed* data — the normalized, authorized path — without coupling to tool names.

- `on_event` receives **all** events: framework events (tool calls/results, deferred-tool batches, enqueued-message deliveries, streaming parts), custom events, and capability events. A bare `@on_event` subscribes to everything (the parameter must then be typed as the full `AgentStreamEvent` union). Capabilities with no subscriptions cost nothing.
- Dispatch order follows capability order; listeners are async and observe-only — they react via side effects (`ctx.enqueue`, emitting further events, mutating the event, see below).
- App code gets the same surface through the `Hooks` capability: `@hooks.on.event(FileReadEvent)` (the existing transform semantics of `hooks.on.event` are deprecated toward `hooks.on.run_event_stream`).

**Dispatch timing.** By default, listeners run at the event's *stream position*: an event enters the run's event buffer at emit time (so stream ordering can never invert — an event emitted at the end of a tool body always precedes that tool's `FunctionToolResultEvent`), and listeners are invoked as events flow by, in stream order, before the next model request at the latest. Emitters aren't slowed down by their listeners.

**Decision events.** An event type that expects answers declares inline dispatch on the class, next to its decision fields — then `await ctx.emit_event(...)` runs the listener chain before returning, and the emitter reads the outcome off the event (the DOM `preventDefault()` model):

```python
@dataclass(kw_only=True)
class BeforeCompactionEvent(CapabilityEvent, namespace=COMPACTION_EVENTS, dispatch='inline'):
    strategy: str
    estimated_tokens: int
    cancelled: bool = False
    cancel_reason: str | None = None

    def cancel(self, reason: str | None = None) -> None:
        self.cancelled = True
        self.cancel_reason = reason

# emitter (a compaction strategy):
event = await ctx.emit_event(BeforeCompactionEvent(strategy='summarizing', estimated_tokens=41_000))
if event.cancelled:
    return  # skip compaction this turn

# listener (e.g. Planning, mid-activity):
@on_event(BeforeCompactionEvent)
async def _hold_compaction(self, ctx: RunContext[Any], event: BeforeCompactionEvent) -> None:
    if self._plan_step_in_flight:
        event.cancel('a plan step is mid-flight')
```

Because the class carries the declaration, an emitter can't forget to wait and read stale fields; the decided state also serializes to the stream, so consumers see the outcome. This subsumes the proposal's `before_compaction` guard — and generalizes it: any capability can define an intervention point without core growing a new hook, with `GuardrailResult`-style vocabularies expressible as fields on the event.

## Guardrails between the two families

The custom/capability split is enforced, not conventional:

- A `CapabilityEvent` with no resolvable owning capability (a plain `@agent.tool`, or `AgentRun.emit_event`) raises `UserError` — app code emits `CustomEvent`s.
- A `CustomEvent` emitted from inside a capability's hook dispatch raises `UserError` — except for app-facing callback capabilities (`Hooks`, `ProcessEventStream`), which run *user* callbacks. That keeps the sanctioned frontend bridge working: capability events do **not** cross to UI frontends by default (the new `UIEventStream.handle_capability_event` default emits nothing — internal coordination signals shouldn't leak internal metadata to browsers), and the way to surface one is an application subscriber that re-emits it as a `CustomEvent`, deliberately choosing the payload.

## Serialization

Same registry + schema-time union-rebuild mechanism as `CustomEvent` (shared `_event_registry.py`, patterned on the native-tools union): `event_kind='capability'` routes the closed `AgentStreamEvent` union to the family, `kind` selects the concrete class, and the flat wire dict revalidates to the typed subclass. An unregistered `kind` — a process that hasn't imported the defining package, e.g. a Temporal worker — degrades to `UnknownCapabilityEvent` with the payload preserved in `data` (re-flattened on re-serialization, so a downstream consumer that has the module imported recovers the typed event) instead of failing validation.

## Relation to the capability-domain-hooks proposal

- `HarnessEvent`/`HarnessEventEmitter` → core `CapabilityEvent` + the run event stream; no per-capability emitter object, no `on_event=` constructor params (the existing `on_spend`/`on_usage`/`on_fire` params get deprecation shims pointing at subscriptions).
- Tier-2 interventions (`GuardrailResult` reuse, `before_compaction`) → inline-dispatched decision events; guardrails remain the right tool where they exist today.
- The proposal's "core ask" (a `CustomEvent` bridge member in `AgentStreamEvent`) → #6258, extended with typed subclasses.
- Logfire angle: one typed stream to instrument; a stock adapter can mirror capability events to spans without per-capability glue.

## Checklist

- [x] Tests (`tests/test_capability_events.py`, `tests/test_on_event.py`, adapter suppression tests, Temporal field-exclusion), docs (`docs/capabilities/overview.md` "Capability events" + "Reacting to events", `docs/hooks.md`, `docs/agent.md` cross-link)
- [ ] AI generated code (checked by the author in the UI)


mpfaffenberger (commented):
Hey! This is really exciting. I audited the lifecycle hooks from Code Puppy against this stack and the proposed `on_event` follow-up. It looks like the full agent-run domain is covered, with the remaining hooks belonging to the application shell rather than capabilities.

I wrote up the complete audit here, including what I think this means for existing Harness capabilities: https://github.com/pydantic/pydantic-ai-notes/pull/15

The nice surprise is that this does not look like a broad Harness rewrite. Most of the migration appears limited to typed event emission at existing commit points, plus replacing a few ad hoc callbacks and cross-capability tool-name couplings.

DouweM:
Consolidated into #6258: the custom-event side and this capability side kept reshaping each other during review, so they now land as one PR (same branch history, description carried over there). Design changes since this PR was opened, all reflected in #6258:

- **Custom events are typed-only.** The `emit_event('name', data)` shorthand and the untyped name+payload event are gone; a quick event is a two-line dataclass. `data` now exists only on the `Unknown*Event` degradation envelopes, and `emit_event` takes exactly one constructed event.
- **Live delivery.** Emitted events now reach stream consumers while the emitting tool is still running, instead of surfacing with the tool's result — this was the flagship progress-reporting use case and it didn't actually work before (caught by @macroscopeapp's review here, thanks).
- **Naming convention:** events use `Start`/`End` (`CompactionStartEvent`), `before_`/`after_` stays hook vocabulary.
- **The compaction event family will live in core**, next to `OpenAICompaction`/`AnthropicCompaction`, with the Harness strategies emitting the same family — stacked PR to follow.

@mpfaffenberger thanks for the [lifecycle coverage audit](https://github.com/pydantic/pydantic-ai-notes/pull/15) — the empty "infeasible" category and the four-hotspot migration scope match what we're building against, and #6258's description now links it. Your test-case list (emission-before-errors, multi-instance ids, observer exception policy) is going into the follow-up work.

Macroscope findings from this PR, addressed in #6258 (commit 198f80a02): implicit-id capability tools can now emit (resolved via the `capability_id` stamped on tool definitions); `@on_event` markers named `on_event` are rejected; `CombinedCapability` subclasses' own listeners dispatch; legacy `hooks.on.event` replacements compose; dotted namespaces derive with `rpartition`; envelope-shadowing payload fields, per-instance `kind` overrides, and non-dataclass subclasses are rejected; `@dataclass(slots=True)` works; `_event_stream_replacements` joined the Temporal/Prefect field audits. Two declined: pre-set `capability_id` stays respected (deliberate, for re-emitting on another instance's behalf — now documented on the field), and Temporal's `emit_event` rejection stays for event-stream handlers too, since they also run inside activities where silently dropping would be worse (error message now says so).

PR #7873, opened 2026-08-28
Batch event-stream dispatch into per-burst durable units

### LINKED ISSUE #7631: Durable execution: allow batching workflow-side event dispatch
Under `durable_exec.temporal`, `_call_event_stream_handler_activity` dispatches one activity per workflow-side event and awaits each in turn.

Because `CallToolsNode._run_stream` yields the `FunctionToolCallEvent`s before `_call_tools` starts any task, the tools of a model response cannot begin until every call event's activity has completed, and the next `ModelRequestNode` cannot begin until every result event's has.

Measured on a chat workload: ~0.15 s and ~25 KB per event, 347 such activities across 33 workflows. A step with four tool calls pays ~0.7 s of dead time on each side.

**Request:** a way to hand the durable boundary a *burst* of events rather than one event — e.g. an opt-in on `TemporalAgent` that collects the ready events and calls the handler activity once with the list. Ordering and per-event semantics are preserved; only the number of round trips changes.

**Context:** the split looks deliberate and is unchanged on v2 (`durable_exec/_base.py`, `_dispatch_event_stream_event` / `_single_event_stream`), and #5477 keeps it. `ProcessEventStream` is not an alternative here — its hook runs in workflow code and must be deterministic, so it cannot persist.

Working around it means subclassing `TemporalAgent` and overriding the `event_stream_handler` property at a pinned version, which is what we have done. Note that draining the stream into a list before dispatching is *not* a valid workaround: the call and result events arrive on one generator with the tools running inside it, so draining would withhold every `tool_call` until all tools had finished. Collecting concurrently and flushing what is ready preserves liveness, but it depends on library internals.

Happy to contribute a PR if the shape is welcome.

DouweM:
Closes #7631.

Under durable execution, every event delivered to an `event_stream_handler` crossed the durable boundary as its own unit (activity/step/task), awaited serially: measured ~0.15s and ~25KB of history per event, with tools unable to start until every call-event unit round-tripped (#7631, thanks @tarik-gtlaw for the measurements and the liveness analysis).

Since #6696 event dispatch is one declared operation owned by the base, so this fixes it once for all engines:

- **Batching by default, no user opt-in**: the base collects ready events and dispatches one durable unit per burst; the user handler is still invoked once per event, in order, inside the unit. A pending-dispatch-plus-accumulator loop gives natural batching under backpressure with no timers, and tool execution proceeds while call-event dispatch is in flight (the stream is never drained up front, per the liveness constraint in the issue).
- **Both wire shapes forever**: new recordings carry lists; decode accepts legacy single-event payloads indefinitely. Persisted operation names and compat rows unchanged.
- **Temporal in-flight histories**: batching changes the workflow command sequence, so dispatch is guarded with `workflow.patched('pydantic-ai-batched-event-dispatch')`: unpatched (pre-change) histories replay per-event; new runs batch. Replay coverage for both.
- **Prefect**: replay-stable sequence counters now count batches; a run recorded pre-change and recovered post-change re-delivers events once at the recovery boundary (documented).

Local durable suite: green modulo the known local-only BlockBuster/Prefect cache-write failures.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7873?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


DouweM:
Note for anyone reading the Macroscope approvability block above: its "Not approved" verdict cites "the unresolved High finding about unbounded buffering during slow dispatch", but that finding is no longer open. Macroscope's own inline comment on `durable_exec/_base.py:597` now carries "No longer relevant as of `ca49001dc`".

That High and the two red `test durable-exec` failures were the same defect seen from two sides: `produce()` ran ahead of the consumer, so the shared stream accumulator advanced underneath events the caller had not seen yet, which both grew the queues without bound and made earlier events observe accumulated text (`'Streamed responseed response'`). The fix restores backpressure: the producer now waits for the consumer to yield each replayed model event before pulling the next one off the model stream.

CI is green, and the dispatcher failure paths that were previously uncovered (handler failure mid-batch, producer failure, the spurious-wake path, and the DBOS non-legacy dispatch line) now have real tests rather than pragmas. The verdict block is stale, not a standing objection.

PR #7890, opened 2026-08-29
List `glm-5.3-flash` among the Z.AI thinking models

ayaangazali:
#7887 added `glm-5.3-flash` to `_known_model_names.py`, but the thinking-mode list in `docs/models/zai.md` still stops at `glm-5.3`, so the new model reads as if it has no thinking support.

It does. `profiles/zai.py` matches on prefixes, not exact ids: `_REASONING_EFFORT_MODEL_PREFIXES = ('glm-5.3', 'glm-5.2')` at `:23`, `_ALWAYS_THINKING_MODEL_PREFIXES = ('glm-5.3',)` at `:31`, and the `startswith('glm-5.3')` branch at `:67`. So `glm-5.3-flash` picks up the whole GLM-5.3 profile, which is what the comment on `tests/models/test_zai.py:340` says too, and #7887 recorded `test_zai_reasoning_effort[glm-5.3-flash].yaml` against it.

The page enumerates specific ids rather than families (it separately calls out the `glm-4.6v` and `glm-4.5v` vision variants), and when #7600 added GLM-5.3 it added `glm-5.3` to this same sentence. This just does the same for the flash variant.

Scope kept to that one list. The paragraph below it already speaks in family terms ("on GLM-5.2 and GLM-5.3"), which covers the flash variant as written, so I left it alone.

Ran `uv run pytest tests/test_examples.py -k zai` (6 passed).

### Checklist

---

About me: I'm a freshman in college, so rather than assume the omission was a mistake I checked how the profile actually resolves and how the same list was handled when GLM-5.3 landed. I used Claude Code to double-check the reasoning and read the diff myself. If you would rather the list stay family-level and drop the variants entirely, that is a fair call and I will close this.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7890?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/models/zai.md
+++ b/docs/models/zai.md
@@ -44,7 +44,7 @@ agent = Agent(model)
 
 ## Thinking mode
 
-Z.AI's `glm-5.3`, `glm-5.2`, `glm-5.1`, `glm-5`, `glm-4.7`, `glm-4.6` (hybrid thinking), and `glm-4.5` (interleaved thinking) models support thinking/reasoning mode, where the model produces reasoning content before the final response. This includes the `glm-4.6v` and `glm-4.5v` vision models. Configure this through the unified [`thinking`][pydantic_ai.settings.ModelSettings.thinking] setting:
+Z.AI's `glm-5.3`, `glm-5.3-flash`, `glm-5.2`, `glm-5.1`, `glm-5`, `glm-4.7`, `glm-4.6` (hybrid thinking), and `glm-4.5` (interleaved thinking) models support thinking/reasoning mode, where the model produces reasoning content before the final response. This includes the `glm-4.6v` and `glm-4.5v` vision models. Configure this through the unified [`thinking`][pydantic_ai.settings.ModelSettings.thinking] setting:
 
 ```python
 from pydantic_ai import Agent


ayaangazali:
Re-checked against current main (c0e4d824e) now that this has been open a week.

Still applies. The thinking-mode sentence in `docs/models/zai.md:47` still runs `glm-5.3`, `glm-5.2`, `glm-5.1`, `glm-5`, `glm-4.7`, `glm-4.6`, `glm-4.5` with no `glm-5.3-flash`, and `profiles/zai.py:23` and `:31` still match on the `glm-5.3` prefix, so the flash variant still picks up always-on thinking and reasoning-effort support. It is still in `_known_model_names.py`.

Clean against main and CI is green.

One thing I noticed since opening this that may be relevant to how you want it worded: the comment at `profiles/zai.py:34` says those prefix tuples should "list concrete released ids only", which came in with #7600. If the same instinct applies to the docs sentence, then naming the flash variant explicitly is the consistent choice. If you would rather the docs list stayed family-level and dropped the variants, that is a fair call and I will close this.

PR #7907, opened 2026-08-30
fix(models): keep PartEndEvent (index, part) pairs consistent with their starts

### LINKED ISSUE #7906: Streamed PartEndEvent carries the wrong part (or IndexError) after a nameless tool-call delta
## Description

When a streaming response contains a tool-call delta **without a function name** (OpenAI-compatible gateways send these — the same shape defended against for empty `choices` per #5165) followed by another part, the streamed `PartEndEvent`s are wrong, and the stream can crash at the end.

Root cause: `PartStartEvent.index` comes from the parts manager's **unfiltered** `_parts` space, but the shared `part_end_event` helper in `models/__init__.py` looked the part up via `get_parts()[index]` — the **filtered** list. A nameless tool-call delta creates a dangling `ToolCallPartDelta` that is excluded from the filtered list, so every later index is off by the number of dangling deltas.

Consequences:
- a later part's `PartEndEvent` carries the **wrong part object** (e.g. the text part's end event carries the earlier tool call) — downstream consumers that merge by index (run_stream_events, AG-UI/Vercel adapters, event replay) corrupt the part;
- at end of stream, `get_parts()[index]` raises `IndexError: list index out of range`, aborting the whole run.

## Reproduction

Stream these chunks (Chat Completions shape) to any agent:

1. `tool_calls=[{index: 0, id: 'call_0', function: {name: 'get_location', arguments: '{}'}}]`
2. `tool_calls=[{index: 1, function: {arguments: '{"city"'}}]` — no name → dangling delta
3. `content='Hello there'`
4. finish

Observed: the text part starts at `index=2` but its end event carries the `ToolCallPart`; with one more trailing part the stream dies with `IndexError`.

Expected: every end event carries the same part its start event announced; dangling deltas are simply not ended.

## Environment

- pydantic-ai: current `main`

simpleqt:
Fixes #7906

## Summary

`PartStartEvent.index` comes from the parts manager's **unfiltered** `_parts` space (`_append_part` returns `len(self._parts)`), but the shared `part_end_event` helper in `models/__init__.py` resolved the part via `get_parts()[index]` — the **filtered** list, which excludes dangling `ToolCallPartDelta`s. One nameless tool-call delta (the shape OpenAI-compatible gateways send; the empty-`choices` twin of this is already defended per #5165) shifts every later index:

- a later part's `PartEndEvent` carried the **wrong part object** (e.g. the text part's end event carried the earlier tool call), corrupting index-merging consumers (`run_stream_events`, AG-UI adapters, event replay);
- at end of stream the lookup raised `IndexError: list index out of range`, aborting the run.

Reachable from the Chat Completions path (args delta before a name arrives), the Responses path (`output_item.added` ordering), and any adapter that permits nameless tool-call deltas. Details and a chunk-by-chunk reproduction in the linked issue.

## Change

`PartsManager.get_part(index)` resolves in the **unfiltered** space (materializing string buffers exactly like `get_parts` does) and returns `None` for indices that are still dangling deltas; `part_end_event` uses it, so every end event carries the same part its start event announced and dangling deltas simply get no end event.

## Tests

Regression test in `tests/models/test_openai.py`: tool call + nameless delta + text + a final turn, asserting per-turn that each end event's part type matches its start's, the text part's end carries the `TextPart` itself, and the stream completes (no `IndexError`). Full affected suites: `test_openai.py` + `test_parts_manager.py` + `test_streaming.py` → 474 passed, 3 xfailed.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7907?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/_parts_manager.py
+++ b/pydantic_ai_slim/pydantic_ai/_parts_manager.py
@@ -136,15 +136,15 @@ class ModelResponsePartsManager:
         return [p for p in self._parts if not isinstance(p, ToolCallPartDelta)]
 
     def get_part(self, index: int) -> ManagedPart | None:
-        """Return the part at the unfiltered stream ``index``.
+        """Return the part at the unfiltered stream `index`.
 
-        Returns None while the part is still a dangling ``ToolCallPartDelta`` (a tool-call
+        Returns None while the part is still a dangling `ToolCallPartDelta` (a tool-call
         delta that never received a name and so was never promoted to a real part).
 
-        ``PartStartEvent.index`` values come from the unfiltered ``_parts`` space (see
-        :meth:`_append_part`), so a lookup by that index must NOT go through :meth:`get_parts`:
+        `PartStartEvent.index` values come from the unfiltered `_parts` space (see
+        `_append_part`), so a lookup by that index must NOT go through `get_parts`:
         the filtered list is shorter whenever a dangling delta exists, which returns
-        the wrong part for the index or raises ``IndexError`` past the end of the
+        the wrong part for the index or raises `IndexError` past the end of the
         filtered list.
         """
         if index < 0 or index >= len(self._parts):
--- a/pydantic_ai_slim/pydantic_ai/_parts_manager.py
+++ b/pydantic_ai_slim/pydantic_ai/_parts_manager.py
@@ -136,14 +136,16 @@ class ModelResponsePartsManager:
         return [p for p in self._parts if not isinstance(p, ToolCallPartDelta)]
 
     def get_part(self, index: int) -> ManagedPart | None:
-        """Return the part at the unfiltered stream ``index``, or None while it is a
-        dangling ``ToolCallPartDelta`` (a tool-call delta that never received a name and so was
-        never promoted to a real part).
+        """Return the part at the unfiltered stream ``index``.
+
+        Returns None while the part is still a dangling ``ToolCallPartDelta`` (a tool-call
+        delta that never received a name and so was never promoted to a real part).
 
         ``PartStartEvent.index`` values come from the unfiltered ``_parts`` space (see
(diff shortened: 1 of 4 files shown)


simpleqt:
Two checks on this PR's CI run are infra ghosts rather than content failures: `check` and `coverage` both failed within ~1s of starting (timestamps 11:08-11:09), while the same run's `quality checks` (lint + pyright + the full pre-commit suite) and all 50+ test lanes are green on the same head `e3e1447` — and the touched suites pass locally (231 tests). I don't have rerun rights on this repo; could someone with access re-run the failed checks?

simpleqt:
Correction to my comment above: the `coverage` failure was **not** an infra ghost — it was a real gap in this change. The repo enforces `fail-under=100.00`, and the report showed `pydantic_ai_slim/pydantic_ai/_parts_manager.py` at 99.16% with lines 151 and 155 (the two `return None` branches of `get_part`: out-of-range index, and a dangling `ToolCallPartDelta` at the requested index) never executed. The `check` job failing in ~2s was just the all-green aggregator reflecting that same coverage failure, not an independent issue — I misread both.

Fixed in b985cd5: `test_get_part_resolves_in_the_unfiltered_index_space` in `tests/test_parts_manager.py` exercises `get_part`'s public contract directly — a dangling delta at its unfiltered index returns `None`, a real part at an index past the dangling one resolves to itself, and out-of-range indexes return `None` instead of raising `IndexError` (the failure mode the filtered `get_parts()[index]` lookup this method replaced could hit). Full `tests/test_parts_manager.py` (58 tests) passes locally; ruff check/format clean.

PR #7915, opened 2026-08-30
fix(openai): keep the cumulative usage snapshot across usage-less chunks

### LINKED ISSUE #7913: continuous_usage_stats: usage-less stream chunk wipes the cumulative usage to zero
## Description

With `openai_continuous_usage_stats=True`, a streaming chunk that carries **no usage** wipes the running cumulative usage to zero.

`OpenAIStreamedResponse` folds each chunk's usage by *replacing* the running total (correct for cumulative stats), but `_map_usage` maps a `usage=None` chunk to an all-zero `RequestUsage` — and the replacement is unconditional:

```python
chunk_usage = self._map_usage(chunk)
if self._model_settings.get('openai_continuous_usage_stats'):
    self._usage = chunk_usage   # ← zero on usage-less chunks
```

Usage-less chunks are normal: the first chunk of some streams, moderation chunks, and gateways that only implement `include_usage`.

The zeroing is masked when the stream finishes (the final chunk restores the true total), but a stream **interrupted before the final chunk** (error or cancellation) commits `partial_response.usage = 0` — losing the tokens actually consumed so far, which is exactly the information a partial-usage report needs.

## Reproduction

Stream with `openai_continuous_usage_stats=True`:

1. chunk with `usage={prompt_tokens: 10, completion_tokens: 5}`
2. chunk with `usage=None`
3. (stream interrupted — error/cancel)

`partial_response.usage` reports `input_tokens=0, output_tokens=0` instead of `10/5`.

## Expected behavior

A chunk without usage leaves the running cumulative snapshot unchanged; only chunks that actually carry usage replace it.

## Environment

- pydantic-ai: current `main`

simpleqt:
Fixes #7913

## Summary

With `openai_continuous_usage_stats=True`, each chunk's cumulative usage *replaces* the running total — but a chunk with `usage=None` maps to an all-zero `RequestUsage`, and the replacement was unconditional, wiping the snapshot. Usage-less chunks are normal (first chunk, moderation chunks, gateways that only implement `include_usage`), and the zeroing surfaces whenever a stream is interrupted before the final chunk (error/cancellation): `partial_response.usage` then reports 0/0 instead of the tokens consumed so far.

## Change

The fold now replaces only when the chunk actually carries usage (`chunk.usage is not None`); usage-less chunks leave the snapshot untouched. Extracted into `_update_usage(chunk)` so `_get_event_iterator` stays under the repo's ruff C901 budget. The default (non-continuous) path is unchanged — adding a zero there was already a no-op.

## Tests

Regression test with a usage-less chunk between two cumulative chunks: the per-step usage snapshot shows the 10/5 cumulative value held across the usage-less chunk (never zero) until the next cumulative chunk replaces it. Suite: `tests/models/test_openai.py` → 231 passed.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7915?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


simpleqt:
The two failing checks on the current head run are not from this change: `test on 3.13 (all-extras)` fails in `tests/evals/test_otel.py::test_span_query_stop_recursing_when_with_multiple_conditions` (IndexError — a file this PR does not touch, and the test skips locally where its OTel deps are absent), and `check`/the durable-exec job are infra flakiness seen across other PRs today. The quality-checks (lint + pyright) lane is green at this head. Happy to rebase if main's test_otel issue lands a fix first.

simpleqt:
Rebased onto main in db71dad (merge) + 9b42b36 — #7916's terminal finish_reason work landed in the same region. The only collision was its `# noqa: C901` on the chat `_get_event_iterator`, which is unused on this branch (the `_update_usage` extraction keeps that iterator under the complexity threshold), so it's dropped. All 235 `tests/models/test_openai.py` tests pass; ruff/format/pyright clean.

PR #7919, opened 2026-08-30
fix(bedrock): omit, don't rewrite, an unsigned trailing thinking block on Anthropic

### LINKED ISSUE #7908: Bedrock: a thinking-only response is re-sent with its thinking block silently downgraded to `<thinking>` text, and the API rejects the retry
### What happens

On Bedrock + Anthropic with extended thinking, a **thinking-only** model response kills the run with a permanent 400:

```
ModelHTTPError 400: messages.1.content.1: thinking or redacted_thinking blocks in
the latest assistant message cannot be modified. These blocks must remain as they
were in the original response.
```

We saw this in production. The user re-sent the identical prompt 42 seconds later and got a byte-identical failure, so it is deterministic rather than transient — the same prompt is simply unanswerable.

### Why

Two pieces that are each defensible on their own.

**1.** Extended thinking cannot be turned off on Sonnet/Opus-class models, so a turn can legally come back thinking-only (no text, no tool call). `_agent_graph` routes that into the single output-validation retry (`is_thinking_only` → `_build_retry_node`, `_agent_graph.py:1947` / `:1829` on `main`), which appends one `RetryPromptPart` and **re-requests with the just-produced assistant message still in history**.

**2.** On that re-request, `BedrockConverseModel._map_messages` (`models/bedrock.py:1306-1327` on `main`) emits a real `reasoningContent` block only when all of

* `item.provider_name == self.system`,
* `item.signature` is truthy, and
* the profile sets `bedrock_send_back_thinking_parts`

hold. Otherwise it falls into the `else` and **silently rewrites the thinking block as a plain text block wrapped in `<thinking>…</thinking>`** — no warning, no log, no error.

For any other position in the history that fallback is harmless. For the **latest assistant message** it is precisely the modification the Anthropic-on-Bedrock API refuses, so the fallback converts a recoverable situation into an unrecoverable one.

A thinking block reaches that `else` branch easily: `BedrockStreamedResponse` passes `provider_name=self.provider_name if signature else None` (`models/bedrock.py`, the `contentBlockDelta`/`reasoningContent` branch), so a block whose signature delta never arrives ends the stream with `signature=None, provider_name=None`.

### Reproduction

No credentials and no network — the boto3 client is a `MagicMock`, and only `_map_messages` (the last step before the Converse call) is driven. The thinking parts are built by the real `ModelResponsePartsManager` using the exact call shape `BedrockStreamedResponse` uses.

```python
import asyncio
from unittest.mock import MagicMock

import pydantic_ai
from pydantic_ai._parts_manager import ModelResponsePartsManager
    ModelRequest,
    ModelResponse,
    RetryPromptPart,
    UserPromptPart,
)

MODEL_ID = "us.anthropic.claude-sonnet-4-5-20250929-v1:0"


def build_model() -> BedrockConverseModel:
    return BedrockConverseModel(
        MODEL_ID, provider=BedrockProvider(bedrock_client=MagicMock())
    )


def streamed_thinking(model, *, deliver_signature: bool):
    """Parts exactly as BedrockStreamedResponse builds them."""
    mgr = ModelResponsePartsManager(ModelRequestParameters())
    list(
        mgr.handle_thinking_delta(
            vendor_part_id=0,
            content="weighing it up",
            signature=None,
            provider_name=None,
        )
    )
    if deliver_signature:
        list(
            mgr.handle_thinking_delta(
                vendor_part_id=0,
                signature="ErUBCkYIBRgCKkD1qX9m",
                provider_name=model.system,
            )
        )
    return list(mgr.get_parts())


async def main() -> None:
    model = build_model()
    print("pydantic-ai", pydantic_ai.__version__)
    print(
        "profile bedrock_send_back_thinking_parts:",
        model.profile.get("bedrock_send_back_thinking_parts"),
    )

    for label, deliver in (("signature delivered", True), ("no signature delta", False)):
        parts = streamed_thinking(model, deliver_signature=deliver)
        history = [
            ModelRequest(parts=[UserPromptPart("hi")]),
            ModelResponse(parts=parts),  # thinking-only -> the output retry
            ModelRequest(parts=[RetryPromptPart(content="Please return text.")]),
        ]
        _, wire = await model._map_messages(
            history, ModelRequestParameters(), dict(model.settings or {})
        )
        assistant = [m for m in wire if m["role"] == "assistant"]
        blocks = [[next(iter(b)) for b in m["content"]] for m in assistant]
        print(f"{label:22} part={parts[0]!r}")
        print(f"{'':22} assistant blocks={blocks}")
        if assistant and "text" in assistant[0]["content"][0]:
            print(f"{'':22} -> {assistant[0]['content'][0]['text']!r}")


asyncio.run(main())
```

On `pydantic-ai-slim[bedrock]` **2.36.0**:

```
pydantic-ai 2.36.0
profile bedrock_send_back_thinking_parts: True
signature delivered    part=ThinkingPart(content='weighing it up', signature='ErUBCkYIBRgCKkD1qX9m', provider_name='bedrock')
                       assistant blocks=[['reasoningContent']]
no signature delta     part=ThinkingPart(content='weighing it up')
                       assistant blocks=[['text']]
                       -> '<thinking>\nweighing it up\n</thinking>'
```

The second row is the request the API rejects. Adding a second, signed block in front of the unsigned one reproduces the exact error position we saw in production (`content.1`): the wire becomes `['reasoningContent', 'text']`.

### Consequence

Nothing fails at the point where the block loses its signature. The failure surfaces one request later, as a 400 on a history the caller cannot repair: **removing the offending block is itself a modification**, and the signature is opaque so it cannot be reconstructed. The run ends with no output at all.

This is adjacent to #7597 / #7601, which stopped an empty-string signature reaching Anthropic. That fix routes such a block into the "don't send it" branch — which on Bedrock *is* the `<thinking>` text fallback, so on this path the block still goes on the wire, just modified.

### Possible directions

Not sure which you'd prefer, happy to open a PR for whichever:

1. **Don't apply the tagged-text fallback to the last assistant message on a provider that enforces thinking-block integrity.** The fallback is a sensible way to preserve thinking as context for older/foreign histories, but for the trailing assistant message it is the one thing that is guaranteed to be rejected.
2. **Don't re-send the offending response at all on the output-validation retry.** `_build_retry_node` could drop a thinking-only response whose blocks can't round-trip, so the retry re-asks the original prompt. That is what we ended up doing locally, as a `ProcessHistory` capability — it costs one extra generation and returns a real answer instead of a dead turn.
3. **At minimum, make the downgrade observable.** A single `warning` when a `ThinkingPart` is rewritten as text would have turned a multi-day investigation into a log line.

### Environment

- `pydantic-ai-slim[bedrock]` 2.36.0 (repro above); originally hit on 2.23.0, and `main` carries the same code at `models/bedrock.py:1306-1327`
- Bedrock Converse, Anthropic Claude with extended thinking, streaming

simpleqt:
Fixes #7908

## What

On Anthropic-via-Bedrock with extended thinking, a thinking-only response whose thinking block lost its signature in streaming (the signature delta never arrives; `_process_response`/streaming leave `provider_name=None`) used to be re-sent on the output-validation retry rewritten as `<thinking>` text — the one payload modification the API deterministically rejects with a 400, killing the run with no output, one request later, silently.

## Change

In `BedrockConverseModel._map_messages`, for the **latest** assistant message on the Anthropic thinking variant, a thinking block with no provider attribution is now **omitted with a `UserWarning`** instead of rewritten:

- a turn the omission empties is dropped entirely (the existing `if content:` guard), so the retry re-asks the original prompt — the two user turns are folded by the existing merge pass — instead of sending a payload the API can never accept;
- **older-history** thinking keeps the tagged-text rewrite (the API only protects the latest assistant message — pinned by a new test);
- **unsigned-reasoning families** (DeepSeek R1, Nova) are untouched: their rewrite was never a modification any API checks, so the omission is gated on `bedrock_thinking_variant == 'anthropic'`, not on `bedrock_send_back_thinking_parts` (DeepSeek R1 sets the latter but its thinking arrives unsigned by design — pinned by new tests);
- foreign-provider thinking parts (`provider_name` set) still take the rewrite even in the trailing position — the API never saw them.

This is the adapter-level boundary from the triage comment: the graph keeps the response + retry; the adapter owes a survivable payload (same invariant as #6139 for OpenAI chat histories), and the omission is now observable, as the issue's direction 3 asked.

## Tests

Seven new `_map_messages` unit tests in `tests/models/test_bedrock.py` assert the exact wire form: the issue's scenario (assistant turn absent, warning raised, user turns merged), unsigned-trailing-beside-a-tool-call (turn kept minus the block), older-history rewrite, signed trailing sent verbatim as `reasoningContent`, foreign trailing rewrite, Nova rewrite, DeepSeek R1 rewrite. Full `tests/models/test_bedrock.py`: 200 passed (the pre-existing DeepSeek VCR test covers the unsigned-reasoning boundary end-to-end as well). `ruff check`/`format` and `pyright` clean.

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7919?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/models/bedrock.py
+++ b/pydantic_ai_slim/pydantic_ai/models/bedrock.py
@@ -1160,6 +1160,19 @@ class BedrockConverseModel(Model[BaseClient]):
         # Most families accept a `status` field on `toolResult` blocks; Writer Palmyra rejects it.
         supports_tool_result_status = profile.get('bedrock_supports_tool_result_status', True)
 
+        # Whether this model's reasoning blocks round-trip: for these families (Anthropic, DeepSeek R1)
+        # thinking parts are re-sent as `reasoningContent`. Anthropic additionally signs every thinking
+        # block and enforces that the latest assistant message re-sends them unmodified — so there a
+        # block whose signature never arrived (the shape Bedrock's response/stream mapping leaves on a
+        # `ThinkingPart` when the signature is missing, i.e. `provider_name=None`) has no survivable wire
+        # form and is omitted instead of rewritten (see the `ThinkingPart` mapping below). DeepSeek R1's
+        # and Nova's reasoning is unsigned, so their tagged-text rewrite is accepted as-is.
+        send_back_thinking = profile.get('bedrock_send_back_thinking_parts', False)
+        enforces_thinking_integrity = profile.get('bedrock_thinking_variant', None) == 'anthropic'
+        last_model_response_index = max(
+            (i for i, message in enumerate(messages) if isinstance(message, ModelResponse)), default=None
+        )
+
         # Media returned from a tool that can't live inside a `toolResult` block (see
         # `bedrock_supported_media_kinds_in_tool_returns`) is emitted as a sibling block. Models like
         # Mistral and Llama require every `toolResult` for a tool-use turn to sit together in the message
@@ -1175,7 +1188,7 @@ class BedrockConverseModel(Model[BaseClient]):
                 bedrock_messages.append({'role': 'user', 'content': [*deferred_media_content]})
                 deferred_media_content.clear()
 
-        for message in messages:
+        for index, message in enumerate(messages):
             if isinstance(message, ModelRequest):
                 for part in message.parts:
                     if isinstance(part, SystemPromptPart):
@@ -1300,15 +1313,12 @@ class BedrockConverseModel(Model[BaseClient]):
             elif isinstance(message, ModelResponse):
                 flush_deferred_media()
                 content: list[ContentBlockOutputTypeDef] = []
+                is_latest_model_response = index == last_model_response_index
                 for item in message.parts:
                     if isinstance(item, TextPart):
                         content.append({'text': item.content})
                     elif isinstance(item, ThinkingPart):
-                        if (
-                            item.provider_name == self.system
-                            and item.signature
-                            and profile.get('bedrock_send_back_thinking_parts', False)
-                        ):
+                        if item.provider_name == self.system and item.signature and send_back_thinking:
                             reasoning_content: ReasoningContentBlockOutputTypeDef
                             if item.id == 'redacted_content':
                                 reasoning_content = {
@@ -1322,6 +1332,22 @@ class BedrockConverseModel(Model[BaseClient]):
                                     }
                                 }
                             content.append({'reasoningContent': reasoning_content})
+                        elif is_latest_model_response and enforces_thinking_integrity and not item.provider_name:
+                            # An Anthropic thinking block whose signature never arrived has no survivable
+                            # wire form: the API rejects any modified thinking in the latest assistant
+                            # message, and without the signature the block can't be re-sent verbatim. The
+                            # tagged-text rewrite used for older history below is exactly the modification
+                            # that turns a recoverable retry into a permanent 400, so omit the block (a
+                            # turn left empty by this is dropped below) and make the omission observable.
+                            # See https://github.com/pydantic/pydantic-ai/issues/7908.
+                            warnings.warn(
+                                'Omitting a thinking block with no signature from the latest assistant message: '
+                                'this model requires thinking blocks in the latest assistant message to be re-sent '
+                                'unmodified, so sending the unsigned block in any form would be rejected. The '
+                                'request proceeds without it, and if this leaves that message empty it is dropped '
+                                'entirely. See https://github.com/pydantic/pydantic-ai/issues/7908',
+                                UserWarning,
+                            )
                         else:
                             start_tag, end_tag = self.profile.get('thinking_tags', DEFAULT_THINKING_TAGS)
                             content.append({'text': '\n'.join([start_tag, item.content, end_tag])})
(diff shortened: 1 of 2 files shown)


DouweM:
> This comment was written by Claude Code (claude-opus-5-5) on behalf of @DouweM.

Thanks @simpleqt for putting this together! I'm marking it as a draft while we pin down the actual trigger in #7908: against the live Bedrock API, the `<thinking>` text downgrade this PR removes is accepted on every Claude model we tried. Omitting the block would drop the model's reasoning from the retry request without a demonstrated failure to prevent. Details and the questions for the reporter are in https://github.com/pydantic/pydantic-ai/issues/7908#issuecomment-5786273989. We'll pick this back up once that's resolved.

PR #7922, opened 2026-08-30
Detect Azure content-filter errors for `AzureProvider` and for `AsyncAzureOpenAI` clients on `OpenAIProvider`

### LINKED ISSUE #7789: Azure-specific behavior in `OpenAIChatModel` is provider-name-keyed, so `OpenAIProvider(openai_client=AsyncAzureOpenAI(...))` misses it

`OpenAIProvider` accepts a pre-built `openai_client=`, and `AsyncAzureOpenAI` subclasses `AsyncOpenAI`, so `OpenAIProvider(openai_client=AsyncAzureOpenAI(...))` typechecks and constructs. The resulting model talks to Azure but reports `name == 'openai'`, and every Azure-specific behavior in `models/openai.py` is selected by the provider *name*.

## Mechanism

Two name gates, both in `pydantic_ai_slim/pydantic_ai/models/openai.py`:

- **line 453** — `_check_azure_content_filter` opens with `if system == 'azure' and e.status_code == 400`. Called from `_completions_create` (1175), `_responses_compact` (2149) and `_responses_create` (2758).
- **line 1915** — `_raise_document_input_not_supported_error` branches on `self._provider.name == 'azure'` to choose the error text.

Reproduced:

```
OpenAIProvider(openai_client=AsyncAzureOpenAI(...))   name=openai
AzureProvider(...)                                    name=azure
```

## What a user loses

**Content filtering (line 453).** An Azure content-filter rejection is meant to become `ModelResponse(finish_reason='content_filter')`. Through `OpenAIProvider` it stays a raw `ModelHTTPError`. Verified end-to-end against an `httpx.MockTransport` 400: `AzureProvider` maps it, `OpenAIProvider` + Azure client raises.

**Document input (line 1915).** Here the divergence happens earlier and the line itself is never reached. `openai_chat_supports_document_input=False` is set by `providers/azure.py:145`, so through `OpenAIProvider` the model keeps the default `True`:

```
OpenAIProvider(openai_client=AsyncAzureOpenAI)   supports_document_input=True   -> guard not reached
AzureProvider(...)                               supports_document_input=False  -> guard reached
```

So the user does not get the wrong *message* — they get no guard at all, and a document is sent to an API that cannot accept it. Same root cause, different surface: the profile is provider-derived too.

## Why this is currently invisible

The content-filter mapping is inert for **everyone**, including `AzureProvider`. `_AzureErrorResponse` (line 393) is shaped `{'error': _AzureError}`, but on `openai` 3.0.0 the SDK's `_make_status_error` already unwraps `body['error']`, so `e.body` is the inner dict:

```
type(e) = BadRequestError
e.body  = {'code': 'content_filter', 'message': 'blocked', 'status': 400, 'innererror': {...}}
```

`model_validate` raises `ValidationError`, which `_check_azure_content_filter` swallows, returning `None` on every real request. `tests/models/test_openai.py::test_azure_prompt_filter_error` passes only because it hand-builds `APIStatusError(..., body={'error': {...}})` rather than going through the SDK's own error construction.

That is a second defect, and it is why the first one has never been noticed: the name gate cannot be observed while nothing downstream of it runs.

## Prior art

#7280 settled the analogous question for Google — `GoogleModel` now derives its Vertex-vs-Gemini-API branching from the client rather than the provider name, after `GoogleProvider(client=genai.Client(vertexai=True))` was found misrouted in #6792. The constraint there was that `Provider.name` flows into `ModelMessage.provider_name` and is checked during history replay, so it cannot simply be recomputed from the client.

Whether the same seam is right here is a separate question and deliberately not proposed in this issue — the Azure surface has its own semantics, and the two defects above interact.

quinnarnold:
## Why we make these changes

Closes #7789. Azure clients supplied through `OpenAIProvider` miss the document-input guard and content-filter mapping because the provider name stays `openai`. The SDK also unwraps Azure error bodies, so the previous parser misses real content-filter errors even through `AzureProvider`.

Detect the Azure client transport while preserving the provider name used in message history. Honor explicit dictionary and callable profile overrides when applying the Azure document-input default.

## New public surface

None.

## User-visible behavior

`Agent(OpenAIChatModel(..., provider=OpenAIProvider(openai_client=azure_client)))` now rejects unsupported document input by default before sending it. Azure content-filter responses become `ContentFilterError` with the finish reason and filter details intact, through both Chat Completions and Responses.

## Verification

- [Azure behavior regression](https://github.com/pydantic/pydantic-ai/pull/7922/files#diff-52dd30774371beafe168f376f79554db9493c353452c52637b6ea74e1647bf08R5613): six profile cases per provider, plus real SDK error mapping through both Chat Completions and Responses. Pins the error message, finish reason, filter details, and provider name.
- Restoring the profile override bug makes both Chat Completions provider cases fail.
- 27 focused Azure/document/profile tests pass; slim installation collects successfully with 234 expected skips. All repository pre-commit hooks pass, including full Pyright.

## What changes for existing users

Azure-backed clients gain the intended default guard and content-filter handling; explicit user profile overrides and persisted provider names remain intact.

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7922?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


dsfaccini:
thank you @quinnarnold

PR #7943, opened 2026-08-31
Expose OpenAI response `service_tier` in `ModelResponse.provider_details`

### LINKED ISSUE #7941: Expose the actual OpenAI response service tier on ModelResponse
## Description

Pydantic AI supports sending OpenAI service-tier settings through both `OpenAIChatModel` and `OpenAIResponsesModel`, including `openai_service_tier="priority"` and the unified `service_tier` setting. However, the actual service tier reported by OpenAI is not exposed on the resulting `ModelResponse`.

OpenAI returns a top-level `service_tier` field on both Chat Completions and Responses API response objects. This is the tier that actually served the request, and it can differ from the requested tier. For example, a Fast/Priority request may be downgraded because of ramp-rate limits; in that case OpenAI returns `service_tier: "default"` and charges Standard rates.

The current OpenAI adapter already parses this field:

- `_ChatCompletion` and `_ChatCompletionChunk` widen `service_tier` to `str | None`, following #4911, so OpenAI-compatible providers can return non-standard values without validation failures.
- The request paths pass `_resolve_openai_service_tier(model_settings)` to both Chat Completions and Responses API calls.

But the response mapping does not copy the provider-reported value into `ModelResponse`. It is therefore lost after the SDK response has been processed.

### Proposed behavior

Expose the actual provider-reported tier in `ModelResponse.provider_details`, for example:

```python
response.provider_details == {
    # existing details...
    "service_tier": "priority",
}
```

`provider_details` seems like the natural location because this is provider-reported response metadata, similar to the existing OpenAI moderation, finish-reason, timestamp, conversation, and background details. A first-class `ModelResponse` field could also work if there is a broader cross-provider design preference.

The value should come from the provider response, not from the requested model settings. It should remain an arbitrary `str` for OpenAI-compatible providers, consistent with #4911. If the provider does not return the field, the key can remain absent.

Ideally this would be consistent across all four OpenAI response paths:

- Chat Completions, non-streaming
- Chat Completions, streaming
- Responses API, non-streaming
- Responses API, streaming

### Use cases

The primary use case is correct request-level cost attribution. Pricing can depend on the service tier:

- Standard/default uses standard rates.
- Flex uses discounted rates.
- Fast/Priority uses premium rates.
- A requested Priority tier can actually be served and billed as default after a downgrade.
- `auto` can depend on project-level configuration, so the request setting alone is not sufficient.

Without the actual response tier, consumers must either assume the requested tier or use the standard model price. Both can produce incorrect costs. The same field is also useful for downgrade diagnostics, observability, and reconciliation with provider usage exports.

A possible consumer flow would be:

```python
actual_tier = response.provider_details.get("service_tier")
cost = calc_price(
    response.usage,
    model_ref=response.model_name,
    provider_id=response.provider_name,
    price_context={"service_tier": actual_tier},
)
```

The `price_context` API above is being proposed separately in `genai-prices`; the requested Pydantic AI feature is only to preserve the provider response metadata needed by such consumers.

## References

- OpenAI Fast mode documentation: https://developers.openai.com/api/docs/guides/fast-mode
  - The response `service_tier` identifies the tier used to process the request.
  - A downgraded Fast/Priority request returns `service_tier: "default"` and is charged Standard rates.
- OpenAI API pricing: https://developers.openai.com/api/docs/pricing
- Pydantic AI request-side OpenAI service-tier support: #1923 and #2368
- Pydantic AI cross-provider service-tier setting: #4926
- Pydantic AI response validation for provider-specific service tiers: #4911
- `genai-prices` OpenAI Flex/Priority pricing request: https://github.com/pydantic/genai-prices/issues/115
- `genai-prices` price-variant proposal: https://github.com/pydantic/genai-prices/pull/547

Wh1isper:
OpenAI reports the service tier that actually served each request, which can differ from the requested tier and determines billing. Pydantic AI accepted the field in the SDK response but discarded it while building `ModelResponse`.

This keeps provider-specific metadata in `ModelResponse.provider_details` and covers Chat Completions and Responses API, both streamed and non-streamed. The implementation was generated by an AI coding agent and reviewed line-by-line by the human PR author.

### New public surface

None.

### User-visible behavior

```python
# Before: the provider-reported tier was unavailable.
# After:
response.provider_details['service_tier'] == 'default'
```

The value comes from the provider response, remains an arbitrary string for OpenAI-compatible providers, and is absent when the provider does not return it.

### Verification

- [Chat non-stream and non-standard provider value](https://github.com/Wh1isper/pydantic-ai/blob/531093a430e10bcc9c8bcb0510c08b1bc15cf469/tests/models/test_openai.py#L5008)
- [Chat stream with the tier only on a choice-less final chunk](https://github.com/Wh1isper/pydantic-ai/blob/531093a430e10bcc9c8bcb0510c08b1bc15cf469/tests/models/test_openai.py#L641)
- [Recorded Chat non-stream/stream responses](https://github.com/Wh1isper/pydantic-ai/blob/531093a430e10bcc9c8bcb0510c08b1bc15cf469/tests/models/test_openai.py#L2228-L2480)
- [Recorded Responses API non-stream/stream responses](https://github.com/Wh1isper/pydantic-ai/blob/531093a430e10bcc9c8bcb0510c08b1bc15cf469/tests/models/test_openai_responses.py#L1318-L1550)
- `483` OpenAI model tests, Ruff, Pyright, and all pre-commit hooks pass.

OpenAI responses that include `service_tier` now expose it in `ModelResponse.provider_details`; request behavior is unchanged.

- Closes #7941

### Checklist

<a href="https://cubic.dev/pr/pydantic/pydantic-ai/pull/7943?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Wh1isper:
CI update: the full core test matrix, coverage, durable-exec, lowest-version, quality, and review checks are green. The only failing required check is `Harness Compat`, which found seven stale researcher integration snapshots after `service_tier` became observable.

Companion PR pydantic/pydantic-ai-harness#742 updates exactly those seven snapshots and is green. Once it is merged into harness `main`, rerunning the failed compatibility workflow should clear the final check here.

PR #7954, opened 2026-08-31
Context-manage models rebuilt inside durable operations

### LINKED ISSUE #7944: `DBOSDurability` does not context-manage models resolved by `ResolveModelId`
### Initial Checks

- [ ] I'm using the latest released version. I reproduced this with `2.32.0`; the latest release is `2.36.0`. `main` at `f4f6ddee242ade6f8146c5255fabe5d257060684` still has the same unentered-model path, but I have not run the reproduction against `2.36.0` yet.
- [x] I've searched the issue tracker before opening this issue.

### Description

`DBOSDurability` supports a model-name string resolved by `ResolveModelId`, which is the documented route for constructing a model inside a durable model step.

The DBOS adapter resolves that model but never enters or exits its async context. Consequently, a resolver-created model cannot release provider-owned resources through its normal lifecycle.

In particular, an `OpenAIResponsesModel` created with an `OpenAIProvider` that owns its HTTP client leaves that client open. Custom async credential callbacks can have the same issue when they own loop-bound resources.

`BaseDurabilityCapability._durable_model_scope` currently does:

```python
with self._durable_run_context_scope(run_context) as ctx:
    model = await self._resolve_model_for_request(model_id, ctx)
    yield model, ctx
```

The DBOS request, streamed-request, and suspended-response-cancellation steps all use this scope.

Expected behavior: a model dynamically returned by `ResolveModelId` is entered before a durable operation and exited after it completes. For a stream, exit must happen after the stream is consumed.

Actual behavior: ordinary and streamed durable requests invoke the resolved model without calling either `__aenter__` or `__aexit__`.

This needs DBOS configured against a local Postgres database. It uses `TestModel` only, with no provider credentials or network model request.

```python
import asyncio
from uuid import uuid4

from dbos import DBOS, DBOSConfig, SetWorkflowID
from pydantic_ai import Agent
from pydantic_ai.capabilities import ResolveModelId
from pydantic_ai.durable_exec.dbos import DBOSDurability


DBOS(
    config=DBOSConfig(
        name='resolved-model-lifecycle-repro',
        system_database_url='postgresql://postgres:dbos@localhost:5432/postgres',
        application_database_url='postgresql://postgres:dbos@localhost:5432/postgres',
    )
)

counters = {'entered': 0, 'exited': 0, 'requests': 0, 'streams': 0}


class TrackingModel(TestModel):
    async def __aenter__(self):
        counters['entered'] += 1
        return await super().__aenter__()

    async def __aexit__(self, *args):
        counters['exited'] += 1
        return await super().__aexit__(*args)

    async def request(self, *args):
        counters['requests'] += 1
        return await super().request(*args)

    def request_stream(self, *args, **kwargs):
        counters['streams'] += 1
        return super().request_stream(*args, **kwargs)


agent = Agent(
    'application-model',
    name='resolved-model-lifecycle-repro',
    capabilities=[
        ResolveModelId(lambda _ctx, _model_id: TrackingModel(custom_output_text='ok')),
        DBOSDurability(),
    ],
)


@DBOS.workflow(name='resolved-model-lifecycle-request')
async def request_workflow():
    return (await agent.run('hello')).output


@DBOS.workflow(name='resolved-model-lifecycle-stream')
async def stream_workflow():
    async with agent.run_stream('hello') as result:
        return await result.get_output()


async def main():
    DBOS.launch()
    with SetWorkflowID(str(uuid4())):
        assert await request_workflow() == 'ok'
    with SetWorkflowID(str(uuid4())):
        assert await stream_workflow() == 'ok'

    print(counters)
    assert counters == {'entered': 2, 'exited': 2, 'requests': 1, 'streams': 1}


asyncio.run(main())
```

Actual result:

```text
{'entered': 0, 'exited': 0, 'requests': 1, 'streams': 1}
AssertionError
```

### Logfire Trace

Not applicable. The reproduction uses `TestModel` and no external model request.

- Python: 3.13.8
- Pydantic AI: 2.32.0
- LLM provider SDK: not applicable (`TestModel`)

dsfaccini:
> Opened by @dsfaccini, whose commit is the shared-scope half. Taken over by @DouweM, who added the Temporal half; see the discussion on #7944.

## Why we make these changes

A durable model unit rebuilds a string-selected model on the worker, through a `ResolveModelId` capability or the `infer_model` backstop, and then used that model without ever entering its async context. Anything the model owns therefore leaks: an `OpenAIResponsesModel` whose `OpenAIProvider` owns its HTTP client leaves that client open, and an async credential callback holding loop-bound resources never releases them. Reported in #7944 with a self-contained reproduction that counts `__aenter__` / `__aexit__` calls and observes zero of each.

The fix is ownership-aware rather than unconditional. A model the durable unit built is owned by that unit and gets its normal lifecycle. The agent's default model and anything registered through `models=` keep their existing external lifecycle owner and are never entered a second time, including when a resolver chain hands back a registered instance (the capability's own registry acts as the durable backstop, so identity, not code path, decides).

There are three places that rebuild models, not one:

- `BaseDurabilityCapability._durable_model_scope`, which DBOS and Prefect go through.
- `TemporalDurability._cancel_suspended_response_without_run_context` in `durable_exec/temporal/_durability.py`. `_cancel_suspended_response_operation` returns early when `params.run_context is None` and so bypasses the shared scope, and Temporal's override then builds a model with `infer_model` and uses it unentered. That `None` is the backward-compat case for activity payloads recorded before the run context was carried, so this fires when an in-flight workflow from an older deploy retries a suspended-response cancellation. The base class only raises there, so Temporal is the sole engine implementing it and the sole one affected.
- `TemporalModel` in `durable_exec/temporal/_model.py`, which resolves independently in three activity bodies. This one serves only the deprecated `TemporalAgent` (its single construction site is `_agent.py:225`), so it is fixed here but goes away in v3.

Note that `TemporalDurability` inherits all four model operations from `BaseDurabilityCapability` and overrides none of them, so the ordinary Temporal request, streaming, and compaction paths are covered by the shared-scope fix rather than by anything Temporal-specific.

All three are fixed here, through one shared `managed_model_scope()` helper in `durable_exec/_utils.py` so the ownership handling exists once rather than in five places.

## New public surface

None. `managed_model_scope` is internal to `pydantic_ai.durable_exec`.

## User-visible behavior

```diff
Agent.run() / Agent.run_stream()
  └─ durable model unit
     ├─ shared scope (DBOS, Prefect, and Temporal's request/stream/compaction)
     ├─ TemporalDurability cancel-without-run-context (backward-compat payloads)
     └─ TemporalModel activities (deprecated TemporalAgent)
-       └─ rebuild model → request, consume stream, or cancel
+       └─ rebuild model → enter → request, consume stream, or cancel → exit
```

For a streamed request the model stays entered until the stream is fully consumed, and it is exited when the operation body raises.

## Verification

- Ordinary, streamed, and failing requests each record one matching enter and exit, asserted on the exact event order.
- Temporal: the backward-compat cancellation path (registered, default, and inferred, including the exception path), plus all three `TemporalModel` activity bodies for the deprecated `TemporalAgent`.
- DBOS: real SQLite-backed ordinary and streamed workflows.
- Prefect: ordinary and streamed lifecycle.
- Ownership: the agent default and a `models=` registration are never entered by the durable unit.
- Teardown edge cases: a truthy `__aexit__` cannot suppress an error raised by the operation body, and a teardown failure surfaces (chained) rather than vanishing.

On the combined branch: `tests/durable_exec` 696 passed, 2 skipped, 5 failed, where all five are the pre-existing Prefect cache-write failures from BlockBuster rejecting `os.mkdir`; `tests/durable_exec/temporal` 285 passed, 2 skipped; lint clean; per-file pyright clean on every changed file.

## What changes for existing users

A model constructed inside a durable unit now receives its normal async lifecycle, so provider-owned clients are closed instead of leaked. Concrete model instances you pass in or register keep their current lifecycle owner, so nothing changes for them.

Fixes #7944

### Checklist

- [x] Any permitted **compatibility impact** has a label, warning, migration, release note, and exact API-check waiver. (Not applicable.)

### DIFF
--- a/pydantic_ai_slim/pydantic_ai/durable_exec/_base.py
+++ b/pydantic_ai_slim/pydantic_ai/durable_exec/_base.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import copy
 from abc import abstractmethod
 from collections.abc import AsyncGenerator, AsyncIterable, AsyncIterator, Awaitable, Callable, Generator, Mapping
-from contextlib import asynccontextmanager, contextmanager
+from contextlib import AsyncExitStack, asynccontextmanager, contextmanager
 from functools import partial
 from typing import Any, ClassVar, Literal, NamedTuple, Protocol, TypeVar, cast, runtime_checkable
 from weakref import ref
@@ -667,16 +667,22 @@ class BaseDurabilityCapability(AbstractCapability[AgentDepsT]):
         both halves, and the guard is not optional for any of them: the unit's recorded result is
         replayed on recovery or a cache hit without re-running its body, so a `ctx.enqueue()` from
         the model -- or from a `resolve_model_id` capability rebuilding it -- would be dropped.
+        Models rebuilt inside the unit are owned and context-managed here; the agent's default and
+        `models=` registry instances keep their existing external lifecycle owner.
         Pairing the two here means a unit can't get its model without the guard, instead of each
         engine remembering to install it per unit (Temporal has its own chokepoint in
         `deserialize_run_context`, so it doesn't use this).
         """
         with self._durable_run_context_scope(run_context) as ctx:
             model = await self._resolve_model_for_request(model_id, ctx)
-            ctx.model = model
-            if isinstance(ctx, _RestrictedRunContext):
-                ctx._expose_field('model')  # pyright: ignore[reportPrivateUsage]
-            yield model, ctx
+            registered, _ = self._registered_model_id(model)
+            async with AsyncExitStack() as stack:
+                if not registered:
+                    model = await stack.enter_async_context(model)
+                ctx.model = model
+                if isinstance(ctx, _RestrictedRunContext):
+                    ctx._expose_field('model')  # pyright: ignore[reportPrivateUsage]
+                yield model, ctx
 
     def _build_resolve_tool_config(self, base_config: Any) -> Callable[[ToolsetTool[Any] | None, str], ToolConfig]:
         """Build the per-tool config resolver from declarative fields (metadata key + polarity)."""
--- a/pydantic_ai_slim/pydantic_ai/durable_exec/_base.py
+++ b/pydantic_ai_slim/pydantic_ai/durable_exec/_base.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import copy
 from abc import abstractmethod
 from collections.abc import AsyncGenerator, AsyncIterable, AsyncIterator, Awaitable, Callable, Generator, Mapping
-from contextlib import AsyncExitStack, asynccontextmanager, contextmanager
+from contextlib import asynccontextmanager, contextmanager
 from functools import partial
 from typing import Any, ClassVar, Literal, NamedTuple, Protocol, TypeVar, cast, runtime_checkable
 from weakref import ref
@@ -106,7 +106,7 @@ from ._toolset import (
     validate_dynamic_tool_args,
     wrap_tool_call_result,
 )
-from ._utils import DurableModel, StreamedActivityResult, capture_event_stream, unwrap_model
+from ._utils import DurableModel, StreamedActivityResult, capture_event_stream, managed_model_scope, unwrap_model
 
 _T = TypeVar('_T')
 
@@ -676,13 +676,11 @@ class BaseDurabilityCapability(AbstractCapability[AgentDepsT]):
         with self._durable_run_context_scope(run_context) as ctx:
             model = await self._resolve_model_for_request(model_id, ctx)
             registered, _ = self._registered_model_id(model)
-            async with AsyncExitStack() as stack:
-                if not registered:
-                    model = await stack.enter_async_context(model)
-                ctx.model = model
+            async with managed_model_scope(model, owned=not registered) as active_model:
+                ctx.model = active_model
                 if isinstance(ctx, _RestrictedRunContext):
                     ctx._expose_field('model')  # pyright: ignore[reportPrivateUsage]
-                yield model, ctx
+                yield active_model, ctx
 
     def _build_resolve_tool_config(self, base_config: Any) -> Callable[[ToolsetTool[Any] | None, str], ToolConfig]:
         """Build the per-tool config resolver from declarative fields (metadata key + polarity)."""
--- a/pydantic_ai_slim/pydantic_ai/durable_exec/_utils.py
+++ b/pydantic_ai_slim/pydantic_ai/durable_exec/_utils.py
@@ -11,6 +11,7 @@ helpers are reserved for the bundled `temporal`, `dbos`, and `prefect`
 integrations.
 """
 
+import sys
 from collections.abc import AsyncGenerator, AsyncIterable, AsyncIterator, Awaitable, Callable
 from contextlib import asynccontextmanager
 from dataclasses import dataclass
@@ -30,8 +31,26 @@ __all__ = [
     'StreamedActivityResult',
     'disable_threads',
     'capture_event_stream',
+    'managed_model_scope',
     'unwrap_model',
 ]
+
+
+@asynccontextmanager
+async def managed_model_scope(model: Model, *, owned: bool) -> AsyncGenerator[Model]:
+    """Context-manage a model when the current durable unit owns it."""
+    if not owned:
+        yield model
+        return
+
+    active_model = await model.__aenter__()
+    try:
+        yield active_model
+    except BaseException:
+        await model.__aexit__(*sys.exc_info())
+        raise
+    else:
+        await model.__aexit__(None, None, None)
 
 
 def unwrap_model(model: Model) -> Model:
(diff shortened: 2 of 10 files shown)


DouweM:
@dsfaccini I had this in flight when you opened it (see my note on #7944), so rather than put up a competing PR I have pushed my half onto your branch as `f41abef32`, using the "allow edits by maintainers" permission. Your commit is untouched and this stays your PR: the shared-scope fix and its tests are yours, and they were right.

There is one real gap, and it is the reason I did not just close mine.

**Temporal is still broken by this PR as it stood.** The description says the changed ownership logic is shared by the callable durable-operation base used by DBOS, Prefect, and Temporal. Temporal does not use `_durable_model_scope` at all: the docstring you extended says so explicitly ("Temporal has its own chokepoint in `deserialize_run_context`, so it doesn't use this"), and `durable_exec/temporal/_model.py` resolves the model independently in three activity bodies, each of which then used it unentered:

- the ordinary request activity,
- the streamed request activity (the model has to stay entered until the stream is fully consumed, so the scope has to wrap the whole `request_stream` block, not just the resolve),
- the suspended-response cancellation activity.

All three are fixed in the new commit, with an identity-based `_is_registered_model` on `TemporalModel` playing the part your `_registered_model_id` plays on the base.

The rest of what I added:

- **One shared helper instead of four copies.** `managed_model_scope(model, *, owned=...)` now lives in `durable_exec/_utils.py`, which is where this package already keeps cross-engine helpers, so Temporal does not have to reach into `_base` for a private name. Your `_durable_model_scope` now calls it; behaviour is unchanged from your `AsyncExitStack` version.
- **Tests for the engines the PR did not reach**: Temporal inferred and registered models, and Prefect ordinary plus streamed lifecycle. Your DBOS integration tests stay as they are.
- **Two teardown edge cases**: a truthy `__aexit__` cannot suppress an error raised by the operation body, and a teardown failure surfaces rather than disappearing. I deliberately kept your simpler semantics here (the teardown exception surfaces with normal chaining) rather than the log-and-swallow variant I had first written; there is no case where the caller is better off with the teardown error only in a log.

Verified on the combined branch: `tests/durable_exec` at 696 passed, 2 skipped, 5 failed, where all five are the pre-existing Prefect cache-write failures from BlockBuster rejecting `os.mkdir`; `tests/durable_exec/temporal` at 285 passed, 2 skipped; lint clean; per-file pyright clean on all six changed files.

Worth updating the PR description's verification section, since the "Temporal is covered by the shared base" reasoning no longer describes what the PR does.

DouweM:
Correcting myself on the earlier comment, @dsfaccini: I said this PR "leaves Temporal broken". That was wrong as stated, and the accurate version is narrower.

`TemporalDurability` inherits `_model_request_operation`, `_model_request_stream_operation`, `_cancel_suspended_response_operation`, and `_compact_messages_operation` from `BaseDurabilityCapability` and overrides none of them, so the modern Temporal path does go through `_durable_model_scope` and was already fixed by your commit alone. I misread the parenthetical in that method's docstring ("Temporal has its own chokepoint in `deserialize_run_context`, so it doesn't use this"), which is about installing the enqueue guard, not about the scope. Your original scoping instinct was better than my correction of it.

What the Temporal commits on this branch actually fix:

1. **`TemporalModel` (`temporal/_model.py`)**: a real leak, but only on the deprecated `TemporalAgent` path. `TemporalModel` has exactly one construction site, `_agent.py:225`, inside the `PydanticAIDeprecationWarning`-carrying `TemporalAgent`. Worth fixing while it exists; it goes away in v3.

2. **`TemporalDurability._cancel_suspended_response_without_run_context` (`temporal/_durability.py:526`, commit `a4d12bb61`)**: this one is on the modern path and is the same #7944 defect. It called `infer_model(model_id)` and used the result without entering it. It is reached from `_base.py:1380`, where `if params.run_context is None:` returns early and bypasses `_durable_model_scope` entirely, and the Temporal transport produces that `None` when `serialized_run_context is None`, which is the backward-compat case for activity payloads recorded before the run context was carried. So it fires when an in-flight workflow from an older deploy retries a suspended-response cancellation. The base class only raises there, so Temporal is the sole engine that implements it and the sole one affected.

Same `managed_model_scope` helper, same ownership rule (registry hit means externally owned and must not be entered; `infer_model` result is owned by the unit), with the ownership flag computed before the reassignment. Tests pin both sides plus the exception path.

Also worth recording, since pydanty filed issues for both: #7956 (Temporal activities do not expose `ctx.model`) and #7957 (`_is_registered_model` duplicating `_registered_model_id`) are both scoped to `temporal/_model.py`, so both are confined to the deprecated `TemporalAgent` path and disappear with it in v3. I would not spend a PR on either.

DouweM:
For the record, one model-construction site was deliberately left out of this PR: `BaseDurabilityCapability._invoke_capability_operation` rebuilds a `Model` at `_base.py:439` (post-merge line numbers) when a `before_model_request` hook swaps `model_id`, and nothing context-manages that instance either.

It is out of scope here because the fix is a different shape, not the `managed_model_scope` ownership rule this PR establishes: that site sits outside the operation body, so it re-runs on every replay and cache hit, and inside a durable container it is already unreachable in the leaking form (`_find_model_id` rejects an unregistered instance two frames later). The leak is only observable outside a durable container, where the whole projection round-trip should not be running at all.

Filed with a reproducer as #7969.

Issue #7985, opened 2026-09-01
Remove the `anthropic` 1.3.0 dependency cutoff after 2026-09-08


dsfaccini:
> This issue was posted by Claude Code using claude-opus-5 on behalf of David.

`pyproject.toml` carries a per-package `exclude-newer` cutoff admitting `anthropic` 1.3.0 ahead of the repo's 7-day resolution cooldown:

```toml
[tool.uv.exclude-newer-package]
anthropic = "2026-09-01T17:37:11Z"
```

1.3.0 is the first SDK release listing `claude-fable-5-1` and `claude-mythos-5-1`, which `test_known_model_names` generates `KnownModelName` from, so the floor bump could not resolve without it.

**Remove the entry after 2026-09-08 17:37 UTC**, once the global `exclude-newer = "7 days"` covers 1.3.0 on its own. Left in place past that date it stops acting as a floor and becomes a ceiling, quarantining every `anthropic` release after 1.3.0.

Same shape as the 1.0.0 cutoff removed in #7921.

PR #7989, opened 2026-09-01
Add Claude Fable 5.1 (`claude-fable-5-1`) and Claude Mythos 5.1 (`claude-mythos-5-1`) support

### LINKED ISSUE #7984: Claude Fable 5.1 rejects replayed thinking blocks when `instructions` or the toolset change

Claude Fable 5.1 binds each thinking block to the conversation prefix that produced it. Replaying message history after the `system` prompt text changes, or after the `tools` array grows, now returns a 400 — and both are things Pydantic AI does by design, via dynamic `@agent.instructions` and via conditional / deferred toolsets.

```
messages.1.content.0: Invalid `signature` in `thinking` block. The block is bound to a
different conversation. Remove the block, or set
`thinking.block_binding.prefix_mismatch_behavior` to "drop_block". The `system` prompt …
```

## Verified live

Two-turn conversation on `claude-fable-5-1`, replaying turn 1's assistant content (a `thinking` block plus text) into turn 2:

| turn 2 differs from turn 1 by | `prefix_mismatch_behavior` unset / `"error"` | `"drop_block"` |
|---|---|---|
| nothing (control) | 200 | 200, no `input_transformations` |
| `system` prompt text | **400** | 200 + `thinking_dropped` |
| one extra entry in `tools` | **400** | 200 + `thinking_dropped` |
| thinking block stripped from history | 200 | 200, no `input_transformations` |

Under `"drop_block"` the drop is reported as `input_transformations: [{"type": "thinking_dropped", "path": "messages.1.content.0", "reason": "prefix_binding_mismatch"}]` and the model no longer sees the reasoning.

## Scope of the exposure

Anthropic gates enforcement on account age: the check is **enforced by default for accounts created on or after 2026-08-31**, and for older accounts the API records the mismatch but acts on it only when the request sets `prefix_mismatch_behavior`. Confirmed against our own (older) key — the default path returns 200 on `claude-fable-5-1`, `claude-fable-5`, and `claude-opus-5`. So this is currently silent for existing users and hard-fails for new accounts, which is the worst shape for a bug report to arrive in.

`thinking.block_binding` is reachable today by passing it through `AnthropicModelSettings.anthropic_thinking`, so there is a user-side workaround. What isn't decided is what Pydantic AI should do by default.

## What needs deciding

1. Does `AnthropicModel` pick a default `prefix_mismatch_behavior`, or stay on Anthropic's `"error"`?
2. If a default is set, `"drop_block"` silently degrades reasoning quality, while `"error"` surfaces a 400 whose message doesn't name the Pydantic AI feature that caused it (dynamic instructions, a changed toolset).
3. Either way, should the 400 be translated into a Pydantic AI error that names the likely cause?

Anything that changes the default moves observable behavior for every Anthropic user who round-trips thinking parts, not only Fable 5.1 users, so it wants a maintainer call before implementation.

Whichever default lands has to ship with documentation in the same PR: which setting Pydantic AI sends, what makes a thinking block fail the check, and — when the answer is `"drop_block"` — that the reasoning is dropped from the model's view rather than replayed, so the loss is visible to users before they hit it.

<details>
<summary>Repro script</summary>

```python
import json, os, urllib.request, urllib.error

KEY = os.environ['ANTHROPIC_API_KEY']
BETA = 'thinking-binding-controls-2026-08-01'
MODEL = 'claude-fable-5-1'
TOOL = {
    'name': 'get_weather',
    'description': 'Get weather for a city.',
    'input_schema': {'type': 'object', 'properties': {'city': {'type': 'string'}}, 'required': ['city']},
}


def call(payload):
    headers = {
        'x-api-key': KEY,
        'anthropic-version': '2023-06-01',
        'content-type': 'application/json',
        'anthropic-beta': BETA,
    }
    request = urllib.request.Request(
        'https://api.anthropic.com/v1/messages?beta=true', data=json.dumps(payload).encode(), headers=headers
    )
    try:
        with urllib.request.urlopen(request, timeout=120) as response:
            return 200, json.loads(response.read().decode())
    except urllib.error.HTTPError as e:
        return e.code, json.loads(e.read().decode())['error']['message']


SYSTEM_A = 'You are a helpful assistant. Answer briefly.'
SYSTEM_B = SYSTEM_A + ' Today is 2026-09-01.'
FIRST_USER = 'Think about it, then say what 17*23 is.'

code, message = call(
    {
        'model': MODEL,
        'max_tokens': 2048,
        'system': SYSTEM_A,
        'tools': [TOOL],
        'messages': [{'role': 'user', 'content': FIRST_USER}],
    }
)
assert code == 200, message
assistant = message['content']

for label, system, tools in [
    ('control', SYSTEM_A, [TOOL]),
    ('system changed', SYSTEM_B, [TOOL]),
    ('tools grew', SYSTEM_A, [TOOL, {'name': 'get_time', 'description': 'Get the time.', 'input_schema': {'type': 'object', 'properties': {}}}]),
]:
    for behavior in ('error', 'drop_block'):
        code, message = call(
            {
                'model': MODEL,
                'max_tokens': 512,
                'system': system,
                'tools': tools,
                'thinking': {'type': 'adaptive', 'block_binding': {'prefix_mismatch_behavior': behavior}},
                'messages': [
                    {'role': 'user', 'content': FIRST_USER},
                    {'role': 'assistant', 'content': assistant},
                    {'role': 'user', 'content': 'And times two?'},
                ],
            }
        )
        transformations = message.get('input_transformations') if code == 200 else None
        print(f'{label:16s} {behavior:12s} {code} {transformations}')
```

</details>

## References

- [What's new in Claude Fable 5.1 — editing earlier turns invalidates thinking blocks](https://platform.claude.com/docs/en/models/fable-5-1/whats-new-fable-5-1#editing-earlier-turns-invalidates-thinking-blocks)
- [Preserved thinking](https://platform.claude.com/docs/en/build-with-claude/thinking#preserved-thinking)

Found while adding `claude-fable-5-1` support.

dsfaccini:
- Closes #7984

### Why we make these changes

Pydantic AI now recognizes Claude Fable 5.1 and Claude Mythos 5.1. Fable 5.1 rejects replayed thinking blocks when the conversation prefix changes, as reported in [#7984](https://github.com/pydantic/pydantic-ai/issues/7984).

Pydantic AI preserves Anthropic’s account default, then retries a rejected Fable 5.1 replay once with `drop_block`. It also corrects forced tool choice: only the two 5.1 models reject it.

### New public surface

- Recognized direct, Gateway, and supported Bedrock Fable 5.1 model identifiers.
- `AnthropicStaleThinkingBlockWarning` and `AnthropicModelProfile.anthropic_binds_thinking_blocks`.

### User-visible behavior

```diff
- A stale Fable 5.1 thinking block could end the request with Anthropic's 400.
+ Pydantic AI retries once without that block, warns, and records the drop in `provider_details`.
```

### Verification

- [Known model names](https://github.com/pydantic/pydantic-ai/blob/fe05372a82a22113139d546395d082d8e61178b3/tests/models/test_model_names.py#L227) and [model-profile / forced-tool-choice coverage](https://github.com/pydantic/pydantic-ai/blob/fe05372a82a22113139d546395d082d8e61178b3/tests/profiles/test_anthropic.py#L369).
- Recorded [legacy-account replay](https://github.com/pydantic/pydantic-ai/blob/fe05372a82a22113139d546395d082d8e61178b3/tests/models/anthropic/test_thinking_block_binding.py#L644) and [sync / streaming retry](https://github.com/pydantic/pydantic-ai/blob/fe05372a82a22113139d546395d082d8e61178b3/tests/models/anthropic/test_thinking_block_binding.py#L676) tests.

### Deliberate decisions

Choices a reviewer could mistake for an oversight. Each was made on evidence, not convenience.

<details>
<summary>Decision log</summary>

**Default sends no `block_binding` at all; a rejection is retried once.** Not a `drop_block` default. For accounts created before 2026-08-31 Anthropic records the prefix mismatch but acts on it *only* when the request sets `prefix_mismatch_behavior` — so defaulting to `drop_block` would discard reasoning on accounts the check never touches, to fix a failure only newer accounts have. The carve-out is pinned by its own recorded cassette rather than resting on the docs sentence.

**`anthropic_binds_thinking_blocks` is Fable 5.1 only — Mythos 5.1 is deliberately excluded.** Mythos 5.1 mirrors Fable 5.1 elsewhere in this PR, so the omission looks like a miss. Anthropic's Fable 5.1 release notes state outright that Mythos 5.1 does not run the check; including it would put a field plus a beta header on the wire of the one model we cannot probe, for a check that never fires.

**The binding narrowing is a client-class boundary, not a base-URL one.** `anthropic_binds_thinking_blocks` is cleared for the Bedrock, Bedrock Mantle, Foundry and Vertex clients, which leaves a proxied or Gateway `AsyncAnthropic` inside it. That reads as an accidental hole and is not one. The beta follows `thinking.block_binding`, which is unset by default, and the retry needs a 400 whose body carries Anthropic's own `The block is bound to a different conversation` — so nothing extra reaches the wire for those clients unless Anthropic itself rejects a replay. Measured through the Pydantic AI Gateway rather than assumed: `claude-fable-5-1` reaches Anthropic, the profile binds, and a forced `prefix_mismatch_behavior='error'` returns that byte-identical 400. Gateway behavior therefore matches the direct API, and gating on the `api.anthropic.com` host would have been the change that broke that parity.

**The container-500 fallback classifies its own error, not the 500 that caused it.** A `[500, stale-thinking-400]` sequence on a history-resolved container previously sent two requests and propagated the 400 with no retry, so this PR's own recovery never fired. The fallback's error is now rebound and classified, and the dropped container is carried forward so the drop-block retry does not resend an id the 500 disowned.

**A model add touching `.github/scripts/` and the `add-new-model` skill is mechanical, not scope creep.** `check_http_dependencies.py` mirrors the declared floors exactly, so raising the `anthropic` extra to `>=1.3.0` made its `contains('1.0.0')` assertion false and reddened quality checks. `pydantic-ai-runner.lock` needed regenerating because the shim smoke step imports `BetaThinkingBlockBindingParam`, which `anthropic` 1.0.0 does not carry. The skill's forced-tool-choice guidance named Fable 5 as the model that rejects forcing, which this PR's own correction falsified.

**The `anthropic.input_transformations` span event is a requirement, not decoration.** A dropped thinking block is otherwise silent. `provider_details` never reaches the span — `messages_to_otel_messages` emits `role` + `parts`, and `response_attributes` lifts only `provider_response_id` and `finish_reason` — and the warning is not trace-visible either, since nothing bridges `warnings` to telemetry. The name is provider-scoped because the transformation is Anthropic's own field, reported verbatim; a `pydantic_ai.*` name would promise a cross-provider shape that has not been designed.

**`block_binding` rides in `extra_body`, not the typed `thinking` param.** Fable 5.1 emits thinking blocks whether or not the request configured thinking, so the retry usually has no `thinking` object for the binding to live in, and the SDK's discriminated union always demands a `type`. Picking one would change how the model thinks; `extra_body` reaches the same JSON key without inventing a `type` the caller never asked for. A `block_binding` the caller set through `extra_body` is honored and never retried over.

**`message_delta` replaces the transformation array; it does not extend it.** The SDK documents that a `message_delta` carries `input_transformations` only after a server-side model fallback mid-stream, and then it holds the serving model's whole array — which repeats what `message_start` already reported. Appending duplicated every request-side entry.

**Forced tool choice is corrected for three models beyond the two being added.** `claude-fable-5`, `claude-mythos-5` and `claude-mythos-preview` can force tool choice again. This moves observable behavior for existing users and was a conscious call, re-verified live before changing it: `claude-fable-5` returns 200 on both `any` and `tool` against the GA and beta endpoints, while `claude-fable-5-1` 400s on all four. The two Mythos ids 404 on our credentials, so they follow Anthropic's published table.

**The narrow `warnings.filterwarnings` in the sanitizer tests is intentional.** `simplefilter('error')` there overrode the suite's deliberate `ResourceWarning` ignores, so delayed event-loop GC failed whichever test happened to collect it. The reason is now stated at both call sites.

**`startswith('claude-fable-5-1')` also matches a hypothetical `claude-fable-5-10`.** Left as-is: every sibling flag in the same function uses the bare-`startswith` convention, no current id resolves incorrectly, and Anthropic's dated-suffix names are handled. Worth revisiting only if such a name appears.

**The dated `anthropic` entry in `[tool.uv.exclude-newer-package]` is temporary.** It admits only the reviewed 1.3.0 release past the repo's supply-chain cooldown. It must be removed after 2026-09-08 17:37 UTC or it stops acting as a floor and becomes a ceiling — tracked in [#7985](https://github.com/pydantic/pydantic-ai/issues/7985).

</details>

Users of `claude-fable-5`, `claude-mythos-5`, and `claude-mythos-preview` can force tool choice again. Fable 5.1 stale-thinking replays now complete after dropping the stale reasoning block and emit a warning; other models are unchanged by default.

### Checklist


DouweM:
@dsfaccini I think we should add the `drop_block` rule for thinking here in this PR, instead of tracking it separately as #7984

dsfaccini:
> This reply was posted by Claude Code using claude-opus-5 on behalf of David.

Done — it's in this PR now, and `Closes #7984` is in the description.

Scoped to the models that actually bind, behind a new `anthropic_binds_thinking_blocks` profile flag (`claude-fable-5-1`, `claude-mythos-5-1`). Live probe: with a thinking block replayed and the `system` text changed, `claude-fable-5`, `claude-opus-5` and `claude-sonnet-5` all return 200 under an explicit `prefix_mismatch_behavior: 'error'` — only `claude-fable-5-1` returns the 400. So every other Anthropic request, including the Bedrock / Mantle / Foundry / Vertex clients where the beta is unverified, is unchanged.

One thing worth your eye. `block_binding` normally rides on the `thinking` config, but Fable 5.1 emits thinking blocks even when the request configures no thinking, and `ModelRequestParameters.thinking` defaults to `None` — so `_translate_thinking` returns `OMIT` for anyone who never touches thinking settings, which is both the common case and the one that fails. The API accepts a `thinking` object carrying `block_binding` alone (verified: with no `type`, `'error'` still 400s and `'drop_block'` still reports `thinking_dropped`), but the SDK's union always requires a `type` and picking one would change how the model thinks. So that shape goes through `extra_body`. That's the one call I'd want a second opinion on.

Covered by a live cassette that records the drop end to end, plus unit tests for the request shape.

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -887,14 +894,46 @@ def _translate_thinking(
...
+
+    def _bind_thinking_blocks(self, thinking: BetaThinkingConfigParam) -> BetaThinkingConfigParam:
+        """Default `prefix_mismatch_behavior` to `drop_block` on models that bind thinking blocks.
+
+        Claude Fable 5.1 binds each thinking block to the conversation prefix that produced it, so
+        replaying history after a dynamic `@agent.instructions` string changes, or after a
+        conditional toolset advertises a new non-deferred tool, fails with a 400. Dropping the stale
+        block costs the model that turn's reasoning; failing the request costs the run.
+        """
+        if thinking['type'] == 'disabled' or 'block_binding' in thinking:
```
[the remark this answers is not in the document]
dsfaccini: > This reply was posted by Claude Code using claude-opus-5 on behalf of David.

Confirmed and fixed in f4d70be91.

Verified against the live API before changing anything — `thinking: {'type': 'adaptive', 'block_binding': {}}` on `claude-fable-5-1` returns `400 thinking.adaptive.block_binding: Extra inputs are not permitted` without `thinking-binding-controls-2026-08-01`, and 200 with it. The SDK documents the empty object as meaning every binding default, so it is a valid request that our truthiness check was silently breaking.

The beta now attaches on membership (`'block_binding' in thinking`), with a regression test pinning the empty case.

dsfaccini:
please clarify this point

> The binding has to be sendable without a thinking config. Fable 5.1 emits thinking blocks whether or not the request configures thinking, and ModelRequestParameters.thinking defaults to None, so _translate_thinking returns OMIT for anyone who never touches thinking settings — the common case, and precisely the one that 400s. Anthropic accepts a thinking object carrying block_binding alone: verified live that with no type, 'error' still 400s and 'drop_block' still reports thinking_dropped. The SDK's discriminated union always requires type, and picking one would change how the model thinks, so that shape rides in extra_body — the one non-obvious routing decision here, flagged for review.

since the dropping won't apply to older user accounts, how can such a user use the model without the thinking being dropped?

dsfaccini:

Short answer to the second question first, because it turned out to be the sharper one: **on an older account the drop does apply, and that is a real consequence of this PR.** Anthropic gates the *default*, not the parameter. From the [Fable 5.1 release notes](https://platform.claude.com/docs/en/models/fable-5-1/whats-new-fable-5-1#editing-earlier-turns-invalidates-thinking-blocks):

> The check is enforced for new accounts created on or after August 31, 2026. For accounts created earlier, the API records the mismatch but acts on it only when the request sets `thinking.block_binding.prefix_mismatch_behavior`.

Setting the field is exactly what this PR does. So an older account that would have replayed a stale block intact now has it dropped.

**To keep the block, ask for nothing** — `block_binding: None` sends the field as `null`, which falls back to the account default:

```python
settings: AnthropicModelSettings = {
    'anthropic_thinking': {'type': 'adaptive', 'block_binding': None}
}
```

Live matrix on `claude-fable-5-1`, one seeded thinking block replayed after the `system` text changed, on our (pre-cutoff) account:

| `thinking.block_binding` | result |
|---|---|
| field absent | 200, block kept |
| `null` | 200, block kept |
| `{}` | 200, block kept |
| `{'prefix_mismatch_behavior': 'error'}` | **400** `The block is bound to a different conversation` |
| `{'prefix_mismatch_behavior': 'drop_block'}` | 200, `input_transformations: [thinking_dropped]` |

The limit is worth being blunt about: **there is no "keep it" behavior.** `BetaThinkingPrefixMismatchBehavior` is `Literal['error', 'drop_block']`. `null` / `{}` / absent all mean "use the account default", which on an account created on or after 31 August 2026 *is* `error`. So keeping a stale block is only available to older accounts, and only for as long as Anthropic leaves them ungated. Now documented in `docs/models/anthropic.md`.

---

### On the `extra_body` point

Three facts stacked, and the last one is the answer:

1. `block_binding` lives inside the `thinking` object. There is nowhere else to put it.
2. Pydantic AI sends no `thinking` at all unless you configure one — `ModelRequestParameters.thinking` defaults to `None`, so `_translate_thinking` returns `OMIT`. Fable 5.1 emits thinking blocks anyway, so the user who never touched a thinking setting is both the common case *and* the one that hits the 400. That user has no `thinking` object for the binding to ride in.
3. The API accepts a `thinking` object holding `block_binding` and nothing else — no `type` (verified: `'error'` still 400s, `'drop_block'` still reports `thinking_dropped`). `BetaThinkingConfigParam` is a union of `enabled` / `adaptive` / `disabled`, each with `type` required, so that shape has no typed representation. Hence `extra_body`, the same route `temperature` / `top_p` / `top_k` already take in this file.

The wire for a default `claude-fable-5-1` request is just:

```json
{"thinking": {"block_binding": {"prefix_mismatch_behavior": "drop_block"}}}
```

**One correction to what I wrote before.** I justified this with "picking a `type` would change how the model thinks". That is too strong: Anthropic documents adaptive thinking as *always on* for Fable 5.1, so `{'type': 'adaptive'}` would be a no-op there. The durable reason is different — `anthropic_binds_thinking_blocks` is a profile flag, so the next model to carry it may not have always-on adaptive thinking, and a `type` we injected would then silently be a thinking config the user never asked for. The type-less shape can never do that. PR body updated.

### And a defect the same doc page exposed

> Claude Mythos 5.1 doesn't run this check.

The flag was `True` for `claude-mythos-5-1` too, on the "same capabilities" mirror. It isn't a capability Mythos 5.1 has — so we were sending it a `block_binding` field and a beta header that can never fire, on the one model here we can't probe. Scoped to `claude-fable-5-1` in `04b3e2a31`.

docs/models/anthropic.md
```diff
@@ -558,6 +558,60 @@ On a model that doesn't support forcing:
...
+
+- a [dynamic instructions](../agent.md#instructions) function whose text differs between runs, and
+- a [filtered toolset](../toolsets.md#filtering-tools) that advertises a new tool mid-conversation, unless the tool uses [deferred loading](../toolsets.md#deferred-loading).
+
+Anthropic enforces the check on accounts created on or after 31 August 2026. On an older account the mismatch is recorded, but acted on only if the request sets a behavior.
+
+On that model Pydantic AI defaults Anthropic's `thinking.block_binding.prefix_mismatch_behavior` to `'drop_block'`, marked by [`anthropic_binds_thinking_blocks=True`][pydantic_ai.profiles.anthropic.AnthropicModelProfile.anthropic_binds_thinking_blocks]. The stale block is dropped and the request proceeds instead of failing. **The model no longer sees that turn's reasoning** — the trade is one turn's thinking against a failed run. Anthropic reports every drop, and Pydantic AI records it on the response:
```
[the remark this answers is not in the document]
dsfaccini: > This reply was posted by Claude Code using claude-opus-5 on behalf of David.

Confirmed live, and it was worse than the docs: streamed runs recorded **no** drop at all.

Probed a real stream on `claude-fable-5-1` replaying a stale block under `drop_block`:

```
message_start   input_transformations at message_start.message: [{"type":"thinking_dropped",...}]
message_delta   no input_transformations
```

So `message_start` is where a live stream reports it, and the `message_delta` branch — the only one we had — never fires. My mock test asserted a shape the API doesn't send.

Fixed in `542c9fc05`: `message_start` records it, `message_delta` appends rather than overwrites (the SDK types the field on both), and there's now a **recorded streaming cassette** — `test_anthropic_fable_5_1_drops_a_stale_thinking_block_streamed` — because which event carries the report is precisely what a mock can't establish.

The docs example itself is unchanged: it uses `run_sync`, which was already correct, and now the streamed path matches it.

dsfaccini:

@DouweM reversing the `drop_block` default in `707989f55`, and the reason is one sentence in Anthropic's release notes that I under-weighted when I implemented your call:

> For accounts created earlier, the API records the mismatch but acts on it only when the request sets `thinking.block_binding.prefix_mismatch_behavior`.

**Setting the field is what makes the API act.** So defaulting to `drop_block` doesn't only rescue accounts that would 400 — it *creates* a loss of reasoning on every account created before 2026-08-31, which the check otherwise never touches. We'd be trading a failure mode newer accounts have for a silent quality regression everyone else gets.

**New behavior:** send nothing. If Anthropic rejects the replay, retry that request once with `drop_block` and warn. Nobody passes a setting, older accounts keep replaying their reasoning, newer accounts don't fail. It costs one extra round trip in the mismatch case and nothing anywhere else.

The warning is actionable, which was the sticking point on #7984 — it now has something to tell you:

```
'claude-fable-5-1' rejected a replayed thinking block because the conversation prefix changed,
which Anthropic enforces for accounts created on or after 2026-08-31. Pydantic AI retried the
request with `prefix_mismatch_behavior='drop_block'`, so the run continued without that turn's
reasoning. To skip the rejected request, ask for the drop yourself: `model_settings={...}`.
To keep the retry and silence this warning: `warnings.simplefilter('ignore',
AnthropicStaleThinkingBlockWarning)`. See <docs link>
```

### The account-age carve-out is now pinned by a cassette

We're resting a default on it, so it's a recorded test rather than a docs quote — `test_anthropic_fable_5_1_replays_a_stale_thinking_block_on_a_legacy_account`. In the recorded wire bytes: the `system` prompt changes between turns, the thinking block is replayed, **no `thinking` param is sent**, and the response reports **no** `input_transformations`. The model still saw the reasoning.

The two `drop_block` cassettes now record the explicit opt-in, which is also the exact shape the retry sends.

<details>
<summary>Scope of the retry</summary>

Only fires when all four hold: the model carries `anthropic_binds_thinking_blocks`, the request set no `block_binding` of its own, the status is 400, and Anthropic's error body contains `The block is bound to a different conversation`. An explicit `'error'` is a caller asking to fail; an explicit `'drop_block'` can't produce the rejection. Every other 400 propagates untouched. One retry, never a loop.

</details>

pydantic_ai_slim/pydantic_ai/models/anthropic.py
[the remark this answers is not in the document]
dsfaccini: > This reply was posted by Claude Code using claude-opus-5 on behalf of David.

Real, and fixed in 3be873044.

Reproduced through a local HTTP server against the pre-fix code: with `extra_body={'thinking': {...}}` the retried request went out byte-identical to the rejected one, so it failed the same way while `AnthropicStaleThinkingBlockWarning` claimed the run had recovered. Post-fix the retry carries both the caller's keys and the added `block_binding`.

I didn't take the suggested diff verbatim, because it fixes one half and breaks the other: `extra_body` is also how a caller reaches `block_binding` itself, which the installed SDK can't type. Unconditionally overriding `extra_body['thinking']` would override that choice. So the fix has two halves:

1. `_build_extra_body` merges as `{**thinking_override, **caller_thinking}` — the caller's keys still win, only the `block_binding` they provably did not set comes from the retry.
2. `_is_stale_thinking_block_error` now refuses to retry when `block_binding` is set through `extra_body`, matching the rule the typed `thinking` config already had.

Two regression tests, one per half.

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -984,24 +1098,41 @@ async def create(
...
-                extra_body=_build_extra_body(model_settings),
+                extra_body=_build_extra_body(model_settings, thinking_override),
             )
 
         with _map_api_errors(self.model_name, self._provider.model_id_namespace):
             try:
-                return await create(container)
-            except APIStatusError as e:
+                return await create(container, thinking, betas, None)
+            except APIStatusError as error:
```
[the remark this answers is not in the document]
dsfaccini: > This reply was posted by Claude Code using claude-opus-5 on behalf of David.

Confirmed and fixed in 114556c55.

Reproduced against the pre-fix code with `[500, stale-thinking-400, success]` on `claude-fable-5-1` with a history container — 2 requests, the 400 propagating, no retry and no warning:

```
[0] container='container_from_history'  -> 500
[1] container=OMIT                      -> 400, propagates
```

One correction to the suggested diff: catching the fallback error and re-running the check isn't sufficient on its own. The drop retry reads `container`, the id the *first* attempt used — the one the 500 just disowned — so a third request would have resent it and 500'd again. The fix therefore also carries the drop forward, and the test pins the container of all three requests as `['container_from_history', OMIT, OMIT]`.

Covered by `test_anthropic_code_execution_files_500_then_stale_thinking_block_still_retries`.

dsfaccini:

Two Macroscope **High** findings landed after the approvals above, both real and both reproduced. Fixed in `3be873044` and `114556c55`; the description now covers them. Flagging here because the threads are resolved and therefore collapsed, and the approvals on this PR predate both commits.

Both are the same defect in different clothes: **the stale-thinking retry re-sent a request that could only fail the same way.**

1. **A caller's own `extra_body['thinking']` swallowed the retry.** `extra_body` wins over anything the model builds, so the retried request went out byte-identical to the one Anthropic had just rejected — while `AnthropicStaleThinkingBlockWarning` claimed the run had recovered. The retry's `block_binding` is now merged *underneath* the caller's keys.

2. **The container-500 fallback's own error skipped the check.** When a `container_upload` on a history-resolved container answers 500, we re-send once without the container; that call sits inside the handler that caught the 500, so a stale-thinking 400 on it propagated instead of retrying.

On (2) I did not apply the suggested diff as written, because it fixes one half and leaves a second failure: the drop retry read the outer `container` — the id the 500 had just disowned — so a third request would have resent it and 500'd again. The fix carries the drop forward, and the test pins all three requests' containers as `['container_from_history', OMIT, OMIT]`.

<details>
<summary>Pre-fix repro for (2)</summary>

`MockAnthropic([500, stale-thinking-400, success])` on `claude-fable-5-1` with a history container:

```
[0] container='container_from_history'  -> 500
[1] container=OMIT                      -> 400, propagates
```

Two requests, no warning, no retry. Post-fix it is three, and the run recovers.

</details>

Still wanted here: **a human review round.** Every approval on this PR is a bot, and it now changes request behavior three times over — the forced-`tool_choice` narrowing, the thinking-block retry, and that retry's container-fallback path.

docs/models/anthropic.md
```diff
@@ -560,6 +560,62 @@ On a model that doesn't support forcing:
...
+
+- a [dynamic instructions](../agent.md#instructions) function whose text differs between runs, and
+- a [filtered toolset](../toolsets.md#filtering-tools) that advertises a new tool mid-conversation, unless the tool uses [deferred loading](../toolsets.md#deferred-loading).
+
+Anthropic enforces the check for accounts created on or after 31 August 2026. For an older account it records the mismatch but acts on it only if the request sets `thinking.block_binding.prefix_mismatch_behavior`.
+
+**Pydantic AI sets nothing by default**, so an older account keeps replaying its reasoning untouched. Where the check is enforced, the rejected request is retried once with `prefix_mismatch_behavior='drop_block'`: the stale block is dropped, the run continues, and a [`AnthropicStaleThinkingBlockWarning`][pydantic_ai.models.anthropic.AnthropicStaleThinkingBlockWarning] explains what happened. **The model no longer sees that turn's reasoning** — the trade is one turn's thinking against a failed run. Models marked [`anthropic_binds_thinking_blocks=True`][pydantic_ai.profiles.anthropic.AnthropicModelProfile.anthropic_binds_thinking_blocks] are the only ones that retry.
```
dsfaccini, line 574: @DouweM

docs/models/anthropic.md
```diff
@@ -560,6 +560,62 @@ On a model that doesn't support forcing:
...
+
+Anthropic reports every drop, and Pydantic AI surfaces it two ways. Under [instrumentation](../logfire.md) the model request span carries an `anthropic.input_transformations` event, so a drop is visible in the trace as it happens. On the response it is recorded in `provider_details`:
+
+```python {title="dropped_thinking_blocks.py" test="skip"}
```
dsfaccini: why are we skipping this one?

dsfaccini: > This reply was posted by Claude Code using claude-opus-5 on behalf of David.

No good reason — removed in 553c94f5a.

I checked rather than reasoned about it: dropping the `test="skip"` and running `tests/test_examples.py -k dropped_thinking_blocks` passes. It's now tested like the two examples below it.

It was the only skip this PR added; the other twelve in the file are pre-existing.

tests/models/test_tool_choice_unit.py
```diff
@@ -946,7 +946,10 @@ async def test_anthropic_single_tool_forcing_under_unified_thinking(
 
 
 # Models that reject a forced `tool_choice` outright, even without thinking (unlike other Anthropic models).
-NO_FORCING_ANTHROPIC_MODELS = ['claude-fable-5', 'claude-mythos-5', 'claude-mythos-preview']
+NO_FORCING_ANTHROPIC_MODELS = [
```
dsfaccini, line 949: so the 5.0 family now allows forcing?

dsfaccini: > This reply was posted by Claude Code using claude-opus-5 on behalf of David.

`claude-fable-5` — yes, verified live today: 200 on both `{'type': 'any'}` and `{'type': 'tool'}`, 3/3 each, on the GA and beta endpoints.

`claude-mythos-5` and `claude-mythos-preview` — **not verified.** Both 404 on our credentials (Project Glasswing), so they come off the list on Anthropic's [forcing-tool-use table](https://platform.claude.com/docs/en/agents-and-tools/tool-use/implement-tool-use) alone, which now names Fable 5.1 and Mythos 5.1 and nothing else.

What I can't tell you is whether Anthropic *changed* the API or whether the entry was always wider than the behavior. I can only report today's probe and today's table.

Unchanged either way: forcing alongside **extended** thinking is still rejected across Anthropic models, and adaptive thinking is still compatible with it. This list is only about models that reject forcing with no thinking configured at all.

docs/models/anthropic.md
```diff
@@ -560,6 +560,64 @@ On a model that doesn't support forcing:
...
+import warnings
+
+from pydantic_ai.models.anthropic import AnthropicStaleThinkingBlockWarning
+
+warnings.simplefilter('ignore', AnthropicStaleThinkingBlockWarning)
+```
+
+There is no third behavior: `prefix_mismatch_behavior` is either `'error'` or `'drop_block'`. Passing `None` is how you ask for Anthropic's account default explicitly, which keeps the block only where the check isn't enforced.
+
+Models that don't bind thinking blocks are unaffected: their requests carry no `block_binding` and no binding beta, and a 400 from them is never retried. Anthropic documents **Claude Mythos 5.1** as not running this check, so it is not marked either.
```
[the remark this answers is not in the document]
dsfaccini: > This reply was posted by Claude Code using claude-opus-5 on behalf of David.

Correct, fixed in 9ae38cabc.

Verified against the code rather than the docs: the beta is attached on `'block_binding' in thinking` with no profile gate at all, and the comment directly above that condition already says "a user who sets `block_binding` through `anthropic_thinking` gets it too". The sentence contradicted our own source.

I extended the wording slightly past the suggestion: an explicitly configured request is also never retried, whatever the model, and that belonged in the same sentence as the retry claim it qualifies.

tests/models/test_anthropic.py
```diff
@@ -4566,6 +4572,535 @@ async def test_anthropic_unified_thinking_false_omits_param(allow_model_requests
...
+async def test_anthropic_fable_5_1_drops_a_stale_thinking_block(allow_model_requests: None, anthropic_api_key: str):
+    """Asking for `drop_block` drops the stale block and lets the run continue.
+
+    This is the shape Pydantic AI retries with after Anthropic rejects a replay, and the shape a
+    caller sets to skip that rejected request altogether.
+    """
+    m = AnthropicModel('claude-fable-5-1', provider=AnthropicProvider(api_key=anthropic_api_key))
+    history = await stale_thinking_block_history(m)
+
+    settings = AnthropicModelSettings(
```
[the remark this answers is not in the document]
dsfaccini: > This reply was posted by Claude Code using claude-opus-5 on behalf of David.

Premise checked and correct — fixed in 9ae38cabc.

There is no `match_on` anywhere in `tests/`, so vcrpy's defaults apply and the body is not matched. The cassette proved causation when it was recorded; on replay it could not, so the recorded `thinking_dropped` would have replayed even with the setting no longer reaching the wire.

The repo already has the tool for this — `RequestCapture` in `tests/conftest.py`, whose docstring names this exact hazard. The test now also asserts the outbound `thinking` object and the `anthropic-beta` header, the latter because the cassette serializer strips `anthropic-*` headers, so the wire is the only place beta gating is visible.

I mutation-tested the fix rather than trusting a green run: dropping `block_binding` from the setting now fails at the new body assertion. Before the change, that mutation passed.

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -1048,6 +1194,14 @@ def _get_betas_and_extra_headers(
         if self._get_task_budget(model_settings) is not None:
             betas.add(_ANTHROPIC_TASK_BUDGETS_BETA)
 
+        # `thinking.block_binding` is a 400 (`Extra inputs are not permitted`) without its beta,
+        # so the beta follows the field rather than the profile flag — a user who sets
+        # `block_binding` through `anthropic_thinking` gets it too.
+        # Membership, not truthiness: `block_binding: {}` is a valid request meaning "every binding
+        # default", and it needs the beta exactly as much as a populated one does.
+        if not isinstance(thinking, Omit) and 'block_binding' in thinking:
```
dsfaccini, line 1234: > This review was posted by Codex Desktop using gpt-5.6-sol on behalf of David.

**Blocking:** `_get_betas_and_extra_headers` only examines the typed `thinking` argument. A caller-provided `extra_body['thinking']['block_binding']` reaches the wire without `thinking-binding-controls-2026-08-01`, which Anthropic rejects with a 400. There is a second precedence failure: typed `anthropic_thinking` containing `block_binding`, overridden by raw `extra_body['thinking']`, suppresses the retry even though the final wire object no longer contains the field. Resolve the effective final thinking object once and use it consistently for beta selection and retry classification; cover both combinations.

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -984,24 +1097,56 @@ async def create(
...
+                    try:
+                        return await create(None, thinking, betas, None)
+                    except APIStatusError as fallback_error:
+                        # Classify what the fallback hit, not the 500 that caused it, so a stale
+                        # thinking block rejected only on this attempt still reaches the retry
+                        # below. The history-resolved container is gone, so it stays dropped there.
+                        error = fallback_error
+                        retry_container = None
+                if not _is_stale_thinking_block_error(anthropic_profile, model_settings, thinking, error):
                     raise
```
dsfaccini: > This review was posted by Codex Desktop using gpt-5.6-sol on behalf of David.

**Blocking:** assigning `error = fallback_error` does not change Python's active exception. This bare `raise` still re-raises the outer container 500, masking an unrelated error from the no-container fallback. Reproduced with 500 → 429: the caller receives `ModelHTTPError(500)`. Preserve and explicitly raise the fallback exception when it is not eligible for stale-thinking recovery, and add a 500 → 429 regression test.

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -984,24 +1097,56 @@ async def create(
...
-                return await create(None)
+
+            warnings.warn(
```
dsfaccini, line 1162: > This review was posted by Codex Desktop using gpt-5.6-sol on behalf of David.

**Blocking:** this warning says the run continued before the retry request has succeeded. A reproduced stale-400 → retry-503 sequence emits `AnthropicStaleThinkingBlockWarning` with that success claim and then raises `ModelHTTPError(503)`. Await a successful retry before emitting the warning, or change the warning to describe an attempted recovery; add the failing-retry regression case.

pydantic_ai_slim/pydantic_ai/profiles/anthropic.py
```diff
@@ -114,12 +114,22 @@ class AnthropicModelProfile(ModelProfile, total=False):
...
     (or an explicit list of tools) raises a `UserError`.
     """
 
+    anthropic_binds_thinking_blocks: bool
+    """Whether the model binds each thinking block to the conversation prefix that produced it. Default: `False`.
+
+    Claude Fable 5.1 rejects a replayed thinking block once the `system` prompt text changes or a
+    non-deferred tool joins the `tools` array — both of which Pydantic AI causes by design, through
+    dynamic `@agent.instructions` and conditional toolsets. When True, requests default
+    `thinking.block_binding.prefix_mismatch_behavior` to `'drop_block'`, so a stale block is dropped
```
dsfaccini: > This review was posted by Codex Desktop using gpt-5.6-sol on behalf of David.

This describes the superseded design. The implementation now sends no binding setting initially and retries with `drop_block` only after Anthropic rejects the replay. Please update the profile field's public docstring to describe the account-preserving retry behavior.

dsfaccini (commented):
> This review was posted by Codex Desktop using gpt-5.6-sol on behalf of David.

Do not merge yet. I reproduced three runtime defects in the stale-thinking recovery path; the inline comments include the fixes and regression cases. The provider-spec assumptions, public API additions, and cassettes otherwise look sound. Current-head CI is green, but it does not exercise these failure combinations.

pydantic_ai_slim/pydantic_ai/models/anthropic.py
dsfaccini: > This reply was posted by Codex Desktop using gpt-5.6-sol on behalf of David.

Fixed in `2fb73f149`. The effective wire `thinking` object now resolves after `extra_body` precedence and drives both beta selection and retry classification. Regression coverage pins raw `block_binding` and typed-thinking-overridden-by-raw-thinking.

pydantic_ai_slim/pydantic_ai/models/anthropic.py
dsfaccini: > This reply was posted by Codex Desktop using gpt-5.6-sol on behalf of David.

Fixed in `2fb73f149`. The no-container fallback error is now raised explicitly; a 500 → 429 regression test confirms the caller receives the 429.

pydantic_ai_slim/pydantic_ai/models/anthropic.py
dsfaccini: > This reply was posted by Codex Desktop using gpt-5.6-sol on behalf of David.

Fixed in `2fb73f149`. The retry is awaited before the recovery warning is emitted; a stale-400 → retry-503 regression confirms the 503 surfaces with no success warning.

pydantic_ai_slim/pydantic_ai/profiles/anthropic.py
dsfaccini: > This reply was posted by Codex Desktop using gpt-5.6-sol on behalf of David.

Updated in `2fb73f149`. The public docstring now describes preserving the account default initially, retrying with `drop_block` only after rejection, and warning only after success.

pydantic_ai_slim/pydantic_ai/models/anthropic.py
```diff
@@ -555,15 +588,110 @@ def _build_extra_body(model_settings: AnthropicModelSettings) -> object | None:
...
+    """Report Anthropic's input transformations on the current span, and return them for `provider_details`.
+
+    A dropped thinking block is otherwise silent: the model answered without the reasoning we
+    replayed, and nothing in the response says so. `provider_details` keeps it inspectable after the
+    run; the span event surfaces it in the trace, on the model request that lost the reasoning.
+    """
+    reported = [transformation.model_dump() for transformation in transformations]
+    span = get_current_span()
+    if span.is_recording():
+        span.add_event(
```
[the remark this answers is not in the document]
dsfaccini: > This reply was posted by Claude Code using claude-opus-5 on behalf of David.

The OTel event is a requirement of this PR, not an optional extra, so it stays — including the `anthropic.` namespace.

`provider_details` alone is not equivalent: it never reaches the span. `InstrumentationSettings.messages_to_otel_messages` emits `role` + `parts` only, and `response_attributes` lifts just `provider_response_id` and `finish_reason` — so a dropped thinking block would be invisible in traces.

`AnthropicStaleThinkingBlockWarning` is not a substitute either. Nothing bridges `warnings` to telemetry, and it covers only the automatic-retry path — not a caller's explicit `block_binding={'prefix_mismatch_behavior': 'drop_block'}`, where the drop is reported with no warning at all.

On the namespace: the transformation is Anthropic's own and forced by the provider — `input_transformations` is their response field, reported verbatim. A `pydantic_ai.*` name would promise a cross-provider shape we have not designed.

@DouweM leaving this open in case you want to weigh in on the adapter-vs-instrumentation-layer convention itself; the event's presence is settled.

tests/models/test_anthropic.py
```diff
@@ -4566,6 +4574,629 @@ async def test_anthropic_unified_thinking_false_omits_param(allow_model_requests
     assert kwargs.get('thinking') is OMIT
 
 
+_THINKING_BINDING_BETA = 'thinking-binding-controls-2026-08-01'
```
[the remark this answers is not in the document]
dsfaccini: > This reply was posted by Claude Code using claude-opus-5 on behalf of David.

Done in `36a9c429d`. The suite, its helpers (`stale_thinking_block_error`, `dropped_thinking_stream`, `sent_betas`, `MockAnthropic` fixtures) and its three cassettes now live in `tests/models/anthropic/test_thinking_block_binding.py`, alongside `test_code_execution_files.py`.

`test_anthropic.py` keeps no binding references, and `mock_anthropic_client` was made public so the sibling files can share it. Same 428 tests pass across `tests/models/anthropic/` and `test_anthropic.py` as before the move.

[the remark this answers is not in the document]
dsfaccini: > This reply was posted by Codex Desktop using gpt-5.6-sol on behalf of David.

Fixed in `36a9c429d`. The 31-test thinking-block-binding suite now lives in `tests/models/anthropic/test_thinking_block_binding.py`, with its three cassettes moved alongside it; affected shared-helper imports were updated. The extracted suite plus neighboring Anthropic modules pass 54 tests.

dsfaccini:

Addressed in `07c2a24ed`. Blocking finding and 5 of 6 required taken; 1 partial, 1 skip, and I promoted one informational to a fix.

**Blocking — runner lock.** Confirmed and regenerated: `requires-dist` now reads `genai-prices>=0.1.6` and resolves 0.1.6.

**Required 2 is the one that mattered.** Verified against the installed SDK 1.3.0 docstring — the `message_delta` carries `input_transformations` only after a mid-stream fallback, and then it "holds the serving model's entries and replaces the one in `message_start`". Now replaces. The comment claiming the delta never carries the field is corrected too. Required 3 and 6 added, and I mutation-tested the new pair: reverting to append fails both, with the duplicate `messages.1.content.0` a real fallback would produce. Required 1 and 4 taken as suggested.

**Required 5 — partial, deliberate.** Not reverting to `simplefilter('error')`: that form overrode the suite's intentional `ResourceWarning` ignores, so delayed event-loop GC failed whichever test collected it — the flake `6ba011972` was written to fix. Your own probe found nothing masked today, so the revert trades a live flake for a theoretical tripwire. The legitimate half was the missing rationale, so both call sites now state it.

**Informational 1 — promoted to a fix, not deferred.** The claim-check rescoped it as pre-existing, but this PR is what falsified it: flipping `claude-fable-5` to `supports_forced_tool_choice=True` is what made the skill's guidance wrong. Fixed with your suggested wording. Informational 2 skipped as you recommended.

One correction: `MockAnthropic` did **not** raise queued exceptions at `create()` time for `stream=True` — that branch never called `raise_if_exception`, so the error surfaced during iteration, which production's `except APIStatusError` around `create()` never sees. Required 6's test would have proven nothing as written, so the mock was made faithful to the SDK first.

The PR body now carries a collapsed decision log covering these and the earlier conscious requirements.

